PR #3605, opened 2020-11-13
Change per column rocksdb block cache size from 180mb to 32mb

pmnoxx:
Rocksdb block size cache gets allocated per column. Currently it's 47*180 mb = 8.5gb.
By decreasing rocksdb block cache size to 32mb, we will decrease total cache size to 1.5gb.

In the future total cache size can be decreased by fine tuning cache sizes per column.

### DIFF
--- a/core/store/src/db.rs
+++ b/core/store/src/db.rs
@@ -485,8 +485,8 @@ fn rocksdb_options() -> Options {
 fn rocksdb_block_based_options() -> BlockBasedOptions {
     let mut block_opts = BlockBasedOptions::default();
     block_opts.set_block_size(1024 * 16);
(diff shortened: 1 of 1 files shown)


frol:
Should this be closed in favor of #3606?

pmnoxx:
> Should this be closed in favor of #3606?

The master branch is currently broken. I want to run nightly tests on both, and see how limiting buffer size affects the performance. We should make a decision once we have all the data.

bowenwang1996 (commented):
this seems to be in direct conflict with #3606. Which one do we want?

bowenwang1996 (approved):

PR #3604, opened 2020-11-13
fix: rectify inflation

bowenwang1996:
Implement the proposal to recalibrate inflation based on time instead of number of blocks. The proposal can be found in https://github.com/nearprotocol/NEPs/pull/132

Test Plan
---------
* Modify existing economics tests to account for the change under the protocol feature flag.

PR #3606, opened 2020-11-13
Limit rocksdb block cache size 1mb

pmnoxx:
We should consider limit Rocksdb per column block cache size to 1mb. This should reduce memory usage even further.


frol (commented):
I am not sure about the disk cache implications, but it fails `test_contract_call_disk` test.

@ailisp You seem to have more experience with these

pmnoxx:
> I am not sure about the disk cache implications, but it fails `test_contract_call_disk` test.
> 
> @ailisp You seem to have more experience with these

Master branch is currently broken. You can see other pull requests failing as well for the same reason: https://github.com/near/nearcore/pull/3604

chefsale:
Master wasn't broken, you just neede to rebase I updated the branch for you. We were fixing some issues with CI testing so, thats why. Test seem to be passing now.

bowenwang1996 (approved):

PR #3607, opened 2020-11-13
Pass cache to viewing contracts as well.

olonho:
* Refactored view calls, grouping most arguments of view call into `ViewApplyState` (similar to non-view calls)
* Added `cache` field to `ViewApplyState` and wrap up store into the cache in view calls from runtimes (regular and standalone)


runtime/runtime/src/state_viewer.rs
```diff
@@ -112,6 +112,11 @@ impl TrieViewer {
         // TODO(#1015): Add ability to pass public key and originator_id
         let originator_id = contract_id;
         let public_key = PublicKey::empty(KeyType::ED25519);
+        let cache = epoch_info_provider.contract_cache();
+        let cache = match &cache {
+            Some(cache) => Some((*cache).as_ref()),
+            None => None,
+        };
```
frol, line 119: This should be enough :thinking: 

```suggestion
        let cache = epoch_info_provider.contract_cache().as_ref();
```

neard/src/runtime.rs
```diff
@@ -110,6 +111,13 @@ impl EpochInfoProvider for SafeEpochManager {
         let mut epoch_manager = self.0.write().expect(POISONED_LOCK_ERR);
         epoch_manager.minimum_stake(prev_block_hash)
     }
+
+    fn contract_cache(&self) -> Option<Arc<dyn CompiledContractCache>> {
+        match self.0.write() {
+            Ok(epoch_manager) => Some(epoch_manager.contract_cache()),
+            Err(_) => None,
+        }
```
frol, line 119:
```suggestion
        self.0.write().ok()?.contract_cache()
```

frol (approved):

runtime/runtime/src/state_viewer.rs
olonho: Rust compiler tends to disagree:
```
error[E0308]: mismatched types
   --> runtime/runtime/src/state_viewer.rs:160:17
    |
160 |                 cache,
    |                 ^^^^^ expected trait object `dyn near_primitives::types::CompiledContractCache`, found struct `std::sync::Arc`
    |
    = note: expected enum `std::option::Option<&dyn near_primitives::types::CompiledContractCache>`
               found enum `std::option::Option<&std::sync::Arc<dyn near_primitives::types::CompiledContractCache>>`
```

frol: Oh, we have Arc there, use `as_deref()` there to force Rust to go through the dereference chain instead of just `Option<T> -> Option<&T>`

```suggestion
        let cache = epoch_info_provider.contract_cache().as_deref();
```

evgenykuzyakov (changes requested):
1) Please add description.

2) Why contract cache belongs to epoch manager? It doesn't make any sense. I don't think this hack is a good idea.

3) `near-vm-runner` can't depend on primitives, because we'll not be able to publish it to crates.

chain/epoch_manager/src/lib.rs
```diff
@@ -56,6 +59,8 @@ pub struct EpochManager {
     epoch_info_aggregator: Option<EpochInfoAggregator>,
     /// Largest final height. Monotonically increasing.
     largest_final_height: BlockHeight,
+    /// Cache of compiled to native code contracts.
+    cache: Arc<dyn CompiledContractCache>,
```
bowenwang1996, line 63: it makes no sense to have the contract cache here

bowenwang1996 (changes requested):
It doesn't make sense to put the cache inside epoch manager by any means. It is better to pass store to `TrieViewer` or to `ViewRuntimeAdapter::call_function` from `NightShadeRuntime`

runtime/runtime/src/state_viewer.rs
olonho: Had to create a temp to hold a value, but applied similar solution.

neard/src/runtime.rs
olonho: `Some(self.0.write().ok()?.contract_cache())` works

chain/epoch_manager/src/lib.rs
olonho: > 1. Please add description.
You mean to PR itself?

> 2. Why contract cache belongs to epoch manager? It doesn't make any sense. I don't think this hack is a good idea.
OK, actually I agree that it's not the best place, however feel that view call miss smth like `ApplyState`, and adding additional argument felt somewhat clumsy. WDYT of replacing `epoch_info_provider` with more general purpose argument of that kind?

> 3. `near-vm-runner` can't depend on primitives, because we'll not be able to publish it to crates.
Hmm, it already depends on `near_primitives::hash::CryptoHash` per your suggestion in previous review. How would you suggest to organize dependency graph?

bowenwang1996: EpochInfoProvider is supposed to provide specifically information related to epoch and validators and we definitely should not tag on the contract cache there.

bowenwang1996, line 63: this is still not removed

bowenwang1996 (approved):

olonho:
* added description
 * removed cache from epoch management
 * still not sure how to address Evgeny's comment `near-vm-runner can't depend on primitives, because we'll not be able to publish it to crates.`

core/primitives/src/types.rs
```diff
@@ -558,6 +558,12 @@ pub enum TransactionOrReceiptId {
     Receipt { receipt_id: CryptoHash, receiver_id: AccountId },
 }
 
+/// Cache for compiled modules
+pub trait CompiledContractCache: Send + Sync {
+    fn put(&self, key: &[u8], value: &[u8]) -> Result<(), std::io::Error>;
+    fn get(&self, key: &[u8]) -> Result<Option<Vec<u8>>, std::io::Error>;
```
evgenykuzyakov, line 564: Can `key` be a `CryptoHash`?

core/primitives/src/views.rs
```diff
@@ -58,6 +59,24 @@ pub struct AccountView {
     pub storage_paid_at: BlockHeight,
 }
 
+#[derive(Debug)]
+pub struct ViewApplyState {
```
evgenykuzyakov, line 63: Please add a comment

core/store/src/lib.rs
```diff
@@ -488,3 +488,19 @@ pub fn set_genesis_state_roots(store_update: &mut StoreUpdate, genesis_roots: &V
         .set_ser::<Vec<StateRoot>>(DBCol::ColBlockMisc, GENESIS_STATE_ROOTS_KEY, genesis_roots)
         .expect("Borsh cannot fail");
 }
+
+pub struct StoreCompiledContractCache {
+    pub store: Arc<Store>,
+}
+
```
evgenykuzyakov, line 495: Please add a comment

neard/src/runtime.rs
```diff
@@ -1397,19 +1397,23 @@ impl node_runtime::adapter::ViewRuntimeAdapter for NightshadeRuntime {
         current_protocol_version: ProtocolVersion,
     ) -> Result<Vec<u8>, Box<dyn std::error::Error>> {
         let state_update = self.get_tries().new_trie_update_view(shard_id, state_root);
+        let view_state = ViewApplyState {
+            block_height: height,
+            last_block_hash: *last_block_hash,
+            epoch_id: epoch_id.clone(),
+            epoch_height: epoch_height,
+            block_timestamp: block_timestamp,
+            current_protocol_version: current_protocol_version,
```
evgenykuzyakov, line 1406:
```suggestion
            epoch_height,
            block_timestamp,
            current_protocol_version,
```

evgenykuzyakov (approved):

core/primitives/src/types.rs
olonho: Good question. Guess we wouldn't win much from that, and probably could use more complex key structure. Let's keep it as is for now.

runtime/runtime-standalone/src/lib.rs
```diff
@@ -16,9 +16,11 @@ use near_primitives::types::{
     StateChangeCause,
 };
 use near_primitives::version::PROTOCOL_VERSION;
+use near_primitives::views::ViewApplyState;
```
willemneal, line 19: This crate was moved to the sdk.  I'll update my PR to use this after it merges.

willemneal (approved):

runtime/runtime-standalone/src/lib.rs
ailisp: should we delete runtime-standalone from nearcore to avoid keep maintain two copy of it?

ailisp (approved):
Looks good!

PR #3609, opened 2020-11-13
limit rocksdb memory usage for no_cache feature

pmnoxx:
Limit write buffer size to 64mb down from 256mb.


bowenwang1996 (commented):
should we change it to explicitly depend on the number of columns?

pmnoxx:
> number



> should we change it to explicitly depend on the number of columns?
I think it's better to have a hard coded constant. We are not going to increase the number of columns significantly in the future by a huge margin, so total memory usage should not be affected by much even if we do.

I added an explanation about how total block cache size limit is calculated.

We should test how my changes affect the performance. If needed we can increase cache limit for specific columns.

core/store/src/db.rs
```diff
@@ -461,7 +461,8 @@ fn rocksdb_options() -> Options {
     opts.set_max_open_files(512);
     opts.set_keep_log_file_num(1);
     opts.set_bytes_per_sync(1048576);
-    opts.set_write_buffer_size(1024 * 1024 * 512 / 2);
+    // The are 2 write buffers by default, so the total limit is 128mb * 2 = 256mb.
```
bowenwang1996, line 464: seems like the comment is wrong. The overall size is 128MB

core/store/src/db.rs
```diff
@@ -485,9 +486,10 @@ fn rocksdb_options() -> Options {
 fn rocksdb_block_based_options() -> BlockBasedOptions {
     let mut block_opts = BlockBasedOptions::default();
     block_opts.set_block_size(1024 * 16);
-    // We create block_cache for each of 47 columns, so the total cache size is 1 * 47 = 47mb
-    let cache_size = 1024 * 1024 * 1;
-    block_opts.set_block_cache(&Cache::new_lru_cache(cache_size).unwrap());
+    // Total block cache size is 1mb * 47 (number of columns) = 47mb. It should me a multiple of
+    // 512kb.
+    let block_cache_size_per_column = 1024 * 1024 * 47 / 47;
```
bowenwang1996, line 491: looks like we already merged the change to 32MB. Let's stick to that

bowenwang1996:
@pmnoxx should we close this PR?

core/store/src/db.rs
```diff
@@ -461,8 +461,17 @@ fn rocksdb_options() -> Options {
     opts.set_max_open_files(512);
     opts.set_keep_log_file_num(1);
     opts.set_bytes_per_sync(1048576);
-    opts.set_write_buffer_size(1024 * 1024 * 512 / 2);
-    opts.set_max_bytes_for_level_base(1024 * 1024 * 512 / 2);
+
+    #[cfg(feature = "no_cache")]
```
bowenwang1996, line 465: is there a reason why we have to make this distinction? Couldn't we simply lower the write buffer size?

pmnoxx: You are right, I changed it to always reduce write buffer to 64mb.

Makefile
```diff
@@ -1,37 +0,0 @@
-docker-nearcore:
```
bowenwang1996, line 1: why is this file changed at all?

core/store/src/db.rs
```diff
@@ -485,7 +485,10 @@ fn rocksdb_options() -> Options {
 fn rocksdb_block_based_options() -> BlockBasedOptions {
     let mut block_opts = BlockBasedOptions::default();
     block_opts.set_block_size(1024 * 16);
+    #[cfg(feature = "no_cache")]
+    let cache_size = 1024 * 1024 * 16;
```
bowenwang1996, line 489: should we set it to an even smaller number? like 2MB?

pmnoxx: I tried setting it to 1mb before, but it affected performance too much, the node couldn't run anymore.
I have another PR, which add performance metrics. We should first measure how those settings affect the performance, before we do any changes.

That's why I don't want to push this PR, before I get the data.

bowenwang1996: I see. Makes sense

PR #3610, opened 2020-11-13
fix(Increase historgram levels up to 30)

mhalambek:
Instaead of default use exponential histogram buckets.
Renaming of existing rpc metrics to all start with near_rpc.

### DIFF
--- a/core/metrics/src/lib.rs
+++ b/core/metrics/src/lib.rs
@@ -109,8 +109,16 @@ pub fn try_create_histogram(name: &str, help: &str) -> Result<Histogram> {
 
 /// Attempts to create a `HistogramVector`, returning `Err` if the registry does not accept the counter
 /// (potentially due to naming conflict).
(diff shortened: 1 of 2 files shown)


chefsale (approved):

core/metrics/src/lib.rs
```diff
@@ -109,8 +109,16 @@ pub fn try_create_histogram(name: &str, help: &str) -> Result<Histogram> {
...
+pub fn try_create_histogram_vec(
+    name: &str,
+    help: &str,
+    labels: &[&str],
+    buckets: Option<Vec<f64>>,
+) -> Result<HistogramVec> {
+    let mut opts = HistogramOpts::new(name, help);
+    if buckets.is_some() {
+        opts = opts.buckets(buckets.unwrap());
+    }
```
frol, line 121:
```suggestion
    if let Some(buckets) = buckets {
        opts = opts.buckets(buckets);
    }
```

frol (approved):

Issue #3598, opened 2020-11-12
Investigate runtime estimator panic after PR#3505


ailisp:

ailisp:
The panic is due to randomness selecting account when add some key value to storage, and in next step metric doesn't have required keys in storage because randomly select other accounts which did not have add key value step. Therefore some metric (trigger when delete key value in storage) didn't trigger at all, and cause either stat calc overflow, or some key doesn't present in metric stat at all and panicked (not every time panicked, and may panicked for one of these two places). @evgenykuzyakov observed it didn't happen when there's a warmup iter, but i think that's warmup iter increase total number of txns and has a higher chance that one time account has key value to delete and thus the delete from store metrics is recorded. And obviously this is not correct as most of time delete from store metrics is not triggered. 

So back to PR, #3505 itself has no issue, above described is a long existing issue in runtime params estimator and described in issue #3601

PR #3616, opened 2020-11-14
fix(tool): fix runtime-params-estimator #3601

### LINKED ISSUE #3601: bug: Param estimator reuses accounts that rely on the storage
Param estimator reuses accounts for measuring different host function gas numbers.
Some of these arguments depend on the existing storage. If we use an account, then the 2nd estimate run on this account will hit a modified storage and may receive a different result.

For example `storage_remove_10b_key_10kib_value_1k` has only 1 account with keys (for some reason). It requires the existing values in the trie to be able to count the number of bytes removed. It causes the cost to either overestimate or underestimate.

ailisp:
Fix #3601. get new numbers which is very different from existing genesis.json, but reasonable compare to run runtime-param-estimator from master branch

Test Plan
------------
runtime-param-estimator works, no #3598 and #3601. And every delete storage actually delete, not noop.


runtime/runtime-params-estimator/src/cases.rs
```diff
@@ -502,6 +540,45 @@ pub fn run(mut config: Config, only_compile: bool) -> RuntimeConfig {
 
     // Measure the speed of all extern function calls.
     for (metric, method_name) in v {
+        // Before read and remove key value, insert key value to all accounts
+        if metric == Metric::storage_read_10b_key_10b_value_1k {
+            testbed = function_call_on_all_accounts(
+                metric,
```
evgenykuzyakov, line 546: The metric should be `warmup`

ailisp: No the sequence of test is:
```
    storage_write_10b_key_10b_value_1k => storage_write_10b_key_10b_value_1k,
    storage_has_key_10b_key_10b_value_1k => storage_has_key_10b_key_10b_value_1k,
    storage_read_10b_key_10b_value_1k => storage_read_10b_key_10b_value_1k,
...
    storage_write_10kib_key_10b_value_1k => storage_write_10kib_key_10b_value_1k,
    storage_has_key_10kib_key_10b_value_1k => storage_has_key_10kib_key_10b_value_1k,
    storage_read_10kib_key_10b_value_1k => storage_read_10kib_key_10b_value_1k,
...
```
before testing read_key_10b_key_10b_value_1k and we need to have all account has key values. Reason of not doing it in warmup is - we also have storage_read_key_10kib_key_10b_value_1k test. And in storage_has_key_10b_key_10b_value_1k we want to measure a state with some key and without some key

ailisp:
@evgenykuzyakov @nearmax  I realized that some tests like storage_remove_10b_key_10b_value_1k and delete_account, warm up txns is questionable because warm up txns could delete too many of them, and cause storage_remove_10b_key_10b_value_1k become many no-ops that underestimate delete cost. Should these metrics always skip warm up iter?

evgenykuzyakov:
> I realized that some tests like storage_remove_10b_key_10b_value_1k and delete_account, warm up txns is questionable because warm up txns could delete too many of them, and cause storage_remove_10b_key_10b_value_1k become many no-ops that underestimate delete cost. Should these metrics always skip warm up iter?

We need warmup iter to compile a contract to cache. Otherwise we don't need warmup. It should be fine to call `noop` as a warmup transaction

ailisp:
> > I realized that some tests like storage_remove_10b_key_10b_value_1k and delete_account, warm up txns is questionable because warm up txns could delete too many of them, and cause storage_remove_10b_key_10b_value_1k become many no-ops that underestimate delete cost. Should these metrics always skip warm up iter?
> 
> We need warmup iter to compile a contract to cache. Otherwise we don't need warmup. It should be fine to call `noop` as a warmup transaction

I see. But I mean a different warmup: the one inside measure_transactions function by calling same `f()` to generate a txn to "warmup". For storage_remove_10b_key_10b_value_1k, the warm up is also storage_remove_10b_key_10b_value_1k:
```
    if config.warmup_iters_per_block > 0 {
        let bar = ProgressBar::new(warmup_total_transactions(config) as _);
        bar.set_style(ProgressStyle::default_bar().template(
            "[elapsed {elapsed_precise} remaining {eta_precise}] Warm up {bar} {pos:>7}/{len:7} {msg}",
        ));
        for block_size in config.block_sizes.clone() {
            for _ in 0..config.warmup_iters_per_block {
                let block: Vec<_> = (0..block_size).map(|_| (*f)()).collect();
                testbed.process_block(&block, allow_failures);
                bar.inc(block_size as _);
                bar.set_message(format!("Block size: {}", block_size).as_str());
            }
        }
        testbed.process_blocks_until_no_receipts(allow_failures);
        bar.finish();
    }
```
My concern is that deleting too many key-value pairs in storage_remove_10b_key_10b_value_1k will cause not many accounts have keys to remove, cause this cost underestimate. Also, key-value removed in warm up is random, two runs of param-estimator can have two different case (assume iter_per_block=10, warmup_iters_per_block=10, block_size=100):
- case 1: warm up remove 1000 keys of 1000 accounts. actual measure_txn randomly choose 1000 accounts that still have keys.
- case 2: warm up remove 1000 keys of 1000 accounts. actual measure_txn randomly choose 1000 accounts, 500 of them still have keys, 500 of them has deleted keys. 
This two cases can happen and cause a incorrect volatility in measure. Although if number of accounts >> iter_per_block*block_size this volatility is trivial.

evgenykuzyakov:
We don't use warmup for the `icount` because it doesn't need warmup due to precise instruction counting anyway.

ailisp:
> We don't use warmup for the `icount` because it doesn't need warmup due to precise instruction counting anyway.

Got it, so warmup_iter is 0, and it's not a problem

ailisp:
Fixed some issues and rerun yesterday evening, but still not finished after more than 12 hours, maybe set this size of key value to all active accounts are too expensive :thinking:

ailisp:
It finally finishes after more than 24 hours run. storage read_value_byte, remove_ret_value_byte was heavily underestimated, due to many of those keys do not exist (because of only randomly selected about 100 accounts to storage write a key value):
Old, already `* multiplier = 3`:
        "storage_write_base": 64196736000,                                                                     
        "storage_write_key_byte": 70482867,                                                                    
        "storage_write_value_byte": 31018539,                                                                  
        "storage_write_evicted_byte": 32117307,                                                                
        "storage_read_base": 56356845750,                                                                      
        "storage_read_key_byte": 30952533,                                                                     
        **"storage_read_value_byte": 5611005,**                                                                    
        "storage_remove_base": 53473030500,                                                                    
        "storage_remove_key_byte": 38220384,                                                                   
        **"storage_remove_ret_value_byte": 11531556,**
        "storage_has_key_base": 54039896625,
        "storage_has_key_byte": 30790845,

new:
      "storage_write_base": 10476448125,                                                                       
      "storage_write_key_byte": 23015098,                                                                      
      "storage_write_value_byte": 10698078,                                                                    
      "storage_write_evicted_byte": 12642023,                                                                  
      "storage_read_base": 35282594250,                                                                        
      "storage_read_key_byte": 12048013,                                                                       
      **"storage_read_value_byte": 42063716,**                                                                     
      "storage_remove_base": 44997444875,                                                                      
      "storage_remove_key_byte": 24992328,                                                                     
      **"storage_remove_ret_value_byte": 49999206,**                                                               
      "storage_has_key_base": 21365012250,                                                                     
      "storage_has_key_byte": 11866106,

ailisp:
I just examine code again, it's not all active accounts (20K) used in deploy code, but only 300, and randomly selected 2 account, not 100, from 300 to write key, value, so the bug mentioned in this PR still exist and fix in this PR is still valid: adding write all key values to 300 accounts, so randomly selected two account can guarantee to have account.

But, write all key values to these 300 accounts should not take long, which means runtime param estimator might already very slow before this PR because of some reason, I'm confirming this by run master on runtime-param-estimator



Note: 
- 300 can be reasoned, it's from:
```
    *good_account.borrow_mut() = true;
    *curr_code.borrow_mut() = code_10k.to_vec();

    testbed = measure_transactions(
        Metric::ActionDeploy10K,
        &mut m,
        &config,
        Some(testbed),
        &mut f,
        false,
    );

    // Deploying more small code accounts. It's important that they are the same size to correctly
    // deduct base
    for _ in 0..2 {
        testbed =
            measure_transactions(Metric::warmup, &mut m, &config, Some(testbed), &mut f, false);
    }

    *good_account.borrow_mut() = false;
```
only when `*good_account=true` account that has test contract deployed will be added to `good_code_accounts`, which is then assign to `ad` as accounts used for measure in `    for (metric, method_name) in v {`. So it's 3 times above, one         `Metric::ActionDeploy10K`, two in for loop. Each time iter_per_block=1, block_sizes=[100], totally 3*1*100=300 good accounts.

- before measuring host functions, block size is set to 2:
```
    config.block_sizes = vec![2];

    // When adding new functions do not forget to rebuild the test contract by running `test-contract/build.sh`.
```
This cause we only randomly choose two account to do all host function measures, which I assume it's the designed behavior.

ailisp:
master still runs in around 6 hours, so the slowdown is indeed because of 300 write_x_key_y_value_1k, so i make a change to only  write_x_key_y_value_1k to account that going to be measure read or remove, instead of do so on 300 accounts, hopefully this will bring execution time back to 6 hours.

evgenykuzyakov:
Can you also take a look why `storage_write_evicted_byte` is lower than `storage_read_value_byte` and why it has decreased?

ailisp:
> Can you also take a look why storage_write_evicted_byte is lower than storage_read_value_byte and why it has decreased?

OK. The numbers above is depreciated as i changed to only set key value to the account that is going to read and delete key,value lazily, will look if new number is still this case.

core/primitives/src/version.rs
```diff
@@ -84,7 +84,7 @@ pub enum ProtocolFeature {
 
 /// Current latest stable version of the protocol.
 #[cfg(not(feature = "nightly_protocol"))]
-pub const PROTOCOL_VERSION: ProtocolVersion = 41;
+pub const PROTOCOL_VERSION: ProtocolVersion = 42;
```
bowenwang1996, line 87: why? What is the protocol change here?

runtime/runtime-params-estimator/src/cases.rs
```diff
@@ -35,26 +36,79 @@ fn measure_function(
...
-        let account_idx = *accounts_deployed.choose(&mut rng).unwrap();
+        let account_idx = loop {
+            let i = rand::thread_rng().gen::<usize>() % accounts_deployed.len();
+            let x = accounts_deployed[i];
+            if accounts_used.contains(&x) {
+                continue;
+            }
+            break x;
+        };
+        accounts_used.insert(account_idx);
```
evgenykuzyakov, line 59: Since `accounts_deployed` is fairly small, we should instead clone it and then draw and remove from it.

runtime/runtime-params-estimator/src/cases.rs
```diff
@@ -413,6 +477,8 @@ pub fn run(mut config: Config, only_compile: bool) -> RuntimeConfig {
 
     let ad: Vec<_> = good_code_accounts.into_iter().collect();
 
+    config.block_sizes = vec![2];
+
```
evgenykuzyakov, line 481: Why move it here?

runtime/runtime-params-estimator/src/cases.rs
```diff
@@ -478,17 +542,17 @@ pub fn run(mut config: Config, only_compile: bool) -> RuntimeConfig {
...
-    storage_read_10kib_key_10b_value_1k => storage_read_10kib_key_10b_value_1k,
     storage_has_key_10kib_key_10b_value_1k => storage_has_key_10kib_key_10b_value_1k,
+    storage_read_10kib_key_10b_value_1k => storage_read_10kib_key_10b_value_1k,
     storage_remove_10kib_key_10b_value_1k => storage_remove_10kib_key_10b_value_1k,
     storage_write_10b_key_10kib_value_1k => storage_write_10b_key_10kib_value_1k,
     storage_write_10b_key_10kib_value_1k_evict => storage_write_10b_key_10kib_value_1k,
-    storage_read_10b_key_10kib_value_1k => storage_read_10b_key_10kib_value_1k,
     storage_has_key_10b_key_10kib_value_1k => storage_has_key_10b_key_10kib_value_1k,
+    storage_read_10b_key_10kib_value_1k => storage_read_10b_key_10kib_value_1k,
     storage_remove_10b_key_10kib_value_1k =>   storage_remove_10b_key_10kib_value_1k ,
```
evgenykuzyakov, line 556: Is there a reason to move it now?

runtime/runtime-params-estimator/src/testbed_runners.rs
```diff
@@ -230,8 +234,9 @@ where
         for block_size in config.block_sizes.clone() {
             let block: Vec<_> = (0..block_size).map(|_| (*f)()).collect();
             let start = start_count(config.metric);
-            testbed.process_block(&block, allow_failures);
-            testbed.process_blocks_until_no_receipts(allow_failures);
+            let mut testbed_inner = testbed_clone.lock().unwrap();
```
evgenykuzyakov, line 237: This line should got before `start`

core/primitives/src/version.rs
ailisp: fees changed

ailisp: but I think let's going to change fees in https://github.com/near/nearcore/pull/3653 and do a one time protocol change to update fees? Since not only this PR, but other PRs have introduces many fee changes since last hard fork, and fees is not updated since then.

runtime/runtime-params-estimator/src/cases.rs
ailisp: Fixed

ailisp:
@evgenykuzyakov addressed all comments, ptal

evgenykuzyakov (approved):

ailisp:
@willemneal @olonho please review

olonho (approved):
LGTM

olonho:
Note that test fails with 
```
[2020-12-07T17:49:15Z] thread 'runtime::test::test_delete_account_after_unstake' panicked at 'Failed to open the database: DBError(Error { message: "IO error: While open directory: /tmp/test_validator_delete_accountrM135v/data: Too many open files" })', core/store/src/lib.rs:292:23```

ailisp:
> `` [2020-12-07T17:49:15Z] thread 'runtime::test::test_delete_account_after_unstake' panicked at 'Failed to open the database: DBError(Error { message: "IO error: While open directory: /tmp/test_validator_delete_accountrM135v/data: Too many open files" })', core/store/src/lib.rs:292:23``` ``

This happen quite often after we switch to use plain cargo test workspace in ci. I asked SRE team to take a look. (For sure unrelated to this PR because all change is in runtime-param-estimator :)

willemneal (approved):

PR #3622, opened 2020-11-16
fix(network): avoid unnecessary verification of sync data

bowenwang1996:
Currently we always first verify sync data (both account announcements and edges) before processing them, which can lead to a lot of repetitive verifications and even slow down the node as a result. This PR addresses the issue by:
* First check if edges or account announements need to be verified before verifying them.
* Move the verification of account signatures from view client to peer manager to reduce overhead.

Test plan
----------
* unit tests
* nayduck http://nayduck.eastus.cloudapp.azure.com:3000/#/run/734


chain/client/tests/process_blocks.rs
```diff
@@ -68,7 +67,6 @@ pub fn create_nightshade_runtimes(genesis: &Genesis, n: usize) -> Vec<Arc<dyn Ru
 /// Runs block producing client and stops after network mock received two blocks.
 #[test]
 fn produce_two_blocks() {
-    init_test_logger();
```
mfornet, line 71: Why removing test_logger everywhere?

mfornet (approved):
Everything looks good. 

It still to be seen the impact on the performance. One way to notice it with a simple pytest is creating one nodes and submitting ~10k edges, then repeating the same edges and checking the time for the second update is negligible compared to the first.

chain/client/tests/process_blocks.rs
bowenwang1996: It actually prevented me from seeing which test was failing due to the amount of log output

neard/src/runtime.rs
```diff
@@ -65,6 +64,12 @@ impl AsRef<RwLock<EpochManager>> for SafeEpochManager {
     }
 }
 
+impl Clone for SafeEpochManager {
+    fn clone(&self) -> Self {
+        SafeEpochManager(self.0.clone())
+    }
+}
```
frol, line 71: This is effective the default implementation for `#[derive(Clone)]`

frol (approved):
I have reviewed everything but the network part. LGTM

bowenwang1996:
> It still to be seen the impact on the performance. One way to notice it with a simple pytest is creating one nodes and submitting ~10k edges, then repeating the same edges and checking the time for the second update is negligible compared to the first.

Added time measurement in the unit test

SkidanovAlex (approved):

PR #3623, opened 2020-11-17
fix(chain): validate chunk mask

### LINKED ISSUE #3586: Block::check_validity should check chunk mask
`Block::check_validity` only checks that the number of chunks included is correct, but doesn't check against chunk mask in `BlockHeaderV2`. This means that with multiple shards it's possible to manipulate `ShardChunkHeader::height_included` values in the block while still having it match the header.

mikhailOK:
Make header validation match chunk mask against chunks_included for V1
headers.
Make block validation match chunk mask in header against computed mask
from block body. Don't check chunks_included anymore, as it's checked
in header validation.
Also refactor chunks_included logic in Block::genesis as it was
unnecessarily complex and did not match the logic for non-genesis
blocks anyway.

Fixes #3586

Test plan
---------
Unit test

### DIFF
--- a/chain/chain/src/chain.rs
+++ b/chain/chain/src/chain.rs
@@ -3181,10 +3181,7 @@ impl<'a> ChainUpdate<'a> {
             return Err(ErrorKind::InvalidChunkMask.into());
         }
 
-        // Actually only needed for BlockHeaderV1
-        if header.chunk_mask().iter().map(|&x| u64::from(x)).sum::<u64>()
-            != header.chunks_included()
-        {
+        if !header.verify_chunks_included() {
             return Err(ErrorKind::InvalidChunkMask.into());
         }
 
--- a/chain/chain/src/chain.rs
+++ b/chain/chain/src/chain.rs
@@ -3181,6 +3181,13 @@ impl<'a> ChainUpdate<'a> {
             return Err(ErrorKind::InvalidChunkMask.into());
         }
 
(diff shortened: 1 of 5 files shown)


chain/chain/src/chain.rs
```diff
@@ -3181,6 +3181,13 @@ impl<'a> ChainUpdate<'a> {
             return Err(ErrorKind::InvalidChunkMask.into());
         }
 
+        // Actually only needed for BlockHeaderV1
```
bowenwang1996, line 3184: why don't we do this only for `BlockHeaderV1` then

core/primitives/src/block_header.rs
```diff
@@ -357,13 +357,14 @@ impl BlockHeader {
...
         chunk_tx_root: MerkleHash,
-        chunks_included: u64,
+        num_shards: u64,
         challenges_root: MerkleHash,
         timestamp: DateTime<Utc>,
         initial_gas_price: Balance,
         initial_total_supply: Balance,
         next_bp_hash: CryptoHash,
     ) -> Self {
+        let chunks_included = if height == 0 { num_shards } else { 0 };
```
bowenwang1996, line 367: why does height matter here?

mikhailOK: old code was `Block::compute_chunks_included(&chunks, 0)` with `0` instead of `height`

bowenwang1996 (approved):

PR #3630, opened 2020-11-19
refactor (indexer): Move receipt_execution_outcomes to IndexerChunkView from StreamerMessage. Bump Indexer version to 0.5.0

khorolets:


chain/indexer/src/streamer/mod.rs
```diff
@@ -51,22 +51,32 @@ async fn build_streamer_message(
...
             chunk;
-
-        let indexer_transactions = transactions
-            .into_iter()
-            .map(|transaction| {
-                IndexerTransactionWithOutcome {
-                    outcome: outcomes.remove(&transaction.hash).expect("The transaction execution outcome should always present in the same block as the transaction itself"),
-                    transaction,
-                }
+        let mut outcomes = shards_outcomes.get_mut(&header.shard_id).unwrap().clone();
```
frol, line 60: 1. Let's always use `.expect` with a message why we believe this cannot fail.
2. Use [`.remove()`](https://doc.rust-lang.org/stable/std/collections/struct.HashMap.html#method.remove) instead of `get()` + `clone()`

chain/indexer/src/streamer/mod.rs
```diff
@@ -51,22 +51,32 @@ async fn build_streamer_message(
...
+        }
+        // let indexer_transactions = transactions
+        //     .into_iter()
+        //     .map(|transaction| {
+        //         IndexerTransactionWithOutcome {
+        //             outcome: outcomes.remove(&transaction.hash).expect("The transaction execution outcome should always present in the same block as the transaction itself"),
+        //             transaction,
+        //         }
+        //     })
+        //     .collect::<Vec<IndexerTransactionWithOutcome>>();
```
frol, line 79:
```suggestion
```

chain/indexer/src/streamer/mod.rs
```diff
@@ -84,30 +94,32 @@ async fn build_streamer_message(
         let mut chunk_receipts = chunk_local_receipts;
         chunk_receipts.extend(chunk_non_local_receipts);
 
+        // Add local receipts to corresponding outcomes
+        for receipt in &local_receipts {
+            if let Some(outcome_position) = outcomes
+                .iter()
+                .position(|outcome| outcome.execution_outcome.id == receipt.receipt_id)
```
frol, line 101: Use [`find`](https://doc.rust-lang.org/stable/std/iter/trait.Iterator.html#method.find) instead of `position`, to get the element itself instead of the position, so you don't need to `get` + `expect`.

frol (changes requested):

chain/indexer/src/streamer/mod.rs
khorolets: find doesn't return mutable, and I need to mutate it

khorolets: ah, sorry `iter_mut` did the trick

frol (approved):

chain/indexer/src/streamer/mod.rs
```diff
@@ -60,16 +60,22 @@ async fn build_streamer_message(
...
-        for transaction in transactions {
-            let outcome_position = outcomes.iter().position(|outcome| outcome.execution_outcome.id == transaction.hash)
-                .expect("The transaction execution outcome should always present in the same block as the transaction itself");
-            indexer_transactions.push(IndexerTransactionWithOutcome {
-                outcome: outcomes.remove(outcome_position),
-                transaction,
+        let mut receipt_outcomes = outcomes.split_off(transactions.len());
+
+        let indexer_transactions = transactions
+            .iter()
```
frol, line 66: Use `into_iter()` to avoid unnecessary transaction clonning

chain/indexer/src/streamer/mod.rs
```diff
@@ -60,16 +60,22 @@ async fn build_streamer_message(
...
+            .enumerate()
+            .map(|(index, transaction)| {
+                assert_eq!(
+                    outcomes[index].execution_outcome.id, transaction.hash,
```
frol, line 70: Use [`zip`](https://doc.rust-lang.org/stable/std/iter/trait.Iterator.html#method.zip) instead of `enumerate` + fetching by index.

frol (changes requested):

frol (approved):

PR #3631, opened 2020-11-19
Removed the no-color flag and added a log_color bool in the config.json

chefsale:
The --no-color with ansi charset disabled didn't work on the messages which we were logging, but only on the default syslog output produced the logger. Now I added a flag which check if the log_color flag is set to log the info colored. By default the coloring is enabled.


chain/client/src/info.rs
```diff
@@ -48,6 +48,8 @@ pub struct InfoHelper {
     validator_signer: Option<Arc<dyn ValidatorSigner>>,
     /// Telemetry actor.
     telemetry_actor: Addr<TelemetryActor>,
+    /// Log coloring enabled
+    log_color: bool,
```
frol, line 52: Considering the whole structure, I find this name to be not very descriptive. Consider something along these lines: `should_use_color_for_logging` (I am not totally sold on it as it is quite long, but at least it is immediately obvious how one may want to use it)

core/chain-configs/src/client_config.rs
```diff
@@ -75,6 +75,8 @@ pub struct ClientConfig {
     pub tracked_shards: Vec<ShardId>,
     /// Not clear old data, set `true` for archive nodes.
     pub archive: bool,
+    /// Enable coloring of the logs
+    pub log_color: bool,
```
frol, line 79: I don't think we need to put it into chain config. The CLI argument should be enough. The same CLI argument should control all the colored output.

My reasoning is that the config is currently used to tweak long-term options while colors can be on in one context and off in another.

chain/client/src/info.rs
```diff
@@ -112,14 +115,26 @@ impl InfoHelper {
         } else {
             White.bold().paint("")
```
frol, line 116: These paints are not gated

chain/client/src/info.rs
```diff
@@ -112,14 +115,26 @@ impl InfoHelper {
...
+            info!(target: "stats", "{} {} {} {} {} {}",
+                  Yellow.bold().paint(display_sync_status(&sync_status, &head, genesis_height)),
+                  validator_info_log,
+                  Cyan.bold().paint(format!("{:2}/{:?}/{:2} peers", network_info.num_active_peers, network_info.highest_height_peers.len(), network_info.peer_max_count)),
+                  Cyan.bold().paint(format!("⬇ {} ⬆ {}", pretty_bytes_per_sec(network_info.received_bytes_per_sec), pretty_bytes_per_sec(network_info.sent_bytes_per_sec))),
+                  Green.bold().paint(format!("{:.2} bps {}", avg_bls, gas_used_per_sec(avg_gas_used))),
+                  Blue.bold().paint(format!("CPU: {:.0}%, Mem: {}", cpu_usage, pretty_bytes(memory_usage * 1024)))
+            );
+        } else {
+            info!(target: "stats", "{} {} {} {} {} {}",
```
frol, line 129: Instead of copy-pasting, I suggest we prepare the values first, then conditionally apply the colors to them, and then print them out.

With the current approach, it is very easy to introduce discrepancy between colored and non-colored output

frol (changes requested):

core/chain-configs/src/client_config.rs
chefsale: So, the problem with using it as an argument we need to propagate it all way around through several strcutures just to tweak on if which is using it, and also having it in the config is kind of long term thing, as if you decide not to use color you dont want to pass the flag everytime to the binary.

chain/client/src/info.rs
chefsale: colored_logging_enabled maybe would be better?

core/chain-configs/src/client_config.rs
frol: Well, I hear you. My point is that you usually want to run it without color in a supervisor, while you may want to run it with colors when you run it directly, and in this scenario, you will need to edit the config file just to tweak this.

I don't want us to spend more time on it than necessary, so I can agree on having it in the config. I would like to ask to have it next to the relevant `log_summary_period` configuration option, so we probably want to call it `log_summary_style` with string options of `"plain"` or `"colored"` (in the future we may introduce more of those, and I think it is easier to reason about the configuration option when it is not a boolean flag). To implement it, just introduce an enum with `#[derive(Serialize, Deserialize)]` and use it as a type of value.

chain/client/src/info.rs
bowenwang1996: > colored_logging_enabled maybe would be better?

Yeah let's do that

chain/client/src/info.rs
```diff
@@ -103,24 +106,51 @@ impl InfoHelper {
...
-            White.bold().paint(format!(
+            format!(
                 "{}/{}",
                 if validator_info.is_validator { "V" } else { "-" },
                 validator_info.num_validators
-            ))
+            )
         } else {
-            White.bold().paint("")
+            String::from("")
```
bowenwang1996, line 117:
```suggestion
            String::new()
```

bowenwang1996 (approved):

core/chain-configs/src/client_config.rs
chefsale: TIL some more rust, added the serde enum deserialization for the config like you suggested.

frol (approved):
Beautiful!

PR #3635, opened 2020-11-20
fix(test): test_ban_peer_for_invalid_block_header

### LINKED ISSUE #3634: test_ban_peer_for_invalid_block_header is flaky
test_ban_peer_for_invalid_block_header is flaky when run with nightly protocol features (`--features nightly_protocol,nightly_protocol_features`)

mikhailOK:
The test messes with the block at height 4, and fails if that height
gets skipped. Change it to the first block at height >=4.
Fixes #3634

Test plan
---------
run locally


bowenwang1996 (approved):
Thanks

PR #3636, opened 2020-11-20
feat (indexer): Extend IndexerConfig to be available to turn off requirement the node to be fully synced before streaming. Update Indexer example

khorolets:


chain/indexer/src/streamer/mod.rs
```diff
@@ -135,13 +135,18 @@ pub(crate) async fn start(
...
 
     'main: loop {
-        time::delay_for(INTERVAL).await;
-        let status = fetch_status(&client).await;
-        if let Ok(status) = status {
-            if status.sync_info.syncing {
-                continue;
+        match indexer_config.await_for_node_synced {
+            AwaitForNodeSyncedEnum::WaitForFullSync => {
+                time::delay_for(INTERVAL).await;
```
frol, line 140: It seems we need to have the delay regardless to the mode

khorolets, line 140: @frol I'm not sure about putting delay only in await for node to be synced case or in any case.

chain/indexer/src/lib.rs
```diff
@@ -30,13 +30,24 @@ pub enum SyncModeEnum {
     BlockHeight(u64),
 }
 
+/// Enum to define whether await for node to be fully synced or stream while syncing (useful for indexing from genesis)
+#[derive(Debug, Clone)]
+pub enum AwaitForNodeSyncedEnum {
```
frol, line 35: Well done!

frol (changes requested):

frol (approved):

PR #3639, opened 2020-11-21
Security advisory fix 1

pmnoxx:


chain/network/src/peer_manager.rs
```diff
@@ -145,6 +148,9 @@ pub struct PeerManagerActor {
     metric_recorder: MetricRecorder,
     edge_verifier_pool: Addr<EdgeVerifier>,
     txns_since_last_block: Arc<AtomicUsize>,
+    pending_incoming_connections_counter: Arc<AtomicUsize>,
+    peer_counter: Arc<AtomicUsize>,
+    test_ignore_reading: bool,
```
bowenwang1996, line 153: Why do we need this? Ideally we should not add something to the struct that is used only for testing purposes.

pmnoxx: Naptha attack relies on manipulating TCP state machine, to make server think it's in connected state, while client doesn't need to keep the state at all. 
pinger.py opens TCP connection, which is managed by Linux kernel. Pinger.py requires root permissions to execute iptables command, which tell Linux to ignore outgoing FIN, RST TCP messages
`
    system('iptables -A OUTPUT -d %s -p tcp --dport %d --tcp-flags FIN FIN -j DROP' %( target_ip, port ))
    system('iptables -A OUTPUT -d %s -p tcp --dport %d --tcp-flags RST RST -j DROP' %( target_ip, port ))
`
That's why we can't use this technique in the unit test, even if we had root permissions, this solution isn't portable.

If you want to create a client which can do this attack in a portable way without modifying the kernel, then you will have to write a non kernel space program implementing TCP [https://tools.ietf.org/html/rfc793](RFC 793)

I found one user, which claims he implemented partially TCP in Rust. https://github.com/jonhoo/rust-tcp/tree/master/src
It may be possible to modify it's source code to reproduce this attack. However, it requires investing a bit of time to do the necessary changes.

I'm open to any other suggestions.

pmnoxx: I watched the description of how  https://github.com/jonhoo/rust-tcp/tree/master/src works. He creates a tun (/dev/net/tun) device, which is a kernel network interface, which provides interface, which behaves like a network adapter.
However, this method requires you root permissions, and it's not portable as well.

bowenwang1996: I am confused. Couldn't we simply send 1000 `TcpInboundConnect` message to `PeerManagerActor` and assert that the peer counter size is bounded by a small number?

pmnoxx: > I am confused. Couldn't we simply send 1000 `TcpInboundConnect` message to `PeerManagerActor` and assert that the peer counter size is bounded by a small number?

In the current implementation we are checking peer counter size before TcpInboundConnect is created. You can take a look at peer_manager.rs:1140 inside PeerManagerActor::started

pmnoxx: I simplified the unit test.

pmnoxx: test_ignore_reading is no longer needed

bowenwang1996 (approved):

PR #3640, opened 2020-11-21
fix(chunk): address some suboptimalities in shards manager

bowenwang1996:
Address a couple small issues in chunk handling:
* `ChunkHash` is already a hash, so it doesn't make sense to hash it again.
* Avoid some unnecessary cloning in `ShardsManager::request_partial_encoded_chunk`

Test plan
---------
existing tests


frol (approved):

core/primitives/src/sharding.rs
```diff
@@ -47,6 +47,12 @@ impl From<CryptoHash> for ChunkHash {
     }
 }
 
+impl Hash for ChunkHash {
+    fn hash<H: Hasher>(&self, state: &mut H) {
```
birchmd, line 51: I'm not certain whether this will perform better than the derived `Hash`. Note that [`Hasher::write` adds bytes to the state of some external hasher](https://doc.rust-lang.org/std/hash/trait.Hasher.html#tymethod.write), and [the hash is actually computed on a call to `Hasher::finish`](https://doc.rust-lang.org/std/hash/trait.Hasher.html#tymethod.finish), so this change is not preventing the computation of any hash.

Granted, this is writing the bytes directly into the `Hasher` state with no extra processing, so it might be marginally faster, but the reason I am not sure is because [the `Hash` derive macro is built in to the rust compiler](https://doc.rust-lang.org/src/core/hash/mod.rs.html#200-202). Generally I assume functionality built in to the compiler itself is heavily optimized so an explicit code version is unlikely to be faster.

birchmd: It's also worth mentioning that could cause code to behave incorrectly: https://rust-lang.github.io/rust-clippy/master/#derive_hash_xor_eq

(though in this case I doubt it would cause a problem)

bowenwang1996: @birchmd I see. So we need to implement our own `Hasher`?

birchmd: Note that the return type of `Hasher::finish` is `u64`, so this only applies to non-cryptographic hash functions, like those involved in `HashMap`s for example. Any `Hasher` implementation will need to do something to the bytes of a `ChunkHash` to meet this interface, so simply returning the bytes is not an option (though it could be as simple as taking the first 8 bytes for example).

The `Hasher` interface also does not allow special treatment of data as far as I can tell, so there would be no way to say something like "given only a `ChunkHash`, return the first 8 bytes". And even if we could do that, I'm not sure it would help very much because it would only be replacing a non-cryptographic hash function, which likely is very fast to compute in the first place.

frol: Well, unless we measure the performance, let's not spend our time optimizing things introducing code that we don't understand the performance characteristics of :thinking:

bowenwang1996: I reverted my change

SkidanovAlex (approved):

PR #3647, opened 2020-11-23
fix(Add neard to release scripts)

mhalambek:
Deprecate near and use neard.


bowenwang1996 (approved):
Can we also remove `keypair-generator` and `genesis-csv-to-json`? Don't really need those

PR #3648, opened 2020-11-23
fix: Fix delete_key storage_usage off by one error

evgenykuzyakov:
Fix `storage_usage` subtraction to be correct. Previously it included one extra byte for type of `Option<AccessKey>` instead of `AccessKey`.

# Test plan:
- CI
- new unit tests

### DIFF
--- a/core/primitives/src/version.rs
+++ b/core/primitives/src/version.rs
@@ -49,6 +49,9 @@ pub const CREATE_HASH_PROTOCOL_VERSION: ProtocolVersion = 38;
 
 pub const SHARD_CHUNK_HEADER_UPGRADE_VERSION: ProtocolVersion = 40;
 
+/// Fix the storage usage of the delete key action.
+pub const DELETE_KEY_STORAGE_USAGE_PROTOCOL_VERSION: ProtocolVersion = 41;
+
 pub struct ProtocolVersionRange {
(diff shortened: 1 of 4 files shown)


bowenwang1996 (approved):

evgenykuzyakov:
> `buildkite/nearcore/db-migration` test 

This test fails because we have a fork between `1.16.2` protocol `40` which is live on `mainnet` with this PR change due to security release. And we have `master` branch with the protocol version `40` which is only live on the `betanet` right now.

This causes a conflict that breaks either `backward-compatible` and `upgradability` tests or `db-migration` test. Since `mainnet` is already on protocol `40` with the `1.16.2` we can't modify what protocol `40` is. So it has to stay `DELETE_KEY_STORAGE_USAGE_PROTOCOL_VERSION`.

While for `SHARD_CHUNK_HEADER_UPGRADE_VERSION` we have to upgrade it to `41`, because it's not released yet.

This is unfortunate and seems we have to think how to do future releases to leave space for security fixes. I think one of the best ways is to skip versions. Then for example, current protocol could have been `40`, `SHARD_CHUNK_HEADER_UPGRADE_VERSION` can be `50` and the security fix can be `41`.

But because we had a conflict we have to ignore the `db-migration` test and merge it anyway.

PR #3652, opened 2020-11-23
fix: cache epoch validators

### LINKED ISSUE #3650: Optimize `EpochManager::get_all_block_producers_settlement`
Profiling shows that `get_all_block_producers_settlement` is very hot and it is called in a number of places, including `NightShadeRuntime::get_part_owner`, `NightShadeRuntime::verify_approval`, `ShardsManager::get_random_target_tracking_shard`, etc.

bowenwang1996:
Cache epoch validators in `EpochManager::get_all_block_producers_settlement`. Fixes #3650.

Test plan
---------
`test_epoch_validators_cache`


chain/epoch_manager/src/lib.rs
```diff
@@ -519,35 +525,42 @@ impl EpochManager {
...
-        let slashed = self.get_slashed_validators(last_known_block_hash)?.clone();
-        let epoch_info = self.get_epoch_info(epoch_id)?;
-        let mut settlement = vec![];
-        for validator_id in epoch_info.block_producers_settlement.iter() {
-            let validator_stake = epoch_info.validators[*validator_id as usize].clone();
-            let is_slashed = slashed.contains_key(&validator_stake.account_id);
-            settlement.push((validator_stake, is_slashed));
+    ) -> Result<&[(ValidatorStake, bool)], EpochError> {
+        // TODO: Revisit this when we enable slashing
+        if self.epoch_validators_ordered.cache_get(epoch_id).is_none() {
```
birchmd, line 530: Could maybe optimize for the case where the `epoch_id` is known by using a `match` (presently there are always two cache lookups, even when already present)

```rust
match self.epoch_validators_ordered.cache_get(epoch_id) {
    Some(settlement) => Ok(settlement),
    None => {
        ... // compute settlement
        self.epoch_validators_ordered.cache_set(epoch_id.clone(), settlement);
        Ok(self.epoch_validators_ordered.cache_get(epoch_id).unwrap())
    }
}
```

chain/epoch_manager/src/lib.rs
```diff
@@ -519,35 +525,42 @@ impl EpochManager {
...
-        let settlement =
-            self.get_all_block_producers_settlement(epoch_id, last_known_block_hash)?;
-        let mut result = vec![];
-        let mut validators: HashSet<AccountId> = HashSet::default();
-        for (validator_stake, is_slashed) in settlement.into_iter() {
-            if !validators.contains(&validator_stake.account_id) {
-                validators.insert(validator_stake.account_id.clone());
-                result.push((validator_stake, is_slashed));
+    ) -> Result<&[(ValidatorStake, bool)], EpochError> {
+        if self.epoch_validators_ordered_unique.cache_get(epoch_id).is_none() {
```
birchmd, line 550: Same here

chain/epoch_manager/src/lib.rs
bowenwang1996: cannot get around lifetime restrictions

birchmd: I see. I guess since the function returns a reference it needs to thread the same lifetime all the way through no matter how you scope it. Ok, I suppose we have to live with 2 cache hits per call. Thanks for checking.

frol (approved):

SkidanovAlex (approved):

PR #3656, opened 2020-11-26
Fix deserialization of cached error.

olonho:
Distinguish deserialized errors and errors coming from deserialization itself.

### DIFF
--- a/runtime/near-vm-runner/src/cache.rs
+++ b/runtime/near-vm-runner/src/cache.rs
@@ -72,7 +72,7 @@ fn compile_and_serialize_wasmer(
 
 /// Deserializes contract or error from the binary data. Signature means that we could either
 /// return module or cached error, which both considered Ok(), or encounter an error during
-/// deserialization process.
+/// the deserialization process.
 fn deserialize_wasmer(
     serialized: &[u8],
 ) -> Result<Result<wasmer_runtime::Module, VMError>, CacheError> {
--- a/runtime/near-vm-runner/src/cache.rs
+++ b/runtime/near-vm-runner/src/cache.rs
@@ -1,11 +1,10 @@
 use crate::errors::IntoVMError;
 use crate::prepare;
 use borsh::{BorshDeserialize, BorshSerialize};
-use log::error;
 use near_primitives::hash::CryptoHash;
 use near_primitives::types::CompiledContractCache;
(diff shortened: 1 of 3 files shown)


runtime/near-vm-runner/src/cache.rs
```diff
@@ -71,19 +71,26 @@ fn compile_and_serialize_wasmer(
     Ok(module)
 }
 
-fn deserialize_wasmer(serialized: &[u8]) -> Result<wasmer_runtime::Module, VMError> {
+/// Deserializes contract or error from the binary data. Signature means that we could either
+/// return module or cached error, which both considered Ok(), or encounter an error during
+/// deserialization process.
+fn deserialize_wasmer(
+    serialized: &[u8],
+) -> Result<Result<wasmer_runtime::Module, VMError>, VMError> {
```
frol, line 79: Can we have a dedicated error enum with: DeserializationError, CompilationError(VMError) instead of Result<Result<_, VMError>, VMError> (it is really hard to reason about this return value type)

olonho: That's hard question. If we think about our API as "stack", than compilation error remembered earlier and stored in the database is indeed success, and not an error in deserialization logic, while it's an error in higher level function. SO IMO current signature is pretty representative to that notion, and fortunately it doesn't escape cache.rs anyway.

runtime/near-vm-runner/src/cache.rs
```diff
@@ -102,7 +109,7 @@ pub(crate) fn compile_module_cached_wasmer(
     match cache.get(&(key.0).0) {
         Ok(serialized) => match serialized {
             Some(serialized) => match deserialize_wasmer(serialized.as_slice()) {
-                Ok(module) => Ok(module),
+                Ok(module_or_error) => module_or_error,
                 Err(e) => {
                     error!(target: "runtime", "Cannot deserialize cached contract: {:?}", e);
```
frol, line 114: It seems that given the current implementation, it is a normal (not an erroneous) operation mode to receive a deserialization error return value.

frol: Or not? Hmm, I got confused here

frol (approved):

bowenwang1996 (approved):

Issue #3180, opened 2020-08-16
Revisit seals implementation


SkidanovAlex:
I was debbugging an issue today related to (but not really caused by) Seals, and discovered that the implementation is likely incorrect.
 Given we do not need seals for the single shard setting, in which the chunk producer is almost always the same as the block producer, I disabled the seals for now.

Specific concerns:

1\. This logic of retaining seals appears incorrect:

```
 self.active_demurs.retain(|chunk_hash, seal| {
 let accepting_period_over = (now - seal.sent).num_milliseconds() > ACCEPTING_SEAL_PERIOD_MS;
 let parts_remain = seal.part_ords.len() > NUM_PARTS_LEFT_IN_SEAL;

 // note chunk producers that failed to make parts available
 if parts_remain && accepting_period_over {
 warn!(target: "client", "Couldn't reconstruct chunk {:?} from {:?}, I'm {:?}", chunk_hash, seal.chunk_producer, me);
 dont_include_chunks_from.cache_set(seal.chunk_producer.clone(), ());
 Self::insert_past_seal(past_seals, seal.height, chunk_hash.clone());

 // Do not retain this demur, it has expired
 false
 } else 

 { true } 

 });
 ```

It seems like `else if !parts_remain 

 { false } 

` is missing. It is not only an optimization (keeping fewer entries), but also a correctness issue. Without this line, a seal that only has 1 part left (which is a sufficient condition to consider it fulfilled) will remain in the `active_demurs`, meaning that it will continue blocking accepting the chunks that have this seal (see below), because the `seal.process` that is used to check if seal is fulfilled only returns `true` if **all** the parts are received.

2\. We likely should not have to block the acceptance of the chunk on absence of seals. Today it is blocked:

```
 let have_all_seal = seal.process(entry);

 if have_all_parts && have_all_receipts && have_all_seal {
 ```

Ideally we should accept the chunk as soon as it has all the parts and the receipts, but then continue accepting seals, and keep on tracking if we succeeded accepting all of them. It appears that it's suffucient to remove the `have_all_seals` above, and add `seal.process` in early exits in the method (specifically the place where we exit if we have all parts and all receipts), but more analysis and testing is needed.

[...]

PR #3657, opened 2020-11-26
fix(seals) - Accept block before all seals obtained, continue tracking seals afterward

birchmd:
@SkidanovAlex pointed out some problems with the seals logic in https://github.com/near/nearcore/issues/3180

In this PR we address those concerns. Fixes https://github.com/near/nearcore/issues/3180

Testing plan:

* Existing tests
* New case for `test_track_seals`


chain/client/src/client.rs
```diff
@@ -803,8 +803,27 @@ impl Client {
                 )?;
 
                 match process_result {
-                    ProcessPartialEncodedChunkResult::Known => Ok(vec![]),
-                    ProcessPartialEncodedChunkResult::HaveAllPartsAndReceipts(prev_block_hash) => {
+                    ProcessPartialEncodedChunkResult::Known(demurs_remain) => {
+                        if demurs_remain {
```
bowenwang1996, line 807: do we actually need to do this? if we have remaining demurs the request is not removed from the request pool, which means that it will be retried later.

chain/chunks/src/lib.rs
```diff
@@ -969,6 +966,18 @@ impl ShardsManager {
         }
     }
 
+    fn validate_parts<'a, I>(&mut self, merkle_root: MerkleHash, parts: I) -> Result<(), Error>
+    where
+        I: IntoIterator<Item = &'a PartialEncodedChunkPart>,
+    {
+        let num_total_parts = self.runtime_adapter.num_total_parts();
+        for part_info in parts {
+            // TODO: only validate parts we care about
```
bowenwang1996, line 975: Can we create an issue for this if you do not plan to address it in this PR?

chain/chunks/src/lib.rs
```diff
@@ -160,24 +161,16 @@ struct ActiveSealDemur {
...
-                    } else {
-                        false
-                    }
-                });
-                res
+                for ord in parts {
+                    demur.part_ords.remove(ord);
+                }
             }
+            Seal::Past => (),
```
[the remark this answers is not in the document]
birchmd: Done in https://github.com/near/nearcore/pull/3657/commits/30be273f0617e4267878988e76573be9ce49516e

chain/chunks/src/lib.rs
```diff
@@ -66,12 +65,14 @@ pub enum ChunkStatus {
...
+pub type DemursRemain = bool;
+
 #[derive(Debug)]
 pub enum ProcessPartialEncodedChunkResult {
-    Known,
+    Known(DemursRemain),
     /// The CryptoHash is the previous block hash (which might be unknown to the caller) to start
     ///     unblocking the blocks from
-    HaveAllPartsAndReceipts(CryptoHash),
+    HaveAllPartsAndReceipts(CryptoHash, DemursRemain),
```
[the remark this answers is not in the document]
birchmd: Done in https://github.com/near/nearcore/pull/3657/commits/30be273f0617e4267878988e76573be9ce49516e

chain/client/src/client.rs
birchmd: Yeah, good point. Removed in https://github.com/near/nearcore/pull/3657/commits/30be273f0617e4267878988e76573be9ce49516e

chain/chunks/src/lib.rs
birchmd: Created https://github.com/near/nearcore/issues/3662

bowenwang1996 (approved):
Let's run nayduck with this change to see if there is any regression

birchmd:
@bowenwang1996 Nayduck run: http://nayduck.eastus.cloudapp.azure.com:3000/#/run/801

Looks like maybe `expensive near-client cross_shard_tx tests::test_cross_shard_tx_drop_chunks` is a new failure. Others seem to be flaky on `master` too.

birchmd:
@SkidanovAlex Something to consider when you are doing your review:

I believe the reason `test_cross_shard_tx_drop_chunks` fails in Nayduck is because the nodes all stop accepting each other's chunks since they fail to fulfil the seals due to the chunks being dropped.

There are a few solutions here, but I wanted to check with you to see what the intention is with this test. The possible solutions are:

1. Increase the seal acceptance time; if we make this long enough seals should always be fulfilled eventually, even with some chunks not getting through
2. Make the fraction of chunks dropped lower; this increases the fraction of fulfilled seals so that hopefully not too many nodes ban each other's chunks
3. Some combination of 1. and 2.

The reason the intention of the test is important is because it tells me what I should change. If the test is to be interpreted as "the network must function when only 90% of chunks are received" then solution 2. is off the table and I must increase the seal acceptance time. If the test is only that "the network must function when some fraction of chunks are received" then I have more options in terms of what I change, and this will be informed by what we want the seal acceptance time to be in production.

chain/chunks/src/lib.rs
```diff
@@ -1266,11 +1276,17 @@ impl ShardsManager {
             // If we do care about the shard, we will remove the request once the full chunk is
             //    assembled.
             if !cares_about_shard {
+                let demurs_remain = self.seals_mgr.has_active_demur(&chunk_hash);
                 self.encoded_chunks.remove_from_cache_if_outside_horizon(&chunk_hash);
-                self.requested_partial_encoded_chunks.remove(&chunk_hash);
-                return Ok(ProcessPartialEncodedChunkResult::HaveAllPartsAndReceipts(
+                // We can only remove this from requested chunks if there are not parts
+                // we are waiting for to fulfil demurs.
+                if !demurs_remain {
```
SkidanovAlex, line 1283: If they have remained, when would we remove the request?
It seems that the next time we receive the seals, we will not get to here, we will exit above when we check if all the parts and all the receipts are present (and we return `Known`), and we do not appear to remove the request there

birchmd: The request is removed on the very next line. This is why I added this `if` statement; we should not remove the request until enough seals have been fulfilled. It used to be the case that we did not enter this `have_all_parts && have_all_receipts` block while there were still active demurs, but since we are not blocking adding the chunk anymore now we do enter this block. However, we still cannot remove the chunk from the set of requests until after we have the seals too.

Unrelated note: please also comment on https://github.com/near/nearcore/pull/3657#issuecomment-738983558 when you have a chance.

SkidanovAlex: My question is specifically about the situation in which some demurs are still left here. We only remove the request on the next line if no demurs remain, if some remain, then we keep the request.

The next time the parts arrive, we will exit early on line 1159. There's new logic there that will process the seals, but not the logic that would remove the request, so it seems like the request will remain forever?

SkidanovAlex:
> @SkidanovAlex Something to consider when you are doing your review:
> 
> I believe the reason `test_cross_shard_tx_drop_chunks` fails in Nayduck is because the nodes all stop accepting each other's chunks since they fail to fulfil the seals due to the chunks being dropped.
> 
> There are a few solutions here, but I wanted to check with you to see what the intention is with this test. The possible solutions are:
> 
>     1. Increase the seal acceptance time; if we make this long enough seals should always be fulfilled eventually, even with some chunks not getting through
> 
>     2. Make the fraction of chunks dropped lower; this increases the fraction of fulfilled seals so that hopefully not too many nodes ban each other's chunks
> 
>     3. Some combination of 1. and 2.
> 
> 
> The reason the intention of the test is important is because it tells me what I should change. If the test is to be interpreted as "the network must function when only 90% of chunks are received" then solution 2. is off the table and I must increase the seal acceptance time. If the test is only that "the network must function when some fraction of chunks are received" then I have more options in terms of what I change, and this will be informed by what we want the seal acceptance time to be in production.

I think we should do (1) anyway, but also not being able to fetch 3 seals in 30 seconds is rather suspicious. It could be that there's an actual bug. We re-request around every 400ms, right?

Though the `cross_shard_tx_*` tests are known to make the system slow, so could be that the actual time it takes to re-request is larger.

If there's indeed a bug, it will remain if we increase the timeout, so increasing it to 60s-120s first is a good start.

If it doesn't help (and I suspect it won't), the way I would approach it is add temporarily some tracing that periodically prints the remaining seals, and for how long they were active, and also trace dropped requests in the test itself. If indeed the hypothesis about 30s not being sufficient is correct, we will see that the chunk is dropped repeatedly, or otherwise will get some insight into what's going on.

chain/chunks/src/lib.rs
birchmd: Ah, ok. Sorry I misunderstood your question. Yes you are right that I was not properly removing the request if the seal was fulfilled after the other parts were known.

I have changed this in https://github.com/near/nearcore/pull/3657/commits/ca7556b5f4e0bfc77f8cee6e84cf6b540e0f55f3
It is a bit larger of a commit than strictly necessary to fix this issue, but I also refactored `track_seals` so that it does not iterate through all active seals on every call and I beefed up the unit testing to capture the requests getting removed.

Please let me know if you notice anything else wrong in the updated logic.

akhi3030:
This PR does not look active to me anymore.

PR #3658, opened 2020-11-27
feat(testing): bridge testing infrastructure major update

Kouprin:


bowenwang1996 (commented):
a not-so-related question: why do we put the e2e test in nearcore? Couldn't we trigger nayduck from some bridge repo? @Kouprin @utka

Kouprin:
@bowenwang1996 not sure what do you mean. We reuse bridge repo if found, otherwise git-clone it.

There are many tests that need Nearcore + Bridge at the same time. There are three options of what we can do:
1. Put the tests to the Bridge (we have couple of sanity e2e tests at Bridge CI).
2. Put the tests to the Nearcore (it's how we work now).
3. Extract pytest to separate repo (it might be complicated and requires additional work).

bowenwang1996:
I see. I think the current arrangement is fine.

pytest/lib/bridge.py
```diff
@@ -1,14 +1,142 @@
...
+        args = ['node', 'index.js', 'TESTING'] + args
+        return self.js_call(args, timeout)
+
+    def js_call_service(self, args, stdout, stderr):
+        if not isinstance(args, list):
+            args = [args]
+        args = ['node', 'index.js', 'start'] + args + ['--daemon', 'false']
+        logger.info(' '.join(args))
+        service = subprocess.Popen(args, stdout=stdout, stderr=stderr, cwd=self.cli_dir)
+        # TODO ping and wait until service really starts
```
mfornet, line 86: This can be done using `pm2 show args[0]` (and parsing the output of that command).

pytest/lib/bridge.py
```diff
@@ -1,14 +1,142 @@
...
+
+
+class BridgeUser(object):
+
+    def __init__(self, name, eth_secret_key, near_sk, near_pk):
+        self.name = name
+        self.eth_secret_key = eth_secret_key
+        self.near_account_name = name + '.test0'
+        self.near_sk = near_sk
+        self.near_pk = near_pk
```
mfornet, line 98: We can reuse the class `Key` from: https://github.com/near/nearcore/blob/49048c733c08e2561db3c8b106c0248f32f479df/pytest/lib/key.py#L5

pytest/lib/bridge.py
```diff
@@ -1,14 +1,142 @@
...
+        args = ['node', 'index.js'] + args
+        return self.js_call(args, timeout)
+
+    def js_call_testing(self, args, timeout=3):
+        if not isinstance(args, list):
+            args = [args]
+        args = ['node', 'index.js', 'TESTING'] + args
+        return self.js_call(args, timeout)
+
+    def js_call_service(self, args, stdout, stderr):
```
mfornet, line 80:
```suggestion
    def js_start_service(self, args, stdout, stderr):
```

pytest/lib/bridge.py
```diff
@@ -94,15 +222,19 @@ def start(self):
                 '--port',
```
mfornet, line 222: Remove both TODO:

```
# TODO use blockTime
# TODO set params
```

pytest/lib/bridge.py
```diff
@@ -223,99 +344,163 @@ def __init__(self, config):
...
+        nonce = node.get_nonce_for_pk(node.signer_key.account_id, node.signer_key.pk)
+        create_account_tx = sign_create_account_with_full_access_key_and_balance_tx(
+                node.signer_key,
+                user.near_account_name,
+                user.near_signer_key,
+                billion_tokens // 10,
+                nonce + 1,
+                h)
+        res = node.send_tx_and_wait(create_account_tx, timeout=15)
+        assert_success(res['result']['status']['SuccessValue'])
```
mfornet, line 369: Is it the following scenario possible?

- You create a tx and send it.
- After 15 seconds it hasn't passed because of some glitch on the network
- The tx is still around and it eventually succeed
- Since it initially failed, the tx is re-issued, and it will fail now because the account was already created in the previous tx.

The way to address this, is checking first if the account exist before trying to issue the tx. Anyway, 15 seconds is probably enough time, such that the described scenario is very unlikely

pytest/lib/bridge.py
```diff
@@ -223,99 +344,163 @@ def __init__(self, config):
...
+            '--eth-client-lock-eth-amount',
+            '1000000000000000000',
+            '--eth-client-lock-duration',
+            '30'], timeout=60))
+        assert_success(self.adapter.js_call_cli('init-eth-prover', timeout=60))
+        assert_success(self.adapter.js_call_cli('init-eth-erc20', timeout=60))
+        assert_success(self.adapter.js_call_cli('init-eth-locker', timeout=60))
+
+    @retry(stop_max_attempt_number=10, wait_fixed=1000)
     def init_near_token_factory(self):
```
mfornet, line 399: Why do we need a `retry` on this method? what is the criteria to use `retry`? I see it in some of the methods but not in all that I expect. 

(tbh ideally they won't be used anywhere, but I guess interfaces are not robust enough)

mfornet:
You are missing the coolest feature from `Path` here: https://github.com/near/nearcore/pull/3658/files#diff-775a9f1990e68594074cb25791d9d5cc0c08ac783767390127fe66e6e01c17a7R334-R345

We can replace:

```
Path(os.path.join(os.path.join(logs_dir,service),'err.log')).touch()
```

with:

```
(Path(logs_dir) / service / 'err.log').touch()
```

pytest/lib/bridge.py
```diff
@@ -150,51 +277,45 @@ def start(self):
...
+    try:
+        message = message.decode('ascii')
+    except:
+        # message is already str
+        pass
+    if message == '':
+        # empty message is considered valid
+        return
+    message = message.strip().split('\n')[-1].strip().split(' ')
+    if not (message[0] in ('Deployed', 'Done') or message[-1] in ('deployed', 'initialized')):
```
mfornet, line 296: The function `assert_success` feels too hacky.

pytest/lib/bridge.py
```diff
@@ -150,51 +277,45 @@ def start(self):
...
         self.cli_dir = os.path.join(self.bridge_dir, 'cli')
         if not os.path.exists(self.bridge_dir):
-            self._git_clone_install()
+            self.git_clone_install()
         if not os.path.exists(os.path.expanduser("~/go")):
-            print('Go must be installed')
+            logger.error('Go must be installed')
             assert False
+        # TODO resolve
+        # This hack is for NayDuck
```
mfornet, line 318: Why is this hack required?

mfornet (commented):
Bridge infrastructure looks much better now, thanks for the refactoring @Kouprin

pytest/lib/bridge.py
Kouprin: `--daemon false` means we don't use pm2 layer (and we are going to get rid of pm2 completely, AFAIK).
Reading logs might be better solution. However, I'm not sure that it needs to be included into this PR.

pytest/lib/bridge.py
Kouprin: ++

pytest/lib/bridge.py
Kouprin: ++

pytest/lib/bridge.py
Kouprin: We still don't use custom params to start ganache. We also don't handle stderr/stdout for cases if we want to simulate Eth forks.
I will remove only `# TODO use blockTime` as we use `ganache_block_prod_time` param from config.

pytest/lib/bridge.py
Kouprin: You're right, it's about robustness. Here's the exact failure: https://nayduckdiag.blob.core.windows.net/logs/66198_v2_stderr.

pytest/lib/bridge.py
Kouprin: It might be possible. However, after bunch of executions at Nayduck I haven't seen this scenario happened.
Anyway, this is a good catch, and I'm going to replace it with
```
        res = node.send_tx(create_account_tx)
        assert 'result' in res
```

pytest/lib/cluster.py
```diff
@@ -313,7 +293,7 @@ def start_proxy_if_needed(self):
     def start(self, boot_key, boot_node_addr, skip_starting_proxy=False):
         if self._proxy_local_stopped is not None:
             while self._proxy_local_stopped.value != 2:
-                logging.debug(f'Waiting for previous proxy instance to close')
+                print(f'Waiting for previous proxy instance to close')
```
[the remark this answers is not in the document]
Kouprin: This is the only place asked for `logging` module. All other use `print`. I guess it's happened by mistake and I reduced entropy by removing `import logging` and replace `logging` with `print` here.

pytest/lib/cluster.py
```diff
@@ -891,7 +871,7 @@ def spin_up_node_and_push(i, boot_key, boot_addr):
 
     return ret
 
-def start_bridge(start_local_ethereum=True, handle_contracts=True, handle_relays=True, config=None):
+def start_bridge(nodes, start_local_ethereum=True, handle_contracts=True, handle_relays=True, config=None):
```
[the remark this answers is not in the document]
Kouprin: I'd love to delegate to bridge how to initialize itself by having a cluster. We may change in future the strategy inside of bridge lib and ask several nodes instead of `nodes[0]` to speed the things up and increase robustness.

pytest/lib/bridge.py
Kouprin: Hm, getting nonce works badly after this fix:
`nonce = node.get_nonce_for_pk(tx.sender.near_account_name, tx.sender.near_signer_key.pk)` returns None. I'm going to investigate.

pytest/lib/bridge.py
Kouprin: You are right. The output of different services is very different. It seems more friendly to have several 
```
assert_success(A) 
assert_success(B)
assert_success(C)
```
instead of
```
assert A.decode('ascii').strip().split('\n')[-1].strip().split(' ')[0] == 'Done'
assert B.decode('ascii').strip().split('\n')[-1].strip().split(' ')[-1] == 'deployed'
assert C == b''
```
So I expect it's better to delegate inner logic to such function and make it a little bit hacky for unification purposes.

pytest/lib/bridge.py
Kouprin: Previously it was `cp`. Pytest folder and Bridge folder appear on different filesystems at Nayduck and it's not allowed to `cp` files between them. Debugging this is extremely time-consumable because each run at Nayduck takes a lot of time and gives small knowledge for reverse-engineering.

I've found that running `wget` ignores the issue with filesystems. I suggest to patch this hack separately of this PR under @utka consulting, if it will be necessary for some reason.

pytest/lib/bridge.py
Kouprin: Let me try this one:
```
        res = node.send_tx(create_account_tx)['result']
        while True:
            try:
                created = node.get_tx(res, user.near_account_name)
                assert_success(created['result']['status']['SuccessValue'])
                break
            except:
                time.sleep(1)
```

Kouprin: Seems that last one works well. I'm surprised it's not enough to make sure that create account tx is just accepted by network.

pytest/lib/bridge.py
mfornet: I'll try to fix in the future if this become problematic.

pytest/lib/bridge.py
mfornet: Interesting. This is not a big problem, and actually we probably don't need to resolve anything here.

mfornet (commented):
@Kouprin Let's include a link to nayduck with all tests passing before merging.

I noticed on http://nayduck.eastus.cloudapp.azure.com:3000/#/run/928 that some tests are flaky yet.

Kouprin:
@mfornet I've done some additional investigation about failures:

1. `pytest bridge/same_block.py 1 2` - failed in the test code (https://nayduckdiag.blob.core.windows.net/logs/68205_v2_stderr), not the module:
```
  File "tests/bridge/same_block.py", line 46, in <module>
    status = nodes[0].get_status(check_storage=False)
```
It's not clear do we need to wrap all methods with `retry` in the code of each test. I double-checked with other pytests like `tests/sanity/block_sync.py` and there are no `retry` wrappers on `get_status`.

2. `pytest bridge/end2end.py 1 5` - failed in `init-near-token-factory` (https://nayduckdiag.blob.core.windows.net/logs/68250_v2_stderr) with the following:
```
[2020-12-21 17:50:48] INFO     31983  js_call            node index.js init-near-token-factory
[2020-12-21 17:50:59] ERROR    31983  js_call            js call failed, exit code 1, msg: b'Secret key for fungible token is not specified. Reusing master secret key.\nDeploying token contract.\nAccount neartokenfactory does not exist creating it.\nCreated account neartokenfactory\nDeployed contract to account neartokenfactory\nFailed to fetch state of the test0 account. Is it initialized?\n'
[2020-12-21 17:51:00] INFO     31983  js_call            node index.js init-near-token-factory
[2020-12-21 17:51:03] ERROR    31983  js_call            js call failed, exit code 1, msg: b'Secret key for fungible token is not specified. Reusing master secret key.\nDeploying token contract.\nFailed to initialize the token factory Error: Smart contract panicked: panicked at &#39;Already initialized&#39;, src&#x2F;lib.rs:116:9\n'
[2020-12-21 17:51:04] INFO     31983  js_call            node index.js init-near-token-factory
...
```
This is related to how `init-near-token-factory` works in the Rainbow Bridge code. Basically, this is a bug of the Bridge CLI. I guess we need to separate factory initialization from token initialization there.

3. `pytest bridge/two_users.py same_amount` - failed in checking balances (https://nayduckdiag.blob.core.windows.net/logs/68262_v2_stderr). This is the most suspicious failure I've ever seen. Please look into the logs carefully. This transaction sends 123 tokens from ETH to NEAR and 61 tokens back. Starting with 1e9 ERC-20 tokens, the final amount expected to be 999999938. However, actual amount is 999998938, which is 1000 less than expected.
The test constantly waits for valid balance until timeout:
```
[2020-12-21 17:56:55] INFO     25584  js_call            node index.js TESTING get-erc20-balance 6704Fbfcd5Ef766B287262fA2281C105d57246a6 erc20
[2020-12-21 17:56:56] INFO     25584  check_balances     alice's expected amount of erc20 token is 999999938, actual amount is 999998938
```
Is it Nayduck issue? There is nothing that even touches 1000 tokens in that test.

4. `pytest bridge/multiple_relays.py` - failed with timeout (https://nayduckdiag.blob.core.windows.net/logs/68286_v2_stderr). It might be issue of Testing Infra or some specific issue with running several Relays. Going to investigate deeper.

Also I saw several times that tests failing because final balances are not equal to expected with no particular reason. Reading the logs, I saw all subprocesses finished properly with 0 exit code which means all necessary steps for token transferring have been completed successfully. I've also removed using shared memory to avoid issues related to race conditions (cc @SkidanovAlex). (It might be related to p.3 above).

I suggest to merge this PR to accept most major changes and fix all minor issues on demand then. At current state it's hard to distinguish Testing Infrastructure bugs from Bridge bugs or even Nayduck bugs. I don't think we should expect that all tests must pass because their failures might be unrelated to the Testing Infrastructure itself.

mfornet (approved):
Ok, let's merge this, and I'll continue from here fixing the CI.

PR #3661, opened 2020-11-28
Implement time travel for NEAR testing infrastructure

pczarn:
Implements https://github.com/near/bounties/issues/31

Document all occurences of `Utc::now`

* Inspect whether they need to be replaced with a time proxy.

---

> confirm with us the correctness of each workitem before moving on to the next one

Please review and confirm the correctness of point one:

* [x] Inspect all the occurrences of Utc::now and documenting which ones need to be replaced with a proxy.
* [x] Design a time-proxy that will return Utc::now without the adversarial feature flag, but otherwise enables manipulating time via an RPC end point
* [x] Implement the proxy, add the meta-test described above.


pczarn:
Upon basic inspection, the occurences of `Instant::now` are all related to timing or testing.

chain/chunks/src/lib.rs
```diff
@@ -260,6 +260,7 @@ impl SealsManager {
                 };
 
                 let chosen = Self::get_random_part_ords(candidates);
+                // `sent` is not accessed, so we don't use the time proxy.
```
bowenwang1996, line 263: It is not used right now, but will be enabled after #3657 is merged

bowenwang1996 (commented):
where is the `time_proxy` module? Also I don't think we need to have a `Utc` submodule there

pczarn:
The `time_proxy` module is to-be-done as the second step. I believe I will invoke this stuff by doing `use time_proxy::UtcProxy`, then calling `UtcProxy::now`, what do you think?

Alas, I caught a cold (this is a cold season after all) so this task will need to wait a few days.

chain/chunks/src/lib.rs
pczarn: will fix this catch

pczarn:
I'm back to coding. I'll pick up this task shortly. Please excuse the delay.

pczarn:
I have the second sub-task compiling, and the last sub-task as a draft.

Could you think of more tests utilizing time travel? If we spend effort implementing the feature that enables testing, we might as well add more testing.

pczarn:
Without time travel, the test completes in 22 seconds. With time travel enabled, the test completes in 50 seconds half of the time, and gets stuck the other half of the time. The core simply does not work well with time changes.

SkidanovAlex:
Hi, @pczarn, I reviewed most of the code, and it appears to be doing exactly what is needed.

Several questions:
1. Why have the `proxify` map, as opposed to not having it, and assuming that all the locations are proxified when time travel is enabled? It seems to make using the time travel machinery in the tests harder.
2. Can you briefly describe basic use cases: as I understand it, it supports both jumps in time (by keeping rate at 0), and slowing time down / speeding it up. Can you confirm those are two primary use cases?
3. I can see one test that stops time, checks that nothing happens, and then make the time work normally again. Can you add tests for another use case as well: not stopping time, but instead slowing it down, say 3x, and making sure that on the same (relatively long) period of time 3x fewer blocks are produced?

@frol @matklad can you guys please review the instances of the `unsafe` code?

pczarn:
Hello, @SkidanovAlex, thanks for a thoughtful review.

1. It's there to assist the developer in case someone needs to add or change code that uses time in several places. With `proxify`, one can try out differences between using `system_time` and using `*Proxy` without recompiling `neard`. I agree that such apparatus is of dubious value, but I needed it when introducing many proxy uses, so I left it in. The scripts `run_time_travel` and `find_time_travel_setting` use `proxify` and I wonder whether , as I left them in as well. They helped me locate one proxy use that was misbehaving. So this all boils down to the question, should I keep things in the codebase that are likely of little value?
2. It supports jumps in absolute time value (`diff`/`const_diff`), as well as switches to another rate of passage of time (`rate`/`speed_diff`). Keep in mind that a switch from non-1 time passage rate to any rate will cause a jump in absolute time, as the previous period of slowdown / speedup henceforth counts as a normal period of passage of time. That's because all passage of time before the moment of the last proxy update is counted at rate 1.0. (I could change this behavior.) Also, you said "it supports jumps in time by keeping rate at 0", sure, but not necessarily just by keeping rate at 0, that is not a necessary condition for a jump -- I wanted to clarify so there is no misunderstanding how jumps can happen. Basically, jumps happen on most changes if it's not the first change to the proxy, or if the change touches absolute `diff`,
3. I will add that test. Can one rely on the approximate rate of block production?

frol (approved):
The unsafe code looks ok to me, but I would still like @matklad to look at it as well.

P.S. Ideally, we want to have ["Safety" comments](https://rust-lang.github.io/rust-clippy/rust-1.51.0/index.html#missing_safety_doc) next to the `unsafe` blocks

core/primitives/src/time.rs
```diff
@@ -0,0 +1,146 @@
...
+    }
+
+    fn set(value: TimeTravelSingleton) {
+        let singleton = SingletonReader { inner: Arc::new(Mutex::new(value)) };
+        unsafe {
+            SINGLETON = mem::transmute(Box::new(singleton));
+        }
+    }
+}
+
```
matklad, line 80: I don't think the unsafe is correct. `TimeTravelSingleton::set` is a public API of the crate, and calling it concurrently results in a race on the `SINGLETON` variable. 

Taking a step back, I don't think this *needs* manually written unsafe. I am a bit confused what's the intent here. It seems like `get` can overwrite a previous set, which feels like a bug.

If you need to initialize something once, once_cell crate will work (it's already among the deps).

If yon needs a gobally swappable singleton, take a look at the `arc_swap` crate.

pczarn: You are right, there's a race. I could address these issues, but I will introduce `arc_swap` in this PR as you suggested because it will eliminate the need for unsafe code.

here's how I'd address the race by changing `SINGLETON`:
```
static mut SINGLETON: AtomicPtr<SingletonReader> = AtomicPtr::new(0 as *const SingletonReader);
```

core/primitives/src/time.rs
```diff
@@ -0,0 +1,146 @@
...
+        }
+    }
+
+    fn set(value: TimeTravelSingleton) {
+        let singleton = SingletonReader { inner: Arc::new(Mutex::new(value)) };
+        unsafe {
+            SINGLETON = mem::transmute(Box::new(singleton));
+        }
+    }
+}
```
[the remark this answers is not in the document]
pczarn: This code can be called from the closure within `get`, so we have two use cases
1. this is the first time the assignment happens, and we must replace the static singleton as there is no inner value yet
2. this is not our first store
I can see how we could improve on the code so that it does not leak memory in the second case. I removed this code in favor of `arc-swap`, though, so the concerns no longer apply.

for the record, the code here is taken from: https://stackoverflow.com/questions/27791532/how-do-i-create-a-global-mutable-singleton

core/primitives/src/time.rs
```diff
@@ -0,0 +1,146 @@
...
+
+impl SingletonReader {
+    fn get() -> SingletonReader {
+        unsafe {
+            ONCE.call_once(|| {
+                SingletonReader::set(TimeTravelSingleton::default());
+            });
+
+            (*SINGLETON).clone()
+        }
```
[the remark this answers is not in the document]
pczarn: ok, but I changed the code so the concerns no longer apply

core/primitives/src/time.rs
```diff
@@ -0,0 +1,146 @@
...
+    }
+
+    fn set(value: TimeTravelSingleton) {
+        let singleton = SingletonReader { inner: Arc::new(Mutex::new(value)) };
+        unsafe {
+            SINGLETON = mem::transmute(Box::new(singleton));
+        }
```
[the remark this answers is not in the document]
pczarn: ok, but I changed the code so the concerns no longer apply

core/primitives/src/time.rs
```diff
@@ -0,0 +1,146 @@
...
+    }
+}
+
+impl UtcProxy {
+    pub fn now(file: &'static str, line: u32) -> chrono::DateTime<chrono::Utc> {
+        let now = chrono::Utc::now();
+        let singleton = SingletonReader::get();
+        let time_travel = singleton.inner.lock().expect("poisoned mutex");
```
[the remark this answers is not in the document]
pczarn: ok, but I changed the code so these concerns no longer apply

pczarn:
I added 3 more commits that address the concerns that you brought up. @frol @matklad @SkidanovAlex 

* used `arc-swap` for time primitives
* added slowdown test (`time_slowdown.py`) that does not behave exactly as I would expect, but still, it slows down by 50% for passage of time at 30% speed. The test accepts one command line argument (rate) so you can play with the passage of time.
* minor fixes

SkidanovAlex:
Hi, @pczarn, the changes look good to me.
Let's remove the `proxify` map, I think it overly complicates the machinery (and make the time machine always affect all the locations when enabled).

pczarn:
done in the latest 2 commits @SkidanovAlex

core/primitives/src/time.rs
matklad: Yup, the new version looks much better :+1: !

pczarn:
@SkidanovAlex please schedule some time for a review

ailisp (approved):

pytest/tests/adversarial/time_travel.py
```diff
@@ -0,0 +1,164 @@
...
+            print(f'approval {msg.Routed.body.BlockApproval.target_height}  {fr}->{to}')
+        if msg.enum == 'Routed' and msg.Routed.body.enum.startswith('PartialEncodedChunk'):
+            num_chunks_sent.value += 1
+        if msg.enum == 'Block':
+            h = msg.Block.BlockV2.header.BlockHeaderV2.inner_lite.height
+            num_blocks_sent.value += 1
+            print(f'block {h}  {fr}->{to}')
+        return True
+
+# consensus_config = {"consensus": {"min_num_peers": 0}}
```
bowenwang1996, line 45: should this be removed?

bowenwang1996 (approved):

SkidanovAlex:
The code looks good!
Thanks @pczarn!

@bowenwang1996 let's not merge until I figure out the issue with cluster.py.

runtime/runtime-params-estimator/src/testbed_runners.rs
```diff
@@ -154,7 +154,7 @@ fn end_count_instructions() -> u64 {
 }
 
 fn start_count_time() -> Consumed {
-    Consumed::Instant(Instant::now())
+    Consumed::Instant(Instant::system_time())
```
olonho, line 157: What's the point of this change? We do not need system time here, we need a way to count monotonic ticks between two moments. 
`Instant::now` does exactly what we need, i.e. 
```
A measurement of a monotonically nondecreasing clock. Opaque and useful only with Duration.
```

While  description of SystemTime 
```
A measurement of the system clock, useful for talking to external entities like the file system or other processes.
```
is not behavior we're looking for here.

olonho (changes requested):
We need to use monotonic clock when measuring system elapsed time in estimator.

runtime/runtime-params-estimator/src/testbed_runners.rs
pczarn: @olonho does my comment resolve your doubts?

pczarn:
@SkidanovAlex what's the status of this PR?

bowenwang1996:
Hey @pczarn, due to the fact that this PR is outdated, we decided to implement it from scratch instead of rebasing this PR on master. See (https://github.com/near/nearcore/pull/5123) for details. Did you get the payment?

pczarn:
@bowenwang1996 yes I did receive the payment for this PR. This PR is a good chance for me to learn since I can look into somebody's reimplementation.

pmnoxx:
@pczarn I think that overall this is a good PR.However, in order to ship it safely a few steps would have been required.

In the ideal scenario we could have done it like this:

- (introduce one source of time): We have multiple ways to get time. `SystemTime`, `chrono:Utc`, `Instant`.
The instant has `Instant::elapsed()` method, which can get current time by itself, so replacing Instant with a wrapper is a must.
However, having multiple types of wrappers in the code base for multiple types introduces unnecessary complexity.
The best approach would be to start with introducing new wrapper for that, that can be used everywhere, take a look at https://github.com/near/nearcore/pull/5177.

- Then do the replacement step by step, ideally separately for each crate or group of crates. Like (chain/network ; chain/json* / chain/client; core/) - Ideally up to 10 files a a time, to make it easy to review

- Add mocking, just like you did, and replace `now()`, `elapsed()` methods. 

- Also rust runs tests in parallel, so we have to handle it somehow. We decided on propagating `TimeFactory` throughout the code. And that will be a big change as well.

pmnoxx:
Just the first 2 steps are done currently.

bowenwang1996:
> @bowenwang1996 yes I did receive the payment for this PR. This PR is a good chance for me to learn since I can look into somebody's reimplementation. I can already see some things that could have been better in this PR.

Thanks for your contribution! If you see places for improvements, feel free to submit new PRs.

PR #3689, opened 2020-12-03
fix(chunks): Validate partial encoded chunk header signatures exactly once

### LINKED ISSUE #3683: Redundant chunk header signature check
There are multiple ways in which we can receive parts of encoded chunks:

* `PartialEncodedChunk`
* `PartialEncodedChunkResponse`
* `PartialEncodedChunkForward`

The latter two do not come with a full chunk header, and instead the chunk header is looked up from a cache before passing through the same logic as `PartialEncodedChunk`. This logic begins with a signature verification of the header, which is clearly redundant if it is the same header which is pulled from a cache (and was previously validated before being put there).

Instead we should verify the header signature exactly once (in the `PartialEncodedChunk` pathway), then not do it when working with `PartialEncodedChunkResponse` and `PartialEncodedChunkForward`.

birchmd:
This PR fixes an issue where the node is doing redundant (and expensive) validation work. This overhead is especially an issue when chunk forwarding is enabled.

Fixes #3683

Testing plan:

* Existing tests

### DIFF
--- a/chain/chunks/src/lib.rs
+++ b/chain/chunks/src/lib.rs
@@ -1117,7 +1117,7 @@ impl ShardsManager {
         // We must check the protocol version every time, since a new value
         // could be passed to the function, whereas the signature check is intrinsic
         // to the header, thus only needs to happen exactly once.
-        let partial_encoded_chunk = partial_encoded_chunk.unwrap();
+        let partial_encoded_chunk = partial_encoded_chunk.extract();
         if !partial_encoded_chunk.header.version_range().contains(protocol_version) {
             return Err(Error::InvalidChunkHeader);
         }
--- a/chain/chunks/src/lib.rs
+++ b/chain/chunks/src/lib.rs
@@ -35,6 +35,7 @@ use near_primitives::types::{
     AccountId, Balance, BlockHeight, BlockHeightDelta, Gas, MerkleHash, ShardId, StateRoot,
     ValidatorStake,
 };
+use near_primitives::utils::MaybeValidated;
 use near_primitives::validator_signer::ValidatorSigner;
 use near_primitives::version::ProtocolVersion;
 use near_primitives::{checked_feature, unwrap_or_return};
@@ -1082,21 +1083,17 @@ impl ShardsManager {
 
     pub fn process_partial_encoded_chunk(
         &mut self,
-        partial_encoded_chunk: PartialEncodedChunkV2,
+        partial_encoded_chunk: MaybeValidated<PartialEncodedChunkV2>,
         chain_store: &mut ChainStore,
         rs: &mut ReedSolomonWrapper,
         protocol_version: ProtocolVersion,
     ) -> Result<ProcessPartialEncodedChunkResult, Error> {
         // Check validity first
 
-        if !partial_encoded_chunk.header.version_range().contains(protocol_version) {
-            return Err(Error::InvalidChunkHeader);
-        }
-        let header = partial_encoded_chunk.header.clone();
-        let chunk_hash = header.chunk_hash();
-
-        // 1. Checking signature validity
-        match self.runtime_adapter.verify_chunk_header_signature(&header) {
+        // 1. Checking signature validity (if needed)
+        let signature_check = partial_encoded_chunk
+            .validate_with(|pec| self.runtime_adapter.verify_chunk_header_signature(&pec.header));
+        match signature_check {
             Ok(false) => {
                 byzantine_assert!(false);
                 return Err(Error::InvalidChunkSignature);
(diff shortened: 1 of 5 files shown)


chain/client/src/client.rs
```diff
@@ -1053,9 +1056,9 @@ impl Client {
         let mut partial_encoded_chunks =
             self.shards_mgr.get_stored_partial_encoded_chunks(next_height);
         for (_shard_id, partial_encoded_chunk) in partial_encoded_chunks.drain() {
-            if let Ok(accepted_blocks) =
-                self.process_partial_encoded_chunk(PartialEncodedChunk::V2(partial_encoded_chunk))
-            {
+            let chunk =
+                MaybeValidated::NotValidated(PartialEncodedChunk::V2(partial_encoded_chunk));
```
bowenwang1996, line 1060: why is this not validated?

core/primitives/src/utils.rs
```diff
@@ -16,10 +16,51 @@ use crate::version::{
...
+    }
+
+    pub fn map<U, F: FnOnce(T) -> U>(self, f: F) -> MaybeValidated<U> {
+        match self {
+            Self::Validated(t) => MaybeValidated::Validated(f(t)),
+            Self::NotValidated(t) => MaybeValidated::NotValidated(f(t)),
+        }
+    }
+
+    pub fn unwrap(self) -> T {
```
bowenwang1996, line 45: Can we name it differently? When reading the code it is very easy to think that `unwrap` implies that it may panic. Maybe `unbox`?

chain/chunks/src/lib.rs
```diff
@@ -1117,6 +1114,16 @@ impl ShardsManager {
             }
         };
 
+        // We must check the protocol version every time, since a new value
+        // could be passed to the function, whereas the signature check is intrinsic
+        // to the header, thus only needs to happen exactly once.
+        let partial_encoded_chunk = partial_encoded_chunk.unwrap();
+        if !partial_encoded_chunk.header.version_range().contains(protocol_version) {
```
bowenwang1996, line 1121: actually what is the purpose of this check? Does this mean that version v1 cannot be processed if you pass a version that is higher than `SHARD_CHUNK_HEADER_UPGRADE_VERSION`?

chain/client/src/client.rs
```diff
@@ -795,8 +799,9 @@ impl Client {
                     return Err(Error::Other("Invalid chunk version".to_string()));
                 };
 
+                let pec = PartialEncodedChunk::clone(&partial_encoded_chunk);
```
bowenwang1996, line 802: not related to this PR per se, but I wonder whether we should clone here. It seems that there are often cases where `process_partial_encoded_chunk` returns early because we already know this partial encoded chunk and therefore there is no need to take ownership here. Maybe passing it as a reference to `ShardsManager::process_partial_encoded_chunk` would be better

chain/client/src/client.rs
birchmd: My understanding is that this covers the case where we were missing some chunks for the block at height `h`, so could not accept it. But in the mean time we have received the block and some chunks for height `h+1`, which will need to be processed once we finish with height `h`. And it looked to me like the signature check does not happen for the `h+1` because they point at a block we haven't added yet. Therefore these should be labelled as `NotValidated`.

chain/chunks/src/lib.rs
birchmd: This check prevents accepting a chunk which uses an old version after the network has collectively decided to upgrade. It should not prevent accepting an old chunk for an old block because you would pass the function the protocol version associated with that block's epoch.

It's also not just about v1 vs v2, we may introduce new versions in the future, and again need to make sure each version is only accepted during the epochs where they were valid.

chain/client/src/client.rs
birchmd: Yes, I agree this clone seems wasteful. Do you want me to make that change here or file an issue and do it in a separate PR?

bowenwang1996: Sure let's file it as a separate issue

chain/client/src/client.rs
bowenwang1996: I see

bowenwang1996 (approved):

core/primitives/src/utils.rs
birchmd: I normally think of something that is boxed as being heap-allocated, but that is not the case here. How about `extract` instead?

chain/client/src/client.rs
birchmd: https://github.com/near/nearcore/issues/3700

core/primitives/src/utils.rs
bowenwang1996: Sure that works

birchmd: Done in https://github.com/near/nearcore/pull/3689/commits/311d7b3c21355396270f496fdb5f8f43e8f49d41

PR #3704, opened 2020-12-05
Fixes in ecrecover, parse_meta_tx, eip-712 and enable meta tx

ailisp:
The construction of eip-712 encoding in ecrecover test was wrong, cause ecrecover test failed. This pr fixed it. 
- ensure correctness of ecrecover by adding more test, compare with metamask generate sig.
- add some eip-712 utilities used in tests. 
- fix implement of parse_meta_tx:
  - reproducing eip-712 hash from give meta_tx args by parsing method definition and parsing RLP encoded method call args, eip-712 parsed args into parsed method&type structure, then verify the signature
  - Re-encode decoded RLP meta tx args to ETH-ABI encoded, so that near evm runner can execute it
- enable meta txn, add a test case showing it verifies signature, execute the meta txn and gives same result as a normal txn. (meta txn charging is not yet implemented, we just parse_meta_tx, get the fee token address/amount and ignore them)

Test Plan
------------
CI and `cargo test --package near-evm-runner --features protocol_feature_evm,nightly_protocol `


ailisp:
@evgenykuzyakov please review this one, so many conflicts so i recreate pr from master

evgenykuzyakov (commented):
- [ ] Please move meta call parsing methods out `utils.rs` to a separate module.
- [x] Add comments for each parsing method and describe the expected input format maybe with the links to the standard.

ailisp:
@ilblackdragon Eugeny is going on vacation. This is what i proposed second approach in slack:
- [x] user provided rlp encoded function args as input, we decode it into nested Vec<u8> arrays
- [x] parse method_name and eip-712 nested Vec<u8> given the parsed method_name
- [ ] re-encode rlp encoded function args with eth-abi, should be a big `match` based on each arg's type, to evm and pass to evm runner as `input`.

artob:
@ailisp To facilitate the review and merge, could you rebase this PR against the [evm-phase2](https://github.com/near/nearcore/tree/evm-phase2) branch, please?

ailisp:
@artob will rebase after i pushed last todo, pull evm_phase2 and resolve conflict

ailisp:
@artob ready for review, (although haven't fully addressed @evgenykuzyakov 's comments)

artob (commented):
> @artob ready for review, (although haven't fully addressed @evgenykuzyakov 's comments)

Thanks, I will endeavor to review today.

runtime/near-evm-runner/tests/standard_ops.rs
```diff
@@ -258,52 +258,89 @@ fn test_solidity_accurate_storage_on_selfdestruct() {
...
-        U256::from(0),
-        U256::from(0),
+        U256::from(14),
+        U256::from(6),
         Address::from_slice(&[0u8; 20]),
-        test_addr,
-        "returnSomeFunds()",
-        vec![],
+        test_addr.clone(),
+        "adopt(uint256 petId)",
```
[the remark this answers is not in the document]
ailisp: Thanks! Corrected in https://github.com/near/nearcore/pull/3704/commits/773d3cc678017d18baff4ae2322227eb0b31737b

runtime/near-evm-runner/tests/standard_ops.rs
```diff
@@ -131,6 +131,7 @@ fn test_precompiles() {
 
     let mut context = create_context(&mut fake_external, &vm_config, &fees_config, accounts(1), 0);
     let (input, _) = soltest::functions::precompile_test::call();
+    println!("input: {:?}", hex::encode(&input));
```
artob, line 134: Perhaps this `println()` could be commented out by default? The tests should be silent by default.

runtime/near-evm-runner/src/utils.rs
```diff
@@ -183,23 +187,428 @@ pub fn format_log(topics: Vec<H256>, data: &[u8]) -> std::result::Result<Vec<u8>
...
+    c.is_ascii_alphabetic() || c == '_'
+}
+
+fn is_arg_char(c: char) -> bool {
+    c.is_ascii_alphanumeric() || c == '_'
+}
+
+/// Return a signature of the method_name
+/// E.g. method_signature(Methods before parse: "adopt(uint256 petId)") -> "adopt(uint256)"
+fn methods_signature(method_and_type: &MethodAndTypes) -> String {
```
artob, line 447: It's not clear to me why we need this function. The [ABI](https://docs.soliditylang.org/en/develop/abi-spec.html) suffices to specify how EVM function selectors look.

Why do we accept non-standard function signatures? Where are they coming from?

I followed the call chain far enough to assume they will be externally provided as an argument to the `meta_call` method. Is that correct? Who will be calling this method?

runtime/near-evm-runner/src/utils.rs
```diff
@@ -315,3 +735,36 @@ pub fn ecrecover_address(hash: &RawHash, signature: &[u8; 65]) -> Address {
...
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_parse_method() {
+        let methods =
+            MethodAndTypes::parse("Blah(Foo q,Bar w)Foo(uint256 x,Bar z)Bar(T ttt)T(string x)")
```
artob, line 746: Maybe I'm misunderstanding what this is for, but it seems incorrectly named: it's not a method type, it's a struct type, right?

And if it's a struct type, then [EIP-712](https://eips.ethereum.org/EIPS/eip-712) mandates alphabetical ordering for referenced types (`Bar` before `Foo`):

> If the struct type references other struct types (and these in turn reference even more struct types), then the set of referenced struct types is collected, sorted by name and appended to the encoding. An example encoding is `Transaction(Person from,Person to,Asset tx)Asset(address token,uint256 amount)Person(address wallet,string name).`

artob:
@ailisp My understanding of this pull is still a work-in-progress. I added some review comments, and will continue the review tomorrow.

artob (commented):
@ailisp My understanding of this pull is still a work-in-progress. I added some review comments, and will continue the review tomorrow.

runtime/near-evm-runner/tests/standard_ops.rs
ailisp: Yep

runtime/near-evm-runner/src/utils.rs
ailisp: We need `adopt(uint256 petId)` format in doing eip-712 hash. It needs to first keccak("adopt(uint256 petId)") then pack petId as uint256 after it.

runtime/near-evm-runner/src/utils.rs
ailisp: Yes method_name is not very accurate, I just inherit it from Illia's implementation

runtime/near-evm-runner/src/utils.rs
ailisp: @artob here is correct, i got from near-pet-shop was a old version where method name is given as "adopt(uint256 petId)" in message, that's why i needed to keccak("adopt(uint256 petId)"), right now our message is "adopt(uint256)" and thus these functions are unnecessary

runtime/near-evm-runner/src/utils.rs
ailisp: Name it method_signature now

runtime/near-evm-runner/src/utils.rs
ailisp: Deleted

ailisp: @artob I just found this cannot be replaced. this:
```
    let types = "NearTx(string evmId,uint256 nonce,uint256 feeAmount,address feeAddress,address contractAddress,string contractMethod,Arguments arguments)".to_string() + &arguments;
    bytes.extend_from_slice(&keccak(types.as_bytes()).as_bytes());

```
must have argument like **evmId** and also argument names in arguments" otherwise generated eip-712 hash doesn't match what metamask generated. For example I tried passing 
```
NearTx(string,uint256,uint256,address,address,string,Arguments)Arguments(uint256) and generated eip-712 hash doesn't match metamask's. You can run this to get metamask generated eip-712 hash:
https://gist.github.com/ailisp/0f65bc1a3827b03141ae6601918a67a2

I consult eip-712 spec, seems metamask is correct here, see:
![image](https://user-images.githubusercontent.com/13259400/102568225-e61f9180-4097-11eb-8aab-eefa5d3e8d01.png)

So, we should still pass method with method name as message (adopt(uint256 petId)) to correctly reproduce the signature, instead of adopt(uint256)

File: runtime/near-evm-runner/src/utils.rs (hunk 4 above)
Comment (ailisp) [reply runtime/near-evm-runner/src/utils.rs:746]: Still trying to think of a better name for it. Thought it can be method_signature. But in functional programming world, signature is something like "adopt(uint)" (without arg name), see above comment eip-712 requires adopt come with arg name, therefore method_signature is inappropriate

Comment (ailisp) [reply runtime/near-evm-runner/src/utils.rs:754]: I name it "method_def" now, to explictly differentiate with method_sig

File: runtime/near-evm-runner/src/utils.rs (hunk 3 above)
Comment (ailisp) [reply runtime/near-evm-runner/src/utils.rs:443]: @artob in these two commits:
https://github.com/near/nearcore/pull/3704/commits/90a6518603d61bcb7559a3153648d189647baf43
https://github.com/near/nearcore/pull/3704/commits/2c2b80d7ec6c7ca919e6519a27c29b6c7d7c3804
i link to spec and explain every use of method_sig vs method_def. Hope this can resolve the confusion.

File: runtime/near-evm-runner/src/runner.rs [hunk 5]
```diff
@@ -264,6 +264,7 @@ impl<'a> EvmContext<'a> {
             &self.fees_config.evm_config,
             self.chain_id,
         )?;
+        println!("{:?}", result);
```
artob, line 267: @ailisp Is this `println()` intentional?

runtime/near-evm-runner/src/utils.rs
artob: @ailisp There is still a significant confusion here. I think you are being misled by EIP-712's structure signature format (`NearTx(...)`) resembling the Solidity function selector signature format (`adopt(uint256)`) . However, there is actually no connection between the two, even if they do superficially resemble each other.

EIP-712 is about structure types, the ABI is about function calls. The structure signature includes the structure's property names; the function signature does _not_ include the function's parameter names, only their types. Does that make sense?

runtime/near-evm-runner/src/utils.rs
artob: @ailisp Please see my other comment just now. EIP-712 is about structure types, the ABI is about function calls. The structure signature includes the structure's property names; the function signature does _not_ include the function's parameter names, only their types.

An ABI function signature `adopt(uint256)` has nothing to do with EIP-712: it describes a function to call, not a structure type. Conversely, EIP-712 has nothing to do with function or method signatures such as `adopt()`. Hence why the naming here is problematic, and renaming it to `method_def` or `method_sig` will not fix that problem.

runtime/near-evm-runner/src/utils.rs
artob: > We need adopt(uint256 petId) format in doing eip-712 hash. It needs to first keccak("adopt(uint256 petId)") then pack petId as uint256 after it.

`adopt` is a function (or method, if you like) name, not a structure name. So, EIP-712 is not relevant to it. As per the Solidity ABI spec, to call a function `adopt(uint256 petId)` in the EVM, you take `keccak("adopt(uint256)")` (_without_ the parameter name!) and then append the argument to that. EIP-712 is about structures, not functions/methods.

ailisp: Yes, but, we need to pass this format:
> include the function's parameter names, append any type signatures for its non-atomic argument types.

in message, because this is the only place we have function types and type definitions that is later used in parse_meta_call to reproduce signature.

runtime/near-evm-runner/src/utils.rs
ailisp: Your comments is mostly correct, but this is not:

> An ABI function signature adopt(uint256) has nothing to do with EIP-712: it describes a function to call, not a structure type. 

It's not because we need to eip-712 adopt(uint256), but because we need to eip-712 this:
```
NearTx(...)Arguments(uint256 petId)
```
And ^ `uint256 petId` is taken from `adopt(uint256)`

runtime/near-evm-runner/src/runner.rs
ailisp: oops, will remove it

ailisp:
Somehow i didn't reply this one:
> And if it's a struct type, then EIP-712 mandates alphabetical ordering for referenced types (Bar before Foo):
> If the struct type references other struct types (and these in turn reference even more struct types), then the set of referenced struct types is collected, sorted by name and appended to the encoding. An example encoding is Transaction(Person from,Person to,Asset tx)Asset(address token,uint256 amount)Person(address wallet,string name).

`Transaction(Person from,Person to,Asset tx)Asset(address token,uint256 amount)Person(address wallet,string name)` is provided client, client code supposed to pass sorted args. rust side it can process either sorted and not sorted.
Assume: 1) If client doesn't eip-712, pass unsorted and use the same unsorted string to `Person(address wallet,string name)Asset(address token,uint256 amount)` generate the sig,  then we in rust side use the same string to keccak and generate same sig as client;
2) if client fully follows eip-712, pass sorted `...Asset(address token,uint256 amount)Person(address wallet,string name)` and was using this string to generate the sig, still we in rust side use same string to keccak and generate same sig as client.

So this is harmless

runtime/near-evm-runner/src/utils.rs
artob: OK, thanks for that. I will read the code again with this in mind :+1:

runtime/near-evm-runner/src/utils.rs
artob: OK, after your comments I believe I understand what we're doing. Thanks, @ailisp.

artob (approved):
LGTM :+1: overall after extensive review. Will endeavor to test this tomorrow prior to the holidays.

PR #3710, opened 2020-12-07
fix(chunks): Do not apply a block until we have all its missing chunks

### LINKED ISSUE #3684: Avoid repeatedly applying a block before all the chunk parts are available
Today when we receive a block X, if it is orphaned for some reason, and we keep receiving its chunk parts before its prev block gets accepted, when the prev block of X is finally accepted, it would trigger the processing of the chunk parts for X and each time all the chunk parts needed for a shard are successfully processed, it would in turn trigger another processing of X, even if we know that X cannot be applied until all the necessary parts for all shards have been processed. Such repetition could lead to a delay in block processing and the delays can cascade, which makes the node fall out of sync quickly. To address this, we could check whether all the chunk parts for all shards are already accepted before triggering the block processing.

birchmd:
Previously the logic was to try applying all blocks which were missing chunks each time we added a new chunk, even if there were still more chunks missing. With this change we now keep track of what chunks each block is missing and only apply the block again after we have all the missing chunks.

Fixes #3684 

Testing plan:

* new tests for the new data structure
* existing unit tests

### DIFF
--- a/chain/chain/src/missing_chunks.rs
+++ b/chain/chain/src/missing_chunks.rs
@@ -1,0 +1,292 @@
+use log::warn;
+use near_primitives::hash::CryptoHash;
+use near_primitives::sharding::ChunkHash;
+use near_primitives::types::BlockHeight;
+use std::cmp::Ordering;
+use std::collections::{
+    btree_map::{self, BTreeMap},
+    hash_map::{self, HashMap},
+    BinaryHeap, HashSet,
+};
+
+type BlockHash = CryptoHash;
+
+const MAX_BLOCKS_MISSING_CHUNKS: usize = 1024;
+
+pub trait BlockLike {
+    fn hash(&self) -> BlockHash;
+    fn height(&self) -> BlockHeight;
+}
+
+#[derive(Debug)]
+struct HeightOrdered<T>(T);
+
+impl<T: BlockLike> PartialEq for HeightOrdered<T> {
+    fn eq(&self, other: &Self) -> bool {
+        self.0.height() == other.0.height()
+    }
+}
+impl<T: BlockLike> Eq for HeightOrdered<T> {}
+impl<T: BlockLike> PartialOrd for HeightOrdered<T> {
+    fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
+        Some(self.0.height().cmp(&other.0.height()))
+    }
+}
+impl<T: BlockLike> Ord for HeightOrdered<T> {
+    fn cmp(&self, other: &Self) -> Ordering {
+        self.0.height().cmp(&other.0.height())
+    }
+}
+
+/// Structure for keeping track of missing chunks.
+/// The reason to have a Block type parameter instead of using the
+/// `block::Block` type is to make testing easier (`block::Block` is a complex structure and I
+/// don't care about most of it).
+#[derive(Debug, Default)]
+pub struct MissingChunksPool<Block: BlockLike> {
+    missing_chunks: HashMap<ChunkHash, HashSet<BlockHash>>,
+    blocks_missing_chunks: HashMap<BlockHash, HashSet<ChunkHash>>,
+    blocks_waiting_for_chunks: HashMap<BlockHash, Block>,
+    blocks_ready_to_process: BinaryHeap<HeightOrdered<Block>>,
+    height_idx: BTreeMap<BlockHeight, HashSet<BlockHash>>,
+}
+
+impl<Block: BlockLike> MissingChunksPool<Block> {
+    pub fn new() -> Self {
+        Self {
+            missing_chunks: Default::default(),
+            blocks_missing_chunks: Default::default(),
+            blocks_waiting_for_chunks: Default::default(),
+            blocks_ready_to_process: BinaryHeap::new(),
+            height_idx: Default::default(),
+        }
+    }
+
+    pub fn contains(&self, block_hash: &BlockHash) -> bool {
+        self.blocks_waiting_for_chunks.contains_key(block_hash)
+    }
+
+    pub fn len(&self) -> usize {
+        self.blocks_waiting_for_chunks.len()
+    }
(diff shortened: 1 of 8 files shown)


chain/chain/src/missing_chunks.rs
```diff
@@ -0,0 +1,292 @@
...
+                        self.mark_block_as_ready(&block_hash);
+                    }
+                }
+                hash_map::Entry::Vacant(_) => {
+                    warn!(target: "chunks", "Invalid MissingChunksPool state. Block with hash {} was still a value of the missing_chunks map, but not present in the blocks_missing_chunks map", block_hash);
+                    self.mark_block_as_ready(&block_hash);
+                }
+            }
+        }
+        // Chunks are accepted in order of height (because they rely on knowing the prev_block),
```
bowenwang1996, line 141: The reasoning in the parenthesis does not imply that chunks are accepted in order of height. Forks do exist

chain/client/src/client.rs
```diff
@@ -804,9 +806,11 @@ impl Client {
 
                 match process_result {
                     ProcessPartialEncodedChunkResult::Known => Ok(vec![]),
-                    ProcessPartialEncodedChunkResult::HaveAllPartsAndReceipts(prev_block_hash) => {
-                        Ok(self
-                            .process_blocks_with_missing_chunks(prev_block_hash, protocol_version))
+                    ProcessPartialEncodedChunkResult::HaveAllPartsAndReceipts(_) => {
+                        self.chain
+                            .blocks_with_missing_chunks
+                            .accept_chunk(&chunk_hash, height_included);
```
bowenwang1996, line 812: this doesn't work because `height_included` is not yet set at this point (it is 0).

chain/chain/src/missing_chunks.rs
```diff
@@ -0,0 +1,292 @@
...
+pub struct MissingChunksPool<Block: BlockLike> {
+    missing_chunks: HashMap<ChunkHash, HashSet<BlockHash>>,
+    blocks_missing_chunks: HashMap<BlockHash, HashSet<ChunkHash>>,
+    blocks_waiting_for_chunks: HashMap<BlockHash, Block>,
+    blocks_ready_to_process: BinaryHeap<HeightOrdered<Block>>,
+    height_idx: BTreeMap<BlockHeight, HashSet<BlockHash>>,
+}
+
+impl<Block: BlockLike> MissingChunksPool<Block> {
+    pub fn new() -> Self {
```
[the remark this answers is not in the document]
birchmd: Yes, I originally was deriving `Default`, but when I switched to using a type parameter it meant that the type you use also needs to have a `Default` implementation (which is dumb because I don't need any instances of `Block` to initialize the data structure). Since there is no `Default` implementation for `block::Block` I needed to put in the `new` boilerplate.

chain/chain/src/missing_chunks.rs
birchmd: The only purpose of this is to prune old blocks from the data structure (i.e. prevent forever unavailable chunks from blowing this up). I'll re-think how to do this in a safe way since your other comment about `height_included` is about this same piece of logic.

The consensus protocol has finalization, I know that. What if instead I pass in the last finalized height here instead of the last added block? I think I could do that at the end of `on_block_accepted` for example.

bowenwang1996: An easy solution is to prune heights below `Chain::tail()` when some block is accepted.

birchmd: I changed the pruning to use `Chain::tail` in https://github.com/near/nearcore/pull/3710/commits/54c6b603d2a0452aee8a7e16f14e7d44e2068526

I made it happen in `receive_block` instead of on block acceptance because we already had a call to `tail` there, so I could use that result again. Of course it could mean we are pruning the `MissingChunksPool` more often than necessary, but I thought that is ok because pruning on an already pruned pool should be a no-op that takes essentially no time. Whereas calling `tail` looks like a DB read and thus could be more expensive to repeat that in some other part of the code.

But let me know what you think.

chain/client/src/client_actor.rs
```diff
@@ -961,6 +961,7 @@ impl ClientActor {
             debug!(target: "client", "dropping block {} that is too far behind. Block height {} current tail height {}", block.hash(), block.header().height(), tail);
             return;
         }
+        self.client.chain.blocks_with_missing_chunks.prune_blocks_below_height(tail);
```
bowenwang1996, line 964: this is not the best place to put it since invalid blocks can also trigger this. I suggest moving this to `on_block_accepted`

bowenwang1996 (commented):
One small issue with pruning using tail is that it is no-op for archival nodes, so we might want to consider switching to something like `head - 2 * epoch_length` instead

chain/client/src/client_actor.rs
birchmd: Done in https://github.com/near/nearcore/pull/3710/commits/18842d1ee8ff975525c3ae883988d1417da5fde4

chain/client/src/client.rs
```diff
@@ -926,7 +926,10 @@ impl Client {
         }
 
         if status.is_new_head() {
-            self.shards_mgr.update_largest_seen_height(block.header().height());
+            let height = block.header().height();
+            self.shards_mgr.update_largest_seen_height(height);
+            let prune_height = height - 2 * self.config.epoch_length;
```
bowenwang1996, line 931: I didn't mean this literally. For one, you need to use at least `saturating_sub` to make sure we don't underflow. Also it is better to extract the constant out

birchmd: Per our offline discussion, I changed to last finalized block height in https://github.com/near/nearcore/pull/3710/commits/95cc9e71fb27f7bdbae794838c88abcb40515d8c

bowenwang1996 (approved):

PR #3711, opened 2020-12-07
fix(chain): verify orphan approvals

mikhailOK:
Partially verify approvals for orphans.

Protocol changes:
- `Vec<Option<Approvals>>` in block header now has elements for slashed validators (must be None)
- Doomslug threshold from 2/3 of non-slashed to 2/3 of all stake
- `BlockHeaderV3` with `prev_height` field

Test plan
---------
unit test


chain/chain/src/chain.rs
```diff
@@ -3258,6 +3267,35 @@ impl<'a> ChainUpdate<'a> {
...
+        Ok(())
+    }
+
+    #[cfg(feature = "protocol_feature_block_header_v3")]
+    fn verify_orphan_header_approvals(&mut self, header: &BlockHeader) -> Result<(), Error> {
+        let prev_hash = header.prev_hash();
+        let prev_height = match header.prev_height() {
+            None => {
+                // this will accept orphans of V1 and V2
+                // TODO: reject header V1 and V2 after a certain height
```
bowenwang1996, line 3281: don't think we need to do this. We currently reject blocks that are older than tail right away, so after the network has been upgraded it wouldn't be long before the tail becomes V3

chain/chain/src/doomslug.rs
```diff
@@ -420,25 +426,27 @@ impl Doomslug {
...
-        let threshold1 = stakes.iter().map(|(x, _)| x).sum::<Balance>() * 2 / 3;
-        let threshold2 = stakes.iter().map(|(_, x)| x).sum::<Balance>() * 2 / 3;
+        let threshold1 = stakes.iter().map(|(x, _, _)| x).sum::<Balance>() * 2 / 3;
+        let threshold2 = stakes.iter().map(|(_, x, _)| x).sum::<Balance>() * 2 / 3;
 
         let approved_stake1 = approvals
             .iter()
             .zip(stakes.iter())
-            .map(|(approval, (stake, _))| if approval.is_some() { *stake } else { 0 })
+            .filter(|(_, (_, _, is_slashed))| !*is_slashed)
```
bowenwang1996, line 441: might be better to use `filter_map` directly https://nnethercote.github.io/perf-book/iterators.html#chaining

chain/epoch_manager/src/lib.rs
```diff
@@ -666,10 +666,30 @@ impl EpochManager {
...
 
+    /// get_heuristic_block_approvers_ordered: block producers for epoch
+    /// get_all_block_approvers_ordered: block producers for epoch, slashing info, sometimes block producers for next epoch
+    pub fn get_heuristic_block_approvers_ordered(
+        &mut self,
+        epoch_id: &EpochId,
+    ) -> Result<Vec<ApprovalStake>, EpochError> {
+        let epoch_info = self.get_epoch_info(epoch_id)?;
+        let mut settlement = vec![];
+        for validator_id in epoch_info.block_producers_settlement.iter() {
```
bowenwang1996, line 677: there is an existing function `get_all_block_producers_settlement`

core/primitives/src/block_header.rs
```diff
@@ -287,6 +360,14 @@ impl BlockHeader {
             next_bp_hash,
             block_merkle_root,
         };
+        #[cfg(not(feature = "protocol_feature_block_header_v3"))]
+        let last_header_v2_version = 9999;
```
bowenwang1996, line 364: can we use an option to avoid this?

neard/src/runtime.rs
```diff
@@ -770,6 +773,47 @@ impl RuntimeAdapter for NightshadeRuntime {
...
+        &self,
+        epoch_id: &EpochId,
+        doomslug_threshold_mode: DoomslugThresholdMode,
+        prev_block_hash: &CryptoHash,
+        prev_block_height: BlockHeight,
+        block_height: BlockHeight,
+        approvals: &[Option<Signature>],
+    ) -> Result<(), Error> {
+        let mut epoch_manager = self.epoch_manager.as_ref().write().expect(POISONED_LOCK_ERR);
+        let info =
```
bowenwang1996, line 787: let's move this two lines in two a block so that the lock is dropped immediately after we get the necessary data.

bowenwang1996 (commented):
Verifying all signatures in a header isn't very cheap. We should make sure that we only do it once for every block, i.e, if a block is orphaned and later processed, we should not verify signature again.

bowenwang1996:
@SkidanovAlex Will this change break the bridge?

chain/chain/src/chain.rs
mikhailOK: Where do we reject them?
Seems like some checks are missing

chain/epoch_manager/src/lib.rs
mikhailOK: `get_all_block_producers_settlement` takes a block hash, which is not available

chain/chain/src/chain.rs
bowenwang1996: in `ClientActor::receive_block`, you might want to merge with master

mikhailOK: can't see the check, made `test_header_version_downgrade` with `should_panic`

chain/epoch_manager/src/lib.rs
bowenwang1996: I see, but we can use the `epoch_validators_ordered` cache

chain/chain/src/chain.rs
bowenwang1996: https://github.com/near/nearcore/blob/4a9ecaed34a9bce3ba7c469af48a78d9d02e8461/chain/client/src/client_actor.rs#L960

mikhailOK: We reject old heights, but don't reject old versions - a malicious node can produce V1 at any height, and it will be approved

bowenwang1996: oh I misunderstood what you meant

frol:
It seems to be in conflict with #3804 which also introduced BlockHeader v3

frol:
What is the state of this PR?

bowenwang1996 (commented):
@mikhailOK you need to bump nightly protocol version again for this change so that it automatically triggers a hard fork when we deploy it to betanet.

chain/client/tests/process_blocks.rs
```diff
@@ -2408,6 +2410,51 @@ fn test_not_broadcast_block_on_accept() {
     assert!(network_adapter.requests.read().unwrap().is_empty());
 }
 
+#[test]
+#[should_panic]
+// TODO: reject header version downgrade
```
bowenwang1996, line 2415: are you going to do this in a separate PR? If so, it is better to add the issue number here

bowenwang1996 (approved):

PR #3715, opened 2020-12-09
fix(chunks & network): Remove some unnecessary cloning

### LINKED ISSUE #3700: Reduce cloning in handling of partial encoded chunks
There are a few places in the code for handling partial encoded chunks where we use cloning as an escape hatch from the borrow checker. For example, @bowenwang1996 pointed out that we clone before passing to `process_partial_encoded_chunk` even though we frequently exit that function early. Ref: https://github.com/near/nearcore/pull/3689#discussion_r536191245

The purpose of this issue is to review usage of `Clone` in the partial encoded chunks pipeline and try to eliminate as many as possible (in particular the one pointed out by Bowen above).

birchmd:
This PR attempts to improve performance by limiting the amount of cloning done in the logic for processing partial encoded chunks. In the process I found there were a couple places in the networking layer where there was excess cloning as well, and fixed there too.

Fixes #3700 

Testing plan:
  * No logic is changed, so we rely on existing testing infrastructure

### DIFF
--- a/chain/chunks/src/chunk_cache.rs
+++ b/chain/chunks/src/chunk_cache.rs
@@ -75,17 +75,14 @@ impl EncodedChunksCache {
     }
 
     // `chunk_header` must be `Some` if the entry is absent, caller must ensure that
-    pub fn get_or_insert_from_header<F>(
+    pub fn get_or_insert_from_header(
         &mut self,
         chunk_hash: ChunkHash,
-        lazy_chunk_header: F,
-    ) -> &mut EncodedChunksCacheEntry
-    where
-        F: FnOnce() -> ShardChunkHeader,
-    {
+        chunk_header: &ShardChunkHeader,
+    ) -> &mut EncodedChunksCacheEntry {
         self.encoded_chunks
             .entry(chunk_hash)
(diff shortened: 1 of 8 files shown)


chain/chunks/src/chunk_cache.rs
```diff
@@ -75,14 +75,17 @@ impl EncodedChunksCache {
     }
 
     // `chunk_header` must be `Some` if the entry is absent, caller must ensure that
-    pub fn get_or_insert_from_header(
+    pub fn get_or_insert_from_header<F>(
         &mut self,
         chunk_hash: ChunkHash,
-        chunk_header: ShardChunkHeader,
-    ) -> &mut EncodedChunksCacheEntry {
+        lazy_chunk_header: F,
```
bowenwang1996, line 81: why do we do this instead of passing a reference?

birchmd: Yeah, good point. This is probably just being fancy for no reason. I likely was in the mindset of closures because I was looking at the closure used in the `or_insert_with` call.

Changed to reference in https://github.com/near/nearcore/pull/3715/commits/d74ac85451bbf0d36b9da649b44da104bab80b0b

bowenwang1996 (approved):

Issue #3706, opened 2020-12-07
Runtime performance regression


bowenwang1996:
Both @pmnoxx and I noticed a significant runtime performance regression on testnet with binary built from latest master https://github.com/near/nearcore/commit/6bac97f5362c4965634f50f666a32784c312f1ef. For example, in one instance, a node takes 13s to apply a chunk with only one receipt. 

This happens on testnet at block height 27362367, the chunk is as follows:
```
{
    "id": "123",
    "jsonrpc": "2.0",
    "result": {
        "author": "jazza.pool.f863973.m0",
        "header": {
            "balance_burnt": "242883773294000000000",
            "chunk_hash": "4bYbBvkBAGZv7MJuCZGfveM65UCtxuEB2UtGsDhWTtjg",
            "encoded_length": 593,
            "encoded_merkle_root": "Hchzigw5NB5m2hbbrQ2GAP7iEuUe7hc8zzdVkhZfDA1d",
            "gas_limit": 1000000000000000,
            "gas_used": 2428837732940,
            "height_created": 27362367,
            "height_included": 27362367,
            "outcome_root": "EL5sj8g8GV6eCjMxub7ivVGFUCTCM6386JRh7D15T4nU",
            "outgoing_receipts_root": "AKLknWijM2WZrxFJczro9aTdMPXcuu5FXGsaZhhsQSgS",
            "prev_block_hash": "4BbbcLg8FQ7YjJBnC8CzR3xtQDjGiBmemf7x59TfHYLx",
            "prev_state_root": "6Mag1h1Vx738AHds5AB9MTWhz7GiBZ5iuLtGy4MtU133",
            "rent_paid": "0",
            "shard_id": 0,
            "signature": "ed25519:2LyhBEFkmjgafAxtP3gvdiJojvGn2N4tKkopA2eNCSrdAt41Vt5hbyrvnzBT9HEmmxaVsAb9FmXgxAkJL9jUxoxD",
            "tx_root": "11111111111111111111111111111111",
            "validator_proposals": [],
            "validator_reward": "0"
        },
        "receipts": [
            {
                "predecessor_id": "jon.testnet",
                "receipt": {
                    "Action": {
                        "actions": [
                            {
                                "FunctionCall": {
[...]

PR #3716, opened 2020-12-10
fix: Remove default features from runtime-param-estimator to avoid leaking `no_cache`

evgenykuzyakov:
Rust cargo assumes all features are beneficial and enables them whenever possible.
If you have a dependency in the workspace cargo toml and compile a crate, it would leak all features to other binaries.
This is what caused regression in performance by enabling `no_cache` on `near-store` crate.

This change moves all features from default to `required` feature and updates scripts for `runtime-params-estimator` 

Fixes: #3706

# Test plan:
- Printed cache size when `TrieCache` is constructed and verified it's set to `10K`.
- CI
- near-store bench shows no regression to pre-EVM branch:
```
running 2 tests
test trie_lookup ... bench:     148,107 ns/iter (+/- 13,365)
test trie_update ... bench:     350,430 ns/iter (+/- 26,119)
```

### DIFF
--- a/runtime/runtime-params-estimator/Cargo.toml
+++ b/runtime/runtime-params-estimator/Cargo.toml
@@ -21,8 +21,8 @@ near-chain-configs = { path = "../../core/chain-configs" }
 near-runtime-fees = { path = "../../runtime/near-runtime-fees" }
 near-crypto = { path = "../../core/crypto" }
 near-vm-logic = {path = "../../runtime/near-vm-logic" , features = ["costs_counting"]}
-near-vm-runner = {path = "../../runtime/near-vm-runner" , features = ["costs_counting", "no_cache", "no_cpu_compatibility_checks", "wasmtime_vm" ]}
-node-runtime = { path = "../../runtime/runtime" , features = ["costs_counting", "no_cache"]}
+near-vm-runner = {path = "../../runtime/near-vm-runner" , features = ["costs_counting" ]}
+node-runtime = { path = "../../runtime/runtime" , features = ["costs_counting"]}
 near-store = { path = "../../core/store" }
 near-primitives = { path = "../../core/primitives" }
 testlib = { path = "../../test-utils/testlib" }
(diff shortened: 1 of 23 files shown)


runtime/runtime-params-estimator/Cargo.toml
```diff
@@ -21,8 +21,8 @@ near-chain-configs = { path = "../../core/chain-configs" }
 near-runtime-fees = { path = "../../runtime/near-runtime-fees" }
 near-crypto = { path = "../../core/crypto" }
 near-vm-logic = {path = "../../runtime/near-vm-logic" , features = ["costs_counting"]}
-near-vm-runner = {path = "../../runtime/near-vm-runner" , features = ["costs_counting", "no_cache", "no_cpu_compatibility_checks", "wasmtime_vm" ]}
```
evgenykuzyakov, line 24: Note, we don't need `wasmtime_vm` because `wasmtime` feature will enable it.

bowenwang1996 (commented):
shouldn't we make the dependency on runtime-params-estimator optional? Which crate depends on it?

evgenykuzyakov:
root `Cargo.toml` workspace depends on it

ailisp:
> root Cargo.toml workspace depends on it

I thought only when cargo build --workspace would include the feature, :bomb: cargo :bomb:

ailisp:
We can consider add a special neard test or feature check, to prevent unexpected feature come into binary?

bowenwang1996:
> root Cargo.toml workspace depends on it

I see. There isn't a way to make this optional?

evgenykuzyakov:
> I see. There isn't a way to make this optional?

Optional features still leaks every dependency feature

ailisp:
Wait, isn't no_cache feature present in runtime-param-estimator's cargo.toml long ago? why it leak after evm PR

willemneal (approved):
I had noticed wasmtime getting compiled when the feature wasn't set so this makes total sense now.

evgenykuzyakov:
> Wait, isn't no_cache feature present in runtime-param-estimator's cargo.toml long ago? why it leak after evm PR

because it was added to the `[dev-dependencies]` section to run nightly features tests 

Also I'll add a test that makes sure that `no_cache` feature is disabled

bowenwang1996 (approved):

PR #3730, opened 2020-12-11
feat: Add an impl for ProfileData and add it as an option to ApplyState

willemneal:
This change will allow the simulator to pass a profile data reference via ApplyState to get the break down of how gas was spent.

Currently this means only function calls are profiled, not view calls because the ViewApplyState is in the primitives crate, while ProfileData and ApplyState are in the `near-vm-logic` crate.  A future PR that reorganizes these core types will allow adding the profile data to ViewApplyState.


runtime/near-vm-logic/src/types.rs
```diff
@@ -50,5 +52,156 @@ pub enum PromiseResult {
...
+  fn borrow(&self) -> InternalProfileData<u64> {
+    *self.data.borrow()
+  }
+
+  fn add_val(&self, index: usize, value: u64) {
+    *self.data.borrow_mut().get_mut(index).unwrap() += value;
+  }
+
+  pub fn add_action_cost(&self, action: ActionCosts, value: u64) {
+    self.add_val(ProfileData:: ACTION_START + action as usize, value);
```
ailisp, line 99: should we add to both action_cost and all_gas here? Also add_ext_cost below

willemneal: It is set after execution as gas costs can be from other sources than these two.

ailisp (approved):

neard/src/runtime.rs
```diff
@@ -451,6 +451,7 @@ impl NightshadeRuntime {
             cache: Some(Arc::new(StoreCompiledContractCache { store: self.store.clone() })),
             #[cfg(feature = "protocol_feature_evm")]
             evm_chain_id: self.evm_chain_id(),
+            profile: None
```
bowenwang1996, line 454: is it better if this is hidden behind a feature flag just like `evm_chain_id`?

runtime/runtime/src/actions.rs
```diff
@@ -165,7 +166,7 @@ pub(crate) fn action_function_call(
         return Err(StorageError::StorageInconsistentState(
             "Account balance integer overflow during function call deposit".to_string(),
         )
-        .into());
+            .into());
```
bowenwang1996, line 169: changes here and below seem unnecessary

neard/src/runtime.rs
willemneal: Having discussed this with @olonho, he made a good point that it might be worth having in general as it could be used in a more detailed outcome via RPC.  I had initially hidden it behind a flag and could switch back.

ailisp: I agree it's better not to hide it. protocol_feature_xxx is used when we eventually want to add this feature to nodes but not right now. This is a feature that is not a protocol upgrade and is a feature we want to enable right now

bowenwang1996: It is not related to the protocol features. It is more about the fact that we don't need `profile` for the node to function in its normal setting. Is the current approach using a runtime flag to guard this?

willemneal: There is already a `costs_counting` feature that will prevent the node from keeping track of the fees in `VMLogic`.  So the "tax" on the nodes is passing around a None value. Currently the main `run` method in the runner checks if the value is `Some` and `run_vm_profiled`.

Would you prefer that this join the `cost_counting` feature and only check for option type with it?

bowenwang1996: Okay then I think it makes sense to put this behind `cost_counting` as well.

runtime/near-vm-runner/src/runner.rs
```diff
@@ -27,8 +27,10 @@ pub fn run<'a>(
     promise_results: &'a [PromiseResult],
     current_protocol_version: ProtocolVersion,
     cache: Option<&'a dyn CompiledContractCache>,
+    #[allow(unused_variables)]
```
bowenwang1996: why is this variable not guarded by the feature flag?

runtime/near-vm-runner/src/runner.rs
```diff
@@ -27,7 +27,39 @@ pub fn run<'a>(
     promise_results: &'a [PromiseResult],
     current_protocol_version: ProtocolVersion,
     cache: Option<&'a dyn CompiledContractCache>,
+    #[allow(unused_variables)] profile: &Option<ProfileData>,
 ) -> (Option<VMOutcome>, Option<VMError>) {
+    #[cfg(feature = "costs_counting")]
```
bowenwang1996, line 32: if we do this why do we not put the function argument behind a feature flag so that we don't have to do something like `#[allow(unused_variables)]`

willemneal: Because the caller always passes the parameter since it's in `ApplyState` and by default is None, but we only use it when the feature is on. 

Making the parameter be behind the feature would require moving each call behind a feature flag too.  Is this what you want?

bowenwang1996 (approved):

runtime/near-vm-runner/src/runner.rs
evgenykuzyakov: It sounds like a good idea. For example on-disk compilation cache introduced a few unwanted changes. Hiding a feature behind a flag makes it safer for non-prod changes.

neard/src/runtime.rs
```diff
@@ -457,6 +457,7 @@ impl NightshadeRuntime {
             cache: Some(Arc::new(StoreCompiledContractCache { store: self.store.clone() })),
             #[cfg(feature = "protocol_feature_evm")]
             evm_chain_id: self.evm_chain_id(),
+            profile: None,
```
evgenykuzyakov, line 460: Let's wrap `profile` field in the `ApplyState` as well with a feature.

evgenykuzyakov (approved):

olonho (approved):

PR #3738, opened 2020-12-14
add performance metrics feature

pmnoxx:
Currently we don't measure how much time it takes to process messages in actors. In this PR we will add an option to print every minute CPU usage of every thread, and printing whenever execution time takes more than 50ms.

### DIFF
--- a/chain/client/src/client_actor.rs
+++ b/chain/client/src/client_actor.rs
@@ -32,6 +32,7 @@ use near_network::types::{NetworkInfo, ReasonForBan};
 use near_network::{
     NetworkAdapter, NetworkClientMessages, NetworkClientResponses, NetworkRequests,
 };
+use near_performance_metrics_macros::perf;
 use near_primitives::hash::CryptoHash;
 use near_primitives::network::{AnnounceAccount, PeerId};
 use near_primitives::types::{BlockHeight, EpochId};
@@ -190,6 +191,7 @@ impl Actor for ClientActor {
 impl Handler<NetworkClientMessages> for ClientActor {
     type Result = NetworkClientResponses;
 
+    #[perf]
     fn handle(&mut self, msg: NetworkClientMessages, ctx: &mut Context<Self>) -> Self::Result {
         #[cfg(feature = "delay_detector")]
         let _d = DelayDetector::new(format!("NetworkClientMessage {}", msg.as_ref()).into());
@@ -352,7 +354,7 @@ impl Handler<NetworkClientMessages> for ClientActor {
                 trace!(target: "sync", "Received state response shard_id: {} sync_hash: {:?} part(id/size): {:?}",
                     shard_id,
                     hash,
-                    state_response.part().as_ref().map(|(part_id,data)|(part_id, data.len()))
+                    state_response.part().as_ref().map(|(part_id, data)| (part_id, data.len()))
                 );
                 // Get the download that matches the shard_id and hash
                 let download = {
@@ -518,6 +520,7 @@ impl Handler<NetworkClientMessages> for ClientActor {
 impl Handler<Status> for ClientActor {
     type Result = Result<StatusResponse, String>;
 
+    #[perf]
     fn handle(&mut self, msg: Status, ctx: &mut Context<Self>) -> Self::Result {
         #[cfg(feature = "delay_detector")]
         let _d = DelayDetector::new("client status".to_string().into());
@@ -592,7 +595,8 @@ impl Handler<Status> for ClientActor {
 impl Handler<GetNetworkInfo> for ClientActor {
     type Result = Result<NetworkInfoResponse, String>;
 
-    fn handle(&mut self, _: GetNetworkInfo, ctx: &mut Context<Self>) -> Self::Result {
+    #[perf]
+    fn handle(&mut self, msg: GetNetworkInfo, ctx: &mut Context<Self>) -> Self::Result {
         #[cfg(feature = "delay_detector")]
         let _d = DelayDetector::new("client get network info".into());
         self.check_triggers(ctx);
--- a/chain/client/src/client_actor.rs
+++ b/chain/client/src/client_actor.rs
@@ -32,6 +32,7 @@ use near_network::types::{NetworkInfo, ReasonForBan};
 use near_network::{
     NetworkAdapter, NetworkClientMessages, NetworkClientResponses, NetworkRequests,
 };
+use near_performance_metrics;
 use near_performance_metrics_macros::perf;
 use near_primitives::hash::CryptoHash;
 use near_primitives::network::{AnnounceAccount, PeerId};
@@ -733,7 +734,7 @@ impl ClientActor {
     fn schedule_triggers(&mut self, ctx: &mut Context<Self>) {
         let wait = self.check_triggers(ctx);
 
-        ctx.run_later(wait, move |act, ctx| {
+        near_performance_metrics::actix::run_later(ctx, file!(), line!(), wait, move |act, ctx| {
             act.schedule_triggers(ctx);
         });
     }
@@ -1113,9 +1114,15 @@ impl ClientActor {
         if self.network_info.num_active_peers < self.client.config.min_num_peers
             && !self.client.config.skip_sync_wait
         {
-            ctx.run_later(self.client.config.sync_step_period, move |act, ctx| {
-                act.start_sync(ctx);
-            });
+            near_performance_metrics::actix::run_later(
+                ctx,
+                file!(),
+                line!(),
+                self.client.config.sync_step_period,
+                move |act, ctx| {
+                    act.start_sync(ctx);
+                },
+            );
             return;
         }
         self.sync_started = true;
@@ -1169,9 +1176,15 @@ impl ClientActor {
             }
         }
 
-        ctx.run_later(self.client.config.catchup_step_period, move |act, ctx| {
-            act.catchup(ctx);
-        });
+        near_performance_metrics::actix::run_later(
+            ctx,
+            file!(),
+            line!(),
+            self.client.config.catchup_step_period,
+            move |act, ctx| {
+                act.catchup(ctx);
+            },
+        );
     }
 
     fn run_timer<F>(
@@ -1203,7 +1216,12 @@ impl ClientActor {
             Ok(v) => v,
             Err(err) => {
                 error!(target: "sync", "Sync: Unexpected error: {}", err);
-                ctx.run_later(self.client.config.sync_step_period, move |act, ctx| {
+
+            near_performance_metrics::actix::run_later(
+                ctx,
+                file!(),
+                line!(),
+                self.client.config.sync_step_period, move |act, ctx| {
                     act.sync(ctx);
                 });
                 return;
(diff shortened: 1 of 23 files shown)


chain/client/src/info.rs.orig
```diff
@@ -0,0 +1,291 @@
+use std::cmp::min;
```
frol, line 1: I think this is an artifact of some merge conflict, so we need to remove it.

chain/client/src/info.rs.rej
```diff
@@ -0,0 +1,30 @@
+--- chain/client/src/info.rs
```
frol, line 1: ^

neard/tests/sync_state_nodes.rs
```diff
@@ -279,6 +279,7 @@ fn sync_empty_state() {
 
                     actix::spawn(view_client1.send(GetBlock::latest()).then(move |res| {
                         match &res {
+                            /**/
```
frol, line 282:
```suggestion
```

performance-metrics/src/stats.rs
```diff
@@ -0,0 +1,311 @@
...
+        self.classes.insert(class_name);
+    }
+
+    fn print_stats(&self, tid: usize, sleep_time: Duration) {
+        let ratio = (self.time.as_nanos() as f64) / (sleep_time.as_nanos() as f64);
+
+        let class_name = format!("{:?}", self.classes);
+
+        if ratio > MIN_OCCUPANCY_RATIO_THRESHOLD {
+            info!(
```
frol, line 73: As far as I remember, we need to specify `target: "performance_metric"` in every log call, so tracing filters apply correctly.

frol (commented):
It feels so intrusive and boilerplaty to the codebase :confused:

performance-metrics/Cargo.toml
```diff
@@ -0,0 +1,17 @@
+[package]
```
frol, line 1: If it is nearcore-specific crate, let's name it with `near-` prefix and move to the `utils` folder, otherwise, I vote for extracting it to a standalone crate with its own git repo and publish on crates.io

performance-metrics/src/stats.rs
bowenwang1996: Yep

chain/client/src/client_actor.rs
```diff
@@ -380,238 +412,226 @@ impl Handler<NetworkClientMessages> for ClientActor {
...
-            if now > block_timestamp {
-                let elapsed = (now - block_timestamp).to_std().unwrap();
-                if elapsed
-                    > Duration::from_millis(
-                        self.client.config.max_block_production_delay.as_millis() as u64
-                            * STATUS_WAIT_TIME_MULTIPLIER,
-                    )
-                {
-                    return Err(format!("No blocks for {:?}.", elapsed));
+        performance(std::any::type_name::<Self>(), msg, |msg| {
```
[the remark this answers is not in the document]
bowenwang1996: @frol what do you think of the current implementation? Is it better to create some procedural macro that we can mark the entire `impl Handler<..>` with?

chain/indexer/src/lib.rs
```diff
@@ -80,13 +81,18 @@ impl Indexer {
...
     pub fn streamer(&self) -> mpsc::Receiver<streamer::StreamerMessage> {
         let (sender, receiver) = mpsc::channel(16);
-        actix::spawn(streamer::start(
-            self.view_client.clone(),
-            self.client.clone(),
-            self.near_config.clone(),
-            self.indexer_config.clone(),
-            sender,
-        ));
+        performance_metrics::stats::spawn(
```
bowenwang1996, line 84: @frol do you need this for indexer?

bowenwang1996 (commented):
it feels to me that a procedural macro implementation would be much cleaner

chain/client/src/info.rs.orig
pmnoxx: removed

chain/indexer/src/lib.rs
pmnoxx: This is a separate binary, so this is no longer needed.

pmnoxx: I'll revert this change.

performance-metrics/Cargo.toml
pmnoxx: I agree, I renamed it to near-performance-metrics and moved it to utils.

performance-metrics/src/stats.rs
pmnoxx: Yes, I did it in config.rs

performance-metrics/Cargo.toml
pmnoxx: I would propose to keep it in utils for now, we can extract it to cargo.io later

chain/client/src/client_actor.rs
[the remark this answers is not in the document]
pmnoxx: I renamed this method to measure_performance

pmnoxx:
> It feels so intrusive and boilerplaty to the codebase

> it feels to me that a procedural macro implementation would be much cleaner

We can use regular macros to make the code less intrusive. This is how the code could look like:
```
generate_handler!(ViewClientActor, Query, Result<Option<QueryResponse>, String>);
/// Handles runtime query.
impl PerfHandler<Query> for ViewClientActor {
    type Result = Result<Option<QueryResponse>, String>;

    fn perf_handle(&mut self, msg: Query, _: &mut Self::Context) -> Self::Result {
        self.handle_query(msg)
    }
}
```
Regular macros are hygienic as opposed to procedural macros. I would suggest to use the above syntrax.

pmnoxx:
> > It feels so intrusive and boilerplaty to the codebase
> 
> > it feels to me that a procedural macro implementation would be much cleaner
> 
> We can use regular macros to make the code less intrusive. This is how the code could look like:
> 
> ```
> generate_handler!(ViewClientActor, Query, Result<Option<QueryResponse>, String>);
> /// Handles runtime query.
> impl PerfHandler<Query> for ViewClientActor {
>     type Result = Result<Option<QueryResponse>, String>;
> 
>     fn perf_handle(&mut self, msg: Query, _: &mut Self::Context) -> Self::Result {
>         self.handle_query(msg)
>     }
> }
> ```
> 
> Regular macros are hygienic as opposed to procedural macros. I would suggest to use the above syntrax.
> @bowenwang1996 @frol

The other alternative would be something like:
```
generate_handler!(ViewClientActor, Query, Result<Option<QueryResponse>, String>,
    fn perf_handle(&mut self, msg: Query, _: &mut Self::Context) -> Self::Result {
        self.handle_query(msg)
    }
}
```
The code becomes shorter, but unfortunately more obsure.

frol:
I haven't put much thought into it myself. I am up to anything that is less intrusive, and easier to deal with. Does the first `generate_handler!` macro generates the code you posted below it meaning that you only need to add a single line to the Handler impl?

pmnoxx:
```+pub trait PerfHandler<M>
+where
+    Self: Actor,
+    M: Message,
+{
+    /// The type of value that this handler will return.
+    type Result: MessageResponse<Self, M>;
+
+    /// This method is called for every message received by this actor.
+    fn perf_handle(&mut self, msg: M, ctx: &mut Self::Context) -> Self::Result;
+}
+
+macro_rules! generate_handler {
+    ($class: ty, $message:ty, $result:ty) => {
+        impl Handler<$message> for $class {
+            type Result = $result;
+
+            fn handle(&mut self, msg: $message, ctx: &mut Self::Context) -> Self::Result {
+                self.perf_handle(msg, ctx)
+            }
+        }
+    };
+}```

**Comment by pmnoxx** [comment]
In this case we would generate
```

        impl Handler<Query> for ViewClientActor {
            type Result =Result<Option<QueryResponse>, String>;

            fn handle(&mut self, msg: Query, ctx: &mut Self::Context) -> Self::Result {
                self.perf_handle(msg, ctx)
            }
        }
```

**Comment by bowenwang1996** [comment]
@pmnoxx I think the macro solution you proposed is fine, but it would be really nice if we can do something like

```rust
impl Handler<Query> for ViewClientActor {
   type Result =Result<Option<QueryResponse>, String>;

    #[perf]
     fn handle(&mut self, msg: Query, ctx: &mut Self::Context) -> Self::Result {
                ...
      }
}
```
but this probably requires a lot more work

pmnoxx:
I can also add macros for:
```
near_performance_metrics::stats::spawn(
                    std::any::type_name::<Self>(),
                    file!(),
                    line!(),
                   arg
);
```
to be 
```
actix_spawn!(arg)
```

pmnoxx:
> ```rust
> impl Handler<Query> for ViewClientActor {
>    type Result =Result<Option<QueryResponse>, String>;
> 
>     #[perf]
>     fn handle(&mut self, msg: Query, ctx: &mut Self::Context) -> Self::Result {
>           ...
>     }
> }
> ```

This probably won't be possible. However, it may be possible to do something like:
```
#[perf]
impl Handler<Query> for ViewClientActor {
   type Result =Result<Option<QueryResponse>, String>;

    fn handle(&mut self, msg: Query, ctx: &mut Self::Context) -> Self::Result {
          ...
    }
}
```
I would have to write a procedural macro that gets 3 items from this:
`Query,  ViewClientActor , Result<Option<QueryResponse>, String>`
to generate `impl Handler<Query> for ViewClientActor {`

This is doable, I can spend a day to try to get it to work.

pmnoxx:
I got the following syntax to work:
```
impl Handler<Query> for ViewClientActor {
    type Result = Result<Option<QueryResponse>, String>;

    #[perf]
    fn handle(&mut self, msg: Query, _: &mut Self::Context) -> Self::Result {
        self.handle_query(msg)
    }
}
```

chain/client/src/client_actor.rs
```diff
@@ -350,9 +352,9 @@ impl Handler<NetworkClientMessages> for ClientActor {
                 let state_response = state_response_info.take_state_response();
 
                 trace!(target: "sync", "Received state response shard_id: {} sync_hash: {:?} part(id/size): {:?}",
-                    shard_id,
-                    hash,
-                    state_response.part().as_ref().map(|(part_id,data)|(part_id, data.len()))
+                       shard_id,
```
bowenwang1996, line 355: this seems unnecessary.

chain/network/src/peer.rs
```diff
@@ -850,40 +853,40 @@ impl StreamHandler<Result<Vec<u8>, ReasonForBan>> for Peer {
...
-                    .then(move |res, act, ctx| {
-                        match res {
-                            Ok(ConsolidateResponse::Accept(edge_info)) => {
-                                act.peer_info = Some(peer_info).into();
-                                act.peer_status = PeerStatus::Ready;
-                                // Respond to handshake if it's inbound and connection was consolidated.
-                                if act.peer_type == PeerType::Inbound {
-                                    act.edge_info = edge_info;
-                                    act.send_handshake(ctx);
+                        .send(Consolidate {
```
bowenwang1996, line 856: seems like just formatting change. Did you run `cargo fmt --all` (I recommend running formatter automatically on save)?

chain/network/tests/runner/mod.rs
```diff
@@ -656,24 +656,26 @@ enum RunnerMessage {
...
-            }
-            RunnerMessage::ChangeAccountId(node_id, account_id) => {
-                self.accounts_id.as_mut().unwrap()[node_id] = account_id.clone();
-                let info = self.info.as_ref().cloned().unwrap();
-                let mut write_info = info.write().unwrap();
-
-                write_info.peers_info[node_id].id = peer_id_from_seed(account_id.as_str());
-                if write_info.peers_info[node_id].account_id.is_some() {
-                    write_info.peers_info[node_id].account_id = Some(account_id);
+            std::any::type_name::<Self>(),
```
bowenwang1996: not necessary here. This file is for tests

chain/network/tests/stress_network.rs
```diff
@@ -114,18 +114,23 @@ fn stress_test() {
                         if !flag.load(Ordering::Relaxed) {
                             let flag1 = flag.clone();
 
-                            actix::spawn(pms[ix].send(GetInfo {}).then(move |info| {
-                                if let Ok(info) = info {
-                                    if info.num_active_peers == num_nodes - 2 {
-                                        flag1.store(true, Ordering::Relaxed);
+                            near_performance_metrics::stats::spawn(
```
bowenwang1996, line 117: same here

chain/client/src/sync.rs
```diff
@@ -853,7 +853,10 @@ impl StateSync {
                 new_shard_sync_download.downloads[0].state_requests_count += 1;
                 new_shard_sync_download.downloads[0].last_target = Some(target.clone());
                 let run_me = new_shard_sync_download.downloads[0].run_me.clone();
-                actix::spawn(
+                near_performance_metrics::stats::spawn(
```
bowenwang1996, line 856: shouldn't this be guarded by the feature flag?

pmnoxx: There is a flag inside near_performance_metrics::stats::spawn

chain/network/src/peer.rs
pmnoxx: I'm running formatted automatically on save. I also ran cargo fmt --all to make sure.

chain/network/src/peer.rs
```diff
@@ -932,7 +935,7 @@ impl StreamHandler<Result<Vec<u8>, ReasonForBan>> for Peer {
                         }
                     }
                     actix::fut::ready(())
-                }).spawn(ctx);
+                    }).spawn(ctx);
```
bowenwang1996, line 938: this looks a bit off

utils/near-performance-metrics-macros/src/lib.rs
```diff
@@ -0,0 +1,44 @@
...
+        output.into()
+    } else {
+        panic!("not a function {:?}");
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    #[test]
+    fn it_works() {
```
bowenwang1996, line 41: this should be removed

utils/near-performance-metrics/src/stats.rs
```diff
@@ -0,0 +1,341 @@
...
+                class_name,
+                TID.with(|x| *x.borrow()),
+                std::any::type_name::<Message>(),
+                took.as_millis(),
+                msg_length,
+                msg_test
+            );
+        }
+
+        STATS.write().unwrap().log(class_name, std::any::type_name::<Message>(), 0, took);
```
bowenwang1996, line 216: Is lock contention a concern here? What a large number of actors try to write at the same time

bowenwang1996 (approved):

utils/near-performance-metrics/src/stats.rs
pmnoxx: That should not be an issue, we don't many messages.

pmnoxx:
I'll run a node for a day for make sure everything is working fine.

utils/near-performance-metrics/src/stats.rs
```diff
@@ -0,0 +1,341 @@
...
+            std::any::type_name::<Message>(),
+            took.as_millis(),
+        );
+    }
+
+    STATS.write().unwrap().log(class_name, std::any::type_name::<Message>(), 0, took);
+    result
+}
+
+pub fn measure_performance2<F, Result>(
```
frol, line 151: Where is it used? Let's have a better name for it :smile:

utils/near-performance-metrics/src/stats.rs
```diff
@@ -0,0 +1,341 @@
...
+                    1;
+                Poll::Ready(x)
+            }
+            Poll::Pending => Poll::Pending,
+        }
+    }
+}
+
+#[allow(unused_variables)]
+pub fn spawn<F>(class_name: &'static str, file: &'static str, line: u32, f: F)
```
frol, line 267: How do you feel about moving these helpers into a separate module? It looks a bit odd when you see `metrics::stats::spawn`.

ailisp (approved):

utils/near-performance-metrics/src/lib.rs
```diff
@@ -0,0 +1,5 @@
+pub mod process;
+pub mod stats;
+
+#[macro_use]
+extern crate lazy_static;
```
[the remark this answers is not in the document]
pmnoxx: I replaced lazy_static with once_cell::sync::Lazy, so those macros are no longer needed.

utils/near-performance-metrics-macros/src/lib.rs
```diff
@@ -0,0 +1,36 @@
...
+
+        let function_body = quote! { #block }.to_string();
+
+        let new_body = format!(
+            "{} {} {}",
+            "fn xxx() { \
+            use near_performance_metrics::stats::measure_performance;\
+            near_performance_metrics::stats::measure_performance(std::any::type_name::<Self>(), msg, move |msg| { ",
+            function_body,
+            "}) }").parse().unwrap();
```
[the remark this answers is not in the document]
pmnoxx: Thanks, I got the `quote!` to work.

utils/near-performance-metrics/src/stats.rs
pmnoxx: I'll change those function to be `metrics::actix::spawn`.

utils/near-performance-metrics/src/stats.rs
pmnoxx: I removed that method, it's not longer needed.

frol (approved):
It is much cleaner and easier to maintain. Thanks for taking this longer path giving us cleaner implementation!

cargo-deny is not happy. I guess, you had run `cargo update` which lead to bumping all the dependencies. To minimize the scope of potential side-effects due to various dependencies update, we configured cargo-deny to guard us against duplicate versions of dependencies. To address cargo-deny errors, try reverting Cargo.lock to the master version and just run `cargo check`, it will only update the relevant pieces, and hopefully this won't introduce duplicate dependencies.

PR #3747, opened 2020-12-15
add feature flag to debug memory usage

pmnoxx:


memory-tracker/Cargo.toml
```diff
@@ -0,0 +1,24 @@
...
+[dependencies]
+rand = "0.7.3"
+libc = "0.2.65"
+backtrace = "0.3"
+arr_macro = "0.1.3"
+log = "0.4"
+tracing = "0.1.13"
+near-logger-utils = { path = "../test-utils/logger" }
+# never versions of heim don't compile, because of uom
+# heim = "0.0.11"
```
bowenwang1996, line 16: should we remove it then?

memory-tracker/src/allocator.rs
```diff
@@ -0,0 +1,301 @@
...
+        let mut addr: Option<*mut c_void> = Some(2 as *mut c_void);
+
+        const MAX_STACK: usize = 30;
+        let mut ary: [*mut c_void; MAX_STACK + 1] = [0 as *mut c_void; MAX_STACK + 1];
+        let mut chosen_i = 0;
+
+        if IN_TRACE.with(|in_trace| *in_trace.borrow()) == 0 {
+            IN_TRACE.with(|in_trace| *in_trace.borrow_mut() = 1);
+            if layout.size() >= ALLOC_LIMIT || rand::thread_rng().gen_range(0, 100) < 1
+            //                || LAST_SIZE.with(|ls| *ls.borrow()) != layout.size()
```
bowenwang1996, line 152: should this be removed?

bowenwang1996 (commented):
I think it is better if we extract memory tracker into its own repo. There is nothing that is nearcore specific there.

near-memory-tracker/src/lib.rs
```diff
@@ -0,0 +1 @@
+pub mod allocator;
```
frol, line 1: Just like with the other profiler, let's put it into the `utils/near-memory-tracker`

nohup.out
```diff
@@ -0,0 +1,9 @@
+Usage: nearup run [OPTIONS] [testnet|localnet|mainnet|betanet|crashnet]
```
frol, line 1: Nit: clean up

near-memory-tracker/src/lib.rs
bowenwang1996: I actually suggest that we put this in a separate repo. There is nothing nearcore specific here

pmnoxx: I'll put it in a separate repo. There are still some features I want to add to debug memory leaks.

utils/near-performance-metrics/src/stats_enabled.rs
```diff
@@ -2,18 +2,25 @@ use log::{info, warn};
...
 use futures::task::Context;
+use near_rust_allocator_proxy::allocator::{
+    current_thread_memory_usage, current_thread_peak_memory_usage, reset_memory_usage_max,
+    thread_memory_usage,
+};
 use once_cell::sync::Lazy;
 use std::pin::Pin;
 use strum::AsStaticRef;
 
+const MEBIBYTE: usize = 1024 * 1024;
```
bowenwang1996, line 21: Do you mean "MEGABYTE"?

pmnoxx: No, the proper term is Mebibyte, also denoted as MiB. It's 1024*1024 bytes.
https://simple.wikipedia.org/wiki/Mebibyte#:~:text=A%20mebibyte%20is%20a%20unit,to%20the%20megabyte%20(MB).
I copied constant from https://docs.rs/size/0.1.2/size/constant.MEBIBYTE.html
Megabyte is 1000*1000 bytes.

neard/src/main.rs
```diff
@@ -1,3 +1,7 @@
+#[cfg(not(feature = "memory_profiling"))]
+#[cfg(jemallocator)]
+extern crate jemallocator;
```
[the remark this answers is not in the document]
pmnoxx: I wanted to keep the behavior the same in case `memory_profiling` is not enabled as before the change.
You can take a look at `core/primitives/src/lib.rs`

utils/near-performance-metrics/src/stats_enabled.rs
```diff
@@ -40,6 +47,20 @@ struct ThreadStats {
     classes: HashSet<&'static str>,
 }
 
+extern "C" {
+    pub fn gettid() -> u32;
+}
+
+pub fn get_tid() -> usize {
+    let res = TID.with(|t| {
+        if *t.borrow() == usize::max_value() {
```
frol, line 56: The initial value is 0. Am I missing something here or this condition will never be true?

utils/near-performance-metrics/src/stats_enabled.rs
```diff
@@ -40,6 +47,20 @@ struct ThreadStats {
     classes: HashSet<&'static str>,
 }
 
+extern "C" {
+    pub fn gettid() -> u32;
+}
```
frol, line 52: I would like to suggest using `gettid` implementation from `nix` crate (https://docs.rs/nix/0.19.1/nix/unistd/fn.gettid.html) instead of depending implicitly on libc, and explicit use of unsafe calling this method

bowenwang1996 (approved):

utils/near-performance-metrics/src/stats_enabled.rs
pmnoxx: Fixed

utils/near-performance-metrics/src/stats_enabled.rs
pmnoxx: Fixed

PR #3753, opened 2020-12-17
feat: update Makefile to use stable

willemneal:
Since wasmtime isn't used in production this overrides the toolchain file when building with Makefile


ailisp:
Sounds good, I think also worth to modify this file to pin a specific version of stable rust?
> https://github.com/near/nearcore/blob/feat/stable_release/rust-toolchain

ailisp (approved):

willemneal:
Yeah the toolchain file would be the easiest but with wasmtime still requiring nightly it would fail the - -all-feature check in CI

ailisp:
:+1: I see, your change is probably the only way

ailisp:
@willemneal  oh before merge, also change:
- Dockerfile
- Dockerfile.nightly

bowenwang1996 (commented):
@willemneal should we update rust-toolchain as well?

willemneal:
@bowenwang1996 As mentioned above since we have a feature for using wasmtime/lightbeam, CI will fail when doing the `--all-features` check if the toolchain file is stable.  This is a work around to still build the release with stable, while allowing experimentation with nightly requirements.  

So I was also hoping that the build with stable could be part of CI.

Makefile
```diff
@@ -5,20 +5,20 @@ docker-nearcore-nightly:
...
-	cargo build -p store-validator --release
+	cargo +stable build -p neard --release
+	cargo +stable build -p keypair-generator --release
+	cargo +stable build -p genesis-csv-to-json --release
+	cargo +stable build -p near-vm-runner-standalone --release
+	cargo +stable build -p state-viewer --release
+	cargo +stable build -p store-validator --release
 
 debug:
-	cargo build -p neard
```
olonho, line 16: Cannot we have set of options extracted to the variable?

willemneal: Can you explain more?  I'm not familiar with Makefiles.

olonho: Just create variable `RUST_OPTIONS`, set it to `+stable` and use `$RUST_OPTIONS` later on.

Makefile
```diff
@@ -5,20 +5,20 @@ docker-nearcore-nightly:
...
-	cargo build -p state-viewer
-	cargo build -p store-validator
+	cargo +stable build -p neard
+	cargo +stable build -p keypair-generator
+	cargo +stable build -p genesis-csv-to-json
+	cargo +stable build -p near-vm-runner-standalone
+	cargo +stable build -p state-viewer
+	cargo +stable build -p store-validator
 
 nightly-release:
```
chefsale, line 25: This should also use the stavle version than, as its just near nightly.

chefsale (changes requested):

chefsale (approved):

PR #3761, opened 2020-12-18
disable performance-stats feature by default

pmnoxx:
performance-stats  feature was not properly disabled, this PR fixes the issue.


pmnoxx:
#3761 Performance metrics feature enabled by default

bowenwang1996 (approved):
seems like there are some formatting changes that are not really necessary

PR #3773, opened 2021-01-02
feat(evm): Raw Ethereum transaction relay support

ilblackdragon:
Adds support for relaying signed transactions from Ethereum.
Can be used by MetaMask configured to the right RPC proxy or NEAR's Web3 provider in `sendRawTransaction` method.

Addresses few collateral issues:
  - converting `chain_id` to u64, as prior it was u128 which didn't match OE EVM interface
  - fixing `view_call` function when it's used for contract deployment (this is used for dry_run)

Test plan
---
Adding unit test for sending signed transactions: transfer, deploy code and function call. Testing invalid signature and chain_id passes inside the transaction.
Adding support for sendRawTransaction in near-web3-provider repo: https://github.com/near/near-web3-provider/pull/70


runtime/near-evm-runner/src/runner.rs
```diff
@@ -249,6 +304,7 @@ impl<'a> EvmContext<'a> {
             return Err(VMLogicError::EvmError(EvmError::InvalidNonce));
         }
         self.add_balance(&meta_call_args.sender, U256::from(self.attached_deposit))?;
+        // TODO: this is wrong?!
```
bowenwang1996, line 307: what exactly is wrong?

runtime/near-evm-runner/src/types.rs
```diff
@@ -96,6 +101,113 @@ pub struct MetaCallArgs {
...
+    result
+}
+
+impl EthSignedTransaction {
+    /// Returns sender of given signed transaction by doing ecrecover on the signature.
+    pub fn sender(&self) -> Address {
+        let mut rlp_stream = RlpStream::new();
+        // See details of CHAIN_ID computation here - https://github.com/ethereum/EIPs/blob/master/EIPS/eip-155.md#specification
+        let chain_id =
+            if [27u64, 28u64].contains(&self.v) { None } else { Some((self.v - 35) / 2) };
```
bowenwang1996, line 156: is it somehow guaranteed that `self.v >= 35`?

bowenwang1996 (commented):
what's the reason for changing the type of `chain_id` from `u128` to `u64`?

runtime/near-evm-runner/src/runner.rs
ilblackdragon: because the attached deposit coming from the relayer. Instead value should be attached by the meta call itself, charging the user that signed meta tx.
I think we are missing the `value` field in the meta call and should not allow to attach funds by relayers

runtime/near-evm-runner/src/types.rs
ilblackdragon: the link above defines what `v` can be: 27/28 or 35 + chain_id*2 + {0, 1}

runtime/near-evm-runner/src/runner.rs
ailisp: parse_meta_call returns fee value as fee_amount now:
```
    Ok(MetaCallArgs { sender, nonce, fee_amount, fee_address, contract_address, input })

```

ailisp (approved):

ilblackdragon:
@bowenwang1996 because `chain_id` is u64: https://github.com/near/nearcore/blob/master/runtime/near-evm-runner/src/near_ext.rs#L74 
It was my original mistake to make it u128 everywhere in our code.

runtime/near-evm-runner/src/runner.rs
ilblackdragon: That's not the `value`.
`Value` is how much signer attached of base token with the call.
This considers attached deposit coming from relayer, which is incorrect.
We need to add another field to the meta call - `value` (or can call it attached_deposit or whatever) that will actually specify how much to deduct from given user (if they have enough balance).
See the implementation of raw_call for example of such deduction.

ailisp: Got it, some subword of `attached_(base_token)_deposit_value` :D

artob:
To be pedantic, _"Chain ID is a 256-bit value"_ per [EIP-1344](https://eips.ethereum.org/EIPS/eip-1344). However, in terms of OpenEthereum's APIs, they do only take a 64-bit value. And all [registered chain IDs](https://chainid.network) do fit within 64 bits, and certainly our own chain IDs do, so probably no good reason to need it be larger.

artob:
> @bowenwang1996 because `chain_id` is u64: https://github.com/near/nearcore/blob/master/runtime/near-evm-runner/src/near_ext.rs#L74
> It was my original mistake to make it u128 everywhere in our code.

To be pedantic, _"Chain ID is a 256-bit value"_ per [EIP-1344](https://eips.ethereum.org/EIPS/eip-1344). However, in terms of OpenEthereum's APIs, they do only take a 64-bit value. And all [registered chain IDs](https://chainid.network) do fit within 64 bits, and certainly our own chain IDs do, so probably no good reason to need it be larger.

artob (approved):
Looks really good overall, and significantly increases our out-of-the-box Eth1 compatibility.

My only comments are some nits that can be addressed in a later iteration.

artob:
@ilblackdragon @ailisp FYI, I've rebased this pull request against the current master branch just now.

runtime/near-evm-runner/src/types.rs
```diff
@@ -96,6 +102,124 @@ pub struct MetaCallArgs {
...
 
+#[derive(Debug, Eq, PartialEq)]
+pub struct EthTransaction {
+    pub nonce: U256,
+    pub gas_price: U256,
+    pub gas: U256,
+    pub to: Option<Address>,
+    pub value: U256,
+    pub data: Vec<u8>,
+}
```
evgenykuzyakov, line 113: Can you please add comments to the fields

runtime/near-evm-runner/src/types.rs
```diff
@@ -96,6 +102,124 @@ pub struct MetaCallArgs {
...
+    }
+}
+
+#[derive(Debug, Eq, PartialEq)]
+pub struct EthSignedTransaction {
+    pub transaction: EthTransaction,
+    pub v: u64,
+    pub r: U256,
+    pub s: U256,
+}
```
evgenykuzyakov, line 141: Add comments here

evgenykuzyakov (approved):
You need to bump EVM protocol version to hardfork betanet.

runtime/near-evm-runner/src/types.rs
artob: Resolved in 311b3c8ce1c18d1662e5549575f2b3ad8657a21a.

runtime/near-evm-runner/src/types.rs
artob: Resolved in 311b3c8ce1c18d1662e5549575f2b3ad8657a21a.

artob:
> You need to bump EVM protocol version to hardfork betanet.

I suggest we merge this at the same time as #3797 and thus only need to bump once (in that PR).

artob (approved):

artob:
@bowenwang1996 Need your review as code owner to be able to merge this.

artob:
> You need to bump EVM protocol version to hardfork betanet.

Done, just pushed now. @bowenwang1996 This is all good to go as far as @ilblackdragon and I are concerned, if you would have a chance to review.

bowenwang1996 (approved):

artob:
@olonho @willemneal Could we get a review from you guys as well, pretty please, to unblock merging this?

willemneal (approved):
LGTM just one logic question.

runtime/near-evm-runner/src/runner.rs
```diff
@@ -241,6 +253,63 @@ impl<'a> EvmContext<'a> {
...
+        }
+        // Retrieve the signer of given transaction.
+        let sender = match signed_transaction.sender() {
+            Some(sender) => sender,
+            None => return Err(VMLogicError::EvmError(EvmError::InvalidEcRecoverSignature)),
+        };
+        if let Some(receiver) = signed_transaction.transaction.to {
+            if signed_transaction.transaction.data.is_empty() {
+                // If data is empty, this is a simple transfer.
+                if let Some(receiver) = signed_transaction.transaction.to {
```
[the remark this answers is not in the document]
ilblackdragon: Good catch. An artifact of restructuring code.

PR #3775, opened 2021-01-04
[WIP] Evm phase2

evgenykuzyakov:


runtime/near-evm-runner/src/utils.rs
```diff
@@ -183,66 +187,497 @@ pub fn format_log(topics: Vec<H256>, data: &[u8]) -> std::result::Result<Vec<u8>
     Ok(result)
 }
 
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum Type {
```
evgenykuzyakov, line 191: What is this type and why is it public?
 - Add a comment
 - Rename to something meaningful or wrap into a module, so it has a context.
 - Consider making it non public visibility.

runtime/near-evm-runner/src/utils.rs
```diff
@@ -183,66 +187,497 @@ pub fn format_log(topics: Vec<H256>, data: &[u8]) -> std::result::Result<Vec<u8>
...
+    String,
+    Bool,
+    Bytes,
+    Byte(u8),
+    Custom(String),
+    Array { length: Option<u64>, inner: Box<Type> },
+}
+
+/// the type string is being validated before it's parsed.
+pub fn parse_type(field_type: &str) -> Result<Type> {
```
evgenykuzyakov, line 204: What's the input format?

runtime/near-evm-runner/src/utils.rs
```diff
@@ -183,66 +187,497 @@ pub fn format_log(topics: Vec<H256>, data: &[u8]) -> std::result::Result<Vec<u8>
...
+        };
+
+        token = Some(type_);
+        lexer.advance();
+    }
+
+    Ok(token.ok_or(VMLogicError::EvmError(EvmError::InvalidMetaTransactionMethodName))?)
+}
+
 pub fn near_erc721_domain(chain_id: U256) -> RawU256 {
```
evgenykuzyakov, line 270: Add a comment

runtime/near-evm-runner/src/utils.rs
```diff
@@ -183,66 +187,497 @@ pub fn format_log(topics: Vec<H256>, data: &[u8]) -> std::result::Result<Vec<u8>
...
+    bytes.extend_from_slice(&near);
+    let version = b"1";
+    let version: RawU256 = keccak(&version).into();
+    bytes.extend_from_slice(&version);
     bytes.extend_from_slice(&u256_to_arr(&chain_id));
     keccak(&bytes).into()
 }
 
-pub fn method_name_to_rlp(method_name: String) -> [u8; 4] {
+pub fn method_sig_to_abi(method_sig: &str) -> [u8; 4] {
```
evgenykuzyakov, line 285: Add a comment. Explain the input format.

runtime/near-evm-runner/src/utils.rs
```diff
@@ -183,66 +187,497 @@ pub fn format_log(topics: Vec<H256>, data: &[u8]) -> std::result::Result<Vec<u8>
...
+
+pub fn encode_string(s: &str) -> Vec<u8> {
+    let mut bytes = vec![];
+    bytes.extend_from_slice(&keccak(s.as_bytes()).as_bytes());
+    bytes
+}
+
+#[derive(Debug, Eq, PartialEq)]
+pub enum Value {
+    Bytes(Vec<u8>),
```
evgenykuzyakov, line 305: Non-descriptive name. Consider rename and add a comment.

runtime/near-evm-runner/src/utils.rs
```diff
@@ -183,66 +187,497 @@ pub fn format_log(topics: Vec<H256>, data: &[u8]) -> std::result::Result<Vec<u8>
...
+        if rlp.is_list() {
+            Ok(Value::List(rlp.as_list()?))
+        } else {
+            Ok(Value::Bytes(rlp.decoder().decode_value(|bytes| Ok(bytes.to_vec()))?))
+        }
+    }
+}
+
+#[derive(Debug, Eq, PartialEq)]
+pub struct Arg {
```
evgenykuzyakov, line 320: ditto

runtime/near-evm-runner/src/utils.rs
```diff
@@ -183,66 +187,497 @@ pub fn format_log(topics: Vec<H256>, data: &[u8]) -> std::result::Result<Vec<u8>
...
+    #[allow(dead_code)]
+    pub name: String,
+    pub type_raw: String,
+    pub t: Type,
+}
+
+#[derive(Debug, Eq, PartialEq)]
+pub struct Method {
```
evgenykuzyakov, line 328: ditto

runtime/near-evm-runner/src/utils.rs
```diff
@@ -183,66 +187,497 @@ pub fn format_log(topics: Vec<H256>, data: &[u8]) -> std::result::Result<Vec<u8>
...
+    pub name: String,
+    pub raw: String,
+    pub args: Vec<Arg>,
+}
+
+#[derive(Debug, Eq, PartialEq)]
+pub struct MethodAndTypes {
```
evgenykuzyakov, line 335: same. It would make sense to put it all into a mod, if it has some context. E.g.:
```
mod MetaParsing {
pub struct MethodAndTypes
...
}

File: runtime/near-evm-runner/src/utils.rs [hunk 9]
```diff
@@ -183,66 +187,497 @@ pub fn format_log(topics: Vec<H256>, data: &[u8]) -> std::result::Result<Vec<u8>
...
+    for c in chars {
+        if !is_arg_char(c) {
+            break;
+        }
+        i += 1;
+    }
+    Ok((text[..i].to_string(), &text[i..]))
+}
+
+fn parse_type_raw(text: &str) -> Result<(String, &str)> {
```
evgenykuzyakov, line 418: Need a comment and a hint towards a format

runtime/near-evm-runner/src/utils.rs
```diff
@@ -183,66 +187,497 @@ pub fn format_log(topics: Vec<H256>, data: &[u8]) -> std::result::Result<Vec<u8>
...
+    let i =
+        text.find(' ').ok_or(VMLogicError::EvmError(EvmError::InvalidMetaTransactionMethodName))?;
+    Ok((text[..i].to_string(), &text[i..]))
```
evgenykuzyakov, line 421: Looks like a `split` method

runtime/near-evm-runner/src/utils.rs
```diff
@@ -183,66 +187,497 @@ pub fn format_log(topics: Vec<H256>, data: &[u8]) -> std::result::Result<Vec<u8>
...
+}
+
+fn consume(text: &str, c: char) -> Result<&str> {
```
evgenykuzyakov, line 424: need a comment

runtime/near-evm-runner/src/utils.rs
```diff
@@ -183,66 +187,497 @@ pub fn format_log(topics: Vec<H256>, data: &[u8]) -> std::result::Result<Vec<u8>
...
+    Ok(&text[1..])
+}
+
+fn is_arg_start(c: char) -> bool {
+    c.is_ascii_alphabetic() || c == '_'
+}
+
+fn is_arg_char(c: char) -> bool {
+    c.is_ascii_alphanumeric() || c == '_'
+}
```
evgenykuzyakov, line 439: comments

runtime/near-evm-runner/src/utils.rs
```diff
@@ -183,66 +187,497 @@ pub fn format_log(topics: Vec<H256>, data: &[u8]) -> std::result::Result<Vec<u8>
...
+
+/// Return a signature of the method_def
+/// E.g. method_signature(Methods before parse: "adopt(uint256 petId)") -> "adopt(uint256)"
+fn methods_signature(method_and_type: &MethodAndTypes) -> String {
+    let method = format!(
+        "{}({})",
+        method_and_type.method.name,
+        itertools::join(method_and_type.method.args.iter().map(|arg| arg.type_raw.as_str()), ",")
+    );
```
evgenykuzyakov, line 448: This can be moved below `types` to avoid extra concatenation

runtime/near-evm-runner/src/utils.rs
```diff
@@ -183,66 +187,497 @@ pub fn format_log(topics: Vec<H256>, data: &[u8]) -> std::result::Result<Vec<u8>
...
+                    ","
+                )
+            )
+        })
+        .collect();
+    method + &types
+}
+
+#[derive(Debug, Clone)]
+pub struct ParseMethodNameError;
```
evgenykuzyakov, line 473: comment

runtime/near-evm-runner/src/utils.rs
```diff
@@ -183,66 +187,497 @@ pub fn format_log(topics: Vec<H256>, data: &[u8]) -> std::result::Result<Vec<u8>
...
+                    element,
+                    types,
+                )?);
+            }
+            Ok(keccak(r).as_bytes().to_vec())
+        }),
+    }
+}
+
+fn eip_712_rlp_list<F>(value: &Value, f: F) -> Result<Vec<u8>>
```
evgenykuzyakov, line 525: comment

runtime/near-evm-runner/src/utils.rs
```diff
@@ -183,66 +187,497 @@ pub fn format_log(topics: Vec<H256>, data: &[u8]) -> std::result::Result<Vec<u8>
...
+where
+    F: Fn(&Vec<Value>) -> Result<Vec<u8>>,
+{
+    match value {
+        Value::Bytes(_) => Err(VMLogicError::EvmError(EvmError::InvalidMetaTransactionFunctionArg)),
+        Value::List(l) => f(l),
+    }
+}
+
+fn eip_712_rlp_value<F>(value: &Value, f: F) -> Result<Vec<u8>>
```
evgenykuzyakov, line 535: comment

evgenykuzyakov (commented):
@ailisp

runtime/near-evm-runner/src/utils.rs
ailisp: I think it's better to name to ArgType or EvmArgType, make private. it is "Type" and public because it's copied from eip-712 crate :D

runtime/near-evm-runner/src/utils.rs
ailisp: A single evm function arg type in string, without the argument name. e.g. `"bytes"` `"uint256[][3]"` `"CustomStructName"`

runtime/near-evm-runner/src/utils.rs
ailisp: seem many comments here, i'll address them in a PR

runtime/near-evm-runner/src/utils.rs
ailisp: split doesn't work because want only split on first space

runtime/near-evm-runner/src/utils.rs
ailisp: It need to be pub to construct and compare in test. pub(crate) maybe better, but we can simply make the whole utils mod private to near-evm-runner

runtime/near-evm-runner/src/utils.rs
birchmd: I think `splitn` is what you are looking for, it allows limiting how many sub-strings are returned.

https://doc.rust-lang.org/std/primitive.str.html#method.splitn

ailisp:
I just learned splitn, thanks! but it still has a problem of allowing "up
to n" not "exactly n" so it isn't easier than find and split (need add
```
if splitted.len() != 2 {
  return Err(...)
}
```

runtime/near-evm-runner/src/utils.rs
ailisp: I just learned splitn, thanks! but it still has a problem of allowing "up
to n" not "exactly n" so it isn't easier than find and split (need add
```
if splitted.len() != 2 {
  return Err(...)
}
```

birchmd: Right, that makes sense. You could also do something like

```
let mut iter = text.splitn(2, ' ');
let (prefix, suffix) = iter.next().and_then(|p| iter.next().map(|s| (p, s))).ok_or(...);
Ok((prefix.to_string(), suffix))
```

But I don't think that is cleaner than your solution with `find`; the mapping over options might even be less clear.

As a side note there is `split_once`, but it is not yet stabilized and probably a small thing like this is not worth adding a nightly feature over.
https://doc.rust-lang.org/std/primitive.str.html#method.split_once

artob (approved):

PR #3785, opened 2021-01-06
Pin the syn dependency to avoid breaking change

artob:
[syn](https://crates.io/crates/syn) [1.0.58](https://crates.io/crates/syn/1.0.58), released yesterday (January 5), breaks our build despite it ostensibly being a semver-compatible update:

Screen Shot 2021-01-06 at 19 25 47

To reproduce the problem, remove your top-level `Cargo.lock` and attempt to rebuild nearcore from scratch.

See https://github.com/nushell/nushell/issues/2867 and https://github.com/nushell/nushell/pull/2868 for another project who already ran into this same trouble.

This pull request pins syn to 1.0.57 for now. The pin can be removed once upstream fixes their broken release management process.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -144,7 +144,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "b4ca8ce00b267af8ccebbd647de0d61e0674b6e61185cc7a592ff88772bed655"
 dependencies = [
  "quote 1.0.7",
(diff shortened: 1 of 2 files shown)


birchmd (approved):
Yes, I saw this issue in the Mocknet tests as well.

https://nayduckdiag.blob.core.windows.net/logs/69828_v2_stderr

PR #3787, opened 2021-01-06
fix: ensure stable toolchain is added for docker images

willemneal:


stable.rust-toolchain
```diff
@@ -0,0 +1 @@
+stable-2020-10-08
```
bowenwang1996, line 1: The naming convention of this file feels a bit strange to me, but maybe there is some convention that I am not aware of?

bowenwang1996 (approved):

stable.rust-toolchain
willemneal: It was the same convention used in the current `rust-toolchain` file.  Instead of the version it's the date of the release.

chefsale (approved):

Issue #3488, opened 2020-10-14
Implement Epoch Sync


SkidanovAlex:
This issue tracks adding support for light speed sync.

Kouprin:
@bowenwang1996 there is discussion I've had with @SkidanovAlex and want to share and synchronize with you our next steps.

-- TL;DR starts here --

This Tuesday I suggested to have Epoch Sync at the start + passive GC headers concept. It resolves two main tasks we want to achieve:
1. making syncing procedure fast;
2. clearing headers to control amount of memory / storage required for them.

To obtain that, I implemented the invariant: "store the headers from Epoch Sync up to the Head". However, Alex proposes different one: "store the headers from the Tail up to the Head". The motivation here is to reach maximum reliability on headers absence. When we have headers starting from Epoch Sync, we have most of the necessary headers all the time - and what's why it will be extremely hard to catch and reproduce any issues related to headers absence. Another thing about it, the invariant I've suggested is too broad and allows different nodes to use very different amount of storage / memory under the same invariant.

Alex's suggestion, basically, means enabling GC headers in realtime + has several important implications.
1. It doesn't lead us to Syncing state machine simplification that was proposed originally with Epoch Sync design. Instead, it needs to update dangerous method reset_data_pre_state_sync (and maybe some more) and keep having deal with it.
2. As far we can't go to the Syncing state machine simplification, there is no actual reason to run Epoch Sync at node start only. Instead, we can run Epoch Sync every time when we need to run State Sync. This is very convenient for our validators because they will not be required to restart their nodes to run Epoch Sync.

In terms of business decisions, I'm okay with both solutions. Unfortunately, there is no way to combine them in improving manner - we have to decide which approach we should use.

-- TL;DR ends here --

Personally, I trust Alex's vision and experience, that's why I prefer to follow his idea. However, it also means that I need to return to redesign / implementation status in my current work. In this case all approximate dates that I brought to you will be shifted.

bowenwang1996:
@Kouprin @SkidanovAlex I would really love to get rid of `reset_data_pre_state_sync`, but I also agree with the invariant that Alex proposed. Could we reset tail to epoch sync start and delete everything in between instead?

Kouprin:
@bowenwang1996 If you want to simply get rid of `reset_data_pre_state_sync`, please explain how you see the new policy of switching to the State Sync. Originally, we planned to disallow to run State Sync in the state machine more than once.

[...]

PR #3804, opened 2021-01-14
Adding block_ordinal to BlockHeaderInnerRest

Kouprin:
This change is necessary for enabling Epoch Sync https://github.com/near/nearcore/issues/3488.

It creates BlockHeaderV3 and BlockHeaderInnerRestV3 and a nightly `protocol_feature_block_ordinal` feature. The actual Block Ordinal of the new Block is (the amount of Blocks already existing on the Canonical Chain + 1) which is taken from Bowen-Merkle Trees.

Tested with `test_block_ordinal` sanity test and CI.


Kouprin:
`test_cache_invalidation` passes locally. Is github cargo-test configured specifically for current version with not including upgrade?

core/primitives/src/version.rs
```diff
@@ -52,6 +52,8 @@ pub const DELETE_KEY_STORAGE_USAGE_PROTOCOL_VERSION: ProtocolVersion = 40;
 
 pub const SHARD_CHUNK_HEADER_UPGRADE_VERSION: ProtocolVersion = 41;
 
+pub const BLOCK_ORDINAL_UPGRADE_VERSION: ProtocolVersion = 45;
```
frol, line 55: @bowenwang1996 what is the current process for the protocol features? Should all of them go through nightly? This version 45 while the stable is 41, and the latest nightly is 47 looks off to me.

frol (approved):
view-client and primitives look good to me :+1:

chain/chain/src/chain.rs
```diff
@@ -3199,10 +3199,6 @@ impl<'a> ChainUpdate<'a> {
             return Err(ErrorKind::InvalidChunkMask.into());
         }
 
-        if !header.verify_chunks_included() {
-            return Err(ErrorKind::InvalidChunkMask.into());
```
bowenwang1996, line 3203: why is this removed?

bowenwang1996 (changes requested):
New features should go through nightly first. See https://github.com/near/NEPs/issues/126#issuecomment-712130812 for the idea and you can take a look at https://github.com/near/nearcore/blob/191559126134075b3d5476410ab061b0f6b8509d/core/primitives/src/version.rs#L78 to see how it is done

chain/chain/src/chain.rs
Kouprin: I wrongly assumed this check is not necessary because it's always true. Reverted.

core/primitives/src/version.rs
Kouprin: Added the feature to nightly.

core/primitives/src/version.rs
```diff
@@ -173,3 +177,42 @@ macro_rules! checked_feature {
         }
     }};
 }
+
+#[macro_export]
+macro_rules! checked_feature_crate {
+    ($feature_name:tt, $feature:ident, $current_protocol_version:expr) => {{
+        #[cfg(feature = $feature_name)]
+        let is_feature_enabled = crate::version::PROTOCOL_FEATURES_TO_VERSION_MAPPING
+            [&crate::version::ProtocolFeature::$feature]
```
frol, line 186: Is the only change here `near_primitives::` being renamed into `crate::`? What is the issue you have faced with the original macro?

Kouprin: It has no `near_primitives::` and unable to compile.

frol: I see. We need to use `$crate` instead of `near_primitives`: https://doc.rust-lang.org/1.5.0/book/macros.html#the-variable-crate

Kouprin: It seems $crate works perfectly. Thank you for the insight. Updated.

frol (approved):

core/primitives/src/block_header.rs
```diff
@@ -434,6 +543,36 @@ impl BlockHeader {
...
+        } else {
+            let inner_rest = BlockHeaderInnerRestV3 {
+                chunk_receipts_root,
+                chunk_headers_root,
+                chunk_tx_root,
+                challenges_root,
+                random_value: CryptoHash::default(),
+                validator_proposals: vec![],
+                chunk_mask: vec![true; chunks_included as usize],
+                block_ordinal: 1, // TODO MOO Bowen please make sure genesis is block #1 in terms of Merkle Proofs
```
bowenwang1996, line 555: this is the case

core/primitives/src/block_header.rs
```diff
@@ -563,6 +719,16 @@ impl BlockHeader {
...
             BlockHeader::BlockHeaderV1(header) => &header.inner_rest.chunk_mask,
             BlockHeader::BlockHeaderV2(header) => &header.inner_rest.chunk_mask,
+            BlockHeader::BlockHeaderV3(header) => &header.inner_rest.chunk_mask,
+        }
+    }
+
+    #[inline]
+    pub fn block_ordinal(&self) -> NumBlocks {
+        match self {
+            BlockHeader::BlockHeaderV1(_) => 0, // not applicable
```
bowenwang1996, line 729: Should we return an option then? It is better to take advantage of the rust type system than to rely on some hardcoded constant.

core/primitives/src/views.rs
```diff
@@ -415,6 +416,7 @@ pub struct BlockHeaderView {
     pub random_value: CryptoHash,
     pub validator_proposals: Vec<ValidatorStakeView>,
     pub chunk_mask: Vec<bool>,
+    pub block_ordinal: NumBlocks,
```
bowenwang1996, line 419: this should ideally also guarded by the feature flag. While not doing so has no impact on the protocol level, rpc users may get confused as to what this is.

core/primitives/src/block_header.rs
Kouprin: I've updated the comment.

core/primitives/src/block_header.rs
Kouprin: We can do that.
However, BlockHeader is very basic primitive that is used everywhere in the code. Once we use Option<NumBlocks> for small piece of semantical sugar here, we will have to unwrap it everywhere for forever. I wouldn't go this way as it may create tons of `match ...` and `unwrap()` code later with no actual purpose.

core/primitives/src/views.rs
Kouprin: Updated.

core/primitives/src/block_header.rs
Kouprin: We can introduce Option wrapper for block_ordinal in BlockHeaderView only. It makes sense to me. Should we go this way?

bowenwang1996: Yeah let's do that

bowenwang1996 (approved):

pytest/lib/messages/block.py
```diff
@@ -266,6 +286,28 @@ class ApprovalInner:
             ]
         }
     ],
+    [
+        BlockHeaderInnerRestV2, {
```
Kouprin, line 290: This is a typo.

Issue #3788, opened 2021-01-07
[Tracking Issue] Review the error handling infrastructure


frol:
*This is a tracking issue preparing to dive into #3517.*

* [x] Learn best practices on error handling (`thiserror` is the best crate I know that fits our needs)
* [x] Review the PR #3204 that we need to "rebase" onto the latest master

khorolets:
@frol I guess we should also add something like "come up with design of error structs" for view client

frol:
@khorolets I guess, we can try to copy what Bowen has already done unless we find a better option during review. Ideally, every function should return only what it can potentially return, meaning that there should not be a huge enum with all possible errors in the world, instead, there should be granular types (it can even be just a struct, not an enum, e.g. in Rust standard library there is [`AddrParseError`](https://doc.rust-lang.org/stable/std/net/struct.AddrParseError.html), which is `struct AddrParseError(())`)

@bowenwang1996 I would like to ensure exhautiveness of the potential error responses. How realistic is this approach in your opinion in regarts to the core?

bowenwang1996:
> How realistic is this approach in your opinion in regarts to the core?

It may be possible, but will likely take several months to complete

PR #3805, opened 2021-01-14
refactor(jsonrpc): Initial refactoring for Structured Error handling for JSON RPC and ViewClient

khorolets:
Start of #3788 
It's just a beginning of refactoring. @frol and I decided that having error structures for each handler would be better solution. This is a first try with `GetBlock` handler, which doesn't change the response from jsonrpc.

To start it, I've done:
* Extract types from `chain/client` to separate crate `client-primitives`
* Extract `messages.rs` from `chain/jsonrpc/client` to separate `jsonrpc-client-primitives` crate
* Refactor `GetBlock` handler to have `GetBlockError`

Work in progress.

@frol I'd like to know what should be changed in `GetBlockError` to make it better

P.S. Commented code in the commit will be cleared while I move further.


chain/client-primitives/src/types.rs
```diff
@@ -150,14 +150,21 @@ impl SyncStatus {
 /// Actor message requesting block by id or hash.
 pub struct GetBlock(pub BlockReference);
 
+pub enum GetBlockError {
+    ChainError(near_chain::Error),
+    ViewClientError(Error),
+    InvalidParams(String),
+    Other(String),
```
frol, line 157: This is the first iteration to make sure it is compilable, right?

chain/client-primitives/Cargo.toml
```diff
@@ -0,0 +1,19 @@
...
+[dependencies]
+actix = "0.9"
+chrono = { version = "0.4.4", features = ["serde"] }
+serde = { version = "1", features = ["derive"] }
+strum = { version = "0.18", features = ["derive"] }
+
+near-chain = { path = "../chain" }
+near-chain-primitives = { path = "../chain-primitives" }
+near-chunks = { path = "../chunks" }
+near-network = { path = "../network" }
```
frol, line 18: Ideally, we don't want to depend on non-primitive dependencies.

Can you review the usage and consider moving certain parts around (e.g. if we moved something unnecessary to the *-primitives, and would be better to leave it in the original place)?

chain/jsonrpc-primitives/src/blocks.rs
```diff
@@ -0,0 +1,43 @@
...
+            }
+            near_client_primitives::types::GetBlockError::NotSyncedYet => {
+                RpcBlockError::NotSyncedYet
+            }
+            near_client_primitives::types::GetBlockError::IOError(s)
+            | near_client_primitives::types::GetBlockError::Unexpected(s) => {
+                near_metrics::inc_counter_vec(
+                    &crate::metrics::RPC_UNEXPECTED_ERROR_COUNT,
+                    &["RpcBlockError", &s],
+                );
```
frol, line 32: This metric should only be triggered on "Unexpected" variant. Also, we need to log it to be able to investigate the specific case since just a number is not useful (unless we confirm that it is always zero).

chain/jsonrpc-primitives/src/blocks.rs
```diff
@@ -0,0 +1,43 @@
...
+    #[error("Block `{0}` is missing")]
+    BlockMissing(near_primitives::hash::CryptoHash),
+    #[error("Block not found")]
+    BlockNotFound(String),
+    #[error("There are no fully synchronized blocks yet")]
+    NotSyncedYet,
+    #[error("Too many requests. Try again later: {0}")]
+    TooManyRequests(actix::MailboxError),
+    #[error("Unexpected error occurred: {0}")]
+    Unexpected(String),
```
frol, line 12: I suggest the following renaming:

* `Unexpected` -> `Unreachable` - this is the common term in Rust, and we should indicate in the error message that we would love to be informed about the error if it happens (e.g. if it only happens on someone else's instance, where we don't collect metrics), so the error message could be something along these lines: "It is a bug if you receive this error type, please, report this incident: https://github.com/near/nearcore/issues/new/choose"
* `TooManyRequests` -> `InternalError` with a help message: "The node reached its limits. Try again later. More details: {0}"

chain/jsonrpc-primitives/src/message.rs
```diff
@@ -130,6 +130,39 @@ impl std::string::ToString for RpcError {
...
+
+        Self::new(-32_000, "Server error".to_string(), error_data)
+    }
+}
+
+impl From<actix::MailboxError> for RpcError {
+    fn from(error: actix::MailboxError) -> Self {
+        Self::new(-32_000, "Server error".to_string(), Some(Value::String(error.to_string())))
+    }
+}
```
frol, line 158: Do we need it?

chain/jsonrpc-primitives/src/metrics.rs
```diff
@@ -0,0 +1,11 @@
+use lazy_static::lazy_static;
+use near_metrics::IntCounterVec;
+
+lazy_static! {
+    pub static ref RPC_UNEXPECTED_ERROR_COUNT: near_metrics::Result<IntCounterVec> =
+        near_metrics::try_create_int_counter_vec(
+            "near_rpc_unexpected_errors_total",
```
frol, line 7: note: `unreachable`

chain/jsonrpc/src/lib.rs
```diff
@@ -649,12 +653,15 @@ impl JsonRpcHandler {
...
-            .map_err(|err| RpcError::server_error(Some(err.to_string())))?
-            .map_err(|err| RpcError::server_error(Some(err)))?;
+        // TODO refactor it. Changed to keep it working before refactoring
+        let result = self.view_client_addr.send(GetBlock(block_reference)).await?;
+
+        if result.is_err() {
+            return Err(RpcError::from(RpcBlockError::from(result.err().unwrap())));
+        }
+
+        let block = result.unwrap();
```
frol, line 663:
```suggestion
        let block = match result {
            Ok(block) => block,
            Err(err) => {
                return Err(RpcError::from(RpcBlockError::from(err)));
            }
        };
```

chain/jsonrpc/src/lib.rs
```diff
@@ -667,12 +674,14 @@ impl JsonRpcHandler {
...
-            .await
-            .map_err(|err| RpcError::server_error(Some(err.to_string())))?
-            .map_err(|err| RpcError::server_error(Some(err)))?;
+        // TODO refactor it. Changed to keep it working before refactoring
+        let result = self.view_client_addr.send(GetBlock(block_reference)).await?;
+        if result.is_err() {
+            return Err(RpcError::from(RpcBlockError::from(result.err().unwrap())));
+        }
+
+        let block = result.unwrap();
```
frol, line 683: ^

core/primitives/src/rpc.rs
```diff
@@ -12,6 +12,38 @@ use crate::views::{
     ExecutionOutcomeWithIdView, LightClientBlockLiteView, QueryRequest, StateChangeWithCauseView,
     StateChangesKindsView, StateChangesRequestView,
 };
+use serde::de::DeserializeOwned;
+use serde_json::Value;
+
+#[derive(Serialize)]
+pub struct RpcParseError(pub String);
```
frol, line 19: Let's extract this file from core primitives and merge it into `near-jsonrpc-primitives`

chain/jsonrpc/src/lib.rs
```diff
@@ -621,13 +627,11 @@ impl JsonRpcHandler {
...
-        let block_reference = if let Ok((block_id,)) = parse_params::<(BlockId,)>(params.clone()) {
-            BlockReference::BlockId(block_id)
-        } else {
-            parse_params::<BlockReference>(params)?
-        };
-        jsonify(self.view_client_addr.send(GetBlock(block_reference)).await)
+    async fn block(
+        &self,
+        request_data: near_primitives::rpc::RpcBlockRequest,
+    ) -> Result<BlockView, RpcBlockError> {
```
frol, line 633: And the final step would be to use `RpcBlockResponse`, which will wrap `BlockView` (flatten). :pray:

chain/client-primitives/src/types.rs
```diff
@@ -150,14 +150,42 @@ impl SyncStatus {
...
+
+impl From<near_chain_primitives::Error> for GetBlockError {
+    fn from(error: near_chain_primitives::Error) -> GetBlockError {
+        match error.kind() {
+            near_chain_primitives::ErrorKind::IOErr(s) => GetBlockError::IOError(s),
+            near_chain_primitives::ErrorKind::DBNotFoundErr(s) => GetBlockError::BlockNotFound(s),
+            near_chain_primitives::ErrorKind::BlockMissing(hash) => {
+                GetBlockError::BlockMissing(hash)
+            }
+            // TODO: add metrics here
```
frol, line 175:
```suggestion
```

frol (changes requested):

chain/jsonrpc-primitives/src/message.rs
khorolets: For a while. Will be possible to delete after all refactoring.

chain/jsonrpc-primitives/src/blocks.rs
```diff
@@ -0,0 +1,67 @@
...
+                RpcBlockError::BlockMissing(block_hash)
+            }
+            near_client_primitives::types::GetBlockError::BlockNotFound(s) => {
+                RpcBlockError::BlockNotFound(s)
+            }
+            near_client_primitives::types::GetBlockError::NotSyncedYet => {
+                RpcBlockError::NotSyncedYet
+            }
+            near_client_primitives::types::GetBlockError::IOError(s) => {
+                RpcBlockError::Unreachable(s)
```
frol, line 37: This should be just `InternalError`. I suggest we use `String` inside the `InternalError` since there is no good in storing actix error inside anyway.

chain/jsonrpc-primitives/src/metrics.rs
```diff
@@ -0,0 +1,11 @@
+use lazy_static::lazy_static;
+use near_metrics::IntCounterVec;
+
+lazy_static! {
+    pub static ref RPC_UNREACHABLE_ERROR_COUNT: near_metrics::Result<IntCounterVec> =
+        near_metrics::try_create_int_counter_vec(
+            "near_rpc_unreachable_errors_total",
+            "Total count of Unexpected RPC errors returned, by method",
```
frol, line 8: unreachable

core/primitives/src/lib.rs
```diff
@@ -1,12 +1,12 @@
 #[cfg(jemallocator)]
 extern crate jemallocator;
 
+pub use borsh;
+
```
frol, line 5: Why is this change?

frol (changes requested):

frol (approved):
I unblock this PR. The minor changes are agreed to be addressed, so other code owners need to review.

Review hits:

* The changes mostly target `chain/jsonrpc` (@frol and @khorolets are code-owners, so we review that)
* The rest of the changes are about extracting modules into separate crates to avoid fat dependencies, fat primitives, and circular dependencies

chain/jsonrpc-primitives/src/types/blocks.rs
```diff
@@ -0,0 +1,99 @@
...
+    #[error("Block `{0}` is missing")]
+    BlockMissing(near_primitives::hash::CryptoHash),
+    #[error("Block not found")]
+    BlockNotFound(String),
+    #[error("There are no fully synchronized blocks yet")]
+    NotSyncedYet,
+    #[error("The node reached its limits. Try again later. More details: {0}")]
+    InternalError(String),
+    #[error("It is a bug if you receive this error type, please, report this incident: https://github.com/near/nearcore/issues/new/choose. Details: {0}")]
+    Unreachable(String),
```
frol, line 15: Sorry for getting here once again, but let's also clarify why it is a bug, and why it should be unreachable.

```suggestion
    // NOTE: Currently, the underlying errors are too broad, and while we tried to handle
    // expected cases, we cannot statically guarantee that no other errors will be returned
    // in the future.
    // TODO #3851: Remove this variant once we can exhaustively match all the underlying errors
    #[error("It is a bug if you receive this error type, please, report this incident: https://github.com/near/nearcore/issues/new/choose. Details: {0}")]
    Unreachable(String),
```

frol (approved):

frol (approved):

khorolets:
@bowenwang1996 could you please take a look at this PR? Thank you.

chain/client-primitives/src/types.rs
```diff
@@ -150,14 +150,41 @@ impl SyncStatus {
...
+    #[error("IO Error: {0}")]
+    IOError(String),
+    #[error("Block `{0}` is missing")]
+    BlockMissing(CryptoHash),
+    #[error("Block not found")]
+    BlockNotFound(String),
+    #[error("There are no fully synchronized blocks yet")]
+    NotSyncedYet,
+    #[error("Unexpected error occurred: {0}")]
+    Unexpected(String),
```
bowenwang1996, line 164: Why do we need `Unexpected`?

chain/jsonrpc/src/lib.rs
```diff
@@ -621,13 +628,11 @@ impl JsonRpcHandler {
...
     }
 
-    async fn block(&self, params: Option<Value>) -> Result<Value, RpcError> {
-        let block_reference = if let Ok((block_id,)) = parse_params::<(BlockId,)>(params.clone()) {
-            BlockReference::BlockId(block_id)
-        } else {
-            parse_params::<BlockReference>(params)?
-        };
-        jsonify(self.view_client_addr.send(GetBlock(block_reference)).await)
+    async fn block(
```
bowenwang1996, line 631: why is this method refactored but not other methods?

bowenwang1996 (approved):

chain/client-primitives/src/types.rs
frol: @khorolets let's rename it to `Unreachable` as well and add the comment as we did for the `RpcBlockError::Unreachable`.

chain/jsonrpc/src/lib.rs
frol: @bowenwang1996 It is to minimize the scope. I asked @khorolets to start with a single method to justify the refactoring (extracting parts into separate *-primitives crates), and then we can purely focus on jsonrpc refactoring scaling the work of refactoring the methods and also paving the road for getting rid of `Unreachable` error states (see #3851)

chain/jsonrpc-primitives/src/errors.rs
```diff
@@ -0,0 +1,77 @@
...
+    pub fn method_not_found(method: String) -> Self {
+        RpcError::new(-32_601, "Method not found".to_owned(), Some(Value::String(method)))
+    }
+}
+
+impl std::string::ToString for RpcError {
+    fn to_string(&self) -> String {
+        format!("{:?}", self)
+    }
+}
```
miraclx, line 65: Any reason why this isn't `fmt::Display`?

khorolets: No reason if I recall correct

miraclx: Resolved in #4772

PR #3806, opened 2021-01-14
feat: add `merge_import` rule to rustfmt.toml

willemneal:
This change will make it easier to read over imports in a given file.

When doing work on dependency reorg this was very useful.


willemneal:
While changing the formatting rules, what are people's thoughts are forcing all imports to be their own line regardless of length?
[See `import_layout`](https://github.com/rust-lang/rustfmt/blob/master/Configurations.md#imports_layout), 

For example with `HorizontalVertical`:

```rust
use vm::{
    ActionParams, ActionType, ActionValue, ContractCreateResult, ExecTrapResult, Ext, GasLeft,
    MessageCallResult, ParamsType, ReturnData, Schedule,
};
```

becomes

```rust 
use vm::{
    ActionParams,
    ActionType,
    ActionValue,
    ContractCreateResult,
    ExecTrapResult,
    Ext,
    GasLeft,
    MessageCallResult,
    ParamsType,
    ReturnData,
    Schedule,
};
```

While this does improve readability, it increases number of lines.   What is your opinion?

frankbraun:
> This change will make it easier to read over imports in a given file.
> 
> When doing work on dependency reorg this was very useful.

This seems like a good change. [merge_import](https://github.com/rust-lang/rustfmt/blob/master/Configurations.md#merge_imports) is deprecated, though. Why not use `imports_granularity = "Crate"` instead.?

Also, did you try `imports_granularity = "Module"` instead? See [imports_granularity](https://github.com/rust-lang/rustfmt/blob/master/Configurations.md#imports_granularity). I think this might lead to better results.

frankbraun:
> While changing the formatting rules, what are people's thoughts on forcing all imports to be their own line regardless of length?

I would stick with the default here. That is, not forcing all imports to be on their own line.

frankbraun (approved):

chain/chain/src/chain.rs
```diff
@@ -1,58 +1,60 @@
-use std::collections::{HashMap, HashSet};
-use std::sync::Arc;
-use std::time::{Duration as TimeDuration, Instant};
+use std::{
+    collections::{HashMap, HashSet},
```
ailisp, line 2: Should this also become
```
use std::{
    collections::{
        HashMap,
        HashSet,
    }
}
```

ailisp (approved):

chain/chain/src/chain.rs
willemneal: That's only if I apply the second rule `imports_layout`.

willemneal:
@frankbraun good catch on the deprecation.  However, when I tired to replace the rule I get: 
```
Warning: Unknown configuration option `imports_granularity`
```

Also looking at the different options `Crate` seems like the best.

chain/chain/src/chain.rs
frol: I beg not to do it! It is especially ugly for `std` imports. There was a discussion recently on reddit about it: https://www.reddit.com/r/rust/comments/kmw6xc/code_style_question_grouping_of_use_statements/

frol: My rational is: it is easier to read a single line than to follow multiline statements, and this way you hit merged conflicts less frequently.

frol:
`imports_granularity = "Module"` looks better to me, but I would still avoid any changes while the options are unstable. Also, I prefer the default formatting and avoid controversial conversations (people have different tastes, so let's just stick to the most common - the default).

frankbraun:
> @frankbraun good catch on the deprecation. However, when I tired to replace the rule I get:
> 
> ```
> Warning: Unknown configuration option `imports_granularity`
> ```

Looks like I mistakenly assumed the documentation is correct, I get the same error. LGTM then as is.


> Also looking at the different options `Crate` seems like the best.

OK.

willemneal:
> Also, I prefer the default formatting and avoid controversial conversations (people have different tastes, so let's just stick to the most common - the default).

@frol Does this mean you think this PR shouldn't happen?  Personally don't have much skin in the Rust format game, but this change makes a large project like this much cleaner.

artob:
> While changing the formatting rules, what are people's thoughts on forcing all imports to be their own line regardless of length? [...]
> 
> For example with `HorizontalVertical`:
> 
> ```rust
> use vm::{
>     ActionParams, ActionType, ActionValue, ContractCreateResult, ExecTrapResult, Ext, GasLeft,
>     MessageCallResult, ParamsType, ReturnData, Schedule,
> };
> ```
> 
> becomes
> 
> ```rust
> use vm::{
>     ActionParams,
>     ActionType,
>     ActionValue,
>     ContractCreateResult,
>     ExecTrapResult,
>     Ext,
>     GasLeft,
>     MessageCallResult,
>     ParamsType,
>     ReturnData,
>     Schedule,
> };
> ```
> 
> While this does improve readability, it increases the number of lines. What is your opinion?

This would be a welcome change, already for the reason that it would simplify dealing with merge conflicts in such import lines. I've sorted out several such conflicts recently, and the latter style is preferable in such a situation.

> Also, I prefer the default formatting and avoid controversial conversations (people have different tastes, so let's just stick to the most common - the default).

This is a good counter-argument, I do agree. When defaults are good, there's usually a poor cost/benefit outcome to be meddling with them. That said, if we did decide to customize anything (i.e., cross that bridge), I think the one-import-per-line will be a good one to include.

artob (approved):

frol:
@willemneal 

> Does this mean you think this PR shouldn't happen?

I would like to understand the reasoning (the problem that we try to solve) first.

> Personally don't have much skin in the Rust format game, but this change makes a large project like this much cleaner.

I agree that imports need to be cleaner, but I think the problem is not the formatting, but our style of imports, i.e. we import the leaf items (struct, enum, function, sometimes even specific enum variants) losing the context of the crate/module this item belongs, and often I find myself lost with the types imported all around the project. My style is to keep enough context on every single line for better review and maintenance, just to give specific examples:

* Rosetta RPC implementation files uses 0-4 `use` statements per file (with an exception of `lib.rs` and `adapters/mod.rs` with 7 `use`s): https://github.com/near/nearcore/blob/master/chain/rosetta-rpc/src
* Indexer implementation: https://github.com/near/nearcore/tree/master/chain/indexer/src

My point is that if we want to make the code cleaner, we need to invest in a better project structure and modules structure instead of trying to format import statements. Import statements are crucial, and I pay a good deal of attention to make them useful, but they work great only when the modules are aligned.

frol:
I want to share my code style: https://github.com/near/near-indexer-for-explorer/blob/master/CONTRIBUTING.md#code-style (see both ordering & grouping and `use` statements sections)

bowenwang1996 (commented):
I don't have a strong opinion on the style, but I agree with @frol that unless there is a very good reason to do this, we should not make such changes since all the open PRs have to rebase on top of this, which could be painful.

willemneal:
Seeing as the options are unstable and there is disagreement, I will close this PR.

I do find it amusing that having one import per line was argued to be both good for merging and the source of merge conflicts.  If they are sorted I am in the first camp, but again closing this PR.

PR #3653, opened 2020-11-24
[WIP] Update fees config

evgenykuzyakov:
It may need to go behind nightly feature first

PR #3818, opened 2021-01-19
Move to cost taking IO into account.

olonho:


core/primitives/src/version.rs
```diff
@@ -92,7 +92,7 @@ pub const PROTOCOL_VERSION: ProtocolVersion = 41;
 
 /// Current latest nightly version of the protocol.
 #[cfg(feature = "nightly_protocol")]
-pub const PROTOCOL_VERSION: ProtocolVersion = 47;
+pub const PROTOCOL_VERSION: ProtocolVersion = 48;
```
bowenwang1996, line 95: what is the new feature? We update nightly protocol version only if we change a feature or add a new feature

runtime/runtime-params-estimator/Cargo.toml
```diff
@@ -50,7 +50,7 @@ costs_counting = ["near-vm-logic/costs_counting",
                   "testlib/costs_counting",
                   "near-evm-runner/costs_counting"]
 # Required feature for proper config, but can't be enabled by default because it is leaked to other release crates.
-required = ["costs_counting", "near-vm-runner/no_cpu_compatibility_checks", "no_cache", "near-store/single_thread_rocksdb"]
+required = ["costs_counting", "near-vm-runner/no_cpu_compatibility_checks", "no_cache"]
```
evgenykuzyakov, line 53: Still don't understand why we don't run single thread rocksdb.

evgenykuzyakov (changes requested):
Changing default will create an issue and not upgrade the existing protocol. There is a PR with a proper fix using a new experimental feature that does live migration: https://github.com/near/nearcore/pull/3653

But we also have a breaking change with the gas price computation based on the total prepaid gas amount, but this PR doesn't change it.

bowenwang1996:
@olonho do we still want to keep this PR?

matklad:
We definitely want to wrap this work up, as the current situation is self-contradictory: #4474. The current params estimator doesn't actually the costs we currently use. Not sure if its better to revive this PR or create a new one, that's up to @olonho

olonho:
No longer relevant.

PR #3819, opened 2021-01-19
make stats about futures less confusing

pmnoxx:
I'll remove the number printed in print_performance_stats function to make stats less confusing.


utils/near-performance-metrics/src/stats_enabled.rs
```diff
@@ -227,7 +227,7 @@ where
 
 pub fn print_performance_stats(sleep_time: Duration) {
     STATS.lock().unwrap().print_stats(sleep_time);
-    info!("Futures waiting for completition {}", REF_COUNTER.lock().unwrap().len());
+    info!("Futures waiting for completition");
```
bowenwang1996, line 230:
```suggestion
    info!("Futures waiting for completion");
```

bowenwang1996 (approved):

PR #3821, opened 2021-01-19
Change receipt id to use current block hash

### LINKED ISSUE #3780: Tracking issue for GHSA-fmqj-frw6-m49v
Tracks security advisory issue https://github.com/near/nearcore/security/advisories/GHSA-fmqj-frw6-m49v

EgorKulikov:
Changes they way receipt id is calculated to use current block hash instead of last block hash. Fixes #3780


runtime/runtime/src/state_viewer.rs
```diff
@@ -126,14 +126,16 @@ impl TrieViewer {
...
+            &view_state.block_hash,
             epoch_info_provider,
             view_state.current_protocol_version,
         );
         let config = Arc::new(RuntimeConfig::default());
         let apply_state = ApplyState {
             block_index: view_state.block_height,
-            last_block_hash: view_state.last_block_hash,
+            // Used for legacy reasons
+            last_block_hash: view_state.block_hash,
```
bowenwang1996, line 137: I think it is better to add `last_block_hash` to `ViewApplyState` and use it here so that in the future, it we need to actually use `last_block_hash` inside runtime, this would not cause any confusion on the rpc side.

core/primitives/src/version.rs
```diff
@@ -52,6 +52,9 @@ pub const DELETE_KEY_STORAGE_USAGE_PROTOCOL_VERSION: ProtocolVersion = 40;
 
 pub const SHARD_CHUNK_HEADER_UPGRADE_VERSION: ProtocolVersion = 41;
 
+/// Updates the way receipt ID is constructed to use current block hash instead of last block hash
+pub const CREATE_RECEIPT_ID_SWITCH_TO_CURRENT_BLOCK_VERSION: ProtocolVersion = 48;
```
bowenwang1996, line 56: @evgenykuzyakov do we want to go through nightly protocol for fixes like this as well?

runtime/runtime/src/state_viewer.rs
frol: I believe we don't expose `ViewApplyState` through RPC, so it is not an issue here. However, the rest of the reasoning on whether to use View* or non-View* structure here might be still relevant (I fail to assess it at the moment).

frol (approved):
LGTM. Bowen's comments need to be addressed (especially the one about the protocol version update since the version is completely off.

EgorKulikov:
> LGTM. Bowen's comments need to be addressed (especially the one about the protocol version update since the version is completely off.

Can you elaborate a bit about version? I just bumped nightly by one

frol:
@EgorKulikov Sure. You set `CREATE_RECEIPT_ID_SWITCH_TO_CURRENT_BLOCK_VERSION` to 48, while the latest stable is 41, which means that to release this feature we will need to adjust 48 to 42 or maybe 43 if some other feature like this is landed.

First of all, these variables (like `CREATE_RECEIPT_ID_SWITCH_TO_CURRENT_BLOCK_VERSION`) should be deprecated (cc @bowenwang1996). Instead, we need to either use `STABLE_PROTOCOL_FEATURES_TO_VERSION_MAPPING` or `nightly_protocol_features_to_version_mapping` (see below in the `version.rs` file) and use `checked_feature` macro instead of manual version comparison (grep to the macro usage)

P.S. I created a tracking issue to document better this flow

core/primitives/src/version.rs
evgenykuzyakov: I don't think so. It seems simple enough and covered by unit tests

core/primitives/src/version.rs
```diff
@@ -92,7 +95,7 @@ pub const PROTOCOL_VERSION: ProtocolVersion = 41;
```
evgenykuzyakov, line 95: You need to bump regular protocol version above, since it's not a nightly feature.

EgorKulikov:
> First of all, these variables (like `CREATE_RECEIPT_ID_SWITCH_TO_CURRENT_BLOCK_VERSION`) should be deprecated (cc @bowenwang1996). Instead, we need to either use `STABLE_PROTOCOL_FEATURES_TO_VERSION_MAPPING` or `nightly_protocol_features_to_version_mapping` (see below in the `version.rs` file) and use `checked_feature` macro instead of manual version comparison (grep to the macro usage)

I think this is just fixed bugs, not really features. They probably should be handled same way as features, but I do not think this is within the scope of what I am doing here

core/primitives/src/version.rs
```diff
@@ -88,11 +91,11 @@ pub enum ProtocolFeature {
 
 /// Current latest stable version of the protocol.
 #[cfg(not(feature = "nightly_protocol"))]
-pub const PROTOCOL_VERSION: ProtocolVersion = 41;
+pub const PROTOCOL_VERSION: ProtocolVersion = 48;
```
bowenwang1996, line 94: this should be updated to 42. Stable protocol version is not in sync with nightly protocol version

EgorKulikov: Fixed

core/primitives/src/version.rs
EgorKulikov: Fixed

core/primitives/src/version.rs
```diff
@@ -52,6 +52,13 @@ pub const DELETE_KEY_STORAGE_USAGE_PROTOCOL_VERSION: ProtocolVersion = 40;
 
 pub const SHARD_CHUNK_HEADER_UPGRADE_VERSION: ProtocolVersion = 41;
 
+/// Updates the way receipt ID is constructed to use current block hash instead of last block hash
+#[cfg(not(feature = "nightly_protocol"))]
+pub const CREATE_RECEIPT_ID_SWITCH_TO_CURRENT_BLOCK_VERSION: ProtocolVersion = 42;
+
+#[cfg(feature = "nightly_protocol")]
+pub const CREATE_RECEIPT_ID_SWITCH_TO_CURRENT_BLOCK_VERSION: ProtocolVersion = 48;
```
frol, line 60: If we release it to stable protocol directly, we don't need the feature gate. It is fine to break nightly compatibility by our feature gate design.

```suggestion
pub const CREATE_RECEIPT_ID_SWITCH_TO_CURRENT_BLOCK_VERSION: ProtocolVersion = 42;
```

core/primitives/src/version.rs
```diff
@@ -88,11 +95,11 @@ pub enum ProtocolFeature {
 
 /// Current latest stable version of the protocol.
 #[cfg(not(feature = "nightly_protocol"))]
-pub const PROTOCOL_VERSION: ProtocolVersion = 41;
+pub const PROTOCOL_VERSION: ProtocolVersion = 42;
 
 /// Current latest nightly version of the protocol.
 #[cfg(feature = "nightly_protocol")]
-pub const PROTOCOL_VERSION: ProtocolVersion = 47;
+pub const PROTOCOL_VERSION: ProtocolVersion = 48;
```
frol, line 102: No need to tick the nightly version.

```suggestion
pub const PROTOCOL_VERSION: ProtocolVersion = 47;
```

frol (approved):

bowenwang1996 (commented):
Please add a test that shows that that receipt ids generated from the same transaction on different forks are now different. This can be done by constructing the following:
              / ------ h +1 ------
h -----
            \---------------h+2 -------
where h+1 and h+2 contains the same chunk.

chain/client/tests/process_blocks.rs
```diff
@@ -2312,39 +2312,62 @@ fn test_query_final_state() {
     assert!(account_state1.amount < TESTING_INIT_BALANCE - TESTING_INIT_STAKE);
 }
 
+#[test]
+fn test_fork_receipt_ids() {
+    let (mut env, tx_hash) = prepare_env_with_transaction();
+
+    let produced_block = env.clients[0].produce_block(1).unwrap().unwrap();
+    env.process_block(0, produced_block.clone(), Provenance::PRODUCED);
```
bowenwang1996, line 2320:
```suggestion
    env.produce_block(0, 1);
```

bowenwang1996 (approved):

bowenwang1996:
@olonho @willemneal please review

chain/client/tests/process_blocks.rs
EgorKulikov: We still need height of said block down the code

willemneal (approved):

evgenykuzyakov (commented):
Need to increase protocol versions for nightly features.
For example in https://github.com/near/nearcore/pull/3881/files#diff-90f9183105b6010d859a9cfbde384be99559b004ed5fffc8c18e6bfaa0db5860

I bumped all nightly protocol version to 100, so they don't create conflicts for stable protocol features

EgorKulikov:
@evgenykuzyakov in my opinion we'd better leave this pr as is and then this changes would be implemented in your pr. But if you insist I can just port them here

evgenykuzyakov (approved):

PR #3822, opened 2021-01-20
fix(runtime): clean up runtime traps, remove unused ones, map wasmer 0.x trap to near error

ailisp:
Some near vm errors were never constructed. Drop some of them, and construct remaining of them when wasmer 0.x trap corresponding traps

Test Plan
---------
near vm tests in CI

### DIFF
--- a/runtime/near-vm-errors/src/lib.rs
+++ b/runtime/near-vm-errors/src/lib.rs
@@ -65,10 +65,6 @@ pub enum WasmTrap {
     MisalignedAtomicAccess,
     /// Breakpoint trap.
     BreakpointTrap,
-    /// Stack overflow.
-    StackOverflow,
-    /// Generic trap.
-    GenericTrap,
 }
 
 #[derive(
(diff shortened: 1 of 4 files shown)


evgenykuzyakov (changes requested):
This will break consensus. You can't change errors without a protocol upgrade, because they have to be deterministic

ailisp:
> This will break consensus. You can't change errors without a protocol upgrade, because they have to be deterministic

Error changes, however it doesn't break consensus:
- when there's no error, same behavior
- when there's error, both this pr and old one would give error, but some WasmerUnknownError becomes more specialized errors, no state change is applied
- there's no case when it was error that would succeed now or it was successful that would fail now
So I don't think a protocol upgrade is necessary?

evgenykuzyakov:
> when there's error, both this pr and old one would give error, but some WasmerUnknownError becomes more specialized errors, no state change is applied.

I think we merklize errors in `ExecutionOutcome`

ailisp:
I see, it's problematic then, and wasmer 1.0 runner must under another
protocol version/feature for same reason

PR #3826, opened 2021-01-20
feat: Add ability to import methods guarded by protocol feature

evgenykuzyakov:
Modifies `wrapped_import` macro to add an ability to introduce new host methods.
This change doesn't modify the protocol.

## Test plan:
- CI
- https://github.com/near/nearcore/pull/3827 for an example of adding a new import

### DIFF
--- a/runtime/near-vm-runner/src/imports.rs
+++ b/runtime/near-vm-runner/src/imports.rs
@@ -1,4 +1,3 @@
-use near_primitives::checked_feature;
 use near_primitives::version::ProtocolVersion;
 use near_vm_logic::VMLogic;
 
@@ -79,6 +78,7 @@ macro_rules! wrapped_imports {
                 )*
             }
 
+            #[allow(unused_variables)]
             pub(crate) fn build_wasmer(
                 memory: wasmer_runtime::memory::Memory,
                 logic: &mut VMLogic<'_>,
@@ -94,7 +94,7 @@ macro_rules! wrapped_imports {
                 let mut ns = wasmer_runtime_core::import::Namespace::new();
                 ns.insert("memory", memory);
                 $({
-                    if true $(&& checked_feature!($feature_name, $feature, protocol_version))* {
+                    if true $(&& near_primitives::checked_feature!($feature_name, $feature, protocol_version))* {
                         ns.insert(stringify!($func), wasmer_runtime::func!(wasmer_ext::$func));
                     }
                 })*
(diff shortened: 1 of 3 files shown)


runtime/near-vm-runner/src/imports.rs
```diff
@@ -24,93 +25,105 @@ macro_rules! rust2wasm {
...
+                let mut ns = wasmer_runtime_core::import::Namespace::new();
+                ns.insert("memory", memory);
+                $({
+                    $(#[cfg(feature = $feature_name)])*
+                    if true $(&& near_primitives::checked_feature!($feature_name, $feature, protocol_version))* {
+                        ns.insert(stringify!($func), wasmer_runtime::func!(wasmer_ext::$func));
+                    }
+                })*
+
+                import_object.register("env", ns);
```
ailisp, line 103: Why using a namespace object now?

evgenykuzyakov: This is based on https://github.com/near/nearcore/pull/2842

evgenykuzyakov: We need to do a dynamic construction of imports, because some feature might be on, but not enabled at the protocol version yet.

runtime/near-vm-runner/src/imports.rs
```diff
@@ -264,4 +277,7 @@ wrapped_imports! {
     // ###############
     validator_stake<[account_id_len: u64, account_id_ptr: u64, stake_ptr: u64] -> []>,
     validator_total_stake<[stake_ptr: u64] -> []>,
+    // ###############
+    // An example to add a protocol feature guarded host method
+    // #["protocol_feature_evm", EVM] test_api<[a: u64] -> []>,
```
ailisp, line 282: This example seems to be confusing, as evm is not executed in near-vm-runner. Should we use another information here?
Also, consider add a test to guarded api?

ailisp (approved):

runtime/near-vm-runner/src/imports.rs
ailisp: Got it

runtime/near-vm-runner/src/imports.rs
evgenykuzyakov: Yes. The test is here: https://github.com/near/nearcore/pull/3827

But I need to compile a contract with this import and check that it fails to run in the previous protocol version and runs in the new protocol version.

We can do this for a real new import. For example `burn_near`

PR #3799, opened 2021-01-12
feat(runtime): Wasmer 1.0 runner

ailisp:
Add Wasmer 1.0 runner, enabled by wasmer1_vm features (similar to wasmtime_vm). Expose wasmer 1.0 runner and wasmtime runner available to neard and vm standalone.

Test Plan
---------
All near-vm-runner tests pass. compatible behavior as wasmer 0.17. neard compiles and work with wasmer 1.0 runner turns on

PR #3831, opened 2021-01-21
fix(runtime): drop pwasm util stack limiter and let wasmer handle stack overflow trap

ailisp:
pwasm util's stack limiter wrap function with
```
// when stack over some limit
if ...
  unreachable
end
// call to original wasm func
```

This cause stackoverflow trap never happens and report as Wasmer Trap unreachable for wasmer1, and WasmerUnknownError in wasmer 0.17. These are confusing since it's not a user introduced unreachable and since both wasmer 0.x and 1.x can properly handle stack overflow and exit, this inject should be avoided.
Also as i test locally, this and #3822 together fix a incorrect comment in wasmer 0.x runner: FailedWithNoError (stack unwind fails, include stackoverflow) and unreachable can be differentiate (the reverse was said in comment), although it cannot tell if it's a stackoverflow or other kind of stack unwind fails. In wasmer 1.x it's even better that stack overflow can be differentiate with other stack unwind fails (see modified test and comment in this PR)

Test Plan
---------
CI, also i had manually incorporate #3822 to show that, without pwasm-util stack limiter, wasmer 0.17 returns FailedWithNoError for stack overflow, WasmTrap::Unreachable for actual unreachable; with pwasm-util stack limiter, wasmer 0.17 returns WasmTrap::Unreachable for both stack overflow and unreachable, which is as expected because of this wrap: https://github.com/paritytech/wasm-utils/blob/master/src/stack_height/mod.rs#L73. without pwasm-util stack limiter, wasmer 1.x returns stk_ovf (stack_overflow) for stack overflow and unreachable for unreachable, with pwasm-util, wasmer 1.x return unreachable for both


olonho (changes requested):
Please avoid making this change, as its benefits seems to be smaller than problems it potentially creates.

bowenwang1996 (commented):
Can you rebase your branch? Currently it is difficult to see what this change is actually about since it also includes #3799

ailisp:
@bowenwang1996 this is blocked by wasmer now, we cannot drop stack limiter until stackoverflow use deterministic amount of gas (one TODO item in the wishlist we sent to wasmer)

bowenwang1996:
> @bowenwang1996 this is blocked by wasmer now, we cannot drop stack limiter until stackoverflow use deterministic amount of gas (one TODO item in the wishlist we sent to wasmer)

Is it still blocked?

ailisp:
Later I talked with @nearmax we still want to use pwasm based stack limiter because we want to be portable on wasm runtime

Issue #3832, opened 2021-01-21
`cargo test` doesn't build


frankbraun:
`cargo test` doesn't build anymore on master:

```
$ cargo test
   Compiling runtime-params-estimator v2.3.0 (/home/frank/work/near/nearcore/runtime/runtime-params-estimator)
error[E0425]: cannot find value `EXT_COSTS_COUNTER` in crate `node_runtime`
  --> runtime/runtime-params-estimator/src/stats.rs:29:39
   |
29 |         let ext_costs = node_runtime::EXT_COSTS_COUNTER
   |                                       ^^^^^^^^^^^^^^^^^ not found in `node_runtime`
   |
help: consider importing this constant
   |
1  | use near_vm_logic::EXT_COSTS_COUNTER;
   |

error[E0425]: cannot find value `EXT_COSTS_COUNTER` in crate `node_runtime`
   --> runtime/runtime-params-estimator/src/testbed_runners.rs:212:19
    |
212 |     node_runtime::EXT_COSTS_COUNTER.with(|f| {
    |                   ^^^^^^^^^^^^^^^^^ not found in `node_runtime`
    |
help: consider importing this constant
    |
1   | use near_vm_logic::EXT_COSTS_COUNTER;
    |

error[E0061]: this function takes 10 arguments but 11 arguments were supplied
  --> runtime/runtime-params-estimator/src/vm_estimator.rs:55:5
   |
55 |     near_vm_runner::run(
   |     ^^^^^^^^^^^^^^^^^^^ expected 10 arguments
56 |         code_hash,
   |         ---------
57 |         code,
   |         ----
58 |         b"cpu_ram_soak_test",
   |         --------------------
59 |         &mut fake_external,
   |         ------------------
60 |         context,
   |         -------
61 |         &config,
   |         -------
62 |         &fees,
   |         -----
63 |         &promise_results,
   |         ----------------
64 |         PROTOCOL_VERSION,
   |         ----------------
65 |         None,
   |         ----
66 |         &None,
   |         ----- supplied 11 arguments

error: aborting due to 3 previous errors

Some errors have detailed explanations: E0061, E0425.
For more information about an error, try `rustc --explain E0061`.
error: could not compile `runtime-params-estimator`

To learn more, run the command again with --verbose.
```

The first commit that introduced that problem is e2e6bd63f3d6ad34e0711d8068d755ac480e0b8e

[...]

PR #3834, opened 2021-01-21
fix: #3832

willemneal:
@frankbraun

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3560,7 +3560,6 @@ dependencies = [
  "node-runtime",
  "openssl-probe",
  "rand 0.7.3",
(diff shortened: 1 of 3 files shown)


olonho (approved):

ailisp (approved):

evgenykuzyakov (commented):
This PR might leak `protocol_feature_evm` from `[workspace]` members. Not sure how it works. This is the reason why it was not in `default`

frankbraun:
Just removing the `default_features = false` in the top-level `Cargo.toml` is not an option?

bowenwang1996:
I think we do not need this PR any more. Please reopen if this is not the case

PR #3835, opened 2021-01-21
chore: reorganize dependencies and prepare crates for publishing

willemneal:
This PR removes circular dependencies between runtime crates and the primitives crate.  It also switches to use rocksdb's crate.io version instead of near's github repo's single_thread branch.  The change in rocksdb has been merged upstream.  Once published the single_threaded feature can be added back.


core/primitives/src/config.rs
```diff
@@ -0,0 +1,626 @@
...
+    stake,
+    add_key,
+    delete_key,
+    value_return,
+    new_receipt,
+}
+
+impl fmt::Display for ActionCosts {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "{}", ActionCosts::name_of(*self as usize))
```
ailisp, line 482: There is some crate or macro that can reduce this code, e.g. https://docs.rs/enum-display-derive/0.1.0/enum_display_derive/

ailisp (approved):

runtime/near-runtime-fees/Cargo.toml
```diff
@@ -14,6 +14,7 @@ Fees applied to near runtime encapsulated in a separate crate. Might merge it la
 
 [dependencies]
 serde = { version = "1", features = ["derive"] }
+near-primitives = { path = "../../core/primitives" }
```
evgenykuzyakov, line 17:
```suggestion
near-primitives = { path = "../../core/primitives", version = "0.1.0" }
```

evgenykuzyakov (commented):
Have you tried to publish using `cargo publish --dry-run`?

core/primitives/Cargo.toml
```diff
@@ -23,7 +23,7 @@ rand = "0.7"
 reed-solomon-erasure = "4"
 jemallocator = { version = "0.3", optional = true }
 hex = "0.4"
-num-rational = "0.2.4"
+num-rational = { version = "0.2.4", features = ["serde"]}
 primitive-types = "0.7"
 
 borsh = "0.7.1"
```
evgenykuzyakov, line 29: There are also 3 dependencies below that we'll have to publish first before we can publish this

runtime/near-runtime-fees/Cargo.toml
willemneal: looking over this crate, it just re-exports the near_primitives so perhaps it's not needed anymore.  Do you think I should just refactor all deps to use `near_primitives::runtime::fees`?

core/chain-configs/src/genesis_config.rs
```diff
@@ -13,15 +13,17 @@ use num_rational::Rational;
...
+use near_primitives::{
+    hash::CryptoHash,
+    runtime::config::RuntimeConfig,
+    serialize::{u128_dec_format, u128_dec_format_compatible},
+    state_record::StateRecord,
+    types::{
+        AccountId, AccountInfo, Balance, BlockHeight, BlockHeightDelta, EpochHeight, Gas,
+        NumBlocks, NumSeats,
+    },
+    version::ProtocolVersion,
```
frol, line 25: I vote to revert styling changes back unless we commit to a certain style.

core/primitives/src/lib.rs
```diff
@@ -11,15 +11,18 @@ pub mod account;
...
 pub mod epoch_manager;
 pub mod errors;
 pub mod hash;
 pub mod logging;
 pub mod merkle;
 pub mod network;
+pub mod profile;
 pub mod receipt;
 pub mod rpc;
+pub mod runtime;
```
frol, line 25: Our core primitives are already enormous (it has 35 flat modules in the root). What is the driving factor for collapsing even more crates into it?

core/primitives/Cargo.toml
willemneal: `near_crypto` seems publishable as is.
`near-primitives` and `near_vm_errors` depends on `near-rpc-error-macro`.  So it would make sense to bring `near_vm_errors` into `near_primitives`.
Leaving `near-rpc-error-macro` to be published, which itself depends on `near-rpc-error-core`, which seems publishable as is.

runtime/near-evm-runner/Cargo.toml
```diff
@@ -34,6 +34,7 @@ ethereum-types = "0.8.0"
 ethabi = "11.0.0"
 
 near-vm-logic = { path = "../near-vm-logic" }
+near-primitives = { path = "../../core/primitives" }
```
frol, line 37: I am not sure about near-evm-runner, but I think it is still preferable to keep it slim and avoid dependency on near-primitives.

runtime/near-vm-logic/Cargo.toml
```diff
@@ -24,19 +24,20 @@ sha3 = ">=0.8,<0.10"
 near-runtime-fees = { path = "../near-runtime-fees", version = "2.3.0" }
 near-vm-errors = { path = "../near-vm-errors", version = "2.3.0" }
 near-runtime-utils = { path = "../near-runtime-utils", version = "2.3.0" }
+near-primitives = { path = "../../core/primitives", version = "0.1.0"}
```
frol, line 27: `near-vm-logic` and `near-runtime-fees` are dependencies of near-sdk-rs which we MUST maintain compilable for Wasm target (and slim) for smart contracts, so they cannot depend on near-primitives.

This is why you observe some code duplication.

runtime/near-runtime-fees/Cargo.toml
frol: Just a reminder: we cannot have near-primitives dependency in near-runtime-fees (see my other comment)

frol (changes requested):
I am afraid I have to ask for reverting some of the changes:

* [x] Most crucial are the changes to near-vm-logic and near-runtime-fees (also, please, double-check what they depend on, so you don't accidentally introduce near-primitives somewhere else down the line)
* [x] `use` statements code-style was not really agreed on, so please revert those changes for this PR

chain/indexer/Cargo.toml
```diff
@@ -8,7 +8,7 @@ edition = "2018"
 actix = "0.9"
 tracing = "0.1.13"
 futures = "0.3.5"
-rocksdb = { git = "https://github.com/nearprotocol/rust-rocksdb", branch="disable-thread" }
+rocksdb = { version = "0.15.0"}
```
bowenwang1996, line 11: isn't this simply `rocksdb = "0.15.0"`?

runtime/near-vm-logic/Cargo.toml
evgenykuzyakov: I thought this is the point. We want to publish `near-primitives` crate

core/primitives/Cargo.toml
evgenykuzyakov: It seems we should move `near-rpc-error-macro` to optional dependency and only use it when we need to regenerate error schema. @frol

runtime/near-vm-logic/Cargo.toml
frol: @evgenykuzyakov

1. I don't believe it will compile near-crypto, sha2, rand, jemalloc for Wasm target
2. We don't help guys porting near-sdk-rs for `no_std` environment which such a change...

core/chain-configs/src/genesis_config.rs
willemneal: Yeah good point; this was just hand crafted a while ago, but I'll revert!

matklad: fwiw, I have a "more than rustfmt" style guide here: https://github.com/rust-analyzer/rust-analyzer/blob/master/docs/dev/style.md

core/primitives/src/lib.rs
willemneal: Lifting deps that are across multiple crates.  E.i. there should only be one set of primitive types.  I solved this by adding a new crate `near-primitives-core` which can be depended on by both `near-primitives` and vm crates.

Cargo.lock
```diff
@@ -10,15 +10,15 @@ dependencies = [
  "actix-rt",
  "actix_derive",
  "bitflags",
- "bytes",
+ "bytes 0.5.6",
```
frol, line 13: I suggest we don't go this long path (#3869) of trying to upgrade dependencies in this PR, so the quick fix here should be:

1. revert the Cargo.lock changes
2. run `cargo check`, so it only updates the relevant parts

core/primitives/src/lib.rs
frol: I love that we remove code duplication! @willemneal thanks for championing this!

@evgenykuzyakov do we have a test in nearcore or near-sdk-rs that confirms that it is compilable with the latest `near-vm-logic` and `near-vm-errors`?

We need to confirm that `near-sdk-rs` is not going to break after this refactoring.

willemneal:
@frol Thanks! I had a meeting today with @evgenykuzyakov and the idea is to publish the vm crates with a new major version and then fix the issues this PR creates.  I've tested it locally and it's only a small change for the sdk.

frol (approved):
@willemneal I guess, you will need to revert the changes in `deny.toml` :wink:

evgenykuzyakov (approved):

core/primitives/src/version.rs
```diff
@@ -98,16 +95,12 @@ pub const PROTOCOL_VERSION: ProtocolVersion = 42;
 
 /// Current latest nightly version of the protocol.
 #[cfg(feature = "nightly_protocol")]
-pub const PROTOCOL_VERSION: ProtocolVersion = 104;
+pub const PROTOCOL_VERSION: ProtocolVersion = 48;
```
evgenykuzyakov, line 98: Seems it was incorrectly reverted

core/primitives/src/version.rs
```diff
@@ -98,16 +95,12 @@ pub const PROTOCOL_VERSION: ProtocolVersion = 42;
...
-    static ref STABLE_PROTOCOL_FEATURES_TO_VERSION_MAPPING: HashMap<ProtocolFeature, ProtocolVersion> =
-        vec![
-            #[cfg(feature = "protocol_feature_lower_storage_cost")]
-            (ProtocolFeature::LowerStorageCost, 42),
-        ]
-        .into_iter()
-        .collect();
+    static ref STABLE_PROTOCOL_FEATURES_TO_VERSION_MAPPING: HashMap<ProtocolFeature, ProtocolVersion> = vec![
+        /* add mapping here */
+    ].into_iter().collect();
```
evgenykuzyakov, line 103: Same here

core/primitives/src/version.rs
```diff
@@ -122,18 +115,18 @@ lazy_static! {
...
+            (ProtocolFeature::ForwardChunkParts, 42),
             #[cfg(feature = "protocol_feature_rectify_inflation")]
-            (ProtocolFeature::RectifyInflation, 102),
+            (ProtocolFeature::RectifyInflation, 43),
             #[cfg(feature = "protocol_feature_evm")]
-            (ProtocolFeature::EVM, 103),
+            (ProtocolFeature::EVM, 46),
             #[cfg(feature = "protocol_feature_block_header_v3")]
-            (ProtocolFeature::BlockHeaderV3, 104),
+            (ProtocolFeature::BlockHeaderV3, 48),
```
evgenykuzyakov, line 129: And here

core/primitives/src/version.rs
```diff
@@ -148,8 +141,6 @@ lazy_static! {
             );
         }
         nightly_protocol_features_to_version_mapping
-            .extend(STABLE_PROTOCOL_FEATURES_TO_VERSION_MAPPING.iter());
-        nightly_protocol_features_to_version_mapping
```
evgenykuzyakov, line 152: And here

core/primitives/src/version.rs
```diff
@@ -87,9 +87,6 @@ pub enum ProtocolFeature {
     EVM,
     #[cfg(feature = "protocol_feature_block_header_v3")]
     BlockHeaderV3,
-    /// Decreases the storage cost of 1 byte by 10X.
-    #[cfg(feature = "protocol_feature_lower_storage_cost")]
-    LowerStorageCost,
```
evgenykuzyakov, line 92: It's needed

core/primitives/Cargo.toml
```diff
@@ -22,26 +28,26 @@ rand = "0.7"
...
 near-crypto = { path = "../crypto" }
 near-vm-errors = { path = "../../runtime/near-vm-errors" }
 near-rpc-error-macro = { path = "../../tools/rpctypegen/macro" }
 
 [features]
-default = ["jemallocator", "stable_protocol_features"]
+default = ["jemallocator"]
 dump_errors_schema = ["near-rpc-error-macro/dump_errors_schema"]
-stable_protocol_features = ["protocol_feature_lower_storage_cost"]
-protocol_feature_lower_storage_cost = []
```
evgenykuzyakov, line 38: Invalid merge?

evgenykuzyakov (changes requested):

core/primitives/Cargo.toml
```diff
@@ -39,8 +39,10 @@ near-vm-errors = { path = "../../runtime/near-vm-errors" }
 near-rpc-error-macro = { path = "../../tools/rpctypegen/macro" }
 
 [features]
-default = ["jemallocator"]
+default = ["jemallocator", "protocol_feature_lower_storage_cost"]
```
evgenykuzyakov, line 42: I think we should add `stable_protocol_features` here instead.
```suggestion
default = ["jemallocator", "stable_protocol_features"]
```

evgenykuzyakov (approved):

core/primitives-core/Cargo.toml
```diff
@@ -0,0 +1,30 @@
...
+serde = { version = "1", features = ["derive"] }
+serde_json = "1"
+sha2 = ">=0.8,<0.10"
+lazy_static = "1.4"
+
+[features]
+default = ["protocol_feature_lower_storage_cost"]
+costs_counting = []
+protocol_feature_evm = []
+protocol_feature_lower_storage_cost = []
```
evgenykuzyakov, line 29: I think `primitives-core` doesn't need this feature, since `RuntimeConfig` lives in `primitives/runtime` which are part of `primitives` and not `primitives-core`

```suggestion
default = []
costs_counting = []
protocol_feature_evm = []
```

willemneal: Ah great thanks for double checking!

core/primitives-core/Cargo.toml
```diff
@@ -0,0 +1,28 @@
...
+[dependencies]
+base64 = "0.11"
+borsh = "0.8.1"
+bs58 = "0.4"
+hex = "0.4"
+derive_more = "0.99.3"
+num-rational = { version = "0.3.1", features = ["serde"]}
+serde = { version = "1", features = ["derive"] }
+serde_json = "1"
+sha2 = ">=0.8,<0.10"
```
bowenwang1996, line 22: why

core/primitives/src/contract.rs
```diff
@@ -1,21 +1 @@
...
-
-    pub fn get_hash(&self) -> CryptoHash {
-        self.hash
-    }
-
-    pub fn get_code(&self) -> &Vec<u8> {
-        &self.code
-    }
-}
+pub use near_primitives_core::contract::*;
```
bowenwang1996, line 1: why do we do this? Is there any point in keeping this file?

core/store/Cargo.toml
```diff
@@ -32,10 +32,12 @@ name = "trie_bench"
 harness = false
 
 [features]
-default = []
+default = ["rocksdb"]
```
bowenwang1996, line 35: why do we need to make this a feature

willemneal: This is so that you could depend on store without requiring rocksdb.  But since other crates in nearcore do i made it default.  I fine to revert this and just make the dep non-optional.

core/primitives/src/contract.rs
willemneal: This was so that other crates that depend on `near-primitives` don't have to depend on `near-primitives-core`.

core/primitives-core/Cargo.toml
willemneal: Not sure, I just copied this from `primitives`.

core/primitives/src/logging.rs
```diff
@@ -1,83 +1 @@
...
-            pretty_str(HI_NEAR, HI_NEAR.chars().count())
-        );
-    }
-
-    #[test]
-    fn test_non_ut8_no_truncation() {
-        assert_eq!(format!("`{}`", HI_NEAR), pretty_str(HI_NEAR, HI_NEAR.len()));
-    }
-}
+pub use near_primitives_core::logging::*;
```
frol, line 1: Instead of keeping the mostly empty files, I suggest we delete these re-export files and re-export right in the `lib.rs` (`pub use near_primitives_core::logging;`)

runtime/near-evm-runner/Cargo.toml
```diff
@@ -68,3 +68,6 @@ required-features = ["protocol_feature_evm"]
 name = "bench"
 required-features = ["protocol_feature_evm"]
 harness = false
+
+[package.metadata.workspaces]
+independent = true
```
frol, line 73: I recall there was some mention that we plan to use `cargo-workspaces`. It would be great to have a short document/section in CONRIBUTION.md walking through the flow and common features (like this `package.metadata.workspaces` special Cargo.toml section) :pray:

core/primitives/src/contract.rs
bowenwang1996: Okay, so why do we have near-primitives-core in the first place

core/store/Cargo.toml
bowenwang1996: Not sure how you do that given the way it is written today. If you want something to depend on store without rocksdb, you need to refactor the crate

willemneal: @bowenwang1996 Reverted!

core/primitives/src/contract.rs
willemneal: To remove duplicated code.  This way the VM crates can depend on this crate and so can `near-primitives`.

bowenwang1996 (approved):

ailisp (approved):

PR #3843, opened 2021-01-22
Cache timing

olonho:


bowenwang1996 (commented):
@olonho just fyi, we have a tool called `delay_detector` that could be used to measure potential slowness in some function. @ailisp already used it in his investigation

runtime/near-vm-runner-standalone/src/main.rs
```diff
@@ -89,6 +95,55 @@ fn default_vm_context() -> VMContext {
...
+    println!("restore cache from {}", file);
+    let bytes = read_file(file).unwrap();
+    let start = Instant::now();
+    let artifact = Artifact::deserialize(&bytes).unwrap();
+    println!("Artifact::deserialize {} μs", start.elapsed().as_micros());
+    unsafe {
+        let compiler = compiler_for_backend(Backend::Singlepass).unwrap();
+        let start = Instant::now();
+        match load_cache_with(artifact, compiler.as_ref()) {
+            Ok(_) => println!("load_cache_with {} μs", start.elapsed().as_micros()),
```
matklad, line 141:
```suggestion
            Ok(_) => println!("load_cache_with {:?}", start.elapsed())),
```

duration type has a nice Debug impl which picks the appropriate time unint automatically. `{:.2?}` can be used to not show overly precise fractional parts.

ailisp (commented):
I suggest to reuse delay_detector too. Note that delay_detector only report delay > 50ms, We should make it configurable since in runtime layer much smaller delay is also problematic.

ailisp (approved):
Anyway, it's very helpful to have this information in runtime standalone

matklad:
JFYI, rust-analyzer has a slightly fancier version of delay detector at https://github.com/rust-analyzer/rust-analyzer/blob/master/crates/profile/src/hprof.rs

The output looks like this:

```
   85ms - handle_completion
       68ms - import_on_the_fly @ 
           67ms - import_assets::search_for_relative_paths
                0ms - crate_def_map:wait (804 calls)
                0ms - find_path (16 calls)
                2ms - find_similar_imports (1 calls)
                0ms - generic_params_query (334 calls)
               59ms - trait_solve_query (186 calls)
            0ms - Semantics::analyze_impl (1 calls)
            1ms - render_resolution (8 calls)
        0ms - Semantics::analyze_impl (5 calls)
```

There's also generic [tracing](https://github.com/tokio-rs/tracing) ecosystem, which is polymorphic in the consumer of instrumentation data.

runtime/near-vm-runner-standalone/src/main.rs
```diff
@@ -89,6 +95,55 @@ fn default_vm_context() -> VMContext {
...
+
+    return Ok(data);
+}
+
+fn write_file(name: &str, data: &Vec<u8>) -> std::io::Result<()> {
+    let mut file = File::create(name)?;
+    file.write(data)?;
+
+    return Ok(());
+}
```
matklad, line 112: * https://doc.rust-lang.org/stable/std/fs/fn.read.html
* https://doc.rust-lang.org/stable/std/fs/fn.write.html

ailisp:
I suggest also do the following changes:
print bootstrap wasmer time, compile time, execution time, either with hprofs or plain print; if with plain print, print it with a special prefix so it's straightforward to parse in proposed scripts.

bowenwang1996:
@olonho what is the status of this PR?

matklad:
Closing: we now use `tracing` in contract runtime, that already gives us visibility into the time spend in the caches:

https://github.com/near/nearcore/blob/master/docs/architecture.md#logging--observability

PR #3845, opened 2021-01-22
feat(evm): Add value to the meta tx

### LINKED ISSUE #3795: Meta call should have value field
Currently there is no way for user to attach `value` (EVM base token amount) to the call that was relayed via `meta_call`.

There is actually a bug where attached deposit of the `meta_call` function gets used as the `value` in the subsequent EVM call, which is not really correct as attached deposit belongs to the relayer not to the meta call signer.

ilblackdragon:
Also switch to use borsh serialization of MetaCall args for consistency.

Fixes #3795

Testing Plan
---
Add test case to check that `value` is indeed been used.
Add test to check that wrongly serialized message gets ArgumentError


runtime/near-evm-runner/tests/standard_ops.rs
```diff
@@ -308,14 +314,15 @@ fn test_meta_call_sig_and_recover() {
...
+        U256::from(0),
         "adopt(uint256 petId)",
         // RLP encode of ["0x09"]
         hex::decode("c109").unwrap(),
     );
 
     // meta_tx[0..65] is eth-sig-util format signature
     // assert signature same as eth-sig-util, which also implies msg before sign (constructed by prepare_meta_call_args, follow eip-712) same
-    assert_eq!(hex::encode(&meta_tx[0..65]), "29b88cd2fab58cfd0d05eacdabaab081257d62bdafe9153922025c8e8723352d61922acbb290bd1dba8f17f174d47cd5cc41480d19a82a0bff4d0b9b9441399b1c");
+    assert_eq!(hex::encode(&meta_tx[0..65]), "4d94263f09bfd6322a633eebbf087fbed32d1b964e2fdab9cc9931fff3b9cd683e0912697e26836007e6ba026acccd9bb6116713959936815b1f6d9496dc5d341c");
```
ailisp, line 325: Did you try the eth-sig-util to take same argument (by add a `value` field here: https://gist.github.com/ailisp/0f65bc1a3827b03141ae6601918a67a2#file-sig-js-L29, `npm init && npm i eth-sig-util && node sig.js`)

ailisp (approved):

PR #3850, opened 2021-01-24
feat(runtime): save transaction hashes in state

bowenwang1996:
Maintain in state the hashes of transactions from last `transaction_validity_period`. Per block hashes are saved under `Triekey::BlockTransactionHashes` while we also maintain `TrieKey::AccountTransactionHash` that allows us to easily query whether some account has already sent a given transaction.

Test plan
----------
* `test_transaction_hash_collision`
* `test_chunk_transaction_validity`
* `test_transaction_hashes_with_forks`
* nayduck


chain/rosetta-rpc/src/adapters/mod.rs
```diff
@@ -187,6 +187,11 @@ fn convert_block_changes_to_transactions(
                     "State Change 'NotWritableToDisk' should never be observed".to_string(),
                 ));
             }
+            #[cfg(feature = "protocol_feature_transaction_hashes_in_state")]
+            StateChangeCauseView::StoreTransactionHashes => {
+                // TODO: is any action required here?
+                unimplemented!()
```
bowenwang1996, line 193: @frol what is the expected behavior here?

frol, line 193: It should just ignore this change completely.

```suggestion
                continue;
```

neard/src/runtime.rs
```diff
@@ -523,6 +532,30 @@ impl NightshadeRuntime {
...
+        {
+            get::<()>(
+                &state_update,
+                &TrieKey::AccountTransactionHash {
+                    account_id: tx.transaction.signer_id.to_string(),
+                    hash: tx.get_hash(),
+                },
+            )
+            .map(|x| x.is_some())
+            .map_err(|e| Error::from(ErrorKind::StorageError(e)))
```
frol, line 552: This feels like abuse of store with assumption that `()` is encoded as nothing in borsh. I believe it would be cleaner if store would expose `contains` method just to check the existance of a given key.

Notes:

1. `keyMayExist` in RocksDB is [confusing](https://github.com/facebook/rocksdb/blob/v5.14.2/include/rocksdb/db.h#L369-L375), so we definitely don't want to use it (see https://github.com/facebook/rocksdb/issues/4149)
2. I think we can use [`get_pinned`](https://docs.rs/rocksdb/0.15.0/rocksdb/struct.DB.html?search=#method.get_pinned) method to avoid memory copies (and borsh deserialization) since we don't even want to read the value itself

frol (approved):

neard/src/runtime.rs
bowenwang1996: > This feels like abuse of store with assumption that () is encoded as nothing in borsh

I don't think I assume that. Can you explain why?

frol: My bad, sorry. I thought that you store something else but use `()` in `get` just to "skip" heavy deserialization

chain/chain/src/chain.rs
```diff
@@ -2451,6 +2455,7 @@ impl Chain {
 /// and decide to accept it or reject it.
 /// If rejected nothing will be updated in underlying storage.
 /// Safe to stop process mid way (Ctrl+C or crash).
+#[allow(unused)]
```
matklad, line 2458: Can we put unsued on a specific field and/or add a comment? Might help with noticing when the unused annotation is itself unused.

chain/client/src/test_utils.rs
```diff
@@ -1170,14 +1170,12 @@ pub fn create_chunk(
...
-            }
-            EncodedShardChunk::V2(ref mut chunk) => {
-                *chunk.header.height_included_mut() = 2;
-            }
+    match &mut chunk {
+        EncodedShardChunk::V1(ref mut chunk) => {
+            chunk.header.height_included = next_height;
+        }
+        EncodedShardChunk::V2(ref mut chunk) => {
+            *chunk.header.height_included_mut() = next_height;
```
matklad, line 1178:
```suggestion
        EncodedShardChunk::V1(chunk) => {
            chunk.header.height_included = next_height;
        }
        EncodedShardChunk::V2(chunk) => {
            *chunk.header.height_included_mut() = next_height;
```

I thing this are no longer necessary.

chain/client/tests/process_blocks.rs
```diff
@@ -2618,3 +2622,245 @@ fn test_block_ordinal() {
     ordinal += 1;
     assert_eq!(next_block.header().block_ordinal(), ordinal);
 }
+
+/// Test that duplicate transactions are properly rejected.
+#[cfg(feature = "protocol_feature_transaction_hashes_in_state")]
+#[test]
+fn test_transaction_hash_collision() {
```
matklad, line 2629: Maybe it makes sense to have only single cfg here:

```
#[cfg(feature = "protocol_feature_transaction_hashes_in_state")]
#[cfg(test)]
mod transaction_hashes_in_state_tests {
  #[test] fn test_transaction_hash_collision() { ... }
}
```

core/primitives/src/trie_key.rs
```diff
@@ -172,6 +195,17 @@ impl TrieKey {
...
+            #[cfg(feature = "protocol_feature_transaction_hashes_in_state")]
+            TrieKey::BlockTransactionHashes { height } => {
+                res.extend(col::BLOCK_TRANSACTION_HASH);
+                res.extend(&height.to_le_bytes());
+            }
+            #[cfg(feature = "protocol_feature_transaction_hashes_in_state")]
+            TrieKey::AccountTransactionHash { account_id, hash } => {
+                res.extend(col::ACCOUNT_TRANSACTION_HASH);
+                res.extend(hash.as_ref());
+                res.extend(account_id.as_bytes());
```
matklad, line 207: Feels a bit odd that the order of hash and account_id is reversed with the mathc statement/variant decl

runtime/runtime/src/lib.rs
```diff
@@ -1246,6 +1271,36 @@ impl Runtime {
...
+                    &mut state_update,
+                    TrieKey::BlockTransactionHashes { height: apply_state.block_index },
+                    &transaction_hash_keys,
+                );
+                for height in apply_state
+                    .prev_block_height
+                    .saturating_sub(apply_state.transaction_validity_period)
+                    ..apply_state
+                        .block_index
+                        .saturating_sub(apply_state.transaction_validity_period)
```
matklad, line 1289: Tiny nit, but feels like this can benefit from introduction of a couple of local variables.

core/primitives/src/trie_key.rs
bowenwang1996: This is intentional to reduce depth in merkle patricia tree.

willemneal (approved):
Looks good to me.  This makes me wonder if a `protocol_feature` macro might be useful, e.g. `protocol_feature!(transaction_hashs_in_state)`. However, this would mean another import statement.

runtime/runtime/src/lib.rs
```diff
@@ -1246,6 +1271,36 @@ impl Runtime {
...
+        checked_feature!(
+            "protocol_feature_transaction_hashes_in_state",
+            TransactionHashesInState,
+            apply_state.current_protocol_version,
+            {
+                set(
+                    &mut state_update,
+                    TrieKey::BlockTransactionHashes { height: apply_state.block_index },
+                    &transaction_hash_keys,
+                );
```
evgenykuzyakov, line 1283: Maybe we don't need to store it if it's empty.

runtime/runtime/src/lib.rs
```diff
@@ -1246,6 +1271,36 @@ impl Runtime {
...
+                for height in apply_state
+                    .prev_block_height
+                    .saturating_sub(apply_state.transaction_validity_period)
+                    ..apply_state
+                        .block_index
+                        .saturating_sub(apply_state.transaction_validity_period)
+                {
+                    let trie_key = TrieKey::BlockTransactionHashes { height };
+                    let transaction_hash_keys =
+                        get::<Vec<(AccountId, CryptoHash)>>(&state_update, &trie_key)?
```
evgenykuzyakov, line 1293: Introduce a type for `Vec<(AccountId, CryptoHash)>`

neard/src/runtime.rs
```diff
@@ -529,6 +538,30 @@ impl NightshadeRuntime {
 
         Ok(result)
     }
+
+    #[allow(unused_variables)]
+    fn check_duplicate_transaction(
```
evgenykuzyakov, line 543: The more canonical name will `is_duplicate_transaction`
```suggestion
    fn is_duplicate_transaction(
```

runtime/runtime/src/lib.rs
evgenykuzyakov: It would be great to comment and draw an example with the boundaries of valid TXs and when they expire the most. Include skipped block in this example as well.

I did it on a paper and it seems fine, but we should include it here for future reference.

EgorKulikov (approved):

EgorKulikov:
As a food for thought: while we do not store ACCOUNT_TRANSACTION_HASH under that account subtree, it is stored in state and nobody "pays" for it, while it clearly have associated account

bowenwang1996:
> As a food for thought: while we do not store ACCOUNT_TRANSACTION_HASH under that account subtree, it is stored in state and nobody "pays" for it, while it clearly have associated account

@evgenykuzyakov what do you think?

abacabadabacaba:
I don't like the entire idea of preventing duplicate transactions at consensus level, for the following reasons:
1. Duplicate transactions are causing the most problems for tools that process historical data, such as block explorer. However, these tools need to be able to process all historical data, and there are already duplicate transactions in the blockchain, so in any case the tools will need to be modified to handle duplicate transactions.
2. If a wallet adds the same key to an account, it is a bug in the wallet. Re-adding the same key is just one of the many ways a buggy wallet can compromise the security of the account, and most other ways cannot be mitigated by consensus changes. In general, a good approach is to only ever add a freshly generated random key to an account, in which case this problem should not happen.
3. This change bloats the state and adds substantial extra work for all full nodes, for not much gain.

frol:
> Duplicate transactions are causing the most problems for tools that process historical data, such as block explorer. However, these tools need to be able to process all historical data, and there are already duplicate transactions in the blockchain, so in any case the tools will need to be modified to handle duplicate transactions.

@abacabadabacaba Well, it causes issues much earlier. Tx status cannot be retrieved by the transaction hash and account id anymore.

bowenwang1996:
The issue is fixed by #4064

PR #3855, opened 2021-01-26
Compress compiled contracts.

olonho:


ailisp (approved):
Looks good. I think we can add bench of a series of different size of contracts, to see for which contract worth a conpiled, can be addressed in later PRs

PR #3859, opened 2021-01-27
Cache contract on deployment.

olonho:
### Problem
 Currently we blindly deploy the contract, and do not check if it is fully correct Wasm
binary that could be linked with NEAR host functions. As result, our compilation caches could
be attacked by incorrect contracts execution attempts both intentionally and non-intentionally.
Also node on which deployment happened will get precompiled contract immediately.

### Solution

Precompile contracts on deployment and return deployment error immediately.

### Test plan

cargo test --all --workspace

### Considerations

As we store precompiled contracts in regular store, not in trie, propagation of compiled contracts across nodes happens gradually. We could come up with a mechanism to actually distribute precompiled contracts, but this looks like a next step.

### DIFF
--- a/runtime/runtime/src/actions.rs
+++ b/runtime/runtime/src/actions.rs
@@ -11,7 +11,7 @@ use near_primitives::transaction::{
     Action, AddKeyAction, DeleteAccountAction, DeleteKeyAction, DeployContractAction,
     FunctionCallAction, StakeAction, TransferAction,
 };
-use near_primitives::types::{AccountId, EpochInfoProvider, ValidatorStake};
+use near_primitives::types::{AccountId, CompiledContractCache, EpochInfoProvider, ValidatorStake};
 use near_primitives::utils::create_random_seed;
 use near_primitives::version::{
     ProtocolVersion, DELETE_KEY_STORAGE_USAGE_PROTOCOL_VERSION,
@@ -29,12 +29,14 @@ use near_store::{
 };
 use near_vm_errors::{CacheError, CompilationError, FunctionCallError, InconsistentStateError};
 use near_vm_logic::types::PromiseResult;
-use near_vm_logic::{VMContext, VMOutcome};
-use near_vm_runner::VMError;
+use near_vm_logic::{VMConfig, VMContext, VMKind, VMOutcome};
+use near_vm_runner::{precompile, VMError};
 
 use crate::config::{safe_add_gas, RuntimeConfig};
 use crate::ext::RuntimeExt;
 use crate::{ActionResult, ApplyState};
+use std::ops::Deref;
+use std::sync::Arc;
 
 /// Runs given function call with given context / apply state.
 /// Precompiles:
@@ -424,6 +426,8 @@ pub(crate) fn action_deploy_contract(
     account: &mut Account,
     account_id: &AccountId,
     deploy_contract: &DeployContractAction,
+    cache: Option<Arc<dyn CompiledContractCache>>,
+    vm_config: &VMConfig,
 ) -> Result<(), StorageError> {
     let code = ContractCode::new(deploy_contract.code.clone(), None);
     let prev_code = get_code(state_update, account_id, Some(account.code_hash))?;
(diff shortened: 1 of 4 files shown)


ailisp (commented):
Would this cause a change in deduct deploy cost?

runtime/runtime/src/actions.rs
```diff
@@ -438,6 +442,18 @@ pub(crate) fn action_deploy_contract(
         })?;
     account.code_hash = code.get_hash();
     set_code(state_update, account_id.clone(), &code);
+    if let Some(cache) = cache {
+        // We only precompile contract for Wasmer VM at the moment.
+        if let Some(err) =
+            precompile(code.get_code(), &code.get_hash(), vm_config, cache.deref(), VMKind::Wasmer)
+        {
+            return Err(StorageError::StorageInconsistentState(format!(
```
evgenykuzyakov, line 450: What if the contract doesn't compile? It looks like `precompile` can throw an error.

evgenykuzyakov (changes requested):
In general, when every PR should contain:
- description explaining the problem
- explanation how this PR solves the problem
- Test Plan - what you did to test this change. E.g. why it doesn't break the protocol.|

For this particular change there are a few issues:
- We're modifying the behavior, so the costs of deploy should be adjusted already. We should keep the old behavior before the protocol version that adjusts fees. So we should guard this behind a compilation feature that will be used for rolling out the new fees config.
- This error handling shouldn't change the protocol, but it does.
- Need to add tests to verify that invalid contract doesn't break the protocol, but the new valid contract gets precompiled.

runtime/runtime/src/actions.rs
```diff
@@ -438,6 +442,18 @@ pub(crate) fn action_deploy_contract(
...
+                "Cannot precompile contract {}: {:?}",
+                code.get_hash().to_string(),
```
matklad, line 452:
```suggestion
                code.get_hash(),
```

`x.to_string` is, by definition, `format!("{}", x)`, so `format!("{}", x.to_string())` is a tautology.

runtime/runtime/src/actions.rs
matklad, line 450: A more idiomatic (but not necessary more readable :) ) version would be `

```
precompile(...).map_err(|err| StorageError(...))?
```

runtime/runtime/src/actions.rs
olonho: Thanks, C still haunts my brain.

runtime/runtime/src/actions.rs
olonho: IIRC originally it was `map_err`, but IDE inspection suggested me to rewrite with 'let`.

olonho: Actually if contract doesn't compile we could do one of the following:
   * disregard error, but it means we'll try to recompile it later on and spend CPU
   * make deployment an error
Second behavior seems to be more correct, as it will also protect our runtime from attempts to recompile later on.
@evgenykuzyakov probably your question was not regarding if we shall return an error, but more liek what kind of error shall be returned here. In review feedback I introduced `DataProcessingError` which tries to distinguish it from the storage error.

matklad: Ah, sorry, I was confused by the peculiar signature of `precompile`. It returns an `Option<VMError>`. I would expect it to return `Result<(), VMError>`. The tangible benefit there would be that it would be impossible to forget to check the error: compiler warns about unused results, but not about unused options. And map_err would also work then.

olonho:
> Would this cause a change in deduct deploy cost?

It should increase deployment cost a bit, need to come up with measurement scheme.

runtime/runtime/src/actions.rs
olonho: Regarding protocol change - that's an interesting question. You mean protocol changes because we now fail earlier than before in case of invalid contracts, right? Could you please recommend what else to be done to migrate to the new protocol, but just bumping `PROTOCOL_VERSION` in version.rs?

bowenwang1996 (commented):
it seems that this changes the protocol

olonho:
Obsolete by now, closing.

PR #3860, opened 2021-01-27
Revert "Compress compiled contracts."

evgenykuzyakov:
Reverts near/nearcore#3855

What do we gain by compressing contracts in cache? It definitely slows down the runtime which decreases the function call TPS. What the cost effects of doing it?

There was no migration done from cache Key1 to Key2, so if a node runs a current protocol, then all contracts becomes uncompiled at once.

### DIFF
--- a/runtime/near-vm-runner/src/cache.rs
+++ b/runtime/near-vm-runner/src/cache.rs
@@ -7,7 +7,6 @@ use near_vm_errors::CacheError::{DeserializationError, ReadError, SerializationE
 use near_vm_errors::{CacheError, VMError};
 use near_vm_logic::{VMConfig, VMKind};
 use std::convert::TryFrom;
-use std::io::{Read, Write};
 use wasmer_runtime::{compiler_for_backend, Backend};
 use wasmer_runtime_core::cache::Artifact;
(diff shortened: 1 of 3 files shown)


runtime/near-vm-runner/src/cache.rs
```diff
@@ -47,11 +36,10 @@ fn get_key(code_hash: &[u8], code: &[u8], vm_kind: VMKind, config: &VMConfig) ->
         // Sometimes caller doesn't compute code_hash, so hash the code ourselves.
         Err(_e) => near_primitives::hash::hash(code),
     };
-    let key = ContractCacheKey::Version2 {
+    let key = ContractCacheKey::Version1 {
```
ailisp, line 39: we can keep this as version1, and hide version2 so node is exactly behaves as before. Then make some benches to see whether we should adopt version 2 for some or all contracts.

evgenykuzyakov: Revert is safer. Once we have proof that this change is needed for performance gains we can proceed to put it. Otherwise any new change should be guarded by a compilation feature in case it has a chance to affect the node performance.

ailisp (approved):

PR #3861, opened 2021-01-28
refactor(jsonrpc): Structured errors for `GetChunk` ViewClient handler and backward compatible response from JSON RPC

khorolets:
Test Plan
--------

* [x] The legacy input parameters parsed correctly (old tests pass)
* [x] The new input parameters parsed correctly (new test added)
* [x] The error messages are backward compatible after refactoring (old tests pass)

### DIFF
--- a/chain/client-primitives/src/types.rs
+++ b/chain/client-primitives/src/types.rs
@@ -215,16 +215,16 @@ impl Message for GetChunk {
 
 #[derive(thiserror::Error, Debug)]
 pub enum GetChunkError {
-    #[error("Shard id {0} does not exist")]
-    InvalidShardId(u64),
-    #[error("{0}")]
-    MismatchedVersion(String),
-    #[error("Chunk {0:?} is missing")]
-    ChunkMissing(ChunkHash),
     #[error("IO Error: {0}")]
     IOError(String),
-    #[error("Block not found")]
-    BlockNotFound(String),
+    #[error("Block has never been observed: {0}")]
+    UnknownBlock(String),
+    #[error("Block with hash `{0}` is unavailable on the node")]
+    UnavailableBlock(CryptoHash),
+    #[error("Shard ID {0} is invalid")]
+    InvalidShardId(u64),
+    #[error("Chunk {0:?} is missing")]
+    UnknownChunk(ChunkHash),
     // NOTE: Currently, the underlying errors are too broad, and while we tried to handle
     // expected cases, we cannot statically guarantee that no other errors will be returned
     // in the future.
@@ -236,13 +236,13 @@ pub enum GetChunkError {
 impl From<near_chain_primitives::Error> for GetChunkError {
     fn from(error: near_chain_primitives::Error) -> Self {
         match error.kind() {
+            near_chain_primitives::ErrorKind::IOErr(s) => Self::IOError(s),
+            near_chain_primitives::ErrorKind::DBNotFoundErr(s) => Self::UnknownBlock(s),
+            near_chain_primitives::ErrorKind::BlockMissing(hash) => Self::UnavailableBlock(hash),
             near_chain_primitives::ErrorKind::InvalidShardId(shard_id) => {
                 Self::InvalidShardId(shard_id)
             }
-            near_chain_primitives::ErrorKind::Other(s) => Self::MismatchedVersion(s),
-            near_chain_primitives::ErrorKind::IOErr(s) => Self::IOError(s),
-            near_chain_primitives::ErrorKind::DBNotFoundErr(s) => Self::BlockNotFound(s),
-            near_chain_primitives::ErrorKind::ChunkMissing(hash) => Self::ChunkMissing(hash),
+            near_chain_primitives::ErrorKind::ChunkMissing(hash) => Self::UnknownChunk(hash),
             _ => Self::Unreachable(error.to_string()),
         }
     }
--- a/chain/client-primitives/src/types.rs
+++ b/chain/client-primitives/src/types.rs
@@ -215,7 +215,7 @@ impl Message for GetChunk {
 
 #[derive(thiserror::Error, Debug)]
 pub enum GetChunkError {
-    #[error("Invalid shardId {0}")]
+    #[error("Shard id {0} does not exist")]
     InvalidShardId(u64),
     #[error("{0}")]
     MismatchedVersion(String),
--- a/chain/client-primitives/src/types.rs
+++ b/chain/client-primitives/src/types.rs
@@ -169,14 +169,12 @@ pub enum GetBlockError {
 }
 
 impl From<near_chain_primitives::Error> for GetBlockError {
-    fn from(error: near_chain_primitives::Error) -> GetBlockError {
+    fn from(error: near_chain_primitives::Error) -> Self {
         match error.kind() {
-            near_chain_primitives::ErrorKind::IOErr(s) => GetBlockError::IOError(s),
-            near_chain_primitives::ErrorKind::DBNotFoundErr(s) => GetBlockError::BlockNotFound(s),
-            near_chain_primitives::ErrorKind::BlockMissing(hash) => {
-                GetBlockError::BlockMissing(hash)
-            }
-            _ => GetBlockError::Unreachable(error.to_string()),
+            near_chain_primitives::ErrorKind::IOErr(s) => Self::IOError(s),
+            near_chain_primitives::ErrorKind::DBNotFoundErr(s) => Self::BlockNotFound(s),
+            near_chain_primitives::ErrorKind::BlockMissing(hash) => Self::BlockMissing(hash),
+            _ => Self::Unreachable(error.to_string()),
         }
     }
 }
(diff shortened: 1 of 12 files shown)


chain/client/src/view_client.rs
```diff
@@ -527,26 +527,31 @@ impl Handler<GetChunk> for ViewClientActor {
...
-            GetChunk::ChunkHash(chunk_hash) => self.chain.get_chunk(&chunk_hash).map(Clone::clone),
+
+        let chunk = match msg {
+            GetChunk::ChunkHash(chunk_hash) => {
+                self.chain.get_chunk(&chunk_hash).map(Clone::clone)?
+            }
             GetChunk::BlockHash(block_hash, shard_id) => {
                 let block = self.chain.get_block(&block_hash).map(Clone::clone);
-                get_chunk_from_block(block, shard_id, &mut self.chain)
+                get_chunk_from_block(block, shard_id, &mut self.chain)?
```
frol, line 537: I suggest we pass the block itself instead of a result of a block into the `get_chunk_from_block` helper function (please, update the function to reflect this change), so the usage would look like:

```rust
let block = self.chain.get_block(&block_hash)?.clone();
get_chunk_from_block(block, shard_id, &mut self.chain)?
```

chain/jsonrpc-primitives/src/types/blocks.rs
```diff
@@ -31,35 +39,45 @@ pub struct RpcBlockResponse {
     pub block_view: near_primitives::views::BlockView,
 }
 
+// near_client_primitives::types::GetBlock wants BlockReference from near_primitives
+// that's why this impl exists
```
frol, line 43: Thanks for explicitly stating this :+1:

chain/jsonrpc-primitives/src/types/chunks.rs
```diff
@@ -0,0 +1,128 @@
...
+#[derive(thiserror::Error, Debug)]
+pub enum RpcChunkError {
+    #[error("Shard id {0} does not exist")]
+    InvalidShardId(u64),
+    #[error("{0}")]
+    MismatchedVersion(String),
+    #[error("Chunk {0:?} is missing")]
+    ChunkMissing(near_primitives::sharding::ChunkHash),
+    #[error("Block not found")]
+    BlockNotFound(String),
```
frol, line 32: ^

chain/jsonrpc-primitives/src/types/chunks.rs
```diff
@@ -0,0 +1,128 @@
...
+            }
+            near_client_primitives::types::GetChunkError::ChunkMissing(hash) => {
+                Self::ChunkMissing(hash)
+            }
+            near_client_primitives::types::GetChunkError::BlockNotFound(s) => {
+                Self::BlockNotFound(s)
+            }
+            near_client_primitives::types::GetChunkError::IOError(s) => Self::InternalError(s),
+            near_client_primitives::types::GetChunkError::Unreachable(s) => {
+                println!("{}", &s);
```
frol, line 93: nit: clean up

chain/jsonrpc/src/lib.rs
```diff
@@ -632,24 +637,17 @@ impl JsonRpcHandler {
         &self,
         request_data: near_jsonrpc_primitives::types::blocks::RpcBlockRequest,
     ) -> Result<near_jsonrpc_primitives::types::blocks::RpcBlockResponse, RpcBlockError> {
```
frol, line 639: Following the style, we should use absolute path:

```suggestion
    ) -> Result<near_jsonrpc_primitives::types::blocks::RpcBlockResponse, near_jsonrpc_primitives::types::blocks::RpcBlockError> {
```

P.S. We may consider `use near_jsonrpc_primitives::types as near_jsonrpc_types;` at the top to shorten the lines a bit, if that helps at all.

chain/network/src/types.rs
```diff
@@ -618,7 +617,7 @@ impl RawRoutedMessage {
     }
 }
 
-#[derive(BorshSerialize, Serialize, PartialEq, Eq, Clone, Debug)]
+#[derive(BorshSerialize, PartialEq, Eq, Clone, Debug)]
```
frol, line 620: *For other reviewers:*

@khorolets and I had a chat about where should `BlockReference` structure belong, and we decided to keep `BlockReference` structure in core-primitives, but remove `serde` serialization/deserialization from it since it was JSON-RPC-specific and could easily confuse people or introduce a backward-incompatible change. We tried to do so, and identified unused serialization/deserialization derives on near-network structures.

@mfornet Do we need serde in `near-network` at all? It feels unnecessary to me, but I see a few uses, which I believe could be either extracted from `near-network` or replaced with Debug display instead of using JSON serialization.

frol (changes requested):

chain/jsonrpc-primitives/src/types/chunks.rs
```diff
@@ -0,0 +1,128 @@
...
+    #[serde(flatten)]
+    pub chunk_view: near_primitives::views::ChunkView,
+}
+
+#[derive(thiserror::Error, Debug)]
+pub enum RpcChunkError {
+    #[error("The node reached its limits. Try again later. More details: {0}")]
+    InternalError(String),
+    #[error("Block not found")]
+    BlockNotFound(String),
```
frol, line 28:
```suggestion
    #[error("Block has never been observed: {0}")]
    UnknownBlock(String),
```

chain/jsonrpc-primitives/src/types/chunks.rs
```diff
@@ -0,0 +1,128 @@
...
+    #[error("Block `{0}` is unavailable on the node")]
+    UnavailableBlock(near_primitives::hash::CryptoHash),
+    #[error("Shard id {0} does not exist")]
+    InvalidShardId(u64),
+    #[error("Chunk {0:?} is missing")]
+    ChunkMissing(near_primitives::sharding::ChunkHash),
```
frol, line 34:
```suggestion
    #[error("Chunk with hash {0:?} has never been observed")]
    UnknownChunk(near_primitives::sharding::ChunkHash),
```

frol (changes requested):

frol (approved):

chain/network/src/types.rs
frol: @bowenwang1996 Ok, we just cleaned up the unused bits that were needed to create a clear separation of concerns, and won't do anything beyond that.

bowenwang1996: What I meant is that this change will completely deprecate the rpc routing and we should be aware of that. I am fine doing with but would like to be sure that we are on the same page.

frol: I don't think rpc routing is relevant here. RPC routing is done through near-network, which uses borsh for serialization (RPC receives the data in JSON, deserializes it, view-client can still route the request through near-network, which will take the structure and serialize it with borsh, and thus JSON serialization is relevant only on the RPC side)

chain/jsonrpc-primitives/src/types/chunks.rs
```diff
@@ -0,0 +1,152 @@
...
+use serde_json::Value;
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(untagged)]
+pub enum ChunkReference {
+    BlockShardId(near_primitives::types::BlockId, near_primitives::types::ShardId),
+    BlockIdShardIdStruct(BlockShardId),
+    Hash(near_primitives::hash::CryptoHash),
+    HashStruct(ChunkId),
+}
```
frol, line 11: Well, this is becoming ugly. Let's have another round of bikeshedding.

To reference a chunk we have two options:

1. Provide a block id (hash or height) and a shard id
2. Provide a chunk hash

There are no other options, so `ChunkReference` should only keep two enum variants. The support of the new API style should be isolated in the `parse` method trying to parse legacy one first, and if that fails, try to parse the new style of params.

Thus, let's go back to two enum variants here:

```suggestion
pub enum ChunkReference {
    BlockIdShardId { block_id: BlockId, shard_id: ShardId },
    ChunkHash { chunk_id: CryptoHash },
}
```

chain/jsonrpc-primitives/src/types/chunks.rs
```diff
@@ -0,0 +1,152 @@
...
+    #[error("The node reached its limits. Try again later. More details: {0}")]
+    InternalError(String),
+    #[error("Block not found")]
+    UnknownBlock(String),
+    #[error("Block `{0}` is unavailable on the node")]
+    UnavailableBlock(near_primitives::hash::CryptoHash),
+    #[error("Shard id {0} does not exist")]
+    InvalidShardId(u64),
+    #[error("Chunk {0:?} is missing")]
+    UnknownChunk(near_primitives::sharding::ChunkHash),
```
frol, line 47: Update the error messages to match those from GetChunkError in view-client.

chain/client-primitives/src/types.rs
```diff
@@ -212,11 +210,46 @@ pub enum GetChunk {
...
+pub enum GetChunkError {
+    #[error("IO Error: {0}")]
+    IOError(String),
+    #[error("Block has never been observed: {0}")]
+    UnknownBlock(String),
+    #[error("Block with hash `{0}` is unavailable on the node")]
+    UnavailableBlock(CryptoHash),
+    #[error("Shard ID {0} is invalid")]
+    InvalidShardId(u64),
+    #[error("Chunk {0:?} is missing")]
```
frol, line 226:
```suggestion
    #[error("Chunk with hash {0:?} has never been observed on this node")]
```

frol (changes requested):

chain/network/src/types.rs
bowenwang1996: Yeah good point. I cannot remember why we need `Serialize` here. Maybe it was simply a mistake

chain/jsonrpc-primitives/src/types/chunks.rs
```diff
@@ -0,0 +1,140 @@
...
+}
+
+impl RpcChunkRequest {
+    pub fn parse(value: Option<Value>) -> Result<Self, crate::errors::RpcParseError> {
+        // Try to parse legacy positioned args and if it fails parse newer named args
+        let chunk_reference = if let Ok((chunk_id,)) =
+            crate::utils::parse_params::<(near_primitives::hash::CryptoHash,)>(value.clone())
+        {
+            ChunkReference::ChunkHash { chunk_id }
+        } else if let Ok((block_id, shard_id)) = crate::utils::parse_params::<(
```
frol, line 69: I believe it should be a tuple of tuple of block id and shard id. Please, confirm that all the expected variants we discussed work

frol (changes requested):

chain/jsonrpc-primitives/src/types/chunks.rs
khorolets: I have checked manually with httpie. I can provide logs on Monday for history and to be 100% sure.

khorolets: We have http based test if recall correctly. Maybe it's worth to be extended with such check, what do you think @frol ?

frol: Yes, let's make sure we have tests for all the legacy and the new ways of passing the parameters.

Have a good weekend!

bowenwang1996 (approved):

frol (approved):

PR #3310, opened 2020-09-11
Cache compiled contracts.

olonho:

PR #3864, opened 2021-01-29
fix(runtime): bring back in memory cache

ailisp:
Mostly undo https://github.com/near/nearcore/pull/3310, on disk cache helper functions are kept and mark as allow(unused), once we plan to re-enable on disk cache it should be obvious to refer and undo this PR.

Test Plan
---------
CI


runtime/near-vm-runner/tests/test_error_cases.rs
```diff
@@ -530,16 +525,16 @@ fn test_external_call_error() {
         );
     });
 }
-#[test]
```
olonho, line 533: Why this change?

olonho (changes requested):
Think in memory cache shall be addition to on-disk cache, not replacing it. Just the deserialized data must be stored in in-memory cache, so code could be way simpler. Let me sketch it.

olonho:
Suggest we'd better use https://github.com/near/nearcore/pull/3865

runtime/near-vm-runner/tests/test_error_cases.rs
ailisp: This assert disk cache count from 0 to 1, which would fail due to on disk cache is not utilized

ailisp:
@olonho Sounds good, I'll close this one

PR #3866, opened 2021-01-29
ERC-721 CryptoZombies benchmark

frankbraun:
The different call context (compared to https://github.com/near/nearcore/blob/master/test-utils/testlib/src/standard_evm_cases.rs#L320-L356) still leads to a weird failure of the actual ERC-721 transfer.


runtime/near-evm-runner/benches/benches/cryptozombies.rs
```diff
@@ -36,3 +38,49 @@ pub fn deploy_code(c: &mut Criterion) {
...
+    let alice_address = near_evm_runner::utils::near_account_id_to_evm_address(&alice_account());
+    let bob_address = near_evm_runner::utils::near_account_id_to_evm_address(&bob_account());
+
+    c.bench_function("transfer_erc721", |b| {
+        // first zombie ID
+        let mut zombie_id = U256::zero();
+        b.iter(|| {
+            // create zombie
+            //let mut ext_copy = ext.clone();
+            evm_call(&mut ext, "alice", "call", args_create.clone(), false).0.unwrap();
```
ilblackdragon, line 68: create only works once per sender in the ZombieAttack contract. all the rest of creates should be failing:
https://github.com/near/nearcore/blob/master/runtime/near-evm-runner/tests/contracts/ZombieAttack.sol#L118

frankbraun: > create only works once per sender in the ZombieAttack contract. all the rest of creates should be failing:
> https://github.com/near/nearcore/blob/master/runtime/near-evm-runner/tests/contracts/ZombieAttack.sol#L118

Thanks for pointing that out, I missed that this was the reason the transfers failed.


> Also next line keeps trying to transfer zombie_id == 0 from alice to bob, but after the first transfer it will be already at bob's, so should fail as well.

I think that's not correct, I was increasing the `zombie_id` below, so that it should always transfer Alice's zombies.

Given that the contract only allows one zombie per sender, do you think it's OK for performance measurement if I just clone the state in every iteration, create a zombie with ID 0 and transfer that one? That would be the easiest and would still give an upper bound on transfer time.

I could also modify the Zombie contract to allow multiple zombies per sender or create more senders (with 1 zombie each).

frankbraun: I went with the first approach, cloning the state in every iteration.

runtime/near-evm-runner/benches/benches/cryptozombies.rs
```diff
@@ -36,3 +38,45 @@ pub fn deploy_code(c: &mut Criterion) {
...
+    let alice_address = near_evm_runner::utils::near_account_id_to_evm_address(&alice_account());
+    let bob_address = near_evm_runner::utils::near_account_id_to_evm_address(&bob_account());
+
+    c.bench_function("transfer_erc721", |b| {
+        b.iter(|| {
+            // create zombie
+            let mut ext_copy = ext.clone();
+            evm_call(&mut ext_copy, "alice", "call", args_create.clone(), false).0.unwrap();
+
+            // transfer ERC-721 zombie token
```
ailisp, line 68: shall we bench create zombie and transfer token separately?

frankbraun: Good point, since we measure creating a zombie already in another benchmark it probably makes sense to just measure the token transfer here. I'll change it accordingly.

frankbraun: I changed it accordingly.

frankbraun:
@artob This is ready to be merged.

artob (approved):
LGTM :+1:

PR #3870, opened 2021-01-29
fix: Fixing a regression in two network tests

SkidanovAlex:
BlockHeaderV3 was introduced as a nightly feature, but the tests were
changed to use it unconditionally, and thus fail when the nightly
features are disabled.

Test plan:
----------
Ran both tests locally before and after the change


Kouprin (approved):

pytest/tests/sanity/network_drop_package.py
```diff
@@ -26,7 +26,7 @@ def __init__(self, *args, **kwargs):
 
     async def handle(self, msg, fr, to):
         if msg.enum == 'Block':
-            h = msg.Block.BlockV2.header.BlockHeaderV3.inner_lite.height
+            h = msg.Block.BlockV2.header.inner_lite().height
```
Kouprin, line 29: do we need to do the same with Block and other structs? like `msg.Block.header().inner_lite().height`

SkidanovAlex: Ideally. But in this PR I only fixed the one that is broken atm.

PR #3874, opened 2021-02-01
refactor: more idiomatic Path handling

matklad:
Rust paths are not strings (they are a strict superset), so it's more
convenient to avoid strings altogether.

* use `.display()` function when printing a `Path`
* use `From<OsString>` when getting path from the outside world
* convert to `Path` as early as possible, and from `Path` as late if
  possible, if at all.


willemneal (approved):
LGTM!  Thanks!

runtime/runtime-params-estimator/src/testbed_runners.rs
```diff
@@ -48,7 +48,7 @@ pub struct Config {
     /// Number of the transactions in the block.
     pub block_sizes: Vec<usize>,
     /// Where state dump is located in case we need to create a testbed.
-    pub state_dump_path: String,
+    pub state_dump_path: PathBuf,
```
ailisp, line 51: Why this is PathBuf instead of Path?

matklad: `PathBuf` and `String` are owned containers (both backed by a `Vec<u8>` internally), `&Path` and `&str` are borrowed (pareametrized by a lifetime) counterparts. So this exchanges one owned container for another borrowed one with a more precise type.

Usning a borrowed container doesn't make much sense here: adding lifetime to a struct wound be a pain, and there isn't actually an existing owner we can borrow from here, so this won't even improve performance.

frol: @ailisp `PathBuf` is an owned version of `Path` (like `String` vs `str`)

frol (approved):

PR #3876, opened 2021-02-01
fix macos build related to nix::unistd::gettid

pmnoxx:
In one of the recent changes we started using `gettid` in `nix::unistd`.
This broke support for macos, this PR fixed that issue.


utils/near-performance-metrics/src/stats_enabled.rs
```diff
@@ -47,16 +47,6 @@ struct ThreadStats {
     classes: HashSet<&'static str>,
 }
 
-pub fn get_tid() -> usize {
-    let res = TID.with(|t| {
```
matklad, line 51: Shouldn't this also remove the TID threadlocal?

pmnoxx: You are right.

bowenwang1996 (approved):

matklad (approved):

PR #3880, opened 2021-02-01
fix(delay_detector): measure both cpu time and real time

ailisp:
Address #3841 , as real time can vary when system load is different. cpu time gives more accurate measurement in this case.

Test Plan
---------
CI. run a delay_detector local node, try both time measuring works.


tools/delay_detector/Cargo.toml
```diff
@@ -6,3 +6,4 @@ edition = "2018"
 
 [dependencies]
 log = "0.4"
+cpu-time = "1.0"
```
bowenwang1996, line 9: no newline here :)

bowenwang1996 (approved):

matklad:
Hm, is using cpu time correct here? I worry about monotonicity guarantees, especially when the process is migrated between CPUs. 

https://docs.rs/cpu-time/1.0.0/src/cpu_time/clock_gettime.rs.html#38

```
unsafe { clock_gettime(CLOCK_PROCESS_CPUTIME_ID, &mut time) }
...
    /// Returns the amount of CPU time used from the previous timestamp.
    pub fn duration_since(&self, timestamp: Self) -> Duration {
        self.0 - timestamp.0
    }
````

https://linux.die.net/man/3/clock_gettime

>The CLOCK_PROCESS_CPUTIME_ID and CLOCK_THREAD_CPUTIME_ID clocks are realized on many platforms using timers from the CPUs (TSC on i386, AR.ITC on Itanium). These registers may differ between CPUs and as a consequence these clocks may return bogus results if a process is migrated to another CPU.

So seems like this can panic?

https://doc.rust-lang.org/stable/src/core/time.rs.html#875-881

bowenwang1996:
@ailisp please take a look at what @matklad said

ailisp:
Good point. 
> The processors in an SMP system do not start all at exactly the same time and therefore the timer registers are typically running at an offset. Some architectures include code that attempts to limit these offsets on bootup. However, the code cannot guarantee to accurately tune the offsets. Glibc contains no provisions to deal with these offsets (unlike the Linux Kernel). Typically these offsets are small and therefore the effects may be negligible in most cases. 

So there's chance for small timing difference  it can panic on underflow. This cannot be prevent even when real time has >50ms difference, because real time can be much more than cpu time. Therefore don't have a good way to avoid the panic. I'm think of manually compare two ProcessTime(Duration) and not print when underflow, but it's not pub, so not doable before fix it in cpu_time crate

matklad:
We might do the same thing that std does for Instant::now:

https://github.com/rust-lang/rust/blob/b593389edbaa9ea0c90f0ed419283842f534e50a/library/std/src/time.rs#L223-L259

Just use some hammed and nails to make it monotonic (this is probably better done at the cpu_time crate's layer, so we need a fork/a pr. At any rate, raising an issue on the crate's issuer tracker would be good, it at last should docucent this footgun (or maintainer can explain why this isn't applicable :) ))

ailisp:
reported https://github.com/tailhook/cpu-time/issues/4, if author
acknowledged the issue, we can fix on cpu_time

ailisp:
@matklad cpu_time author replied that monotonicity holds for modern OSes (linux 2.4+): https://github.com/tailhook/cpu-time/issues/4#issuecomment-773225589. So it's good, WDYT?

matklad:
Yeah, lgtm!

bowenwang1996:
@ailisp can we merge this PR?

PR #3881, opened 2021-02-02
feat: Lower storage cost 10X

evgenykuzyakov:
Introduces a protocol change to lower storage cost 10X, to 1e19 per byte (current cost is 1e20 per byte).

This introduces a stable protocol feature instead of nightly, since we want to release it directly to testnet.

Gov forum discussion: https://gov.near.org/t/storage-staking-price/399

## Test plan:

- [x] CI
- [x] nayduck http://nayduck.eastus.cloudapp.azure.com:3000/#/run/1082
- [X] unit_test


core/runtime-configs/src/lib.rs
```diff
@@ -120,4 +148,18 @@ mod tests {
...
+    fn test_lower_cost() {
+        let config = Arc::new(RuntimeConfig::default());
+        let config_same = RuntimeConfig::from_protocol_version(&config, 0);
+        assert_eq!(
+            config_same.as_ref().storage_amount_per_byte,
+            config.as_ref().storage_amount_per_byte
+        );
+        let config_lower = RuntimeConfig::from_protocol_version(&config, ProtocolVersion::MAX);
+        assert!(
+            config_lower.as_ref().storage_amount_per_byte < config.as_ref().storage_amount_per_byte
```
bowenwang1996, line 162: can we assert that one is 10x of the other

bowenwang1996 (approved):
I thought we planned to reduce it by 20x. @ilblackdragon

core/runtime-configs/src/lib.rs
evgenykuzyakov: Default is different from mainnet genesis

olonho:
BTW, what is the motivation behind this change?

evgenykuzyakov:
> BTW, what is the motivation behind this change?

@olonho https://gov.near.org/t/storage-staking-price/399 and linked video discussion https://www.youtube.com/watch?v=Ge1Cfq-0y5w

ilblackdragon:
I would vote to go with 20x reduction as we discussed during the call.

evgenykuzyakov:
@ilblackdragon  @bowenwang1996 I didn't agree to 20X on the call. We should do official voting process to discuss 10X vs 20X. We also need to do this asap if we still want to ship it.

evgenykuzyakov:
After discussion on the forum and in NEAR Dev telegram group, we've decided to keep 10X reduction as implemented in this PR. It allows us to move to 20X reduction later when we need it.

PR #3882, opened 2021-02-02
rewrite graph struct to be 200 times faster

pmnoxx:
It looks like it's possible to rewrite graph class to be 200 times faster. Do we really need to move computation to another thread?


chain/network/src/routing.rs
```diff
@@ -855,6 +857,176 @@ impl Graph {
...
+            }
+            Entry::Occupied(v) => *v.get(),
+        }
+    }
+
+    fn add_directed_edge(&mut self, peer0: PeerId, peer1: PeerId) {
+        let id0 = self.get_id(&peer0);
+        let id1 = self.get_id(&peer1);
+
+        self.adjacency.entry(id0).or_insert_with(HashSet::new).insert(id1);
```
matklad, line 922:
```suggestion
        self.adjacency.entry(id0).or_default().insert(id1);
```

chain/network/src/routing.rs
```diff
@@ -855,6 +857,176 @@ impl Graph {
...
+ */
+
+#[derive(Clone)]
+pub struct Graph {
+    pub source: PeerId,
+    p2id: HashMap<PeerId, u64>,
+    id2p: HashMap<u64, PeerId>,
+    adjacency: HashMap<u64, HashSet<u64>>,
+    total_active_edges: u64,
+    nodes: u64,
```
matklad, line 869:
```suggestion
    id2p: Vec<PeerId>,
    adjacency: HashMap<u64, HashSet<u64>>,
    total_active_edges: u64,
```

This seems to be a more minimal representation -- `nodes == id2p.len()`, and looks like we don't actually need a hash map as ids are sequential?

chain/network/src/routing.rs
```diff
@@ -855,6 +857,176 @@ impl Graph {
...
+        res.nodes += 1;
+        res.id2p.insert(0, source.clone());
+        res.p2id.insert(source, 0);
+
+        res
+    }
+
+    fn contains_edge(&mut self, peer0: &PeerId, peer1: &PeerId) -> bool {
+        if let Some(id0) = self.p2id.get(&peer0) {
+            if let Some(id1) = self.p2id.get(&peer1) {
```
matklad, line 891: Seems like this either wants to be `&self`, or wants to use `get_id` fn.

chain/network/src/routing.rs
```diff
@@ -855,6 +857,176 @@ impl Graph {
...
+    // TODO(MarX, #1363): This is too slow right now. (See benchmarks)
+    /// Compute for every node `u` on the graph (other than `source`) which are the neighbors of
+    /// `sources` which belong to the shortest path from `source` to `u`. Nodes that are
+    /// not connected to `source` will not appear in the result.
+    pub fn calculate_distance(&self) -> HashMap<PeerId, HashSet<PeerId>> {
+        let mut queue = vec![];
+        let mut distance: HashMap<u64, i32> = HashMap::new();
+        // TODO(MarX, #1363): Represent routes more efficiently at least while calculating distances
+        //let mut routes: HashMap<u32, HashSet<u32>> = HashMap::new();
+        let mut routes: HashMap<u64, u128> = HashMap::new();
```
matklad, line 957: As ids are sequential, these could be `Vec`s

chain/network/src/routing.rs
```diff
@@ -855,6 +857,176 @@ impl Graph {
...
+            if let Some(neighbors) = self.adjacency.get(&cur_peer) {
+                for neighbor in neighbors {
+                    if let Entry::Vacant(entry) = distance.entry(*neighbor) {
+                        queue.push(*entry.key());
+                        entry.insert(cur_distance + 1);
+                        routes.insert(neighbor.clone(), 0);
+                    }
+                    // If this edge belong to a shortest path, all paths to
+                    // the closer nodes are also valid for the current node.
+                    if *distance.get(neighbor).unwrap() == cur_distance + 1 {
```
matklad, line 998: Seems like this can be written as a single `*entry().get_or_insert()` to avoid double lookup in the hashmap and unwrap.

chain/network/src/routing.rs
```diff
@@ -855,6 +857,176 @@ impl Graph {
...
 }
 
+ */
+
+#[derive(Clone)]
+pub struct Graph {
+    pub source: PeerId,
+    p2id: HashMap<PeerId, u64>,
+    id2p: HashMap<u64, PeerId>,
+    adjacency: HashMap<u64, HashSet<u64>>,
```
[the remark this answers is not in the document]
pmnoxx: Rust's default hash map used to be slow until https://github.com/rust-lang/rust/pull/58623
https://github.com/rust-lang/hashbrown has map implementation got merge into rust.

I checked performance by replacing default HashMap  with hashbrown implementation and performance was the same.

chain/network/src/routing.rs
pmnoxx: Nodes may get added and removed. The current code may leek memory in id2p, p2id if all edges of given code are removed. This code is just a proof of concept to show that performance can be optimized. This was not meant to be reviewed yet, as it's a draft.

pmnoxx: I'll have to add reference counting for Peers to know when they get removed. Once I do that, I can add another data structure with unused ids, and then I would be able to id2p to vector.

chain/network/src/routing.rs
pmnoxx: Yes, or we could make distance/routes into an array once I make other changes

chain/network/src/routing.rs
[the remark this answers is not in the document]
matklad: There are two things that can vary for the hash map

* hash map implementation
* hash function

Rust's hash map *implementation* is indeed the fastest general hashmap in the world. But it is generic over hash function, and, depending on the specific hash function, the preformance may vary. 

In particular, `std::collections::HashMap` uses `SipHasher`, and `rustc_hash::HashMap` is using `FxHash`. The unerlying hash map implementation is still `std::collections::HashMap`:

https://github.com/rust-lang/rustc-hash/blob/5e09ea0a1c7ab7e4f9e27771f5a0e5a36c58d1bb/src/lib.rs#L43

``` 
pub type FxHashMap<K, V> = HashMap<K, V, BuildHasherDefault<FxHasher>>;
```

The choice of hash function varies, the unscientific benchmarg shows that FxHasher is 2x faster for lookup of integer keys:

```
build std::collections::HashSet: 4.32s
build rustc_hash::FxHashSet:     3.94s

lookup std::collections::HasSet: 5.32s
lookup rustc_hash::FxHashSet:    1.93s
```

https://github.com/matklad/hashbench

That being said, `hashbrown` crate uses aHash by default (unlike std, it doesn't guarantee DOS resistance), which should be in the same ballpark as FxHash, so, it there's no difference, than hash map lookups are not the bottleneck in this case.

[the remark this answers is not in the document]
pmnoxx: I know. That exactly the reason why I tested changing HashMap to the one using hashbrown.

Most likely the current bottleneck is constructing the result itself, which can be quite large. In the benchmark test with 100 nodes, we end up creating a HashMap<PeerId, HashSet<PeerId>>  containing around 10k instances PeerId. Most time is likely spend there. We could instead use HashMap<u64, HashSet<u64>>. However, this change is self contained and doesn't require refactoring rest of the code.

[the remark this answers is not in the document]
matklad: Ah, sorry, I misunderstood the comment about hashbrown!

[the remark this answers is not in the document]
pmnoxx: Actually rustc_hash did help. Previously when I was testing hashmap vs hashbrown, I was testing the old code. It was bottlenecked by PeerId. Now that I use u64 ids it does help.

chain/network/src/routing.rs
```diff
@@ -758,100 +759,184 @@ pub struct RoutingTableInfo {
...
+
+    fn compute_result(&self, routes: &mut Vec<u128>) -> HashMap<PeerId, HashSet<PeerId>> {
+        let source_id = 0;
+        let mut result = HashMap::new();
+        for key in 0..routes.len() {
+            let cur_route = routes[key];
+            let mut peer_set: HashSet<PeerId> = HashSet::new();
+            if let Some(neighbors) = self.adjacency.get(&source_id) {
+                let mut id = 0;
+                for neighbor in neighbors {
```
matklad, line 923:
```suggestion
                for (id, neighbor) in neighbors.iter().enumerate().take(128) {
```

chain/network/src/routing.rs
```diff
@@ -758,100 +759,180 @@ pub struct RoutingTableInfo {
...
+    fn add_edge_internal(&mut self, peer0: PeerId, peer1: PeerId) {
+        if peer0 != peer1 {
+            let id0 = self.get_id(&peer0);
+            let id1 = self.get_id(&peer1);
+
+            self.adjacency.entry(id0).or_default().insert(id1);
+            self.adjacency.entry(id1).or_default().insert(id0);
+
+            self.change_counter(id0, 1);
+            self.change_counter(id1, 1);
```
matklad, line 835: Is the graph undirected? If that's the case, I thing we don't need to maintain counts, as the count seems to always be equal to `self.adjecency[id].len()`? Can we use `adjecency[id].is_empty()` as a gc condition(/

chain/network/src/routing.rs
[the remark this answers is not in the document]
matklad: This gets curiouser and curiouser, I don't understand why lookups are more costly than constructing a (std) hash map

https://users.rust-lang.org/t/building-hash-map-is-cheaper-than-doing-lookups/55044


But this is unrelateated to the pr

chain/network/src/routing.rs
```diff
@@ -161,13 +164,17 @@ impl Edge {
...
+                //     let _d =
+                //           NewDelayDetector::new("verify_edge_added".into(), Duration::from_millis(1));
                 let data = self.hash();
 
                 self.removal_info.is_none()
                     && self.signature0.verify(data.as_ref(), &self.peer0.public_key())
                     && self.signature1.verify(data.as_ref(), &self.peer1.public_key())
             }
             EdgeType::Removed => {
+                //let _d =
```
bowenwang1996, line 176: should this be removed?

chain/network/src/routing.rs
```diff
@@ -739,8 +756,18 @@ impl RoutingTable {
         }
     }
 
-    pub fn get_raw_graph(&self) -> &HashMap<PeerId, HashSet<PeerId>> {
-        &self.raw_graph.adjacency
+    // This method is slow
+    pub fn get_raw_graph(&self) -> HashMap<PeerId, HashSet<PeerId>> {
+        let mut result = HashMap::new();
```
bowenwang1996, line 761: it is probably faster to use `with_capacity` to avoid multiple allocations

chain/network/src/routing.rs
```diff
@@ -758,100 +785,180 @@ pub struct RoutingTableInfo {
...
-        distance.insert(&self.source, 0);
-
-        // Add active connections
-        if let Some(neighbors) = self.adjacency.get(&self.source) {
-            for neighbor in neighbors {
-                queue.push(neighbor);
-                distance.insert(neighbor, 1);
-                routes.insert(neighbor.clone(), vec![neighbor.clone()].drain(..).collect());
+    pub fn calculate_distance(&self) -> HashMap<PeerId, Vec<PeerId>> {
+        let _d = NewDelayDetector::new("calculate_distance".into(), Duration::from_millis(1));
```
bowenwang1996, line 890: if you want to keep this it is better to hide it behind a feature flag

utils/near-performance-metrics/src/stats_enabled.rs
```diff
@@ -351,3 +360,24 @@ pub fn print_performance_stats(sleep_time: Duration) {
         }
     }
 }
+
+pub struct NewDelayDetector<'a> {
```
bowenwang1996, line 364: what is the difference between this and `DelayDetector`?

utils/deepsize/Cargo.toml.orig
```diff
@@ -0,0 +1,27 @@
+[package]
```
bowenwang1996, line 1: why do we put this inside nearcore?

pmnoxx: In Rust you cannot declare theirs for external packages. By having deepsize here, we are able to do that.
I'll remove this crate for now and write a new crate from scratch.

chain/network/src/routing.rs
```diff
@@ -758,100 +780,175 @@ pub struct RoutingTableInfo {
...
         }
 
-        routes.into_iter().filter(|(_, hops)| !hops.is_empty()).collect()
+        self.compute_result(&mut routes, &distance)
+    }
+
+    fn compute_result(
+        &self,
+        routes: &mut Vec<u128>,
+        distance: &Vec<i32>,
```
matklad, line 928:
```suggestion
        routes: &[u128],
        distance: &[i32],
```

`&Vec<_>` is a "useless type": https://github.com/rust-analyzer/rust-analyzer/blob/master/docs/dev/style.md#useless-types

chain/network/src/routing.rs
```diff
@@ -758,100 +780,175 @@ pub struct RoutingTableInfo {
...
+            let neighbors = Vec::from_iter(neighbors.into_iter());
+            for (key, cur_route) in routes.iter().enumerate() {
+                if key == source_id || distance[key] == -1 {
+                    continue;
+                }
+                let mut peer_set: Vec<PeerId> = Vec::with_capacity(cur_route.count_ones() as usize);
+
+                for (id, neighbor) in neighbors.iter().enumerate().take(128) {
+                    if (cur_route & ((1 as u128) << id)) != 0 {
+                        peer_set.push(self.id2p.get(neighbor).unwrap().clone());
```
matklad, line 943:
```suggestion
                        peer_set.push(self.id2p[neighbor].clone());
```

chain/network/src/routing.rs
```diff
@@ -758,100 +780,175 @@ pub struct RoutingTableInfo {
...
+    fn compute_result(
+        &self,
+        routes: &mut Vec<u128>,
+        distance: &Vec<i32>,
+    ) -> HashMap<PeerId, Vec<PeerId>> {
+        let source_id = 0;
+        let mut result = HashMap::with_capacity(routes.len());
+
+        if let Some(neighbors) = self.adjacency.get(&source_id) {
+            let neighbors = Vec::from_iter(neighbors.into_iter());
```
matklad, line 934:
```suggestion
            let neighbors = Vec::from_iter(neighbors.into_iter());
```

why do we need this extra copy into a Vec? Seems like it should work without?

chain/network/src/routing.rs
```diff
@@ -758,100 +780,175 @@ pub struct RoutingTableInfo {
...
+            for (key, cur_route) in routes.iter().enumerate() {
+                if key == source_id || distance[key] == -1 {
+                    continue;
+                }
+                let mut peer_set: Vec<PeerId> = Vec::with_capacity(cur_route.count_ones() as usize);
+
+                for (id, neighbor) in neighbors.iter().enumerate().take(128) {
+                    if (cur_route & ((1 as u128) << id)) != 0 {
```
matklad, line 942:
```suggestion
                    if (cur_route & (1u128 << id)) != 0 {
```

chain/network/src/routing.rs
```diff
@@ -758,100 +780,175 @@ pub struct RoutingTableInfo {
...
+                        peer_set.push(self.id2p.get(neighbor).unwrap().clone());
+                    };
+                }
+                if peer_set.len() != 0 {
```
matklad, line 946:
```suggestion
                if !peer_set.is_empty() {
```

matklad (commented):
Mostly nits, but `total_active_edges` seems like it might affect correctness. 

Also, 100X perf improvement is nice :D

Also, I am wondering, does it make sense to try petgraph here? Although I myself also actually prefer to code simple graph routines myself.

chain/network/src/routing.rs
pmnoxx: Thanks, yes it's undirected. I simplified the code.

bowenwang1996 (commented):
with this change we cannot have more than 128 peers, so it might be worth adding a check when we load the network config.

chain/network/src/routing.rs
```diff
@@ -758,100 +780,175 @@ pub struct RoutingTableInfo {
...
         false
     }
 
-    fn add_directed_edge(&mut self, peer0: PeerId, peer1: PeerId) {
-        self.adjacency.entry(peer0).or_insert_with(HashSet::new).insert(peer1);
+    fn remove_if_unused(&mut self, id: usize) {
+        let entry = self.adjacency.entry(id).or_default();
+
+        if entry.len() == 0 && id != 0 {
+            let peer = self.id2p.get(&id).take().cloned().unwrap();
```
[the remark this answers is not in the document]
pmnoxx: You are right.

chain/network/src/routing.rs
```diff
@@ -739,8 +751,18 @@ impl RoutingTable {
         }
     }
 
-    pub fn get_raw_graph(&self) -> &HashMap<PeerId, HashSet<PeerId>> {
-        &self.raw_graph.adjacency
+    // This method is slow
+    pub fn get_raw_graph(&self) -> HashMap<PeerId, HashSet<PeerId>> {
+        let mut result = HashMap::with_capacity(self.raw_graph.adjacency.len());
```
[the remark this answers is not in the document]
pmnoxx: Doing a search in our project on `let res = ` and `let result =` gave 65 and 73 results. I'll switch to use res in this file to be consistent in one file

pmnoxx:
> Mostly nits, but `total_active_edges` seems like it might affect correctness.
> 
> Also, 100X perf improvement is nice :D
> 
> Also, I am wondering, does it make sense to try petgraph here? Although I myself also actually prefer to code simple graph routines myself.
It's probably not worth it. We care about performance in this piece of code, so it's better if we write it ourselves.

chain/network/src/routing.rs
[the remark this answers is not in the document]
pmnoxx: We are not consisted on using one naming schema.

chain/network/src/routing.rs
```diff
@@ -758,100 +780,175 @@ pub struct RoutingTableInfo {
 #[derive(Clone)]
 pub struct Graph {
     pub source: PeerId,
-    adjacency: HashMap<PeerId, HashSet<PeerId>>,
+    p2id: HashMap<PeerId, usize>,
+    id2p: FxHashMap<usize, PeerId>,
+    adjacency: FxHashMap<usize, FxHashSet<usize>>,
     total_active_edges: u64,
+    unused: FxHashSet<usize>,
```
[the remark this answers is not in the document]
pmnoxx: It did improve performance by 20% and make code cleaner. Although I had to rewrite that code to make it work.

chain/network/src/routing.rs
```diff
@@ -758,100 +780,175 @@ pub struct RoutingTableInfo {
...
             self.total_active_edges += 1;
         }
     }
 
     pub fn remove_edge(&mut self, peer0: &PeerId, peer1: &PeerId) {
         if self.contains_edge(&peer0, &peer1) {
-            self.remove_directed_edge(&peer0, &peer1);
-            self.remove_directed_edge(&peer1, &peer0);
+            self.remove_edge_internal(&peer0, &peer1);
             self.total_active_edges -= 1;
```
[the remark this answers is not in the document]
pmnoxx: Why do you think this function can panic?

pmnoxx:
> with this change we cannot have more than 128 peers, so it might be worth adding a check when we load the network config.

 I added a check

chain/network/src/routing.rs
```diff
@@ -752,106 +766,172 @@ pub struct ProcessEdgeResult {
...
 
 #[derive(Clone)]
 pub struct Graph {
     pub source: PeerId,
-    adjacency: HashMap<PeerId, HashSet<PeerId>>,
+    source_id: usize,
+    p2id: HashMap<PeerId, usize>,
+    id2p: Vec<PeerId>,
+    used: Vec<bool>,
+    unused: Vec<usize>,
```
matklad, line 779: If we have some kind of utils modue, it might make sense to extract these three fields into a `struct NeArena<T>`, such that they are independent of the graph details. No strong preference either way here.

chain/network/src/routing.rs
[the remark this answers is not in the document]
matklad: Now it cant, but in the previous revision doing something like 

```
let mut g = Graph::new(source);
g.remove_edge()
```

would have executed `0u64 - 1`, and that panics if overflow checks are on.

chain/network/src/routing.rs
pmnoxx: I implemented this structure already we can bring it back: https://github.com/near/nearcore/blob/3749f2b640a7bad48e1843c8c4b99a0a91465b14/chain/network/src/freelist.rs

@bowenwang1996 The question is what would we do with that code.
a) leave it as it is in chain/network/src/freelist.rs
b) move it to crates.io, have someone finish it
c) rewrite Arena to not uses indexes add it to our code
d) rewrite Arena to not uses indexes and publish it
e) do nothing

pmnoxx: If we have some free time to spend, we can optimize the performance of this code further by another 2-4x. It would probably benefit us more.

bowenwang1996: let's not do that for now. It is not a priority

chain/network/src/routing.rs
```diff
@@ -739,8 +741,19 @@ impl RoutingTable {
         }
     }
 
-    pub fn get_raw_graph(&self) -> &HashMap<PeerId, HashSet<PeerId>> {
-        &self.raw_graph.adjacency
+    pub fn get_raw_graph(&self) -> HashMap<PeerId, HashSet<PeerId>> {
```
bowenwang1996, line 744: looks like this function is only used when we enable `metric_recorder`. It might be better to guard the function with the feature so that we don't use it anywhere else (as it is expensive)

bowenwang1996 (approved):
looks good to me. @mfornet can you also review?

chain/network/src/routing.rs
```diff
@@ -752,106 +765,177 @@ pub struct ProcessEdgeResult {
...
-        // Add active connections
-        if let Some(neighbors) = self.adjacency.get(&self.source) {
-            for neighbor in neighbors {
-                queue.push(neighbor);
-                distance.insert(neighbor, 1);
-                routes.insert(neighbor.clone(), vec![neighbor.clone()].drain(..).collect());
-            }
-        }
+    pub fn calculate_distance(&self) -> HashMap<PeerId, Vec<PeerId>> {
+        // TODO add removal of unreachable nodes
```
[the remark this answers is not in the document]
pmnoxx: No, it's also not trivial to do this correctly.
Let's say we our node is A.
We add an edge B<->C;
We run calculate_distance that also removes unreachable nodes. Removing B and C.
Then we add A<->B.

This would lead to us to removing a completely valid edges.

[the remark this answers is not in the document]
pmnoxx: I added a warning, when number of unreachable nodes is greater than 1000.

PR #3883, opened 2021-02-02
feat (jsonrpc): Add EXPERIMENTAL_receipt jsonrpc method

khorolets:
Add `EXPERIMENTAL_receipt` jsonrpc method

Test for method is in progress

### DIFF
--- a/chain/jsonrpc-primitives/src/types/receipts.rs
+++ b/chain/jsonrpc-primitives/src/types/receipts.rs
@@ -1,0 +1,82 @@
+use serde::{Deserialize, Serialize};
+use serde_json::Value;
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct ReceiptReference {
+    pub receipt_id: near_primitives::hash::CryptoHash,
+}
+
+#[derive(Serialize, Deserialize)]
+pub struct RpcReceiptRequest {
+    #[serde(flatten)]
+    pub receipt_reference: ReceiptReference,
+}
+
+#[derive(Serialize, Deserialize)]
+pub struct RpcReceiptResponse {
+    #[serde(flatten)]
+    pub receipt_view: Option<near_primitives::views::ReceiptView>,
+}
+
+#[derive(thiserror::Error, Debug)]
+pub enum RpcReceiptError {
+    #[error("The node reached its limits. Try again later. More details: {0}")]
+    InternalError(String),
+    // NOTE: Currently, the underlying errors are too broad, and while we tried to handle
+    // expected cases, we cannot statically guarantee that no other errors will be returned
+    // in the future.
+    // TODO #3851: Remove this variant once we can exhaustively match all the underlying errors
+    #[error("It is a bug if you receive this error type, please, report this incident: https://github.com/near/nearcore/issues/new/choose. Details: {0}")]
+    Unreachable(String),
+}
+
+impl From<ReceiptReference> for near_client_primitives::types::GetReceipt {
+    fn from(receipt_reference: ReceiptReference) -> Self {
+        Self { receipt_id: receipt_reference.receipt_id }
+    }
+}
+
+impl RpcReceiptRequest {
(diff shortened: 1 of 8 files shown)


chain/jsonrpc-primitives/src/types/receipts.rs
```diff
@@ -0,0 +1,82 @@
...
+#[derive(Serialize, Deserialize)]
+pub struct RpcReceiptRequest {
+    #[serde(flatten)]
+    pub receipt_reference: ReceiptReference,
+}
+
+#[derive(Serialize, Deserialize)]
+pub struct RpcReceiptResponse {
+    #[serde(flatten)]
+    pub receipt_view: Option<near_primitives::views::ReceiptView>,
```
frol, line 18: I think it is better to respond with a `ReceiptView`, and if there is no receipt found (`Option::None` variant) we can respond with `RpcReceiptError::UnknownReceipt`.

chain/jsonrpc/client/src/lib.rs
```diff
@@ -189,6 +189,8 @@ jsonrpc_client!(pub struct JsonRpcClient {
     pub fn chunk(&self, id: ChunkId) -> RpcRequest<ChunkView>;
     pub fn validators(&self, block_id: MaybeBlockId) -> RpcRequest<EpochValidatorInfo>;
     pub fn gas_price(&self, block_id: MaybeBlockId) -> RpcRequest<GasPriceView>;
+    #[allow(non_snake_case)]
+    pub fn EXPERIMENTAL_receipt(&self, id: near_jsonrpc_primitives::types::receipts::ReceiptReference) -> RpcRequest<near_primitives::views::ReceiptView>;
```
frol, line 193: This style is deprecated since it only generates legacy positional arguments (sorry for the missing comment). Please, add the `impl` below following the example of `fn query`, where you pass `RpcReceiptRequest` and return `RpcRequest<RpcReceiptResponse>`.

frol (changes requested):

chain/jsonrpc-primitives/src/types/receipts.rs
```diff
@@ -0,0 +1,88 @@
...
+    }
+}
+
+impl From<RpcReceiptError> for crate::errors::RpcError {
+    fn from(error: RpcReceiptError) -> Self {
+        let error_data = match error {
+            RpcReceiptError::InternalError(_) => Some(Value::String(error.to_string())),
+            RpcReceiptError::UnknownReceipt(_) => Some(Value::String(error.to_string())),
+            RpcReceiptError::Unreachable(s) => Some(Value::String(s)),
+        };
```
frol, line 84: It seems that this should be fine:

```suggestion
        let error_data = Some(Value::String(error.to_string()));
```

chain/jsonrpc/src/lib.rs
```diff
@@ -652,6 +660,25 @@ impl JsonRpcHandler {
...
+    ) -> Result<
+        near_jsonrpc_primitives::types::receipts::RpcReceiptResponse,
+        near_jsonrpc_primitives::types::receipts::RpcReceiptError,
+    > {
+        match self
+            .view_client_addr
+            .send(GetReceipt { receipt_id: request_data.receipt_reference.receipt_id })
+            .await??
+        {
+            Some(receip_view) => Ok(receip_view.into()),
```
frol, line 675:
```suggestion
            Some(receipt_view) => Ok(receipt_view.into()),
```

frol (changes requested):

bowenwang1996 (approved):

chain/jsonrpc/tests/rpc_query.rs
```diff
@@ -1,10 +1,11 @@
 use std::convert::TryFrom;
 
 use actix::{Actor, System};
+use borsh::BorshSerialize;
```
frol, line 4: I believe we should just revert all the changes to this file.

frol (approved):

PR #3884, opened 2021-02-02
epoch_sync_data_hash

Kouprin:


neard/src/runtime.rs
```diff
@@ -1048,6 +1048,19 @@ impl RuntimeAdapter for NightshadeRuntime {
         Ok(epoch_manager.get_epoch_info(epoch_id)?.minted_amount)
     }
 
+    fn get_epoch_sync_data_hash(
+        &self,
+        prev_block_hash: &CryptoHash,
+        prev_epoch_id: &EpochId,
+        epoch_id: &EpochId,
+    ) -> Result<CryptoHash, Error> {
```
bowenwang1996, line 1056: I think you want prev epoch info and prev prev epoch info instead of curren epoch info and prev epoch info

Kouprin:
This one won't be merged.

PR #3885, opened 2021-02-02
fix: `cargo t` works in the root of the repo

matklad:
This is a minimal diff (plus a couple of drive-by retactors) to squash the compilation error. A more thorough fix here is to actually not use cargo features, but I'll pursue that separately as it needs more deep understanding of the code that I have right now.

The core issue is that `cd runtime/runtime-params-estimator/ && cargo check --no-default-features` fails. That is the params-estimator crate itself uses its own feature wrong.


runtime/runtime-params-estimator/src/vm_estimator.rs
```diff
@@ -135,10 +136,7 @@ pub fn load_and_compile(
     }
 }
 
-#[cfg(feature = "lightbeam")]
-const USING_LIGHTBEAM: bool = true;
-#[cfg(not(feature = "lightbeam"))]
-const USING_LIGHTBEAM: bool = false;
+const USING_LIGHTBEAM: bool = cfg!(feature = "lightbeam");
```
willemneal, line 139: Nice!

willemneal (approved):
Thanks for looking into this.  It's been a real thorn in my side.

runtime/runtime-params-estimator/src/stats.rs
```diff
@@ -26,8 +26,11 @@ impl Measurements {
         block_size: usize,
         block_cost: ExecutionCost,
     ) {
+        #[cfg(feature = "costs_counting")]
         let ext_costs = node_runtime::EXT_COSTS_COUNTER
             .with(|f| f.borrow_mut().drain().collect::<HashMap<_, _>>());
+        #[cfg(not(feature = "costs_counting"))]
+        let ext_costs = HashMap::new();
```
olonho, line 33: What is expected semantics here? Guess without costs counting this code shalln't make much sense.

olonho (approved):
Guess we shall get rid of costs_counting eventually.

runtime/runtime-params-estimator/src/stats.rs
matklad: I don't know, this is just the simplest patch that makes this compilable, without introducing unused/unreachable warnings )

I think the right high-level semantics would be to print "cost counting is disabled" when trying to *get* the metrics, but I want to tackle that together with removing the cargo feature (I am prepping a little [likbez](https://en.wikipedia.org/wiki/Likbez) for that...)

matklad:
Here's a writeup about features, which suggest a more principled pattern for dealing with similar problems: https://gist.github.com/matklad/8adcf295f40297622e2164836f359c87

runtime/runtime-params-estimator/src/stats.rs
ailisp: Simply exit when cost_counting is not enabled?

ailisp: or exit much earlier in param estimator when compile without cost counting, and unreachable here. what exactly we want is to enforce "it must compile with cost counting"

ailisp (approved):

runtime/runtime-params-estimator/src/stats.rs
matklad: > "it must compile with cost counting"

This should do it: aa21023cae90142226ae618f2e5dbef96885fef6

evgenykuzyakov (approved):

olonho (approved):

PR #3888, opened 2021-02-02
fix memory_stats on mac

pmnoxx:
Disable memory_stats on mac


bowenwang1996 (approved):
Can you check that this fixes the issue on mac?

PR #3891, opened 2021-02-02
change peer_manager not to recompute routing multiple times

pmnoxx:
We should fix peer_manager not to schedule multiple computations of routing table at the same time.


chain/network/src/peer_manager.rs
```diff
@@ -568,18 +570,22 @@ impl PeerManagerActor {
...
-                file!(),
-                line!(),
-                duration,
-                |act, _ctx| {
-                    act.routing_table.update();
-                    #[cfg(feature = "metric_recorder")]
-                    act.metric_recorder.set_graph(act.routing_table.get_raw_graph())
-                },
-            );
+        if self.scheduled_routing_table_update == false {
```
bowenwang1996, line 573:
```suggestion
        if !self.scheduled_routing_table_update {
```

bowenwang1996 (approved):

PR #3894, opened 2021-02-03
fix: address circular dependency in PROTOCOL_FEATURE_MAPPING

### LINKED ISSUE #3893: Circular dependency in lazy static
It appears that there is some circular dependency in the definition of `PROTOCOL_FEATURES_TO_VERSION_MAPPING` https://github.com/near/nearcore/blob/6b49b20a844b6294b4b3e07a6119936d3cb7fec1/core/primitives/src/version.rs#L114 and it gets stuck when `STABLE_PROTOCOL_FEATURES_TO_VERSION_MAPPING` is not empty, since it tries to access itself in the loop where it is not yet defined.

bowenwang1996:
Addresses circular dependency mentioned in #3893. Fixes #3893 

Test plan
---------
CI

### DIFF
--- a/core/primitives/src/version.rs
+++ b/core/primitives/src/version.rs
@@ -131,7 +131,9 @@ lazy_static! {
             STABLE_PROTOCOL_FEATURES_TO_VERSION_MAPPING.iter()
         {
             assert!(
-                PROTOCOL_FEATURES_TO_VERSION_MAPPING[&stable_protocol_feature]
+                *nightly_protocol_features_to_version_mapping
+                    .get(&stable_protocol_feature)
+                    .unwrap_or(&stable_protocol_version)
                     >= *stable_protocol_version
             );
         }


frol (approved):

core/primitives/src/version.rs
```diff
@@ -131,7 +131,9 @@ lazy_static! {
             STABLE_PROTOCOL_FEATURES_TO_VERSION_MAPPING.iter()
         {
             assert!(
-                PROTOCOL_FEATURES_TO_VERSION_MAPPING[&stable_protocol_feature]
+                *nightly_protocol_features_to_version_mapping
+                    .get(&stable_protocol_feature)
```
matklad, line 135:
```suggestion
                    .get(stable_protocol_feature)
```

nit: this deref is probably unnneded.

matklad (commented):
LGTM!

I've seen both `once_cell` and `lazy_static` being used in the code. These crates have overlapping functionality, so I suggest normalizing to one of them. My 100% unbiased vote would go for once_cell.

Looking a bit beyond this tactical change, recursively inintialized statics looks complicated. I think this module would benefit from refactoring to use only a *single* lazy value, which holds a struct which keeps both maps as fields.

bowenwang1996:
@matklad why is `once_cell` preferred? I don't think I have used it before

matklad:
@bowenwang1996 we use it a bunch here: https://github.com/near/nearcore/blob/6ac5a2cc3464efde45e2e4bd3a0e67f329399efe/utils/near-performance-metrics/src/stats_enabled.rs#L28-L35

There's no anything particular earth shattering about once_cell, lazy_static works fine. But there's a number of smaller benefits that add up.

* it doesn't use macros, so it's friendlier for IDEs
* it is more flexible. In particular, you can use it to implement lazy fields, which is not possible with lazy static.
* it has Sync and !Sync versions
* it's on path for inclusion into std: https://github.com/rust-lang/rust/issues/74465

PR #3899, opened 2021-02-03
refactor: simplify gas_counter

matklad:


runtime/near-vm-logic/src/gas_counter.rs
```diff
@@ -16,12 +16,7 @@ thread_local! {
 
 #[cfg(all(feature = "costs_counting", feature = "protocol_feature_evm"))]
 pub fn reset_evm_gas_counter() -> u64 {
-    let mut ret = 0;
-    EVM_GAS_COUNTER.with(|f| {
-        ret = *f.borrow();
-        *f.borrow_mut() = 0;
-    });
-    ret
+    EVM_GAS_COUNTER.with(|f| f.replace(0))
```
matklad, line 19: [Jones's trick](http://giphygifs.s3.amazonaws.com/media/MS0fQBmGGMaRy/giphy.gif) can simplify a lot of things. Many types have `.replace` method: https://doc.rust-lang.org/stable/std/?search=replace. There's also `std::mem::replace`.

runtime/near-vm-logic/src/gas_counter.rs
```diff
@@ -95,56 +90,45 @@ impl GasCounter {
...
-
-    #[cfg(feature = "costs_counting")]
     #[inline]
     fn update_profile_action(&mut self, action: ActionCosts, value: u64) {
-        match &self.profile {
-            Some(profile) => profile.add_action_cost(action, value),
-            None => {}
-        };
+        if cfg!(feature = "costs_counting") {
+            match &self.profile {
```
bowenwang1996, line 125: is it better to use `if let` here

matklad: Yeah, let me refactor it as well while I am at it.

My rule of thumb is "use if let iff there's no else". With else, `match` actually reads better and I like that ignoring (`_`) is explicit.

frol (approved):

frol:
I have already restarted the `db-migration` CI test 4 times and it still fails. I don't expect this change to have any effect on the test, but it seems @matklad is super unlucky :smile:

matklad:
ping @evgenykuzyakov, I *think* this is blocked on review from you (although I might be missing some process here). If this looks good, I'll rebase.

frol:
@matklad please, rebase, I will merge it since Evgeny is out of the office.

matklad:
Ah! rebased @frol !

matklad:
And now rebased *properly*

PR #3903, opened 2021-02-04
fix(runtime): fix receipt id change

bowenwang1996:
#3821 introduces a bug in data id calculation as it directly passes in the current block hash instead of passing the prev block hash first and then upgrade to current block hash. This causes nodes with that commit to produce different receipts and therefore different state after they apply a block. The reason why we didn't catch it in CI is that backward_compatible.py does not have a function call that uses promise.

Test plan
----------
`backward_compatible.py` fails after the fix to the test on master but passes after the change in this PR.

### DIFF
--- a/core/primitives/src/utils.rs
+++ b/core/primitives/src/utils.rs
@@ -153,13 +153,14 @@ pub fn create_action_hash(
 pub fn create_data_id(
     protocol_version: ProtocolVersion,
     action_hash: &CryptoHash,
+    prev_block_hash: &CryptoHash,
     block_hash: &CryptoHash,
     data_index: usize,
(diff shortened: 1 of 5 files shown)


EgorKulikov (approved):
lgtm. Though I was pretty sure it used current block hash before receipt id change as well, guess I was mistaken

bowenwang1996:
> Though I was pretty sure it used current block hash before receipt id change as well, guess I was mistaken

It uses `last_block_hash` from `RuntimeExt`, `RuntimeExt` is constructed with `apply_state.last_block_hash` and before #3821 it was prev_block_hash

evgenykuzyakov (approved):

PR #3904, opened 2021-02-04
fix tests depending on feature = costs_counting

pmnoxx:
Tests are broken :/


matklad:
This should be the same as https://github.com/near/nearcore/pull/3885

pmnoxx:
> This should be the same as #3885

Tests are still not compiling. Now in one place instead of 3 places.

PR #3908, opened 2021-02-04
Migrate nearcore buildkite configs to the repo

chefsale:


matklad (approved):
LGTM as is!

But here is a bunch of further improvements we might do (probably better left to the follow-up PR), in the order of impact

* set `RUSTFLAGS='-D warnings' ` for *all* steps, and not only for sanity checks. wanrings depend on the sef of features activated, and we do have one unsilenced warning on master because of this.
* move `label`  to be the first element of each step -- purely a cosmetic change, but will improve readability a lot
* maybe add `cargo test --locked --workspace --no-run` before `cargo test --locked --workspace`. That way, we can easily track "how long does it take to compile" vs "how long does it take to run the unit tests"
* There's also a bunch of unobvious trick with caches to make compilation faster.

mhalambek (approved):

chefsale:
Yeah, @matklad please just create issues in near-ops repo and feel free to assign them to me. We need to do some cleaning up for sure.

PR #3912, opened 2021-02-04
tool(delay_detector): add configurable min delay

ailisp:
In runtime we often want to cache delay < 50ms, make it configurable so delays from other module still filtered at 50ms.

Test Plan
---------
Build with delay_detector it still works. use delay detector with shorter delay works

### DIFF
--- a/tools/delay_detector/src/lib.rs
+++ b/tools/delay_detector/src/lib.rs
@@ -12,7 +12,13 @@ pub struct DelayDetector<'a> {
 
 impl<'a> DelayDetector<'a> {
     pub fn new(msg: Cow<'a, str>) -> Self {
-        Self { msg, started: Instant::now(), snapshots: vec![], last_snapshot: None, min_delay: Duration::from_millis(50) }
+        Self {
+            msg,
+            started: Instant::now(),
+            snapshots: vec![],
+            last_snapshot: None,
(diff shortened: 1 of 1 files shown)


bowenwang1996 (approved):

tools/delay_detector/src/lib.rs
```diff
@@ -3,15 +3,20 @@ use std::borrow::Cow;
...
 
 impl<'a> DelayDetector<'a> {
+    pub fn new_with_min_delay(msg: Cow<'a, str>, min_delay: Duration) -> Self {
+        Self { msg, started: Instant::now(), snapshots: vec![], last_snapshot: None, min_delay }
+    }
+
     pub fn new(msg: Cow<'a, str>) -> Self {
-        Self { msg, started: Instant::now(), snapshots: vec![], last_snapshot: None }
+        Self::new_with_min_delay(msg, Duration::from_millis(50))
     }
```
matklad, line 20: I'd maybe go with a more composable API:


```rust
    pub fn new(msg: Cow<'a, str>) -> Self {
        Self { msg, started: Instant::now(), snapshots: vec![], last_snapshot: None, min_delay: Duration::from_millis(50) }
    }
    pub fn min_delay(mut self, min_delay: Duration) -> Self {
        self.min_delay = min_delay;
        self
    }
```

PR #3919, opened 2021-02-05
feat(rpc): add protocol config rpc endpoint

### LINKED ISSUE #3918: RPC for getting the current config
We currently have `EXPERIMENTAL_genesis_config` rpc endpoint which returns the genesis config. However, when we do upgrades, we might change the values in the config and therefore it is desirable to have some rpc endpoint that returns the config given a block.

bowenwang1996:
Add `EXPERIMENTAL_protocol_config` rpc endpoint that returns the protocol config given some block, since protocol config can change after genesis due to protocol upgrades. This will allow users to get the up-to-date protocol-level configs. Fixes #3918.

Test plan
----------
`test_protocol_config_rpc`


chain/jsonrpc-primitives/src/types/config.rs
```diff
@@ -0,0 +1,78 @@
...
+            near_client_primitives::types::GetProtocolConfigError::UnknownBlock(s) => {
+                Self::BlockNotFound(s)
+            }
+            near_client_primitives::types::GetProtocolConfigError::IOError(s) => {
+                Self::InternalError(s)
+            }
+            near_client_primitives::types::GetProtocolConfigError::Unreachable(s) => {
+                near_metrics::inc_counter_vec(
+                    &crate::metrics::RPC_UNREACHABLE_ERROR_COUNT,
+                    &["RpcBlockError", &s],
```
khorolets, line 52: This should be RpcProtocolConfigError

khorolets, line 52:
```suggestion
                    &["RpcProtocolConfigError", &s],
```

khorolets (approved):

chain/jsonrpc-primitives/src/types/config.rs
```diff
@@ -0,0 +1,78 @@
...
+#[derive(Serialize, Deserialize)]
+pub struct RpcProtocolConfigResponse {
+    #[serde(flatten)]
+    pub config_view: near_chain_configs::ProtocolConfigView,
+}
+
+#[derive(thiserror::Error, Debug)]
+pub enum RpcProtocolConfigError {
+    #[error("Block not found")]
+    BlockNotFound(String),
```
frol, line 29: We use `UnknownBlock` and `UnavailableBlock` to distinguish two possible `BlockNotFound` reasons, so I suggest we keep using it:

```suggestion
    #[error("Block has never been observed: {0}")]
    UnknownBlock(String),
```

@khorolets we need to go over the blocks implementation and fix it there (we already did it for chunks)

core/chain-configs/src/genesis_config.rs
```diff
@@ -272,3 +272,102 @@ pub fn get_initial_supply(records: &[StateRecord]) -> Balance {
...
+
+// Note: this type cannot be placed in primitives/src/view.rs because of `RuntimeConfig` dependency issues.
+// Ideally we should create `RuntimeConfigView`, but given the deeply nested nature and the number of fields inside
+// `RuntimeConfig`, it should be its own endeavor.
+#[derive(Serialize, Deserialize, Debug)]
+pub struct ProtocolConfigView {
+    /// Current Protocol Version
+    pub protocol_version: ProtocolVersion,
+    /// Official time of blockchain start.
+    pub genesis_time: DateTime<Utc>,
```
frol, line 284: We need to make sure that dates, `u64`-like types (BlockHeight) are serialized as strings due to JSON numbers limitation.

frol (approved):

core/chain-configs/src/genesis_config.rs
bowenwang1996: I believe it is. For example, https://github.com/near/nearcore/blob/3ad75b5bc8e71b3316507cf0d0fce6b21d3f8ad5/neard/res/genesis_config.json#L3

bowenwang1996:
@frol do you think we should deprecate `EXPERIMENTAL_genesis_config` (make it an alias to `EXPERIMENTAL_protocol_config`)? Or should we introduce `EXPERIMENTAL_protocol_config` separately in near-api-js? If we do the latter, we also need to make sure that wallet uses the new endpoint.

frol:
@bowenwang1996 Good point. While the naming `geneis_config` implies that it should be *genesis* config, I think it makes sense to actually return the latest runtime config since that is what most of the users wanted to have in the first place; thus, I think we should go with the alias :+1:

frol:
Well, on the second thought, it would not be sane to expose the latest protocol config under `genesis_config` endpoint, and it is better to address the issue on `near-api-js` level (add deprecation note on the method, and use `EXPERIMENTAL_protocol_config({"sync_checkpoint": "genesis"})` explicitly). I will take it from here.

bowenwang1996:
> Well, on the second thought, it would not be sane to expose the latest protocol config under genesis_config endpoint, and it is better to address the issue on near-api-js level (add deprecation note on the method, and use EXPERIMENTAL_protocol_config({"sync_checkpoint": "genesis"}) explicitly). I will take it from here.

Thanks @frol !

SkidanovAlex (approved):

frol:
Just for the reference, near-api-js support is coming: https://github.com/near/near-api-js/pull/508

PR #3928, opened 2021-02-08
fix: remove unused file

matklad:
The directory is otherwise empty, so I suppose this is a leftover?


willemneal (approved):
Good catch!

evgenykuzyakov (approved):

PR #3929, opened 2021-02-08
fix: update example contract

matklad:


matklad:
Ideally, I would like to add an integration test to `near-vm-runner-standalone` which runs `cargo build example-contract`, to make sure this doesn't regress in the future. I am not sure that would be a good idea. I have a couple of questions regarding to that:

* can we stomach CI time increase by CPU minute? Compiling even the empty contract takes a lot of time, as it uses full syn, and `near_sdk` is pretty heavy.
* what is the compatibility story between near-sdk and nearcore? The contract has to depend on a published version of nearsdk. Can it be the case that the published version of nearsdk becomes incompatible with the master of nearcore?

willemneal:
I think it would be a good idea to not depend an the sdk to compile the contract. @evgenykuzyakov had a recent example, but I can't find it.

matklad:
Yeah, actually, I've found a bunch of better maintained contracts in this repository. I think we can remove this example instead of upgrading it to the modern sdk version?

ailisp (approved):

frol (approved):

olonho (approved):

PR #3931, opened 2021-02-08
Improve print_stats method in near-performance-metrics.

pmnoxx:
We print every minute performance stats for all threads. It will be useful to print this information split by enum type.

### DIFF
--- a/utils/near-performance-metrics/src/stats_enabled.rs
+++ b/utils/near-performance-metrics/src/stats_enabled.rs
@@ -297,7 +297,7 @@ where
         0,
         took,
         ended,
-        msg_text.unwrap_or(&""),
+        msg_text.unwrap_or_default(),
     );
     result
 }
--- a/utils/near-performance-metrics/src/stats_enabled.rs
+++ b/utils/near-performance-metrics/src/stats_enabled.rs
@@ -42,7 +42,7 @@ struct Entry {
 }
 
 pub struct ThreadStats {
-    stat: HashMap<(&'static str, u32), Entry>,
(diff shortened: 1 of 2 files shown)


utils/near-performance-metrics/src/stats_enabled.rs
```diff
@@ -289,7 +291,14 @@ where
...
     let ended = Instant::now();
     let took = ended - now;
-    stat.lock().unwrap().log(class_name, std::any::type_name::<Message>(), 0, took, ended);
+    stat.lock().unwrap().log(
+        class_name,
+        std::any::type_name::<Message>(),
+        0,
+        took,
+        ended,
+        msg_text.unwrap_or(&""),
```
matklad, line 300:
```suggestion
        msg_text.unwrap_or_default(),
```

might also work

utils/near-performance-metrics/src/stats_enabled.rs
```diff
@@ -42,7 +42,7 @@ struct Entry {
 }
 
 pub struct ThreadStats {
-    stat: HashMap<(&'static str, u32), Entry>,
+    stat: HashMap<(&'static str, u32, &'static str), Entry>,
```
matklad, line 45: At this point we might want to introduce a struct with name fields for `(&'static str, u32, &'static str)`

pmnoxx: We could, but we are not using any of the fields by name. We simply print those 3 fields in that order.

bowenwang1996 (approved):

PR #3941, opened 2021-02-10
move StateSync out of client actor

pmnoxx:
We should be able to move StateSync out of client actor.


pmnoxx:
The change is not ready yet. I need to figure out how to synchronize `StateSync` between `ClientActor` and `SyncStateActor`.
I may be a good idea to check whenever anything else is missing.

PR #3942, opened 2021-02-10
Bowen error demo

Kouprin:


Kouprin:
The example is not quite vanilla as I expected to build. `prev_epoch_prev_last_block_info` is used in Epoch Sync finalization, not only in `save_block`. Closing.

Issue #3901, opened 2021-02-03
After actix upgrade, panic are not always properly propagated, and tests that `should_panic` fail.


bowenwang1996:
After #3869, `chunks_recovered_from_full_timeout_too_short` fails, which is likely related to the block timeout we set and the speed up from actix version upgrade.

SkidanovAlex:
@frol after looking into it, I observe the following:

#3869 updates all the tests except for the ones in `chunk_management.rs` to use `System::buider()` instead of `System::run()`. I applied the same change, in the following way:

```
fn chunks_produced_and_distributed_common(
    validator_groups: u64,
    drop_from_1_to_4: bool,
    block_timeout: u64,
) {
    init_test_logger();
    System::builder().stop_on_panic(true).run(move || {
```

It is consistent with how you updated other tests.

With this, when I run the test, I can see in the output that it indeed panics

```
thread 'chunks_recovered_from_full_timeout_too_short' panicked at 'assertion failed: `(left == right)`
  left: `2`,
 right: `3`', chain/client/tests/chunks_management.rs:106:21
stack backtrace:
```

This is expected behavior. However, the test doesn't stop and continues running. The panic doesn't propagate, and doesn't cause the test to fail.

bowenwang1996:
@matklad do you happen to know what may have caused this?

SkidanovAlex:
This is not a new problem, and we had infrastructure to handle it, but it was (inadvertently?) removed in the linked PR:

https://github.com/near/nearcore/pull/3869/files#diff-6e234c46051e0a6a189a87fdf9c34622301856c8df033bb8d3eac07688f71904L217

frol:
@SkidanovAlex the old way of setting up `init_stop_on_panic` was not working with the new version of actix, so I replaced it with what I assumed would work and which seemed to work (since CI passed even for `should_panic` tests).  I wanted to investigate the issue with this specific test (https://github.com/near/nearcore/pull/3869#discussion_r569700140), but after a brief chat with @bowenwang1996 I was under impression that the test itself is broken, and as he mentioned, the plan was to address this test in a separate issue.

@bowenwang1996 would you like me to take a look into it?

SkidanovAlex:
[...]

PR #3204, opened 2020-08-18
fix(rpc): structured errors

bowenwang1996:
Make rpc return structured errors instead of string and update the rpc error schema accordingly. Errors returned by rpc are now of the form
```
{type: <error_type>, content: <error_content>}
```
where content may contain inner errors.
 
Fixes #2976.
Fixes #2536 

Test plan
---------
Nightly tests and near-api-js tests

PR #3944, opened 2021-02-10
refactor(jsonrpc): Add structured errors to query method in ViewClient and RPC

### LINKED ISSUE #3848: View query times out for older block
**Describe the bug**
Running view query on a block ~60 days ago returns "Query has timed out", without providing guidance on what to do next.

**To Reproduce**
```
http post http://rpc.mainnet.near.org method=query params:='{"request_type": "call_function", "block_id": 23308486, "account_id": "staked.poolv1.near", "method_name": "get_total_staked_balance", "args_base64": "e30="}' jsonrpc=2.0 id=123
HTTP/1.1 200 OK
Via: 1.1 google
content-length: 106
content-type: application/json
date: Sun, 24 Jan 2021 13:31:31 GMT

{
    "error": {
        "code": -32000,
        "data": "query has timed out",
        "message": "Server error"
    },
    "id": "123",
    "jsonrpc": "2.0"
}
```

**Expected behavior**
Get result of view call or report that node is not archival and doesn't contain such data.

- mainnet

khorolets:
* add some errors to `runtime/runtime/state_viewer`
* introduce `QueryError` into chain primitives
* refactor runtime `query` method implementation in `neard`
* refactor `view_client` `handle_query` method
* remove routing for `query` method (resolves #3848)

Still need to implement errors on the RPC and Rosetta side. Work in progress. 

I'm creating this draft PR to ask for a review of error structs and error messages.

### DIFF
--- a/chain/chain-primitives/src/error.rs
+++ b/chain/chain-primitives/src/error.rs
@@ -22,12 +22,15 @@ pub enum QueryError {
     ContractCodeDoesNotExist { contract_account_id: near_primitives::types::AccountId },
     #[error("Access key for public key #{public_key} does not exist while viewing")]
     AccessKeyDoesNotExist { public_key: String },
-    #[error("Storage error occurred: {0:?}")]
-    StorageError(#[from] near_primitives::errors::StorageError),
+    #[error("Storage error occurred: #{storage_error:?}")]
+    StorageError {
+        #[from]
+        storage_error: near_primitives::errors::StorageError,
+    },
     #[error("Internal error occurred: #{error_message}")]
     InternalError { error_message: String },
-    #[error("VM error occurred: {0}")]
-    VMError(String),
+    #[error("VM error occurred: #{error_message}")]
+    VMError { error_message: String },
 }
 
 #[derive(Debug)]
--- a/chain/chain-primitives/src/error.rs
+++ b/chain/chain-primitives/src/error.rs
@@ -14,18 +14,18 @@ use near_primitives::types::{BlockHeight, EpochId, ShardId};
 
 #[derive(thiserror::Error, Debug)]
 pub enum QueryError {
-    #[error("Invalid account ID {0}")]
-    InvalidAccount(near_primitives::types::AccountId),
-    #[error("Account ID {0} does not exist while viewing")]
-    AccountDoesNotExist(near_primitives::types::AccountId),
-    #[error("Contract ID {0} code does not exist while viewing")]
-    ContractCodeDoesNotExist(near_primitives::types::AccountId),
-    #[error("Access key for public key {0} does not exist while viewing")]
-    AccessKeyDoesNotExist(String),
+    #[error("Invalid account ID #{requested_account_id}")]
+    InvalidAccount { requested_account_id: near_primitives::types::AccountId },
+    #[error("Account ID #{requested_account_id} does not exist while viewing")]
+    AccountDoesNotExist { requested_account_id: near_primitives::types::AccountId },
+    #[error("Contract ID #{contract_account_id} code does not exist while viewing")]
+    ContractCodeDoesNotExist { contract_account_id: near_primitives::types::AccountId },
+    #[error("Access key for public key #{public_key} does not exist while viewing")]
+    AccessKeyDoesNotExist { public_key: String },
     #[error("Storage error occurred: {0:?}")]
     StorageError(#[from] near_primitives::errors::StorageError),
-    #[error("Internal error occurred: {0}")]
-    InternalError(String),
+    #[error("Internal error occurred: #{error_message}")]
+    InternalError { error_message: String },
     #[error("VM error occurred: {0}")]
     VMError(String),
 }
--- a/chain/chain-primitives/src/error.rs
+++ b/chain/chain-primitives/src/error.rs
@@ -12,6 +12,24 @@ use near_primitives::serialize::to_base;
 use near_primitives::sharding::{ChunkHash, ShardChunkHeader};
 use near_primitives::types::{BlockHeight, EpochId, ShardId};
 
+#[derive(thiserror::Error, Debug)]
+pub enum QueryError {
+    #[error("Invalid account ID {0}")]
+    InvalidAccount(near_primitives::types::AccountId),
+    #[error("Account ID {0} does not exist while viewing")]
+    AccountDoesNotExist(near_primitives::types::AccountId),
+    #[error("Contract ID {0} code does not exist while viewing")]
+    ContractCodeDoesNotExist(near_primitives::types::AccountId),
+    #[error("Access key for public key {0} does not exist while viewing")]
+    AccessKeyDoesNotExist(String),
+    #[error("Storage error occurred: {0:?}")]
+    StorageError(#[from] near_primitives::errors::StorageError),
+    #[error("Internal error occurred: {0}")]
+    InternalError(String),
+    #[error("VM error occurred: {0}")]
+    VMError(String),
+}
+
 #[derive(Debug)]
 pub struct Error {
     inner: Context<ErrorKind>,
--- a/chain/chain-primitives/src/error.rs
+++ b/chain/chain-primitives/src/error.rs
@@ -14,14 +14,14 @@ use near_primitives::types::{BlockHeight, EpochId, ShardId};
 
 #[derive(thiserror::Error, Debug)]
 pub enum QueryError {
-    #[error("Invalid account ID #{requested_account_id}")]
+    #[error("Account ID #{requested_account_id} is invalid")]
     InvalidAccount { requested_account_id: near_primitives::types::AccountId },
     #[error("Account ID #{requested_account_id} does not exist while viewing")]
     AccountDoesNotExist { requested_account_id: near_primitives::types::AccountId },
     #[error("Contract ID #{contract_account_id} code does not exist while viewing")]
     ContractCodeDoesNotExist { contract_account_id: near_primitives::types::AccountId },
     #[error("Access key for public key #{public_key} does not exist while viewing")]
-    AccessKeyDoesNotExist { public_key: String },
+    AccessKeyDoesNotExist { public_key: near_crypto::PublicKey },
     #[error("Storage error occurred: #{storage_error:?}")]
     StorageError {
         #[from]
--- a/chain/chain-primitives/src/error.rs
+++ b/chain/chain-primitives/src/error.rs
@@ -16,21 +16,18 @@ use near_primitives::types::{BlockHeight, EpochId, ShardId};
 pub enum QueryError {
     #[error("Account ID #{requested_account_id} is invalid")]
     InvalidAccount { requested_account_id: near_primitives::types::AccountId },
-    #[error("Account ID #{requested_account_id} does not exist while viewing")]
-    AccountDoesNotExist { requested_account_id: near_primitives::types::AccountId },
-    #[error("Contract ID #{contract_account_id} code does not exist while viewing")]
-    ContractCodeDoesNotExist { contract_account_id: near_primitives::types::AccountId },
+    #[error("Account #{requested_account_id} does not exist while viewing")]
+    UnknownAccount { requested_account_id: near_primitives::types::AccountId },
+    #[error(
+        "Contract code for contract ID #{contract_account_id} has never been observed on the node"
+    )]
+    NoContractCode { contract_account_id: near_primitives::types::AccountId },
     #[error("Access key for public key #{public_key} does not exist while viewing")]
-    AccessKeyDoesNotExist { public_key: near_crypto::PublicKey },
-    #[error("Storage error occurred: #{storage_error:?}")]
-    StorageError {
-        #[from]
-        storage_error: near_primitives::errors::StorageError,
-    },
+    UnknownAccessKey { public_key: near_crypto::PublicKey },
     #[error("Internal error occurred: #{error_message}")]
     InternalError { error_message: String },
-    #[error("VM error occurred: #{error_message}")]
-    VMError { error_message: String },
+    #[error("Function call returned an error: #{error_message}")]
+    ContractExecutionError { error_message: String },
 }
 
 #[derive(Debug)]
--- a/chain/chain-primitives/src/error.rs
+++ b/chain/chain-primitives/src/error.rs
@@ -15,19 +15,43 @@ use near_primitives::types::{BlockHeight, EpochId, ShardId};
 #[derive(thiserror::Error, Debug)]
 pub enum QueryError {
     #[error("Account ID #{requested_account_id} is invalid")]
-    InvalidAccount { requested_account_id: near_primitives::types::AccountId },
+    InvalidAccount {
+        requested_account_id: near_primitives::types::AccountId,
+        block_height: near_primitives::types::BlockHeight,
+        block_hash: near_primitives::hash::CryptoHash,
+    },
     #[error("Account #{requested_account_id} does not exist while viewing")]
-    UnknownAccount { requested_account_id: near_primitives::types::AccountId },
+    UnknownAccount {
+        requested_account_id: near_primitives::types::AccountId,
+        block_height: near_primitives::types::BlockHeight,
+        block_hash: near_primitives::hash::CryptoHash,
+    },
     #[error(
         "Contract code for contract ID #{contract_account_id} has never been observed on the node"
     )]
-    NoContractCode { contract_account_id: near_primitives::types::AccountId },
+    NoContractCode {
+        contract_account_id: near_primitives::types::AccountId,
+        block_height: near_primitives::types::BlockHeight,
+        block_hash: near_primitives::hash::CryptoHash,
+    },
     #[error("Access key for public key #{public_key} does not exist while viewing")]
-    UnknownAccessKey { public_key: near_crypto::PublicKey },
+    UnknownAccessKey {
+        public_key: near_crypto::PublicKey,
+        block_height: near_primitives::types::BlockHeight,
+        block_hash: near_primitives::hash::CryptoHash,
+    },
     #[error("Internal error occurred: #{error_message}")]
-    InternalError { error_message: String },
+    InternalError {
+        error_message: String,
+        block_height: near_primitives::types::BlockHeight,
+        block_hash: near_primitives::hash::CryptoHash,
+    },
     #[error("Function call returned an error: #{error_message}")]
-    ContractExecutionError { error_message: String },
+    ContractExecutionError {
+        error_message: String,
+        block_height: near_primitives::types::BlockHeight,
+        block_hash: near_primitives::hash::CryptoHash,
+    },
 }
 
 #[derive(Debug)]
--- a/chain/client-primitives/src/types.rs
+++ b/chain/client-primitives/src/types.rs
@@ -279,17 +279,37 @@ pub enum QueryError {
     #[error("The node does not track the shard")]
     UnavailableShard { requested_shard_id: near_primitives::types::ShardId },
     #[error("Account ID #{requested_account_id} is invalid")]
-    InvalidAccount { requested_account_id: near_primitives::types::AccountId },
+    InvalidAccount {
+        requested_account_id: near_primitives::types::AccountId,
+        block_height: near_primitives::types::BlockHeight,
+        block_hash: near_primitives::hash::CryptoHash,
+    },
     #[error("Account #{requested_account_id} does not exist while viewing")]
-    UnknownAccount { requested_account_id: near_primitives::types::AccountId },
+    UnknownAccount {
+        requested_account_id: near_primitives::types::AccountId,
+        block_height: near_primitives::types::BlockHeight,
+        block_hash: near_primitives::hash::CryptoHash,
+    },
     #[error(
         "Contract code for contract ID #{contract_account_id} has never been observed on the node"
     )]
-    NoContractCode { contract_account_id: near_primitives::types::AccountId },
+    NoContractCode {
+        contract_account_id: near_primitives::types::AccountId,
+        block_height: near_primitives::types::BlockHeight,
+        block_hash: near_primitives::hash::CryptoHash,
+    },
     #[error("Access key for public key #{public_key} has never been observed on the node")]
-    UnknownAccessKey { public_key: near_crypto::PublicKey },
+    UnknownAccessKey {
+        public_key: near_crypto::PublicKey,
+        block_height: near_primitives::types::BlockHeight,
+        block_hash: near_primitives::hash::CryptoHash,
+    },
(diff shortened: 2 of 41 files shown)


chain/jsonrpc-primitives/src/types/query.rs
```diff
@@ -0,0 +1,151 @@
...
+    #[error("Invalid account ID {0}")]
+    InvalidAccount(near_primitives::types::AccountId),
+    #[error("Account ID {0} has never been observed on the node")]
+    AccountDoesNotExist(near_primitives::types::AccountId),
+    #[error("Contract code for contract ID {0} has never been observed on the node")]
+    ContractCodeDoesNotExist(near_primitives::types::AccountId),
+    #[error("Access key for public key {0} has never been observed on the node")]
+    AccessKeyDoesNotExist(String),
+    #[error("VM error occurred: {0}")]
+    VMError(String),
```
frol, line 29: I suggest we apply the following reasoning:

* Each individual error variant should be self-contained, e.g. `NotSyncedYet`, `DoesNotTrackShard`, and even `VMError` imply a certain context. I suggest we capture the context in the naming:
  * `NoSyncedBlocks` ("There are no fully synchronized blocks on the node yet")
  * `UnavailableShard { requested_shard_id: ShardId }` ("The node does not track the shard #{requested_shard_id} where the requested account resides")
  * `FailedFunctionCall { error_message: String }` - we may consider returning `VMError` structure here since It is what we may want to do in `tx` and `broadcast_tx_*` endpoints anyway, and having a debug-formatted string of the enum is not helpful anyway, so we may just as well just go and dump `VMError`
* I suggest we keep our terminology around "does-not-exist" consistent and always use either "unavailable" or "unknown": `UnknownAccount`, `UnknownAccessKey`
* Contract code can be missing on an account, so it is not part of the "does not exist" crew, but rather just: `NoContractCode { contract_account_id: AccountId }`

chain/jsonrpc-primitives/src/types/query.rs
```diff
@@ -0,0 +1,151 @@
...
+
+#[derive(thiserror::Error, Debug)]
+pub enum RpcQueryError {
+    #[error("IO Error: {0}")]
+    IOError(String),
+    #[error("There are no fully synchronized blocks yet")]
+    NotSyncedYet,
+    #[error("The node does not track the shard")]
+    DoesNotTrackShard,
+    #[error("Invalid account ID {0}")]
```
frol, line 20:
```suggestion
    #[error("Account ID {0} is invalid")]
```

chain/jsonrpc-primitives/src/types/query.rs
```diff
@@ -0,0 +1,151 @@
...
+    #[serde(flatten)]
+    pub query_response: near_primitives::views::QueryResponse,
+}
+
+impl RpcQueryRequest {
+    pub fn parse(value: Option<Value>) -> Result<RpcQueryRequest, crate::errors::RpcParseError> {
+        let query_request = if let Ok((path, data)) =
+            crate::utils::parse_params::<(String, String)>(value.clone())
+        {
+            let data = near_primitives_core::serialize::from_base(&data)
```
frol, line 51: Please, preserve the comment we had in the original `jsonrpc/lib.rs` implementation about soft-deprecation

chain/jsonrpc-primitives/src/types/query.rs
```diff
@@ -0,0 +1,151 @@
...
+            near_client_primitives::types::QueryError::AccessKeyDoesNotExist(public_key) => {
+                Self::AccessKeyDoesNotExist(public_key)
+            }
+            near_client_primitives::types::QueryError::VMError(s) => Self::VMError(s),
+            near_client_primitives::types::QueryError::Unreachable(s) => {
+                near_metrics::inc_counter_vec(
+                    &crate::metrics::RPC_UNREACHABLE_ERROR_COUNT,
+                    &["RpcQueryError", &s],
+                );
+                Self::Unreachable(s)
```
frol, line 133: I recall we had some discussion about adding logging for such unreachable states, so not only we can see the counter, but actually address those based on which case we missed. Do we log it anywhere?

chain/jsonrpc/src/lib.rs
```diff
@@ -557,86 +557,16 @@ impl JsonRpcHandler {
...
-        })?
+    async fn query(
+        &self,
+        request_data: near_jsonrpc_primitives::types::query::RpcQueryRequest,
+    ) -> Result<
+        near_jsonrpc_primitives::types::query::RpcQueryResponse,
+        near_jsonrpc_primitives::types::query::RpcQueryError,
+    > {
+        let query = Query::new(request_data.block_reference, request_data.request);
+        let query_response = self.view_client_addr.send(query.clone()).await??;
```
frol, line 568:
```suggestion
        let query_response = self.view_client_addr.send(query).await??;
```

neard/src/lib.rs
```diff
@@ -28,6 +28,7 @@ use near_store::migrations::{
 
 #[cfg(feature = "protocol_feature_rectify_inflation")]
 use near_store::migrations::migrate_16_to_rectify_inflation;
+pub use runtime::errors as runtime_errors;
```
frol, line 31: Why?

neard/src/runtime/errors.rs
```diff
@@ -0,0 +1,84 @@
+use near_chain::near_chain_primitives::error::QueryError;
+
+#[derive(thiserror::Error, Debug)]
+#[error(transparent)]
+pub struct RuntimeQueryError(#[from] pub QueryError);
```
frol, line 5: Do we need to be able to construct `RuntimeQueryError` from the `QueryError`?

```suggestion
pub struct RuntimeQueryError(pub QueryError);
```

I also suggest we rename it to `WrappedQueryError` to better indicate the intention that we just want to have a wrapped type here, and also add a comment explaining that we cannot implement `From<>` trait for the original `QueryError` struct since it is a foreign type.

runtime/runtime/src/state_viewer/errors.rs
```diff
@@ -0,0 +1,95 @@
...
+#[derive(thiserror::Error, Debug)]
+pub enum ViewContractCodeError {
+    #[error("Account ID {0} is invalid")]
+    InvalidAccountId(near_primitives::types::AccountId),
+    #[error("Account ID {0} does not exist")]
+    AccountDoesNotExist(near_primitives::types::AccountId),
+    #[error("Storage error: {0}")]
+    StorageError(near_primitives::errors::StorageError),
+    #[error("Contract code for contract ID {0} does not exist")]
+    ContractCodeDoesNotExist(near_primitives::types::AccountId),
```
frol, line 20: Let's prefer using named arguments inside the enums like:

```rust
    InvalidAccountId { requested_account_id: near_primitives::types::AccountId },
    NoContractCode { contract_account_id: near_primitives::types::AccountId },
```

runtime/runtime/src/state_viewer/errors.rs
```diff
@@ -0,0 +1,95 @@
...
+    fn from(storage_error: near_primitives::errors::StorageError) -> Self {
+        Self::StorageError(storage_error)
+    }
+}
+
+impl From<near_primitives::errors::StorageError> for CallFunctionError {
+    fn from(storage_error: near_primitives::errors::StorageError) -> Self {
+        Self::StorageError(storage_error)
+    }
+}
```
frol, line 95: This can be replaced with `#[from]` attribute in the relevant enum variant, I believe

runtime/runtime/src/state_viewer/mod.rs
```diff
@@ -55,23 +58,60 @@ impl TrieViewer {
...
+                        .map_err(|_| {
+                            errors::ViewAccessKeyError::InternalError(format!(
+                                "Unexpected invalid public key {:?} received from store",
+                                public_key
+                            ))
+                        })
+                        .map(|key| (key, access_key))
+                })
+                .collect::<Result<Vec<_>, errors::ViewAccessKeyError>>(),
+            Err(e) => Err(e.into()),
```
frol, line 102: Replace this match with `?` (`state_update.iter(&prefix)?.iter().map(...)`)

chain/client-primitives/src/types.rs
```diff
@@ -264,7 +264,42 @@ impl Query {
...
+    #[error("Invalid account ID {0}")]
+    InvalidAccount(near_primitives::types::AccountId),
+    #[error("Account ID {0} has never been observed on the node")]
+    AccountDoesNotExist(near_primitives::types::AccountId),
+    #[error("Contract code for contract ID {0} has never been observed on the node")]
+    ContractCodeDoesNotExist(near_primitives::types::AccountId),
+    #[error("Access key for public key {0} has never been observed on the node")]
+    AccessKeyDoesNotExist(String),
+    #[error("VM error occurred: {0}")]
+    VMError(String),
```
frol, line 287: just a reminder to align these accordingly to the comment I left to the enum in jsonrpc-primitives

frol (approved):
We are on the right track!

frol:
@khorolets Another reminder as we discussed last week: Add tests covering the error responses (at least the legal ones, no need to test IOErrors and Unexpected errors); see `neard/tests/rpc_nodes.rs`

neard/src/runtime/mod.rs
```diff
@@ -1176,125 +1176,105 @@ impl RuntimeAdapter for NightshadeRuntime {
...
+                        &self.epoch_manager,
+                        current_protocol_version,
+                        #[cfg(feature = "protocol_feature_evm")]
+                        self.evm_chain_id(),
+                    )
+                    .map_err(errors::WrappedQueryError::from)?;
+                Ok(QueryResponse {
+                    kind: QueryResponseKind::CallResult(CallResult {
+                        result: call_function_result,
+                        logs,
```
frol, line 1235: I am not quite sure it is the way we want to wrap things. Can we change the `CallResult(CallResult { result, log })` into 

```rust
struct QueryResponseKind {
   // ...
   CallResult(CallResult)
}

#[serde(tag = "execution_status", rename="SCREAMING_SNAKE_CASE")]
enum CallResult {
    Success { result, logs },
    Failure { error, logs },
}
```

I see it that the JSON RPC response for a failing view function call will be:

```json
{
  "jsonrpc": "2.0",
  "result": {
    "execution_status": "FAILURE",
    "error": ...,
    "logs": ...
  }
}
```

It does not make sense to me to have a failing CallResult as part of the QueryError since it is not the query failed, but the contract code.

frol (approved):
Just a few nits, and we are ready to merge it!

khorolets:
Backward compatible messages from jsonrpc needs to be done to finish this PR

chain/client/src/view_client.rs
```diff
@@ -229,50 +219,31 @@ impl ViewClientActor {
...
+                    shard_id,
+                    &state_root,
+                    header.height(),
+                    header.raw_timestamp(),
+                    header.prev_hash(),
+                    header.hash(),
+                    header.epoch_id(),
+                    &msg.request,
+                )
+                .map_err(RuntimeQueryError::from)?)
```
frol, line 244: I wonder if it is the only place where we need this `RuntimeQueryError`. If so, it might be easier to move the `match` that is currently in the `from` implementation right here inline and avoid the `RuntimeQueryError`. This will reduce the indirection, but will bloat this implementation :disappointed: 

WDYT?

chain/client/src/view_client.rs
```diff
@@ -446,14 +417,46 @@ impl Actor for ViewClientActor {
...
 
     #[perf]
     fn handle(&mut self, msg: Query, _: &mut Self::Context) -> Self::Result {
         self.handle_query(msg)
     }
 }
 
+#[derive(thiserror::Error, Debug)]
+#[error(transparent)]
+struct RuntimeQueryError(#[from] pub near_chain::near_chain_primitives::error::QueryError);
```
frol, line 430: 0. Consider removing this wrapper (see my other comment in the `view_client.rs`)
1. If we keep it, let's document it
2. If we keep it, let's rename the struct to be more explicit

```suggestion
/// This is a helper struct which is used to implement `From<near_chain_primitives::error::QueryError>`
/// for `near_client_primitives::types::QueryError` (both are foreign types) to be able to use `?` operator.
#[derive(thiserror::Error, Debug)]
#[error(transparent)]
struct WrappedChainQueryError(#[from] pub near_chain::near_chain_primitives::error::QueryError);
```

chain/chain-primitives/src/error.rs
```diff
@@ -13,6 +13,27 @@ use near_primitives::serialize::to_base;
...
+    AccountDoesNotExist { requested_account_id: near_primitives::types::AccountId },
+    #[error("Contract ID #{contract_account_id} code does not exist while viewing")]
+    ContractCodeDoesNotExist { contract_account_id: near_primitives::types::AccountId },
+    #[error("Access key for public key #{public_key} does not exist while viewing")]
+    AccessKeyDoesNotExist { public_key: near_crypto::PublicKey },
+    #[error("Storage error occurred: #{storage_error:?}")]
+    StorageError {
+        #[from]
+        storage_error: near_primitives::errors::StorageError,
+    },
```
frol, line 30: I feel like it should be merged into the `InternalError` variant. I feel uneasy about merging a perfectly structured variant into a string, but it seems that after all the `StorageError` is implemented as an internal error representation anyway.

chain/jsonrpc-primitives/src/types/receipts.rs
```diff
@@ -52,12 +53,13 @@ impl From<near_client_primitives::types::GetReceiptError> for RpcReceiptError {
             near_client_primitives::types::GetReceiptError::UnknownReceipt(hash) => {
                 Self::UnknownReceipt(hash)
             }
-            near_client_primitives::types::GetReceiptError::Unreachable(s) => {
+            near_client_primitives::types::GetReceiptError::Unreachable(error_message) => {
+                error!(target: "jsonrpc", "Unreachable error occurred: {}", &error_message);
```
frol, line 57: I prefer thinking of the logging levels in the following way:

* error - it is a full-stop, there is a need for manual recovery, we should not progress further
* warning - we want to bring the information to the attention, but the program can keep going
* info - whatever we think a user may want to know during the normal execution
* debug - more details on what is happening
* trace - the level at which we want to know every single meaningful step the program makes

Thus, let's make them `warn` instead of `error`

neard/tests/rpc_nodes.rs
```diff
@@ -728,3 +728,42 @@ fn test_protocol_config_rpc() {
...
+                            ),
+                            request: near_primitives::views::QueryRequest::ViewAccount {
+                                account_id: "test.near".to_string(),
+                            },
+                        })
+                        .await
+                        .unwrap();
+                    assert_eq!(false, true);
+                    System::current().stop();
+                });
```
frol, line 765: todo for @frol: This is blocked on #3901

runtime/runtime/src/state_viewer/errors.rs
```diff
@@ -0,0 +1,81 @@
...
+pub enum ViewContractCodeError {
+    #[error("Account ID #{requested_account_id} is invalid")]
+    InvalidAccountId { requested_account_id: near_primitives::types::AccountId },
+    #[error("Account ID #{requested_account_id} does not exist")]
+    AccountDoesNotExist { requested_account_id: near_primitives::types::AccountId },
+    #[error("Storage error: #{storage_error}")]
+    StorageError {
+        #[from]
+        storage_error: near_primitives::errors::StorageError,
+    },
```
frol, line 24: I would move `StorageError` to be the last or even consider immediately convert it to `InternalError { error_message: String }` (or `InternalError { #[from] storage_error: StorageError }` wherever possible, and only `ViewAccessKeyError` will have it as a `String`). I don't have a strong preference here. Either way, I would put it the last variant on the list, and definitely name it as an `InternalError`, as from the external user point of view, that is the only thing they care about.

frol (changes requested):

khorolets:
@frol I hope I've finalized error structs according to your suggestions. I'm going to deal with backward now.

neard/src/runtime/errors.rs
```diff
@@ -0,0 +1,107 @@
+use easy_ext::ext;
+
+use near_chain::near_chain_primitives::error::QueryError;
+
+#[ext(FromStateViewerErrors)]
```
frol, line 5: Let's just use it by absolute path since it is the only place, and we don't need to import it. It also follows the style we already have

```suggestion
#[easy_ext::ext(FromStateViewerErrors)]
```

frol (changes requested):

chain/jsonrpc/src/lib.rs
```diff
@@ -571,90 +568,56 @@ impl JsonRpcHandler {
...
+    async fn query(
+        &self,
+        request_data: near_jsonrpc_primitives::types::query::RpcQueryRequest,
+    ) -> Result<
+        near_jsonrpc_primitives::types::query::RpcQueryResponse,
+        near_jsonrpc_primitives::types::query::RpcQueryError,
+    > {
+        let query = Query::new(request_data.block_reference, request_data.request);
+        // This match is used here to give backward compatible error message for specific
+        // error variants. Should be refactored once structured errors fully shipped
```
frol, line 580: :cry: 

I think our only option going forward would be to return both `result` and `error` in the same JSON RPC response. (`result` for backward compatibility, and `error` for future sanity)

chain/jsonrpc-primitives/src/types/query.rs
```diff
@@ -0,0 +1,204 @@
...
+    // TODO #3851: Remove this variant once we can exhaustively match all the underlying errors
+    #[error("It is a bug if you receive this error type, please, report this incident: https://github.com/near/nearcore/issues/new/choose. Details: {error_message}")]
+    Unreachable { error_message: String },
+}
+
+#[derive(Serialize, Deserialize)]
+pub struct RpcQueryResponse {
+    #[serde(flatten)]
+    pub query_response: near_primitives::views::QueryResponse,
+}
```
frol, line 69: I feel that the nesting is completely unnecessary here. We should either (1) implement our own `RpcQueryResponse` or (2) just type-alias the `views::QueryResponse`. For clear separation of concerns and following our previous style, we should consider (1) and also removing the serde serialization from `view::QueryResponse`.

frol (changes requested):
It is a massive effort! Given there is only one comment from me, I think it is time to mark it is ready for further review.

chain/client/src/view_client.rs
```diff
@@ -229,50 +219,52 @@ impl ViewClientActor {
...
-                    self.network_adapter.do_send(NetworkRequests::Query {
-                        query_id: msg.query_id,
-                        account_id: validator,
-                        block_reference: msg.block_reference,
-                        request: msg.request,
-                    });
-                }
-
-                Ok(None)
+        if self.runtime_adapter.cares_about_shard(
```
bowenwang1996, line 224: I am sorry but please do not try to change the logic here. What you did is not an equivalent transformation.

chain/jsonrpc/src/lib.rs
```diff
@@ -250,6 +241,9 @@ impl JsonRpcHandler {
             "EXPERIMENTAL_changes_in_block" => self.changes_in_block(request.params).await,
             "EXPERIMENTAL_check_tx" => self.check_tx(request.params).await,
             "EXPERIMENTAL_genesis_config" => self.genesis_config().await,
+            "EXPERIMENTAL_light_client_proof" => {
```
bowenwang1996, line 244: what is this change for

chain/jsonrpc/src/lib.rs
bowenwang1996: why couldn't we get rid of `QueryResponseKind::Error` and emulate it here instead?

chain/client/src/view_client.rs
khorolets: @bowenwang1996 can you assist here to transform it equivalent, please?

chain/jsonrpc/src/lib.rs
khorolets: Alphabetical ordering of methods, we found it to be easier to read and maintain

chain/jsonrpc/src/lib.rs
frol: @bowenwang1996 good point.

@khorolets Can we get rid of the `Error` variant and move this handling to a separate backward-compatibility-wrapper called on the `process_request` level? (`serde_json::to_value(query_response)` can moved into this wrapper)

neard/tests/rpc_nodes.rs
```diff
@@ -728,3 +727,42 @@ fn test_protocol_config_rpc() {
...
     });
 }
+
+#[test]
+fn test_query_rpc() {
+    init_integration_logger();
+    heavy_test(|| {
+        System::builder()
+            .stop_on_panic(true)
+            .run(move || {
```
frol, line 737: hint: see #4013 on how to use `run_actix_until_stop`  (after the PR is merged)

chain/client/src/view_client.rs
frol: @bowenwang1996 we removed query-routing here following your take in https://github.com/near/nearcore/pull/3204/files#diff-34fd63bb882408a6cd040297ab48d277c848f434f9f72da6696aa17e86099405R212-R230

frol (approved):

core/primitives/src/views.rs
```diff
@@ -246,14 +246,13 @@ impl std::iter::FromIterator<AccessKeyInfoView> for AccessKeyList {
...
-#[derive(BorshSerialize, BorshDeserialize, Serialize, Deserialize, Debug, PartialEq, Eq, Clone)]
-#[serde(untagged)]
+#[derive(BorshSerialize, BorshDeserialize, Debug, PartialEq, Eq, Clone)]
 pub enum QueryResponseKind {
     ViewAccount(AccountView),
     ViewCode(ContractCodeView),
     ViewState(ViewStateResult),
     CallResult(CallResult),
-    Error(QueryError),
+    // Error(QueryError),
```
frol, line 255: Clean it up

neard/tests/rpc_nodes.rs
```diff
@@ -712,3 +711,135 @@ fn test_protocol_config_rpc() {
...
+                        },
+                    })
+                    .await;
+                assert!(query_response.is_err());
+                let error_message = match query_response {
+                    Ok(result) => panic!("expected error but received Ok: {:?}", result.kind),
+                    Err(err) => err.data.unwrap(),
+                };
+                assert!(
+                    error_message.to_string().contains("does not exist while viewing"),
```
frol, line 837: Let's do strict matching, ensuring that we don't break the error message even slightly

frol (approved):

bowenwang1996 (approved):

PR #3948, opened 2021-02-11
refactor: simplify access to COSTS_CONTER thread local

matklad:
Practically, this gets rid of `#[cfg]` here and there.

In general, it's a good rule of thumb to protect access to global and
thread local data with accessor functions. That way, the clients don't
need to deal with RefCells and Mutexes themselves.

### DIFF
--- a/runtime/near-vm-logic/src/gas_counter.rs
+++ b/runtime/near-vm-logic/src/gas_counter.rs
@@ -7,15 +7,27 @@ use near_primitives_core::{
     profile::ProfileData,
     types::Gas,
 };
+use std::collections::HashMap;
 use std::fmt;
 
 #[cfg(feature = "costs_counting")]
 thread_local! {
-    pub static EXT_COSTS_COUNTER: std::cell::RefCell<std::collections::HashMap<ExtCosts, u64>> =
-        Default::default();
-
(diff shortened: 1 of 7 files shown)


ailisp (approved):
beautiful!

runtime/near-vm-logic/src/gas_counter.rs
```diff
@@ -7,17 +7,29 @@ use near_primitives_core::{
...
     #[cfg(feature = "protocol_feature_evm")]
     pub static EVM_GAS_COUNTER: std::cell::RefCell<EvmGas> = Default::default();
 }
 
+#[inline]
+pub fn with_ext_cost_counter(f: impl FnOnce(&mut HashMap<ExtCosts, u64>)) {
+    #[cfg(feature = "costs_counting")]
+    {
+        thread_local! {
+            static EXT_COSTS_COUNTER: std::cell::RefCell<HashMap<ExtCosts, u64>> =
```
willemneal, line 24: Wow didn't realize you could have a static variable inside a function.

willemneal (approved):

matklad:
ping @evgenykuzyakov, I think this is blocked on your review :)

olonho (approved):

frol (approved):

Issue #3953, opened 2021-02-12
Add BLAKE2b support to the Math API


artob:
To speed up and lower the cost of the [EVM-as-a-contract](https://github.com/near/near-evm) implementation, we will need to add `blake2b()` as a [Math API](https://nomicon.io/RuntimeSpec/Components/BindingsSpec/MathAPI.html) primitive so that the EVM contract itself will not need to implement this hash function.

References:
- https://ethereum.stackexchange.com/questions/15479/list-of-pre-compiled-contracts/15483#15483
- https://en.wikipedia.org/wiki/BLAKE_(hash_function)

Issue #3921, opened 2021-02-05
Add ECRecover support to the Math API


artob:
As discussed with @ilblackdragon, @djsatok, and @frankbraun on 2021-02-04, to speed up and lower the cost of the [EVM-as-a-contract](https://github.com/near/near-evm) implementation, we will need to add `ecrecover()` as a [Math API](https://nomicon.io/RuntimeSpec/Components/BindingsSpec/MathAPI.html) primitive so that the EVM contract itself will not need to implement this function. More broadly, this function will also be useful to any contracts dealing with Eth1/EVM interoperability.

References:
- https://ethereum.stackexchange.com/questions/15479/list-of-pre-compiled-contracts/15483#15483
- https://docs.soliditylang.org/en/latest/units-and-global-variables.html#mathematical-and-cryptographic-functions
- https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/cryptography/ECDSA.sol
- https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/test/cryptography/ECDSA.test.js

PR #3954, opened 2021-02-12
Extend the Math API with EVM precompiles

artob:
To speed up performance and lower the cost of [EVM-as-a-contract](https://github.com/near/near-evm), this adds several functions to the contract runtime [Math API] in order to support implementing the standard EVM precompiled contracts:

- #3922 `ripemd160()` (RIPEMD-160 hash function)
- #3953 `blake2b()` (BLAKE2b hash function)
- #3921 `ecrecover()` (ECRecover ECDSA signer recovery)

References:

- https://ethereum.stackexchange.com/questions/15479/list-of-pre-compiled-contracts/15483#15483
- https://docs.soliditylang.org/en/latest/units-and-global-variables.html#mathematical-and-cryptographic-functions
- https://en.wikipedia.org/wiki/RIPEMD
- https://en.wikipedia.org/wiki/BLAKE_(hash_function)
- https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/cryptography/ECDSA.sol

[Math API]: https://nomicon.io/RuntimeSpec/Components/BindingsSpec/MathAPI.html


ailisp (commented):
Code looks good. This is considered as a protocol version change, so we need to introduce a new protocol feature & protocol version and import it with guard:
https://github.com/near/nearcore/blob/483af53f7fa3d028371021f33e4149e8ee46cebe/runtime/near-vm-runner/src/imports.rs#L282

artob:
@ailisp Please re-review, this is substantially closer to graduating from draft status.

The top open question I still have is how, concretely, I should go about measuring and assigning the gas costs for the new functions. (Also, the runtime-params-estimator exits with a failure, so I may still be missing something in that regard.)

> Code looks good. This is considered as a protocol version change, so we need to introduce a new protocol feature & protocol version and import it with guard:

Thanks. I've guarded everything new behind the EVM feature for now.

core/primitives-core/src/config.rs
```diff
@@ -215,6 +215,19 @@ pub struct ExtCostsConfig {
     /// Cost of getting sha256 per byte
     pub keccak512_byte: Gas,
 
+    /// Cost of getting ripemd160 base
+    pub ripemd160_base: Gas,
```
ilblackdragon, line 219: These should be guarded by feature as well. Otherwise backward compat tests fail - https://buildkite.com/nearprotocol/nearcore/builds/5739#ee3b3336-f269-4c34-ac2b-fb736925da7e

ilblackdragon:
I also think this should be a separate feature from EVM, so that it can be released separately from the full precompiled EVM.

This can be (and probably should be) batched with bn128 host functions.
And as I mentioned a really useful would be to have host function for storage cost.

runtime/runtime-params-estimator/test-contract/src/lib.rs
```diff
@@ -439,6 +442,66 @@ pub unsafe fn keccak512_10kib_10k() {
...
+
+// Function to measure `ecrecover_base`. Also measures `base`, `write_register_base`, and
+// `write_register_byte`. However `ecrecover` computation is more expensive than register writing
+// so we are okay overcharging it.
+// Compute ecrecover 10k times.
+#[no_mangle]
+pub unsafe fn ecrecover_10k() {
+    let hash = [0u8; 32];
+    let signature = [0u8; 65];
+    let (r, s, v) = (&signature[0..32], &signature[32..64], 27);
```
ailisp, line 499: Should this test measure recovering a valid signature

ailisp:
Looks like runtime param estimator tests failed because it isn't compiled with protocol-feature-evm enabled, so those new functions isn't exported:

https://buildkite.com/nearprotocol/nearcore/builds/5739#17b95e82-bc41-4b95-b620-44d8c5b42c89

I suggest:
1. change ci config to make it compile and pass. On CI it is using `Time metric`, which is fast to run, but isn't the one we use as actual cost.
2. Actual cost we use "count of CPU instructions on qemu", this takes 6+ hours to run, boot a cloud instance (at least 4 cpu 16g ram) and update  runtime-params-estimator/estimate.sh to compile with protocol-feature-evm. Then run estimate.sh. Steps are here:

Use this to build a docker image:
https://github.com/near/nearcore/blob/master/runtime/runtime-params-estimator/emu-cost/README.md#usage

once in docker, disregard above steps, simply run `estimate.sh`
https://github.com/near/nearcore/blob/master/runtime/runtime-params-estimator/estimate.sh

runtime/runtime-params-estimator/test-contract/src/lib.rs
artob: Initially I had wanted to do that, but this test contract apparently shouldn't have any dependencies (e.g., on the `hex-literals` crate) so I ripped that code out. I plan to reformulate later using more cumbersome code to initialize the hash and signature word by word.

abacabadabacaba:
There is a problem with the BLAKE2b function added here. While this PR adds the whole BLAKE2b function, what EVM should provide to smart contracts is the compression function F (see [EIP 152](https://eips.ethereum.org/EIPS/eip-152)), and you can't use the whole BLAKE2b function to implement just the compression function.

While we can add just the function F as a host function, it would be preferred that the function is usable not just for the EVM but also for other contracts that may want to compute BLAKE2 hashes. So, I think that a better idea is to add a function that evaluates a sequence of calls to the compression function in a way similar to how the BLAKE2 function itself does it. Such function could be used to perform a single compression function call (as needed by the EVM), and it can also compute the BLAKE2 hash itself in a single call.

Another thing that I would have changed is to charge gas proportional to the number of blocks, not bytes.

artob:
@joshuajbouw I will not have time to come back to this for a bit, so please take over this task and push it to completion. The subtasks I currently know about include the following:

- Investigate if/how we can batch this with #3971, already merged (and also a good example of adding new host functions)
- Add any and all missing feature guards, as per @ilblackdragon's [comment](https://github.com/near/nearcore/pull/3954#pullrequestreview-594203564)
- Change the feature guards in this PR to be based on a new distinct feature name, as per @ilblackdragon's [comment](https://github.com/near/nearcore/pull/3954#issuecomment-782151255)
- Flesh out the `ecrecover_10k()` test case for the runtime-params-estimator, as per @ailisp's [comment](https://github.com/near/nearcore/pull/3954#pullrequestreview-594402622)
- Add support for [EIP-152](https://eips.ethereum.org/EIPS/eip-152) (the BLAKE2 F compression function) as per @abacabadabacaba's [comment](https://github.com/near/nearcore/pull/3954#issuecomment-784399636); I would retain the already implemented BLAKE2b function, so this would be a new function
- Measure and assign the gas costs for the new functions using the runtime-params-estimator, as per @ailisp's [comment](https://github.com/near/nearcore/pull/3954#issuecomment-782226357); for any support on that, ping @ailisp
- Move the PR out of Draft status and re-request reviews once ready to go

joshuajbouw:
Ok, yes will go through all that today and this weekend.

joshuajbouw:
Got a question, doing all the feature gating. I figure, that in the wrapped_imports that ripemd160, blake2b and ecrecover could be used for more than just EVM?

In: https://github.com/near/nearcore/blob/evm-math-api/runtime/near-vm-runner/src/imports.rs#L173

artob:
> Got a question, doing all the feature gating. I figure, that in the wrapped_imports that ripemd160, blake2b and ecrecover could be used for more than just EVM?

Right, as we discussed on the call just now, all additions here should be behind the new feature guard. Wherever you see `protocol_feature_evm` from the original commits here, change to the new feature guard you're defining.

joshuajbouw:
Do we have ETH ABI packed encoding to make it a reality or were you suggesting something much like the original implementation of F? @abacabadabacaba

joshuajbouw:
Another question, @artob had added this to to the genesis_config.js https://github.com/near/nearcore/blob/2dcb7383b4c0e2d29faa7b08bc10570673c7522d/neard/res/genesis_config.json#L157. This gets a bit tricky as we can't exactly conditionally use this file. Is there some sort of upgrade we use to add more to the `wasm_config`?

artob:
> Another question, @artob had added this to to the genesis_config.js
> 
> https://github.com/near/nearcore/blob/2dcb7383b4c0e2d29faa7b08bc10570673c7522d/neard/res/genesis_config.json#L157-L161
> 
> This gets a bit tricky as we can't exactly conditionally use this file. Is there some sort of upgrade we use to add more to the `wasm_config::ext_costs` section? And I am assuming that it must also be added to `mainnet_genesis.json` if so? I see a lot of examples of changing values, but not adding.

A question for @evgenykuzyakov, given also the additions in #3971. (Didn't see that changing `genesis_config.js`.)

abacabadabacaba:
@joshuajbouw Some clarification regarding the BLAKE2 function.

You are right that the compression F-function is implemented in the BLAKE2 library. However, this implementation uses a fixed number of rounds (12 for BLAKE2b), while EIP-152 permits any number of rounds (up to 2<sup>32</sup>-1). So, we cannot use the library, we need to copy the code and change it to work with arbitrary number of rounds.

On the other hand, the Ethereum precompile only evaluates the compression function once, but it may be more useful to evaluate it multiple times (this is needed when computing a BLAKE2 hash of a message which is longer than one block). For this reason, I propose to have a host function that can evaluate the compression function multiple times (in the same way as when computing a BLAKE2 hash) but also permit setting an arbitrary number of rounds (like in Ethereum). Such function will be useful both for BLAKE2 hashing and for the EVM.

joshuajbouw:
Makes sense. I'll add it to the cryptography library we have now and make the changes. Hopefully, I can get those changes done to the parent blake2 lib approved and we can use it directly again.

artob:
> For this reason, I propose to have a host function that can evaluate the compression function multiple times (in the same way as when computing a BLAKE2 hash) but also permit setting an arbitrary number of rounds (like in Ethereum). Such function will be useful both for BLAKE2 hashing and for the EVM.

Yes, this seems the best approach. Thanks for weighing in, @abacabadabacaba.

ailisp (approved):

joshuajbouw:
ITS ALIVE

joshuajbouw:
Note that I still am getting results for gas, hold on.

joshuajbouw:
Since the other comment went to hidden I'll post it again, I still need to update the gas costs which is happening right now. Don't merge until that is done and of course, post review.

joshuajbouw:
The `estimate.sh` script broke and I think this is culprit. It tries to get `CARGO_MANIFEST_DIR` but thats an auto ENV set by `cargo` but, I don't see `cargo` used in the very last part of the script. @ailisp https://github.com/near/nearcore/pull/4125

runtime/near-vm-logic/src/logic.rs
```diff
@@ -948,6 +948,146 @@ impl<'a> VMLogic<'a> {
...
+    ///
+    /// Takes as an argument the state vector "h", message block vector "m" (the
+    /// last block is padded with zeros to full block size, if required), 2w-bit
+    /// offset counter "t", and final block indicator flag "f". Local vector
+    /// v[0..15] is used in processing. F returns a new state vector. The number
+    /// of rounds, "r", is 12 for BLAKE2b and 10 for BLAKE2s. Rounds are
+    /// numbered from 0 to r - 1.
+    ///
+    /// # Cost
+    /// base + write_register_base + write_register_byte * 64 + blake2b_f_base
```
matklad, line 1007: The cost comment doesn't mention that the cost depends on the number of rounds, it probably should?

It's also interesting that, if we pass in zero rounds, we don't pay blake2b_f specific feee

runtime/near-vm-logic/src/logic.rs
```diff
@@ -948,6 +948,146 @@ impl<'a> VMLogic<'a> {
...
+        rounds_ptr: u64,
+        h_ptr: u64,
+        m_ptr: u64,
+        t_ptr: u64,
+        f_ptr: u64,
+        register_id: u64,
+    ) -> Result<()> {
+        use std::convert::TryFrom;
+
+        let rounds = self.memory_get_u32(rounds_ptr)?;
```
matklad, line 1020: Is there a reason we pass `rounds` and `f` by pointer here, while we pass `v`  by value in `ecrecover`. My understanding of this specific part of code is very vague, but, intuitively, if we can pass by value, we should?

bowenwang1996 (approved):

core/crypto/Cargo.toml
```diff
@@ -6,7 +6,7 @@ edition = "2018"
 
 [dependencies]
 arrayref = "0.3"
-blake2 = "0.9.1"
+blake2 = { git = "https://github.com/near/near-blake2.git", branch = "f-standard-out", version = "0.9.1" }
```
willemneal, line 9: Eventually need to republish this to `near-blake2` and then change to `{ package = "near-blake2", version = "0.9.1" }`, as this crate needs to be published for the simulator.

willemneal (approved):
Just one comment which could be addressed in a future PR.

runtime/near-vm-logic/src/logic.rs
artob: `v` is just single byte (`u8`).

core/primitives/src/version.rs
```diff
@@ -103,7 +103,7 @@ pub enum ProtocolFeature {
 
 /// Current latest stable version of the protocol.
 #[cfg(not(feature = "nightly_protocol"))]
-pub const PROTOCOL_VERSION: ProtocolVersion = 43;
+pub const PROTOCOL_VERSION: ProtocolVersion = 44;
```
frol, line 106: Is there a need to bump the nightly version?

runtime/near-vm-errors/src/lib.rs
```diff
@@ -186,6 +186,8 @@ pub enum HostError {
     ContractSizeExceeded { size: u64, limit: u64 },
     /// The host function was deprecated.
     Deprecated { method_name: String },
+    /// Invalid ECDSA signature.
+    InvalidECDSASignature,
```
frol, line 190: Should it also be feature-guarded?

frol (approved):

runtime/near-vm-logic/src/logic.rs
matklad: `f` is also a single byte (`bool`), and `rounds` is a single `u32`. I still don't see a reason to not just pass them directly, instead of passing pointers to them (which doesn't mean that there isn't such a reason).

artob: Sure, makes sense to me. @joshuajbouw?

abacabadabacaba (changes requested):
Here are the issues that I already mentioned during the EVM team meeting on Monday:

**BLAKE2**

For BLAKE2, we want to provide a host function that can perform a single compression function evaluation, partial hashing, or full hashing. Currently, there are two functions: one that evaluates the compression function once, and one that computes a full hash with a fixed parameters. This is not what we want. Also, the compression function takes some parameters from memory, that could be passed directly.

I think that the function we need for BLAKE2b should have a signature similar to this:

```rust
fn blake2b(rounds: u32, state_len: u64, state_ptr: u64, value_len: u64, value_ptr: u64, t0: u64, t1: u64, f0: u64, f1: u64, register_id: u64): Result<()>
```

It should do the following:
* Panic if the length of the state is not 64 bytes.
* Pad the value with zero bytes so that the length is a multiple of 128 bytes. If the value has zero length, pad it with 128 zero bytes.
* For each 128-byte block of the padded value, call the compression function with the following arguments:
  * Number of rounds: same as passed to the function.
  * State: input state for the first call, the result of the previous call for subsequent calls.
  * Counter: increment `(t0, t1)` by the number of bytes in the current block (counting only bytes from the value but not the padding bytes) and use the resulting values.
  * Finalization flags: pass zero values for all blocks other than the last one. Use `f0` and `f1` for the last block.
* Save the resulting state in `register_id`.

There should be three gas costs: base cost, per-block cost and per-round cost.

There are other issues with the current BLAKE2 implementation:
* We may want to keep it in nearcore repository (either as a separate crate or as a part of near-crypto) for ease of maintenance.
* The current implementation has a bug which will make it panic if the number of rounds is large.
* We don't have is sufficiently covered by tests.
* We want a similar host function for BLAKE2s.

**RIPEMD160**

For this function, we want to charge gas proportional to the number of blocks rather than bytes.

**ecrecover**

This function should return an error instead of aborting the execution if the input is not valid.

joshuajbouw:
Thanks @abacabadabacaba for the valuable feedback. I'll be getting all that in as soon as possible, I got 2 issues that I need to solve with the Aurora engine and I'll add that in.

joshuajbouw:
Just an update as I'm working on this again, going to be doing the following changes:
- Everything that @abacabadabacaba has suggested, including moving the blake2 lib into the nearcore crypto folder
- Resolving the questions that @frol posed
- The first point resolves @willemneal's remark
- Update descriptions with correct costs as per @matklad

Will try my earnest to get this done as soon as possible, we are quite short on time on the EVM side so, I can not guarantee it will be done by the beginning of May, but I'll put in a best effort.

runtime/near-vm-errors/src/lib.rs
bowenwang1996: Yeah looks like it should be @artob

matklad: Wait, should we have this at all? As per @abacabadabacaba  comment, `ecrecover` should return an error, rather than abort execution. My understanding is that, if we emit an `HostError`, we unrecoverably abort transaction?

I might be misunderstading something, but I think what we need here is something like

```diff
diff --git a/runtime/near-vm-logic/src/logic.rs b/runtime/near-vm-logic/src/logic.rs
index 9bb74980..aa5cc40e 100644
--- a/runtime/near-vm-logic/src/logic.rs
+++ b/runtime/near-vm-logic/src/logic.rs
@@ -1097,10 +1097,10 @@ impl<'a> VMLogic<'a> {
         r_ptr: u64,
         s_ptr: u64,
         register_id: u64,
-    ) -> Result<()> {
+    ) -> Result<bool> {
         self.gas_counter.pay_base(ecrecover_base)?;
         if v != 27 && v != 28 {
-            return Err(HostError::InvalidECDSASignature.into());
+            return Ok(false);
         }
         let hash = self.memory_get_vec(hash_ptr, 32)?;
         let v = (v & 0xFF) as u8;
@@ -1115,12 +1115,12 @@ impl<'a> VMLogic<'a> {
         let signature = secp256k1::Signature::parse(&signature);
 
         let recovery_id = match secp256k1::RecoveryId::parse_rpc(v) {
-            Err(_) => return Err(HostError::InvalidECDSASignature.into()),
+            Err(_) => return Ok(false),
             Ok(rid) => rid,
         };
 
         let public_key = match secp256k1::recover(&hash, &signature, &recovery_id) {
-            Err(_) => return Err(HostError::InvalidECDSASignature.into()),
+            Err(_) => return Ok(false),
             Ok(pk) => pk,
         };
 
@@ -1129,7 +1129,8 @@ impl<'a> VMLogic<'a> {
         let mut address = [0u8; 20];
         address.copy_from_slice(&result[12..]);
 
-        self.internal_write_register(register_id, address.to_vec())
+        self.internal_write_register(register_id, address.to_vec());
+        Ok(true)
     }
 
     /// Called by gas metering injected into Wasm. Counts both towards `burnt_gas` and `used_gas`.
```

?

bowenwang1996 (commented):
@abacabadabacaba mentioned that we may introduce some tests where we implement these crypto functions in a different language and fuzz test the rust implementation and assert that the outputs from different implementations are the same.

joshuajbouw:
A note on the estimated values. These were taking from ETH values that they had estimated for their gas. It should be close, and the blake2 ones probably are grossly overestimated but, I think we are ok with overcharging for now.

runtime/near-vm-logic/src/logic.rs
```diff
@@ -943,14 +943,190 @@ impl<'a> VMLogic<'a> {
...
+    /// Hashes the given value using BLAKE2b and returns it into `register_id`.
+    ///
+    /// # Errors
+    ///
+    /// If `value_len + value_ptr` points outside the memory or the registers use more memory than
+    /// the limit with `MemoryAccessViolation`.
+    ///
+    /// # Cost
+    ///
+    /// `base + write_register_base + write_register_byte * num_bytes + blake2b_base + blake2b_byte * num_bytes`
```
willemneal, line 985: Should this be updated since it is 
```
self.gas_counter.pay_per(blake2b_block, message_blocks)?; 
self.gas_counter.pay_per(blake2b_round, (rounds as u64) * message_blocks)?
```

willemneal (approved):
LGTM! Just my one comment.

core/crypto/blake2/src/consts.rs
```diff
@@ -0,0 +1,50 @@
...
+    0x5be0cd19137e2179,
+];
+
+/*
+pub const BLAKE2B_BLOCKBYTES : usize = 128;
+pub const BLAKE2B_OUTBYTES : usize = 64;
+pub const BLAKE2B_KEYBYTES : usize = 64;
+pub const BLAKE2B_SALTBYTES : usize = 16;
+pub const BLAKE2B_PERSONALBYTES : usize = 16;
+*/
```
bowenwang1996, line 37: What is the purpose of these?

bowenwang1996 (commented):
Why do we include the entire implementation of blake2? Is there no existing library for it?

bowenwang1996 (approved):

joshuajbouw:
So, just got to a few last things before this is 100% good to go, but none of the logic that would need to be scrutinised will be affected.

* blake2 lib needs to be removed and put back into its own lib at https://github.com/near/near-blake2. This will prevent the sanity checker from panicking at the SIMD stuff in the library, and we can retain it.
* I need to double-check the documentation
* I need clarity if we are setting a limit on the amount of rounds of blake2.

joshuajbouw:
Ok this PR is good to go.

ilblackdragon:
Let's also not forget to document these changes in Nomicon (https://github.com/near/neps)

runtime/near-vm-errors/src/lib.rs
nearmax: Sounds like @matklad  is right.

Btw, it is extremely important to categorize error messages correctly with the runtime. Almost any incorrect miscategorization will either lead to a network split/slashing or Mainnet shutdown.

nearmax (changes requested):

joshuajbouw:
@matklad PR got a bit long and missed your comment that was pointed out by Max. I had changed that now.

runtime/near-vm-logic/src/logic.rs
```diff
@@ -943,14 +943,200 @@ impl<'a> VMLogic<'a> {
...
+        f0: u64,
+        f1: u64,
+        register_id: u64,
+    ) -> Result<()> {
+        use blake2::VarBlake2b;
+        use std::convert::TryFrom;
+
+        if state_len != 8 {
+            return Err(HostError::Blake2InvalidStateLength { length: state_len }.into());
+        }
```
matklad, line 1016: We probably should do this after paying the base cost? Or, rather, why do we even pass `state_len` if it *has* to be `8`?

I think that's the right pattern to use here:

https://github.com/near/nearcore/blob/f797a3b5585f1bec47bbe0348efd7e3728df2382/runtime/near-vm-logic/src/logic.rs#L1424

runtime/near-vm-logic/src/logic.rs
```diff
@@ -943,14 +943,200 @@ impl<'a> VMLogic<'a> {
...
+    ///
+    ///  Where `message_blocks` is `max(message_len / 8, 1)`.
+    ///
+    /// `base + write_register_base + write_register_byte * num_bytes + ripemd160_base + ripemd160_block * message_blocks`
+    #[cfg(feature = "protocol_feature_evm")]
+    pub fn ripemd160(&mut self, value_len: u64, value_ptr: u64, register_id: u64) -> Result<()> {
+        self.gas_counter.pay_base(ripemd160_base)?;
+        let value = self.get_vec_from_memory_or_register(value_ptr, value_len)?;
+
+        let message_blocks = std::cmp::max(value_len / 8, 1);
```
matklad, line 971: Ignoring leftover after `/ 8` feels odd here. Naively, I'd expect that we want [ceiling rather than floor](https://en.wikipedia.org/wiki/Floor_and_ceiling_functions) here. Might make sense to leave a comment explainign why we take floor here?

runtime/near-vm-logic/src/logic.rs
```diff
@@ -943,14 +943,200 @@ impl<'a> VMLogic<'a> {
...
+    ) -> Result<()> {
+        use blake2::VarBlake2b;
+        use std::convert::TryFrom;
+
+        if state_len != 8 {
+            return Err(HostError::Blake2InvalidStateLength { length: state_len }.into());
+        }
+
+        // Change to per block for gas.
+        let message_blocks = std::cmp::max(message_len / 128, 1);
```
matklad, line 1019: Same here, why we round down?

runtime/near-vm-logic/src/logic.rs
```diff
@@ -943,14 +943,200 @@ impl<'a> VMLogic<'a> {
...
+
+        self.gas_counter.pay_base(blake2b_base)?;
+        self.gas_counter.pay_per(blake2b_block, message_blocks)?;
+        self.gas_counter.pay_per(blake2b_round, (rounds as u64) * message_blocks)?;
```
matklad, line 1023: This multiplication can overflow

runtime/near-vm-logic/src/logic.rs
```diff
@@ -943,14 +943,200 @@ impl<'a> VMLogic<'a> {
...
+
+        let state = <[u64; 8]>::try_from(self.memory_get_vec_u64(state_ptr, 8)?)
```
matklad, line 1025: Use `memory_get_into` here to avoid allocation of intermediate vec.

runtime/near-vm-logic/src/logic.rs
```diff
@@ -943,14 +943,200 @@ impl<'a> VMLogic<'a> {
...
+
+        self.gas_counter.pay_base(blake2b_base)?;
+        self.gas_counter.pay_per(blake2b_block, message_blocks)?;
+        self.gas_counter.pay_per(blake2b_round, (rounds as u64) * message_blocks)?;
+
+        let state = {
+            let mut buf = [0u32; 8];
+            let values = self.memory_get_vec_u64(state_ptr, 8)?;
+            for (x, y) in buf.iter_mut().zip(values.into_iter()) {
+                *x = <u32>::try_from(y).map_err(|_| HostError::Blake2HashDataOverflow)?;
```
matklad, line 1084: Why `HostError::Blake2HashDataOverflow` needs to exist? Seems like we can ask the caller to pass  a bunch of `u32`, rather than asking them to pass a bunch of `u64` and then checking if those `u64`s are actually `u32`s?

matklad (changes requested):

joshuajbouw:
Had to import two additional crates `num-integer` and `byte-slice-cast`, they are listed in our Cargo.lock as our dependencies do use it as well. Would be great if we had our own library or something similar to add in this missing functionality / unstable API from std.

joshuajbouw:
Closing in favour of a new PR with blake2 descoped as we have to ensure that it is completely correct. Ecrecover is the most important method that we need in the PR for Aurora.

PR #3958, opened 2021-02-15
fix: paper over #3925

matklad:


bowenwang1996 (approved):

frol (approved):
@bowenwang1996 Is it the only test that failed? I recall removing the `join`s from a few more places :thinking:

bowenwang1996:
I am not sure either. Let's see if it comes up again

PR #3960, opened 2021-02-15
fix(rpc): address slowness in validator rpc

### LINKED ISSUE #3614: Address slowness in validator rpc endpoint
Currently the rpc endpoint `validators` can be quite slow if the request does not query the latest block. Furthermore, such request can also slow down the client because it takes the lock on epoch manager. To address this issue, we can do the following:
1) implement a new endpoint `EXPERIMENTAL_validators` that only allows querying a specific epoch or the latest data and internally we cache validator data per epoch.
2) For the current endpoint, we return an error if it does not query the latest block.

Note: there is no fundamental reason why we cannot resolve the lock issue, but that requires a bigger change and still will not address the issue of an arbitrary query being slow.

bowenwang1996:
Fix #3614 by disallowing querying the validator rpc endpoint through block hash. Instead, valid queries are either for an epoch or for latest data. Inside epoch manager, validator info for past epochs are stored when we finalize an epoch to ensure that queries will not cause recomputation, which can potentially block the client actor.

Test plan
---------
* `test_get_validator_info`
* `test_get_validator_info_rpc`


chain/jsonrpc/src/lib.rs
```diff
@@ -279,7 +279,15 @@ impl JsonRpcHandler {
             "query" => self.query(request.params).await,
             "status" => self.status().await,
             "tx" => self.tx_status_common(request.params, false).await,
-            "validators" => self.validators(request.params).await,
+            "validators" => {
```
bowenwang1996, line 282: @frol should I introduce an alias here or is it fine as is?

chain/epoch_manager/src/lib.rs
```diff
@@ -929,34 +915,83 @@ impl EpochManager {
...
+                    .enumerate()
+                    .map(|(validator_id, info)| {
+                        let validator_stats = epoch_summary
+                            .validator_block_chunk_stats
+                            .get(&info.account_id)
+                            .map(|stats| stats.block_stats.clone())
+                            .unwrap_or_else(|| ValidatorStats { produced: 0, expected: 0 });
+                        let mut shards = validator_to_shard[validator_id]
+                            .clone()
+                            .into_iter()
```
matklad, line 937:
```suggestion
                            .iter()
                            .cloned()
```

this'll avoid unnecessary clone of the hash map itself

chain/epoch_manager/src/lib.rs
```diff
@@ -929,34 +915,83 @@ impl EpochManager {
...
-                    num_expected_blocks: validator_stats.expected,
-                })
-            })
-            .collect::<Result<Vec<CurrentEpochValidatorInfo>, EpochError>>()?;
+
+        // This ugly code arises because of the incompatible types between `block_tracker` in `EpochInfoAggregator`
+        // and `validator_block_chunk_stats` in `EpochSummary`. Rust currently has no support for Either type
+        // in std.
+        let (current_validators, next_epoch_id, all_proposals) = match epoch_identifier {
+            EpochIdentifier::EpochId(ref id) => {
```
matklad, line 923: a modern Rust style avoids `ref`s and leans on match ergonomics

```suggestion
        let (current_validators, next_epoch_id, all_proposals) = match &epoch_identifier {
            EpochIdentifier::EpochId(id) => {
```

this way, you specifiy ownersip mode once, in the match head, and not for every match arm.

chain/epoch_manager/src/lib.rs
```diff
@@ -929,34 +915,83 @@ impl EpochManager {
...
-                    shards,
-                    num_produced_blocks: validator_stats.produced,
-                    num_expected_blocks: validator_stats.expected,
-                })
-            })
-            .collect::<Result<Vec<CurrentEpochValidatorInfo>, EpochError>>()?;
+
+        // This ugly code arises because of the incompatible types between `block_tracker` in `EpochInfoAggregator`
+        // and `validator_block_chunk_stats` in `EpochSummary`. Rust currently has no support for Either type
+        // in std.
```
matklad, line 921: we have `either` crate in crate graph. But also, urgh, I dont' know how to make this meaningfully better :man_shrugging:

bowenwang1996: Thought about it but decide not to introduce unnecessary dependencies :)

chain/jsonrpc-primitives/src/types/validator.rs
```diff
@@ -0,0 +1,89 @@
...
+    pub fn parse(
+        value: Option<Value>,
+    ) -> Result<RpcValidatorRequest, crate::errors::RpcParseError> {
+        let epoch_reference = if let Ok((block_id,)) =
+            crate::utils::parse_params::<(near_primitives::types::MaybeBlockId,)>(value.clone())
+        {
+            if block_id.is_some() {
+                return Err(crate::errors::RpcParseError(
+                    "Require at least one parameter".to_owned(),
+                ));
```
frol, line 68: Can we still let people specify block hash/block height and resolve it to the epoch id? I mean, we won't provide the information as of the arbitrary block hash/height, but rather round it to the epoch boundary instead.

chain/jsonrpc/src/lib.rs
frol: I think it is fine to modify the original method as long as we keep certain backward compatibility.

chain/chain/src/types.rs
```diff
@@ -644,6 +644,13 @@ pub struct LatestKnown {
     pub seen: u64,
 }
 
+/// Either an epoch id or latest block hash
+#[derive(Debug)]
+pub enum EpochIdentifier {
```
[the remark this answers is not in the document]
bowenwang1996: The problem is that while we can resolve specific epoch id, epoch manager does not have the concept of chain and therefore does not know about the latest block. I don't want to refactor epoch manager to add that just for this purpose.

frol (approved):
I approve to unblock this PR. We discussed it offline, and it seems there is a clear path to implement it in a backward-compatible way (`[null]`, `["block-hash"]`, `[block-height-u64]` should keep working fine, though the latter two will resolve to the `next_epoch_id` of the given block, representing the `validators` stats for the whole epoch, but that is fine)

SkidanovAlex (approved):

PR #3965, opened 2021-02-16
refactor: remove roundabout way to connect to socket

matklad:
Seems like this Resolver stuff is just ... unnecessary? It sounds like
it does DNS resolution (with its notoriously sync-only APO), but this
isn't needed in this case. addr is a `SocketAddr`, ie, a resolved IP.

Note the 1 second timeout: it's the same value that was used by actix:
https://github.com/actix/actix/blob/v0.11.0-beta.1/src/actors/resolver.rs#L382

This is a second attempt at https://github.com/near/nearcore/pull/3961, which was reverted in https://github.com/near/nearcore/pull/3964. I've looked at the code more closely, and noticed that atix hard-codded a one-second timeout here.

This is def a problem in theory -- if the peer just drops packets (instead of actively refusing the connection), the whole peer manager get's stuck in the `.wait(ctx);` call for several minutes. Not sure if this is exactly a problem we've observed thoug!

Note that, even with a timeout, it feels like this code is vulnerable to DOS, as it's pretty easy to freeze the peer manager for one second this way.

### DIFF
--- a/chain/network/src/peer_manager.rs
+++ b/chain/network/src/peer_manager.rs
@@ -10,7 +10,7 @@ use std::time::{Duration, Instant};
 use actix::io::FramedWrite;
 use actix::{
     Actor, ActorFuture, Addr, Arbiter, AsyncContext, Context, ContextFutureSpawner, Handler,
-    Recipient, Running, StreamHandler, SyncArbiter, SyncContext, SystemService, WrapFuture,
+    Recipient, Running, StreamHandler, SyncArbiter, SyncContext, WrapFuture,
 };
 use chrono::Utc;
 use futures::task::Poll;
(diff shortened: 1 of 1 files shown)


matklad:
cc @chefsale, still figuring out how check if this fixes the problem

matklad:
Ok, I've verified manually by running `nearup run testnet  --binary-path ./target/release` that

* old version of the patch fails to connect to peers
* new version works
* the error indeed is due to timeout (that is, while we connect to testnet, some connections fail with a timeout)

(btw, `nearup` tool is really nice!)

Couple of questions:

* should we surface this timeout via some config?
* is it OK that connecting to testnet actually triggers timeout? (probably yes -- most likely some peer's port is blocked by some firewall and drops `SYN`, instead of `RST`ing)
* Do we want to test it? I am not sure what's the best way to do this -- we need some kind of integration test which has a peer that doesn't RST the TCP connection. I am out of my OSI depth here )
* Do we want to track an issuer that `triggers timeout` at the moment means that PeerManager does nothing until timeout expires?

frol (approved):

mfornet (approved):
> Note that, even with a timeout, it feels like this code is vulnerable to DOS, as it's pretty easy to freeze the peer manager for one second this way.

Is it really freezing Peer Manager if it is a Future (Timeout)? because I believe while the future is not resolved the PeerManager can still make progress.

In addition:

This is the case for outbound connections which are bounded locally and only triggered by the node itself, not from external nodes trying to connect. So it is not trivial to DOS a node this way since node select peer that it wants to connect.

The worst scenario is that this peer (A) always tries to connect to adversary B, and B doesn't accept/reject connection so full timeout is going to pass.

matklad:
>Is it really freezing Peer Manager if it is a future (Timeout)? because I believe while the future is not resolved the PeerManager can still make progress.

I believe so. At least, `PeerManager` won't process any additional messages until the future is resolved. The `.wait` call returns immediately, but (according to the docs and observed behavior, although I didn't verify this tightly), it prevens the actor from accepting new messages. 

> This is the case for outbound connections which are bounded locally and only triggered by the node itself, not from external nodes trying to connect. 

As, I missed that bit, thanks for pointing out. Yeah, in this case, this is much less worrisome. Although we do observe timeouts routinelly on the testnet here.

chain/network/src/peer_manager.rs
```diff
@@ -1647,10 +1647,7 @@ impl Handler<OutboundTcpConnect> for PeerManagerActor {
         let _d = DelayDetector::new("outbound tcp connect".into());
         debug!(target: "network", "Trying to connect to {}", msg.peer_info);
         if let Some(addr) = msg.peer_info.addr {
-            #[allow(deprecated)]
-            // There is not clear migration path at the moment: https://github.com/actix/actix/issues/451
-            actix::actors::resolver::Resolver::from_registry()
-                .send(actix::actors::resolver::ConnectAddr(addr))
+            tokio::time::timeout(Duration::from_secs(1), TcpStream::connect(addr))
```
bowenwang1996, line 1650: can we add a comment on why we put `Duration::from_secs(1)` here?

Issue #3938, opened 2021-02-10
State query is too slow


bowenwang1996:
As reported by @vgrichina, currently on mainnet the following query always times out:

```
http post https://rpc.mainnet.near.org jsonrpc=2.0 id=dontcare method=query \
  params:='{
    "request_type": "view_state",
    "finality": "final",
    "account_id": "contract.paras.near",
    "prefix_base64": ""
  }'
HTTP/1.1 200 OK
Alt-Svc: clear
Via: 1.1 google
content-length: 111
content-type: application/json
date: Wed, 10 Feb 2021 00:01:49 GMT

{
    "error": {
        "code": -32000,
        "data": "query has timed out",
        "message": "Server error"
    },
    "id": "dontcare",
    "jsonrpc": "2.0"
}
```

I tried it with a node that is running current master node and it still times out. It look like this is likely caused by the slowness in trie iteration, but requires further investigation.

bowenwang1996:
@mikhailOK I did some further investigation and found that this query resulted in 9292 trie iterations and takes a total of 16s, so each iteration takes about 1.7ms and I am not sure how much we can improve on that. What do you think?

mikhailOK:
Trie iterator is pretty slow, we can try to optimize it with some pre-fetching from rocksdb but it's still trie. NEAR Indexer approach would be more efficient for this type of queries if we indexed contract data.

bowenwang1996:
> NEAR Indexer approach would be more efficient for this type of queries if we indexed contract data.

@frol what do you think?

vgrichina:
I think apart from low level performance issue we at least need some way to do pagination to query this as reasonably small chunks for example. 

Otherwise this RPC endpoint not going to be useful in practice. 

BTW is there specific reason we have to store state in trie vs more efficient key value storage? Merkle proofs? Or something else?

frol:
@bowenwang1996 you can do anything you want with indexer, which is basically a thin wrapper on top of the view-client. I only fear that in this case we may find ourselves re-implementing state storage, then accounts lookup, then what is next, view calls? Do we give up on ability to query historical data or do we duplicate data to have fast access to the whole state at once? There are tons of questions that need to be answered based on the usage profile

bowenwang1996:
> BTW is there specific reason we have to store state in trie vs more efficient key value storage? Merkle proofs? Or something else?

We need the state to be verifiable. There is some discussions around moving to more efficient solutions, but that is unlikely to happen any time soon.

[...]

PR #3969, opened 2021-02-17
feat(jsonrpc): Add warn log to jsronrpc methods where time out can happen

### LINKED ISSUE #3951: Log timed out JSON RPC requests
This is a feature that will help us investigate timeout issues like #3938 with more information.

khorolets:
Closes #3951 

Doesn't affect #3938 because that particular method (`query`) is being refactored right now in #3944 

As for the `query` timeout can appear there because of network routing but in my PR which refactors `query` method we totally removed routing (decision made by @bowenwang1996 and @frol) so it seems like timeout from jsonrpc side wouldn't be possible.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3188,6 +3188,7 @@ dependencies = [
  "serde",
  "serde_json",
  "tokio",
+ "tracing",
  "validator",
 ]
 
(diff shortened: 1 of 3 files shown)


frol (approved):

bowenwang1996:
> As for the query timeout can appear there because of network routing but in my PR which refactors query method we totally removed routing (decision made by @bowenwang1996 and @frol) so it seems like timeout from jsonrpc side wouldn't be possible.

While it is possible to have routing today, it is de facto removed because every rpc node tracks the shard today. Moreover, it appears that we often hear query timeout reported by the wallet team, so it would be really helpful if we can log that as well.

PR #3971, opened 2021-02-17
feat: Add precompiled contracts for alt_bn128 curve [rebased]

evgenykuzyakov:
Merged latest.

Based on https://github.com/near/nearcore/pull/2842 by @snjax

This feature adds to near following precompiled contracts:

- alt_bn128_g1_multiexp
- alt_bn128_pairing_check
Example of usage is at https://github.com/zeropoolnetwork/near-groth16-verifier

TODO:
- [x] Update Cargo.toml files to properly expose feature
- [x] Update test-contract to use feature
- [x] Use latest VMLogic to pass protocol version
- [x] Expose nightly feature
- [x] Add contract tests 
- [x] Wrap errors with feature
- [x] Update param estimator to use different versions of test contracts
- [x] Re-run param estimator.

## Test plan:
- CI
- Added tests to test-contract-rs in near-vm-runner.
- Doc tests for alt_bn128 within near-vm-logic


artob:
@evgenykuzyakov Depending on when this would be merged, let's coordinate with #3954.

runtime/runtime-params-estimator/test-contract/build.sh
```diff
@@ -16,10 +16,33 @@ dd if=/dev/urandom of=./res/medium_payload bs=$(expr 102400 - ${bare_wasm}) coun
...
+rm -rf target
+RUSTFLAGS='-C link-arg=-s' cargo +nightly build --target wasm32-unknown-unknown --release --features nightly_protocol_features
+# bare_wasm=$(stat -c%s target/wasm32-unknown-unknown/release/test_contract.wasm)
+bare_wasm=$(stat -f%z target/wasm32-unknown-unknown/release/test_contract.wasm)
+echo ${bare_wasm}
+
+# Generate several files of various sizes. We will compile these files into the Wasm binary to
+# bloat its size to the given values.
+# 10KiB
+dd if=/dev/urandom of=./res/small_payload bs=$(expr 10240 - ${bare_wasm}) count=1
```
ailisp, line 37: these *_payload files should be not committed

ailisp: also res/*_contract.wasm

ailisp: ah seems not related to this PR. We should recompile contract with stable now, as near-sdk-rs examples already using `stable-2020-10-08`: 
https://github.com/near/near-sdk-rs/blob/master/rust-toolchain
https://github.com/near/near-sdk-rs/blob/master/examples/fungible-token/build.sh

ailisp (approved):

runtime/runtime-params-estimator/test-contract/build.sh
evgenykuzyakov: Looks like a bug

evgenykuzyakov: > also `res/*_contract.wasm`

They are required for sanity check, because we directly embed them into the param estimator, we should probably switch to lazy initialization.

runtime/near-vm-logic/src/logic.rs
```diff
@@ -780,6 +780,97 @@ impl<'a> VMLogic<'a> {
...
+        value_ptr: u64,
+        register_id: u64,
+    ) -> Result<()> {
+        self.gas_counter.pay_base(alt_bn128_g1_multiexp_base)?;
+        let value_buf = self.get_vec_from_memory_or_register(value_ptr, value_len)?;
+        let len = value_buf.len() as u64;
+        self.gas_counter.pay_per_byte(alt_bn128_g1_multiexp_byte, len)?;
+
+        let discount = (alt_bn128_g1_multiexp_base as u64
+            + alt_bn128_g1_multiexp_byte as u64 * len)
```
bowenwang1996, line 812: is it possible for this to overflow?

bowenwang1996 (approved):

runtime/near-vm-logic/src/logic.rs
evgenykuzyakov: no

core/primitives-core/src/config.rs
```diff
@@ -290,6 +291,31 @@ pub struct ExtCostsConfig {
     pub validator_stake_base: Gas,
     /// Cost of calling `validator_total_stake`.
     pub validator_total_stake_base: Gas,
+
+    // #############
+    // # Alt BN128 #
+    // #############
+    /// Base cost for multiexp
+    #[cfg(feature = "protocol_feature_alt_bn128")]
```
willemneal, line 299: This could be addressed in a refactor PR, but @matklad might know a way we can wrap all of these fields additions with just one flag instead of each.

core/primitives-core/src/config.rs
```diff
@@ -348,6 +374,20 @@ impl Default for ExtCostsConfig {
             promise_return: SAFETY_MULTIPLIER * 186717462,
             validator_stake_base: SAFETY_MULTIPLIER * 303944908800,
             validator_total_stake_base: SAFETY_MULTIPLIER * 303944908800,
+            #[cfg(feature = "protocol_feature_alt_bn128")]
```
willemneal, line 377: Same here.

core/primitives-core/src/config.rs
```diff
@@ -404,6 +444,20 @@ impl ExtCostsConfig {
             promise_return: 0,
             validator_stake_base: 0,
             validator_total_stake_base: 0,
+            #[cfg(feature = "protocol_feature_alt_bn128")]
```
willemneal, line 447: And here.

core/primitives-core/src/config.rs
olonho: Why not just have an inner struct under feature?

runtime/near-vm-runner/tests/test-contract-rs/src/lib.rs
```diff
@@ -502,12 +551,19 @@ pub unsafe fn recurse() {
     value_return(data.len() as u64, data.as_ptr() as u64);
 }
 
+/// Rust compiler is getting smarter and starts to optimize my deep recursion.
+/// We're going to fight it with a more obscure implementations.
```
olonho, line 555: :)

olonho (approved):

core/primitives-core/src/config.rs
matklad: Yeah, given that all these start with `alt_bn128_g1_`, it really wants to be

```rust
pub alt_bn128_g1: AltBn128g1
```

regardless of a feature flags.

runtime/near-vm-runner/tests/test-contract-rs/src/lib.rs
matklad: Curiously, `#[inline(never)]` doesn't help: https://godbolt.org/z/7sjn9x

core/primitives-core/src/config.rs
evgenykuzyakov: This might break cost counter right now

evgenykuzyakov:
Here is the param estimator run with 200000/20000 accounts: https://gist.github.com/evgenykuzyakov/5ac71f68da3eab1d37dae33c60c35eda

```
      "alt_bn128_g1_multiexp_base": 237668976500,
      "alt_bn128_g1_multiexp_byte": 1111697487,
      "alt_bn128_g1_sum_base": 1058438125,
      "alt_bn128_g1_sum_byte": 25406181,
      "alt_bn128_g1_multiexp_sublinear": 1441698,
      "alt_bn128_pairing_check_base": 3228502967000,
      "alt_bn128_pairing_check_byte": 8858396182
```

core/primitives-core/src/config.rs
evgenykuzyakov: Discussed that we'll refactor it later

abacabadabacaba:
There is an issue with the way these functions accept arguments. Basically, the arguments to all these functions are arrays of fixed-sized structures. The functions take their arguments in Borsh-serialized form: first 4 bytes encode the length, and the remaining bytes are the concatenation of encoded elements.

It is easy to notice that the length is redundant: since the size of each element is the same, the number of elements can be easily computed from the length of the input. Also, any code that is going to call these functions (e.g. SDK code) will already have these structures as an array, but without the length prepended. So, it will probably have to copy the entire array just to prepend this value which is useless anyway.

Also, this code charges gas per byte, while it would be more correct to charge it per element. So, replacing Borsh deserialization with a trivial one (check that the input length is a multiple of the number of bytes per element, then decode each element) would make the code more efficient both on the contract side and on the runtime side.

bowenwang1996:
@evgenykuzyakov please take a look at what @abacabadabacaba said

evgenykuzyakov:
> There is an issue with the way these functions accept arguments. Basically, the arguments to all these functions are arrays of fixed-sized structures. The functions take their arguments in Borsh-serialized form: first 4 bytes encode the length, and the remaining bytes are the concatenation of encoded elements.

But all other parameters still has to be serialized with Borsh. You don't just copy them in. So you have to manually serialize every element to save 4 bytes. You still need a copy.

abacabadabacaba:
@evgenykuzyakov Isn't it possible to take the values directly from a slice, without Borsh serialization? Suppose that you have a Rust slice with those structures inside a smart contract, and you pass its address and length (in bytes) to the host function, can't the host function just decode this data? In this case, no Borsh is needed on the smart contract side at least.

evgenykuzyakov:
>  Isn't it possible to take the values directly from a slice, without Borsh serialization? Suppose that you have a Rust slice with those structures inside a smart contract, and you pass its address and length (in bytes) to the host function, can't the host function just decode this data? In this case, no Borsh is needed on the smart contract side at least.

it's possible to do, but the inner structure are still Borsh encoded. The inner structure memory alignment is not guaranteed to match Borsh alignment. While Borsh specific serialization gives such guarantee.

abacabadabacaba:
As far as I understand, the inner structures basically consist of fixed-size integers, and their in-memory representation matches their Borsh representation. So Borsh serialization will mostly just copy data in memory (but in an inefficient way), and that can be avoided just by not using Borsh serialization and instead passing the in-memory data directly. In any case, the alignment of any values in smart contract memory must be deterministic and repeatable (because the execution of smart contracts in general is deterministic and repeatable).

joshuajbouw:
Just to add into this a little bit, shouldn't we maintain our own `bn` lib since we rely on a not very well used `zeropool-bn` crate which forked `bn`? I already had a head start over here: https://github.com/aurora-is-near/aurora-bn which can be used.

PR #3972, opened 2021-02-18
feat(runtime): white list some always in memory contract

ailisp:
For example, near_evm contract. It's determined by contract's account id.

Test Plan
---------
near vm runner tests; 
doing debug print as in `cache-slow-repro-whitelist` branch and see no deserialization caused latency even when in memory cache is disabled for near_evm.wasm contract. And when it's not in whitelist, this deserialization latency happens, as before.


nearmax:
I suggest to extract `ALWAYS_IN_MEM_CONTRACT` into a config file.

runtime/near-vm-runner/src/cache.rs
```diff
@@ -3,16 +3,55 @@ use crate::prepare;
...
 use wasmer_runtime_core::load_cache_with;
 
+lazy_static! {
+    static ref ALWAYS_IN_MEM_CONTRACT: RwLock<HashMap<AccountId, Option<(CryptoHash, Result<wasmer_runtime::Module, VMError>)>>> = {
+        let mut m = HashMap::new();
+        m.insert("evm".to_string(), None);
+        // Add more always in memory contract account ids
+        RwLock::new(m)
+    };
+}
```
matklad, line 27: Software-architecture wise, I am somewhat surprised that we use a global variable here, instead of storing this as a field of some kind of context struct. As we might want to have several runtime instances (for sharding), I fear that this'll bite us down the road. But this isn't specific to the PR, the `MODULES` below suffer from the same problem.

olonho: Overall it doesn't look like proper long-term approach. 
First: the name is not proper contract identifier, content hash is.
Second: this kind of information belongs to configuration, not to the binary code.

ailisp: Yes it's problematic for multiple runner instance, existing code assumed there's only one runner

ailisp:
> I suggest to extract ALWAYS_IN_MEM_CONTRACT into a config file.

@nearmax I suppose this goes to config.json instead of genesis config, to not introduce protocol change

runtime/near-vm-runner/src/cache.rs
bowenwang1996: As @matklad mentioned, it becomes a problem when we introduce parallel runtimes, which may be very soon

neard/src/runtime.rs
```diff
@@ -128,6 +129,7 @@ pub struct NightshadeRuntime {
     genesis_runtime_config: Arc<RuntimeConfig>,
 
     store: Arc<Store>,
+    in_memory_contracts: InMemoryContracts,
```
bowenwang1996, line 132: why don't we make it per shard to reduce potentially lock contention?

neard/src/config.rs
```diff
@@ -550,6 +552,7 @@ pub struct NearConfig {
     pub client_config: ClientConfig,
     pub network_config: NetworkConfig,
     pub rpc_config: RpcConfig,
+    pub cache_config: CacheConfig,
```
matklad, line 555: This is a too general name for this context. Ie, it's not clear it cache applies to chain/network/vm. `VMCacheConfig` would be better.

As an aside, it actually should be `VmCacheConfig`, according to Rust naming rules, but we have `VM` everywhere, so I guess its better to roll with that.

runtime/near-vm-runner/src/runner.rs
```diff
@@ -30,6 +31,7 @@ pub fn run<'a>(
     promise_results: &'a [PromiseResult],
     current_protocol_version: ProtocolVersion,
     cache: Option<&'a dyn CompiledContractCache>,
+    in_mem_contract: InMemoryContracts,
```
matklad, line 34: General observation: function with many arguments are also functions with frequently changing set of arguments. We should look into packing some of the args here into a struct, to reduce code churn:

https://github.com/rust-analyzer/rust-analyzer/blob/master/docs/dev/style.md#functions-with-many-parameters

but this is unrelated to the PR

neard/src/runtime.rs
matklad: I'd probably suggest to YAGNI that, unless the code is proven guilty for a perf regression. With per-shard, we'll be using N times more memory storing the same bytes. I think a more promising optimization avenue here (should we need it) is to precompile stuff at stratup, and then share a read-only cache.

core/primitives/src/runtime/in_memory_contract.rs
```diff
@@ -0,0 +1,48 @@
+use std::collections::HashMap;
+use std::sync::RwLock;
+use std::sync::Arc;
+use crate::{
+    hash::CryptoHash,
+    types::AccountId,
+};
+use near_vm_errors::VMError;
+
+pub type InMemoryContracts = Arc<RwLock<HashMap<AccountId, Option<(CryptoHash, Result<wasmer_runtime::Module, VMError>)>>>>;
```
matklad, line 10: Hm, taking a bird's eye view, the fact fact the datastructure for in-memory cache which is used by the vm-runner crate, is defined outside the vm-runner crate feels wrong. 

I guess it's reasonable place to do this, given the current structure of code. But the structure itself is suspicious .

ailisp: this should be struct InMemoryContracts(Arc ...) instead of type. but it should be here because primitivies/src/runtime/apply_state.rs is the place to use it, if it were in runtime crate there would be an additional circular dependency primitives depends on runtime

neard/src/config.rs
ailisp: Sounds good

neard/src/runtime.rs
ailisp: Yeah, make it per shard sounds good

runtime/near-vm-runner/src/runner.rs
ailisp: Will do in later refactor PRs, you're also welcome to do, your refactor PRs look very good

neard/src/config.rs
ailisp: renamed

neard/src/runtime.rs
ailisp: Github didn't show @matklad 's comment when i comment previously. I thinks per shard mem is better. We don't try to tell whether it's a same bytes before cache it (it's cached based on account name, not contract bytes), so won't spend more memory.

runtime/near-vm-runner/src/cache.rs
ailisp: Has been redo to keep in mem in NightshadeRuntime, and pass in as arguments

core/primitives/src/runtime/in_memory_contract.rs
```diff
@@ -0,0 +1,56 @@
...
+        match self.0.read().unwrap().get(account_id) {
+            Some(Some((k, r))) => {
+                if k == key {
+                    Some(Some(r.clone()))
+                } else {
+                    None
+                }
+            }
+            Some(None) => Some(None),
+            None => None,
```
EgorKulikov, line 43: Why do we return None when there is a contract with another key and when there is no such account_id, but Some(None) when there is no contract for this account? This seem weird

neard/src/runtime.rs
bowenwang1996: @matklad yeah there will be no repetition yet unless we introduce contracts that are globally accessible/

core/primitives/src/runtime/in_memory_contract.rs
EgorKulikov: it may well be intended, but I think it is at least worth documenting

core/primitives/src/runtime/in_memory_contract.rs
```diff
@@ -0,0 +1,56 @@
+use crate::{hash::CryptoHash, types::AccountId};
+use near_vm_errors::VMError;
+use std::collections::HashMap;
+use std::sync::Arc;
+use std::sync::RwLock;
+
+#[derive(Clone)]
+pub struct InMemoryContracts(
+    pub  Arc<
+        RwLock<HashMap<AccountId, Option<(CryptoHash, Result<wasmer_runtime::Module, VMError>)>>>,
```
bowenwang1996, line 10: it is better to extract the value of the hashmap into a separate type.

bowenwang1996 (commented):
I don't think we should use the same cache for both client and view client. We should either use no cache for view client or use a separate cache.

core/primitives/src/runtime/in_memory_contract.rs
ailisp: Good catch, it should be Some(None) too. I also add a doc string

ailisp:
@bowenwang1996 
this pr didn't change view_client, do you mean state_viewer? if so state_viewer is an ops tool, I think better to pass an empty contract cache, to avoid making main api `run_vm` etc takes Option<InMemoryContracts> (add additional `Option`). Use a separate cache for state_viewer turns this into a trait, also add unnecessary cost to `run_vm`.

EgorKulikov (approved):

core/primitives/src/runtime/in_memory_contract.rs
```diff
@@ -0,0 +1,61 @@
...
+            map.insert(account_id.clone(), None);
+        }
+        Self(Arc::new(RwLock::new(map)))
+    }
+
+    /// Get contract in InMemoryContracts cache
+    /// Returns Some(ContractResult) when account is in whitelist and contract cache is valid
+    /// Returns Some(None) when account is in whitelist but contract cache isn't exist (first compile),
+    /// or invalid (another contract being deployed and compiled)
+    /// Returns None when account is not in whitelist
```
frol, line 33: Can we actually just use a custom enum with 3 variants?

bowenwang1996:
> this pr didn't change view_client, do you mean state_viewer?

Oh I see you already pass the empty cache here https://github.com/near/nearcore/pull/3972/files#diff-9811f5bb9cc495ab24e1f7ab510a6ff683aeef10dee2a59698f3dd05fdbb08c5R146

neard/src/runtime.rs
matklad: Ack! might make sense to shard this from the start then, if this doesn't make the code more complex.

runtime/near-vm-runner/Cargo.toml
```diff
@@ -26,6 +26,7 @@ near-primitives = { path = "../../core/primitives" }
 log = "0.4"
 near-evm-runner = { path = "../near-evm-runner", optional = true }
 cached = "0.23.0"
+lazy_static = "1.4"
```
matklad, line 29: No longer needed?

core/primitives/src/runtime/in_memory_contract.rs
```diff
@@ -0,0 +1,51 @@
...
+pub struct InMemoryContracts(pub Arc<RwLock<HashMap<AccountId, InMemoryContractEntry>>>);
+
+impl std::fmt::Debug for InMemoryContracts {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "InMemoryContracts")
+    }
+}
+
+impl InMemoryContracts {
+    pub fn new(in_memory_account_ids: &Vec<AccountId>) -> Self {
```
matklad, line 20:
```suggestion
    pub fn new(in_memory_account_ids: &[AccountId]) -> Self {
```

`&Vec<T>` is useless type: https://github.com/rust-analyzer/rust-analyzer/blob/master/docs/dev/style.md#useless-types. 

`s/&vec![]/&[]/g` after this

bowenwang1996:
@ailisp do we still want to proceed with this PR?

ailisp:
@bowenwang1996 I think let's keep it open. If 1) there's prove that wasmer2 deserialize fast enough that we can drop in memory cache, and 2) we've fully tested wasmer2 runner to supersede wasmer 0.17, then we're sure we don't need this PR.

If there is N+1 number of active contracts on mainnet that evict evm contract frequently, before we're ready with 1) 2) then we will need this.

bowenwang1996:
> If there is N+1 number of active contracts on mainnet that evict evm contract frequently, before we're ready with 1) 2) then we will need this.

How exactly does this help? The fees are the same

ailisp:
@bowenwang1996 the problem is not fee, it's when evm contract not in memory cache deserialization takes too long

ailisp:
ping

PR #3974, opened 2021-02-18
feat: add `tracing` infrastructure to contract runtime

matklad:
(builds on top of #3973)

`tracing` is a generic instrumentation framework. It allows manually
instrumented code to signal point-in-time events and nested spans.
Separately, a consumer for those spans and events can be defined.

In this particular case, we use `tracing` to measure time of various
phases of contract execution, with the output ending up like this (read
it bottom up):

        398.15ms compile_module
      398.19ms compile_module_cached_wasmer_impl
      140.19µs run_wasmer/call
    407.21ms run_wasmer

The output's format can certainly be improved, but the beauty of
`tracing` is that it can be done separately from instrumentation.

`tracing` is a bit over-engineered to my personal taste, but other than
that it is a well-maintained crate with a large ecosystem buy-in, and is
a solid bet.

The following pattern is suggested for using tracing:

```rust
  fn some_function(arg: Arg) {
    let span = tracing::info_span!("some_function");
    let _span = span.enter();

    let foo = {
        let span = tracing::info_span!("some_function/foo");
        let _span = span.enter();

        complex_computation_a() + long_computation_b()
    }
  }
```

That is:

* use `info` level (no strong motivation here, maybe we should switch to
  debug?)
* use function name as the message (helps finding it using "goto
  symbol" in an IDE)
* use `func_name/section` format to instrument sections of code, but
  prefer to extract them into a separate function.
* do not use `#[instrument]` attribute (it requires all args to be
  `Debug`, and works nicely only for common case).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3633,6 +3633,7 @@ dependencies = [
  "near-vm-logic",
  "parity-wasm",
  "pwasm-utils",
+ "tracing",
  "wabt",
  "wasmer",
  "wasmer-compiler-singlepass",
@@ -3655,6 +3656,8 @@ dependencies = [
  "serde",
  "serde_json",
  "strum",
+ "tracing",
+ "tracing-subscriber",
 ]
 
 [[package]]
@@ -5840,9 +5843,9 @@ checksum = "360dfd1d6d30e05fda32ace2c8c70e9c0a9da713275777f5a4dbb8a1893930c6"
 
 [[package]]
 name = "tracing"
-version = "0.1.22"
+version = "0.1.25"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "9f47026cdc4080c07e49b37087de021820269d996f581aac150ef9e5583eefe3"
+checksum = "01ebdc2bb4498ab1ab5f5b73c5803825e60199229ccba0698170e3be0e7f959f"
 dependencies = [
  "cfg-if 1.0.0",
  "pin-project-lite",
@@ -5852,9 +5855,9 @@ dependencies = [
 
 [[package]]
 name = "tracing-attributes"
-version = "0.1.11"
+version = "0.1.13"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "80e0ccfc3378da0cce270c946b676a376943f5cd16aeba64568e7939806f4ada"
+checksum = "a8a9bd1db7706f2373a190b0d067146caa39350c486f3d455b0e33b431f94c07"
 dependencies = [
  "proc-macro2 1.0.24",
  "quote 1.0.8",
--- a/runtime/near-vm-runner-standalone/Cargo.toml
+++ b/runtime/near-vm-runner-standalone/Cargo.toml
@@ -24,6 +24,8 @@ clap = "2.33.0"
 base64 = "0.13"
 strum = "0.20"
 num-rational = { version = "0.3" }
+tracing = { version = "0.1"}
+tracing-subscriber = "0.2"
 
 near-vm-logic = { path = "../near-vm-logic", version = "3.0.0", features = ["costs_counting"]}
 near-vm-runner = { path = "../near-vm-runner", version = "3.0.0", features = ["wasmtime_vm", "wasmer1_vm"] }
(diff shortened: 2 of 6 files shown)


runtime/near-vm-runner/src/cache.rs
```diff
@@ -17,6 +17,9 @@ pub(crate) fn compile_module(
     code: &[u8],
     config: &VMConfig,
 ) -> Result<wasmer_runtime::Module, VMError> {
+    let span = tracing::info_span!("compile_module");
+    let _span = span.enter();
```
ailisp, line 21: Can this be simplified to `tracing::info_span!("compile_module").enter();`

ailisp (approved):
Beautiful API and clearly show which function is called by which one, nice!

bowenwang1996 (commented):
@matklad should we extract `tracing_timings.rs` and put it in `tools` for example? We can also potentially replace delay detector with this. cc @pmnoxx

willemneal:
I agree with Bowen, this seems like the start of a more general purpose tool which I could be used for code compiled to Wasm.

pmnoxx:
> (builds on top of #3973)
> 
> `tracing` is a generic instrumentation framework. It allows manually
> instrumented code to signal point-in-time events and nested spans.
> Separately, a consumer for those spans and events can be defined.
> 
> In this particular case, we use `tracing` to measure time of various
> phases of contract execution, with the output ending up like this (read
> it bottom up):
> 
> ```
>     398.15ms compile_module
>   398.19ms compile_module_cached_wasmer_impl
>   140.19µs run_wasmer/call
> 407.21ms run_wasmer
> ```
> 
> The output's format can certainly be improved, but the beauty of
> `tracing` is that it can be done separately from instrumentation.
> 
> `tracing` is a bit over-engineered to my personal taste, but other than
> that it is a well-maintained crate with a large ecosystem buy-in, and is
> a solid bet.
> 
> The following pattern is suggested for using tracing:
> 
> ```rust
>   fn some_function(arg: Arg) {
>     let span = tracing::info_span!("some_function");
>     let _span = span.enter();
> 
>     let foo = {
>         let span = tracing::info_span!("some_function/foo");
>         let _span = span.enter();
> 
>         complex_computation_a() + long_computation_b()
>     }
>   }
> ```
> 
> That is:
> 
> * use `info` level (no strong motivation here, maybe we should switch to
>   debug?)
> * use function name as the message (helps finding it using "goto
>   symbol" in an IDE)
> * use `func_name/section` format to instrument sections of code, but
>   prefer to extract them into a separate function.
> * do not use `#[instrument]` attribute (it requires all args to be
>   `Debug`, and works nicely only for common case).

I was thinking about implementing something similar with a small different. I would like to integrate those stats with `utils/near-performance-metrics` and print them once a minute.

I would like to see something like this 
```

>     compile_module 398.15ms/30
>   compile_module_cached_wasmer_impl 398.19ms/15 
>   run_wasmer/call 140.19µs(199µs)/10 
> run_wasmer 407.21ms(350ms)/2
```
We could print both time data in following format avg time(max time)/number of calls.
I was going to implement something like for the purpose of investigating how much times is spend in different methods.

runtime/near-vm-runner/src/cache.rs
matklad: hopefully soon: https://github.com/tokio-rs/tracing/pull/1252!

matklad:
>should we extract tracing_timings.rs and put it in tools for example?

@bowenwang1996 I prefer to let the general purpose bits mature near the call-site and abstracting them away after they are sufficiently ready. In particular, creating a whole crate for a single file seems overkill. Long-term, I think this'll be moved directly to a separate repo & crates.io crate, or maybe even folded directly with https://github.com/davidbarsky/tracing-tree. 

>I was going to implement something like for the purpose of investigating how much times is spend in different methods.

@pmnoxx I would suggest taking a look at the https://tracing.rs/tracing/, I think their infra is right for  this task. In particular, with tracing you can easily vary the consumers of the events streams. This is useful, because I imagine we want different consumers for "monitoring in production" and "debugging performance problems locally".

frol (approved):

olonho (approved):

PR #3979, opened 2021-02-19
Print RocksDb memory usage every minute

pmnoxx:
In this PR we add printing memory real RocksDB memory usage every minute.
To make this feature work, you need to start near with
`LD_PRELOAD=${PWD}/bins/near-c-allocator-proxy.so nearup ...`
from https://github.com/near/near-memory-tracker/blob/master/near-dump-analyzer


pmnoxx:
This approach will not work.

utils/near-performance-metrics/src/stats_enabled.rs
```diff
@@ -206,6 +239,10 @@ impl Stats {
             other_memory_size / MEBIBYTE,
             other_memory_count
         );
+        let rocksdb_memory_usage = get_rocksdb_memory_usage();
+        if rocksdb_memory_usage > 0 {
+            info!("    RocksDB memory usage: {}MiB", rocksdb_memory_usage / MEBIBYTE);
```
matklad, line 244: You might want to steal the `Bytes` type from rust-analyzer to avoid `/ MEBIBYTE` on two different call-sites: 

https://github.com/rust-analyzer/rust-analyzer/blob/20a911f3cc2beb0409ab71cc1560648374745f7f/crates/profile/src/memory_usage.rs#L43-L80

In general, I've found that in Rust adding small well typed structs for quantities pays of. With derives, the boilerplate is relatively slim.

pmnoxx: How about we use https://github.com/hyunsik/bytesize/blob/master/src/lib.rs
Their code seems small and clean.

matklad: Agree, that one is a solid crate!

utils/near-performance-metrics/src/stats_enabled.rs
```diff
@@ -168,6 +181,34 @@ pub(crate) fn get_entry() -> Arc<Mutex<ThreadStats>> {
     LOCAL_STATS.with(|x| (*x).clone())
 }
 
+#[cfg(target_os = "linux")]
+fn get_c_memory_usage_cur_thread() -> ByteSize {
```
bowenwang1996, line 185: It seems that with this change, if we run neard with `--performance_stats` without the special way of starting it described in the comment, then it would crash, which I don't think is a good idea

pmnoxx: This code wouldn't crash, malloc is guaranteed to return NULL when allocating given chunks of memory larger than can be allocated.

pmnoxx: I added a feature "c_memory_stats" just in case.

bowenwang1996 (approved):

utils/near-performance-metrics/src/stats_enabled.rs
```diff
@@ -16,11 +16,12 @@ use std::task::Poll;
 use std::time::Duration;
 use std::time::Instant;
 
+use bytesize::ByteSize;
 use strum::AsStaticRef;
 
-const MEBIBYTE: usize = 1024 * 1024;
-const MEMORY_LIMIT: usize = 512 * MEBIBYTE;
-const MIN_MEM_USAGE_REPORT_SIZE: usize = 100 * MEBIBYTE;
+static MEMORY_LIMIT: u64 = 512 * bytesize::MIB;
```
matklad, line 22: Worth sending a PR to `bytesize` with `const`s to make`static MEMORY_LIMIT: ByteSize = ByteSize::mib(512);` work.(obv no need to block PR overt this)

utils/near-performance-metrics/src/stats_enabled.rs
```diff
@@ -48,17 +49,19 @@ pub struct ThreadStats {
...
-            stat: HashMap::new(),
-            cnt: 0,
+            stat: Default::default(),
+            cnt: Default::default(),
             time: Duration::default(),
-            classes: HashSet::new(),
-            in_progress_since: None,
+            classes: Default::default(),
+            in_progress_since: Default::default(),
             last_check: Instant::now(),
```
matklad, line 63: Lol, I *hate* when you could've derived default if not for a singly pesky field )

utils/near-performance-metrics/src/stats_enabled.rs
matklad: Yeah, naively I'd expect a call do a different symbol. But we need [weak linkage](https://en.wikipedia.org/wiki/Weak_symbol) for that, and it's not available in rust at the moment.

utils/near-performance-metrics/src/stats_enabled.rs
```diff
@@ -168,6 +175,34 @@ pub(crate) fn get_entry() -> Arc<Mutex<ThreadStats>> {
...
+}
+
+#[cfg(all(target_os = "linux", feature = "c_memory_stats"))]
+fn get_c_memory_usage() -> ByteSize {
+    // hack to get memory usage stats for c memory usage
+    // This feature will only work if near is started with environment
+    // LD_PRELOAD=${PWD}/bins/near-c-allocator-proxy.so nearup ...
+    // from https://github.com/near/near-memory-tracker/blob/master/near-dump-analyzer
+    unsafe { ByteSize::b(libc::malloc(usize::MAX) as u64) }
+}
```
matklad, line 199: Btw, in rust-analyzer we use jemalloc's API for that: https://github.com/rust-analyzer/rust-analyzer/blob/master/crates/profile/src/memory_usage.rs#L27-L31

obv, works only with jemalloc.

In theory, theres' API in glibc for that as well (mallinfo) but in practice it's just broken

matklad (approved):

utils/near-performance-metrics/src/stats_enabled.rs
pmnoxx:
```
error[E0015]: calls in statics are limited to constant functions, tuple structs and tuple variants
  --> utils/near-performance-metrics/src/stats_enabled.rs:22:34
   |
22 | static MEMORY_LIMIT2: ByteSize = ByteSize::mb(512);
   |                                  ^^^^^^^^^^^^^^^^^

```

utils/near-performance-metrics/src/stats_enabled.rs
pmnoxx: Yes, there is nothing we can do about it.

pmnoxx: Also by doing this approach, we can get memory stats per thread, which is nice.

utils/near-performance-metrics/src/stats_enabled.rs
pmnoxx: Also `+=` operator is missing

pmnoxx: https://github.com/hyunsik/bytesize/pull/21

PR #3983, opened 2021-02-19
feat: allow more sophisticated vm benchmarks 

matklad:
We now can run several contracts in a row which allows us to, eg,
observe caching effects:

        374.26ms compile_module
      374.29ms compile_module_cached_wasmer_impl
      123.06µs run_wasmer/call
    383.05ms run_wasmer

      52.46µs run_wasmer/call
    8.45ms run_wasmer

      57.97µs run_wasmer/call
    8.47ms run_wasmer

      128.15µs run_wasmer/call
    8.59ms run_wasmer

This includes #3974 just to get nice numbers, but can be landed separately. 

At the moment, the `Script` has only programmatic interface -- the best way to create your custom test scenario is to write a `#[test]`. We *can* add some kind of interpreter here, so that the script can be read as text, but I feel that just using Rust would be more productive.

### DIFF
--- a/runtime/near-vm-runner-standalone/src/script.rs
+++ b/runtime/near-vm-runner-standalone/src/script.rs
@@ -1,0 +1,220 @@
+use std::fs;
+use std::path::Path;
+
+use near_primitives::types::CompiledContractCache;
+use near_primitives_core::profile::ProfileData;
+use near_primitives_core::runtime::fees::RuntimeFeesConfig;
+use near_vm_logic::mocks::mock_external::MockedExternal;
+use near_vm_logic::types::PromiseResult;
+use near_vm_logic::{ProtocolVersion, VMConfig, VMContext, VMKind, VMOutcome};
+use near_vm_runner::{run_vm_profiled, MockCompiledContractCache, VMError};
+
+use crate::State;
+
+#[derive(Clone, Copy)]
+pub struct Contract(usize);
+
+/// Constructs a "script" to execute several contracts in a row. This is mainly
+/// intended for VM benchmarking.
+pub struct Script {
+    contracts: Vec<Vec<u8>>,
+    vm_kind: VMKind,
+    vm_config: VMConfig,
+    protocol_version: ProtocolVersion,
+    profile: ProfileData,
+    contract_cache: Option<Box<dyn CompiledContractCache>>,
+    initial_state: Option<State>,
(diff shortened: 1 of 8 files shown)


ailisp (approved):
Didn't think of making a script mod and a "script DSL in rust", very cool!

matklad:
Added support for contract cache

runtime/near-vm-runner-standalone/src/script.rs
```diff
@@ -0,0 +1,222 @@
...
+use near_vm_runner::{run_vm_profiled, MockCompiledContractCache, VMError};
+
+use crate::State;
+
+#[derive(Clone, Copy)]
+pub struct Contract(usize);
+
+/// Constructs a "script" to execute several contracts in a row. This is mainly
+/// intended for VM benchmarking.
+pub struct Script {
```
olonho, line 19: Neat DSL!

runtime/near-vm-runner-standalone/src/script.rs
```diff
@@ -0,0 +1,222 @@
...
+impl Default for Script {
+    fn default() -> Self {
+        Script {
+            contracts: Vec::new(),
+            vm_kind: VMKind::default(),
+            vm_config: VMConfig::default(),
+            protocol_version: ProtocolVersion::MAX,
+            profile: ProfileData::new_disabled(),
+            contract_cache: None,
+
```
olonho, line 54: Why extra space?

runtime/near-vm-runner-standalone/src/script.rs
```diff
@@ -0,0 +1,222 @@
...
+        self
+    }
+    #[allow(unused)]
+    pub(crate) fn repeat(&mut self, n: u32) -> &mut Step {
+        self.repeat = n;
+        self
+    }
+}
+
+fn default_vm_context() -> VMContext {
```
olonho, line 175: Maybe we could make it better tunable in the future.

runtime/near-vm-runner-standalone/src/script.rs
matklad: No good reason really, removed it! 

My original thinking was to separate global readonly "static" info like `VmKind` from mutable things likes state, but that clearly won't survive the first refacroring of this code

olonho (approved):

PR #3984, opened 2021-02-19
fix: minor soundness issue

matklad:
These two impls are unsound -- they might lead to undefined behavior in
safe rust. Luckily, we only use `ProfileData` with a non-default feature
enabled, so this isn't a problem in practice.

Doubly lucky, seems like these impls are actually unused?

If, in the future, we actually need a SyncVersion of this, we should use
an array of atomics with Relaxed fetch_adds.


matklad:
Let's see if this passes CI -- I tried to check locally with various features, but one never knows when there isn't yet another feature combination :)

matklad:
Ok, this fails compilation of tests, so this is going to be slightly trickier to handle...

matklad:
should be better now

frol (approved):

bowenwang1996 (approved):

frol:
@matklad please, fmt

evgenykuzyakov (approved):

PR #3985, opened 2021-02-19
Concurrent preloader for contracts

olonho:
To minimize deserialization delays we introduce mechanism of pipelined constructing
WASM modules before execution.

On simulated tests (with disabled in-memory instance cache) we get the following results.
```
time cargo test --release  --color=always --test test_vm_runner test_vm_run_sequential --no-fail-fast --manifest-path runtime/near-vm-runner/Cargo.toml -- --format=json --exact -Z unstable-options --show-output
time cargo test --release  --color=always --test test_vm_runner test_vm_run_preloaded --no-fail-fast --manifest-path runtime/near-vm-runner/Cargo.toml -- --format=json --exact -Z unstable-options --show-output
```

Sequential execution (0ms cache delay):

```
real	0m1.152s
user	0m0.934s
sys	0m0.241s
```

Preloaded execution (0ms cache delay):

```
real	0m0.810s
user	0m1.202s
sys	0m0.367
```

Sequential execution (5ms cache delay):

```
real	0m2.726s
user	0m1.277s
sys	0m0.305s
```

Preloaded execution (5ms cache delay):

```
real	0m1.021s
user	0m1.133s
sys	0m0.309s
```

So with realistic cache delays we get almost 3x speedup.


runtime/near-vm-runner/src/preload.rs
```diff
@@ -0,0 +1,136 @@
...
+        for request in requests {
+            let index = self.prepared.len();
+            let (tx, rx) = channel();
+            self.prepared.push(CallInner { rx });
+            let copy_request = request.clone();
+            let tx = tx.clone();
+            self.pool.execute(move || {
+                prepare_in_thread(copy_request, vm_kind, tx);
+            });
+            result.push(ContractCallPrepareResult { handle: Some(index), error: None });
```
ailisp, line 70: why handle need to wrap in `Some`?

runtime/near-vm-runner/src/preload.rs
```diff
@@ -0,0 +1,136 @@
...
+        ext: &mut dyn External,
+        context: VMContext,
+        vm_config: &'a VMConfig,
+        fees_config: &'a RuntimeFeesConfig,
+        promise_results: &'a [PromiseResult],
+        current_protocol_version: ProtocolVersion,
+        profile: ProfileData,
+    ) -> (Option<VMOutcome>, Option<VMError>) {
+        match &prepared.error {
+            Some(error) => return (None, Some(error.clone())),
```
ailisp, line 88: Looks like this would never be error because it's constructed by:
```
             result.push(ContractCallPrepareResult { handle: Some(index), error: None });
```
Prepare error would happen here:
```
     tx.send(VMCallData { result }).unwrap();
```
and are saved in self.prepared (`Vec<CallInner>`). so i think to process error, this should refer to `self.prepared[prepared.handle]`?

ailisp (approved):
Nice! I think the benefit is significant. a few nit comment

runtime/near-vm-runner/src/preload.rs
```diff
@@ -0,0 +1,136 @@
...
+    pool: ThreadPool,
+    prepared: Vec<CallInner>,
+}
+
+impl ContractCaller {
+    pub fn new(num_threads: usize) -> ContractCaller {
+        ContractCaller { pool: ThreadPool::new(num_threads), prepared: Vec::new() }
+    }
+
+    pub fn preload<'a>(
```
matklad, line 55:
```suggestion
    pub fn preload(
```

unused lifetime?

runtime/near-vm-runner/src/preload.rs
```diff
@@ -0,0 +1,136 @@
...
+        self: &mut ContractCaller,
```
matklad, line 56:
```suggestion
        &mut self,
```

The explicit `self: Self` forme is idiomatically used only when there isn't a shorthand (eg, in `self: Arc<Self>`)

runtime/near-vm-runner/src/preload.rs
```diff
@@ -0,0 +1,136 @@
...
+        let mut result: Vec<ContractCallPrepareResult> = Vec::new();
+        for request in requests {
+            let index = self.prepared.len();
+            let (tx, rx) = channel();
+            self.prepared.push(CallInner { rx });
+            let copy_request = request.clone();
+            let tx = tx.clone();
+            self.pool.execute(move || {
+                prepare_in_thread(copy_request, vm_kind, tx);
+            });
```
matklad, line 69: The standard idiom for cloning data into closure is to use a block experssion:

```suggestion
            self.pool.execute({
                let request = request.clone();
                let tx = tx.clone();
                move || prepare_in_thread(request, vm_kind, tx)
            });
```

runtime/near-vm-runner/src/preload.rs
```diff
@@ -0,0 +1,136 @@
...
+    pub code: Vec<u8>,
+    pub vm_config: VMConfig,
+    pub cache: Option<Arc<dyn CompiledContractCache>>,
+}
+
+#[derive(Clone)]
+pub struct ContractCallPrepareResult {
+    pub handle: Option<usize>,
+    pub error: Option<VMError>,
+}
```
matklad, line 43: Seems like we can simplify this to just 
```suggestion
pub struct ContractCallPrepareResult {
    rx: Receiver<VMCallData>,
}
```

that is, we don't need `CallInner` and the `prepared` vector, `PrepareResult` itself could carry all the data necessary for resolution.

runtime/near-vm-runner/src/preload.rs
```diff
@@ -0,0 +1,136 @@
...
+                            ext,
+                            context,
+                            vm_config,
+                            fees_config,
+                            promise_results,
+                            profile,
+                            current_protocol_version,
+                        ),
+                    },
+                };
```
matklad, line 110:
```suggestion
                match call_data.result {
                    Err(err) => (None, Some(err)),
                    Ok(module) => match module {
                        Wasmer0(module) => run_wasmer_module(
                            module,
                            method_name,
                            ext,
                            context,
                            vm_config,
                            fees_config,
                            promise_results,
                            profile,
                            current_protocol_version,
                        ),
                    },
                }
```

runtime/near-vm-runner/src/preload.rs
```diff
@@ -0,0 +1,136 @@
...
+fn prepare_in_thread(request: ContractCallPrepareRequest, vm_kind: VMKind, tx: Sender<VMCallData>) {
+    let cache = request.cache.as_deref();
+    let result = match vm_kind {
+        VMKind::Wasmer0 => cache::wasmer0_cache::compile_module_cached_wasmer(
+            &(request.code_hash.0).0,
+            request.code.as_slice(),
+            &request.vm_config,
+            cache,
+        )
+        .map(|module| VMModule::Wasmer0(module)),
```
matklad, line 132:
```suggestion
        .map(VMModule::Wasmer0),
```

runtime/near-vm-runner/src/wasmer_runner.rs
```diff
@@ -259,6 +259,59 @@ pub fn run_wasmer<'a>(
...
+    current_protocol_version: ProtocolVersion,
+) -> (Option<VMOutcome>, Option<VMError>) {
+    if method_name.is_empty() {
+        return (
+            None,
+            Some(VMError::FunctionCallError(FunctionCallError::MethodResolveError(
+                MethodResolveError::MethodEmptyName,
+            ))),
+        );
+    }
```
matklad, line 280: That's interesting! In principle, we could move this error to prepare stage. But that would be a bad idea, because we need to preserve semantics of sequential execution. More generally, prepare call should never fail, all errors should be moved to the moment we actually run the call.

runtime/near-vm-runner/src/preload.rs
olonho: Nope, don't want to expose implementation details (such as VM implementation dependencies) in public API of preloader.

runtime/near-vm-runner/src/wasmer_runner.rs
olonho: Also current semantics of preloader is that we preload the contract, and select method to execute later on.

runtime/near-vm-runner/src/preload.rs
olonho: Indeed, made it somewhat simpler.

runtime/near-vm-runner/src/preload.rs
matklad: But the `rx` can be private, no?

matklad (approved):

Cargo.toml
```diff
@@ -63,6 +63,10 @@ near-chain = { path = "./chain/chain" }
 
 node-runtime = { path = "./runtime/runtime" }
 
+near-vm-logic = { path = "./runtime/near-vm-logic" }
+near-vm-errors = { path = "./runtime/near-vm-errors" }
+near-vm-runner = { path = "./runtime/near-vm-runner" }
```
matklad, line 68: Hm, wait, why we need these here?

matklad: Ah, right, thats' because we add a test to the root package. Naively, I'd expect the test to be within `near-vm-runner` crate, like this:

```diff
diff --git a/Cargo.lock b/Cargo.lock
index 032d7eb9..60f5dd02 100644
Binary files a/Cargo.lock and b/Cargo.lock differ
diff --git a/Cargo.toml b/Cargo.toml
index b19e030e..1dd38eb4 100644
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -63,10 +63,6 @@ near-chain = { path = "./chain/chain" }
 
 node-runtime = { path = "./runtime/runtime" }
 
-near-vm-logic = { path = "./runtime/near-vm-logic" }
-near-vm-errors = { path = "./runtime/near-vm-errors" }
-near-vm-runner = { path = "./runtime/near-vm-runner" }
-
 near-jsonrpc = { path = "./chain/jsonrpc" }
 near-network = { path = "./chain/network" }
 
diff --git a/runtime/near-vm-runner/src/tests.rs b/runtime/near-vm-runner/src/tests.rs
index 34d6636a..7cbaf3cf 100644
--- a/runtime/near-vm-runner/src/tests.rs
+++ b/runtime/near-vm-runner/src/tests.rs
@@ -1,5 +1,6 @@
 mod error_cases;
 mod invalid_contracts;
+mod contract_preload;
 mod rs_contract;
 mod ts_contract;
 
diff --git a/tests/test_vm_runner.rs b/runtime/near-vm-runner/src/tests/contract_preload.rs
similarity index 96%
rename from tests/test_vm_runner.rs
rename to runtime/near-vm-runner/src/tests/contract_preload.rs
index 1cd42463..3ece2215 100644
--- a/tests/test_vm_runner.rs
+++ b/runtime/near-vm-runner/src/tests/contract_preload.rs
@@ -1,7 +1,6 @@
 use near_primitives::hash::hash;
 use near_primitives::runtime::fees::RuntimeFeesConfig;
 use near_vm_logic::{ProtocolVersion, VMConfig, VMContext, VMKind, VMOutcome};
-use near_vm_runner::{run_vm, ContractCallPrepareRequest, ContractCaller, VMError};
 
 use near_primitives::borsh::BorshSerialize;
 use near_primitives::types::CompiledContractCache;
@@ -13,10 +12,13 @@ use std::sync::{Arc, Mutex};
 use std::thread::sleep;
 use std::time::Duration;
 
+use crate::{run_vm, ContractCallPrepareRequest, ContractCaller, VMError};
+
+
 const TEST_CONTRACT_1: &'static [u8] =
-    include_bytes!("../runtime/near-vm-runner/tests/res/test_contract_rs.wasm");
+    include_bytes!("../../tests/res/test_contract_rs.wasm");
 const TEST_CONTRACT_2: &'static [u8] =
-    include_bytes!("../runtime/near-vm-runner/tests/res/test_contract_ts.wasm");
+    include_bytes!("../../tests/res/test_contract_ts.wasm");
 
 fn default_vm_context() -> VMContext {
     return VMContext {
```

But I don't actually know the rule we use to decide if we want to put a test into the top-level tests directory, or into a specific crate.

olonho: Actually that's exactly how it used to be, and you suggested to move it here in earlier comments. I'm kind of OK with either approach, but keep moving it forward and backward doesn't make much sense.

olonho: OK, done

PR #3986, opened 2021-02-20
fix(trie): combining state parts memory usage

mikhailOK:
After getting all state parts, write them to storage one at a time
instead of combining them in memory.

Test plan
---------
test for trie seek and iteration
test that creating parts behavior is the same
test that combining parts behavior is the same


chain/chain/src/types.rs
```diff
@@ -566,11 +566,13 @@ pub trait RuntimeAdapter: Send + Sync {
...
     /// Should be executed after accepting all the parts to set up a new state.
-    fn confirm_state(
+    fn apply_state_part(
         &self,
         shard_id: ShardId,
         state_root: &StateRoot,
-        parts: &Vec<Vec<u8>>,
+        part_id: u64,
+        num_parts: u64,
+        part: &Vec<u8>,
```
bowenwang1996, line 575:
```suggestion
        part: &[u8],
```

core/store/src/trie/iterator.rs
```diff
@@ -52,95 +52,74 @@ impl<'a> TrieIterator<'a> {
...
-                    return Ok(());
+                    if existing_key < key {
+                        self.key_nibbles.extend(existing_key.iter());
+                        *status = CrumbStatus::Exiting;
+                    }
+                    break;
                 }
-                TrieNode::Branch(mut children, _) => {
+                TrieNode::Branch(children, _) => {
                     if key.is_empty() {
```
bowenwang1996, line 87: why does it make sense for key to be empty?

core/store/src/trie/iterator.rs
```diff
@@ -150,94 +129,215 @@ impl<'a> TrieIterator<'a> {
...
+        trie: &Trie,
+        map: &BTreeMap<Vec<u8>, Vec<u8>>,
+        state_root: &CryptoHash,
+        seek_key: &[u8],
+    ) {
+        let mut iterator = trie.iter(&state_root).unwrap();
+        iterator.seek(&seek_key).unwrap();
+        let result1: Vec<_> = iterator.map(Result::unwrap).take(5).collect();
+        let result2: Vec<_> =
+            map.range(seek_key.to_vec()..).map(|(k, v)| (k.clone(), v.clone())).take(5).collect();
```
bowenwang1996, line 340: why do we take 5 here?

core/store/src/trie/state_parts.rs
```diff
@@ -56,26 +55,24 @@ impl Trie {
...
-                }
-                Some(Ok(_)) => {
-                    // The last iteration actually reads a value we don't need.
-                }
-            }
-            // TODO #1603 this is bad for large keys
-            if iterator.key_nibbles >= path_end {
-                break;
+        let path_end = if size_end == root_node.memory_usage {
+            vec![16]
```
bowenwang1996, line 59: can you add a comment explaining what happens here?

core/store/src/trie/state_parts.rs
```diff
@@ -167,133 +165,49 @@ impl Trie {
...
+        if state_root == &CryptoHash::default() {
+            return Ok(TrieChanges::empty(CryptoHash::default()));
+        }
+        let trie = Trie::from_recorded_storage(PartialStorage { nodes: PartialState(part) });
+        let root_node = trie.retrieve_node(&state_root)?;
+        let total_size = root_node.memory_usage;
+        let size_start = (total_size + num_parts - 1) / num_parts * part_id;
+        let size_end = min((total_size + num_parts - 1) / num_parts * (part_id + 1), total_size);
+        let path_begin = trie.find_path(&root_node, size_start)?;
+        let path_end = if size_end == root_node.memory_usage {
```
bowenwang1996, line 193: maybe we should extract this part into a separate function?

bowenwang1996 (commented):
where do we actually save the state parts to storage?

mikhailOK (commented):
part downloader logic validates parts and writes them to `ColStateParts`, after collecting all parts we do `set_state_finalize` which calls `apply_state_part` on each of them. `apply_state_part` reads from `ColStateParts` and writes to `ColState`.

bowenwang1996 (approved):
I don't fully understand the code, but the tests look good to me. Could we start writing a test that involves syncing 30GB of state and see how it works?

mikhailOK:
Manually tested with testnet, memory usage was low and combining parts took about a minute (2.5GB state)

core/store/src/trie/mod.rs
```diff
@@ -63,6 +62,15 @@ enum NodeHandle {
     Hash(CryptoHash),
 }
 
+impl NodeHandle {
+    fn unwrap_hash(&self) -> &CryptoHash {
+        match self {
+            Self::Hash(hash) => hash,
+            Self::InMemory(_) => unreachable!(),
+        }
+    }
```
matklad, line 71: tiny nit, but a more idiomatic version would probably be 

```rust
impl NodeHandle {
    fn as_hash(&self) -> Option<&CryptoHash> {
        match self {
            Self::Hash(hash) => Some(hash),
            Self::InMemory(_) => None,
        }
    }
}
```

that way, we get to re-use Option's `unwrap`.

Kouprin: @matklad I guess this way we're losing important knowledge about `unreachable!()` situation.

matklad: There'll be `.unwrap()` on the call-site, so this just moves the impossibility assertion from the definition site to the call-site. 

There's no much semantic difference between `unreachable!()` and `Option::unwrap` -- both mean that the programmer asserts something at runtime, which can't be explained using rust's type system. 

But, again, this is at most a minor code-style point, in the grand scheme of things, both options are totally fine/

Kouprin: @matklad sure. My point in using `unreachable!()` shows that something is impossible by design, but each `unwrap` must be explained why it's safe. It's not about rust, it's about being clear to devs.

(My understanding might be outdated. If it's possible to point that something is not reachable by design another better way (with no using `unreachable!()`) please let me know.)

matklad: I understand `unreachable!` the same way. But, in my mind, `.unwrap()` also signals impossibility by design? That is, if we look at the call-site,

```rust
hash = *child.unwrap_hash();

/* versus */

hash = *child.as_hash().unwrap();
```

it seems me that both versions say that, by design, `child` is guaranteed to be `Hash` in this particular case. 

The difference is that, when I see `foo.unwrap()`, I immediately know that is signals an impossibility of something. When I see `foo.unwrap_hash()`, I need to check the definition of `unwrap_hash` to realize that it has an unreachable inside.

Kouprin: @matklad not sure is it the standard way of understanding in the team. For example, in my latest work, there are explanations of `unwrap` used (https://github.com/near/nearcore/compare/master...gc_headers_2?expand=1#diff-c26b0a777803f79360d3f469862cfcd95deb3c390809095670f6db88d00332f4R236-R239) and I found them useful.

Basically, I work with concepts written by others as black boxes. That's why, in my opinion, `unwrap`s must be explained, otherwise we may end up with lack of understanding what's possible and why. Seeing `unreachable!()` helps me to dive into details much better than guessing can I reuse the same pattern in another place.

One more example - we use tons of `unwrap`s in testing because we don't want to process errors in any other way. It doesn't mean the methods used in tests should never return errors. :)

matklad: I agree that unwraps should be explained (or, better still, refactored out of existence). But I would also think that, for the same reason, a call to `.unwrap_hash()` should be explained. 

Ie, if I see an uncommitted call to `.unwrap_hash()`, I understand that the author believes, for some good reason, that `unreachable` case won't happen here. But that's exactly the same info I get from seeing `.unwrap()`. That is, `.unwrap()` and `.unwrap_hash()` signal the same level of intentionallyty to me.

Issue #3742, opened 2020-12-14
Improve ViewClient trie cache


mikhailOK:
Reverting #3593 means that it's easier to hit cache inconsistency in view client. ViewClient already has to be tolerant to this type of errors for independent reasons (not reading from snapshots). The goal of this issue is to track possible improvements to the cache.

mikhailOK:
#3879 reduces lock contention.
ViewClient instances still share one cache, so we should keep this open

bowenwang1996:
@mikhailOK is there a plan to address the shared cache there?

bowenwang1996:
@mikhailOK do we still have any action items here? If not, let's close the issue

PR #3988, opened 2021-02-22
fix(store): reset_data_pre_state_sync memory usage

mikhailOK:
Add delete_all to StoreUpdate and implement it as a rocksdb range
delete. Use it in reset_data_pre_state_sync to clear ColState.

Test plan
---------
unit test
existing tests


core/store/src/db.rs
```diff
@@ -381,6 +385,18 @@ impl Database for RocksDB {
...
                 },
+                DBOp::DeleteAll { col } => {
+                    let cf_handle = unsafe { &*self.cfs[col as usize] };
+                    let opt_first = self.db.iterator_cf(cf_handle, IteratorMode::Start).next();
+                    let opt_last = self.db.iterator_cf(cf_handle, IteratorMode::End).next();
+                    assert_eq!(opt_first.is_some(), opt_last.is_some());
+                    if let (Some((min_key, _)), Some((max_key, _))) = (opt_first, opt_last) {
+                        if min_key != max_key {
+                            batch.delete_range_cf(cf_handle, &min_key, &max_key)
+                        }
```
frol, line 396: > Removes the database entries in the range ["begin_key", "end_key"), i.e., including "begin_key" and excluding "end_key". It is not an error if no keys exist in the range ["begin_key", "end_key").

https://docs.rs/rocksdb/0.15.0/rocksdb/struct.WriteBatch.html#method.delete_range_cf

1. It might be a good idea to put this comment into the code for future reference
2. `min_key != max_key` seem redundant, though it does not hurt

frol (approved):

core/store/src/trie/shard_tries.rs
```diff
@@ -73,10 +73,12 @@ impl ShardTries {
...
                 DBOp::Insert { col, .. } if *col == DBCol::ColState => unreachable!(),
-                DBOp::Delete { col, key } if *col == DBCol::ColState => {
+                DBOp::Delete { col, .. } if *col == DBCol::ColState => unreachable!(),
+                DBOp::DeleteAll { col } if *col == DBCol::ColState => {
                     // Delete is possible in reset_data_pre_state_sync
-                    let (shard_id, hash) = TrieCachingStorage::get_shard_id_and_hash_from_key(key)?;
-                    shards[shard_id as usize].push((hash, None));
+                    for shard_id in 0..self.caches.len() {
+                        self.caches[shard_id].clear();
+                    }
```
Kouprin, line 81: Are there any tests checking cache clearing / discrepancy between cache and column?

Kouprin:
`store_update.delete_all` looks very desirable. Thank you @mikhailOK.

Kouprin (approved):

core/store/src/trie/shard_tries.rs
mikhailOK: A discrepancy manifests itself as `StorageInconsistentState`: the invariant is that if a trie node is in cache+storage, then its children have to be there too, and forgetting to clear cache can break it.
At the moment ViewClient has its own cache and allows the error, and regular Client panics if it happens.

bowenwang1996: what was the reason that we allow the error in view client (instead of cleaning the cache)?

bowenwang1996 (approved):

core/store/src/db.rs
```diff
@@ -381,6 +385,17 @@ impl Database for RocksDB {
                 DBOp::Delete { col, key } => unsafe {
                     batch.delete_cf(&*self.cfs[col as usize], key);
                 },
+                DBOp::DeleteAll { col } => {
+                    let cf_handle = unsafe { &*self.cfs[col as usize] };
+                    let opt_first = self.db.iterator_cf(cf_handle, IteratorMode::Start).next();
+                    let opt_last = self.db.iterator_cf(cf_handle, IteratorMode::End).next();
+                    assert_eq!(opt_first.is_some(), opt_last.is_some());
```
ailisp, line 392: would this be None any time

ailisp (approved):

core/store/src/trie/shard_tries.rs
mikhailOK: Tracked by #3742. Two reasons:
1. Invalidating the cache requires taking a lock and we don't want ViewClient to block Client 
2. The error is theoretically possible in ViewClient because it doesn't read from snapshots. Client is single threaded and the only one who can write to storage, so it doesn't have that problem. If we ever decide to use snapshots we can bring back #3092.

PR #3990, opened 2021-02-22
feat(build): Added CARGO_PROFILE_RELEASE_DEBUG=true to nightly and pe…

chefsale:
…rf releases

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,31 +13,31 @@ release:
 	cargo $(RUST_OPTIONS) build -p store-validator --release
 
 debug:
-	cargo $(RUST_OPTIONS) build -p neard
+	CARGO_PROFILE_RELEASE_DEBUG=true cargo $(RUST_OPTIONS) build -p neard
 	cargo $(RUST_OPTIONS) build -p near-vm-runner-standalone
 	cargo $(RUST_OPTIONS) build -p state-viewer
 	cargo $(RUST_OPTIONS) build -p store-validator
 
 perf-release:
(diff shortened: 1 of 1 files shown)


matklad:
LGTM. Note that `debug = true` implies full debug info, which weighs a lot. *If* this is a concern, `debug = 1` (include only line number info) might be an interesting middleground.

chefsale:
cc: @frol and @bowenwang1996  on the debug=1 or debug=true, which makes more sense to you guys?

chefsale:
cc: @frol and @bowenwang1996 on the debug=true or debug=1, whatever works for you?

bowenwang1996 (commented):
Probably better to go with `debug=true` since @pmnoxx also needs it

frol:
@chefsale I had just checked the release binary that was produced by the latest CI and it appears to be 29M in size, which means that it does not have the debug symbols: https://s3-us-west-1.amazonaws.com/build.nearprotocol.com/nearcore/Linux/master/774632825d69f1928b0e7707ebcda42fc1673951/neard

image

chefsale:
That is expected, this is the normal stable release without debug, the one with debug is under perf release folder: https://s3-us-west-1.amazonaws.com/build.nearprotocol.com/nearcore/Linux/master/774632825d69f1928b0e7707ebcda42fc1673951/perf/neard with size of 440MB.

frol:
@chefsale Ouch, I misread the PR. Sorry for the false alarm

Issue #3621, opened 2020-11-16
Avoid unnecessary verification of edges and account announcements


bowenwang1996:
Currently we always first try to verify edges and account announcements when we receive them and due to the broadcast nature of how this information is propagated, we are doing a lot of repetitive verification which could be avoided. Ideally we should not verify the same data twice or verify sync data that are not needed.

bowenwang1996:
This could be combined with the change to update the way we exchange routing table information #3934

bowenwang1996:
This issue is not closed. I think we still do unnecessary verification of account announcement

pmnoxx:
> This issue is not closed. I think we still do unnecessary verification of account announcement

```
    let new_accounts = accounts
            .into_iter()
            .filter_map(|announce_account| {
                if let Some(current_announce_account) =
                    self.routing_table.get_announce(&announce_account.account_id)
                {
                    if announce_account.epoch_id == current_announce_account.epoch_id {
                        None
                    } else {
                        Some((announce_account, Some(current_announce_account.epoch_id)))
                    }
                } else {
                    Some((announce_account, None))
                }
            })
            .collect();
```
It looks like we are already filtering existing accounts.

bowenwang1996:
No I meant that we still potentially send multiple of them to view client to verify at the same time.

pmnoxx:
That makes sense, we need to do something similar to what I did for edges in https://github.com/near/nearcore/pull/4098

PR #3997, opened 2021-02-24
Don't validate edges multiple times

pmnoxx:
We are validating edges multiple times, even if we already have them in our routing table.


chain/network/src/peer_manager.rs
```diff
@@ -1450,12 +1450,24 @@ impl Handler<NetworkRequests> for PeerManagerActor {
...
+                    .filter(|edge| {
+                        if let Some(cur_edge) =
+                            self.routing_table.get_edge(edge.peer0.clone(), edge.peer1.clone())
+                        {
+                            // only consider edges with bigger nonce
+                            return cur_edge.nonce < edge.nonce;
+                        }
+                        true
+                    })
+                    .collect();
```
matklad, line 1464:
```suggestion
                let SyncData { mut edges, accounts } = sync_data;
                edges.retain(|edge| {
                    match self.routing_table.get_edge(edge.peer0.clone(), edge.peer1.clone()) {
                        // only consider edges with bigger nonce
                        Some(cur_edge) => cur_edge.nonce < edge.nonce,
                        None => true,
                    }
                });
```

mfornet (approved):

chain/network/src/peer_manager.rs
```diff
@@ -1450,12 +1450,24 @@ impl Handler<NetworkRequests> for PeerManagerActor {
...
 
                 self.edge_verifier_pool.send(EdgeList(edges.clone()))
-                    .into_actor(self)
-                    .then(move |response, act, ctx| {
-                        match response {
-                            Ok(false) => act.try_ban_peer(ctx, &peer_id, ReasonForBan::InvalidEdge),
+                            .into_actor(self)
```
bowenwang1996, line 1467: why?

bowenwang1996 (commented):
Please try avoiding unnecessary reformatting. Also I'd like to mention that this is related to #3621

chain/network/src/peer_manager.rs
pmnoxx: fixed, Intellij Rust sometimes automatically reformats code wrongly

bowenwang1996 (approved):

PR #4000, opened 2021-02-24
refactor: significantly reduce costs_conting blast radius

matklad:
The goal here is to make sure that only the leaf crates (primitives-core
and vm-logic) and the topmost crate (params-estimator) know about this
feature.

Before this PR, every intermediary crate would have to know about this
feature as well.

To fix this, we encapsulate all conditional compilation logic inside
ProfileData type, while making sure it compiles to no-op unless the
feature is enabled.

### DIFF
--- a/core/primitives-core/src/profile.rs
+++ b/core/primitives-core/src/profile.rs
@@ -19,46 +19,92 @@ impl<T> Clone for FixedArray<T> {
 }
 
 /// Profile of gas consumption.
-pub type ProfileData = FixedArray<u64>;
+#[derive(Clone)]
+pub struct ProfileData {
+    repr: Repr,
+}
 
-pub fn clone_profile(profile: &ProfileData) -> ProfileData {
-    profile.clone()
+#[derive(Clone)]
+enum Repr {
+    #[cfg(feature = "costs_counting")]
+    Enabled {
+        data: FixedArray<u64>,
+    },
+    Disabled,
+}
+
+#[cfg(not(feature = "costs_counting"))]
+const _ASSERT_NO_OP_IF_COUNTING_DISABLED: [(); 0] = [(); std::mem::size_of::<ProfileData>()];
+
+impl Default for ProfileData {
+    fn default() -> ProfileData {
+        ProfileData::new_disabled()
+    }
 }
 
 impl ProfileData {
-    pub fn new() -> Self {
-        Self { data: Rc::new(RefCell::new([0u64; 1 + ExtCosts::count() + ActionCosts::count()])) }
-    }
-
     const EXT_START: usize = 1;
     const ACTION_START: usize = ProfileData::EXT_START + ExtCosts::count();
     // const LENGTH: usize = PROFILE_DATA_LEN;
-    pub fn all_gas(&self) -> Gas {
-        (*self.data.borrow())[0]
+
+    #[inline]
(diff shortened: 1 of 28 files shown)


core/primitives-core/src/profile.rs
```diff
@@ -85,7 +131,7 @@ impl ProfileData {
     }
 
     pub fn set_burnt_gas(&self, burnt_gas: u64) {
-        *self.data.borrow_mut().get_mut(0 as usize).unwrap() = burnt_gas;
+        self.with_data(|data| data[0] = burnt_gas);
```
willemneal, line 134:
```suggestion
        self.add_val(0, burnt_gas);
```

willemneal (approved):
LGTM!  Thanks for cleaning this up!  I have one question about this approach in general, will the rust compiler remove code paths that lead to no-ops like this?

frol (approved):

core/primitives-core/src/profile.rs
matklad: I'd rather keep the manual `=` -- otherwise, setting gas twice in a row would be surprising. Granted, one probably shouldn't set the gas twice right now, but who knows what happens with code in the future.

matklad:
@willemneal the short answer is yes, this is guaranteed. 

The long answer is that this is guaranteed, with a caveat about inlining. 

If the data structure has size of zero (is a ZST), than it is absolutely guaranteed that it won't be present in the resulting program immage, and that even things like `Vec<MyZST>` would not allocate, and would be just glorified counters. 

The situation with functions is a bit more complicated. An empty function will be removed, unless all three are true:

* it is non-generic
* it is defined in another crate
* it isn't `#[inline]`

So if you have

```rust
pub fn noop() {
}
```

in crate foo

and 

```
fn main() {
  foo::noop()
}
```

in crate bar, then the compiled code will contain a `call`. This is the fallout from Rust ultimately relying on the C compilation model. When a function is non-generic, the library for the corresponding crate will contain machine code for a function, and it won't be possible to inline it into some other crate.

However, if the function is generic (or marked with `#[inline]`), the compiled library will also include MIR (rustc's IR) of the function. With access to MIR, compiler will be able to do cross-crate inlining, and ultimately, to optimze the function calls away.

core/primitives-core/src/profile.rs
willemneal: Fair enough.

ailisp (approved):
Nice!

core/primitives-core/src/profile.rs
```diff
@@ -19,46 +19,92 @@ impl<T> Clone for FixedArray<T> {
...
+impl Default for ProfileData {
+    fn default() -> ProfileData {
+        ProfileData::new_disabled()
     }
+}
 
+impl ProfileData {
     const EXT_START: usize = 1;
     const ACTION_START: usize = ProfileData::EXT_START + ExtCosts::count();
     // const LENGTH: usize = PROFILE_DATA_LEN;
```
bowenwang1996, line 48: should this be removed?

bowenwang1996 (approved):

core/primitives-core/src/profile.rs
matklad: Maybe? Haven't really looked into the actual impl here)

evgenykuzyakov (approved):

PR #4006, opened 2021-02-25
fix: Fixing an error in cross_shard_tx

SkidanovAlex:
There were three issues, of which one is an actual bug.

The bug was my regression in the latest rewrite of the block sync: if
the current head is not on the canonical header chain, the block sync
would continuously try to fetch a block that build on top of the current
head, which is not what needs to be fetched. This is the main fix in
this change (changes in sync.rs)

The smaller issues fixed:
a) With the recent changes the node became faster, and 80ms was no longer
adequate for the acceptable forkfulness. Changed to 70ms.
b) There was a bug that was querying the balance at the very start of
the test from a wrong node. In the past it was working because the query
was processed before the first block was produced, and all nodes can
respond to any queries on the genesis state. But with the changes making
the node faster the first block now was produced before some queries
were processed, and the query could not be properly processed.


SkidanovAlex:
Still see some issues, closing.

PR #4007, opened 2021-02-25
refactor: typify MethodUTF8Error out of existence

matklad:
runtime actually stores method name as `String`, so we have a static
proof that utf8 validation already happened.

Wasm import/export tables are specked to use utf8, so we don't need to
worry about supporting non-utf8 even in theory.

### DIFF
--- a/runtime/near-vm-runner/src/wasmtime_runner.rs
+++ b/runtime/near-vm-runner/src/wasmtime_runner.rs
@@ -114,7 +114,7 @@ pub mod wasmtime_runner {
     pub fn run_wasmtime<'a>(
         _code_hash: &[u8],
         code: &[u8],
-        method_name: &[u8],
+        method_name: &str,
         ext: &mut dyn External,
         context: VMContext,
         wasm_config: &'a VMConfig,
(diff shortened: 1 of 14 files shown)


runtime/near-vm-runner/src/wasmtime_runner.rs
```diff
@@ -226,7 +215,7 @@ pub mod wasmtime_runner {
                 None => (
                     None,
                     Some(VMError::FunctionCallError(FunctionCallError::MethodResolveError(
-                        MethodResolveError::MethodUTF8Error,
+                        MethodResolveError::MethodNotFound,
```
matklad, line 218: :warning: this fixes a bug, we reported a wrong error before.

willemneal (approved):
LGTM! glad to see the interface match Wasm!

ailisp (approved):

bowenwang1996 (commented):
@matklad is the mapping between `String` and `&[u8]` bijective? If there is for some reason, a string that after converting to bytes cannot be converted back, then we need to do a protocol upgrade for this.

matklad:
@bowenwang1996 no, the mapping is injective, but not surjective. Every `str` is a `[u8]`, but the converse is not true. But I don't think we need a protocol change here, as it seems that we already use Strings in the protocol, and just loose this info somewhere along the way. 

In terms of code, here's the single place where we modify the code outside of contract runtime:

https://github.com/near/nearcore/pull/4007/files#diff-318756fbcf83731e0211cbc53dad981140af78c2b06dce9de383c06cbd93d6efR138

As you see, the transformation is lossless.

bowenwang1996:
> As you see, the transformation is lossless.

Yeah that is what I wanted to make sure.

bowenwang1996 (approved):

evgenykuzyakov (approved):

PR #4012, opened 2021-02-25
Add new FramedWrite with buffer limits and stats

pmnoxx:
We should limit the amount of memory each peer is using by limiting size of it's `FrameWrite's` buffer.
In addition for each `PeerActor` we will print stats to detect whenever we are not draining it.


pmnoxx:
I extracted unrelated changed into https://github.com/near/nearcore/pull/4015

neard/src/config.rs
```diff
@@ -13,6 +13,8 @@ use log::info;
 use num_rational::Rational;
 use serde::{Deserialize, Serialize};
 
+#[cfg(feature = "delay_detector")]
```
matklad, line 16: Wait, we removed `, optional = true` from Cargo.tom, so ther's no longer `delay_detector` feature I think

chain/network/src/peer.rs
```diff
@@ -255,7 +256,15 @@ impl Peer {
                 #[cfg(feature = "metric_recorder")]
                 self.peer_manager_addr.do_send(metadata.set_size(bytes.len()));
                 self.tracker.increment_sent(bytes.len() as u64);
-                self.framed.write(bytes);
+                let bytes_len = bytes.len();
+                if !self.framed.write(bytes) {
+                    error!(
+                        "{} Failed to send message {} of size {}",
+                        get_tid(),
+                        format!("routed message from {}", strum::AsStaticRef::as_static(msg)),
```
bowenwang1996, line 264: the message we send here is not necessarily routed

utils/near-performance-metrics/src/framed_write.rs
```diff
@@ -0,0 +1,369 @@
...
+);
+
+impl<T: AsyncWrite, E: From<io::Error>, K: EncoderCallBack> Clone for UnsafeWriter<T, E, K> {
+    fn clone(&self) -> Self {
+        UnsafeWriter(self.0.clone(), self.1.clone())
+    }
+}
+
+pub trait EncoderCallBack {
+    fn drained(&mut self, bytes: usize, buf_len: usize, buf_capacity: usize);
```
bowenwang1996, line 79: what is this supposed to do

utils/near-performance-metrics/src/framed_write.rs
```diff
@@ -0,0 +1,369 @@
...
+    buffer: BytesMut,
+    error: Option<E>,
+    low: usize,
+    high: usize,
+    handle: SpawnHandle,
+    task: Option<task::Waker>,
+    callback: K,
+}
+
+pub struct EmptyCallBack {}
```
bowenwang1996, line 61: why do we need this

utils/near-performance-metrics/src/stats_enabled.rs
```diff
@@ -126,11 +149,27 @@ impl ThreadStats {
...
+            }
+            self.write_buf_added = Default::default();
+            self.write_buf_drained = Default::default();
+
             let mut stat: Vec<_> = self.stat.iter().collect();
             stat.sort_by(|x, y| (*x).0.cmp(&(*y).0));
 
             for entry in stat {
-                warn!(
+                info!(
```
bowenwang1996, line 172: why

chain/jsonrpc/src/lib.rs
```diff
@@ -1038,6 +1039,7 @@ pub fn start_http(
     view_client_addr: Addr<ViewClientActor>,
 ) {
     let RpcConfig { addr, cors_allowed_origins, polling_config, limits_config } = config;
+    info!("Starting http server at {}", addr);
```
bowenwang1996, line 1042: Do we actually need this? cc @frol

chain/network/src/codec.rs
```diff
@@ -26,6 +44,22 @@ impl Encoder<Vec<u8>> for Codec {
...
+                    buf.len(),
+                    buf.capacity(),
+                );
+            }
+            if buf.capacity() >= MAX_CAPACITY as usize
+                && item.len() + 4 + buf.len() > buf.capacity()
+            {
+                error!("{} throwing away message, because buffer is full item.len(): {} buf.capacity: {}", get_tid(), item.len(), buf.capacity());
+
+                return Err(Error::new(ErrorKind::Other, "Buf max capacity exceeded"));
```
bowenwang1996, line 61: might make sense to use a prometheus metric counter here

chain/jsonrpc/src/lib.rs
pmnoxx: I moved this out to https://github.com/pulls

utils/near-performance-metrics/src/stats_enabled.rs
pmnoxx: Those messages should have been `info` instead of `warn`. I would only use `warn`/`error`. For event's that aren't supposed to happen, like taking too much time, etc.

I moved this change out to https://github.com/near/nearcore/pull/4015

utils/near-performance-metrics/src/framed_write.rs
pmnoxx: We don't strictly need `EmptyCallBack `. However, we could make a PR to `actix_util`, with new optional argument to `FrameWrite`, which is the struct used to reporting when buffer gets drained. That way we could completely remove this file, and keep using one from `actix_util`.

I'm also fine with removing `EmptyCallBack` if you prefer.

chain/network/src/peer.rs
pmnoxx: fixed

neard/src/config.rs
pmnoxx: We removed `, optional = true` from `delay_detector` package, the `delay_detector` feature is independent from the package with the same name.

We were not using `delay_detector` package inside neard before, I was getting compilation errors with `, optional = true`. Anyway, I moved this code to another PR.

utils/near-performance-metrics/src/framed_write.rs
pmnoxx: I added a new callback function, which is given as parameter to `FramedWriter`. Whenever buffer decreased in size, we get information about the number of bytes consumed, and size of Buffer, and capacity of the butter.

We also have `Encoder` trait, which get called whenever new data gets added to the buffer. So, we know size/capacity of the buffer, as well as the number of bytes added to the buffer during that time. 

With both `Encoder` and `EncoderCallBack` traits you have enough information to know the current state of the buffer, there is no need to have login inside `FramedWrite`.

chain/network/src/codec.rs
pmnoxx: done

chain/network/src/codec.rs
```diff
@@ -26,6 +45,23 @@ impl Encoder<Vec<u8>> for Codec {
...
+                stat.lock().unwrap().log_add_write_buffer(
+                    item.len() + 4,
+                    buf.len(),
+                    buf.capacity(),
+                );
+            }
+            if buf.capacity() >= MAX_CAPACITY as usize
+                && item.len() + 4 + buf.len() > buf.capacity()
+            {
+                error!("{} throwing away message, because buffer is full item.len(): {} buf.capacity: {}", get_tid(), item.len(), buf.capacity());
```
bowenwang1996, line 60:
```suggestion
                error!(target: "network", "{} throwing away message, because buffer is full item.len(): {} buf.capacity: {}", get_tid(), item.len(), buf.capacity());
```

chain/network/src/peer.rs
```diff
@@ -255,7 +256,15 @@ impl Peer {
                 #[cfg(feature = "metric_recorder")]
                 self.peer_manager_addr.do_send(metadata.set_size(bytes.len()));
                 self.tracker.increment_sent(bytes.len() as u64);
-                self.framed.write(bytes);
+                let bytes_len = bytes.len();
+                if !self.framed.write(bytes) {
+                    error!(
```
bowenwang1996, line 261:
```suggestion
                    error!(target: "network",
```

utils/near-performance-metrics/src/framed_write.rs
```diff
@@ -0,0 +1,369 @@
...
+use std::task::{Context, Poll};
+use std::{io, task};
+
+use actix::{Actor, ActorContext, ActorFuture, AsyncContext, Running, SpawnHandle};
+use bitflags::bitflags;
+use bytes::BytesMut;
+use tokio::io::{AsyncWrite, AsyncWriteExt};
+use tokio_util::codec::Encoder;
+
+// This file was copied from actix-0.11.0-beta.1/src/io.rs
```
bowenwang1996, line 15: Please copy their license and prepend it to this file. We will eventually extract all of that and put it into some `LICENSES` file on the top level, but for now let's do this in the individual files.

bowenwang1996 (commented):
Please address https://github.com/near/nearcore/pull/4012/files#r585101813

utils/near-performance-metrics/src/framed_write.rs
pmnoxx: I added it

pmnoxx:
> Please address https://github.com/near/nearcore/pull/4012/files#r585101813

done, all checks also pass

bowenwang1996 (approved):

chain/network/Cargo.toml
```diff
@@ -20,6 +20,8 @@ byteorder = "1.2"
 lazy_static = "1.4"
 tracing = "0.1.13"
 strum = { version = "0.20", features = ["derive"] }
+near-rust-allocator-proxy = "0.2.9"
+size = "0.1.2"
```
matklad, line 24: Hm, why are we using both `size` and `bytesize` crates? They seem to do essentially the same thing?

pmnoxx: I agree, we should use `bytesize` to be consistent. I would like to use `ByteInfo` global constants, the PR is submitted and will be included in next release of `bytesize` crate. Once it gets published to `crates.io`, it would be a good time to replace `size.

pmnoxx:
https://github.com/near/nearcore/pull/4029/files

PR #4013, opened 2021-02-25
fix: Fix tests unable to fail

### LINKED ISSUE #3901: After actix upgrade, panic are not always properly propagated, and tests that `should_panic` fail.

frol:
Resolves #3901.


frol:
The tests finally pass! The changes are essentially the following:

1. I updated actix to the next beta
2. I introduced `near-actix-test-utils` (in `test-utils/actix-test-utils` folder) with two helpers: `run_actix_until_stop` and `run_actix_until_panic` (the latter just does not print the stack trace)
3. Replaced all the previous `System::builder().stop_on_panic(true)` with `run_actix_until_stop` or `run_actix_until_panic` (which  stops actix System once a panic is caught in any of the actors)
4. I increased some timeouts since given the implementation of `run_actix_until_stop`, we start hitting the node earlier than we used to do it
5. I also replaced some `.unwrap()`s with proper handling since we often query the node before the chain is booted up (thus returning us an error that the latest block does not exist)

Overall, the tests were just slightly touched and now we have the CI passing and catching panic correctly.

P.S. To review, you may want to disable white-space changes in the diff

khorolets (approved):

bowenwang1996 (approved):

SkidanovAlex (approved):

PR #4014, opened 2021-02-25
fix: Fixing an error in cross_shard_tx

SkidanovAlex:
There were several issues, of which one is an actual bug.

The bug was my regression in the latest rewrite of the block sync: if
the current head is not on the canonical header chain, the block sync
would continuously try to fetch a block that build on top of the current
head, which is not what needs to be fetched. This is the main fix in
this change (changes in sync.rs)

The smaller issues fixed:
a) With the recent changes the node became faster, and 80ms was no longer
adequate for the acceptable forkfulness. Changed to 70ms. With 70 the
forkfulness sometimes was below 2.4, but reducing the timeout below 70
was sometimes resulting in nodes not being able to keep up, so I had to
decrease the min forkfulness to 2.3.
b) There was a bug that was querying the balance at the very start of
the test from a wrong node. In the past it was working because the query
was processed before the first block was produced, and all nodes can
respond to any queries on the genesis state. But with the changes making
the node faster the first block now was produced before some queries
were processed, and the query could not be properly processed.
c) The nodes were configured to report each other as non-archival,
resulting in them not having any viable peers to request older blocks.
d) `gc_stop_height` was five epochs behind, while the test sometimes
generates forks longer than that, making it impossible to apply blocks.
Changing it to return 0 in `KeyValueRuntime`.

Test plan
---------
http://nayduck.eastus.cloudapp.azure.com:3000/#/run/1202

### DIFF
--- a/chain/chain/src/test_utils.rs
+++ b/chain/chain/src/test_utils.rs
@@ -68,6 +68,7 @@ pub struct KeyValueRuntime {
     validator_groups: u64,
     num_shards: NumShards,
     epoch_length: u64,
+    no_gc: bool,
 
     // A mapping state_root => {account id => amounts}, for transactions and receipts
     state: RwLock<HashMap<StateRoot, KVState>>,
@@ -108,6 +109,24 @@ impl KeyValueRuntime {
         validator_groups: u64,
         num_shards: NumShards,
         epoch_length: u64,
+    ) -> Self {
+        Self::new_with_validators_and_no_gc(
+            store,
+            validators,
+            validator_groups,
+            num_shards,
+            epoch_length,
+            false,
+        )
+    }
+
+    pub fn new_with_validators_and_no_gc(
+        store: Arc<Store>,
+        validators: Vec<Vec<AccountId>>,
+        validator_groups: u64,
+        num_shards: NumShards,
+        epoch_length: u64,
+        no_gc: bool,
     ) -> Self {
         let tries = ShardTries::new(store.clone(), num_shards);
         let mut initial_amounts = HashMap::new();
@@ -163,6 +182,7 @@ impl KeyValueRuntime {
             hash_to_next_epoch: RwLock::new(map_with_default_hash1),
             hash_to_valset: RwLock::new(map_with_default_hash3),
             epoch_start: RwLock::new(map_with_default_hash2),
+            no_gc,
         }
     }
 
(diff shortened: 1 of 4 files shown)


chain/chain/src/test_utils.rs
```diff
@@ -917,13 +917,11 @@ impl RuntimeAdapter for KeyValueRuntime {
...
 
-    fn get_gc_stop_height(&self, block_hash: &CryptoHash) -> BlockHeight {
-        let block_height = self
-            .get_block_header(block_hash)
-            .unwrap_or_default()
-            .map(|h| h.height())
-            .unwrap_or_default();
-        block_height.saturating_sub(NUM_EPOCHS_TO_KEEP_STORE_DATA * self.epoch_length)
+    fn get_gc_stop_height(&self, _block_hash: &CryptoHash) -> BlockHeight {
+        // KeyValueRuntime tests do not test GC, thus no need to simulate the 5 epochs from the NightshadeRuntime.
```
SkidanovAlex, line 921: I'm actually not 100% this statement is true

bowenwang1996 (approved):

chain/chain/src/test_utils.rs
```diff
@@ -918,12 +938,16 @@ impl RuntimeAdapter for KeyValueRuntime {
     }
 
     fn get_gc_stop_height(&self, block_hash: &CryptoHash) -> BlockHeight {
-        let block_height = self
-            .get_block_header(block_hash)
-            .unwrap_or_default()
-            .map(|h| h.height())
-            .unwrap_or_default();
-        block_height.saturating_sub(NUM_EPOCHS_TO_KEEP_STORE_DATA * self.epoch_length)
+        if !self.no_gc {
```
matklad, line 941: Nit, but I'd :bulb: "invert if condition" here: double negation in `!no_gc` is awkward.

chain/client/src/sync.rs
```diff
@@ -400,6 +400,49 @@ impl BlockSync {
...
+            }
+
+            // Then go forward for as long as we known the next block
+            let mut ret_hash = candidate.1;
+            loop {
+                match chain.mut_store().get_next_block_hash(&ret_hash) {
+                    Ok(hash) => {
+                        let hash = hash.clone();
+                        if chain.block_exists(&hash)? {
+                            ret_hash = hash;
```
matklad, line 429:
```suggestion
                        if chain.block_exists(hash)? {
                            ret_hash = hash.clone();
```

chain/client/src/sync.rs
```diff
@@ -400,6 +400,49 @@ impl BlockSync {
...
+                        } else {
+                            break;
+                        }
+                    }
+                    Err(e) => match e.kind() {
+                        near_chain::ErrorKind::DBNotFoundErr(_) => {
+                            break;
+                        }
```
matklad, line 437:
```suggestion
                        near_chain::ErrorKind::DBNotFoundErr(_) => break,
```

chain/client/src/sync.rs
SkidanovAlex: I applied the change, but will undo it, a reference to hash keeps a reference on `chain`, and cannot be used in `chain.block_exists`.

PR #4015, opened 2021-02-25
Add a few important logs

pmnoxx:
In this change I want to add logging http server addr, I've seen a cases, where a node would not start, because addr was already in use. 

I added a `DelayDetector` to print the time it takes to read config.

In near-performance-metrics I also changed a few `warn!` to be `info!`


pmnoxx:
This PR was extracted from https://github.com/near/nearcore/pull/4012/files

chain/jsonrpc/Cargo.toml
```diff
@@ -17,6 +17,7 @@ serde_json = "1"
 validator = "0.12"
 tracing = "0.1.13"
 borsh = "0.8.1"
+log = "0.4"
```
matklad, line 20: Why do we need both `tracing` and `log`? `tracing` is a struct super-set

matklad: https://docs.rs/tracing/0.1.25/tracing/#for-log-users

bowenwang1996 (commented):
I think it is better to add `target` in the logging macro

pmnoxx:
> I think it is better to add `target` in the logging macro

done

Issue #3725, opened 2020-12-11
Memory spike


bowenwang1996:
We observed several times that for nodes on mainnet that they sometimes experience a very drastic memory spike that causes the node to run out of memory in 10-20 minutes. @pmnoxx could you run a node on mainnet with your memory profiling tools to see if there is something abnormal?

- [x] write code for detecting memory spikes
- [ ] start dedicated node running on mainnet for week or two to see if there are any spikes
- [ ] wait for results

pmnoxx:
I'll start the node today.

pmnoxx:
I've been running a few nodes during last few days looks for any memory spikes. So far I found one such spike:
```
Dec 20 11:02:40.262  INFO near_performance_metrics::actix: Slow function call run_later:2 chain/network/src/peer_manager.rs:580 took: 2505ms    
Dec 20 11:02:40.751  WARN near_memory_tracker::allocator: Thread 27632 reached new record of memory usage 3205MiB

 2: alloc::raw_vec::RawVec<T,A>::with_capacity_in
             at $HOME/.rustup/toolchains/nightly-2020-10-08-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/alloc/src/raw_vec.rs:163:9
      alloc::raw_vec::RawVec<T>::with_capacity
             at $HOME/.rustup/toolchains/nightly-2020-10-08-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/alloc/src/raw_vec.rs:93:9
      alloc::vec::Vec<T>::with_capacity
             at $HOME/.rustup/toolchains/nightly-2020-10-08-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/alloc/src/vec.rs:363:20
      alloc::slice::hack::to_vec
             at $HOME/.rustup/toolchains/nightly-2020-10-08-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/alloc/src/slice.rs:161:23
      alloc::slice::<impl [T]>::to_vec
             at $HOME/.rustup/toolchains/nightly-2020-10-08-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/alloc/src/slice.rs:397:9
      near_store::trie::state_parts::<impl near_store::trie::Trie>::combine_state_parts::{{closure}}
             at $HOME/repos/nearcore.measure-mem/core/store/src/trie/state_parts.rs:278:25
[...]

Issue #3777, opened 2021-01-04
Rethink state sync


bowenwang1996:
Currently when we do state sync, we request state parts of size at most 1MB. However, when the state size is on the order of gigabytes, it means that a node needs to request thousands of parts and each part takes some time to compute. Since we request all parts at once, this causes unnecessary increase in network traffic, especially since the requests and responses are also routed. We should maybe reconsider the entire approach and see how we make it better when the state size is very large.

bowenwang1996:
cc @mikhailOK

mikhailOK:
Potential ideas:
- Try to speed up computing state parts with a faster trie iterator
- Pre-compute state parts in the background when epoch starts (currently we compute state parts upon receiving a request)
- Manage incoming state part requests. Don't respond twice to duplicate requests, reject instantly if there are too many queued up.

Longarithm:
Tracked around #8545.

PR #4016, opened 2021-02-26
fix(view_client): throttle state requests

bowenwang1996:
@pmnoxx did some investigation and found that #3725 is indeed caused by a large number of `StateResponse` messages. While @pmnoxx is working on rewriting `FramedWriter` to address this on the network layer, here is a (temporary) fix that throttles state sync requests to reduce the number of messages sent over the network. The strategy here is quite aggressive and will likely reject more requests than necessary, but if there are sufficient number of nodes in the network regular state syncs should not be affected.

Test plan
-----------
Sanity test `test_state_request`


SkidanovAlex:
Why throttle on `(sync_hash, shard_id)`, why not just throttle all requests?
As in, say there are two shards and two active sync_hashes, is there a reason why we would throttle the same request four times more aggressively that four different requests?

SkidanovAlex:
Also, 30 seconds throttling will almost certainly break stress.py. It should either be configurable, or lower.

chain/client/src/view_client.rs
```diff
@@ -97,6 +97,8 @@ pub struct ViewClientActor {
     network_adapter: Arc<dyn NetworkAdapter>,
     pub config: ClientConfig,
     request_manager: Arc<RwLock<ViewClientRequestManager>>,
+    state_header_request_cache: Arc<Mutex<TimedCache<(CryptoHash, ShardId), ()>>>,
```
pmnoxx, line 100: I don't see how this would help. I saw a situation where we were sending to peer 11k messages of size 2mb each over period of 30m for a node with 32gb ram. So that would be less than 7m for a node with 8 gb ram.

I don't have stat's on how many of those requests are unique so this may not help at all. Even if we slow down the amount the grown of data by factor of 10x, the node will die in an hour instead of 7m, so likely we won't have much affect at all. Though by looking at rate of requests, we see 6 per seconds, so either there are a lot of duplicates, or those are requests to different peers for the same CryptoHash.

chain/client/src/view_client.rs
```diff
@@ -112,6 +114,9 @@ impl ViewClientRequestManager {
 }
 
 impl ViewClientActor {
+    /// Number of seconds between state requests. More frequent requests will be rejected.
+    const TIME_BETWEEN_STATE_REQUESTS: u64 = 30;
```
matklad, line 118:
```suggestion
    const TIME_BETWEEN_STATE_REQUESTS: Duration = Duration::from_secs(30);
```

It's better to call `as_secs` when using `TimedCache`.

I don't really understand why `TImedCache` is not using Duration already, this feels like a questionable API design, opened https://github.com/jaemk/cached/issues/74

chain/client/src/view_client.rs
bowenwang1996: > so either there are a lot of duplicates

That is very likely the case given the suboptimal state sync design (#3777). Also this protects against spams

bowenwang1996:
> As in, say there are two shards and two active sync_hashes, is there a reason why we would throttle the same request four times more aggressively that four different requests?

That seems too aggressive to me. The number of possible valid `(shard_id, sync_hash)` peer is bounded by a small number and I don't think that will cause any issues.

chain/client/src/view_client.rs
pmnoxx: Can you provide us with stats of what percentage of requests are duplicates?

bowenwang1996: I don't have any stats, but the total number of state parts should be around ~3k on testnet. cc @mikhailOK

mikhailOK: With 3k parts and 30 second throttle per part we can still get 100 requests/second, this seems like a problem

pmnoxx: On one node that ran out of memory where we send 11k StateRequests to one shard I had a print which shows time, length of message send to given peer. For example:
```

[2mFeb 24 22:07:49.761[0m [32m INFO[0m near_network::peer: send message type: StateResponse length: 886107
[2mFeb 24 22:07:49.762[0m [32m INFO[0m near_network::peer: send message type: StateResponse length: 1029446
[2mFeb 24 22:07:49.763[0m [32m INFO[0m near_network::peer: send message type: StateResponse length: 4112816
[2mFeb 24 22:07:49.765[0m [32m INFO[0m near_network::peer: send message type: StateResponse length: 1891264
[2mFeb 24 22:07:49.765[0m [32m INFO[0m near_network::peer: send message type: StateResponse length: 695405
[2mFeb 24 22:07:49.766[0m [32m INFO[0m near_network::peer: send message type: StateResponse length: 764465
[2mFeb 24 22:07:49.767[0m [32m INFO[0m near_network::peer: send message type: StateResponse length: 3754816
[2mFeb 24 22:07:49.769[0m [32m INFO[0m near_network::peer: send message type: StateResponse length: 2124892
[2mFeb 24 22:07:49.769[0m [32m INFO[0m near_network::peer: send message type: StateResponse length: 924100
[2mFeb 24 22:07:49.770[0m [32m INFO[0m near_network::peer: send message type: StateResponse length: 1905068
[2mFeb 24 22:07:49.771[0m [32m INFO[0m near_network::peer: send message type: StateResponse length: 1814682
```

Here is the total number of requests:
```bash
> cat xxx.txt | grep StateResponse | wc -l
 11464
```

If the group requests by the minute, and length, and print unique entries (that will be an upper bound)
```bash
> cat xxx.txt | grep StateResponse | cut -c 13-17,100- | sort | uniq-c | wc -l
6404
```

Therefore the solution suggested in in request would reduce the number of requests by up to 45%, so this would not have much of an effect on the issue we were trying to solve.

pmnoxx: We do see that the same chunks were requested multiple times over period of 30 minutes, here is an example for chunk of size `3661359`.
```
[2mFeb 24 21:38:28.101[0m [32m INFO[0m near_network::peer: send message type: StateResponse length: 3661359
[2mFeb 24 21:38:42.753[0m [32m INFO[0m near_network::peer: send message type: StateResponse length: 3661359
[2mFeb 24 21:39:19.962[0m [32m INFO[0m near_network::peer: send message type: StateResponse length: 3661359
[2mFeb 24 21:39:42.534[0m [32m INFO[0m near_network::peer: send message type: StateResponse length: 3661359
[2mFeb 24 21:39:57.039[0m [32m INFO[0m near_network::peer: send message type: StateResponse length: 3661359
[2mFeb 24 21:41:26.905[0m [32m INFO[0m near_network::peer: send message type: StateResponse length: 3661359
[2mFeb 24 21:41:44.142[0m [32m INFO[0m near_network::peer: send message type: StateResponse length: 3661359
[2mFeb 24 21:42:55.326[0m [32m INFO[0m near_network::peer: send message type: StateResponse length: 3661359
[2mFeb 24 21:42:55.401[0m [32m INFO[0m near_network::peer: send message type: StateResponse length: 3661359
[2mFeb 24 21:43:36.778[0m [32m INFO[0m near_network::peer: send message type: StateResponse length: 3661359
[2mFeb 24 21:43:42.491[0m [32m INFO[0m near_network::peer: send message type: StateResponse length: 3661359
[2mFeb 24 21:44:31.613[0m [32m INFO[0m near_network::peer: send message type: StateResponse length: 3661359
[2mFeb 24 21:45:35.039[0m [32m INFO[0m near_network::peer: send message type: StateResponse length: 3661359
[2mFeb 24 21:46:21.572[0m [32m INFO[0m near_network::peer: send message type: StateResponse length: 3661359
[2mFeb 24 21:47:10.639[0m [32m INFO[0m near_network::peer: send message type: StateResponse length: 3661359
[2mFeb 24 21:47:33.760[0m [32m INFO[0m near_network::peer: send message type: StateResponse length: 3661359
[2mFeb 24 21:50:14.535[0m [32m INFO[0m near_network::peer: send message type: StateResponse length: 3661359
[2mFeb 24 21:50:19.451[0m [32m INFO[0m near_network::peer: send message type: StateResponse length: 3661359
[2mFeb 24 21:50:36.836[0m [32m INFO[0m near_network::peer: send message type: StateResponse length: 3661359
[2mFeb 24 21:52:35.851[0m [32m INFO[0m near_network::peer: send message type: StateResponse length: 3661359
[2mFeb 24 21:55:17.819[0m [32m INFO[0m near_network::peer: send message type: StateResponse length: 3661359
[2mFeb 24 21:55:20.559[0m [32m INFO[0m near_network::peer: send message type: StateResponse length: 3661359
[2mFeb 24 21:55:35.185[0m [32m INFO[0m near_network::peer: send message type: StateResponse length: 3661359
[2mFeb 24 21:56:22.179[0m [32m INFO[0m near_network::peer: send message type: StateResponse length: 3661359
[2mFeb 24 21:56:27.402[0m [32m INFO[0m near_network::peer: send message type: StateResponse length: 3661359
[2mFeb 24 21:57:14.904[0m [32m INFO[0m near_network::peer: send message type: StateResponse length: 3661359
[2mFeb 24 21:57:15.551[0m [32m INFO[0m near_network::peer: send message type: StateResponse length: 3661359
[2mFeb 24 21:58:19.494[0m [32m INFO[0m near_network::peer: send message type: StateResponse length: 3661359
[2mFeb 24 21:58:29.978[0m [32m INFO[0m near_network::peer: send message type: StateResponse length: 3661359
[2mFeb 24 21:59:33.337[0m [32m INFO[0m near_network::peer: send message type: StateResponse length: 3661359
[2mFeb 24 22:00:40.557[0m [32m INFO[0m near_network::peer: send message type: StateResponse length: 3661359
[2mFeb 24 22:00:41.837[0m [32m INFO[0m near_network::peer: send message type: StateResponse length: 3661359
[2mFeb 24 22:00:42.152[0m [32m INFO[0m near_network::peer: send message type: StateResponse length: 3661359
[2mFeb 24 22:01:25.373[0m [32m INFO[0m near_network::peer: send message type: StateResponse length: 3661359
[2mFeb 24 22:02:48.739[0m [32m INFO[0m near_network::peer: send message type: StateResponse length: 3661359
[2mFeb 24 22:03:26.730[0m [32m INFO[0m near_network::peer: send message type: StateResponse length: 3661359
[2mFeb 24 22:03:38.884[0m [32m INFO[0m near_network::peer: send message type: StateResponse length: 3661359
[2mFeb 24 22:03:46.039[0m [32m INFO[0m near_network::peer: send message type: StateResponse length: 3661359
[2mFeb 24 22:04:44.783[0m [32m INFO[0m near_network::peer: send message type: StateResponse length: 3661359
[2mFeb 24 22:05:27.521[0m [32m INFO[0m near_network::peer: send message type: StateResponse length: 3661359
[2mFeb 24 22:05:28.175[0m [32m INFO[0m near_network::peer: send message type: StateResponse length: 3661359
[2mFeb 24 22:07:39.217[0m [32m INFO[0m near_network::peer: send message type: StateResponse length: 3661359
[2mFeb 24 22:07:49.780[0m [32m INFO[0m near_network::peer: send message type: StateResponse length: 3661359
```
We do see that we were trying to send the same StateResponse multiple times, usually every 30s, there were some duplicates, so in theory we could reduce the traffic in half.

pmnoxx: If we were to send each chunk to given peer at most once in period of 30m, we would reduce the number of chunks send to
```
> cat xxx.txt | grep StateResponse | cut -c 100- | sort | uniq-c | wc -l
353
```

bowenwang1996: @mikhailOK I updated the implementation to serve at most 30 requests every 30s.

SkidanovAlex:
> That seems too aggressive to me.

In practice we only have one shard today, and the sync hash during a specific epoch will be the same. Thus, today this change is identical to throttling all requests together. My suggestion is no more aggressive, but is simpler, and doesn't hide the fact that we are throttling all the requests together :)

pmnoxx:
> > That seems too aggressive to me.
> 
> In practice we only have one shard today, and the sync hash during a specific epoch will be the same. Thus, today this change is identical to throttling all requests together. My suggestion is no more aggressive, but is simpler, and doesn't hide the fact that we are throttling all the requests together :)

We could also implement deduplicating messages in the write buffer. This would be simpler and more generic.

bowenwang1996:
Close in favor of #4012

bowenwang1996:
Discussed with @mikhailOK and we decide to pursue this PR again for 2 reasons:
* Throttle inside view client allows us to avoid doing unnecessary computation for parts which are going to be dropped on the network side.
* It is more robust and can prevent memory peaks within the channels before they even get to the network buffer.

bowenwang1996:
@SkidanovAlex addressed your comments

SkidanovAlex:
@bowenwang1996 did you see this comment above?

> Also, 30 seconds throttling will almost certainly break stress.py. It should either be configurable, or lower.

bowenwang1996:
> @bowenwang1996 did you see this comment above?
> 
> >

@SkidanovAlex Yes it is addressed. I introduced `view_client_throttle_period` in config.json. Only the number of requests within this period is hardcoded. This pretty much allows us to have arbitrary control of throttling from config.json.

SkidanovAlex (approved):

PR #4018, opened 2021-02-26
refactor: move code style checks to tests from CI

matklad:
It's more convenient when checks are done during ~~`cargo check`~~ `cargo test`, as
that's what folks are running anyway.


chefsale (approved):
Interesting, wouldn't expect this is the case, but let's go with this if this makes sense. Also please add @frol as reviewer to check if this is fine.

frol (approved):
This will move the formatting check from the sanity checks pipeline to tests pipeline :thinking: 

I don't have any hard feelings, just wanted to bring it up.

PR #4023, opened 2021-03-01
feat(chain): Epoch Sync / GC Headers skeleton

Kouprin:
This change is related to #3488 and contains the following features:

- `ColHeaderHashesByHeight` DB column is added that stores map from Block Header Height to the list of available Headers (GC at chunk_tail is enabled)
- BlockHeaderV3 is updated with Epoch Manager data needed by Epoch Sync
- Network Epoch Sync requests and responses skeleton
- EpochSync class and its initialization
- `save_header_head_forcibly`, `save_block_header_no_update_tree`, `epoch_sync_init_epoch_manager` and other methods needed by Epoch Sync
- BlockInfo is updated to store its hash
- `migrate_16_to_17` migration
- Store Validator update
- `BlockMissing` error is removed
- `update_*head*` methods need Header instead of Block
- `EpochOutOfBounds` error is extended with EpochId
- tests fixes and other minor updates/fixes

This change is important for two reasons:
1. The nodes will apply migrations early and will be prepared to the following protocol upgrade.
2. Due to many files affected (58), this commit solidifies the scope of Epoch Sync / GC Headers work, helps to manage following tasks and decreases the number of merge conflicts in future.

Also tested locally on `gc_sync_after_sync.py` and `gc_after_sync.py` as those tests are most sensitive to the changes.

### DIFF
--- a/chain/chain-primitives/src/error.rs
+++ b/chain/chain-primitives/src/error.rs
@@ -8,10 +8,9 @@ use log::error;
 use near_primitives::block::BlockValidityError;
 use near_primitives::challenge::{ChunkProofs, ChunkState};
 use near_primitives::errors::{EpochError, StorageError};
-use near_primitives::hash::CryptoHash;
 use near_primitives::serialize::to_base;
 use near_primitives::sharding::{ChunkHash, ShardChunkHeader};
-use near_primitives::types::{BlockHeight, ShardId};
+use near_primitives::types::{BlockHeight, EpochId, ShardId};
 
 #[derive(Debug)]
 pub struct Error {
@@ -26,9 +25,6 @@ pub enum ErrorKind {
     /// Orphan block.
     #[fail(display = "Orphan")]
     Orphan,
-    /// Block is not available (e.g. garbage collected)
-    #[fail(display = "Block Missing (unavailable on the node): {}", _0)]
-    BlockMissing(CryptoHash),
     /// Chunk is missing.
     #[fail(display = "Chunk Missing (unavailable on the node): {:?}", _0)]
     ChunkMissing(ChunkHash),
@@ -154,7 +150,7 @@ pub enum ErrorKind {
     ValidatorError(String),
     /// Epoch out of bounds. Usually if received block is too far in the future or alternative fork.
     #[fail(display = "Epoch Out Of Bounds")]
-    EpochOutOfBounds,
+    EpochOutOfBounds(EpochId),
     /// A challenged block is on the chain that was attempted to become the head
     #[fail(display = "Challenged block on chain")]
     ChallengedBlockOnChain,
@@ -221,15 +217,13 @@ impl Error {
         match self.kind() {
             ErrorKind::Unfit(_)
             | ErrorKind::Orphan
-            | ErrorKind::BlockMissing(_)
             | ErrorKind::ChunkMissing(_)
             | ErrorKind::ChunksMissing(_)
             | ErrorKind::InvalidChunkHeight
             | ErrorKind::IOErr(_)
             | ErrorKind::Other(_)
             | ErrorKind::ValidatorError(_)
-            // TODO: can be either way?
-            | ErrorKind::EpochOutOfBounds
+            | ErrorKind::EpochOutOfBounds(_)
             | ErrorKind::ChallengedBlockOnChain
             | ErrorKind::StorageError(_)
             | ErrorKind::GCError(_)
@@ -305,7 +299,7 @@ impl std::error::Error for Error {}
 impl From<EpochError> for Error {
     fn from(error: EpochError) -> Self {
         match error {
-            EpochError::EpochOutOfBounds => ErrorKind::EpochOutOfBounds,
+            EpochError::EpochOutOfBounds(epoch_id) => ErrorKind::EpochOutOfBounds(epoch_id),
             EpochError::MissingBlock(h) => ErrorKind::DBNotFoundErr(to_base(&h)),
             err => ErrorKind::ValidatorError(err.to_string()),
         }
--- a/chain/chain/src/store.rs
+++ b/chain/chain/src/store.rs
@@ -7,7 +7,6 @@ use std::sync::Arc;
 use borsh::{BorshDeserialize, BorshSerialize};
 use cached::{Cached, SizedCache};
 use chrono::Utc;
-use tracing::debug;
 
 use near_chain_primitives::error::{Error, ErrorKind};
 use near_primitives::block::{Approval, Tip};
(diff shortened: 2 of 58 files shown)


chain/chain-primitives/src/error.rs
```diff
@@ -221,15 +217,14 @@ impl Error {
...
             ErrorKind::Unfit(_)
             | ErrorKind::Orphan
-            | ErrorKind::BlockMissing(_)
             | ErrorKind::ChunkMissing(_)
             | ErrorKind::ChunksMissing(_)
             | ErrorKind::InvalidChunkHeight
             | ErrorKind::IOErr(_)
             | ErrorKind::Other(_)
             | ErrorKind::ValidatorError(_)
             // TODO: can be either way?
```
bowenwang1996, line 226: can we remove this TODO

chain/chain/src/store.rs
```diff
@@ -789,6 +760,17 @@ impl ChainStoreAccess for ChainStore {
...
 
+    fn get_all_header_hashes_by_height(
+        &mut self,
+        height: BlockHeight,
+    ) -> Result<HashSet<CryptoHash>, Error> {
+        match self.store.get_ser(ColHeaderHashesByHeight, &index_to_bytes(height)) {
+            Ok(Some(hash_set)) => Ok(hash_set),
+            Ok(None) => Ok(HashSet::new()),
+            Err(e) => Err(e.into()),
+        }
```
bowenwang1996, line 771:
```suggestion
        Ok(self.store.get_ser(ColHeaderHashesByHeight, &index_to_bytes(height))?.unwrap_or_default())
```

chain/chain/src/store.rs
```diff
@@ -1750,6 +1739,19 @@ impl<'a> ChainStoreUpdate<'a> {
         }
     }
 
+    /// Save header head in Epoch Sync
+    /// Checking validity of header head is delegated to Epoch Sync methods
+    pub fn save_header_head_forcibly(&mut self, t: &Tip) -> Result<(), Error> {
```
bowenwang1996, line 1744: Maybe force_save_header_head is a better name

chain/chain/src/store.rs
```diff
@@ -1750,6 +1739,19 @@ impl<'a> ChainStoreUpdate<'a> {
...
+        self.try_save_latest_known(t.height)?;
+
+        self.chain_store_cache_update.height_to_hashes.insert(t.height, Some(t.last_block_hash));
```
bowenwang1996, line 1747: It seems that `height_to_hashes` is used to update `ColBlockHeight`, which stores blocks, not block headers, by height. Therefore I wonder whether this line here breaks some invariant since now we potentially don't have the corresponding block in storage.

chain/chunks/src/lib.rs
```diff
@@ -1101,8 +1101,7 @@ impl ShardsManager {
             Ok(true) => (),
             Err(chain_error) => {
                 return match chain_error.kind() {
-                    near_chain::ErrorKind::BlockMissing(_)
-                    | near_chain::ErrorKind::DBNotFoundErr(_) => {
+                    near_chain::ErrorKind::DBNotFoundErr(_) => {
```
bowenwang1996, line 1104: why?

chain/client-primitives/src/types.rs
```diff
@@ -123,6 +123,8 @@ pub enum SyncStatus {
     AwaitingPeers,
     /// Not syncing / Done syncing.
     NoSync,
+    /// Syncing using light-client headers to a recent epoch
+    EpochSync { epoch_ord: u64 },
```
bowenwang1996, line 127: why do we use epoch ordinal instead of epoch id?

core/primitives/src/merkle.rs
```diff
@@ -131,7 +131,7 @@ pub fn compute_root_from_path_and_item<T: BorshSerialize>(
 /// The root can be computed by folding `path` from right but is not explicitly
 /// maintained to save space.
 /// The size of the object is O(log(n)) where n is the number of leaves in the tree, i.e, `size`.
-#[derive(Default, Clone, BorshSerialize, BorshDeserialize)]
+#[derive(Default, Clone, Serialize, BorshSerialize, BorshDeserialize, Eq, PartialEq, Debug)]
```
bowenwang1996, line 134: why do you need `Serialize`?

core/primitives/src/syncing.rs
```diff
@@ -188,6 +191,34 @@ impl ShardStateSyncResponseV1 {
     }
 }
 
+#[derive(Serialize, BorshSerialize, BorshDeserialize, Eq, PartialEq, Debug, Clone)]
```
bowenwang1996, line 194: Why do we need `Serialize` here?

core/primitives/src/syncing.rs
```diff
@@ -188,6 +191,34 @@ impl ShardStateSyncResponseV1 {
...
+pub struct EpochSyncFinalizationResponse {
+    pub cur_epoch_header: BlockHeader,
+    pub prev_epoch_headers: Vec<BlockHeader>,
```
bowenwang1996, line 197: is this bounded in size?

core/store/src/migrations.rs
```diff
@@ -512,14 +514,103 @@ pub fn migrate_14_to_15(path: &String) {
...
+    .unwrap();
+
+    // Add ColHeaderHashesByHeight
+    let chunk_tail =
+        store.get_ser::<BlockHeight>(ColBlockMisc, CHUNK_TAIL_KEY).unwrap().unwrap_or(0);
+    let mut m = HashMap::new();
+
+    map_col(&store, DBCol::ColBlockHeader, |header: BlockHeader| {
+        let height = header.height();
+        // TODO MOO check the sign carefully
```
bowenwang1996, line 584: is this checked?

core/store/src/migrations.rs
```diff
@@ -512,14 +514,103 @@ pub fn migrate_14_to_15(path: &String) {
...
+            latest_protocol_version: old_block_info.latest_protocol_version,
+            slashed: old_block_info.slashed,
+            total_supply: old_block_info.total_supply,
+        }
+    })
+    .unwrap();
+
+    // Add ColHeaderHashesByHeight
+    let chunk_tail =
+        store.get_ser::<BlockHeight>(ColBlockMisc, CHUNK_TAIL_KEY).unwrap().unwrap_or(0);
```
bowenwang1996, line 579: shouldn't it be unwrap_or(genesis_height)?

core/primitives/src/merkle.rs
Kouprin: Removed.

core/primitives/src/syncing.rs
Kouprin: Removed.

core/primitives/src/syncing.rs
Kouprin: Cannot be longer than the length of the Epoch.

core/store/src/migrations.rs
Kouprin: This is a corner case for very small tests.

BTW, how is it possible to store Headers with Height lower than genesis_height?

core/store/src/migrations.rs
Kouprin: I put this comment for reviewers.

chain/chain-primitives/src/error.rs
bowenwang1996: It doesn't look like there is any action item here -- we cannot attribute epoch out of bound to malicious activities.

core/store/src/migrations.rs
bowenwang1996: I see. I think I misunderstood how it works

core/store/src/migrations.rs
```diff
@@ -512,14 +514,103 @@ pub fn migrate_14_to_15(path: &String) {
...
+    let mut m = HashMap::new();
```
bowenwang1996, line 580: Can we use a more informative name than `m`?

core/store/src/migrations.rs
bowenwang1996: looks good to me

bowenwang1996 (approved):

core/store/src/migrations.rs
Kouprin: Renamed to `heights_to_hashes`.

chain/chain-primitives/src/error.rs
Kouprin: Done.

Kouprin:
Squashed the commits into one due to merge conflicts with 272b3143484e1ade33fe528a515d9a254f8b2245.

frol (approved):

chain/client/src/client.rs
```diff
@@ -111,6 +116,22 @@ impl Client {
...
+            runtime_adapter
+                .get_epoch_block_producers_ordered(
+                    &genesis_block.header().epoch_id(),
+                    &genesis_block.hash(),
+                )?
+                .iter()
+                .map(|x| x.0.clone().into())
+                .collect(),
+            Duration::from_millis(EPOCH_SYNC_REQUEST_TIMEOUT_MS),
+            Duration::from_millis(EPOCH_SYNC_PEER_TIMEOUT_MS),
```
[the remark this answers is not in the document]
Kouprin: Sure. I'm changing them to Duration type.

PR #4028, opened 2021-03-03
feat(primitives): Add versioning to Account

EgorKulikov:
Just like we have different versions of Block or BlockHeader types we'll now have same for accounts. As accounts are saved in state we have to be able to deserialize old Account structure as well

### Test plan
Tests added/updated to check serialization/deserialization, including from older version. Also for this to work new version should always serialize into longer sequence of bytes, such check added as well


core/primitives-core/src/account.rs
```diff
@@ -20,6 +23,173 @@ pub struct Account {
...
+    pub fn storage_usage(&self) -> StorageUsage {
+        match self {
+            Account::AccountV1(acc) => acc.storage_usage,
+        }
+    }
+
+    #[cfg(not(feature = "protocol_feature_add_account_versions"))]
+    pub fn set_amount(&mut self, amount: Balance) {
+        self.amount = amount;
+    }
```
matklad, line 119: Given that we introduce functions here anyway, perhaps we can introduce 

```
pub fn inc_amount(&mut self, by: Balance) -> Result<(), OverflowError> {
    self.amount = self.amount.checked_add(by).ok_or(())?;
    Ok());
}
pub fn dec_amount(&mut self, by: Balance) -> Result<(), OverflowError>
``` 

instead? That way, "not forget overflow checks" becomes less of an issue. 

They do seem to help in the common case:

```diff
diff --git a/core/primitives-core/src/account.rs b/core/primitives-core/src/account.rs
index 2d5442fa..909fdd06 100644
--- a/core/primitives-core/src/account.rs
+++ b/core/primitives-core/src/account.rs
@@ -113,9 +113,15 @@ impl Account {
         }
     }
 
+    // #[cfg(not(feature = "protocol_feature_add_account_versions"))]
+    // pub fn set_amount(&mut self, amount: Balance) {
+    //     self.amount = amount;
+    // }
+
     #[cfg(not(feature = "protocol_feature_add_account_versions"))]
-    pub fn set_amount(&mut self, amount: Balance) {
-        self.amount = amount;
+    pub fn inc_amount(&mut self, by: Balance) -> Result<(), ()> {
+        self.amount = self.amount.checked_add(by).ok_or(())?;
+        Ok(())
     }
 
     #[cfg(feature = "protocol_feature_add_account_versions")]
diff --git a/runtime/runtime/src/actions.rs b/runtime/runtime/src/actions.rs
index 8338a56a..3d77b090 100644
--- a/runtime/runtime/src/actions.rs
+++ b/runtime/runtime/src/actions.rs
@@ -329,9 +329,9 @@ pub(crate) fn action_transfer(
     account: &mut Account,
     transfer: &TransferAction,
 ) -> Result<(), StorageError> {
-    account.set_amount(account.amount().checked_add(transfer.deposit).ok_or_else(|| {
+    account.inc_amount(transfer.deposit).map_err(|()| {
         StorageError::StorageInconsistentState("Account balance integer overflow".to_string())
-    })?);
+    })?;
     Ok(())
 }
 
diff --git a/runtime/runtime/src/lib.rs b/runtime/runtime/src/lib.rs
index 1e466789..c06b9bca 100644
--- a/runtime/runtime/src/lib.rs
+++ b/runtime/runtime/src/lib.rs
@@ -7,6 +7,7 @@ use log::debug;
 pub use near_crypto;
 use near_crypto::PublicKey;
 pub use near_primitives;
+use near_primitives::errors::IntegerOverflowError;
 use near_primitives::runtime::get_insufficient_storage_stake;
 use near_primitives::{
     account::{AccessKey, Account},
@@ -604,7 +605,7 @@ impl Runtime {
                 // account holder. If the account doesn't exist by the end of the execution, the
                 // validators receive the full reward.
                 tx_burnt_amount -= receiver_reward;
-                account.set_amount(safe_add_balance(account.amount(), receiver_reward)?);
+                account.inc_amount(receiver_reward).map_err(|()| IntegerOverflowError)?;
                 set_account(state_update, account_id.clone(), account);
                 state_update.commit(StateChangeCause::ActionReceiptGasReward {
                     receipt_hash: receipt.get_hash(),
@@ -984,12 +985,9 @@ impl Runtime {
                         .checked_sub(return_stake)
                         .ok_or_else(|| RuntimeError::UnexpectedIntegerOverflow)?,
                 );
-                account.set_amount(
-                    account
-                        .amount()
-                        .checked_add(return_stake)
-                        .ok_or_else(|| RuntimeError::UnexpectedIntegerOverflow)?,
-                );
+                account
+                    .inc_amount(return_stake)
+                    .map_err(|()| RuntimeError::UnexpectedIntegerOverflow);
 
                 set_account(state_update, account_id.clone(), &account);
             } else if *max_of_stakes > 0 {
@@ -1050,12 +1048,9 @@ impl Runtime {
                             account_id
                         ))
                     })?;
-                account.set_amount(
-                    account
-                        .amount()
-                        .checked_add(treasury_reward)
-                        .ok_or_else(|| RuntimeError::UnexpectedIntegerOverflow)?,
-                );
+                account
+                    .inc_amount(treasury_reward)
+                    .map_err(|()| RuntimeError::UnexpectedIntegerOverflow)?;
                 set_account(state_update, account_id.clone(), &account);
             }
         }
```

Though, there seems to be the case where we genuinely need `set` sadly. 

But this refactor is probably not related to this PR and can be done separatly.

core/primitives-core/src/account.rs
```diff
@@ -20,6 +23,173 @@ pub struct Account {
...
+    pub fn amount(&self) -> Balance {
+        self.amount
+    }
+
+    #[cfg(feature = "protocol_feature_add_account_versions")]
+    pub fn amount(&self) -> Balance {
+        match self {
+            Account::AccountV1(acc) => acc.amount,
+        }
+    }
```
matklad, line 78: Some Rust perf trivia: 

Rust compilation model is close enough to C++: in a compiled library, generic functions are present in the "source form" (a-la C++ templates in headers) but non-generic functions are compiled down to machine code. 

The latter means that non-generic functions can't be inlined across the crate. So, a `pub` getter without `#[inline]` annotation can really drag the performance down. Inline helps, because it instructs the compiler to include "source" of the function into the compiled library, thus making it eligible for compilation *at all*.

So, as a rule-of-thumb, small public functions should be `#[inline]`. Now, this doesn't entirely apply to us, as we compile with `lto=true` and `codegen-units=1`, which makes *all* the source code available to compiler (sort-of amalgamation build). However, this makes our compile times in release mode *really* slow, as it kills any parallelism.

ailisp: we often comment out `lto=true` when we had to debug a release build, it would be nice to have these inlines so that without lto=true version has better performance

ailisp (approved):

core/primitives-core/src/account.rs
```diff
@@ -20,6 +23,173 @@ pub struct Account {
...
+}
+
+#[cfg(feature = "protocol_feature_add_account_versions")]
+impl BorshDeserialize for Account {
+    fn deserialize(buf: &mut &[u8]) -> Result<Self, Error> {
+        if buf.len() == std::mem::size_of::<AccountV1>() {
+            // This should only ever happen if we have pre-transition account serialized in state
+            // See test_account_size
+            Ok(Account::AccountV1(<AccountV1 as BorshDeserialize>::deserialize(buf)?))
```
bowenwang1996, line 179: is `as BorshDeserialize` necessary?

core/primitives-core/src/account.rs
```diff
@@ -20,6 +23,173 @@ pub struct Account {
...
+        } else {
+            #[derive(BorshDeserialize)]
+            enum DeserializableAccount {
```
bowenwang1996, line 182: this feels somewhat suboptimal to me. Essentially every time `Account` changes, this struct needs to be changed correspondingly. @matklad do you see a better way to do this?

bowenwang1996 (commented):
Can we add a test to make sure the the size of `Account` cannot be 80 bytes?

core/primitives-core/src/account.rs
EgorKulikov: Yes, as there is method with same signature in (serde) Serialize

core/primitives-core/src/account.rs
EgorKulikov: I actually discussed this with @matklad beforehand. Alternative is for Account to be an additional one-member tuple around enum, and I do think it is more hassle

EgorKulikov:
@bowenwang1996 I think test_account_size and test_account_deserialization cover that

core/primitives-core/src/account.rs
EgorKulikov: Done

core/primitives-core/src/account.rs
matklad: No, I don't see a better way to do this(doesn't mean there isn't one), and this generally seems OK to me. This seems to be inherent complexity: there are bytes in some specific format in the persistent storage, and there's a domain object which, due to backwards compatibility concerns, does not precisely match the "on-disk" format. Something somewhere needs to handle this discrepancy. serde's derive only really works out of the box if you don't have backwards comparability concerns. 

What does feel suboptimal to me is that *the rest* of the code needs to care about account versions now. Let me try to explain this. I am not sure if this is a problem or a feature, as, to know that, I need to understand the code in depth :-)

`Account` is a public enum with versions, so any code *could* care if account is `V1` or `V2`. Naively, I'd expect that we want to enforce that most of the code is oblivious to the version of an account. But, naturally, we must be backwards compatible with the data already existing in the world, parts of which were created before `V2` existed. The way I usually approach this is by making `Account` a *struct* which contains only the latest and greatest version of all the fields, and having a separate `AccountRepr` solely for purposes of deserializaion. 

To give a concrere example, let's say we had had `struct Account { name: String }` and then decided to evolve that to `struct Account { first_name:String, last_name: String }`. We could start with this:

```rust
pub struct Account {
    pub name: String,
}

impl Account {
    pub fn serialize(self) -> Vec<u8> {
        AccountRepr::from(self).serialize()
    }
    pub fn deserialize(data: Vec<u8>) -> Result<Account> {
        AccountRepr::deserialize(data).map(Account::from)
    }
}

// NB: not pub
#[derive(Serialize, Deserialize)]
enum AccountRepr {
    V1 { name: String },
}

impl From<AccountRepr> for Account {
    fn from(repr: AccountRepr) -> Self {
        match repr {
            AccountRepr::V1 { name } => Account { name },
        }
    }
}

impl From<Account> for AccountRepr {
    fn from(account: Account) -> Self {
        AccountRepr::V1 { name: account.name }
    }
}
```

and then, once we realize we neen first/last name, we change the code to

```rust
// V2

pub struct Account {
    pub first_name: String,
    pub last_name: String,
}

impl Account {
    pub fn serialize(self) -> Vec<u8> {
        AccountRepr::from(self).serialize()
    }
    pub fn deserialize(data: Vec<u8>) -> Result<Account> {
        AccountRepr::deserialize(data).map(Account::from)
    }
}

// NB: not pub
#[derive(Serialize, Deserialize)]
enum AccountRepr {
    V1 { name: String },
    V2 { first_name: String, last_name: String },
}

impl From<AccountRepr> for Account {
    fn from(repr: AccountRepr) -> Self {
        match repr {
            AccountRepr::V2 { first_name, last_name } => Account { first_name, last_name },
            AccountRepr::V1 { name } => {
                // The tricky bit: whats' the actual rule for staying backwards
                // compatible?
                //
                // Let's set the last name to an empty string, bug, eg,
                // splitting on `" "` might or might not be a valid ruler
                Account { first_name: name, last_name: String::new() }
            }
        }
    }
}

impl From<Account> for AccountRepr {
    fn from(account: Account) -> Self {
        AccountRepr::V2 { first_name: account.first_name, last_name: account.last_name }
    }
}
```

That way, there's single place in code that cares about existence of two different versions of accounts.


This scheme might, or might not work for our use-case. To work, there needs to be an injective mapping from old accounts to new ones. If there isn't one, and there's genuinely some business logic that needs to work differently for two acccounts types, we might need to add something like:

```rust
pub struct Account {
    pub first_name: String,
    pub last_name: String,
    pub creation_version: AccountCreationVersion
}

pub enum AccountCreationVersion {
    V1,
    V2
}
```

Making the `Account` itself an `enum` I think might be benefitial only if we explicitelly want all the logic to be different for different versions, so that each bit code is forced to `match` and than deal with v1 and v2 separatelly. 

Based on my priors of how the world usually works, I think there's a high probability that we are in case one (injective mapping) or case two (version flag is enough).

But, again, I know almost nothing about how the thing is actually used, so this might not be applicable.

EgorKulikov: While I see merits of what you propose I think we need to be consistent as to which approach we use

bowenwang1996: @matklad I don't think what you proposed works well for our use cases. For example, the change that @EgorKulikov does in this PR does not actually change the account structure (i.e, the new version has exactly the same structure) but rather changes the logic of how a field is updated. This means that while we can go with what you proposed, it does not save us much because we still need to do things differently based on protocol versions, which means that it is better to introduce getters and setters, as done in this PR.

bowenwang1996 (approved):

core/primitives-core/src/account.rs
matklad: SGTM, def do not want to block on this. Though, I must the fact that the `enum` is public does make me uneasy. I'll take some time some time to read on the similar pattern in Block, to be more sure if this is essential or accidental complexity.

frol (approved):

PR #4035, opened 2021-03-04
chore: update each internal dep with a version number

willemneal:


matklad (approved):

runtime/runtime/Cargo.toml
```diff
@@ -19,14 +23,14 @@ ethereum-types = "0.11.0"
...
-near-vm-errors = { path = "../../runtime/near-vm-errors" }
+near-crypto = { path = "../../core/crypto", version = "0.1.0" }
+near-primitives = { path = "../../core/primitives", version = "0.1.0"}
+near-store = { path = "../../core/store", version = "2.2.0" }
+near-metrics = { path = "../../core/metrics", version = "0.1.0" }
+near-runtime-utils = { path = "../../runtime/near-runtime-utils" , version = "3.0.0" }
+near-vm-logic = { path = "../../runtime/near-vm-logic", version = "3.0.0" }
+near-vm-runner = { path = "../../runtime/near-vm-runner", version = "3.0.0" }
+near-vm-errors = { path = "../../runtime/near-vm-errors", version = "3.0.0" }
 near-evm-runner = { path = "../../runtime/near-evm-runner", optional = true }
```
frol, line 34: It seems that it needs to be updated with the version as well.

frol (approved):

runtime/runtime/Cargo.toml
willemneal: I was going to ask about this, but I was hoping to publish the runtime _without_ `near-evm-runner` because it has github deps.

@frol @matklad is this possible?   If not, then I'll have to publish the two github deps it uses.

matklad: I'd just comment it out before publishing )

willemneal: LPT! Thanks!

bowenwang1996 (approved):

evgenykuzyakov:
Have to rename `node-runtime` to `near-runtime`, because `node-runtime` was published 2 days ago by someone else and we need to publish this crate.

ailisp (approved):

bowenwang1996:
@willemneal @evgenykuzyakov what is the status of this PR? Do we still need it?

matklad:
cc https://github.com/near/nearcore/issues/4379, where we discuss what we actually need to publish.

willemneal:
Yeah this is stale.

PR #4038, opened 2021-03-05
feat: improve visibility for runtime performance

matklad:
tracing::span records how long it took to run a block of code and prints
this into into log.

See

https://github.com/near/nearcore/blob/ff01c544ce2675fadbf0284af52b356a383b27a8/docs/architecture.md#runtimenear-vm-runner

for the info on the tracing infra

Test Plan
----------

Only active with debug logging level enabled, doesn't need additional testing.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3818,6 +3818,7 @@ dependencies = [
  "tempfile",
  "testlib",
  "thiserror",
+ "tracing",
 ]
 
 [[package]]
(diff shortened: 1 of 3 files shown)


bowenwang1996 (commented):
Will `info_span` print the log on the INFO level? That is probably not what we want

bowenwang1996 (approved):

olonho (approved):

bowenwang1996:
@matklad please resolve the merge conflict and fix CI :)

matklad:
Rebased, applying auto merge as I don't think anything changed here. 

cc @near/nodex (@bowenwang1996, should we create this team alias for GH pings?), I think we might want more of this throughout the stack!

matklad:
The issue for "have more of this" is https://github.com/near/nearcore/issues/4040

PR #4041, opened 2021-03-05
refactor: reduce code duplication

matklad:
After a series of previous refactors, we can finally consolidate all
entry points. We still need two functions: `run` is a public API, and it
picks the appropriate VM automatically. `run_vm` is an internal API
which allows us to check that all supported VMs work.


ailisp (approved):
:+1

frol (approved):

PR #4048, opened 2021-03-05
fix(test): BlockHeaderV3 in pytest schema

### LINKED ISSUE #4047: proxy_simple.py fails
`pytest sanity/proxy_simple.py --features nightly_protocol --features nightly_protocol_features`

mikhailOK:
Also reorder fields a bit.
Fixes #4047

Test plan
---------
proxy_simple.py
repro_2916.py


Kouprin (approved):

frol (approved):
We don’t have running nodes with v3, right? I mean, we won’t split the network due to the old and the new fields order since there are no nodes running this code yet

Kouprin:
@frol it requires enabling `protocol_feature_block_header_v3`

PR #4052, opened 2021-03-05
fix(test): validator_switch.py

### LINKED ISSUE #4051: validator_switch.py fails
Started failing after #3960 
http://nayduck.eastus.cloudapp.azure.com:3000/#/test/102172
```
  File "tests/sanity/validator_switch.py", line 43, in <module>
    validator_info['result']
KeyError: 'result'
```

mikhailOK:
Fixes #4051

Test plan
---------
validator_switch.py

### DIFF
--- a/pytest/tests/sanity/state_sync1.py
+++ b/pytest/tests/sanity/state_sync1.py
@@ -68,9 +68,7 @@ while cur_height <= EPOCH_LENGTH:
 if not synced:
     assert False, "Nodes are not synced"
 
-status = nodes[0].get_status()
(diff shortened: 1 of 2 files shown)


bowenwang1996 (commented):
Can you explain why this fixes the test?

mikhailOK:
#3960 changed the API for epoch query: querying the latest block hash is no longer valid, because now the block hash must be the last block in an epoch, and it must have a block after it. With the new API these tests get the old behavior by passing `latest`.

bowenwang1996 (approved):
Thanks

Issue #3996, opened 2021-02-24
Replace all usage of `include_bytes` with `lazy_static_include_bytes`


evgenykuzyakov:
`include_bytes` is slowing compilation time and require recompilation when a dependency is touched

It'll save ton of $$$ on CI

rishank-shah:
Hi. I am new to contributions. I would like to give this issue a try if it's okay with you.

KaiserKarel:
Mind if I pick this up?

bowenwang1996:
@KaiserKarel please feel free to take it

matklad:
Note that lazy_static_include is not *exactly* what we need for tests. It has this "if release, just `include!`" logic, which isn't necessary and slows down compilation in release (when benchmarking). What we need is just "read file at runtime", and this doesn't need to be a macro. There's a once_cell recipe for that:

https://github.com/matklad/once_cell/blob/master/src/lib.rs#L167-L200

If we have some common dependency for our tests, we can put it there. 

But the actual diff is minor here, lazy_static_inculude_bytes works ok, it just does a couple of unnecessary hops.

bowenwang1996:
Fixed in #4053

PR #4053, opened 2021-03-06
feat: use lazy_static_include

KaiserKarel:
In debug mode, using `lazy_static_include!` saves on compilation time versus `include_bytes!`, as mentioned in #3996. I have refrained from changing  `include_bytes` in benches, as they are compiled in release mode anyway,

Test plan
---------
Current tests already cover this change. However we can no longer distribute debug mode binaries (I am not sure if this is currently done, but I highly doubt it).


KaiserKarel:
`lazy_static_include` not checking if paths exist during tests/compilations did bite me in the butt a bit. At least I got familiar with the test framework. Let's see if the builds work now, else I will have a look next weekend.

KaiserKarel:
I'm a bit too unfamiliar with what db-migration is testing, but I am suspecting that the build is spinning up a node in debug compilation?

bowenwang1996 (approved):

test-utils/testlib/Cargo.toml
```diff
@@ -16,6 +16,7 @@ rand = "0.7"
 serde_json = "1"
 futures = "0.3"
 lazy_static = "1.4"
```
bowenwang1996, line 18: is `lazy_static` still needed after this change?

bowenwang1996:
> I'm a bit too unfamiliar with what db-migration is testing, but I am suspecting that the build is spinning up a node in debug compilation?

Can you try rebasing your change on top of master?

test-utils/testlib/Cargo.toml
KaiserKarel: Yeah, it's used in some places for test globals:

```
lazy_static! {
    static ref HEAVY_TESTS_LOCK: Mutex<()> = Mutex::new(());
}
```

ailisp (approved):

bowenwang1996 (approved):

PR #4055, opened 2021-03-06
update_migration_17_18

### LINKED ISSUE #4054: DB migration 17 -> 18 is too disruptive
17 -> 18 migration requires going through all of the headers in storage. Since currently we do not garbage collect headers, on mainnet for example there are more than 20 million headers and this migration is expected to take hours to finish. In addition, every node has to apply this migration and it may cause a major disruption to the network. We should consider alternative options, such as populating `ColHeaderHashesByHeight` lazily or not introducing this column until we garbage collect headers.

Kouprin:
Fixes #4054.

Tested on `gc_sync_after_sync.py` and `state_sync_late.py` tests.


bowenwang1996 (commented):
@Kouprin it appears to me that this would still loop over all key values in `ColBlockHeader`. Did I miss something?

Kouprin:
@bowenwang1996 yes. I do think reading 20 mln Headers shouldn't take up for several hours. Please check my comment https://github.com/near/nearcore/issues/4054#issuecomment-792055027.

frol (approved):
The code looks good to me. As to the approach, I let the chain team decide.

bowenwang1996:
@Kouprin did you have a chance to test it with mainnet data?

frol:
Off-topic: @bowenwang1996 @chefsale can we add CI jobs that try booting up a node from testnet and mainnet backups (RPC and archival) with some timeout? We had quite a few issues with `config.json` parsing, sync, migrations, etc. Well, I think what I just described is designed to be the canary build. I wonder if we can automate canary builds to be manually triggered for PRs.

bowenwang1996 (approved):

Issue #2816, opened 2020-06-09
Prevent rocksdb corruption on running out of disk


ilblackdragon:
Currently even though nearcore handles storage carefully and wouldnt commit only when everything succeeded. 
 Rocksdb can fail though to commit when it the node runs out of space and do that in a way that's super hard to recover (Wal file being behind the sst for example).

My suggestion we check every few seconds how much available space is there and panic when that go under 10mb for example.

KaiserKarel:
I could have a look at this as well, if this is still a relevant issue.

I suppose we'd want the following? 

 - [ ] Start a task during initialization to poll rocksdb for available storage.
 - [ ] Use some form of signalling to shutdown the server.
 
Alternatively, would it not be better to simply check if Rocksdb has enough space before committing the transaction, and else aborting gracefully?

bowenwang1996:
> Alternatively, would it not be better to simply check if Rocksdb has enough space before committing the transaction, and else aborting gracefully?

That sounds good to me

KaiserKarel:
So that would basically involve adding a check to the total free memory in the data directory, on each transaction commit. Are we concerned with the performance overhead? I could also add the check every nth commit, for that it would be nice to know the median and average transaction size (on disk).

I think doing some heuristic to minimize the number of checks might be a bit premature, so I'll just check every single time first, and see what the impact is.

bowenwang1996:
@KaiserKarel as suggested in the description of the issue, you can check it once in a while and panic if it goes under some threshold

KaiserKarel:
Gotcha. :+1:

KaiserKarel:
I've got an implementation working, but still deciding on how to create an appropriate test, as checking disk failure is a bit difficult to set up. Perhaps adding a sanity check where I run the containerized node with `--storage-opt size=0G`, and checking for exit 1 would suffice?

bowenwang1996:
> I was a bit naive in thinking it would be a simple setting. Apparently --storage-opt size requires a different overlay and filesystem, so I am not sure if we can run this in CI. Alternatively, I could create a loop device, mount that and use it as data directory.

@chefsale can we potentially create a special test in CI for such scenarios?

mikhailOK:
Several questions for how this feature is supposed to work:
1. Can we make it so that partial writes/inconsistent storage are never possible, or a heuristic is the best option? Does rocksdb have any solution built in?
[...]

PR #4058, opened 2021-03-08
(WIP) feat: check available memory in store/rocksdb

KaiserKarel:
RocksDB may become unrecoverable if the available disk memory is exhausted. This PR adds a check every `interval` writes to see if we still have enough space left.

Since this adds some configuration options, I have also added a builder to RocksDB for setting these. Fixes https://github.com/near/nearcore/issues/2816

Test plan
---------

Still figuring this out with @chefsale and @bowenwang1996. Will have to configure some low-disk situation in the CI. A script such as:

```
dd if=/dev/zero of=low_storage.img bs=5M count=1
du -sh low_storage.img
sudo losetup -fP low_storage.img
losetup -a
mkfs.ext4 low_storage.img
mkdir loopfs
sudo mount low_storage.img loopfs
sudo chmod -R a+rw loopfs
cargo run
```
should do the trick too


core/store/src/db.rs
```diff
@@ -550,35 +680,60 @@ impl RocksDB {
...
+        if self.check_free_space_interval <= counter {
+            return Ok(());
         }
-        let cfs =
-            cf_names.iter().map(|n| db.cf_handle(n).unwrap() as *const ColumnFamily).collect();
-        Ok(Self { db, cfs, _pin: PhantomPinned })
+        self.check_free_space_counter.swap(0, Ordering::Relaxed);
+
+        let available = available_space(self.db.path())?;
+        if available.as_u64() < 16 * self.free_space_threshold.as_u64() {
```
bowenwang1996, line 701: Why 16?

KaiserKarel:
I'll try to find the cause of these failing tests this sunday. Seems due to rocksdb's `path` method returning an error.

core/store/src/db.rs
```diff
@@ -380,6 +509,7 @@ impl Database for RocksDB {
     }
 
     fn write(&self, transaction: DBTransaction) -> Result<(), DBError> {
+        self.pre_write_check().unwrap();
```
ailisp, line 512: Since we can catch disk low, we can doing a graceful shutdown of node here, instead of panic?

core/store/src/db.rs
```diff
@@ -550,35 +680,62 @@ impl RocksDB {
...
+        self.check_free_space_counter.swap(0, Ordering::Relaxed);
+
+        let available = available_space(self.db.path())?;
+
+        if available < 16_u64 * self.free_space_threshold {
+            warn!("remaining disk space is running low ({} left)", available);
+        }
+
+        if available < self.free_space_threshold {
+            Err(PreWriteCheckErr::LowMemory(available))
```
ailisp, line 707: I thin this should be named LowDiskSpace. Low memory -> Low RAM

ailisp (approved):

core/store/src/db.rs
```diff
@@ -550,35 +680,62 @@ impl RocksDB {
...
+            PreWriteCheckErr::LowMemory(available) => {
+                write!(f, "Low memory: {} available", available)
+            }
+        }
+    }
+}
+
+impl From<std::io::Error> for PreWriteCheckErr {
+    fn from(err: std::io::Error) -> Self {
+        PreWriteCheckErr::IO(err)
```
frol, line 738: BTW. we use `thiserror` to reduce such boilerplate.

frol (approved):
I second @ailisp question about a graceful shutdown. @bowenwang1996 do we have an infrastructure to do that? I think panic here might not really stop the node since it might only crash an actor, and we should stop actix, though core/store is definitely not the right place to do that.

Overall, this change looks good to me.

bowenwang1996:
> do we have an infrastructure to do that

I don't think so

core/store/src/db.rs
KaiserKarel: Agree!

core/store/src/db.rs
KaiserKarel: Do we have a way to initiate shutdown from there?

PR #4060, opened 2021-03-09
feat: time wasmer deserialization costs

matklad:
As that's where we spend the bulk of our time


bowenwang1996 (commented):
@matklad I thought we agreed to use `debug_span!` instead of `info_span!`

matklad:
@bowenwang1996 yeah, we should change this, but in a separate PR. At the moment the rest of vm-runner is using info, so I'd rather merge this and change to debug later. 

We don't enable logging for vm-runner on master anyway, so this shouldn't affect actual log output I think (#4038 is not merged)

matklad:
ok, let me merge this (there's a bunch of info calls anyway, so this doesn't make the things worse), and I'll immediatelly send a follow up to switch to debug.

Issue #4057, opened 2021-03-08
Database migration 17 -> 18 failure


bowenwang1996:
On betanet we observed the following
```
2021-03-08 14:23:35.593 INFO tailer - show_logs: 21473|neard|betanet 
Mar 08 14:04:17.448  INFO near: Version: 1.2.0, Build: 1ee021d7, Latest Protocol: 105    
Mar 08 14:04:17.859  INFO near: Opening store database at "/home/ubuntu/.near/betanet/data"    
Mar 08 14:04:18.385  INFO near: Migrate DB from version 16 to 17    
Mar 08 14:04:21.332  INFO near: Migrate DB from version 17 to 18    
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Custom { kind: InvalidData, error: "Not all bytes read" }', core/store/src/migrations.rs:559:69 
stack backtrace: 
  0: rust_begin_unwind 
            at /rustc/18bf6b4f01a6feaf7259ba7cdae58031af1b7b39/library/std/src/panicking.rs:475 
  1: core::panicking::panic_fmt 
            at /rustc/18bf6b4f01a6feaf7259ba7cdae58031af1b7b39/library/core/src/panicking.rs:85 
  2: core::option::expect_none_failed 
            at /rustc/18bf6b4f01a6feaf7259ba7cdae58031af1b7b39/library/core/src/option.rs:1221 
  3: core::result::Result<T,E>::unwrap 
            at /rustc/18bf6b4f01a6feaf7259ba7cdae58031af1b7b39/library/core/src/result.rs:973 
  4: near_store::migrations::migrate_17_to_18::{{closure}} 
            at /var/lib/buildkite-agent/builds/buildkite-i-068dcccb7eab0b579-1/nearprotocol/nearcore-nightly-release/core/store/src/migrations.rs:559 
  5: near_store::migrations::map_col_from_key 
            at /var/lib/buildkite-agent/builds/buildkite-i-068dcccb7eab0b579-1/nearprotocol/nearcore-nightly-release/core/store/src/migrations.rs:322 
  6: near_store::migrations::migrate_17_to_18 
            at /var/lib/buildkite-agent/builds/buildkite-i-068dcccb7eab0b579-1/nearprotocol/nearcore-nightly-release/core/store/src/migrations.rs:556 
  7: neard::apply_store_migrations 
[...]

PR #4065, opened 2021-03-09
T4057

Kouprin:
Fixes https://github.com/near/nearcore/pull/4057.


bowenwang1996:
https://github.com/near/nearcore/issues/4057#issuecomment-796280916

PR #4071, opened 2021-03-10
refactor: remove unused dependency

matklad:


ailisp (approved):
Is there some way to automatic remove unused crate from Cargo.toml?

matklad:
There's https://crates.io/crates/cargo-udeps for that (and I think we might run that somewhere? there's some udeps-related meta in some cargo toml). I am not sure it's worth it to install on CI: the main part of dependency hygiene is still figuring out if dependency actually makes sense.  For example, udeps wouldn't help with https://github.com/near/nearcore/pull/4072

PR #4073, opened 2021-03-10
feat(runtime): use native engine in wasmer 1

ailisp:
This changes compilation artifact format and would have a better execution performance. Deserialization performance is going to be measured separately, but even if it's slower, it'll replaced by flatdata based artifact (in near/wasmer), which has a constant ~0.2ms "deserialization"

Test Plan
---------
CI near-vm-runner tests should still work

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3627,6 +3627,7 @@ dependencies = [
  "wabt",
  "wasmer",
  "wasmer-compiler-singlepass",
+ "wasmer-engine-native",
  "wasmer-runtime-core-near",
  "wasmer-runtime-near",
  "wasmer-types",
--- a/runtime/near-vm-runner/Cargo.toml
+++ b/runtime/near-vm-runner/Cargo.toml
@@ -20,6 +20,7 @@ wasmer-runtime-core = {version = "0.17.1", package = "wasmer-runtime-core-near" 
 wasmer = { version = "1.0.2", optional = true }
 wasmer-types = { version = "1.0.2", optional = true }
 wasmer-compiler-singlepass = { version = "1.0.2", optional = true }
+wasmer-engine-native = { version = "1.0.2", optional = true }
 pwasm-utils = "0.12"
 parity-wasm = "0.41"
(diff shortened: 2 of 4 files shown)


ailisp:
bench se/deser:
https://github.com/wasmerio/wasmer/compare/master...near:bench-cache?expand=1
bench shows read from disk and deser a native artifact of a 2mb wasm takes 4.2ms. It's improved compare to same jit artifact takes 10ms (6ms if replace deser with borsh). Still far slower than mmap flatdata, so i'm going optimze deser NativeArtifact

olonho:
I'd suggest to use https://github.com/near/nearcore/pull/4074.

matklad:
@ailisp am I correct that the benchmarked some more and realised that JIT  engine is good enough? I recall we talked about this in one of the meetings, but don't remember details. Should this be closed then?

ailisp:
@matklad yes execution performance is similar to jit engine. the enhancement of native engine is it deserialize faster than jit engine (before optimize is 4ms vs 6ms, after https://github.com/wasmerio/wasmer/pull/2190 optiomization is 0.4ms vs 6ms). However the huge problem is it doesn't work and require wasmer fix https://github.com/wasmerio/wasmer/issues/2179

ailisp:
I think we're still not complete sure if wasmer will fix it or not, and will they do the artifact deserialization optimization on jit-artifact, we need to sync with them on a certain answer. 
- simply introducing deserialization cost with JIT artifact is the least coding effort.
- optimize away deserialization cost of JIT artifact to avoid introduce deserialization cost is the least protocol change effort compare to the first, and doable with a similar PR like wasmerio/wasmer#2190
- Most ideal case, is they can merge wasmerio/wasmer#2190 and fix wasmerio/wasmer#2179. under this case we simply switch to native engine without introduce deserialize cost and without additional optimization

bowenwang1996:
@ailisp given that we are going to move to wasmer2, do we still need this PR?

ailisp:
No. Wasmer team didn't make native engine work with single pass compiler in wasmer 1, we'll retry tests in wasmer2

PR #4074, opened 2021-03-11
Wasmer1 implementation of preloader.

olonho:


runtime/near-vm-runner/src/wasmer1_runner.rs
```diff
@@ -239,8 +238,77 @@ fn run_method(module: &Module, import: &ImportObject, method_name: &str) -> Resu
...
+) -> (Option<VMOutcome>, Option<VMError>) {
+    // Do we really need that code?
+    if method_name.is_empty() {
+        return (
+            None,
+            Some(VMError::FunctionCallError(FunctionCallError::MethodResolveError(
+                MethodResolveError::MethodEmptyName,
+            ))),
+        );
+    }
```
matklad, line 284: :+1: at minimum, we need a

```
struct MethodName<'a>(&'a str);

impl<'a> TryFrom<&'a str> for MethodName<'a> {
    type Error = MethodResolveError;
}
```

to write it once. But I think just removing this to  get `MethodNotFoundError` in `check_method` would be fine? We don't seem to charge any fees in that case.

runtime/near-vm-runner/Cargo.toml
```diff
@@ -44,7 +46,7 @@ base64 = "0.13"
 default = ["wasmer0_vm", "wasmtime_vm", "wasmer1_vm"]
 wasmer0_vm = [ "wasmer-runtime" ]
 wasmtime_vm = [ "wasmtime", "anyhow"]
-wasmer1_vm = [ "wasmer", "wasmer-types", "wasmer-compiler-singlepass" ]
+wasmer1_vm = [ "wasmer", "wasmer-types", "wasmer-compiler-singlepass",  "wasmer-compiler-cranelift", "wasmer-engine-native", ]
```
matklad, line 49: This will enable & compile the cranelift dependency even if we don't use it. Would it be better to do something like

```
        #[cfg(feature = "wasmer_compiler_cranelift")]
        let engine = wasmer_engine_native::Native::new(wasmer_compiler_cranelift::Cranelift::default())
                .engine();
        #[cfg(feature = "wasmer_compiler_singlepass")]
        let engine = wasmer_engine_native::Native::new(wasmer_compiler_singlepass::Singlepass::default())
                .engine();
```

In this case, compile-time features allow us to avoid compiling dependency altogether, so the pay off is relatively high.

matklad (approved):

runtime/near-vm-runner/src/wasmer1_runner.rs
olonho: But this will change protocol, unfortunately, we must keep returning `MethodEmptyName `.

runtime/near-vm-runner/Cargo.toml
olonho: That's true, however experiments with various combinations of backend/serialization formats are easier the current way.

olonho:
Also we may want to include backend/compiler information into the caching key, if we'll ever consider switching it. Will do in separate PR.

runtime/near-vm-runner/src/wasmer1_runner.rs
matklad: Hm, are we actually sensitive to the *specific* error kind, and not just the fact that the error occurred? If we are, that's scary: in https://github.com/matklad/nearcore/blob/3f47dcd3d7c84195272cd67f426e8061260e2319/runtime/near-vm-errors/src/lib.rs#L37-L43

we more-or-less dump wasmer's internal errors to strings, and I bet their representation will change in the future. I guess, @bowenwang1996 is the best person to ask about that?

olonho: Yes, we are sensitive to the exact error kind, that's why every VM change shall be done very careful.

olonho:
Done in https://github.com/near/nearcore/pull/4077

runtime/near-vm-runner/src/wasmer1_runner.rs
ailisp: Hmm my understand from previously talked to @evgenykuzyakov  is "whether error happens or not" is merklelized in trie, not "what kind of error happened", so this is not a protocol change.

ailisp (approved):
Looks good overall, but before merge we should figure out whether we want enable different kind of wasmer backend based on feature, and if so, with_vm_variant should test both kind

runtime/near-vm-runner/src/wasmer1_runner.rs
bowenwang1996: The exact type of error does not matter, but it matters whether it is an error or not, so we cannot just remove this block.

matklad: > so we cannot just remove this block.

The error would happen down the line, where we won't be able to find a method with an empty name in the list of methods exported by module. 

I guess the high-order bit here that we can refactor this, as long as we preserve the fact of error, and None vm outcome. We'll look into this after the PR is merged.

PR #4077, opened 2021-03-11
Introduce compiler backend data into cache key.

olonho:
As we have certain flexibility in the way how we produce the cached contracts let's introduce an additional field in cache key, so that we could distinguish data compiled with different backend configurations.


runtime/near-vm-runner/src/wasmer1_runner.rs
```diff
@@ -244,3 +244,8 @@ pub fn compile_module(code: &[u8]) -> bool {
     let store = Store::new(&engine);
     Module::new(&store, code).is_ok()
 }
+
+pub(crate) fn wasmer1_vm_hash() -> u64 {
+    // TODO: take into account compiler and engine used to compile the contract.
+    53
```
ailisp, line 250: Since we start to consider this, also helpful to add wasmer crate version

ailisp (approved):
Looks good

matklad (commented):
LGTM, although it feels a bit weird that we specify hash numbers manually. A more rust-like approach would be to define a struct with derifeved `Hash` and use that, [like we do here](https://github.com/near/nearcore/blob/df2d8bac977461c3abded5ef52ac3000f53e9097/core/primitives-core/src/config.rs#L93-L100). 

Or, given that we compute the final hash using [borsh](https://github.com/near/nearcore/blob/df2d8bac977461c3abded5ef52ac3000f53e9097/runtime/near-vm-runner/src/cache.rs#L33-L39) anyway, I guess we can avoid hash computation altogrether and just use a `#[derive(BorshDeserialize, BorshSerialize)]` data?

runtime/near-vm-runner/src/wasmer1_runner.rs
olonho: We could encode all relevant info in this version.

olonho:
> LGTM, although it feels a bit weird that we specify hash numbers manually. A more rust-like approach would be to define a struct with derifeved `Hash` and use that, [like we do here](https://github.com/near/nearcore/blob/df2d8bac977461c3abded5ef52ac3000f53e9097/core/primitives-core/src/config.rs#L93-L100).
> 
> Or, given that we compute the final hash using [borsh](https://github.com/near/nearcore/blob/df2d8bac977461c3abded5ef52ac3000f53e9097/runtime/near-vm-runner/src/cache.rs#L33-L39) anyway, I guess we can avoid hash computation altogrether and just use a `#[derive(BorshDeserialize, BorshSerialize)]` data?

That's not exactly my intention. Point is that we shall manually control whenever given configuration is incompatible.

olonho:
PTAL

runtime/near-vm-runner/src/wasmer1_runner.rs
```diff
@@ -243,11 +245,31 @@ pub(crate) fn compile_wasmer1_module(code: &[u8]) -> bool {
...
+impl Wasmer1Config {
+    fn config_hash(self: Self) -> u64 {
+        let mut s = DefaultHasher::new();
+        self.hash(&mut s);
+        s.finish()
+    }
+}
+
+const WASMER1_CONFIG: Wasmer1Config =
+    Wasmer1Config { seed: 53, use_cranelift: false, use_native_engine: false };
```
matklad, line 264: If we are being fancy, we can also do

```rust
impl Wasmer1Config {
    const CURRENT: Self = Self { ... }
}
```

Seems a touch namespace-er, but doesn't really matter.

matklad (approved):

PR #4079, opened 2021-03-11
refactor(jsonrpc): Refactor view_client structured errors according to removal of BlockMissing for GetBlockError, GetChunkError, QueryError

khorolets:
According to recent removal of `BlockMissing` we had to refactor some structures to be up to date


chain/client/src/view_client.rs
```diff
@@ -203,20 +203,39 @@ impl ViewClientActor {
...
             BlockReference::Finality(ref finality) => {
-                let block_hash = self.get_block_hash_by_finality(&finality)?;
+                let block_hash = match self.get_block_hash_by_finality(&finality) {
+                    Ok(block_hash) => block_hash,
+                    Err(_) => {
+                        return Err(QueryError::UnknownBlock {
+                            block_reference: msg.block_reference,
+                        })
+                    }
+                };
```
frol, line 213: You can use `.map_err(...)?` instead of an explicit match, btw.

frol (approved):

CODEOWNERS
```diff
@@ -1,6 +1,7 @@
 # CODEOWNERS: https://help.github.com/articles/about-codeowners/
 
 /chain/ @SkidanovAlex @bowenwang1996
+/chain/client/src/view_client.rs @frol @khorolets
```
bowenwang1996, line 4:
```suggestion
/chain/client/src/view_client.rs @frol @khorolets @bowenwang1996
```

bowenwang1996 (approved):

CODEOWNERS
khorolets: You already owner of entire `/chain/` does it makes sense to add you as owner of specific file in it? Is our change overrides your ownership?

frol (approved):

PR #4080, opened 2021-03-11
refactor: use more natural interface for contract runner

matklad:


runtime/near-vm-runner/src/runner.rs
```diff
@@ -33,8 +33,8 @@ pub fn run<'a>(
     profile: &ProfileData,
 ) -> (Option<VMOutcome>, Option<VMError>) {
     run_vm(
-        code_hash,
-        code,
+        code.hash.as_ref().to_vec(),
+        code.code.as_slice(),
```
matklad, line 37: I'll refactor the rest of vm-runner in a separate PR, after this interface PR is merged. That change would be large-ish, but it wont't touch runtime codewners.

runtime/runtime-params-estimator/src/vm_estimator.rs
```diff
@@ -48,12 +45,9 @@ fn call(code: &[u8]) -> (Option<VMOutcome>, Option<VMError>) {
 
     let promise_results = vec![];
 
-    let mut hash = DefaultHasher::new();
-    code.hash(&mut hash);
-    let code_hash = hash.finish().to_le_bytes().to_vec();
+    let code = ContractCode::new(code.to_vec(), None);
```
matklad, line 48: Not sure whether this is important or not, but this now use the correct sha256 hash and not SipHash (which is what DefaultHasher currently is).

willemneal (approved):

runtime/runtime-params-estimator/src/vm_estimator.rs
ailisp: Your change is the correct behavior

ailisp (approved):

bowenwang1996 (approved):

PR #4089, opened 2021-03-12
feat(primitives): Add versioning to Account

EgorKulikov:
Just like we have different versions of Block or BlockHeader types we'll now have same for accounts. As accounts are saved in state we have to be able to deserialize old Account structure as well

### Migration
Script scripts/migrations/30-add-account-versioning.py added to migrate genesis file for betanet

### Test plan
Tests added/updated to check serialization/deserialization, including from older version. Also for this to work new version should always serialize into longer sequence of bytes, such check added as well
Tests that current genesis is deserializable


neard/tests/load_genesis.rs
```diff
@@ -0,0 +1,9 @@
+use near_chain_configs::Genesis;
+
+#[test]
+fn test_load_genesis() {
```
bowenwang1996, line 4: wouldn't this test fail if we enable `protocol_feature_add_account_versions`?

scripts/migrations/30-add-account-versioning.py
```diff
@@ -0,0 +1,23 @@
+"""
```
bowenwang1996, line 1: @chefsale @mhalambek do you want this script to be part of nearcore or part of near-ops? Also @EgorKulikov the naming of this file is a bit confusing -- the number is used to mean the new protocol version. We can either drop it or change it to 107

neard/tests/load_genesis.rs
EgorKulikov: Why will it? And I think it should've ran as part of CI, no?

scripts/migrations/30-add-account-versioning.py
EgorKulikov: Name changed

bowenwang1996 (approved):

bowenwang1996:
@chefsale please make sure that you run the migration script provided in this file when doing hard fork on betanet

neard/tests/load_genesis.rs
bowenwang1996: I misread and thought we are using the same file in both cases :)

scripts/migrations/107-add-account-versioning.py
```diff
@@ -0,0 +1,23 @@
+"""
```
frol, line 1: @bowenwang1996 does this mean that everyone (every RPC node, every Bisontrails instance, every indexer node, etc) will be required to run this script or re-download the genesis file and records once we enable this feature? It feels like quite a cumbersome process. It seems that I had the wrong impression that the genesis file never changes. Do nodes need the file only on their first start and it is never used/loaded beyond that point?

frol (approved):

neard/tests/load_genesis.rs
frol: Should the node with the feature enabled be able to parse the existing genesis file?

scripts/migrations/107-add-account-versioning.py
bowenwang1996: This only happens when we do hard fork. Also unless it is mainnet, there is no reason for other people to run the migration themselves and they can wait for us to provide the new genesis. If it is mainnet, then first of all we do not expect to do hardforks there and also even if we do it requires much more careful coordination

ailisp (approved):

core/primitives-core/src/account.rs
```diff
@@ -20,6 +23,189 @@ pub struct Account {
...
+            enum DeserializableAccount {
+                AccountV1(AccountV1),
+            };
+            let deserialized_account = DeserializableAccount::deserialize(buf)?;
+            match deserialized_account {
+                DeserializableAccount::AccountV1(account) => Ok(Account::AccountV1(account)),
+            }
+        }
+    }
+}
```
evgenykuzyakov, line 207: It worth moving it up to line 46, so you don't forget to add a field to `DeserializableAccount`.

core/primitives-core/src/account.rs
```diff
@@ -20,6 +23,189 @@ pub struct Account {
...
+    #[serde(with = "u128_dec_format_compatible")]
+    pub amount: Balance,
+    /// The amount locked due to staking.
+    #[serde(with = "u128_dec_format_compatible")]
+    pub locked: Balance,
+    /// Hash of the code stored in the storage for this account.
+    pub code_hash: CryptoHash,
+    /// Storage used by the given account, includes account id, this struct, access keys and other data.
+    pub storage_usage: StorageUsage,
+}
```
evgenykuzyakov, line 39: Maybe it makes sense to make fields private for this `AccountV1` and old account `Account`. It shouldn't affect anything in the code and will catch any potential bugs with the direct access to the fields for files where the feature might not be enabled.

neard/res/genesis_for_test.json
```diff
@@ -0,0 +1,286 @@
+{
```
evgenykuzyakov, line 1: The name doesn't explain what's different with this genesis. Should it explicitly call which test or protocol version?

scripts/migrations/30-add-account-versioning.py
chefsale: Not really sure what is the use case  of this script. I see its a one time migration, but not sure what else would we use it in near-ops. If its just a one off migration it can stay in nearcore

core/primitives-core/src/account.rs
EgorKulikov: I guess adding new version will lead to changing new, and that will fail serialization/deserialization tests

core/primitives-core/src/account.rs
EgorKulikov: Done

neard/res/genesis_for_test.json
EgorKulikov: Done

chefsale (approved):

mhalambek (approved):

PR #4090, opened 2021-03-12
refactor: switch the rest of vm-runner to use ContractCode

matklad:


runtime/near-vm-runner/src/preload.rs
```diff
@@ -28,12 +28,8 @@ struct VMCallData {
 struct CallInner {
     rx: Receiver<VMCallData>,
 }
-
-// TODO: consider using https://crates.io/crates/scoped_threadpool and not clone the data.
-#[derive(Clone)]
 pub struct ContractCallPrepareRequest {
-    pub code_hash: CryptoHash,
-    pub code: Vec<u8>,
+    pub code: Arc<ContractCode>,
```
matklad, line 32: The call site has an arc already, so this clears the cloning concern:

https://github.com/matklad/nearcore/blob/3011e716d780b8f66273b9a0f8865765e78457ca/runtime/runtime/src/actions.rs#L80-L97

ailisp: Hmm maybe i need to upgrade clion and intellij-rust. I don't have these inlay hints

ailisp (approved):

runtime/near-vm-runner/src/preload.rs
matklad: I think you just need to enable them in the config, they were implemented many years ago, while I was still at JetBrains. You need this config:

image

olonho (approved):

PR #4098, opened 2021-03-14
Validate each edge only once

pmnoxx:
Currently we may validate each edge a few times after a node connects to a network. This PR will guarantee that each node edge will be only verified once on node joining the network.

### DIFF
--- a/chain/network/src/routing.rs
+++ b/chain/network/src/routing.rs
@@ -1,6 +1,6 @@
 use std::collections::{hash_map::Entry, HashMap, HashSet, VecDeque};
 use std::ops::Sub;
-use std::sync::Arc;
+use std::sync::{Arc, RwLock};
 use std::time::{Duration, Instant};
 
 use serde::Serialize;
@@ -308,6 +308,7 @@ impl RoutingTable {
             account_peers: SizedCache::with_size(ANNOUNCE_ACCOUNT_CACHE_SIZE),
             peer_forwarding: HashMap::new(),
             edges_info: HashMap::new(),
+            edges_info_shared: Default::default(),
             route_back: RouteBackCache::new(
                 ROUTE_BACK_CACHE_SIZE,
                 ROUTE_BACK_CACHE_EVICT_TIMEOUT,
--- a/chain/network/src/routing.rs
+++ b/chain/network/src/routing.rs
@@ -1,6 +1,6 @@
 use std::collections::{hash_map::Entry, HashMap, HashSet, VecDeque};
 use std::ops::Sub;
-use std::sync::{Arc, RwLock};
+use std::sync::{Arc, Mutex};
 use std::time::{Duration, Instant};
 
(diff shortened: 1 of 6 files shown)


pmnoxx:
#3621

chain/network/src/routing.rs
```diff
@@ -262,6 +262,10 @@ pub struct RoutingTable {
     pub peer_forwarding: HashMap<PeerId, Vec<PeerId>>,
     /// Store last update for known edges.
     pub edges_info: HashMap<(PeerId, PeerId), Edge>,
+    /// Shared version of edges_info used by multiple threads
+    edges_info_shared: Arc<Mutex<HashMap<(PeerId, PeerId), u64>>>,
+    /// List of edges that should be added
+    edges_to_add_shared: Arc<Mutex<Vec<Edge>>>,
```
pmnoxx, line 268: I would like to use a Lock-free unbounded multiple writers one reader queue, for `edges_to_add_shared`, similar to one that `Actix` uses. I couldn't find one on crates.io, but it's there probably somewhere.

pmnoxx: Otherwise, we may block `PeerManager` while doing `edges_to_add_shared.lock()`.

pmnoxx: Though I think `Mutex` is fine. We are only going to add a small number of edges in total.

chain/network/src/types.rs
```diff
@@ -1282,7 +1283,11 @@ pub enum PeerManagerRequest {
     UnregisterPeer,
 }
 
-pub struct EdgeList(pub Vec<Edge>);
+pub struct EdgeList(
```
bowenwang1996, line 1286: it is better to use named fields here. Otherwise it is hard to read `EdgeVerifier::handle`

bowenwang1996 (commented):
@mfornet can you review this PR?

chain/network/src/types.rs
pmnoxx: Fixed

bowenwang1996 (commented):
What does `QueueSender` do and why do we need it?

pmnoxx:
> What does `QueueSender` do and why do we need it?

I noticed that on my machine we would spend 20s in `PeerManager`, processing response from `EdgeVerifier`. That would probably be over 1m on Google Cloud.
Btw, my performance stats code is not tracking time spend in futures, which where spawned with `.spawn(ctx)`. That's why we didn't notice that. 

Therefore I decided to add MPSC lock-free unbounded queue for sending edges from `EdgeVerifier` to `PeerManager`. My intention is to limit the amount of time we block `PeerManager`.

I'll change it so we spend at most 50ms at a time processing verified edges in `PeerManager`. Then we yield, and then we continue again 50ms later.

pmnoxx:
https://github.com/near/nearcore/pull/4111

PR #4101, opened 2021-03-14
VM Mock: export Receipt fields + rustfmt autoformatting

robert-zaremba:
In unit tests, we can use `near_sdk::test_utils::get_created_receipts` ([ref](https://docs.rs/near-sdk/3.0.0-pre.2/near_sdk/test_utils/fn.get_created_receipts.html)) to get transaction receipts created in the most recent smart contract call. This would be extremely useful if the returned structure: [Receipt](https://docs.rs/near-vm-logic/4.0.0-pre.1/src/near_vm_logic/mocks/mock_external.rs.html#217),  would have it's fields exported. This would allow inspecting receipts in unit tests and do more things without running the simulation framework.

Changeset in this PR:
* exporting `Receipt` fields, and it's all nested types.
* formatting file with `rustfmt`


bowenwang1996 (commented):
I don't have all the context here to judge whether we need this change or not. @austinabell @mikedotexe please take a look

matklad:
I am fine landing this PR: these are mocks specifically intended for testing, so making them public wouldn't hurt. 

However, the bigger issue that sdk public, verisioned API ends up defined in nearcore repository is I think a big problem. 

@austinabell I wonder if you might want to just define your own `pub struct MockedExternal` in the sdk?

austinabell:
> I am fine landing this PR: these are mocks specifically intended for testing, so making them public wouldn't hurt.
> 
> However, the bigger issue that sdk public, verisioned API ends up defined in nearcore repository is I think a big problem.
> 
> @austinabell I wonder if you might want to just define your own `pub struct MockedExternal` in the sdk?

Honestly, from the SDK perspective, I don't see why this would change anything for us. `MockedExternal`'s API is not exposed at all, so any update or change to the interface would only have to be handled by us. This seems like a simpler option than duplicating and maintaining a copy. 

Let me know if I'm overlooking something here, just a quick look over.

austinabell:
> I am fine landing this PR: these are mocks specifically intended for testing, so making them public wouldn't hurt.
> 
> However, the bigger issue that sdk public, verisioned API ends up defined in nearcore repository is I think a big problem.
> 
> @austinabell I wonder if you might want to just define your own `pub struct MockedExternal` in the sdk?

Will just define a receipt type within the SDK to have more control of this going forward. I'll link here when I open

austinabell (approved):
Would need these changes to create the wrapper type in SDK. Can bundle this and updating near-vm-logic dep

bowenwang1996 (approved):

mikedotexe (approved):
Awesome idea, Robert, thank you for your contribution

bowenwang1996:
@robert-zaremba could you run cargo fmt? looks like CI failed because of formatting

austinabell:
removing automerge for discussion as I've put in a PR here https://github.com/near/near-sdk-rs/pull/479 which removes the need for this altogether, based on a suggestion from @matklad. Feel free to add automerge again, but this isn't needed for the SDK anymore

bowenwang1996:
> removing automerge for discussion as I've put in a PR here [near/near-sdk-rs#479](https://github.com/near/near-sdk-rs/pull/479) which removes the need for this altogether, based on a suggestion from @matklad. Feel free to add automerge again, but this isn't needed for the SDK anymore

I see. Could you close this PR once near/near-sdk-rs#479 is merged?

austinabell:
[near/near-sdk-rs#479](https://github.com/near/near-sdk-rs/pull/479) is merged and 4.0.0-pre.1 is released so try the existing usage with this version :)

PR #4107, opened 2021-03-15
Set transaction size limit

Longarithm:


core/primitives/src/transaction.rs
```diff
@@ -72,6 +72,14 @@ impl Action {
             _ => 0,
         }
     }
+    #[cfg(feature = "protocol_feature_tx_size_limit")]
```
nearmax, line 75: I suggest for the transaction size to use the size of the transaction serialized as borsh. This will remove potential tricky angles of attack, e.g. it is still possible to create a heavy transaction that adds a very large number of function access keys with long method names. Also, that removes ambiguity in interpreting what transaction size is, so e.g. we can say that block size is sum of the transaction sizes plus header data.

Ideally, we would add a trait to Borsh called `BorshSize` that implements `borsh_size() -> u64` method which we can use similarly to `BorshSerialize` and `BorshDeserialize`, like this: `#[derive(BorshSize)]`. AFAIR there might some places in our code where we measure the size of the objects by first serializing them in Borsh and then counting bytes, so it might be useful there too. It shouldn't be difficult to add to borsh and will probably be largely copy-paste of https://github.com/near/borsh-rs/blob/master/borsh/src/ser/mod.rs

nearmax (changes requested):

core/primitives/src/transaction.rs
bowenwang1996: Yeah agree with what Max said here

runtime/runtime/src/verifier.rs
```diff
@@ -326,6 +328,18 @@ pub(crate) fn validate_actions(
         });
     }
 
+    #[cfg(feature = "protocol_feature_tx_size_limit")]
+    {
+        let total_size =
+            total_size(actions).map_err(|_| ActionsValidationError::IntegerOverflow)? as u64;
+        if total_size > limit_config.max_transaction_size {
```
matklad, line 335: Intuitively, it seems that `IntegerOverflow` should result in `TransactionSizeExceeded` instead of being a separate error?

core/primitives-core/src/config.rs
```diff
@@ -68,6 +68,9 @@ pub struct VMLimitConfig {
     pub max_length_returned_data: u64,
     /// Max contract size
     pub max_contract_size: u64,
+    /// Max transaction size
+    #[cfg(feature = "protocol_feature_tx_size_limit")]
+    pub max_transaction_size: u64,
```
matklad, line 73: That's mostly unrelated to the PR, but I am somewhat surprised that we use raw `u64` for quantities, instead of using strongly typed wrappers a-la `pub struct Bytes(pub u64);`, `pub struct Gas(pub u64);`, etc. For byte sizes, we started to use https://github.com/hyunsik/bytesize somewhere in network.

runtime/runtime/src/verifier.rs
Longarithm: I agree, but in the code above `total_prepaid_gas` is handled in the same fashion.
@bowenwang1996 is the `IntegerOverflow` error important enough to track each case when it happens?

core/primitives-core/src/config.rs
Longarithm: I guess `Gas` and `Balance` have been separated into such types, because they are widely used and it makes sense to change type size everywhere, if we want to.
But it's not simple to find cases where other quantities in `VMLimitConfig` can be united into other types.
What is the common sense beneath replacing raw types with strongly typed wrappers?
Also, apparently `VMLimitConfig` is so widely used so it could be simpler to create a separate issue for this.

runtime/runtime/src/verifier.rs
bowenwang1996: In this case we can return `TransactionSizeExceeded`.

core/primitives-core/src/config.rs
matklad: There are big benefits, but a bunch of small ones. 

Units become unambiguous. If you have `size: Bytes`, you don't need to ask whether the unit is byte or page, you can *ask* for what you need: `size.as_bytes()`, `size.as_kilobytes`. You can say `timeout: Duration` rather then `timeout_ms: u64`.

Common functionality becomes available. I see a couple of places that did `/ 1024` for formatting purposes. With a type, you can just write `size.as_kilobytes` or even `format!("{:?}", duration)` and have the type automatically pick a human-readable formatting. 

Reduced probability of type confusion error: it's impossible to add apples and oranges if the two are different types. 

(If we implement this our selves): we can avoid providing unchecked arithmetics. Instead of various `safe_add_stuff` utils which you need to remember to use, there'll be *only* checked methods like `size.add(other_size)?`.

runtime/runtime/src/verifier.rs
Longarithm: Closing the thread because tx size computation was changed (see above).

core/primitives-core/src/config.rs
Longarithm: Makes sense. Created a separate issue https://github.com/near/nearcore/issues/4139

runtime/runtime/src/verifier.rs
```diff
@@ -57,6 +57,19 @@ pub fn validate_transaction(
         return Err(InvalidTxError::InvalidSignature.into());
     }
 
+    #[cfg(feature = "protocol_feature_tx_size_limit")]
```
bowenwang1996, line 60: This check should only be done once. There is no reason to check transactions for which you already checked. For example, inside `prepare_transactions` it doesn't make sense to check again because this check should have been done when the transaction is received. To further reduce the overhead, you could change `NetworkClientMessage::Transaction` to include the transaction size. It probably even makes sense to implement custom borsh deserializer that would fail when the a larger-than-allowed transaction arrives on the network (this can potentially be done in a separate PR)

core/primitives/src/transaction.rs
Longarithm: Thank you!
Changed the way of size computation and added a separate issue for adding new trait.

runtime/runtime/src/verifier.rs
Longarithm: It seems that at least one serialization is already happening inside `SignedTransaction::new`, to save `Transaction` hash. So I saved `size` there in the same fashion and started to use it, which shouldn't give significant overhead.

bowenwang1996: Yeah that makes sense

bowenwang1996 (approved):

bowenwang1996:
Please run nayduck

bowenwang1996:
@Longarithm is this PR ready for review?

Longarithm:
Not yet, I got some Nayduck tests failing: http://nayduck.eastus.cloudapp.azure.com:3000/#/run/1401

Longarithm:
http://nayduck.eastus.cloudapp.azure.com:3000/#/run/1414 - runs for `master`, 26 failed tests in nightly
http://nayduck.eastus.cloudapp.azure.com:3000/#/run/1413 - runs for this branch, 24 failed tests in nightly
The only new test failing is `pytest sanity/epoch_switches.py`
Adding @mikhailOK

mikhailOK (approved):

core/primitives/src/transaction.rs
```diff
@@ -205,22 +205,31 @@ pub struct SignedTransaction {
...
-        let mut signed_tx = Self { signature, transaction, hash: CryptoHash::default() };
+        let mut signed_tx =
+            Self { signature, transaction, hash: CryptoHash::default(), size: u64::default() };
         signed_tx.init();
         signed_tx
     }
 
     pub fn init(&mut self) {
-        self.hash = self.transaction.get_hash();
+        let hash_and_size = self.transaction.get_hash_and_size();
```
nearmax, line 221:
```rust
let (hash, size) = self.transaction.get_hash_and_size();
```
It is uncommon in Rust to not decompose the tuple, especially if you are accessing all elements.

nearmax (approved):

core/primitives/src/transaction.rs
Longarithm: Yes, makes sense.

frol (approved):

PR #4111, opened 2021-03-15
Validate each edge only once

pmnoxx:
Currently we may validate each edge a few times after a node connects to a network. This PR will guarantee that each node edge will be only verified once on node joining the network.

Reposting #4098 to fir CI.


chain/network/src/routing.rs
```diff
@@ -262,6 +263,11 @@ pub struct RoutingTable {
     pub peer_forwarding: HashMap<PeerId, Vec<PeerId>>,
     /// Store last update for known edges.
     pub edges_info: HashMap<(PeerId, PeerId), Edge>,
+    /// Shared version of edges_info used by multiple threads
+    edges_info_shared: Arc<Mutex<HashMap<(PeerId, PeerId), u64>>>,
+    /// Queue of edges verified, but not added yes
+    pub edges_to_add_receiver: QueueReceiver<Edge>,
+    pub edges_to_add_sender: QueueSender<Edge>,
```
bowenwang1996, line 270: Do you think it is better to put these three fields in a separate struct?

chain/network/Cargo.toml
```diff
@@ -21,6 +21,7 @@ tracing = "0.1.13"
 strum = { version = "0.20", features = ["derive"] }
 near-rust-allocator-proxy = "0.2.9"
 bytesize = "1.0.1"
+conqueue = "0.4.0"
```
bowenwang1996, line 24: How confident are we about this crate? cc @matklad

bowenwang1996 (approved):

chain/network/Cargo.toml
pmnoxx: @bowenwang1996 It's a small package just only 200 lines of code. I read the code it looks good to me.

chain/network/src/routing.rs
pmnoxx: We always use those fields with `self.routing_table.<FIELD>`, we don't even touch them inside `routing.rs`.
It makes sense to put them in a separate struct inside `PeerManager`.

pmnoxx: I put them to `EdgeVerifierHelper`. The purpose of those fields is to communicate with `EdgeVerifier`,  that's much better now. Thanks.

pmnoxx:
@bowenwang1996 I split broadcasting edges and accounts. The code is much cleaner now.

I also made is so, we run a trigger every 100ms to add edges for at most 50ms. This should ensure that PeerManager is reponsive.

chain/network/src/peer_manager.rs
```diff
@@ -202,6 +227,92 @@ impl PeerManagerActor {
...
+                        EdgeType::Added => {
+                            self.wait_peer_or_remove(ctx, edge.clone());
+                            continue;
+                        }
+                        EdgeType::Removed => {}
+                    }
+                }
+            }
+            new_edges.push(edge);
+            if start.elapsed() >= Duration::from_millis(50) {
```
bowenwang1996, line 291: let's make this an associated constant of `PeerManagerActor`. Also might be better to use 100ms

pmnoxx: @bowenwang1996 We can configure it as we want. I configured it to do 50ms of work with 50ms delay.

If you prefer I can set it to do 100ms of work, with 50ms rest in between. 
Though I think it's better to do less work more often, for example 20 ms of work with 1ms rest.

chain/network/Cargo.toml
matklad: Yeah, I'd probably go for https://docs.rs/crossbeam/0.8.0/crossbeam/queue/index.html instead. That's 200 lines of `unsafe` and relaxed atomics, and that's hard to get right. crossbeam channel received much more scrutiny . Additionally, each new crate is a supply chain attack vector -- you can steal/fish crates.io credentials, publish a patch release, and it'll get picked up by a next cargo update. Popular crates also suffer from this problem, but their releases have more eyes on them. But we already have crossbeam as a transitive dependency. 

Crossbeam is mpmc, while this is mpsc, so in theory there might be a perf difference, but for that we need to look at the benchmarks first. 

More generally, I am a bit unnerved by the number of dependencies we already have. There's 3 different versions of rand, for example :)

bowenwang1996: @pmnoxx

chain/network/src/peer_manager.rs
bowenwang1996: Okay that's fine

chain/network/Cargo.toml
pmnoxx: I looked at that library, it's using a spin lock, which could cause serious performance issues if another thread gets suspended by the operating system while having the lock. I would not recommend to use it. I would be if sacrificing performance, if it means that the locking mechanism by a constant factor of 10 times slower. However, using spin lock can potentially cause delays of up to few hundreds milliseconds.  This can happen often since` EdgeVerifier` can request lock quite often.

I saw `Actix` uses similar approach like `conqueue`: 
https://github.com/actix/actix/blob/master/actix/src/address/queue.rs
It's a lock-free mpsc queue, they copied data from: https://docs.rs/futures/0.1.31/src/futures/unsync/mpsc.rs.html#1-474
This approach doesn't use spin lock, which so this is a preferred way for process communication.

If we don't want to use `conqueue`, then I think we should copy the implementation from https://docs.rs/futures/0.1.31/src/futures/unsync/mpsc.rs.html
That approach has been well tested, and it's a preferred way for multi thread communication, which both `actix` and `futures` use.

We could extend this implement prioritization. Instead of having one queue, we could have a set of `k` queues, which priorities 0 to k -1. We could for example, every 1ms check whenever any of the queues has data, process data from lowest queue, and keep doing that until all queue are empty. That way we could add prioritization on top of Actix/Tokio. If I were to implement my own IO loop, I would also use queues like that as well.

matklad: Oh dear, unbounded spin-loops are indeed a pretty serious problem. I am surprised, to say the least, to learn that crossbeam seg-queue is susceptible, without this being mentioned in the docs.  Opened https://github.com/crossbeam-rs/crossbeam/issues/675

Yeah, in this case this dependency seems reasonable!

PR #4113, opened 2021-03-15
Reset max memory usage after loading config

pmnoxx:
We want to know when given thread reaches peak memory usage. However, the main thread reaches peak memory usage while loading config, then it frees memory. We will reset peak memory usage after `start_with_config`, so we can see report when the main thread reaches peak memory usage.

### DIFF
--- a/neard/src/lib.rs
+++ b/neard/src/lib.rs
@@ -26,6 +26,7 @@ use near_store::migrations::{
     migrate_7_to_8, migrate_8_to_9, migrate_9_to_10, set_store_version,
 };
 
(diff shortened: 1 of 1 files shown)


neard/src/lib.rs
```diff
@@ -294,5 +295,8 @@ pub fn start_with_config(
 
     trace!(target: "diagnostic", key="log", "Starting NEAR node with diagnostic activated");
 
+    // We probably reached peak memory once on this thread, we want to see when it happens again.
+    reset_memory_usage_max();
```
bowenwang1996, line 299: let's put this behind `performance_stats`

frol (approved):

PR #4114, opened 2021-03-15
Fix build.sh failing

pmnoxx:
Bare wasm contract is 35k on my system, which causes script to fail, due to `dd` value being negative.

```

++ filesize target/wasm32-unknown-unknown/release/test_contract.wasm
++ local file=target/wasm32-unknown-unknown/release/test_contract.wasm
+++ stat -c%s target/wasm32-unknown-unknown/release/test_contract.wasm
++ size=39792
++ '[' 0 -eq 0 ']'
++ echo 39792
++ return 0
+ bare_wasm=39792
+ echo 39792
39792
++ expr 10240 - 39792
+ dd if=/dev/urandom of=./res/payload bs=-29552 count=1
dd: invalid number: ‘-29552’
```

I also switched `#!/bin/bash` to `#!/usr/bin/env bash`, which should be portable and work for both linux and macOS.

### DIFF
--- a/runtime/runtime-params-estimator/test-contract/build.sh
+++ b/runtime/runtime-params-estimator/test-contract/build.sh
@@ -30,6 +30,7 @@ echo ${bare_wasm}
 
 # 10KiB
 
+# hotfix for people building cargo with debug flags, causing binary to be 35kb
 if [ "${bare_wasm}" -ge 10239 ]; then
     bare_wasm=10239
 fi
--- a/runtime/runtime-params-estimator/test-contract/build.sh
+++ b/runtime/runtime-params-estimator/test-contract/build.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 set -ex
 
 function filesize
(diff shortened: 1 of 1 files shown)


runtime/runtime-params-estimator/test-contract/build.sh
```diff
@@ -29,6 +29,12 @@ echo ${bare_wasm}
 # bloat its size to the given values.
 
 # 10KiB
+
+# hotfix for people building cargo with debug flags, causing binary to be 35kb
```
willemneal, line 33: Shouldn't users always build release when testing performance?

pmnoxx: I'm running release build with debug symbols on, so the performance shouldn't be affected. It's useful for debugging though.

We need files produced by this build script for running `cargo test`. 
That wasn't the case in the past. 
By default all tests run when you execute `cargo test`, you have no control over it. So this is a regression.

pmnoxx: I would also be fine if we don't merge `build.sh` fix, but instead disable broken tests mentioned in https://github.com/near/near-ops/issues/440
Alternatively, we can commit 
```

49 | |     CODE_10K => "test-contract/res/stable_small_contract.wasm",
50 | |     CODE_100K => "test-contract/res/stable_medium_contract.wasm",
51 | |     CODE_1M => "test-contract/res/stable_large_contract.wasm",
``` to the repo

matklad: >I'm running release build with debug symbols on, so the performance shouldn't be affected. It's useful for debugging though.

I think there's a bit of confusion here. There are two "debug" related cargo toml fields. `debug` controls debug symbols (`-g`). `debug-assertions` controls `debug_assert`. The `lazy_static_include` crate uses the latter to switch between compile time and runtime inclusion.

matklad: https://github.com/near/nearcore/pull/4125 fixes the compilation error.

pmnoxx: I'll close this PR then.

PR #4115, opened 2021-03-15
Versioning for ValidatorStake

birchmd:
The `ValidatorStake` struct will need to be extended to include a new field in future work related to the new block producer and chunk producer selection (see https://github.com/near/NEPs/pull/167 ).

In this PR we replace the struct with an enum with each variant representing a different version. This will allow easy iteration on `ValidatorStake` fields, while keeping necessary code paths to process old blocks and chunks.


core/store/src/migrations.rs
```diff
@@ -585,9 +626,9 @@ pub fn migrate_17_to_18(path: &String) {
 
 #[cfg(feature = "protocol_feature_rectify_inflation")]
 pub fn migrate_18_to_rectify_inflation(path: &String) {
```
birchmd, line 628: @bowenwang1996 I'm not really sure about this one. Whether it should use the old structures or new depends on which migration is applied first.

core/primitives/src/views.rs
```diff
@@ -1075,7 +1112,32 @@ impl From<FinalExecutionOutcomeWithReceiptView> for FinalExecutionOutcomeView {
 }
 
 #[derive(BorshSerialize, BorshDeserialize, Serialize, Deserialize, Debug, Clone, Eq, PartialEq)]
-pub struct ValidatorStakeView {
+pub enum ValidatorStakeView {
```
birchmd, line 1115: @frol I don't know if updating `ValidatorStakeView` to be an enum has any backwards compatibility implications for json rpc. If so, let me know and we can discuss how to resolve.

bowenwang1996 (commented):
Why do you need `protocol_feature_block_header_v4`? It is sufficient to change v3

birchmd:
> Why do you need protocol_feature_block_header_v4? It is sufficient to change v3

`protocol_feature_block_header_v4` was already there as `protocol_feature_block_header_v3`. I only renamed it so that my change to BlockHeader could become v3 because I was unsure about the stabilization timeline for the existing v3. Are you suggesting that I do not add a new version to block header at all, and only change the type in the existing v3? In this case, the versioning ValidatorStake change would be conditional on whatever changes `protocol_feature_block_header_v3` was making.

bowenwang1996:
> Are you suggesting that I do not add a new version to block header at all, and only change the type in the existing v3

Yes

birchmd:
@bowenwang1996 Done. I changed the existing block header v3 and hid all the other data structure changes behind that feature flag as well.

chain/chain/src/chain.rs
```diff
@@ -357,7 +361,7 @@ impl Chain {
         self.doomslug_threshold_mode = DoomslugThresholdMode::NoApprovals
     }
 
-    pub fn compute_bp_hash_inner(bps: Vec<ValidatorStake>) -> Result<CryptoHash, Error> {
+    pub fn compute_bp_hash_inner<T: BorshSerialize>(bps: Vec<T>) -> Result<CryptoHash, Error> {
```
bowenwang1996, line 364: If you make this change then the function name doesn't make sense any more. We should either just get rid of this function or rename it and move to utils

chain/chain/src/chain.rs
```diff
@@ -367,7 +371,22 @@ impl Chain {
         last_known_hash: &CryptoHash,
     ) -> Result<CryptoHash, Error> {
         let bps = runtime_adapter.get_epoch_block_producers_ordered(&epoch_id, last_known_hash)?;
-        Chain::compute_bp_hash_inner(bps.iter().map(|(bp, _)| bp).cloned().collect::<Vec<_>>())
+        #[cfg(not(feature = "protocol_feature_block_header_v3"))]
```
bowenwang1996, line 374: there is a macro `checked_feature` that you can use

chain/chain/src/chain.rs
```diff
@@ -3067,14 +3087,24 @@ impl<'a> ChainUpdate<'a> {
...
+                .flat_map(|chunk| chunk.validator_proposals()),
+        ) {
+            // Can only occur if there were a different number of proposals in the header
+            // and chunks
+            if hp.is_none() || cp.is_none() {
+                return Err(ErrorKind::InvalidValidatorProposals.into());
+            }
+            // Proposals differed!
+            if hp.unwrap() != cp.unwrap() {
+                return Err(ErrorKind::InvalidValidatorProposals.into());
```
bowenwang1996, line 3106: It's probably better to write this as a match to avoid using `unwrap`

core/primitives/src/epoch_manager.rs
```diff
@@ -49,117 +40,975 @@ pub struct EpochConfig {
...
+        pub fn timestamp_nanosec(&self) -> &u64 {
+            match self {
+                Self::V1(v1) => &v1.timestamp_nanosec,
+                Self::V2(v2) => &v2.timestamp_nanosec,
+            }
+        }
+    }
+
+    #[derive(Default, BorshSerialize, BorshDeserialize, Serialize, Eq, PartialEq, Clone, Debug)]
+    pub struct BlockInfoV2 {
```
bowenwang1996, line 239: Can you add a comment on what changed between the versions?

core/primitives/src/epoch_manager.rs
```diff
@@ -49,117 +40,975 @@ pub struct EpochConfig {
...
+        pub seat_price: Balance,
+        /// Current protocol version during this epoch.
+        #[default(PROTOCOL_VERSION)]
+        pub protocol_version: ProtocolVersion,
+    }
+
+    #[derive(
+        SmartDefault, BorshSerialize, BorshDeserialize, Serialize, Clone, Debug, PartialEq, Eq,
+    )]
+    pub struct EpochInfoV2 {
```
bowenwang1996, line 551: Same here

core/primitives/src/version.rs
```diff
@@ -166,6 +166,14 @@ lazy_static! {
...
     };
 }
+#[cfg(all(feature = "nightly_protocol", feature = "protocol_feature_block_header_v3"))]
+lazy_static! {
+    pub static ref BLOCK_HEADER_V3_VERSION: ProtocolVersion =
+        *PROTOCOL_FEATURES_TO_VERSION_MAPPING.get(&ProtocolFeature::BlockHeaderV3).unwrap();
+}
+
+#[cfg(not(feature = "protocol_feature_block_header_v3"))]
+pub const BLOCK_HEADER_V3_VERSION: &'static ProtocolVersion = &u32::MAX;
```
bowenwang1996, line 176: What is this for?

core/store/src/migrations.rs
bowenwang1996: This one will be applied first

neard/src/lib.rs
```diff
@@ -193,6 +196,10 @@ pub fn apply_store_migrations(path: &String, near_config: &NearConfig) {
         // version 18 => rectify inflation: add `timestamp` to `BlockInfo`
         migrate_18_to_rectify_inflation(&path);
     }
+    #[cfg(feature = "protocol_feature_block_header_v3")]
+    if db_version <= 18 {
+        migrate_18_to_new_validator_stake(&path);
+    }
```
bowenwang1996, line 202: this should be done in the nightly_protocol block

bowenwang1996 (commented):
Please run nayduck

core/primitives/src/views.rs
```diff
@@ -1069,23 +1107,115 @@ impl From<FinalExecutionOutcomeWithReceiptView> for FinalExecutionOutcomeView {
...
+    use borsh::{BorshDeserialize, BorshSerialize};
+    use near_crypto::PublicKey;
+    use near_primitives_core::types::{AccountId, Balance};
+    use serde::{Deserialize, Serialize};
+
+    #[derive(
+        BorshSerialize, BorshDeserialize, Serialize, Deserialize, Debug, Clone, Eq, PartialEq,
+    )]
+    #[serde(tag = "version")]
+    pub enum ValidatorStakeView {
```
bowenwang1996, line 1123: You can simply add one more field to `ValidatorStakeView`. In terms of json serialization it will be backward compatible

core/primitives/src/version.rs
birchmd: `BLOCK_HEADER_V3_VERSION` is used to select runtime behaviour based on the protocol version. For example it is used [here](https://github.com/near/nearcore/pull/4115/files#diff-bf94debb856e8e44f62c9316b1e0c96d25e57610e3cc02481abe3737c5d2b82bR1147) to decide what version of the chunk header to produce.

Needing to choose logic based on the protocol version (as opposed to compile-time features) is also done elsewhere. For example, it is done [here](https://github.com/near/nearcore/blob/master/core/primitives/src/block_header.rs#L382) to decide which version of the block header to produce. However, in that code, there is an explicit runtime lookup of the `PROTOCOL_FEATURES_TO_VERSION_MAPPING`, which seems unnecessary to me because it is a constant. So I have introduced `BLOCK_HEADER_V3_VERSION` which essentially caches the version lookup.

To make the code compile I also need `BLOCK_HEADER_V3_VERSION` to be defined when the `protocol_feature_block_header_v3` is not enabled, so I inserted a value which says that the block header v3 version is impossibly far in the future. I could probably eliminate the need to have this constant defined when the feature is not enabled, but it might mean making some of the `if` statements where it would be used less clear because additional compile-time feature gating would be needed.

core/primitives/src/views.rs
birchmd: Yes, that's right. The `#[serde(tag = "version")]` attribute makes it serialize the enum variant as an additional field (e.g. `"version": "v1"`) instead of as nested objects. I think this is useful because it means clients can act differently depending on the version of the structure (during the transition period `version` not being a present field will also be relevant). For example, in our [lightclient pytest](https://github.com/near/nearcore/blob/master/pytest/lib/lightclient.py#L145) we validate the `next_bp_hash` field, and this can only be done properly if the version is present in the json structure (because the version -- or lack of version -- influences the borsh serialization). I actually need to fix that in my PR here (thanks for reminding me!).

The reason to make `ValidatorStakeView` an enum at all is because it also implements borsh serialization, where adding an extra field is not backwards compatible. Presently it appears to not be used, so I didn't worry about backwards compatibility in the borsh serialization when introducing the enum in this PR, but it looks like it will be used for [syncing](https://github.com/near/nearcore/blob/master/core/primitives/src/syncing.rs#L217) in the future and so I expect we will be happy it is an enum down the road.

chain/chain/src/chain.rs
birchmd: I don't think I should use the macro in this case because the macro requires being passed the protocol version. The protocol version is not already present in the scope of this function, but it can be looked up because we have the epoch information present. However, in the case `protocol_feature_block_header_v3` is not enabled the protocol version does not need to be looked up, so it would be doing extra work for no reason.

core/store/src/migrations.rs
birchmd: Ok, should be correct as written then.

chain/chain/src/chain.rs
birchmd: Done in https://github.com/near/nearcore/pull/4115/commits/e62cbf26d10b9d19bcc30e07dab5c0fe4ee76b20

Though I did not move the function. Since the error type is specific to `Chain`, I thought it made sense to leave it where it was.

chain/chain/src/chain.rs
birchmd: Done in https://github.com/near/nearcore/pull/4115/commits/95d8768defc681c1a6acf7d36615ce5b13081cfe

core/primitives/src/epoch_manager.rs
birchmd: Done in https://github.com/near/nearcore/pull/4115/commits/a9f5f9004348c8e8175d8e974303f682d06b3942

core/primitives/src/epoch_manager.rs
birchmd: Done in https://github.com/near/nearcore/pull/4115/commits/a9f5f9004348c8e8175d8e974303f682d06b3942

neard/src/lib.rs
birchmd: Done in https://github.com/near/nearcore/pull/4115/commits/0da96728250c49c0dc6e6ae834d90cf9f43f343f

core/primitives/src/views.rs
birchmd: You can see me using `version` in the json serialization now: https://github.com/near/nearcore/pull/4115/commits/8256535d11bf08a803fa5462fbf4f870fe6beb10

The lightclient python test now passes for both stable and nightly binaries.

bowenwang1996: > but it looks like it will be used for syncing in the future and so I expect we will be happy it is an enum down the road.

I don't think syncing will use this struct. `view` struct should only be used in rpc responses.

bowenwang1996: @frol what do you think?

chain/chain/src/chain.rs
```diff
@@ -3067,15 +3087,29 @@ impl<'a> ChainUpdate<'a> {
...
+                .filter(|chunk| block_height == chunk.height_included())
+                .flat_map(|chunk| chunk.validator_proposals()),
+        ) {
+            match (hp, cp) {
+                (None, Some(_)) | (Some(_), None) | (None, None) => {
+                    // Can only occur if there were a different number of proposals in the header
+                    // and chunks
+                    return Err(ErrorKind::InvalidValidatorProposals.into());
+                }
+                (Some(hp), Some(cp)) => {
```
bowenwang1996, line 3105: If you switch the order of the two cases you can simply use `_` for the second case

chain/client/tests/process_blocks.rs
```diff
@@ -998,10 +1008,30 @@ fn test_bad_orphan() {
             };
             let chunk = match &mut body.chunks[0] {
                 ShardChunkHeader::V1(_) => unreachable!(),
+                #[cfg(not(feature = "protocol_feature_block_header_v3"))]
                 ShardChunkHeader::V2(chunk) => chunk,
+                #[cfg(feature = "protocol_feature_block_header_v3")]
+                ShardChunkHeader::V2(_) => unreachable!(),
```
bowenwang1996, line 1014: why is this unreachable?

core/primitives/src/version.rs
bowenwang1996: My question is why this is done the same way as how other protocol upgrades (you can check where `ProtocolFeatures` variants are used and how they used to determine the version). Is there something specific about this change?

core/primitives/src/views.rs
birchmd: > I don't think syncing will use this struct. `view` struct should only be used in rpc responses.

The skeleton for using it is already in place. `EpochSyncResponse` is in `near_primitives::syncing` and uses `LightClientBlockView`, which contains `ValidatorStakeView`. Is it expected this will change?

core/primitives/src/version.rs
birchmd: The main difference is there are multiple possible behaviours, i.e. it is not binary. We need to select either version 1, 2, or 3 depending on the protocol version for both block headers and chunk headers, and this is implemented as `if (cond1) ... else if (cond2) ... else`, where the conditions are checking the current protocol version against known points where the logic changed.

As I mentioned in my previous comment, the logic for block header was already done in this fashion, but used runtime lookup on `PROTOCOL_FEATURES_TO_VERSION_MAPPING` instead of a constant. I thought using a constant was better for both readability and performance.

chain/client/tests/process_blocks.rs
birchmd: Because if the feature is enabled then the code will only ever create `ShardChunkHeader::V3` in tests.

chain/chain/src/chain.rs
birchmd: Done in https://github.com/near/nearcore/pull/4115/commits/70bc050f9bdc187030da0ba5795b3bf646ac46ed

core/primitives/src/version.rs
bowenwang1996: I think what you did here defeats the purpose of having `ProtocolFeature` and `PROTOCOL_FEATURES_TO_VERSION_MAPPING`. If done this way, we might as well have a constant for each version to begin with.

core/primitives/src/views.rs
bowenwang1996: That is unfortunate. We should address that issue instead of perpetuating it.

birchmd: Alright, so it does not need to be an enum for syncing reasons. What do you think about my usage of the version field in the lightclient test? The test attempts to manually verify a hash so it needs to know the details of serialization (of the real ValidatorStake, not just the view), and hence the version is necessary information.

core/primitives/src/version.rs
birchmd: Yes, you are right that it does not effectively make use of those structures. But the main question for me is whether it would be better if I were to use them more directly. I don't think it does, but if you do then I will change it to match what is done with block headers.

bowenwang1996: I think the reason why we use `ProtocolFeature` and `PROTOCOL_FEATURES_TO_VERSION_MAPPING` is to put all versions in one place so that it is easier to manage. For performance, I think it is a valid concern. @matklad do you have any suggestions here?

core/primitives/src/views.rs
bowenwang1996: You can achieve the same by checking whether the new field exists right?

birchmd: > You can achieve the same by checking whether the new field exists right?

That's true, though it seems a little hacky to me that someone would need to know what fields were introduced in which version to figure out what the correct serialization is. But then again, it's a little hacky to try and ad-hoc reconstruct the borsh serialization in the first place.

Let's see what @frol has to say about whether or not a `version` field is useful in the view struct.

chain/chain/src/validate.rs
```diff
@@ -124,27 +123,31 @@ pub fn validate_chunk_with_chunk_extra(
...
-    if prev_chunk_extra.outcome_root != chunk_header.outcome_root() {
+    if *prev_chunk_extra.outcome_root() != chunk_header.outcome_root() {
         return Err(ErrorKind::InvalidOutcomesProof.into());
     }
 
-    if prev_chunk_extra.validator_proposals != chunk_header.validator_proposals() {
+    let chunk_extra_proposals = prev_chunk_extra.validator_proposals();
+    let chunk_header_proposals = chunk_header.validator_proposals();
+    if chunk_header_proposals.len() != chunk_extra_proposals.len()
+        || !chunk_extra_proposals.zip(chunk_header_proposals).all(|(a, b)| a == b)
```
ailisp, line 137: impl PartialEq/Eq on ValidatorStakeIter and simply compare `!=` here?

core/primitives/src/epoch_manager.rs
```diff
@@ -49,117 +40,977 @@ pub struct EpochConfig {
...
+
+        #[inline]
+        pub fn validator_reward(&self) -> &HashMap<AccountId, Balance> {
+            match self {
+                Self::V1(v1) => &v1.validator_reward,
+                Self::V2(v2) => &v2.validator_reward,
+            }
+        }
+
+        pub fn validators_iter(&self) -> ValidatorStakeIter {
```
ailisp, line 701: should these and following match v1/v2 methods be inline too?

core/primitives/src/types.rs
```diff
@@ -401,17 +390,275 @@ pub struct ApprovalStake {
...
+        pub stake: Balance,
+    }
+
+    pub type ValidatorStakeV1 = ValidatorStake;
+
+    pub struct ValidatorStakeIter<'a> {
+        collection: &'a [ValidatorStake],
+        curr_index: usize,
+        len: usize,
+    }
```
ailisp, line 576: I think can define ValidatorStakeIter as `struct ValidatorStakeIter<'a>(std::slice::iter<'a, ValidatorStake>), so don't need keep track of curr_index and len and reuse https://doc.rust-lang.org/std/slice/struct.Iter.html#impl-Iterator to impl Iterator

ailisp (approved):

core/primitives/src/views.rs
frol: I believe it is a great foundation to have everything versioned. Having the internally tagged enum representation plays fine for the backward compatibility :+1:

My only ask would be to use a more explicit tag name: `validator_stake_struct_version` (I am not super happy about the naming, but plain `version` is super confusing, so if you have a better idea, go for it).

frol (approved):
Overall, looks great! Please, find a few nits in the comments

khorolets (approved):

core/primitives/src/version.rs
matklad: If the primary concern here is runtime lookup, it can be eliminated by using a (const-evalable) flat array instead of hash map:

```rust
#[derive(Clone, Copy)]
#[repr(usize)]
pub enum Feature {
    A,
    #[cfg(feature = "b")]
    B,
    __N_Features,
}

pub const fn feature_to_version(f: Feature) -> u32 {
    const MAPPING: [u32; Feature::__N_Features as usize] = {
        let mut res = [0; Feature::__N_Features as usize];
        let map = [
            (Feature::A, 100),
            #[cfg(feature = "b")]
            (Feature::B, 200),
            ];

        let mut idx = 0;
        while idx < map.len() {
            res[map[idx].0 as usize] = map[idx].1;
            idx += 1;
        }

        let mut idx = 0;
        while idx < res.len() {
            let is_set = res[idx] != 0;
            let _assert= [()][!is_set as usize];
            idx += 1;
        }

        res
    };
    MAPPING[f as usize]
}
```

Alternatively, we might want to just make protocol version feature's discriminant?

```rust
#[derive(Clone, Copy)]
#[repr(u32)]
pub enum Feature {
    A = 100,
    #[cfg(feature = "b")]
    B = 200,
}
```

Taking a step back, I think that:

* the current API for feature checking feels awkward, and probably can be improved in terms of ergonomics, runtime preformance, and runtime semantics
* there's a large benefit in uniformity (even if it results in not the best possible code)
* we probably should look into cleaning this up in a separate PR

matklad:
Drive by procedural note: I've noticed that this is a second PR which make something versionable, and both PRs consist of two parts:

* introducing getters everywhere and changing call-sites (the bulk of changes)
* actually changing some struct definitions and adding the corresponding protocol feature. 

On the first blush (and, as I haven't done similar work, I might be very wrong here), it seems that it would be beneficial to split the two bits into separate PRs. The PR which just introduces getters can be reviewed relatively quicker, so that one doesn't have to deal with merege conflicts when merging relatively more involved second part.

core/primitives/src/version.rs
bowenwang1996: Yeah I agree with what @matklad said. Let's just follow the current way of doing the protocol version check in this PR and then we can address the issue separately. Created #4171

core/primitives/src/epoch_manager.rs
```diff
@@ -49,117 +40,977 @@ pub struct EpochConfig {
...
+        /// Validators slashed since the start of epoch or in previous epoch.
+        pub slashed: HashMap<AccountId, SlashState>,
+        /// Total supply at this block.
+        pub total_supply: Balance,
+        #[cfg(feature = "protocol_feature_rectify_inflation")]
+        pub timestamp_nanosec: u64,
+    }
+
+    #[derive(Default, BorshSerialize, BorshDeserialize, Serialize, Eq, PartialEq, Clone, Debug)]
+    pub struct BlockInfoV1 {
```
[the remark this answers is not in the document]
birchmd: Done in https://github.com/near/nearcore/pull/4115/commits/9db73bf9f9c41c8c40a6785dc3a49b69914f844f

core/primitives/src/epoch_manager.rs
```diff
@@ -49,117 +40,977 @@ pub struct EpochConfig {
...
+    impl Default for EpochInfo {
+        fn default() -> Self {
+            Self::V2(EpochInfoV2::default())
+        }
+    }
+
+    #[derive(
+        SmartDefault, BorshSerialize, BorshDeserialize, Serialize, Clone, Debug, PartialEq, Eq,
+    )]
+    pub struct EpochInfoV1 {
```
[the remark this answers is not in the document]
birchmd: Done in https://github.com/near/nearcore/pull/4115/commits/9db73bf9f9c41c8c40a6785dc3a49b69914f844f

core/primitives/src/types.rs
```diff
@@ -422,39 +669,197 @@ pub struct BlockExtra {
...
+    use near_primitives_core::hash::CryptoHash;
+    use near_primitives_core::types::{Balance, Gas};
+    use serde::Serialize;
+
+    /// Information after chunk was processed, used to produce or check next chunk.
+    #[derive(Debug, PartialEq, BorshSerialize, BorshDeserialize, Serialize, Clone, Eq)]
+    pub enum ChunkExtra {
+        V1(ChunkExtraV1),
+        V2(ChunkExtraV2),
+    }
```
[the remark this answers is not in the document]
birchmd: I removed `Serialize` from most structs in https://github.com/near/nearcore/pull/4115/commits/f9be8cb1cf5adfa198f86d8df78b5a78d335d51b

`ValidatorStake` was left because its serialization is used (at least removing `Serialize` made the compiler unhappy). But I added the tagged attribute same as in `ValidatorStakeView`.

core/primitives/src/types.rs
birchmd: Yes, reusing the iterator is a good idea. Done in https://github.com/near/nearcore/pull/4115/commits/14273cdea40d52322e38656b4be88ce9f1245caa

I kept the `len` field though because there are places in the code where knowing the number of elements without consuming the iterator is important.

core/primitives/src/epoch_manager.rs
birchmd: Yes, good catch. Added missing inlines in https://github.com/near/nearcore/pull/4115/commits/0af94dcc7eafef6304c2612f0aea35e2f29bb3b0

chain/chain/src/validate.rs
birchmd: I don't think implementing `PartialEq` makes sense because the `eq` method signature uses `&self`, whereas to access elements in the iterator (via `all` or `next`) I need `&mut self`. This is also the only place where we compare two of these objects, so it's not saving much to implement the logic elsewhere.

core/primitives/src/views.rs
birchmd: Name changed to `validator_stake_struct_version` in https://github.com/near/nearcore/pull/4115/commits/b288f74bf34b0a8aa50da75ee9a3698179259623

core/primitives/src/utils.rs
```diff
@@ -69,6 +69,35 @@ impl<T: Sized> Deref for MaybeValidated<T> {
...
 
+/// Similar to https://doc.rust-lang.org/std/iter/struct.Zip.html
+/// but it does not short-circuit when one of the iterators returns None,
+/// instead it will only stop when _both_ iterators return None.
+/// Notably, this can be used to lazily discover if two iterators
+/// contain a different number of elements.
+pub struct FullZip<A: Iterator, B: Iterator> {
+    iter_a: A,
+    iter_b: B,
+}
```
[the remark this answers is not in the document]
birchmd: Done in https://github.com/near/nearcore/pull/4115/commits/de0a80aa2b2d54c84f15a2e6d10c80cbe7d59e36

core/primitives/src/version.rs
birchmd: Sounds good @bowenwang1996 

Changed to using the mapping directly in https://github.com/near/nearcore/pull/4115/commits/a21a48b96f40be81449cfbacf4e961d590ef6240

chain/chain/src/validate.rs
matklad: You might make use of https://doc.rust-lang.org/stable/std/iter/trait.Iterator.html#method.eq though I think.

birchmd: Ah, yes, that is nicer! Done in https://github.com/near/nearcore/pull/4115/commits/321c4cb9569ade9daf2dabf3ba52eb465c9760f1

bowenwang1996 (approved):

PR #4116, opened 2021-03-15
fix(runtime): disallow deletion of account with large state

bowenwang1996:
For account whose state is large, deleting them is not cheap and this PR prevents accounts with more than 10kb of state (excluding contracts) from being deleted. Contract developers should develop their own logic to delete part of the state if needed.

Test plan
---------
* `test_delete_account_too_large`
* `test_delete_account_with_contract_and_small_state`
* `test_delete_account_with_contract_and_large_state`
*  nayduck


runtime/runtime/src/actions.rs
```diff
@@ -445,7 +446,28 @@ pub(crate) fn action_delete_account(
...
+        >= STABLE_PROTOCOL_FEATURES_TO_VERSION_MAPPING[&ProtocolFeature::DeleteActionRestriction]
+    {
+        let account = account.as_ref().unwrap();
+        let mut account_storage_usage = account.storage_usage;
+        let contract_code = get_code(state_update, account_id, Some(account.code_hash))?;
+        if let Some(code) = contract_code {
+            // account storage usage should be larger than code size
+            let code_len = code.code.len() as u64;
+            debug_assert!(account_storage_usage > code_len);
+            account_storage_usage -= code_len;
```
evgenykuzyakov, line 461: Considering previous issue with state size due to delete key action, we should use `saturated_sub` to be safe

core/primitives/src/version.rs
```diff
@@ -94,21 +94,23 @@ pub enum ProtocolFeature {
...
-pub const PROTOCOL_VERSION: ProtocolVersion = 42;
+pub const PROTOCOL_VERSION: ProtocolVersion = 43;
 
 /// Current latest nightly version of the protocol.
 #[cfg(feature = "nightly_protocol")]
 pub const PROTOCOL_VERSION: ProtocolVersion = 106;
 
 lazy_static! {
-    static ref STABLE_PROTOCOL_FEATURES_TO_VERSION_MAPPING: HashMap<ProtocolFeature, ProtocolVersion> =
+    pub static ref STABLE_PROTOCOL_FEATURES_TO_VERSION_MAPPING: HashMap<ProtocolFeature, ProtocolVersion> =
```
evgenykuzyakov, line 109: This is not needed. nightly merges stable features in

runtime/runtime/src/actions.rs
```diff
@@ -445,7 +446,28 @@ pub(crate) fn action_delete_account(
     result: &mut ActionResult,
     account_id: &AccountId,
     delete_account: &DeleteAccountAction,
+    current_protocol_version: ProtocolVersion,
 ) -> Result<(), StorageError> {
+    if current_protocol_version
+        >= STABLE_PROTOCOL_FEATURES_TO_VERSION_MAPPING[&ProtocolFeature::DeleteActionRestriction]
```
evgenykuzyakov, line 452:
```suggestion
        >= PROTOCOL_FEATURES_TO_VERSION_MAPPING[&ProtocolFeature::DeleteActionRestriction]
```

core/primitives/src/version.rs
bowenwang1996: Good point.

core/primitives/src/errors.rs
```diff
@@ -677,7 +679,8 @@ impl Display for ActionErrorKind {
                 write!(f, "An new action receipt created during a FunctionCall is not valid: {}", e)
             }
             ActionErrorKind::InsufficientStake { account_id, stake, minimum_stake } => write!(f, "Account {} tries to stake {} but minimum required stake is {}", account_id, stake, minimum_stake),
-            ActionErrorKind::OnlyImplicitAccountCreationAllowed { account_id } => write!(f, "CreateAccount action is called on hex-characters account of length 64 {}", account_id)
+            ActionErrorKind::OnlyImplicitAccountCreationAllowed { account_id } => write!(f, "CreateAccount action is called on hex-characters account of length 64 {}", account_id),
+            ActionErrorKind::DeleteAccountWithLargeState { account_id } => write!(f, "The state of account {} is too large and therefore cannot be deleted", account_id)
```
matklad, line 683:
```suggestion
            ActionErrorKind::DeleteAccountWithLargeState { account_id } => write!(f, "The state of account {} is too large and therefore cannot be deleted", account_id),
```

neard/res/genesis_config.json
```diff
@@ -1,239 +1,239 @@
 {
-  "protocol_version": 42,
-  "genesis_time": "1970-01-01T00:00:00.000000000Z",
-  "chain_id": "sample",
-  "genesis_height": 0,
-  "num_block_producer_seats": 50,
+  "protocol_version": 43, 
+  "genesis_time": "1970-01-01T00:00:00.000000000Z", 
+  "chain_id": "sample",
```
matklad, line 4: trailing ws?

runtime/runtime/src/actions.rs
```diff
@@ -445,7 +445,28 @@ pub(crate) fn action_delete_account(
     result: &mut ActionResult,
     account_id: &AccountId,
     delete_account: &DeleteAccountAction,
+    current_protocol_version: ProtocolVersion,
 ) -> Result<(), StorageError> {
+    if current_protocol_version
+        >= PROTOCOL_FEATURES_TO_VERSION_MAPPING[&ProtocolFeature::DeleteActionRestriction]
```
matklad, line 451: unrelated to PR, but feels like we can have a nicer API: `current_protocol_version.has_feature(ProtocolFeature::DeleteActionRestriction)`

neard/res/genesis_config.json
bowenwang1996: hmm this file is auto generated

runtime/runtime/src/actions.rs
bowenwang1996: That is a good idea!

runtime/runtime/src/actions.rs
```diff
@@ -458,7 +457,7 @@ pub(crate) fn action_delete_account(
             // account storage usage should be larger than code size
             let code_len = code.code.len() as u64;
             debug_assert!(account_storage_usage > code_len);
```
evgenykuzyakov, line 459: We don't run `debug_assert` in prod, right? So it's only for tests

evgenykuzyakov (approved):

runtime/runtime/src/actions.rs
bowenwang1996: yes

PR #4117, opened 2021-03-16
fix(test): fix test_rejoin

mikhailOK:
Fix the tests that run neard binary in nayduck.

Test plan
---------
the test works again

### DIFF
--- a/test-utils/testlib/src/node/process_node.rs
+++ b/test-utils/testlib/src/node/process_node.rs
@@ -47,6 +47,7 @@ impl Node for ProcessNode {
     fn start(&mut self) {
         match self.state {
             ProcessNodeState::Stopped => {
+                std::env::set_var("ADVERSARY_CONSENT", "1");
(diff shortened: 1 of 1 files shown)


test-utils/testlib/src/node/process_node.rs
```diff
@@ -47,6 +47,7 @@ impl Node for ProcessNode {
     fn start(&mut self) {
         match self.state {
             ProcessNodeState::Stopped => {
+                std::env::set_var("ADVERSARY_CONSENT", "1");
```
bowenwang1996, line 50: What is this for?

bowenwang1996 (approved):

test-utils/testlib/src/node/process_node.rs
mikhailOK: nayduck builds the binary with adversarial features, this is required to run it

Issue #4063, opened 2021-03-09
state_sync.py: Unreachable error occurred


mikhailOK:
http://nayduck.eastus.cloudapp.azure.com:3000/#/test/104936
```
[2mMar 09 17:12:50.914[0m [34mDEBUG[0m chain: Verifying challenges []    
[2mMar 09 17:12:50.914[0m [33m WARN[0m jsonrpc: Unreachable error occurred: DB Not Found Error: CHUNK EXTRA: 8veiAXzFM9bBvFotemJDzpcUzj3gRPmuN1y2WipvQJ4s:0 
 Cause: Unknown
thread 'actix-rt|system:0|arbiter:0' panicked at 'called `Result::unwrap()` on an `Err` value: InconsistentCardinality { expect: 1, got: 2 }', /home/azureuser/.cargo/registry/src/github.com-1ecc6299db9ec823/prometheus-0.11.0/src/vec.rs:258:49
stack backtrace:
[2mMar 09 17:12:50.914[0m [34mDEBUG[0m chain: Catching up: removing prev=`8veiAXzFM9bBvFotemJDzpcUzj3gRPmuN1y2WipvQJ4s` from the queue. I'm Some("test0")    
[2mMar 09 17:12:50.915[0m [34mDEBUG[0m chain: Check orphans: from 8veiAXzFM9bBvFotemJDzpcUzj3gRPmuN1y2WipvQJ4s, # orphans 0    
   0: rust_begin_unwind
             at /rustc/91a79fb29ac78d057d04dbe86be13d5dcc64309a/library/std/src/panicking.rs:483
   1: core::panicking::panic_fmt
             at /rustc/91a79fb29ac78d057d04dbe86be13d5dcc64309a/library/core/src/panicking.rs:85
   2: core::option::expect_none_failed
             at /rustc/91a79fb29ac78d057d04dbe86be13d5dcc64309a/library/core/src/option.rs:1234
   3: core::result::Result<T,E>::unwrap
             at /rustc/91a79fb29ac78d057d04dbe86be13d5dcc64309a/library/core/src/result.rs:973
   4: prometheus::vec::MetricVec<T>::with_label_values
             at /home/azureuser/.cargo/registry/src/github.com-1ecc6299db9ec823/prometheus-0.11.0/src/vec.rs:258
   5: near_metrics::inc_counter_vec
```
Failing after #3944

mikhailOK:
[...]

PR #4119, opened 2021-03-16
fix(jsonrpc): Fix some errors inconsistency with unknown blocks or untracked shards for handle_query view_client method

khorolets:
This PR improves errors in `handle_query` method after refactoring. Solves some problems found in the issue #4063


bowenwang1996 (approved):

chain/jsonrpc/src/lib.rs
```diff
@@ -212,6 +212,24 @@ fn process_query_response(
                 "block_height": block_height,
                 "block_hash": block_hash,
             })),
+            near_jsonrpc_primitives::types::query::RpcQueryError::UnknownBlock {
+                ref block_reference,
+            } => match block_reference {
+                near_primitives::types::BlockReference::BlockId(block_id) => Err(RpcError::new(
+                    -32_000,
```
chefsale, line 219: Can we make this status code maybe a 
```
const RPC_ERROR_CODE:i32 = -32_200; 
```
with the name a bit more descriptive as I don't really know what does the number represent? 😄

frol: @chefsale Well, this function documentation goes the following:

```
/// This function processes response from query method to introduce
/// backward compatible response in case of specific errors
```

This number is mostly useless and means nothing, so we just keep it there for compatibility reasons.

@khorolets you may probably provide a reference to the `RpcError::server_error` method in a comment.

frol (approved):

PR #4120, opened 2021-03-16
refactor(pytest): Update sanity/transactions.py test to pass according to dropped routing in handle_query ViewClient method

khorolets:
In merged PR #3944 I've dropped routing for the `Query` handler in ViewClient (@frol and @bowenwang1996 discussed it and approved this change).

After that test `pytest/tests/sanity/transactions.py` started failing. @frol and I investigated it and found out that the fails related to the attempts to test transactions between shards, all the test relies on the querying jsonrpc `query` method to get accounts from different shards and check their balances, but since routing is not present there anymore it's impossible to do it.

In this PR I've made the smallest change introducing only one shard and all the accounts on that shard so the test is not failing anymore, but I'm not sure if it not brakes the main idea of the test.

@bowenwang1996 @SkidanovAlex could you assist in refactoring this test so it could still be useful by checking what it supposed to be checking, please?

### DIFF
--- a/pytest/tests/sanity/transactions.py
+++ b/pytest/tests/sanity/transactions.py
@@ -16,8 +16,8 @@ TIMEOUT = 240
 
 nodes = start_cluster(
     num_nodes=4,
-    num_observers=0,
-    num_shards=1,
+    num_observers=1,
+    num_shards=4,
     config=None,
     genesis_config_changes=[
         ["min_gas_price", 0],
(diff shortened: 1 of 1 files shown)


khorolets:
@mikhailOK I think you might want to follow this, I see you are fixing some tests maybe you have some knowledge to help here?

bowenwang1996 (commented):
The number of shards should not change. You can add an observer node that tracks all shards and do rpc query from it. I can do this if needed

khorolets:
> The number of shards should not change. You can add an observer node that tracks all shards and do rpc query from it. I can do this if needed

Would be very helpful as I don't know what to do :(

frol (approved):

bowenwang1996:
30 runs all pass http://nayduck.eastus.cloudapp.azure.com:3000/#/run/1348

bowenwang1996 (approved):

PR #4121, opened 2021-03-16
Switch all script to use /usr/bin/env bash as /bin/bash is not portable

chefsale:
cc: @pmnoxx @matklad

### DIFF
--- a/chain/jsonrpc/build_errors_schema.sh
+++ b/chain/jsonrpc/build_errors_schema.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 rm ../../target/rpc_errors_schema.json
 cargo build --features dump_errors_schema
 cp ../../target/rpc_errors_schema.json ./res/rpc_errors_schema.json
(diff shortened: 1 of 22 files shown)


matklad (approved):
That's a lot of bash scripts!

matklad:
JFYI, in rust-analyzer we avoided bash scritps and instead did all task automation in Rust, [using this trick](https://github.com/matklad/cargo-xtask).

chefsale:
> JFYI, in rust-analyzer we avoided bash scritps and instead did all task automation in Rust, [using this trick](https://github.com/matklad/cargo-xtask).

That would be a bit more work :)

bowenwang1996 (approved):

chefsale:
cc: @ailisp @frol

pmnoxx (approved):

ailisp:
Looks good! I read somewhere, some os has env only in /bin/env, but should be good for most commonly seen OSs :D

willemneal:
Given our reliance on Bash, perhaps we should add the following to CI to see what it says about the scripts: https://github.com/koalaman/shellcheck

willemneal (approved):

chefsale:
@willemneal this is cool, i will look into integrating this 🥇

PR #4122, opened 2021-03-17
fix(test): rpc_state_changes access key nonce

mikhailOK:
Update expected results with protocol_feature_access_key_nonce_range.

Test plan
---------
nayduck


pytest/tests/sanity/rpc_state_changes.py
```diff
@@ -155,11 +158,12 @@ def test_changes_with_new_account_with_access_key():
     # Step 3
     status = nodes[0].get_status()
     latest_block_hash = status['sync_info']['latest_block_hash']
+    nonce += 8
```
bowenwang1996, line 161: why 8?

mikhailOK: keep test exactly the same without nightly_protocol, it had `nonce=8` below.
`+=1` passes too

bowenwang1996 (approved):

PR #4124, opened 2021-03-17
refactor(jsonrpc): Refactor gas_price method to be typed and errors structured

chefsale:


chain/client/src/view_client.rs
```diff
@@ -1203,14 +1203,16 @@ impl Handler<NetworkViewClientMessages> for ViewClientActor {
...
 
     #[perf]
     fn handle(&mut self, msg: GetGasPrice, _ctx: &mut Self::Context) -> Self::Result {
         let header = self
             .maybe_block_id_to_block_hash(msg.block_id)
             .and_then(|block_hash| self.chain.get_block_header(&block_hash));
-        header.map(|b| GasPriceView { gas_price: b.gas_price() }).map_err(|e| e.to_string())
+        header
+            .map(|b| GasPriceView { gas_price: b.gas_price() })
+            .map_err(|e| GetGasPriceError::from(e))
```
frol, line 1215: Given we now have the implementation for `From<near_chain_primitives::Error> for GetGasPriceError`, we can use `?` operator, and simplify the code to the following:

```suggestion
        Ok(GasPriceView { gas_price: header?.gas_price() })
```

frol (approved):
Perfecto!

PR #4125, opened 2021-03-17
fix: fix compilation of params estimator in release mode

matklad:
lazy_static_include_bytes includes resources either at compile time or
at runtime, depending on the value of `debug_assertions` cfg value. This
is useful if you want to ship a single binary to the users with all
resources embedded, but want to develop locally with faster compile
times.

However, our "resources" are just test contracts, so we can just always
load them at run time.

Bimodal loading is problematic, because it can create compile-time
errors *only* if `--release` is passed.


runtime/runtime-params-estimator/src/evm_estimator.rs
```diff
@@ -209,10 +209,7 @@ pub fn measure_evm_deploy(
 use_contract!(soltest, "../near-evm-runner/tests/build/SolTests.abi");
 use_contract!(precompiled_function, "../near-evm-runner/tests/build/PrecompiledFunction.abi");
 
-lazy_static_include_str! {
-    TEST => "../near-evm-runner/tests/build/SolTests.bin",
-    PRECOMPILED_TEST => "../near-evm-runner/tests/build/PrecompiledFunction.bin",
```
matklad, line 214: Note that this one was genuinely unused, but the macro silenced the warning.

willemneal (approved):

runtime/runtime-params-estimator/src/lib.rs
```diff
@@ -15,3 +15,35 @@ pub mod testbed;
...
+        TestContract { path, cell: OnceCell::new() }
+    }
+}
+
+impl std::ops::Deref for TestContract {
+    type Target = [u8];
+
+    fn deref(&self) -> &[u8] {
+        self.cell
+            .get_or_init(|| {
```
willemneal, line 40: Does this get called every time?  Or does `OnceCell` ensure that it's only loaded once?

ailisp (approved):

joshuajbouw:
This PR may have broken the script since we do not have `CARGO_MANIFEST_DIR` as it isn't ran with `cargo`.

PR #4126, opened 2021-03-17
feat: infrastructure for reusing memory and store between contracts

olonho:
feat: infrastructure for reusing memory and store between contracts

To prepare for optimized pipelined execution of contracts we need to be able to share reusable parts of state.
This is a preparation PR allowing to share certain parts of the VM state between runs. By default, memory state is turned 
off, as we need to ensure that no memory data of one contract execution can be seen in another execution.
Once https://github.com/near/nearcore/pull/4128 will be merged, we could test performance impact of different
sharing practises.

Test plan
---------

cargo test
more specifically tests::contract_preload::test_run_preloaded which tests the behavior of the preloader.

### DIFF
--- a/runtime/near-vm-runner/src/preload.rs
+++ b/runtime/near-vm-runner/src/preload.rs
@@ -58,7 +58,7 @@ pub struct ContractCaller {
     vm_config: VMConfig,
     vm_data_private: VMDataPrivate,
     vm_data_shared: VMDataShared,
-    prepared: Vec<CallInner>,
+    preloaded: Vec<CallInner>,
 }
 
 impl ContractCaller {
@@ -105,7 +105,7 @@ impl ContractCaller {
             vm_config,
             vm_data_private: private,
             vm_data_shared: shared,
-            prepared: Vec::new(),
+            preloaded: Vec::new(),
         }
     }
 
@@ -115,15 +115,15 @@ impl ContractCaller {
     ) -> Vec<ContractCallPrepareResult> {
         let mut result: Vec<ContractCallPrepareResult> = Vec::new();
         for request in requests {
-            let index = self.prepared.len();
+            let index = self.preloaded.len();
             let (tx, rx) = channel();
-            self.prepared.push(CallInner { rx });
+            self.preloaded.push(CallInner { rx });
             self.pool.execute({
                 let tx = tx.clone();
                 let vm_config = self.vm_config.clone();
                 let vm_data_shared = self.vm_data_shared.clone();
                 let vm_kind = self.vm_kind.clone();
-                move || prepare_in_thread(request, vm_kind, vm_config, vm_data_shared, tx)
+                move || preload_in_thread(request, vm_kind, vm_config, vm_data_shared, tx)
             });
             result.push(ContractCallPrepareResult { handle: index });
         }
@@ -132,7 +132,7 @@ impl ContractCaller {
 
     pub fn run_preloaded<'a>(
         self: &mut ContractCaller,
-        prepared: &ContractCallPrepareResult,
+        preloaded: &ContractCallPrepareResult,
         method_name: &str,
         ext: &mut dyn External,
         context: VMContext,
@@ -141,7 +141,7 @@ impl ContractCaller {
         current_protocol_version: ProtocolVersion,
         profile: ProfileData,
     ) -> (Option<VMOutcome>, Option<VMError>) {
-        match self.prepared.get(prepared.handle) {
+        match self.preloaded.get(preloaded.handle) {
             Some(call) => {
                 let call_data = call.rx.recv().unwrap();
                 match call_data.result {
@@ -222,7 +222,7 @@ impl Drop for ContractCaller {
     }
 }
 
-fn prepare_in_thread(
+fn preload_in_thread(
     request: ContractCallPrepareRequest,
     vm_kind: VMKind,
     vm_config: VMConfig,
--- a/runtime/near-vm-runner/src/preload.rs
+++ b/runtime/near-vm-runner/src/preload.rs
@@ -59,7 +59,12 @@ pub struct ContractCaller {
 
 impl ContractCaller {
     pub fn new(num_threads: usize) -> ContractCaller {
-        ContractCaller { pool: ThreadPool::new(num_threads), vm_data_private: None, vm_data_shared: None, prepared: Vec::new() }
+        ContractCaller {
+            pool: ThreadPool::new(num_threads),
+            vm_data_private: None,
+            vm_data_shared: None,
+            prepared: Vec::new(),
+        }
     }
 
     pub fn preload(
@@ -78,9 +83,9 @@ impl ContractCaller {
                             vm_config.limit_config.initial_memory_pages,
                             vm_config.limit_config.max_memory_pages,
                         )
-                            .unwrap(),
+                        .unwrap(),
                     ));
-                },
+                }
                 VMKind::Wasmer1 => {
                     let store = default_wasmer1_store();
                     let store_clone = store.clone();
@@ -90,7 +95,9 @@ impl ContractCaller {
                             &store_clone,
                             vm_config.limit_config.initial_memory_pages,
                             vm_config.limit_config.max_memory_pages,
-                        ).unwrap()));
+                        )
+                        .unwrap(),
+                    ));
                 }
                 _ => panic!("Not currently supported"),
             };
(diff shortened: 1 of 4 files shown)


runtime/near-vm-runner/src/wasmer1_runner.rs
```diff
@@ -286,6 +286,7 @@ pub(crate) fn default_wasmer1_store() -> Store {
 pub(crate) fn run_wasmer1_module<'a>(
     module: &Module,
     store: &Store,
+    memory: &mut Wasmer1Memory,
```
matklad, line 289: Wasmer memory is ref-counted object, so passing an exclusive reference to shared data doesn't make much sens. I'd make this accept memory by value. That should also simplify the calling code, as that no longer needs to be `&mut`

runtime/near-vm-runner/src/preload.rs
```diff
@@ -91,29 +140,48 @@ impl ContractCaller {
...
-                            ext,
-                            context,
-                            vm_config,
-                            fees_config,
-                            promise_results,
-                            profile,
-                            current_protocol_version,
-                        ),
+                        Wasmer0(module) => match &mut self.vm_data_private {
+                            Some(data) => match data {
```
matklad, line 144: I think you can match on tupples here:

```
match (module, &mut self.vm_data_private) {
  (Wasmer0(module), VMDataPrivate::Wasmer0(memory)) => { ... }
  (Wasmer1(module), VMDataPrivate::Wasmer1(memory)) => { ... }
  _ => panic!(),
}
```

I don't think there's a super-good way to remove the unreachable state altogether here.

runtime/near-vm-runner/src/wasmer1_runner.rs
ailisp: I think this PR would be better to rebase on #4127 now. Also, we should have a benchmark on this PR to see reuse actually saving non-trivial time

runtime/near-vm-runner/src/preload.rs
```diff
@@ -12,13 +12,25 @@ use near_vm_logic::types::PromiseResult;
...
+
+enum VMDataPrivate {
+    Wasmer0(WasmerMemory),
+    Wasmer1(Wasmer1Memory),
+}
+
+#[derive(Clone)]
+enum VMDataShared {
+    Wasmer0,
+    Wasmer1(wasmer::Store),
```
ailisp, line 33: I feel it can be problematic to share store between instances. From this doc: https://docs.rs/wasmer/1.0.2/wasmer/struct.Store.html:

> The store represents all global state that can be manipulated by WebAssembly programs. It consists of the runtime representation of all instances of functions, tables, memories, and globals that have been allocated during the lifetime of the abstract machine.

Would there be name conflictions, and if not clean up store in each run, could it grow infinitely?

olonho: Note that intended usage of `ContractCaller` is the timespan of one block, it shall not be reused between blocks. What shall be reused need to be decided based on actual perf data, agree with that. Current infrastructure allows rather flexible selection of what is shared and what is recreated, that's the point of `vm_data_shared` and `vm_data_private`.

bowenwang1996: @olonho do we cache any `ContractCallPrepareResult`?

olonho: No, it's pretty transient, like handle.

runtime/near-vm-runner/src/preload.rs
```diff
@@ -41,27 +54,76 @@ pub struct ContractCallPrepareResult {
...
+        let (shared, private) = match vm_kind {
+            VMKind::Wasmer0 => (
+                VMDataShared::Wasmer0,
+                VMDataPrivate::Wasmer0(if SHARE_MEMORY_INSTANCE {
+                    Some(
+                        WasmerMemory::new(
+                            vm_config.limit_config.initial_memory_pages,
+                            vm_config.limit_config.max_memory_pages,
+                        )
+                        .unwrap(),
```
matklad, line 75: (unrelated to PR per se) urgh, we should pull these unwraps up, into `WasmerMemory::new`. They indeed can not fail as far as I can see, but we should assert that once.

runtime/near-vm-runner/src/preload.rs
```diff
@@ -41,27 +54,76 @@ pub struct ContractCallPrepareResult {
...
+                                vm_config.limit_config.max_memory_pages,
+                            )
+                            .unwrap(),
+                        )
+                    } else {
+                        None
+                    }),
+                )
+            }
+            _ => panic!("Not currently supported"),
```
matklad, line 100:
```suggestion
            VMKind::Wasmer => panic!("Not currently supported"),
```

This way, if we add a new `VMKind` like `Wasmer2`, we'll get a compile time error, which would give us a chance to consider if  we should panic or if we should just implement this. 

In general, `_` top-level patters should be avoided.

runtime/near-vm-runner/src/preload.rs
```diff
@@ -12,13 +12,27 @@ use near_vm_logic::types::PromiseResult;
...
+    Wasmer1(wasmer::Module),
+}
+
+enum VMDataPrivate {
+    Wasmer0(Option<WasmerMemory>),
+    Wasmer1(Option<Wasmer1Memory>),
+}
+
+#[derive(Clone)]
+enum VMDataShared {
```
matklad, line 33: An alternative is to "transpose" these two:

```rust
enum VMData {
  Wasmer0 { private: Option<WasmerMemory>, shared: () }
  Wasmer1 { private: Option<Wasmer1Memory>, shared: wasmer::Store }
}
```

runtime/near-vm-runner/src/preload.rs
```diff
@@ -70,46 +132,83 @@ impl ContractCaller {
...
+                                    if SHARE_MEMORY_INSTANCE {
+                                        memory.as_mut().unwrap()
+                                    } else {
+                                        new_memory = WasmerMemory::new(
+                                            self.vm_config.limit_config.initial_memory_pages,
+                                            self.vm_config.limit_config.max_memory_pages,
+                                        )
+                                        .unwrap();
+                                        &mut new_memory
+                                    },
```
matklad, line 168: It feels like we do a the same check twice here: `SHARE_MEMORY_INSTANCE` is true iff `memory.as_mut()` is not None. So perhaps use this instead?

```suggestion
                                    module.clone(),
                                    memory.as_mut().unwrap_or_else(|| {
                                        new_memory = WasmerMemory::new(
                                            self.vm_config.limit_config.initial_memory_pages,
                                            self.vm_config.limit_config.max_memory_pages,
                                        )
                                        .unwrap();
                                        &mut new_memory
                                    }),
```

matklad (approved):
Left a couple of style nits, but LGTM as is!

runtime/near-vm-runner/src/preload.rs
olonho: Well, for me it was rather important to distinguish shared vs. non-shared state, as they have different concurrent access policies.

runtime/near-vm-runner/src/preload.rs
olonho: Ack.

runtime/near-vm-runner/src/preload.rs
olonho: Interesting question, see benefits in either approach.

runtime/near-vm-runner/src/preload.rs
olonho: Indeed, somewhat reworked this place.

SkidanovAlex:
If this is intended to be pushed, the commit message needs to follow the template from CONTRIBUTING.md (in particular, include a description of what it does and how it was tested).

runtime/near-vm-runner/src/preload.rs
```diff
@@ -41,27 +54,76 @@ pub struct ContractCallPrepareResult {
...
 impl ContractCaller {
-    pub fn new(num_threads: usize) -> ContractCaller {
-        ContractCaller { pool: ThreadPool::new(num_threads), prepared: Vec::new() }
+    pub fn new(num_threads: usize, vm_kind: VMKind, vm_config: VMConfig) -> ContractCaller {
+        let (shared, private) = match vm_kind {
+            VMKind::Wasmer0 => (
+                VMDataShared::Wasmer0,
+                VMDataPrivate::Wasmer0(if SHARE_MEMORY_INSTANCE {
+                    Some(
+                        WasmerMemory::new(
```
ailisp, line 71: This looks confusing, `private = Some(WasmerMemory::new(...))` when it's `SHARE_MEMORY_INSTANCE`, and when not `private = None` 🤔

runtime/near-vm-runner/src/preload.rs
```diff
@@ -41,27 +54,76 @@ pub struct ContractCallPrepareResult {
...
         requests: Vec<ContractCallPrepareRequest>,
-        vm_kind: VMKind,
     ) -> Vec<ContractCallPrepareResult> {
         let mut result: Vec<ContractCallPrepareResult> = Vec::new();
         for request in requests {
-            let index = self.prepared.len();
+            let index = self.preloaded.len();
             let (tx, rx) = channel();
-            self.prepared.push(CallInner { rx });
+            self.preloaded.push(CallInner { rx });
```
ailisp, line 120: preloaded is better name than prepared 👍 preload = prepare + compile or deserialize

ailisp (approved):
Looks good, although the one naming maybe changed to be more intuitive

runtime/near-vm-runner/src/preload.rs
olonho: Here `private` means `not-shared-between-threads`.

PR #4127, opened 2021-03-17
refactor: prepare to share resources between contract executions

matklad:
This commit introduces a `WasmerMachine` -- an entity which is created
once at the start of the block through which the contracts are executed.

The goal here is to re-use VM resources between contract executions. At
minimum, we would want to re-use virtual memory mapping crated by wasm
runtimes. But we might want to introduce bigger bits of state as well,
for, for example, contract prefetching.

Another way to look at it is that contract execution requires to pay certain fixed-costs.
By sharing the WasmerMachine, we pay these costs once per apply, and not
once per contract call.

### DIFF
--- a/runtime/near-vm-runner/src/runner.rs
+++ b/runtime/near-vm-runner/src/runner.rs
@@ -8,6 +8,8 @@ use near_vm_errors::{CompilationError, FunctionCallError, VMError};
 use near_vm_logic::types::PromiseResult;
 use near_vm_logic::{External, VMContext, VMKind, VMOutcome};
 
+/// Runs a single contract
+///
 /// `run` does the following:
 /// - deserializes and validate the `code` binary (see `prepare::prepare_contract`)
 /// - injects gas counting into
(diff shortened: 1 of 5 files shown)


runtime/near-vm-runner/src/runner.rs
```diff
@@ -8,43 +10,53 @@ use near_vm_errors::{CompilationError, FunctionCallError, VMError};
...
+}
+
+impl WasmMachine {
+    pub fn new(vm_config: VMConfig, cache: Option<Arc<dyn CompiledContractCache>>) -> WasmMachine {
+        WasmMachine { vm_config, cache }
+    }
+
+    /// `run` does the following:
+    /// - deserializes and validate the `code` binary (see `prepare::prepare_contract`)
+    /// - injects gas counting into
```
willemneal, line 25:
```suggestion
    /// - injects gas counting into binary
```

willemneal (approved):
LGTM

runtime/near-vm-runner/src/runner.rs
```diff
@@ -8,43 +10,53 @@ use near_vm_errors::{CompilationError, FunctionCallError, VMError};
...
+        code: &ContractCode,
+        method_name: &str,
+        ext: &mut dyn External,
+        context: VMContext,
+        fees_config: &RuntimeFeesConfig,
+        promise_results: &[PromiseResult],
+        current_protocol_version: ProtocolVersion,
+        profile: &ProfileData,
+    ) -> (Option<VMOutcome>, Option<VMError>) {
+        run_vm(
```
ailisp, line 46: > The goal here is to re-use VM resources between contract executions. At
minimum, we would want to re-use virtual memory mapping crated by wasm
runtimes. 

This PR so far looks good. However `machine.run` still doesn't reuse wasmer instance and wasmer memory, which is the major thing common to any contract execution, that should consider storing them in machine to reuse. And I think we should do a quick benchmark to see how fast is `virtual memory mapping crated by wasm
runtimes` and init wasmer/wasmtime instance, if it's quicker than cleanup existing wasm memory modules, or if it's trivial as bottleneck, then `WasmerMachine {vm_config,cache}` in this PR is sufficient

ailisp (approved):

runtime/near-vm-runner/src/runner.rs
matklad: Yeah, this PR intentionally avoids implementing any actual re-use, to keep it smaller and to just establish the API.

bowenwang1996 (commented):
This PR doesn't change how the preparation is actually done right?

olonho:
As discussed offline, agreed on somewhat less stateful execution engine, while keeping some ability to share state.

runtime/near-vm-runner/src/runner.rs
```diff
@@ -47,6 +49,48 @@ pub fn run<'a>(
     )
 }
 
+/// Runs a batch of contracts.
+///
+/// Re-using the same `WasmMachine` is potentially more efficient than calling
+/// `run` repeatedly.
+pub struct WasmMachine {
+    _private: (),
```
olonho, line 57: Placeholder, right?

runtime/near-vm-runner/src/runner.rs
```diff
@@ -47,6 +49,48 @@ pub fn run<'a>(
...
+}
+
+impl WasmMachine {
```
olonho, line 60: `WasmMachine` is a bit misleading, maybe once we shall rename it. IMO, technically it's more like `ContractRunner` or `ContractExecutor`, isn't it?

olonho (approved):

PR #4135, opened 2021-03-18
fix(test): transactions.py with nightly_protocol

### LINKED ISSUE #4133: transactions.py with nightly_protocol
`pytest sanity/transactions.py --features nightly_protocol --features nightly_protocol_features`

http://nayduck.eastus.cloudapp.azure.com:3000/#/test/117583
```
  File "tests/sanity/transactions.py", line 81, in <module>
    sent_height, height)
AssertionError: Balances before: [999999999999999999999999999999773, 999999999999999999999999999999786, 1000000000000000000000000000000876, 999999999999999999999999999999429, 1000000000000000000000000000000136]
Expected balances: [999999999999999999999999999999245, 999999999999999999999999999999057, 1000000000000000000000000000001438, 999999999999999999999999999999813, 1000000000000000000000000000000447]
Current balances: [1000000000000000000000000000000187, 999999999999999999999999999999786, 1000000000000000000000000000000876, 999999999999999999999999999999439, 999999999999999999999999999999712]
Sent at height: 20, cur height: 31
```

mikhailOK:
Use get_epoch_block_approvers_ordered instead of get_epoch_block_producers_ordered.
Make some changes to the test: query the observer node, disable storage
checking and set a low state sync timeout.
Fixes #4133

Test plan
---------
nayduck

### DIFF
--- a/pytest/tests/sanity/transactions.py
+++ b/pytest/tests/sanity/transactions.py
@@ -25,7 +25,14 @@ nodes = start_cluster(
         ["epoch_length", 10],
         ["block_producer_kickout_threshold", 70]
     ],
-    client_config_changes={ 4: {"tracked_shards": [0, 1, 2, 3]}}
+    client_config_changes={
+                                  0: {"consensus": {"state_sync_timeout": {"secs": 2, "nanos": 0}}},
+                                  1: {"consensus": {"state_sync_timeout": {"secs": 2, "nanos": 0}}},
+                                  2: {"consensus": {"state_sync_timeout": {"secs": 2, "nanos": 0}}},
+                                  3: {"consensus": {"state_sync_timeout": {"secs": 2, "nanos": 0}}},
+                                  4: {"consensus": {"state_sync_timeout": {"secs": 2, "nanos": 0}},
+                                      "tracked_shards": [0, 1, 2, 3]}
+    }
 )
 
 started = time.time()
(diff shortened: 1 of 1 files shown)


mikhailOK:
test run: http://nayduck.eastus.cloudapp.azure.com:3000/#/run/1375

chain/chunks/src/lib.rs
```diff
@@ -1366,7 +1365,7 @@ impl ShardsManager {
         );
 
         let block_producers =
-            self.runtime_adapter.get_epoch_block_producers_ordered(&epoch_id, &parent_hash)?;
+            self.runtime_adapter.get_epoch_block_approvers_ordered(&parent_hash)?;
```
mikhailOK, line 1368: The test passes without this change. Keep it or revert?

birchmd: This change made the test pass before you made changes to the test instead? This seems like a good change to me; at the time I was writing this logic I was not thinking about the subtle difference between the producer and approver sets. But approvers is the right set to send to because the point of this logic is to help nodes tracking the shard to have enough parts to reconstruct the chunk without sending requests and I assume approvals will not be sent out until the chunks for shards they track are reconstructed.

mikhailOK: It changed from 0/10 pass to 2/10 pass, but need to change a bit more code there: `get_all_block_producers_ordered` uses a caching crutch that `get_all_block_approvers_ordered` doesn't have so it could increase cpu

bowenwang1996: Yeah this change makes sense

bowenwang1996 (approved):

Issue #4030, opened 2021-03-03
Possible performance regression in contract execution


birchmd:
Our nightly load testing on mocknet has been consistently failing for a long time. Many of these have been related to problems compiling contracts, however the last several runs have been because there are multiple blocks which take more than 1s to process (the test fails if this happens because you cannot have a 1s block time if blocks take more than 1s to process).

The blocks which take too long to process only happen during the phase where multiple transaction types (including contract calls) are being sent to the network; the phase where all transactions are simple transfers still works well. So it might be the case that there is a performance regression in the execution of contracts. The other transaction types sent during this phase are staking and account creation, so it could also be that one of those has seen a regression instead (or perhaps in addition to).

References:

*   <span class="error">[Load test history]</span>([http://nayduck.eastus.cloudapp.azure.com:3000/#/test_history/99461](http://nayduck.eastus.cloudapp.azure.com:3000/#/test_history/99461))
*   Example failure output (scroll all the way to the bottom to see the `Number of blocks processing for more than 1s`): <span class="error">[1]</span>([https://nayduckdiag.blob.core.windows.net/logs/96888_v2_stdout](https://nayduckdiag.blob.core.windows.net/logs/96888_v2_stdout)), <span class="error">[2]</span>([https://nayduckdiag.blob.core.windows.net/logs/97660_v2_stdout](https://nayduckdiag.blob.core.windows.net/logs/97660_v2_stdout)), <span class="error">[3]</span>([https://nayduckdiag.blob.core.windows.net/logs/97664_v2_stdout](https://nayduckdiag.blob.core.windows.net/logs/97664_v2_stdout)), <span class="error">[4]</span>([https://nayduckdiag.blob.core.windows.net/logs/97923_v2_stdout](https://nayduckdiag.blob.core.windows.net/logs/97923_v2_stdout)), <span class="error">[5]</span>([https://nayduckdiag.blob.core.windows.net/logs/98947_v2_stdout](https://nayduckdiag.blob.core.windows.net/logs/98947_v2_stdout)), <span class="error">[6]</span>([https://nayduckdiag.blob.core.windows.net/logs/99461_v2_stdout](https://nayduckdiag.blob.core.windows.net/logs/99461_v2_stdout))
[...]

PR #4136, opened 2021-03-18
feature: more runtime performance visibility into runtime

matklad:
- add a couple of spans that turned out to be potentially significant
  during #4030 investigation, notably dropping of the instance

- refactor code to make sure that preloading code path also has the same
  spans.

See [1] for how to read the timing info from spans

[1]: https://github.com/near/nearcore/tree/ad3de8f3a597b6cc5c01871a4709973828de3b9f/runtime/near-vm-runner#profiling

### DIFF
--- a/runtime/near-vm-runner/src/cache.rs
+++ b/runtime/near-vm-runner/src/cache.rs
@@ -46,6 +46,7 @@ fn vm_hash(vm_kind: VMKind) -> u64 {
 }
 
 fn get_key(code: &ContractCode, vm_kind: VMKind, config: &VMConfig) -> CryptoHash {
+    let _span = tracing::debug_span!("get_key").entered();
     let key = ContractCacheKey::Version2 {
         code_hash: code.hash,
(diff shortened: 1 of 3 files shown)


bowenwang1996 (commented):
@matklad do we want to merge this? It feels like this change would make code less readable

matklad:
Lets see what @olonho and @ailisp think. 

I agree that it makes the code somewhat less readable, as we want to track each operation, and that introduces "unnecessary" blocks.  That's because this is the place where we call from "our" code into wasmer code, so we can't just use a single span at the function entry. 

At the same time, timing these operations would be helpful in both day-to-day performance optimization work, as well as in monitoring and debugging canaries and production instances, *particularly* because this is not the code written by us. 

On my personal scales, observability wins in this particular case by a large margin.

olonho:
> On my personal scales, observability wins in this particular case by a large margin.
Agree, that it looks somewhat ugly, but indeed - that's ugliness for few, while many will benefit from results of traceability. So LGTM.

olonho (approved):

PR #4143, opened 2021-03-19
fix(primitives-core): ProfileData default uses feature and add tests

willemneal:
The simulator was failing due to default calling `new_disabled` and then debug crashed when dividing by zero. Now debug prints a helpful statement when `costs_counting` is not enabled.

### DIFF
--- a/core/primitives-core/src/profile.rs
+++ b/core/primitives-core/src/profile.rs
@@ -14,7 +14,9 @@ pub struct FixedArray<T> {
 
 impl<T> Clone for FixedArray<T> {
     fn clone(&self) -> Self {
-        Self { data: Rc::clone(&self.data) }
+        Self {
+            data: Rc::clone(&self.data),
+        }
     }
 }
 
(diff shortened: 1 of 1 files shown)


frol (approved):

core/primitives-core/src/profile.rs
```diff
@@ -37,9 +37,15 @@ enum Repr {
...
 
 impl Default for ProfileData {
+    #[cfg(not(feature = "costs_counting"))]
     fn default() -> ProfileData {
         ProfileData::new_disabled()
     }
+
+    #[cfg(feature = "costs_counting")]
+    fn default() -> ProfileData {
+        ProfileData::new_enabled()
```
matklad, line 47: @willemneal I am not completely sure that switching default semantics based on the feature flag is the right thing to do. That's a spooky action at a distance kind of thing, as the result of `ProfileData::default` now is "it depends". 

The original thinking was that the call-sites which need cost-counting would call `ProfileData::new_enabled`, like this:

```rust
diff --git a/near-sdk-sim/src/runtime.rs b/near-sdk-sim/src/runtime.rs
index 3abd685..546edad 100644
--- a/near-sdk-sim/src/runtime.rs
+++ b/near-sdk-sim/src/runtime.rs
@@ -236,7 +236,7 @@ impl RuntimeStandalone {
 
     /// Processes one block. Populates outcomes and producining new pending_receipts.
     pub fn produce_block(&mut self) -> Result<(), RuntimeError> {
-        let profile_data = ProfileData::default();
+        let profile_data = ProfileData::new_enabled();
         let apply_state = ApplyState {
             block_index: self.cur_block.block_height,
             prev_block_hash: Default::default(),
```

Would this work for near-sdk-rs?

willemneal: @matklad Here is a PR to do just that: https://github.com/near/nearcore/pull/4161

PR #4145, opened 2021-03-19
Revert "Validate each edge only once"

pmnoxx:
Reverts near/nearcore#4111

Reverting change, that could break exchange of routing tables.

In 4111, we switches, how we add edges to PeerManager. We will now send only some edges to PeerManager in one request, not the whole graph. 

Inside `try_save_edges`, we run code, which tries to prune edges that are unreachable. We may remove some valid edges, because we didn't get full graph yet.

@bowenwang1996


bowenwang1996 (approved):

pmnoxx:
I'll resubmit this change later, but first we need to figure out what to do with `try_save_edges`

PR #4150, opened 2021-03-20
fix: Update runtime config default storage cost

### LINKED ISSUE #4148: RuntimeConfig has wrong storage amount per byte by default
**Describe the bug**
[Runtime Config] default has wrong storage amount per byte by default.
This leads to issues like near sdk simulator and runtime standalone calculating it wrong as well.

**To Reproduce**
See https://github.com/near/nearcore/blob/master/core/primitives/src/runtime/config.rs#L34

**Expected behavior**
It should be 1e19 yN.

CC @willemneal

willemneal:
fixes #4148


bowenwang1996 (commented):
@evgenykuzyakov can we actually update this? Will this change how the old version works?

bowenwang1996:
@willemneal I assume this is outdated. Please reopen if that is not the case

PR #4151, opened 2021-03-20
refactor(crypto): impl FromStr instead of TryFrom<&str> for near_crypto::Signature and CryptoHash

frol:
I keep refactoring traits usage in near-crypto. `FromStr` is what is used whenever someone wants to parse strings into a structure. I need `near_crypto::Signature` to be parsable in order to use it with clap (CLI parsing).


frol:
While I was at it, I also refactored CryptoHash the same way (TryFrom<&str> and TryFrom<String> -> FromStr)

bowenwang1996 (approved):

khorolets (approved):

PR #4154, opened 2021-03-22
Optimize hash function for PeerId

pmnoxx:
We've been running Borsh Serializer every time we compute cache for PeerId.


frol (approved):
What was the original motivation to use a custom Hash implementation?

bowenwang1996:
> What was the original motivation to use a custom Hash implementation?

I believe it was intended to avoid running the hash computation, but I don't think it achieved that goal since it only changed what is written to the hasher (and in a suboptimal way).

PR #4160, opened 2021-03-24
refactor: centralize&improve infrastructure for building test contracts

matklad:
* Move test contracts to the dedicated `runtime/near-test-contracts` folder
* Do not commit `.wasm` files and instead re-build them on the fly
  when necessary
* Use build.rs rather than build.sh for compilation
* Replace lazy_static_include calls with centralized functions


matklad:
Backstory here: I've noticed that our [`sum_n`](https://github.com/near/nearcore/blob/9dcc926b6a43352d43ecf0eaeb779b81940a8e72/runtime/near-vm-runner/tests/test-contract-rs/src/lib.rs#L478-L481) method doesn't actually sum the n numbers, as LLVM can compute sums of polynomial terms in closed form. I wanted to fix that (as we do use that method for benchmarking), but realized that I would need to build wasm locally and commit it.

matklad:
One thing I am not sure here is whether our python tests are guaranteed to run only after `cargo build`. Some of them read .wasm directly from the `res` folder, and, to populate res, building rust code is required.

test-utils/testlib/src/runtime_utils.rs
```diff
@@ -27,18 +24,11 @@ pub fn evm_account() -> AccountId {
...
 }
 
 lazy_static::lazy_static! {
-    static ref DEFAULT_TEST_CONTRACT_HASH: CryptoHash = hash(&DEFAULT_TEST_CONTRACT);
-}
-
-lazy_static_include::lazy_static_include_bytes! {
-    DEFAULT_TEST_CONTRACT => "../../runtime/near-vm-runner/tests/res/test_contract_rs.wasm"
+    static ref DEFAULT_TEST_CONTRACT_HASH: CryptoHash = hash(near_test_contracts::rs_contract());
 }
```
matklad, line 32: This seems to be some pre-existing duplication: `default_code_hash` and `DEFAULT_TEST_CONTRACT` compute the same thing. Left as is, as out of scope for the PR.

chefsale (changes requested):
Seems like tests are failing, lets fix that first. 😄

runtime/near-test-contracts/Cargo.toml
```diff
@@ -0,0 +1,10 @@
+[package]
+name = "near-test-contracts"
+version = "0.0.0"
```
bowenwang1996, line 3: Does 0.0.0 make sense?

runtime/near-vm-runner/src/tests/rs_contract.rs
```diff
@@ -13,18 +13,13 @@ use crate::tests::{
...
-}
-
-#[cfg(not(feature = "protocol_feature_alt_bn128"))]
-lazy_static_include::lazy_static_include_bytes! {
-    TEST_CONTRACT => "tests/res/test_contract_rs.wasm",
-}
-
 fn test_contract() -> ContractCode {
-    ContractCode::new(TEST_CONTRACT.to_vec(), None)
+    let code = if cfg!(feature = "protocol_feature_alt_bn128") {
```
bowenwang1996, line 17: This seems off to me. It should be based on whether `nightly_protocol` is enabled.

matklad: Probably! I've tried to keep pre-existing semantics in this PR, I can change this and (that)[https://github.com/near/nearcore/pull/4160/files#r600402157] in a follow up.

runtime/near-test-contracts/Cargo.toml
matklad: It does -- we are not going to publish this crate, so it doesn't have a meaningful semver. So putting a default `0.1.0` would be (a minor and insignificant) lie. `0.0.0` and `publish = false` explicitly say that we don't care about semver in this case.

matklad:
Got this past CI, including the "One thing I am not sure here is whether our python tests are guaranteed to run only after cargo build" bit.

runtime/near-test-contracts/build.rs
```diff
@@ -0,0 +1,50 @@
...
+    let status = cmd.status()?;
+    if !status.success() {
+        return Err(io::Error::new(io::ErrorKind::Other, "cargo build failed"));
+    }
+    fs::copy(
+        format!("./{}/target/wasm32-unknown-unknown/release/{}.wasm", dir, dir.replace('-', "_")),
+        format!("./res/{}.wasm", output),
+    )?;
+    println!("cargo:rerun-if-changed=./{}/src/lib.rs", dir);
+    println!("cargo:rerun-if-changed=./{}/Cargo.toml", dir);
```
ailisp, line 41: What are these println?

ailisp (approved):

runtime/near-test-contracts/build.rs
matklad: They tell cargo when it needs to rebuild the test contracts, These are documented in https://doc.rust-lang.org/cargo/reference/build-scripts.html#outputs-of-the-build-script

In theory, we maybe can just always run `cargo build` here, as that should be a no-op, but it wasn't hard to save an extra process invocation here.

bowenwang1996 (approved):

chefsale (approved):

runtime/near-vm-runner/src/tests/rs_contract.rs
matklad: Took a look here, I am not sure how to fix it. This crate doesn't have `nightly_protocol` feature: 

```toml
[features]
# all vms enabled for tests, but only one default vm, specified by runtime crate
default = ["wasmer0_vm", "wasmtime_vm", "wasmer1_vm"]
wasmer0_vm = [ "wasmer-runtime" ]
wasmtime_vm = [ "wasmtime", "anyhow"]
wasmer1_vm = [ "wasmer", "wasmer-types", "wasmer-compiler-singlepass",  "wasmer-compiler-cranelift", "wasmer-engine-native", ]
lightbeam = ["wasmtime/lightbeam"]
no_cpu_compatibility_checks = []
protocol_feature_evm = ["near-primitives/protocol_feature_evm", "near-evm-runner/protocol_feature_evm"]

no_cache = []

protocol_feature_alt_bn128 = [
    "near-vm-logic/protocol_feature_alt_bn128",
    "near-primitives/protocol_feature_alt_bn128",
    "near-vm-errors/protocol_feature_alt_bn128"
]
```

At this point, I don't know the purpose of `nightly_protocol` and `nightly_protocol_features` cargo features and why they are set up that way, so I can't immediately fix that. Learning more about our feature setup is on my todo list, so I might come back to this later.

bowenwang1996: @matklad nightly_protocol is supposed to be a marker feature https://github.com/near/NEPs/issues/126#issuecomment-712130812

PR #4161, opened 2021-03-24
fix(primitives-core): make default profile disabled

willemneal:
As per @matklad suggested that default shouldn't be dependent on feature flag.

### DIFF
--- a/core/primitives-core/src/profile.rs
+++ b/core/primitives-core/src/profile.rs
@@ -225,7 +225,6 @@ mod test {
 
     #[test]
     fn test_profile_data_debug() {
-        #[cfg(not(feature = "costs_counting"))]
         println!("{:#?}", ProfileData::new_enabled());
     }
 }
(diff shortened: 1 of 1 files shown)


core/primitives-core/src/profile.rs
```diff
@@ -232,6 +226,6 @@ mod test {
     #[test]
     fn test_profile_data_debug() {
         #[cfg(not(feature = "costs_counting"))]
-        println!("{:#?}", ProfileData::default());
+        println!("{:#?}", ProfileData::new_enabled());
```
matklad, line 229: This is just to check that we don't panic, right? I guess we might as well remove the ` #[cfg(not(feature = "costs_counting"))]` on the previous line then?

matklad (approved):

core/primitives-core/src/profile.rs
willemneal: Yep.  I'll remove it.

core/primitives-core/src/profile.rs
```diff
@@ -232,6 +226,6 @@ mod test {
     #[test]
     fn test_profile_data_debug() {
         #[cfg(not(feature = "costs_counting"))]
```
willemneal, line 228:
```suggestion
```

frol (approved):

PR #4162, opened 2021-03-24
refactor: don't build parameter for integration tests

matklad:
runtime-params-estimator is a tricky thing to build, because it requires
that cost-counting features are enabled in the rest of nearcore. As
features in Cargo are contagions (cargo always takes a union of all
enabled features), having params estimator as a dependency is
problematic.

But looks like we just don't need to?

More specifically, this PR fixes the compilation error for the following
command:

  cargo test test_add_access_key_function_call_testnet \
      --features nightly_protocol,nightly_protocol_features,expensive_tests

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3688,7 +3688,6 @@ dependencies = [
  "neard",
  "node-runtime",
  "rand 0.7.3",
- "runtime-params-estimator",
  "testlib",
 ]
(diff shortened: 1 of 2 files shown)


matklad:
Closing. #4130 is a less principled fix. I want to do something better here, but that would be a bigger change.

PR #4163, opened 2021-03-24
fix(test): fix nightly access key nonce tests

### LINKED ISSUE #4158: test_cases_testnet_rpc tests fail with nightly features on NayDuck
See:
http://nayduck.eastus.cloudapp.azure.com:3000/#/test/123689
http://nayduck.eastus.cloudapp.azure.com:3000/#/test/123703
http://nayduck.eastus.cloudapp.azure.com:3000/#/test/123707

We missed it due to rust tests were not run with features enabled

bowenwang1996:
Fixes #4158.

Test plan
---------
http://nayduck.eastus.cloudapp.azure.com:3000/#/


test-utils/testlib/src/standard_test_cases.rs
```diff
@@ -840,6 +840,30 @@ pub fn test_delete_key_last(node: impl Node) {
...
+            key
+        }
+        #[cfg(not(feature = "protocol_feature_access_key_nonce_range"))]
+        {
+            let _ = (user, result);
+            access_key.clone()
+        }
+    };
+    assert_eq!(access_key_view, key.into());
+}
```
EgorKulikov, line 866:
```suggestion
}

```

test-utils/testlib/src/standard_test_cases.rs
```diff
@@ -840,6 +840,30 @@ pub fn test_delete_key_last(node: impl Node) {
...
+            let mut key = access_key.clone();
+            let block = user.get_block_by_hash(result.transaction_outcome.block_hash);
+            if let Some(b) = block {
+                key.nonce = (b.header.height - 1) * AccessKey::ACCESS_KEY_NONCE_RANGE_MULTIPLIER;
+            }
+            key
+        }
+        #[cfg(not(feature = "protocol_feature_access_key_nonce_range"))]
+        {
+            let _ = (user, result);
```
EgorKulikov, line 861: I am not sure what is happening here, are we trying to avoid user and result being not used?

EgorKulikov (approved):

test-utils/testlib/src/standard_test_cases.rs
```diff
@@ -840,6 +840,30 @@ pub fn test_delete_key_last(node: impl Node) {
...
+            access_key.clone()
+        }
```
matklad, line 863:
```suggestion
    let key = if cfg!(feature = "protocol_feature_access_key_nonce_range") {
            let mut key = access_key.clone();
            let block = user.get_block_by_hash(result.transaction_outcome.block_hash);
            if let Some(b) = block {
                key.nonce = (b.header.height - 1) * AccessKey::ACCESS_KEY_NONCE_RANGE_MULTIPLIER;
            }
            key
        } else  {
            let _ = (user, result);
            access_key.clone()
        }
```

Seems like this *might* be enough? The benefit of `if cfg!()` is that both branches are type-checked (but then one is eliminated during codegen).

test-utils/testlib/src/standard_test_cases.rs
```diff
@@ -840,6 +840,30 @@ pub fn test_delete_key_last(node: impl Node) {
     assert!(node_user.get_access_key(&account_id, &node.signer().public_key()).is_err());
 }
 
+fn assert_access_key(
+    access_key: &AccessKey,
+    access_key_view: AccessKeyView,
+    result: &FinalExecutionOutcomeView,
+    user: Box<dyn User>,
```
matklad, line 847:
```suggestion
    user: &dyn User,
```

seems like just ref is enough?

test-utils/testlib/src/standard_test_cases.rs
bowenwang1996: Yes

test-utils/testlib/src/standard_test_cases.rs
```diff
@@ -847,19 +847,19 @@ fn assert_access_key(
...
-            let _ = (user, result);
-            access_key.clone()
+        let mut key = access_key.clone();
+        let block = user.get_block_by_hash(result.transaction_outcome.block_hash);
+        if let Some(b) = block {
+            key.nonce = (b.header.height - 1) * AccessKey::ACCESS_KEY_NONCE_RANGE_MULTIPLIER;
         }
+        key
+    } else {
+        let _ = (user, result);
```
matklad, line 857:
```suggestion
        let _ = (user, result);
```

You no longer need this

PR #4165, opened 2021-03-25
fix: count vm_runner cache operation as busy, not as idle

matklad:
Unentered spans count towrads "idle" measure, which is not what we want. We want "busy", and for that the span must be entered.


ailisp (approved):
Hmm why before this change there's still some deserialize_wasmer1 tracing numbers?

matklad:
>Hmm why before this change there's still some deserialize_wasmer1 tracing numbers?

@ailisp sorry, missed this comment. We don't use memory cache for wasmer1, it always goes via `CompliedContractCache` which incures deserilization. That tripped me up during benchmarking as well.

PR #4167, opened 2021-03-25
feat(rpc): viewing large contract state is now prohibted to avoid excessive resource consumption

### LINKED ISSUE #3950: Reject view_state JSON RPC requests for big accounts
This is a short-term workaround to avoid DoS from #3938

bowenwang1996:
Fixes #3950.

Test plan
---------
* `test_view_state_too_large`
* Nayduck


neard/src/runtime/errors.rs
```diff
@@ -74,6 +75,12 @@ impl QueryError {
             node_runtime::state_viewer::errors::ViewStateError::InternalError { error_message } => {
                 Self::InternalError { error_message, block_height, block_hash }
             }
+            ViewStateError::AccountDoesNotExist { requested_account_id } => {
+                Self::InvalidAccount { requested_account_id, block_height, block_hash }
+            }
+            ViewStateError::AccountStateTooLarge { requested_account_id } => {
+                Self::InvalidAccount { requested_account_id, block_height, block_hash }
```
bowenwang1996, line 82: @frol @khorolets should this be a different error or is it fine to reuse `InvalidAccount`?

frol: Let's use a separate error variant to properly communicate the reason. Following the style of InvalidAccount, UnknownAccount NoContractCode, I suggest TooLargeContractState

frol (approved):

khorolets (approved):

runtime/runtime/src/state_viewer/mod.rs
```diff
@@ -124,6 +124,25 @@ impl TrieViewer {
...
+            Some(account) => {
+                let code_len = get_code(state_update, account_id, Some(account.code_hash()))?
+                    .map(|c| c.code.len() as u64)
+                    .unwrap_or_default();
+                if account.storage_usage() > Account::MAX_ACCOUNT_DELETION_STORAGE_USAGE + code_len
+                {
+                    return Err(errors::ViewStateError::AccountStateTooLarge {
+                        requested_account_id: account_id.clone(),
+                    });
+                }
```
evgenykuzyakov, line 137: How then I should clean up my state, if I can't view it either?

This will likely break the berryclub wayback machine that relies on the prefix query.

We should bump this limit to something higher since this is an RPC call

runtime/runtime/src/state_viewer/mod.rs
```diff
@@ -25,6 +25,9 @@ pub mod errors;
 pub struct TrieViewer {}
 
 impl TrieViewer {
+    /// Upper bound of the size of contract state that is still viewable.
+    const CONTRACT_STATE_SIZE_LIMIT: u64 = 50_000;
```
frol, line 29: How complicated is it to be exposed via the config.json file, so we can recommend people to run their own RPC node with the limit bumped to the values they feel is necessary?

frol (approved):

evgenykuzyakov (approved):

Issue #3421, opened 2020-10-01
Upgrade `DeleteAccount` action to generate regular transfer


evgenykuzyakov:
Currently `DeleteAccount` action generates a refund. A refund can't create an implicit account.
If we switch it to generate regular accounts, then beneficiary receiver account can be created. This allows to support UTXO like transactions on NEAR. Where every transaction on an implicit account instead a `Transfer` and `DeleteAccount` action that removes current account and splits the amount into a receiver and the remainder. The order of the receiver and the remainder can be randomized.

- [ ] Create a NEP for the proposal
- [ ] Implement a protocol upgrade
- [ ] Commit NEP and update the spec.

nearmax:
As discussed today, this proposal does not explain what is the benefit of this feature, and given that it makes the protocol more complex it shouldn't have been implemented. Since this feature has already caused problems for NEAR network, @bowenwang1996 decided to revert it.

PR #4169, opened 2021-03-26
Upgrade `DeleteAccount` action to generate regular transfer

Longarithm:
Solution for #3421


runtime/runtime/src/config.rs
```diff
@@ -132,7 +152,10 @@ pub fn total_send_fees(
                 }
             },
             DeleteKey(_) => cfg.delete_key_cost.send_fee(sender_is_receiver),
-            DeleteAccount(_) => cfg.delete_account_cost.send_fee(sender_is_receiver),
+            DeleteAccount(_) => {
+                // TODO add cfg[feature]
+                cfg.delete_account_cost.send_fee(sender_is_receiver) + cfg.transfer_cost.send_fee(sender_is_receiver)
```
evgenykuzyakov, line 157: You should use `send_transfer_fee(cfg, sender_is_receiver, receiver_id, current_protocol_version)` here

runtime/runtime/src/config.rs
```diff
@@ -186,7 +202,10 @@ pub fn exec_fee(
             AccessKeyPermission::FullAccess => cfg.add_key_cost.full_access_cost.exec_fee(),
         },
         DeleteKey(_) => cfg.delete_key_cost.exec_fee(),
-        DeleteAccount(_) => cfg.delete_account_cost.exec_fee(),
+        DeleteAccount(_) => {
+            // TODO add cfg[feature]
+            cfg.delete_account_cost.exec_fee() + cfg.transfer_cost.exec_fee()
```
evgenykuzyakov, line 207: And `exec_transfer_fee(cfg, receiver_id, current_protocol_version)` here

runtime/runtime/src/config.rs
Longarithm: Discussed offline

bowenwang1996 (commented):
Can you add a transaction to delete account and create implicit account in `pytests/sanity/upgradable.py`?

runtime/near-vm-logic/src/logic.rs
```diff
@@ -24,6 +25,51 @@ pub type Result<T> = ::std::result::Result<T, VMLogicError>;
...
+            cfg.create_account_cost
+                .exec_fee()
+                .checked_add(cfg.add_key_cost.full_access_cost.exec_fee())
+                .ok_or(HostError::IntegerOverflow)?
+                .checked_add(cfg.transfer_cost.exec_fee())
+                .ok_or(HostError::IntegerOverflow)?
+        } else {
+            cfg.transfer_cost.exec_fee()
+        },
+    )
```
matklad, line 46: nit and mostly personal preference, but I usually write these as

```
    let res = if current_protocol_version >= IMPLICIT_ACCOUNT_CREATION_PROTOCOL_VERSION
        && is_account_id_64_len_hex(&receiver_id)
    {
        cfg.create_account_cost
            .exec_fee()
            .checked_add(cfg.add_key_cost.full_access_cost.exec_fee())
            .ok_or(HostError::IntegerOverflow)?
            .checked_add(cfg.transfer_cost.exec_fee())
            .ok_or(HostError::IntegerOverflow)?
    } else {
        cfg.transfer_cost.exec_fee()
    };
    Ok(res)
```

If you put a complex expression inside `Ok`  / `Some`, there's a chance that it'll grow *more* complex in the future, and you would want to pull parts of it out of `Ok`, which would shift the indentation. So, a more future-proof version is to extract the inner value into a variable from the start.

But, again, this is more "consider if you might like this alternative way", it not necessary better.

runtime/runtime/src/actions.rs
```diff
@@ -444,10 +448,14 @@ pub(crate) fn action_delete_account(
...
     actor_id: &mut AccountId,
     receipt: &Receipt,
+    #[cfg(feature = "protocol_feature_allow_create_account_on_delete")]
+    // Comment to work-around rustfmt bug: https://github.com/rust-lang/rustfmt/issues/4783
+    action_receipt: &ActionReceipt,
     result: &mut ActionResult,
     account_id: &AccountId,
     delete_account: &DeleteAccountAction,
     current_protocol_version: ProtocolVersion,
+    #[cfg(feature = "protocol_feature_allow_create_account_on_delete")] config: &RuntimeFeesConfig,
```
matklad, line 458: Not sure whether it makes sense to gate these arguments with  feature flags. I think the following would be simpler? 

```diff
diff --git a/runtime/near-vm-logic/src/logic.rs b/runtime/near-vm-logic/src/logic.rs
index f050ea42..13cf9072 100644
--- a/runtime/near-vm-logic/src/logic.rs
+++ b/runtime/near-vm-logic/src/logic.rs
@@ -29,7 +29,7 @@ pub fn transfer_exec_fee(
     cfg: &ActionCreationConfig,
     receiver_id: &AccountId,
     current_protocol_version: ProtocolVersion,
-) -> Result<Gas> {
+) -> Result<Gas, OverflowEr> {
     Ok(
         if current_protocol_version >= IMPLICIT_ACCOUNT_CREATION_PROTOCOL_VERSION
             && is_account_id_64_len_hex(&receiver_id)
diff --git a/runtime/runtime/src/actions.rs b/runtime/runtime/src/actions.rs
index ea4e4411..bb3e7eba 100644
--- a/runtime/runtime/src/actions.rs
+++ b/runtime/runtime/src/actions.rs
@@ -6,7 +6,6 @@ use near_primitives::checked_feature;
 use near_primitives::contract::ContractCode;
 use near_primitives::errors::{ActionError, ActionErrorKind, ExternalError, RuntimeError};
 use near_primitives::hash::CryptoHash;
-#[cfg(feature = "protocol_feature_allow_create_account_on_delete")]
 use near_primitives::receipt::ReceiptEnum;
 use near_primitives::receipt::{ActionReceipt, Receipt};
 use near_primitives::runtime::config::AccountCreationConfig;
@@ -37,7 +36,6 @@ use near_vm_logic::types::PromiseResult;
 use near_vm_logic::{VMContext, VMOutcome};
 
 use crate::config::{safe_add_gas, RuntimeConfig};
-#[cfg(feature = "protocol_feature_allow_create_account_on_delete")]
 use crate::config::{transfer_exec_fee, transfer_send_fee};
 use crate::ext::RuntimeExt;
 use crate::{ActionResult, ApplyState};
@@ -448,14 +446,12 @@ pub(crate) fn action_delete_account(
     account: &mut Option<Account>,
     actor_id: &mut AccountId,
     receipt: &Receipt,
-    #[cfg(feature = "protocol_feature_allow_create_account_on_delete")]
-    // Comment to work-around rustfmt bug: https://github.com/rust-lang/rustfmt/issues/4783
-    action_receipt: &ActionReceipt,
+    action_receipt: &ActionReceipt,
     result: &mut ActionResult,
     account_id: &AccountId,
     delete_account: &DeleteAccountAction,
     current_protocol_version: ProtocolVersion,
-    #[cfg(feature = "protocol_feature_allow_create_account_on_delete")] config: &RuntimeFeesConfig,
+    config: &RuntimeFeesConfig,
 ) -> Result<(), StorageError> {
     if current_protocol_version >= ProtocolFeature::DeleteActionRestriction.protocol_version() {
         let account = account.as_ref().unwrap();
@@ -478,52 +474,47 @@ pub(crate) fn action_delete_account(
     // We use current amount as a pay out to beneficiary.
     let account_balance = account.as_ref().unwrap().amount();
     if account_balance > 0 {
-        checked_feature!(
+        if checked_feature!(
             "protocol_feature_allow_create_account_on_delete",
             AllowCreateAccountOnDelete,
-            current_protocol_version,
-            {
-                let sender_is_receiver = account_id == &delete_account.beneficiary_id;
-                let exec_gas = config.action_receipt_creation_config.send_fee(sender_is_receiver)
-                    + transfer_send_fee(
-                        &config.action_creation_config,
-                        sender_is_receiver,
-                        &delete_account.beneficiary_id,
-                        current_protocol_version,
-                    );
-                result.gas_burnt += exec_gas;
-                result.gas_used += exec_gas
-                    + config.action_receipt_creation_config.exec_fee()
-                    + transfer_exec_fee(
-                        &config.action_creation_config,
-                        &delete_account.beneficiary_id,
-                        current_protocol_version,
-                    );
-
-                result.new_receipts.push(Receipt {
-                    predecessor_id: account_id.clone(),
-                    receiver_id: delete_account.beneficiary_id.clone(),
-                    receipt_id: CryptoHash::default(),
-
-                    receipt: ReceiptEnum::Action(ActionReceipt {
-                        signer_id: action_receipt.signer_id.clone(),
-                        signer_public_key: action_receipt.signer_public_key.clone(),
-                        gas_price: action_receipt.gas_price,
-                        output_data_receivers: vec![],
-                        input_data_ids: vec![],
-                        actions: vec![Action::Transfer(TransferAction {
-                            deposit: account_balance,
-                        })],
-                    }),
-                });
-            },
-            {
-                result.new_receipts.push(Receipt::new_balance_refund(
+            current_protocol_version
+        ) {
+            let sender_is_receiver = account_id == &delete_account.beneficiary_id;
+            let exec_gas = config.action_receipt_creation_config.send_fee(sender_is_receiver)
+                + transfer_send_fee(
+                    &config.action_creation_config,
+                    sender_is_receiver,
                     &delete_account.beneficiary_id,
-                    account_balance,
-                ));
-            }
-        )
+                    current_protocol_version,
+                );
+            result.gas_burnt += exec_gas;
+            result.gas_used += exec_gas
+                + config.action_receipt_creation_config.exec_fee()
+                + transfer_exec_fee(
+                    &config.action_creation_config,
+                    &delete_account.beneficiary_id,
+                    current_protocol_version,
+                );
+
+            result.new_receipts.push(Receipt {
+                predecessor_id: account_id.clone(),
+                receiver_id: delete_account.beneficiary_id.clone(),
+                receipt_id: CryptoHash::default(),
+
+                receipt: ReceiptEnum::Action(ActionReceipt {
+                    signer_id: action_receipt.signer_id.clone(),
+                    signer_public_key: action_receipt.signer_public_key.clone(),
+                    gas_price: action_receipt.gas_price,
+                    output_data_receivers: vec![],
+                    input_data_ids: vec![],
+                    actions: vec![Action::Transfer(TransferAction { deposit: account_balance })],
+                }),
+            });
+        } else {
+            result
+                .new_receipts
+                .push(Receipt::new_balance_refund(&delete_account.beneficiary_id, account_balance));
+        }
     }
     remove_account(state_update, account_id)?;
     *actor_id = receipt.predecessor_id.clone();
diff --git a/runtime/runtime/src/lib.rs b/runtime/runtime/src/lib.rs
index 9c2df71b..160381c1 100644
--- a/runtime/runtime/src/lib.rs
+++ b/runtime/runtime/src/lib.rs
@@ -423,13 +423,11 @@ impl Runtime {
                     account,
                     actor_id,
                     receipt,
-                    #[cfg(feature = "protocol_feature_allow_create_account_on_delete")]
                     action_receipt,
                     &mut result,
                     account_id,
                     delete_account,
                     apply_state.current_protocol_version,
-                    #[cfg(feature = "protocol_feature_allow_create_account_on_delete")]
                     &apply_state.config.transaction_costs,
                 )?;
             }
```

runtime/runtime/src/actions.rs
```diff
@@ -802,15 +853,24 @@ mod tests {
         let mut actor_id = account_id.clone();
         let mut action_result = ActionResult::default();
         let receipt = Receipt::new_balance_refund(&"alice.near".to_string(), 0);
+        #[cfg(feature = "protocol_feature_allow_create_account_on_delete")]
+        let action_receipt = match receipt.receipt {
+            ReceiptEnum::Action(ref action_receipt) => action_receipt,
```
matklad, line 858:
```suggestion
        let action_receipt = match &receipt.receipt {
            ReceiptEnum::Action(action_receipt) => action_receipt,
```

runtime/runtime/src/config.rs
```diff
@@ -189,6 +205,48 @@ pub fn exec_fee(
...
+pub fn prepaid_exec_fee(
+    config: &RuntimeFeesConfig,
+    action: &Action,
+    receiver_id: &AccountId,
+    current_protocol_version: ProtocolVersion,
+) -> Gas {
+    let exec_gas = exec_fee(config, action, receiver_id, current_protocol_version);
+
+    let extra_prepaid_gas = match action {
+        Action::DeleteAccount(DeleteAccountAction { beneficiary_id }) => checked_feature!(
```
matklad, line 218: On the first glance, it seems like this could be `if checked_feature!()` as well?

test-utils/testlib/src/fees_utils.rs
```diff
@@ -155,11 +163,27 @@ impl FeeHelper {
         self.gas_to_balance(exec_gas + send_gas)
     }
 
-    pub fn delete_account_cost(&self) -> Balance {
+    pub fn prepaid_delete_account_cost(
+        &self,
+        #[cfg(feature = "protocol_feature_allow_create_account_on_delete")]
+        implicit_account_created: bool,
```
matklad, line 169: Code smell: this boolean is always passed as literal `true` or `false` on the call site. This means that it might be better to split this function in two with different names:

```rust
pub fn prepaid_delete_account_cost_for_implicit_account(&self) {
    self.prepaid_delete_account_cost(false)
}
pub fn prepaid_delete_account_cost_for_explicit_account(&self) {
    self.prepaid_delete_account_cost(true)
}
fn self.prepaid_delete_account_cost(&self, is_accout_implicet) {
    ...
}

```

runtime/near-vm-logic/src/logic.rs
```diff
@@ -1450,6 +1496,14 @@ impl<'a> VMLogic<'a> {
         Ok(())
     }
 
+    // TODO document
```
bowenwang1996, line 1499: is this supposed to be done?

runtime/near-vm-logic/src/logic.rs
```diff
@@ -24,6 +25,51 @@ pub type Result<T> = ::std::result::Result<T, VMLogicError>;
 const LEGACY_DEFAULT_PROTOCOL_VERSION: ProtocolVersion = 34;
 const IMPLICIT_ACCOUNT_CREATION_PROTOCOL_VERSION: ProtocolVersion = 35;
 
+pub fn transfer_exec_fee(
```
bowenwang1996, line 28: why does this function need to be public? Also it probably makes more sense to implement this in `runtime/src/config.rs`

runtime/runtime/src/actions.rs
```diff
@@ -470,9 +478,52 @@ pub(crate) fn action_delete_account(
...
+                    + transfer_exec_fee(
+                        &config.action_creation_config,
+                        &delete_account.beneficiary_id,
+                        current_protocol_version,
+                    );
+
+                result.new_receipts.push(Receipt {
+                    predecessor_id: account_id.clone(),
+                    receiver_id: delete_account.beneficiary_id.clone(),
+                    receipt_id: CryptoHash::default(),
```
bowenwang1996, line 506: Receipt id must be globally unique.

runtime/runtime/src/config.rs
```diff
@@ -73,6 +70,41 @@ macro_rules! safe_add_balance_apply {
     }
 }
 
+pub fn transfer_exec_fee(
```
bowenwang1996, line 73: I am confused. You implement this here and again in `VMLogic`?

runtime/near-vm-logic/src/logic.rs
Longarithm: Done

runtime/runtime/src/actions.rs
Longarithm: Done. I was sure that previously I got "unused variable"-like warnings.

runtime/runtime/src/actions.rs
Longarithm: Done

runtime/runtime/src/config.rs
Longarithm: Yes, that's right

runtime/near-vm-logic/src/logic.rs
Longarithm: Yes, done now

runtime/runtime/src/config.rs
Longarithm: Yes. I would unite these functions, but the version in `VMLogic` is supposed to throw `HostError::IntegerOverflow` in case of overflows, and version in `Runtime` doesn't.
Is there actually any confidence that addition of such fees will never cause overflow? If so, I could just reuse this function.

test-utils/testlib/src/fees_utils.rs
Longarithm: Done

runtime/runtime/src/actions.rs
Longarithm: Discussed offline: receipt ids are assigned to all `new_receipts` in `apply_action_receipt` using `create_receipt_id_from_receipt` function.

runtime/runtime/src/config.rs
Longarithm: Discussed offline, code duplication was removed

runtime/near-vm-logic/src/logic.rs
Longarithm: Moved to `primitives-core/src/runtime/fees.rs`, because the `runtime` package depends on `near-vm-logic`.

runtime/runtime/src/actions.rs
matklad: Yeah, that's the differences between `cfg!()` and `#[cfg()]`. Consider 

```rust
fn a(param: u32) {
    if cfg!(cond) {
        println!("{}", param)
    }
}

fn b(param: u32) {
    #[cfg(cond)] 
    {
        println!("{}", param)
    }
}
```

Depending on the compile-time value of `cond`, a looks to the compiler like this:

```rust
fn a(param: u32) {
    if true {
        println!("{}", param)
    }
}

fn a(param: u32) {
    if false {
        println!("{}", param)
    }
}
```

Both versions "use" `param`, it's just that one usage is dead code (behind `if false`) and is eliminated during codegen.

In contrast, the `#[cfg]` version looks likes this:

```rust
fn b(param: u32) {
    {
        println!("{}", param)
    }
}

fn b(param: u32) {
}
```

depending on `cfg`, param is used or unused, so you might get a warning depending on specific compilation flags.

runtime/near-vm-logic/src/logic.rs
```diff
@@ -1481,31 +1484,18 @@ impl<'a> VMLogic<'a> {
...
-                )?;
-            }
-        }
-        self.gas_counter.pay_action_base(
-            &self.fees_config.action_creation_config.transfer_cost,
-            sir,
-            ActionCosts::transfer,
-        )?;
+        let receiver_id = self.get_account_by_receipt(&receipt_idx);
+        let is_receiver_implicit = self.current_protocol_version
```
EgorKulikov, line 1488: Consider using checked_feature! macro

runtime/runtime/src/actions.rs
Longarithm: Thank you for the clarification!

Longarithm:
> Can you add a transaction to delete account and create implicit account in `pytests/sanity/upgradable.py`?

Done @bowenwang1996

Longarithm:
Spec update: https://github.com/near/NEPs/pull/196
Nayduck run: http://nayduck.eastus.cloudapp.azure.com:3000/#/run/1503

Newly failed tests: 
1 in Debug - test_consensus_with_epoch_switches
3 in Debug/Nightly - rpc_state_changes, skip_epoch, cross_shard_tx
All passed locally.

runtime/near-vm-logic/src/logic.rs
```diff
@@ -1481,31 +1484,18 @@ impl<'a> VMLogic<'a> {
...
+            >= IMPLICIT_ACCOUNT_CREATION_PROTOCOL_VERSION
```
EgorKulikov, line 1489: Consider introducing fn is_implicit_account_creation_enabled in version.rs

EgorKulikov (approved):

runtime/near-vm-logic/src/logic.rs
Longarithm: Done

pytest/lib/branches.py
```diff
@@ -87,6 +87,8 @@ def download_binary(uname, branch):
     """Download binary for given platform and branch."""
     url = f'https://s3-us-west-1.amazonaws.com/build.nearprotocol.com/nearcore/{uname}/{branch}/near'
     print(f'Downloading near & state-viewer for {branch}@{uname}')
+    # TODO: make platform-independent.
```
bowenwang1996, line 90: you can check what the underlying OS is and do things accordingly, although there may be a better solution. @chefsale @frol

test-utils/testlib/src/standard_test_cases.rs
```diff
@@ -1196,7 +1199,21 @@ pub fn test_fail_not_enough_balance_for_storage(node: impl Node) {
...
+    let node_user = node.user();
+    let _ = node_user.create_account(
+        alice_account(),
+        eve_dot_alice_account(),
+        node.signer().public_key(),
+        money_used,
+    );
+    let transaction_result =
+        node_user.delete_account(eve_dot_alice_account(), eve_dot_alice_account()).unwrap();
+    assert_eq!(transaction_result.status, FinalExecutionStatus::SuccessValue(to_base64(&[])));
```
bowenwang1996, line 1213:
```suggestion
    assert_eq!(transaction_result.status, FinalExecutionStatus::SuccessValue(String::new()));
```

test-utils/testlib/src/standard_test_cases.rs
```diff
@@ -1196,7 +1199,21 @@ pub fn test_fail_not_enough_balance_for_storage(node: impl Node) {
...
+}
```
bowenwang1996, line 1214: Please also assert that the account is actually deleted

runtime/near-vm-logic/src/logic.rs
```diff
@@ -1268,6 +1266,14 @@ impl<'a> VMLogic<'a> {
         self.checked_push_promise(Promise::Receipt(new_receipt_idx))
     }
 
+    /// Helper function to return the account id towards which the receipt is directed.
+    fn get_account_by_receipt(&self, receipt_idx: &ReceiptIndex) -> AccountId {
+        self.receipt_to_account
+            .get(receipt_idx)
+            .expect("promises and receipt_to_account should be consistent.")
+            .clone()
```
bowenwang1996, line 1274: why do you need to clone here?

runtime/near-vm-logic/src/logic.rs
```diff
@@ -1736,7 +1726,7 @@ impl<'a> VMLogic<'a> {
     /// # Cost
     ///
     /// `burnt_gas := base + dispatch action base fee + dispatch action per byte fee * num bytes + cost of reading and parsing account id from memory `
-    /// `used_gas := burnt_gas + exec action base fee + exec action per byte fee * num bytes`
+    /// `used_gas := burnt_gas + exec action base fee + exec action per byte fee * num bytes + receipt creation fees + transfer fees`
```
bowenwang1996, line 1729: those fees are only added if the beneficiary is an implicit account

runtime/runtime/src/actions.rs
```diff
@@ -517,9 +520,45 @@ pub(crate) fn action_delete_account(
...
+                );
+            result.gas_burnt += exec_gas;
+            result.gas_used += exec_gas
+                + config.action_receipt_creation_config.exec_fee()
+                + transfer_exec_fee(&config.action_creation_config, is_receiver_implicit);
+
+            result.new_receipts.push(Receipt {
+                predecessor_id: account_id.clone(),
+                receiver_id: delete_account.beneficiary_id.clone(),
+                receipt_id: CryptoHash::default(),
```
bowenwang1996, line 546: Could you add a comment here on where the receipt id is actually set?

runtime/runtime/src/config.rs
```diff
@@ -189,6 +176,44 @@ pub fn exec_fee(
...
+                let is_receiver_implicit =
+                    is_implicit_account_creation_enabled(current_protocol_version)
+                        && is_account_id_64_len_hex(&beneficiary_id);
+                config.action_receipt_creation_config.send_fee(sender_is_receiver)
+                    + config.action_receipt_creation_config.exec_fee()
+                    + transfer_send_fee(
+                        &config.action_creation_config,
+                        sender_is_receiver,
+                        is_receiver_implicit,
+                    )
```
bowenwang1996, line 205: This doesn't feel right. Could you explain why we need to pay action_receipt_creation fee if the beneficiary id is not implicit?

Longarithm: This is because now we create `ActionReceipt` with `Transfer` in any case (actions.rs:543) instead of refund.
If beneficiary is not implicit, we just pay for it, otherwise we also pay for implicit account creation.

test-utils/testlib/src/standard_test_cases.rs
Longarithm: Done

runtime/near-vm-logic/src/logic.rs
Longarithm: I thought about returning `AccountId` without reference to simplify code at some place. It didn't help, but I forgot to revert this change.

runtime/runtime/src/config.rs
bowenwang1996: I see. Why did we decide to do this?

Longarithm: I don't understand why this transfer should be free for remover, because it is the action they trigger and thus have to pay for it. I guess when the tokens might be burned it might not be fair, but now tokens are always transferred to some destination. 
Another minor opinion is that introducing a condition on account type on `action_delete_account` side will make the logic harder to understand. 
@evgenykuzyakov it was actually your change :) Is there another logic behind it?

pytest/lib/branches.py
Longarithm: Seems like '\=https' worked.

Longarithm: Done

runtime/runtime/src/config.rs
bowenwang1996: After thinking more about this I think the change makes sense. We should not generate a refund (which is free) to begin with.

bowenwang1996 (approved):

PR #4170, opened 2021-03-26
fix: off by one in db refcounting

matklad:
This feels like a typo to me, but I have next to zero understanding of this code, so I would prefer that someone else takes over this, if this is a bug at all.

### DIFF
--- a/core/store/src/db/refcount.rs
+++ b/core/store/src/db/refcount.rs
@@ -50,8 +50,8 @@ pub(crate) fn merge_refcounted_records(result: &mut Vec<u8>, val: &[u8]) {
 }
 
 /// Returns
-/// (Some(value), rc) if rc > 0
-/// (None, rc) if rc <= 0
+/// (Some(value), rc) if rc >= 0
+/// (None, rc) if rc < 0
 pub fn decode_value_with_rc(bytes: &[u8]) -> (Option<&[u8]>, i64) {
     if bytes.len() < 8 {
         debug_assert!(bytes.is_empty());


bowenwang1996 (commented):
I believe this is intentional. @mikhailOK can explain more

matklad:
I suspected that! then, the comment above the function might need an update?

mikhailOK:
It's sort of impossible case. `rc == 0` is serialized as empty:
```
/// Refcount record format:
/// rc = 0 => empty
/// rc < 0 => 8 bytes little endian rc
/// rc > 0 => value followed by 8 bytes little endian rc
```

`merge_refcounted_records` and `encode_value_with_rc` have this invariant, so before that line of code we can have `debug_assert!(rc != 0)`.

Also, `rc < 0` is possible in individual records we insert, but never in the output of `merge_refcounted_records` because rocksdb merges in the order of insertion.

matklad:
Makes sense! Changed the commit to fix the comment instead!

core/store/src/db/refcount.rs
```diff
@@ -50,8 +50,8 @@ pub(crate) fn merge_refcounted_records(result: &mut Vec<u8>, val: &[u8]) {
 }
 
 /// Returns
-/// (Some(value), rc) if rc > 0
-/// (None, rc) if rc <= 0
+/// (Some(value), rc) if rc >= 0
+/// (None, rc) if rc < 0
```
bowenwang1996, line 54: but if rc == 0 the value doesn't exist, so the orginal comment seems correct?

matklad: Hmmmm, right, we then hit the first branch, and return `(None, 0)`, that makes sesne!

PR #4173, opened 2021-03-26
fix: prevent LLVM from optimizing the benchmark to O(1) computation

matklad:
LLVM is able to compute closed form for polynomial sums like

    let mut sum = 0i
    for i in 0..n {
        sum += a_0 + a_1*i + a_2*i*i ... + a_k*i^k;
    }

https://godbolt.org/z/4Wx7G8vEx

### DIFF
--- a/runtime/near-test-contracts/test-contract-rs/src/lib.rs
+++ b/runtime/near-test-contracts/test-contract-rs/src/lib.rs
@@ -477,7 +477,12 @@ pub unsafe fn sum_n() {
 
     let mut sum = 0u64;
     for i in 0..n {
-        sum += i;
+        // LLVM optimizes sum += i into O(1) computation, use volatile to thwart
+        // that.
+        unsafe {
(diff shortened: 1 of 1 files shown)


ailisp (approved):

bowenwang1996 (approved):

runtime/near-test-contracts/test-contract-rs/src/lib.rs
```diff
@@ -477,7 +477,8 @@ pub unsafe fn sum_n() {
 
     let mut sum = 0u64;
     for i in 0..n {
-        sum += i;
+        // NOTE: just sum += i optimizes into O(1) computation.
```
olonho, line 480: Another option is to do something like naive prime factorization and adjust results per known complexity.

matklad:
Discussion with olonho revealed a more direct approach -- using volatile opos.

PR #4176, opened 2021-03-29
(feat) Account V2 and storage usage recalculation

EgorKulikov:
When we touch some account for the first time after this is introduced (i. e. when we naturally will update it's storage usage) we will migrate it to AccountV2 and recalculate it's storage usage from scratch

### Test plan
See test_account_migration


runtime/runtime/src/actions.rs
```diff
@@ -686,11 +711,73 @@ pub(crate) fn check_account_existence(
...
+                usage += usage_config.num_extra_bytes_record
+                    + public_key.try_to_vec().unwrap().len() as u64
+                    + get_access_key(state_update, account_id, &public_key)
+                        .unwrap()
+                        .try_to_vec()
+                        .unwrap()
+                        .len() as u64;
+                Ok(())
+            })
+            .collect::<Result<Vec<_>, _>>()?;
```
matklad, line 746: Why are we collecting to a Vec and throwing the vector itself away? Is this just for side-effects (exhausting the iterator)? In that case, you can use `try_for_each` instead of `map/collect`. 

Useless bit of trivia: `.collect::<Result<(), _>>` would also work. Collecting into `()` works like a for/for_each.

EgorKulikov: That's exactly what I needed, thanks. I guess I just wanted to make that work and then forgot to clean up

runtime/runtime/src/actions.rs
```diff
@@ -417,11 +430,21 @@ pub(crate) fn action_implicit_account_creation_transfer(
...
 
 pub(crate) fn action_deploy_contract(
+    apply_state: &ApplyState,
     state_update: &mut TrieUpdate,
     account: &mut Account,
     account_id: &AccountId,
     deploy_contract: &DeployContractAction,
 ) -> Result<(), StorageError> {
+    #[cfg(feature = "protocol_feature_add_account_versions")]
+    recalculate_usage(
```
bowenwang1996, line 440: protocol version also needs to be checked here

runtime/runtime/src/actions.rs
```diff
@@ -687,11 +714,74 @@ pub(crate) fn check_account_existence(
...
+    usage_config: &StorageUsageConfig,
+    state_update: &TrieUpdate,
+) -> Result<(), StorageError> {
+    if let Account::AccountV1(_) = account {
+        // As this is valid account that fits into state we can use unchecked operations here
+        let mut usage: StorageUsage = usage_config.num_bytes_account;
+        if let Some(code) = get_code(state_update, account_id, Some(account.code_hash()))? {
+            usage += code.code.len() as u64;
+        }
+        state_update
```
bowenwang1996, line 730: what is the reason for going over all the account storage? Couldn't we simply check the access key part?

bowenwang1996 (commented):
It seems to be that this feature can be combined with account version v2 (under the same feature flag) since they will be stabilized together

runtime/runtime/src/actions.rs
EgorKulikov: We do not store storage usage by type anywhere, only total

runtime/runtime/src/actions.rs
EgorKulikov: Done

PR #4180, opened 2021-03-30
refactor(jsonrpc): Refactor broadcast_tx jsonrpc handler to add structured errors

khorolets:
This PR introduces structured errors to JSON RPC methods which deals with transaction broadcasting:
* `broadcast_tx_async`
* `broadcast_tx_commit`
* `EXPERIMENTAL_broadcast_tx_sync`
* `EXPERIMENTAL_check_tx`
* `EXPERIMENTAL_tx_status`
* `tx`


chain/jsonrpc-primitives/src/types/transactions.rs
```diff
@@ -0,0 +1,119 @@
...
+}
+
+#[derive(thiserror::Error, Debug)]
+pub enum RpcTransactionError {
+    // TxStatusError
+    // NetworkClientResponses::InvalidTx
+    //NetworkClientResponses::NoResponse
+    // DoesNotTrackShard,
+    // Ban { ban_reason: ReasonForBan },
+    // ServerError (Timeout, Closed)
```
bowenwang1996, line 17: What happened here

chain/jsonrpc-primitives/src/types/transactions.rs
```diff
@@ -0,0 +1,119 @@
...
+    pub final_execution_outcome: near_primitives::views::FinalExecutionOutcomeViewEnum,
+}
+
+#[derive(Debug)]
+pub struct RpcTransactionCheckValidResponse;
+
+#[derive(Serialize, Deserialize, Debug)]
+pub struct RpcBroadcastTxSyncResponse {
+    pub transaction_hash: String,
+    pub is_routed: bool,
```
bowenwang1996, line 51: I thought we disabled routing?

chain/jsonrpc-primitives/src/types/transactions.rs
khorolets: Sorry it's WIP

chain/jsonrpc-primitives/src/types/transactions.rs
khorolets: Only in query. I haven't heard of disabling routing for broadcasting transactions.

khorolets:
@vgrichina 

We have methods:
* `EXPERIMENTAL_broadcast_tx_sync`
* `EXPERIMENTAL_check_tx`

I assume they were created for Wallet needs. Currently `EXPERIMENTAL_check_tx` returns OK with empty body in case if everything is and error in other cases. `EXPERIMENTAL_broadcast_tx_sync` returns OK with tx hash and `is_routed: bool` in case if tx is valid and in case if tx was routed.

@frol and I suggest to return an error in case of tx was routed as well because from an end-user perspective it was unable to check if tx is valid because it was routed and the user needs to check tx to ensure it is valid. Any objections if we throw an error there?

Another suggestion to return the same structure for `EXPERIMENTAL_check_tx` as we return in `EXPERIMENTAL_broadcast_tx_sync` instead of empty body.

/cc @bowenwang1996

chain/jsonrpc-primitives/src/types/transactions.rs
bowenwang1996: @frol @khorolets it is effective disabled because we disabled receipt responses https://github.com/near/nearcore/blob/ab4f0c1c481b503e725798e994a58964ed6bc49b/chain/client/src/view_client.rs#L984. After you route the transaction, you won't be able to get its status

bowenwang1996:
> @frol and I suggest returning an error in case tx was routed as well because from an end-user perspective it was unable to check if tx is valid because it was routed and the user needs to check tx to ensure it is valid. Any objections if we throw an error there?
>
> Another suggestion to return the same structure for EXPERIMENTAL_check_tx as we return in EXPERIMENTAL_broadcast_tx_sync instead of empty body.

Sounds good to me

chain/jsonrpc-primitives/src/types/transactions.rs
frol: @bowenwang1996 and we don't try to get its status here, as you can see, if `RequestRouted` is returned, we give up on waiting and return it to the user. Do you mean that we cannot even get `RequestRouted` result?

bowenwang1996: I see. I think I misremembered how it works

chain/jsonrpc-primitives/src/types/transactions.rs
```diff
@@ -0,0 +1,152 @@
...
+            if !near_runtime_utils::is_valid_account_id(&account_id) {
+                return Err(crate::errors::RpcParseError(format!(
+                    "Invalid account id: {}",
+                    account_id
+                )));
+            }
+            let transaction_info = TransactionInfo::TransactionId { hash, account_id };
+            Ok(Self { transaction_info })
+        } else {
+            // TODO: move this code to crate::utils::parse_signed_transaction
```
frol, line 87: When TODO? I think it makes sense to inline `from_base64_or_parse_err` and extract this code into `parse_signed_transaction` as you suggest :+1:

chain/jsonrpc-primitives/src/types/transactions.rs
```diff
@@ -0,0 +1,152 @@
+use near_primitives::borsh::BorshDeserialize;
+use serde::{Deserialize, Serialize};
+use serde_json::Value;
+
+#[derive(Debug, Clone)]
+pub struct RpcTransactionRequest {
```
frol, line 6: I suggest we rename this to `RpcBroadcastTransactionRequest` to avoid confusion (I had to look around to make sure it is indeed used for broadcast endpoints (well, it is also used for `check_tx`, but that is fine, I believe, to use `RpcBroadcastTransactionRequest` there as well)

frol (approved):

chain/jsonrpc-primitives/src/types/transactions.rs
```diff
@@ -0,0 +1,132 @@
...
+            near_client_primitives::types::TxStatusError::ChainError(err) => {
+                Self::InternalError { debug_info: format!("{:?}", err) }
+            }
+            near_client_primitives::types::TxStatusError::MissingTransaction(
+                requested_transaction_hash,
+            ) => Self::UnknownTransaction { requested_transaction_hash },
+            near_client_primitives::types::TxStatusError::InvalidTx(context) => {
+                Self::InvalidTransaction { context }
+            }
+            near_client_primitives::types::TxStatusError::InternalError => {
```
frol, line 97: Let's add `debug_message` to the `TxStatusError::InternalError` variant, so we can pass it down

chain/jsonrpc-primitives/src/types/transactions.rs
```diff
@@ -0,0 +1,132 @@
...
+#[derive(Serialize, Deserialize, Debug)]
+pub struct RpcTransactionResponse {
+    #[serde(flatten)]
+    pub final_execution_outcome: near_primitives::views::FinalExecutionOutcomeViewEnum,
+}
+
+#[derive(Serialize, Deserialize, Debug)]
+pub struct RpcBroadcastTxSyncResponse {
+    pub transaction_hash: String,
+    pub is_routed: bool,
```
frol, line 54: I feel that `is_routed` is misleading, and it is better to just drop it and only return `transaction_hash` while the method is still EXPERIMENTAL.

frol (commented):
1. I have minor suggestions in the comments
2. I want to ask you to add notable changes to the `chain/jsonrpc/CHANGELOG`

chain/jsonrpc-primitives/src/types/transactions.rs
khorolets: Do we still want to have a structure for that instead of returning `CryptoHash` on the jsonrpc side?

frol: @khorolets yes, we still want to have a structure since with structure we always have an opportunity to provide extra context if necessary in the future.

chain/jsonrpc-primitives/src/types/transactions.rs
```diff
@@ -0,0 +1,131 @@
...
+    #[error("The node reached its limits. Try again later. More details: {debug_info}")]
+    InternalError { debug_info: String },
+    #[error("Timeout")]
+    TimeoutError,
+    // NOTE: Currently, the underlying errors are too broad, and while we tried to handle
+    // expected cases, we cannot statically guarantee that no other errors will be returned
+    // in the future.
+    // TODO #3851: Remove this variant once we can exhaustively match all the underlying errors
+    #[error("It is a bug if you receive this error type, please, report this incident: https://github.com/near/nearcore/issues/new/choose. Details: {error_message}")]
+    Unreachable { error_message: String },
```
frol, line 42: Do we need it? I don't see it ever being constructed.

chain/jsonrpc/CHANGELOG.md
```diff
@@ -1,5 +1,21 @@
...
+  * `broadcast_tx_commit`
+  * `EXPERIMENTAL_broadcast_tx_sync`
+  * `EXPERIMENTAL_check_tx`
+  * `EXPERIMENTAL_tx_status`
+  * `tx`
+
+## Breaking changes
+
+Response from `EXPERIMENTAL_broadcast_tx_sync` and `EXPERIMENTAL_check_tx` doesn't return `is_routed` 
+field anymore. In case if transaction wes routes an error will be returned. Also `EXPERIMENTAL_check_tx`
```
frol, line 16:
```suggestion
field anymore. In case of a transaction getting routed an error will be returned. Also, `EXPERIMENTAL_check_tx` 
```

chain/jsonrpc/CHANGELOG.md
```diff
@@ -1,5 +1,21 @@
...
+  * `EXPERIMENTAL_tx_status`
+  * `tx`
+
+## Breaking changes
+
+Response from `EXPERIMENTAL_broadcast_tx_sync` and `EXPERIMENTAL_check_tx` doesn't return `is_routed` 
+field anymore. In case if transaction wes routes an error will be returned. Also `EXPERIMENTAL_check_tx` 
+returns response with transaction hash instead of empty body.
+
 ## Unreleased
```
frol, line 19: Let's remove `## Unreleased`. The changes below are included in this 0.2.1 release. I know we missed quite a few changes, we will need to get better at writing changelogs.

chain/jsonrpc/src/lib.rs
```diff
@@ -530,79 +551,90 @@ impl JsonRpcHandler {
...
-                Err(RpcError::server_error(Some(ServerError::TxExecutionError(err.into()))))
+                Err(near_jsonrpc_primitives::types::transactions::RpcTransactionError::RequestRouted {
+                    transaction_hash: (request_data.signed_transaction.get_hash()).to_string(),
+                })
             }
-            NetworkClientResponses::DoesNotTrackShard => {
-                Err(RpcError::server_error(Some(does_not_track_shard_err.to_string())))
+            network_client_responses=> Err(
+                near_jsonrpc_primitives::types::transactions::RpcTransactionError::from_network_client_responses(network_client_responses)
+            )
```
frol, line 574: I wonder why this is not getting formatted properly

chain/jsonrpc/src/lib.rs
```diff
@@ -530,79 +551,90 @@ impl JsonRpcHandler {
...
-                self.tx_polling(TransactionInfo::Transaction(tx)).await
-            }
-            NetworkClientResponses::InvalidTx(err) => {
-                Err(RpcError::server_error(Some(ServerError::TxExecutionError(err.into()))))
+                self.tx_polling(near_jsonrpc_primitives::types::transactions::TransactionInfo::Transaction(tx)).await
             }
-            NetworkClientResponses::NoResponse => {
-                Err(RpcError::server_error(Some(ServerError::Timeout)))
+            network_client_response=> {
+                Err(near_jsonrpc_primitives::types::transactions::RpcTransactionError::from_network_client_responses(network_client_response))
```
frol, line 636: Formatting here is off as well

frol (approved):

PR #4181, opened 2021-03-31
feat: rework contract runtime error processing

olonho:
Make sure errors processed in the similar manner between Wasmer 0.x and Wasmer 1.x.
Introduced explicit non-deterministic error.

Test plan
---------
Reworked error cases tests to make them uniform, then `cargo test`.


runtime/near-vm-runner/src/wasmer1_runner.rs
```diff
@@ -87,26 +90,140 @@ impl IntoVMError for wasmer::InstantiationError {
     }
 }
 
+// Copy-pasted, shall be exported by Wasmer instead.
```
matklad, line 93: Hm, it is public? https://docs.rs/wasmer-vm/1.0.2/wasmer_vm/enum.TrapCode.html

runtime/near-vm-errors/src/lib.rs
```diff
@@ -389,6 +383,9 @@ impl fmt::Display for WasmTrap {
             WasmTrap::GenericTrap => write!(f, "Generic trap."),
             WasmTrap::BreakpointTrap => write!(f, "Breakpoint trap."),
             WasmTrap::StackOverflow => write!(f, "Stack overflow."),
+            WasmTrap::NondeterministicTrap(message) => {
+                write!(f, "Non-deterministic trap: {}", message)
+            }
```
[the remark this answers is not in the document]
olonho: Reworked, but requires looking on that in separate PR.

runtime/near-vm-runner/src/wasmer1_runner.rs
olonho: Indeed, fixed.

olonho: Indeed, fixed.

runtime/near-vm-errors/src/lib.rs
```diff
@@ -34,12 +32,8 @@ pub enum FunctionCallError {
...
     HostError(HostError),
     EvmError(EvmError),
-    /// An error message when wasmer 1.0 returns a wasmer::RuntimeError
-    WasmerRuntimeError(String),
-    /// A trap in Wasmer 1.0, not same as WasmTrap above, String is a machine readable form like "stk_ovf"
-    /// String is used instead of wasmer internal enum is because of BorshSerializable.
-    /// It can be convert back by wasmer_vm::TrapCode::from_str
-    Wasmer1Trap(String),
+    /// Non-deterministic error.
+    Nondeterministic(String),
```
bowenwang1996, line 36: Is this error deterministic, i.e, is it guaranteed that a nondeterministic error will be wrapped in this variant?

runtime/near-vm-runner/src/wasmer1_runner.rs
```diff
@@ -94,19 +96,58 @@ impl IntoVMError for wasmer::RuntimeError {
...
-                    // A trap
-                    VMError::FunctionCallError(FunctionCallError::Wasmer1Trap(
-                        trap_code.to_string(),
-                    ))
-                } else {
-                    // A general error
-                    VMError::FunctionCallError(FunctionCallError::WasmerRuntimeError(error_msg))
-                }
-            }
+            Ok(e) => return (&e).into(),
```
bowenwang1996, line 99: can probably just do an `if let` here

runtime/near-vm-runner/Cargo.toml
```diff
@@ -22,6 +22,7 @@ wasmer-types = { version = "1.0.2", optional = true }
 wasmer-compiler-singlepass = { version = "1.0.2", optional = true }
 wasmer-compiler-cranelift = { version = "1.0.2", optional = true }
 wasmer-engine-native = { version = "1.0.2", optional = true }
+wasmer-vm = { version = "1.0.2" }
```
matklad, line 25: should probably be `optional = true` for now, like the rest of wasmer 1 crates?

runtime/near-vm-runner/src/wasmer1_runner.rs
```diff
@@ -94,19 +96,58 @@ impl IntoVMError for wasmer::RuntimeError {
...
+            }
+            Some(TrapCode::Interrupt) => {
+                FunctionCallError::Nondeterministic("Wasmer interrupt".to_string())
+            }
+            Some(TrapCode::UnalignedAtomic) => {
+                FunctionCallError::WasmTrap(WasmTrap::MisalignedAtomicAccess)
+            }
+            Some(TrapCode::VMOutOfMemory) => {
+                FunctionCallError::Nondeterministic("Wasmer out of memory".to_string())
+            }
```
matklad, line 147:
```suggestion
            Some(TrapCode::UnalignedAtomic) => {
                FunctionCallError::WasmTrap(WasmTrap::MisalignedAtomicAccess)
            }
            Some(TrapCode::Interrupt) => {
                FunctionCallError::Nondeterministic("Wasmer interrupt".to_string())
            }
            Some(TrapCode::VMOutOfMemory) => {
                FunctionCallError::Nondeterministic("Wasmer out of memory".to_string())
            }
```

nit: can group non-deterministic errors together.

runtime/near-vm-runner/src/wasmer_runner.rs
```diff
@@ -114,8 +113,11 @@ impl IntoVMError for wasmer_runtime::error::RuntimeError {
...
                 // A trap that Wasmer knows about occurred.
                 // As of 0.17.1, can be thrown on C signals caught, for example OOM.
-                InvokeError::TrapCode { code: _, srcloc: _ } => {
-                    VMError::FunctionCallError(WasmUnknownError)
+                InvokeError::TrapCode { code, srcloc } => {
+                    VMError::FunctionCallError(FunctionCallError::Nondeterministic(format!(
+                        "Wasmer trap {} at {}",
+                        *code as u32, srcloc
+                    )))
                 }
```
matklad, line 121: Hm, from the light reading of the code, I am not sure that this should *always* be nondeterministic. It seems that, eg, division by zero generates this trap, no? Do we have a test for division by zero?

runtime/runtime/src/actions.rs
```diff
@@ -191,10 +191,21 @@ pub(crate) fn action_function_call(
...
-        Some(VMError::FunctionCallError(err)) => {
-            result.result = Err(ActionErrorKind::FunctionCallError(err).into());
-            false
-        }
+        Some(VMError::FunctionCallError(err)) => match err {
+            FunctionCallError::Nondeterministic(msg) => {
+                panic!("Contract runner returned non-deterministic error '{}', aborting", msg)
+            }
+            FunctionCallError::CompilationError(_)
+            | FunctionCallError::LinkError { msg: _ }
```
matklad, line 199: Unrelated to the PR, but:

* LinkError should probably be a tuple -- it's the only one with `msg` field, which feels inconsistent
* I am not entirely sure we want to distinguish between CompilationError and LinkError, they feel sufficiently similar to me in the context of wasm execution.

runtime/near-vm-runner/Cargo.toml
olonho: Would love to, but unfortunately cannot get it compiled in such a way.

runtime/near-vm-errors/src/lib.rs
ailisp, line 36: I doubt of still keep it WasmUnkownError is a good for wasmer 1. The internal of it is really unknown in wasmer 0.x (UnknownTrap etc.) but it's indeed known in wasmer 1. I suggest name it as WasmerRuntimeError or WasmerInternalError

runtime/near-vm-runner/src/wasmer1_runner.rs
```diff
@@ -94,19 +96,58 @@ impl IntoVMError for wasmer::RuntimeError {
...
+            None => FunctionCallError::Nondeterministic(error_msg),
```
ailisp, line 148: This is incorrect, this only happens when wasmer returns a Generic Error or User Error that failed to downcast to VMLogicError: https://github.com/wasmerio/wasmer/blob/9e0122e1f7bdb6fedf5fc17cc9a3a0fda30031a0/lib/engine/src/trap/error.rs#L18
And Generic Error & User Error must come from we construct it manually in contract, which is unreachable, because we only construct VMLogicError from there. So this is not an Nondeterministic, should be debug_assert!(false) or unreachable!()

runtime/near-vm-runner/src/wasmer_runner.rs
```diff
@@ -135,7 +137,9 @@ impl IntoVMError for wasmer_runtime::error::RuntimeError {
                 // upon the middleware or backend being used.
                 // As of 0.17.0, thrown only from Singlepass BE and wraps RuntimeError
                 // instance.
-                InvokeError::Breakpoint(_) => VMError::FunctionCallError(WasmTrap(BreakpointTrap)),
+                InvokeError::Breakpoint(_) => VMError::FunctionCallError(
+                    FunctionCallError::Nondeterministic("Wasmer breakpoint".to_string()),
+                ),
```
ailisp, line 142: Why break point is nondeterministic error?

ailisp:
Overall looks good, this specific one is questionable: https://github.com/near/nearcore/pull/4181#discussion_r605937213

runtime/near-vm-errors/src/lib.rs
olonho: Yes, now it is.

runtime/near-vm-runner/src/wasmer_runner.rs
olonho: Correct, reworked that piece.

runtime/runtime/src/actions.rs
olonho: Guess so, but this PR already become bigger than I want it to be.

runtime/near-vm-errors/src/lib.rs
olonho: Reworked, so that `Nondeterministic` is properly returned.

runtime/near-vm-runner/src/wasmer1_runner.rs
olonho: Reworked code around this place.

runtime/near-vm-runner/src/wasmer_runner.rs
olonho: Because it's not something decided by the contract, I presume.

runtime/near-vm-runner/src/wasmer1_runner.rs
```diff
@@ -94,19 +96,57 @@ impl IntoVMError for wasmer::RuntimeError {
...
+        let error = match trap_code {
+            Some(TrapCode::StackOverflow) => FunctionCallError::WasmTrap(WasmTrap::StackOverflow),
+            Some(TrapCode::HeapSetterOutOfBounds) => {
+                FunctionCallError::WasmTrap(WasmTrap::MemoryOutOfBounds)
+            }
+            Some(TrapCode::HeapAccessOutOfBounds) => {
+                FunctionCallError::WasmTrap(WasmTrap::MemoryOutOfBounds)
+            }
+            Some(TrapCode::HeapMisaligned) => {
+                FunctionCallError::WasmTrap(WasmTrap::MisalignedAtomicAccess)
```
matklad, line 111: nit: `WasmTrap::IllegalAlignment` probably makes more sense.

runtime/near-vm-runner/src/wasmer1_runner.rs
```diff
@@ -94,19 +96,57 @@ impl IntoVMError for wasmer::RuntimeError {
...
+            Some(TrapCode::UnalignedAtomic) => {
+                FunctionCallError::WasmTrap(WasmTrap::MisalignedAtomicAccess)
+            }
+            Some(TrapCode::Interrupt) => {
+                FunctionCallError::Nondeterministic("Wasmer interrupt".to_string())
+            }
+            Some(TrapCode::VMOutOfMemory) => {
+                FunctionCallError::Nondeterministic("Wasmer out of memory".to_string())
+            }
+            None => panic!("Unknown error: {}", error_msg),
```
matklad, line 147: nit: I'd pull that out out the match:

```
let trap_code = trap_code.unwrap_or_else(|| panic!("Unknown error: {}", error_msg))
```

No need to wrap each arm into Some that way.

runtime/near-vm-runner/src/wasmer_runner.rs
```diff
@@ -102,7 +100,9 @@ impl IntoVMError for wasmer_runtime::error::RuntimeError {
                 // invoke returns false and doesn't fill error info what Singlepass BE doesn't.
                 // Failed unwinder may happen in the case of deep recursion/stack overflow.
                 // Also can be thrown on unreachable instruction, which is quite unfortunate.
-                InvokeError::FailedWithNoError => VMError::FunctionCallError(WasmUnknownError),
+                InvokeError::FailedWithNoError => VMError::FunctionCallError(
+                    FunctionCallError::Nondeterministic("FailedWithNoError".to_string()),
+                ),
```
matklad, line 105: Hm, the following two cases from the comments worry me: 

> Failed unwinder may happen in the case of deep recursion/stack overflow.
> Also can be thrown on unreachable instruction, which is quite unfortunate.

I see that we have tests for these cases, so presumably these don't result in non-deterministic errors, and the comment is wrong? Or is there some specific pattern of stack overflow which might cause us to always hit this case?

runtime/near-vm-runner/src/wasmer_runner.rs
matklad: I wonder if we should `unreachable!()` here?

runtime/near-vm-runner/src/wasmer_runner.rs
```diff
@@ -113,9 +113,16 @@ impl IntoVMError for wasmer_runtime::error::RuntimeError {
...
+                InvokeError::TrapCode { code, srcloc } => {
+                    VMError::FunctionCallError(match *code as u32 {
+                        0 /* Unreachable */ => FunctionCallError::WasmTrap(WasmTrap::Unreachable),
+                        1 /* IncorrectCallIndirectSignature */ => FunctionCallError::WasmTrap(WasmTrap::IncorrectCallIndirectSignature),
+                        2 /* MemoryOutOfBounds */ => FunctionCallError::WasmTrap(WasmTrap::MemoryOutOfBounds),
+                        3 /* CallIndirectOOB */ => FunctionCallError::WasmTrap(WasmTrap::CallIndirectOOB),
+                        4 /* IllegalArithmetic */ => FunctionCallError::WasmTrap(WasmTrap::IllegalArithmetic),
+                        5 /* MisalignedAtomicAccess */ => FunctionCallError::WasmTrap(WasmTrap::MisalignedAtomicAccess),
+                        _ => FunctionCallError::WasmUnknownError(format!("Wasmer trap {} at {}", code, srcloc)),
+                    })
```
matklad, line 125:
```suggestion
                    VMError::FunctionCallError(FunctionCallError::WasmTrap(match code {
                        ExceptionCode::Unreachable => WasmTrap::Unreachable,
                        ExceptionCode::IncorrectCallIndirectSignature => WasmTrap::IncorrectCallIndirectSignature,
                        ExceptionCode::MemoryOutOfBounds => WasmTrap::MemoryOutOfBounds,
                        ExceptionCode::CallIndirectOOB => WasmTrap::CallIndirectOOB,
                        ExceptionCode::IllegalArithmetic => WasmTrap::IllegalArithmetic,
                        ExceptionCode::MisalignedAtomicAccess => WasmTrap::MisalignedAtomicAccess,
                    }))

```

This seems to work?

runtime/near-vm-runner/src/wasmer1_runner.rs
olonho: Not sure if we have it, atm.

runtime/near-vm-runner/src/wasmer1_runner.rs
olonho: Done.

runtime/near-vm-runner/src/wasmer_runner.rs
olonho: It's called from two places: `call_func_with_index_inner` when `error_out` is None, or in `signal_trap_handler` when cannot unwind.

runtime/near-vm-runner/src/wasmer_runner.rs
olonho: Not for me, seems it's really exported.

matklad (approved):

ailisp (approved):

runtime/near-vm-runner/src/wasmer1_runner.rs
```diff
@@ -94,19 +96,53 @@ impl IntoVMError for wasmer::RuntimeError {
...
-                    ))
-                } else {
-                    // A general error
-                    VMError::FunctionCallError(FunctionCallError::WasmerRuntimeError(error_msg))
-                }
-            }
+            Ok(e) => return (&e).into(),
+            _ => {}
         }
+        let trap_code = trap_code.unwrap_or_else(|| panic!("Unknown error: {}", error_msg));
```
bowenwang1996, line 102: Why does it make sense to panic here?

olonho: Because we cannot distinguish if this error is deterministic or not.

matklad (approved):

evgenykuzyakov:
Adding @volovyk-s to check on error reporting, since it affects near-api-js

nearmax (changes requested):

frol (changes requested):

runtime/runtime/src/actions.rs
olonho: The intention here is to have error separated between contract executor and the rest of logic. We can make it a enum, but the only purpose of this error is to show error to humans, that's why I made this change.

runtime/runtime/src/actions.rs
```diff
@@ -192,10 +192,22 @@ pub(crate) fn action_function_call(
         false,
     );
     let execution_succeeded = match err {
-        Some(VMError::FunctionCallError(err)) => {
-            result.result = Err(ActionErrorKind::FunctionCallError(err).into());
-            false
-        }
+        Some(VMError::FunctionCallError(err)) => match err {
```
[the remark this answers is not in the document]
olonho: That's intentional, and for that particular distinction we can introduce CompilationError and ExecutionError, if this difference is indeed important. Currently, the only use of the error is showing it to developers in logs.

core/primitives/src/errors.rs
```diff
@@ -384,8 +384,8 @@ pub enum ActionErrorKind {
         #[serde(with = "u128_dec_format")]
         minimum_stake: Balance,
     },
-    /// An error occurred during a `FunctionCall` Action.
-    FunctionCallError(FunctionCallError),
+    /// An error occurred during a `FunctionCall` Action, parameter is debug message.
+    FunctionCallError(String),
```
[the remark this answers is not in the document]
olonho: Do not think, that protocol changes in any way, as for protocol it's important presence of FunctionCallError  error, not exact type of it.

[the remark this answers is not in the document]
frol: I think for proof it is important to have an identical hash for the execution outcome, so it will require a protocol change unless I misunderstand something.

runtime/runtime/src/actions.rs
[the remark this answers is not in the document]
frol: > Currently, the only use of the error is showing it to developers in logs.

This is exactly what we want to improve by introducing sane granularity.

> That's intentional, and for that particular distinction we can introduce CompilationError and ExecutionError, if this difference is indeed important.

Yes, that would be ideal. I think end-users care about three types of errors: UnknownMethod, CompilationError, ExecutionError.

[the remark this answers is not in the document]
nearmax: Sounds good, please introduce these 3 levels of errors.

runtime/near-vm-runner/src/wasmer1_runner.rs
olonho: Done.

runtime/near-vm-runner/src/wasmer1_runner.rs
olonho: Done.

runtime/runtime/src/actions.rs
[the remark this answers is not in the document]
olonho: Cool, let's do that!

core/primitives/src/errors.rs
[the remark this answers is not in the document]
olonho: IIRC, when computing hash we explicitly ignore error messages, @matklad shown me the place where it's done.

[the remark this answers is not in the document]
matklad: >@matklad shown me the place where it's done.

Over here: 

https://github.com/near/nearcore/blob/b95eedf81894d87802285cbd331515ee69a0f51f/core/primitives/src/transaction.rs#L324

olonho:
PTAL, merged sugestions.

runtime/near-vm-errors/src/lib.rs
```diff
@@ -31,15 +29,11 @@ pub enum FunctionCallError {
     MethodResolveError(MethodResolveError),
     /// A trap happened during execution of a binary
     WasmTrap(WasmTrap),
-    WasmUnknownError,
+    WasmUnknownError(String),
```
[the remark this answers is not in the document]
olonho: Done

core/primitives/src/errors.rs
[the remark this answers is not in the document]
bowenwang1996: The variant of the error doesn't matter since it doesn't go into the merkle root

runtime/runtime/src/actions.rs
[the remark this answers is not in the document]
volovyk-s: +1 for structured errors, it's one of the pain points in `near-api-js` and all projects that are using it.
What do you think about adding the `BadArgument` error? It's a common mistake for developers.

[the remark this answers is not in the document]
nearmax: @volovyk-s  `BadArgument` is a contract-specific error (because each contract decides to interpret it differently). We cannot differentiate it on a blockchain-level.

runtime/near-vm-runner/src/wasmer1_runner.rs
nearmax: We should leave a comment that says, that if this happens that it is a bug in Wasmer, since their errors should have an exhaustive coverage, without "unknown errors".

runtime/runtime/src/actions.rs
```diff
@@ -192,10 +192,22 @@ pub(crate) fn action_function_call(
...
+            FunctionCallError::Nondeterministic(msg) => {
+                panic!("Contract runner returned non-deterministic error '{}', aborting", msg)
+            }
+            FunctionCallError::CompilationError(_)
+            | FunctionCallError::LinkError { msg: _ }
+            | FunctionCallError::MethodResolveError(_)
+            | FunctionCallError::WasmTrap(_)
+            // TODO: shall we abort on unknown errors also?
```
nearmax, line 214: What is the resolution with this TODO?

runtime/runtime/src/actions.rs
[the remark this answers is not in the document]
nearmax: @olonho It seems like we are still combining all three kinds of errors into `ActionErrorKind::FunctionCallError` with string. Did you forget to push the commit?

nearmax (changes requested):

core/primitives/src/errors.rs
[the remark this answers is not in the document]
frol: Great, we don't need a protocol upgrade!

What about database migration?

I also see various ExecutionOutcome types used in chain/network. (ExecutionOutcome uses TxExecutionError which uses ActionError which uses ActionErrorKind and all of that gets borsh-serialized and should be compatible with the existing running nodes, right?)

frol (changes requested):
I would suggest to:

* [ ] run nayduck
* [ ] check that you can boot a node from an old data folder (and query `tx` JSON RPC endpoint to trigger reading execution outcomes - let's assume we can tolerate the change in the schema, but it should not crash and should return sane results)
* [ ] confirm that this new node can still communicate with the old ones

runtime/runtime/src/actions.rs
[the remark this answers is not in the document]
matklad: > and we prefer to have error in RPC as structured as possible

I want to push back a little bit against "as structured as possible errors". This is not strictly related to the PR, just the  pattern I observer in Rust's error handling. 

The issue with _maximally_ stuctured errors is that they increase coupling, and sometimes it's better to stop at the goldilocks errors, which are structured just enough. 

In the context of this PR, if we make error maximally structured, than we need to expose wasmer 0.17 and wasmer 1.0 as is. And when we remove support for wasmer 0.17 or upgrade to (hypothetical) wasmer 2.0, we'd have to change the definition of the error type as well, and update the call-sites. This PR in some sense is the consequence of the fact that we did change error definition when upgrading to wasmer 1.0. This is a code smell, pointing out that the original  error definition probably wasn't encapsulated enough. 

But, clearly, exposing *just* the error message goes to far in the opposite direction, as it forces the clients to parse out the messages. 

There's a goldilocks error pattern -- `ErrorKind`. The best example of it is std::io::Error:

image

The `Error` type itself is an opaque struct, which can be converted to string via `Display`. There's a `.kind()` method, which exposes **non-exhaustive**, **field-less** enum. This structure allows the client to behave differently for different sorts of errors, but allows the implementor to modify the errors  without changes to the client. 

To re-iterate, this is definitely not something we should do in the scope of this PR, and it's not necessary something we want to do even eventually (serializability of errors add an extra design dimension here which might change the tradeoffs). But it's useful to keep this bit of info in mind -- just exposing errors as is is the path of least resistance, but often not something you want to have at the API _boundary_. 

References: 

* https://github.com/rust-lang-nursery/failure/blob/master/book/src/error-errorkind.md
* https://matklad.github.io/2020/10/15/study-of-std-io-error.html
* https://blog.burntsushi.net/rust-error-handling/

runtime/runtime/src/actions.rs
olonho: Done.

olonho:
@bowenwang1996 @frol could you please help with running real node with those changes?

core/primitives/src/errors.rs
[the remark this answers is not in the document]
olonho: Also do not think we shall store anything related to exact error kind in DB. If we do - guess it's a bug.

runtime/runtime/src/actions.rs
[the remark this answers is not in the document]
frol: @matklad thanks for the detailed overview. I believe after some discussions we already arrived at the middle ground of introducing exactly the necessary variants that the consumer cares about :+1:

core/primitives/src/errors.rs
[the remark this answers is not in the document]
frol: @olonho we definitely store the errors in the storage since we receive them from RPC: https://explorer.near.org/transactions/GcRiQb5PFwMAogtu8te6wDKMRg9pefbWV4nJx3iqijny

image

```bash
http post https://archival-rpc.mainnet.near.org/ jsonrpc=2.0 method=tx params:='["GcRiQb5PFwMAogtu8te6wDKMRg9pefbWV4nJx3iqijny", "tom.zest.near"]' id=dontcare
```

olonho:
Likely done all the above steps.
nayduck  run is complete.
Have the node up and running on top of the old database.

core/primitives/src/errors.rs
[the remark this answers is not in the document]
bowenwang1996: We do store them in database and if we've changed how the error serializes to borsh we should do a migration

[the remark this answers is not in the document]
olonho: Right, so could we do migration in the followup CL? And would be amazing, if we could cooperate on the migration effort.

runtime/near-vm-runner/src/wasmer1_runner.rs
olonho: Done

core/primitives/src/errors.rs
```diff
@@ -384,8 +393,8 @@ pub enum ActionErrorKind {
         #[serde(with = "u128_dec_format")]
         minimum_stake: Balance,
     },
-    /// An error occurred during a `FunctionCall` Action.
-    FunctionCallError(FunctionCallError),
+    /// An error occurred during a `FunctionCall` Action, parameter is debug message.
+    FunctionCallError(ContractCallError),
```
olonho, line 397: The place where serialization changes.

bowenwang1996:
@olonho @frol instead of doing database migration, I think it may be easier to do the following:
* We maintain the old `FunctionCallError` and implement `BorshDeserialize` for `ContractCallError` which first tries to deserialize into `ContractCallError` and if it fails, it tries to deserialize into `FunctionCallError`
* We also implement `Into<ContractCallError>` for `FunctionCallError`

olonho:
Reworked serialization format, so that we don't need to upgrade database anymore. Sketched possible universal mechanism of handling such incompatibilities in the future.

runtime/near-vm-errors/src/lib.rs
```diff
@@ -67,12 +85,12 @@ pub enum WasmTrap {
...
     /// Misaligned atomic access trap.
     MisalignedAtomicAccess,
-    /// Breakpoint trap.
-    BreakpointTrap,
     /// Stack overflow.
     StackOverflow,
     /// Generic trap.
     GenericTrap,
+    /// Indirect call to null.
+    IndirectCallToNull,
```
matklad, line 93: Let's shift this error to where (impossible)  `BreakpointTrap` was, to avoid changing (serialized) error codes for overflow and generic trap.

olonho: Done.

core/primitives/src/errors.rs
```diff
@@ -332,6 +332,48 @@ pub struct ActionError {
...
+#[derive(
+    BorshSerialize, BorshDeserialize, Debug, Clone, PartialEq, Eq, Deserialize, Serialize, RpcError,
+)]
+pub enum ContractCallError {
+    MethodResolveError(MethodResolveError),
+    CompilationError(CompilationError),
+    ExecutionError(String),
+}
+
+impl Into<FunctionCallErrorSer> for ContractCallError {
```
bowenwang1996, line 344: @matklad Is my understanding correct that implementing `Into` is not rust-idiomatic? I think it is always preferred to implement `From`

runtime/near-vm-errors/src/lib.rs
```diff
@@ -17,10 +17,33 @@ pub enum VMError {
     CacheError(CacheError),
 }
 
-#[derive(
-    Debug, Clone, PartialEq, Eq, BorshDeserialize, BorshSerialize, Deserialize, Serialize, RpcError,
-)]
+// TODO: remove serialization derives, once fix compilation caching.
+#[derive(Debug, Clone, PartialEq, Eq, BorshDeserialize, BorshSerialize, Serialize, Deserialize)]
 pub enum FunctionCallError {
```
bowenwang1996, line 22: is this enum still needed given the presence of `FunctionCallErrorSer` and `ContractCallError`

core/primitives/src/errors.rs
matklad: Oh, good catch @bowenwang1996! Yeah, this should go the other way arond, `From<ContractCallError> for FunctionCallErrorSer`.

runtime/near-vm-errors/src/lib.rs
olonho: Yes, it's used inside the contract runtime and have precise execution problem description. We don't want it to leak beyond contract runtime.

core/primitives/src/errors.rs
olonho: Replaced.

frol:
> We maintain the old FunctionCallError and implement BorshDeserialize for ContractCallError which first tries to deserialize into ContractCallError and if it fails, it tries to deserialize into FunctionCallError

@bowenwang1996 I just want to understand the reasoning here. Do we need to shortcut here and get this PR merged asap or we just don't want migrations if we can work around them keeping the old reach data in the database in case we need to update it once again?

core/primitives/src/errors.rs
```diff
@@ -332,6 +332,48 @@ pub struct ActionError {
...
 }
 
+#[derive(
+    BorshSerialize, BorshDeserialize, Debug, Clone, PartialEq, Eq, Deserialize, Serialize, RpcError,
+)]
+pub enum ContractCallError {
+    MethodResolveError(MethodResolveError),
+    CompilationError(CompilationError),
+    ExecutionError(String),
+}
```
ailisp, line 342: This is really good structure, self documented of what could be wrong in contract call :+!

runtime/near-vm-errors/src/lib.rs
```diff
@@ -17,10 +17,33 @@ pub enum VMError {
...
+    },
+    HostError(HostError),
+    EvmError(EvmError),
+    /// Non-deterministic error.
+    Nondeterministic(String),
+}
+
+/// Serializable version of `FunctionCallError`. Must never reorder/remove elements, can only
+/// add new variants at the end (but do that very carefully). This type must be never used
+/// directly, and must be converted to `ContractCallError` instead using `into()` converter.
```
ailisp, line 44: One missing in doc: since it's never used directly, when it should be used?

runtime/near-vm-errors/src/lib.rs
```diff
@@ -67,8 +86,8 @@ pub enum WasmTrap {
     IllegalArithmetic,
     /// Misaligned atomic access trap.
     MisalignedAtomicAccess,
-    /// Breakpoint trap.
-    BreakpointTrap,
+    /// Indirect call to null.
+    IndirectCallToNull,
```
ailisp, line 90: Add a test case to trigger this error?

runtime/near-vm-errors/src/lib.rs
```diff
@@ -363,12 +382,13 @@ impl fmt::Display for FunctionCallError {
             FunctionCallError::HostError(e) => e.fmt(f),
             FunctionCallError::LinkError { msg } => write!(f, "{}", msg),
             FunctionCallError::WasmTrap(trap) => write!(f, "WebAssembly trap: {}", trap),
-            FunctionCallError::WasmUnknownError => {
-                write!(f, "Unknown error during Wasm contract execution")
+            FunctionCallError::WasmUnknownError { debug_message } => {
+                write!(f, "Unknown error during Wasm contract execution: {}", debug_message)
```
ailisp, line 386: Still feel we should rename this to WasmInternalError, it's a paradox that it's unknown then error detail (debug_message) is given

runtime/near-vm-runner/src/tests/error_cases.rs
```diff
@@ -130,25 +131,22 @@ fn trap_contract() -> Vec<u8> {
...
-                Some(vm_outcome_with_gas(47105334)),
-                Some(VMError::FunctionCallError(FunctionCallError::WasmUnknownError))
-            )
-        ),
-        VMKind::Wasmer1 => assert_eq!(
+    with_vm_variants(|vm_kind: VMKind| {
+        match vm_kind {
+            VMKind::Wasmer0 | VMKind::Wasmer1 => {}
+            // All contracts leading to hardware traps can not run concurrently on Wasmtime and Wasmer,
+            // Restore, once get rid of Wasmer 0.x.
```
ailisp, line 138: Can we restore even after get rid of wasmer 0.x (does wasmer 1 and wasmtime conflict on hardware traps)

runtime/near-vm-runner/src/tests/error_cases.rs
```diff
@@ -130,25 +131,22 @@ fn trap_contract() -> Vec<u8> {
...
+            VMKind::Wasmtime => return,
+        }
+        assert_eq!(
             make_simple_contract_call_vm(&trap_contract(), "hello", vm_kind),
             (
                 Some(vm_outcome_with_gas(47105334)),
-                Some(VMError::FunctionCallError(FunctionCallError::Wasmer1Trap(
-                    "unreachable".to_string()
+                Some(VMError::FunctionCallError(FunctionCallError::WasmTrap(
+                    WasmTrap::Unreachable
```
ailisp, line 146: This test returns WasmUnknownError for wasmer 0.x, did you change somewhere in code to make it return unreachable too?

runtime/near-vm-runner/src/tests/error_cases.rs
```diff
@@ -284,27 +318,21 @@ fn stack_overflow() -> Vec<u8> {
 
 #[test]
 fn test_stack_overflow() {
+    const COST: u64 = 63226248177;
```
ailisp, line 321: Why this one have a separated const COST, while other tests don't

runtime/near-vm-errors/src/lib.rs
olonho: OK, described that it is used in serialization.

runtime/near-vm-errors/src/lib.rs
olonho: Makes sense, will do.

runtime/near-vm-errors/src/lib.rs
olonho: Any renames like that would lead to serialization issues again. Think is general we shall go in direction of separatimg runtime and serialized data structures.

core/primitives/src/errors.rs
```diff
@@ -332,6 +332,48 @@ pub struct ActionError {
     pub kind: ActionErrorKind,
 }
 
+#[derive(
+    BorshSerialize, BorshDeserialize, Debug, Clone, PartialEq, Eq, Deserialize, Serialize, RpcError,
+)]
```
matklad, line 337: Should we

```suggestion
#[derive(Debug, Clone, PartialEq, Eq)]
```
?

My understanding is that, ideally we indeed would prefer to use this serialization format. But as we are not using this serialization right now, we might avoid derives? Seems easy enough to add them when we need them.

runtime/near-vm-errors/src/lib.rs
```diff
@@ -34,13 +57,9 @@ pub enum FunctionCallError {
     WasmUnknownError,
     HostError(HostError),
     EvmError(EvmError),
-    /// An error message when wasmer 1.0 returns a wasmer::RuntimeError
-    WasmerRuntimeError(String),
-    /// A trap in Wasmer 1.0, not same as WasmTrap above, String is a machine readable form like "stk_ovf"
-    /// String is used instead of wasmer internal enum is because of BorshSerializable.
-    /// It can be convert back by wasmer_vm::TrapCode::from_str
-    Wasmer1Trap(String),
+    ExecutionError(String),
```
matklad, line 60: I'd do

```suggestion
    ExecutionError { msg: String },
```

just in case. This thing is serializable to JSON, and, tuple variants are not extensible:


```rust
#[derive(serde::Serialize, serde::Deserialize)]
enum E1 {
    V(String),
}

#[derive(serde::Serialize, serde::Deserialize)]
enum E2 {
    V { msg: String },
}

#[derive(serde::Serialize, serde::Deserialize)]
enum E3 {
    V { msg: String, code: u32 },
}


fn main() {
    println!("{}", serde_json::to_string(&E1::V("hello".into())).unwrap());
    println!("{}", serde_json::to_string(&E2::V { msg: "hello".into() }).unwrap());
    println!("{}", serde_json::to_string(&E3::V { msg: "hello".into(), code: 92 }).unwrap());
}

// OUTPUT
{"V":"hello"}
{"V":{"msg":"hello"}}
{"V":{"msg":"hello","code":92}}
```

Migrating `E2 -> E3` is easier that `E1 -> E3`.

Now, because this is also borsh serializable, this might not be super-relevant, as adding a field to a borsh struct is not compatible, but it's better to be on the safe side here (and consistent with `LinkError`).

runtime/near-vm-runner/src/tests/rs_contract.rs
```diff
@@ -285,12 +285,10 @@ pub fn test_out_of_memory() {
...
             match vm_kind {
-                VMKind::Wasmer0 =>
-                    Some(VMError::FunctionCallError(FunctionCallError::WasmUnknownError)),
-                VMKind::Wasmer1 => Some(VMError::FunctionCallError(
-                    FunctionCallError::Wasmer1Trap("unreachable".to_string())
+                VMKind::Wasmer0 | VMKind::Wasmer1 => Some(VMError::FunctionCallError(
+                    FunctionCallError::WasmTrap(WasmTrap::Unreachable)
                 )),
-                _ => unreachable!(),
+                VMKind::Wasmtime => unreachable!(),
```
matklad, line 291: :+1:

runtime/near-vm-runner/src/tests/error_cases.rs
olonho: Need to recheck, updated comment.

matklad (approved):

runtime/near-vm-errors/src/lib.rs
```diff
@@ -17,10 +17,33 @@ pub enum VMError {
     CacheError(CacheError),
 }
 
-#[derive(
-    Debug, Clone, PartialEq, Eq, BorshDeserialize, BorshSerialize, Deserialize, Serialize, RpcError,
-)]
+// TODO: remove serialization derives, once fix compilation caching.
```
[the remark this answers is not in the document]
olonho: Not addressed here, added TODO(4217).

runtime/near-vm-runner/src/tests/error_cases.rs
olonho: Fixed.

core/primitives/src/errors.rs
olonho: Exactly the case, I'd love that to be serialization format, but it's not.

runtime/near-vm-errors/src/lib.rs
olonho: OK, done.

bowenwang1996 (approved):

frol (approved):

olonho:
Addresses Max's request.

PR #4183, opened 2021-04-01
refactor: reduce the number of deps for public libraries

matklad:
primitives-core is an API crate, so all of its deps become deps of
consumers of the library. This makes deps significantly more costly.

For derive-more in particular, it seems like it is not used that
heavily.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3440,7 +3440,6 @@ dependencies = [
  "base64 0.11.0",
  "borsh",
  "bs58",
- "derive_more",
  "hex",
  "lazy_static",
  "num-rational",
--- a/core/primitives-core/Cargo.toml
+++ b/core/primitives-core/Cargo.toml
@@ -15,7 +15,6 @@ base64 = "0.11"
 borsh = "0.8.1"
 bs58 = "0.4"
 hex = "0.4"
-derive_more = "0.99.3"
 num-rational = { version = "0.3.1", features = ["serde"]}
 serde = { version = "1", features = ["derive"] }
 serde_json = "1"
--- a/core/primitives-core/src/hash.rs
+++ b/core/primitives-core/src/hash.rs
@@ -7,12 +7,16 @@ use serde::{Deserialize, Deserializer, Serialize, Serializer};
 use crate::logging::pretty_hash;
 use crate::serialize::{from_base, to_base, BaseDecode};
 
-#[derive(Copy, Clone, PartialOrd, PartialEq, Eq, Ord, derive_more::AsRef)]
-#[as_ref(forward)]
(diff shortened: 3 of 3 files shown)


matklad:
@frol this is more of a suggestion PR: I don't have enough experience to understand whether we want to do this or not. I think you are the better person to make a judgement call here :)

bowenwang1996 (approved):
I think this PR makes sense

frol (changes requested):
I am only fine with this change if we measure the improvement and enforce it. My measurement is that with `derive_more` I need to read less code, and I don't believe at this point that you can notice any improvement with the proposed change. As to the enforcement, the next time someone goes to this file, they potentially will introduce `derive_more`, and we will be back to square one.

I think the problem you raised needs to be addressed systematically, and make sure that our crates granularity is correct before we try to save some bits.

matklad:
> our crates granularity is correct before we try to save some bits.

That's a great point, agree that this is premature optimization at this point, closing!

> measure the improvement

I've tried `cd core/primitives-core && t cargo build --lib -j 1`, the results look like this:

```rust
# With derive more
real 32.10s
cpu  33.02s (30.78s user + 2.24s sys)
rss  452.62mb

# Without derive more
real 28.18s
cpu  28.93s (26.89s user + 2.04s sys)
rss  392.70mb
```

So that's 4 seconds absolute / 14 % relative difference in *the worst-case* (from scratch debug build). `derive-more` seems to be among the bigger ones in this corener of the crate graps:

image

(`env RUSTC_BOOTSTRAP=1 cargo build --lib -j 1 -Ztimings`)

> and enforce it

The enforcement that works well for rust-analayzer is three-fold:

* dependency stance [is documented](https://github.com/rust-analyzer/rust-analyzer/blob/master/docs/dev/style.md#cratesio-dependencies)
* by the time folks get review permissions, they generally know that we don't do deps
* once in a while I read Cargo.lock to see if I need to raise some issues for our dependencies :)

matklad:
Forgot to mention one more thing: 

>My measurement is that with derive_more I need to read less code

I personally feel the opposite: with derive_more, I need to read *more* code. It's not immediately obvious if it implements `AsRef<[u8]>` or `AsRef<[u8; 32]>`, and if I "goto definition", than I land into proc macro, which is not too transparent.

frol:
Great points! I would love to see those measures applied. Currently, we only use `cargo-deny` to protect us from too many duplicate dependencies (we [cannot avoid them completely since](https://github.com/near/nearcore/blob/master/deny.toml) we pull out quite massive third-party dependencies like wasmer, wasmtime, evm, and various crypto implementations)

PR #4185, opened 2021-04-01
Profile by default

matklad:
Having costs breakdown (how much gas was spend for each operation type)
is helpful for both increasing reliability (as we can eyeball if the
costs "make sense") and developer friendliness (as devs can understand
where all the gas is going).

Currently this is hidden behind a feature flag, but I would love to
argue that we should just always enable it.

Performance cost seems to be negligible: we heap-allocate an array once
per apply state, and actual cost increasse are just counter increments,
they should be dwarfed by the host functions themselves.

Correctness-wise, the code is semi-sensitive. It is executed by runtime,
but doesn't affect the results. So it's enough if it doesn't panic.

# Test Plan

All of existing tests now exercise costs councting code. An additional unit-test is added to smoke-test that costs profiling can't panic due to overflow.

### DIFF
--- a/core/primitives-core/src/profile.rs
+++ b/core/primitives-core/src/profile.rs
@@ -71,7 +71,7 @@ impl ProfileData {
     fn add_val(&self, index: usize, value: u64) {
         self.with_slot(index, |slot| {
             let old = slot.get();
-            slot.set(old + value);
+            slot.set(old.saturating_add(value));
         })
     }
 
@@ -220,5 +220,15 @@ mod test {
     fn test_profile_data_debug_no_data() {
         let profile_data = ProfileData::new_enabled();
         println!("{:#?}", &profile_data);
+    }
+
+    #[test]
+    fn test_no_panic_on_overflow() {
+        let profile_data = ProfileData::new_enabled();
+        profile_data.add_action_cost(ActionCosts::function_call, u64::MAX);
+        profile_data.add_action_cost(ActionCosts::function_call, u64::MAX);
+
+        let res = profile_data.get_action_cost(ActionCosts::function_call as usize);
+        assert_eq!(res, u64::MAX);
     }
 }
--- a/core/primitives-core/src/profile.rs
+++ b/core/primitives-core/src/profile.rs
@@ -1,33 +1,21 @@
+use std::rc::Rc;
 use std::{cell::Cell, fmt};
 
 use crate::config::{ActionCosts, ExtCosts};
 use crate::types::Gas;
 
 /// Data for total gas burnt (index 0), and then each external cost and action
-#[allow(dead_code)]
 type DataArray = [Cell<u64>; ProfileData::LEN];
 
 /// Profile of gas consumption.
 #[derive(Clone)]
 pub struct ProfileData {
-    repr: Repr,
+    data: Rc<DataArray>,
 }
-
-#[derive(Clone)]
-enum Repr {
-    #[cfg(feature = "costs_counting")]
-    Enabled {
-        data: std::rc::Rc<DataArray>,
-    },
-    Disabled,
-}
-
-#[cfg(not(feature = "costs_counting"))]
-const _ASSERT_NO_OP_IF_COUNTING_DISABLED: [(); 0] = [(); std::mem::size_of::<ProfileData>()];
 
 impl Default for ProfileData {
     fn default() -> ProfileData {
-        ProfileData::new_disabled()
+        ProfileData::new()
     }
 }
 
@@ -37,26 +25,12 @@ impl ProfileData {
     const LEN: usize = 1 + ActionCosts::count() + ExtCosts::count();
 
     #[inline]
-    #[cfg(feature = "costs_counting")]
-    pub fn new_enabled() -> Self {
+    pub fn new() -> Self {
         // We must manually promote to a constant for array literal to work.
         const ZERO: Cell<u64> = Cell::new(0);
 
-        let data = std::rc::Rc::new([ZERO; ProfileData::LEN]);
-        let repr = Repr::Enabled { data };
-        ProfileData { repr }
-    }
-
-    #[inline]
-    #[cfg(not(feature = "costs_counting"))]
-    pub fn new_enabled() -> Self {
-        Self::new_disabled()
-    }
-
-    #[inline]
-    pub fn new_disabled() -> Self {
-        let repr = Repr::Disabled;
-        ProfileData { repr }
+        let data = Rc::new([ZERO; ProfileData::LEN]);
+        ProfileData { data }
     }
 
     #[inline]
@@ -69,24 +43,14 @@ impl ProfileData {
     }
     #[inline]
     fn add_val(&self, index: usize, value: u64) {
-        self.with_slot(index, |slot| {
-            let old = slot.get();
-            slot.set(old.saturating_add(value));
-        })
+        let slot = &self.data[index];
+        let old = slot.get();
+        slot.set(old.saturating_add(value));
     }
 
-    #[inline]
-    fn with_slot(&self, index: usize, f: impl FnOnce(&Cell<u64>)) {
-        match &self.repr {
-            #[cfg(feature = "costs_counting")]
-            Repr::Enabled { data } => f(&data[index]),
-            Repr::Disabled => drop((index, f)),
-        }
-    }
     fn read(&self, index: usize) -> u64 {
-        let mut res = 0;
-        self.with_slot(index, |slot| res = slot.get());
-        res
+        let slot = &self.data[index];
+        slot.get()
     }
 
     pub fn all_gas(&self) -> Gas {
@@ -121,18 +85,12 @@ impl ProfileData {
     }
 
     pub fn set_burnt_gas(&self, burnt_gas: u64) {
-        self.with_slot(0, |slot| slot.set(burnt_gas));
+        let slot = &self.data[0];
+        slot.set(burnt_gas)
     }
 }
 
 impl fmt::Debug for ProfileData {
-    #[cfg(not(feature = "costs_counting"))]
-    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        writeln!(f, "ERROR: cost_counting feature is not enabled in near-primitives-core, cannot print profile data")?;
-        Ok(())
-    }
-
-    #[cfg(feature = "costs_counting")]
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         use num_rational::Ratio;
         let all_gas = self.all_gas();
@@ -201,30 +159,27 @@ mod test {
     use super::*;
     #[test]
     fn test_profile_all_gas() {
-        let profile_data = ProfileData::new_enabled();
+        let profile_data = ProfileData::new();
         profile_data.set_burnt_gas(42);
-        #[cfg(not(feature = "costs_counting"))]
-        assert_eq!(profile_data.all_gas(), 0);
-        #[cfg(feature = "costs_counting")]
         assert_eq!(profile_data.all_gas(), 42);
     }
 
     #[test]
     fn test_profile_data_debug() {
-        let profile_data = ProfileData::new_enabled();
+        let profile_data = ProfileData::new();
         profile_data.set_burnt_gas(42);
         println!("{:#?}", &profile_data);
     }
 
     #[test]
     fn test_profile_data_debug_no_data() {
-        let profile_data = ProfileData::new_enabled();
+        let profile_data = ProfileData::new();
         println!("{:#?}", &profile_data);
     }
 
     #[test]
     fn test_no_panic_on_overflow() {
-        let profile_data = ProfileData::new_enabled();
+        let profile_data = ProfileData::new();
         profile_data.add_action_cost(ActionCosts::function_call, u64::MAX);
         profile_data.add_action_cost(ActionCosts::function_call, u64::MAX);
 
--- a/core/primitives-core/src/profile.rs
+++ b/core/primitives-core/src/profile.rs
@@ -120,10 +120,6 @@ impl ProfileData {
         self.all_gas() - self.host_gas() - self.action_gas()
     }
 
-    pub fn unaccounted_gas(&self) -> u64 {
-        self.all_gas() - self.wasm_gas() - self.action_gas() - self.host_gas()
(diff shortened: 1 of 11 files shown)


willemneal:
Yes! I also suggest that this becomes available via RPC.

matklad:
> I also suggest that this becomes available via RPC.

That's precisely my secret agenda here ;)

ailisp (approved):

bowenwang1996 (commented):
I am still not sure that it is a good idea to enable this for every node, even if we want rpc to use this data. Ideally in runtime there should not be extraneous code that does not do exactly what needs to be done. If we want to surface information to rpc, we can maybe consider a runtime flag that can be controlled through config.json.

matklad:
> in runtime there should not be extraneous code that does not do exactly what needs to be done

Hm, yeah, strongly agree here! I'll come back to this next week, this indeed needs a more thorough approach.

core/primitives-core/src/profile.rs
```diff
@@ -1,70 +1,36 @@
...
-        Self { data: Rc::clone(&self.data) }
-    }
-}
+/// Data for total gas burnt (index 0), and then each external cost and action
+type DataArray = [Cell<u64>; ProfileData::LEN];
 
 /// Profile of gas consumption.
 #[derive(Clone)]
 pub struct ProfileData {
-    repr: Repr,
```
olonho, line 23: Why this change?

core/primitives-core/src/profile.rs
```diff
@@ -77,21 +43,14 @@ impl ProfileData {
     }
     #[inline]
     fn add_val(&self, index: usize, value: u64) {
-        self.with_data(|data| data[index] += value)
+        let slot = &self.data[index];
+        let old = slot.get();
+        slot.set(old.saturating_add(value));
```
olonho, line 48: Practically speaking u64 saturation is not very likely, but as you like.

olonho (approved):
Looks good!

core/primitives-core/src/profile.rs
matklad: This is (was) the main idea of the PR: if we enable profiling by default, we don't need a `Repr`  that can be either enabled or disabled. But, as bowenwang1996 noticed, we probably don't want to execute *anything* not strictly necessary in the runtime, so I'll roll back this particular change, to allow disabling the extra code.

nearmax:
> I am still not sure that it is a good idea to enable this for every node, even if we want rpc to use this data. Ideally in runtime there should not be extraneous code that does not do exactly what needs to be done. If we want to surface information to rpc, we can maybe consider a runtime flag that can be controlled through config.json.

@bowenwang1996  @matklad  I do not agree with this reasoning. Collecting cost breakdown comes at no cost, so it is not a performance issue. It is also not clear what is meant by "not be extraneous code that does not do exactly what needs to be done". Runtime computing and exposing things beyond protocol needs already. For example extensive compilation&execution error messaging is computed and exposed by the runtime and is eventually exposed in RPC and explorer, but it does not have effect on the protocol. From this perspective extra information about compilation&execution failures is no different from extra information about gas cost breakdown.

bowenwang1996:
> For example extensive compilation&execution error messaging is computed and exposed by the runtime and is eventually exposed in RPC and explorer, but it does not have effect on the protocol. From this perspective extra information about compilation&execution failures is no different from extra information about gas cost breakdown.

I agree with your reasoning but not your conclusion. This information serve no purpose from the point of view of a validator and it probably should not be part of the binary that they run. We can have a different type of node that collects all the additional information we need for purposes other than validating the chain.

nearmax:
>This information serve no purpose from the point of view of a validator and it probably should not be part of the binary that they run.

So are you suggesting to hide compilation&execution errors with flag too? What is the advantage for hiding it from validators if they are not gaining any CPU utilization and since RPC is not their primary interface anyway? How are we making life of validator better by disabling it?

bowenwang1996:
> > This information serve no purpose from the point of view of a validator and it probably should not be part of the binary that they run.
> 
> So are you suggesting to hide compilation&execution errors with flag too? What is the advantage for hiding it from validators if they are not gaining any CPU utilization and since RPC is not their primary interface anyway? In general, I am confused why we are saying we are doing something for validators' sake when it has no effect on validators. I think we might be over-optimizing here.

I think there are two factors at play here: performance and security. Some code that is unnecessary for validator node may not incur much performance overhead and may increase the chance of bugs or security vulnerabilities. I am not saying that it is the case here, but that is just a general observation. You could say that we are over optimizing here, but then what is the boundary at which we can say "this code doesn't make sense to be put in runtime"?

nearmax:
I suggest the following criteria for rejecting code from being added into transaction runtime:
* It adds significant performance overhead;
* It can be easily implemented outside the runtime codebase;
* It makes runtime less maintainable and more complex.

If either of these criteria are satisfied then we reject this code from the runtime. AFAIU gas breakdown counter does not satisfy any of the criteria:
* It does not add any noticeable performance overhead as reported by @olonho ;
* We cannot easily implement it outside the runtime codebase, since we need to have gas breakdown counters to be called right next to where the fees are applied, which are all over the runtime codebase. Having tools that re-implement runtime but with the counters will lead to large code duplication;
* We need to have counters in the runtime code anyway, because this is how param estimator works, so it is not the question of having it in runtime code, but whether we disable it through a flag or expose it via RPC for explorer along with compilation/execution error. Guarding code with flag actually makes it more complex and harder to reason about, so if we are optimizing for runtime simplicity then we should remove the flag. Piping this data via ExecutionOutcome shouldn't be a problem for any other code outside the runtime since it does not care what is inside the ExecutionOutcome (compilation/execution errors and gas breakdown are stripped away when computing merkle hash).

bowenwang1996:
> It adds significant performance overhead;

I think this may be a bit hard to determine. Sometimes seemingly innocent code can lead to performance overhead. For example, if you compute some sha256 per receipt, then it may appear that there is not much overhead, but when there are thousands of receipts to be processed, it adds up.

matklad:
I don't think I can make a judgement call here, so let me just add a couple of more observations

I don't think "easy to main & fast & essential" quite captures it -- I've written a bunch of easy to maintain code that turned out to be accidentally quadratic in pathalogical cases. It feels like we need something stronger here, like: "relatively obvious to not contain unknown vulns / dos opportunities" (good luck checking that :) ).

In this PR, this additional facet is demonstrated by overflow handling. If I am writing code for non-validators, I just don't care about it. If I am writing for validators, I am thinking "wait, what if someone misuses the profiler and calls profile *before* deducting gas? In that case, if I just do +=, someone will craft malicious input to make profiling infra deterministically panic. So let's use saturating ops here to be sure".

On the other hand, it's also not true that gating the code solves this problem completely. Making code conditional significantly increases complexity. We may end up in a situation where we *think* the code is not used by the runtime, so we are not scrutinizing it closely, but, due to some misconfiguration, actually end up using it. 


Finally (I know, I promised not to make judgment calls), I think if we allow non-trivial non-essential code in principle, then this PR probably passes the mark: I am not sure which metric of reliability do we want, but this particular code seems to be almost as simple as it can get (to clarify, the claim is not that it is bug free, it is that any other non-trivial code would have not fewer bugs).

bowenwang1996:
> In this PR, this additional facet is demonstrated by overflow handling. If I am writing code for non-validators, I just don't care about it. If I am writing for validators, I am thinking "wait, what if someone misuses the profiler and calls profile _before_ deducting gas? In that case, if I just do +=, someone will craft malicious input to make profiling infra deterministically panic. So let's use saturating ops here to be sure".

I think this is a good point. @matklad it seems that you are saying that we also put more burden on our shoulders because we need to be thinking very carefully about even seemingly simple code.

nearmax:
@bowenwang1996 

> I think this may be a bit hard to determine. Sometimes seemingly innocent code can lead to performance overhead. For example, if you compute some sha256 per receipt, then it may appear that there is not much overhead, but when there are thousands of receipts to be processed, it adds up.

In this case it is quite obvious that this does not add performance overhead for several reasons:
* @olonho provided to me an argument almost on the level of the CPU instructions on why this is negligible. Given that it is super clear that the overhead is just following a couple of pointers and incrementing an integer value, I don't see how this is even an argument;
* It is a common practice to have counters exactly like this in the production code in large companies enabled by default and people don't think twice before adding them unless it is some super low-level code, which is far from being the case here. Even more, people have much more complex counters, e.g. those that collect complex data points into aggregated distributions and rarely think of them as adding performance overhead. See [Streamz](https://news.ycombinator.com/item?id=24303422), Borgmon, Monarch. You can also talk to @chefsale about it, or I can give you more details in non-written form.

I think saying that we don't know how much overhead global counter like this adds is very nitpicky. I am personally puzzled with how we can look at the code that does a couple of pointer traversals and one integer increment and say that we don't know its performance overhead.

@matklad 

> In this PR, this additional facet is demonstrated by overflow handling. If I am writing code for non-validators, I just don't care about it. If I am writing for validators, I am thinking "wait, what if someone misuses the profiler and calls profile before deducting gas? In that case, if I just do +=, someone will craft malicious input to make profiling infra deterministically panic. So let's use saturating ops here to be sure".

We should not forget that we are talking about incrementing a single integer number from the array, and not some complex monitoring code. It does not sound too difficult to have an overflow check or saturating addition for that single integer in one place. Besides, it is impossible to saturate that number in practice.

> wait, what if someone misuses the profiler and calls profile before deducting gas?

Contract developers do not control whether profile counter is incremented before or after deducing gas. Also, it is currently controlled in one place so there is minimal space for mistake. We might be imagining hypothetical code that is more complex than the code at hand, I suggest we keep the discussion focused on this specific code and not hypothetical one.

> On the other hand, it's also not true that gating the code solves this problem completely. Making code conditional significantly increases complexity. We may end up in a situation where we think the code is not used by the runtime, so we are not scrutinizing it closely, but, due to some misconfiguration, actually end up using it.

This sounds like a strong argument for not making this code conditional, and I agree with it.

> Finally (I know, I promised not to make judgment calls), I think if we allow non-trivial non-essential code in principle, then this PR probably passes the mark: I am not sure which metric of reliability do we want, but this particular code seems to be almost as simple as it can get (to clarify, the claim is not that it is bug free, it is that any other non-trivial code would have at least as many bugs).

I agree, it is hard to imagine more trivial code.

> I think this is a good point. @matklad it seems that you are saying that we also put more burden on our shoulders because we need to be thinking very carefully about even seemingly simple code.

We need to think about it for all runtime code anyway. Runtime has a lot overflow checks and other safety measures, because that's how runtime code is written. It should not be an argument on its own against adding new code into runtime. Most of the arguments so far were not against this specific code, but either against adding code into runtime in general or about some hypothetical code.

matklad:
>Contract developers do not control whether profile counter is incremented before or after deducing gas. Also, it is currently controlled in one place so there is minimal space for mistake. 

I think this specific point is incorrect. Looking at the call-sites, there actually are a bunch of places where we update profile before we deduct gas, here's one:

https://github.com/near/nearcore/blob/fd058703d3415d7b3c25d2cbe7f98480b5112cea/runtime/near-vm-logic/src/gas_counter.rs#L186-L188

So it seems like that a malicious actor can craft an action that would be so costly that it'd overeflow u64, and which would cause profile counter to overflow *before* we deduct gas and return an error, and it is really important (and not just defense in depth) to do checked arithmetics while profiling.

bowenwang1996:
> In this case it is quite obvious that this does not add performance overhead for several reasons:
> 
> * @olonho provided to me an argument almost on the level of the CPU instructions on why this is negligible. Given that it is super clear that the overhead is just following a couple of pointers and incrementing an integer value, I don't see how this is even an argument;
> * It is a common practice to have counters exactly like this in the production code in large companies enabled by default and people don't think twice before adding them unless it is some super low-level code, which is far from being the case here. Even more, people have much more complex counters, e.g. those that collect complex data points into aggregated distributions and rarely think of them as adding performance overhead. See [Streamz](https://news.ycombinator.com/item?id=24303422), Borgmon, Monarch. You can also talk to @chefsale about it, or I can give you more details in non-written form.
> 
> I think saying that we don't know how much overhead global counter like this adds is very nitpicky. I am personally puzzled with how we can look at the code that does a couple of pointer traversals and one integer increment and say that we don't know its performance overhead.

To be clear, I am not talking about this change specifically. I agree that this change does not impose much overhead. I was more talking about the general case -- what is the general approach we take to determine whether a change like this is reasonable?

nearmax:
@matklad 

> So it seems like that a malicious actor can craft an action that would be so costly that it'd overeflow u64, and which would cause profile counter to overflow before we deduct gas and return an error

Actions have fixed cost, it is not possible for an actor to define an action with a custom cost that would cause u64 overflow.    Number of actions are also limited by physical means like number of blocks in Wasm bytecode. I doubt we can even have u32 overflow here (not that I am advocating to use u32 :) ).

@bowenwang1996 

> To be clear, I am not talking about this change specifically. I agree that this change does not impose much overhead. I was more talking about the general case -- what is the general approach we take to determine whether a change like this is reasonable?

I don't think we can have a general rule, we might need to do case by case decision, unfortunately.

matklad:
Meta note: @bowenwang1996 @nearmax sorry for messing up the process on this one! I feel this could have been handled better by:

* noticing earlier that this feature actually needs a design discussion and facilitating it
* as implementer, being proactive about picking the right decision, rather than waiting for design decision to happen

With this in mind, lets continue discussion in this PR. 

I feel moderately strongly that we should proceed with it as is. I understand the concern that it's better to have as few things in runtime as possible, but, at the same time:

* the present code is the simplest non-trivial extra thing we can add. We do need to review it carefully and code it defensively, but its not an insurmountable obstacle. 
* the benefits here are substantiation -- the code improves observability of the running system by a lot. Observabiity I think is super important for long-runinng distributed systems.

core/primitives-core/src/profile.rs
```diff
@@ -120,23 +84,13 @@ impl ProfileData {
         self.all_gas() - self.host_gas() - self.action_gas()
     }
 
-    pub fn unaccounted_gas(&self) -> u64 {
```
willemneal, line 123: Is this removed since it is calling all public methods?

matklad: No, this is just a misleading method which always returns zero due to the way `wasm_gas` is defined.

willemneal: Ah I see.

willemneal (approved):

PR #4186, opened 2021-04-01
refactor: simplify protocol feature checking API

### LINKED ISSUE #4171: Improve protocol version check
Currently new protocol features are added in `ProtocolFeature` and there is a separate hashmap `PROTOCOL_FEATURES_TO_VERSION_MAPPING` that maps features to their corresponding protocol versions. @birchmd noted that this way of checking protocol versions for a given feature incurs unnecessary overhead and we should find ways to avoid this. @matklad has some suggestions [here](https://github.com/near/nearcore/pull/4115#discussion_r602217817), which seems quite reasonable.

matklad:
closes #4171 

# Test Plan

Pure reafactor, covered by existing tests.

### DIFF
--- a/core/primitives/src/sharding.rs
+++ b/core/primitives/src/sharding.rs
@@ -12,7 +12,7 @@ use crate::types::validator_stake::{ValidatorStake, ValidatorStakeIter, Validato
 use crate::types::{Balance, BlockHeight, Gas, MerkleHash, ShardId, StateRoot};
 use crate::validator_signer::ValidatorSigner;
 #[cfg(feature = "protocol_feature_block_header_v3")]
-use crate::version::{ProtocolFeature, PROTOCOL_FEATURES_TO_VERSION_MAPPING};
+use crate::version::ProtocolFeature;
 use crate::version::{ProtocolVersion, ProtocolVersionRange, SHARD_CHUNK_HEADER_UPGRADE_VERSION};
 use reed_solomon_erasure::ReconstructShard;
 use std::sync::Arc;
@@ -530,17 +530,16 @@ impl ShardChunkHeader {
 
     #[cfg(feature = "protocol_feature_block_header_v3")]
     pub fn version_range(&self) -> ProtocolVersionRange {
-        let block_header_v3_version =
-            PROTOCOL_FEATURES_TO_VERSION_MAPPING.get(&ProtocolFeature::BlockHeaderV3).unwrap();
+        let block_header_v3_version = ProtocolFeature::BlockHeaderV3.protocol_version();
         match &self {
             ShardChunkHeader::V1(_) => {
                 ProtocolVersionRange::new(0, Some(SHARD_CHUNK_HEADER_UPGRADE_VERSION))
             }
             ShardChunkHeader::V2(_) => ProtocolVersionRange::new(
                 SHARD_CHUNK_HEADER_UPGRADE_VERSION,
-                Some(*block_header_v3_version),
+                Some(block_header_v3_version),
             ),
-            ShardChunkHeader::V3(_) => ProtocolVersionRange::new(*block_header_v3_version, None),
+            ShardChunkHeader::V3(_) => ProtocolVersionRange::new(block_header_v3_version, None),
         }
     }
 }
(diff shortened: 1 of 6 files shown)


matklad:
This feels too easy.... Is there some requirenment I am missing here? In particular, there's this assert:

https://github.com/near/nearcore/blob/master/core/primitives/src/version.rs#L158-L167

it looks like it tries to handle the case where a protocol feature is both stable and nightly...

bowenwang1996:
> This feels too easy.... Is there some requirenment I am missing here? In particular, there's this assert:
> 
> https://github.com/near/nearcore/blob/master/core/primitives/src/version.rs#L158-L167
> 
> it looks like it tries to handle the case where a protocol feature is both stable and nightly...

I think that is not intended and a protocol feature should not be on both nightly and stable.

frol (approved):
:heart_eyes:

matklad:
>I think that is not intended and a protocol feature should not be on both nightly and stable.

Then I believe this is indeed all we need here.

core/primitives/src/sharding.rs
```diff
@@ -1127,8 +1126,7 @@ impl EncodedShardChunk {
         #[cfg(not(feature = "protocol_feature_block_header_v3"))]
         let block_header_v3_version = None;
         #[cfg(feature = "protocol_feature_block_header_v3")]
-        let block_header_v3_version =
-            PROTOCOL_FEATURES_TO_VERSION_MAPPING.get(&ProtocolFeature::BlockHeaderV3).copied();
+        let block_header_v3_version = Some(ProtocolFeature::BlockHeaderV3.protocol_version());
```
matklad, line 1129: cc @birchmd 

I *think* the way to do what you want to do using existing patterns would be:

```diff
λ git diff
diff --git a/core/primitives/src/sharding.rs b/core/primitives/src/sharding.rs
index 2e5d0561..fd70d8ed 100644
--- a/core/primitives/src/sharding.rs
+++ b/core/primitives/src/sharding.rs
@@ -4,6 +4,7 @@ use serde::Serialize;
 
 use near_crypto::Signature;
 
+use crate::checked_feature;
 use crate::hash::{hash, CryptoHash};
 use crate::merkle::{combine_hash, merklize, MerklePath};
 use crate::receipt::Receipt;
@@ -1123,11 +1124,6 @@ impl EncodedShardChunk {
         content.reconstruct(rs).unwrap();
         let (encoded_merkle_root, merkle_paths) = content.get_merkle_hash_and_paths();
 
-        #[cfg(not(feature = "protocol_feature_block_header_v3"))]
-        let block_header_v3_version = None;
-        #[cfg(feature = "protocol_feature_block_header_v3")]
-        let block_header_v3_version = Some(ProtocolFeature::BlockHeaderV3.protocol_version());
-
         if protocol_version < SHARD_CHUNK_HEADER_UPGRADE_VERSION {
             #[cfg(feature = "protocol_feature_block_header_v3")]
             let validator_proposals =
@@ -1150,9 +1146,11 @@ impl EncodedShardChunk {
             );
             let chunk = EncodedShardChunkV1 { header, content };
             (Self::V1(chunk), merkle_paths)
-        } else if block_header_v3_version.is_none()
-            || protocol_version < block_header_v3_version.unwrap()
-        {
+        } else if !checked_feature!(
+            "protocol_feature_block_header_v3",
+            BlockHeaderV3,
+            protocol_version
+        ) {
             #[cfg(feature = "protocol_feature_block_header_v3")]
             let validator_proposals =
                 validator_proposals.into_iter().map(|v| v.into_v1()).collect();
```

But I agree that it looks ugly for cases like 

```rust
if protocol_version < feat1 {
} else if protocol_version < feat2 {
} else {
}
```

The core issues here is that features themselves are gated by cfg, so you can't just look at the features' protocol version. 

It might look slightly better if the order of cases is reversed from newest to oldest:

```rust
if check_feature(feat1) {
} else if check_feature(feat2) {
} else {
}
```

But in this case, the problem is that `SHARD_CHUNK_HEADER_UPGRADE_VERSION` is not like other features :(

birchmd: The pattern I was following was the one from `block_header.rs`, where you have a variable which is `None` when the feature is not enabled, and `Some` otherwise. But I don't like that pattern and I agree this is nicer. And yes, the shard chunk header upgrade was done before `checked_feature` was developed so it is somewhat ad-hoc.

bowenwang1996 (approved):

PR #4192, opened 2021-04-02
feat(tool): dump given contract state by state-viewer

ailisp:
This is initial step to  "dump and reproduce data for repro contract execution locally". In state viewer this become available for any one who run a node. We plan to also expose this data in RPC so developer can find out this information without run a node, for example, on explorer

Usage and Test Plan
------
For contracts written with near-sdk-rs,
```
target/debug/state-viewer dump_account_storage --account test.near
```

For contracts written with near-sdk-as, storage key is defined by developer, for example this contract: https://github.com/near-examples/counter/blob/f72865e7cdaf09515ef310ae02bbf97dbb8432e2/assembly/main.ts#L6
state can be retrieved by:
```
target/debug/state-viewer dump_account_storage --account test.near --storage_key counter
```

File is output as binary file contains exact bytes in storage, to given output file (default to output.bin)


bowenwang1996 (commented):
This always dumps the latest state on the node. Is this what we want?

test-utils/state-viewer/src/main.rs
```diff
@@ -585,6 +611,30 @@ fn main() {
                 account_id
             );
         }
+        ("dump_account_storage", Some(args)) => {
+            let account_id = args.value_of("account").expect("account is required");
```
matklad, line 615: It's better to mark clap arg as required and just unwrap here: https://docs.rs/clap/2.33.3/clap/struct.Arg.html#method.required

test-utils/state-viewer/src/main.rs
```diff
@@ -586,6 +619,46 @@ fn main() {
...
+                    account_id: account_id.to_string(),
+                    key: storage_key.as_bytes().to_vec(),
+                };
+                let item = trie.get(state_root, &key.to_vec());
+                let value = item.unwrap();
+                if let Some(value) = value {
+                    let record = StateRecord::from_raw_key_value(key.to_vec(), value).unwrap();
+                    match record {
+                        StateRecord::Data { account_id: _, data_key: _, value } => {
+                            fs::write(output, &value).unwrap();
```
bowenwang1996, line 648: why do we only care about value but not key?

ailisp: The user who run this contract provided key: account and storage_key arg, so they already know the key

bowenwang1996 (approved):

matklad (approved):
The code looks good to me now!

I don't have an opinion about "is this what we want?". I *think* we want to expose this via RPC, so exposing via CLI might not be necessary? There's also bowen's question about latest vs historical state. But you know more about this than I do :)


Aside: we probably should be switching for clap 3 beta for non-core CLI tools, clap's stringly-typed API is too verbose.

ailisp:
> I don't have an opinion about "is this what we want?". I think we want to expose this via RPC, so exposing via CLI might not be necessary? 

It is helpful in the case of dump state from a localnet node, or dump state from testnet/mainnet node when state is too big to download in rpc/explorer.

> There's also bowen's question about latest vs historical state. But you know more about this than I do :)

This is solved, there's an option of given height now

PR #4193, opened 2021-04-02
feat(chain): Chunk Only Producers

birchmd:
Chunk-only producers are an important stepping stone towards sharding in mainnet. See https://gov.near.org/t/block-and-chunk-producer-selection-algorithm-in-simple-nightshade/66 for more details. Also see https://github.com/near/NEPs/pull/167 for the spec this work is based on.

This PR does most of the work towards landing this feature. Much of the work with this PR was updating tons of tests because they were using the assumption that validators produce blocks/chunk in a cyclic order. That is no longer true because the randomness is done on the fly at each height instead of when processing the proposals.

This PR is not yet suitable for merging to master; missing items are listed below:

- [ ] Nayduck failures (looks like some tests are failing -- http://nayduck.eastus.cloudapp.azure.com:3000/#/run/1452)
- [ ] Writing a new pytest to see this feature working end-to-end. This PR adds some new tests, and fixes a lot of old tests, so it probably works, but it's always nice to see an integration test.

List of (possible) Nayduck failures to be addressed:
```
expensive nearcore test_rejoin test::test_4_20_kill1_two_shards	
pytest sanity/one_val.py	
pytest sanity/rpc_state_changes.py	
pytest sanity/staking2.py	
pytest sanity/staking_repro1.py	
pytest sanity/state_sync2.py	
pytest sanity/sync_chunks_from_archival.py	
pytest stress/stress.py 3 3 3 0 staking transactions local_network packets_drop	
pytest stress/stress.py 3 3 3 0 staking transactions node_restart packets_drop	
pytest stress/stress.py 3 3 3 0 staking transactions node_restart wipe_data	
pytest sanity/gc_after_sync.py
pytest sanity/gc_sync_after_sync.py swap_nodes
```


bowenwang1996:
Nayduck run http://nayduck.eastus.cloudapp.azure.com:3000/#/run/2015

chain/epoch_manager/src/shard_assignment.rs
```diff
@@ -0,0 +1,274 @@
...
+/// best when the number of chunk producers is greater than
+/// `num_shards * min_validators_per_shard`.
+pub fn assign_shards<T: HasStake + Eq + Clone>(
+    chunk_producers: Vec<T>,
+    num_shards: usize,
+    min_validators_per_shard: usize,
+) -> Result<Vec<Vec<T>>, NotEnoughValidators> {
+    // Initially, sort by number of validators, then total stake
+    // (i.e. favour filling under-occupied shards first).
+    let mut shard_index: MinHeap<(usize, Balance, ShardId)> =
```
SkidanovAlex, line 20: This is a very confusing name. `shard_validators_heap`, or something along the lines?

chain/epoch_manager/src/shard_assignment.rs
```diff
@@ -0,0 +1,274 @@
...
+
+/// Assign chunk producers (a.k.a validators) to shards. The i-th element
+/// of the output corresponds to the validators assigned to the i-th shard.
+/// This function ensures that every shard has at least `min_validators_per_shard`
+/// assigned to it, and attempts to balance the stakes between shards (keep the total
+/// stake assigned to each shard approximately equal). This function performs
+/// best when the number of chunk producers is greater than
+/// `num_shards * min_validators_per_shard`.
+pub fn assign_shards<T: HasStake + Eq + Clone>(
+    chunk_producers: Vec<T>,
```
SkidanovAlex, line 14: Are `chunk_producers` sorted in any particular way (in particular, in descending order of stake)?
It would be good to document here.
If they are not sorted in descending stake, it might be worth doing?

chain/epoch_manager/src/shard_assignment.rs
```diff
@@ -0,0 +1,274 @@
...
+    while shard_index.peek().unwrap().0 < min_validators_per_shard {
+        let (assignment_index, cp) = cp_iter
+            .next()
+            .expect("cp_iter should contain enough elements to minimally fill each shard");
+        let (least_validator_count, shard_stake, shard_id) =
+            shard_index.pop().expect("shard_index should never be empty");
+
+        if assignment_index < num_chunk_producers {
+            // no need to worry about duplicates yet; still on first pass through validators
+            shard_index.push((least_validator_count + 1, shard_stake + cp.get_stake(), shard_id));
```
SkidanovAlex, line 114: Between here, and two similar places above, you use inconsistent naming for the first two values you pop (you arbitrarily use or not use the word `least`).

chain/epoch_manager/src/shard_assignment.rs
```diff
@@ -0,0 +1,274 @@
...
+            num_shards,
+            min_validators_per_shard,
+            num_chunk_producers,
+        );
+    }
+
+    Ok(result)
+}
+
+fn assign_with_possible_repeats<T: HasStake + Eq, I: Iterator<Item = (usize, T)>>(
```
SkidanovAlex, line 95: This method seems overly complex.

```
random_shuffle(validators) // optionally
last_ord = 0
for (shard_id in ...) {
    for validator_ord in 0..min_validators_per_shard {
        result[shard_id].push(last_ord);
        last_ord += 1
        last_ord %= num_chunk_producers;
    }

}
```

SkidanovAlex (commented):
Only reviewed the first half so far, some comments inline.

core/primitives/src/rand.rs
```diff
@@ -0,0 +1,187 @@
...
+        let weighted_index = WeightedIndex::new(weights);
+
+        assert_eq!(weighted_index.get_aliases(), &[1, 0, 3, 1, 3, 3, 3]);
+
+        assert_eq!(weighted_index.get_no_alias_odds(), &[35, 40, 28, 29, 28, 28, 35]);
+    }
+
+    #[test]
+    fn test_sample_should_produce_correct_distribution() {
+        let weights = vec![5, 1, 1];
```
SkidanovAlex, line 161: This test fails for [4, 6, 1, 1] (with asserts properly updated).

If you look at the aliases section in the page you linked (https://www.keithschwarz.com/darts-dice-coins/), this is exactly the example on which he explains it, but with 6 and 4 swapped. If I don't swap them (leave as 6, 4, 1, 1), your code works correctly (it 6,4,1,1 -> 4,4,3,1 -> 3,4,3,2 -> 3,3,3,3), but if I do, you hit the issue he explains at the very beginning of the section (4,6,1,1 -> 3,6,1,2 -> 3,4,3,2 -> ???).

`big` needs to be a maxheap.

Also, the test itself is extremely flacky (fails with relative error 0.7%-0.8% one out of 5 executions). Though I ran it with `rand::thread_rng` instead of hashes, may be it changes the distribution.

chain/epoch_manager/src/validator_selection.rs
```diff
@@ -0,0 +1,820 @@
...
+        max_cp_selected,
+        min_stake_ratio,
+        num_shards,
+    );
+
+    // since block producer proposals could become chunk producers, their actual stake threshold
+    // is the smaller of the two thresholds
+    let bp_stake_threshold = cmp::min(bp_stake_threshold, cp_stake_threshold);
+
+    // proposals remaining chunk_producer_proposals were not selected for either role
```
SkidanovAlex, line 65: This comment is very confusing. I can't tell if it's intended to be such and I just can't parse it, or if it has something misspelled :)

SkidanovAlex (commented):
Still haven't reviewed `validator_selection.rs`.

chain/chain/src/store_validator/validate.rs
```diff
@@ -678,26 +678,17 @@ pub(crate) fn outcome_indexed_by_block_hash(
...
-                    ) {
-                        outcome_ids.extend(unwrap_or_err_db!(
-                            sv.store.get_ser::<Vec<CryptoHash>>(
-                                ColOutcomeIds,
-                                &get_block_shard_id(block.hash(), chunk_header.shard_id())
-                            ),
-                            "Can't get Outcome ids by Block Hash"
-                        ));
-                    }
+                if let Ok(Some(_)) = sv.store.get_ser::<ChunkExtra>(
```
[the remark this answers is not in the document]
bowenwang1996: That is the difficult part -- there is a test that fails because the catch up has not finished downloading the state and as a result we don't actually have chunk extra for the block. There is nothing wrong with that and I am not sure what would be a better condition to check

chain/client/src/client.rs
```diff
@@ -1425,6 +1425,8 @@ impl Client {
         let head = self.chain.head()?;
         if let Some(next_epoch_id) = self.get_next_epoch_id_if_at_boundary(&head)? {
             self.forward_tx(&next_epoch_id, tx)?;
+        } else {
```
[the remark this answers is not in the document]
bowenwang1996: > looking at the place where this is called in process_tx_internal it appears that if an active validator received a tx directly, the tx will never be forwarded before this change, resulting in very long processing times

Yes that is exactly the intention. One python test failed because of this.

> But if this is indeed your intent here, the method needs to be renamed?

What do you suggest we rename it to?

bowenwang1996 (approved):
Approve to unblock this PR and avoid having to deal with merge conflicts with master. If there are any further testing needed, we can do it after it is merged as a nightly protocol feature

bowenwang1996:
@frol looks like it needs your approval 🙏

frol (approved):
The changes in Rosetta-RPC and primitives LGTM

bowenwang1996:
@nikurt please review

nikurt (approved):

PR #4199, opened 2021-04-06
feat(config): configurable trie viewer state size limit

ailisp:
As required by exposing download contract state for a local repro.

Test Plan
------------
- [x] Tests pass.
- [x] new binary work in old config.
- [x] new binary with a new config file that has very small size limit return size too big error.
- [x] Added a debug print (removed in PR) to see config.json without `trie_viewer_state_size_limit` get limit of Some(50000), with `"trie_viewer_state_size_limit": null` get limit of `None`.


runtime/runtime/src/state_viewer/mod.rs
```diff
@@ -22,14 +22,21 @@ use crate::{actions::execute_function_call, ext::RuntimeExt};
 
 pub mod errors;
 
-pub struct TrieViewer {}
+pub struct TrieViewer {
+    /// Upper bound of the byte size of contract state that is still viewable. None is no limit
+    state_size_limit: Option<u64>,
```
matklad, line 27: Unsure if we need an `Option` here. Practically speaking, no limit and very large limit are indistinguishable, so maybe it makes sense to just use `u64` here and let the caller pass `u64::MAX`?

runtime/runtime/src/state_viewer/mod.rs
```diff
@@ -22,14 +22,21 @@ use crate::{actions::execute_function_call, ext::RuntimeExt};
...
 
     pub fn new() -> Self {
-        Self {}
+        Self { state_size_limit: Self::DEFAULT_CONTRACT_STATE_SIZE_LIMIT }
+    }
+
+    pub fn state_size_limit(mut self, limit: Option<u64>) -> Self {
+        self.state_size_limit = limit;
+        self
     }
```
matklad, line 40: It feels a bit odd to have `DEFAULT_CONTRACT_STATE_SIZE_LIMIT` be specified here. In the only "real" use-case, we immediately override the default value:

https://github.com/near/nearcore/pull/4199/files#diff-f0e8081b926da5d399c24bd42912c9cf2cb268e2d783b4f5f3184ecfbec33ee1R150

Logically, defaults belong to config as well. 

I think (70% confidence) that it's better to change 

```rust
impl TrieViewer {
    pub const DEFAULT_CONTRACT_STATE_SIZE_LIMIT: Option<u64> = Some(50_000);

    pub fn new() -> Self {
        Self { state_size_limit: Self::DEFAULT_CONTRACT_STATE_SIZE_LIMIT }
    }

    pub fn state_size_limit(mut self, limit: Option<u64>) -> Self {
        self.state_size_limit = limit;
        self
    }
}
```

to

```rust
impl TrieViewer {
    pub fn new_with_state_size_limit(state_size_limit: Option<u64>) -> Self {
        Self { state_size_limit }
    }
}
```

That is, it seems the fact that the `DEFAULT_CONTRACT_STATE_SIZE_LIMIT` const lives here makes sense if the limit is not configurable. If it *is* configurable, `TrieViewer` doesn't need to know the default.

runtime/runtime/src/state_viewer/mod.rs
```diff
@@ -132,10 +139,12 @@ impl TrieViewer {
                 let code_len = get_code(state_update, account_id, Some(account.code_hash()))?
                     .map(|c| c.code.len() as u64)
                     .unwrap_or_default();
-                if account.storage_usage() > Self::CONTRACT_STATE_SIZE_LIMIT + code_len {
-                    return Err(errors::ViewStateError::AccountStateTooLarge {
-                        requested_account_id: account_id.clone(),
-                    });
+                if let Some(limit) = self.state_size_limit {
+                    if account.storage_usage() > limit + code_len {
```
matklad, line 143: nit:

```suggestion
                    if account.storage_usage() - code_len > limit {
```

now that `limit` is configurable, this could in theory overflow for large limits. Unlikely, but possible. Moving `code_len` to the left is valid if account.storage_usage() guarantees to include `code_len`. If that's not the case, `-` should be replace with saturating_sub.

runtime/runtime/src/state_viewer/mod.rs
ailisp: u64::MAX is not easy to pass in config.json. while this one it's "trie_viewer_state_size_limit": null

runtime/runtime/src/state_viewer/mod.rs
ailisp: Unlikely hit but still a good security check: people would either choose some far smaller limit like 50K, 50M, or unlimited `None`,

runtime/runtime/src/state_viewer/mod.rs
ailisp: I chose `new` because there's only one place override state_size_limit:
image

matklad: >I chose new because there's only one place override state_size_limit:

uhu, but not how all *other* places are test code. The single non-test usage uses `new` immediately followed by `state_size_limit`. 

This is what makes me feel uneasy about the code: you would feel that `DEFAULT_CONTRACT_STATE_SIZE_LIMIT` is something that is used by default, but in the actual production code it is never used and is always overwritten by something else (although most likely it is overwritten with the same value).

runtime/runtime/src/state_viewer/mod.rs
matklad: I guess, we can have it as `Option` in the config.json, and then to `limit.unwrap_or(u64::MAX)` when converting from config to something we pass around. But yeah, it's also convenient to specify `None` as a default for tests. I guess I don't have a strong opinion one way or another here.

runtime/runtime/src/state_viewer/mod.rs
ailisp: Addressed in ad5a057

runtime/runtime/src/state_viewer/mod.rs
ailisp: I feel to say" limit is None" is more intuitive than limit is "u64::max" so i just passed that around

bowenwang1996 (commented):
As mentioned before, this is not enough because it will timeout on the jsonrpc side first. You can test it with some contract whose state is > 10MB.

neard/src/config.rs
```diff
@@ -426,6 +430,8 @@ pub struct Config {
     pub epoch_sync_enabled: bool,
     #[serde(default = "default_view_client_throttle_period")]
     pub view_client_throttle_period: Duration,
+    #[serde(default = "default_trie_viewer_state_size_limit")]
+    pub trie_viewer_state_size_limit: Option<u64>,
 }
```
matklad, line 435: Unrelated to the PR, but the odd thing I've noticed is that we use both `#[serde(default)]` on the struct itself and `#[serde(default = "default_trie_viewer_state_size_limit")]`  on the field. That's tautological and can lead to an inconsistency between `Default` impl and the defaults during deserialization. I think for this particular struct we want to remove all `sered(default` from fields and just inline the the default functions into `impl Default for Config`. 

In a similar vein, some structs in this file (eg Network) have `impl Default` but do not have struct-level `serde(default)`. This also feels odd a bit.

ailisp:
@bowenwang1996 I didn't see any timeout configured for view_client query and in async fn query. (If there were I expect to see something like `timeout(self.polling_config.polling_timeout, async {` in async fn query. Is it an actix_web option? If so I also don't see any option that config timeout: https://docs.rs/actix-web/3.3.2/actix_web/struct.HttpServer.html 🤔 Do you have some clues where set timeout (our code or actix-web)? I'll also experiment locally with some 10M state

bowenwang1996:
> @bowenwang1996 I didn't see any timeout configured for view_client query and in async fn query. (If there were I expect to see something like `timeout(self.polling_config.polling_timeout, async {` in async fn query. Is it an actix_web option? If so I also don't see any option that config timeout: https://docs.rs/actix-web/3.3.2/actix_web/struct.HttpServer.html 🤔 Do you have some clues where set timeout (our code or actix-web)? I'll also experiment locally with some 10M state

@frol @khorolets did we remove timeout on queries?

bowenwang1996 (approved):

neard/src/config.rs
frol: Yeah, clearly needs some look. Should not be part of this PR, though, but good call!

neard/src/runtime/mod.rs
```diff
@@ -144,9 +144,10 @@ impl NightshadeRuntime {
         genesis: &Genesis,
         initial_tracking_accounts: Vec<AccountId>,
         initial_tracking_shards: Vec<ShardId>,
+        trie_viewer_state_size_limit: Option<u64>,
```
frol, line 147: Just wanted to share my pain: I really struggle when we keep dropping new arguments to the methods, and soon end up with `new(1, None, Some(3), "node.test", 42, "", None)` :cry: 

Well, we are already there:

```
neard::NightshadeRuntime::new(
        Path::new("."),
        store1,
        &genesis,
        vec![],
        vec![],
        None,
    )
```

It will require another round of refactoring later outside of this PR, I believe, I just wanted to bring it up for future reference.

frol (approved):

neard/src/runtime/mod.rs
matklad: :+1: here's the pattern we use in rust-analyzer to fight this problem: https://github.com/rust-analyzer/rust-analyzer/blob/master/docs/dev/style.md#functions-with-many-parameters

EgorKulikov (approved):

runtime/runtime/src/state_viewer/mod.rs
```diff
@@ -132,10 +132,12 @@ impl TrieViewer {
                 let code_len = get_code(state_update, account_id, Some(account.code_hash()))?
                     .map(|c| c.code.len() as u64)
                     .unwrap_or_default();
-                if account.storage_usage() > Self::CONTRACT_STATE_SIZE_LIMIT + code_len {
-                    return Err(errors::ViewStateError::AccountStateTooLarge {
-                        requested_account_id: account_id.clone(),
-                    });
+                if let Some(limit) = self.state_size_limit {
+                    if account.storage_usage() - code_len > limit {
```
evgenykuzyakov, line 136: Or move it to the right
```suggestion
                    if account.storage_usage().saturating_sub(code_len) > limit {
```

evgenykuzyakov (approved):

PR #4200, opened 2021-04-07
test(jsonrpc): Cover JSON-RPC transaction related methods with tests for error messages (to watch for backward compatibility after refactor)

khorolets:
I've refactored transaction related methods on JSON-RPC in this PR #4180

Here I've added tests to check that error messages are the same, so we can ensure that refactor is backward compatible. We need to merge this PR before #4180

### DIFF
--- a/neard/tests/rpc_nodes.rs
+++ b/neard/tests/rpc_nodes.rs
@@ -847,3 +847,233 @@ fn test_query_rpc_account_view_account_doesnt_exist_must_return_error() {
         });
     });
 }
+
+#[test]
+fn test_tx_not_enough_balance_must_return_error() {
+    init_integration_logger();
+    heavy_test(|| {
+        run_actix_until_stop(async move {
+            let num_nodes = 1;
+            let dirs = (0..num_nodes)
+                .map(|i| {
+                    tempfile::Builder::new()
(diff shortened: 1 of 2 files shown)


frol (approved):
LGMT

neard/tests/rpc_nodes.rs
```diff
@@ -847,3 +847,232 @@ fn test_query_rpc_account_view_account_doesnt_exist_must_return_error() {
...
+                    let res = view_client.send(GetBlock::latest()).await;
+                    if let Ok(Ok(block)) = res {
+                        if block.header.height > 10 {
+                            let _ = client
+                                .EXPERIMENTAL_check_tx(to_base64(&bytes))
+                                .map_err(|err| {
+                                    assert_eq!(
+                                        err.data.unwrap().to_string(),
+                                        "\"Node doesn\'t track this shard. Cannot determine whether the transaction is valid\"".to_string()
+                                    );
```
matklad, line 1065: A possible alternative: 

```suggestion
                                    assert_eq!(
                                        err.data.unwrap(),
                                        serde_json::json!("Node doesn't track this shard. Cannot determine whether the transaction is valid")
                                    );
```

PR #4203, opened 2021-04-08
feat(contract): contracts to cover all cases of local reproduce contract call with downloaded states

ailisp:
WIP, contracts mostly done, will test contracts and state dump on testnet tomorrow


runtime/near-test-contracts/README.md
```diff
@@ -15,3 +16,14 @@ to the Cargo.toml and use `near_test_contract::rs_contract()`.
...
+
+## Contracts for testing local reproduce transaction execution
+There're three contracts that reflect different state-dependent transaction execution on testnet.
+These contracts are supposed to be deployed on testnet and then obtain a copy of state with rpc, and 
+reproduce its transaction locally with (WIP) near-vm-standalone with custom scripts.
+- simple-state: a simple, one huge state stored for contract, simple contract get/set
+- multiple-state: contract store multiple states under different keys on trie with get/set method
+- cross-contract: cross contract call to read state from another contract or update state of another contract
+
+Above scenarios should cover enough use cases of developer downloading states of testnet contracts, and reproduce it locally.
```
ailisp, line 28: @matklad I could think is single state, multiple state, cross contract read, cross contract write. What other scenarios that have a different workflows in reproducing can you think of?

matklad: Looking at the https://github.com/near/nearcore/blob/92703b38b6ce131c1bcc69f9cadd6734c089418c/runtime/near-test-contracts/test-contract-rs/src/lib.rs#L7-L159,

I think `random_seed, block_index, block_timestamp, epoch_height, storage_usage` are things we want to specifically test for. I expect that most of those will fall out naturally and won't require any special handlig, but still make sense to double-check. 

With respect to storage, I think doing a `read` of a non-existing key is another case we need to cover.

runtime/near-test-contracts/cross-contract/Cargo.toml
```diff
@@ -0,0 +1,22 @@
...
+name = "cross-contract"
+version = "0.0.0"
+authors = ["Near Inc <hello@nearprotocol.com>"]
+edition = "2018"
+
+[lib]
+crate-type = ["cdylib", "rlib"]
+
+[dependencies]
+near-sdk = "3.1.0"
```
matklad, line 11: Hm, I am unsure about this. 

On the one hand, this contract has a dedicated Cargo.lock and is re-build on CI, so depending on an sdk would work fine. It would be reproducible, won't require real cyclic dependencies, and would be easy to use (the build process for the contract is automated). 

On the other hand, it feels like this pulls in too much stuff for the thing we actually want to test. At least, there's a problem of increased compile times. And then we'd have to update Cargo.tomls when the new sdk  version is released. 

More philosophically, here we test functionality which is independent of a particular sdk, so it feels odd  that we use a specific SDK.


I wonder, can we use existing `test-contract-rs` here instead? `write_key_value` and `read_value` methods seem like they cover `simple-state` and `multi-state`. Can we use several instances of this contract to cover cross-contract as well?

runtime/near-test-contracts/simple-state/src/lib.rs
```diff
@@ -0,0 +1,72 @@
...
+
+    #[test]
+    fn get_nonexistent_message() {
+        let context = get_context(true);
+        testing_env!(context);
+        let contract = SimpleState::default();
+        assert_eq!(None, contract.get_status("francis.near".to_string()));
+        assert_eq!(get_logs(), vec!["get_status for account_id francis.near"])
+    }
+}
```
matklad, line 72: I don't think we actually run this tests anywhere, so, if we do move forward with sdk-based contracts, it's better to remove these.

runtime/near-test-contracts/README.md
ailisp: Good points!

runtime/near-test-contracts/cross-contract/Cargo.toml
ailisp: The audience and purpose is different. test-contract-rs is unit testing contract runtime, but developer almost always use near-sdk-rs to write contracts (and reproduce given its state), so these contracts are both a test and a demonstration to contract developers that how they're going to download states and reproduce execution. For example, multi-state, it used nested UnorderedMap, a common pattern in a real world contract, and it translate to multiple trie items: every key value pair in UnorderedMap is flattened to be a trie key and value. This seems non trivial to express with plain write_key_value (and even if you do so you still need to show it works in a real contract).
As regarding concern of build time and extra dependency and odd to depend on specific sdk, I think you're right, maybe make sense to place these contracts in another repo, like contract-debug-demo or just near-sdk-rs, as a highlevel, developer facing documentation thing than i test.
For now I'll locally use these contracts to test implementing vm-standalone + rust scripts reproduce

matklad: Aha, great points! Yeah, it seems like this then belonges elsewheer. I wonder if `near-sdk-rs` would be an appropriate place, as it is essentailly an example?

ailisp: Sounds good, I'll remake examples in near-sdk-rs

PR #4205, opened 2021-04-08
feat: precompile contracts on deploy

olonho:
Implement logic for precompilation of contracts in deploy action.
We precompile with the current default VM and config, if it changes -
contract executor still keeps the fallback to recompile contracts
as needed, so no protocol upgrades needed.
Cost reevaluation will be done separately.

Test plan
---------
cargo test --all


matklad (changes requested):
This feels overall, but I think it would be prudent to tidy up the interface a bit  (hide VMKind & introduce a dedicated  error)

runtime/near-vm-runner/src/cache.rs
```diff
@@ -292,3 +292,46 @@ pub mod wasmer1_cache {
...
+    if cache.is_none() {
+        return Ok(false);
+    }
+    let cache = cache.unwrap();
+    let key = get_key(wasm_code, vm_kind, config);
+    // Check if we already cached with such a key.
+    match cache.get(&(key.0).0) {
+        // If so - do not override.
+        // TODO: is it correct?
+        Ok(_) => return Ok(false),
```
matklad, line 311: Huh, should this be `Ok(None)`? It seems at the moment we *always* hit the `return` branch.

runtime/near-vm-runner/src/cache.rs
```diff
@@ -292,3 +292,46 @@ pub mod wasmer1_cache {
         }
     }
 }
+
+pub fn cache_contract(
+    vm_kind: VMKind,
+    wasm_code: &ContractCode,
+    config: &VMConfig,
+    cache: Option<&dyn CompiledContractCache>,
+) -> Result<bool, VMError> {
```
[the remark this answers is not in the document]
bowenwang1996: In addition to what @matklad said, we should also document what the return value means. For me it is not clear what the boolean value means here

runtime/near-vm-runner/src/cache.rs
```diff
@@ -292,3 +292,46 @@ pub mod wasmer1_cache {
         }
     }
 }
+
+pub fn cache_contract(
```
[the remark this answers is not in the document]
bowenwang1996: yeah it does seem `cache_contract` may be a bit misleading here as we do compile the contract in some cases (when it is not cached)

runtime/near-vm-runner/src/cache.rs
```diff
@@ -292,3 +292,46 @@ pub mod wasmer1_cache {
...
+    vm_kind: VMKind,
+    wasm_code: &ContractCode,
+    config: &VMConfig,
+    cache: Option<&dyn CompiledContractCache>,
+) -> Result<bool, VMError> {
+    if cache.is_none() {
+        return Ok(false);
+    }
+    let cache = cache.unwrap();
```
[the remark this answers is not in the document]
olonho: Done.

runtime/runtime/src/actions.rs
```diff
@@ -436,7 +438,16 @@ pub(crate) fn action_deploy_contract(
     );
     account.set_code_hash(code.get_hash());
     set_code(state_update, account_id.clone(), &code);
-    Ok(())
+    // Precompile the contract.
+    match cache_contract(
+        VMKind::default(),
```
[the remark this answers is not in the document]
olonho: Agreed, fixed.

runtime/near-vm-runner/src/cache.rs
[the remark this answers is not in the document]
olonho: Right, changed.

runtime/near-vm-runner/src/cache.rs
olonho: Right, fixed.

runtime/near-vm-runner/src/cache.rs
```diff
@@ -292,3 +292,46 @@ pub mod wasmer1_cache {
...
+
+pub fn cache_contract(
+    vm_kind: VMKind,
+    wasm_code: &ContractCode,
+    config: &VMConfig,
+    cache: Option<&dyn CompiledContractCache>,
+) -> Result<bool, VMError> {
+    if cache.is_none() {
+        return Ok(false);
+    }
```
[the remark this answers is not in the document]
olonho: Well, here I'm kind of both agree and disagree.  In a sense, idea was to delegate decision on "right" behavior if cache isn't found to contract runtime.

runtime/near-vm-runner/src/cache.rs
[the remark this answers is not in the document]
olonho: Done

runtime/near-vm-runner/src/tests/contract_preload.rs
```diff
@@ -188,3 +193,35 @@ pub fn test_run_preloaded() {
...
+
+fn test_precompile_vm(vm_kind: VMKind) {
+    let mock_cache = MockCompiledContractCache::new(0);
+    let cache: Option<&dyn CompiledContractCache> = Some(&mock_cache);
+    let vm_config = VMConfig::default();
+    let code1 = ContractCode::new(near_test_contracts::rs_contract().to_vec(), None);
+    let code2 = ContractCode::new(near_test_contracts::ts_contract().to_vec(), None);
+
+    let result = precompile_contract_impl(vm_kind, &code1, &vm_config, cache);
+    assert_eq!(result, Result::Ok(true));
```
matklad, line 205: nit: here and elsewhere

```suggestion
    assert_eq!(result, Ok(true));
```

runtime/runtime/src/actions.rs
```diff
@@ -436,6 +438,11 @@ pub(crate) fn action_deploy_contract(
     );
     account.set_code_hash(code.get_hash());
     set_code(state_update, account_id.clone(), &code);
+    // Precompile the contract and store result (compiled code or error) in the database.
+    #[allow(unused_must_use)]
+    {
+        precompile_contract(&code, &apply_state.config.wasm_config, apply_state.cache.as_deref());
+    }
```
matklad, line 445: This is usually written as 

```suggestion
    let _ = precompile_contract(&code, &apply_state.config.wasm_config, apply_state.cache.as_deref());
```

runtime/near-vm-runner/src/cache.rs
```diff
@@ -292,3 +292,61 @@ pub mod wasmer1_cache {
         }
     }
 }
+
+pub(crate) fn precompile_contract_impl(
+    vm_kind: VMKind,
+    wasm_code: &ContractCode,
+    config: &VMConfig,
```
matklad, line 299: I think this is OK for the initial implementation, but I am not entirely happy that we are passing the whole `VMConfig` here. In partticular, we hash `ExtCosts` into the key, but we don't actually use them during compilation. I think we might want to have

```rust
struct VMConfig {
    compile: VMCompileConfig,
    run: VMRunConfig,
}
```

and pass only the `Compile` part here. But this definitely can, and should be done separately.

runtime/runtime/src/actions.rs
```diff
@@ -436,6 +438,11 @@ pub(crate) fn action_deploy_contract(
     );
     account.set_code_hash(code.get_hash());
     set_code(state_update, account_id.clone(), &code);
+    // Precompile the contract and store result (compiled code or error) in the database.
```
matklad, line 441: Do we need to gate this behind some kind of runtime or compile-time flag? In particular, is it OK to enable this withoud adjusting the costs? 

I don't know the answer here, I guess that's a question for @bowenwang1996?

matklad (approved):

runtime/near-vm-runner/src/cache.rs
```diff
@@ -292,3 +292,61 @@ pub mod wasmer1_cache {
...
+    cache: Option<&dyn CompiledContractCache>,
+) -> Result<bool, ContractPrecompilatonError> {
+    let cache = match cache {
+        None => return Ok(false),
+        Some(it) => it,
+    };
+    let key = get_key(wasm_code, vm_kind, config);
+    // Check if we already cached with such a key.
+    match cache.get(&(key.0).0) {
```
bowenwang1996, line 308: what is the type of `key`? `(key.0).0` doesn't look very nice

runtime/runtime/src/actions.rs
bowenwang1996: Good point. This should come together with the change to the cost of deploy action

olonho: I guess practically we shall recompute the costs along with other recent modifications and enabling Wasmer1.

runtime/near-vm-runner/src/cache.rs
olonho: We need to get raw bits of the CryptoHash. Any better way?

runtime/runtime/src/actions.rs
olonho: But I can comment it out, if you feel it dangerous, and submit everything else.

runtime/near-vm-runner/src/cache.rs
ailisp: `impl CryptoHash { fn raw_bytes(&self) -> _ { self.0 }` ?

runtime/near-vm-runner/src/cache.rs
```diff
@@ -292,3 +292,61 @@ pub mod wasmer1_cache {
...
 }
+
+pub(crate) fn precompile_contract_impl(
+    vm_kind: VMKind,
+    wasm_code: &ContractCode,
+    config: &VMConfig,
+    cache: Option<&dyn CompiledContractCache>,
+) -> Result<bool, ContractPrecompilatonError> {
+    let cache = match cache {
+        None => return Ok(false),
```
ailisp, line 303: I think it's better to have this fn with `cache: &dyn CompiledContractCache`, there is nothing to do when cache is None

runtime/near-vm-runner/src/cache.rs
```diff
@@ -292,3 +292,61 @@ pub mod wasmer1_cache {
...
+        Some(it) => it,
+    };
+    let key = get_key(wasm_code, vm_kind, config);
+    // Check if we already cached with such a key.
+    match cache.get(&(key.0).0) {
+        // If so - do not override.
+        // TODO: is it correct?
+        Ok(Some(_)) => return Ok(false),
```
ailisp, line 311: `false` meaning is not obvious, maybe document it: Ok(true) if compile happens, Ok(false) if compile not happen. This also show `cache: Option<&dyn CompiledContractCache>` is better without `Option`, otherwise Ok(false) mean both cache hit (no compilation) and no cache (no compilation), but these two are different regarding subsequent cost

runtime/near-vm-runner/src/errors.rs
```diff
@@ -1,5 +1,24 @@
...
 use near_vm_errors::VMError;
 
 pub trait IntoVMError {
     fn into_vm_error(self) -> VMError;
 }
+
+#[derive(Debug, PartialEq)]
+pub struct ContractPrecompilatonError {
+    // TODO: actually make this a distinct error
+    inner: VMError,
```
ailisp, line 11: What is the reason of design error this way instead of make it as a variant of VMError?

runtime/near-vm-runner/src/cache.rs
ailisp: I think this TODO is correct given key is `get_key(wasm_code, vm_kind, config)`

ailisp (approved):

runtime/near-vm-runner/src/errors.rs
matklad: Yeah, that's what todo is about, it's just  we don't want to conflict with error rework PR here.

runtime/near-vm-runner/src/cache.rs
olonho: Ok, added

runtime/near-vm-runner/src/cache.rs
olonho: My feeling is that this way it's somewhat easier to use.

runtime/near-vm-runner/src/cache.rs
olonho: Actually it's somewhat logically matching to boolean in http://www.cplusplus.com/reference/map/map/insert/, i.e. if modification of cache has happened.

olonho: Removed TODO.

runtime/near-vm-runner/src/cache.rs
bowenwang1996: couldn't you do `.as_ref`?

bowenwang1996:
This PR should be bundled with the PR that updates cost (not necessarily the same PR, but needs to be released together). Otherwise deploy action cost will be incorrectly calibrated.

olonho:
Yes, the plan is to enable precompilation along with Wasmer1 enabling. Now it's disabled.

runtime/near-vm-runner/src/cache.rs
```diff
@@ -292,3 +292,60 @@ pub mod wasmer1_cache {
         }
     }
 }
+
+pub(crate) fn precompile_contract_impl(
+    vm_kind: VMKind,
+    wasm_code: &ContractCode,
+    config: &VMConfig,
+    cache: Option<&dyn CompiledContractCache>,
+) -> Result<bool, ContractPrecompilatonError> {
```
frol, line 301: What does `bool` indicate here? (well, I got the answer reading the code and the comment below) I would love to have an enum to easier reason about the return value

core/primitives-core/src/hash.rs
```diff
@@ -15,6 +15,12 @@ pub struct Digest(pub [u8; 32]);
 #[as_ref(forward)]
 pub struct CryptoHash(pub Digest);
 
+impl CryptoHash {
+    pub fn raw_bytes(&self) -> &[u8; 32] {
+        &self.0 .0
+    }
+}
```
frol, line 22: There is already `AsRef` implemented for `CryptoHash`, just use `x.as_ref()` and drop this impl.

frol (approved):
I only have a few nits to the code. The code itself looks fine to me, but I have no idea on the implications this can have in terms of costs etc, so I cannot review that part.

runtime/near-vm-runner/src/cache.rs
olonho: Not really, as some use cases need exactly sized array, i.e. ` SerializationError { hash: [u8; 32] }` and `as_ref` returns `&[u8]`. It may be worth reworking, but looks like something not belonging to this PR.

runtime/near-vm-runner/src/cache.rs
olonho: Right, enum looks way better - fixed.

core/primitives-core/src/hash.rs
olonho: Not possible, as some use cases need exactly sized array, i.e. `SerializationError { hash: [u8; 32] }` and `as_ref` returns `&[u8]`. It may be worth reworking, but looks like something not belonging to this PR.

runtime/near-vm-runner/src/errors.rs
olonho: Not sure why we shall ever make it variant of VMError, guess we uniformly rely on `VMError` for most of error passing in runtime. But I'm open to suggestions, what could be improved here.

EgorKulikov (approved):

PR #4209, opened 2021-04-09
feat(vm-standalone): add state-file option

ailisp:
It's impractical to pass state as arguments when state is large. So also add as state-file.

Test Plan
---------
Create a state.json and `--state-file state.json` works same as `--state $(cat state.json)`


matklad (approved):

olonho (approved):

nearmax:
@matklad @ailisp AFAIR we were going to get rid of this tool in favor of https://github.com/near/near-sdk-rs/tree/master/near-sdk-sim which is supposed to supersede it.

ailisp:
@nearmax It is not, near-sdk-sim is used for okr "The breakdown of the gas costs for contract execution are accessible in the simulation tests". vm standalone and tool built on top of it is used for okr "The gas cost breakdown of a deployed contract is easily reproducible on a local machine" & "Contract developers can dry-run non-deployed contracts"

PR #4210, opened 2021-04-09
fix: apply migration of epoch validator info for archival nodes

### LINKED ISSUE #4197: Validator rpc is broken for archival nodes
#3960 unfortunately introduces the problem that rpc can no longer respond to requests about historical data since it relies on the existence of keys in `ColEpochValidatorInfo`. A migration should be done for archival nodes to populate this column for all historical data.

bowenwang1996:
After #3960 archival nodes are not able to serve validator information for historical data since `ColEpochValidatorInfo` was not populated. This PR does the migration for archival nodes. Fixes #4197.

Test plan
---------
Tested with an archival node on mainnet. It works.


chain/epoch_manager/src/reward_calculator.rs
```diff
@@ -25,6 +26,22 @@ pub struct RewardCalculator {
...
+            epoch_length: config.epoch_length,
+            protocol_reward_rate: config.protocol_reward_rate,
+            protocol_treasury_account: config.protocol_treasury_account.to_string(),
+            online_max_threshold: config.online_max_threshold,
+            online_min_threshold: config.online_min_threshold,
+            #[cfg(feature = "protocol_feature_rectify_inflation")]
+            num_seconds_per_year: NUM_SECONDS_IN_A_YEAR,
+        }
+    }
+}
```
matklad, line 43: This feels like a (minor) abuse of the language/library feature.  `From`  generally implies "lossless conversion", rather that projection, and often is associated with ownership transfer. In other words, From usually goes from a type to an equivalent or bigger type. In this case, we do the opposite and take subset of fields out of GenesisConfig

In this particular case, just an inherent `pub fn new`  would work?

Inherent methods are more discoverable, and more future-proof. Suppose in the future it turns out that we need something else, besides the GenessisConfig, to construct RewardCalculator. With `new`, we'd just add an extra argument. With `From`, we need to first refactor to using `new`. 

Finally, in general it's best to use the simplest tool that gets the job done. A function is simpler than a trait. 

To clarify, this isn't at all important in terms of actual, meaningful difference in this specific example. It's just an element of Rust style :)

core/primitives-core/src/hash.rs
```diff
@@ -116,6 +116,12 @@ impl From<&CryptoHash> for Vec<u8> {
     }
 }
 
+impl From<CryptoHash> for [u8; 32] {
+    fn from(hash: CryptoHash) -> [u8; 32] {
+        (hash.0).0
+    }
+}
```
matklad, line 123: In contrast, this impl is 100% warranted: the conversion here matches the spirit of the From trait (lossless owned value to owned value) and we can't add `fn new` method to `[u8; 32]`.

chain/epoch_manager/src/reward_calculator.rs
bowenwang1996: I see. Thanks for pointing this out!

chain/epoch_manager/src/lib.rs
```diff
@@ -1082,6 +1082,24 @@ impl EpochManager {
         let stake_divisor = self.config.minimum_stake_divisor as Balance;
         Ok(self.get_epoch_info(&next_epoch_id)?.seat_price() / stake_divisor)
     }
+
+    // Note: this function should only be used in 18 -> 19 migration and should be removed in the
+    // next release
+    /// `block_header_info` must be the header info of the last block of an epoch.
+    pub fn migrate_18_to_19(
```
bowenwang1996, line 1089: @matklad btw I wonder whether there is any better way to do this. Essentially I need a non-public function in some migration code that only needs to be run once. Current plan is to remove this code later

matklad: I don't think there's any specific great solution here. Though, I'd probably just make `finalize_epoch` public (with a comment), such that I could write migration code in the migration file. I can also imagine some kind of `#[cfg(feature = "migration_api")]`, but that'd be overkill -- we have enough features already :) 

Although there's one thing I don't quite understand. From the reading of it, we have all the migrations, and in theory can migrate 1 database all the way to 19. Does removing this code in the future also mean that we forgo the ability to forward-migrate version 1 database?

neard/src/lib.rs
```diff
@@ -193,6 +193,11 @@ pub fn apply_store_migrations(path: &String, near_config: &NearConfig) {
         // version 17 => 18: add `hash` to `BlockInfo` and ColHeaderHashesByHeight
         migrate_17_to_18(&path);
     }
+    if db_version <= 18 {
+        info!(target: "near", "Migrate DB from version 18 to 19");
+        // version 18 => 19: populate ColEpochValidatorInfo for archival nodes
+        migrate_18_to_19(&path, near_config);
```
matklad, line 199: Hm, something odd jumps at me. Here we migrate 18 to 19, but in the next if we migrate forward from `18`. Naively, I'd expect to see `migrate_18_to_rectify_inflation` and friends first, and `migrate_18_to_19` last. Not sure if it's just unclear naming or a real problem.

chain/epoch_manager/src/lib.rs
bowenwang1996: > Does removing this code in the future also mean that we forgo the ability to forward-migrate version 1 database?

I don't think that is necessarily an issue because there is no need for that. You only need forward compatibility if for example, you expect that you can run the code corresponds to database version 1 and it still works, but that is simply not the case. The protocol version and changed in between and if you run a node with some commit that corresponds to database version 1, it just won't be able to join the network.

neard/src/lib.rs
bowenwang1996: yeah. For nightly feature we need to change the name of the migration function when we stabilize them. Please feel free to suggest better ways to name those functions.

chain/epoch_manager/src/lib.rs
matklad: Yeah, I don't have an opinion if this is an issues or not :)

I was just surprised to see that we seemingly keep ability to migrate from version 1, while discussing removing the ability to migrate from version 18 some time in the future.  

If we dont't need to keep comapt with DB version 1 forever, than the problem indeed is much easier, and temporary function seems to be a perfect solution.

core/chain-configs/src/genesis_config.rs
```diff
@@ -134,6 +138,28 @@ pub struct GenesisConfig {
     pub minimum_stake_divisor: u64,
 }
 
+impl From<&GenesisConfig> for EpochConfig {
+    fn from(config: &GenesisConfig) -> Self {
```
bowenwang1996, line 142: @matklad here I am not sure how to avoid abusing `From`. `EpochConfig` is defined in a dependency of this crate (where `GenesisConfig`) is defined and unless we do some larger refactoring, it seems that there is no easy way to get around this.

neard/src/lib.rs
frol: Nightly are conditional migrations, and at the time of writing we implement a migration path from version X (e.g. 18) to the state that nightly wants, so it seems fine to me to name them `migrate_18_to_nightly_feature`. If it ever gets in conflict, we need to update the migration and the function name, but otherwise, we can keep it and indicate that it was initially implemented on top of db version 18. Not perfect, but it seems to do the trick for us.

Thugh, explicit `db_version <= 18` checks seem weird to me. Should it be `db_version < 10000`?

core/chain-configs/src/genesis_config.rs
frol: Yeah, it is a slight abuse, but I think it is fine (I wonder if this should become idiomatic at some point). Alternatively, you can do:

1. implement `to_epoch_config(&self)` on the `GenesisConfig`
2. implement a new trait like:

```rust
trait NewEpochConfig {
    fn from_genesis_config(&GenesisConfig) -> EpochConfig
}

impl NewEpochConfig for EpochConfig {
    fn from_genesis_config(&GenesisConfig) -> EpochConfig {
       ...
    }
}
```

note: you will need to import this trait where you want to call `from_genesis_config` on `EpochConfig` value.

P.S. we use [easy-ext](https://docs.rs/easy-ext/) to shink the duplication

frol (approved):

core/chain-configs/src/genesis_config.rs
matklad:
```rust
impl GenesisConfig {
    fn [to_]epoch_config(&self) -> EpochConfig
}
```

is how I'd personally write this. I would definitely stay away from using extension traits, they seem to be pure accidental complexity for this use-case.

But, again, it's not that From impl is wrong :)

mikhailOK (approved):

PR #4211, opened 2021-04-09
fix(chain): fix get_block_hash_by_height

mikhailOK:
Fix a bug where the first block in the epoch could get incorrectly
rejected:
process_block replaces the main chain in the ChainStoreUpdate
and calls create_light_client_block, which uses get_header_by_height.
Due to the bug it read directly from ChainStore and could fail.

Test plan
---------
add debug assert, run tests before and after the change


chain/client/src/client_actor.rs
```diff
@@ -989,6 +989,9 @@ impl ClientActor {
                 warn!(target: "client", "receive bad block: {}", err);
             }
             Err(ref err) if err.is_error() => {
+                if let near_chain::ErrorKind::DBNotFoundErr(msg) = err.kind() {
+                    debug_assert!(!msg.starts_with("BLOCK HEIGHT"), "{:?}", err);
```
bowenwang1996, line 993: why is this not possible? `DBNotFoundErr` can originate from many different places

mikhailOK: assert only for "BLOCK HEIGHT" column

bowenwang1996 (commented):
Can we add a test that reproduces the scenario you mentioned?

bowenwang1996 (approved):

PR #4213, opened 2021-04-12
refactor(jsonrpc): Structured errors from other JSON-RPC methods

khorolets:
* `status`
* `health`
* `EXPERIMENTAL_changes_in_block`
* `EXPERIMENTAL_changes`
* `EXPERIMENTAL_light_client_proof`
* `EXPERIMENTAL_validators_ordered`
* `light_client_proof`


chain/client-primitives/src/types.rs
```diff
@@ -339,8 +339,48 @@ pub struct Status {
     pub is_health_check: bool,
 }
 
+#[derive(thiserror::Error, Debug)]
+pub enum StatusError {
+    #[error("IO Error: {error_message}")]
+    IOError { error_message: String },
```
frol, line 345: I think it does not help us much and we may just merge it into the InternalError variant. (I think it is fine to have just IOError, but once we have InternalError on the list, it feels like we have two variants representing the same thing, and IOError does not even record the original error)

chain/jsonrpc/src/lib.rs
```diff
@@ -725,54 +774,50 @@ impl JsonRpcHandler {
...
+                block_hash,
+                state_changes_request: request.state_changes_request,
+            })
+            .await
+            .map(|v| {
+                v.map(|changes| near_jsonrpc_primitives::types::changes::RpcStateChangesResponse {
+                    block_hash: block.header.hash,
+                    changes,
+                })
+            })??)
```
frol, line 820: Those maps were unfortunate since `jsonify` wanted to have `Result<Result<...>>`, but now we can refactor it into:

```suggestion
        let changes = self
            .view_client_addr
            .send(GetStateChanges {
                block_hash,
                state_changes_request: request.state_changes_request,
            })
            .await??;
        Ok(near_jsonrpc_primitives::types::changes::RpcStateChangesResponse {
            block_hash: block.header.hash,
            changes,
        })
```

frol (approved):
Great job! Just a few notes on where we can improve even further.

chain/jsonrpc-primitives/src/types/status.rs
```diff
@@ -0,0 +1,71 @@
...
+    #[serde(flatten)]
+    pub status_response: near_primitives::views::StatusResponse,
+}
+
+#[derive(thiserror::Error, Debug)]
+pub enum RpcStatusError {
+    #[error("Node is syncing")]
+    NodeIsSyncing,
+    #[error("No blocks for {elapsed:?}")]
+    NoBlocks { elapsed: std::time::Duration },
```
frol, line 15: I propose to rename it to `StalledNetwork` or at least `NoNewBlocks`.

chain/jsonrpc-primitives/src/errors.rs
```diff
@@ -75,3 +88,35 @@ impl From<crate::errors::RpcParseError> for RpcError {
...
+impl Display for ServerError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> Result<(), std::fmt::Error> {
+        match self {
+            ServerError::TxExecutionError(e) => write!(f, "ServerError: {}", e),
+            ServerError::Timeout => write!(f, "ServerError: Timeout"),
+            ServerError::Closed => write!(f, "ServerError: Closed"),
+            ServerError::InternalError => write!(f, "ServerError: Internal Error"),
+        }
+    }
+}
```
matklad, line 101:
```suggestion
impl fmt::Display for ServerError {
    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
        match self {
            ServerError::TxExecutionError(e) => write!(f, "ServerError: {}", e),
            ServerError::Timeout => write!(f, "ServerError: Timeout"),
            ServerError::Closed => write!(f, "ServerError: Closed"),
            ServerError::InternalError => write!(f, "ServerError: Internal Error"),
        }
    }
}
```

importing just `use std::fmt;` is the least annoying way to write Display impls manually.

chain/jsonrpc-primitives/src/types/transactions.rs
```diff
@@ -0,0 +1,152 @@
...
+        final_execution_outcome: near_primitives::views::FinalExecutionOutcomeViewEnum,
+    ) -> Self {
+        Self { final_execution_outcome }
+    }
+}
+
+impl From<RpcTransactionError> for crate::errors::RpcError {
+    fn from(error: RpcTransactionError) -> Self {
+        let error_data = match error {
+            RpcTransactionError::InvalidTransaction { ref context } => Some(
```
matklad, line 136:
```suggestion
        let error_data = match &error {
            RpcTransactionError::InvalidTransaction { context } => Some(
```

Seems simpler to me personally.

chain/jsonrpc-primitives/src/types/transactions.rs
```diff
@@ -0,0 +1,152 @@
...
+                serde_json::to_value(crate::errors::ServerError::TxExecutionError(
+                    near_primitives::errors::TxExecutionError::InvalidTxError(context.clone()),
+                ))
+                .unwrap_or(Value::String(error.to_string())),
```
matklad, line 140: Looks like there's some leftover code here?


```suggestion
            RpcTransactionError::InvalidTransaction { ref context } =>
                serde_json::to_value(crate::errors::ServerError::TxExecutionError(
                    near_primitives::errors::TxExecutionError::InvalidTxError(context.clone()),
                )
```

Or should this be `.ok()` rather then `Some()`?

chain/jsonrpc/src/lib.rs
```diff
@@ -342,8 +356,18 @@ impl JsonRpcHandler {
                 let query_response = self.query(rpc_query_request).await;
                 process_query_response(query_response)
             }
-            "status" => self.status().await,
-            "tx" => self.tx_status_common(request.params, false).await,
+            "status" => {
+                let status_response = self.status().await?;
+                serde_json::to_value(status_response)
+                    .map_err(|err| RpcError::parse_error(err.to_string()))
```
matklad, line 370: Hm, why is this pares error rather than serialization error / Internal error? From the look of it, seems like serialization should return internal error or maybe just `.unwrap` even, and rely on the framework to catch panics and 500? Assuming that serialization failure indicates a bug in our code.

khorolets:
@frol I've created `RpcError::serialization_error` method which created "Server error" and replaced usages of `parse_error` in handlers. Could you have a look and say about client's `call_method` method, should I replace usages of `parse_error` there as well?

@matklad appreciate your suggestions!

frol (approved):
Yay! :tada: 

We did it! (For those who follows this: this is the last PR in the series of internal JSON RPC error handling refactoring, and the next step would be to expose those errors through JSON RPC with exhaustive and easy-to-reason error kinds)

frol:
@bowenwang1996 your review is required, FYI

bowenwang1996 (commented):
Please run nayduck to make sure that there is no regression

khorolets:
@bowenwang1996 does it look ok http://nayduck.eastus.cloudapp.azure.com:3000/#/run/1521 ?

bowenwang1996 (approved):

PR #4216, opened 2021-04-13
(fix) Inconsistent storage state if node version has account versioning enabled and chain version does not

EgorKulikov:
Deleted test will be reintroduced with V2

Due to concerns about patching genesis versioning switched from enum of structs to struct with separate field for enum

@chefsale please note that for betanet scripts/migrations/107-2-revert-add-account-versioning.py should be applied to state dump/new genesis


core/primitives-core/src/account.rs
```diff
@@ -39,7 +43,7 @@ pub struct AccountV1 {
 }
 
 #[cfg(feature = "protocol_feature_add_account_versions")]
-#[derive(BorshSerialize, Serialize, Deserialize, PartialEq, Eq, Debug, Clone)]
+#[derive(Serialize, Deserialize, PartialEq, Eq, Debug, Clone)]
```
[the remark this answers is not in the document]
EgorKulikov: As I understand it Borsch is used for state, serde for genesis and dump.

core/primitives-core/src/account.rs
```diff
@@ -194,14 +198,16 @@ impl BorshDeserialize for Account {
...
-                DeserializableAccount::AccountV1(account) => Ok(Account::AccountV1(account)),
-            }
+            unreachable!();
+        }
+    }
+}
+
+#[cfg(feature = "protocol_feature_add_account_versions")]
+impl BorshSerialize for Account {
+    fn serialize<W: Write>(&self, writer: &mut W) -> io::Result<()> {
```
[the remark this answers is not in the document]
EgorKulikov: Done

core/primitives-core/src/account.rs
```diff
@@ -188,20 +188,22 @@ impl Account {
 
 #[cfg(feature = "protocol_feature_add_account_versions")]
 impl BorshDeserialize for Account {
-    fn deserialize(buf: &mut &[u8]) -> Result<Self, Error> {
+    fn deserialize(buf: &mut &[u8]) -> Result<Self, io::Error> {
```
bowenwang1996, line 191: With this change, it seems that we don't need to gate this implementation with a feature flag.

core/primitives-core/src/account.rs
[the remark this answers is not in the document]
bowenwang1996: @matklad yes borsh serialization is what this important here because if we don't do it correctly the state will diverge.

bowenwang1996 (approved):
Please take a look at my comment (if it makes sense)

core/primitives-core/src/account.rs
EgorKulikov: Without feature Account is struct and have derived BorshDeserialize

EgorKulikov:
http://nayduck.eastus.cloudapp.azure.com:3000/#/run/1497

chefsale:
Just on thing which is not clear to me: 
Do we need to bump the version here to trigger the hardfork?

chefsale:
Ah, realized we will have to do the manual hardfork due to migration script. Please let me know before you plan to land it, as the automated betanet release is at 1PM London time, we should ideally land this afterwards and manually hardfork and release again. Maybe let's land this on Monday?

core/primitives-core/src/account.rs
```diff
@@ -5,27 +5,30 @@ use crate::hash::CryptoHash;
...
+    fn default() -> Self {
+        AccountVersion::V1
+    }
+}
+
+/// Per account information stored in the state.
+#[derive(Serialize, Deserialize, PartialEq, Eq, Debug, Clone)]
+#[cfg_attr(
+    not(feature = "protocol_feature_add_account_versions"),
+    derive(BorshSerialize, BorshDeserialize)
```
bowenwang1996, line 29: probably better to remove derive altogether to prevent accidental change to state

EgorKulikov: Can you elaborate? Derive only kicks in if feature is not enabled

EgorKulikov: Also this is covered by tests

chefsale (commented):
lgtm

chefsale (approved):

PR #4218, opened 2021-04-13
fix: make param estimator runnable without cargo

matklad:
env! macro, unlike std::env::var function, works at compile time, so the
resulting binary will be able to find test contracts even when run
without cargo.

# Test Plan

Should fix CI for params estimator, which is currently broken.


olonho (approved):
it's clear improvement, but does it actually fixes param estimator so that it could run and produce sensible results?

willemneal (approved):

matklad:
>it's clear improvement, but does it actually fixes param estimator so that it could run and produce sensible results?

Not sure, we'll see. I've subscribe to param estimator build on CI, so I'll make sure that it at least won't fail in the future.

PR #4220, opened 2021-04-13
refactor: remove unnecessary indirection in CryptoHash

matklad:
Archeological notes:

1. We started with `pub type CryptoHash = Digest;` where `Digest` was a
   type form `exonum_sodiumoxide`

   https://github.com/near/nearcore/blame/fff46c2413d7371c0d4b155d6533126d4aa0ffab/core/primitives/src/hash.rs

2. We changed that to `struct CryptoHash(Digest)` to enable Default in

   https://github.com/near/nearcore/commit/0aef66679616398c07ee2230de7345287c2751ea

3. And then we replace Digest use with `struct Digest([u8; 32])` in

   https://github.com/near/nearcore/commit/a50cbc21b023ef82d55236bed6139404d83b079e


matklad:
Will finish this up after #4205 , to avoid conflicts

frol (approved):

matklad:
Ready for review. Added a second commint that removes 

```
impl<'a> From<&'a CryptoHash> for String {
    fn from(h: &'a CryptoHash) -> Self {
        to_base(&h.0)
    }
}
```

impl: its unnecessary, unidiomatic, and actually makes call sites more awkward in comparison to just using `ToString`.

core/primitives-core/src/hash.rs
```diff
@@ -7,50 +7,29 @@ use serde::{Deserialize, Deserializer, Serialize, Serializer};
...
+#[derive(Copy, Clone, PartialEq, Eq, PartialOrd, Ord, derive_more::AsRef, derive_more::AsMut)]
 #[as_ref(forward)]
-pub struct Digest(pub [u8; 32]);
-
-#[derive(Copy, Clone, PartialOrd, Ord, derive_more::AsRef)]
-#[as_ref(forward)]
-pub struct CryptoHash(pub Digest);
-
-impl CryptoHash {
-    pub fn raw_bytes(&self) -> &[u8; 32] {
```
matklad, line 19: `.raw_bytes()` is not spelled as `.0`.

ailisp (approved):
Looks good!

ailisp:
Looks like our PR failed at this flaky timeout quite often recently: sync_empty_state

bowenwang1996 (approved):

PR #4221, opened 2021-04-13
feat(rpc): expose epoch height for local reconstruct vmcontext

ailisp:
reproduce contract call with vm standalone requires a vmcontext, which takes epoch height instead of epoch id. This only introduce new field in view, should not cause any protocol change or breaking existing code use `validator` rpc

Test Plan
------------
CI


matklad (approved):

ailisp:
@bowenwang1996 @frol please take a look

chain/chain/src/test_utils.rs
```diff
@@ -1026,6 +1026,7 @@ impl RuntimeAdapter for KeyValueRuntime {
             current_proposals: vec![],
             prev_epoch_kickout: vec![],
             epoch_start_height: 0,
+            epoch_height: 1,
```
frol, line 1029: It is an unfortunate naming collision here (`epoch_start_height` is a block height)

core/primitives/src/views.rs
```diff
@@ -1324,8 +1324,10 @@ pub struct EpochValidatorInfo {
     pub current_proposals: Vec<ValidatorStakeView>,
     /// Kickout in the previous epoch
     pub prev_epoch_kickout: Vec<ValidatorKickoutView>,
-    /// Epoch start height
+    /// Epoch start block height
     pub epoch_start_height: BlockHeight,
+    /// Epoch height
+    pub epoch_height: EpochHeight,
```
frol, line 1330: I would not say that `validators` endpoint is the best place to expose this value, but I don't have a better place in mind either.

frol (approved):

core/primitives/src/views.rs
ailisp: Same to me, expose in block api there's significantly more change (Block and BlockView stores nothing about epoch info, and have to refer back to RuntimeAdapter, but RuntimeAdaptor is in neard). Another more proper option maybe add a EXPERIMENTAL_epoch_info and in future supersede validators since epoch_info also covers everything validators in epoch, plus specific epoch information

bowenwang1996 (approved):

PR #4224, opened 2021-04-14
refactor: improve error reporting in the build script

matklad:
Before, the error looked like

  No such file or directory (os error 2)

Now, it is

  command `“rustup” “target” “add” “wasm32-unknown-unknown”` failed to run: No such file or directory (os error 2)

This helps to debug issues like rustup not being present on the system.

# Test Plan

The code in question is a part of an internal test suite, will be exercised by the existing CI

### DIFF
--- a/runtime/near-test-contracts/build.rs
+++ b/runtime/near-test-contracts/build.rs
@@ -9,11 +9,9 @@ fn main() {
 }
 
 fn try_main() -> io::Result<()> {
-    let status =
-        Command::new("rustup").args(&["target", "add", "wasm32-unknown-unknown"]).status()?;
-    if !status.success() {
-        return Err(io::Error::new(io::ErrorKind::Other, "rustup target add failed"));
-    }
+    let mut cmd = Command::new("rustup");
+    cmd.args(&["target", "add", "wasm32-unknown-unknown"]);
+    check_status(cmd)?;
 
     build_contract("./test-contract-rs", &[], "test_contract_rs")?;
     build_contract(
@@ -29,10 +27,8 @@ fn build_contract(dir: &str, args: &[&str], output: &str) -> io::Result<()> {
     let mut cmd = cargo_build_cmd();
     cmd.args(args);
     cmd.current_dir(dir);
-    let status = cmd.status()?;
-    if !status.success() {
-        return Err(io::Error::new(io::ErrorKind::Other, "cargo build failed"));
-    }
+    check_status(cmd)?;
+
     fs::copy(
         format!("./{}/target/wasm32-unknown-unknown/release/{}.wasm", dir, dir.replace('-', "_")),
         format!("./res/{}.wasm", output),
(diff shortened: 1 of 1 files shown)


matklad:
cc @Longarithm

runtime/near-test-contracts/build.rs
```diff
@@ -48,3 +44,16 @@ fn cargo_build_cmd() -> Command {
...
 }
+
+fn check_status(mut cmd: Command) -> io::Result<()> {
+    let status = cmd.status().map_err(|err| {
+        io::Error::new(io::ErrorKind::Other, format!("command `{:?}` failed to run: {}", cmd, err))
+    })?;
+    if !status.success() {
+        return Err(io::Error::new(
+            io::ErrorKind::Other,
+            format!("command `{:?}` exited with non-zero status: {:?}", cmd, status),
```
ailisp, line 55: Directly output vector looks ugly, and user cannot copy and paste to shell to rerun & debug. Use cmd.join(" ") or use a more correct `shlex.join`: https://docs.rs/shlex/1.0.0/shlex/fn.join.html?

ailisp (approved):

runtime/near-test-contracts/build.rs
matklad: Agree that it's a bit ugly, but it is ugly specifically such that copy&paste works even if there are things spaces in the command name. 

```
λ git diff
diff --git a/runtime/near-test-contracts/build.rs b/runtime/near-test-contracts/build.rs
index 116c61cfb..78f966b36 100644
--- a/runtime/near-test-contracts/build.rs
+++ b/runtime/near-test-contracts/build.rs
@@ -10,7 +10,7 @@ fn main() {
 
 fn try_main() -> io::Result<()> {
     let mut cmd = Command::new("rustup");
-    cmd.args(&["target", "add", "wasm32-unknown-unknown"]);
+    cmd.args(&["target", "addd", "wasm32-unknown-unknown"]);
     check_status(cmd)?;
 
     build_contract("./test-contract-rs", &[], "test_contract_rs")?;

20:13:46|~/projects/nearcore|better-build-errors⚡*
λ cargo build -q -p near-test-contracts
error: failed to run custom build command for `near-test-contracts v0.0.0 (/home/matklad/projects/nearcore/runtime/near-test-contracts)`

Caused by:
  process didn't exit successfully: `/home/matklad/projects/nearcore/target/debug/build/near-test-contracts-f8128a735147dfcb/build-script-build` (exit code: 1)
  --- stdout
  error: The subcommand 'addd' wasn't recognized
  	Did you mean 'add'?

  If you believe you received this message in error, try re-running with 'rustup target -- addd'

  USAGE:
      rustup target <SUBCOMMAND>

  For more information try --help

  --- stderr
  command `"rustup" "target" "addd" "wasm32-unknown-unknown"` exited with non-zero status: ExitStatus(ExitStatus(256))

20:13:48|~/projects/nearcore|better-build-errors⚡*
λ "rustup" "target" "addd" "wasm32-unknown-unknown"
error: The subcommand 'addd' wasn't recognized
	Did you mean 'add'?

If you believe you received this message in error, try re-running with 'rustup target -- addd'

USAGE:
    rustup target <SUBCOMMAND>

For more information try --help
```

I don't think it's worth pulling an extra dependency here. JFYI, a more general solution for "I want to run a bunch of shell commands in rust and get reasonable errors bace" would be https://github.com/matklad/xshell


A proper solution here would be to implement `Display` for `std::process::Command`, thats on my todo list: https://github.com/rust-lang/rust/issues/72009.

frol (approved):

PR #4225, opened 2021-04-14
refactor: remove knowledge about particular VM from vm-logic

matklad:
As near-vm-logic docs say

> It is not dependent on the specific way the smart contract code is executed, e.g. through Wasmer or whatnot

So VMKind really does not belong here. It's a private impl detail of
vm-runner, and it's the runner job to select the specific right VM.


matklad:
Backstory: started writing high-level docs for the vm-runner, looked at the vm-logic docs, noticed that this type is in the wrong crate.

willemneal (approved):
LGTM!  Makes much more sense.

ailisp (approved):
Makes sense

olonho (approved):

EgorKulikov (approved):

PR #4226, opened 2021-04-15
[DO NOT MERGE] Fix apply chunks

bowenwang1996:
Submit a PR so that it is easier to see what has changed


runtime/runtime/src/lib.rs
```diff
@@ -1120,6 +1125,23 @@ impl Runtime {
                 &mut stats,
             )?;
         }
+        if !apply_state.is_new_chunk
+            && apply_state.current_protocol_version
+                >= PROTOCOL_FEATURES_TO_VERSION_MAPPING[&ProtocolFeature::FixApplyChunks]
+        {
```
frol, line 1131: Should we use `checked_feature!`?

frol (approved):
I only reviewed it on a high level. LGTM.

runtime/runtime/src/lib.rs
bowenwang1996: checked_feature does not work when there is no associated rust feature

frol:
Already released as 1.18.3: https://github.com/near/nearcore/compare/1.18.2...1.18.3

PR #4229, opened 2021-04-16
refactor(neard/tests): reduce boilerplate with intuitive structures

miraclx:
- **reduce boilerplate with intuitive structures**: barely changing patterns repeated over and over in `neard/tests/*` are simplified with easy to reason about structures.
- **run_actix_until** (near-actix-test-utils): cancel all active async tasks on SIGINT (<kbd>ctrl</kbd>+<kbd>c</kbd>) and panic pending ones.


khorolets (commented):
Overall looks goods!

neard/tests/run_nodes.rs
```diff
@@ -17,22 +19,24 @@ fn run_nodes(
...
-                    .unwrap()
-            })
-            .collect::<Vec<_>>();
-        let (_, _, clients) =
-            start_nodes(num_shards, &dirs, num_validators, 0, epoch_length, genesis_height);
+
+    let cluster = NodeCluster::new()
+        .mkdirs_with(num_nodes as usize, |index| {
+            format!("run_nodes_{}_{}_{}", num_nodes, num_validators, index)
+        })
```
frol, line 26: Let's just merge `mkdirs_with` into `new`. It does not make any sense to have a cluster without nodes at all.

```suggestion
    let cluster = NodeCluster::new(num_nodes as usize, |index| {
            format!("run_nodes_{}_{}_{}", num_nodes, num_validators, index)
        })
```

neard/tests/node_cluster.rs
```diff
@@ -0,0 +1,106 @@
+use futures::future;
+
+use near_actix_test_utils::{run_actix_until_stop, spawn_interruptible as spawn};
```
frol, line 3: Let's not hide the fact that we use `spawn_interruptible` helper. ("Explicit is better than implicit", by the Zen of Python)

```suggestion
use near_actix_test_utils::{run_actix_until_stop, spawn_interruptible};
```

frol (approved):
@miraclx Great job! I have a few ideas in the comments

neard/tests/run_nodes.rs
miraclx: 😅, you're totally right. I'm on it.

neard/tests/rpc_nodes.rs
```diff
@@ -42,86 +45,85 @@ macro_rules! panic_on_rpc_error {
...
+        .mkdirs_with(4, |index| format!("tx_propagation{}", index))
+        .with(HeavyTest(true))
+        .with(Shards(4))
+        .with(ValidatorSeats(2))
+        .with(LightClients(2))
+        .with(EpochLength(10))
+        .with(GenesisHeight(0));
+
+    cluster.exec_until_stop(|genesis, rpc_addrs, clients| async move {
+        let view_client = clients[0].1.clone();
```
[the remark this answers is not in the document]
miraclx: 🤔I'll look into it

neard/tests/node_cluster.rs
```diff
@@ -0,0 +1,106 @@
...
+        self
+    }
+
+    pub fn mkdirs_with<F: Fn(usize) -> String>(mut self, capacity: usize, gen_dirname: F) -> Self {
+        self.dirs = Vec::with_capacity(capacity);
+        self.dirs.extend(
+            (0..capacity).map(|index| {
+                tempfile::Builder::new().prefix(&gen_dirname(index)).tempdir().unwrap()
+            }),
+        );
```
matklad, line 53:
```suggestion
        self.dirs = 
            (0..capacity).map(|index| {
                tempfile::Builder::new().prefix(&gen_dirname(index)).tempdir().unwrap()
            }).collect();
```

neard/tests/node_cluster.rs
```diff
@@ -0,0 +1,106 @@
...
+    pub fn with(mut self, config: ClusterConfigVariant) -> Self {
+        match config {
+            HeavyTest(is_heavy) => self.is_heavy = is_heavy,
+            Shards(n) => self.num_shards = Some(n),
+            ValidatorSeats(n) => self.num_validator_seats = Some(n),
+            LightClients(n) => self.num_lightclient = Some(n),
+            EpochLength(l) => self.epoch_length = Some(l),
+            GenesisHeight(h) => self.genesis_height = Some(h),
+        };
+        self
```
matklad, line 44: As an alternative, I think we can avoid defining `ClusterConfigVariant` struct and just create a bunch of `with` functions. Either variant works, but I'd personally go with a bunch of functions, as that feels less fancy and doesn't require an extra import.

neard/tests/node_cluster.rs
```diff
@@ -0,0 +1,106 @@
...
+                "cluster config: expected a non-zero number of directories"
+            );
+            let (genesis, rpc_addrs, clients) = start_nodes(
+                self.num_shards.expect("cluster config: [num_shards] undefined"),
+                &self.dirs,
+                self.num_validator_seats.expect("cluster config: [num_validator_seats] undefined"),
+                self.num_lightclient.expect("cluster config: [num_lightclient] undefined"),
+                self.epoch_length.expect("cluster config: [epoch_length] undefined"),
+                self.genesis_height.expect("cluster config: [genesis_height] undefined"),
+            );
```
matklad, line 82: Nit: I'd pull `asserts` and `.expects` out of `async` block, so that the panic happens immediatly when `exec` is called, rather then when the future is polled.

neard/tests/node_cluster.rs
```diff
@@ -0,0 +1,106 @@
...
+                tempfile::Builder::new().prefix(&gen_dirname(index)).tempdir().unwrap()
+            }),
+        );
+        self
+    }
+
+    fn exec<F, R>(self, f: F)
+    where
+        R: future::Future<Output = ()> + 'static,
+        F: Fn(
```
matklad, line 60: Could be an `FnOnce` is seems?

matklad (commented):
:+1: on turning setup *code* into setup *data*!

neard/tests/node_cluster.rs
miraclx: Oh right, `Vec` get's the `size_hint` through `Map`'s proxy on `Range`, so it's reserved to capacity.

neard/tests/node_cluster.rs
miraclx: @frol, what do you think?

<table>
<tr>
<td>

``` rust
cluster
    .with(HeavyTest(true))
    .with(Shards(1))
    .with(ValidatorSeats(1))
    .with(LightClients(0))
    .with(EpochLength(10))
    .with(GenesisHeight(0));
```
</td>
<td>

``` rust
cluster
    .heavy_test(true)
    .shards(1)
    .validator_seats(1)
    .light_clients(1)
    .epoch_length(10)
    .genesis_height(0);
```
</td>
</tr>
<tr>
<td> Current </td> <td> Proposed </td>
</tr>
</table>

Or are they to be prefixed with `set_` e.g `.set_genesis_height(0)`

frol: Yeah, I would go with the proposed alt variant. `is_heavy_test(true)` could be just `set_heavy_test()` (don't need to even specify the argument)

neard/tests/node_cluster.rs
```diff
@@ -0,0 +1,104 @@
...
+                    tempfile::Builder::new().prefix(&gen_dirname(index)).tempdir().unwrap()
+                })
+                .collect(),
+            ..Default::default()
+        }
+    }
+
+    add_mut_helpers! {
+        set_heavy_test(self) => self.is_heavy = true,
+        set_shards(self, n: NumShards) => self.num_shards = Some(n),
```
frol, line 44: set_shards is too broad naming. Let’s go with set_shards_number.

I also propose to rename set_validator_seats_count

neard/tests/node_cluster.rs
```diff
@@ -0,0 +1,104 @@
...
+            dirs: (0..capacity)
+                .map(|index| {
+                    tempfile::Builder::new().prefix(&gen_dirname(index)).tempdir().unwrap()
+                })
+                .collect(),
+            ..Default::default()
+        }
+    }
+
+    add_mut_helpers! {
```
frol, line 42: We tend to avoid macroses if possible since it creates burden of “learning” the macro implementation. In this case, the functions are small, and I believe that even the total lines count could be the same without the macro while the complexity will be much lower, so I argue that it does not buy us anything here. Also, we cannot properly document the methods

neard/tests/node_cluster.rs
```diff
@@ -0,0 +1,104 @@
...
+        set_heavy_test(self) => self.is_heavy = true,
+        set_shards(self, n: NumShards) => self.num_shards = Some(n),
+        set_validator_seats(self, n: NumSeats) => self.num_validator_seats = Some(n),
+        set_lightclient(self, n: usize) => self.num_lightclient = Some(n),
```
frol, line 46: Looking into testlib run_nodes implementation: https://github.com/near/nearcore/blob/92be6b559237530123851dcf829a3b9c7de5a281/test-utils/testlib/src/lib.rs#L63

I have two ideas:

* rename this method to set_lightclient_nodes_count
* remove this method and require users to pass “tracking_nodes_count” and “lightclient_nodes_count” in `NodeCluster:new` instead of “capacity”

neard/tests/node_cluster.rs
```diff
@@ -0,0 +1,104 @@
...
+            pub fn $method(mut $self$(, $arg: $type)?) -> Self {
+                $expr;
+                $self
+            }
+        )+
+    }
+}
+
+impl NodeCluster {
+    pub fn new<F: Fn(usize) -> String>(capacity: usize, gen_dirname: F) -> Self {
```
frol, line 31: Let’s go with `nodes_count` instead of capacity unless we go with the plan in another comment about lightclint_nodes

neard/tests/node_cluster.rs
```diff
@@ -0,0 +1,104 @@
...
+                .map(|index| {
+                    tempfile::Builder::new().prefix(&gen_dirname(index)).tempdir().unwrap()
+                })
+                .collect(),
+            ..Default::default()
+        }
+    }
+
+    add_mut_helpers! {
+        set_heavy_test(self) => self.is_heavy = true,
```
frol, line 43: Actually, do we ever use cluster in non-heavy-test way? Maybe it is redundant?

miraclx: Not right now, but unless every test that uses a NodeCluster is always considered a heavy test, I thought to leave the option open.

neard/tests/node_cluster.rs
miraclx: here's a couple of probable names, ticked are the ones I think I prefer, what do you think?

- ∞
  - [X] `set_shard_count`
  - [ ] `set_num_shards`
  - [ ] `set_shards_number`
- ∞
  - [X] `set_validator_seat_count`
  - [ ] `set_validator_seats_count`
  - [ ] `set_num_validator_seats`

neard/tests/node_cluster.rs
miraclx: On documentation, I figured it was an internal implementation detail and with reasonable names, we wouldn't really need those, but I get it

> and I believe that even the total lines count could be the same

you're ~right, only +10 LoC

neard/tests/node_cluster.rs
miraclx: I'll go with the first. The second, I think, adds extra verbosity that's not that necessary.

neard/tests/node_cluster.rs
miraclx: I think I'll go with this instead of the other option suggested in the comment prior.

neard/tests/node_cluster.rs
miraclx: Then again, `set_num_` prefixed names seems to be the most intuitive.

frol (approved):

bowenwang1996 (approved):

neard/tests/rpc_nodes.rs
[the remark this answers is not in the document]
miraclx: After thinking about this for a while, I think it's best to leave it as-is.

PR #4232, opened 2021-04-16
chore: upgrade wabt dev dependency

matklad:
We use wabt in test to convert textual representation of Wasm to binary.
wabt is implemented in C++ and, in 0.9, has a perplexing bug where
syntax error in a test wasm file would cause Rust to die with a C++
exception and a scary looking message. 0.10 fixes that.

While upgrading wabt, I've noticed that we instruct `git diff` to ignore
changes in Cargo.lock:

    16:59:19|~/projects/nearcore|master⚡*
    λ git status --short --branch
    ## master...upstream/master
     M Cargo.lock
     M runtime/near-vm-runner/Cargo.toml

    16:59:21|~/projects/nearcore|master⚡*
    λ git diff
    index 857032bdd..90a0dbe71 100644
    Binary files a/Cargo.lock and b/Cargo.lock differ
    diff --git a/runtime/near-vm-runner/Cargo.toml b/runtime/near-vm-runner/Cargo.toml
    index 030bc37f2..ffa4ac3f0 100644
    --- a/runtime/near-vm-runner/Cargo.toml
    +++ b/runtime/near-vm-runner/Cargo.toml
    @@ -39,7 +39,7 @@ threadpool = "1.8.1"
     [dev-dependencies]
     near-test-contracts = { path = "../near-test-contracts" }
     assert_matches = "1.3"
    -wabt = "0.9"
    +wabt = "0.10"
     bencher = "^0.1.5"
     base64 = "0.13"

I think that's not helpful: when updating dependencies, it is useful to
inspect what exactly get changed in Cargo.lock. Otherwise, for all I
know wabt 0.10 can bring a host of shady new dependencies.

So this commit removes `-diff` (and `liguist-generated=true` is default
for Cargo.lock anyway).

(Note that Cargo.lock files became more diff-friendly some time ago)

# Testing Plan

This affects only existing tests anyway


ailisp:
@chefsale Looks like failed due to missing new cmake. amazon linux used a very old version of cmake 2.8, I installed a cmake 3.10 or 3.13 in /usr/local/bin/cmake in buildkite image before, but don't know why it is not in path now. Could you please check and update buildkite image?

bowenwang1996:
@chefsale ping

matklad:
I think the root problem here is using C++ code, so let's close this in favor of #4471

PR #4236, opened 2021-04-16
fix(logging): improve view function call logging

bowenwang1996:
Currently we log `VMOutcome` with the derived `Debug` implementation, which may log too much data and slow down neard. This PR bounds the amount of data we log for `VMOutcome` and also removes an unnecessary `clone`.

Test plan
---------
CI


EgorKulikov (approved):

runtime/near-vm-logic/src/logic.rs
```diff
@@ -2430,3 +2430,18 @@ pub struct VMOutcome {
     pub used_gas: Gas,
     pub logs: Vec<String>,
 }
+
+impl std::fmt::Display for VMOutcome {
```
frol, line 2434: I would say it is an abuse of `Display`. The information is useful in logging. Let's consider IPv4 address as an example; Display makes perfect sense there since its internal implementation is a 4-byte integer which can be displayed to humans as `127.0.0.1`. Here, I argue that "VMOutcome: balance ..." is not a display of the VMOutcome, there is no meaningful Display for this structure, so I would avoid implementing it. Instead, I would implement a simple `.format_log_message()` / `.display_log_message()` / `.to_log_message()` / `.fmt_log_message()` (naming is hard, you know). We also may just as well replace the derived `Debug` implementation with this custom one.

frol (approved):

runtime/near-vm-logic/src/logic.rs
matklad: Yeah, agree that there's no meaningful Display possible here. Hand-writing debug is what I'd do personally. Here's an example of compressed debug from rust-analyzer: 

https://github.com/rust-analyzer/rust-analyzer/blob/fc709c8b210c551c97ca3b7c7daeaabc15cd80aa/crates/project_model/src/workspace.rs#L53-L79

If we want do be fancy (we don't), we can even write it such that `{:?}` prints compressed representation and `{:#?}` prints the full one. 

I would include the length of `Value` into the debug repr though.

runtime/runtime/src/state_viewer/mod.rs
```diff
@@ -264,12 +264,13 @@ impl TrieViewer {
...
-            let mut result = vec![];
-            if let ReturnData::Value(buf) = &outcome.return_data {
-                result = buf.clone();
-            }
+            debug!(target: "runtime", "(exec time {}) result of execution: {}", time_str, outcome);
+            let VMOutcome { logs: outcome_logs, return_data, .. } = outcome;
+            logs.extend(outcome_logs);
+            let result = match return_data {
+                ReturnData::Value(buf) => buf,
+                _ => vec![],
```
matklad, line 272: Nit, but I'd write this explicitly as 

```suggestion
                ReturnData::ReceiptIndex(_) | ReturnData::None => vec![],
```

as per this classical lovely interlaced advice: https://youtu.be/-J8YyfrSwTk?t=1819

runtime/runtime/src/state_viewer/mod.rs
```diff
@@ -264,12 +264,13 @@ impl TrieViewer {
...
-            debug!(target: "runtime", "(exec time {}) result of execution: {:#?}", time_str, outcome);
-            logs.extend(outcome.logs);
-            let mut result = vec![];
-            if let ReturnData::Value(buf) = &outcome.return_data {
-                result = buf.clone();
-            }
+            debug!(target: "runtime", "(exec time {}) result of execution: {}", time_str, outcome);
+            let VMOutcome { logs: outcome_logs, return_data, .. } = outcome;
+            logs.extend(outcome_logs);
+            let result = match return_data {
```
matklad, line 270:
```suggestion
            logs.extend(outcome.logs);
            let result = match outcome.return_data {
```

Seems like there's no need to destruct here?

bowenwang1996: interesting. I didn't realize that the borrow checker is smart enough to understand that :)

matklad: Yeah, borrow checker understands partial moves and partial borrows. What it doesn't understand is partial closure captures:

```rust
let outcome = ...
let closure = || outcome.logs.drain(...);
// The whole outcome is borrowed here, you can't use `.return_data`
```

Though, even this will be fixed later next year.

PR #4239, opened 2021-04-19
Fix for account revert script

EgorKulikov:
Previous version deletes account_id by mistake


chefsale (approved):

chefsale:
Works with this change, thanks.

PR #4245, opened 2021-04-21
refactor(indexer): Update Indexer Framework 0.9.0 (IndexerShard)

khorolets:
* Add IndexerShard structure
* Bump version to 0.9.0
* update CHANGELOG

Also I've updated `example-indexer` to keep it up to date and changed the commented example of `StreamerMessage` it was a bit outdated.

### DIFF
--- a/chain/indexer/CHANGELOG.md
+++ b/chain/indexer/CHANGELOG.md
@@ -1,5 +1,22 @@
 # Changelog
 
+## 0.9.0
+
+* Introduce `IndexerShard` structure which contains corresponding chunks and `IndexerExecutionOutcomeWithReceipt`
+
+## Breaking change
+
+* `IndexerExecutionOutcomeWithReceipt`field `receipt` is not optional anymore and isn't included to chunk
+* `IndexerChunkView` doesn't contain field `receipt_execution_outcomes` anymore, this field has been moved to `IndexerShard`
+* `StreamerMessage` structure now looks like:
+  ```
+  StreamerMessage {
+    block: BlockView,
+    shards: Vec<IndexerShard>,
+    state_changes: StateChangesView,  
+  }
+  ```
+
 ## 0.8.1
 
 * Add `InitConfigArgs` and `indexer_init_configs`
--- a/chain/indexer/CHANGELOG.md
+++ b/chain/indexer/CHANGELOG.md
@@ -3,15 +3,15 @@
 ## 0.9.0
 
 * Introduce `IndexerShard` structure which contains corresponding chunks and `IndexerExecutionOutcomeWithReceipt`
-* `IndexerExecutionOutcomeWithReceipt` field `receipt` is always required
+* `receipt` field in `IndexerExecutionOutcomeWithReceipt` is no longer optional as it used to be always set anyway, 
+so now we explicitly communicate this relation ("every outcome has a corresponding receipt") through the type system
 * Introduce `IndexerExecutionOutcomeWithOptionalReceipt` which is the same as `IndexerExecutionOutcomeWithReceipt`
 but with optional `receipt` field.
 
-## Breaking change
+## Breaking changes
 
-* `IndexerExecutionOutcomeWithReceipt`field `receipt` is not optional anymore and isn't included to chunk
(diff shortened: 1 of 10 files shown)


chain/indexer/src/streamer/mod.rs
```diff
@@ -85,13 +100,11 @@ async fn build_streamer_message(
         )
         .await?;
 
-        local_receipts.extend_from_slice(&chunk_local_receipts);
-
-        let mut chunk_receipts = chunk_local_receipts;
+        let mut chunk_receipts = chunk_local_receipts.clone();
```
frol, line 103: I suggest we re-order this `chunk_receipts` with the loop below, so you don't need to clone `chunk_local_receipts` (you just loop over `&chunk_local_receipts` as before.

chain/indexer/src/streamer/types.rs
```diff
@@ -17,17 +17,40 @@ pub struct IndexerChunkView {
...
+impl From<IndexerExecutionOutcomeWithOptionalReceipt> for IndexerExecutionOutcomeWithReceipt {
+    fn from(outcome_with_optional_receipt: IndexerExecutionOutcomeWithOptionalReceipt) -> Self {
+        Self {
+            execution_outcome: outcome_with_optional_receipt.execution_outcome,
+            receipt: outcome_with_optional_receipt
+                .receipt
+                .expect("`receipt` must be present at this moment"),
+        }
+    }
+}
```
frol, line 56: I am not 100% against `From` implementation here, but it might be easier to reason about if we in fact inline this transformation where we need it, since "receipt must be present at **this** moment" is not super accurate without the context of "this moment" given that `From` does not imply that information.

frol (approved):
I have a few nits, but overall, it is going to work just fine without any changes.

chain/indexer/CHANGELOG.md
```diff
@@ -1,5 +1,25 @@
 # Changelog
 
+## 0.9.0
+
+* Introduce `IndexerShard` structure which contains corresponding chunks and `IndexerExecutionOutcomeWithReceipt`
+* `IndexerExecutionOutcomeWithReceipt` field `receipt` is always required
+* Introduce `IndexerExecutionOutcomeWithOptionalReceipt` which is the same as `IndexerExecutionOutcomeWithReceipt`
+but with optional `receipt` field.
```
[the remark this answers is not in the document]
khorolets: We expose it. Because users may want to use it while working with `IndexerTransactionWithOutcome`. `outcome` field there has type `IndexerExecutionOutcomeWithOptionalReceipt` so I would prefer to have this structure exposed.

chain/indexer/src/streamer/types.rs
khorolets: Agree, I had some doubts, you've just approved my theory about implicit way

frol (approved):

bowenwang1996 (commented):
So it looks like this won't work before we restore the receipts that are generated due to the bug?

frol:
@bowenwang1996 Why is so? If you refer to the outcomes requiring receipts, we don't have the outcomes since we don't have the receipts, so we are "good"

bowenwang1996:
> @bowenwang1996 Why is so? If you refer to the outcomes requiring receipts, we don't have the outcomes since we don't have the receipts, so we are "good"

Right. That works fine after the fix, but for the two blocks where the issue occurred it would still be a problem

chain/indexer/src/streamer/mod.rs
```diff
@@ -50,21 +52,33 @@ async fn build_streamer_message(
...
             .remove(&header.shard_id)
             .expect("Execution outcomes for given shard should be present");
+
+        // Take execution outcomes for receipts from the vec and keep only the ones for transactions
         let mut receipt_outcomes = outcomes.split_off(transactions.len());
 
         let indexer_transactions = transactions
             .into_iter()
-            .zip(outcomes.into_iter())
+            .zip(outcomes.clone().into_iter())
```
frol, line 81: Is this clone still necessary?

bowenwang1996 (approved):

frol:
@khorolets May I ask you to communicate the changes, the implications, and the timelines (there is no hurry, but it is better to migrate to the new version before the next protocol upgrade) to the partners?

PR #4247, opened 2021-04-21
(wip) (fix) Fix Storage Usage

EgorKulikov:
Storage usage is incorrect for some accounts on mainnet due to oversubtracting storage usage on access key deletion. Bug itself was already fixed. Delta between actual and expected storage usage was calculated on mainnet dump. Next time storage usage will be changed for such accounts they will be migrated to V2 and storage usage will be adjusted

### Test plan
See test_storage_usage_delta


core/primitives/src/account.rs
```diff
@@ -1,9 +1,17 @@
+#[cfg(feature = "protocol_feature_add_account_versions")]
+use crate::checked_feature;
+use crate::serialize::u128_dec_format_compatible;
 use borsh::{BorshDeserialize, BorshSerialize};
+#[cfg(feature = "protocol_feature_add_account_versions")]
+use core::default::Default;
+#[cfg(feature = "protocol_feature_add_account_versions")]
+use core::result::Result;
+#[cfg(feature = "protocol_feature_add_account_versions")]
+use core::result::Result::Ok;
```
matklad, line 10:
```suggestion
```

I don't think these are needed, these types are in prelude and dont' require an import.

core/primitives/src/account.rs
```diff
@@ -55,14 +67,25 @@ impl Account {
         locked: Balance,
         code_hash: CryptoHash,
         storage_usage: StorageUsage,
+        protocol_version: ProtocolVersion,
     ) -> Self {
-        Account {
+        #[cfg(not(feature = "protocol_feature_add_account_versions"))]
+        let _ = protocol_version;
```
matklad, line 73:
```suggestion
```

I think the `checked_feature` macro does this internally

core/primitives/src/account.rs
```diff
@@ -159,91 +199,73 @@ impl BorshSerialize for Account {
                 storage_usage: self.storage_usage,
             }
             .serialize(writer),
+            _ => SerializableAccount {
```
matklad, line 202:
```suggestion
            AccountVersion::V2 => SerializableAccount {
```

It's better to default to exhaustive matches: https://youtu.be/-J8YyfrSwTk?t=1819

EgorKulikov: This logic would be the same for all future versions, so this is by design

core/primitives/src/account.rs
EgorKulikov: becuase without this feature we have no corresponding field it would not

core/primitives/src/account.rs
```diff
@@ -128,6 +145,16 @@ struct LegacyAccount {
     storage_usage: StorageUsage,
 }
 
+#[cfg(feature = "protocol_feature_add_account_versions")]
+#[derive(BorshSerialize, BorshDeserialize)]
+struct SerializableAccount {
```
bowenwang1996, line 150: should we implement `From<SerializableAccount>` for `Account`?

bowenwang1996 (commented):
where did you fix storage usage?

EgorKulikov:
> where did you fix storage usage?

It is not in yet, safely stashed until this builds, that's why is added wip to it

core/primitives/src/account.rs
EgorKulikov: Considering it should ever be used only once I am not sure we need this

core/primitives/src/runtime/config.rs
```diff
@@ -42,13 +52,45 @@ lazy_static::lazy_static! {
...
+lazy_static::lazy_static! {
+    static ref STORAGE_USAGE_DELTA_FROM_FILE: HashMap<AccountId, u64> = RuntimeConfig::read_storage_usage_delta();
+    static ref EMPTY_STORAGE_USAGE_DELTA: HashMap<AccountId, u64> = HashMap::new();
+    // static ref STORAGE_USAGE_DELTA_CSV: str = include_str!("../../res/storage_usage_delta.csv");
+}
+
+#[cfg(feature = "protocol_feature_add_account_versions")]
+lazy_static_include::lazy_static_include_str! {
+    STORAGE_USAGE_DELTA_CSV => "res/storage_usage_delta.csv"
+}
```
frol, line 65: This looks like a hotfix for mainnet. Is it? Is it so messed up that there is no other way but to have a handcrafted list of accounts?

core/primitives/src/runtime/config.rs
```diff
@@ -42,13 +52,45 @@ lazy_static::lazy_static! {
     static ref LOWER_STORAGE_COST_CONFIG: Mutex<Option<Arc<RuntimeConfig>>> = Mutex::new(None);
 }
 
+#[cfg(feature = "protocol_feature_add_account_versions")]
+lazy_static::lazy_static! {
+    static ref STORAGE_USAGE_DELTA_FROM_FILE: HashMap<AccountId, u64> = RuntimeConfig::read_storage_usage_delta();
+    static ref EMPTY_STORAGE_USAGE_DELTA: HashMap<AccountId, u64> = HashMap::new();
+    // static ref STORAGE_USAGE_DELTA_CSV: str = include_str!("../../res/storage_usage_delta.csv");
```
frol, line 59: Seems to be a dead comment

frol (changes requested):

frol:
Can it be a migration?

core/primitives/src/runtime/config.rs
EgorKulikov: Removed

core/primitives/Cargo.toml
```diff
@@ -16,10 +16,12 @@ bs58 = "0.4"
 base64 = "0.13"
 byteorder = "1.3"
 chrono = { version = "0.4.4", features = ["serde"] }
+csv = "1.1"
 derive_more = "0.99.3"
 easy-ext = "0.2"
 sha2 = "0.9"
 lazy_static = "1.4"
+lazy-static-include = "3"
```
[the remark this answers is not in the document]
EgorKulikov: I think this is not yet possible, see https://github.com/rust-lang/cargo/issues/5954

core/primitives/res/storage_usage_delta.csv
```diff
@@ -0,0 +1,3112 @@
+0.near,1
```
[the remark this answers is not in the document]
EgorKulikov: > Can it be a migration?

I do not see how it can be considering this changes state

EgorKulikov:
> Can it be a migration?

Considering we are changing state, I do not see how it can be a migration

core/primitives/src/runtime/config.rs
EgorKulikov: It is not handcrafted, but generating it on the fly is not viable. I'll add description

core/primitives/src/runtime/config.rs
```diff
@@ -24,6 +26,13 @@ pub struct RuntimeConfig {
     pub wasm_config: VMConfig,
     /// Config that defines rules for account creation.
     pub account_creation_config: AccountCreationConfig,
+    /// Storage usage delta that need to be added on migration of account from v1 to v2
+    /// See https://github.com/near/nearcore/issues/3824
+    /// Account list for mainnet generated by utils/storage-usage-delta-calculator
+    #[serde(default = "RuntimeConfig::empty_storage_usage_delta")]
+    #[serde(skip)]
```
bowenwang1996, line 33: why do you have both default and skip?

core/primitives/src/runtime/config.rs
```diff
@@ -24,6 +26,13 @@ pub struct RuntimeConfig {
...
+    #[cfg(feature = "protocol_feature_add_account_versions")]
+    pub storage_usage_delta: &'static HashMap<AccountId, u64>,
```
bowenwang1996, line 35: What is the reason for doing this? Couldn't we have it as a global variable? I believe `&'static` should be used with caution

bowenwang1996 (commented):
Some thoughts:
* I am not sure whether reading the csv into `RuntimeConfig` is a good idea. This is a one-off fix that we do and because of this, I think it could be better if we just use a global variable that is populated when neard starts.
* How do you plan to test this change?

core/primitives/res/storage_usage_delta.csv
```diff
@@ -0,0 +1,3112 @@
...
+ziyou.near,1
+zjouraolian.near,1
+zkp.near,2
+zoek.near,1
+zorro.near,1
+zpool.near,4
+zyzygy.near,1
+zzangpk.near,1
+zzf.near,1
+zztom.near,1
```
ailisp, line 3112: I'm not totally understanding the background here and from this PR my guess is some user's storage usage is wrong and adjust by this file. Will there be more accounts have incorrect storage since this file is made?

core/primitives/src/runtime/config.rs
EgorKulikov: default for deserialization, skip for serialization

core/primitives/src/runtime/config.rs
EgorKulikov: Yes, global variable seems to be better suited

core/primitives/res/storage_usage_delta.csv
EgorKulikov: The bug that led to this discrepancy was fixed

EgorKulikov:
> * How do you plan to test this change?

I'll add test to actions.rs

core/primitives/res/storage_usage_delta.csv
EgorKulikov: The original bug was already fixed, so delta of storage usage should not change

EgorKulikov:
> * How do you plan to test this change?

I've introduced test, see test_storage_usage_delta

bowenwang1996 (commented):
Can we add at least an integration test to test that this protocol feature works fine with upgrades? You can do it in `client/tests/process_blocks.rs`

runtime/runtime/src/actions.rs
```diff
@@ -644,6 +673,13 @@ pub(crate) fn action_add_key(
         }
     );
     let storage_config = &apply_state.config.transaction_costs.storage_usage_config;
+    #[cfg(feature = "protocol_feature_add_account_versions")]
+    migrate_account(
```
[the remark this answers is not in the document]
EgorKulikov: Not in every action, only in those that change storage usage

EgorKulikov:
> Can we add at least an integration test to test that this protocol feature works fine with upgrades? You can do it in `client/tests/process_blocks.rs`

Will do

runtime/runtime/src/actions.rs
```diff
@@ -773,13 +809,67 @@ pub(crate) fn check_account_existence(
...
+    ) && account.version() == V1
+    {
+        account.set_version(V2);
+        if is_mainnet && STORAGE_USAGE_DELTA_FROM_FILE.contains_key(account_id) {
+            account.set_storage_usage(
+                account
+                    .storage_usage()
+                    .saturating_add(*STORAGE_USAGE_DELTA_FROM_FILE.get(account_id).unwrap()),
+            );
+        }
```
frol, line 855:
```suggestion
        if is_mainnet {
            if let Some(storage_usage_delta) = STORAGE_USAGE_DELTA_FROM_FILE.get(account_id) {
                account.set_storage_usage(
                    account
                        .storage_usage()
                        .saturating_add(*storage_usage_delta),
                );
            }
        }
```

utils/storage-usage-delta-calculator/src/main.rs
```diff
@@ -0,0 +1,31 @@
...
+use near_primitives::state_record::StateRecord;
+use node_runtime::Runtime;
+use std::fs::OpenOptions;
+use std::io::prelude::*;
+use std::io::Error;
+
+fn main() -> Result<(), Error> {
+    println!("Start");
+
+    let genesis = Genesis::from_file("output.json");
```
frol, line 12: I feel like you expect to have a state dump in this `output.json`, but it is not clear. Please, provide more details in the doc comments as to how to use it.

utils/storage-usage-delta-calculator/src/main.rs
```diff
@@ -0,0 +1,31 @@
...
+    let mut writer = OpenOptions::new().append(true).create(true).open("delta.csv").unwrap();
+
+    println!("Genesis read");
+    let storage_usage =
+        Runtime::compute_storage_usage(&genesis.records.0[..], &RuntimeConfig::default());
+    println!("Storage usage calculated");
+    for record in genesis.records.0 {
```
frol, line 19: I believe this should work:

```suggestion
    let storage_usage =
        Runtime::compute_storage_usage(genesis.records.as_ref(), &RuntimeConfig::default());
    println!("Storage usage calculated");
    for record in genesis.records.into() {
```

P.S. `GenesisRecords` needs one more derive `DeriveMore::Into` to support `.into()`

runtime/runtime/src/actions.rs
```diff
@@ -773,13 +809,67 @@ pub(crate) fn check_account_existence(
...
+        let account_id = &record[0];
+        let delta = record[1].parse::<u64>().expect("Malformed storage_usage_delta.csv");
+        result.insert(account_id.to_string() as AccountId, delta);
+    }
+    result
+}
+
+#[cfg(feature = "protocol_feature_add_account_versions")]
+lazy_static::lazy_static! {
+    static ref STORAGE_USAGE_DELTA_FROM_FILE: HashMap<AccountId, u64> = read_storage_usage_delta();
```
frol, line 832: We already did enough hacks around by pulling a random CSV file right into the binary, let's at least have the HashMap properly propagated through the configuration. `neard` should be the single source of configuration, we should not hijack runtime and get changed its operation due to some random global variable. Let `neard` hold the file, parse it conditionally when the chain id is `mainnet`, and pass it through the runtime configs down here if necessary.

frol (changes requested):
Is it possible to avoid the hardcoded CSV file and instead reproduce it on the fly? How long will it take? If that is not the way to go, how about generating `.rs` file instead of `.csv`? (I really don't like having more dependencies than necessary, and `csv` in `runtime` looks completely odd)

EgorKulikov:
> Is it possible to avoid the hardcoded CSV file and instead reproduce it on the fly? How long will it take? If that is not the way to go, how about generating `.rs` file instead of `.csv`? (I really don't like having more dependencies than necessary, and `csv` in `runtime` looks completely odd)

I'd rather parse csv manually than put more than 3k lines of data directly into rs file

matklad:
>I'd rather parse csv manually than put more than 3k lines of data directly into rs file

Another option is to convert CSV to borsh once, and then, at/in runtime, `include!(data.borsh)` and deserialize that. Does conceptually the same thing as embedding & parsing CSV, but re-usese existing deps

frol:
> Another option is to convert CSV to borsh

Well, it will be even more magical and impossible to review, so I would not go with it

EgorKulikov:
> Well, it will be even more magical and impossible to review, so I would not go with it

We can go with serde, I guess

bowenwang1996:
> Is it possible to avoid the hardcoded CSV file and instead reproduce it on the fly? How long will it take? If that is not the way to go, how about generating .rs file instead of .csv? (I really don't like having more dependencies than necessary, and csv in runtime looks completely odd)

@frol it is impossible to reproduce it while the node is running as it is prohibitively expensive to iterate over the state. More complex solutions are possible, but I don't think it is worth the effort here.

frol:
> We can go with serde, I guess

JSON or manual CSV parsing is fine. Let's move the preparation logic and the artifact file to `neard` crate and pass the adjustment delta-table to runtime through configuration.

PR #4228, opened 2021-04-16
Fix(chain): fix apply chunks

bowenwang1996:
Port https://github.com/near/nearcore/pull/4226 to master.

Test plan
----------
Nayduck http://nayduck.eastus.cloudapp.azure.com:3000/#/run/1508

Issue #4243, opened 2021-04-20
Epoch start is not correctly saved


bowenwang1996:
Currently we save epoch start whenever the block is the first block of an epoch [https://github.com/near/nearcore/blob/3744f07e13bf43a9522fb39fa8f6f128396d0e1f/chain/epoch_manager/src/lib.rs#L499](https://github.com/near/nearcore/blob/3744f07e13bf43a9522fb39fa8f6f128396d0e1f/chain/epoch_manager/src/lib.rs#L499) but this is not necessarily correct because there could be a fork and the fork can overwrite the epoch start on the canonical chain, which would affect the rpc output for validators endpoint.

PR #4248, opened 2021-04-22
Restore receipts after recent fix

Longarithm:
Currently rebased on https://github.com/near/nearcore/pull/4274 to reuse `MigrationData`.

This PR has two goals:

1. Re-introduce receipts to the chain, which were previously lost because of the bug in apply_chunks fixed here: https://github.com/near/nearcore/pull/4228
2. Provide a script allowing to verify that added receipts were actually lost. Because usual nodes clean old data, it can be launched only on mainnet archival node dumps.

**Test plan**

Check that
- incorrect receipts don't pass verification and there are no still missing receipts: `restored-receipts-verifier -> test_checking_differences`
- new receipts are correctly added to chain: the epoch is right, missing chunks don't affect the process, all receipts are added: `process_blocks -> test_restoring_receipts_mainnet`


bowenwang1996 (commented):
Please add a description to the PR on what this PR does and what the test plan is

utils/restored-receipts-verifier/src/main.rs
```diff
@@ -0,0 +1,106 @@
...
+    }
+
+    // Check that receipts from repo were actually generated
+    let receipt_result_in_repo_json =
+        include_str!("../../../neard/res/fix_apply_chunks_receipts.json");
+    let receipt_result_in_repo = serde_json::from_str::<ReceiptResult>(receipt_result_in_repo_json)
+        .expect("File with receipts restored after apply_chunks fix have to be correct");
+    let receipts_in_repo = receipt_result_in_repo.get(&shard_id).unwrap();
+    let receipt_hashes_in_repo =
+        HashSet::<_>::from_iter(receipts_in_repo.into_iter().map(|receipt| receipt.get_hash()));
```
matklad, line 91: Nit: this bit of code is a bit hard to follow. The smell here is that we have a whole bunch of variables here, and they all start with the same prefix. So, to a fast-reading human mind, all this variables look essentially the same, and you need to concentrate to actaully read the names fully and to understand what goes where. 

A nice trick to make the data flow here more obvious is to use a block expression: 

```rust
    let receipt_hashes_missing: HashSet<CryptoHash> =
        receipts_missing.into_iter().map(|receipt| receipt.get_hash()).collect();

    let receipt_hashes_in_repo: HashSet<CryptoHash> = {
        // Check that receipts from repo were actually generated
        let receipt_result_in_repo_json =
            include_str!("../../../neard/res/fix_apply_chunks_receipts.json");
        let receipt_result_in_repo =
            serde_json::from_str::<ReceiptResult>(receipt_result_in_repo_json)
                .expect("File with receipts restored after apply_chunks fix have to be correct");
        let receipts_in_repo = receipt_result_in_repo.get(&shard_id).unwrap();
        receipts_in_repo.into_iter().map(|receipt| receipt.get_hash()).collect()
    };
```

There's still a bunch of names here, and they are still too repetitive, but now it is clear to the reader that only the two top-level ones are worth paying attention to.

utils/restored-receipts-verifier/src/main.rs
```diff
@@ -0,0 +1,106 @@
...
+    let receipt_hashes_missing =
+        HashSet::<_>::from_iter(receipts_missing.into_iter().map(|receipt| receipt.get_hash()));
+
+    let receipt_hashes_not_verified: Vec<CryptoHash> =
+        receipt_hashes_in_repo.difference(&receipt_hashes_missing).cloned().collect();
+    assert_eq!(
+        receipt_hashes_not_verified.len(),
+        0,
```
matklad, line 99:
```suggestion
    assert!(
        receipt_hashes_not_verified.is_empty(),
```

utils/restored-receipts-verifier/src/main.rs
```diff
@@ -0,0 +1,106 @@
...
+        "Some of receipt hashes in repo were not verified successfully: {:?}",
+        receipt_hashes_not_verified
```
matklad, line 101: :+1: it's always a good idea to provide enough context in the assertion message!

utils/restored-receipts-verifier/src/main.rs
```diff
@@ -0,0 +1,106 @@
...
+use near_primitives::hash::CryptoHash;
+use near_primitives::receipt::Receipt;
+use near_store::create_store;
+use neard::{get_default_home, get_store_path, load_config, NightshadeRuntime};
+use std::collections::HashSet;
+use std::io::Error;
+use std::iter::FromIterator;
+use std::path::Path;
+
+fn main() -> Result<(), Error> {
```
matklad, line 12:
```suggestion
fn main() -> io::Result<()> {
```

chain/chain/src/chain.rs
```diff
@@ -2734,9 +2733,24 @@ impl<'a> ChainUpdate<'a> {
...
 
+        // Re-introduce receipts missing before apply_chunks fix (see https://github.com/near/nearcore/pull/4228)
+        #[cfg(feature = "protocol_feature_restore_receipts_after_fix")]
+        if self.runtime_adapter.is_next_block_epoch_start(prev_block.hash()).unwrap_or(false)
+            && prev_protocol_version < ProtocolFeature::RestoreReceiptsAfterFix.protocol_version()
+            && ProtocolFeature::RestoreReceiptsAfterFix.protocol_version() <= protocol_version
+        {
+            let receipt_result_json =
+                include_str!("../../../neard/res/fix_apply_chunks_receipts.json");
+            let receipt_result = serde_json::from_str::<ReceiptResult>(receipt_result_json)
```
EgorKulikov, line 2749: Seems like you apply receipts saved from mainnet on all nets, this may lead to weird results

utils/restored-receipts-verifier/src/main.rs
```diff
@@ -0,0 +1,106 @@
...
+    let mut chain_store = ChainStore::new(store.clone(), near_config.genesis.config.genesis_height);
+    let runtime = NightshadeRuntime::new(
+        &home_dir,
+        store,
+        &near_config.genesis,
+        near_config.client_config.tracked_accounts.clone(),
+        near_config.client_config.tracked_shards.clone(),
+    );
+
+    let mut receipts_missing: Vec<Receipt> = vec![];
```
EgorKulikov, line 39: Nit: I think Vec::new() is much more logical here

utils/restored-receipts-verifier/Cargo.toml
```diff
@@ -0,0 +1,16 @@
+[package]
+name = "restored-receipts-verifier"
+version = "0.1.0"
+authors = ["Aleksandr Logunov <alex.logunov@near.org>"]
+edition = "2018"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+clap = "2.33"
```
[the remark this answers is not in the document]
Longarithm: Simple replace of dependency with `clap = { git = "https://github.com/clap-rs/clap/" }` led to

```
error: failed to select a version for `proc-macro2`.
    ... required by package `clap_derive v3.0.0-beta.2 (https://github.com/clap-rs/clap/#92f744cc)`
    ... which is depended on by `clap v3.0.0-beta.2 (https://github.com/clap-rs/clap/#92f744cc)`
    ... which is depended on by `restored-receipts-verifier v0.1.0 (/Users/Shared/OldLaptop/Programming/nearcore/utils/restored-receipts-verifier)`
versions that meet the requirements `^1.0.26` are: 1.0.26
```

chain/chain/src/chain.rs
```diff
@@ -2734,9 +2733,24 @@ impl<'a> ChainUpdate<'a> {
...
             self.runtime_adapter.get_epoch_protocol_version(block.header().epoch_id())?;
 
+        // Re-introduce receipts missing before apply_chunks fix (see https://github.com/near/nearcore/pull/4228)
+        #[cfg(feature = "protocol_feature_restore_receipts_after_fix")]
+        if self.runtime_adapter.is_next_block_epoch_start(prev_block.hash()).unwrap_or(false)
+            && prev_protocol_version < ProtocolFeature::RestoreReceiptsAfterFix.protocol_version()
+            && ProtocolFeature::RestoreReceiptsAfterFix.protocol_version() <= protocol_version
+        {
+            let receipt_result_json =
+                include_str!("../../../neard/res/fix_apply_chunks_receipts.json");
```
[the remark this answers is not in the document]
Longarithm: Agree, such cross-package dependency was not good.
Moved initialization of `receipt_result` to `neard`. It should look a bit less hacky now, and also it helped to distinguish between mainnet and not-mainnet cases easier.

chain/chain/src/chain.rs
```diff
@@ -2734,9 +2733,24 @@ impl<'a> ChainUpdate<'a> {
...
+            let receipt_result = serde_json::from_str::<ReceiptResult>(receipt_result_json)
+                .expect("File with receipts restored after apply_chunks fix have to be correct");
+            self.chain_store_update.save_outgoing_receipt(&block.hash(), 0, receipt_result);
+        }
```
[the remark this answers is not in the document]
Longarithm: Added comments and moved `.json` reading to the `neard` package closer to reading other configs data.

neard/src/config.rs
```diff
@@ -779,12 +783,32 @@ pub fn init_configs(
...
             let genesis: Genesis = serde_json::from_slice(*MAINNET_GENESIS_JSON)
                 .expect("Failed to deserialize MainNet genesis");
+
             if let Some(account_id) = account_id {
                 generate_validator_key(account_id, &dir.join(config.validator_key_file));
             }
 
+            // Put receipts which were lost because of a bug in apply_chunks to the runtime config.
+            // See https://github.com/near/nearcore/pull/4248/ for more details
+            checked_feature!(
```
EgorKulikov, line 800: I think we want to use only bool version of this macro (and if) going forward

neard/src/config.rs
```diff
@@ -744,7 +745,10 @@ fn generate_validator_key(account_id: &str, path: &Path) {
 }
 
 lazy_static_include::lazy_static_include_bytes! {
-    MAINNET_GENESIS_JSON => "res/mainnet_genesis.json"
+    MAINNET_GENESIS_JSON => "res/mainnet_genesis.json",
+    // File with receipts which were lost because of a bug in apply_chunks to the runtime config.
+    // Follows the ReceiptResult format which is HashMap<ShardId, Vec<Receipt>>.
+    MAINNET_RESTORED_RECEIPTS_JSON => "res/mainnet_restored_receipts.json",
```
EgorKulikov, line 751: should this be hidden behind a feature?

neard/src/config.rs
```diff
@@ -779,12 +783,32 @@ pub fn init_configs(
             config.write_to_file(&dir.join(CONFIG_FILENAME));
 
             // TODO: add download genesis for mainnet
+            #[cfg(feature = "protocol_feature_restore_receipts_after_fix")]
+            let mut genesis: Genesis = serde_json::from_slice(*MAINNET_GENESIS_JSON)
+                .expect("Failed to deserialize MainNet genesis");
+
+            #[cfg(not(feature = "protocol_feature_restore_receipts_after_fix"))]
             let genesis: Genesis = serde_json::from_slice(*MAINNET_GENESIS_JSON)
                 .expect("Failed to deserialize MainNet genesis");
```
EgorKulikov, line 792:
```suggestion
            #[cfg(feature = "protocol_feature_restore_receipts_after_fix")]
            let mut genesis: Genesis;
            #[cfg(not(feature = "protocol_feature_restore_receipts_after_fix"))]
            let genesis: Genesis;
            
             genesis = serde_json::from_slice(*MAINNET_GENESIS_JSON)
                .expect("Failed to deserialize MainNet genesis");
```

utils/restored-receipts-verifier/Cargo.toml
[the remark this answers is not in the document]
matklad: You don't need a git dependency here, beta is published:

https://github.com/near/nearcore/blob/1fd8e1a56ef3206e4acfe3ae2f8d47597942cafc/tools/indexer/example/Cargo.toml#L9

neard/src/config.rs
EgorKulikov: I wonder if this is correct place to do it at all, this only runs on init I think, not on run, and besides changing genesis is a bad idea

chain/chain/src/chain.rs
```diff
@@ -2734,9 +2733,43 @@ impl<'a> ChainUpdate<'a> {
...
+                            .receipts_to_restore;
+                        self.chain_store_update.save_outgoing_receipt(
+                            &block.hash(),
+                            0,
+                            receipt_result,
+                        );
+                    }
+                }
+            );
+        }
```
bowenwang1996, line 2771: I am not sure why you want to save those as outgoing receipts. Even if you decide to do that, the implementation is not correct -- you need to change the outgoing receipt roots as well. Otherwise the invariant that outgoing receipts for a given block and shard match its outgoing receipt root in the corresponding chunk would be broken.

bowenwang1996 (changes requested):

neard/src/config.rs
Longarithm: Agree, moved it to `MigrationData`

bowenwang1996:
@Longarithm if you think your PR is ready for review, please mark it as ready for review.

chain/chain/src/chain.rs
Longarithm: Agree, moved receipts addition to `process_state_update`.

neard/src/runtime/mod.rs
```diff
@@ -193,6 +198,41 @@ impl NightshadeRuntime {
...
+        let mut epoch_manager = self.epoch_manager.as_ref().write().expect(POISONED_LOCK_ERR);
+        let mut candidate_hash = hash.clone();
+        loop {
+            let block_info = epoch_manager.get_block_info(&candidate_hash).unwrap().clone();
+            if block_info.chunk_mask()[shard_id as usize] {
+                break Ok(epoch_manager.get_epoch_info(block_info.epoch_id())?.protocol_version());
+            }
+            candidate_hash = block_info.prev_hash().clone();
+        }
+    }
```
bowenwang1996, line 234: Why is this implemented in `RuntimeAdapter` but not `Chain`?

neard/src/runtime/mod.rs
```diff
@@ -415,6 +461,53 @@ impl NightshadeRuntime {
...
+            ) {
+            let prev_protocol_version = self
+                .get_protocol_version_of_last_block_with_chunk(prev_block_hash, shard_id)
+                .unwrap();
+
+            let mut receipts_to_restore = if !checked_feature!(
+                "protocol_feature_restore_receipts_after_fix",
+                RestoreReceiptsAfterFix,
+                prev_protocol_version
+            ) {
```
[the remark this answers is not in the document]
Longarithm: No, I check above that the protocol version of previous epoch is old, to run this code only in one epoch.
Here I check that the protocol version of previous block with existing chunk is old, so I will add receipts to the first suitable block and will not do it for other ones.

chain/client/tests/process_blocks.rs
```diff
@@ -2746,6 +2890,112 @@ fn test_congestion_receipt_execution() {
...
+        let mut height: BlockHeight = 1;
+        let mut last_update_height: BlockHeight = 0;
+
+        // If some receipts are still not applied, upgrade already happened, and no new receipt was
+        // applied in some last blocks, consider the process stuck to avoid any possibility of infinite loop.
+        while !receipt_hashes_to_restore.is_empty() && height - last_update_height < height_timeout
+        {
+            let mut block = env.clients[0].produce_block(height).unwrap().unwrap();
+            block.mut_header().get_mut().inner_rest.latest_protocol_version =
+                ProtocolFeature::RestoreReceiptsAfterFix.protocol_version();
```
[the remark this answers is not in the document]
Longarithm: Thank you, removed.

chain/chain/src/chain.rs
```diff
@@ -62,8 +50,20 @@ use near_primitives::views::{
...
+use crate::types::{
+    AcceptedBlock, ApplyTransactionResult, Block, BlockEconomicsConfig, BlockHeader,
+    BlockHeaderInfo, BlockStatus, ChainGenesis, Provenance, RuntimeAdapter,
+};
+use crate::validate::{
+    validate_challenge, validate_chunk_proofs, validate_chunk_with_chunk_extra,
+    validate_transactions_order,
+};
+use crate::{byzantine_assert, create_light_client_block_view, Doomslug};
+use crate::{metrics, DoomslugThresholdMode};
```
bowenwang1996, line 66: @matklad do you have suggestions on how we can avoid re-arranging imports indefinitely :)

chain/chain/src/migrations.rs
```diff
@@ -0,0 +1,82 @@
...
+
+/// Get epoch id of the last block in which chunk for the given shard was presented.
+fn get_epoch_id_of_last_block_with_chunk(
+    chain_store: &mut dyn ChainStoreAccess,
+    hash: &CryptoHash,
+    shard_id: ShardId,
+) -> Result<EpochId, Error> {
+    let mut candidate_hash = hash.clone();
+    loop {
+        let block_header = chain_store.get_block_header(&candidate_hash)?.clone();
```
bowenwang1996, line 16: You don't need to clone here

chain/chain/src/migrations.rs
```diff
@@ -0,0 +1,82 @@
...
+        if block_header.chunk_mask()[shard_id as usize] {
+            break Ok(block_header.epoch_id().clone());
+        }
+        candidate_hash = block_header.prev_hash().clone();
+    }
+}
+
+/// Get epoch id by hash of previous block.
```
bowenwang1996, line 24: Do you mean get previous epoch id?

chain/chain/src/migrations.rs
```diff
@@ -0,0 +1,82 @@
...
+    chain_store: &mut dyn ChainStoreAccess,
+    runtime_adapter: &dyn RuntimeAdapter,
+    block_hash: &CryptoHash,
+) -> Result<EpochId, Error> {
+    let epoch_start_header = {
+        let height = runtime_adapter.get_epoch_start_height(block_hash)?;
+        let hash = chain_store.get_block_hash_by_height(height)?;
+        chain_store.get_block_header(&hash)?.clone()
+    };
+    if runtime_adapter.is_next_block_epoch_start(block_hash).unwrap() {
```
bowenwang1996, line 35: why does it make sense to unwrap here?

chain/chain/src/migrations.rs
```diff
@@ -0,0 +1,82 @@
...
+/// Get epoch id by hash of previous block.
+fn get_prev_epoch_id_from_prev_block(
+    chain_store: &mut dyn ChainStoreAccess,
+    runtime_adapter: &dyn RuntimeAdapter,
+    block_hash: &CryptoHash,
+) -> Result<EpochId, Error> {
+    let epoch_start_header = {
+        let height = runtime_adapter.get_epoch_start_height(block_hash)?;
+        let hash = chain_store.get_block_hash_by_height(height)?;
+        chain_store.get_block_header(&hash)?.clone()
```
bowenwang1996, line 33: This unfortunately does not work due to #4243

chain/chain/src/migrations.rs
```diff
@@ -0,0 +1,82 @@
...
+        protocol_version
+    ) && !checked_feature!(
+        "protocol_feature_restore_receipts_after_fix",
+        RestoreReceiptsAfterFix,
+        prev_epoch_protocol_version
+    ) {
+        let prev_protocol_version = runtime_adapter.get_epoch_protocol_version(
+            &get_epoch_id_of_last_block_with_chunk(chain_store, prev_block_hash, shard_id)?,
+        )?;
+        Ok(shard_id == 0
```
bowenwang1996, line 73: It's better to extract this out to the start of the function so that we don't do unnecessary computation if shard id is not `0`.

chain/chain/src/migrations.rs
```diff
@@ -0,0 +1,82 @@
+use crate::store::ChainStoreAccess;
+use crate::types::RuntimeAdapter;
+use near_chain_primitives::error::Error;
+use near_primitives::checked_feature;
+use near_primitives::hash::CryptoHash;
+use near_primitives::types::{EpochId, ShardId};
+
+/// Get epoch id of the last block in which chunk for the given shard was presented.
+fn get_epoch_id_of_last_block_with_chunk(
```
bowenwang1996, line 9: What's the reason for implementing these functions as standalone functions rather than methods of `ChainUpdate` or `Chain`?

Longarithm: We have calls of `runtime.apply_transactions` in many places, like `apply_block_at_height` in `neard/migrations`, `state-viewer`. Also there is `validate_chunk_state_challenge`, where we don't call it yet. 
In all these places we currently create only `ChainStore` or just block headers.

chain/chain/src/chain.rs
matklad: I think two strategies work for big projects for dealing with "code normalization", including imports:

* just don't care about imports, let everyone arrange & rearrage them however they like
* make sure that all existing code is using consistent imports, so that the path of the least resistance for the new code is to be consistent. 

To make the second case work best, you need 

* simple opinionated rules which doesn't leave much place for judgement
* de-facto usage of those rule by the majority of the project
* some document which spells the rules out, to like during code reviews

rust-analyzer uses the second approach, it works OK. Here are our rules: https://github.com/rust-analyzer/rust-analyzer/blob/ad0648dc95d2d1e9d35cd825c2661fe002c02fc9/docs/dev/style.md#order-of-imports

At some point, this might be enforceable via rustfmt (https://github.com/rust-lang/rustfmt/issues/2979), but it doesnt' look like it'll be done soon. 

Practically, for nearcore, I suggest:

* Creating `style.md` document, and writing down *some* strategy we use (even if it is "don't care, don't touch existing imports")
* Maybe looking at the existing code and writing down a strategy we aspire to, and start passively normalizing the code to that. I think that might work out in the long-run.

chain/chain/src/migrations.rs
```diff
@@ -0,0 +1,54 @@
...
+    // Check that block belongs to the first epoch where the protocol feature was enabled
+    // to avoid get_epoch_id_of_last_block_with_chunk call in the opposite case
+    if checked_feature!(
+        "protocol_feature_restore_receipts_after_fix",
+        RestoreReceiptsAfterFix,
+        protocol_version
+    ) && !checked_feature!(
+        "protocol_feature_restore_receipts_after_fix",
+        RestoreReceiptsAfterFix,
+        prev_epoch_protocol_version
```
bowenwang1996, line 41: I think it is better to use separate variables for these two booleans to make the code more readable

neard/src/runtime/mod.rs
```diff
@@ -315,6 +314,10 @@ impl NightshadeRuntime {
         }
     }
 
+    pub fn migration_data(&self) -> Arc<MigrationData> {
```
bowenwang1996, line 317: where is this used?

Longarithm: I use it in `protocol_feature_restore_receipts_after_fix_tests` test in `process_blocks`, to extract migration data from runtime directly and avoid unnecessary data loading.

bowenwang1996: If it is used in test please either decorate it with #[cfg(test)] or find other ways to implement it so that it does not get used accidentally anywhere else.

runtime/runtime/src/lib.rs
```diff
@@ -1101,8 +1105,27 @@ impl Runtime {
...
+        let _ = incoming_receipts;
+        #[cfg(feature = "protocol_feature_restore_receipts_after_fix")]
+        if ProtocolFeature::RestoreReceiptsAfterFix.protocol_version() == protocol_version
+            && migration_flags.is_first_block_with_chunk_of_version
+        {
+            let restored_receipts = migration_data
+                .restored_receipts
+                .get(&0u64)
+                .expect("Receipts to restore must contain an entry for shard 0")
+                .clone();
```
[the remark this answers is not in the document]
Longarithm: Because `get` returns only a reference to `Vec<Receipt>`. I can call `remove` instead but don't think that modifying `migration_data` is a good idea.

chain/chain/src/migrations.rs
```diff
@@ -0,0 +1,38 @@
...
+/// Check that block is the first one with existing chunk for the given shard in the chain with its protocol version.
+/// We assume that current block contain the chunk for shard with the given id.
+pub fn check_if_block_is_first_with_chunk_of_version(
+    chain_store: &mut dyn ChainStoreAccess,
+    runtime_adapter: &dyn RuntimeAdapter,
+    prev_block_hash: &CryptoHash,
+    shard_id: ShardId,
+) -> Result<bool, Error> {
+    // At first, check that shard id = 0 and don't do unnecessary computation otherwise
+    if shard_id != 0 {
```
EgorKulikov, line 16: I am not sure this is still correct way to do this, as in the future we may have migrations that need to be applied on first block with chunk for other shards as well

Longarithm: But later we can replace it with the following, right?
```
if shard_id != 0 && !checked_feature(SomeMigrationWithOtherShard, ...) {
    return Ok(false);
}
```

chain/chain/src/migrations.rs
```diff
@@ -0,0 +1,38 @@
...
+    }
+
+    let epoch_id = runtime_adapter.get_epoch_id_from_prev_block(prev_block_hash)?;
+    let protocol_version = runtime_adapter.get_epoch_protocol_version(&epoch_id)?;
+    let prev_epoch_id = match runtime_adapter.get_prev_epoch_id_from_prev_block(prev_block_hash) {
+        Ok(epoch_id) => epoch_id,
+        _ => {
+            return Ok(false);
+        }
+    };
```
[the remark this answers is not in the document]
Longarithm: Because `prev_epoch_id` may not exist, e.g. during the first epoch in scope of `
sync::test::test_block_sync` test.
I actually think it's better to return `true` in this case, moved this part to function `is_first_epoch_with_protocol_version`.

chain/chain/Cargo.toml
```diff
@@ -17,6 +17,7 @@ num-rational = "0.3"
 tracing = "0.1.13"
 thiserror = "1.0"
 strum = "0.20"
+serde_json = "1"
```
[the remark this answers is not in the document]
Longarithm: It is a legacy import, removed.

runtime/runtime/src/lib.rs
```diff
@@ -1139,16 +1162,16 @@ impl Runtime {
...
+        let mut incoming_receipts = incoming_receipts.to_vec();
+        let gas_used_for_migrations = self
+            .apply_migrations(
+                &mut state_update,
+                &mut incoming_receipts,
+                &apply_state.migration_data,
+                &apply_state.migration_flags,
+                apply_state.current_protocol_version,
+            )
+            .map_err(|e| RuntimeError::StorageError(e))?;
```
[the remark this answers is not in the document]
Longarithm: Fixed.

chain/chain/src/migrations.rs
EgorKulikov: I am not sure time savings would outweigh additional code complexity. Have you measured how long this runs?

chain/chain/src/store.rs
```diff
@@ -278,6 +278,22 @@ pub trait ChainStoreAccess {
...
         }
     }
+
+    /// Get epoch id of the last block with existing chunk for the given shard id.
+    fn get_epoch_id_of_last_block_with_chunk(
+        &mut self,
+        hash: &CryptoHash,
+        shard_id: ShardId,
+    ) -> Result<EpochId, Error> {
+        let mut candidate_hash = hash.clone();
```
EgorKulikov, line 288: Why do we do all the clones here if we ever only want &CryptoHash?

core/primitives/src/runtime/config.rs
```diff
@@ -1,12 +1,11 @@
 //! Settings of the parameters of the runtime.
-use serde::{Deserialize, Serialize};
-
 use crate::checked_feature;
 use crate::config::VMConfig;
 use crate::runtime::fees::RuntimeFeesConfig;
 use crate::serialize::u128_dec_format;
 use crate::types::{AccountId, Balance};
 use crate::version::ProtocolVersion;
+use serde::{Deserialize, Serialize};
```
EgorKulikov, line 8: This file only has changes in imports, maybe revert?

core/primitives/src/runtime/migration_data.rs
```diff
@@ -11,10 +13,21 @@ pub struct MigrationData {
...
         f.debug_struct("MigrationData").finish()
     }
 }
+
+#[derive(Debug, Default)]
+pub struct MigrationFlags {
+    // True iff the current block is the first one in the chain with the current version
+    pub is_first_block_of_version: bool,
+    // True iff the current block containing chunk for some specific shard is the first one in the
+    // chain with the current version
```
EgorKulikov, line 31: This seems to be only true if first epoch of the current version have block with chunk (and also currently only for shard 0)

neard/src/migrations.rs
```diff
@@ -272,15 +282,29 @@ lazy_static_include::lazy_static_include_bytes! {
     MAINNET_STORAGE_USAGE_DELTA => "res/storage_usage_delta.json",
 }
 
+#[cfg(feature = "protocol_feature_restore_receipts_after_fix")]
+lazy_static_include::lazy_static_include_bytes! {
```
EgorKulikov, line 286: Maybe swap this with next constant so that things related to single migration is consecutive

chain/chain/src/store.rs
Longarithm: When we call `self.get_block_header` in loop, we borrow `*self` mutably, and if we store just a reference to needed hash, we can't unborrow `*self`, which leads to 
```
error[E0499]: cannot borrow `*self` as mutable more than once at a time
```

Longarithm: Confirmed with @matklad

runtime/runtime/src/lib.rs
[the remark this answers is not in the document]
bowenwang1996: I thought you only need a reference here. Where do you need the ownership of those receipts?

chain/chain/src/migrations.rs
Longarithm: It usually takes `50µs`, which is usually about 1/20 of the `apply_chunks` call time in the test environment.
I expect that in the mainnet `apply_chunks` call time must grow, but  `check_if_block_is_first_with_chunk_of_version` call time shouldn't change, so this optimization can be removed.

runtime/runtime/src/lib.rs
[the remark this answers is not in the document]
Longarithm: Later, I need to create a list with all receipts (`restored + incoming`).
So I think I have to clone `restored` ones anyway, if I don't modify `migration_data`.

chain/chain/src/migrations.rs
```diff
@@ -0,0 +1,41 @@
...
+) -> Result<bool, Error> {
+    match runtime_adapter.get_prev_epoch_id_from_prev_block(prev_block_hash) {
+        Ok(prev_epoch_id) => {
+            let epoch_id = runtime_adapter.get_epoch_id_from_prev_block(prev_block_hash)?;
+            let protocol_version = runtime_adapter.get_epoch_protocol_version(&epoch_id)?;
+            let prev_epoch_protocol_version =
+                runtime_adapter.get_epoch_protocol_version(&prev_epoch_id)?;
+            Ok(protocol_version != prev_epoch_protocol_version)
+        }
+        Err(_) => Ok(true),
```
bowenwang1996, line 20: This looks very suspicious to me. Have you accounted for all possible ways an attacker can maliciously send you some block that such `get_prev_epoch_id_from_prev_block` returns an error?

chain/chain/src/migrations.rs
```diff
@@ -0,0 +1,41 @@
...
+    runtime_adapter: &dyn RuntimeAdapter,
+    prev_block_hash: &CryptoHash,
+    shard_id: ShardId,
+) -> Result<bool, Error> {
+    // Check that block belongs to the first epoch with current protocol version
+    // to avoid get_epoch_id_of_last_block_with_chunk call in the opposite case
+    if is_first_epoch_with_protocol_version(runtime_adapter, prev_block_hash)? {
+        // Compare only epochs because we already know that current epoch is the first one with current protocol version
+        Ok(chain_store.get_epoch_id_of_last_block_with_chunk(prev_block_hash, shard_id)?
+            != runtime_adapter.get_epoch_id_from_prev_block(prev_block_hash)?)
```
bowenwang1996, line 37: Please use separate variables to avoid very long expressions like this

bowenwang1996 (changes requested):

chain/chain/src/migrations.rs
Longarithm: I've tried to handle error coming from `KeyValueRuntime.get_prev_epoch_id_from_prev_block` but now found that my implementation was incorrect. Fixed

neard/src/runtime/mod.rs
Longarithm: It's problematic to use `#[cfg(test)]` currently: https://github.com/near/nearcore/issues/4305
So I used `load_migration_data`.

olonho (approved):

frol:
@Longarithm @bowenwang1996 Will the restored receipts be stored in the storage (e.g. accessible via [`GetReceipt` view client method](https://github.com/near/nearcore/blob/c90548f761be6ca362a493b000a19b8030ad137a/chain/client/src/view_client.rs#L879-L890))?

/cc @khorolets

Longarithm:
Result of applying receipts to current mainnet state:

Commit used: https://github.com/near/nearcore/commit/6d2530b472d2e7726f760fcdec44b55a86573466

Logs:

```
$ ./target/debug/state-viewer apply --height 41798999 --shard_id 0
Aug 21 00:17:33.763  INFO near: Opening store database at "/home/Aleksandr1/.near/data"
Aug 21 00:17:34.060  INFO runtime: Tracking shards: {0}
Pre-balance: b4a0c9f6f783bfcde1b1d4b75c85341c1c0d2e075f58d7a231601193d14ff91a 75182679016125035577525
Pre-balance: b828e1925025d7e12d1cf1268d6d9316c9972f85913bf302458a33516c3ebaef 75031270125311804664489
Pre-balance: tom.zest.near 2110354731541133588492954595
Receipts to restore len = 383
Incoming receipts len = 390
Processed: 390, Delayed: 0
Post-balance: b4a0c9f6f783bfcde1b1d4b75c85341c1c0d2e075f58d7a231601193d14ff91a 77427050088811915307435
Post-balance: b828e1925025d7e12d1cf1268d6d9316c9972f85913bf302458a33516c3ebaef 77315256944165011573409
Post-balance: tom.zest.near 2110455899816030404631797163
apply chunk for shard 0 at height 41798999, resulting chunk extra ChunkExtraV1 { state_root: `3bGmsULynNUdN1jye64FoD1c2SdNg94q6tCAuSrhhKyy`, outcome_root: `Gsn5v5Aopgrj6bLXjnMDv7H5CAjMtW4R2NVqCWfKsnZx`, validator_proposals: [], gas_used: 267292638915101, gas_limit: 1000000000000000, balance_burnt: 22938360794841900000000 }
Existing chunk extra: ChunkExtraV1 { state_root: `C9FWq6sHK2QGrvyux5LQQA8fTZXkyYtagJdT8ZXLL1gr`, outcome_root: `J9rwHeE6pv3oRyYWMVPtKgrLTxNS2JyPB68rFrVwj1Ae`, validator_proposals: [], gas_used: 52966524587327, gas_limit: 1000000000000000, balance_burnt: 4001453174034300000000 }
``` 

"Processed: 390, Delayed: 0" shows that all restored receipts were applied, and exit code 0 shows that no issues were happened.

Some research of accounts balance.
Compute post-balance minus pre-balance minus all deposits made in restored receipts, we expect to get 0.

```
b4a0c9f6f783bfcde1b1d4b75c85341c1c0d2e075f58d7a231601193d14ff91a:
mainnet_restored_receipts.json contains one refund for it;
>>> 77427050088811915307435-75182679016125035577525-2244371072686879729910 
0

b828e1925025d7e12d1cf1268d6d9316c9972f85913bf302458a33516c3ebaef:
mainnet_restored_receipts.json contains one refund for it;
>>> 77315256944165011573409-75031270125311804664489-2283986818853206908920 
0

tom.zest.near
mainnet_restored_receipts.json contains two refunds for it;
>>> 2110455899816030404631797163-2110354731541133588492954595-10000000000000000000000-22561702411076162935108
68606572485739975907460
```

But receipts in [considered block](https://explorer.near.org/blocks/4YfXc3K3xkPeg1xqpdAY5cgoY61MNfBdVt8yMBs9NxxM) also affect this account. There are two such ones, and
```
>>> 68606572485739975907460-18606572485739975907460-50000000000000000000000
0
```

Same thing was checked for two previous blocks.

PR #4250, opened 2021-04-23
feat: stabilize multiple nightly features

bowenwang1996:
Stabilize `protocol_feature_rectify_inflation`, `protocol_feature_access_key_nonce_range`, and `protocol_feature_forward_chunk_parts`

Test plan
---------
Nayduck http://nayduck.eastus.cloudapp.azure.com:3000/#/run/1557


chain/chain/src/chain.rs
```diff
@@ -2826,24 +2826,19 @@ impl<'a> ChainUpdate<'a> {
...
+                        for transaction in transactions {
+                            self.chain_store_update
+                                .get_chain_store()
+                                .check_transaction_validity_period(
+                                    prev_block.header(),
+                                    &transaction.transaction.block_hash,
+                                    transaction_validity_period,
+                                )
+                                .map_err(|_| Error::from(ErrorKind::InvalidTransactions))?;
                         }
```
matklad, line 2840:
```suggestion
                    if checked_feature!("stable", AccessKeyNonceRange, protocol_version) {
                        let transaction_validity_period = self.transaction_validity_period;
                        for transaction in transactions {
                            self.chain_store_update
                                .get_chain_store()
                                .check_transaction_validity_period(
                                    prev_block.header(),
                                    &transaction.transaction.block_hash,
                                    transaction_validity_period,
                                )
                                .map_err(|_| Error::from(ErrorKind::InvalidTransactions))?;
                        }
```

I think it's better to remove `("stable", $feature:ident, $current_protocol_version:expr, $feature_block:block) => {{` branch of macro definition altogether, as a plain `if` works in this case, and is more obvious.

chain/chunks/src/lib.rs
```diff
@@ -1215,14 +1209,9 @@ impl ShardsManager {
...
-            "protocol_feature_forward_chunk_parts",
-            ForwardChunkParts,
-            protocol_version,
-            {
-                self.send_partial_encoded_chunk_to_chunk_trackers(partial_encoded_chunk)?;
-            }
-        );
+        checked_feature!("stable", ForwardChunkParts, protocol_version, {
+            self.send_partial_encoded_chunk_to_chunk_trackers(partial_encoded_chunk)?;
+        });
```
matklad, line 1214:
```suggestion
        if checked_feature!("stable", ForwardChunkParts, protocol_version) {
            self.send_partial_encoded_chunk_to_chunk_trackers(partial_encoded_chunk)?;
        }
```

core/primitives/src/version.rs
```diff
@@ -147,6 +141,16 @@ impl ProtocolFeature {
...
+    ("stable", $feature:ident, $current_protocol_version:expr) => {{
+        $crate::version::ProtocolFeature::$feature.protocol_version() <= $current_protocol_version
+    }};
+    ("stable", $feature:ident, $current_protocol_version:expr, $feature_block:block) => {{
+        if $crate::version::ProtocolFeature::$feature.protocol_version()
+            <= $current_protocol_version
+        {
+            $feature_block
+        }
+    }};
```
matklad, line 153:
```suggestion
```

(see comment on https://github.com/near/nearcore/pull/4250/files#diff-ef9c6aaa80a330e446c5365f42be9bff37ba4f898cf519dadd7e17545783c77cR2839)

runtime/runtime/src/verifier.rs
```diff
@@ -131,24 +131,19 @@ pub fn verify_and_charge_transaction(
...
-            if let Some(height) = block_height {
-                let upper_bound =
-                    height * near_primitives::account::AccessKey::ACCESS_KEY_NONCE_RANGE_MULTIPLIER;
-                if transaction.nonce >= upper_bound {
-                    return Err(InvalidTxError::NonceTooLarge {
-                        tx_nonce: transaction.nonce,
-                        upper_bound,
-                    }
-                    .into());
+    checked_feature!("stable", AccessKeyNonceRange, current_protocol_version, {
```
matklad, line 134:
```suggestion
    if checked_feature!("stable", AccessKeyNonceRange, current_protocol_version) {
```

chain/epoch_manager/src/reward_calculator.rs
```diff
@@ -60,28 +56,23 @@ impl RewardCalculator {
             if use_hardcoded_value { Rational::new_raw(1, 20) } else { self.max_inflation_rate };
         let protocol_reward_rate =
             if use_hardcoded_value { Rational::new_raw(1, 10) } else { self.protocol_reward_rate };
-        let epoch_total_reward: u128 = checked_feature!(
-            "protocol_feature_rectify_inflation",
-            RectifyInflation,
-            protocol_version,
-            {
+        let epoch_total_reward: u128 =
+            if checked_feature!("stable", RectifyInflation, protocol_version) {
```
EgorKulikov, line 60: Why you've switched from macro to if/else?

EgorKulikov (approved):

frol (approved):
Releasing a few features at once feels a bit scary, but it seems that we should be good given that we reviewed all the places where the features used to alter the behavior.

bowenwang1996:
> Releasing a few features at once feels a bit scary

I thought we agreed on this :)

chain/epoch_manager/src/lib.rs
```diff
@@ -369,7 +368,6 @@ impl EpochManager {
             next_version,
         } = epoch_summary;
 
-        #[cfg(feature = "protocol_feature_rectify_inflation")]
```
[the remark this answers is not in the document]
bowenwang1996: No. `calculate_reward` has the check inside

pmnoxx (approved):
LGTM
Do you think it would be safer to ship those features separately?
Or is there any reason to ship them together in one PR?

bowenwang1996:
> Do you think it would be safer to ship those features separately?
> Or is there any reason to ship them together in one PR?

With the nightly features design it is expected that we may stabilize more than one feature in one release. Also we are a bit behind here as some of the features stabilized in this PR have been implemented half a year ago.

chain/chain/src/types.rs
```diff
@@ -118,7 +118,6 @@ pub struct BlockHeaderInfo {
     pub chunk_mask: Vec<bool>,
     pub total_supply: Balance,
     pub latest_protocol_version: ProtocolVersion,
-    #[cfg(feature = "protocol_feature_rectify_inflation")]
```
olonho, line 121: How those changes affect databases and RPC layouts?

olonho: Have we considered removing all protocol-specific fields and replacing those with single extension point, like I sketeched in this example
```
#[derive(Debug)]
struct Data {
    pub extensions: Vec<Option<Extensions>>,
}

fn set_ext(d: &mut Data, e: Extensions) {
    let index = index_of(&e);
    println!("index is {}", index);
    if index <= d.extensions.len() {
        d.extensions.resize_with(index + 1, || None);
    }
    d.extensions[index] = Some(e);
}

fn get_F2(d: &Data) -> i64 {
    match d.extensions[2] {
        Some(Extensions::F2(value)) => value,
        _ => panic!()
    }
}

#[test]
fn extensions() {
    let mut d = Data { extensions: vec!() };
    set_ext(&mut d, Extensions::F1(1));
    set_ext(&mut d, Extensions::F2(2));
    set_ext(&mut d, Extensions::F3("Hello".to_string()));
    println!("d = {:?}", d);
    println!("f2 = {:?}", get_F2(&d));
}
```

Maybe @matklad can suggest better idioms here.

chain/chain/src/chain.rs
```diff
@@ -2819,24 +2819,19 @@ impl<'a> ChainUpdate<'a> {
...
-                                        &transaction.transaction.block_hash,
-                                        transaction_validity_period,
-                                    )
-                                    .map_err(|_| Error::from(ErrorKind::InvalidTransactions))?;
-                            }
+                    if checked_feature!("stable", AccessKeyNonceRange, protocol_version) {
+                        let transaction_validity_period = self.transaction_validity_period;
+                        for transaction in transactions {
+                            self.chain_store_update
+                                .get_chain_store()
```
matklad, line 2826: Unrelated to the PR, but `get_` prefixes are not used in Rust. `get` (unlikne `get_mut`) also typically implies no mutation. 

Looking at the scary comment on the method, it seems a slightly better name would be `.raw_chain_store()` -- `raw` signals that this might break some highere-level invariant.

core/store/src/migrations.rs
```diff
@@ -656,25 +664,46 @@ pub fn migrate_18_to_rectify_inflation(path: &String) {
...
-            *block_header.last_final_block(),
-            *block_header.prev_hash(),
-            block_header.validator_proposals().map(|v| v.into_v1()).collect(),
-            block_header.chunk_mask().to_vec(),
-            vec![],
-            block_header.total_supply(),
-            block_header.latest_protocol_version(),
-            block_header.raw_timestamp(),
-        )
+        if key == &[0; 32] {
```
olonho, line 669: Looks somewhat strange, maybe introduce some specifically named constant or checker function `.is_all_zeros()`?

neard/src/runtime/mod.rs
```diff
@@ -1863,44 +1854,26 @@ mod test {
...
-                            * U256::from(num_ns_in_second)))
-                    .as_u128();
-                let per_epoch_protocol_treasury = per_epoch_total_reward
-                    * *self.runtime.genesis_config.protocol_reward_rate.numer() as u128
-                    / *self.runtime.genesis_config.protocol_reward_rate.denom() as u128;
-                let per_epoch_per_validator_reward =
-                    (per_epoch_total_reward - per_epoch_protocol_treasury) / num_validators as u128;
-                (per_epoch_per_validator_reward, per_epoch_protocol_treasury)
-            }
+            let num_seconds_per_year = 60 * 60 * 24 * 365;
```
olonho, line 1859: Does it account for leap years? Shall it?

neard/src/runtime/mod.rs
```diff
@@ -1863,44 +1854,26 @@ mod test {
...
+            let num_ns_in_second = 1_000_000_000;
+            let per_epoch_total_reward =
+                (U256::from(*self.runtime.genesis_config.max_inflation_rate.numer() as u64)
```
olonho, line 1862: Maybe use `Ratio`, if computation shall be precise?

olonho (commented):
Overall LGMT, few nits

chain/chain/src/types.rs
matklad: I guess there are two points here:

* how do we add fields to internal code 
* how do we add fields to schemes we use for serialization

For the first one, I think "just add a field" is probably the right solution -- it's valuable that stabilization is just removal of gating, rather than a change from a dynamic lookup in a vector to a field lookup.

For serialization, I think we might already have the desired extensibility because the top-level of things we serialize are `enums`? That's a different pattren that proposed here, it's not super clear to me which one is better, I guess that depends also on the specific format we use. For example, in JSON having an additional `data` array doesn't seem that much different from just adding additional fields to the object itself. For borsh, it might make sense, as it, in a sense, transformes borsh into a self-describing format, where you encode each object as a list of key/value pairs.


From my POV, the primary problem is not the seriazliation format per se, but rather that we use `derive(Serialize)` types throughout codebase. So, unrelated change to the internals might compile in isolation, but break compatibility in a bad way. I prefer a setup where no types used internally a `Serizlize` and instead a bit of boilerplate code is written explicitly to convert between storage format and domain objects (as outilned in https://github.com/near/nearcore/pull/4028#discussion_r587352294)

matklad (approved):

chain/chain/src/types.rs
bowenwang1996: > How those changes affect databases and RPC layouts?

There is a database migration

neard/src/runtime/mod.rs
bowenwang1996: No. This is not supposed to be 100% accurate and some small differences don't really matter.

neard/src/runtime/mod.rs
bowenwang1996: How does it make it more precise? This number must be an integer

olonho: You use integer division with multiple division and multiplication, Ratio can try make intermediate steps more precise.

SkidanovAlex (approved):

olonho (approved):

PR #4251, opened 2021-04-23
Change genesis json hash Closes #4249

### LINKED ISSUE #4249: Change Genesis::json_hash to allow streaming records
In #4190 the current plan is to process records one by one as we read them from the file. The current hash becomes difficult to compute in this model.
Updating the hash function will mean that this check https://github.com/near/nearcore/blob/master/neard/src/runtime/mod.rs#L271 will fail on node restart, so we also need to do a database migration to remove the old genesis hash.

posvyatokum:


core/chain-configs/src/genesis_config.rs
```diff
@@ -264,6 +266,37 @@ impl GenesisRecords {
...
+
+    pub fn process_genesis(&mut self, genesis: &Genesis) {
+        self.process_config(&genesis.config);
+        for record in genesis.records.as_ref() {
+            self.process_record(record)
+        }
+    }
+
+    pub fn get_hash(self) -> CryptoHash {
+        CryptoHash(self.digest.finalize().into())
```
bowenwang1996, line 296: what's the complexity of this function? Is `finalize` an expensive operation?

neard/src/lib.rs
```diff
@@ -203,6 +203,11 @@ pub fn apply_store_migrations(path: &String, near_config: &NearConfig) {
         // version 19 => 20: fix execution outcome
         migrate_19_to_20(&path, &near_config);
     }
+    if db_version <= 20 {
+        info!(target: "near", "Migrate DB from version 20 to 21");
+        // version 20 => 21: delete genesis json hash
```
bowenwang1996, line 208: can you add a bit more explanation as to why we need to delete genesis json hash?

core/chain-configs/src/genesis_config.rs
bowenwang1996: nvm didn't realize that it takes `self` instead of `&self`

posvyatokum: Complexity stays the same as before because this function can only be called once. Although it's not clear from the name, I'll change that.

bowenwang1996 (approved):

core/chain-configs/src/genesis_config.rs
```diff
@@ -264,6 +266,37 @@ impl GenesisRecords {
     }
 }
 
+pub struct GenesisJsonHasher {
+    pub digest: sha2::Sha256,
```
bowenwang1996, line 270: Do we actually need to make this field public?

posvyatokum: Yeah, better make it private

core/chain-configs/src/genesis_config.rs
```diff
@@ -264,6 +266,37 @@ impl GenesisRecords {
...
+        Self { digest: sha2::Sha256::new() }
+    }
+
+    pub fn process_config(&mut self, config: &GenesisConfig) {
+        let mut ser = Serializer::pretty(&mut self.digest);
+        config.serialize(&mut ser).expect("Error serializing the genesis config.");
+    }
+
+    pub fn process_record(&mut self, record: &StateRecord) {
+        let mut ser = Serializer::pretty(&mut self.digest);
```
matklad, line 284: Is there a reason why we are using `pretty` here? My understanding is that we are changing the hash anyway, so we wight as well use the compact form, to hash less bytes.

core/chain-configs/src/genesis_config.rs
```diff
@@ -264,6 +266,37 @@ impl GenesisRecords {
...
+        record.serialize(&mut ser).expect("Error serializing the genesis record.");
+    }
+
+    pub fn process_genesis(&mut self, genesis: &Genesis) {
+        self.process_config(&genesis.config);
+        for record in genesis.records.as_ref() {
+            self.process_record(record)
+        }
+    }
```
matklad, line 293: I'd write this as 

```rust
impl GenesisJsonHasher {
    pub fn process_genesis(genesis: &Genesis) -> CryptoHash {
      let mut hasher = Self::new();
      hasher.process_config(&genesis.config);
      genessis.records.iter().for_each(|it| hasher.process_record(it));
      hasher.finalize()
    }
}
```

`&mut self` here doesn't really make sense, as the only correct way to use this function is starting with an empty hasher and calling `.finalize` in the end. So we might as well just make it impossible to call this the wrong way.

matklad (commented):
Took a second look, noticed a couple of minor points, not really requiring fixes.

PR #4252, opened 2021-04-23
Add helper for gas value

ilyar:


core/primitives-core/src/convert.rs
```diff
@@ -0,0 +1,24 @@
+use crate::types::Gas;
+
+pub trait ToGas<T> {
+    fn tera(_: T) -> Self;
+}
+
+impl ToGas<f64> for Gas {
+    fn tera(tera_gas: f64) -> Self {
+        (tera_gas * f64::powf(10_f64, 12_f64)) as Gas
```
frol, line 9: floats are definitely not the way to go

frol (commented):
I don't feel like it is a good abstraction. Please, provide more details and use-cases.

matklad:
@frol I think https://github.com/near/near-sdk-rs/pull/368 is the required context. 

My understanding (though I am second guessing here; @ilyar if you can provide more details here, that would be helpful) is that the problem this solves is "writing tests with near-sdk-sim". As such, I think this functionality belongs to the near-sdk-rs repo, rather than the nearcore repo (our current crate setup is to blame for confusion though).

core/primitives-core/src/convert.rs
ilyar: I think this should be done in near-sdk-rs
if not float what should be used, my first version was made for a string, but @matklad  [showed that it is better to use float](https://github.com/near/near-sdk-rs/pull/368#discussion_r617464033), I was confused

frol:
Let's continue the discussion on near-sdk-rs side. Given that this implementation is not used inside nearcore, we should not include it at least until we refactor primitives to be a well-defined crate of its own with all the nice helpers.

Issue #3456, opened 2020-10-05
Support blockhash API


ilblackdragon:
Currently we return hash of 0x0 for `blockhash` API for Solidity.

Challenge with this, is that we don't have access to blocks data at all from `NearExt`.

ilblackdragon:
Also given most of the times, this used for randomness and time in Ethereum, it may be useful to add separate precompiles for those two things.

ilblackdragon:
Alternative, is to raise Invalid instruction.

bowenwang1996:
> Challenge with this, is that we don't have access to blocks data at all from NearExt.

Why is it not possible to pass it to `NearExt`?

frankbraun:
On the [weekly update 2021-02-26](https://github.com/near/nearcore/discussions/4019) @abacabadabacaba suggested that we use a "simulated blockhash" for this. For example, we could simply hash the block height (`block.number`) to get a unique block hash for every block.

I think that makes sense. However, we could have EVM contracts deployed on multiple shards in the future, or even in the same shard. Therefore, in order to have unique block hashes for every EVM, I suggest something like this: `blockhash = hash(account_id|block.number)` where `account_id` is the account identifier the EVM is deployed under.

nearmax:
I suggest to use `hash(account_id, block.hash)` as EMV hash to take care of potential rare situation where we have short-term NEAR fork.

joshuajbouw:
@nearmax I'm addressing this in #4256. Did you mean to have the `hash(account_id, block_hash)` as part of the EVM? For sake of consistency, I think its fine for the `vm_logic` just to return the `block_hash` from `VMContext`.

ilblackdragon:
@joshuajbouw `block.blockhash(uint blockNumber) returns (bytes32)` in Solidity takes the block number up to 256 most recent.

I remember the idea by @abacabadabacaba was discussed to completely simulate `blockhash` in a deterministic way. 
Is that not correct? Otherwise your PR should be modified to also return block hashes based on block index (up to last 2 epochs) on something like that.

joshuajbouw:
@ilblackdragon right but it was suggested to instead do `hash(account_id, block.hash)` as per @nearmax's comment given the rare situation where we could have a short-term NEAR fork.

ilblackdragon:
The problem is not in how to prevent hash of NEAR's block.
The problem is having hashes for the past blocks.

[...]

PR #4256, opened 2021-04-27
feat: add block hash API to near_vm_logic

joshuajbouw:
The PR is relatively straight forward, it just exposes two new methods `block_hash` and `prev_block_hash` to the VM logic, which may be useful.

I do need clarification that they work as intended, returning the latest block hash however, and not n-1.


core/primitives-core/src/types.rs
```diff
@@ -18,6 +18,9 @@ pub type Nonce = u64;
 pub type BlockHeight = u64;
 /// Height of the epoch.
 pub type EpochHeight = u64;
+/// Block hash.
+#[cfg(feature = "protocol_feature_vm_hash")]
+pub type BlockHash = [u8; 32];
```
frol, line 23: Why is it a new type instead of a `CryptoHash`?

joshuajbouw: Oh good catch. I did use it before finding `CryptoHash`. Forgot to remove it.

core/primitives-core/src/types.rs
```diff
@@ -18,6 +18,9 @@ pub type Nonce = u64;
 pub type BlockHeight = u64;
 /// Height of the epoch.
 pub type EpochHeight = u64;
+/// Block hash.
+#[cfg(feature = "protocol_feature_vm_hash")]
```
willemneal, line 22: Does a type alias need a feature check? It seems that it adds more complexity than its worth.

runtime/near-vm-logic/src/logic.rs
```diff
@@ -628,6 +630,28 @@ impl<'a> VMLogic<'a> {
         Ok(self.context.block_index)
     }
 
+    /// Returns the current block hash.
+    ///
+    /// # Cost
+    ///
+    /// `base`
+    #[cfg(feature = "protocol_feature_vm_hash")]
+    pub fn block_hash(&mut self) -> Result<BlockHash> {
```
willemneal, line 639: I'm a fan of full type path here so that you don't need the feature check at the top of the file.  But again it's a personal preference.

ilblackdragon: If we want to have block hashes - let's surface all the available block hashes - e.g. 2 last epochs.

core/primitives-core/src/types.rs
joshuajbouw: Fair point

runtime/near-vm-logic/src/logic.rs
joshuajbouw: Good call

Cargo.toml
```diff
@@ -117,6 +117,11 @@ protocol_feature_access_key_nonce_range = ["neard/protocol_feature_access_key_no
 protocol_feature_tx_size_limit = ["near-primitives/protocol_feature_tx_size_limit", "node-runtime/protocol_feature_tx_size_limit", "neard/protocol_feature_tx_size_limit"]
 protocol_feature_allow_create_account_on_delete = ["testlib/protocol_feature_allow_create_account_on_delete", "near-primitives/protocol_feature_allow_create_account_on_delete", "node-runtime/protocol_feature_allow_create_account_on_delete", "neard/protocol_feature_allow_create_account_on_delete"]
 protocol_feature_add_account_versions = ["near-primitives/protocol_feature_add_account_versions"]
+protocol_feature_vm_hash = [
+    "node-runtime/protocol_feature_vm_hash",
```
EgorKulikov, line 121: Seems like duplication

runtime/near-vm-logic/src/logic.rs
joshuajbouw: @ilblackdragon You are right, I do need to add in the method to get a block from an index. Ach.

EgorKulikov (approved):

Cargo.toml
joshuajbouw: I had realised that, but just entered them all anyways.

joshuajbouw:
I'll put this on WIP for now. We are working out on an alternative solution for EVM, and there is a change that needs to be made in order to call any block hash from a given number.

runtime/near-vm-logic/src/logic.rs
nearmax: Please sync with @bowenwang1996 to make sure we are not attempting to surface more block hashes than what is available with GC.

I also don't think surfacing more than let's say 10 block hashes is safe. We need to be super conservative with what we are exposing to contracts, because it cannot be undone. If we surface 2 epochs worth of hashes, then we cannot ever change the length of the epoch after that because some contracts may start breaking. And we did change the length of the epoch in past. We also do not want to surface number which is too large, because who knows how aggressive we would want our GC to be in the future. But surfacing something like 10 hashes seems arbitrary. Therefore I suggest we do not surface more than one or two hashes.

nearmax (changes requested):

nearmax (changes requested):
Also, we need to add it to imports: https://github.com/near/nearcore/blob/8d04bcb53f63ca72179efd57a280004340884686/runtime/near-vm-runner/src/imports.rs#L189

Then please also add it to runtime-params estimator (if you want in a separate PR to unblock you, since it is urgent). Please check with @olonho

runtime/near-vm-logic/src/logic.rs
artob: @nearmax I understand the concern. However, if we can't get at the last 256 hashes (needed by the [EVM's `BLOCKHASH` instruction](http://ethervm.io/)), then we will not be able to go with the scheme you outlined in https://github.com/near/nearcore/issues/3456#issuecomment-793048499, but will have to compute the EVM's block hashes based on the NEAR block number instead of the NEAR block hash. Sounds like that'll be the safer thing to do, despite the temporary fork concern you raised there.

bowenwang1996: I think exposing last 256 blocks should not be a problem, although it requires more effort in the implementation to make sure that it is not expensive. @artob is there any alternative solution to exposing the last 256 block hashes?

core/primitives-core/src/types.rs
ailisp: `type` in rust introduce alias, `pub type BlockHash = CryptoHash` plus `type CryptoHash = [u8;32]` will have equivalent semantic result as `type BlockHash=[u8;32]` but more robust if some day we change CryptoHash's definition

artob:
After calls with @nearmax, @bowenwang1996, @birchmd, and @djsatok today, the scope of this somewhat expanded. I'm reassigning to @birchmd given his previous Chain experience, and will write up an NEP detailing the current plan.

runtime/near-vm-logic/src/logic.rs
```diff
@@ -628,6 +630,28 @@ impl<'a> VMLogic<'a> {
...
+        Ok(self.context.block_hash)
+    }
+
+    /// Returns the previous block hash.
+    ///
+    /// # Cost
+    ///
+    /// `base`
+    #[cfg(feature = "protocol_feature_vm_hash")]
+    pub fn prev_block_hash(&mut self) -> Result<BlockHash> {
```
matklad, line 650: During discussion with nearmax, we realized that returning a `BlockHash` directly might not *just* work. All our other APIs only return primitive types (u64 integers), that's the first case where we want to return something bigger. Looks like we need either to use an out-parameter here, or [multiple return values](https://github.com/WebAssembly/multi-value), the latter approach being more promising imo.

birchmd: The pattern used in other examples, [such as `sha256`](https://github.com/near/nearcore/blob/d336b3fc1b9dc2167aa2eb9a89e24e4a2a09e27d/runtime/near-vm-logic/src/logic.rs#L901), is to use the register API, so I think we should probably stick to that for consistency. Though wasm handling multiple return values now is really cool!

matklad: Ah, I somehow miss the fact that we already do use out parameters. Clearly, we should use the same design here as well.

birchmd:
Closing this in favour of https://github.com/near/nearcore/pull/4273

PR #4262, opened 2021-04-29
Add config option for gas profiling

matklad:
This is an alternative version to #4185, which enables profiling via config flag.

This is not finished, but I am feeling like I am hopelessly bogged down in the weeds, so I am opening this PR to ask question mostly. 

What happens here is:

* remove compile-time flag for ProfileData
* cleanup it's interface somewhat (in particular, `ProfileData::non_zero_costs` is what we want to ship to users)
* add `gas_profiling` config and thread it to TrieViewer
* expose profile data for the **view** RPC calls

Questions I need help with, in order of priority:

* how do I expose this for non-view calls? Should I store profile data in the database (together with other outcome records)? Should we add a dedicated `replay` call, which works like view, but can reply things (without modifying the state)?
* what's our final strategy for enabling this functionalitty? Do we enable it by default, or do we have a runtime flag? 
* If we add a runtime flag, where it should be stored? at the moment it's a field of TrieViewer, but I don't feel like that's neccessary the right thing to do
* what's the best way to check that this works? The PR description doesn't mention TestPlan because I don't know what's the best place to test this functionality.

### DIFF
--- a/chain/chain/src/test_utils.rs
+++ b/chain/chain/src/test_utils.rs
@@ -847,10 +847,7 @@ impl RuntimeAdapter for KeyValueRuntime {
                 block_hash: *block_hash,
             }),
             QueryRequest::CallFunction { .. } => Ok(QueryResponse {
-                kind: QueryResponseKind::CallResult(CallResult {
-                    result: Default::default(),
-                    logs: Default::default(),
-                }),
+                kind: QueryResponseKind::CallResult(CallResult::default()),
                 block_height,
                 block_hash: *block_hash,
             }),
--- a/core/primitives-core/Cargo.toml
+++ b/core/primitives-core/Cargo.toml
@@ -24,7 +24,6 @@ lazy_static = "1.4"
 
 [features]
 default = []
-costs_counting = []
 protocol_feature_add_account_versions = []
 protocol_feature_evm = []
 protocol_feature_alt_bn128 = []
--- a/core/primitives-core/src/profile.rs
+++ b/core/primitives-core/src/profile.rs
@@ -1,3 +1,4 @@
+use std::collections::BTreeMap;
 use std::{cell::Cell, fmt};
 
 use crate::config::{ActionCosts, ExtCosts};
@@ -15,15 +16,9 @@ pub struct ProfileData {
 
 #[derive(Clone)]
 enum Repr {
-    #[cfg(feature = "costs_counting")]
-    Enabled {
-        data: std::rc::Rc<DataArray>,
-    },
+    Enabled { data: std::rc::Rc<DataArray> },
     Disabled,
 }
-
-#[cfg(not(feature = "costs_counting"))]
-const _ASSERT_NO_OP_IF_COUNTING_DISABLED: [(); 0] = [(); std::mem::size_of::<ProfileData>()];
 
 impl Default for ProfileData {
     fn default() -> ProfileData {
@@ -37,7 +32,15 @@ impl ProfileData {
     const LEN: usize = 1 + ActionCosts::count() + ExtCosts::count();
 
     #[inline]
-    #[cfg(feature = "costs_counting")]
+    pub fn new(enabled: bool) -> ProfileData {
+        if enabled {
+            ProfileData::new_enabled()
+        } else {
+            ProfileData::new_disabled()
+        }
+    }
+
+    #[inline]
     pub fn new_enabled() -> Self {
         // We must manually promote to a constant for array literal to work.
         const ZERO: Cell<u64> = Cell::new(0);
@@ -45,12 +48,6 @@ impl ProfileData {
         let data = std::rc::Rc::new([ZERO; ProfileData::LEN]);
         let repr = Repr::Enabled { data };
         ProfileData { repr }
-    }
-
-    #[inline]
-    #[cfg(not(feature = "costs_counting"))]
-    pub fn new_enabled() -> Self {
-        Self::new_disabled()
     }
 
     #[inline]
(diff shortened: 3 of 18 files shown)


bowenwang1996:
> expose profile data for the view RPC calls

I am not sure how useful this is given the limitation of view calls. Also if something can be done through a view call, usually there is no need to send a transaction to achieve the same thing. Therefore I don't see why we want to have profile data for view calls.

> how do I expose this for non-view calls? Should I store profile data in the database (together with other outcome records)? Should we add a dedicated replay call, which works like view, but can reply things (without modifying the state)?

That seems reasonable. Then we can potentially modify the `tx` rpc (which queries the status of a transaction) to include this data or add a new rpc endpoint for this purpose.

> what's our final strategy for enabling this functionalitty? Do we enable it by default, or do we have a runtime flag?

That seems reasonable to me. What's the reason for preferring runtime flag over compile-time flag though?

> If we add a runtime flag, where it should be stored? at the moment it's a field of TrieViewer, but I don't feel like that's neccessary the right thing to do

You can add that to `Runtime`

matklad:
> That seems reasonable. Then we can potentially modify the tx rpc (which queries the status of a transaction) to include this data or add a new rpc endpoint for this purpose.

Oh, I think I am being confusing here. I see two *different appraches*.

* Approach 1 is to store this information into the database on every transaction execution, and fetch it for RPC.
* Approach 2 is to extend our view rpc (or, more likely, add a separate RPC call) with a "dry re-run" mode, where we can re-run some existing transaction, but record profiling information this time. 

From your reply, I reason that approach 1 is reasonable (and it indeed seems simpler). I recall @frol having concerns that remembering this data for every function in a database might be too much data, but, if we are not going to do this for validators, than it's probably fine?

>I am not sure how useful this is given the limitation of view calls. 

Yeah, it was mostly "because I can", to get something working, rather than "because I should". 

>What's the reason for preferring runtime flag over compile-time flag though?

I assume we want to ship a single `neard` binary which you can run in validating or in non-validating mode? We *can* make this purely a compile-time flag, but that would be operationally more complicated it seems. Though, I don't have a super well-informed opinion here. 

>You can add that to Runtime

Is it really OK to add flags to Runtime? I feel like it is deliberatelly a zero-sized struct right now? Just want to clarify in case I misunderstand.

bowenwang1996:
> I recall @frol having concerns that remembering this data for every function in a database might be too much data, but, if we are not going to do this for validators, than it's probably fine?

How much data do you write per transaction? We already save execution outcomes for transactions. It doesn't seem like it would be a concern.

> Is it really OK to add flags to Runtime? I feel like it is deliberatelly a zero-sized struct right now? Just want to clarify in case I misunderstand.

I am not sure whether it is "deliberate". It is more like we don't see the need of adding any thing in the struct, but if there is a need, I don't see why we couldn't do that.

nearmax:
I am still confused why profiling is conditional on anything. It seems like [we agreed](https://github.com/near/nearcore/pull/4185) that it does cause performance overhead, and any kind of concerns that this opens angle of attack is in my view nitpicking, because it is not possible based on how runtime works. I think we are adding unnecessary logic to combat non-existing problems.

nearmax:
> I am not sure how useful this is given the limitation of view calls. Also if something can be done through a view call, usually there is no need to send a transaction to achieve the same thing. Therefore I don't see why we want to have profile data for view calls.

I agree, view calls are limited, e.g. they cannot access storage or create receipts (e.g. for cross-contract calls). We need to have a solution that works both for regular calls and view calls and then pipe this data as an additional field in `ExecutionOutcome` https://github.com/near/nearcore/blob/4088a98a5a6ba46b0580a5649779c6d9f31c2e8e/core/primitives/src/transaction.rs#L333 so it gets treated with exactly the same logic as we treat `FunctionCallErrorSer`.

> That seems reasonable. Then we can potentially modify the tx rpc (which queries the status of a transaction) to include this data or add a new rpc endpoint for this purpose.

Please see the above proposal to include it in `ExecutionOutcome`, it then will be automatically exposed in RPC.

> Should we add a dedicated replay call, which works like view, but can reply things (without modifying the state)?

I am strongly against replay calls, our existing architecture provides an easy out of the box solution by just threading gas breakdown into `ExecutionOutcome`.

> Approach 1 is to store this information into the database on every transaction execution, and fetch it for RPC.

This exactly what will happen if we thread gas breakdown into `ExecutionOutcome`.

> That seems reasonable to me. What's the reason for preferring runtime flag over compile-time flag though?

I am strongly against flags, as explained in [the above post](https://github.com/near/nearcore/pull/4262#issuecomment-833686945).

> How much data do you write per transaction? We already save execution outcomes for transactions. It doesn't seem like it would be a concern.

It will add 528 bytes per transaction (66 fees storing u64 each). Take a typical transaction: https://explorer.near.org/transactions/AH7EoC7ygFgvwwvCJMGtGdb4iuT4kvZg5nYcJcgyZVrj (not even a large function call).  You can fetch its entire status using:
```bash
http post https://rpc.mainnet.near.org jsonrpc=2.0 id=dontcare method=tx \
    params:='[ "AH7EoC7ygFgvwwvCJMGtGdb4iuT4kvZg5nYcJcgyZVrj", "sender.testnet"]'
```
which is 2205 bytes in JSON. Which is of-course not how indexer stores it, but given all the flattening that indexer is doing, which I doubt will be significant storage increase since indexer stores all kinds of stuff too.

More heavy transactions like https://explorer.near.org/transactions/C4CLNae3VxSPexmCnC7wymdzp3shcm1C2PsFovZWnZN1
are 50KB in JSON.

frol:
> It will add 528 bytes per transaction (66 fees storing u64 each).

That is good to know. My concern was not baked by this kind of data. My general feeling was that any overhead during transaction execution pulls us back on higher TPS goal. Also, I would not say that we need profiling for every single transaction, which brought me to idea that we can limit profiling to dry-run calls (re-execute individual receipts on demand)

matklad:
Closed in favor of #4185, next step is #4299

PR #4263, opened 2021-04-29
refactor: improve compilation time for neard tests

matklad:
See #4027 for discussion why this layout is better. TL;DR is that we
don't re-link neard repeatedly.

As a cherry on top, we now get to proprely share NodeCluster code
between the tests, instead of it being compiled afresh for each test.


bowenwang1996 (approved):

frol (approved):

miraclx (approved):

matklad:
Hehe, turns out that this is not trivial -- there was a bit of a global state, which caused interverence between the tests in the same process. The last commit removes that global state.

matklad:
@bowenwang1996 or @evgenykuzyakov , could you take another look at this commit: d8d0fd86a4624bcdd1a7654c268e9ec4a0275dc4

It's a bit more than "just move the code around", as we need to fix deps between the tests as well.

core/primitives/src/runtime/config.rs
```diff
@@ -26,6 +27,34 @@ pub struct RuntimeConfig {
...
+pub struct CurrentRuntimeConfig {
+    genesis_runtime_config: Arc<RuntimeConfig>,
+    with_lower_storage_cost: OnceCell<Arc<RuntimeConfig>>,
+}
+
+impl CurrentRuntimeConfig {
+    pub fn new(genesis_runtime_config: RuntimeConfig) -> Self {
+        Self {
+            genesis_runtime_config: Arc::new(genesis_runtime_config),
+            with_lower_storage_cost: OnceCell::new(),
```
mina86, line 44: Alternatively just go ahead and construct the alternative configuration. It would increase memory usage if the protocol feature is not used but simplify things (and minimally speed things up) for the case of the feature being enabled.

mina86:
> could you take another look at this commit: d8d0fd8 ?

For what its worth, LGTM.

core/primitives/src/runtime/config.rs
matklad: Yeah, that's an extremely reasonable suggestion, to say the least! There's really no need for any kind of laziness here.

bowenwang1996:
@matklad any objections to merging this PR?

matklad:
I still want to merge this, but the specific change no longer is the best way to fix this after https://github.com/near/nearcore/pull/4381/files#r657012158. Let me close this and create an issue instead.

PR #4264, opened 2021-04-29
doc: document protocol upgrade in nearcore

bowenwang1996:
Document the process of introducing and stabilizing a protocol change.

Test plan
---------
N/A


docs/protocol_upgrade.md
```diff
@@ -0,0 +1,150 @@
+## Protocol Upgrade
+
+This document describes the entire cycle of how a protocol upgrade is done, from the initial PR to the final release.
+It is important for everyone who contributes to the development of the protocol and its client(s) to understand this process.
+
+### Background
+
+At NEAR, we use protocol version to mean the version of the blockchain protocol and is separate from the version of some specific client (such as nearcore),
+since the protocol version defines the protocol rather than some specific implementation of the protocol.
+Our upgrade scheme dictates that protocol version X is backward compatible with protocol version X-1,
```
matklad, line 10: Could you clarify if we try to keep `X` and `X-2` compatible?

docs/protocol_upgrade.md
```diff
@@ -0,0 +1,150 @@
...
+            let _ = $current_protocol_version;
+            $non_feature_block
+        }
+    }};
+}
+```
+With this macro, the previous example can be simplified into
+```rust
+checked_feature!(“protocol_feature_evm”, EVM, {do_evm_stuff();}, {do_old_stuff();});
+```
```
matklad, line 124: To my taste, this is a bit too fine-grained and a bit too tied to specific API. I feel that this will get slightly wrong relatively quickly. 

I suggest trimming this section down to painting a big picture (we use compile-time features) and leaving *pointers* to more fine grained documentation ("see `ProtocolFeature` type and `checked_feature` enum for how to check features in code"). 

The specific API docs can then be moved to a doc string in an `.rs` file.

docs/protocol_upgrade.md
```diff
@@ -0,0 +1,150 @@
...
+cargo build -p neard --release --features nightly_protocol,<protocol_feature>
+```
+
+In practice, we have all nightly protocol features enabled for Nayduck tests and on betanet, which is updated daily.
+
+### Feature Stabilization
+
+When we are ready to make a new release, we examine which features should be stabilized and for those that should be, we
+remove the feature gating everywhere. One difference is that we now use this rule of the `checked_feature` macro:
```
matklad, line 143: This is a bit confusing. We still have protocol version checks, because we need to maintain behavior for old blocks. It's just that we remove cargo featrures. 

I think it maybe makes sense to call the two kinds of feature gating at the start of the document:

* for stable features, we want to check if the features is enabled for the block's protocol version
* for unstable features, we shoudl check if we are runing nightly binary as well. 

Hm, now that I've spelled this out, it seems like it would be useful to discuss where protocol version comes from. THat is, that we associated each block with a protocol version at which it was produced.

matklad (commented):
Lovely :heart_eyes: 

There's one doc writing tip I want to drop here. For my docs, I use a sentence-per-line format. That is, add a `\n` after every `.`.  This affects only the raw view, not the rendenred view. It makes the docs easier to diff and to edit. It also gives an immediate feedback about line lengths. See example here: https://raw.githubusercontent.com/rust-analyzer/rust-analyzer/master/docs/dev/style.md

I don't think we'll be able to consistenly use this style for near, but it's usuful to keep in mind as an option

matklad:
>Interesting. I was told completely the opposite back in school :)

I learned the opposite from asciidoc:

https://asciidoctor.org/docs/asciidoc-recommended-practices/#one-sentence-per-line

:-)

matklad (approved):
LGTM!

As a minor point, it usually is better to not @mention folks directly in the commit messages -- github may send *repeated* notifications for those messages if someone rebases the commit in their fork (GitHub will see a new hash and re-ping).

PR #4268, opened 2021-05-01
refactor: Upgrade actix-web

### LINKED ISSUE #4240: Update actix-web dependencies
We need to update actix-web dependencies to the latest beta releases and migrate from `actix-web::client` to `awc`, it seems `actix-web::client` has been dropped in the recent `actix-web` releases.

This will unblock us so we can address other issues with jsonrpc-client.

Hint: we need to update the dependencies in all the nearcore crates, and use `cargo deny --all-features check bans` to check sanity that we don't have duplicate dependencies (update `deny.toml` if necessary). Some actix dependencies are patched, so let me know if you bump into some actix dependency that needs to be bumped on my fork.

miraclx:
Fixes #4240

- Update `actix-web` to latest beta `v4.0.0-beta.6`
- Migrate to using `awc` directly instead of previously re-exported `actix_web::client`
- Pulled upstream updates for [`near/actix-extras`](https://github.com/near/actix-extras) and updated `actix-web` dependency on all crates (https://github.com/actix/actix-extras/compare/master...near:actix-web-4-beta.6)
  - Reimplemented cookie adapter (https://github.com/near/actix-extras/commit/f3e6f09c77d1b7cbc2f1160181b40754c15ef1bb) that was removed from `actix-http` (https://github.com/actix/actix-web/pull/2065)
- Forked [`paperclip`](https://github.com/near/paperclip) and updated `actix-web`, adding and defaulting to the `actix4` feature flag (https://github.com/near/paperclip/commit/b9589c6e520eeee4a07ed94f7e32c476990d6536)


Cargo.toml
```diff
@@ -70,8 +70,8 @@ neard = { path = "./neard"}
 
 testlib = { path = "./test-utils/testlib" }
 
-[replace]
-"ethereum-types:0.10.0" = { path = "./patches/ethereum-types-0.10.0-to-0.11.0" }
+[patch.crates-io]
+ethereum-types = { path = "./patches/ethereum-types-0.10.0-to-0.11.0" }
```
frol, line 74: It is weird if this works fine since it does not specify that it needs to only patch the 0.10.0 version :thinking:

frol (approved):
Great! We only need to make sure all the existing test pass. Let's run nayduck: https://github.com/near/nearcore/blob/master/nightly/nayduck.py

Just sign up on the nayduck website and run `python nightly/nayduck.py`

bowenwang1996 (commented):
@frol why do we want to upgrade to some beta version? Is it better to wait for a stable release?

chefsale (commented):
LGTM :) yes, only not sure if we want to use beta? Agree with @bowenwang1996.

miraclx:
NayDuck link: http://nayduck.eastus.cloudapp.azure.com:3000/#/run/1572

Cargo.toml
miraclx: Not a lot changes, I think. Plus, `[replace]` is [deprecated](https://doc.rust-lang.org/cargo/reference/overriding-dependencies.html#the-replace-section).

```patch
index 8fd5c4e1..d6bbb532 100644
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -70,8 +70,8 @@ neard = { path = "./neard"}
 
 testlib = { path = "./test-utils/testlib" }
 
-[replace]
-"ethereum-types:0.10.0" = { path = "./patches/ethereum-types-0.10.0-to-0.11.0" }
+[patch.crates-io]
+ethereum-types = { path = "./patches/ethereum-types-0.10.0-to-0.11.0" }
 
 [profile.release]
 lto = true        # Enable full link-time optimization.
index 17dbadda..bd315595 100644
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1694,7 +1694,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "53d4e679d6864bc26210feb5cf044e245741cd9d7701b35c00440a6e84d61399"
 dependencies = [
  "anyhow",
- "ethereum-types 0.10.0 (registry+https://github.com/rust-lang/crates.io-index)",
+ "ethereum-types 0.10.0",
  "hex",
  "serde",
  "serde_json",
@@ -1743,13 +1743,6 @@ dependencies = [
  "ethereum-types 0.11.0",
 ]
 
-[[package]]
-name = "ethereum-types"
-version = "0.10.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "05dc5f0df4915fa6dff7f975a8366ecfaaa8959c74235469495153e7bb1b280e"
-replace = "ethereum-types 0.10.0"
-
 [[package]]
 name = "ethereum-types"
 version = "0.11.0"
```

khorolets (approved):

frol:
@bowenwang1996 we are already on actix betas (that was to migrate to tokio 1.x). This upgrade needed to acquire some new configuration options of awc (actix client), such as configurable size of payload to fix jsonrpc-client, and potentially resolve telemetry issue.

bowenwang1996 (approved):

PR #4269, opened 2021-05-02
Optimise memory usage while loading genesis

posvyatokum:
1. Adding option of streaming records to Genesis
2. Applying genesis records in chunks (not only in case of streaming)


posvyatokum:
Probably should adjust the chunk size. With this patch, the node is starting in 15 minutes, without it -- in 8. As for memory, the new code consumes 3 GB (old code -- 14 GB), so there is room to go.

core/chain-configs/src/genesis_config.rs
```diff
@@ -175,11 +177,29 @@ impl From<&GenesisConfig> for EpochConfig {
...
+
 #[derive(Debug, Clone, Default, Serialize, Deserialize)]
 pub struct Genesis {
     #[serde(flatten)]
     pub config: GenesisConfig,
     pub records: GenesisRecords,
+    /// Genesis object may not contain records
+    /// In this case records can be found in records_file
+    /// The idea is that all records consume too much memory
+    /// So they should be processed in streaming fashion with stream_records_with_callback
```
matklad, line 200: Nit: let's add `.` to the sentences. I don't think proper grammar and punctuation are useful in and of themselves, however, (at least for myself), if you try to not just "add comments" but write proper sentences, the end result is better. This is basically a trick to make yourself be more thoughtful when translating your internal understanding of the problem to another human, who might lack the context.

core/chain-configs/src/genesis_config.rs
```diff
@@ -175,11 +177,29 @@ impl From<&GenesisConfig> for EpochConfig {
...
+pub enum GenesisRecordsFileType {
+    FullGenesis,
+    RecordsArray,
+}
+
+#[derive(Debug, Clone)]
+pub struct GenesisRecordsFile {
+    pub path: PathBuf,
+    pub file_type: GenesisRecordsFileType,
+}
```
matklad, line 190: Am I correct that `path` only makes sense when the type is `RecordsArray`? If that's the case, let's make illegal states unrepresentable: 

```suggestion
#[derive(Debug, Copy, Clone)]
pub enum GenesisRecordsFile {
    FullGenesis,
    RecordsArray { path: PathBuf },
}
```

core/chain-configs/src/genesis_config.rs
```diff
@@ -266,6 +286,93 @@ impl GenesisRecords {
...
+            }
+        }
+        Err(de::Error::custom("missing field: records"))
+    }
+}
+
+pub fn stream_records_from_records_file(reader: impl Read, mut callback: impl FnMut(StateRecord)) {
+    let mut deserializer = serde_json::Deserializer::from_reader(reader);
+    let records_processor = RecordsProcessor { sink: &mut callback };
+    deserializer.deserialize_seq(records_processor).expect("error while processing records");
```
matklad, line 366: I'd return an `serde_json::Result`  out of this function, rather than just unwrapping. It's useful when `.unwraps` are visible at the call-site, especially for such generic and re-usable functions. 

Generally speaking, internal unwraps are fine for such config loading code, but, if propagating errors is not to onerous, one might as well just do it.

core/chain-configs/src/genesis_config.rs
```diff
@@ -337,6 +457,33 @@ impl Genesis {
         hasher.process_genesis(self);
         hasher.finalize()
     }
+
+    pub fn stream_records_with_callback(&self, callback: impl FnMut(StateRecord)) {
```
matklad, line 461: Same here, I'd return an `io::Result` here  (you can convert `serde_json::Error` to `io::Error` https://docs.rs/serde_json/1.0.64/src/serde_json/error.rs.html#129-168)

core/chain-configs/src/genesis_config.rs
```diff
@@ -337,6 +457,33 @@ impl Genesis {
...
+                stream_records_from_genesis_file(reader, callback);
+            }
+            GenesisRecordsFileType::RecordsArray => {
+                stream_records_from_records_file(reader, callback);
+            }
+        }
+    }
+
+    pub fn process_records(&self, mut callback: impl FnMut(&StateRecord)) {
+        if !self.records.as_ref().is_empty() {
```
matklad, line 476: :bulb: invert if

One less `!` that way.

core/chain-configs/src/genesis_config.rs
matklad, line 461:
```suggestion
    fn stream_records_with_callback(&self, callback: impl FnMut(StateRecord)) {
```

Seems like it doesn't need to be `pub`?

neard/src/config.rs
```diff
@@ -1047,13 +1050,21 @@ pub fn download_genesis(url: &String, path: &PathBuf) {
...
+pub fn load_config_without_genesis_records(dir: &Path) -> NearConfig {
     let config = Config::from_file(&dir.join(CONFIG_FILENAME));
-    let genesis = if let Some(ref genesis_records_file) = config.genesis_records_file {
-        Genesis::from_files(&dir.join(&config.genesis_file), &dir.join(genesis_records_file))
-    } else {
-        Genesis::from_file(&dir.join(&config.genesis_file))
-    };
+    let genesis_config = GenesisConfig::from_file(&dir.join(&config.genesis_file));
+    let genesis_records_file =
+        if let Some(ref genesis_records_file_path) = config.genesis_records_file {
```
matklad, line 1057: As an alternative, to me this version seems simpler: 

```suggestion
        if let Some(genesis_records_file_path) = &config.genesis_records_file {
```

neard/src/genesis_validate.rs
```diff
@@ -1,104 +1,133 @@
...
-    );
-    assert!(!validators.is_empty(), "no validators in genesis");
-
-    let mut total_supply = 0;
-    let mut staked_accounts = HashMap::new();
-    let mut account_ids = HashSet::new();
-    let mut access_key_account_ids = HashSet::new();
-    let mut contract_account_ids = HashSet::new();
-    for record in genesis.records.0.iter() {
+pub struct GenesisValidator<'a> {
```
matklad, line 7:
```suggestion
struct GenesisValidator<'a> {
```

In general, I find it very helpful to keep visibility as low as possible -- it makes reading the code later easier, as you can see which is public interface, and which are impl details.

core/chain-configs/src/genesis_config.rs
matklad: Ah, reading more into this, I see that's not the case: `Path` might point either to the genesis file itself, or to a separaet file with records.

core/chain-configs/src/genesis_config.rs
```diff
@@ -266,6 +286,93 @@ impl GenesisRecords {
...
+                    return Ok(());
+                }
+                _ => {
+                    map.next_value::<IgnoredAny>()?;
+                }
+            }
+        }
+        Err(de::Error::custom("missing field: records"))
+    }
+}
```
matklad, line 361: I wonder if it makes sense to collapse these two to a single visitor, which can process either a map or a seq? Just so that we don't have two things and two public functions?

runtime/runtime/src/lib.rs
```diff
@@ -1294,62 +1297,125 @@ impl Runtime {
...
+    result: HashMap<String, u64>,
+    config: &'a StorageUsageConfig,
+}
+
+impl<'a> StorageComputer<'a> {
+    pub fn new(config: &'a RuntimeConfig) -> Self {
+        Self { result: HashMap::new(), config: &config.transaction_costs.storage_usage_config }
+    }
+
+    pub fn process_record<Record: Borrow<StateRecord>>(&mut self, record: Record) {
```
matklad, line 1312:
```suggestion
    pub fn process_record(&mut self, record: StateRecord) {
```

I see that this is a pre-existing pattern, but it's not great. Making function generic makes compile times worse, and makes the code harder to understand. 

`Borrow` is basically only needed when your are implementing a hash map or other associative container.

runtime/runtime/src/lib.rs
```diff
@@ -1294,62 +1297,125 @@ impl Runtime {
             })?;
         Ok(())
     }
+}
+
+struct StorageComputer<'a> {
+    result: HashMap<String, u64>,
+    config: &'a StorageUsageConfig,
+}
```
matklad, line 1305: Do we need this doer object? At least in the current code, seems like it can be just a function? Even when we add straming here, I feel we might get by with a function?

```rust
    /// It's okay to use unsafe math here, because this method should only be called on the trusted
    /// state records (e.g. at launch from genesis)
    pub fn compute_storage_usage(
        &self,
        genessis: &Genesis,
        config: &RuntimeConfig,
    ) -> HashMap<AccountId, u64> {
        let mut res = HashMap::new();
        genessis.process_records(|record| {
            ...
        });
        res
    }
```

More generally, a lot of code in this PR adds objects which shoudl be used according to a `create, process*, finalize` protocol. I think some of that is inherent to the problem, but it'd be cool to minimize this where possible.

matklad: Ah, we *do* need this for `GenesisStateApplier`... :thinking: yeah, don't have any suggestions on how to make this better...

runtime/runtime/src/lib.rs
```diff
@@ -1434,22 +1502,55 @@ impl Runtime {
...
+        tries: ShardTries,
+        shard_id: ShardId,
+        validators: &[(AccountId, PublicKey, Balance)],
+        genesis: &Genesis,
+        config: &RuntimeConfig,
+        shard_account_ids: HashSet<AccountId>,
+    ) -> StateRoot {
+        let mut genesis_state_applier =
+            GenesisStateApplier::new(tries, shard_id, validators, config);
+        genesis_state_applier.apply(genesis, shard_account_ids)
```
matklad, line 1553: Ok, at least it seems that *this* dosen't need to be a stateful object, as we apply all the stuff in one go anyway.

core/chain-configs/src/genesis_config.rs
matklad, line 461: Let's inline the function into the caller? At the moment, the public API looks like this:

```
    pub fn stream_records_with_callback(&self, callback: impl FnMut(StateRecord))
    pub fn process_records(&self, mut callback: impl FnMut(&StateRecord))
```

It's not really clear what's the differences between the two fns. Making one private (or maybe just getting rid of it altogether) makes the API much clearer.

core/chain-configs/src/genesis_config.rs
```diff
@@ -337,6 +457,33 @@ impl Genesis {
...
+            GenesisRecordsFileType::FullGenesis => {
+                stream_records_from_genesis_file(reader, callback);
+            }
+            GenesisRecordsFileType::RecordsArray => {
+                stream_records_from_records_file(reader, callback);
+            }
+        }
+    }
+
+    pub fn process_records(&self, mut callback: impl FnMut(&StateRecord)) {
```
matklad, line 475:
```suggestion
    pub fn for_each_record(&self, mut callback: impl FnMut(&StateRecord)) {
```

sounds like a bit more rustic name.

matklad (commented):
Left some Rust-related comments (although most of them are "oh, that's actually harder than I thought, I see now why you do this the way you did").

One thing I am worrying a lot is that now processing records might fail with an `io::Error` and we do not reflect that in the types. We just unwrap inside.I feel that's a significant change of geneissis contract, and that's its better to expose the fact we can fail at the call site. But I see how this will make this already quite a big change larger. I don't necessary want to block on this, but it does seems that silencing IO errors that way increases the technical debt substantionaly.

runtime/runtime/src/lib.rs
```diff
@@ -1294,62 +1297,125 @@ impl Runtime {
...
+
+        let (store_update, new_state_root) = self
+            .tries
+            .apply_all(&trie_changes, self.shard_id)
+            .expect("Failed to apply genesis chunk");
+        store_update.commit().expect("Store update failed on genesis initialization");
+        self.current_state_root = new_state_root;
+    }
+
+    fn apply_chunk(&mut self, genesis: &Genesis, chunk_account_ids: HashSet<&AccountId>) {
```
bowenwang1996, line 1405: Nomenclature: `chunk` could be confusing because in the context of nearcore, it usually refers to a shard block. I think it is better to use another name here.

runtime/runtime/src/lib.rs
posvyatokum: Yeah, for now StorageComputer stays

core/chain-configs/src/genesis_config.rs
posvyatokum: Made stream_records_with_callback private

core/chain-configs/src/genesis_config.rs
posvyatokum: I wasn't able to do it when I started writing this code, but now I tried again and everything worked. It's much simpler and prettier this way.

runtime/runtime/src/lib.rs
posvyatokum: Changed it to batch

runtime/runtime/src/lib.rs
posvyatokum: I'm not sure if I've done what you wanted, it still looks weird. Also this part of code moved to runtime/runtime/src/genesis.rs

core/chain-configs/src/genesis_config.rs
```diff
@@ -299,11 +375,20 @@ impl GenesisJsonHasher {
...
 impl Genesis {
     pub fn new(config: GenesisConfig, records: GenesisRecords) -> Self {
-        let mut genesis = Self { config, records, phantom: PhantomData };
+        let mut genesis =
+            Self { config, records, records_file: PathBuf::new(), phantom: PhantomData };
         genesis.config.total_supply = get_initial_supply(&genesis.records.as_ref());
         genesis
     }
 
+    pub fn new_as_is(
```
bowenwang1996, line 384: I don't think this is the best name :)

bowenwang1996 (approved):
What is the peak memory usage when neard loads testnet genesis after this change?

posvyatokum:
> What is the peak memory usage when neard loads testnet genesis after this change?

5 GB

core/chain-configs/src/genesis_config.rs
```diff
@@ -266,6 +274,74 @@ impl GenesisRecords {
...
+
+    fn deserialize<D>(self, deserializer: D) -> Result<Self::Value, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        deserializer.deserialize_seq(self)
+    }
+}
+
+fn stream_records_from_file(
```
EgorKulikov, line 336: Maybe this is logically static method of RecordsProcessor

frol:
@posvyatokum 

> Probably should adjust the chunk size. With this patch, the node is starting in 15 minutes, without it -- in 8. As for memory, the new code consumes 3 GB (old code -- 14 GB), so there is room to go.

I just started a fresh testnet node based on 1.20.0-rc.2 release (which includes this PR, I believe), and observed that it consumed 10GB of RAM on `neard init`. Is that expected?

PR #4270, opened 2021-05-03
fix(jsonrpc): Validators by epoch_id current epoch is not returning Unknown Epoch anymore

khorolets:
I've noticed that if I request `validators` method and pass `epoch_id` in params I'm getting `Unknown Epoch` if epoch I request is current one.

I thinks it has to work so this fix makes it possible.

The cause was in the epoch manager. If we request validators with epoch id we try to use `EpochSummary` which deals with "cached" epochs. But we request ongoing epoch which is not cached yet. So on the `view_client` I compare current epoch id with requested one and in case of match I act like it was a request for latest to trigger `EpochInfoAggregator` instead of `EpochSummary`

### DIFF
--- a/neard/tests/rpc_nodes.rs
+++ b/neard/tests/rpc_nodes.rs
@@ -10,7 +10,7 @@ use futures::future::join_all;
 use futures::{future, FutureExt, TryFutureExt};
 
 use near_actix_test_utils::spawn_interruptible;
-use near_client::{GetBlock, GetExecutionOutcome, TxStatus};
+use near_client::{GetBlock, GetExecutionOutcome, GetValidatorInfo, TxStatus};
 use near_crypto::{InMemorySigner, KeyType};
 use near_jsonrpc::client::new_client;
 use near_logger_utils::init_integration_logger;
@@ -19,7 +19,9 @@ use near_primitives::hash::{hash, CryptoHash};
 use near_primitives::merkle::{compute_root_from_path_and_item, verify_path};
 use near_primitives::serialize::{from_base64, to_base64};
 use near_primitives::transaction::{PartialExecutionStatus, SignedTransaction};
-use near_primitives::types::{BlockId, BlockReference, Finality, TransactionOrReceiptId};
+use near_primitives::types::{
+    BlockId, BlockReference, EpochId, EpochReference, Finality, TransactionOrReceiptId,
+};
 use near_primitives::views::{
     ExecutionOutcomeView, ExecutionStatusView, FinalExecutionOutcomeViewEnum, FinalExecutionStatus,
 };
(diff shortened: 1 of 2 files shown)


frol (approved):

matklad:
Is a test possible for this?

neard/tests/rpc_nodes.rs
```diff
@@ -1049,3 +1051,39 @@ fn test_check_tx_on_lightclient_must_return_does_not_track_shard() {
...
+                            final_block.header.epoch_id,
+                        )),
+                    })
+                    .await;
+
+                if let Ok(Ok(validators)) = res {
+                    assert_eq!(validators.current_validators.len(), 1);
+                    System::current().stop();
+                }
+                sleep(std::time::Duration::from_millis(500)).await;
```
matklad, line 1085: Urgh, it feel really wrong to add 0.5 seconds to the CI time just to test for an edge case. I wonder if 

> "Test suite time is bounded by available CPU resources (in parciucular, tests do not rely on timeouts for sychronization)"

would be a good OKR for after we upgrade actix to the non-beta version.

khorolets: Oh, let's assume it's leftover code, I don't see a reason to have loop there.

khorolets: But we don't have other a way to ensure the blockchain has really started yet, I hope we'll be able to avoid it in future.

matklad: Yeah, those things are always annoying and hard to design. Generally, you need some kind of a signal which can be emitted on readiness, and which you can listen to. That way, you can act immediately after the thing is ready, and not after some delay. This can either be some protocol level API (ie, there might be an rpc call is_ready, which returns the response only when the node is ready, whatever that means), or some internal observability. For example, we can pull the logs out out nodes until we noticed that all participants logged that they are ready.

bowenwang1996: Don't think you need to sleep 500 milliseconds here. You can wait until block height is higher than 0 and then do the query.

khorolets:
@bowenwang1996 @matklad Guys, could you help with failing `tests/sanity/backward_compatible.py`

From what I can gather it takes "stable" branch which is `1.19.0` at the moment and my branch to compare if tx sending result is similar. But it fails with

```
File "tests/sanity/backward_compatible.py", line 80, in main
    assert 'error' not in res, res
AssertionError: {'jsonrpc': '2.0', 'error': {'code': -32000, 'message': 'Server error', 'data': {'TxExecutionError': {'InvalidTxError': {'InvalidNonce': {'tx_nonce': 1, 'ak_nonce': 7000000}}}}}, 'id': 'dontcare'}
```

which happens right after 

```python
res = stable_node.send_tx_and_wait(tx, timeout=20)
```

I think there is ain't much I can do about it in my branch. 

Can someone help to figure out what is the reason of this error and how to fix it. Is it possible in this PR?

birchmd:
> @bowenwang1996 @matklad Guys, could you help with failing `tests/sanity/backward_compatible.py`
> 
> From what I can gather it takes "stable" branch which is `1.19.0` at the moment and my branch to compare if tx sending result is similar. But it fails with
> 
> ```
> File "tests/sanity/backward_compatible.py", line 80, in main
>     assert 'error' not in res, res
> AssertionError: {'jsonrpc': '2.0', 'error': {'code': -32000, 'message': 'Server error', 'data': {'TxExecutionError': {'InvalidTxError': {'InvalidNonce': {'tx_nonce': 1, 'ak_nonce': 7000000}}}}}, 'id': 'dontcare'}
> ```
> 
> which happens right after
> 
> ```python
> res = stable_node.send_tx_and_wait(tx, timeout=20)
> ```
> 
> I think there is ain't much I can do about it in my branch.
> 
> Can someone help to figure out what is the reason of this error and how to fix it. Is it possible in this PR?

I am also seeing this on my PR (#4273 ) as well. I was confused because I didn't think anything I did should have messed with the nonce, and having @khorolets encounter the same problem makes me feel like it is not something in my PR after all.

matklad:
Seems like #4277 should fix this?

khorolets:
> Seems like #4277 should fix this?

Thanks for the heads-up! It's working locally, updating the PR.

PR #4271, opened 2021-05-03
fix: Fixing lightclnt.py

SkidanovAlex:
The interface has changed, and the missing light client block now
returns {}, not None, adjusting the test

Also, removing checking storage, it makes the test slow, but the test heavily relies on block being produced slower than the checks we run.

Test plan
--------
http://nayduck.eastus.cloudapp.azure.com:3000/#/run/1563

### DIFF
--- a/pytest/tests/sanity/lightclnt.py
+++ b/pytest/tests/sanity/lightclnt.py
@@ -38,6 +38,9 @@ nodes = start_cluster(
     [["epoch_length", 6], ["block_producer_kickout_threshold", 40],
      ["chunk_producer_kickout_threshold", 40]], client_config_changes)
 
+for node in nodes:
+    node.stop_checking_store()
+
 started = time.time()
 
 hash_to_height = {}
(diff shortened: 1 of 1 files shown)


pytest/tests/sanity/lightclnt.py
```diff
@@ -56,7 +59,7 @@ def get_light_client_block(hash_, last_known_block):
     global block_producers_map
 
     ret = nodes[0].json_rpc('next_light_client_block', [hash_])
-    if ret['result'] is not None and last_known_block is not None:
+    if ret['result'] != {} and last_known_block is not None:
```
matklad, line 62:
```suggestion
    if ret['result'] and last_known_block is not None:
```

This'd be more Pythonic, but not sure if we use falsy idioms in our Python code.

SkidanovAlex: We generally don't have as defined guidelines for python code. I personally prefer explicit comparisons.

bowenwang1996 (approved):

PR #4273, opened 2021-05-04
Feat(runtime): block_hash host function

birchmd:
This PR is meant to supersede https://github.com/near/nearcore/pull/4256

The goal is to provide a `block_hash` host function in the Near runtime which can be used by the Aurora project to provide the `BLOCKHASH` op-code in the EVM.

Note: the implementation here is very simple (it essentially just reads the value directly out of the node's DB), which means the cost of this operation will be much higher than it is on Ethereum. We could optimize this in the future (e.g. via an in-memory structure that is maintained by the chain Client), which is why I chose to hide the details of the implementation behind the trait object. The fact that this implementation is naïve is likely fine because the `BLOCKHASH` op-code is not frequently used in real Ethereum contracts.

Testing Plan:
- Aurora tests (see https://github.com/birchmd/aurora-engine/blob/block-hash-testing/tests/test_block_hash.rs this won't make it into the main repo until after the feature is stabilized in nearcore, but the test is written at least)


runtime/near-vm-logic/src/logic.rs
```diff
@@ -696,6 +696,23 @@ impl<'a> VMLogic<'a> {
         Ok(self.current_storage_usage)
     }
 
+    /// Returns the block hash for the block at the given height (if any). It can
+    /// only return a result for the most recent 256 blocks.
+    ///
+    /// # Cost
+    ///
+    /// `base + storage_read_base + storage_read_key_byte * 8 + write_register_base + write_register_byte * 32
```
birchmd, line 704: @nearmax @olonho @matklad 

I'm interested if you have any comments on the cost here.
Since the simple implementation I wrote for the backend uses the `Store` directly I figured the storage read costs would be appropriate.

ailisp (approved):

core/primitives/src/runtime/apply_state.rs
```diff
@@ -41,4 +44,6 @@ pub struct ApplyState {
     pub evm_chain_id: u64,
     /// Data collected from making a contract call
     pub profile: crate::profile::ProfileData,
+    #[cfg(feature = "protocol_feature_block_hash_host_fn")]
```
EgorKulikov, line 47: Can you add description here?

runtime/near-vm-logic/src/logic.rs
```diff
@@ -696,6 +696,23 @@ impl<'a> VMLogic<'a> {
...
+    #[cfg(feature = "protocol_feature_block_hash_host_fn")]
+    pub fn block_hash(&mut self, block_height: u64, register_id: u64) -> Result<()> {
+        self.gas_counter.pay_base(base)?;
+        // we assume that under the hood this is a DB read, where the block height
+        // is the key, and it is 64 bits = 8 bytes
+        self.gas_counter.pay_base(storage_read_base)?;
+        self.gas_counter.pay_per_byte(storage_read_key_byte, 8)?;
+        let result = self.ext.block_hash(block_height)?.unwrap_or_default();
```
matklad, line 712: Hm, is this the best ABI possible? My understanding is that the hash might, or might not be there, so essentailly we are returning an `Option<Vec<u8>>`. But we encode optionnness in the length of the vector, so the user needs to explicitly check for it. 

I guess, at minimum we need to explicitly document "if the block's hash is unavailble, the returned vector has zero length". But maybe a more direct API is possible? Could we return a `bool` here, and only write a vector if the hash actually exists?

runtime/near-vm-logic/src/logic.rs
matklad: The reasoning here seems correct. Though, the specified formula is not exactly correct -- for the case where there isn't a hash, we don't pay `write_register_byte * 32` part, as we store zero-length vector.

runtime/near-vm-logic/src/mocks/mock_external.rs
```diff
@@ -211,6 +216,11 @@ impl External for MockedExternal {
     fn validator_total_stake(&self) -> Result<Balance> {
         Ok(self.validators.values().sum())
     }
+
+    #[cfg(feature = "protocol_feature_block_hash_host_fn")]
+    fn block_hash(&self, _block_height: BlockHeight) -> Result<Option<CryptoHash>> {
```
matklad, line 221: As an aletrnaive, you can avoid conditional imports and use fully qualified names here:

```suggestion
    fn block_hash(&self, _block_height: near_primitives_core::types::BlockHeight) -> Result<Option<near_primitives::hash::CryptoHash>> {
```

runtime/runtime/src/state_viewer/mod.rs
```diff
@@ -526,6 +538,8 @@ mod tests {
             cache: None,
             #[cfg(feature = "protocol_feature_evm")]
             evm_chain_id: 0x99,
+            #[cfg(feature = "protocol_feature_block_hash_host_fn")]
+            block_hash_provider: Arc::new(near_primitives::types::NullBlockHashProvider),
```
matklad, line 542: Unrelated to the PR, but it seems like these tests would benefit from refactoring the construction of the ViewApplyState to a sligle place. It feels like a fairly frequently changing thing, and updating all the call sites is tedious.

matklad:
>We could optimize this in the future (e.g. via an in-memory structure that is maintained by the chain Client), which is why I chose to hide the details of the implementation behind the trait object.

I personally try to avoid trait objects in Rust, they have two problems:

* they make the code harder to read as, when you are calling `.bar` on a `dyn Foo`, it's hard to figure out what actually happens there -- goto definition will bring you to the trait methond, not to the specific implementation. 
* in Rust specifically, they come with an ergonomic hit, as `dyn Trait` is an unsized type, so the caller hash to manage references or choose a box.

In this particular case, it seems that we maybe don't need polymorhism, and an abstract type wold  be sufficient? Ie, something like

```rust
pub struct BlockHashProvider {
    store: Arc<Store>,
    current_block_height: near_primitives::types::BlockHeight,
}

impl BlockHashProvider {
    pub fn new(store: Arc<Store>, current_block_height: near_primitives::types::BlockHeight) -> Self;
    pub fn block_hash(
        &self,
        height: near_primitives::types::BlockHeight,
    ) -> Result<Option<CryptoHash>, std::io::Error>;
}
```

If we want to optimize this in the future, we can just change how the `block_hash` function is implemented internally, by adding caches or what not.  Note that the `Store` iteslf is already polimorphic (it stores `dyn Database` inside, such non-virtual interface pattern is a great way to use `dyn` in Rust). So all the placese which use NullProvider can probably use the normal provider with the mock store they already use.

To be clear, this is not a blocking concern, the current version is fine with me!

core/store/src/lib.rs
```diff
@@ -516,6 +516,38 @@ impl CompiledContractCache for StoreCompiledContractCache {
...
+}
+
+#[cfg(feature = "protocol_feature_block_hash_host_fn")]
+impl near_primitives::types::BlockHashProvider for StoreBlockHashProvider {
+    fn block_hash(
+        &self,
+        height: near_primitives::types::BlockHeight,
+    ) -> Result<Option<CryptoHash>, std::io::Error> {
+        if height >= self.current_block_height
+            || height < self.current_block_height.saturating_sub(256)
```
EgorKulikov, line 543: Maybe we need to have a named constant for 256

EgorKulikov (approved):

runtime/near-vm-logic/src/logic.rs
birchmd: It's not returning a vector of 0 length. It always returns a `[u8; 32]` (i.e. 256-bit sequence), but in the case where the hash is missing it returns `[0u8; 32]`. This is still perhaps not the best way to do this because the consumer of the API still has to check that the result is non-zero. I like the suggestion of returning a `bool` and only writing to the register if the hash is present!

birchmd:
@matklad 

> In this particular case, it seems that we maybe don't need polymorhism, and an abstract type would be sufficient

I didn't know this was possible in Rust. I thought the only way to have abstract methods was via traits. In your example, where does the implementation of `block_hash` actually live? Or have I misunderstood, and the implementation would be there, but you mean that future optimizations would simply change that implementation in a way which is not visible outside the struct?

> all the places which use NullProvider can probably use the normal provider with the mock store they already use

I kind of like the explicit nature of NullProvider. It makes clear that I am not expecting the column in the DB to be present in the tests. Using the mock store would maybe give the false impression to someone looking at the code in the future that this host function should work in test environments.

That said, I agree trait objects are a little messy for the reasons you said and I would much prefer to get to a place where it is not needed. The reason I went for a trait object in the first place was to follow a similar pattern to `EpochInfoProvider`.

matklad:
> you mean that future optimizations would simply change that implementation in a way which is not visible outside the struct?

Yeah, that's what I mean. You are correct that traits are the only way to actually do polymorphism (there's also a [linker hack](https://github.com/rust-lang/rust/blob/88b99dec2a4d103b8153ca9300bf0fbebdf65bda/library/std/src/panicking.rs#L33-L52) you can use though -- there's a linker hack for everything).

>The reason I went for a trait object in the first place was to follow a similar pattern to EpochInfoProvider.

Yeah, and the EpochInfoProvider does this to cut the crate dependencies (as per comment), not to provide future extensibility. Cutting dependencies might be a good justification for introducing a dyn trait with a single impl. Though, I wonder if this is needen in this case? For `EpochInfoProvider`, the type that implements that is EpochManager, and Runtime doesn't have access to that. For BlockHeightProvider, we only need the Store, and it seems that Runtime can actually use it?

birchmd:
> For BlockHeightProvider, we only need the Store, and it seems that Runtime can actually use it?

That's true for now. What I was imagining for a future optimization would live in `chain` though, so then the trait object would be needed to break the dependency cycle. I was thinking that a caching layer in the Provider struct would not really do anything because I think it will be uncommon to hit `block_hash` with the same input more than once in a single chunk. So instead, an optimization would be to have an implementation which is entirely in-memory (which is fine since it only needs the last 256 heights) and is maintained by the chain logic (since it needs to know what the main chain is).

Although, maybe using the trait object now when it is not needed is a form of premature optimization, and it's not worth the complexity for something we may or may not do in the future. @bowenwang1996 what are your thoughts on this?

core/store/src/lib.rs
birchmd: Done in https://github.com/near/nearcore/pull/4273/commits/1c14806c9aeb06542c726d343ae83f369ccca553

runtime/near-vm-logic/src/mocks/mock_external.rs
birchmd: Done in https://github.com/near/nearcore/pull/4273/commits/a27e725e66f5c82b4f60635f3852b0f9a84d8c8e

runtime/near-vm-logic/src/logic.rs
birchmd: Done in https://github.com/near/nearcore/pull/4273/commits/401d95197344e4f28695d3283603911938bda108

core/primitives/src/runtime/apply_state.rs
birchmd: Done in https://github.com/near/nearcore/pull/4273/commits/c095edd5af317348d6d1249c6a193ae4c8f769bf

core/store/src/lib.rs
```diff
@@ -516,6 +516,43 @@ impl CompiledContractCache for StoreCompiledContractCache {
     }
 }
 
+/// Constant for the maximum number of heights back from the current height for which
+/// blockhashes will be served. (See usage in `BlockHashProvider` impl below).
+#[cfg(feature = "protocol_feature_block_hash_host_fn")]
+const MAX_HEIGHT_HISTORY: u64 = 256;
```
bowenwang1996, line 522: I think it is probably better to make this an associated constant of `BlockHashProvider`

runtime/near-vm-logic/src/logic.rs
bowenwang1996: I don't think `storage_read_base` is necessarily the best base cost here -- Trie reads do not directly translate to database reads, but I think it is also fine to overestimate here.

bowenwang1996 (commented):
Just want to make sure that we have actually agreed on this approach @artob @abacabadabacaba

core/store/src/lib.rs
birchmd: Unfortunately this does not work with the trait object approach we are currently using

```
error[E0038]: the trait `BlockHashProvider` cannot be made into an object
    --> core/primitives/src/utils.rs:373:21
     |
373  | impl fmt::Debug for dyn crate::types::BlockHashProvider {
     |                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ `BlockHashProvider` cannot be made into an object
     |
     = help: consider moving `MAX_HEIGHT_HISTORY` to another trait
note: for a trait to be "object safe" it needs to allow building a vtable to allow the call to be resolvable dynamically; for more information visit <https://doc.rust-lang.org/reference/items/traits.html#object-safety>
    --> core/primitives/src/types.rs:1000:11
     |
999  | pub trait BlockHashProvider {
     |           ----------------- this trait cannot be made into an object...
1000 |     const MAX_HEIGHT_HISTORY: u64 = 256;
     |           ^^^^^^^^^^^^^^^^^^ ...because it contains this associated `const`
```

Speaking of which, do you have comments on https://github.com/near/nearcore/pull/4273#issuecomment-832750663 ?

bowenwang1996:
> As far as I know this is Arto's preferred approach. However, it is up to you if you want an additional host function in nearcore, if not then the Aurora team will pursue another solution.

@artob can you confirm? I thought @abacabadabacaba has some different opinions.

ilblackdragon:
Just wanted to link why this is a usefull host function to have independent of Aurora - https://github.com/near/NEPs/issues/208

joshuajbouw:
@birchmd maybe its a good idea to revive this PR?

birchmd:
> @birchmd maybe its a good idea to revive this PR?

This PR was ready to go (obviously there are now conflicts that need to be resolved), and I was only waiting for the thumbs up from @artob and @bowenwang1996 . I am happy to put the work in resolving the conflicts if it is decided we do in fact want this PR to be merged.

bowenwang1996:
@artob @abacabadabacaba please take a look at this and voice your concerns if there are any

joshuajbouw:
@abacabadabacaba can you please review this?

nearmax:
I actually think that @ilblackdragon 's argument is an argument against implementation of this host function, see my reply here: https://github.com/near/NEPs/issues/208#issuecomment-894582803

nearmax (changes requested):
Since this functionality is not urgent and can be replicated without the host function, it is blocked on going through a proper quality control process described here https://gov.near.org/t/quality-control-for-protocol-api-changes/1941#the-format-4 , which requires submission of NEP that requires outlining alternatives and side-effects. Right now we have a meaningful push-back from @abacabadabacaba and @ilblackdragon 's argument is actually an argument against such host function, see my comment https://github.com/near/NEPs/issues/208#issuecomment-894582803.

joshuajbouw:
Yes this isn't required anymore. We found an alternative that is suitable without adding to nearcore.

joshuajbouw:
@birchmd probably should close this then

PR #4274, opened 2021-05-04
(fix) Fix storage usage bug with migration

EgorKulikov:
On some further though we are actually able to migrate this inside a single block

### Test plan
Sanity test in migrations.rs
Integration test in process_blocks


core/primitives/src/runtime/migration_data.rs
```diff
@@ -0,0 +1,39 @@
...
+    #[cfg(feature = "protocol_feature_fix_storage_usage")]
+    pub storage_usage_delta: Vec<(AccountId, u64)>,
+}
+
+impl MigrationData {
+    pub const EMPTY: MigrationData = MigrationData {
+        #[cfg(feature = "protocol_feature_fix_storage_usage")]
+        storage_usage_delta: Vec::new(),
+    };
+}
```
matklad, line 17: Seems like you can use derive(Default) here?

core/primitives/src/views.rs
```diff
@@ -1505,6 +1506,9 @@ impl From<StateChangeCause> for StateChangeCauseView {
             }
             StateChangeCause::UpdatedDelayedReceipts => Self::UpdatedDelayedReceipts,
             StateChangeCause::ValidatorAccountsUpdate => Self::ValidatorAccountsUpdate,
+            StateChangeCause::Migration { description } => {
+                Self::Migration { description: description.clone() }
```
matklad, line 1510: Seems like this'll work

```suggestion
                Self::Migration { description }
```

neard/src/runtime/mod.rs
```diff
@@ -417,6 +428,11 @@ impl NightshadeRuntime {
             #[cfg(feature = "protocol_feature_evm")]
             evm_chain_id: self.evm_chain_id(),
             profile: Default::default(),
+            migration_context: MigrationContext {
+                is_first_block_with_current_version: current_protocol_version
+                    != prev_block_protocol_version,
+                migration_data: self.migration_data.clone(),
```
matklad, line 434: Some style guides recommend spelling this as `Arc::clone(&self.migration_data)`, to make it clear that we are not deep-cloning large vectors here.

core/primitives/src/types.rs
```diff
@@ -152,6 +152,9 @@ pub enum StateChangeCause {
     /// State change that happens when we update validator accounts. Not associated with with any
     /// specific transaction or receipt.
     ValidatorAccountsUpdate,
+    /// State change that is happens due to migration that happens in first block of an epoch
+    /// after protocol upgrade
+    Migration { description: String },
```
matklad, line 157: Here, and in `StateChangeCauseView` I worry a bit that we add a free-form string description to a serialized field. This would be the first such string for both enums it seems. Perhaps future proof this a bit with:

```rust
enum StateChangeCauseView {
   Migration { id: MigrationId },
}

#[derive(Serialize,Deserialize)]
enum MigrationId { StorageUsageFix }

impl fmt::Display for MigrationId {}
```

That is, decouple what the migration is from a user-friendly text description.

runtime/runtime/src/lib.rs
```diff
@@ -1070,6 +1070,49 @@ impl Runtime {
...
+        state_update: &mut TrieUpdate,
+        apply_state: &ApplyState,
+    ) -> Result<Gas, StorageError> {
+        #[cfg(feature = "protocol_feature_fix_storage_usage")]
+        let mut gas_used: Gas = 0;
+        #[cfg(not(feature = "protocol_feature_fix_storage_usage"))]
+        let gas_used: Gas = 0;
+        #[cfg(feature = "protocol_feature_fix_storage_usage")]
+        if ProtocolFeature::FixStorageUsage.protocol_version()
+            == apply_state.current_protocol_version
```
Longarithm, line 1091: What do you think about this approach?
```
let gas_used: Gas = 0;
#[cfg(feature = "protocol_feature_fix_storage_usage")]
let gas_used = if ProtocolFeature::FixStorageUsage.protocol_version()
            == apply_state.current_protocol_version {
   ...
   Runtime::GAS_USED_FOR_STORAGE_USAGE_DELTA_MIGRATION;
}
```

EgorKulikov: It would work while there is single migration, but what if there's more?

core/primitives/src/runtime/migration_data.rs
EgorKulikov: Not working for const (although I used it for MigrationContext)

core/primitives/src/types.rs
```diff
@@ -152,6 +154,28 @@ pub enum StateChangeCause {
...
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        match self {
+            Self::StorageUsageFix => write!(f, "Storage usage fix"),
+        }
+    }
+}
+
+impl fmt::Debug for MigrationId {
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        (self as &dyn fmt::Display).fmt(f)
```
matklad, line 177: `wirite!(f, "{}", self)` or `fmt::Display::fmt(self, f)` would be a more direct way to write this.

runtime/runtime/src/lib.rs
Longarithm: Got it, thank you!

core/primitives/src/types.rs
EgorKulikov: I removed Display anyway as it is not really needed, only Debug

EgorKulikov: Apparently it is needed after all, done

core/primitives/src/types.rs
```diff
@@ -152,6 +154,28 @@ pub enum StateChangeCause {
...
     /// specific transaction or receipt.
     ValidatorAccountsUpdate,
+    /// State change that is happens due to migration that happens in first block of an epoch
+    /// after protocol upgrade
+    Migration { migration_id: MigrationId },
+}
+
+#[derive(Serialize, Deserialize, Clone, BorshSerialize, BorshDeserialize)]
+pub enum MigrationId {
+    StorageUsageFix,
```
bowenwang1996, line 164: Can we add a comment on what the fix is (maybe provide some link as well)?

neard/src/runtime/mod.rs
```diff
@@ -193,6 +197,11 @@ impl NightshadeRuntime {
         epoch_manager.get_epoch_info(&epoch_id).map(|info| info.epoch_height()).map_err(Error::from)
     }
 
+    fn get_epoch_id(&self, parent_hash: &CryptoHash) -> Result<EpochId, Error> {
```
bowenwang1996, line 200: `get_epoch_from_prev_block` is already a method in `RuntimeAdapter`

runtime/runtime/src/lib.rs
```diff
@@ -1070,6 +1072,48 @@ impl Runtime {
...
```
bowenwang1996, line 1091: should we move this logic to the construction of `ApplyState`?

runtime/runtime/src/lib.rs
```diff
@@ -1070,6 +1072,48 @@ impl Runtime {
...
+                        // of u64, and maximal delta is 4196, se we can add here without checking
+                        // for overflow
+                        account.set_storage_usage(account.storage_usage() + delta);
+                        set_account(state_update, account_id.clone(), &account);
+                    }
+                    // Account could have been deleted in the meantime
+                    None => {}
+                }
+            }
+            gas_used += Runtime::GAS_USED_FOR_STORAGE_USAGE_DELTA_MIGRATION;
```
bowenwang1996, line 1108: Do we have to hardcode this value instead of computing it as we apply the migrations?

neard/src/runtime/mod.rs
EgorKulikov: Yes, but I need epoch of prev block, and I do not have hash of block before prev

runtime/runtime/src/lib.rs
EgorKulikov: I am not really sure how. For my calculations I did everything through VMLogic

runtime/runtime/src/lib.rs
EgorKulikov: I do not think so, tbh, but can do that if you insist

runtime/runtime/src/lib.rs
```diff
@@ -1120,11 +1164,19 @@ impl Runtime {
             });
         }
 
+        let gas_used_for_migrations =
+            if apply_state.migration_context.is_first_block_with_current_version {
```
Longarithm, line 1168: Is it possible that this code will not be reached in the desired block because of this condition above?

`if !apply_state.is_new_chunk
            && apply_state.current_protocol_version
                >= ProtocolFeature::FixApplyChunks.protocol_version()`

AFAIU it means that the block doesn't have a presented chunk, which may be the case.

runtime/runtime/src/lib.rs
EgorKulikov: After some consideration I agree

EgorKulikov: After some more consideration I again think this is more appropriate place to select exact migration. On the other hand passing non-empty MigrationData only on new version blocks seems reasonable

EgorKulikov: See my new implementation

runtime/runtime/src/lib.rs
EgorKulikov: Thanks, fixed

core/primitives/src/types.rs
EgorKulikov: Done

neard/src/runtime/mod.rs
bowenwang1996: I see. Then `parent_hash` is not the best name :)

runtime/runtime/src/lib.rs
bowenwang1996: @EgorKulikov could you explain why you think it is better to put this logic here? My reasoning is as follows:
* If we move this logic to when `ApplyState` is constructed, all the logic for determining whether migration data should be populated will be in one place and it is easier to read and reason about.
* You don't need to have the `unreachable` branch here -- `None` would mean that no migration is needed.

tools/storage-usage-delta-calculator/src/main.rs
```diff
@@ -0,0 +1,39 @@
...
+use std::io::Error;
+
+/// Calculates delta between actual storage usage and one saved in state
+/// output.json should contain dump of current state,
+/// run state-viewer --home ~/.near/mainnet/ dump_state
+/// to get it
+fn main() -> Result<(), Error> {
+    env_logger::Builder::new().filter(None, LevelFilter::Debug).init();
+
+    debug!("Start");
```
bowenwang1996, line 16: Should we remove these?

EgorKulikov: I amnpersonally concerned when I start some tool and it seemingly do nothing for 15 minutes

EgorKulikov: For some reason it seems to be a single conversation and I cannot reply directly to comment above. Your approach may well be better if we only have one item in MigrationData, but how do you think we should deal with multiple such elements?

runtime/runtime/src/lib.rs
EgorKulikov: Your approach may well be better if we only have one item in MigrationData, but how do you think we should deal with multiple such elements?

neard/src/runtime/mod.rs
EgorKulikov: Fixed

runtime/runtime/src/lib.rs
EgorKulikov: Refactored apply_migrations for more clarity

bowenwang1996 (approved):

core/primitives/src/runtime/migration_data.rs
```diff
@@ -0,0 +1,20 @@
...
+use crate::types::AccountId;
+#[cfg(feature = "protocol_feature_fix_storage_usage")]
+use crate::types::Gas;
+use std::fmt;
+use std::fmt::{Debug, Formatter};
+
+#[derive(Default)]
+pub struct MigrationData {
+    #[cfg(feature = "protocol_feature_fix_storage_usage")]
+    pub storage_usage_delta: Vec<(AccountId, u64)>,
```
frol, line 11: I would prefer to have a struct with named arguments as `u64` is not really descriptive

frol (changes requested):
I don't see what prevents this migration from being applied on every single epoch start. Do I miss something?

Another corner case I have in mind is: what if the very first block on the epoch won't be produced? Will this still work fine? I am not very familiar with the block production code.

EgorKulikov:
> I don't see what prevents this migration from being applied on every single epoch start. Do I miss something?

migration_data in ApplyState in only Some when protocol version changes, not just epoch

EgorKulikov:
> Another corner case I have in mind is: what if the very first block on the epoch won't be produced? Will this still work fine? I am not very familiar with the block production code.

It follows same logic as changes to validators at the start of an epoch

core/primitives/src/runtime/migration_data.rs
EgorKulikov: If we use named structure in json this will lead to additional 70kb of data included executable

core/primitives/src/types.rs
```diff
@@ -152,6 +152,9 @@ pub enum StateChangeCause {
     /// State change that happens when we update validator accounts. Not associated with with any
     /// specific transaction or receipt.
     ValidatorAccountsUpdate,
+    /// State change that is happens due to migration that happens in first block of an epoch
```
frol, line 155: nit:

```suggestion
    /// State change that happens due to state-correction migration that happens in the first block of an epoch
```

Ideally, reference this PR as an example for the reader, so they don't need to do git-blame to look up what kind of "state-corrections" happen.

chain/client/tests/process_blocks.rs
```diff
@@ -43,6 +43,8 @@ use near_primitives::sharding::{EncodedShardChunk, ReedSolomonWrapper, ShardChun
 #[cfg(feature = "protocol_feature_block_header_v3")]
 use near_primitives::sharding::{ShardChunkHeaderInner, ShardChunkHeaderV3};
 
+#[cfg(feature = "protocol_feature_fix_storage_usage")]
```
olonho, line 46: Generally feel that we somewhat abuse Rust features here, don't we?

runtime/runtime/src/lib.rs
```diff
@@ -1070,6 +1072,39 @@ impl Runtime {
...
+        #[cfg(not(feature = "protocol_feature_fix_storage_usage"))]
+        let gas_used: Gas = 0;
+        #[cfg(feature = "protocol_feature_fix_storage_usage")]
+        if ProtocolFeature::FixStorageUsage.protocol_version() == protocol_version {
+            for (account_id, delta) in &migration_data.storage_usage_delta {
+                match get_account(state_update, account_id)? {
+                    Some(mut account) => {
+                        // Storage usage is saved in state, hence it is nowhere close to max value
+                        // of u64, and maximal delta is 4196, se we can add here without checking
+                        // for overflow
```
olonho, line 1092: Maybe checked add would be just easier than long comment why it's not needed :)?

chain/client/tests/process_blocks.rs
```diff
@@ -2267,6 +2275,135 @@ fn test_epoch_protocol_version_change() {
     assert_eq!(protocol_version, PROTOCOL_VERSION + 1);
 }
 
+#[test]
+#[cfg(feature = "protocol_feature_fix_storage_usage")]
```
bowenwang1996, line 2279: It is better to group the tests for some protocol feature in one mod so that the imports are cleaner, e.g.

```rust
#[cfg(test)]
mod protocol_feature_fix_storage_usage_tests {
   ....
}

**Review by bowenwang1996 — CHANGES_REQUESTED** [review]

File: chain/client/tests/process_blocks.rs [hunk 18]
```diff
@@ -2267,6 +2275,135 @@ fn test_epoch_protocol_version_change() {
...
+                .get(&TrieKey::Account { account_id: "test0".to_string() })
+                .unwrap()
+                .unwrap()
+                .clone();
+            let account_test0 = Account::try_from_slice(&mut account_test0_raw).unwrap();
+            assert_eq!(account_near.storage_usage(), if i >= 11 { 4560 } else { 364 });
+            assert_eq!(account_test0.storage_usage(), 182);
+        }
+    }
+    {
```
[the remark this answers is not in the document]
EgorKulikov: They are kinda the same, just with different chain ids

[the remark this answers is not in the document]
bowenwang1996: Then it makes sense to refactor the test to avoid code duplication. Something like
```rust
fn test_with_chain_id(chain_id: &str) {
  ....
}
```

runtime/runtime/src/lib.rs
```diff
@@ -1070,6 +1072,39 @@ impl Runtime {
...
+        state_update: &mut TrieUpdate,
+        migration_data: &Arc<MigrationData>,
+        protocol_version: ProtocolVersion,
+    ) -> Result<Gas, StorageError> {
+        #[cfg(feature = "protocol_feature_fix_storage_usage")]
+        let mut gas_used: Gas = 0;
+        #[cfg(not(feature = "protocol_feature_fix_storage_usage"))]
+        let gas_used: Gas = 0;
+        #[cfg(feature = "protocol_feature_fix_storage_usage")]
+        if ProtocolFeature::FixStorageUsage.protocol_version() == protocol_version {
```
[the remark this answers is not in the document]
bowenwang1996: I don't think we have. Just want to point out that so far we do not rely on that in our codebase. If we do want to rely on this assumption, we should at least add a comment to `PROTOCOL_VERSION`

chain/client/tests/process_blocks.rs
```diff
@@ -2267,6 +2275,135 @@ fn test_epoch_protocol_version_change() {
...
+
+            let validator_signer = InMemoryValidatorSigner::from_seed(
+                &"test0".to_string(),
+                KeyType::ED25519,
+                &"test0".to_string(),
+            );
+
+            block.mut_header().get_mut().inner_rest.latest_protocol_version =
+                ProtocolFeature::FixStorageUsage.protocol_version();
+            block.mut_header().resign(&validator_signer);
```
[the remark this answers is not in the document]
EgorKulikov: I need to update protocol version, so no, I cannot just use TestEnv::produce_block

chain/client/tests/process_blocks.rs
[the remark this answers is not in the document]
EgorKulikov: I did something like that, not sure it is more understandable though

chain/client/tests/process_blocks.rs
EgorKulikov: I moved this test to its own mod to simplify imports

chain/client/tests/process_blocks.rs
```diff
@@ -2879,3 +2878,107 @@ mod access_key_nonce_range_tests {
...
+            let mut block = env.clients[0].produce_block(i).unwrap().unwrap();
+
+            let validator_signer = InMemoryValidatorSigner::from_seed(
+                &"test0".to_string(),
+                KeyType::ED25519,
+                &"test0".to_string(),
+            );
+
+            block.mut_header().get_mut().inner_rest.latest_protocol_version =
+                ProtocolFeature::FixStorageUsage.protocol_version();
```
bowenwang1996, line 2930: Could you add a comment on the difference between this and `TestEnv::produce_block`?

core/primitives/src/version.rs
```diff
@@ -101,15 +101,19 @@ pub enum ProtocolFeature {
     TransactionSizeLimit,
     #[cfg(feature = "protocol_feature_allow_create_account_on_delete")]
     AllowCreateAccountOnDelete,
+    #[cfg(feature = "protocol_feature_fix_storage_usage")]
+    FixStorageUsage,
 }
 
 /// Current latest stable version of the protocol.
+/// Some features (e. g. FixStorageUsage) require that there is at least one epoch with exactly
+/// corresponding version
```
bowenwang1996, line 110:
```suggestion
/// the corresponding version
```

bowenwang1996 (approved):

PR #4278, opened 2021-05-06
fix: revert part of #4181 to classify FailedWithNoError as deterministic

matklad:


matklad:
cc @olonho

chefsale:
Can we just bump the version to: `1.19.9-rc.2`?

frol (approved):

PR #4279, opened 2021-05-06
fix: wasmer1_default feature actually selects wasmer1

matklad:
We use this feature in ./runtime/near-vm-runner/src/vm_kind.rs.

So, it should be declared in `vm_runner` Cargo.toml rather than in
`vm_logic`.


matklad:
I think I orginialy broke that in #4225

chefsale (approved):

EgorKulikov (approved):

bowenwang1996 (approved):

matklad:
My understanding is that we don't actually need this for 1.19.0 release, as #4278 fixed the problem. I am going to close this PR and open a new one against the master branch.

PR #4281, opened 2021-05-06
fix: wasmer1_default feature actually selects wasmer1

matklad:
We use this feature in ./runtime/near-vm-runner/src/vm_kind.rs.

So, it should be declared in `vm_runner` Cargo.toml rather than in
`vm_logic`.

This is #4279 submitted against master. 

Not sure how to make sure that similar errors don't happen in the future. The culprit here is that you can have `cfg(feature = "some-typo")`, and you won't see any warnings about this anywhere...

### DIFF
--- a/runtime/near-vm-logic/Cargo.toml
+++ b/runtime/near-vm-logic/Cargo.toml
@@ -37,10 +37,6 @@ protocol_feature_evm = ["near-primitives-core/protocol_feature_evm"]
 protocol_feature_alt_bn128 = ["bn", "near-primitives-core/protocol_feature_alt_bn128", "near-vm-errors/protocol_feature_alt_bn128"]
 protocol_feature_allow_create_account_on_delete = []
 
-wasmer0_default = []
-wasmtime_default = []
-wasmer1_default = []
-
 # Use this feature to enable counting of fees and costs applied.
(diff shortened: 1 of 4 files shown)


EgorKulikov (approved):

ailisp (approved):

ailisp:
I think we were having some vm-logic <> vm-runner refactor that forgot moving features (VMKind moving to vm-runner):
```
impl Default for VMKind {
    #[cfg(all(
        feature = "wasmer0_default",
        not(feature = "wasmer1_default"),
        not(feature = "wasmtime_default")
    ))]
    fn default() -> Self {
        VMKind::Wasmer0
    }
...
```
rust cannot detect this kind of problem and none of us catch it in PR review. Is there some tool or code pattern that can help us detect unexpected feature present / unexpected feature missing?

matklad:
I don't know a good simple immediate fix, longer term, I think the following two fixes might make sense:

* move config to run-time, or mostly runtime 
* change near workspace structure, so that it's a worksapce with a viretual manifest, such that featruse can be enabled globally (so that there's no need to forward features from one package to another package)

bowenwang1996 (approved):

olonho (approved):

PR #4286, opened 2021-05-07
test: add tests for FP conversions in WASM VM

olonho:


matklad:
One thing I am worried here is "how do we make sure these are all the tests we need?". Would this specific case be covered by some existing test suite? Quick googling revealed that this exists:

https://github.com/WebAssembly/testsuite/blob/01efde81028c5b0d099eb836645a2dc5e7755449/conversions.wast#L125-L132

Can we integrate this in our test suite?

ailisp (approved):

ailisp:
> One thing I am worried here is "how do we make sure these are all the tests we need?". Would this specific case be covered by some existing test suite? Quick googling revealed that this exists:
https://github.com/WebAssembly/testsuite/blob/01efde81028c5b0d099eb836645a2dc5e7755449/conversions.wast#L125-L132
Can we integrate this in our test suite?

Hmm for this test suite wasmer already include it in their tests:

https://github.com/wasmerio/wasmer/tree/master/tests/wast/spec

https://github.com/wasmerio/wasmer/blob/ac357e4a23f871c51adab75ae2e5defa964321df/tests/compilers/wast.rs#L22

PR #4287, opened 2021-05-07
feat(rpc): patch contract code/state, account/access key for sandbox node

ailisp:
This PR is both showing how sandbox node would be build as a different neard and also add a feature specific to sandbox node, that is patch state in rpc call. The rpc request format is providing state record to override with serde serialized StateRecords.

Test Plan
------------
Add chain test to show patch_state is effective in sandbox node by testing patch state can be than queried and used. 
Add pytest to show patch state rpc works


chain/client/tests/ganache.rs
```diff
@@ -0,0 +1,173 @@
...
+use near_primitives::utils::to_timestamp;
+use near_primitives::validator_signer::{InMemoryValidatorSigner, ValidatorSigner};
+use near_primitives::version::PROTOCOL_VERSION;
+use near_primitives::views::{
+    BlockHeaderView, FinalExecutionStatus, QueryRequest, QueryResponseKind,
+};
+use near_store::get;
+use near_store::test_utils::create_test_store;
+use neard::config::{GenesisExt, TESTING_INIT_BALANCE, TESTING_INIT_STAKE};
+use neard::NEAR_BASE;
```
matklad, line 66: Unrelated to the PR, but it seems that there's something wrong with our testing infra if adding a simple smoke test for new functionality requires more imports than fit on my screen. 

I think this smell points to the problem that we use *code* to setup the test. The better solution is usually to do setup by specifying *data* which is required, and having a single generic driver that interprets the data and sets up the in-memory objects. `NodeCluster` refactor was a great step in that direction:

https://github.com/near/nearcore/blob/0687126597dbf9c61af33ee270086d60dd645911/neard/tests/rpc_nodes.rs#L51-L57

chain/chain/src/chain.rs
```diff
@@ -2466,6 +2473,45 @@ impl Chain {
...
+                StateRecord::Data { account_id, data_key, value } => store_update.set(
+                    ColState,
+                    &(TrieKey::ContractData { account_id, key: data_key }.to_vec()),
+                    &value,
+                ),
+                _ => unimplemented!("patch_state can only patch Account, AccessKey, Contract and Data kind of StateRecord")
+            }
+        }
+        // store_update.set(ColStateParts, &key, &state_part);
+        store_update.commit().expect("Failed to patch state");
```
matklad, line 2510: Why is this an expect rather than `?`

chain/jsonrpc/src/lib.rs
```diff
@@ -958,6 +976,21 @@ impl JsonRpcHandler {
     }
 }
 
+#[cfg(feature = "ganache")]
+impl JsonRpcHandler {
+    async fn ganache_patch_state(&self, params: Option<Value>) -> Result<Value, RpcError> {
+        let records = parse_params::<Vec<StateRecord>>(params)?;
```
matklad, line 982: When designing RPC, it makes sense to think about forward compatibility. "If we are to add new parameter here, can we avoid breaking existing clients?". If what gets send on the wire is an array, then the answer is "no". For this reason it makes sense to introduce something like

```rust
#[derive(Serialize)]
pub struct RpcGanachePatchState {
    pub records: Vec<StateRecord>
}
```

A side benefit is that the API signature is clearly visible in the type defitniton, we don't need to hunt for it in method's implementation

chain/jsonrpc/src/lib.rs
```diff
@@ -230,6 +233,21 @@ impl JsonRpcHandler {
...
+            let res = match request.method.as_ref() {
+                // Ganache controls
+                "ganache_patch_state" => Some(self.ganache_patch_state(params).await),
+                _ => None,
+            };
+
+            if let Some(res) = res {
+                return res;
+            }
+        }
```
matklad, line 249: There are two existing code patterns this might be modeled after -- the `adversarial` block above, and the normal block bellow. What's bellow looks like a better model here, for two reasons:

* the code just reads more "production ready". For example, parameter parsing happens in the caller, which improves readability and better layers error
* once we stabilize this, we would want to move it to the bellow block, as that's the pattern we use for "production" requests.

Cargo.toml
```diff
@@ -126,3 +126,5 @@ wasmtime_default = ["node-runtime/wasmtime_default"]
 wasmer1_vm = ["node-runtime/wasmer1_vm"]
 wasmer0_vm = ["node-runtime/wasmer0_vm"]
 wasmtime_vm = ["node-runtime/wasmtime_vm"]
+
+ganache = ["neard/ganache", "near-jsonrpc/ganache"]
```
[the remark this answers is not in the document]
ailisp: No. ganache is a good name for now that straightforward to understand, and easy to search-replace once we have a better name

chain/chain/src/chain.rs
ailisp: Made it not return Result to be clear

chain/client/tests/ganache.rs
matklad: I usually copy-paste the code without imports and then <kbd>Alt+Enter</kbd> missing ones.

chain/client/tests/ganache.rs
```diff
@@ -0,0 +1,174 @@
...
+
+    for i in 9..20 {
+        last_block = env.clients[0].produce_block(i).unwrap().unwrap();
+        env.process_block(0, last_block.clone(), Provenance::PRODUCED);
+    }
+
+    let runtime_adapter = env.clients[0].runtime_adapter.clone();
+    let state2 =
+        query_state(&mut env.clients[0].chain, runtime_adapter.clone(), "test0".to_string());
+    println!("{:?}", state2);
```
matklad, line 173: Hm, I am having second thoughts about testing here. The current test has this problem. The "essence of test" are this lines:

```rust
        .patch_state(&[StateRecord::Data {
            account_id: "test0".to_string(),
            data_key: from_base64(&state[0].key).unwrap(),
            value: vec![40u8],
        }])

    let state2 =
        query_state(&mut env.clients[0].chain, runtime_adapter.clone(), "test0".to_string());
```

All the rest is setup boilerplate. If all the test would have this essense/boilerplate ratio, then adding new tests would be painfunl. This is OK for draft, but, before merging this, we need to make sure that:

* there are at least two tests
* common setup code is extracted into a function
* the tests themselves are relatively small. 

Rust-analyzer's integration test are a good model here:

https://github.com/rust-analyzer/rust-analyzer/blob/9a431c26f4528e2649de0ca171a38c93e473c94e/crates/rust-analyzer/tests/rust-analyzer/main.rs#L40-L73

Each test uses a DSL to setup context/submit quireis. The interpreter of the DSL is in a separate file. It's [rather gnarly](https://github.com/rust-analyzer/rust-analyzer/blob/master/crates/rust-analyzer/tests/rust-analyzer/support.rs#L21-L277), but it isn't repeated for every test.

chain/chain/src/chain.rs
```diff
@@ -2466,6 +2473,41 @@ impl Chain {
...
```
bowenwang1996, line 2507: I don't think this is what you want. This implementation will only change what is stored in the database but will not update trie. Also implementing this in `Chain` breaks the invariant that state update is only done in runtime. Even for features that are not enabled by default I think it is better to uphold those invariants.

ailisp: Yeah I just figured out it doesn't update trie so test didn't pass

> Also implementing this in Chain breaks the invariant that state update is only done in runtime. Even for features that are not enabled by default I think it is better to uphold those invariants.

Good point, I'm thinking of a better way to do this. By adding a ganache node only `state_patches` argument to `runtime.apply` that result in trie_update and state change. In chain just pass this argument to runtime. state_patches should not be considered as a type of transaction or receipt because it's fundamentally different. WDYT?

ailisp: I took a further look on this, maybe it's also okay to make patch_state a free, only allowed in ganache-node action in transaction. Because trie changes are applied to store on each block by `Chain::process_block_single`, so block is the single unit of work that caused a trie and state changed. Therefore state_patch should be part of a block and handled by `apply_chunks->apply_transactions->runtime.apply` that returns `{..., trie_changes, state_changes}` and resulting trie_changes commit by `Chain::process_block_single` makes patch state effective.
There're two ways to include a patch state in block: as a new field, or as a special kind of action in transactions.

chain/client/tests/ganache.rs
ailisp: The DSL like approach in rust-analyzer test looks near! I guess this DSL shall be more general to most of `chain/client/tests/*.rs`. patch_state specifically only need one test: init with some state => patch state => check state patch is effective. And it borrows the boilerplate from `chain/client/tests/process_blocks.rs` where many tests there are heavily boilerplated in a similar manner

chain/chain/src/chain.rs
ailisp: @bowenwang1996 Would like to see your thought on the potential impact of chain and keep most of the storage invariants on one of the two two approaches:

>  There're two ways to include a patch state in block: as a new field, or as a special kind of action in transactions.

bowenwang1996: Let's figure out why we need this operation first.

ailisp: We confirmed we do need this operation because ethereum developers use this feature in writing tests and they would expect this feature. And discussed offline for an implementation plan to store patch_state on chain instead of in block. Because we should avoid adding thing to block and transaction even when it's under a feature. And when client actor got a patch state request from rpc, it only has access to chain client and chain. So it can store this info in chain, I name the field Chain.states_to_patch , then in Chain::process_block_single it create a ChainUpdate and pass the states_to_patch to ChainUpdate, and ChainUpdate can pass states_to_patch to RuntimeAdapter::apply_transactions. So I'll add Chain.states_to_patch and ChainUpdate.states_to_patch field. 
This described is implemented, WIP clean up + test.

chain/client/tests/ganache.rs
matklad: Definitely agree that the problem is pre-existing, and that refactoring of existing tests should be done separately. 

However, I think we can de-boilerplate new tests without necessary touching the old ones. Doing something simple like just extracting the common `setup` function can be done independently. Extending that to support the rest of the tests could be done separately. 

I think that, long term, that's the less risky approach. If you have a bunch of tests you want to refactor to a DSL, there are two ways you can proceed:

* design the DSL upfront and re-write exciting tests to it in one go
* design DSL for *new* test, and iterate on it for a while; move the rest of tests later. 

Second one is usually better, as it allows for a small change and experience-based refinement of the design. 


The more general point here is that I am not a fan of merging functionality which isn't very easy to test. Psychologically, adding tests is not as fun as adding code, so I think it's important to make writing new tests easy, otherwise follow up code fixes might go untested. 

So, I don't think we need to go all in on designing a full test-framework here, but I do think that adding one more test for *this* specific functionality could be made easy by just exctracting some common code in this file.

ailisp: Sounds good! I'll try DSL for the new tests

core/primitives/src/views.rs
```diff
@@ -1474,13 +1474,25 @@ pub type StateChangesKindsView = Vec<StateChangeKindView>;
...
+    ReceiptProcessing {
+        receipt_hash: CryptoHash,
+    },
+    PostponedReceipt {
+        receipt_hash: CryptoHash,
+    },
     UpdatedDelayedReceipts,
     ValidatorAccountsUpdate,
+    #[cfg(feature = "ganache")]
+    PatchState,
```
bowenwang1996, line 1495: I don't think we need this here as I cannot see a use case where you need to query the causes of state changes when you run neard with `ganache`.

ailisp: This was for the sake of keep state change reason realistic, but yeah user run test on ganache won't really query reason for patch_state so we don't need this

runtime/runtime/src/lib.rs
```diff
@@ -1295,6 +1301,36 @@ impl Runtime {
...
+                    set_code(state_update, account_id, &code);
+                    assert_eq!(code.get_hash(), acc.code_hash());
+                }
+                StateRecord::AccessKey { account_id, public_key, access_key } => {
+                    set_access_key(state_update, account_id, public_key, &access_key);
+                }
+                _ => unimplemented!("patch_state can only patch Account, AccessKey, Contract and Data kind of StateRecord")
+            }
+        }
+        state_update.commit(StateChangeCause::PatchState);
```
ailisp, line 1331: @bowenwang1996 Although we may not need StateChangeCauseView, but we seem to need a reason for state_update.commit here, and that we need to add StateChangeCause::PatchState. Given StateChangeCause::PatchState need to be added in primitives, adding StateChangeCauseView::PatchState doesn't hurt

frol (changes requested):
I wonder if it is necessary given that you can construct genesis file with whatever state. Do you really need a hot-patch functionality?

I will properly review the code only after we decide whether we really need it.

ailisp:
> I wonder if it is necessary given that you can construct genesis file with whatever state. Do you really need a hot-patch functionality?
> 
> P.S. I will properly review the code after we decide whether we really need it.

I don't see patch-state on the fly, aka RPC absolutely necessary either, but it's an ask from DevX team for the reason "ethereum has equivalent feature in test tools: smoddit, and developer will ask for it"

Cargo.toml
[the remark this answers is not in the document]
ailisp: it's now named sandbox

chain/chain/src/chain.rs
```diff
@@ -2466,6 +2473,45 @@ impl Chain {
     }
 }
 
+/// Ganache node specific operations
+#[cfg(feature = "ganache")]
+impl Chain {
+    pub fn patch_state<Record: Borrow<StateRecord>>(
+        &mut self,
+        records: &[Record],
+    ) -> Result<(), Error> {
```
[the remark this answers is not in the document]
ailisp: it's now taking Vec<Record> because chain needs to take ownership of it.

chain/client/tests/ganache.rs
ailisp: Removed. Alt+Enter is quite helpful

chain/jsonrpc/src/lib.rs
ailisp: done

chain/jsonrpc/src/lib.rs
ailisp: done

chain/client/tests/ganache.rs
ailisp: Done

ailisp:
Major thing left: add pytest to test patch state rpc works

chain/client/tests/sandbox.rs
```diff
@@ -0,0 +1,131 @@
+#[cfg(test)]
+#[cfg(feature = "sandbox")]
+mod tests {
```
matklad, line 3: It's better to specify this in Cargo.toml via required features:

```
13:08:57|~/tmp/p|HEAD⚡?
λ bat -p Cargo.toml
[package]
name = "p"
version = "0.1.0"
edition = "2018"

[[test]]
name = "bar"
required-features = ["bar"]

[features]
bar = []

13:08:59|~/tmp/p|HEAD⚡?
λ exa -l tests/
.rw-r--r-- 0 matklad 20 May 13:06 bar.rs
.rw-r--r-- 0 matklad 20 May 13:06 foo.rs

13:09:01|~/tmp/p|HEAD⚡?
λ cargo test
    Finished test [unoptimized + debuginfo] target(s) in 0.00s
     Running unittests (target/debug/deps/p-e026952768e4795a)
     Running tests/foo.rs (target/debug/deps/foo-b7f734e93608b823)

13:09:05|~/tmp/p|HEAD⚡?
λ cargo test --features bar
    Finished test [unoptimized + debuginfo] target(s) in 0.00s
     Running unittests (target/debug/deps/p-3c96b03ceb15b495)
     Running tests/bar.rs (target/debug/deps/bar-47b80eb63f74c2b2)
     Running tests/foo.rs (target/debug/deps/foo-36c0da0c7c83c9a2)
```

chain/client/tests/sandbox.rs
```diff
@@ -0,0 +1,131 @@
...
+        test1.set_amount(10);
+
+        env.clients[0].chain.patch_state(vec![StateRecord::Account {
+            account_id: "test1".to_string(),
+            account: test1,
+        }]);
+        do_blocks(&mut env, 9, 20);
+        let test1_after = env.query_account("test1".to_string());
+        assert_eq!(test1_after.amount, 10);
+    }
```
matklad, line 130: :+1: reads much better now, thanks!

matklad:
It's unclear to me what's the status of the "do we need this?" discussion here, so let me add some checkboxes to make consensus more visible.

Does near-sandbox node (ganache equivalent) needs an RPC call to patch state of the chain on the fly (as opposed to just starting from a pre-determined, pre-patched genesis)? 

- [ ] @bowenwang1996 
- [ ] @DiscRiskandBisque

DiscRiskandBisque:
> It's unclear to me what's the status of the "do we need this?" discussion here, so let me add some checkboxes to make consensus more visible.
> 
> Does near-sandbox node (ganache equivalent) needs an RPC call to patch state of the chain on the fly (as opposed to just starting from a pre-determined, pre-patched genesis)?

I've seen this feature be useful in debuggers, but am unsure how it would be useful over other approaches in tests. To be clear, I'm making some assumptions:
1. Tests that need to return to the initial state can rewind rather than patch. I thought we have time travel or a way to snapshot/revert.
2. Tests that rely on state that is not initialized can fire txs in their testing framework's "beforeTest", "setup", etc. equivalent.

Happy to hear if I'm missing something. If not we could leave patching state for later.

Though I will add, since this is a PR: If it's already done then we might as well ship it.

ailisp:
> I thought we have time travel or a way to snapshot/revert.

We have an theoretically working way to do this, but in practice it's quite inconvenient and borying for developer to do so:
1. make a backup of `~/.near/data`
2. start node, run test that would change some state
3. stop node, revert `~/.near/data`, restart node for the next test.

So you must shutdown the node for revert, and right now you have to do 1-3 manually. Consider if we are about to have truffle equivalent `near test` command, then `near test` could do 1-3 for you underneath.

patch_state rpc provides something you can use in `beforeTest` or `beforeEach`, that's an alternative way of restarting node to run another test or `near test` script helper. This use case isn't absolutely necessary and still not as convenient to use. Most intuitive API for this would be implement `snapshot` and `revert` API, use them in `beforeEach`, as ganache did.

time travel is something different than snapshot/revert, we refer it as simulate the block timestamp on chain, for example, move forward 2 days to test some time-depend logic of contract.

Major use case for patch_state is @abacabadabacaba mentioned, as equivalent of ethereum's smoddit, which some ethereum developer use and expect on NEAR

chain/client/tests/sandbox.rs
```diff
@@ -0,0 +1,126 @@
...
+use near_chain::{ChainGenesis, Provenance, RuntimeAdapter};
+use near_chain_configs::Genesis;
+
+use near_client::test_utils::TestEnv;
+
+use near_crypto::{InMemorySigner, KeyType};
+
+#[cfg(not(feature = "protocol_feature_block_header_v3"))]
+#[cfg(feature = "protocol_feature_block_header_v3")]
+use near_primitives::sharding::{ShardChunkHeaderInner, ShardChunkHeaderV3};
```
EgorKulikov, line 14: What happened with this cfgs?

runtime/runtime/src/lib.rs
```diff
@@ -1343,6 +1352,36 @@ impl Runtime {
...
+        state_update: &mut TrieUpdate,
+        states_to_patch: Vec<StateRecord>,
+    ) {
+        for record in states_to_patch {
+            match record {
+                StateRecord::Account { account_id, account } => {
+                    set_account(state_update, account_id, &account);
+                }
+                StateRecord::Data { account_id, data_key, value } => {
+                    state_update.set(TrieKey::ContractData { key: data_key, account_id }, value);
```
EgorKulikov, line 1367: Why do we check existence of account for contract and not for data or access key?

state.json
```diff
@@ -0,0 +1,2 @@
+{"U1RBVEU=":"AgAAAAMAAABib2IFAAAAd29ybGQNAAAAcmVwcm8udGVzdG5ldAUAAABoZWxsbw"}
```
EgorKulikov, line 1: What is this and why it is in the root?

runtime/runtime/src/lib.rs
ailisp: Because for contract code we also update account's code hash. This is same as GenesisStateApplier::apply_batch

state.json
ailisp: sorry, added by mistake when resolve conflicts

chain/client/tests/sandbox.rs
ailisp: removed

bowenwang1996:
> We have an theoretically working way to do this, but in practice it's quite inconvenient and borying for developer to do so:
> 
> 1. make a backup of `~/.near/data`
> 2. start node, run test that would change some state
> 3. stop node, revert `~/.near/data`, restart node for the next test.
> 
> So you must shutdown the node for revert, and right now you have to do 1-3 manually. Consider if we are about to have truffle equivalent `near test` command, then `near test` could do 1-3 for you underneath.

How does this relate to state patching? My understanding is that in that rpc there is no way for you to delete state. Also if the goal is to revert to some previous state, it is easier to just revert to the state as of some previous block, which you can easily do by setting the new state root to some old state root.

ailisp:
> How does this relate to state patching? My understanding is that in that rpc there is no way for you to delete state. Also if the goal is to revert to some previous state, it is easier to just revert to the state as of some previous block, which you can easily do by setting the new state root to some old state root.

This can be done by call state patching. You don't need to remove state instead you do is update value of `STATE` key on the trie. I can add delete state in patch_state to cover remaining cases but patch_state is not intended for this use cases. It's for dynamic patch state in test in an adding/modifying manner, not for reverting state and do not need deleting. Reverting state as we both point out, it's easier to do by setting the new state root to older, and this should be triggered in a different sandbox RPC, I call it `sandbox_snapshot` and `sandbox_revert`. These are not part of this PR

runtime/runtime/src/lib.rs
frol: @ailisp it actually hurts when you want to have an exhaustive match for all the possible enum variants and now you have to deal with the case that cannot happen in production. Please, hide it behind a feature flag. Alternatively, I can suggest we abuse some other patch reason (`Migration` or `InitialState`).

EgorKulikov (approved):

bowenwang1996 (approved):
lgtm. Please address frol's comments

runtime/runtime/src/lib.rs
ailisp: Ah sorry I didn't see this comment. I could not figure out the feature magic to make `cargo check` and `cargo test --workspace --all-features` work at the same time, on the earlier commit where StateChangeCause::PatchState is under a feature. I'll abuse Migration reason

chain/jsonrpc/src/lib.rs
```diff
@@ -958,6 +968,23 @@ impl JsonRpcHandler {
...
+        &self,
+        patch_state_request: near_jsonrpc_primitives::types::sandbox::RpcSandboxPatchStateRequest,
+    ) -> Result<Value, RpcError> {
+        actix::spawn(
+            self.client_addr
+                .send(NetworkClientMessages::Sandbox(NetworkSandboxMessage::SandboxPatchState(
+                    patch_state_request.records,
+                )))
+                .map(|_| ()),
+        );
```
frol, line 983: Why do we want to spawn the task in background instead of awaiting till it completes before responding the client with any status?

chain/jsonrpc/src/lib.rs
```diff
@@ -958,6 +968,23 @@ impl JsonRpcHandler {
...
+        Ok(Value::String("".to_string()))
```
frol, line 984: Following the example of all the refactored JSON RPC methods, we should return some normal structure ([even if it is an empty struct](https://github.com/near/nearcore/blob/2c0c4973ea3b6730c40654b6299d599e2631e6d8/chain/jsonrpc-primitives/src/types/status.rs#L10-L11)), and then serialize it to the Value in `process_request` (in the future, we will write a macro attribute to avoid repetitive code in the `process_request`, but meanwhile, it is the contract we want to maintain in order to make things cleaner.

P.S. Empty string is a really unhelpful response (null is not helpful either, but it indicates that there is nothing to be reported back while string might be any string).

frol (approved):
Thanks for implementing it! :heart: 

Please, address the remaining nits in the JSON RPC implementation

chain/jsonrpc/src/lib.rs
ailisp: Yeah awaiting the change would be easier for use, I'm figuring out how to do it

chain/jsonrpc/src/lib.rs
ailisp: I see, good to know!

chain/jsonrpc/src/lib.rs
frol: Try this:

```suggestion
        self.client_addr
            .send(NetworkClientMessages::Sandbox(NetworkSandboxMessage::SandboxPatchState(
                patch_state_request.records,
            ))).await?;
```

chain/jsonrpc/src/lib.rs
```diff
@@ -958,6 +968,26 @@ impl JsonRpcHandler {
...
+            self.client_addr
+                .send(NetworkClientMessages::Sandbox(NetworkSandboxMessage::SandboxPatchState(
+                    patch_state_request.records,
+                )))
+                .map(|_| ()),
+        );
+        Ok(serde_json::to_value(
+            near_jsonrpc_primitives::types::sandbox::RpcSandboxPatchStateResponse {},
+        )
+        .unwrap())
```
frol, line 987: Please, see `health` endpoint as an example. Return the instance of the type itself:

```suggestion
    Ok(near_jsonrpc_primitives::types::sandbox::RpcSandboxPatchStateResponse {})
```

And in the `fn process_request` encode the result with `serde_json::to_value`, e.g., see `health` method:

https://github.com/near/nearcore/blob/322f6d80b9a6d0f15c3e374a5d368108166a2503/chain/jsonrpc/src/lib.rs#L278-L279

ailisp:
@frol now I followed how send_tx_commit did, wait pending patch state been applied to the block. However this is still not truly sync version of patch state, because patch state applied to the latest block, but view_function is based on final block, still need to wait 2s on client side. I'm trying to think of a deterministic way that "state patch has been applied and no wait is need on client side". Maybe let me revert patch_state to without polling patch state status, and add a patch_state_commit RPC in another PR

ailisp:
@frol made sandbox_patch_state rpc a sync operation now, at the time of rpc return, patch is already applied

ailisp:
Although the ci pass, `cargo build -p neard --features sandbox` fails since 0e45c2c, I should be add to ci

ailisp:
@chefsale @mhalambek @olonho Can you take a look and approve? Thanks

mhalambek (approved):

.gitignore
```diff
@@ -6,7 +6,7 @@ docker-build
 **/target/
 /storage/
 /test-utils/node/storage
-/test-utils/sandbox/storage
+/test-utils/ganache/storage
```
bowenwang1996: Why do we change it from sandbox to ganache?

ailisp: ganache is correct one: https://github.com/near/nearcore/blob/master/.gitignore#L9. I accidentally replaced it with sandbox so i double check every ganache->sandbox replace is doing correct.

chain/chain/src/chain.rs
```diff
@@ -2481,6 +2514,8 @@ pub struct ChainUpdate<'a> {
     genesis: &'a Block,
     #[allow(unused)]
     transaction_validity_period: BlockHeightDelta,
+    #[cfg(feature = "sandbox")]
```
olonho, line 2517: Guess general consensus was that we'll try to avoid Rust features whenever possible.
cc: @matklad

ailisp: I removed most of them on https://github.com/near/nearcore/pull/4287/commits/81fa0e065cde46ff941bdf0b5575e54080d6e2f4, but this place I think it's better to have sandbox feature. chain need a sandbox feature because Chain is the place of adding sandbox related features, Chain needs to know whether sandbox feature present and implement logic only when sandbox present, so it makes sense some fields of Chain is only available under sandbox.

tools/restored-receipts-verifier/src/main.rs
```diff
@@ -108,6 +108,7 @@ fn main() -> Result<()> {
                 *block.header().random_value(),
                 false,
                 false, // because fix was not applied in for the blocks analyzed here
+                None,
```
olonho, line 111: Another convention we seem to follow is to avoid having `Option` in input parameters list for APIs like that. Do we really need it?

ailisp: ah this one is hard to avoid. It was no argument in case of not `sandbox`, but the trouble of different signature under different feature was worse. So I move it to always have this argument, and it has to be an Option to differentiate with/without sandbox state patches.

chain/chain/src/chain.rs
olonho: As it's not core of this change, I'm OK putting feature for now, although still believe that in longer run it doesn't make sense to use features.

tools/restored-receipts-verifier/src/main.rs
olonho: OK, although it correlate with my another advice of not using features. Generally this way of coding prevents tooling from being able to behave properly. But would really love to hear @matklad opinion here.

chain/chain/src/chain.rs
```diff
@@ -2466,6 +2481,24 @@ impl Chain {
...
+        match self.pending_states_to_patch.take() {
+            None => self.pending_states_to_patch = Some(records),
+            Some(mut pending) => {
+                pending.extend_from_slice(&records);
+                self.pending_states_to_patch = Some(pending);
+            }
+        }
+    }
+
+    pub fn patch_state_finished(&self) -> bool {
```
matklad, line 2497:
```
    pub fn patch_state_in_progress(&self) -> bool {
        self.pending_states_to_patch.is_some()
    }
```

`finished` is a slight misnomer -- the function returns `true` even if patch state had never been scheduled in the first place.

chain/jsonrpc/src/lib.rs
```diff
@@ -958,6 +971,45 @@ impl JsonRpcHandler {
...
+                    SandboxResponse::SandboxPatchStateFinished(true),
+                )) = patch_state_finished
+                {
+                    break;
+                }
+                let _ = sleep(self.polling_config.polling_interval).await;
+            }
+        })
+        .await
+        .expect("patch state should happen at next block, never timeout");
```
matklad, line 1007: Hm, this gives me a pause. The problem with poling loops is that they are slow -- you generally add at least half of the polling interval to the waiting. This tends to kill test performance --- if each test has to wait 100 milliseconds, but you have a thousands of tests, your test suite spends over a minute waiting! 

This code is going to be hit by contract authors who write the tests. That is, it's not our test suite which is going to be slow (which we can live with), but the test suite of developers building on top of NEAR. 

Is there some existing issue which tracks the reason we do polling loops here? My understanding is that this all happens withing a single process/address spaces, so I don't see an architectural reason for us to poll via a loop.

chain/network/src/types.rs
```diff
@@ -1473,6 +1485,16 @@ pub enum NetworkClientResponses {
     DoesNotTrackShard,
     /// Ban peer for malicious behavior.
     Ban { ban_reason: ReasonForBan },
+
+    /// Sandbox controls
+    #[cfg(feature = "sandbox")]
+    SandboxResult(SandboxResponse),
```
matklad, line 1491: Position of `#[cfg(feature = "sandbox")]` in `NetworkClientMessages` and `NetworkClientResponses` is inconsistent. Let's keep it next to adversarial in both cases.

pytest/tests/sandbox/patch_state.py
```diff
@@ -0,0 +1,63 @@
...
+    "records": [{'Data': {
+        'account_id': "test0",
+        'data_key': base64.b64encode(k).decode('ascii'),
+        'value': base64.b64encode(new_v).decode('ascii'),
+    }}]
+})
+
+# patch should succeed
+res = nodes[0].call_function("test0", "read_value", base64.b64encode(k).decode('ascii'))
+assert(res['result']['result'] == list(new_v))
```
matklad, line 63: Sanity check: do we need modify some CI config somewhere to make this test work?

tools/restored-receipts-verifier/src/main.rs
matklad: The biggest smell I see here is not an addition of an argument per se, but rather the fact that this change requires us to update a dozen of unrelated tests. At least, my impression is that most of `None` come from tests. 

However, I also do wonder why do we add an extra parameter, rather than an extra method? That's, I'd expected this to work more like

```rust
diff --git a/chain/chain/src/chain.rs b/chain/chain/src/chain.rs
index 87b4eac7..719e3869 100644
--- a/chain/chain/src/chain.rs
+++ b/chain/chain/src/chain.rs
@@ -2911,13 +2911,13 @@ impl<'a> ChainUpdate<'a> {
                             *block.header().random_value(),
                             true,
                             is_first_block_with_chunk_of_version,
-                            #[cfg(feature = "sandbox")]
-                            self.states_to_patch.take(),
-                            #[cfg(not(feature = "sandbox"))]
-                            None,
                         )
                         .map_err(|e| ErrorKind::Other(e.to_string()))?;
 
+                    if let Some(states_to_patch) = self.states_to_patch {
+                        self.runtime_adapter.patch_state(states_to_patch)
+                    }
+
                     let (outcome_root, outcome_paths) =
                         ApplyTransactionResult::compute_outcomes_proof(&apply_result.outcomes);
 
```


I don't really understand the code in question though, so I am fine with the current impl.

chain/chain/src/chain.rs
```diff
@@ -203,6 +206,8 @@ pub struct Chain {
     /// Block economics, relevant to changes when new block must be produced.
     pub block_economics_config: BlockEconomicsConfig,
     pub doomslug_threshold_mode: DoomslugThresholdMode,
+    #[cfg(feature = "sandbox")]
+    pub pending_states_to_patch: Option<Vec<StateRecord>>,
```
matklad, line 210:
```suggestion
    pending_states_to_patch: Option<Vec<StateRecord>>,
```

`pub` means that the item is visible from outside of the crate. So, the absence of `pub` helps with understanding the code a lot, because, at a glance, you immediately know that anything that happens to an item happens in the current crate.

chain/chain/src/chain.rs
```diff
@@ -2077,6 +2090,8 @@ impl Chain {
             self.doomslug_threshold_mode,
             &self.genesis,
             self.transaction_validity_period,
+            #[cfg(feature = "sandbox")]
+            self.pending_states_to_patch.clone(),
```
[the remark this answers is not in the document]
ailisp: This was `take`. I was debugging apply pending patch state was not a synchronize operation and was suspecting the `take` happen in middle of another check. Finally i fixed patch state not sync issue in other place and didn't revert this back. I'll revert this to take and run pytest again to ensure it still works.

chain/network/src/types.rs
ailisp: Good catch

chain/jsonrpc/src/lib.rs
ailisp: No, this not the test noticably, patch_state have to happen in a block, which takes at least a second (block time). user expected to wait 2-3 seconds for block confirmation even this is without pool.
Also it's not possible to synchronously wait for chain to confirm "this state has been patched", current architecture rpc server can send NetworkMessage to Chain, chain doing the job (process blocks, patch states), but chain cannot notify back the operation is done if operation is async, such as submit a txn or patch state, both are pooled and effective later when block is produced. So, rpc server has to send another kind of message to Chain to pool this status.

pytest/tests/sandbox/patch_state.py
ailisp: Good point. This is not doable inside the repo, as we always assume binaries are built before test. I ping nightly instance maintainers to add a pre build step `make sandbox`.

tools/restored-receipts-verifier/src/main.rs
ailisp: Hmm did a try this is not trivial to add, apply_state must be part of the runtime_adapter.apply_transaction, (calls runtime.apply) because this one returns the ApplyResult, which is most importantly `{trie_changes, state_changes, proof}`. If do patch_state after apply_transaction, then trie_changes & proof need to be recalculated, which were calculated as last step in runtime.apply. So it's still most neat to do patch_state as part of runtime.apply. So I'll keep this as is

matklad (approved):

neard/src/main.rs
```diff
@@ -243,6 +243,19 @@ fn main() {
                 near_config.client_config.archive = true;
             }
 
+            #[cfg(feature = "sandbox")]
+            {
```
olonho, line 247: Well, this check is a bit fragile.

olonho (approved):
LGTM, modulo general discussion of getting rid of features.

neard/src/main.rs
ailisp: :) the purpose of the check is to avoid misoperation by not malicious user, sandbox node slip into mainnet/testnet is dangerous.

PR #4292, opened 2021-05-10
refactor(neard): decouple into individual crates

### LINKED ISSUE #4233: Decouple `nearcore` into a lib-crate
As part of the bigger initiative #1881, we want to break `neard` crate into two:

* `neard` binary
* `nearcore` library

We use `neard` as a library in Indexer Framework, and it might also improve testability and maintainability. This will also lay a path towards better dependencies handling.

miraclx:
Dissociates `neard` into its binary and library constituents, with the binary depending on the library. Fixes #4233.

The intent is to have the binary named `neard` and the library named `nearcore`. However, because the workspace is itself `nearcore`, we can either move the library into the root or use a different name for either the workspace or the library itself.


neard/bin/Cargo.toml
```diff
@@ -0,0 +1,55 @@
+[package]
+name = "neard-bin"
```
bowenwang1996, line 2: Let's keep using `neard` as the name of the binary. Otherwise a bunch of docs and code that rely on this will break

miraclx: Understood. And yeah, that's the plan.
Although within the workspace, a lot equally depends on `neard` as the lib name. With the _first pass_, it's like this for now while we make sure we have a proper structure for where we want the lib to go.

The options available are:
- Have the lib named `nearcore`, for which we have two options:
  - Move the lib into the root crate.
  - Rename the workspace.
- Rename the lib. I mentioned something like `libnear` but @frol rightly pointed out that seems more appropriate for a C interop we might not want to block by taking that name.

WDYT?

matklad:
>However, because the workspace is itself nearcore

I feel that that's unfortunate. The root of our workspace is a crate, but this crate only contains integration tests, it isn't a binary or a library. That's surprising (most big projects are set-up differently) and obscures the actual dependencies between the things. 

The typical setup is to make the root of the workspace a "virtual manifest", without the `[package]` section. See examples: [rust-analyzer](https://github.com/rust-analyzer/rust-analyzer/blob/master/Cargo.toml), [rustc](https://github.com/rust-lang/rust/blob/master/Cargo.toml), [substrate](https://github.com/paritytech/substrate/blob/master/Cargo.toml), [firecracker](https://github.com/firecracker-microvm/firecracker/blob/main/Cargo.toml). That way, you don't have "primary" crate, which often makes sense for large projects, as they have several entry points. `cargo test` in the workspace by default test everything. And you get somewhat more sane `--feature` flag behavior (with the second version of feature resolver, we should be able to even get rid of `foo = [bar/foo, baz/foo]` feature propagations).

Large project with non-virtual manifest typically have a top-level crate of substance, with `src` top-level directory which contains stuff, like [cargo](https://github.com/rust-lang/cargo).

So, if we want to introduce a genuine `nearcore` library, I would suggest to get rid of tests-only top-level package first. Between "let's make top-level `nearcore` a real library" and "let's have a virtual manifest", I think "virtual manifest" has more benefits, as we do have several entry points (neard, params estimator, tools like genesis-populate). 

I understand that this derails immediate work of extracing a library out of neard, but "fix workspace, *then* split neard" seems to be less work overall.

matklad:
Here's a my suggested plan for virtual manifest:

* check which tooling depends on the project layout and might break. Quick look at `pipeline.yml` shows that it should work with virtual manifest as is, but we might have some other automation.a
* move tests from `/test` folder somewhere -- either to some of the existing packages (neard perhaps?) or into a new `integration-tests` package
* re-read the features chapter of the cargo book, especially this section: https://doc.rust-lang.org/cargo/reference/features.html#resolver-version-2-command-line-flags
* check that features in top-level toml will work when we switch to virtual manifest (with resolevr = "2" that should be the case)
* add `[workspace] resolver = "2"` to the top-level toml
* remove `[package]`, `[dependencies]`, `[features]` section from top-level toml (`[profile]` should stay)
* check that everything works as expected. 

Is this reasonable? Let's do a checklist to make sure we decide one way or another:

Do you agree that we should move to virtual manifest setup *before* introducing proper `nearcore` library? 

- [ ] @bowenwang1996 
- [ ] @frol 

(tagging a couple of folks who I know have opinions about code organization)

frol:
I totally agree with the proposals from @matklad.

I don't expect any breakage from dropping `[package]` from the workspace Cargo.toml, and I even consider it is a great thing to do since it was quite confusing when you would call `cargo test` or `cargo build` which did nothing useful.

I suggest we move the current `tests/` into `nearcore` (lib crate) if possible, otherwise fall back to `neard` (bin).

frol (approved):
We are on the right track! Once CI is green, we can merge it

chefsale (approved):
LGTM

ailisp (approved):
Looks good!

.buildkite/pipeline.yml
```diff
@@ -124,11 +124,12 @@ steps:
   #  artifact_paths:
   #  - "rainbow-bridge/logs/**/*.log"
 
-  - label: "cargo release check"
+  - label: "cargo check nearcore library (without Cargo.lock)"
```
matklad, line 127: Thanks for clarifying what this does!

chain/client/Cargo.toml
```diff
@@ -6,7 +6,7 @@ edition = "2018"
 
 [dependencies]
 ansi_term = "0.12"
-actix = "0.11.0-beta.2"
+actix = "=0.11.0-beta.2"
```
matklad, line 9: :+1: it's good idea to pin pre-releases which are used in production.

Unrelated to this PR, but I've noticed that we need to update this in a lot of Cargo.tomls. It seems that some deps shouldn't be there? For example, the one in chain/chunks seems to be removable.

neard/Cargo.toml
```diff
@@ -5,61 +5,41 @@ authors = ["Near Inc <hello@nearprotocol.com>"]
 edition = "2018"
 default-run = "neard"
 
+[[bin]]
+path = "src/main.rs"
+name = "neard"
+
+[[bin]]
+path = "src/../src/main.rs"
+name = "near"
```
matklad, line 14: Unrelated to the PR, but there's a chance we no-longer need this work-around. If that's the case, we can remove both bins.

neard/Cargo.toml
```diff
@@ -76,14 +56,3 @@ protocol_feature_restore_receipts_after_fix = ["near-primitives/protocol_feature
 protocol_feature_cap_max_gas_price = ["near-primitives/protocol_feature_cap_max_gas_price", "near-chain/protocol_feature_cap_max_gas_price"]
 nightly_protocol_features = ["nightly_protocol", "near-primitives/nightly_protocol_features", "near-client/nightly_protocol_features", "near-epoch-manager/nightly_protocol_features", "near-store/nightly_protocol_features", "protocol_feature_evm", "protocol_feature_block_header_v3", "protocol_feature_alt_bn128", "protocol_feature_add_account_versions", "protocol_feature_tx_size_limit", "protocol_feature_allow_create_account_on_delete", "protocol_feature_fix_storage_usage", "protocol_feature_restore_receipts_after_fix", "protocol_feature_cap_max_gas_price"]
 nightly_protocol = ["near-primitives/nightly_protocol", "near-jsonrpc/nightly_protocol"]
```
matklad, line 58: We probably want all this features to be 

```
feature_name = ["nearcore/feature_name"]
```

This can be landed in a follow up PR, but it definitely has to happen as soon as possible, otherwise feature lists will diverge.

matklad (approved):

neard/Cargo.toml
frol: Good point, but I want this PR to keep things identical to the previous behavior wherever possible, so we don't spend time on unrelated discussions and don't revert the whole refactoring if we happen to decide to revert this change for some reason. Indeed, I plan to submit a PR dropping the old `near` binary.

khorolets (approved):
LGTM!

bowenwang1996 (approved):

Cargo.toml
```diff
@@ -101,38 +73,3 @@ opt-level = 3 # BLS library is too slow to use in debug
...
-protocol_feature_add_account_versions = ["near-primitives/protocol_feature_add_account_versions"]
-protocol_feature_fix_storage_usage = ["near-primitives/protocol_feature_fix_storage_usage", "neard/protocol_feature_fix_storage_usage", "node-runtime/protocol_feature_fix_storage_usage"]
-protocol_feature_restore_receipts_after_fix = ["near-primitives/protocol_feature_restore_receipts_after_fix", "near-chain/protocol_feature_restore_receipts_after_fix", "neard/protocol_feature_restore_receipts_after_fix", "node-runtime/protocol_feature_restore_receipts_after_fix", "restored-receipts-verifier/protocol_feature_restore_receipts_after_fix"]
-protocol_feature_cap_max_gas_price = ["near-primitives/protocol_feature_cap_max_gas_price", "neard/protocol_feature_cap_max_gas_price", "near-chain/protocol_feature_cap_max_gas_price"]
-
-# enable this to build neard with wasmer 1.0 runner
-# now if none of wasmer0_default, wasmer1_default or wasmtime_default is enabled, wasmer0 would be default
```
olonho, line 130: Those features are needed, why they are removed?

matklad: Good catch, I didn't notice that! I think it'd be good indeed to move this over to `neard/Cargo.toml` instead of just deleting. 

That being said, I wouldn't block on that, as the following now works without feature forwarding:

```
cargo build -p neard --features node-runtime/wasmer1_default
```

frol: @olonho Great catch!

It seems it should be in both `nearcore` (defined as `node-runtime/wasmer1_default`) and in `neard` (defined as `nearcore/wasmer1_default`).

olonho (approved):

PR #4293, opened 2021-05-10
fix: disable threading on wasmer 1

matklad:
Wasmer 1 uses rayon for parallel execution, which we don't want to use.
We've upstreammed the patch making rayon support optional, but it is not
yet released. So let's use our fork at the moment, which patches the
existing 1.0.2 release we use.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -72,6 +72,10 @@ testlib = { path = "./test-utils/testlib" }
 
 [patch.crates-io]
 ethereum-types = { path = "./patches/ethereum-types-0.10.0-to-0.11.0" }
+# Temporary wasmer fork which disables rayon (parallel compilation)
+# https://github.com/wasmerio/wasmer/pull/2262
+# Revome this when the new version of wasmer is published.
+wasmer-compiler-singlepass = { git = "https://github.com/near/wasmer", branch = "1.0.2-single-threaded"}
 
 [profile.release]
 lto = true        # Enable full link-time optimization.
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6264,15 +6264,13 @@ dependencies = [
 [[package]]
 name = "wasmer-compiler-singlepass"
 version = "1.0.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "426ae6ef0f606ca815510f3e2ef6f520e217514bfb7a664defe180b9a9e75d07"
+source = "git+https://github.com/near/wasmer?branch=1.0.2-single-threaded#a98937b4e9e149e4a7773689313b7e43800f6eb9"
 dependencies = [
  "byteorder",
  "dynasm",
  "dynasmrt",
  "lazy_static",
  "more-asserts",
- "rayon",
  "serde",
  "smallvec",
  "wasmer-compiler",
--- a/runtime/near-vm-runner/Cargo.toml
+++ b/runtime/near-vm-runner/Cargo.toml
@@ -20,7 +20,7 @@ wasmer-runtime = { version="0.17.1", features = ["default-backend-singlepass"], 
 wasmer-runtime-core = {version = "0.17.1", package = "wasmer-runtime-core-near" }
 wasmer = { version = "1.0.2", optional = true }
 wasmer-types = { version = "1.0.2", optional = true }
-wasmer-compiler-singlepass = { version = "1.0.2", optional = true }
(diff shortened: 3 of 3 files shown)


Cargo.toml
```diff
@@ -72,6 +72,10 @@ testlib = { path = "./test-utils/testlib" }
 
 [patch.crates-io]
 ethereum-types = { path = "./patches/ethereum-types-0.10.0-to-0.11.0" }
+# Temporary wasmer fork which disables rayon (parallel compilation)
+# https://github.com/wasmerio/wasmer/pull/2262
+# Revome this when the new version of wasmer is published.
+wasmer-compiler-singlepass = { git = "https://github.com/near/wasmer", branch = "1.0.2-single-threaded"}
```
matklad, line 78: Should I publish this as `wasmer-compiler-singlepass-near`, or are git deps ok?

ailisp: @willemneal

ailisp: @matklad Does this also patches wasmer-, wasmer-types, etc. to this branch?

willemneal: git deps are okay at the top level since it isn't published to crates.io.

ailisp: > so all other crates in the crate graph will depend on this patched version

not sure this is sufficient, because crate like wasmer-types are not depend on wasmer-compiler-singlepass. Might be safer to just patch every `wasmer-*` 1.0.2

olonho (commented):
Practically, could we consider always use Near owned Wasmer repo forks? It seems like right thing to do in general.

Cargo.toml
matklad: My PR for wasmer is based on the master branch. We *could* use master, but, imo, that's risky -- unlike specific 1.0.2 version, wasmer didn't promise that some random commit from master is stable enough for general use. 

So what I did in our repo is I checked out the 1.0.2 tag, and rebased my patch on top that. Patching specific tags I think is a reliable strategy: we get assuarance from wasmer that 1.0.2 is stable, and we already did some of our own testing.

We can switch other crates to use my branch, but that wouldn't meaningfully change anything. The branch contains exactly the same code as the released 1.0.2 wasmer crates, except for the single patched crate:

```
12:56:56|~/projects/wasmer|1.0.2-single-threaded✓
λ git log -n 4
commit a98937b4e9e149e4a7773689313b7e43800f6eb9 (HEAD -> 1.0.2-single-threaded, near/1.0.2-single-threaded)
Author: Aleksey Kladov <aleksey.kladov@gmail.com>
Date:   Mon May 10 18:10:18 2021 +0300

    Make fork usable with patch.crates-io syntax

commit aa093a299671d3c107a3161b738e08269445d60e
Author: Aleksey Kladov <aleksey.kladov@gmail.com>
Date:   Mon Apr 26 18:12:13 2021 +0300

    Make threading optional in single pass backend

commit 7cc7dd2dad291ba99d0dd1f1f5ba8bb87912691c
Author: Aleksey Kladov <aleksey.kladov@gmail.com>
Date:   Mon Apr 26 18:03:22 2021 +0300

    Prepare to abstract rayon away by minimizing the API we use

commit 8b2c3852b66e4eaf0267762632aeb78c42632449 (tag: 1.0.2)
Merge: 77f2e8547 fb429e0a6
Author: bors[bot] <26634292+bors[bot]@users.noreply.github.com>
Date:   Fri Feb 5 03:57:43 2021 +0000

    Merge #2094
    
    2094: Prepare for 1.0.2 release r=nlewycky a=nlewycky
    
    # Description
    Prepare for 1.0.2 release
    # Review
    
    - [x] Add a short description of the the change to the CHANGELOG.md file
    
    
    Co-authored-by: Nick Lewycky <nicholas@mxc.ca>
```

matklad:
>Practically, could we consider always use Near owned Wasmer repo forks?

Yes, we can consider that. It would be relatively straightforward to do on the Cargo level. That would create maintenance burden on us to do our own releases of all the crates every time wasmer does theirs. 

I don't see much benefits though -- one of the strengths of Rust ecosystem are dependable dependencies, and the strategy of working with upstream to fix their issues seems to me to be more beneficial in the long run. That is "submit patches upstream, *then* point to the temp fork until the next release" feels better than "patch the fork, then submit a PR upstream" from (eco-)systems design perspective.

Cargo.toml
ailisp: OK, makes sense to me

ailisp (approved):

PR #4296, opened 2021-05-12
fix: remove unwrap and alloc from bytes string serialization

austinabell:
Was going to open up an issue, but there was at least this to fix.

My question is how is this safe to use for `Promise` and return data? Is there some guarantee that it's for example JSON formatted and thus always utf8? This seems like something that would be better serialized as base64 or something else, to allow flexibility in the future (even if restricted now) to be some binary encoding


austinabell:
I also took the liberty to reduce the potential allocations of some of the deserialization using `Cow` in this commit [`431fee1` (#4296)](https://github.com/near/nearcore/pull/4296/commits/431fee1411b6a69e9f30364f2650bf63048674c5) but happy to remove that commit if that specifically is unwanted. I only really care about removing the unwrap and understanding why this is used :)

runtime/near-vm-logic/src/serde_with.rs
```diff
@@ -13,20 +14,20 @@ pub mod bytes_as_base64 {
...
     {
-        let s = String::deserialize(deserializer)?;
-        Ok(base64::decode(&s).expect("Failed to deserialize base64 string"))
+        let s = Cow::<'_, str>::deserialize(deserializer)?;
+        Ok(base64::decode(s.as_ref()).expect("Failed to deserialize base64 string"))
     }
 }
 
 /// Serialize `Vec<u8>` as `String`.
 pub mod bytes_as_str {
```
matklad, line 23: Oh dear, this module and `vec_bytes_as_str` do not make sense. It very easy to accidentally panic when using them. Filed https://github.com/near/nearcore/issues/4297

matklad (approved):
LGTM, but we really should just remove bytes_as_str, as they don't make much sense.

runtime/near-vm-logic/src/serde_with.rs
austinabell: I went ahead and made that quick change since it's inline with these changes, hope you don't mind me pushing after you reviewed:
[`05148f3` (#4296)](https://github.com/near/nearcore/pull/4296/commits/05148f3893abf0d720fa9383590422e2dcd68ab4)

austinabell: oh oops I misread your comment, I thought you were pointing at the `expect`s from the other modules. 

Yeah this PR should already fix #4297 and I do agree with wanting to move away from serializing bytes as a string, but it's currently used in the sdk API and it seems a few internal uses.

runtime/near-vm-logic/src/serde_with.rs
```diff
@@ -56,7 +56,7 @@ pub mod bytes_as_base58 {
         D: Deserializer<'de>,
     {
         let s = Cow::<'_, str>::deserialize(deserializer)?;
-        Ok(bs58::decode(s.as_ref()).into_vec().expect("Failed to deserialize base58 string"))
+        Ok(bs58::decode(s.as_ref()).into_vec().map_err(de::Error::custom)?)
```
matklad, line 59: :+1:

runtime/near-vm-logic/src/serde_with.rs
matklad: >Yeah this PR should already fix #4297

I wouldn't say that -- it does improve the situation by turning a panic into error, but it doesn't fix the underlying problem: if UTF validity is assumed for the purposes of serialization, then `String` should be used instead of vector. 

There are a few internal uses, and that's exactly the problem with the API -- all those uses necessary have surprising failure behaviors.

austinabell: > > Yeah this PR should already fix #4297
> 
> I wouldn't say that -- it does improve the situation by turning a panic into error, but it doesn't fix the underlying problem: if UTF validity is assumed for the purposes of serialization, then `String` should be used instead of vector.
> 
> There are a few internal uses, and that's exactly the problem with the API -- all those uses necessary have surprising failure behaviors.

Sorry to have worded my response poorly, I completely agree with what you're saying, I just mean fix in the case that there won't be random panics and the error will at least be handled well. 

I assume, although not a great API, that the interface is expected on both ends, which is how there hasn't been a bigger issue yet. But yes I agree that moving away from this would be ideal.

bowenwang1996 (approved):

PR #4298, opened 2021-05-13
fix: backport wasmer 0.17 FailWithNoError fix to master

matklad:
Initially, we treated this error case as
deterministic (so, we stored this error in our state,
etc.)

Then, in
https://github.com/near/nearcore/pull/4181#discussion_r606267838
we reasoned that this error actually happens
non-deterministically, so it's better to panic in this
case.

However, when rolling this out, we noticed that this
error happens deterministically for at least one
contract. So here we roll this back to a previous
behavior and emit some deterministic error, which won't
cause the node to panic.


matklad:
@olonho please take a look.

I am not really happy backporting this with a :shrug:, but I don't see anything better we can do here.

olonho (approved):
OK, seems it's the best option we got.

ailisp (approved):

matklad:
Discussed during the contract runtime meeting, decided that this is the way to go.

PR #4299, opened 2021-05-13
feat: expose gas breakdown in API

matklad:
WIP, follow up to #4299

The goal here is to expose profile cost breakdown in the API.

Design:

* profile data is stored in the `ExecutionOutcome` (but not in the `PartialExecutionOutcome`).
* for storing data in the database, we use a compact `Vec<u64>` representation. We don't try to optimize for sparseness
* for serializing data to the JSON for the RPC, we use a human-friendly map representation: `{ "touching_trie_node": 9000, ... }`
* execution outcome is recoreded per receipt, not per individual action within receipt. We merge breakdowns for individual actions together. 

Questions:

* is merging a good idea? It seems like ideally we don't merge. But, if we want to have per-action data, we probably want to restructue the whole ExecutionOutcome, so I want to go with a simple solution in this PR and just  merge things
* Adding a field to `ExecutionOutcome` is backwards compatible for JSON output, but is breaking for borsh serialization. Should I do anything about this? If yes, what exactly should I do? 

Test Plan
--------

Stay Tuned, ideally this comes with an intergration test for RPC.


matklad:
Checkbox for the outlined design approval:

- [ ] @bowenwang1996 
- [ ] @matklad 

(adding my, as the design is currently is incomplete from my POV: I don't know if we need to migrate something or otherwise maintain backwards compat with borsh for ExectutionOutcome).

bowenwang1996:
> I don't know if we need to migrate something or otherwise maintain backwards compat with borsh for ExectutionOutcome

We do. Alternatively we can implement custom borsh serializer for `ExecutionOutcome`.

bowenwang1996 (commented):
Design looks fine, but it requires a database migration

matklad:
This actually have been completely taken over by @ailisp , so closing.

PR #4300, opened 2021-05-13
fix(telemetry) Decrease connection lifetime to remove stale connections

### LINKED ISSUE #2798: Telemetry sending stuck once and never deliver after that: "not a result of an error"
**Describe the bug**
nearcore constantly (every 10 seconds as Telemetry gets tried to be sent) reports:

```
INFO telemetry: Telemetry data could not be sent due to: protocol error: not a result of an error
```

**To Reproduce**
Unknown.

**Expected behavior**
Telemetry should not fail to get delivered.

- all versions (master, devnet, betanet, testnet, mainnet)

**Additional context**
It appears to me that there is something wrong with Keep-Alive implementation since once the error get reported, it never recovers even though the network is fine and the requests to https://explorer.devnet.near.org/api/nodes can be sent just fine from the same server.

mhalambek:
Closes #2798

### DIFF
--- a/chain/telemetry/src/lib.rs
+++ b/chain/telemetry/src/lib.rs
@@ -46,11 +46,7 @@ impl TelemetryActor {
 
         let client = Client::builder()
             .timeout(CONNECT_TIMEOUT)
-            .connector(
-                Connector::new()
-                    .conn_lifetime(Duration::from_secs(u64::max_value()))
-                    .conn_keep_alive(Duration::from_secs(30)),
-            )
+            .connector(Connector::new().max_http_version(actix_web::http::Version::HTTP_11))
             .finish();
         Self { config, client }
     }


frol (approved):
I confirm that this resolves the issue and it seems hanging file descriptors are not created.

@mhalambek please, provide some helpers about `netstat` and the reproduction.

chain/telemetry/src/lib.rs
```diff
@@ -48,8 +48,9 @@ impl TelemetryActor {
             .timeout(CONNECT_TIMEOUT)
             .connector(
                 Connector::new()
-                    .conn_lifetime(Duration::from_secs(u64::max_value()))
-                    .conn_keep_alive(Duration::from_secs(30)),
+                    .conn_lifetime(Duration::from_secs(15))
+                    .conn_keep_alive(Duration::from_secs(30))
+                    .finish(),
```
bowenwang1996, line 53: Why did we not have `.finish` before?

bowenwang1996 (approved):

chain/telemetry/src/lib.rs
mhalambek: Something changed in the lib, I've rebased from the older version of nearcore before lib upgrade

mhalambek:
I've reproduced the bug with disconnecting and reconnecting to VPN, @frol managed to do the same turning WiFi on and off.
For seeing socket count I've used
```bash
netstat -atn | grep 34.83.64.96
```
With the ip being that of explorer
I would like to post the output of it during this fix, but it's empty since tcp connection is not reused anymore.

Issue #4076, opened 2021-03-11
Fully switch to wasmer 1 by default


matklad:
In https://github.com/near/nearcore/pull/3799, @ailisp implementing support for wasmer 1.0 :tada: 

However, we are still using wasmer 0.17 by default -- major upgrade of vm version is not a laughing matter, we must make sure that the two behave identically. This issue exists to track our progress on switching to wasmer 1.0 by default, and eventual removal of support for wasemer 0.17 from the tree. 

The next step in transition is to deploy a *fraction* of nodes with `wasmer1` to the beta net.

matklad:
@ailisp am I correct that enabling wasmer1 is just `cargo build --release --features wasmer1_default`? Is anyone actively looking into deploying & monitoring that on betanet?

matklad:
@bowenwang1996 is there anything else we can do besides "deploy to beta and observe" to make sure that transition goes smoothly?

bowenwang1996:
@chefsale @mhalambek let's do the following:
- Deploy a node on testnet and a node on mainnet using this commit https://github.com/near/nearcore/commit/79f173271665534288a5ddf126b386668fa5682f
- Run them until the next release
- Set up alerts to make sure we are notified when they crash

This should give us ~1 month of testing in production. @matklad are there other tests that you would like to see done?

ailisp:
@matklad correct

matklad:
>are there other tests that you would like to see done?

If one node behaves OK, we should try to deploy *half* of the nodes, to make sure that network with mixed versions work.

bowenwang1996:
> If one node behaves OK, we should try to deploy half of the nodes, to make sure that network with mixed versions work.

I don't think that is needed because whether it works solely depends on the transactions, which means that it is sufficient to use a one-node canary that will apply all the transactions. Having more nodes does not improve testing coverage in this case.

chefsale:
The current releases on testnet and mainnet do not have the wasmer1 file. Should w:
1. wait until this get release in the proper cycle and than try to create a machine which will be running this and do the rollout
2. @matklad can create a branch `testing-wasmer` with all the cherry picks needed to run the thing and I can add ` --features wasmer1_default` to the makefile on top of that and build the binary and release that.

cc: @bowenwang1996 @mhalambek

chefsale:
[...]

PR #4301, opened 2021-05-13
feat: enable in-memory cache for wasmer1

matklad:
This PR ports wasmer0 implementation of in-memory cache to wasme1 as
well.

The goal here is to keep the behavior between two wasmers as close as
possible. We absolutely do want to get rid of the in-memory cache, but
it's safer not to do that while also doing a major wasmer upgrade!

See https://github.com/near/nearcore/issues/4076#issuecomment-839762577
for context

Test Plan
----------- 

```
$ cargo test -p near-vm-runner --features wasmer0_vm,wasmer1_vm,no_cache
$ $ cargo test -p near-vm-runner --features wasmer0_vm,wasmer1_vm
```

### DIFF
--- a/runtime/near-vm-runner/src/cache.rs
+++ b/runtime/near-vm-runner/src/cache.rs
@@ -97,6 +97,9 @@ impl fmt::Debug for MockCompiledContractCache {
     }
 }
 
+#[cfg(not(feature = "no_cache"))]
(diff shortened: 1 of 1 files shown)


ailisp (approved):
Looks good to me

PR #4307, opened 2021-05-19
fix: make it possible to enable debug near_vm_runner logs in neard

matklad:
At the moment, `RUST_LOG=debug neard` won't actually enable debug
logging from the vm runner.

This is because near_vm_runner doesn't set the `target` explicitly,
so the target gets inferred from the crate name. As we have `near=info`
as a logging directive, that means that only info and higher levels are
possible to enable via envirenment variable. The filtering is based on
the *prefix* on the module name.

As we use explicit targets elsewhere, let's fix it by setting a
different target, `vm`.

While we are at at, let's also document the logging policy, so that new
developers know that we set target explicitly.


chefsale (approved):
LgTm

bowenwang1996 (approved):

olonho (approved):

Issue #4302, opened 2021-05-13
Cap maximum gas price


bowenwang1996:
Currently the transaction cost may rise too much too quickly when there is congestion on the network, which may result in very suboptimal user experience. We have discussed this and decide to cap maximum gas price to 10x minimum gas price to address this issue in the short term.

PR #4308, opened 2021-05-20
Cap maximum gas price

Longarithm:
Cap maximum gas price: https://github.com/near/nearcore/issues/4302

**Test plan**

- `test_cap_max_gas_price` checks that `max_gas_price <= 10 * min_gas_price` for the new protocol version.
- `process_blocks.rs:cap_max_gas_price_tests` check that `gas_price` may exceed `10 * min_gas_price` for the old protocol version but may not for the new one.


bowenwang1996 (changes requested):
Please add a test that asserts the gas price does not exceed maximum gas price even when there is congestion.

Longarithm:
Added new tests and moved creation of congestion to a separate function

chain/client/tests/process_blocks.rs
```diff
@@ -77,6 +77,79 @@ pub fn create_nightshade_runtimes(genesis: &Genesis, n: usize) -> Vec<Arc<dyn Ru
         .collect()
 }
 
+/// Create environment and set of transactions which cause congestion on the chain.
+fn create_env_with_congestion(
```
bowenwang1996, line 83: Nice

chain/client/tests/process_blocks.rs
```diff
@@ -2745,6 +2761,45 @@ fn test_congestion_receipt_execution() {
...
+            env.produce_block(0, i);
+            let block = env.clients[0].chain.get_block_by_height(i).unwrap().clone();
+            let protocol_version = env.clients[0]
+                .runtime_adapter
+                .get_epoch_protocol_version(block.header().epoch_id())
+                .unwrap();
+            let min_gas_price =
+                env.clients[0].chain.block_economics_config.min_gas_price(protocol_version);
+            if block.header().gas_price() > 10 * min_gas_price {
+                return true;
```
bowenwang1996: Please also assert that congestion actually happens. You can do this by asserting that gas usage is at least as large as gas limit.

bowenwang1996 (approved):
Well done!

PR #4316, opened 2021-05-24
refactor(indexer): Indexer Framework tries to travel back to find missing local receipt in case of delayed receipt

khorolets:


chain/indexer/Cargo.toml
```diff
@@ -1,11 +1,12 @@
 [package]
 name = "near-indexer"
-version = "0.8.1"
+version = "0.9.1"
```
frol, line 3: There is no breaking change, so I suggest we only bump the patch version.

```suggestion
version = "0.8.2"
```

frol (approved):

frol:
It is a good hot-fix. Once this is rolled out, we can work on a better fix.

@khorolets did it resolve the original issue in the wild?

khorolets:
@frol I've checked on our special blocks. Everything is fine.

chain/indexer/Cargo.toml
khorolets: I forgot to change version when we were releasing 0.9.0, so this is correct line

chain/indexer/src/streamer/mod.rs
```diff
@@ -116,15 +118,44 @@ async fn build_streamer_message(
...
+            let IndexerExecutionOutcomeWithOptionalReceipt { execution_outcome, receipt } = outcome;
+            let receipt = if let Some(receipt) = receipt {
+                receipt
+            } else {
+                // Receipt might be missing only in case of delayed local receipt
+                // that appeared in some of the previous blocks
+                // we will be iterating over previous blocks until we found the receipt
+                'find_local_receipt: loop {
+                    let prev_block =
+                        match fetch_block_by_hash(&client, block.header.prev_hash).await {
```
frol, line 132: @khorolets It seems that this loop will always query only one block before the one that the streamer works with. It should be `prev_block.header.prev_hash` instead of `block.header.prev_hash`. This will lead to an endless loop if there is no receipt in the previous block.

khorolets: oops

Issue #4321, opened 2021-05-26
Restoring receipts should not cause network to crash


bowenwang1996:
Betanet crashed with
```
thread 'main' panicked at 'Receipts to restore must contain an entry for shard 0', runtime/runtime/src/lib.rs:1122:18
```
It appears that we assert `restored_receipt` is nonempty for every network, which is not the case. It might be a good idea to add to `MigrationFlags` whether the network is mainnet.

bowenwang1996:
@EgorKulikov please check whether there is a similar issue for restoring storage usage.

EgorKulikov:
> @EgorKulikov please check whether there is a similar issue for restoring storage usage.

I have no similar checks

Longarithm:
I removed this assertion, because it's already checked in `test_restored_receipts_data` test, so it actually doesn't make sense to check it again in `Runtime`.

bowenwang1996:
> I removed this assertion, because it's already checked in `test_restored_receipts_data` test, so it actually doesn't make sense to check it again in `Runtime`.

That is not necessarily the case. We sometimes have asserts in our code to prevent the node from getting into an inconsistent state such that it cannot recover. For example if the node tries to access a trie node that does not exist, it will simply crash rather than trying to do something else because some fundamental invariant is broken. 

In this case though, that does not apply and yes, we should not panic if there is no receipts to be restored.

PR #4322, opened 2021-05-26
Fix protocol upgrade for non-mainnet chains

Longarithm:
Prevent non-mainnet chains from crashing during protocol upgrade to `RestoreReceiptsAfterFix`: https://github.com/near/nearcore/issues/4321

- Remove unnecessary assertion because it's already checked in `test_restored_receipts_data`
- Test protocol upgrade on testnet in `protocol_feature_restore_receipts_after_fix_tests::test_run_for_testnet`

### DIFF
--- a/runtime/runtime/src/lib.rs
+++ b/runtime/runtime/src/lib.rs
@@ -1116,11 +1116,7 @@ impl Runtime {
             == protocol_version
             && migration_flags.is_first_block_with_chunk_of_version
         {
-            migration_data
-                .restored_receipts
-                .get(&0u64)
-                .cloned()
-                .unwrap_or_default()
+            migration_data.restored_receipts.get(&0u64).cloned().unwrap_or_default()
         } else {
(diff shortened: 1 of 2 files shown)


runtime/runtime/src/lib.rs
```diff
@@ -1116,11 +1116,7 @@ impl Runtime {
             == protocol_version
             && migration_flags.is_first_block_with_chunk_of_version
         {
-            migration_data
-                .restored_receipts
-                .get(&0u64)
-                .expect("Receipts to restore must contain an entry for shard 0")
-                .clone()
+            migration_data.restored_receipts.get(&0u64).cloned().unwrap_or_default()
```
EgorKulikov, line 1119: Can you add comment here about why it can be None?

EgorKulikov (approved):

bowenwang1996 (approved):
Please address Egor's comment

PR #4323, opened 2021-05-26
test: indirect call to a host function

matklad:


ailisp (approved):
So this does no error on wasmer 0 (correct behavior) and panic on wasmer 1?

matklad:
exactly

olonho (approved):

PR #4331, opened 2021-05-27
fix broken release due to wrong rust version and resolver conflit, remove the dockerfile caching as it is not used anyway.

chefsale:
builds to fail:
https://buildkite.com/nearprotocol/nearcore-release/builds/601

Created a build to verify if it works as a quick fix to unblock master failing: https://buildkite.com/nearprotocol/nearcore-release/builds/607

As well remove the mount=type=cache as they get removed eitherways every time once we destroy the container, so there aren't useful at all here.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -29,16 +29,12 @@ COPY . .
 ENV CARGO_TARGET_DIR=/tmp/target
 ENV RUSTC_FLAGS='-C target-cpu=x86-64'
 ENV PORTABLE=ON
-RUN --mount=type=cache,target=/tmp/target \
-    --mount=type=cache,target=/usr/local/cargo/git \
-    --mount=type=cache,target=/usr/local/cargo/registry \
-    cargo +"$(cat /tmp/rust-toolchain)" build -p neard --release && \
+RUN cargo +"$(cat /tmp/rust-toolchain)" build -p neard --release && \
     mkdir /tmp/build && \
     cd /tmp/target/release && \
     mv ./neard /tmp/build
 
 COPY scripts/run_docker.sh /tmp/build/run.sh
-
 
 # Actual image
 FROM ubuntu:18.04
--- a/Dockerfile.nightly
+++ b/Dockerfile.nightly
@@ -29,10 +29,7 @@ COPY . .
 ENV CARGO_TARGET_DIR=/tmp/target
 ENV RUSTC_FLAGS='-C target-cpu=x86-64'
 ENV PORTABLE=ON
-RUN --mount=type=cache,target=/tmp/target \
-    --mount=type=cache,target=/usr/local/cargo/git \
-    --mount=type=cache,target=/usr/local/cargo/registry \
-    cargo +"$(cat /tmp/rust-toolchain)" build -p neard --release --features nightly_protocol --features nightly_protocol_features && \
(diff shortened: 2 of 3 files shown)


chefsale:
cc: @matklad to take a look and help 😄

frol:
@chefsale it should not really use `+stable` in the command line. Can you remove it and try just `cargo build --package neard --release`?

chefsale:
OK @frol , yes I agree. I will add the cat rust toolchain as the dockerfile does shortly.

chefsale:
@frol changed the makefile to use the toolchain provided in the file, didn't realize it wasn't the case there, this seems to solve the build release issue: https://buildkite.com/nearprotocol/nearcore-release/builds/607#f6e8f42b-a385-4dd9-9662-c70eb57d049a

chefsale:
Added verbose logging for the docker build and increased timeout to 10 hours to see if they build takes longer for some reason and what are the logs saying: https://buildkite.com/nearprotocol/nearcore-release/builds/609

matklad:
Looking into this now. 

Good build: https://buildkite.com/nearprotocol/nearcore-release/builds/600#b6617b72-bed8-4f49-aab3-a3c0202c41db
Bad build: https://buildkite.com/nearprotocol/nearcore-release/builds/609#c54f348e-1301-41da-8ca6-fde8b5158a29

Seems that the build is just stuck for some reason :thinking:. I've tried running a couple of clean builds locally without docker:

old a668e4399655af513b0d90e0be694dad2448e6cd

```
λ t cargo build -q --release -p neard
real 347.37s
cpu  2719.58s (2611.37s user + 108.21s sys)
rss  3595.57mb
```

new 4009a1422b706a1a0156a4134b4aa5f37deaa2ab

```
λ t cargo build -q --release -p neard
real 289.45s
cpu  2681.82s (2574.72s user + 107.10s sys)
rss  3537.24mb
```

So far, no clue as to what's going on

chefsale:
So regukar build works now, but the docker build takes 3 plus hours so far, it is running though. It could be issue in the dockerfile also I believe, one thing is we shouldn't be using the volume at all, as it slows things down. But could be something else as well, but most likely due to something with docker:

https://buildkite.com/nearprotocol/nearcore-release/builds/609

matklad:
I think I've reprodced the issue in a local docker build (it also seems like it's stuck compiling nearcore), investigating .

chefsale:
Yes, it gets stuck on:
```
[2021-05-28T08:32:22Z] #14 312.9    Compiling nearcore v1.2.0 (/near/nearcore)
```
for hours 😅

matklad:
will prepare a fix later today

chefsale:
@matklad you are the best. Can I get an approval on this changes for now and than leave this docker fix to you? cc: @frol @mhalambek

matklad:
The change looks good to me. Although, rather than using `+` syntax, I'd just set `RUSTUP_TOOLCHAIN` env var, but that's not really relevant.

matklad:
* https://github.com/near/nearcore/pull/4332 is the minimal fix
* https://github.com/near/nearcore/pull/4333 is the fix for the underlying root cause (compiling test code into neard)

bowenwang1996 (approved):

matklad:
Stable release got through: https://buildkite.com/nearprotocol/nearcore-release/builds/611#acae2d30-1188-4731-8453-fde4df42d6e9

Nightly is still broken though, will look into that next week .

PR #4334, opened 2021-05-28
doc: update architecture.md

bowenwang1996:
Update `architecture.md` to include some high-level information about nearcore and two ways state updates can happen. Also add codeowners for `/docs`.

Test plan
---------
N/A


docs/architecture.md
```diff
@@ -105,6 +116,15 @@ It is also worth noting that `NightshadeRuntime` is the struct that implements `
 
 ## Cross Cutting Concerns
 
+## State update
```
matklad, line 119: This section is misplaced, `### Logging & Observability` is intended to be under `## Cross Cutting Concerns`

docs/architecture.md
```diff
@@ -13,6 +13,17 @@ More specifically, the neard binary can do the following:
...
 - start a node with a given folder that contains required data
 
+There are three major components of nearcore:
+
+- Network. We implement a peer-to-peer network that powers communications between blockchain nodes.
+This includes initiating connections with other nodes, maintain a view of the entire network, routing messages to the right nodes, etc.
+Network is a somewhat standalone module, though it uses information about validators to propagate and validate certain messages.
+- Chain. Chain is responsible for building and maintaining the blockchain data structure.
+This includes block and chunk production and processing, consensus, and validator selection.
+However, chain is not responsible for actually applying transactions and receipts.
```
matklad, line 23: Nice, it's very helpful to understand what modules are *not* responsible for :+1: !

docs/architecture.md
bowenwang1996: Ah sorry I actually intend to have `### State update`, unless I misunderstand what you mean by `Cross Cutting Concerns`

matklad: Yeah, I am not sure I'd call this a cross-cutting concern. It seems that this is in chain's purview, as it calls both the catchup and apply?

That is, in my mind cross-cutting concerns are for "aspect-oriented" things which touch otherwise unrelated sections of code:

* what common logging framework do we use?
* what is the standard testing approach?
* how do we monitor performance?
* what's the common error handling strategy (speaking of which, documenting deterministic vs non-deterministic vs coordinated failures here would be a helpful)

matklad (approved):
Looks ok to me as is, though I'd move state update section out of cross-cutting concerns and into chain (at least, my understanding is that it is `chain` where both kinds of state update take place).

docs/architecture.md
bowenwang1996: > It seems that this is in chain's purview

Not really. It can be updated in both runtime and chain. Where do you think is the best place to put it?

matklad: >Not really. It can be updated in both runtime and chain. 

But it seems that the "caller" is, in both cases, the `chain`? Ie, it's the chain which calls `Runtime::apply` (indirectly, via `RuntimeAdapter::apply_transactions`). 

In the overview, I'd expect pointers to the call-sites: 

> two code paths may trigger state update. Normal update (processing of a new block) is initiated by chain in `???` and ultimately calls `Runtime::apply`. If a node is lagging behind, the chain engages the `catchup` routine.

bowenwang1996: > But it seems that the "caller" is, in both cases, the chain? Ie, it's the chain which calls Runtime::apply (indirectly, via RuntimeAdapter::apply_transactions).

Yeah good point

frol (approved):

PR #4338, opened 2021-05-28
fix(release): fix the failing builds on an error due to non-existing …

chefsale:
…features. Created a build to verify the fix works: https://buildkite.com/nearprotocol/nearcore-perf-release/builds/235. It seems before the features which don't exist would get ignored, but now it fails due to changes in tooling.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,24 +22,24 @@ debug:
 
 perf-release:
 	CARGO_PROFILE_RELEASE_DEBUG=true cargo $(RUST_OPTIONS) build -p neard --release --features performance_stats,memory_stats
-	cargo $(RUST_OPTIONS) build -p near-vm-runner-standalone --release --features performance_stats,memory_stats
-	cargo $(RUST_OPTIONS) build -p state-viewer --release --features performance_stats,memory_stats
-	cargo $(RUST_OPTIONS) build -p store-validator --release --features performance_stats,memory_stats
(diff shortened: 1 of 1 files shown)


frol (changes requested):
This fix does not seem right. We need to fix the feature names, I believe. The features themselves are definitely required unless I miss something

miraclx:
Yeah, we could directly activate them with `nearcore/<feature>` or at least proxy them in their respective Cargo.toml files instead of removing them outright. However, `near-vm-runner-standalone` doesn't depend on nearcore so we'll have to activate the required features on whatever dependencies it has.

chefsale:
@frol I believe that only neard contains the perf stats feature flags. Maybe I'm wrong. Cc: @bowenwang1996

miraclx:
FWIK, you're right, but not entirely, it's used in `nearcore`, `near-network` and `near-performance-metrics-macros`. So, `neard` and anything else that depends on those can just activate them via `nearcore`. The only crate that won't be needing those features is `near-vm-runner-standalone` because it depends on neither of those. But `state-viewer` and `store-validator` would.

chefsale:
Yes, near-vm-standalone-runner doesn't use it at all, but the two other binaries even they use it, I don't think there is any benefit of enabling it, as it is used to debug neard perfomance issues which neither state

PR #4333, opened 2021-05-28
fix: remove test code from neard

matklad:
testlib is the library we use for testing, it makes no sense that we
compile it into our production code. Admittedly, I lack the context as
to *why* it was added as a non-dev dep in the first place. I have a
suspicion that it was a workaround some feature bugs.

Now that we use resolever=2, it makes sense to try to fix this in a more
principled way, so let's see if some tests start failing due to this
change.

PR #4339, opened 2021-05-29
fix(neard): fixed failed CI tests due to features mismatches

### LINKED ISSUE #4325: Remove nearcore features from `neard` Cargo.toml
With the decoupling of `neard` and `nearcore` we did a copy of features in the Cargo.toml file. Instead, we should just ~make the features in `neard/Cargo.toml` proxy~ remove them from `neard/Cargo.toml` file. UPD: see further conversation for details.

~after #4292 they look like:~

~`nightly_protocol = ["near-primitives/nightly_protocol", "near-jsonrpc/nightly_protocol"]`~

~which may diverge if we forget to maintain the list in both `neard/Cargo.toml` and `nearcore/Cargo.toml`~

miraclx:
We're sort of in a weird spot as it is, we have features within `nearcore` that are required but can't be enabled if they depend on crates listed under `dev-dependencies` (bug in cargo: https://github.com/rust-lang/cargo/issues/6915). Temporarily listing them under `dependencies`, fixed the CI testing failures we had while working on #4292 (same error @ailisp highlighted: https://github.com/near/nearcore/pull/4333#issuecomment-850545773), but as pointed out in https://github.com/near/nearcore/pull/4331#issuecomment-850393673, this method should be out of the question. If we remove it, however, we can't work with the tests that depend on those features.

This PR moves the _previously top-level_ tests into a new crate to have better dependency and feature handling as @matklad suggested

> * [move tests from `/test` folder somewhere -- either to some of the existing packages (neard perhaps?) or into a new `integration-tests` package](https://github.com/near/nearcore/pull/4292#issuecomment-838249062)

This would ensure we have dependencies like `testlib`, `runtime-params-estimator` and `restored-receipts-verifier` that are only needed for testing and depend on nearcore without cyclic dependency issues while having all features relevant to testing in order.

The features within `neard` have been reduced to proxies to `nearcore` features, which partially resolves #4325 (though the intent for that has morphed a bit). This also makes for a cleaner dependency graph from `neard` perspective, removing extraneous dependencies that were previously required.

I also noticed removed the `old_tests` feature that should've been removed along with the rest of it in #928.

Updated some docs and code comments referencing the old `neard` path too.


miraclx:
All required CI checks pass so far ✔️ except the untested `runtime-params-estimator` which hasn't passed in a while anyways, I see the issues with this are as follows:

* First, we add the `rustfmt` component for a missing toolchain https://github.com/near/nearcore/blob/4f300bb91399f987205276cc7ddad44fad767205/runtime/runtime-params-estimator/emu-cost/Dockerfile#L12 I'm not sure what toolchain is installed within the docker image but is this line even necessary?

* Next, we need to properly escape this https://github.com/near/nearcore/blob/4f300bb91399f987205276cc7ddad44fad767205/runtime/runtime-params-estimator/emu-cost/ci.sh#L17 else bash does the cmd substitution outside of the nested bash invocation.

chefsale (approved):

chefsale:
The onky thing I believe we can remove performance stats feature flag from all the binaries except neard. @bowenwang1996

frol (approved):

neard/Cargo.toml
```diff
@@ -20,51 +20,30 @@ tracing = "0.1.13"
...
+protocol_feature_alt_bn128 = ["nearcore/protocol_feature_alt_bn128"]
+protocol_feature_block_header_v3 = ["nearcore/protocol_feature_block_header_v3"]
+protocol_feature_add_account_versions = ["nearcore/protocol_feature_add_account_versions"]
+protocol_feature_tx_size_limit = ["nearcore/protocol_feature_tx_size_limit"]
+protocol_feature_allow_create_account_on_delete = ["nearcore/protocol_feature_allow_create_account_on_delete"]
+protocol_feature_fix_storage_usage = ["nearcore/protocol_feature_fix_storage_usage"]
+nightly_protocol_features = ["nearcore/nightly_protocol_features"]
+nightly_protocol = ["nearcore/nightly_protocol"]
+protocol_feature_restore_receipts_after_fix = ["nearcore/protocol_feature_restore_receipts_after_fix"]
+protocol_feature_cap_max_gas_price = ["nearcore/protocol_feature_cap_max_gas_price"]
```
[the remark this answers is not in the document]
miraclx: Yeah, I thought to leave that to @chefsale, to ensure all CI checks remain intact. Should we do this here instead?

nearcore/Cargo.toml
```diff
@@ -72,20 +68,10 @@ protocol_feature_add_account_versions = ["near-primitives/protocol_feature_add_a
...
-# enable this to build neard with wasmer 1.0 runner
-# now if none of wasmer0_default, wasmer1_default or wasmtime_default is enabled, wasmer0 would be default
-wasmer1_default = ["node-runtime/wasmer1_default"]
-wasmer0_default = ["node-runtime/wasmer0_default"]
-wasmtime_default = ["node-runtime/wasmtime_default"]
-# enable some of these to run runtime tests with wasmer 1.0, 0.x and wasmtime vm enabled
-# but would not change default runner used by neard
-wasmer1_vm = ["node-runtime/wasmer1_vm"]
-wasmer0_vm = ["node-runtime/wasmer0_vm"]
-wasmtime_vm = ["node-runtime/wasmtime_vm"]
```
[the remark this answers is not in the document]
miraclx: Aren't these only used for the tests in `integration-tests`? I moved them there.

neard/Cargo.toml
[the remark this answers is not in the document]
frol: @miraclx I see. Let's merge this first and then address the issue in question separately.

nearcore/Cargo.toml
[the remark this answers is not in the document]
frol: @matklad @olonho are these documented knobs that we want someone to play around easily?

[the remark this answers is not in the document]
olonho: Well, think as with most of our features, they are ad-hoc knobs to configure certain behavior.

olonho (approved):

nearcore/Cargo.toml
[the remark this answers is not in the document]
matklad: Yeah, we should leave this features here so that we can decide to build `neard` with `wasmer0` or `wasmer1`. 

On the other hand, I don't think we need this set of features in `integration-tests/Cargo.toml` -- integration tests should use whatever our default VM is. We test several different vms inside `near_vm_runner`.

neard/Cargo.toml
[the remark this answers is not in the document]
matklad: Yeah, I am no longer so sure if removing featrue delegation here would be a good idea. Turns out, the ` -p a --feature b/f` trick works only for direct dependencies: https://github.com/rust-lang/cargo/issues/9514.

integration-tests/Cargo.toml
```diff
@@ -0,0 +1,45 @@
...
+funty = "=1.1.0" # Pin dependency to avoid compilation errors: https://github.com/myrrlyn/funty/issues/3
+rand = "0.7"
+
+near-logger-utils = { path = "../test-utils/logger" }
+near-chain-configs = { path = "../core/chain-configs" }
+near-crypto = { path = "../core/crypto" }
+near-primitives = { path = "../core/primitives" }
+near-store = { path = "../core/store" }
+runtime-params-estimator = { path = "../runtime/runtime-params-estimator", default_features = false }
+restored-receipts-verifier = { path = "../tools/restored-receipts-verifier" }
```
matklad, line 17: Why do the need these two deps here? Seems like they can be safely removed?

neard/Cargo.toml
```diff
@@ -20,51 +20,30 @@ tracing = "0.1.13"
...
-# enable this to build neard with wasmer 1.0 runner
-# now if none of wasmer0_default, wasmer1_default or wasmtime_default is enabled, wasmer0 would be default
-wasmer1_default = ["node-runtime/wasmer1_default"]
-wasmer0_default = ["node-runtime/wasmer0_default"]
-wasmtime_default = ["node-runtime/wasmtime_default"]
-# enable some of these to run runtime tests with wasmer 1.0, 0.x and wasmtime vm enabled
-# but would not change default runner used by neard
-wasmer1_vm = ["node-runtime/wasmer1_vm"]
-wasmer0_vm = ["node-runtime/wasmer0_vm"]
-wasmtime_vm = ["node-runtime/wasmtime_vm"]
```
matklad, line 70: Let's keep wasm-related features in neard.

integration-tests/Cargo.toml
```diff
@@ -0,0 +1,45 @@
...
+# enable this to build neard with wasmer 1.0 runner
+# now if none of wasmer0_default, wasmer1_default or wasmtime_default is enabled, wasmer0 would be default
+wasmer1_default = ["node-runtime/wasmer1_default"]
+wasmer0_default = ["node-runtime/wasmer0_default"]
+wasmtime_default = ["node-runtime/wasmtime_default"]
+# enable some of these to run runtime tests with wasmer 1.0, 0.x and wasmtime vm enabled
+# but would not change default runner used by neard
+wasmer1_vm = ["node-runtime/wasmer1_vm"]
+wasmer0_vm = ["node-runtime/wasmer0_vm"]
+wasmtime_vm = ["node-runtime/wasmtime_vm"]
```
matklad, line 45: Let's remove wasme-related features from `integration-tests` (unless I am missing a reason for why they should be here?)

matklad (changes requested):

Makefile
```diff
@@ -22,24 +22,24 @@ debug:
...
+	cargo $(RUST_OPTIONS) build -p store-validator --release --features nearcore/performance_stats,nearcore/memory_stats
 
 perf-debug:
 	CARGO_PROFILE_RELEASE_DEBUG=true cargo $(RUST_OPTIONS) build -p neard --features performance_stats,memory_stats
-	cargo $(RUST_OPTIONS) build -p near-vm-runner-standalone --features performance_stats,memory_stats
-	cargo $(RUST_OPTIONS) build -p state-viewer --features performance_stats,memory_stats
-	cargo $(RUST_OPTIONS) build -p store-validator --features performance_stats,memory_stats
+	cargo $(RUST_OPTIONS) build -p near-vm-runner-standalone
+	cargo $(RUST_OPTIONS) build -p state-viewer --features nearcore/performance_stats,nearcore/memory_stats
+	cargo $(RUST_OPTIONS) build -p store-validator --features nearcore/performance_stats,nearcore/memory_stats
```
ailisp, line 33: we probably should check these build in .buildkite/pipeline.yml:
```
cargo check -p state-viewer --features nearcore/performance_stats,nearcore/memory_stats
cargo check -p store-validator --features nearcore/performance_stats,nearcore/memory_stats
```
because existing check doesn't check these feature combinations

Makefile
```diff
@@ -22,24 +22,24 @@ debug:
...
 
 nightly-debug:
 	CARGO_PROFILE_RELEASE_DEBUG=true cargo $(RUST_OPTIONS) build -p neard --features nightly_protocol,nightly_protocol_features,performance_stats,memory_stats
-	cargo $(RUST_OPTIONS) build -p near-vm-runner-standalone --features nightly_protocol,nightly_protocol_features,performance_stats,memory_stats
-	cargo $(RUST_OPTIONS) build -p state-viewer --features nightly_protocol,nightly_protocol_features,performance_stats,memory_stats
-	cargo $(RUST_OPTIONS) build -p store-validator --features nightly_protocol,nightly_protocol_features,performance_stats,memory_stats
+	cargo $(RUST_OPTIONS) build -p near-vm-runner-standalone --features nightly_protocol,nightly_protocol_features
+	cargo $(RUST_OPTIONS) build -p state-viewer --features nearcore/nightly_protocol,nearcore/nightly_protocol_features,nearcore/performance_stats,nearcore/memory_stats
+	cargo $(RUST_OPTIONS) build -p store-validator --features nearcore/nightly_protocol,nearcore/nightly_protocol_features,nearcore/performance_stats,nearcore/memory_stats
```
ailisp, line 45: Same check in buildkite here

runtime/runtime-params-estimator/emu-cost/ci.sh
```diff
@@ -14,7 +14,7 @@ docker run \
      /usr/bin/env bash -c "
 set -ex
 cd /host/${srcdir}/runtime/runtime-params-estimator
-$(cd test-contract && ./build.sh)
+: "'$(cd test-contract && ./build.sh)'"
```
ailisp, line 17: What is this magic and difference compare to `$()` :D I googled but this is hard to search

ailisp (approved):

runtime/runtime-params-estimator/emu-cost/ci.sh
frol: @ailisp when you do `$(echo pwd)` you will get the `$()` resolved to `pwd`, and it will get executed:

image

So the current version of the script resolves `./build.sh` and then all the output is treated as a command that needs to be executed.

`:` (colon) is a built-in no-op command, so this trick basically will end up calling no-op with `pwd` argument, which is fine.

It is better to re-write this piece of code into something like:

```
pushed test-contract
./build.sh
popd
```

instead of crafting magic here :smile:

runtime/near-vm-runner-standalone/Cargo.toml
```diff
@@ -39,7 +39,16 @@ near-test-contracts = { path = "../near-test-contracts" }
...
+protocol_feature_evm = ["near-primitives/protocol_feature_evm"]
+protocol_feature_block_header_v3 = ["near-primitives/protocol_feature_block_header_v3"]
+protocol_feature_add_account_versions = ["near-primitives/protocol_feature_add_account_versions"]
+protocol_feature_tx_size_limit = ["near-primitives/protocol_feature_tx_size_limit"]
+protocol_feature_fix_storage_usage = ["near-primitives/protocol_feature_fix_storage_usage"]
+nightly_protocol_features = ["nightly_protocol", "near-primitives/nightly_protocol_features", "protocol_feature_evm", "protocol_feature_block_header_v3", "protocol_feature_alt_bn128", "protocol_feature_add_account_versions", "protocol_feature_tx_size_limit", "protocol_feature_fix_storage_usage", "protocol_feature_restore_receipts_after_fix", "protocol_feature_cap_max_gas_price"]
+protocol_feature_restore_receipts_after_fix = ["near-primitives/protocol_feature_restore_receipts_after_fix"]
+protocol_feature_cap_max_gas_price = ["near-primitives/protocol_feature_cap_max_gas_price"]
+nightly_protocol = ["near-primitives/nightly_protocol"]
```
bowenwang1996, line 51: why do we need all these features?

miraclx: While fixing CI checks, lines like these;
Before now, lines like these enabled features that prior to the decoupling, existed in the workspace.
https://github.com/near/nearcore/blob/928f15df25ca4128d88b98e930d580ed2023d576/Makefile#L37
Now those features are defined in `nearcore`. This crate, however, doesn't directly depend on `nearcore`. So we have to adapt the needed features to directly match the available dependencies.

Makefile
miraclx: Aren't these covered by this? https://github.com/near/nearcore/blob/928f15df25ca4128d88b98e930d580ed2023d576/.buildkite/pipeline.yml#L30
And @chefsale, what do you think? Regarding your remark in https://github.com/near/nearcore/pull/4338#issuecomment-850899095, if those lines aren't necessary, checks for them might not as well.

matklad (approved):

matklad:
Added auto-merge -- feels like the faster we do this, the better )

Makefile
ailisp: No it doesn't because there's no workspace features, see this two commits' buildkite:
https://github.com/near/nearcore/pull/4287/commits/1a1256ce65cfc6a0076da19ecbeb16790ce214df
https://github.com/near/nearcore/pull/4287/commits/08e7e31e2482850f515acaacb8d04a8d30b4a74b

So the first one only checks
```
       RUSTFLAGS='-D warnings' cargo check --workspace --all-targets --all-features 
```
It passed. The second commit add a `cargo check -p neard --features sandbox` feature and it failed. This means sandbox feature of neard was not checked by ` cargo check --workspace --all-targets --all-features `

ailisp: @miraclx I think you're correct, all-features do include all features, mine case was failing because it's missing some features instead of `--all-features` didn't work. So the ci doesn't check all feature combinations. `cargo check --workspace --all-targets --all-features` ensures when all features are included, all packages can compile. But real build includes package build with no feature, or a subset of features, and they can fail if feature dependency wasn't correctly specified.

So for example this one from Makefile
```
cargo build -p state-viewer --features nearcore/performance_stats,nearcore/memory_stats
```
since we don't do `make` on ci, it makes sense to cargo check these compile, not introduced by your PR, but a thing we probably should do to gurantee `make` works on master commits.

PR #4344, opened 2021-05-31
Compile contracts on writing them to state

Longarithm:
Put each contract to the `CompiledContractCache` when it is written to the state.
This will help to decrease costs for function calls.

**Plan**
- Add caching to `RuntimeAdapter::apply_state_part` for state downloaded from other node
- Enable caching in `apply_deploy_contract`

**Test**
Check that contract is precompiled and cached:
- in `Runtime.apply`
- on state sync


Longarithm:
Please note that it's only a sketch and not ready for review yet

core/store/src/trie/iterator.rs
```diff
@@ -185,7 +185,15 @@ impl<'a> TrieIterator<'a> {
                 }
                 Some(IterStep::Continue)
             }
-            _ => panic!("Should never see Entering or AtChild without a Branch here."),
+            _ => None,
```
bowenwang1996, line 188: what's the reason for changing this panic to `None`

core/store/src/trie/iterator.rs
```diff
@@ -212,7 +220,11 @@ impl<'a> TrieIterator<'a> {
         let mut nodes_list = Vec::new();
         // Actually (self.key_nibbles[..] == path_begin) always because path_begin always ends in a node
         if &self.key_nibbles[..] >= path_begin {
-            nodes_list.push(last_hash);
+            let is_contract_code_node = match self.current_iter_step() {
+                Some(IterStep::Value(_)) => is_contract_code_key(&self.key()),
+                _ => false,
+            };
+            nodes_list.push((last_hash, is_contract_code_node));
```
bowenwang1996, line 227: Don't think it is great to have this function care about some specific trie key. It may make sense to pass some function here as an argument.

core/primitives/src/errors.rs
```diff
@@ -58,6 +58,8 @@ pub enum RuntimeError {
     ReceiptValidationError(ReceiptValidationError),
     /// Error when accessing validator information. Happens inside epoch manager.
     ValidatorError(EpochError),
+    /// Error when precompiling contract and putting result to the contract cache.
+    ContractPrecompilationError(String),
```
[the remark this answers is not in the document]
Longarithm: No, I need this to handle `ContractPrecompilationError` which may come from `precompile_contract`.

core/store/src/trie/iterator.rs
```diff
@@ -197,22 +208,37 @@ impl<'a> TrieIterator<'a> {
...
         let mut prefix = Self::common_prefix(path_end, &self.key_nibbles);
         if self.key_nibbles[prefix..] >= path_end[prefix..] {
             return Ok(vec![]);
         }
         let mut nodes_list = Vec::new();
+
         // Actually (self.key_nibbles[..] == path_begin) always because path_begin always ends in a node
         if &self.key_nibbles[..] >= path_begin {
-            nodes_list.push(last_hash);
+            let last_predicate_result = match self.get_current_iter_step() {
```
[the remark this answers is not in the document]
Longarithm: Yeah, there is a problem. I reverted this change and chose another implementation.

core/primitives/src/errors.rs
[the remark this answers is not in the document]
bowenwang1996: Where would this error occur? Changing `RuntimeError` may result in a protocol change and we need to be very careful here.

core/store/src/trie/iterator.rs
```diff
@@ -130,6 +130,15 @@ impl<'a> TrieIterator<'a> {
         result
     }
 
+    fn has_value(&self) -> bool {
+        let b = self.trail.last().expect("Trail finished.");
```
bowenwang1996, line 134: is this function guaranteed to not be called when trail is empty?

core/store/src/trie/iterator.rs
```diff
@@ -130,6 +130,15 @@ impl<'a> TrieIterator<'a> {
...
+        match (b.status.clone(), &b.node.node) {
```
bowenwang1996, line 135: Why do we need to clone here? Could we do match on `&b.status`?

core/store/src/trie/iterator.rs
```diff
@@ -197,22 +206,33 @@ impl<'a> TrieIterator<'a> {
...
         let mut prefix = Self::common_prefix(path_end, &self.key_nibbles);
         if self.key_nibbles[prefix..] >= path_end[prefix..] {
             return Ok(vec![]);
         }
         let mut nodes_list = Vec::new();
+
         // Actually (self.key_nibbles[..] == path_begin) always because path_begin always ends in a node
         if &self.key_nibbles[..] >= path_begin {
-            nodes_list.push(last_hash);
+            nodes_list.push((last_hash, self.has_value() && self.eval_predicate(predicate)));
```
bowenwang1996, line 235: It is still not no-op when there is no predicate. Also I still don't like this API. @matklad I wonder whether you have ideas regarding how this could be done better.

core/store/src/trie/iterator.rs
Longarithm: No, but AFAIU this case corresponds to the end of trie, so it can be handled trivially.
@abacabadabacaba is that correct?

core/store/src/trie/iterator.rs
matklad: :thinking: yeah, I am having hard time wrapping my head around the API here. It feels suspicious that we seem to do `retrieve_raw_bytes(&hash)?;` twice, for example. We also return internal node hashes, rather than proper key-values.

Don't have immediate suggestions here, besides:

* return a tripple of `(NodeHash, Value, Key)`, to let the caller figure how to deal with them
* rather than returning a vec, maybe cast the thing as a proper iterator:

```
struct TrieIntervalIterator {
    inner: TrieIterator,
    interval: std::ops::Range<&[u8]>
}
```

matklad: Hm, one think that feels a bit strange is that we tend to allocate all over the place, when reading both keys and values. I wonder if this can benefit from *not* using the iterator API, and instead using "streaming" iterators, which borrow from the current iterator object. Something like

```rust
struct StreamingTrieIterator {
    trie: &'a Trie,
    trail: Vec<Crumb>,
    pub(crate) key_nibbles: Vec<u8>,
    root: CryptoHash,
    
    key_buf: Vec<u8>,
    val_buf: Vec<u8>,
}

struct Item<'a> {
    pub key: &'a [u8],
    pub value: &'a [u8],
}

impl StreamingTrieIterator {
    pub fn next<'a>(&mut self) -> Item<'a> // Item borrows *current* key and value from `buf` fields
}
```

It seems that this can make access to the current key cheap, which means that we'll be able to just always return the full key alongside the value, which means the caller will be able to calculate arbitrary predicates using always available key.

core/primitives/src/errors.rs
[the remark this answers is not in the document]
ailisp: Before, Compilation error was happen at contract function call, so it is a enum of FunctionCallError. Switch to precompile contracts this error doesn't fit FunctionCallError any more

ailisp (approved):

core/store/src/trie/iterator.rs
bowenwang1996: @Longarithm do we actually need the hash if the predicate returns false?

Longarithm: @bowenwang1996 We need to return all hashes from this method. They are later converted to `TrieChanges` and nodes mentioned there are applied to `ShardTries`.

nearcore/src/runtime/mod.rs
```diff
@@ -1408,14 +1438,18 @@ impl RuntimeAdapter for NightshadeRuntime {
...
             .expect("Part was already validated earlier, so could never fail here");
-        let trie_changes = Trie::apply_state_part(&state_root, part_id, num_parts, part)
-            .expect("combine_state_parts is guaranteed to succeed when each part is valid");
+        let (trie_changes, contract_codes) =
+            Trie::apply_state_part(&state_root, part_id, num_parts, part)
+                .expect("combine_state_parts is guaranteed to succeed when each part is valid");
         let tries = self.get_tries();
         let (store_update, _) =
             tries.apply_all(&trie_changes, shard_id).expect("TrieChanges::into never fails");
+        let protocol_version = self.get_epoch_protocol_version(epoch_id)?;
```
EgorKulikov, line 1451: Is this the only reason we propagate epoch_id? If yes, why not just propogate epoch version instead?

core/store/src/trie/state_parts.rs
```diff
@@ -183,27 +183,35 @@ impl Trie {
         part_id: u64,
         num_parts: u64,
         part: Vec<Vec<u8>>,
-    ) -> Result<TrieChanges, StorageError> {
+    ) -> Result<(TrieChanges, Vec<Vec<u8>>), StorageError> {
```
EgorKulikov, line 186: Maybe introduce struct as well, but at the very least change doc

EgorKulikov (changes requested):
I still need to look more thoroughly at changes to state and some other, but this is what I have for now

core/store/src/trie/iterator.rs
Longarithm: After discussion with @matklad I moved logic of computing predicate out of there and started returning pairs (hash, key) from it, which makes `visit_nodes_interval` more clear

The average overhead of this change in `visit_nodes_interval` is not more than 1%, which was computed by running 500 iterations of `test_combine_state_parts` test.

chain/client/Cargo.toml
```diff
@@ -55,10 +56,11 @@ expensive_tests = []
 adversarial = ["near-network/adversarial", "near-chain/adversarial"]
 metric_recorder = ["near-client-primitives/metric_recorder"]
 delay_detector = ["near-chain/delay_detector", "near-network/delay_detector", "delay-detector"]
-protocol_feature_block_header_v3 = ["near-primitives/protocol_feature_block_header_v3", "near-chain/protocol_feature_block_header_v3", "near-store/protocol_feature_block_header_v3"]
+protocol_feature_block_header_v3 = ["near-primitives/protocol_feature_block_header_v3", "near-chain/protocol_feature_block_header_v3", "near-store/protocol_feature_block_header_v3", "nearcore/protocol_feature_block_header_v3"]
```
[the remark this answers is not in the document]
Longarithm: Yeah, I tried to apply some hack to test `near-client` separately: https://github.com/near/nearcore/issues/4351#issuecomment-855140473
But it doesn't work anyway because `neard` is not building with it

chain/client/tests/process_blocks.rs
```diff
@@ -34,6 +34,8 @@ use near_network::{
 };
 use near_primitives::block::{Approval, ApprovalInner};
 use near_primitives::block_header::BlockHeader;
+#[cfg(feature = "protocol_feature_precompile_contracts")]
```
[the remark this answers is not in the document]
Longarithm: I agree. It should be covered by this broader issue https://github.com/near/nearcore/issues/4357

core/store/src/trie/iterator.rs
```diff
@@ -197,22 +208,33 @@ impl<'a> TrieIterator<'a> {
...
+    /// Visits all nodes belonging to the interval [path_begin, path_end) in depth-first search order.
+    /// Returns vector of pairs with node hash and evaluation of predicate on node key for all nodes found.
+    /// Used to generate and apply state parts for state sync.
     pub(crate) fn visit_nodes_interval(
         &mut self,
         path_begin: &[u8],
         path_end: &[u8],
-    ) -> Result<Vec<CryptoHash>, StorageError> {
+        predicate: Option<&dyn Fn(&[u8]) -> bool>,
+    ) -> Result<Vec<(CryptoHash, bool)>, StorageError> {
```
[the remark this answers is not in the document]
Longarithm: Which Vec do you mean?
I've changed the implementation a bit, now my intention is to return Vec of pairs `(node_hash, Some(node_key) when some value is presented in the node)`.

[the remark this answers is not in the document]
EgorKulikov: I meant Vec of some struct, instead of tuple (but you can use angle brackets in github comments it seems), anyway, this is outdated by now

core/store/src/trie/iterator.rs
```diff
@@ -197,22 +208,25 @@ impl<'a> TrieIterator<'a> {
...
-    /// Returns hashes of nodes with paths in [path_begin, path_end). Used by state parts
+    /// Visits all nodes belonging to the interval [path_begin, path_end) in depth-first search order.
+    /// Returns vector of pairs with node hash and node key which is not None if node stores a value.
+    /// Used to generate and apply state parts for state sync.
     pub(crate) fn visit_nodes_interval(
         &mut self,
         path_begin: &[u8],
         path_end: &[u8],
-    ) -> Result<Vec<CryptoHash>, StorageError> {
+    ) -> Result<Vec<(CryptoHash, Option<Vec<u8>>)>, StorageError> {
```
bowenwang1996, line 218: I think it is better to introduce a struct here to have more clarity on what the return value is

core/store/src/trie/state_parts.rs
```diff
@@ -176,34 +177,44 @@ impl Trie {
...
-    /// Returns the storage changes for the state part.
+    /// Returns the storage changes for the state part and all contract codes extracted from it.
     /// Writing all storage changes gives the complete trie.
     pub fn apply_state_part(
         state_root: &StateRoot,
         part_id: u64,
         num_parts: u64,
         part: Vec<Vec<u8>>,
-    ) -> Result<TrieChanges, StorageError> {
+    ) -> Result<(TrieChanges, Vec<ContractCode>), StorageError> {
```
bowenwang1996, line 187: Same here

nearcore/src/runtime/mod.rs
```diff
@@ -500,6 +505,29 @@ impl NightshadeRuntime {
...
+            .par_iter()
+            .map(|code| {
+                precompile_contract(
+                    &code,
+                    &runtime_config.wasm_config,
+                    compiled_contract_cache.as_deref(),
+                )
+                .ok();
+            })
+            .collect::<()>();
```
bowenwang1996, line 528: What's the reason for not writing a loop here? The code here is hard to read.

nearcore/src/runtime/mod.rs
```diff
@@ -1413,14 +1441,26 @@ impl RuntimeAdapter for NightshadeRuntime {
...
         let tries = self.get_tries();
         let (store_update, _) =
             tries.apply_all(&trie_changes, shard_id).expect("TrieChanges::into never fails");
+
+        let protocol_version = self.get_epoch_protocol_version(epoch_id)?;
+        if checked_feature!(
+            "protocol_feature_precompile_contracts",
+            PrecompileContracts,
+            protocol_version
+        ) {
```
bowenwang1996, line 1460: Why do we need to do a protocol change here?

bowenwang1996 (commented):
I don't see why we need to change the protocol here. Compiling the contracts should not affect how the protocol works (without changing the gas cost).

core/store/src/trie/state_parts.rs
```diff
@@ -176,34 +177,44 @@ impl Trie {
...
         if state_root == &CryptoHash::default() {
-            return Ok(TrieChanges::empty(CryptoHash::default()));
+            return Ok((TrieChanges::empty(CryptoHash::default()), vec![]));
         }
-        let trie = Trie::from_recorded_storage(PartialStorage { nodes: PartialState(part) });
+        let trie =
+            Trie::from_recorded_storage(PartialStorage { nodes: PartialState(part.clone()) });
```
EgorKulikov, line 192: why do you need clone here? I do not see any more usages of part

nearcore/src/runtime/mod.rs
```diff
@@ -1413,14 +1441,17 @@ impl RuntimeAdapter for NightshadeRuntime {
...
         num_parts: u64,
         data: &[u8],
+        epoch_id: &EpochId,
     ) -> Result<(), Error> {
         let part = BorshDeserialize::try_from_slice(data)
             .expect("Part was already validated earlier, so could never fail here");
-        let trie_changes = Trie::apply_state_part(&state_root, part_id, num_parts, part)
-            .expect("combine_state_parts is guaranteed to succeed when each part is valid");
+        let (trie_changes, contract_codes) =
+            Trie::apply_state_part(&state_root, part_id, num_parts, part)
```
EgorKulikov, line 1449: I wonder why we even bother with Result here if we always just panic on error

runtime/runtime/src/lib.rs
```diff
@@ -1335,6 +1335,31 @@ impl Runtime {
...
+            .par_iter()
+            .map(|state_change| match state_change.trie_key {
+                TrieKey::ContractCode { .. } => {
+                    let code = state_change
+                        .changes
+                        .last()
+                        .expect("Committed entry should have at least one change")
+                        .data
+                        .as_ref();
+                    if let Some(code) = code {
```
EgorKulikov, line 1354: What happens if this is None and is it possible at all?

EgorKulikov (approved):
LGTM, see my and @bowenwang1996 comments though

nearcore/src/runtime/mod.rs
matklad: That's `par_iter()` -- rayon's parallel iterator. A for loop would be more readable, but sequential.

runtime/runtime/src/lib.rs
Longarithm: If code is None it means that it was just removed from the state, thus there is nothing to precompile. I will add comment to clarify it

core/store/src/trie/state_parts.rs
Longarithm: Yeah, it is an artifact of debugging.

nearcore/src/runtime/mod.rs
Longarithm: Could you tell more about what can be improved here?
I see that even on the `Trie::apply_state_part` level `StorageError` can be returned at many places, so it makes sense to process it here in one place.

nearcore/src/runtime/mod.rs
Longarithm: Also, I changed `map + collect` to `for_each`.

nearcore/src/runtime/mod.rs
EgorKulikov: maybe it would be better to have a wrapper function then?

runtime/runtime/src/lib.rs
```diff
@@ -1335,6 +1335,30 @@ impl Runtime {
...
+                    .as_ref();
+                // If code is None, it means that it was just removed from the state, so there is
+                // nothing to precompile.
+                if let Some(code) = code {
+                    let contract_code = ContractCode::new(code.clone(), None);
+                    precompile_contract(&contract_code, &wasm_config, cache).ok();
+                }
+            }
+            _ => {}
+        });
```
bowenwang1996, line 1360: We already compile contracts when they are deployed. You don't need to recompile them again here

nearcore/src/runtime/mod.rs
```diff
@@ -500,6 +504,27 @@ impl NightshadeRuntime {
...
+        &self,
+        epoch_id: &EpochId,
+        contract_codes: Vec<ContractCode>,
+    ) -> Result<(), Error> {
+        let protocol_version = self.get_epoch_protocol_version(epoch_id)?;
+        let runtime_config =
+            RuntimeConfig::from_protocol_version(&self.genesis_runtime_config, protocol_version);
+        let compiled_contract_cache: Option<Arc<dyn CompiledContractCache>> =
+            Some(Arc::new(StoreCompiledContractCache { store: self.store.clone() }));
+        contract_codes.par_iter().for_each(|code| {
```
bowenwang1996, line 518: I don't think using rayon here is necessarily a good idea. At least we should limit how many cores it can use at the same time. When we state sync during catch up, the node still needs to function normally and cannot afford to spend all its cpus on this.

bowenwang1996 (changes requested):

chain/client/tests/process_blocks.rs
```diff
@@ -3182,3 +3184,96 @@ mod cap_max_gas_price_tests {
...
+    use super::*;
+    use near_primitives::contract::ContractCode;
+    use near_primitives::types::CompiledContractCache;
+    use near_store::{Store, StoreCompiledContractCache};
+    use near_vm_runner::{get_contract_cache_key, VMKind};
+
+    /// Deploys test contract to the first client, perform state sync for the second client and
+    /// check existence of contract in both caches.
+    #[test]
+    fn test_caching() {
```
bowenwang1996, line 3199: Please also check that the client that does state sync can also invoke the contract (process a function call transaction) without any issues

bowenwang1996 (commented):
Please also test the following two cases:
* A contract is deployed to some account, and after some number of epochs, a new contract is deployed to the same account. When a node state syncs after the new contract is deployed, it will have the new contract compiled locally but not the old one.
* A contract is deployed to some account and after some number of epochs, the account is deleted. Then if a node state syncs after the account is deleted, it should not have the contract compiled locally.

nearcore/src/runtime/mod.rs
matklad: Question: is this part running concurrently with "functioning normally" part? 

As the caller of this code calls `store_update.commit()`, it seems like this actually blocks normal processing, and, as such, the faster we deal with it, the faster we can resume normal operation, and, from perf pov, more threads is better. 

That said, even if more threads is better for pert, we'd better to restrict the number of threads to not accidentally use more resources than we specify we need. Ie, the failure mode here is that, accidentally, near starts to require 4 cores rather than two. 

I guess, my advice is:

* figure out what our actual budget here is. Can we just compile everything in one thread? If not how many threads we need at minimum? 
* if there isn't strict budgeting, and we just want to go as fast as possible, we need to configure the global thread pool to use only 2 threads (or however many cores we say we need to run `neard`), to not be overly-optimistic about the actual performance. 
* If there is struct budgeting, and we need to hit it, we might do one of the two things:
  * spawn a single thread for all background work, and submit the work to this thread. That way, the work doesn't block processing, but is still done in an efficient manner, interleaving with other operations. In this apporach, it's important that there's a thread for "background work", and not just a thread for compilation -- otherwise we might end up, in the future, with several different background workers, where each worker individually tries to not over-subscribe, but alltogether they do. 
  * split the work in chunks. Rather than compiling all contracts in batch, have a queue on uncompiled contracts, and, on each iteraton of the event loop, pick up up to N kilobytes worth of contract and compiling those.

bowenwang1996: > As the caller of this code calls store_update.commit(), it seems like this actually blocks normal processing,

That is not the case. We also call this function in `apply_state_part`, which is called outside of normal block processing. I don't have a specific budget in mind, but probably not a good idea to use more than half of the available cpus.

Longarithm: I limited number of threads used here to `rayon::current_num_threads() / 2`.

bowenwang1996 (approved):

PR #4348, opened 2021-06-03
Take CapMaxGasPrice feature on nightly

Longarithm:
Actually update nightly protocol version to catch CapMaxGasPrice feature which I forgot to do here: https://github.com/near/nearcore/pull/4308


bowenwang1996 (commented):
How is `protocol_feature_block_header_v3` relevant here?

Longarithm:
@bowenwang1996 
I couldn't run `cap_max_gas_price_tests` because of compilation errors related to this feature. This should be related to recent refactoring of nearcore but I didn't find how yet.

EgorKulikov (approved):

bowenwang1996 (approved):

PR #4356, opened 2021-06-08
refactor(neard): switch to derive clap

### LINKED ISSUE #4347: Refactor `neard` CLI arguments handling to clap 3.0
Currently, the CLI arguments are handled ad-hoc and the error handling and parsing are verbose:

https://github.com/near/nearcore/blob/7c1080c1eef9b31cb6c0a06f5900962f59a53f2f/neard/src/main.rs#L85-L261

I propose we use the derive macro defining structs for subcommands and such from clap 3.0-beta: https://github.com/clap-rs/clap#using-derive-macros

Consider taking inspiration from https://github.com/frol/near-cli/blob/master/src/main.rs

austinabell:
- Refactors CLI to use the derive flavour of clap (which is basically just structopt)
- Separates subcommand into separate modules (to keep main concise and allow cleaner expansion on the CLI)
  - Commit [`8d822b8` (#4356)](https://github.com/near/nearcore/pull/4356/commits/8d822b82c2a94e0a023548d610039f1c871ed74f) which I'm happy to revert if wanted

I will decide now whether or not to
- Move some of the validation logic into these new structures, to avoid passing strings and parsing later

Any input would be appreciated what is ideal :)

I wrote a script to run on all sub-commands, and although testing functionality would be tedious, here is their outputs if anyone wants to diff:

<details>
  <summary>Before</summary>
  
```
NEAR Protocol Node 1.2.0 (build 07b4fd45) (protocol 45) (db 23)

USAGE:
    neard [OPTIONS] <SUBCOMMAND>

FLAGS:
    -h, --help       Prints help information
    -V, --version    Prints version information

OPTIONS:
        --home <home>          Directory for config and data (default "~/.near") [default: /home/austinabell/.near]
        --verbose <verbose>    Verbose logging

SUBCOMMANDS:
    help                 Prints this message or the help of the given subcommand(s)
    init                 Initializes NEAR configuration
    run                  Runs NEAR node
    testnet              Setups testnet configuration with all necessary files (validator key, node key, genesis and
                         config)
    unsafe_reset_all     (unsafe) Remove all the config, keys, data and effectively removing all information about
                         the network
    unsafe_reset_data    (unsafe) Remove all the data, effectively resetting node to genesis state (keeps genesis
                         and config)
neard-init 
Initializes NEAR configuration

USAGE:
    neard init [FLAGS] [OPTIONS]

FLAGS:
        --download-genesis    Download the verified NEAR genesis file automatically.
        --fast                Makes block production fast (TESTING ONLY)
    -h, --help                Prints help information
    -V, --version             Prints version information

OPTIONS:
        --account-id <account-id>                        Account ID for the validator key
        --chain-id <chain-id>                            Chain ID, by default creates new random
        --download-genesis-url <download-genesis-url>    Specify a custom download URL for the genesis-file.
        --genesis <genesis>
            Genesis file to use when initialize testnet (including downloading)

        --num-shards <num-shards>                        Number of shards to initialize the chain with
        --test-seed <test-seed>                          Specify private key generated from seed (TESTING ONLY)
neard-testnet 
Setups testnet configuration with all necessary files (validator key, node key, genesis and config)

USAGE:
    neard testnet [OPTIONS]

FLAGS:
    -h, --help       Prints help information
    -V, --version    Prints version information

OPTIONS:
        --n <n>              Number of non-validators to initialize the testnet with (default 0)
        --prefix <prefix>    Prefix the directory name for each node with (node results in node0, node1, ...) (default
                             "node")
        --shards <s>         Number of shards to initialize the testnet with (default 4)
        --v <v>              Number of validators to initialize the testnet with (default 4)
neard-run 
Runs NEAR node

USAGE:
    neard run [FLAGS] [OPTIONS]

FLAGS:
        --archive    Keep old blocks in the storage (default false)
    -h, --help       Prints help information
    -V, --version    Prints version information

OPTIONS:
        --boot-nodes <boot-nodes>                        Set the boot nodes to bootstrap network from
        --min-peers <min-peers>                          Minimum number of peers to start syncing / producing blocks
        --network-addr <network-addr>
            Customize network listening address (useful for running multiple nodes on the same machine)

        --produce-empty-blocks <produce-empty-blocks>
            Set this to false to only produce blocks when there are txs or receipts (default true)

        --rpc-addr <rpc-addr>
            Customize RPC listening address (useful for running multiple nodes on the same machine)

        --telemetry-url <telemetry-url>                  Customize telemetry url
neard-unsafe_reset_data 
(unsafe) Remove all the data, effectively resetting node to genesis state (keeps genesis and config)

USAGE:
    neard unsafe_reset_data

FLAGS:
    -h, --help       Prints help information
    -V, --version    Prints version information
neard-unsafe_reset_all 
(unsafe) Remove all the config, keys, data and effectively removing all information about the network

USAGE:
    neard unsafe_reset_all

FLAGS:
    -h, --help       Prints help information
    -V, --version    Prints version information
```
</details>

<details>
  <summary>After</summary>
  
```
neard 1.2.0 (build 6ee4308a) (protocol 45) (db 23)
NEAR Protocol Node

USAGE:
    neard [OPTIONS] <SUBCOMMAND>

FLAGS:
    -h, --help       Prints help information
    -V, --version    Prints version information

OPTIONS:
        --home <home>          Directory for config and data (default "~/.near")
        --verbose <verbose>    Verbose logging

SUBCOMMANDS:
    help                 Prints this message or the help of the given subcommand(s)
    init                 Initializes NEAR configuration
    run                  Runs NEAR node
    testnet              Sets up testnet configuration with all necessary files (validator key,
                         node key, genesis and config)
    unsafe_reset_all     (unsafe) Remove all the config, keys, data and effectively removing all
                         information about the network
    unsafe_reset_data    (unsafe) Remove all the data, effectively resetting node to the genesis
                         state (keeps genesis and config)
neard-init 
Initializes NEAR configuration

USAGE:
    neard init [FLAGS] [OPTIONS]

FLAGS:
        --download-genesis    Download the verified NEAR genesis file automatically
        --fast                Makes block production fast (TESTING ONLY)
    -h, --help                Prints help information
    -V, --version             Prints version information

OPTIONS:
        --account-id <account-id>                        Account ID for the validator key
        --chain-id <chain-id>                            Chain ID, by default creates new random
        --download-genesis-url <download-genesis-url>
            Specify a custom download URL for the genesis-file

        --genesis <genesis>
            Genesis file to use when initializing testnet (including downloading)

        --num-shards <num-shards>
            Number of shards to initialize the chain with [default: 1]

        --test-seed <test-seed>
            Specify private key generated from seed (TESTING ONLY)

neard-testnet 
Sets up testnet configuration with all necessary files (validator key, node key, genesis and config)

USAGE:
    neard testnet [OPTIONS]

FLAGS:
    -h, --help       Prints help information
    -V, --version    Prints version information

OPTIONS:
        --n <non-validators>    Number of non-validators to initialize the testnet with [default: 0]
        --prefix <prefix>       Prefix the directory name for each node with (node results in node0,
                                node1, ...) [default: node]
        --shards <shards>       Number of shards to initialize the testnet with [default: 4]
        --v <validators>        Number of validators to initialize the testnet with [default: 4]
neard-run 
Runs NEAR node

USAGE:
    neard run [FLAGS] [OPTIONS]

FLAGS:
        --archive    Keep old blocks in the storage (default false)
    -h, --help       Prints help information
    -V, --version    Prints version information

OPTIONS:
        --boot-nodes <boot-nodes>
            Set the boot nodes to bootstrap network from

        --min-peers <min-peers>
            Minimum number of peers to start syncing/producing blocks

        --network-addr <network-addr>
            Customize network listening address (useful for running multiple nodes on the same
            machine)

        --produce-empty-blocks <produce-empty-blocks>
            Set this to false to only produce blocks when there are txs or receipts (default true)

        --rpc-addr <rpc-addr>
            Customize RPC listening address (useful for running multiple nodes on the same machine)

        --telemetry-url <telemetry-url>                  Customize telemetry url
neard-unsafe_reset_data 
(unsafe) Remove all the data, effectively resetting node to the genesis state (keeps genesis and
config)

USAGE:
    neard unsafe_reset_data

FLAGS:
    -h, --help       Prints help information
    -V, --version    Prints version information
neard-unsafe_reset_all 
(unsafe) Remove all the config, keys, data and effectively removing all information about the
network

USAGE:
    neard unsafe_reset_all

FLAGS:
    -h, --help       Prints help information
    -V, --version    Prints version information
```
</details>

Closes #4347


austinabell:
So the tests are failing primarily because some CLI commands are run with syntax like: `--chain-id=` without any value, and the new version of Clap errors with this (can support `--chain-id` without a value, the `=` is causing issues) without any clear way to get around.

Also, the other issue is that the `verbose` and `home` flags are not included in the subcommands, but are still parsed in the previous impl. This is more easily fixed but was just misleading that the help docs were incorrect

neard/src/main.rs
```diff
@@ -1,25 +1,41 @@
...
 use near_primitives::version::{Version, DB_VERSION, PROTOCOL_VERSION};
 #[cfg(feature = "memory_stats")]
 use near_rust_allocator_proxy::allocator::MyAllocator;
-use nearcore::config::{init_testnet_configs, load_config_without_genesis_records};
-use nearcore::genesis_validate::validate_genesis;
-use nearcore::{get_default_home, get_store_path, init_configs, start_with_config};
+use nearcore::{get_default_home, get_store_path};
+
+#[macro_use]
+extern crate lazy_static;
```
matklad, line 25: Don't use `extern crate` syntax, it is not needed in 2018 edition. Instead, `use lazy_static::lazy_static`.

neard/src/main.rs
```diff
@@ -1,25 +1,41 @@
...
+        version: crate_version!().to_string(),
+        build: git_version!(fallback = "unknown").to_string(),
+    };
+    static ref NEARD_VERSION_STRING: String = {
+        format!(
+            "{} (build {}) (protocol {}) (db {})",
+            NEARD_VERSION.version, NEARD_VERSION.build, PROTOCOL_VERSION, DB_VERSION
+        )
+    };
+}
```
matklad, line 38: It's a shame that clap's API forces us to use lazy static for a version, but I don't see a good solution here.

matklad (commented):
LGTM, this is a big improvement, thanks for tackling this!

Rather than keeping each command in a separate module, I'd create a single `app.rs` which contains definitions of both `NeardOpts` and all other subcommands. That way, it's easy to see the whole CLI at a glance and make sure that, eg, common options across different subcommands are named consistently. But I don't feel particularly strongly about this.

khorolets (approved):
Wow, I'm happy you found time to refactor this!

neard/src/main.rs
austinabell: oops!

austinabell:
> Rather than keeping each command in a separate module, I'd create a single `app.rs` which contains definitions of both `NeardOpts` and all other subcommands. That way, it's easy to see the whole CLI at a glance and make sure that, eg, common options across different subcommands are named consistently. But I don't feel particularly strongly about this.

Yeah, that can definitely make sense, I'll throw them in a `cli.rs` file. This solves my want to keep the main file clean while also keeping CLI commands together, I like it!

Also, I still have to resolve the syntax allowed between this version as this isn't backwards compatible yet, so I'll do that before opening the PR again :)

austinabell:
Okay so I can't seem to find a way to make the `--<option>=` without a value syntax work to be completely backwards compatible, but I will try to outline possible options and leave it to others to choose the direction

1. Keep the strict validation as is and not allow using flag if no value passed.

In this case, the flags cannot be passed unless there is a value associated with it.

I personally think this is the best way to avoid misuse because it's not clear that the flag does nothing, but this is the most breaking of these options.

2. Allow the flag to be used, but just disallow the `=` variant of the flag usage without a value

This can be some by wrapping the optional flags with another option (`Option<Option<T>>` and flattening).

 `neard init --chain-id` would still parse (as long as it doesn't have the `=`)

3. Keep the CLI as is to not make a breaking change (or someone try to get the syntax to parse using the derive syntax, I've set up a simple test)

The test I added is `equal_no_value_syntax`, which uses the exact arguments used in one of the CI tests, so feel free to play around with it :)

frol:
I vote for fixing https://github.com/near/nearcore/blob/07b4fd45f90570d9c4f306f724cf84499cc519c7/scripts/start_unittest.py#L40 since it is super ugly and hacky.

austinabell:
> I vote for fixing
> 
> https://github.com/near/nearcore/blob/07b4fd45f90570d9c4f306f724cf84499cc519c7/scripts/start_unittest.py#L40
> 
> since it is super ugly and hacky.
> My first through is to introduce a separate boolean flag like `--random-chain-id`, but feel free to suggest other options

What is the tolerance for breakages of the neard CLI? Externally there could be scripts that expect this syntax (it's written a few times in docs throughout the codebase) and I want to be sure this either would come with a breaking release or have some way to mitigate this difference.

The `--random-chain-id` flag can be an option, and I'm happy to do this, but it doesn't help the backwards compatibility issue

frol (approved):
This is beautiful! :heart_eyes:

frol:
> What is the tolerance for breakages of the neard CLI? Externally there could be scripts that expect this syntax (it's written a few times in docs throughout the codebase) and I want to be sure this either would come with a breaking release or have some way to mitigate this difference.

I think we can tolerate breaking changes in non-`run` subcommands, so this `init` subcommand is ok to be broken as long as we update all the references.

austinabell:
> > What is the tolerance for breakages of the neard CLI? Externally there could be scripts that expect this syntax (it's written a few times in docs throughout the codebase) and I want to be sure this either would come with a breaking release or have some way to mitigate this difference.
> 
> I think we can tolerate breaking changes in non-`run` subcommands, so this `init` subcommand is ok to be broken as long as we update all the references.

Okay cool sounds good, and just to be clear, the `run` command does have the same breakages if called into this same way but can mitigate these as much as possible (point 2 above) if some breakages are acceptable.

frol:
I don't believe anyone used that hack in `run` command. Let's make this breaking change since it looks super odd anyway

austinabell:
> I don't believe anyone used that hack in `run` command. Let's make this breaking change since it looks super odd anyway

I'd love some input from you or others whether or not the flag without a value should be allowed

for example, the `telemetry-url` flag, should this be accepted:
```bash
neard run --telemetry-url
```
Which would behave the same as not passing the flag? On one hand, it doesn't make logical sense to use through a CLI, but from a script, it could be quicker to handle optional values rather than checking existence before including

frol:
I don't like `--telemetry-url` flag meaning that there should be no telemetry URL set, if any, it should be `--no-telemetry` flag then.

austinabell:
> I don't like `--telemetry-url` flag meaning that there should be no telemetry URL set, if any, it should be `--no-telemetry` flag then.

Sorry I suppose I misworded, I just mean the flag would do nothing, as in:
`neard run --telemetry-url` == `neard run`

Happy to keep it strict validating and fix the weird usage scripts and docs tomorrow if agreed upon, just want to be clear about the tradeoffs :)

frol:
@chefsale @bowenwang1996 any objections about breaking changes to weird use-cases to solidify our CLI interface?

austinabell:
My Python is rusty, please double check [`ec7cb0c` (#4356)](https://github.com/near/nearcore/pull/4356/commits/ec7cb0c18df5ac19f77b7996a8d19b4265d34654).

Also noticed the `get_chain_id_from_flags` seems borked, if someone can confirm I can open an issue?

frol (approved):
LGTM :+1:

neard/src/cli.rs
```diff
@@ -0,0 +1,322 @@
...
+    opts: NeardOpts,
+    #[clap(subcommand)]
+    subcmd: NeardSubCommand,
+}
+
+impl NeardCmd {
+    pub(super) fn parse_and_run() {
+        let neard_cmd = Self::parse();
+        neard_cmd.opts.init();
+        info!(target: "near", "Version: {}, Build: {}, Latest Protocol: {}", NEARD_VERSION.version, NEARD_VERSION.build, PROTOCOL_VERSION);
```
bowenwang1996, line 29: should we change target to `neard`?

neard/src/cli.rs
```diff
@@ -0,0 +1,322 @@
...
+        let sys = actix::System::new();
+        sys.block_on(async move {
+            nearcore::start_with_config(home_dir, near_config);
+        });
+        sys.run().unwrap();
+    }
+}
+
+#[derive(Clap)]
+pub(super) struct TestnetCmd {
```
bowenwang1996, line 223: `Testnet` is a misnomer here. The intentional is to create a local network for testing. This should not be confused with the testnet that has many validators.

bowenwang1996 (commented):
Looks good overall. Would be great if we can address the (longstanding) nomenclature clash

neard/src/cli.rs
frol: Should we remove this subcommand in the future altogether? I don't want this PR to introduce too many breaking changes. Let's create a separate issue for that. I was always confused by that `testnet` subcommand.

chefsale: Make it a new bootcamp issue 😸

chefsale:
I believe as long as we don't break init and run functionality I'm fine with the changes. If we broke these two, that would be pretty bad for all the automation we and partners have in place. 😸

chefsale (approved):
Otherwise LGTM.

austinabell:
> I believe as long as we don't break init and run functionality I'm fine with the changes. If we broke these two, that would be pretty bad for all the automation we and partners have in place.

Just so everyone is clear, this does break interactions depending on how the CLI is used because it has stricter validation. Flags that take a value no longer allow including the flag if no value is passed (it broke some of the scripts within the codebase that were set up strangely).

As for the internal functionality aside from parsing, nothing should be different, unless I'm missing some indirect implications that should not be upheld

chefsale:
I believe that should be fine. I don't think any of our critical automation passes empty flags on init or run.

bowenwang1996:
> In this PR or in a following one? My goal was to have it be as close to the functionality that currently exists, but happy to make whatever changes are agreed upon and when :)

Let's do it in a separate PR, as @frol suggested.

frol:
@olonho @ailisp @evgenykuzyakov Please, review

bowenwang1996 (approved):

ailisp (approved):
Nice refactor!

chefsale:
@austinabell I believe you have all the needed approvals to land this, please resolve the conflict and feel free to merge it.

bowenwang1996:
We still need approval from @olonho

neard/Cargo.toml
```diff
@@ -14,13 +14,14 @@ path = "src/../src/main.rs"
 name = "near"
 
 [dependencies]
-clap = "2.32"
+clap = "=3.0.0-beta.2"
```
olonho, line 17: Any specific reason for exact version dependency?

neard/src/cli.rs
olonho: Sounds reasonable.

neard/src/cli.rs
```diff
@@ -0,0 +1,335 @@
...
+
+        match neard_cmd.subcmd {
+            NeardSubCommand::Init(cmd) => cmd.run(&home_dir),
+            NeardSubCommand::Testnet(cmd) => cmd.run(&home_dir),
+            NeardSubCommand::Run(cmd) => cmd.run(&home_dir),
+
+            NeardSubCommand::UnsafeResetData => {
+                let store_path = get_store_path(&home_dir);
+                info!(target: "near", "Removing all data from {}", store_path);
+                fs::remove_dir_all(store_path).expect("Removing data failed");
```
olonho, line 52: Wow, couldn't it be dangerous? Shall we validate `store_path` to be meaningful file?

olonho (approved):
Estimator part is OK, few nits in other places.

neard/src/cli.rs
```diff
@@ -0,0 +1,335 @@
...
+    pub(super) fn parse_and_run() {
+        let neard_cmd = Self::parse();
+        neard_cmd.opts.init();
+        info!(target: "near", "Version: {}, Build: {}, Latest Protocol: {}", NEARD_VERSION.version, NEARD_VERSION.build, PROTOCOL_VERSION);
+
+        #[cfg(feature = "adversarial")]
+        {
+            error!("THIS IS A NODE COMPILED WITH ADVERSARIAL BEHAVIORS. DO NOT USE IN PRODUCTION.");
+
+            if env::var("ADVERSARY_CONSENT").unwrap_or_else(|_| "".to_string()) != "1" {
```
olonho, line 35: Maybe
```
if env::var("ADVERSARY_CONSENT").unwrap_or_default() != "1" {
        std::process::exit(1);
    }
```
(thanks @matklad ).

neard/Cargo.toml
frol: I guess it is a good idea to pin pre-release versions to avoid breaking changes causing compilation errors. It is less relevant for binary crates, though. Still, I guess it is a good rule of thumb.

neard/src/cli.rs
frol: Yeah, it is great that now we can easily spot such cruel shortcuts. Let's create a separate issue to address this since the implementation is not new, and it has been part of neard for quite a long time already. I guess we should printing the folder that is about to be removed and add a confirmation dialog.

frol:
@austinabell Please, review the comments and extract non-relevant ones into separate issues (I believe we can have a few new `bootcamp` issues from here :smile:). Apply `automerge` label to this PR once you are ready.

neard/Cargo.toml
austinabell: Yeah, I'm just not confident that there will be no breaking changes coming from the beta release so my intuition is that this would be more safe. This isn't used as a lib so the pin shouldn't matter

chefsale:
@austinabell Please for the new issues add a bootcamp label and add them to this PR in the comments so we can classify them properly.

austinabell:
> @austinabell Please for the new issues add a bootcamp label and add them to this PR in the comments so we can classify them properly.

I was hesitant to add that label because a few of them were ambiguous and not defined clearly, but I'll link forward to each issue in comments now

neard/src/cli.rs
austinabell: #4369

neard/src/cli.rs
austinabell: #4372

PR #4360, opened 2021-06-11
Set block protocol version in a separate function in process_blocks

Longarithm:
`set_block_protocol_version` was used several times in the same way, made a separate function for it. 
Also simplify `process_blocks_with_storage_usage_fix`


bowenwang1996 (approved):
Nice!

PR #4364, opened 2021-06-14
feat(jsonrpc): stabilize `EXPERIMENTAL_tx_status` endpoint into previously existed `tx` (no breaking changes)

frol:
I reviewed the `EXPERIMENTAL_tx_status` changes: `receipts` field was added which only exposes non-local receipts and its structure is similar to the `transaction` structure. It is yet to be desired to use tagged enum representations, but given that `transaction` structure is already there using non-tagged representation, it is better to keep them aligned.

Checkpoints:

* [x] The `EXPERIMENTAL_` endpoint is left in the code-base for compatibility reasons (we don't need to remove it and existing clients may keep using it just fine)
* [x] CHANGELOG is updated
* [x] jsonrpc-client is updated (it has a minor breaking change, but it is an internal crate used for testing purposes)

# Test Plan

* [ ] All existing tests pass
* [x] Updated near-api-js to be compatible with the change: https://github.com/near/near-api-js/pull/623 (it expected that `broadcast_tx_commit` output will exactly match `tx` status endpoint output)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3274,7 +3274,7 @@ dependencies = [
 
 [[package]]
 name = "near-jsonrpc"
-version = "0.2.1"
+version = "0.2.2"
 dependencies = [
  "actix",
  "actix-cors",
@@ -3306,7 +3306,7 @@ dependencies = [
 
 [[package]]
 name = "near-jsonrpc-client"
-version = "0.1.0"
+version = "0.2.2"
 dependencies = [
  "actix-web",
  "awc",
--- a/chain/chain/src/chain.rs
+++ b/chain/chain/src/chain.rs
@@ -1059,9 +1059,9 @@ impl Chain {
             Ok((head, needs_to_start_fetching_state)) => {
                 chain_update.chain_store_update.save_block_height_processed(block_height);
                 chain_update.commit()?;
-            
+
                 self.pending_states_to_patch = None;
-                
+
                 if needs_to_start_fetching_state {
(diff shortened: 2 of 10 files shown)


chefsale (approved):
Lgtm

khorolets (approved):

frol:
@bowenwang1996 Do you have any obvious place I should look into to fix those timing out tests of tx status with light-client? https://buildkite.com/nearprotocol/nearcore/builds/1547#0e808822-7fec-4270-9868-cd58932339dd

bowenwang1996 (commented):
Can you check what the `result` is in the `map_ok` block? It is possible that somehow receipts cannot be fetched

frol:
@bowenwang1996 light client routes the request and our network layer is only capable to fetch transactions without receipts. This means that we need to do extend the network layer with the command if we want `tx` to return `receipts` :disappointed: 

@mikedotexe Stabilization of `EXPERIMENTAL_tx_status` is not going to make it to 1.20.0 as we uncovered a problem with `tx`.

bowenwang1996:
> @bowenwang1996 light client routes the request and our network layer is only capable to fetch transactions without receipts. This means that we need to extend the network protocol with a new command or patch the existing one if we want `tx` to return `receipts` 😞

That is unfortunately. Maybe we should rethink the design a little bit. I am not sure whether we should even route the request in the first place.

bowenwang1996:
@frol should we close this PR?

PR #4365, opened 2021-06-15
Increase JsonRpcClient payload receivable size to 100mb

### LINKED ISSUE #4241: Increase Rust JSON-RPC client limit on receiving payload
**Describe the bug**

Currently, our jsonrpc-client has a limit of 256k on receiving payload, which causes Overflow errors on trying to download contract code, state, or state changes even when a server is fine sending such payloads. We need to configure the limit [here](https://github.com/near/nearcore/blob/1b77673548ebaf884e25169dbafae25ce4713472/chain/jsonrpc/client/src/lib.rs#L43-L46) somehow.

**To Reproduce**

1. Clone https://github.com/FroVolod/near-cli
2. `cargo run -- view contract-code network testnet contract meta.pool.testnet download ./meta.pool.testnet.wasm`

You will get:

```
Failed to fetch query for view contract: RpcError { code: -32700, message: "Parse error", data: Some(String("Failed to retrieve payload: Overflow")) }
```

Here is the code that uses jsonrpc-client: https://github.com/FroVolod/near-cli/blob/5d32f348e727ae52053a2b044cf67f11c1ad8938/src/commands/view/view_contract_code/contract/download_mode/download/mod.rs#L50-L57

**Expected behavior**

It should download the payload just fine, and it does it for smaller payloads (contracts).

- nearcore: any
- rust version (if local): any
- docker (if using docker): n/a
- mainnet/testnet/betanet/local: any

ChaoticTempest:
fixes #4241

### DIFF
--- a/chain/jsonrpc/client/src/lib.rs
+++ b/chain/jsonrpc/client/src/lib.rs
@@ -27,9 +27,9 @@ pub enum ChunkId {
 /// Timeout for establishing connection.
 const CONNECT_TIMEOUT: Duration = Duration::from_secs(30);
 
-/// Max size of payload for JsonRpcClient. Set to max usize since the server was already 
-/// fine sending it at whatever size it chose. The client just cares about receiving it.
-const PAYLOAD_LIMIT: usize = usize::MAX;
+/// Max size of the payload JsonRpcClient can receive. Be careful adjusting this value since
+/// smaller values can raise overflow messages.
+const PAYLOAD_LIMIT: usize = 100 * 1024 * 1024;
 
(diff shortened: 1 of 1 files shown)


khorolets (approved):

frol (approved):

bowenwang1996 (commented):
@frol 100MB seems dangerous. A rpc node can be easily attacked if we allow 100MB payloads

frol:
@bowenwang1996 that is client side pulling the payload, and previously the limit was 256k, which was not even sufficient to download a contract binary.

bowenwang1996:
I see

Issue #4367, opened 2021-06-15
Cargo build failure is not covered by buildkite/nearcore


Longarithm:
https://github.com/near/nearcore/commit/377908d18e0d4ed46eab2f0cdabf3f9a648179e3 successfully passed all buildkite/nearcore checks, but `cargo build -p neard --features nightly_protocol --features nightly_protocol_features` failed here: https://buildkite.com/nearprotocol/nearcore-nightly-release/builds/382

It led to the betanet crash. IMO adding corresponding check to the `buildkite/nearcore` would make such errors easier to catch.

PR #4368, opened 2021-06-15
Cover cargo build failures in buildkite/nearcore

Longarithm:
Attempt to solve https://github.com/near/nearcore/issues/4367

Check both stable and nightly `cargo build -p neard`

Checked that it covers previous commit:
```
$ git checkout 377908d18e0d4ed46eab2f0cdabf3f9a648179e3
$ cargo check -p neard --features nightly_protocol,nightly_protocol_features
error: failed to select a version for `near-client`.
    ... required by package `nearcore v1.2.0 (/home/Aleksandr1/nearcore/nearcore)`
    ... which is depended on by `neard v1.2.0 (/home/Aleksandr1/nearcore/neard)`
versions that meet the requirements `=0.1.0` are: 0.1.0

the package `nearcore` depends on `near-client`, with features: `nearcore` but `near-client` does not have these features.
```

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -29,7 +29,7 @@ steps:
       fi
       RUSTFLAGS='-D warnings' cargo check --workspace --all-targets --all-features
       RUSTFLAGS='-D warnings' cargo check -p neard
-      RUSTFLAGS='-D warnings' cargo check -p neard --features
+      RUSTFLAGS='-D warnings' cargo check -p neard --features nightly_protocol,nightly_protocol_features
       # build a sandbox node should succeed
       RUSTFLAGS='-D warnings' cargo check -p neard --features sandbox
       python3 scripts/state/update_res.py check
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -28,6 +28,8 @@ steps:
         cargo-deny --all-features check bans
       fi
       RUSTFLAGS='-D warnings' cargo check --workspace --all-targets --all-features
+      RUSTFLAGS='-D warnings' cargo check -p neard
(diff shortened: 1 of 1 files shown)


chefsale:
I have started both nightly and normal release pipeline on buildkite just to verify this works: 
https://buildkite.com/nearprotocol/nearcore-nightly-release/builds/387
https://buildkite.com/nearprotocol/nearcore-release/builds/628

chefsale (approved):
Both builds have succeeded. Thanks for fixing this. Ship it 🗡️

.buildkite/pipeline.yml
```diff
@@ -28,6 +28,8 @@ steps:
         cargo-deny --all-features check bans
       fi
       RUSTFLAGS='-D warnings' cargo check --workspace --all-targets --all-features
+      RUSTFLAGS='-D warnings' cargo check -p neard
+      RUSTFLAGS='-D warnings' cargo check -p neard --features nightly_protocol,nightly_protocol_features
```
bowenwang1996, line 32: This feels repetitive to me. I think we don't have to run all these checks together. cc @matklad

matklad: Syntactically, `RUSTFLAGS='-D warnings'` is repetitive. It'd be cool to set it once globally for all jobs, not sure if buildkite syntax allows for that. 

Semantically, the two extra jobs are not repetitive. It's totally possible that `cargo check --workspace` works, but `cargo check -p neard` doesn't, due to how features are activated in Cargo. The backstory here is that we did broke release build, which we didn't notice because CI was only testing workspace mode. 

Cost-wise, I'd expect the two additional checks to be realtively cheap -- they should be able to re-use artifacts for all of the dependencies.

PR #4373, opened 2021-06-16
refactor(indexer): Optimize delayed local receipts tracking in Indexer Framework

### LINKED ISSUE #4317: Optimize delayed local receipts tracking in Indexer Framework
In #4316, we hot-fixed fetching local receipts that were delayed due to network congestion, but it would be ideal to track delayed local receipts in intermediate in-memory storage (hashmap), so we don't need to rebuild the whole streamer message for the previous blocks just to look up for a single local receipt.

Obviously, if we start from the block that already has an execution outcome for a delayed local receipt (and the list of tracked local receipts that are delayed is still empty), we can fall back to this current implementation.

# Context

Local receipts are not stored in the storage (e.g. `EXPERIMENTAL_receipts` method cannot provide any information about them)

khorolets:
Fixes #4317

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3254,7 +3254,7 @@ dependencies = [
 
 [[package]]
 name = "near-indexer"
-version = "0.9.1"
+version = "0.9.2"
 dependencies = [
  "actix",
  "async-recursion",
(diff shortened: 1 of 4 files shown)


frol (approved):

frol:
@khorolets run cargo check and commit Cargo.lock file (it is changed due to the version bump in Cargo.toml)

PR #4376, opened 2021-06-17
feat(primitive): prepare for adding profile data to execution outcome

ailisp:
This PR does not introduce any functional change. It adds a metadata field to ExecutionOutcome which is versioned. Add a V1 ExecutionMetadata with is empty. Next PR will add a V2 ExecutionMetadata which add ProfileData into ExecutionOutcome. Add migrations to change exisiting outcome in storage to new format

Test Plan
------------
CI

### DIFF
--- a/core/primitives/src/transaction.rs
+++ b/core/primitives/src/transaction.rs
@@ -285,7 +285,7 @@ impl Default for ExecutionStatus {
     }
 }
 
-/// ExecutionOutcome for proof. Excludes logs.
+/// ExecutionOutcome for proof. Excludes logs and metadata
 #[derive(BorshSerialize, BorshDeserialize, Serialize, PartialEq, Clone)]
 struct PartialExecutionOutcome {
     pub receipt_ids: Vec<CryptoHash>,
@@ -348,6 +348,20 @@ pub struct ExecutionOutcome {
     /// NOTE: Should be the latest field since it contains unparsable by light client
     /// ExecutionStatus::Failure
     pub status: ExecutionStatus,
+    /// Execution metadata, versioned
+    pub metadata: ExecutionMetadata,
+}
+
+#[derive(BorshSerialize, BorshDeserialize, Serialize, Deserialize, PartialEq, Clone, Eq, Debug)]
+pub enum ExecutionMetadata {
+    // V1: Empty Metadata
+    ExecutionMetadataV1,
+}
+
+impl Default for ExecutionMetadata {
+    fn default() -> Self {
+        ExecutionMetadata::ExecutionMetadataV1
+    }
 }
 
 impl ExecutionOutcome {
@@ -370,6 +384,7 @@ impl fmt::Debug for ExecutionOutcome {
             .field("burnt_gas", &self.gas_burnt)
             .field("tokens_burnt", &self.tokens_burnt)
             .field("status", &self.status)
+            .field("meatdata", &self.metadata)
             .finish()
     }
 }
@@ -508,6 +523,7 @@ mod tests {
             gas_burnt: 123,
             tokens_burnt: 1234000,
             executor_id: "alice".to_string(),
+            metadata: ExecutionMetadata::ExecutionMetadataV1,
         };
         let hashes = outcome.to_hashes();
         assert_eq!(hashes.len(), 3);
--- a/core/primitives/src/views.rs
+++ b/core/primitives/src/views.rs
@@ -39,8 +39,8 @@ use crate::sharding::{ChunkHash, ShardChunk, ShardChunkHeader, ShardChunkHeaderI
 use crate::sharding::{ShardChunkHeaderInnerV2, ShardChunkHeaderV3};
 use crate::transaction::{
     Action, AddKeyAction, CreateAccountAction, DeleteAccountAction, DeleteKeyAction,
-    DeployContractAction, ExecutionOutcome, ExecutionOutcomeWithIdAndProof, ExecutionStatus,
-    FunctionCallAction, SignedTransaction, StakeAction, TransferAction,
+    DeployContractAction, ExecutionMetadata, ExecutionOutcome, ExecutionOutcomeWithIdAndProof,
+    ExecutionStatus, FunctionCallAction, SignedTransaction, StakeAction, TransferAction,
 };
 use crate::types::{
     AccountId, AccountWithPublicKey, Balance, BlockHeight, CompiledContractCache, EpochHeight,
@@ -1017,6 +1017,8 @@ pub struct ExecutionOutcomeView {
     pub executor_id: AccountId,
     /// Execution status. Contains the result in case of successful execution.
     pub status: ExecutionStatusView,
+    /// Execution metadata, versioned
+    pub metadata: ExecutionMetadata,
 }
 
 impl From<ExecutionOutcome> for ExecutionOutcomeView {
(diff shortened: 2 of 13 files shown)


matklad:
Thanks for picking up this work from me @ailisp and sorry for blocking your work on sandbox with it.

I have two high-level commets here:

* we need a migration to back-fill the metadata for all existing execution outcomes. 
* I believe this should *not* be a protocol feature -- the idea about metadata is exactly to store the info which is *not* part of the protocol.

bowenwang1996 (commented):
yeah same as what @matklad commented

ailisp:
@matklad @bowenwang1996 ptal :)

core/primitives/src/transaction.rs
```diff
@@ -348,6 +348,20 @@ pub struct ExecutionOutcome {
...
+pub enum ExecutionMetadata {
+    // V1: Empty Metadata
+    ExecutionMetadataV1,
+}
+
+impl Default for ExecutionMetadata {
+    fn default() -> Self {
+        ExecutionMetadata::ExecutionMetadataV1
+    }
 }
```
matklad, line 365: I think it's better to remove this default impl -- we want to include metadata with every outcome, so we'll have to explicitly create a metadata instance at every call-site anyway.

core/store/src/migrations.rs
```diff
@@ -790,3 +794,70 @@ pub fn migrate_18_to_new_validator_stake(store: &Store) {
...
+    let outcomes: Vec<(_, Vec<OldExecutionOutcomeWithIdAndProof>)> = store
+        .iter(DBCol::ColTransactionResult)
+        .map(|(key, value)| {
+            (
+                key,
+                Vec::<OldExecutionOutcomeWithIdAndProof>::try_from_slice(&value)
+                    .expect("BorshDeserialize should not fail"),
+            )
+        })
+        .collect();
```
matklad, line 853: This loads all the outcomes to memory in the same time, is this OK to do? I'd expect that we either:

* iterate and update records in-place
* get the full key-set at once, and then iterate one by one


This might be premature optimization though, I'll let @bowenwang1996 to make a judgment call on whether we should bother with O(1) memory solution here.

ailisp: Good point

core/primitives/src/transaction.rs
ailisp: I remembered the reason for this is ExecutionOutcome and structs containing ExecutionOutcome are deriving(Default), so ExecutionMetadata has to impl Default

matklad: Yeah, than our hand is forced here! I wonder what is the root cause of ExecutionOutcome needing to be Default, but that's out of scope here!

EgorKulikov (approved):
LGTM, although I would prefer in-place update instead of loading everything to memory

core/primitives/src/transaction.rs
```diff
@@ -348,6 +348,20 @@ pub struct ExecutionOutcome {
...
     pub status: ExecutionStatus,
+    /// Execution metadata, versioned
+    pub metadata: ExecutionMetadata,
+}
+
+#[derive(BorshSerialize, BorshDeserialize, Serialize, Deserialize, PartialEq, Clone, Eq, Debug)]
+pub enum ExecutionMetadata {
+    // V1: Empty Metadata
+    ExecutionMetadataV1,
+}
```
bowenwang1996, line 359: @matklad do we need a enum here? I think it might be better to just have a struct and have custom borsh serializer/deserializer

core/store/src/migrations.rs
bowenwang1996: We absolutely cannot load all of them in memory. In fact, I am not sure what the migration is for. Are we going to repopulate the old execution outcomes with metadata? If not, do we actually need to have this migration? Could we implement a custom deserializer instead?

bowenwang1996 (changes requested):

core/primitives/src/transaction.rs
matklad: We do need an enum, exactly to make this migration the only required. In the future, we'll be able to add new bits of data to Metadata in a compatible, migration less way.

Maybe I am missing something, but I don't see how a custom serializer can help us here. `ExecutionOutcome` is a borsh-serialized struct, we can't look at the bytes of serialized execution outcome and figure out if it has an extra field in the end or not?

bowenwang1996: I was thinking that maybe we could try to deserialize it into the new one first and if it does not work we deserialize it into the old one, but that might not actually work now that I think about it again :(

core/store/src/migrations.rs
```diff
@@ -790,3 +794,61 @@ pub fn migrate_18_to_new_validator_stake(store: &Store) {
...
+    for (key, value) in store.iter(DBCol::ColTransactionResult) {
+        let old_outcomes = Vec::<OldExecutionOutcomeWithIdAndProof>::try_from_slice(&value)
+            .expect("BorshDeserialize should not fail");
+        let outcomes: Vec<ExecutionOutcomeWithIdAndProof> =
+            old_outcomes.into_iter().map(|outcome| outcome.into()).collect();
+        let mut store_update = store.store_update();
+        store_update
+            .set_ser(DBCol::ColTransactionResult, key.as_ref(), &outcomes)
+            .expect("BorshSerialize should not fail");
+        store_update.commit().expect("Failed to migrate");
```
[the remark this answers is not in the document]
ailisp: make it batched, ptal

bowenwang1996 (approved):

core/store/src/migrations.rs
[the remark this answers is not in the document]
frol: @bowenwang1996 thanks for the heads up!

@khorolets Let's be prepared for that our next migration might take a while and also don't forget to notify our indexer users about this is coming (we will need to run it on out end to provide some figures about the downtime)

core/primitives/src/views.rs
```diff
@@ -1017,6 +1017,8 @@ pub struct ExecutionOutcomeView {
     pub executor_id: AccountId,
     /// Execution status. Contains the result in case of successful execution.
     pub status: ExecutionStatusView,
+    /// Execution metadata, versioned
+    pub metadata: ExecutionMetadata,
```
frol, line 1021: Please, skip serde serialization and deserialization for now, so we don't leak this field to the JSON RPC users just yet.

frol (approved):
Overall looks good to me! @ailisp Thanks for making these gradual steps which are much easier to review and iterate

core/primitives/src/views.rs
ailisp: OK! Note that we'll need this eventually because the final purpose is to expose profile data via RPC.

matklad: I think we'd actually want to use a different struct for JSON RPC. versioned enum is a borsh-friendly format. For JSON we probably want just a struct here. It's funny how borsh and json are duals in a sense that the most natural forward compat field in JSON is a struct to which you can add fields, while in borsh it's an enum to which you can add variants.

matklad:
Question: does anyone know why `ExecutionOutcome: Serialize`? It seems like removing this bound works: https://github.com/near/nearcore/commit/21a32edde770aa3b8e4a14d78ab2a0bb707442be.

Am I missing some external usage, or should I just send a PR? Structs which are both BorshSerialize and Serialize feel like an anti-pattern to me -- the best serialization formats often are different in borsh and json.

core/primitives/src/views.rs
frol: @matklad That is totally on my radar of refactoring JSON RPC! :+1:

frol:
@matklad I believe that is left-over from some early days or just a copy-paste from somewhere else. We should be more strict about the derives, especially serialization derives since it makes it hard to review sometimes.

PR #4377, opened 2021-06-18
Support disabling RPC endpoint at runtime as well as compile time

mina86:
1. Add `--disable-rpc` command line switch to `neard run` subcommand.
2. Add `json_rpc` build feature which can be disabled to completely remove RPC endpoint support from the binary.

### DIFF
--- a/core/primitives/src/views.rs
+++ b/core/primitives/src/views.rs
@@ -317,8 +317,9 @@ pub struct StatusResponse {
     pub protocol_version: u32,
     /// Latest protocol version that this client supports.
     pub latest_protocol_version: u32,
-    /// Address for RPC server.
-    pub rpc_addr: String,
+    /// Address for RPC server.  None if node doesn’t have RPC endpoint enabled.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub rpc_addr: Option<String>,
     /// Current epoch validators.
     pub validators: Vec<ValidatorInfo>,
     /// Sync status of the node.
--- a/nearcore/src/config.rs
+++ b/nearcore/src/config.rs
@@ -14,6 +14,7 @@ use tracing::info;
 
 use near_chain_configs::{ClientConfig, Genesis, GenesisConfig, LogSummaryStyle};
 use near_crypto::{InMemorySigner, KeyFile, KeyType, PublicKey, Signer};
+#[cfg(feature = "json_rpc")]
 use near_jsonrpc::RpcConfig;
 use near_network::test_utils::open_port;
 use near_network::types::ROUTED_MESSAGE_TTL;
@@ -408,6 +409,7 @@ pub struct Config {
     pub genesis_records_file: Option<String>,
     pub validator_key_file: String,
     pub node_key_file: String,
+    #[cfg(feature = "json_rpc")]
     #[serde(skip_serializing_if = "Option::is_none")]
     pub rpc: Option<RpcConfig>,
     #[cfg(feature = "rosetta_rpc")]
@@ -438,6 +440,7 @@ impl Default for Config {
             genesis_records_file: None,
             validator_key_file: VALIDATOR_KEY_FILE.to_string(),
             node_key_file: NODE_KEY_FILE.to_string(),
+            #[cfg(feature = "json_rpc")]
             rpc: Some(RpcConfig::default()),
             #[cfg(feature = "rosetta_rpc")]
             rosetta_rpc: None,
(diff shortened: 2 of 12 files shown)


nearcore/src/config.rs
```diff
@@ -948,7 +949,7 @@ pub fn create_testnet_configs_from_seeds(
         if local_ports {
             config.network.addr =
                 format!("127.0.0.1:{}", if i == 0 { first_node_port } else { open_port() });
-            config.rpc.addr = format!("127.0.0.1:{}", open_port());
+            config.rpc.get_or_insert(Default::default()).addr = format!("0.0.0.0:{}", open_port());
```
frol, line 952: I think it is better to keep 127.0.0.1 here.

```suggestion
            config.rpc.get_or_insert(Default::default()).addr = format!("127.0.0.1:{}", open_port());
```

core/primitives/src/views.rs
```diff
@@ -317,7 +317,8 @@ pub struct StatusResponse {
     pub protocol_version: u32,
     /// Latest protocol version that this client supports.
     pub latest_protocol_version: u32,
-    /// Address for RPC server.
+    /// Address for RPC server.  Empty if node does not have RPC endpoint
+    /// enabled.
```
frol, line 321: I think it is better to make it `Option<String>`

frol (approved):
@mina86 Thanks for the contribution!

frol:
@mina86 BTW, some CI tests failed (compilation errors)

core/primitives/src/views.rs
mina86: The reason I didn’t turn it into option is the ‘These types should only change when we cannot avoid this.’ suggestion near the top of the page. As far as I understand `StatusResponse` is used in inter-process messaging so changing it would be a breaking change?

nearcore/src/config.rs
mina86: D’oh. Fixed.

mina86:
> @mina86 BTW, some CI tests failed (compilation errors)

Yeah. I’m not sure why I didn’t catch it before sending the PR. Anyway, I’ve dropped the second commit for now. It still needs a bit more work.

mina86:
So there’s a failure in near-api-js in test/providers.test.js but honestly I’m not even sure how to run that test locally to investigate. ;/

frol:
I re-run the failing test. I believe that could have been just a glitch

mina86:
Thanks.

nearcore/src/config.rs
```diff
@@ -471,6 +475,24 @@ impl Config {
...
+        }
+        None
+    }
+
+    #[allow(unused_variables)]
+    pub fn set_rpc_addr(&mut self, addr: String) {
+        #[cfg(feature = "json_rpc")]
+        {
+            // TODO(mina86): Switch to get_or_insert_default() once it’s stable.
+            // See <https://github.com/rust-lang/rust/issues/82901>
```
bowenwang1996, line 492: If there is todo usually it is best to create an issue for it and put the issue number in the parentheses

bowenwang1996 (commented):
Do we want to toggle rpc both at compile time and runtime? Seems a bit redundant to me

core/primitives/src/views.rs
frol: @mina86 This particular one is not used for inter-process messaging (you can tell that by two factors: (1) there is no BorshSerialization implementation on it, (2) reviewing all the usages you can see that it is only used in JSON RPC, which serializes the responses into JSON, so in most of the cases, where RPC is enabled, there will be no structural changes, and in the case when RPC is disabled one cannot get that response in the first place :smile:

bowenwang1996 (approved):
Approve to unblock, but with sync with @frol that it is desirable to have both compile time and runtime flags.

core/primitives/src/views.rs
mina86: Done.

nearcore/src/config.rs
mina86: I hear you, though normally I find it too much hassle for this kind of minimal-impact aesthetic changes. I’ve removed the comment for the time being.

mina86:
> Approve to unblock, but with sync with @frol that it is desirable to have both compile time and runtime flags.

One thing the issue mentioned is the ‘potential [to] reduce compile time’ which is why I’ve added compile time feature. I’d be happy with just the runtime flag.

PR #4378, opened 2021-06-18
refactor: remove heavy dependency from vm-logic

matklad:
primitives are much heavier that primitives-core, so it's best to avoid
depending on them. In this case, the dependency was mostly accidental,
as we only used primitives for ProtocolVersion decisions.

The idea here is exposte the set of knobs to let the caller make
decisions about how protocol version affects VMLogic behavior. As a side
benefit, VMLogic is now oblivious of the protocol version concept at
all.

### DIFF
--- a/runtime/near-vm-logic/Cargo.toml
+++ b/runtime/near-vm-logic/Cargo.toml
@@ -21,7 +21,9 @@ serde = { version = "1", features = ["derive"] }
 sha2 = ">=0.8,<0.10"
 sha3 = ">=0.8,<0.10"
 
-near-primitives = { path = "../../core/primitives" }
+# If you want to uncomment this, think twice: `near-primitives` are a heavy dependency,
+# we are intentionally avoiding it here.
+# near-primitives = { path = "../../core/primitives" }
 near-primitives-core = { path = "../../core/primitives-core", version = "0.1.0" }
 near-vm-errors = { path = "../near-vm-errors", version = "3.0.0" }
 near-runtime-utils = { path = "../near-runtime-utils", version = "3.0.0" }
--- a/runtime/near-vm-logic/src/logic.rs
+++ b/runtime/near-vm-logic/src/logic.rs
@@ -5,17 +5,13 @@ use crate::types::{PromiseIndex, PromiseResult, ReceiptIndex, ReturnData};
 use crate::utils::split_method_names;
 use crate::ValuePtr;
 use byteorder::ByteOrder;
-use near_primitives::checked_feature;
-use near_primitives::version::is_implicit_account_creation_enabled;
 use near_primitives_core::config::ExtCosts::*;
 use near_primitives_core::config::{ActionCosts, ExtCosts, VMConfig};
 use near_primitives_core::profile::ProfileData;
 use near_primitives_core::runtime::fees::{
     transfer_exec_fee, transfer_send_fee, RuntimeFeesConfig,
 };
-use near_primitives_core::types::{
-    AccountId, Balance, EpochHeight, Gas, ProtocolVersion, StorageUsage,
-};
+use near_primitives_core::types::{AccountId, Balance, EpochHeight, Gas, StorageUsage};
 use near_runtime_utils::is_account_id_64_len_hex;
 use near_vm_errors::InconsistentStateError;
 use near_vm_errors::{HostError, VMLogicError};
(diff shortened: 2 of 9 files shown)


runtime/near-vm-logic/src/logic.rs
```diff
@@ -147,12 +152,12 @@ impl<'a> VMLogic<'a> {
...
-            current_protocol_version,
+            protocol_features,
         }
     }
 
-    /// Legacy initialization method that doesn't pass the protocol version and uses the last
-    /// protocol version before the change was introduced.
+    /// Legacy initialization method that doesn't pass the protocol protocol features,
+    /// just disabling them to stay compatible with the old behavior.
     pub fn new(
```
matklad, line 161: Can we just remove this method now? I don't understand why we need it.

EgorKulikov:
I am not sure I like where this is going. Should we create Features for each component of the system? If the problem is primitives as heavy dependency maybe better solution is to move version.rs to primitives-core or create separate crate for features?

runtime/near-vm-logic/Cargo.toml
```diff
@@ -21,7 +21,9 @@ serde = { version = "1", features = ["derive"] }
 sha2 = ">=0.8,<0.10"
 sha3 = ">=0.8,<0.10"
 
-near-primitives = { path = "../../core/primitives" }
+# If you want to uncomment this, think twice: `near-primitives` are a heavy dependency,
+# we are intentionally avoiding it here.
```
frol, line 25: Maybe we want to bring more context here like here:

https://github.com/near/nearcore/blob/7e24d5d758be41ff12528964ebdcad12e32b990d/runtime/near-runtime-utils/Cargo.toml#L16-L17

frol (commented):
I cannot make a good judgment on the design direction we go here, so I only chip in as much as I can.

matklad:
That's a great question. To my mind, per-component features is indeed a good direction to go in. Today, protocol version is a cross-cutting concern, as everything has access to it, and it increases dependencies of things on each other. Specifically, `ProtocolFeature` enum mentions concerns from different components, but each specific component which depends on `ProtocolFeature` cares only about subset of the concerns. 

It seems that only chain fundamentally cares about protocol version and needs access to `ProtocolVersion` type. It feels proper to me if chain is the single place which makes protocol-version related decisions, and then passes those decisions down to subcomponents.

At the same time, the problem is that today we have a globally accessible protocol version, and changing that pattern for only *one* component might not improve things. Although in this specific case the logic.rs change seems an incremental improvement. 

Originally, the motivation for not putting versions stuff in primitives-core was that it was a runtime dependency of near-sdk-rs, and it didn't make sense to compile nearcore internal feature-flag stuff into the contract code. As near-primitives-core [will stop being](https://github.com/near/near-sdk-rs/pull/417/files#diff-c3fb7631cde4c892fbaeff8da210a49530a6c0b60333f46dc22e302f1ec7024fR32-R34) runtime dep of the sdk soon, we can be more lenient with what we put there, so you suggestion of just moving versions there also works!

runtime/near-vm-logic/src/logic.rs
bowenwang1996: @Longarithm I believe you made the change. Is there any reason to keep this method?

bowenwang1996 (approved):

ailisp (approved):

runtime/near-vm-logic/src/logic.rs
Longarithm: @bowenwang1996 I don't see a reason to keep it

runtime/near-vm-logic/src/logic.rs
```diff
@@ -68,8 +62,19 @@ pub struct VMLogic<'a> {
...
+    /// Protocol features that are enabled for the function call.
+    protocol_features: VMLogicProtocolFeatures,
+}
+
+/// Some parts of VMLogic depend on the current protocol version. Rather than
+/// directly passing the protocol version in, we use [`VMLogicProtocolFeatures`],
+/// and let the caller decide which protocol features to activate. That way,
+/// [`VMLogic`] doesn't need to know anything about protocol versions, which
+/// helps with cutting crate dependencies.
+#[derive(Clone, Copy)]
```
olonho, line 74: Maybe `Debug` as well?

olonho (approved):

EgorKulikov (approved):
I am still not sure I like having separate features, otherwise lgtm

Longarithm (approved):

bowenwang1996:
@matklad could we merge this PR?

matklad:
Let's close this for now.

According to https://github.com/near/nearcore/issues/4379, we are still in the state of confusion regarding the overall requirements here. However, it seems that we've got rid of "compiles to wasm" requirement, so "logic should not depend on primitives" invariant is no longer relevant. 

So it seems best to avoid needless changes for now. 

That being said, I still think that the overall approach suggested here (caller picks the features, rather than callee looks at the current protocol version) is, in principle, better. But, even if it is better, it doesn't make much sense to apply it to one isolated crate.

PR #4380, opened 2021-06-20
Math API `ecrecover` and `ripemd160` extension

joshuajbouw:
## Summary

As a successor to https://github.com/near/nearcore/pull/3954, this PR is meant to speed up and lower the cost of the [Aurora EVM contract] cost by adding several new methods to the `nearcore`  [Math API] logic. Notably, `ecrecover` is extremely important as it is the most expensive method.

- #3922 `ripemd160()` (RIPEMD-160 hash function)
- #3921 `ecrecover()` (ECRecover ECDSA signer recovery)

## Todo
- [x] Removed blake2
- [x] Rework ecrecover to use near-crypto and not a new lib
- [x] Remove protocol_feature_evm conditional compilation which were added
- [x] Add `protocol_feature_math_extension` feature
- [x] Params estimator

References:

- https://ethereum.stackexchange.com/questions/15479/list-of-pre-compiled-contracts/15483#15483
- https://docs.soliditylang.org/en/latest/units-and-global-variables.html#mathematical-and-cryptographic-functions
- https://en.wikipedia.org/wiki/RIPEMD
- https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/cryptography/ECDSA.sol

[Aurora EVM contract]: https://github.com/aurora-is-near/aurora-engine
[Math API]: https://nomicon.io/RuntimeSpec/Components/BindingsSpec/MathAPI.html


core/crypto/Cargo.toml
```diff
@@ -6,7 +6,7 @@ edition = "2018"
 
 [dependencies]
 arrayref = "0.3"
-blake2 = "0.9.1"
+blake2 = { package = "near-blake2", git = "https://github.com/near/near-blake2.git", rev = "29a15a" }
```
bowenwang1996, line 9: why do we still have blake2 here

runtime/near-vm-logic/Cargo.toml
```diff
@@ -14,9 +14,13 @@ This crate implements the specification of the interface that Near blockchain ex
 
 [dependencies]
 base64 = "0.13"
+blake2 = { package = "near-blake2", git = "https://github.com/near/near-blake2.git", rev = "29a15a", optional = true }
```
bowenwang1996, line 17: also here

joshuajbouw: Fixed: https://github.com/near/nearcore/pull/4380/commits/2d1ae301c922abe3d851722639af915e3c4255f2

core/crypto/Cargo.toml
joshuajbouw: Fixed in: https://github.com/near/nearcore/pull/4380/commits/951e7670b98fa639e18f29c9f25a6aae02dd665c

Yeah, this one I certainly missed. Forgot to revert it. Good eyes.

core/crypto/src/signature.rs
```diff
@@ -556,6 +558,81 @@ impl<'de> serde::Deserialize<'de> for SecretKey {
...
+        if s > secp256k1_half_n {
+            return false;
+        }
+
+        r < secp256k1_n && s < secp256k1_n && (v == 0 || v == 1)
+    }
+
+    pub fn recover(
+        &self,
+        msg: [u8; 32],
```
bowenwang1996, line 607: why is msg not passed as a reference?

core/primitives-core/src/config.rs
```diff
@@ -353,6 +361,9 @@ impl Default for ExtCostsConfig {
             keccak256_byte: SAFETY_MULTIPLIER * 7157035,
             keccak512_base: SAFETY_MULTIPLIER * 1937129412,
             keccak512_byte: SAFETY_MULTIPLIER * 12216567,
+            ripemd160_base: SAFETY_MULTIPLIER * 15136567500, // 10 x sha256_base
+            ripemd160_block: SAFETY_MULTIPLIER * 80391170 * 8, // 10 x sha256_byte x 8 bytes
+            ecrecover_base: SAFETY_MULTIPLIER * 75682837500, // 50 x sha256_base
```
bowenwang1996, line 366: These should be behind a feature flag? Also how are they determined?

core/crypto/src/signature.rs
ailisp: Pass as a reference - you'll need `Message::from_slice`, The `Message::from<ThirtyTwoByteHash>` is faster. Check their source:
https://www.wpsoftware.net/rustdoc/src/secp256k1/lib.rs.html#470-479
https://www.wpsoftware.net/rustdoc/src/secp256k1/lib.rs.html#502-507

core/primitives-core/src/config.rs
ailisp: Same question here :)

runtime/runtime-params-estimator/src/lib.rs
```diff
@@ -1,20 +1,21 @@
...
-// Encapsulates the runtime so that it can be run separately from the rest of the node.
+/// Encapsulates the runtime so that it can be run separately from the rest of the node.
 pub mod testbed;
-// Prepares transactions and feeds them to the testbed in batches. Performs the warm up, takes care
-// of nonces.
+/// Prepares transactions and feeds them to the testbed in batches. Performs the warm up, takes care
+/// of nonces.
 pub mod testbed_runners;
+/// Runs a VM (Default: Wasmer) on the given contract and measures the time it takes to do a single operation.
+pub mod vm_estimator;
```
ailisp, line 18: should these be `///`, Looks like they are comments instead of docstrings.

ailisp (approved):

runtime/runtime-params-estimator/src/lib.rs
joshuajbouw: Was something that `cargo fmt` picked up that isn't related to this work.

core/primitives-core/src/config.rs
joshuajbouw: I didn't know that they needed to be behind the feature flag. Which do you recommend? The idea is to remove `protocol_feature_evm` entirely. Where it is in the code now is where the old EVM logic is when it was part of core and not as a contract. A later PR will be created to remove it in full. 

The costs are WIP as we had issues with the estimator before. I think, they should be ok now that the feature flags were removed, possibly won't have that issue again. I can attempt to run it to get accurate results. But, the cost is currently based off of ETH's costs using their sha256 cost as a base. They should be overestimates.

core/crypto/src/signature.rs
joshuajbouw: Yeah, I agree with this. Should in fact actually be passed in as `impl AsRef<[u8}>`. This commit does that: https://github.com/near/nearcore/pull/4380/commits/434145b5135c644c4e0d239a41618c78b59dfbb0.

core/crypto/src/signature.rs
```diff
@@ -556,6 +558,83 @@ impl<'de> serde::Deserialize<'de> for SecretKey {
...
+        if s > secp256k1_half_n {
+            return false;
+        }
+
+        r < secp256k1_n && s < secp256k1_n && (v == 0 || v == 1)
+    }
+
+    pub fn recover(
+        &self,
+        msg: impl AsRef<[u8]>,
```
matklad, line 607: Nit:

```suggestion
        msg: &[u8],
```

Generic functions blow up compile time and binary size. For internal APIs, `AsRef` polymorphism is a net-negative. For public APIs, it sometimes makes sense to wire an `pub fn f(param: impl AsRef<T>)` which is a one-liner delegating to `fn f_(param: &T)`.

runtime/near-vm-logic/src/logic.rs
```diff
@@ -943,14 +944,86 @@ impl<'a> VMLogic<'a> {
...
+    /// Hashes the given value using RIPEMD-160 and returns it into `register_id`.
+    ///
+    /// # Errors
+    ///
+    /// If `value_len + value_ptr` points outside the memory or the registers use more memory than
+    /// the limit with `MemoryAccessViolation`.
+    ///
+    /// # Cost
+    ///
+    ///  Where `message_blocks` is `max(message_len / 8, 1)`.
```
matklad, line 964: Misplaced comment? The `message_blocks` calculation looks completely different.

core/crypto/src/signature.rs
matklad: Actually, why do we need `&[u8]` at all here? Seems like this can take `msg: [u8; 32]`, which will allow changing `Message::from_slice` to `Message::from`, removing one error condition altogether?

runtime/near-vm-logic/src/logic.rs
```diff
@@ -943,14 +944,86 @@ impl<'a> VMLogic<'a> {
...
+        }
+
+        let mut hash_bytes = [0u8; 32];
+        self.memory_get_into(hash_ptr, &mut hash_bytes)?;
+        if let Ok(pk) = signature.recover(&hash_bytes) {
+            self.internal_write_register(register_id, pk.as_ref().to_vec())?;
+            return Ok(true as u64);
+        };
+
+        return Ok(false as u64);
```
matklad, line 1024: Nit:

```suggestion
        false as u64
```

runtime/runtime-params-estimator/test-contract/src/lib.rs
```diff
@@ -47,6 +47,8 @@ extern "C" {
     fn sha256(value_len: u64, value_ptr: u64, register_id: u64);
     fn keccak256(value_len: u64, value_ptr: u64, register_id: u64);
     fn keccak512(value_len: u64, value_ptr: u64, register_id: u64);
+    fn ripemd160(value_len: u64, value_ptr: u64, register_id: u64);
+    fn ecrecover(hash_ptr: u64, v: u32, r_ptr: u64, s_ptr: u64, register_id: u64);
```
matklad, line 51: This signature does not match the one in logic.rs

runtime/near-vm-logic/src/logic.rs
```diff
@@ -943,14 +944,86 @@ impl<'a> VMLogic<'a> {
...
+        let signature = Secp256K1Signature::from(signature_bytes);
+
+        if malleability_flag == 1 {
+            if !signature.check_signature_values() {
+                return Ok(false as u64);
+            }
+        }
+
+        let mut hash_bytes = [0u8; 32];
+        self.memory_get_into(hash_ptr, &mut hash_bytes)?;
```
matklad, line 1018: Let's move this reading form memory *before* the if where we check `mallability_flag`. As written now, the code has weird behavior that, when it passes an out of bounds `hash_ptr`, *but* also passes invalid signature, it actually works. It seems better if invalid memory accesses are always caught.

core/crypto/src/signature.rs
```diff
@@ -556,6 +558,83 @@ impl<'de> serde::Deserialize<'de> for SecretKey {
...
+            0xd0, 0x36, 0x41, 0x41,
+        ]);
+
+        let secp256k1_half_n = secp256k1_n / U256::from(2);
+        // Reject upper range of s values (ECDSA malleability)
+        if s > secp256k1_half_n {
+            return false;
+        }
+
+        r < secp256k1_n && s < secp256k1_n && (v == 0 || v == 1)
```
matklad, line 602:
```suggestion
        r < secp256k1_n && s < secp256k1_n
```

The last clause is dead code which is always true -- this check is already done by `check_v`. I guess, `assert!(v == 0 || v == 1)` could be added for readability?

core/crypto/src/signature.rs
joshuajbouw: Ok, then I'll just revert it to that.

runtime/near-vm-logic/src/logic.rs
joshuajbouw: Ah yeah, this was changed. Good eye.

core/crypto/src/signature.rs
joshuajbouw: Fixed: https://github.com/near/nearcore/pull/4380/commits/9c281503ececd4295b166f6b4cbccab1a9b5ab28

I had that before, but then was asked why I didn't add a reference.

runtime/near-vm-logic/src/logic.rs
joshuajbouw: I'm pretty sure that these have to be results?

runtime/runtime-params-estimator/test-contract/src/lib.rs
joshuajbouw: Fixed: https://github.com/near/nearcore/pull/4380/commits/8c9341fe89506c030d56a02dafe8fae62e6eecbd

runtime/near-vm-logic/src/logic.rs
joshuajbouw: Yep, was late when I added that in. Good catch. Fixed: https://github.com/near/nearcore/pull/4380/commits/0acaa6700500576bd116b6fb8bfa6c0b47833838

core/crypto/src/signature.rs
joshuajbouw: Good eye, I just took that from original reference material. You are right, should be fine.

Fixed https://github.com/near/nearcore/pull/4380/commits/0acaa6700500576bd116b6fb8bfa6c0b47833838

core/crypto/src/signature.rs
```diff
@@ -556,6 +558,80 @@ impl<'de> serde::Deserialize<'de> for SecretKey {
...
+        let r = U256::from(r_bytes);
+
+        let mut s_bytes = [0u8; 32];
+        s_bytes.copy_from_slice(&self.0[32..64]);
+        let s = U256::from(s_bytes);
+
+        let _v = match self.check_v() {
+            Ok(v) => v,
+            Err(_e) => return false,
+        };
```
EgorKulikov, line 583:
```suggestion
        if let Err(_) = match self.check_v() {
            return false;
        };
```

core/crypto/src/signature.rs
```diff
@@ -556,6 +558,80 @@ impl<'de> serde::Deserialize<'de> for SecretKey {
...
+            0xd0, 0x36, 0x41, 0x41,
+        ]);
+
+        let secp256k1_half_n = secp256k1_n / U256::from(2);
+        // Reject upper range of s values (ECDSA malleability)
+        if s > secp256k1_half_n {
+            return false;
+        }
+
+        r < secp256k1_n && s < secp256k1_n
```
EgorKulikov, line 601: How s can be not less then secp256k1_n if it is not more than half of it?

runtime/near-vm-logic/src/logic.rs
EgorKulikov:
```suggestion
        Ok(false as u64)
```

core/crypto/src/signature.rs
joshuajbouw: That's from the ETH malleability checks, I have been adding tests for it. I'll double check it.

core/crypto/src/signature.rs
```diff
@@ -556,6 +558,80 @@ impl<'de> serde::Deserialize<'de> for SecretKey {
...
 
+impl Secp256K1Signature {
+    fn check_v(&self) -> Result<u8, crate::errors::ParseSignatureError> {
+        match self.0[64] {
+            27 | 28 => Ok(self.0[64] - 27),
+            _ => Err(crate::errors::ParseSignatureError::InvalidData {
+                error_message: "invalid `v` value".to_string(),
+            }),
+        }
+    }
```
abacabadabacaba, line 569: This interpretation of `v` is different from what the code below uses (specifically, `Signature::verify`). It is a very bad design if values the same type are interpreted differently by different functions.

core/crypto/src/signature.rs
```diff
@@ -556,6 +558,80 @@ impl<'de> serde::Deserialize<'de> for SecretKey {
...
+            0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff,
+            0xff, 0xfe, 0xba, 0xae, 0xdc, 0xe6, 0xaf, 0x48, 0xa0, 0x3b, 0xbf, 0xd2, 0x5e, 0x8c,
+            0xd0, 0x36, 0x41, 0x41,
+        ]);
+
+        let secp256k1_half_n = secp256k1_n / U256::from(2);
+        // Reject upper range of s values (ECDSA malleability)
+        if s > secp256k1_half_n {
+            return false;
+        }
```
abacabadabacaba, line 599: Can we make this check optional? The standard doesn't require it.

core/crypto/src/signature.rs
```diff
@@ -556,6 +558,80 @@ impl<'de> serde::Deserialize<'de> for SecretKey {
...
+        let msg = Message::from(msg);
+
+        let res = SECP256K1
+            .recover(&msg, &recoverable_sig)
+            .map_err(|err| crate::errors::ParseSignatureError::InvalidData {
+                error_message: err.to_string(),
+            })?
+            .serialize_vec(&SECP256K1, false);
+
+        let pk = Secp256K1PublicKey::try_from(&res[0..64]).map_err(|err| {
```
abacabadabacaba, line 627: You need to skip the first byte.

core/primitives-core/src/config.rs
```diff
@@ -223,6 +223,14 @@ pub struct ExtCostsConfig {
     /// Cost of getting sha256 per byte
     pub keccak512_byte: Gas,
 
+    /// Cost of getting ripemd160 base
+    pub ripemd160_base: Gas,
+    /// Cost of getting ripemd160 per message block
+    pub ripemd160_block: Gas,
+
+    /// Cost of calling ecrecover
+    pub ecrecover_base: Gas,
```
abacabadabacaba, line 232: Shouldn't these be behind a feature flag?

runtime/near-vm-logic/Cargo.toml
```diff
@@ -17,10 +17,13 @@ base64 = "0.13"
 borsh = "0.8.1"
 bs58 = "0.4"
 byteorder = "1.2"
+num-integer = "0.1"
```
abacabadabacaba, line 20: IMHO importing a library just to get ceiling division is an overkill.

runtime/near-vm-logic/src/logic.rs
```diff
@@ -943,14 +944,87 @@ impl<'a> VMLogic<'a> {
...
+    ///  Where `message_blocks` is `(value_len + 9).div_ceil(64)`.
+    ///
+    /// `base + write_register_base + write_register_byte * num_bytes + ripemd160_base + ripemd160_block * message_blocks`
+    pub fn ripemd160(&mut self, value_len: u64, value_ptr: u64, register_id: u64) -> Result<()> {
+        use num_integer::Integer;
+
+        self.gas_counter.pay_base(ripemd160_base)?;
+        let value = self.get_vec_from_memory_or_register(value_ptr, value_len)?;
+
+        let message_blocks = (value_len + 9).div_ceil(&64);
```
abacabadabacaba, line 973: Please implement ceiling division manually (be careful to avoid overflows). Also, `value_len` may not be the length of the value, as it may come from a register.

runtime/near-vm-logic/src/logic.rs
```diff
@@ -943,14 +944,87 @@ impl<'a> VMLogic<'a> {
...
+    /// Returns a bool indicating success or failure as a `u64`.
+    ///
+    /// # Errors
+    ///
+    /// * If `hash_ptr`, `r_ptr`, or `s_ptr` point outside the memory or the registers use more
+    ///   memory than the limit, then returns `MemoryAccessViolation`.
+    ///
+    /// # Cost
+    ///
+    /// `base + write_register_base + write_register_byte * 20 + ecrecover_base`
```
abacabadabacaba, line 997: This is not correct (the length is not 20, and read operations also charge gas).

runtime/near-vm-logic/src/logic.rs
```diff
@@ -943,14 +944,87 @@ impl<'a> VMLogic<'a> {
...
+            }
+        }
+
+        if let Ok(pk) = signature.recover(hash_bytes) {
+            self.internal_write_register(register_id, pk.as_ref().to_vec())?;
+            return Ok(true as u64);
+        };
+
+        return Ok(false as u64);
+    }
```
abacabadabacaba, line 1026: There are multiple issues:
1. Unless there is a good reason, please use register API (`get_vec_from_memory_or_register`) to get the inputs.
2. Because different protocols encode the `v` value differently, pass it as a separate argument rather than as a part of the signature. The signature should be 64 bytes.
3. The value of `v` should be in the range [0, 3]. This is what most secp256k1 implementations expect.
4. Now, it can be combined with `malleability_flag` into a single argument, `flags`. Bits 0 and 1 would be `v`, and bit 2 would be the malleability flag.
5. Also, as there is no `u8` type in WASM, this argument's type should be `u32`.
6. The malleability flag should only control the malleability check. The `check_signature_values` performs other checks as well. However, they seems to be redundant, because `signature.recover` performs them as well.

abacabadabacaba (changes requested):

core/crypto/src/signature.rs
ailisp: Hmm I still think calling Message::from<[u8;32]> is faster than pass in as reference, as it doesn't deal with a Result and introduce a AsRef trait is slower to compile. I guess difference is trivial, though

core/primitives-core/src/config.rs
nearmax: That's correct, they should be behind feature flag.

core/primitives-core/src/config.rs
nearmax: AFAIK this was resolved in Slack. EVM precompiles need to have their own feature flag.

nearmax: Unfortunately, we need to be precise with the fees, because if we are not and we underestimate them someone will be able to grind our blockchain and slow block production to the point where it causes issues. Eyeballing costs through what they are on Ethereum does not provide guarantees and so we cannot do it. I am trying to run param estimator right now to see if we can speed up review of this PR.

runtime/near-vm-logic/Cargo.toml
```diff
@@ -17,10 +17,13 @@ base64 = "0.13"
...
+ripemd160 = "0.9.0"
 serde = { version = "1", features = ["derive"] }
 sha2 = ">=0.8,<0.10"
 sha3 = ">=0.8,<0.10"
 
+near-crypto = { path = "../../core/crypto" }
```
nearmax, line 26: FYI, this makes near-vm-logic not compilable to Wasm, which is needed for AS SDK and near-sdk-sim. It would require some ugly refactoring to avoid doing it. Given that we are replacing simulator and mock-vm in our contract testing infrastructure with the sandbox it is not problematic, but Developer Platform team needs to be aware of it. CC @mikedotexe @austinabell @willemneal @DiscRiskandBisque @ChaoticTempest 

This PR will be a forcing function to move all testing infrastructure to the sandbox.

runtime/near-vm-logic/src/logic.rs
```diff
@@ -943,14 +944,87 @@ impl<'a> VMLogic<'a> {
...
+    ///   memory than the limit, then returns `MemoryAccessViolation`.
+    ///
+    /// # Cost
+    ///
+    /// `base + write_register_base + write_register_byte * 20 + ecrecover_base`
+    pub fn ecrecover(
+        &mut self,
+        hash_ptr: u64,
+        sig_ptr: u64,
+        malleability_flag: u8,
```
nearmax, line 1002: Wasm does not support u8 for Wasm interface, please explicitly use u64 for arguments and coerce to u8 inside the method.

nearmax (changes requested):

runtime/near-vm-logic/src/logic.rs
abacabadabacaba: Wouldn't `u32` work as well? Also, the value should be checked and rejected if the high bits are not zero.

core/crypto/src/signature.rs
austinabell:
```suggestion
        if match self.check_v().is_err() {
            return false;
        };
```

Same as above, but no need for the let binding

runtime/near-vm-logic/Cargo.toml
austinabell: but if this functionality is put under a feature flag as mentioned elsewhere in this PR, this wouldn't be required for sim/unit testing in the SDK, would it? (just going off intuition based on scanning the PR, can dig deeper to understand later)

bowenwang1996: It will be stabilized soon

core/primitives-core/src/config.rs
birchmd: Feature flag introduced in https://github.com/near/nearcore/pull/4380/commits/3f08a35fa65f1a24be2fb46f6d72ca1b5c587a3b

core/primitives-core/src/config.rs
birchmd: Feature flag added in https://github.com/near/nearcore/pull/4380/commits/3f08a35fa65f1a24be2fb46f6d72ca1b5c587a3b

Can't speak to the values themselves, so I'll leave this conversation open.

runtime/near-vm-logic/src/logic.rs
birchmd: Fixed in https://github.com/near/nearcore/pull/4380/commits/4ceab9927161613027b27789100e2a5d29a87112

runtime/near-vm-logic/Cargo.toml
birchmd: Fixed in https://github.com/near/nearcore/pull/4380/commits/4ceab9927161613027b27789100e2a5d29a87112

abacabadabacaba:
@birchmd You checked division code is still too complex. Something like `let message_blocks = (value.len() as u64).checked_add(8).ok_or(VMLogicError::HostError(HostError::IntegerOverflow))? + 1;` should work (and not need the `div_ceil` function).

runtime/near-vm-logic/src/logic.rs
```diff
@@ -943,14 +945,104 @@ impl<'a> VMLogic<'a> {
...
+    ///
+    ///  Where `message_blocks` is `(value_len + 9).div_ceil(64)`.
+    ///
+    /// `base + write_register_base + write_register_byte * num_bytes + ripemd160_base + ripemd160_block * message_blocks`
+    #[cfg(feature = "protocol_feature_math_extension")]
+    pub fn ripemd160(&mut self, value_len: u64, value_ptr: u64, register_id: u64) -> Result<()> {
+        self.gas_counter.pay_base(ripemd160_base)?;
+        let value = self.get_vec_from_memory_or_register(value_ptr, value_len)?;
+
+        let message_blocks = value
```
birchmd, line 973: > @birchmd You ceiling division code is still too complex. Something like `let message_blocks = (value.len() as u64).checked_add(8).ok_or(VMLogicError::HostError(HostError::IntegerOverflow))? + 1;` should work (and not need the `div_ceil` function).

@abacabadabacaba I don't understand your simplification, where did the division by 64 go? If the division by 64 is incorrect logic (I don't know if it is or not I was only looking to address superficial comments) then of course we don't need any division function, but that's not what the original comment was about.

abacabadabacaba: Sorry, fixed the snippet.

birchmd: Is it ok to always have the `+ 1` even in the case that `value.len() + 9` is evenly divisible by 64? Technically that is not ceiling division because if the result is evenly divisible then we shouldn't need the `+ 1`, but if you think that's ok then I'm happy to accept the simplification.

abacabadabacaba: That's why I add 8 instead of 9.
(len+9).ceiling_div(64) = (len+9+63) / 64 = (len+8+64) / 64 = (len+8) / 64 + 1

birchmd: Ah, makes sense. Thanks!

Done in https://github.com/near/nearcore/pull/4380/commits/aa1b8c5321594be675a28eedab1679759523689b

runtime/near-vm-logic/Cargo.toml
willemneal: This isn't an issue for `sdk-sim` as the host side is still compiled to native.  However, a feature flag for non-wasm features would suffice.

@abacabadabacaba Is there any way to make the crypto code compile to Wasm just for testing purposes?

runtime/near-vm-logic/src/logic.rs
birchmd: (and `div_ceil` function removed in https://github.com/near/nearcore/pull/4380/commits/1ec6416efb9f6b62299aa170524561b4d35cbf4d )

runtime/near-vm-logic/Cargo.toml
abacabadabacaba: @willemneal Why doesn't it compile now? Ideally it should be entirely portable.

core/crypto/src/signature.rs
joshuajbouw: Ah, not sure why that hadn't clicked in my mind earlier that the 26 / 27 value is an ETH thing.

core/crypto/src/signature.rs
joshuajbouw: So, `1..65` if I'm not mistakened. The first byte being a BTC legacy thing.

core/primitives-core/src/config.rs
joshuajbouw: @nearmax Let me know how it goes either here or in Slack.

core/crypto/src/signature.rs
joshuajbouw: No longer required as this functionality was removed. https://github.com/near/nearcore/pull/4380/commits/1f2a3141f83913957822ab45e4f421405519871f

core/crypto/src/signature.rs
joshuajbouw: This should be ok in this https://github.com/near/nearcore/pull/4380/commits/1f2a3141f83913957822ab45e4f421405519871f

core/crypto/src/signature.rs
joshuajbouw: Made optional here https://github.com/near/nearcore/pull/4380/commits/1f2a3141f83913957822ab45e4f421405519871f

runtime/near-vm-logic/src/logic.rs
joshuajbouw: We have it as `u32` here, I assume that should work too? https://github.com/near/nearcore/pull/4380/commits/f6d566aa60ac8d358dded6410584a79748360efe

Else, we can change it to `u64`.

runtime/near-vm-logic/Cargo.toml
joshuajbouw: Hm. Shoot. Would I need to move all this logic directly then instead of from `near-crypto`?

runtime/near-vm-logic/src/logic.rs
joshuajbouw: 1. I had done so to prevent another vec initialisation, as recommended by @matklad. I'll change this back though.
2. Good call
3. Right I'll add this in.
4. I'm unsure if that is a good idea as it isn't exactly explicit enough without reading the documentation that the `v` is in there and I don't exactly how they are directly related in that sense. Let me know if you still want to do that.
5. That has been changed.
6. I'll take a look at that and remove the ones that are already done.

matklad: >Unless there is a good reason, please use register API (get_vec_from_memory_or_register) to get the inputs.

I believe there's good reason to not do this here, but this isn't a blocker. 

In this case, we have inputs of _fixed length_: the caller doesn't provide the len, its hard-coded to 64 at the call-site. So, `get_vec_from_memory_or_register` is actually guaranteed to fetch the vec from memory, not from register (that function uses `len` to decide between memory and register). By reading directly into fixed-size buffer we gain two benefits: better correctness (we track the size at compile time) and better performance (no need to allocate a vec). 

However, as it's impossible to use registers with this API, the choice here doesn't affect the ABI and so can be revised later. 

To be clear, using `get_vec_from_memory_or_register` is not wrong, its just a bit suboptimal and confusing for fixed-sized inputs.

runtime/near-vm-logic/src/logic.rs
```diff
@@ -943,14 +945,117 @@ impl<'a> VMLogic<'a> {
...
+        self.gas_counter.pay_base(ecrecover_base)?;
+
+        let signature = {
+            let vec = self.get_vec_from_memory_or_register(sig_ptr, 64)?;
+            let mut bytes = [0u8; 65];
+            bytes[0..64].copy_from_slice(&vec);
+
+            match v {
+                0 | 1 | 2 | 3 => {
+                    bytes[65] = v as u8;
```
matklad, line 1027:
```suggestion
                    bytes[64] = v as u8;
```

65 is out of bounds I think?

runtime/near-vm-logic/Cargo.toml
willemneal: I haven't tried, but given this comment, I thought that was Max's implication.  In the past some crypto code relies on inline assembly.

runtime/near-vm-logic/src/logic.rs
joshuajbouw: 65 bytes, so 0 -> 64, then yes. Just got back from picking up my son from his cousins, getting this in now.

nearmax (changes requested):
Fees need to be estimated with param estimator. If they are underestimated our blockchain can be abusable and block production can be slowed down.

runtime/near-vm-logic/src/logic.rs
joshuajbouw: I do agree with this @matklad. If @abacabadabacaba is good with that, then I'll add that back in.

runtime/near-vm-logic/src/logic.rs
joshuajbouw: Fixed here https://github.com/near/nearcore/pull/4380/commits/88fd6aae84bdd019ff183980fc8e6c0eaa111f53

runtime/near-vm-logic/Cargo.toml
joshuajbouw: He had said that it should be ok in the end on Slack, unless someone objects.

nearmax: @austinabell When features stabilize they become enabled by default and then they also need to be compiled to Wasm. However, as I explained above we should stop compiling nearcore code to Wasm.

core/crypto/src/signature.rs
```diff
@@ -556,6 +558,62 @@ impl<'de> serde::Deserialize<'de> for SecretKey {
...
+            0xff, 0xfe, 0xba, 0xae, 0xdc, 0xe6, 0xaf, 0x48, 0xa0, 0x3b, 0xbf, 0xd2, 0x5e, 0x8c,
+            0xd0, 0x36, 0x41, 0x41,
+        ]);
+
+        if reject_upper {
+            let secp256k1_half_n = secp256k1_n / U256::from(2);
+            // Reject upper range of s values (ECDSA malleability)
+            if s > secp256k1_half_n {
+                return false;
+            }
```
abacabadabacaba, line 582: We can always check `s` against just one upper bound: `s < if reject_upper { secp256k1_half_n+1 } else { secp256k1_n }` (where `secp256k1_half_n+1` should be a static).

core/crypto/src/signature.rs
```diff
@@ -556,6 +558,62 @@ impl<'de> serde::Deserialize<'de> for SecretKey {
...
+        let res = SECP256K1
+            .recover(&msg, &recoverable_sig)
+            .map_err(|err| crate::errors::ParseSignatureError::InvalidData {
+                error_message: err.to_string(),
+            })?
+            .serialize_vec(&SECP256K1, false);
+
+        let pk = Secp256K1PublicKey::try_from(&res[1..65]).map_err(|err| {
+            crate::errors::ParseSignatureError::InvalidData { error_message: err.to_string() }
+        })?;
```
abacabadabacaba, line 611: This can never fail (any slice of length 64 will be accepted).

runtime/near-vm-logic/src/logic.rs
```diff
@@ -943,14 +945,117 @@ impl<'a> VMLogic<'a> {
...
+    /// # Cost
+    ///
+    /// `base + write_register_base + write_register_byte * 64 + ecrecover_base`
+    #[cfg(feature = "protocol_feature_math_extension")]
+    pub fn ecrecover(
+        &mut self,
+        hash_ptr: u64,
+        sig_ptr: u64,
+        v: u64,
+        malleability_flag: u64,
```
abacabadabacaba, line 1015: Please combine them into a single `flags` parameter: low 2 bits are `v`, next bit is malleability check flag, remaining bits must be zero.

runtime/near-vm-logic/src/logic.rs
```diff
@@ -943,14 +945,117 @@ impl<'a> VMLogic<'a> {
...
+    ) -> Result<u64> {
+        self.gas_counter.pay_base(ecrecover_base)?;
+
+        let signature = {
+            let vec = self.get_vec_from_memory_or_register(sig_ptr, 64)?;
+            let mut bytes = [0u8; 65];
+            bytes[0..64].copy_from_slice(&vec);
+
+            match v {
+                0 | 1 | 2 | 3 => {
```
abacabadabacaba, line 1026:
```suggestion
if v < 4 {
    [...]
} else {
    [...]
}
```

runtime/near-vm-logic/src/logic.rs
```diff
@@ -943,14 +945,117 @@ impl<'a> VMLogic<'a> {
...
+        hash_ptr: u64,
+        sig_ptr: u64,
+        v: u64,
+        malleability_flag: u64,
+        register_id: u64,
+    ) -> Result<u64> {
+        self.gas_counter.pay_base(ecrecover_base)?;
+
+        let signature = {
+            let vec = self.get_vec_from_memory_or_register(sig_ptr, 64)?;
```
abacabadabacaba, line 1021: We should allow specifying a register here (and for `hash` as well).

runtime/near-vm-logic/src/logic.rs
```diff
@@ -943,14 +945,117 @@ impl<'a> VMLogic<'a> {
...
+            let vec = self.get_vec_from_memory_or_register(hash_ptr, 32)?;
+            let mut bytes = [0u8; 32];
+            bytes.copy_from_slice(&vec);
+            bytes
+        };
+
+        if (malleability_flag & 1) == 1 {
+            if !signature.check_signature_values((malleability_flag & 2) == 2) {
+                return Ok(false as u64);
+            }
```
abacabadabacaba, line 1048: The check itself should always be performed. The only thing that should be configurable is `reject_upper`.

abacabadabacaba (changes requested):

runtime/near-vm-logic/src/logic.rs
joshuajbouw: Besides the other feedback, this is the only one that I am a bit confused about. I'm struggling to understand why you want to combine a flag and a parameter? Out of curiosity, why?

abacabadabacaba: Well, it sounds kinda wasteful to have 2 64-bit parameters for 3 bits of data. Also, if the hash and the signature are passed using register API (so each of them is represented by two parameters because you cannot omit the length because it is needed to distinguish a memory argument from a register argument), there would be 5 other parameters, and I don't like having too many parameters. This is not required, though.

joshuajbouw: I do absolutely agree that it is wasteful but not quite uncommon across the other host functions. I also do want to button down the API and not have to change it in the future. 5 is usually my limit, I hate going over 5. I do see others using `u32`, Max I think suggested `u64` instead. I know we can't do `u8`.

core/crypto/src/signature.rs
```diff
@@ -556,6 +558,62 @@ impl<'de> serde::Deserialize<'de> for SecretKey {
...
+
+        let mut s_bytes = [0u8; 32];
+        s_bytes.copy_from_slice(&self.0[32..64]);
+        let s = U256::from(s_bytes);
+
+        let secp256k1_n = U256::from([
+            0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff,
+            0xff, 0xfe, 0xba, 0xae, 0xdc, 0xe6, 0xaf, 0x48, 0xa0, 0x3b, 0xbf, 0xd2, 0x5e, 0x8c,
+            0xd0, 0x36, 0x41, 0x41,
+        ]);
```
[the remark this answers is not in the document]
joshuajbouw: Fixed in https://github.com/near/nearcore/pull/4380/commits/e5f7468abd9ea4c94da78f2101f415a2ab4054f9

core/crypto/src/signature.rs
joshuajbouw: Thats right. https://github.com/near/nearcore/pull/4380/commits/2091bfafcae395ea2395e4c0f65c6dd21a7843a3

runtime/near-vm-logic/src/logic.rs
joshuajbouw: Yeah, that nit fix should be better. https://github.com/near/nearcore/pull/4380/commits/bdc906663a678fa86fab7b83848a4bf0e6929def

runtime/near-vm-logic/src/logic.rs
joshuajbouw: What do you mean?

runtime/near-vm-logic/src/logic.rs
joshuajbouw: Got it. https://github.com/near/nearcore/pull/4380/commits/5e76c8bf512bb63242279b6df8addd04fd63be30

core/crypto/src/signature.rs
joshuajbouw: Ahh ok yeah I get it.

joshuajbouw: https://github.com/near/nearcore/pull/4380/commits/38fc3626d696ebacbb9f03c5bb35bfb9412b7feb

runtime/near-vm-logic/src/logic.rs
joshuajbouw: If I am not mistaken, this should handle that. https://github.com/near/nearcore/pull/4380/commits/ce1f63cebb6e73037752efb2ac0177d4dc1ab8cd

runtime/near-vm-logic/src/logic.rs
joshuajbouw: @bowenwang1996 Do you have thoughts on this?

core/primitives-core/src/config.rs
```diff
@@ -223,6 +223,17 @@ pub struct ExtCostsConfig {
     /// Cost of getting sha256 per byte
     pub keccak512_byte: Gas,
 
+    #[cfg(feature = "protocol_feature_math_extension")]
```
olonho, line 226: Oh, another field under feature :(. Can we avoid that?

joshuajbouw: Alt BN128 is also under a feature in the same vein?

runtime/runtime-params-estimator/test-contract/src/lib.rs
```diff
@@ -424,6 +428,56 @@ pub unsafe fn keccak512_10kib_10k() {
     }
 }
 
+// Function to measure `ripemd160_base` and `ripemd160_block`. Also measures `base`, `write_register_base`,
+// and `write_register_byte`. However `ripemd160` computation is more expensive than register writing
+// so we are okay overcharging it.
+// Compute ripemd160 on 10b 10k times.
+#[cfg(feature = "protocol_feature_math_extension")]
+#[no_mangle]
+pub unsafe fn ripemd160_10b_10k() {
```
olonho, line 437: Why do we measure those costs calling them from contract? Please just measure cost directly, as suggested in this patch.
[math.zip](https://github.com/near/nearcore/files/6717594/math.zip)

olonho (changes requested):

runtime/near-vm-logic/src/logic.rs
joshuajbouw: Bowen informed me that he is good with this.

core/primitives-core/src/config.rs
joshuajbouw: https://github.com/near/nearcore/pull/4380/commits/bffcbebcdbbb406c91115f0643f83ae4114b5bc4

runtime/runtime-params-estimator/test-contract/src/lib.rs
joshuajbouw: Because this has nothing to do with the EVM code that will be stripped in a future PR. This is adding it directly to the Math API.

joshuajbouw: I misunderstood, its too late to get that in right now. In a future PR when all the others are remeasured directly, just include this as well.

core/crypto/src/signature.rs
```diff
@@ -553,9 +555,64 @@ impl<'de> serde::Deserialize<'de> for SecretKey {
...
+        r_bytes.copy_from_slice(&self.0[0..32]);
+        let r = U256::from(r_bytes);
+
+        let mut s_bytes = [0u8; 32];
+        s_bytes.copy_from_slice(&self.0[32..64]);
+        let s = U256::from(s_bytes);
+
+        let s_check = if reject_upper {
+            // Reject upper range of s values (ECDSA malleability)
+            SECP256K1_N_HALF + U256::one()
```
abacabadabacaba, line 580: Please make this sum (`SECP256K1_N_HALF + U256::one()`) a static constant.

runtime/near-vm-logic/src/logic.rs
```diff
@@ -943,14 +945,131 @@ impl<'a> VMLogic<'a> {
...
+                    ),
+                }));
+            }
+
+            let mut bytes = [0u8; 32];
+            bytes.copy_from_slice(&vec);
+            bytes
+        };
+
+        if !signature.check_signature_values((malleability_flag & 1) == 1) {
```
abacabadabacaba, line 1061: Please check that `malleability_flag` is 0 or 1.

abacabadabacaba (changes requested):

runtime/near-vm-logic/src/logic.rs
nearmax: I suggest we keep arguments separately, it is better to have cleaner API than do micro-optimizations.

core/primitives-core/src/config.rs
nearmax: For some reason this was reverted in https://github.com/near/nearcore/commit/ac53dcbafcc5c0f405eeaf0dccfcdbf5ba19161a#diff-0337a7fe1272b09cad7bda003159822a95e0609b9716fa987b78dacfaa3cd181

nearmax (changes requested):

core/primitives-core/src/config.rs
joshuajbouw: Was reverted as I was asked to revert this change in this library. See: https://github.com/near/nearcore/pull/4380#discussion_r658889331

This has since be re-reverted. Please code owners do not suggest changes that do not actually help progress. I had assumed that change would've helped the params estimator given conversation context.

Revert: https://github.com/near/nearcore/pull/4380/commits/f25fe99e6421e6131291bd5f1bc4ea1567d13d56

joshuajbouw:
Provided CI is good to go, I just need the params estimator numbers, and good to push in.

bowenwang1996 (approved):
Just need the new fees :)

runtime/near-vm-logic/src/logic.rs
nearmax: Fixed in https://github.com/near/nearcore/pull/4380/commits/73380e1f1d565bb8c97e1b1b87b2710e71560221

nearmax (approved):

Issue #3080, opened 2020-08-04
Expose `max_gas_burnt_view` as part of client config


evgenykuzyakov:
`max_gas_burnt_view` defines the maximum burnt gas for a view call.

Since it doesn't influence the protocol (only affects RPC), it should be part of the client config instead of genesis config.

alsakhaev:
I ran into the GasLimitExceeded error in my contract when calling the read function.

I started looking for where to change the gas limit in the node configuration. I have not found documentation for client configs anywhere. By examining the source code, it turned out that this parameter lies in the genesis (genesis.json), which is 5 GB in size. The built-in Linux text editors didn't want to work with such a large file. I discovered a HEX editor that can handle this. Replaced 200Tgas with 999Tgas.
![image](https://user-images.githubusercontent.com/15713727/117123608-ec40b100-ad9f-11eb-9de0-3ab66aca02c5.png)

Launched the client. He noticed the change by comparing the genesis hashes and cleared the entire state.
![image](https://user-images.githubusercontent.com/15713727/117123667-011d4480-ada0-11eb-9aa3-d9fa6e7bc07b.png)

Well, the synchronization of the node started from the very beginning.
![image](https://user-images.githubusercontent.com/15713727/117123689-08445280-ada0-11eb-96cf-945ab2495d18.png)

So, now it's impossible to change the gas limit for view functions.

bowenwang1996:
> So, now it's impossible to change the gas limit for view functions.

The config you modified is in genesis and is part of the protocol, so that doesn't work. What is proposed here is to add a separate client config that does not affect the protocol to increase the limit on view calls.

ethernian:
Yes, please fix the issue. 

It should be a no-brainer fix, max 10 lines of code to replace the source where the read-only gas limit comes from, isn't it? 
I do really not understand why the read-only gas limit is part of genesis and thus part of the consensus.

mina86:
This isn’t yet available in a production release but if anyone wishes to test it in betanet et al with a nightly build `neard` will now accept `--max-gas-burnt-view` switch for its `init` and `run` commands. It’s also possible to add `max_gas_burnt_view` key to the `config.json` file.

PR #4381, opened 2021-06-20
Introduce ‘max_gas_burnt_view’ client config and command line flag

mina86:
The ‘max_gas_burnt_view’ affects the RPCs only and therefore doesn’t
need to be part of the consensus.  In other words, all nodes could
have this value customised rather than using an agreed upon value
from genesis configuration.

Introduce a ‘max_gas_burnt_view’ field to ClientConfig and
a corresponding flag to ‘init’ and ‘run’ subcommands.  When present,
it will override the limit used by the node ignoring whatever is
stored in genesis configuration.

Fixes: https://github.com/near/nearcore/issues/3080

### DIFF
--- a/chain/client/tests/process_blocks.rs
+++ b/chain/client/tests/process_blocks.rs
@@ -85,6 +85,7 @@ pub fn create_nightshade_runtimes(genesis: &Genesis, n: usize) -> Vec<Arc<dyn Ru
                 vec![],
                 vec![],
                 None,
+                None,
             )) as Arc<dyn RuntimeAdapter>
         })
         .collect()
@@ -1804,6 +1805,7 @@ fn test_incorrect_validator_key_produce_block() {
         vec![],
         vec![],
         None,
+        None,
     ));
     let signer = Arc::new(InMemoryValidatorSigner::from_seed("test0", KeyType::ED25519, "seed"));
     let mut config = ClientConfig::test(true, 10, 20, 2, false, true);
@@ -2938,6 +2940,7 @@ mod protocol_feature_restore_receipts_after_fix_tests {
             &genesis,
             vec![],
             vec![],
+            None,
             None,
         );
         // TODO #4305: get directly from NightshadeRuntime
--- a/nearcore/src/config.rs
+++ b/nearcore/src/config.rs
@@ -431,6 +431,9 @@ pub struct Config {
     pub view_client_throttle_period: Duration,
     #[serde(default = "default_trie_viewer_state_size_limit")]
     pub trie_viewer_state_size_limit: Option<u64>,
+    /// If set, overrides value in genesis configuration.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub max_gas_burnt_view: Option<Gas>,
 }
 
 impl Default for Config {
@@ -456,6 +459,7 @@ impl Default for Config {
             view_client_threads: default_view_client_threads(),
             view_client_throttle_period: default_view_client_throttle_period(),
             trie_viewer_state_size_limit: default_trie_viewer_state_size_limit(),
+            max_gas_burnt_view: None,
         }
     }
 }
@@ -642,6 +646,7 @@ impl NearConfig {
                 epoch_sync_enabled: config.epoch_sync_enabled,
                 view_client_throttle_period: config.view_client_throttle_period,
                 trie_viewer_state_size_limit: config.trie_viewer_state_size_limit,
+                max_gas_burnt_view: config.max_gas_burnt_view,
             },
             network_config: NetworkConfig {
                 public_key: network_key_pair.public_key,
@@ -787,6 +792,7 @@ pub fn init_configs(
     genesis: Option<&str>,
     download: bool,
     download_genesis_url: Option<&str>,
+    max_gas_burnt_view: Option<Gas>,
 ) {
     fs::create_dir_all(dir).expect("Failed to create directory");
     // Check if config already exists in home dir.
@@ -798,12 +804,15 @@ pub fn init_configs(
     let chain_id = chain_id
         .and_then(|c| if c.is_empty() { None } else { Some(c.to_string()) })
         .unwrap_or_else(random_chain_id);
+    let mut config = Config::default();
+    if max_gas_burnt_view.is_some() {
+        config.max_gas_burnt_view = max_gas_burnt_view;
+    }
     match chain_id.as_ref() {
         "mainnet" => {
             if test_seed.is_some() {
                 panic!("Test seed is not supported for MainNet");
             }
-            let mut config = Config::default();
             config.telemetry.endpoints.push(MAINNET_TELEMETRY_URL.to_string());
             config.write_to_file(&dir.join(CONFIG_FILENAME));
 
@@ -824,7 +833,6 @@ pub fn init_configs(
             if test_seed.is_some() {
                 panic!("Test seed is not supported for official TestNet");
             }
-            let mut config = Config::default();
             config.telemetry.endpoints.push(NETWORK_TELEMETRY_URL.replace("{}", &chain_id));
             config.write_to_file(&dir.join(CONFIG_FILENAME));
 
@@ -858,7 +866,6 @@ pub fn init_configs(
         }
         _ => {
             // Create new configuration, key files and genesis for one validator.
-            let mut config = Config::default();
             config.network.skip_sync_wait = true;
             if fast {
                 config.consensus.min_block_production_delay =
--- a/nearcore/src/runtime/mod.rs
+++ b/nearcore/src/runtime/mod.rs
@@ -122,7 +122,10 @@ impl EpochInfoProvider for SafeEpochManager {
 /// TODO: this possibly should be merged with the runtime cargo or at least reconciled on the interfaces.
 pub struct NightshadeRuntime {
     genesis_config: GenesisConfig,
-    genesis_runtime_config: Arc<RuntimeConfig>,
+    /// Runtime configuration.  Note that it may be slightly different than
+    /// `genesis_config.runtime_config`.  Consider `max_gas_burnt_view` value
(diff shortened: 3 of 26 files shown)


mina86:
I haven’t yet fully tested that the limit actually takes hold. Any hints how to do that welcome. :)

ailisp:
> I haven’t yet fully tested that the limit actually takes hold. Any hints how to do that welcome. :)

I think the best way will be add a pytest (in pytest dir), where you can write a python test to override config.json and command line options, and send rpc request to check the setting is effective

mina86:
> and send rpc request to check the setting is effective

I’m struggling a bit with figuring out an RPC to test the setting.

mina86:
The check failure looks related to <https://github.com/actix/actix-web/issues/2185>. I’ve tried updating `actix-web` release but that didn’t solve it. The issue is closed but also states that ‘-extras crates to follow’ so we might need to wait for a proper fix.

bowenwang1996 (commented):
I am a bit unsure whether it is a good idea to make `max_gas_burnt_view` part of RuntimeConfig since `RuntimeConfig` is part of the protocol right now and mixing it with non-protocol related config could cause confusion

mina86:
It already is part of the RuntimeConfig (and is included in genesis config). This change will make it so that the actual value used by the runtime will be overridden.

bowenwang1996:
oh I didn't realize that. @evgenykuzyakov what was the motivation to have it as part of `RuntimeConfig`?

bowenwang1996 (commented):
If you want to properly test this change, you can first write a contract that has some method that costs more than 200Tgas to run and then either write a python test or a unit in `process_blocks.rs` to call that method through a view call.

ailisp:
> The check failure looks related to actix/actix-web#2185. I’ve tried updating actix-web release but that didn’t solve it. The issue is closed but also states that ‘-extras crates to follow’ so we might need to wait for a proper fix.

I encountered this fail in my PR too, one way of fix this is to pin actix-http: https://github.com/near/nearcore/pull/4376/commits/7dcfd0d12b525bcf311a5b63067d1e6975c5c974 and also update Cargo.lock https://github.com/near/nearcore/pull/4376/commits/6fa4bd8157295cc32fbf0251a136a7061a84e4b7

tools/indexer/example/Cargo.toml
```diff
@@ -14,3 +14,4 @@ tracing = "0.1.13"
 tracing-subscriber = "0.2.4"
 
 near-indexer = { path = "../../../chain/indexer" }
+near-primitives = { path = "../../../core/primitives" }
```
frol, line 17: `near_primitives` are re-exported from `near-indexer`, so instead of adding it as a direct dependency, just `use near_indexer::near_primitives::types::Gas`.

tools/indexer/example/src/configs.rs
```diff
@@ -1,5 +1,6 @@
 use clap::Clap;
 
+use near_primitives::types::Gas;
```
frol, line 3: ^^^

frol (approved):
Overall looks good to me, so I unblock the PR, but please address the comments about the Indexer example.

nearcore/src/runtime/mod.rs
```diff
@@ -122,7 +122,10 @@ impl EpochInfoProvider for SafeEpochManager {
 /// TODO: this possibly should be merged with the runtime cargo or at least reconciled on the interfaces.
 pub struct NightshadeRuntime {
     genesis_config: GenesisConfig,
-    genesis_runtime_config: Arc<RuntimeConfig>,
+    /// Runtime configuration.  Note that it may be slightly different than
+    /// `genesis_config.runtime_config`.  Consider `max_gas_burnt_view` value
+    /// which may be configured per node.
+    runtime_config: Arc<RuntimeConfig>,
```
matklad, line 128: I feel that config handling here accumulated quite a bit of technical debt, as we now have to separate paths that "patch" config: one with `max_gas_burnt`, and another with `RuntimeConfig::from_protocol_version` (which is especcially confusing, as it relies on hidden global state). I have a stalled PR which cleans parts of this up in https://github.com/near/nearcore/pull/4263#issuecomment-832113787.

This isn't strictly related to the PR, so it isn't a blocker, but I want to raise awareness and signal that straightening this up is non-urgent, but important ;-)

nearcore/src/runtime/mod.rs
```diff
@@ -142,11 +145,14 @@ impl NightshadeRuntime {
...
         initial_tracking_shards: Vec<ShardId>,
         trie_viewer_state_size_limit: Option<u64>,
+        max_gas_burnt_view: Option<Gas>,
     ) -> Self {
         let runtime = Runtime::new();
         let trie_viewer = TrieViewer::new_with_state_size_limit(trie_viewer_state_size_limit);
         let genesis_config = genesis.config.clone();
-        let genesis_runtime_config = Arc::new(genesis_config.runtime_config.clone());
+        let runtime_config = Arc::new(
+            genesis_config.runtime_config.with_custom_max_gas_burnt_view_maybe(max_gas_burnt_view),
```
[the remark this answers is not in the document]
mina86: Nah, you’re right to call me up on this one. Removed.

tools/indexer/example/Cargo.toml
mina86: Done.

mina86:
Added tests though also added a `fib.wasm` blob which I’m not super happy about but haven’t yet figure out how to use runtime/near-test-contracts.

pytest/tests/sanity/rpc_max_gas_burnt.py
```diff
@@ -0,0 +1,63 @@
...
+def test_max_gas_burnt_view():
+    nodes = start_cluster(2, 0, 1,
+                          config=None,
+                          genesis_config_changes=[],
+                          client_config_changes={
+                              1: {'max_gas_burnt_view': int(1e10)}
+                          })
+
+    contract_key = nodes[0].signer_key
+    fib_contract = load_binary_file('testdata/fib.wasm')
```
matklad, line 27: Yeah, I'd rather not add new blobs to the repo if possible. https://github.com/near/nearcore/pull/4397 adds a new function to the contract, you can use it as an example. 

The following impl of fib should work with our test contract: 



```
diff --git a/runtime/near-test-contracts/test-contract-rs/src/lib.rs b/runtime/near-test-contracts/test-contract-rs/src/lib.rs
index 27f6a0c82..8e10cc91b 100644
--- a/runtime/near-test-contracts/test-contract-rs/src/lib.rs
+++ b/runtime/near-test-contracts/test-contract-rs/src/lib.rs
@@ -294,6 +294,26 @@ pub unsafe fn write_key_value() {
     value_return(size_of::<u64>() as u64, &result as *const u64 as u64);
 }
 
+#[no_mangle]
+pub unsafe fn fib() {
+    input(0);
+    if register_len(0) != size_of::<u64>() as u64 {
+        panic()
+    }
+    let data = [0u8; size_of::<u64>()];
+    read_register(0, data.as_ptr() as u64);
+    let key = u64::from_le_bytes(data);
+    let result = fib_inner(key).to_le_bytes();
+    value_return(result.len() as u64, result.as_ptr() as u64);
+
+    fn fib_inner(n: u64) -> u64 {
+        match n {
+            0 | 1 => 1,
+            _ => fib_inner(n - 1) + fib_inner(n - 1),
+        }
+    }
+}
+
 #[no_mangle]
 pub unsafe fn write_block_height() {
     let block_height = block_index();
```

mina86: Thanks. I was trying to get it done in AssemblyScript but Rust is much easier. Done.

bowenwang1996 (approved):
Great work!

chain/client/tests/process_blocks.rs
```diff
@@ -85,6 +85,7 @@ pub fn create_nightshade_runtimes(genesis: &Genesis, n: usize) -> Vec<Arc<dyn Ru
                 vec![],
                 vec![],
                 None,
+                None,
```
olonho, line 88: Seems signature changes like that one could be very broad. Could we agree that instead we pass an aggregating structure containing fields instead of long arg list?

mina86: Yes, I have considered doing that. I still might but I don’t want to include such refactoring in this commit since it’s an unrelated change.

nearcore/src/config.rs
```diff
@@ -431,6 +431,9 @@ pub struct Config {
     pub view_client_throttle_period: Duration,
     #[serde(default = "default_trie_viewer_state_size_limit")]
     pub trie_viewer_state_size_limit: Option<u64>,
+    /// If set, overrides value in genesis configuration.
+    #[serde(skip_serializing_if = "Option::is_none")]
```
olonho, line 435: Hmm, so intention is to allow keeping backward and forward compatibility with serilized data if option is None?

olonho (approved):
Overall looks good, modulo potential argument refactoring and subtle serialized format change, I'm unsure about.

nearcore/src/config.rs
mina86: Mostly yes. Technically it’s probably not necessary (since IIUC unknown fields are ignored and missing fields get None value if they are Option type) but I didn’t actually verify that. There are other Option fields which are skipped if None so precedent is there.

chain/client/tests/process_blocks.rs
mina86: Note however that introducing a structure with arguments won’t fix the issue of changes to the structure affecting all call sites. It will make the calls more readable though thanks to the use of named fields. Issue is that some of the arguments are required and thus there won’t be any sensible implementation of Default for such structure. Builder would be the only solution here but that’s even larger refactoring.

ailisp (approved):
Looks good!

PR #4382, opened 2021-06-20
fix(chain): change max gas price

bowenwang1996:
In #4348 we set the maximum gas price to be 10 * minimum gas price. However, one issue with this change is that it changes max gas price to be 10^9 yotcoNEAR, which means that per block max cost is 1N (10^9 yotcoNEAR/gas * 10^15 gas) and this means in a year the maximal burnt tokens in transaction fees is ~31 million NEAR, which will still be smaller than the total inflation (~50 million NEAR). This PR changes the multiplier to 20x to ensure that the network at least does not inflate at maximal capacity.

Note that not changing nightly protocol version is intentional here -- there is no congestion on betanet anyways and we don't need to do a hard fork there.

Test plan
----------
Existing tests.


chain/chain/src/types.rs
```diff
@@ -172,7 +174,10 @@ impl BlockEconomicsConfig {
     pub fn max_gas_price(&self, protocol_version: ProtocolVersion) -> Balance {
         if checked_feature!("protocol_feature_cap_max_gas_price", CapMaxGasPrice, protocol_version)
         {
-            std::cmp::min(self.max_gas_price, 10 * self.min_gas_price(protocol_version))
+            std::cmp::min(
+                self.max_gas_price,
+                Self::MAX_GAS_MULTIPLIER * self.min_gas_price(protocol_version),
```
EgorKulikov, line 179: Not likely to matter, but maybe do checked multiplication?

EgorKulikov (approved):

Longarithm (approved):

chain/chain/src/types.rs
bowenwang1996: well we know the constants here and they will not cause overflow

PR #4387, opened 2021-06-21
fix(rpc): avoid null response in sandbox rpc

ailisp:
sandbox rpc currently still returns `{"result":null, ...}`, this breaks current release of near-api-js (which is fixed in https://github.com/near/near-api-js/pull/603 but it's not yet released). Change it to make it return `{"result":{}}` so even current release near-api-js works.

Test Plan
------------
CI. Curl on a sandbox node


frol (changes requested):
Given that this endpoint is specific to the sandbox usage, which is quite new, I believe it should be enough to update `near-api-js` and don't push workarounds into JSON RPC implementation. WDYT?

ailisp:
@frol wait a new near-api-js release would work, but I do think `"result":{}` is semantically better than `"result":null` because `{}` is true in javascript, and means an empty response object, while null means None or nothing. WDYT?

frol:
@ailisp to me `null` means there is no meaningful information that can be provided. `{}` means that there is something, but it is missing in the output, so it is something useless. If you want to return something meaningful, I can suggest we return a block height / hash when the patched state got effective or some other useful information.

ailisp:
@frol Makes sense. I read more discussions on return `null` or `{}` in jsonrpc response, both have bunch of supporters. I think there isn't a need to return a block height goes effective is needed because it's synced, user can expect view state changed after this rpc returned. So say it's still return empty response. I'm think of `null` means void or doesn't return anything, while `{}` means empty object, Unit, or returns an empty value. Since we do have a `RpcSandboxPatchStateResponse` object instead of a fn with no return type, I think it's more match `{}` than `null`.

matklad:
As an example, LSP uses `null` for void-returning rpc request: https://microsoft.github.io/language-server-protocol/specifications/specification-current/#responseMessage.

I'd probably even go as far as defining `type RpcSandboxPatchStateResponse = ();`

No strong opinion on what we should do here, *but* I think it'd be good to document the decision somewhere, so that, if we have another void RPC in the future, it uses the same pattern.

frol:
Well, on second thought, I guess `{}` is more future-compatible. Let's merge this one.

frol (approved):

PR #4389, opened 2021-06-21
Introduce `TrieRefcountChange` for better readability

Longarithm:
One step to solve https://github.com/near/NEPs/issues/217

It took a while to understand what exactly `(CryptoHash, Vec<u8>, u32)` stores, so I make a struct which self-describes it.

### DIFF
--- a/core/store/src/trie/mod.rs
+++ b/core/store/src/trie/mod.rs
@@ -409,6 +409,17 @@ pub struct Trie {
     pub counter: TouchedNodesCounter,
 }
 
+/// Stores reference count increase for the given key and value.
+#[derive(BorshSerialize, BorshDeserialize, Clone, PartialEq, Eq, PartialOrd, Ord, Debug)]
+pub struct TrieRefcountChange {
+    /// Hash of TrieKey.
+    key_hash: CryptoHash,
+    /// Value corresponding to the TrieKey stored in Trie.
+    value: Vec<u8>,
+    /// Reference count increase.
+    rc: u32,
+}
+
 ///
 /// TrieChanges stores delta for refcount.
 /// Multiple versions of the state work the following way:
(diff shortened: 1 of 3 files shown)


core/store/src/trie/mod.rs
```diff
@@ -683,13 +695,17 @@ impl Trie {
...
         let mut insertions = Vec::new();
         for (key, (value, rc)) in changes.into_iter() {
             match rc.cmp(&0) {
-                Ordering::Greater => insertions.push((key, value, rc as u32)),
-                Ordering::Less => deletions.push((key, value, (-rc) as u32)),
+                Ordering::Greater => {
+                    insertions.push(TrieRefcountChange { key_hash: key, value, rc: rc as u32 })
+                }
+                Ordering::Less => {
+                    deletions.push(TrieRefcountChange { key_hash: key, value, rc: (-rc) as u32 })
```
bowenwang1996, line 707: This feels a bit weird. Why do we want to cast it as u32?

core/store/src/trie/state_parts.rs
```diff
@@ -542,11 +544,11 @@ mod tests {
...
             assert!(changes_set.deletions.is_empty(), "state parts only have insertions");
-            for (key, value, rc) in changes_set.insertions {
-                map.entry(key).or_insert_with(|| (value, 0)).1 += rc as i32;
+            for TrieRefcountChange { key_hash, value, rc } in changes_set.insertions {
+                map.entry(key_hash).or_insert_with(|| (value, 0)).1 += rc as i32;
             }
-            for (key, value, rc) in changes_set.deletions {
-                map.entry(key).or_insert_with(|| (value, 0)).1 -= rc as i32;
+            for TrieRefcountChange { key_hash, value, rc } in changes_set.deletions {
+                map.entry(key_hash).or_insert_with(|| (value, 0)).1 -= rc as i32;
```
bowenwang1996, line 551: I am not sure why we need to cast back and forth between u32 and i32.

core/store/src/trie/mod.rs
Longarithm: Here changes are separated to insertions and deletions.
Not sure why it is needed, but for insertion or deletion `refcount` change number is always positive. `u32` seems suitable to emphasize this fact.

bowenwang1996: > refcount change number is always positive

Not entirely sure what you mean. Do you mean that the result is always nonnegative? That does not change the fact that deletion will decrease rc, which means that it is negative here. Also later on we cast it back to i32 to calculate the resulting rc. So I am not sure why we need this casting here.

bowenwang1996 (approved):

core/store/src/trie/mod.rs
Longarithm: Discussed offline

Issue #4379, opened 2021-06-18
Add CI test that checks that attempt publishing crates and compile them to Wasm


nearmax:
Some of the crates in nearcore are made publishable to crates.io. This is done so that external tools like near-sdk-sim can use them without pulling cross-repo github dependency which frequently break.

Also, some of the crates need to be compilable to Wasm because they are used for test in near-sd-as.

Unfortunately, there are no CI checks that make sure this is correct, and PRs like this one https://github.com/near/nearcore/commit/d336b3fc1b9dc2167aa2eb9a89e24e4a2a09e27d#r52341955 can break the above invariants. We need to add the following checks into our CI:
* A dry-run attempt to publish publishable crates to crates.io;
* A Wasm compilation of certain crates.

mina86:
Adding a dry-run publish test won’t work at this stage.  Before such
check is added we need to:

* publish a new `near-primitives` release version 0.1.0 (alternatively
  update dependencies to in Cargo.toml to 0.1.0-pre.1),

* publish a new ‘near-vm-errors’ release which includes
  `protocol_feature_alt_bn128` feature and

* publish a new `near-vm-runner` release version 3.0.0,

* create a new `near-evm-runner` crate.

To be honest I’m sceptical that such test will actually be beneficial.
For example, it will force us to publish crates as soon as a new
feature is added to them.  Similarly, make new crates even if they’re
at a very early experimental stages.

matklad:
>Also, some of the crates need to be compilable to Wasm because they are used for test in near-sd-as.

Oh wow, I didn't realize that assembly script tests work by compiling our vmlogic to wasm to implement mocked blockchain in the javascript environment. That makes sense, but also indeed places a rather big constraint on what we are allowed to do in the relevant parts. 

Publishing to crates.io and wasm compatability are rather big asks, because the establish a boundary in the code. It's best if this boundary is physically reflected in code layout, rather than it being an informal agreement that "we publish this, this, and this crate, and we don't publish that crate". As such, I think the best long-term solution is to introduce a single umbrella crate for publishing & wasm compat. 

[...]

PR #4392, opened 2021-06-21
Improve ‘publishability’ of crates

mina86:
To be able to publish a crate all its dependencies must be versioned.
Make sure that this is the case for all packages listed in the
`runtime/publish.sh` script.

Note that this does not fix the issue completely and there are still
other problems stopping the crates from being published.  Most
notably, some of the dependencies not being published yet.

Issue: https://github.com/near/nearcore/issues/4379


bowenwang1996 (commented):
I am confused by this change. I think we want to have `near-primitives-core` instead of `near-primitives` because the dependency is smaller, not the other way around

mina86:
Yes, but if a crate already depends on `near-primitives` there’s no point in *also* depending on `near-primitives-core`.  The former re-exports everything the latter offers.  This commit does not add any new dependencies.  It only removes `near-primitives-core` where it’s redundant and makes sure that all dependencies have version specified (which is required to be able to publish a crate).

If you prefer I kept both dependencies I can alter this commit to only add the version where it’s needed.

mina86:
Though now that I see https://github.com/near/nearcore/pull/4378 perhaps it’s best to wait with this commit.

bowenwang1996:
> Yes, but if a crate already depends on near-primitives there’s no point in also depending on near-primitives-core. The former re-exports everything the latter offers. This commit does not add any new dependencies. It only removes near-primitives-core where it’s redundant and makes sure that all dependencies have version specified (which is required to be able to publish a crate).

Yes but the current dependency in vm logic is a mistake that is being fixed

PR #4394, opened 2021-06-22
Add `--dry-run` option to `runtime/publish.sh` script

mina86:
The `--dry-run` flag will allow to run `runtime/publish.sh` script to
test whether all the crates are in a state where they can be published.
Note that at the moment, this check fails due to problems described in
the GitHub Issue.

Issue: https://github.com/near/nearcore/issues/4379

### DIFF
--- a/runtime/publish.sh
+++ b/runtime/publish.sh
@@ -1,10 +1,31 @@
 #!/usr/bin/env bash
-set -ex
-for p in near-runtime-utils near-vm-errors near-vm-logic near-vm-runner near-vm-runner-standalone
-do
-pushd ./${p}
-cargo publish
-popd
-# Sleep a bit to let the previous package upload to crates.io. Otherwise we fail publishing checks.
-sleep 30
+
+set -eu
+
+dry_run_arg=
+if [ "$#" -ge 1 ]; then
+	case "${1-}" in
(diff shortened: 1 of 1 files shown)


bowenwang1996 (approved):

matklad:
Historically, I wasn't able to successfully use `cargo publish --dry-run` in workspaces, as, when you are dry-running publishing of a crate, cargo assumes that all the deps were actually published. We can try this, but I am not sure we'll succeed. 

A plausible alternative is automated publishing. For example, we publish rust-analyzer crates every week: https://crates.io/crates/ra_ap_ide/versions. But that's a big change of a model from "we maintain semver" to "we don't care about API compatibility just always increment major".

mina86:
Yeah, this is something I’ve hinted at [in the issue](https://github.com/near/nearcore/issues/4379). Having this flag still may make some doze of sense since it allows to first test whether `publish.sh` would work in the first place. It’s hard for me to really judge as I’m not familiar with the publish process or policy.

It seems that what we want is to make sure that the crates that are published don’t have dependencies on anything that is not published so people don’t add undesired dependencies by mistake. This could potentially be achieved by perhaps parsing `cargo tree`.

PR #4164, opened 2021-03-24
Test(runtime): repro of evm contract slow issue, bottleneck is on deser

ailisp:
This shows the slowness of deploy a common sized evm contract (the same test Illia did), Previous repro was a unit test: https://github.com/near/nearcore/compare/cache-slow-repro. This is rewritten with Aleksey's runtime standalone rust script facility which clearly show bottleneck is in deserialization:
```
running 1 test
      13.01µs get_key
      80.94ms compile_and_serialize_wasmer
        938.04µs run_method/instantiate
        4.06ms run_method/call
        949.55µs run_method/drop_instance
      6.00ms run_method
    89.79ms run_wasmer
  89.81ms run_vm

      3.18µs get_key
      19.49ms deserialize_wasmer
        951.93µs run_method/instantiate
        3.98ms run_method/call
        1.03ms run_method/drop_instance
      6.01ms run_method
    28.94ms run_wasmer
  28.96ms run_vm

      3.49µs get_key
      17.68ms deserialize_wasmer
        949.29µs run_method/instantiate
        3.99ms run_method/call
        710.68µs run_method/drop_instance
      5.69ms run_method
    26.24ms run_wasmer
  26.25ms run_vm

test script::evm_slow_deserialize_repro ... ok
```

It's improving in non-optimized wasmer1 (result below), and hopefully going to be fast enough in optimized wasmer1 (untested):
```
running 1 test
      13.57µs get_key
      49.76ms compile_and_serialize_wasmer1
        534.65µs run_method/instantiate
        2.38ms run_method/call
        4.16µs run_method/drop_instance
      2.96ms run_method
    53.52ms run_wasmer1
  53.55ms run_vm

      3.61µs get_key
      2.79ms deserialize_wasmer1
        364.05µs run_method/instantiate
        2.12ms run_method/call
        2.64µs run_method/drop_instance
      2.53ms run_method
    6.34ms run_wasmer1
  6.35ms run_vm

      2.99µs get_key
      1.80ms deserialize_wasmer1
        292.43µs run_method/instantiate
        2.10ms run_method/call
        2.65µs run_method/drop_instance
      2.44ms run_method
    5.32ms run_wasmer1
  5.33ms run_vm

test script::evm_slow_deserialize_repro ... ok
```

Test Plan
---------
```
cargo test --release --package near-vm-runner-standalone --bin near-vm-runner-standalone --features wasmer0_default,no_cache -- script::evm_slow_deserialize_repro --exact --nocapture

[...]

Issue #4358, opened 2021-06-09
Avoid building all three wasm engines for production use


matklad:
At the moment, we enable all three vms using default features in near-vm-runner:

[https://github.com/near/nearcore/blob/07b4fd45f90570d9c4f306f724cf84499cc519c7/runtime/near-vm-runner/Cargo.toml#L48](https://github.com/near/nearcore/blob/07b4fd45f90570d9c4f306f724cf84499cc519c7/runtime/near-vm-runner/Cargo.toml#L48)

The problem with `default` features is that they are impossible to disable. Our current setup is more or less equivalent to just always using all three vm. Indeed, if you remove this default feature, you'll notice that `near-vm-runner` doesn't even compile, as it misses some important `cfg`s.

At the same time, wasm runtimes are very heavy dependencies (take long time to compile, and often use global resources like `#<span class="error">[no_mangle]</span>` functions or signal handlers), so using conditional compilation here makes sense.

I think to fix this we need:

*   change nightly CI setup to run `cargo test --workspace --features nightly_protocol,nightly_protocol_features,protocol_feature_evm,wasmer0_vm,wasmer1_vm,wasmtime_vm`
*   change to `default = <span class="error">["wasmer0_vm"]</span>` only
*   fix compilation time errors

matklad:
cc @olonho @ailisp

ailisp:
we haven't reach consensus on it at last meeting, @olonho @matklad let's make a conclusion on this today

matklad:
Discussed in today's meeting. Summary:

* for contract runtime team, it's more convenient if all vms are enabled by default
* for other teams, this creates extra compilation time tax (b/c vm deps are heavy)
* a nice middleground is to ensure that we test all vms on CI, but ship only one

Specific proposal: 

* remove `wasmer1_vm` and `wasmtime_vm` from the set of [default](https://github.com/near/nearcore/blob/ff01c544ce2675fadbf0284af52b356a383b27a8/runtime/near-vm-runner/Cargo.toml#L47-L49) features in Cargo.toml
* add those features to the [nightly tests](https://github.com/near/nearcore/blob/ff01c544ce2675fadbf0284af52b356a383b27a8/.buildkite/pipeline.yml#L15) instead.

matklad:
cc @akhi3030

PR #4395, opened 2021-06-22
feat(runtime): wasmer2

ailisp:
Replace wasmer1 with wasmer2 since it's never went to mainnet. Some breaking changes, major ones:
Four traps are removed in these three commits:
https://github.com/wasmerio/wasmer/commit/657a6cd69b1a8829b9223beba9eaa79bfadb29cb
https://github.com/wasmerio/wasmer/commit/c45b5f804c46306729fe248bff74e462b1d27251
https://github.com/wasmerio/wasmer/commit/8b86f795b734f2cf437bb1ed76fd21d564a9d7ab

JIT engine is renamed to Universal engine: https://github.com/wasmerio/wasmer/pull/2340
Native engine is renamed to Dylib: https://github.com/wasmerio/wasmer/pull/2356

As for error and traps: same or become a subset of wasmer 1's.
As for potential regressions: should be covered by existing near-vm-runner error cases tests.
As for performance: Run https://github.com/near/nearcore/pull/4164 (need add wasmer1/wasmer2_default feature to cargo, and compile with release, no-cache feature), wasmer2 has similar runtime and slightly faster deserialization.

Test Plan
---------
near-vm-runner tests


ailisp:
near-vm-runner tests failed, however, if run test one module at a time then every near-vm-runner tests passed except error_cases test, which must be run one test at a time
image
If run them together fails:
image
The problem seems to be one trap in error_cases test is not caught and cause entire binary exit

Cargo.toml
```diff
@@ -48,10 +48,6 @@ members = [
 
 [patch.crates-io]
 ethereum-types = { path = "./patches/ethereum-types-0.10.0-to-0.11.0" }
-# Temporary wasmer fork which disables rayon (parallel compilation)
-# https://github.com/wasmerio/wasmer/pull/2262
-# Revome this when the new version of wasmer is published.
-wasmer-compiler-singlepass = { git = "https://github.com/near/wasmer", branch = "1.0.2-single-threaded" }
```
matklad, line 54: :tada:

runtime/near-vm-runner/src/wasmer2_runner.rs
```diff
@@ -286,53 +274,54 @@ fn run_method(module: &Module, import: &ImportObject, method_name: &str) -> Resu
...
+struct Wasmer2Config {
     seed: i32,
     use_cranelift: bool,
     use_native_engine: bool,
 }
 
-impl Wasmer1Config {
+impl Wasmer2Config {
     fn config_hash(self: Self) -> u64 {
         let mut s = DefaultHasher::new();
```
matklad, line 291: Unrelated to this PR, but I now realize that using `DefaultHasher` is suboptimal here -- Rust doesn't guarantee stability of the hasher, so, rust version upgrade might bust the cache. 

We should steal this module from Cargo: 

https://github.com/rust-lang/cargo/blob/58a961314437258065e23cb6316dfc121d96fb71/src/cargo/util/hasher.rs

This will both guarantee hash stability in the future, as well as compatibility with the current hash we de-facto use.

cc @olonho

runtime/near-vm-runner/src/wasmer2_runner.rs
```diff
@@ -286,53 +274,54 @@ fn run_method(module: &Module, import: &ImportObject, method_name: &str) -> Resu
...
         self.hash(&mut s);
         s.finish()
     }
 }
 
-const WASMER1_CONFIG: Wasmer1Config =
-    Wasmer1Config { seed: 53, use_cranelift: false, use_native_engine: false };
+const WASMER2_CONFIG: Wasmer2Config =
+    Wasmer2Config { seed: 53, use_cranelift: false, use_native_engine: false };
```
matklad, line 298: Let's change the seed here to actually bust the cache.

runtime/near-vm-runner/src/vm_kind.rs
```diff
@@ -9,13 +9,13 @@ pub enum VMKind {
     /// Wasmtime VM.
     Wasmtime,
     /// Wasmer 1.x VM.
-    Wasmer1,
+    Wasmer2,
 }
```
matklad, line 13: Let's preserve serialized format here by explicitelly changing discriminants:

```rust
#[derive(Clone, Copy, Debug, Hash, Serialize, Deserialize, BorshSerialize, BorshDeserialize)]
pub enum VMKind {
    Wasmer0 = 0,
    Wasmtime = 1,
//  Wasmer1 = 2,
    Wasmer2 = 3,
}
```

Let's also remove `Deserialize` and `BorshDeserialize` -- I don't belive they are actually used anywhere (transitive structs can have them removed as well), and that'll statically guarantee that removing vm kind won't blow up.

matklad (dismissed):

ailisp:
I further investigate by running test sequentially and move around/comment/rename: `cargo test --package near-vm-runner --lib tests::error_cases -- --test-threads 1 --nocapture`. The tests are run in alphabetic order this way.
Very weird finding:
test_trap_initializer must be put before several other tests to pass all tests, for example, rename it as `test_a_trap_initializer` to make it alhabetically first, otherwise those several test will fail at:
```
thread 'main' panicked at 'not within a catch_unsafe_unwind scope', /home/bo/.cargo/registry/src/github.com-1ecc6299db9ec823/wasmer-runtime-core-near-0.17.1/src/fault.rs:212:10
```

And this doesn't happen when there's no wasmer 2. Still analyzing this puzzle, can't reason it so far :thinking:

ailisp:
Obviously wasmer 0 and wasmer 2 have installed different global trap handlers. And they don't play well together. 
When only run wasmer 0 tests it passes. When only run wasmer 2 tests it also passes. I still not fully figure out in which way they conflict and how to reason about the failure. A minimum repro example is to only keep these two tests:
```
fn test_trap_contract()
fn test_bad_import_1()
```
Adjust their names & run them sequentially in alphabetic order with:
```
cargo test --package near-vm-runner --lib tests::error_cases -- --test-threads 1 --nocapture
```
And when trap_contract is executed first it passes, fails not when the reverse.
Considering we'll fully replace wasmer0 with wasmer2 in future, and when only run wasmer 2, tests passed, so I make a hack to prerun test_trap_contract to "warm up wasmer 0's trap handler when wasmer 2's handler is present" to make all tests pass.

bowenwang1996 (commented):
We need to figure out the difference in trap handlers to ensure that when we do the switch, we cannot be attacked by someone exploiting the differences in trap handling.

ailisp:
ping

runtime/near-vm-runner/src/tests.rs
```diff
@@ -24,14 +26,27 @@ const PREDECESSOR_ACCOUNT_ID: &str = "carol";
 const LATEST_PROTOCOL_VERSION: ProtocolVersion = ProtocolVersion::MAX;
 
 fn with_vm_variants(runner: fn(VMKind) -> ()) {
+    #[cfg(feature = "wasmer0_vm")]
+    static WASMER0_TRAP_HANDLER_SETUP: OnceCell<()> = OnceCell::new();
+    #[cfg(feature = "wasmer0_vm")]
+    WASMER0_TRAP_HANDLER_SETUP.get_or_init(|| {
```
matklad, line 32: A more minimal way to do this would be:

```rust
static WASMER0_TRAP_HANDLER_SETUP: std::sync::Once = std::sync::Once::new();
```

* no need to depend on external library that way
* no need to gate the import.

matklad (commented):
This looks good to me. I am debating however, whether we should finish #4358 first?

The trap handling issue is I think explained by the fact that wasm0 and wasm2 race to install the handler. Wasmer0 overwrites the handler, while wasmer2 chains. So, we should make sure that wasmer0 handler is installed first.

In theory, if we only run wasm0 at runtime, we should be safe. But I think it still makes sense to put this on hold until we only use wasmer0 in produciton.

ailisp:
Recently discussed with @matklad we decide to have wasmer2 before figure out #4358

runtime/near-vm-runner/src/tests.rs
ailisp: Sounds good. Replace with std Once now. However install the handler only in test does not bring correct behavior in runner, and install wasmer0 handerl indrectly via invoke `error_cases::trap_contract()` is not very good, therefore I'll move this to a direct install and to vm-runner.

ailisp:
@matklad  I fixed the handler in a not very hacking way: https://github.com/near/nearcore/pull/4395/commits/6d9ec2aa8e9f793ba2e2fd1228f2d6c239bcaf16. And after check code I see wasmer 2 chains trap handlers while wasmer 0 replace. So pre register wasmer 0 trap handler at the first run_vm ensures wasmer 0 trap handler is registered first and only once. However although it fix the tests, I don't see why. In tests wasmer 0 is run before wasmer 2:
```
    #[cfg(feature = "wasmer0_vm")]
    runner(VMKind::Wasmer0);
//... wasmtime
    #[cfg(feature = "wasmer2_vm")]
    runner(VMKind::Wasmer2);
```
Therefore for a contract trap case, it run into trap for both wasmer 0 and wasmer 2, and at first it runs for wasmer 0, it should register wasmer 0's trap handler but the fact is without 6d9ec2aa8e9f793ba2e2fd1228f2d6c239bcaf16 or explicit pre-run a trap-contract on wasmer 0 it doesn't  :thinking: 

Also I found a weird bug when running wasmer 2 tests. It does not related to wasmer 0 as comment wasmer 0 runner produce same result. The bug is, if run 
```
cargo test -p near-vm-runner -- --test-threads=1
```
It failed at:
```
test tests::error_cases::test_div_by_zero_contract ... fatal runtime error: failed to initiate panic, error 5
error: test failed, to rerun pass '-p near-vm-runner --lib
```
If run with more than one threads, or only all error_cases tests with one test-thread, it passes. And RUST_BACKTRACE has no effect. Do you have some idea on what's the problem

matklad:
Could you try using this version of wasmer0: https://github.com/near/wasmer/pull/42? That PR adds signal chaining, in theory that should solve the problem

Cargo.toml
```diff
@@ -47,7 +47,10 @@ members = [
 ]
 
 [patch.crates-io]
-ethereum-types = { path = "./patches/ethereum-types-0.10.0-to-0.11.0" }
+# Temporary wasmer fork which disables rayon (parallel compilation)
+# https://github.com/wasmerio/wasmer/pull/2262
+# Revome this when the new version of wasmer is published.
```
bowenwang1996:
```suggestion
# Remove this when the new version of wasmer is published.
```

ailisp: This lines and comments are dropped in a later commit (new version of wasmer shipped with feature to disable rayon) Thanks!

ailisp:
@matklad unfortunately Nikolay's patch doesn't work. First I comment out pre-install wasmer 0 handler:
```
    //    #[cfg(feature = "wasmer0_vm")]
    //    wasmer_runtime_core::fault::ensure_sighandler();
```
Then,
test-threads=1, wasmer 0 enabled: failed with same error
no test-threads limit, wasmer 0 enabled: failed with same error (This case passed on this PR, when pre-install wasmer 0 handler exists: 6d9ec2a 2eb8dab. This may indicate chain wasmer 0 trap handler implementation is incompatible with wasmer 2's in some way)

ailisp:
@matklad I created a binary to skip cargo test managed multi-thread/single-thread testing and possible cargo test installed signal handlers (I'm not entirely sure if there's, but I guess there should be because `cargo test` can contiue run other tests when a test failed on some SIGXXX): https://github.com/near/nearcore/pull/4625. Run it by
```
cargo run --package near-vm-runner --bin trap-handlers
```

There are no errors that happened when `cargo test` with `test-threads=1`. And 
```
    #[cfg(feature = "wasmer0_vm")]
    wasmer_runtime_core::fault::ensure_sighandler();
```
is necessary otherwise this binary fails, which exactly reproduce the failure of above minimum cargo test:
> In our codebase, a minimum repro example is to only keep these two tests:
fn test_trap_contract()
fn test_bad_import_1()
Adjust their names & run them sequentially in alphabetic order with:
cargo test --package near-vm-runner --lib tests::error_cases -- --test-threads 1 --nocapture

So my conclusion there is:
- `cargo test ... -- -test-threads 1` to test wasmer trap handlers is problematic, since binary target doesn't have the issue
- PR of current state is good, wasmer 0 and 2's trap handler doesn't conflict and it can continue run wasmer 0 & wasmer 2 after trap happens without crash, as proved by `cargo run --package near-vm-runner --bin trap-handlers`.

matklad:
Hm, it still feels rather wrong to me that `cargo t -p near-vm-runner --lib -- --test-threads 1` passes on master, but fails on this PR. I think that "test never fail, when run in any reasonable way" is an important baseline which we need to preserve. From the disccussion and quick testing (commenting out wasmtime and wasmer0 in `with_vm_variants`), it seems that wasmer2 alone is enough to cause `cargo t -p near-vm-runner --lib -- --test-threads 1`.

So perhaps this is a wasmer2 bug which should be reported upstream?

ailisp:
@olonho If you plan to work on wasmer 2 test, here is the status in summary:
1) This PR's currently pre register wasmer 0 trap handler before wasmer 2's, and cargo test pass.
2) This PR when cargo test with `--test-threads=1`, it failed. It's failed even when there's only wasmer 2, no wasmer 0.
3) When I call each test function sequentially as a rust binary, it pass: https://github.com/near/nearcore/pull/4625. This means wasmer 2's trap handler doesn't play well with cargo test harness.

Therefor wasmer 2 tests pass CI (without `--test-threads=1`) and works in binary that runs in a single thread, so `neard`, as a binary should be safe to upgrade to wasmer2. So my opinion is merge this PR and address `2.` as a separate issue

runtime/near-vm-runner/Cargo.toml
```diff
@@ -18,12 +18,16 @@ serde = { version = "1", features = ["derive"] }
...
+# wasmer = { version = "2.0.0", optional = true }
+wasmer = { git = "https://github.com/olonho/wasmer", branch = "signalless", optional = true  }
+# wasmer-types = { version = "2.0.0", optional = true }
+wasmer-types = { git = "https://github.com/olonho/wasmer", branch = "signalless", optional = true  }
+# wasmer-compiler-singlepass = { version = "2.0.0", optional = true, default-features = false, features = ["std"] } # disable `rayon` feature.
+wasmer-compiler-singlepass = { git = "https://github.com/olonho/wasmer", branch = "signalless",  optional = true }
+# wasmer-engine-universal = { version = "2.0.0", optional = true }
+wasmer-engine-universal = { git = "https://github.com/olonho/wasmer", branch = "signalless", optional = true }
+# wasmer-vm = "2.0.0"
+wasmer-vm = { git = "https://github.com/olonho/wasmer", branch = "signalless" }
```
ailisp, line 30: Nice! so the magic is there?

olonho: Pretty much, implemented in https://github.com/wasmerio/wasmer/pull/2517/

bowenwang1996:
What is the status of this PR?

ailisp:
@olonho had a great work on a signaless wasmer2 which avoid signal handling confliction. However, more efforts need to have Wasmer team merge the work.

runtime/near-vm-runner/src/vm_kind.rs
matklad: Ouch, turns out bors ignores explicit discrimintants:

```rust
Ouch, so turns out that `borsh` ignores explicit discrimintants: 

```rust
#[derive(Clone, Copy, Debug, Hash, Serialize, Deserialize, BorshSerialize, BorshDeserialize)]
pub enum VMKind {
    /// Wasmer 0.17.x VM.
    Wasmer0 = 0,
    /// Wasmtime VM.
    Wasmtime = 1,
    /// Wasmer 1.x VM.
    Wasmer1 = 92,
}

#[test]
fn smoke() {
    let v = VMKind::Wasmer1;
    let data = v.try_to_vec().unwrap();
    assert_eq!(data, vec![2]); // 2, not 92
}
```

So this doesn't actually presereve serialization fromat. Let's remove the discrimintant then, and either leave an unused `wasmer1` or (if that's possible) just make a serialization breaking change here.

runtime/near-vm-runner/src/wasmer2_runner.rs
```diff
@@ -284,53 +272,70 @@ fn run_method(module: &Module, import: &ImportObject, method_name: &str) -> Resu
...
-        } else {
-            wasmer_engine_native::Native::new(wasmer_compiler_singlepass::Singlepass::default())
-                .engine()
-        };
-        Store::new(&engine)
-    } else {
-        let engine = JIT::new(Singlepass::default()).engine();
-        Store::new(&engine)
-    }
+pub(crate) fn default_wasmer2_store() -> Store {
```
matklad, line 325: though: unrelated to pr, but it's probably better to spell this as `wasmer::Store` to not confuse with the store we already have: https://github.com/near/nearcore/blob/3a5aa9132145d6411f1b9b30abc98a6a7a609ea5/core/store/src/lib.rs#L48-L52

matklad (approved):

olonho (approved):

runtime/near-vm-runner/src/wasmer2_runner.rs
```diff
@@ -284,53 +272,70 @@ fn run_method(module: &Module, import: &ImportObject, method_name: &str) -> Resu
...
+}
+
+#[derive(Hash, PartialEq, Debug)]
+#[allow(unused)]
+enum WasmerCompiler {
+    Singlepass = 1,
+    Cranelift = 2,
+    Llvm = 3,
+}
+
```
bowenwang1996, line 295: Could we add some comments on what these enums are?

runtime/near-vm-runner/src/wasmer2_runner.rs
```diff
@@ -284,53 +272,70 @@ fn run_method(module: &Module, import: &ImportObject, method_name: &str) -> Resu
...
         s.finish()
     }
 }
 
-const WASMER1_CONFIG: Wasmer1Config =
-    Wasmer1Config { seed: 53, use_cranelift: false, use_native_engine: false };
+// We use following scheme for the bits forming seed:
+//  kind << 10, kind is 1 for Wasmer
+//  major version << 6
+//  minor version
```
bowenwang1996, line 314: this feels hacky :(

bowenwang1996 (approved):

runtime/near-vm-runner/src/wasmer2_runner.rs
olonho: Yeah, somewhat.

runtime/near-vm-runner/src/wasmer2_runner.rs
olonho: Sure, will fix that. Generally, just different ways to compile Wasm to native.

PR #4397, opened 2021-06-22
Add test to test infinite loop contracts

### LINKED ISSUE #2224: test: Add a python tests that tests Infinite loops in contracts
Create a test that deploys a contract with an infinite loop, and then sends hundreds of transactions invoking the contract. Make sure that the blocks are still produced, and that the transactions are included.

To see examples of contracts deploying in a pytest see `pytest/tests/sanity/deploy_call_smart_contract.py`.

mzhangmzz:
resolves #2224


bowenwang1996 (approved):
Please add the test to `nightly/complete_nayduck_set.txt`

pytest/tests/contracts/infinite_loops.py
```diff
@@ -0,0 +1,64 @@
...
+sys.path.append('lib')
+from cluster import start_cluster
+from transaction import sign_deploy_contract_tx, sign_function_call_tx
+from utils import load_binary_file, compile_rust_contract
+
+nodes = start_cluster(
+    4, 0, 1, None,
+    [["epoch_length", 10], ["block_producer_kickout_threshold", 80], ["transaction_validity_period", 10000]], {})
+
+wasm_file = compile_rust_contract(''' metadata! {
```
matklad, line 17: Using nearsdk just to make a looping contract is an overkill. A specific issue here is that this test will be slower to run as it'll need to invoke cargo machinery to compile the contract. 

Instead, I suggest adding:

```rust
#[no_mangle]
pub unsafe fn loop_forever() {
    loop {}
}
```

To the canonical test contract: 

https://github.com/near/nearcore/blob/7e24d5d758be41ff12528964ebdcad12e32b990d/runtime/near-test-contracts/test-contract-rs/src/lib.rs#L561-L575

See this example for how to use this contract in a python test

https://github.com/matklad/nearcore/blob/21a32edde770aa3b8e4a14d78ab2a0bb707442be/pytest/tests/contracts/deploy_call_smart_contract.py#L21-L23/

pytest/tests/contracts/infinite_loops.py
```diff
@@ -0,0 +1,64 @@
...
+  #[near_bindgen]
+  #[derive(Default, BorshDeserialize, BorshSerialize)]
+  pub struct InifiniteLoop {}
+}
+
+#[near_bindgen]
+impl InifiniteLoop {
+  pub fn start_loop(&self) {
+    loop {
+      env::log(b"hello world");
```
matklad, line 27: I don't think we should `log` here, as that will make us hit the "maximum number of log messages" limit pretty soon, without burning much CPU. An empty loop seems better here.

nightly/complete_nayduck_set.txt
```diff
@@ -104,6 +104,7 @@ pytest contracts/deploy_call_smart_contract.py
 pytest contracts/deploy_call_smart_contract.py --features nightly_protocol --features nightly_protocol_features
 pytest --timeout=240 contracts/gibberish.py
 pytest --timeout=240 contracts/gibberish.py --features nightly_protocol --features nightly_protocol_features
+pytest --timeout=400 contracts/infinite_loops.py --features nightly_protocol --features nightly_protocol_features
```
bowenwang1996, line 107: Please also add a line without features so that it will run for the stable build as well

pytest/tests/contracts/infinite_loops.py
mzhangmzz: fixed, thanks for the suggestion!

mzhangmzz:
@matklad I addressed all your comments, would you mind taking a look at it again?

matklad (approved):
Looks good to me now, thanks!

PR #4399, opened 2021-06-23
refactor(jsonrpc): Expose structured errors (VError)

### LINKED ISSUE #2976: nearcore doesn't return structured errors for most use cases of API
Structured errors work (https://github.com/nearprotocol/nearcore/issues/1839) is not complete and it's hard to know what to expect.

Example of error from nearcore:

```
    response.error {
      code: -32000,
      message: 'Server error',
      data: 'Other Error: block DtdfWgnt3QDEjoeARUK25vg8qh31kc3Lrg6qREsoJsST is ahead of head block 34eT8dRWfDPhZBiRGYvkHDj65ThaXCBTrnM4LQjn2Zab \n' +
        ' Cause: Unknown \n' +
        ' Backtrace:    0: failure::backtrace::Backtrace::new\n' +
        '   1: <actix::address::envelope::SyncEnvelopeProxy<A,M> as actix::address::envelope::EnvelopeProxy>::handle\n' +
        '   2: <actix::contextimpl::ContextFut<A,C> as core::future::future::Future>::poll\n' +
        '   3: tokio::runtime::task::raw::poll\n' +
        '   4: <core::future::from_generator::GenFuture<T> as core::future::future::Future>::poll\n' +
        '   5: actix_rt::runtime::Runtime::block_on\n' +
        '   6: near::main\n' +
        '   7: std::rt::lang_start_internal::{{closure}}::{{closure}}\n' +
        '             at rustc/a74d1862d4d87a56244958416fd05976c58ca1a8/src/libstd/rt.rs:52\n' +
        '      std::sys_common::backtrace::__rust_begin_short_backtrace\n' +
        '             at rustc/a74d1862d4d87a56244958416fd05976c58ca1a8/src/libstd/sys_common/backtrace.rs:130\n' +
        '   8: main\n' +
        '   9: __libc_start_main\n' +
        '  10: _start\n'
    }
```

Note that even actual error message `Other Error: block DtdfWgnt3QDEjoeARUK25vg8qh31kc3Lrg6qREsoJsST is ahead of head block 34eT8dRWfDPhZBiRGYvkHDj65ThaXCBTrnM4LQjn2Zab` doesn't go into proper place in `message`.

This effectively makes current error message strings part of API, which is very suboptimal.

Here's how we have to handle errors because of this:

*near-api-js*

```
src/account.ts
120:                if (!error.message.match(/Transaction \w+ doesn't exist/)) {
205:            if (e.message.includes('does not exist while viewing')) {
```

```
src/wallet-account.ts
193:                if (e.message.includes('does not have enough balance')) {
```

*near-wallet*

```
src/actions/account.js
78:            if (error.message.indexOf('does not exist while viewing') !== -1) {
```

```
src/reducers/allAccounts.js
16:        if (payload.message.match('does not exist')) {
```

Note that we also handle errors explicitly only in few cases (degrading user experience) because we want to avoid depending on error messages in this way. This problem also degrades quality of every app built on NEAR, as everybody has same problem.

khorolets:
Fixed #2976


chain/jsonrpc-primitives/src/errors.rs
```diff
@@ -59,18 +76,67 @@ impl RpcError {
...
+        RpcError {
+            code: -32_000,
+            message: "Server error".to_owned(),
+            data: Some(Value::String(e.clone())),
+            error_struct: Some(RpcErrorKind::HandlerError(serde_json::json!({
+                "name": "SERIALIZATION_ERROR",
+                "info": serde_json::json!({
+                    "error_message": e
+                })
+            }))),
```
frol, line 108: Let's not introduce weird corner cases. Let's define `InternalError` variant next to `ValidationError` and `HandlerError`. We may also consider to map `RpcErrorKind -> HandlerError -> * -> InternalError` (just a pseudocode) into `RpcErrorKind -> InternalError`.

I also want to bring another idea here: consider also merging `Rpc*::Unreachable` into `Rpc*::InternalError` (on the RPC error types, not the view-client error types), so end-users don't need to handle two similar types and rather handle Internal Errors as something being wrong on the server side and the only thing client can do is to retry.

chain/jsonrpc-primitives/src/errors.rs
```diff
@@ -34,7 +51,7 @@ impl RpcError {
     ///
     /// Mostly for completeness, doesn't do anything but filling in the corresponding fields.
     pub fn new(code: i64, message: String, data: Option<Value>) -> Self {
-        RpcError { code, message, data }
+        RpcError { code, message, data, error_struct: None }
```
frol, line 54: Are there any users of this `new` method? Can we drop it? That would probably mean that we can also drop `Option` around `error_struct`

chain/jsonrpc-primitives/src/errors.rs
```diff
@@ -15,11 +15,28 @@ pub struct RpcParseError(pub String);
 #[serde(deny_unknown_fields)]
 pub struct RpcError {
     pub code: i64,
+    #[serde(flatten)]
+    pub error_struct: Option<RpcErrorKind>,
```
frol, line 19: Let's move it one line above, and also mark all the other fields as deprecated. Consider using [`#[deprecated(...)]` attribute](https://doc.rust-lang.org/edition-guide/rust-2018/the-compiler/an-attribute-for-deprecation.html) or just doc comments.

frol:
@khorolets It might be worthwhile to boot an RPC node (no need for it to be archival) for mainnet with this patch and let @vgrichina and @volovyk-s to play around. WDYT?

volovyk-s:
@frol good idea!

khorolets:
> @khorolets It might be worthwhile to boot an RPC node (no need for it to be archival) for mainnet with this patch and let @vgrichina and @volovyk-s to play around. WDYT?

I guess I can, but it means that it won't be definitely included in 1.20.0 then.

frol:
@khorolets address the current comments, and while we are collecting the code review from other nearcore devs, @volovyk-s will have a chance to sent a few requests. Ask Mario or Sandi to have a synced node for you, so you don’t battle with terraform and just update the binary there

chain/jsonrpc-primitives/src/errors.rs
khorolets: `adversarial` feature uses `RpcError::server_error` a lot and `server_error()` uses `new()`. Don't this this is a perfect time to refactor that.

chain/jsonrpc-primitives/src/errors.rs
```diff
@@ -59,18 +80,84 @@ impl RpcError {
...
+                "info": serde_json::json!({
+                    "error_message": e
+                })
+            }))),
+        }
+    }
+
+    /// Helper method to define extract INTERNAL_ERROR in separate RpcErrorKind
+    /// Returns HANDLER_ERROR if the error is not internal one
+    pub fn new_internal_or_handler_error(error_data: Option<Value>, error_struct: Value) -> Self {
```
frol, line 117: I am not a huge fan of such methods, but at least it is explicit about what it does. Let's go ahead with it for now

chain/jsonrpc-primitives/src/errors.rs
```diff
@@ -59,18 +80,84 @@ impl RpcError {
...
+        if error_struct["name"] == "INTERNAL_ERROR" {
+            Self::new_internal_error(error_data, error_struct["info"].clone())
```
frol, line 119: Consider using [`.take()`](https://docs.serde.rs/serde_json/value/enum.Value.html#method.take) instead of `.clone()`

chain/jsonrpc-primitives/src/errors.rs
```diff
@@ -34,7 +55,7 @@ impl RpcError {
     ///
     /// Mostly for completeness, doesn't do anything but filling in the corresponding fields.
     pub fn new(code: i64, message: String, data: Option<Value>) -> Self {
-        RpcError { code, message, data }
+        RpcError { code, message, data, error_struct: None }
     }
     /// Create an Invalid Param error.
     pub fn invalid_params(data: impl Serialize) -> Self {
```
frol, line 61: Let's update it to return `ValidationError`.

Also, update `server_error` to return `InternalError`

chain/jsonrpc-primitives/src/errors.rs
```diff
@@ -59,18 +80,84 @@ impl RpcError {
     }
     /// Create an invalid request error.
     pub fn invalid_request() -> Self {
```
frol, line 82: It is only used in a single place, and it indicates that the user submitted not a JSON RPC request (`{"jsonrpc": "2.0", "id": "qwe", "method": "asd", "params": {}}`), but something else, so I suggest we just use `ParseError` with a message that JSON RPC Request format was expected.

chain/jsonrpc-primitives/src/errors.rs
```diff
@@ -14,12 +14,33 @@ pub struct RpcParseError(pub String);
...
     pub message: String,
+    /// Deprecated please use the `error_struct` instead
     #[serde(skip_serializing_if = "Option::is_none")]
     pub data: Option<Value>,
 }
 
+#[derive(Debug, Serialize, Deserialize, Clone, PartialEq)]
+#[serde(tag = "name", content = "cause", rename_all = "SCREAMING_SNAKE_CASE")]
+pub enum RpcErrorKind {
+    ValidationError(ValidationErrorKind),
```
frol, line 31: I suggest we consider the following naming to disambiguate the "validation" term.

```suggestion
    RequestValidationError(RpcRequestValidationErrorKind),
```

chain/jsonrpc-primitives/src/errors.rs
```diff
@@ -14,12 +14,33 @@ pub struct RpcParseError(pub String);
...
+    HandlerError(Value),
+    InternalError(Value),
```
frol, line 33: Well, now we have another problem, our structure falls apart and we will have:

```json
{
  "name": "INTERNAL_ERROR",
  "cause": {
    "error_message": "..."
  }
}
```

We probably want:

```json
{
  "name": "INTERNAL_ERROR",
  "cause": {
    "name": "???",
    "info": {
      "error_message": "..."
    }
  }
}
```

, yet the only candidate for the second `"name"` field I have in mind is `INTERNAL_ERROR`.

To be more explicit about the structure we may want to write:

```rust
    InternalError {
        name: String,
        cause: { // This is not a valid Rust syntax, though
            error_message: String 
        }
    },
```

or

```rust
    InternalError(RpcInternalErrorKind),
```

chain/jsonrpc-primitives/src/errors.rs
khorolets: Arguing about mutability in this case, I'll leave `.clone()`

chain/jsonrpc-primitives/src/errors.rs
```diff
@@ -49,28 +72,86 @@ impl RpcError {
...
+        error_data: Option<Value>,
+        validatation_error_kind: RpcRequestValidationErrorKind,
+    ) -> Self {
+        RpcError {
+            code: -32_000,
+            message: "Server error".to_owned(),
+            data: error_data,
+            error_struct: Some(RpcErrorKind::RequestValidationError(validatation_error_kind)),
+        }
+    }
```
frol, line 143: It seems that this method is not used

chain/jsonrpc-primitives/src/types/blocks.rs
```diff
@@ -81,18 +82,27 @@ impl From<actix::MailboxError> for RpcBlockError {
...
             }
         };
 
-        Self::new(-32_000, "Server error".to_string(), error_data)
+        let error_data_value = match serde_json::to_value(error) {
+            Ok(value) => value,
+            Err(_err) => {
+                return Self::new_internal_error(
+                    None,
+                    "Failed to serialize RpcBlockError".to_string(),
```
frol, line 100: It would be helpful to include the `err` in the message for debugging purposes.

chain/jsonrpc-primitives/src/types/blocks.rs
```diff
@@ -61,13 +62,13 @@ impl From<near_client_primitives::types::GetBlockError> for RpcBlockError {
             near_client_primitives::types::GetBlockError::IOError { error_message } => {
                 Self::InternalError { error_message }
             }
-            near_client_primitives::types::GetBlockError::Unreachable { error_message } => {
+            near_client_primitives::types::GetBlockError::Unreachable { ref error_message } => {
```
frol, line 65: Why? I mean, it won't hurt, but I think we don't need it, so I wonder about the reason to be explicit here

frol (approved):
There are a few nits, but I believe all my major concerns are addressed.

chain/jsonrpc-primitives/src/errors.rs
```diff
@@ -14,12 +14,33 @@ pub struct RpcParseError(pub String);
...
+    RequestValidationError(RpcRequestValidationErrorKind),
+    HandlerError(Value),
+    InternalError(Value),
+}
+
+#[derive(Debug, Serialize, Deserialize, Clone, PartialEq)]
+#[serde(tag = "name", content = "info", rename_all = "SCREAMING_SNAKE_CASE")]
+pub enum RpcRequestValidationErrorKind {
+    MethodNotFound { method_name: String },
+    InvalidRequest,
```
frol, line 40: This variant is not used anymore (we replaced it with ParseError)

chain/jsonrpc-primitives/src/errors.rs
```diff
@@ -49,28 +72,86 @@ impl RpcError {
...
     pub fn serialization_error(e: String) -> Self {
-        RpcError::new(-32_000, "Server error".to_owned(), Some(Value::String(e)))
+        RpcError::new_internal_error(Some(Value::String(e.clone())), e)
+    }
+
+    /// Helper method to define extract INTERNAL_ERROR in separate RpcErrorKind
+    /// Returns HANDLER_ERROR if the error is not internal one
+    pub fn new_internal_or_handler_error(error_data: Option<Value>, error_struct: Value) -> Self {
+        if error_struct["name"] == "INTERNAL_ERROR" {
+            Self::new_internal_error(error_data, error_struct["info"]["error_message"].to_string())
```
frol, line 106: I believe `to_string` will produce unnecessary quotes like `"error_message": "\"some message\""`. Use `.get()` instead of `[]` and `as_str()` instead of `to_string()`, and handle potential None values and gracefully return a new InternalError with a message that InternalError happened during serializing InternalError :shrug:

chain/jsonrpc-primitives/src/types/blocks.rs
khorolets: "value partially moved here"

frol: I see! All good

khorolets:
The test rpc node (non-archival) with the binary from this PR is running

`http://35.204.24.164:3030/`

/cc @vgrichina @volovyk-s

frol:
@khorolets There is a missing structured error (please, add a test once you resolve it):

```
curl http://35.204.24.164:3030/ -H 'Content-Type: application/json' --data '{"jsonrpc": "2.0", "method": "query", "params": {"request_type": "view_account", "account_id": "test.near", "block_id": 10}, "id": "dontcare"}' | jq

{
  "jsonrpc": "2.0",
  "error": {
    "code": -32000,
    "message": "Server error",
    "data": "DB Not Found Error: BLOCK HEIGHT: 10 \n Cause: Unknown"
  },
  "id": "dontcare"
}
```

(same goes about any `query` method when you specify `block_id` that does not exist)

frol:
I am testing the node now, so I wanted to share some examples with those of you who don't want to craft the queries:

```
$ curl http://35.204.24.164:3030/ -H 'Content-Type: application/json' --data '{"jsonrpc": "2.0", "method": "query", "params": {"request_type": "call_function", "account_id": "tipbot.app.near", "finality": "finalwe", "method_name": "get_deposit", "args_base64": "eyJhY2NvdW50X2lkIjogImZyb2wubmVhciJ9"}, "id": "dontcare"}' | jq

{
  "jsonrpc": "2.0",
  "error": {
    "name": "REQUEST_VALIDATION_ERROR",
    "cause": {
      "name": "PARSE_ERROR",
      "info": {
        "error_message": "Failed parsing args: unknown variant `finalwe`, expected one of `optimistic`, `near-final`, `final`"
      }
    },
    "code": -32700,
    "message": "Parse error",
    "data": "Failed parsing args: unknown variant `finalwe`, expected one of `optimistic`, `near-final`, `final`"
  },
  "id": "dontcare"
}
```

```
$ curl http://35.204.24.164:3030/ -H 'Content-Type: application/json' --data '{"jsonrpc": "2.0", "method": "query", "params": {"request_type": "call_function", "account_id": "tipbot.app.near", "block_id": 34669523, "method_name": "get_deposit", "args_base64": "eyJhY2NvdW50X2lkIjogImZyb2wubmVhciJ9"}, "id": "dontcare"}' | jq

{
  "jsonrpc": "2.0",
  "error": {
    "name": "HANDLER_ERROR",
    "cause": {
      "info": {
        "requested_shard_id": 0
      },
      "name": "UNAVAILABLE_SHARD"
    },
    "code": -32000,
    "message": "Server error",
    "data": "The node does not track the shard ID 0"
  },
  "id": "dontcare"
}
```

(NOTE: it is a weird error for this case when we actually query the garbage-collected data which is not available anymore, but receive a misleading error, but that was the case before as well, so it should be tracked separately)

```
$ http post http://35.204.24.164:3030/ jsonrpc=2.0 method=tx params:='["56t8ceK2B8BympLJ6gPL3CYsdPMJiWSegsCEFsAhcAj7", "near"]' id=dontcare

{
    "error": {
        "name": "HANDLER_ERROR",
        "cause": {
            "info": {
                "requested_transaction_hash": "56t8ceK2B8BympLJ6gPL3CYsdPMJiWSegsCEFsAhcAj7"
            },
            "name": "UNKNOWN_TRANSACTION"
        },
        "code": -32000,
        "data": "Transaction 56t8ceK2B8BympLJ6gPL3CYsdPMJiWSegsCEFsAhcAj7 doesn't exist",
        "message": "Server error"
    },
    "id": "dontcare",
    "jsonrpc": "2.0"
}
```

(this one is garbage collected, and so far it is the best error the node can give)

```
$ http post http://35.204.24.164:3030/ jsonrpc=2.0 method=tx params:='["56t8ceK2B8BympLJ6gPL3CYsdPMJiWSegsCEFsAhcAj7", "!W@near"]' id=dontcare

{
    "error": {
        "name": "REQUEST_VALIDATION_ERROR",
        "cause": {
            "info": {
                "error_message": "Invalid account id: !W@near"
            },
            "name": "PARSE_ERROR"
        },
        "code": -32700,
        "data": "Invalid account id: !W@near",
        "message": "Parse error"
    },
    "id": "dontcare",
    "jsonrpc": "2.0"
}
```

```
http http://35.204.24.164:3030/ jsonrpc=2.0 id=dontcare method=block params:='{"block_id": "BEzouvv6xJ9FWkGGgKN98bqEUde3ZKJx51KiWUZbtdQU"}'
HTTP/1.1 200 OK
content-length: 247
content-type: application/json
date: Fri, 25 Jun 2021 16:32:02 GMT

{
    "error": {
        "name": "HANDLER_ERROR",
        "cause": {
            "info": {},
            "name": "UNKNOWN_BLOCK"
        },
        "code": -32000,
        "data": "DB Not Found Error: BLOCK: BEzouvv6xJ9FWkGGgKN98bqEUde3ZKJx51KiWUZbtdQU \n Cause: Unknown",
        "message": "Server error"
    },
    "id": "dontcare",
    "jsonrpc": "2.0"
}
```

(we would love to provide more information here like which block was requested, but currently, we don't have it implemented)

frol:
@khorolets Let's `#[serde(skip)]` the `context`, so we don't block the release and have more time to think through those transaction execution errors:

```
$ http post http://35.204.24.164:3030/ jsonrpc=2.0 id=dontcare method=broadcast_tx_commit 'params:=["CQAAAGZyb2wubmVhcgBe7rr+mXVb7dVoP+rBXtfA1E9WPok+Zrfs5o2lUFYb3QAAAAAAAAAADgAAAGtob3JvbGV0cy5uZWFy+WQ5DxNLDlRnkZbWdRJr9qUj8tIwd3p4xL0pm8ZvBzABAAAAAwAAAKHtzM4bwtMAAAAAAAAASyivpnygqqH3Ue8fJXqvYhE8iTW9q/ElzUSw17vgW+h4G+ZrWD25Sco98ENxpO3rgMbN7ksvXKO1cQDgLELKBA=="]'

{
    "error": {
        "cause": {
            "info": {
                "context": {
                    "InvalidNonce": {
                        "ak_nonce": 0,
                        "tx_nonce": 0
                    }
                }
            },
            "name": "INVALID_TRANSACTION"
        },
        "code": -32000,
        "data": {
            "TxExecutionError": {
                "InvalidTxError": {
                    "InvalidNonce": {
                        "ak_nonce": 0,
                        "tx_nonce": 0
                    }
                }
            }
        },
        "message": "Server error",
        "name": "HANDLER_ERROR"
    },
    "id": "dontcare",
    "jsonrpc": "2.0"
}
```

khorolets:
> @khorolets There is a missing structured error (please, add a test once you resolve it):
> 
> ```
> curl http://35.204.24.164:3030/ -H 'Content-Type: application/json' --data '{"jsonrpc": "2.0", "method": "query", "params": {"request_type": "view_account", "account_id": "test.near", "block_id": 10}, "id": "dontcare"}' | jq
> 
> {
>   "jsonrpc": "2.0",
>   "error": {
>     "code": -32000,
>     "message": "Server error",
>     "data": "DB Not Found Error: BLOCK HEIGHT: 10 \n Cause: Unknown"
>   },
>   "id": "dontcare"
> }
> ```
> 
> (same goes about any `query` method when you specify `block_id` that does not exist)

Fixed

chain/jsonrpc-primitives/src/types/transactions.rs
```diff
@@ -20,10 +20,14 @@ pub enum TransactionInfo {
...
 }
 
-#[derive(thiserror::Error, Debug)]
+#[derive(thiserror::Error, Debug, Serialize)]
+#[serde(tag = "name", content = "info", rename_all = "SCREAMING_SNAKE_CASE")]
 pub enum RpcTransactionError {
     #[error("An error happened during transaction execution: {context:?}")]
-    InvalidTransaction { context: near_primitives::errors::InvalidTxError },
+    InvalidTransaction {
+        #[serde(skip)]
```
miraclx, line 28: Any reason why we currently `skip` this field? I'm looking to implement `Deserialize` for this struct.

miraclx: Resolved with https://github.com/near/nearcore/pull/4770/files#diff-b0dd5d6f8f4d2a7b135860666232d02f02a60ee1dd9d7fd65ee6afe79550e98eR30

PR #4402, opened 2021-06-23
pytest: Move hello.wasm to location where pytest tests can find it

mina86:
Commit [7c1080c1: fixed failed CI tests due to features mismatches]
moved the hello.wasm file around without updating pytest tests which
read that file.  This caused some of the tests to start failing.

Move the file back to pytest directory, specifically to
pytest/testdata.

Since none of the integration tests refer to that file it’s safe to
move it out of the integration-test directory.

This change fixes the following three tests:
* sanity/rpc_light_client_execution_outcome_proof.py,
* sanity/rpc_state_changes.py and
* sanity/rpc_tx_status.py.


matklad:
I wonder if hello.wasm blob makes sense? We have a couple of canonical contracts (which are alerady build from source by the testing infra) here: 

https://github.com/near/nearcore/tree/master/runtime/near-test-contracts

I think it makes sense to rewrite pytests currently using hello.wasm to use `test_contract_rs.wasm` to reduce the number of blobs in the repo? See this example:

https://github.com/near/nearcore/blob/7e24d5d758be41ff12528964ebdcad12e32b990d/pytest/tests/contracts/deploy_call_smart_contract.py#L19-L23

mina86:
Makes sense. Do you suggest than dropping this change and waiting for the rewrite? I certainly can take a look at changing the tests but that might take several days.

matklad:
I think the change is valuable as is!

bowenwang1996 (approved):
@matklad do you mind creating an issue for what you suggested?

chefsale (approved):
LGTM

PR #4405, opened 2021-06-23
count refund receipts toward gas limit calculation

mzhangmzz:


runtime/runtime/src/lib.rs
```diff
@@ -1602,6 +1617,55 @@ mod tests {
...
+                    &epoch_info_provider,
+                    None,
+                )
+                .unwrap();
+            let (store_update, new_root) = tries.apply_all(&apply_result.trie_changes, 0).unwrap();
+            root = new_root;
+            store_update.commit().unwrap();
+            let state = tries.new_trie_update(0, root);
+            let account = get_account(&state, &alice_account()).unwrap().unwrap();
+            // Check that refund receipts are processed relayed if CountRefundReceiptsInGasLimit is enabled,
```
bowenwang1996, line 1652:
```suggestion
            // Check that refund receipts are delayed if CountRefundReceiptsInGasLimit is enabled,
```
Not sure what you intended to say here :(

bowenwang1996 (commented):
Looks good overall. I think the test might be better written as an integration test (see `process_blocks.rs`)

runtime/runtime/src/lib.rs
```diff
@@ -1224,7 +1239,7 @@ impl Runtime {
                 outgoing_receipts.push(receipt);
             }
 
-            total_gas_burnt += outcome_with_id.outcome.gas_burnt;
+            total_gas_limit_used += outcome_with_id.outcome.gas_burnt;
```
EgorKulikov, line 1242: it seems a bit weird to add bunt gas to used gas here and elsewhere, especially considering those are 2 different fields of ActionResult

mzhangmzz: The variable name is to keep track of total gas burnt for the purpose of gas limit, it has nothing to do with gas_used. I agree that this implementation will make the code a little confusing. That's because gas_burnt is used for two purposes, one to keep account the real gas burnt to charge the signer, one for keeping track of gas limit, and these two concepts are different for refund receipts, but they are stored in one field `gas_burnt`. Another solution is to add another field in ActionResult and ExecutionOutcome specifically for the use of gas limit tracking. I thought that might be overkill and that ExecutionOutcome is stored as part of Chain state which I am not super familiar with, so I don't know what effect that will bring there.

runtime/runtime/src/lib.rs
mzhangmzz: Yep that is what I meant, sorry for the typo

runtime/runtime/src/lib.rs
EgorKulikov: I think maybe rename leaving variable name as it was (total_gas_burnt) or use some other name (like total_execution_gas) to disambiguate with result.gas_used

EgorKulikov:
As to PR in general - lgtm, but I agree with Bowen that integration test would be in order

chain/client/tests/process_blocks.rs
```diff
@@ -2289,6 +2291,117 @@ fn test_block_execution_outcomes() {
...
+        env.produce_block(0, block_height);
+        let block = env.clients[0].chain.get_block_by_height(block_height).unwrap().clone();
+        let prev_block = env.clients[0].chain.get_block_by_height(block_height-1).unwrap().clone();
+        let chunk_extra = env.clients[0].chain.get_chunk_extra(prev_block.hash(), 0).unwrap().clone();
+        let state_update =
+            env.clients[0].runtime_adapter.get_tries().new_trie_update(0, *chunk_extra.state_root());
+        let delayed_indices =
+            get::<DelayedReceiptIndices>(&state_update, &TrieKey::DelayedReceiptIndices)
+                .unwrap();
+        let finished_all_delayed_receipts = match delayed_indices{
```
bowenwang1996, line 2339: how did this pass the fmt checker :)

chain/client/tests/process_blocks.rs
```diff
@@ -2289,6 +2291,117 @@ fn test_block_execution_outcomes() {
...
+                println!("delayed_indices {}-{}", delayed_indices.first_index, delayed_indices.next_available_index);
+                delayed_indices.next_available_index>0 && delayed_indices.first_index == delayed_indices.next_available_index
+            },
+        };
+        let chunk = env.clients[0].chain.get_chunk(&block.chunks()[0].chunk_hash()).unwrap().clone();
+        println!("block {} receipts {:#?}", block_height, chunk.receipts());
+        if chunk.receipts().len() == 0 && chunk.transactions().len() == 0 && finished_all_delayed_receipts {
+            break;
+        }
+        block_height+=1;
```
bowenwang1996, line 2351: Also here...

chain/client/tests/process_blocks.rs
```diff
@@ -2289,6 +2291,117 @@ fn test_block_execution_outcomes() {
...
+                .unwrap();
+        let finished_all_delayed_receipts = match delayed_indices{
+            None => false,
+            Some(delayed_indices) => {
+                println!("delayed_indices {}-{}", delayed_indices.first_index, delayed_indices.next_available_index);
+                delayed_indices.next_available_index>0 && delayed_indices.first_index == delayed_indices.next_available_index
+            },
+        };
+        let chunk = env.clients[0].chain.get_chunk(&block.chunks()[0].chunk_hash()).unwrap().clone();
+        println!("block {} receipts {:#?}", block_height, chunk.receipts());
```
bowenwang1996, line 2347: maybe it is a good idea to remove this print. It looks to me that a lot of stuff get printed

chain/client/tests/process_blocks.rs
```diff
@@ -2289,6 +2291,117 @@ fn test_block_execution_outcomes() {
...
+    for (_, id) in tx_hashes.into_iter().enumerate() {
+        let execution_outcome = env.clients[0].chain.get_execution_outcome(&id).unwrap();
+        assert_eq!(execution_outcome.outcome_with_id.outcome.receipt_ids.len(), 1);
+        assert!(match execution_outcome.outcome_with_id.outcome.status {
+            ExecutionStatus::SuccessReceiptId(id) => {
+                let receipt_outcome = env.clients[0].chain.get_execution_outcome(&id).unwrap();
+                assert!(match receipt_outcome.outcome_with_id.outcome.status {
+                    ExecutionStatus::Failure(TxExecutionError::ActionError(_)) => true,
+                    _ => false,
+                });
```
bowenwang1996, line 2364:
```suggestion
                assert!(matches!(receipt_outcome.outcome_with_id.outcome.status, ExecutionStatus::Failure(TxExecutionError::ActionError(_)));
```

chain/client/tests/process_blocks.rs
```diff
@@ -2289,6 +2291,117 @@ fn test_block_execution_outcomes() {
...
+            break;
+        }
+        block_height+=1;
+    }
+
+    let mut refund_receipt_ids = HashSet::new();
+    for (_, id) in tx_hashes.into_iter().enumerate() {
+        let execution_outcome = env.clients[0].chain.get_execution_outcome(&id).unwrap();
+        assert_eq!(execution_outcome.outcome_with_id.outcome.receipt_ids.len(), 1);
+        assert!(match execution_outcome.outcome_with_id.outcome.status {
```
bowenwang1996, line 2358: this makes code a bit harder to read. You might want to simply panic if it is some other variant

bowenwang1996 (approved):
Looks good! Just a few small issues

EgorKulikov (approved):

bowenwang1996:
@mzhangmzz looks like there are some conflicts you need to resolve

Issue #2780, opened 2020-06-03
Currently most of our Python code base is using print instead of proper logging


chefsale:
Currently most of our Python code base is using print instead of proper logging, we should change that.

cc: @ailisp @bowenwang1996

ailisp:
Good point, logging gives us error and verbose level. timing is probably more important in our tests

bowenwang1996:
Good point.

mikedotexe:
hey @bowenwang1996 it seems that Nayduck is wanting Phuong to have access to the `nearprotocol` GitHub org. Can we give this access to him and/or give him access to Azure in order to change this setting?

bowenwang1996:
@mikedotexe is this resolved?

ChaoticTempest:
@bowenwang1996 nayduck is working for me. Thanks for the help!

PR #4414, opened 2021-06-25
Switch over python print calls for logging calls

ChaoticTempest:
Addresses #2780. This swaps `print` statements into `logger.info` statements. Note that I also switched over the print statements for python scripts inside the `scripts/` folder. This required a bit of some pathing magic to import the logger from the `pytests` folder since the python files are spread across the whole project without `__init__.py` for proper modules.


pytest/lib/mocknet.py
```diff
@@ -83,7 +85,7 @@ def list_validators(node):
 
 def setup_python_environment(node, wasm_contract):
     m = node.machine
-    print(f'INFO: Setting up python environment on {m.name}')
+    logger.info(f'INFO: Setting up python environment on {m.name}')
```
ailisp, line 88: I guess `INFO` can be removed since logger.info will tell?

pytest/lib/cluster.py
```diff
@@ -206,7 +208,7 @@ def validators(self):
                 self.get_status()['validators']))
 
     def stop_checking_store(self):
-        print("WARN: Stopping checking Storage for inconsistency for %s:%s" %
+        logger.info("WARN: Stopping checking Storage for inconsistency for %s:%s" %
```
ailisp, line 211: Also this, use `logger.warn("Stopping checking Storage ...")` instead?

pytest/lib/cluster.py
```diff
@@ -218,7 +220,7 @@ def check_store(self):
                 pass
             else:
                 if res['result'] == 0:
-                    print(
+                    logger.info(
```
ailisp, line 223: logger.error

scripts/state/update_res.py
```diff
@@ -9,14 +9,18 @@
...
+
 
 def main():
     if len(sys.argv) == 1:
         update_res()
     elif len(sys.argv) == 2 and sys.argv[1] == 'check':
         check_res()
     else:
-        print('Usage: update-res.py | update-res.py check')
+        logger.info('Usage: update-res.py | update-res.py check')
```
ailisp, line 23: This probably should use `print` or print to stderr

ailisp (approved):
Overall looks good, some logging can be improved

bowenwang1996 (commented):
Please run nayduck to make sure that this change does not accidentally break anything

ChaoticTempest:
So, I changed some strings from using `%s` to an f-string instead, but that throws invalid syntax since the script itself is using python2 via `#!/usr/bin/env python`. But some other scripts use python3. The mixing of python versions will be a potential issue down the line. Do you guys know of any problems that would arise from switching all of scripts over to `#!/usr/bin/env python3`?

pytest/lib/mocknet.py
ChaoticTempest: Currently, I don't have the logger formatted to show the what the log level is at. I can added it if necessary, but would also end up being added to every other log as well that didn't have `INFO:` prefixed before

chefsale:
Please move them all to Python 3.

pytest/lib/branches.py
```diff
@@ -120,6 +121,6 @@ def prepare_ab_test(other_branch):
         if not os.getenv('NAYDUCK'):
             compile_binary(str(other_branch))
         else:
-            print('RC binary should be downloaded for NayDuck.')
+            logger.info('RC binary should be downloaded for NayDuck.')
```
mina86, line 124: `logging.critical`. After all, we’re aborting the script.

Here and throughout the commit where we log before aborting with a `sys.exit(1)`.

pytest/lib/cluster.py
```diff
@@ -508,10 +510,10 @@ def reset_validator_key(self, new_key):
...
 
 def github_auth():
-    print("Go to the following link in your browser:")
-    print()
-    print("http://nayduck.eastus.cloudapp.azure.com:3000/local_auth")
-    print()
+    logger.info("Go to the following link in your browser:")
+    logger.info("")
+    logger.info("http://nayduck.eastus.cloudapp.azure.com:3000/local_auth")
+    logger.info("")
```
mina86, line 516: Like before, this is an interactive function. I don’t think this should be converted to use logger.

pytest/lib/network.py
```diff
@@ -16,17 +18,17 @@ def init_network_pillager():
...
                 "Make sure the current user has access to it, e.g. by changing the owner:"
             )
-            print("")
-            print(
+            logger.info("")
+            logger.info(
                 "    chown <group>.<user> /sys/fs/cgroup/net_cls/block/net_cls.classid"
             )
-            print("")
+            logger.info("")
```
mina86, line 31: `logger.critical`

pytest/lib/configured_logger.py
```diff
@@ -0,0 +1,44 @@
...
+        handler = logging.StreamHandler(sys.stdout)
+        handler.setLevel(level)
+    log.addHandler(handler)
+    log.propagate = False
+
+    return log
+
+
+# global logger for testing purposes:
+logger = new_logger("test")
```
mina86, line 46: `test_logger` perhaps? Not sure though; maybe `logger` is fine.

pytest/lib/bridge.py
```diff
@@ -175,9 +175,9 @@ def cleanup(self):
                 self.stderr.close()
             self.kill()
         except BaseException:
-            print("Kill %s failed on cleanup!" % (self.__class__.__name__))
+            logger.info("Kill %s failed on cleanup!" % (self.__class__.__name__))
             traceback.print_exc()
```
mina86, line 179: logger.error('Kill %s failed on cleanup', type(self).__name__, exc_info=sys.exc_info())

No need for `traceback.print_exc()` or the two empty lines.

pytest/lib/bridge.py
```diff
@@ -137,11 +137,11 @@ def __init__(self,
         self.custom = dict()
 
 def atexit_cleanup(obj):
-    print("Cleaning %s on script exit" % (obj.__class__.__name__))
+    logger.info("Cleaning %s on script exit" % (obj.__class__.__name__))
```
mina86, line 140: logging does %-formatting so this would be better written as:

    logger.info('Cleaning %s on script exit', type(obj).__name__)

Here and throughout.

pytest/tests/bridge/end2end.py
```diff
@@ -4,9 +4,10 @@
 # If `no_txs_in_parallel`, no txs will be executed in parallel.
 
 import sys, time
+from configured_logger import logger
 
 if len(sys.argv) < 3:
-    print("python end2end.py <eth2near_tx_number> <near2eth_tx_number> [...]")
+    logger.info("python end2end.py <eth2near_tx_number> <near2eth_tx_number> [...]")
```
mina86, line 10: This should stay print.

pytest/lib/test.py
```diff
@@ -0,0 +1,6 @@
+
+
+from configured_logger import new_logger
+
+L = new_logger(outfile="/tmp/something.txt")
+L.info("womp")
```
mina86, line 6: Was this file intended to be included? Same with `welp.txt`?

pytest/tests/mocknet/load_testing.py
```diff
@@ -81,79 +83,79 @@ def check_tps(measurement, expected_in, expected_out=None, tolarance=0.05):
...
     time.sleep(5)
-    print('INFO: All transaction types results:')
+    logger.info('INFO: All transaction types results:')
     all_tx_measurement = measure_tps_bps(nodes)
     test_passed = (all_tx_measurement['bps'] > 0.5) and test_passed
     test_passed = check_memory_usage(nodes[0]) and test_passed
     test_passed = check_slow_blocks(all_tx_initial_metrics,
                                     final_metrics) and test_passed
-    print('')
+    logger.info('')
```
mina86, line 154: I think we should be dropping empty log lines.

pytest/lib/cluster.py
```diff
@@ -361,9 +363,9 @@ def cleanup(self):
         try:
             self.kill()
         except:
-            print("Kill failed on cleanup!")
+            logger.info("Kill failed on cleanup!")
             traceback.print_exc()
-            print("\n\n")
+            logger.info("\n\n")
```
mina86, line 368: logger.critical('Kill failed on cleanup!', exc_info=sys.exc_info())

should do the trick, no?

pytest/tests/sanity/lightclnt.py
```diff
@@ -96,7 +97,7 @@ def get_up_to(from_, to):
 
     for i in range(from_, to + 1):
         hash_ = height_to_hash[i]
-        print(i, hash_, hash_to_epoch[hash_], hash_to_next_epoch[hash_])
+        logger.info(i, hash_, hash_to_epoch[hash_], hash_to_next_epoch[hash_])
```
mina86, line 100: This needs to be changed to use format string. Here and below in this file.

pytest/tests/sanity/proxy_simple.py
```diff
@@ -19,9 +20,9 @@ class Handler(ProxyHandler):
     async def handle(self, msg, fr, to):
         if msg.enum == 'Block':
             h = msg.Block.BlockV2.header.inner_lite().height
-            print("Height:", h)
+            logger.info("Height: {h}")
```
mina86, line 23: You’ve missed `f`.

mina86:
> Note that I also switched over the print statements for python scripts inside the scripts/ folder.

I’d argue those should be left alone. They are interactive tools after all.

> So, I changed some strings from using %s to an f-string instead,

Wouldn’t %-formatting be more idiomatic for logging though? I mean, logging performs the formatting if it’s called with multiple arguments as in `logging.info('foo %s baz', 'bar')`.

pytest/lib/cluster.py
chefsale: As mentioned below by @mina86 , we can also drop the `logger.info("")`

nightly/nayduck.py
```diff
@@ -69,10 +73,10 @@ def get_tests(fl):
...
 
 def github_auth():
-    print("Go to the following link in your browser:")
-    print()
-    print("http://nayduck.eastus.cloudapp.azure.com:3000/local_auth")
-    print()
+    logger.info(
+        "Go to the following link in your browser:\n"
+        "http://nayduck.eastus.cloudapp.azure.com:3000/local_auth\n"
+    )
```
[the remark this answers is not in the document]
ChaoticTempest: I'll just revert this file then

pytest/lib/test.py
ChaoticTempest: Good catch. Accidentally committed these

pytest/tests/sanity/lightclnt.py
ChaoticTempest: Wow, I thought I got all these with my regex matching, but guess not

mina86: https://xkcd.com/1171/

ChaoticTempest:
> I’d argue those should be left alone. They are interactive tools after all.

Good point. I just ported them over since they were the same when I was just outputting them to stdout, but they also include timestamps now.

> Wouldn’t %-formatting be more idiomatic for logging though? I mean, logging performs the formatting if it’s called with multiple arguments as in `logging.info('foo %s baz', 'bar')`.

I don't think it's idiomatic though, since it's still using `msg % args` underneath which still has some gotchas. Quote from the python docs:
> The formatting operations described here exhibit a variety of quirks that lead to a number of common errors (such as failing to display tuples and dictionaries correctly). Using the newer formatted string literals, the str.format() interface, or template strings may help avoid these errors.

pytest/lib/bridge.py
ChaoticTempest: I think it'd be consistent to just change these all over to just f-strings and not have all the gotchas I mentioned, but was trying not to get this ticket out of scope by doing that. Since this ticket was to just change over to logging

ChaoticTempest:
@chefsale @mina86 So what do you guys think. Should I just revert the `scripts/` folder back to using `print` since its meant to be interactive, but the logging still outputs to stdout either ways. Or we can cherrypick the most interactive pieces and just have them be in print calls instead

chefsale:
I think it is fine to revert the scripts and leave the rest. We eitherways have to delete some of the scriots there as they are obsolete. I will take a look at that onxe we land this one.

ChaoticTempest:
@chefsale Alright sounds good. It's reverted

chefsale (approved):
This is an improvement, so LGTM.

Issue #4408, opened 2021-06-24
Use standard test contract in pytest


matklad:
We have a directory with standard test contracts we use to test nearcore: 

https://github.com/near/nearcore/tree/master/runtime/near-test-contracts

It contains pre-compiled wasm blob for assembly-script contract and a rust contract which is build automatically during testing. 

However, we currently use another blob, `hello.wasm` in some of our pytest (https://github.com/near/nearcore/pull/4402/commits/f553303924df9d2a51d4685317c86007ea25eabb contains a list of those).

We should move pytest to use the standard rust contract, and remove hello.wasm from the repo. See, for example, this pytest which already uses the standard contract: 

https://github.com/near/nearcore/blob/7e24d5d758be41ff12528964ebdcad12e32b990d/pytest/tests/contracts/deploy_call_smart_contract.py#L19-L23

Strech goal -- grep the repo for other refernces to `.wasm` files and figure out if they can be replaced with standard test contracts.

PR #4416, opened 2021-06-25
tests: remove unused WASM blobs

mina86:
The add.wasm and singlepass_crash.wasm files are not used by anything
and thus can be safely deleted.

Issue: https://github.com/near/nearcore/issues/4408


mina86:
Let’s wait with this till I get rid of hello.wasm usage as well so all three can be deleted.

PR #4418, opened 2021-06-26
pytest: migrate away from using WASM blobs

mina86:
Migrate all pytest tests which used hello.wasm contract (and thus
relied on a blob stored in the repository) to use test-contract-rs
instead (which is stored as source and built).

With that done, remove all WASM blobs stored in the tree.  This
includes aforementioned hello.wasm as well as add.wasm and
singlepass_crash.wasm files.

Fixes: https://github.com/near/nearcore/issues/4408


bowenwang1996 (commented):
I believe singlepass_crash.wasm is needed in one of the tests. cc @matklad @olonho

mina86:
Nope. It’s only user was removed in https://github.com/near/nearcore/commit/3978e027c2ab5e8606cd1c920e1bddce5c6837de

bowenwang1996 (approved):
Thanks!

PR #4424, opened 2021-06-28
Get rid of internal call in `EncodedShardChunk`

Longarithm:
Get rid of internal `Self::from_parts_and_metadata` passing almost the same set of arguments, which should improve readability


core/primitives/src/sharding.rs
```diff
@@ -1128,7 +1110,7 @@ impl EncodedShardChunk {
         #[cfg(feature = "protocol_feature_block_header_v3")]
         let block_header_v3_version = Some(ProtocolFeature::BlockHeaderV3.protocol_version());
 
-        if protocol_version < SHARD_CHUNK_HEADER_UPGRADE_VERSION {
+        let (new_chunk, merkle_paths) = if protocol_version < SHARD_CHUNK_HEADER_UPGRADE_VERSION {
```
bowenwang1996, line 1113: It seems that you pattern match here only to return the same value. What is the point of doing this?

Longarithm: Yeah, I missed it while moving this part of code.

bowenwang1996 (approved):

PR #4426, opened 2021-06-29
fix(test): test vm standalone script test with wasmer1

ailisp:
There used to be a default_wasmer1 feature in near-vm-standalone but it's removed. Add wasmer1 vm to run test in test code

Test Plan
---------
CI


matklad (approved):

matklad:
I wonder how this test's before/after numbers looks for #4448?

Issue #3266, opened 2020-08-31
Ability to shutdown neard cleanly


bowenwang1996:
Neard should shutdown cleanly. More specifically,
* SIGTERM and SIGINT should be enough to shutdown neard. One should not have to resort to SIGKILL to kill neard.
* When neard is killed, the database integrity should not be damaged. Currently we sometimes see DBNotFound errors after stopping and restarting a node.

nearmax:
We need to use this new feature: https://github.com/rust-rocksdb/rust-rocksdb/pull/459
Kudos to @ailisp

frol:
In #4229 we implemented ctrl+c handler for the tests infrastructure, though we clearly did not gracefully shutdown RocksDB, yet it might be still useful to take some inspiration from there.

mina86:
Issue here is that wasmer 0.17.1 is catching SIGINT and interferes with us trying to catch it as well. This was fixed upstream but our fork still does that. All in all, this is now blocked on https://github.com/near/wasmer/pull/38 which fixes this in our fork as well.

bowenwang1996:
cc @matklad

miraclx:
This is the same issue I encountered on #4229, the workaround was more direct by listening for that signal on a dedicated thread and proceeding to inform the tasks that depend on it. https://github.com/near/nearcore/pull/4229/files#diff-c4d3d011b8925d7128b2a5779e866237fa4627a9655c60ff3ce01d7f37d8bdae

mina86:
The process will now gracefully handle SIGINT (i.e. ^C) and SIGTERM (i.e. default signal used by `kill` command).

`cancel_all_background_work` is still waiting for review.

bowenwang1996:
@mina86 looks like this is still not fixed. Today I shutdown a node that is running https://github.com/near/nearcore/commit/3216f6ed1a7e139770cd68d091b65b150719cb04 and tried to load it from state-viewer and got
```
thread 'main' panicked at 'Failed to open the database: DBError(Error { message: "Corruption: Corruption: IO error: No such file or directoryWhile open a file for random read: /home/ubuntu/.near/data/534603.ldb: No such file or directory" })', core/store/src/lib.rs:299:42
```

mina86:
I’m going to close this in favour of https://github.com/near/nearcore/issues/5340

PR #4428, opened 2021-06-29
neard: handle SIGINT by stopping the node

mina86:
Stop the system once a SIGINT is received.  This should allow for graceful
termination since System::stop will stop all the arbiters and that in turn
will stop all the actors (leading them through stopping and stopped states
thus allowing all the necessary cleanups).

Fixes: https://github.com/near/nearcore/issues/3266


mina86:
Actually it doesn’t work reliably since wasmer will install its own SIGINT handler overwriting tokio.

PR #4429, opened 2021-06-29
store: cancel background RocksDB tasks before closing the database

mina86:
Per RocksDB FAQ:

> Q: Is it safe to close RocksDB while another thread is issuing read,
>    write or manual compaction requests?
> A: No.  The users of RocksDB need to make sure all functions have
>    finished before they close RocksDB.  You can speed up the waiting
>    by calling CancelAllBackgroundWork().

Better be safe than sorry so add the call before the rocksdb::DB object
is dropped.

Fixes: https://github.com/near/nearcore/issues/3266


bowenwang1996 (commented):
@mina86 looks like the implementation is for feature `single_thread_rocksdb`, which is not enabled by default.

mina86:
D’oh. So now it turns out we need to update rocksdb to 0.16 to get this to work.

bowenwang1996:
> D’oh. So now it turns out we need to update rocksdb to 0.16 to get this to work.

Let's do that then

mina86:
Yep, already in review queue: https://github.com/near/nearcore/pull/4442

bowenwang1996 (approved):

Issue #4432, opened 2021-06-29
Nayduck regression


bowenwang1996:
Looks like #4418 caused some python tests to regress:
* http://nayduck.eastus.cloudapp.azure.com:3000/#/test/169005
* http://nayduck.eastus.cloudapp.azure.com:3000/#/test/169025

austinabell:
Wasn't aware this infra used the master branch of the SDK. Is there a channel I should notify of these kinds of changes in the future?

bowenwang1996:
Actually I don't think we should depend on the master branch of near-sdk-rs. cc @mina86

mina86:
I’ve been wondering about that. The dependency is only in the tests so perhaps it could be viewed as testing the SDK as well. But yeah, otherwise I can see how that's surprising.

mina86:
The two failures are now fixed, though it still remains an open question whether we want to depend on master branch of near-sdk-rs. I don’t have strong opinions and the change should be trivial.

austinabell:
> The two failures are now fixed, though it still remains an open question whether we want to depend on master branch of near-sdk-rs. I don’t have strong opinions and the change should be trivial.

I don't have an opinion, but it would be very rare for another breaking change like this where this contract would break.

PR #4437, opened 2021-06-29
pytest: make run_handshake handle unsolicited Block messages

mina86:
This addresses the following failure:

      File "tests/spec/network/peers_request.py", line 44, in <module>
        asyncio.run(main())
      File "/usr/lib/python3.7/asyncio/runners.py", line 43, in run
        return loop.run_until_complete(main)
      File "/usr/lib/python3.7/asyncio/base_events.py", line 579, in run_until_complete
        return future.result()
      File "tests/spec/network/peers_request.py", line 35, in main
        await run_handshake(conn1, nodes[0].node_key.pk, key_pair_1, listen_port=12346)
      File "lib/peer.py", line 182, in run_handshake
        assert response.enum == 'Handshake', response.enum if response.enum != 'HandshakeFailure' else response.HandshakeFailure[1].enum
    AssertionError: Block

Fixes: https://github.com/near/nearcore/issues/4432


bowenwang1996 (commented):
This is weird. Handshake is the first message we exchange when establishing a connection. Therefore this implies that the test may not intercept the messages correctly. cc @mfornet @pmnoxx

mina86:
I think what happened was that between flagging handshake as finished and sending handshake to the pear, another thread notice the new peer and sent the block. The flow of the handshake is that it first goes to Peer which then sends Consolidate to PeerManager which registers a new peer and responds back to Peer with an accept and only then Peer sends handshake back to the other node.

pytest/lib/peer.py
```diff
@@ -179,6 +179,13 @@ async def run_handshake(conn: Connection,
         await conn.send(handshake)
         response = await conn.recv()
```
mfornet, line 180: Since we are only expecting `Handshake` or `HandshakeFailure` we can specify `expected` field in `conn.recv`

```suggestion
        response = await conn.recv(lambda msg : msg.enum.startswith('Handshake')
```

mfornet: This needs to be done in the **three** calls to `conn.recv` in this function.

mina86: > This needs to be done in the three calls to conn.recv in this function.

In practice the first two will always be HandshakeFailure but sure.

bowenwang1996 (approved):

PR #4438, opened 2021-06-30
 feat(runtime): expose profile data in execution outcome 

ailisp:
Expose profile data in execution outcome. So it's stored in database and available in RPC response where there's ExecutionOutcome and FinalExecutionOutcome. Exposed format is detailed and the most compact form, indexer/explorer will need to format it as their need, for example one possible formatting: https://github.com/near/nearcore/blob/b0945d609aced3d0b32ba204f55c306f36bd199b/core/primitives-core/src/profile.rs#L94

Test Plan
---------
CI

### DIFF
--- a/core/primitives-core/src/profile.rs
+++ b/core/primitives-core/src/profile.rs
@@ -1,6 +1,9 @@
 use std::rc::Rc;
 use std::{cell::Cell, fmt};
 
+use borsh::{BorshDeserialize, BorshSerialize};
+use serde::{Deserialize, Deserializer, Serialize, Serializer};
+
 use crate::config::{ActionCosts, ExtCosts};
 use crate::types::Gas;
 
@@ -8,9 +11,56 @@ use crate::types::Gas;
 type DataArray = [Cell<u64>; ProfileData::LEN];
 
 /// Profile of gas consumption.
-#[derive(Clone)]
+#[derive(Clone, PartialEq, Eq)]
 pub struct ProfileData {
     data: Rc<DataArray>,
+}
+
+impl BorshSerialize for ProfileData {
+    fn serialize<W: std::io::Write>(&self, writer: &mut W) -> std::io::Result<()> {
+        let mut v = vec![];
+        for i in 0..ProfileData::LEN {
+            let slot = &self.data[i];
+            v.push(slot.get());
+        }
+        BorshSerialize::serialize(&v, writer)
+    }
+}
+
+impl BorshDeserialize for ProfileData {
+    fn deserialize(buf: &mut &[u8]) -> std::io::Result<ProfileData> {
+        let v: Vec<u64> = BorshDeserialize::deserialize(buf)?;
+        if v.len() != ProfileData::LEN {
+            return Err(std::io::Error::new(
+                std::io::ErrorKind::InvalidData,
+                "Invalid ProfileData",
+            ));
+        }
+        let profile_data = ProfileData::new();
+        for i in 0..ProfileData::LEN {
+            let slot = &profile_data.data[i];
+            slot.set(v[i]);
+        }
+        Ok(profile_data)
+    }
+}
+
+impl Serialize for ProfileData {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        unimplemented!()
+    }
+}
+
+impl<'de> Deserialize<'de> for ProfileData {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        unimplemented!()
+    }
 }
 
 impl Default for ProfileData {
@@ -87,6 +137,17 @@ impl ProfileData {
     pub fn set_burnt_gas(&self, burnt_gas: u64) {
         let slot = &self.data[0];
         slot.set(burnt_gas)
+    }
+
+    /// Cut up counter values up to now
+    /// get and reset all counter to zero
+    pub fn cut(&self) -> ProfileData {
+        let ret = self.clone();
+        for i in 0..ProfileData::LEN {
+            let slot = &self.data[i];
+            slot.set(0);
+        }
+        ret
     }
 }
 
--- a/core/primitives-core/src/profile.rs
+++ b/core/primitives-core/src/profile.rs
@@ -2,7 +2,7 @@ use std::rc::Rc;
 use std::{cell::Cell, fmt};
 
 use borsh::{BorshDeserialize, BorshSerialize};
-use serde::{Deserialize, Deserializer, Serialize, Serializer};
+use serde::{Deserialize, Serialize};
 
 use crate::config::{ActionCosts, ExtCosts};
 use crate::types::Gas;
@@ -16,52 +16,9 @@ pub struct ProfileData {
     data: Rc<DataArray>,
 }
 
-impl BorshSerialize for ProfileData {
-    fn serialize<W: std::io::Write>(&self, writer: &mut W) -> std::io::Result<()> {
-        let mut v = vec![];
-        for i in 0..ProfileData::LEN {
-            let slot = &self.data[i];
-            v.push(slot.get());
-        }
-        BorshSerialize::serialize(&v, writer)
-    }
-}
-
-impl BorshDeserialize for ProfileData {
-    fn deserialize(buf: &mut &[u8]) -> std::io::Result<ProfileData> {
-        let v: Vec<u64> = BorshDeserialize::deserialize(buf)?;
-        if v.len() != ProfileData::LEN {
-            return Err(std::io::Error::new(
-                std::io::ErrorKind::InvalidData,
-                "Invalid ProfileData",
-            ));
-        }
-        let profile_data = ProfileData::new();
-        for i in 0..ProfileData::LEN {
-            let slot = &profile_data.data[i];
-            slot.set(v[i]);
-        }
-        Ok(profile_data)
-    }
-}
-
-impl Serialize for ProfileData {
-    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
-    where
-        S: Serializer,
-    {
-        unimplemented!()
-    }
-}
-
-impl<'de> Deserialize<'de> for ProfileData {
-    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
-    where
-        D: Deserializer<'de>,
-    {
-        unimplemented!()
-    }
-}
+/// Result of a ProfileData stat
+#[derive(Clone, BorshSerialize, BorshDeserialize, Serialize, Deserialize, PartialEq, Eq, Debug)]
+pub struct ProfileDataResult(pub Vec<u64>);
 
 impl Default for ProfileData {
     fn default() -> ProfileData {
@@ -141,13 +98,15 @@ impl ProfileData {
 
     /// Cut up counter values up to now
     /// get and reset all counter to zero
-    pub fn cut(&self) -> ProfileData {
-        let ret = self.clone();
+    pub fn cut(&self) -> ProfileDataResult {
+        let mut ret = vec![];
         for i in 0..ProfileData::LEN {
             let slot = &self.data[i];
+            let old = slot.get();
+            ret.push(old);
             slot.set(0);
         }
-        ret
+        ProfileDataResult(ret)
     }
 }
 
--- a/core/primitives/src/views.rs
+++ b/core/primitives/src/views.rs
@@ -1019,7 +1019,6 @@ pub struct ExecutionOutcomeView {
     /// Execution status. Contains the result in case of successful execution.
     pub status: ExecutionStatusView,
     /// Execution metadata, versioned
-    #[serde(skip)]
     pub metadata: ExecutionMetadata,
 }
 
--- a/runtime/runtime/src/lib.rs
+++ b/runtime/runtime/src/lib.rs
@@ -264,7 +264,7 @@ impl Runtime {
                         gas_burnt: verification_result.gas_burnt,
                         tokens_burnt: verification_result.burnt_amount,
                         executor_id: transaction.signer_id.clone(),
-                        metadata: ExecutionMetadata::ExecutionMetadataV1,
+                        metadata: ExecutionMetadata::ExecutionMetadataV2(apply_state.profile.cut()),
                     },
                 };
                 Ok((receipt, outcome))
--- a/core/primitives/src/transaction.rs
+++ b/core/primitives/src/transaction.rs
@@ -14,6 +14,7 @@ use crate::logging;
 use crate::merkle::MerklePath;
 use crate::serialize::{base64_format, u128_dec_format, u128_dec_format_compatible};
 use crate::types::{AccountId, Balance, Gas, Nonce};
+use near_primitives_core::profile::ProfileData;
 
 pub type LogEntry = String;
 
@@ -356,7 +357,14 @@ pub struct ExecutionOutcome {
 pub enum ExecutionMetadata {
     // V1: Empty Metadata
     ExecutionMetadataV1,
+
+    // V2: With ProfileData
+    ExecutionMetadataV2(ProfileData),
 }
+
+// Metadata SHOULD NOT be mutated after creation
+unsafe impl std::marker::Send for ExecutionMetadata {}
+unsafe impl std::marker::Sync for ExecutionMetadata {}
 
 impl Default for ExecutionMetadata {
     fn default() -> Self {
--- a/core/primitives/src/transaction.rs
+++ b/core/primitives/src/transaction.rs
@@ -14,7 +14,7 @@ use crate::logging;
 use crate::merkle::MerklePath;
 use crate::serialize::{base64_format, u128_dec_format, u128_dec_format_compatible};
 use crate::types::{AccountId, Balance, Gas, Nonce};
-use near_primitives_core::profile::ProfileData;
+use near_primitives_core::profile::ProfileDataResult;
 
 pub type LogEntry = String;
 
@@ -359,12 +359,8 @@ pub enum ExecutionMetadata {
     ExecutionMetadataV1,
 
     // V2: With ProfileData
-    ExecutionMetadataV2(ProfileData),
+    ExecutionMetadataV2(ProfileDataResult),
 }
-
-// Metadata SHOULD NOT be mutated after creation
-unsafe impl std::marker::Send for ExecutionMetadata {}
-unsafe impl std::marker::Sync for ExecutionMetadata {}
 
 impl Default for ExecutionMetadata {
     fn default() -> Self {


bowenwang1996:
Please address #4450 first

bowenwang1996 (approved):

ailisp:
@frol @khorolets What do you think of the format of execution outcome's profile data. My opinion is raw data is the best as it's compact and can be either further aggregate on explorer side or show the break down as is on explorer side.

khorolets:
@ailisp I'm not sure, would like to see an example of output. However

> My opinion is raw data is the best as it's compact and can be either further aggregate on explorer side or show the break down as is on explorer side.

sounds good to me. Let's wait for @frol opinion.

core/primitives-core/src/profile.rs
```diff
@@ -13,6 +16,10 @@ pub struct ProfileData {
     data: Rc<DataArray>,
 }
 
+/// Result of a ProfileData stat
+#[derive(Clone, BorshSerialize, BorshDeserialize, Serialize, Deserialize, PartialEq, Eq, Debug)]
+pub struct ProfileDataResult(pub Vec<u64>);
```
matklad, line 21: This type is going to become part of the state, so I think it makes sense to spend some effort designing it with proper encapsulation and such. Just exposing the raw `Vec<u64>` feels like it might make changing this later complicated. 

I see the following issues here:

* The name is `??Result`, but this isn't an alias to `std::result::Result`. It's better to resolve Result suffix for result *enums*. Here, I think the better name would be `GasProfile`. 
* The field is public, which makes it inconvenient for clients to use this (as the data is in a very raw form), and for the implementation to evolve (ie, if we decide to use sparse, rather than dense, storage here, we'd have to change every call-site)
* Derived serialization and deserialization representations don't take evolvability into the account. For example, if we add new `ActionCosts` variant, then all `ExtCosts` indices will shift by one, which will change the meaning of data already stored in the database. 

I think we should do the following:

* move this type into a dedeicated module (it *can* be a well-isolated value type, so it makes sense to enforce it)
* rename it to `GasProfile`
* make the impl private and provide accessors. I think we need two assessors here: an `impl fmt::Display for GasProfile`, so that it's easy to debug stuff in the repo. Additionally, we need something like `fn nonzero_costs(&self) -> BTreeMap<Cost, u64>`, where `Cost` is either an `ActionCost` or an `ExtCost`. Logically, the data here is a mapping from cost metrics to their values, and we should present it to the clients as such. Probably the best interface here would look like this:

```rust
impl Cost {
    const ALL: &'static [COST] = &[...];
}

impl Index<Cost> for GasProfile {
    type Output = u64;
}
```
* do *not* derive `Serialize` -- for JSON RPC, I think we want to serialize this just as a map from string keys to numeric values, and that should be doable in the rpc crate using the ALL/Index API. If `serialize` is required here, because the RPC just want to dump ExecutionOutcome to JSON as is, we should implement `Serialize` manually such that the representation is a map with string keys. 
* do implement `BorshSerialize`, and make sure that its forward compatible with addition of new acction and ext costs. In parcitular, the current format where we first store all actions and then all costs is *not* forward compatible, as adding any action invalidates all ext costs.

bowenwang1996: > This type is going to become part of the state

I was under the impression that it was not going to be part of the state. Could you point me to where it is saved to the state @matklad ?

matklad: We are stuffing this into `ExecutionMetadata`, which is a part of `ExecutionOutcome`, which is stored in the `ColTransactionResult` column. I might be using terminology incorrectly: it is a part of the state we store in rocksdb, but it isn't the part of the state which affects merkle root. That is, it's not a part of `PartialExecutionOutcome`. 

Do we have terms for:

* stuff we store in rocks and which doesn't affect merkle tree
* stuff we store in merkle tree?

bowenwang1996: > stuff we store in merkle tree?

That is the blockchain state.

> stuff we store in rocks and which doesn't affect merkle tree

I don't think we have a good term for that.

ailisp: Yeah move cost to a new type and provide proper access through public method is a good design

ailisp: Looks like after drop cell on ProfileData, there's no need for a GasProfile (ProfileDataResult).

ailisp:
@matklad Addressed your points, ptal. a couple of updates:
> move this type into a dedeicated module (it can be a well-isolated value type, so it makes sense to enforce it)
> rename it to GasProfile

No longer need as ProfileData doesn't have cell.

> make the impl private and provide accessors. I think we need two assessors here: an impl fmt::Display for GasProfile, so that it's easy to debug stuff in the repo. 

Also no longer need as fmt::Debug for ProfileData is pretty good for debugging in the repo

> Additionally, we need something like fn nonzero_costs(&self) -> BTreeMap<Cost, u64>, where Cost is either an ActionCost or an ExtCost. Logically, the data here is a mapping from cost metrics to their values, and we should present it to the clients as such. Probably the best interface here would look like this ... 

Good point, added to ProfileData

>  do not derive Serialize -- for JSON RPC, I think we want to serialize this just as a map from string keys to numeric values, and that should be doable in the rpc crate using the ALL/Index API. If serialize is required here, because the RPC just want to dump ExecutionOutcome to JSON as is, we should implement Serialize manually such that the representation is a map with string keys.

Make sense, added

>  do implement BorshSerialize, and make sure that its forward compatible with addition of new acction and ext costs. In parcitular, the current format where we first store all actions and then all costs is not forward compatible, as adding any action invalidates all ext costs.

Good point. I break action costs and ext cost to two vectors to ensure forward compatible.


Other changes & notes:
- `ProfileData::cut`, which was intended for reset ProfileData to zero and produce GasProfile per txn / receipt, is no longer needed and dropped. Instead, per receipt profile comes from ActionResult as designed in https://github.com/near/nearcore/pull/4582
- We do not have per txn profile. When process txn, it produces a VerificationResult but not an ActionResult, also apply_action is not invoked and there is no `action_result.profile` collected for resulting ExecutionOutcome. Therefore I added a TODO, and will address this problem in the next PR.

core/primitives-core/src/profile.rs
```diff
@@ -1,15 +1,25 @@
...
 /// Profile of gas consumption.
-#[derive(Clone, PartialEq, Eq)]
+/// Vecs are used for forward compatible. should only append new costs and never remove old costs
+#[derive(Clone, PartialEq, Eq, BorshSerialize, BorshDeserialize)]
 pub struct ProfileData {
-    data: DataArray,
+    all_gas: u64,
+    ext_costs: Vec<u64>,
+    action_costs: Vec<u64>,
+}
```
matklad, line 17: I don't think that's the best representation for serialization. What this code does is it aligns the stored data with the *current* shape of code we use to access it. The problem is, the code is mutable, while the data lives forever. Serialized representation should be designed relative to the actual data we are storing, not relative to the actual code. 

In this particular case, the serialized repl reflects the fact that we store `all_gas` separately, and that we separte concpets for execution & action costs. The first is clearly just an accident -- we shouldn't be storing `all_gas` in the rocksdb at all, because that is alerady stored spearaetly as a part of `ExecutionOutcome`. 

I also don't think that it makes sense to separate action and execution costs in the database schema -- in the future, we might add a new kind of cost, or we might merge action and execution costs into one enum. 


I suggest using `Vec<u8>` as the borsh serialization -- this representation directly reflects the essential data we are storing -- a bunch of counters.

matklad (changes requested):
Like how we no longer need a separate gas profele! However, the serializad representations need more work I think

core/primitives-core/src/profile.rs
ailisp: > In this particular case, the serialized repl reflects the fact that we store all_gas separately, and that we separte concpets for execution & action costs. The first is clearly just an accident -- we shouldn't be storing all_gas in the rocksdb at all, because that is alerady stored spearaetly as a part of ExecutionOutcome.

But ProfileData::wasm_gas() do need total gas to figure out the wasm gas?

Other points sounds good.

ailisp: I check again, ProfileData::wasm_gas is not that important, we can drop it too. explorer/frontend can figure out that number from action cost and ext cost.

ailisp: I dropped all_gas and methods that refer to it. And also make all costs in one field, does not separate action cost and ext cost

> I suggest using Vec<u8> as the borsh serialization -- this representation directly reflects the essential data we are storing -- a bunch of counters.

Tried, however this make `Index` trait for `ProfileData` doesn't work well. We cannot tell arbitrary action cost is where quickly. I suggest store serialized repl also as `BTreeMap<Cost, u64>`

ailisp: @matklad I keep 2), the minimum changed one. Would like to know if you have a better solution or thought on approach 1) and 3).

runtime/runtime/src/lib.rs
```diff
@@ -270,6 +270,8 @@ impl Runtime {
                         gas_burnt: verification_result.gas_burnt,
                         tokens_burnt: verification_result.burnt_amount,
                         executor_id: transaction.signer_id.clone(),
+                        // TODO: profile data is only counted in apply_action, which only happened at process_receipt
+                        // VerificationResult needs updates to incorporate profile data to support profile data of txns
                         metadata: ExecutionMetadata::ExecutionMetadataV1,
```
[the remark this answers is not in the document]
matklad: Ah, I see it now, converting transaction into receipt burns gas, and, ideally should also have a gas profile. Makes sense, thanks for explaining!

core/primitives/src/views.rs
```diff
@@ -1019,7 +1019,6 @@ pub struct ExecutionOutcomeView {
     /// Execution status. Contains the result in case of successful execution.
     pub status: ExecutionStatusView,
     /// Execution metadata, versioned
-    #[serde(skip)]
     pub metadata: ExecutionMetadata,
```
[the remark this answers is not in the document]
ailisp: Done. Note that the key is `Cost` instead of `String` to serde serialialize. As a result it need to be Vec<(Cost, u64)> instead of BTreeMap as serde-json doesn't accept nested enum as json keys. And we do not need any lookup from the view so a vector of key value pair just works. An example output:
```
"costs": [[{"ExtCost": "base"}, 1323840555], [{"ExtCost": "contract_compile_base"}, 35445963], ...
```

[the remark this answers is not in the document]
frol:
```
[[{"ExtCost": "base"}, 1323840555], [{"ExtCost": "contract_compile_base"}, 35445963]
```

1. This is not a self-explanatory format at all; I am completely lost here
2. Please, avoid using the default externally tagged enum representations since it is quite inconvenient format for the client code. Internally tagged or adjacently tagged representations are much easier to work with. More details about enum representations: https://serde.rs/enum-representations.html
3. Please, don't expose unnamed fields (`[{}, 1323840555]`)

frol (changes requested):
Sorry being late to the conversation, but I still don't see the final picture of how this fits into JSON RPC response. I am fine with any representation you need for the internal needs, but I care about the thing we expose through the API boundary since it is really hard to undo, and we stuck with all sorts of weird outputs.

Ideally, I would prefer to decouple the internal representation from JSON RPC outputs just like @matklad pointed out, since it was already causing us too much of a trouble ranging from unexpected type changes to leaky dependencies across the crates which in theory should not leak outside. Should we consider splitting this PR into: (1) store in the database, (2) expose it through JSON RPC?

@ailisp May I ask you to draft some telling examples of the JSON output you want to expose? I assume it will be part of the `tx` endpoint output, right?

core/primitives-core/src/profile.rs
```diff
@@ -19,154 +21,394 @@ impl Default for ProfileData {
...
+    pub fn nonzero_costs(&self) -> Vec<(Cost, u64)> {
+        let mut data = Vec::new();
+        for i in Cost::ALL {
+            if self[*i] > 0 {
+                data.push((*i, self[*i]));
+            }
+        }
+        data
+    }
+}
```
frol, line 403: This looks unfortunate to me... How often do we do these operations? I would estimate that these indexes would introduce some runtime and code size penalty due to the fact that compiler will add boundary checks everywhere (since it is not Vec of any length rather than an array of a constant length), another unnecessary panic source is `unreachable!` which is there only to save us some compilation time (really?) since we introduced `__count` instead of using some macro which would just generate a const for us.

core/primitives/src/views.rs
```diff
@@ -1000,6 +1001,32 @@ impl From<ExecutionStatus> for ExecutionStatusView {
     }
 }
 
+#[derive(BorshSerialize, BorshDeserialize, Serialize, Deserialize, PartialEq, Clone, Eq, Debug)]
+pub struct ProfileDataView {
+    costs: Vec<(Cost, u64)>,
+}
```
matklad, line 1007: Hm, the JSON serialization doesn't look right to me:

```rust
#[test]
fn profile_view_json_repr() {
    let pdv = ProfileDataView { costs: vec![(Cost::ExtCost(ExtCosts::write_memory_base), 92)] };
    let json = serde_json::to_string(&pdv).unwrap();
    assert_eq!(&json, r#"{"costs":[[{"ExtCost":"write_memory_base"},92]]}"#)
}
```

I think what we want to see here is just `{"write_memory_base": 92}`. The current representation includes a bunch of details about how our code is structured today. But we should base the JSON repr on what the actual data are we storing, and how to present the data to the user. The data is a *mapping* of costs to values, it's not an array with pairs of objects. 

I guess it probably makes sense to serialize u64 as string, as we do for gas throughout the API.

matklad (changes requested):

core/primitives-core/src/profile.rs
```diff
@@ -1,15 +1,17 @@
...
 
 /// Profile of gas consumption.
-#[derive(Clone, PartialEq, Eq)]
+/// Vecs are used for forward compatible. should only append new costs and never remove old costs
+/// When add new cost, the new cost should also be append to ProfileData::ALL
+#[derive(Clone, PartialEq, Eq, BorshSerialize, BorshDeserialize)]
 pub struct ProfileData {
-    data: DataArray,
+    costs: Vec<u64>,
 }
```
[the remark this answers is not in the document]
matklad: Taking https://github.com/near/nearcore/pull/4438#discussion_r688456128 into account, why do we need a Vec in the first place here? Seems `Box<[u64; LEN]>` is what's needed.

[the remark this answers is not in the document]
frol: I think the idea behind using `Vec` was to enable deserialization from arrays that are longer or shorter than the current one. Though, I think it should be implemented through custom deserialization, which would initialize the missing costs (if the array is shorter) with zeros. We should never arrive to the point when we run an older code with a newer database (it won't be compatible for tons of different reasons anyway)

ailisp:
> @ailisp May I ask you to draft some telling examples of the JSON output you want to expose? I assume it will be part of the tx endpoint output, right?

Yes. For example (note the metadata part of an outcome)
```
{
  "jsonrpc": "2.0",
  "result": {
    "receipts_outcome": [
      {
        "block_hash": "4ynsYUCi4y38HyL16rpDarfbfaNTxf4NocpLj26vegkw",
        "id": "FWKdWZbUV6AuTm2ZQApTw29pur2LSVWfEiUciF2ffRHD",
        "outcome": {
          "executor_id": "test.near",
          "gas_burnt": 2593738087975,
          "logs": [],
          "metadata": {
            "ExecutionMetadataV2": {
              "costs": [
                [
                  {
                    "ExtCost": "base"
                  },
                  1323840555
                ],
                [
                  {
                    "ExtCost": "contract_compile_base"
                  },
                  35445963
                ],
                [
                  {
                    "ExtCost": "contract_compile_bytes"
                  },
                  18573524250
                ],
                [
                  {
                    "ExtCost": "read_memory_base"
                  },
                  7829589600
                ],
                [
                  {
                    "ExtCost": "read_memory_byte"
                  },
                  91231992
                ],
                [
                  {
                    "ExtCost": "write_memory_base"
                  },
                  2803794861
                ],
                [
                  {
                    "ExtCost": "write_memory_byte"
                  },
                  43580352
                ],
                [
                  {
                    "ExtCost": "read_register_base"
                  },
                  2517165186
                ],
                [
                  {
                    "ExtCost": "read_register_byte"
                  },
                  1576992
                ],
                [
                  {
                    "ExtCost": "write_register_base"
                  },
                  2865522486
                ],
                [
                  {
                    "ExtCost": "write_register_byte"
                  },
                  60825024
                ],
                [
                  {
                    "ExtCost": "storage_write_base"
                  },
                  64196736000
                ],
                [
                  {
                    "ExtCost": "storage_write_key_byte"
                  },
                  563862936
                ],
                [
                  {
                    "ExtCost": "storage_write_value_byte"
                  },
                  248148312
                ],
                [
                  {
                    "ExtCost": "touching_trie_node"
                  },
                  64407823704
                ]
              ]
            }
          },
          "receipt_ids": [
            "2ZJm6abvUYEWPopTXNCyBvMJ5qLaPLKRpZt3Hnm7j4aa"
          ],
          "status": {
            "SuccessValue": "AAAAAAAAAAA="
          },
          "tokens_burnt": "2593738087975000000000"
        },
        "proof": [
          {
            "direction": "Left",
            "hash": "2e2Mue2m7BtjaFryaFEmoXVKYQ49tHqHCndGRVy7ZSgS"
          }
        ]
      },
      {
        "block_hash": "FUjK1VZgZrhv9PJC2UQ8ZfTcNMwup1pPqenAHrHwRNjN",
        "id": "2ZJm6abvUYEWPopTXNCyBvMJ5qLaPLKRpZt3Hnm7j4aa",
        "outcome": {
          "executor_id": "test.near",
          "gas_burnt": 223182562500,
          "logs": [],
          "metadata": {
            "ExecutionMetadataV2": {
              "costs": []
            }
          },
          "receipt_ids": [],
          "status": {
            "SuccessValue": ""
          },
          "tokens_burnt": "0"
        },
        "proof": []
      }
    ],
    "status": {
      "SuccessValue": "AAAAAAAAAAA="
    },
    "transaction": {
      "actions": [
        {
          "FunctionCall": {
            "args": "DgAAAAAAAAAOAAAAAAAAAA==",
            "deposit": "100000000000",
            "gas": 10000000000000,
            "method_name": "write_key_value"
          }
        }
      ],
      "hash": "4K73HY29S85Xki8xQRgSiyEtQNXyBjAbRyFfgE6WtHA1",
      "nonce": 140,
      "public_key": "ed25519:J5d4Pfp2By5fPig4HyDWMDynFPCeEYpzrdHQe9KaiJkh",
      "receiver_id": "test.near",
      "signature": "ed25519:3ouhPGLfB31o918n1ViTz2rMfDZXt5DeX3LsAHN3CyLZiCt2yujtdx3SR9U24AWPqXQfAQ5Nsp6BxLghqg7d3sqH",
      "signer_id": "test.near"
    },
    "transaction_outcome": {
      "block_hash": "4ynsYUCi4y38HyL16rpDarfbfaNTxf4NocpLj26vegkw",
      "id": "4K73HY29S85Xki8xQRgSiyEtQNXyBjAbRyFfgE6WtHA1",
      "outcome": {
        "executor_id": "test.near",
        "gas_burnt": 2427990313954,
        "logs": [],
        "metadata": "ExecutionMetadataV1",
        "receipt_ids": [
          "FWKdWZbUV6AuTm2ZQApTw29pur2LSVWfEiUciF2ffRHD"
        ],
        "status": {
          "SuccessReceiptId": "FWKdWZbUV6AuTm2ZQApTw29pur2LSVWfEiUciF2ffRHD"
        },
        "tokens_burnt": "2427990313954000000000"
      },
      "proof": [
        {
          "direction": "Right",
          "hash": "47EjpZZA7HcUJpxwui92v2sxf1MxyLvZkYP6hAZcWsw8"
        }
      ]
    }
  },
  "id": "dontcare"
}
```

core/primitives/src/views.rs
ailisp: I particularly considered `{"write_memory_base": 92}` form, but it's losing the of ExtCost type information. I think it's worth given future add new type of cost other than ActionCosts and ExtCosts, there're risks of collision. Also, A clear glance to separate action cost and ext cost is important to help debugging cost difference

core/primitives-core/src/profile.rs
ailisp: > another unnecessary panic source is unreachable! which is there only to save us some compilation time (really?) since we introduced __count instead of using some macro which would just generate a const for us.

I don't understand this statement, why `__count => unreachable!` is unnecessary? Without it match cases isn't complete.

core/primitives/src/views.rs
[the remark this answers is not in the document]
ailisp: >  This is not a self-explanatory format at all; I am completely lost here
    Please, avoid using the default externally tagged enum representations since it is quite inconvenient format for the client code. Internally tagged or adjacently tagged representations are much easier to work with. More details about enum representations: https://serde.rs/enum-representations.html

This is rather a personal thing. I do have read this document before this design. 

If you are mean to replace enum part, from externally to internally tagged repr, such as from `{"ExtCost": "base"}` to `{"category":"ExtCost", "kind":"base"}`, I do think the previous one is more clear, since later you also need to document, to document what a category means.
If you mean the entire costs, It does not fit the model since costs is a BTreeMap<Enum, u64>, the tagged or adjacently tagged enum repr for enum is for the enum itself. In json, the resulting enum `{"ExtCost": "base"}` cannot be a map key, therefore, it's natural for the key-value map listed as array. 

> Please, don't expose unnamed fields ([{}, 1323840555])

This does not happen

core/primitives-core/src/profile.rs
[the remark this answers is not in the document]
ailisp: Vec has a benefit to avoid migration after adding a new cost. Vec, borsh serialized, has length encoded as first bytes. So consider this example: given the testnet database, we add a new cost in the new binary. The new cost for all old txn is 0 or more acturately, "-1" since it's not counted. In the new binary, deserialize it as vec, we can know up to which costs are available and remaining costs are considered unavailable. This reduce the need on everytime we introduce a new cost, we have to add a database migration, which is slow and unnecessary to run.

ailisp:
> Sorry being late to the conversation, but I still don't see the final picture of how this fits into JSON RPC response. I am fine with any representation you need for the internal needs, but I care about the thing we expose through the API boundary since it is really hard to undo, and we stuck with all sorts of weird outputs.

For your concerns to JSON RPC response format changes, you can see it's all limited to the metadata -> V2 section, it's trivial to upgrade it to V3, V4, ... of completely different format without the backward compatible issue. Therefore it's not a problem to undo, just bump version metadata version for significant changes. 
At the same time, V2 is designed as key value pairs of costs, which is quite flexible and forward compatible: we can add new kind of cost while still under the v2 structure. There is no harm in explorer side, suppose you're having a diagram for all "known" costs, the newly added unknown ones can be ignored.

> Ideally, I would prefer to decouple the internal representation from JSON RPC outputs just like @matklad pointed out, since it was already causing us too much of a trouble ranging from unexpected type changes to leaky dependencies across the crates which in theory should not leak outside. 

I'm disagree on this for two reasons: 1) action cost (quite stable, kind of different NEAR txn actions) / ext cost (host function) classification is quite mature in our system and unlikely to change and not something we're not confident to expose. 2) even it do change in case, a metadata v2 => v3 update will smooth the confusion to explorer and external users.

> Should we consider splitting this PR into: (1) store in the database, (2) expose it through JSON RPC?

I think at this stage it's unwise to do so. There's mostly some adjustment to existing format type of changes in the PR, so I suggest we do further discussion here and come to a satisfactory JSON RPC format

core/primitives-core/src/profile.rs
frol: That part of my comment was more about the pattern we decided to use to count the number of enum variants by this workaround with `__count`. I recall I had a conversation with someone (@matklad was it you?) about using this workaround in the first place, and the alternatives we discussed: (1) keep the number of variants as a variable that has to be manually updated (prone to errors, especially on merge conflicts), (2) use macro (potential increase in compilation time was the argument against it). In this context I just highlighted that this current workaround injects runtime costs (panic handler) in, what I assume, a hot execution path

core/primitives-core/src/profile.rs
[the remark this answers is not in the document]
frol: @ailisp You don't need a database migration if you implement a custom deserialization; here is some pseudocode:

```rust
struct DataArray([u64; LEN]);

impl Deserialize for DataArray {
  fn deserialize(data) {
      let data_vec = Vec::<u64>::deserialize(data)?;
      let mut data_array = [0; LEN];
      data_array.copy_from(&data_vec[..LEN.min(data_vec.len())]);
      Ok(Self(data_array))
  }
}

struct ProfileData {
  costs: DataArray
}
```

frol (changes requested):
```js
          "metadata": {
            "V2": {
              "costs": [
                [
                  {
                    "ExtCost": "base"
                  },
                  1323840555
                ],
                [
                  {
                    "ExtCost": "contract_compile_base"
                  },
                  35445963
                ],
            ]
          }
```

From the API consumer point of view, I would prefer to have the following structure:

```js
          "metadata": {
            "version": "V2",
            "gas_profiling_data": [
                {
                    "cost_kind": "EXT_COST",
                    "ext_cost_kind": "BASE",
                    "gas_used": 1323840555
                },
                {
                    "cost_kind": "EXT_COST",
                    "ext_cost_kind": "CONTRACT_COMPILE_BASE",
                    "gas_used": 35445963
                },
            ]
          }
```

1. We are committed to keep backward compatibility, and enable future compatibility (we don't use unnamed tuples/arrays, where position means something)
2. We are committed to internally tagged and adjacently tagged representations for JSON RPC (we only keep externally tagged representations for legacy reasons)
3. We use `SCREAMING_SNAKE_CASE` for the enum variant names to indicate that those are unique identifiers
4. We spend time to make the naming self-explanatory and non-ambiguous

What is the meaning of "Ext"? (Even after reading the comment "Strongly-typed representation of the fees for counting" I cannot tell what "ext" implies in this context, which indicates to me that the naming can be improved, and it is even more so when it comes to JSON RPC APIs where you don't have enough context about the relations between the objects)

> For your concerns to JSON RPC response format changes, you can see it's all limited to the metadata -> V2 section, it's trivial to upgrade it to V3, V4, ... of completely different format without the backward compatible issue. Therefore it's not a problem to undo, just bump version metadata version for significant changes.

The old code around metadata V2 will just stop working if new outcomes will produce V3+ metadata, so it is going to be a breaking change for all the applications relying on it. It is even more cruel when you want to bump the metadata version just for the internal needs and the costs/fees remain the same. This is where internally tagged or adjacently tagged representations help us in the long run.

> even it do change in case, a metadata v2 => v3 update will smooth the confusion to explorer and external users.

Any breaking change on the JSON RPC API layer affects **unknown** amount of services. Thus, we commit to avoid breaking changes if we can.

> I think at this stage it's unwise to do so. There's mostly some adjustment to existing format type of changes in the PR, so I suggest we do further discussion here and come to a satisfactory JSON RPC format

Ok, I am fine with that. I was just speaking from my experience that any discussion takes a while and thus I suggested to consider extracting the API discussion if the rest of the implementation is good to go.

core/primitives-core/src/profile.rs
ailisp: Ah I see, so even though `__count=>unreachable!` is really unreachable in execution of the node, have it here make rustc cannot do best optimization to the whole function, which you mentioned  as" injects runtime costs (panic handler)" ?

core/primitives-core/src/profile.rs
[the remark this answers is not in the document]
ailisp: Sounds good then!

ailisp:
@frol OK I think you proprosed structure is indeed better! I'll change to that format. About the API design I think your proposed is more about easy readability and self documented, but don't see it helps address backward compatible better, so I still have some question want to discuss with you:

> The old code around metadata V2 will just stop working if new outcomes will produce V3+ metadata ... This is where internally tagged or adjacently tagged representations help us in the long run.

Hmm why internally/adjacently tagged help? Won't it also stop working if new outcomes produce v3 metadata?

> Any breaking change on the JSON RPC API layer affects unknown amount of services. Thus, we commit to avoid breaking changes if we can.

Breaking APIs are provided usually as addition not replacement, therefore user would have time period to upgrade in a depreciated but not yet dropped period of time. In our code base, it's provided by a normal RPC and a `EXPERIMENT_*` RPC. One potential benefit of the external tagged api is you can provide `v2` and `v3` at same time, for some protocol versions to wait user updates. 
Also I believe current format (either internally tagged or externally, semantically they're equivalently expressive) is good to endure breaking changes. We're adding and never removing existing type of cost, so document to API users should say that there can be more cost added, program consumes this API should ignore unknown new costs to avoid breaking.

ailisp:
@frol now rpc output is this format:
```
          "metadata": {
            "gas_profiling_data": [
              {
                "cost_kind": "ExtCost",
                "ext_cost_kind": "base",
                "gas_used": 264768111
              },
 ...
            ],
            "version": "V2"
          },
```
and profile data is a sized array instead of vec. Please take another look

frol (approved):
> now rpc output is this format:

Yay, thanks for implementing it!

> profile data is a sized array instead of vec

Looks good to me :+1: 

@matklad WDYT?

> Hmm why internally/adjacently tagged help? Won't it also stop working if new outcomes produce v3 metadata? One potential benefit of the external tagged api is you can provide v2 and v3 at same time.

Well, having v2 and v3 at the same time will be quite convoluted from the Rust structs point of view. With internally/adjacently tagged representation you may keep the compatibility by just adding new fields (the most common scenario is that we introduce something extra). Imaging we change to this:

```json
"v2": { "gas_profiling_data": [...] }
```

```json
"v3": { "gas_profiling_data": [...], "total_gas_used": 123 }
```

With externally typed variant, we have to upgrade **all** the clients to read the data from "v3" field instead of "v2". If we keep "v2" along side with "v3", it will require quite some support on the server side. Yet, if we use internally/adjacently tagged version:

```json
{ "version": "v2", "gas_profiling_data": [...] }
```

```json
{ "version": "v3", "gas_profiling_data": [...], "total_gas_used": 123 }
```

all the existing clients will be fine reading from `metadata["gas_profiling_data"]` (no changes required, and the server implementation is as straightforward as it could be).

This PR looks good to me modulo the minor comments I left in the code.

core/primitives-core/src/profile.rs
```diff
@@ -19,154 +59,464 @@ impl Default for ProfileData {
...
+        for i in 0..DataArray::LEN {
             self.data[i] = self.data[i].saturating_add(other.data[i]);
         }
     }
 
     #[inline]
     pub fn add_action_cost(&mut self, action: ActionCosts, value: u64) {
-        self.add_val(ProfileData::ACTION_START + action as usize, value);
+        self[Cost::ActionCost { action_cost_kind: action }] =
+            self[Cost::ActionCost { action_cost_kind: action }].saturating_add(value);
```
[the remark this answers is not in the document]
ailisp: Index is suggested by @matklad on this comment: https://github.com/near/nearcore/pull/4438#discussion_r674051345

So if having index it's natural to also have index_mut. 

I feel good for both `get()` and `index` flavor, but slightly prefer index, since it indicates constant time accessing.

core/primitives-core/src/profile.rs
```diff
@@ -1,13 +1,53 @@
...
+impl BorshSerialize for DataArray {
+    fn serialize<W: std::io::Write>(&self, writer: &mut W) -> Result<(), std::io::Error> {
+        let v: Vec<_> = self.0.into();
+        BorshSerialize::serialize(&v, writer)
+    }
+}
 
 /// Profile of gas consumption.
-#[derive(Clone, PartialEq, Eq)]
+/// Vecs are used for forward compatible. should only append new costs and never remove old costs
```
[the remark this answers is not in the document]
ailisp: Good catch!

core/primitives-core/src/profile.rs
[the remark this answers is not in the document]
frol: I see. Well, I can live with `index` implementation

matklad:
I want to spend more iterations on the specifics of the JSON format. The current solution is indeed workable, but I don't think it's optimal. As this is a public API (so we can't just change *our* code, there are 3rd party users of the API), it makes sense to invest a bit more time here. 

For the reference, the current format is:

```JSON
          "metadata": {
            "gas_profiling_data": [
              {
                "cost_kind": "EXT_COST",
                "ext_cost_kind": "BASE",
                "gas_used": 264768111
              },
 ...
            ],
            "version": "V2"
          },
```

I propose a couple tweaks: 

**First**, my understanding is that `cost_kind` / `ext_cost_kind` split is to a significant degree an implementation detail. From the user's perspective, there are just various costs with different names. So I suggest we use the following format for the single entry:

```JSON
{
  "cost": "EXT_OR_ACTION_COST_NAME",
  "gas_used": 264768111,
}
```

I wouldn't think too much about specific names at the moment -- we can always change the names later, and we will change them, to make sure that each cost's name is understandable

**Second**, I don't think we need `version` field in the JSON. In JSON, unlike Borsh, we can just add new fields. 

To put it in other words, the `enum Foo { V1 }` pattern we use is the artifact due to the fact that we store data in borsh in rocks. The design of the JSON API shouldn't be affected by the way we store data internally. 

So, just removing `version` field should be fine. 

_However_, the semantics of metadata is that each bit of it is entirely optional. That means that the clients should gracefully handles additions to the metadata. This is not true for current Rust impl. The following will fail to deserialize:

```JSON
"metadata": {
  "version": "V3",
  "new_field": 92,
},
```

A client in any language should gracefully handle metadata changes. That's the purpose of metadata infra -- adding non-correctness-critical information without the need to go through the painful upgrade process.  For Rust, I think this can be achieved with the following setup:

```rust
pub struct ExecutionMetadataView {
   #[serde(default)]
    gas_profiling_data: Vec<CostGasUsed>,
}
```

So: I don't think `version` field is necessary, but I am fine with keeping it there. However, the Rust struct must handle changes gracefully, it should not fail on future's metadatas. 


**Third**, I'd name it just `gas_profile`, the `_data` is redundant as the name goes.

**Fourth**, I explicitly like frol's suggestion to use array of objects rather than a map -- it's less cute, but allows us to add new fields to objects in the future, and also gives names to keys and values.

core/primitives-core/src/profile.rs
```diff
@@ -19,154 +58,464 @@ impl Default for ProfileData {
...
+            #[cfg(feature = "protocol_feature_alt_bn128")]
+            Cost::ExtCost { ext_cost_kind: ExtCosts::alt_bn128_g1_sum_byte } => &mut self.data[68],
+            Cost::ExtCost { ext_cost_kind: ExtCosts::__count } => unreachable!(),
+        }
+    }
+}
+
+#[derive(BorshSerialize, BorshDeserialize, Serialize, Deserialize, PartialEq, Clone, Eq, Debug)]
+pub struct CostGasUsed {
+    #[serde(flatten)]
```
matklad, line 495: Let's keep the bits related to JSON serilization in the views.rs file. profile.rs doesn't need to care what the public JSON API looks like.

core/primitives-core/src/profile.rs
```diff
@@ -19,154 +58,464 @@ impl Default for ProfileData {
...
+            Cost::ExtCost { ext_cost_kind: ExtCosts::alt_bn128_g1_sum_byte } => &self.data[68],
+            Cost::ExtCost { ext_cost_kind: ExtCosts::__count } => unreachable!(),
+        }
+    }
+}
+
+impl IndexMut<Cost> for ProfileData {
+    fn index_mut(&mut self, index: Cost) -> &mut Self::Output {
+        match index {
+            Cost::ActionCost { action_cost_kind: ActionCosts::create_account } => &mut self.data[0],
```
matklad, line 371: One way to not duplicate the giant match is to have

```
impl Cost {
  fn index(self) -> usize
}

impl IndexMut<Cost> for ProfileData {
    fn index_mut(&mut self, cost: Cost) -> &mut Self::Output {
        &mut self.data[cost.index()]
    }
```

core/primitives-core/src/profile.rs
matklad: >In this context I just highlighted that this current workaround injects runtime costs (panic handler) in, what I assume, a hot execution path

[At the hot call-sites](https://github.com/near/nearcore/blob/96afe0329006a1dd3b4be065c687407268bbfb24/runtime/near-vm-logic/src/logic.rs#L2300), the costs are compile-time constants,  so the compiler should be able to constant-fold both the match *and* the bounds check on the actual array access. 

In a separate PR, I want to merge both costs enums into one, so that all explicit matching goes away, we can write just `data[cost as usize]`, and adding a new cost (of either kind) can be done in backwards-compatible fashion by appending to the end.

core/primitives-core/src/profile.rs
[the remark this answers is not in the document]
matklad: > I feel it is not the Rust way to abuse Index ([]) operator for something that is not a number or a range, 

> since it indicates constant time accessing.

I don't think either of this statements is true. Here's stdlib counter example for both:

```rust
fn main() {
    let map: std::collections::BTreeMap<_, _> = std::iter::once(("hello", 92)).collect();
	let value = map["hello"];
	println!("{}", value);
}
```

here, the `[]` is neither constant time, nor with range/integer. 

`[]` is a general operator for accessing a mapping in an *infallible* way, while `get` usually return options. In this case, we guarantee that the operation can never fail

That said, I don't have a strong opinion between Index or get, either is fine by me!

core/primitives-core/src/profile.rs
ailisp: Cool solution! I also wonder is it possible to derive Index from IndexMut?

core/primitives-core/src/profile.rs
ailisp: OK!

ailisp:
> First, my understanding is that cost_kind / ext_cost_kind split is to a significant degree an implementation detail. From the user's perspective, there are just various costs with different names. So I suggest we use the following format for the single entry:

Hmm I don't agree action_cost/ext_costing dividing is implementation detail, they're clearly different and contract developer need to learn that. Action cost is inherent cost of doing one of ten near transaction action, such as create receipt, create account, deploy a contract, etc. They happen to all transactions. Ext cost is wasm contract call a host function, and if that host function being called, how much gas was spent, such as write to storage, doing a hash, etc. They only happen in transaction that do smart contract call.

> So: I don't think version field is necessary, but I am fine with keeping it there. However, the Rust struct must handle changes gracefully, it should not fail on future's metadatas.

I consider your and @frol's format again. I think let's do this: keep a version, but that is only for adding something completely different to the profile_data, or in case at a moment we consider to completely reshape profile_data due to some dramatic protocol upgrade. If it's only adding a new cost, then we **do not bump metadata version**. I think this combines the benefit of both approach.

> Third, I'd name it just gas_profile, the _data is redundant as the name goes.

Sounds good!

> Fourth, I explicitly like frol's suggestion to use array of objects rather than a map -- it's less cute, but allows us to add new fields to objects in the future, and also gives names to keys and values.

I agree with that too!

core/primitives-core/src/profile.rs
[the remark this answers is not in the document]
ailisp: @matklad Thanks! I didn't know BTreeMap can also be accessed with `[]`, but given so I think that was a bad decision. People get surprised of index is not a constant time operation, but breemap.get is fine and `get` is not surprising to be not O(1)

ailisp:
Hi @frol I had a talk with @matklad to discuss some more enhancement on the format, and we mostly went into consensus, can you take a look and see whether it makes sense?

Proposed new format:
```json
"metadata": {
  "version": 2,
  "gas_profile": [
    {
      "cost_category": "EXT_COST"
      "cost": "ERC_RECOVER",
      "gas_used": 264768111
    },
  ],
},
```

Major change: "cost" for both ext_cost and action cost. rename the cost_kind as cost_category. Also version "v2" -> 2. Do you think it's better? A benefit we could see is easy to compare versions in a range.

Also, @matklad has convinced me to use string instead of cost in rust, to never cause a serialize/deserialize failure in rust, then we can freely rename cost without breaking:
```
#[derive(Serialize, Deserialize)]
struct CostEntry {
  cost: String,
}
```

The last point from @matklad  is we do not guarantee cost name will remain unchanged to RPC users. This way we can rename them freely in subsequent PRs. We might give this guarantee in future once name of cost is stable enough, but once we gives the guarantee we can't break. @matklad do this also indicate we'll not do renaming in this PR?

matklad:
>do this also indicate we'll not do renaming in this PR?

Yeah, I suggest doing renames separaetly -- there are a lot of costs, and we must make sure that every one has a resaonable name which makes sense outiside of nearcore

core/primitives-core/src/profile.rs
matklad: Not really -- rust shares this limitation with C++, there isn't an easy way to be polymorphic over const-ness, sometimes you need to just write the same code twice. This rarely happens though.

ailisp:
Hi @frol could you take a look at the question we have in https://github.com/near/nearcore/pull/4438#issuecomment-901783480

@matklad Since the rename can happen after this PR, what would you think is required to merge this PR?

frol:
> Major change: "cost" for both ext_cost and action cost. rename the cost_kind as cost_category.

LGTM

> Also version "v2" -> 2. Do you think it's better? A benefit we could see is easy to compare versions in a range.

Yeah, I like it better than "V2" (I am also fine if we drop the version field altogether since in JSON it does not help much, people would still ignore it in most of the cases)

> Another thing we consider rename is EXT_COST, what is EXT mean? Possibly the host cost that out of wasm sandbox, so it's very intuitive to user. And some cost under EXT, such as base, isn't clear what it means. This means we may need to adjust these cost namings in later PRs.

I am all for the clear naming! As I already pointed out "ext" does not help at all, so I would love to rename it in this PR. All the individual cost names are indeed too much of an ask to refactor right now.

> To do this in a backward compatible way, matklad has convinced me to use string instead of cost in rust, to never cause a serialize/deserialize failure in rust, then we can freely rename cost without breaking:

Sounds like a sane suggestion while we are still figuring out the naming :+1: 

> `"gas_used": 264768111`

I suggest we use string encoding for `u64` values since otherwise we may hit rounding errors. Even though, `gas_used` is not expected to be more than (2^53 ~= 9007 TGas)

core/primitives-core/src/profile.rs
```diff
@@ -1,13 +1,51 @@
...
+use borsh::{BorshDeserialize, BorshSerialize};
+use serde::{Deserialize, Serialize};
 
 use crate::config::{ActionCosts, ExtCosts};
-use crate::types::Gas;
 
-/// Data for total gas burnt (index 0), and then each external cost and action
-type DataArray = [u64; ProfileData::LEN];
+#[derive(Clone, PartialEq, Eq)]
+pub struct DataArray([u64; Self::LEN]);
```
matklad, line 10:
```suggestion
pub struct DataArray(Box<[u64; Self::LEN]>);
```

This struct has rather big size_of, so it's going to be faster to box this once,  and then pass the pointer around.

PR #2831, opened 2020-06-10
feat(core): Defined AccountId to hold a contract of being valid

frol:
**This is work-in-progress, there are tons of compilation errors (though primitives crate compiles). I decided to push it to git just in case, I am currently busy with other tasks.**

*Resolves #2074.*

PR #4440, opened 2021-06-30
refactor: Introduce strictly typed and assuredly validated AccountId

### LINKED ISSUE #2074: Define AccountId to hold a contract of being valid
We should leverage strong typing and safety contracts of Rust, and make AccountId type that enforces its validity. This way, we improve node security, especially, around RPC boundary.

miraclx:
This PR introduces a strictly typed structure - `AccountId` that enforces account ID validation.
This way, we can ensure validity and thereby fail fast before an invalid account name is used.

This also, optionally, ensures validation on {,de}serialization so, yeah..

The structure is defined within its own self-contained crate `near-account-id` to allow for easy publishing.
The extra dependencies - `borsh`, `serde` are made optional to allow the crate to be lightweight enough to be depended upon.

Within `nearcore`, the structure is reexported via `near_primitives::AccountId` and the crate via `near_primitives{,_core}::account::id` with default features (`borsh`, `serde`) but you can opt-out of those by depending on it directly.

Resolves #2074, supersedes #2831.

# Testing Plan

* [x] Existing unit tests pass with minor updates
* [x] Nayduck tests pass without modifications
* [x] Manually ensured that the protocol is not changed


miraclx:
Since we now preemptively perform account ID validation, cases like [these](https://github.com/near/nearcore/blob/7d16cf589ab8595c025c11da794373f0dbff787b/test-utils/testlib/src/standard_test_cases.rs#L623-L651) are ensured and extra testing is unnecessarily redundant. However, this, in turn, leads to having variants like `InvalidTxError::InvalidReceiverId` that are unused: https://github.com/near/nearcore/blob/a6bf293727fd5189dee19a7b45514d09cf7c0085/core/primitives/src/errors.rs#L119
Would be nice if we could clean that up a bit, without breaking a lot.

miraclx:
Also, throughout the repo, a debug view has been used on `AccountId` before this change. Back when it was simply a `String` this was okay.. But now, it's a lot more verbose (perhaps more descriptive, depending on how you look at it)

As an example:

<table>
<tr>
	<td rowspan=2> <pre lang="rust"> format!("{:?}", AccountId("miraclx.near")) </pre>
    <td> before	
    <td> <code> "miraclx.near" </code>
<tr>
	<td> after
	<td> <code> "AccountId(\"miraclx.near\")" </code> 
</table>

We could manually impl `Debug` and remain consistent with the former behavior but the proper `Debug` view seems more reasonable to me, what do you think?

I bring this up because it's one of the issues with failing tests in CI

<details>
<summary> snapshot <a href="https://buildkite.com/nearprotocol/nearcore/builds/1895#1b4cbf96-f7c4-4cd5-9caa-228c38d53517">(CI)</a> </summary>

``` diff
---- adapters::tests::test_convert_block_changes_to_transactions stdout ----
━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ Snapshot Differences ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
Snapshot file: chain/rosetta-rpc/src/adapters/snapshots/near_rosetta_rpc__adapters__tests__validators_update_transaction.snap
Snapshot: validators_update_transaction
Source: chain/rosetta-rpc/src/adapters/mod.rs:992
-old snapshot
+new results
────────────────────────────────────────────────────────────────────────────────
validators_update_transaction
────────────┬───────────────────────────────────────────────────────────────────
   12    12 │             status: Some(
   13    13 │                 Success,
   14    14 │             ),
   15    15 │             account: AccountIdentifier {
-  16       │-                address: "nfvalidator1.near",
+        16 │+                address: AccountId(
+        17 │+                    "nfvalidator1.near",
+        18 │+                ),
   17    19 │                 sub_account: None,
   18    20 │             },
   19    21 │             amount: Some(
   20    22 │                 Amount {
┈┈┈┈┈┈┈┈┈┈┈┈┼┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈
   37    39 │             status: Some(
   38    40 │                 Success,
   39    41 │             ),
   40    42 │             account: AccountIdentifier {
-  41       │-                address: "nfvalidator2.near",
+        43 │+                address: AccountId(
+        44 │+                    "nfvalidator2.near",
+        45 │+                ),
   42    46 │                 sub_account: None,
   43    47 │             },
   44    48 │             amount: Some(
   45    49 │                 Amount {
────────────┴───────────────────────────────────────────────────────────────────
thread 'adapters::tests::test_convert_block_changes_to_transactions' panicked at 'snapshot assertion for 'validators_update_transaction' failed in line 992', /var/lib/buildkite-agent/.cargo/registry/src/github.com-1ecc6299db9ec823/insta-1.6.3/src/runtime.rs:1026:9
```

</details>

miraclx:
Also, `node_key.json` has a field `account_id` that previously held an empty string. CI fails for this exact reason. To mitigate this, this PR uses the `"node"` account id instead to ensure validity. This file has to be reset on CI systems.

cc @chefsale

Or should I engineer a workaround within this PR?

chefsale:
So, the CI machines are destroyed on every run, so we don't store the keys, but rather generate them on every run. I believe the issue is rather caused by some code change where we have missed something. I assume this failure comes from account ids being invalid and this is throwing an error now? I assume there is some part of code which generates this node key which we need to adapt. cc: @bowenwang1996 @frol 

Also just to verify we do not change the account id structure in the current generated key json files at all, but just have this additional check? If we do, we should ensure to be backwards compatible.

chefsale (changes requested):
Let's figure out how to fix the tests.

nearcore/src/config.rs
```diff
@@ -723,8 +724,10 @@ impl NearConfig {
             validator_signer.write_to_file(&dir.join(&self.config.validator_key_file));
         }
 
-        let network_signer =
-            InMemorySigner::from_secret_key("".to_string(), self.network_config.secret_key.clone());
+        let network_signer = InMemorySigner::from_secret_key(
+            "node".parse().unwrap(),
+            self.network_config.secret_key.clone(),
+        );
         network_signer.write_to_file(&dir.join(&self.config.node_key_file));
```
miraclx, line 731: > I assume there is some part of code which generates this node key which we need to adapt.

I believe that's here, and it's been substituted for `"node"`.

chefsale: It seems like it is throwing this error:
```


[2021-06-30T16:05:02Z] Jun 30 16:04:51.611  INFO neard: Version: 1.2.0, Build: c9752738, Latest Protocol: 45
--
  | [2021-06-30T16:05:02Z] thread 'main' panicked at 'Failed to deserialize KeyFile: Error("the value is too short for account ID", line: 2, column: 18)', core/crypto/src/key_file.rs:36:40
  | [2021-06-30T16:05:02Z] stack backtrace:
  | [2021-06-30T16:05:02Z]    0: rust_begin_unwind
  | [2021-06-30T16:05:02Z]              at /rustc/2fd73fabe469357a12c2c974c140f67e7cdd76d0/library/std/src/panicking.rs:493:5
  | [2021-06-30T16:05:02Z]    1: core::panicking::panic_fmt
  | [2021-06-30T16:05:02Z]              at /rustc/2fd73fabe469357a12c2c974c140f67e7cdd76d0/library/core/src/panicking.rs:92:14
  | [2021-06-30T16:05:02Z]    2: core::option::expect_none_failed
  | [2021-06-30T16:05:02Z]              at /rustc/2fd73fabe469357a12c2c974c140f67e7cdd76d0/library/core/src/option.rs:1300:5
  | [2021-06-30T16:05:02Z]    3: core::result::Result<T,E>::expect
  | [2021-06-30T16:05:02Z]              at /rustc/2fd73fabe469357a12c2c974c140f67e7cdd76d0/library/core/src/result.rs:997:23
  | [2021-06-30T16:05:02Z]    4: near_crypto::key_file::KeyFile::from_file
  | [2021-06-30T16:05:02Z]              at /var/lib/buildkite-agent/builds/buildkite-i-0a49689549841a3a9-1/nearprotocol/nearcore/core/crypto/src/key_file.rs:36:9
  | [2021-06-30T16:05:02Z]    5: near_crypto::signer::InMemorySigner::from_file
  | [2021-06-30T16:05:02Z]              at /var/lib/buildkite-agent/builds/buildkite-i-0a49689549841a3a9-1/nearprotocol/nearcore/core/crypto/src/signer.rs:62:9
  | [2021-06-30T16:05:02Z]    6: nearcore::config::load_config_without_genesis_records
  | [2021-06-30T16:05:02Z]              at /var/lib/buildkite-agent/builds/buildkite-i-0a49689549841a3a9-1/nearprotocol/nearcore/nearcore/src/config.rs:1107:26
  | [2021-06-30T16:05:02Z]    7: near::cli::RunCmd::run
  | [2021-06-30T16:05:02Z]              at /var/lib/buildkite-agent/builds/buildkite-i-0a49689549841a3a9-1/nearprotocol/nearcore/neard/src/../src/cli.rs:197:31
  | [2021-06-30T16:05:02Z]    8: near::cli::NeardCmd::parse_and_run
  | [2021-06-30T16:05:02Z]              at /var/lib/buildkite-agent/builds/buildkite-i-0a49689549841a3a9-1/nearprotocol/nearcore/neard/src/../src/cli.rs:47:42
  | [2021-06-30T16:05:02Z]    9: near::main
  | [2021-06-30T16:05:02Z]              at /var/lib/buildkite-agent/builds/buildkite-i-0a49689549841a3a9-1/nearprotocol/nearcore/neard/src/../src/main.rs:42:5
  | [2021-06-30T16:05:02Z]   10: core::ops::function::FnOnce::call_once
  | [2021-06-30T16:05:02Z]              at /rustc/2fd73fabe469357a12c2c974c140f67e7cdd76d0/library/core/src/ops/function.rs:227:5
  | [2021-06-30T16:05:02Z] note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
  | [2021-06-30T16:05:02Z]
  | [2021-06-30T16:05:05Z] node 3 started
  | [2021-06-30T16:05:18Z] Traceback (most recent call last):
  | [2021-06-30T16:05:18Z]   File "tests/sanity/upgradable.py", line 167, in <module>
  | [2021-06-30T16:05:18Z]     main()
  | [2021-06-30T16:05:18Z]   File "tests/sanity/upgradable.py", line 83, in main
  | [2021-06-30T16:05:18Z]     assert 'error' not in res, res
```

so I believe the "node" substitution is not working. I assume it expects most likely some real ed25519 key, but than again confusing why does empty work. Let me check a bit more.

chefsale: Also one other test is failing on this one:
```


[2021-06-30T16:10:14Z] Snapshot: validators_update_transaction
--
  | [2021-06-30T16:10:14Z] Source: chain/rosetta-rpc/src/adapters/mod.rs:992
  | [2021-06-30T16:10:14Z] -old snapshot
  | [2021-06-30T16:10:14Z] +new results
  | [2021-06-30T16:10:14Z] ────────────────────────────────────────────────────────────────────────────────
  | [2021-06-30T16:10:14Z] validators_update_transaction
  | [2021-06-30T16:10:14Z] ────────────┬───────────────────────────────────────────────────────────────────
  | [2021-06-30T16:10:14Z]    12    12 │             status: Some(
  | [2021-06-30T16:10:14Z]    13    13 │                 Success,
  | [2021-06-30T16:10:14Z]    14    14 │             ),
  | [2021-06-30T16:10:14Z]    15    15 │             account: AccountIdentifier {
  | [2021-06-30T16:10:14Z]    16       │-                address: "nfvalidator1.near",
  | [2021-06-30T16:10:14Z]          16 │+                address: AccountId(
  | [2021-06-30T16:10:14Z]          17 │+                    "nfvalidator1.near",
  | [2021-06-30T16:10:14Z]          18 │+                ),
  | [2021-06-30T16:10:14Z]    17    19 │                 sub_account: None,
  | [2021-06-30T16:10:14Z]    18    20 │             },
  | [2021-06-30T16:10:14Z]    19    21 │             amount: Some(
  | [2021-06-30T16:10:14Z]    20    22 │                 Amount {
  | [2021-06-30T16:10:14Z] ┈┈┈┈┈┈┈┈┈┈┈┈┼┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈
  | [2021-06-30T16:10:14Z]    37    39 │             status: Some(
  | [2021-06-30T16:10:14Z]    38    40 │                 Success,
  | [2021-06-30T16:10:14Z]    39    41 │             ),
  | [2021-06-30T16:10:14Z]    40    42 │             account: AccountIdentifier {
  | [2021-06-30T16:10:14Z]    41       │-                address: "nfvalidator2.near",
  | [2021-06-30T16:10:14Z]          43 │+                address: AccountId(
  | [2021-06-30T16:10:14Z]          44 │+                    "nfvalidator2.near",
  | [2021-06-30T16:10:14Z]          45 │+                ),
  | [2021-06-30T16:10:14Z]    42    46 │                 sub_account: None,
  | [2021-06-30T16:10:14Z]    43    47 │             },
  | [2021-06-30T16:10:14Z]    44    48 │             amount: Some(
  | [2021-06-30T16:10:14Z]    45    49 │                 Amount {
  | [2021-06-30T16:10:14Z] ────────────┴───────────────────────────────────────────────────────────────────
  | [2021-06-30T16:10:14Z] thread 'adapters::tests::test_convert_block_changes_to_transactions' panicked at 'snapshot assertion for 'validators_update_transaction' failed in line 992', /var/lib/buildkite-agent/.cargo/registry/src/github.com-1ecc6299db9ec823/insta-1.6.3/src/runtime.rs:1026:9
```

I'm not really sure what the issue is. It seems like there are two seperate issues ones seems to be that it is unable to create a `InMemorySigner::from_file` due to something, the other seems like the accountId format is clashing with the old string only version.

I believe best people to talk to are either @bowenwang1996 @abacabadabacaba or @mfornet maybe, as frol is out of the office?

chefsale: Also @EgorKulikov would be a good person as he was working a lot on the AccountId refactor, so could be familiar with this code parts.

chefsale: It seems to me that this tests actually needs to be changed/fixed here: https://github.com/near/nearcore/blob/master/chain/rosetta-rpc/src/adapters/mod.rs#L968. As now the account type is different, but will leave it to them, as they have better understanding.

miraclx: The second issue as pointed out here is caused by having a non-String Debug representation of `AccountId`. It's properly explained in the dropdown here https://github.com/near/nearcore/pull/4440#issuecomment-871606219. We could maintain the previous `"id"` representation or use the new `AccountId("id")` format.

chefsale: In that case I assume it is fine just to implement that. Still than the questions is why does the other failure happen.

ailisp:
@miraclx as far as i understand, backward_compatible.py upgradable.py need to pass, even when it's a not a backward compatible change. In this case, you can introduce a new protocol version, and based on protocol version do different thing. New protocol version should handle all previous protocol version, when protocol version from peers is an older version, this ensures situation like: one of the validator have upgrade with your changes, while other validator nodes not yet, then the new node behaves in the old way (no account id validity check)

frol:
I believe `"AccountId(\"miraclx.near\")"` is better, so unless there is something that I don't take into account, we should use this explicit debug formatting.

As to `""` -> `"node"`, I just realized that we have tons of already existing `node_key.json` files with `""` in them. Those should be deserializable; we cannot break all the existing nodes for this reason. I suggest we consider using a dedicated structure for `node_key.json`, which only has `public_key` and `secret_key` fields, and thus ignores the `account_id` if it is present in that file.

runtime/near-vm-logic/src/logic.rs
```diff
@@ -2191,7 +2191,7 @@ impl<'a> VMLogic<'a> {
         let buf = self.get_vec_from_memory_or_register(ptr, len)?;
         self.gas_counter.pay_base(utf8_decoding_base)?;
         self.gas_counter.pay_per(utf8_decoding_byte, buf.len() as u64)?;
-        let account_id = AccountId::from_utf8(buf).map_err(|_| HostError::BadUTF8)?;
+        let account_id = AccountId::try_from(buf.as_slice()).map_err(|_| HostError::BadUTF8)?;
```
bowenwang1996, line 2194: looks like this changes the protocol

bowenwang1996 (changes requested):
This change is so invasive that I am not even sure whether it is a good idea to do it. It likely changes the protocol in many places unknowingly. The possible causes are:
* change in serialization format (not sure whether we guarantee that AccountId serializes in the same way as before)
* Reject some input that was not previously rejected.
* Return an error where we previously do not return any error

Ideally we need to go through every single place where the change happens and verify that it does not change the protocol.

runtime/near-vm-logic/src/logic.rs
miraclx: > change in serialization format (not sure whether we guarantee that AccountId serializes in the same way as before)

It's a rather simple struct, so it's represented in the same way. And, so far, from the tests I've executed, both serde + borsh serialization formats remain consistent.

> Reject some input that was not previously rejected.

Before working on this PR, I verified that all `AccountId`-s currently used on both `mainnet` and `testnet` were all valid to ensure a case like this doesn't happen. Although, I realize there could be offline cases like these, but I think, if it's gonna be used, it shouldn't have been invalid anyways. The tests had some cases like these, but they've all been updated to use valid AccountIds.

> Return an error where we previously do not return any error

Returning an error from ID validation already happens, just further down the line e.g https://github.com/near/nearcore/blob/2ca2684ad7cf400b5c617f62ac118aa105f16193/runtime/runtime/src/verifier.rs#L40-L42
This change would allow us ensure validation preemptively even before interacting with nearcore. In some cases, it could help reduce the effect of returning Errors wherever applicable. Effectively, we shift the responsibility of validation over to the user instead of propagating that through like we previously did. And in cases like jsonrpc, we can return more quantitative errors regarding invalid Ids instead of situations where it feels like your query got lost in the wind unbeknownst to you that there was an issue with ID validation. cc @frol

miraclx:
> change in serialization format (not sure whether we guarantee that AccountId serializes in the same way as before)

It's a rather simple struct, so it's represented in the same way. And, so far, from the tests I've executed, both serde + borsh serialization formats remain consistent.

> Reject some input that was not previously rejected.

Before working on this PR, I verified that all `AccountId`-s currently used on both `mainnet` and `testnet` were all valid to ensure a case like this doesn't happen. Although, I realize there could be offline cases like these, but I think, if it's gonna be used, it shouldn't have been invalid anyways. The tests had some cases like these, but they've all been updated to use valid AccountIds.

> Return an error where we previously do not return any error

Returning an error from ID validation already happens, just further down the line e.g https://github.com/near/nearcore/blob/2ca2684ad7cf400b5c617f62ac118aa105f16193/runtime/runtime/src/verifier.rs#L40-L42
This change would allow us ensure validation preemptively even before interacting with nearcore. In some cases, it could help reduce the effect of returning Errors wherever applicable. Effectively, we shift the responsibility of validation over to the user instead of propagating that through like we previously did. And in cases like jsonrpc, we can return more quantitative errors regarding invalid Ids instead of situations where it feels like your query got lost in the wind unbeknownst to you that there was an issue with ID validation. cc @frol

bowenwang1996:
@miraclx maybe I didn't make myself clear. What I meant is that we cannot simply change the behavior of the protocol without a proper protocol upgrade, even if the previous logic is flawed. For example, if the current code improperly accepts some input that it should not accept, we cannot just change to return an error without a protocol upgrade. Otherwise a malicious attacker can take advantage of this and cause the network to split or state to diverge, which are extremely bad. We should always strive for security first.

frol:
@bowenwang1996 I don't see any protocol-breaking changes here. The serialization is byte-identical, and account ID validations just moved to a slightly earlier stage. I agree that we need to express caution with such changes (e.g. we cannot introduce `method_name` validation since we already have some transactions where `method_name` has NULL-bytes, which already caused some issues on the Indexer for Explorer side). We confirmed that all the existing account IDs observed on mainnet and testnet are valid, and our p2p network does not operate with invalid accounts (while manually validates the input on receiving data from the network), and we definitely need to pass all the nayduck tests and canary tests. This change should only reduce the chance that we forget account validation at some point, so it is strictly positive. This change is expected to be 100% backward-compatible with the existing network on the protocol level.

P.S. I run nayduck (@miraclx FYI, I had to push the branch to the upstream repo, and run `python nightly/nayduck.py` from the root of the project):

Success. http://nayduck.eastus.cloudapp.azure.com:3000/#/run/1774

bowenwang1996 (commented):
@EgorKulikov @mzhangmzz could you check whether there is any change to the protocol in runtime?

runtime/near-vm-logic/src/logic.rs
```diff
@@ -2191,7 +2192,9 @@ impl<'a> VMLogic<'a> {
         let buf = self.get_vec_from_memory_or_register(ptr, len)?;
         self.gas_counter.pay_base(utf8_decoding_base)?;
         self.gas_counter.pay_per(utf8_decoding_byte, buf.len() as u64)?;
-        let account_id = AccountId::from_utf8(buf).map_err(|_| HostError::BadUTF8)?;
+        let account_id =
+            AccountId::try_from(String::from_utf8(buf).map_err(|_| HostError::BadUTF8)?)
+                .map_err(|_| HostError::InvalidAccountId)?;
```
bowenwang1996, line 2197: This still changes the protocol. We now return an error that we won't return previously if the account id parsed is 65 bytes for example. cc @frol

core/account-id/src/lib.rs
```diff
@@ -0,0 +1,527 @@
...
+                    b'-' | b'_' | b'.' => true,
+                    _ => return Err(ParseAccountError::Invalid),
+                };
+                if current_char_is_separator && last_char_is_separator {
+                    return Err(ParseAccountError::Invalid);
+                }
+                last_char_is_separator = current_char_is_separator;
+            }
+
+            (!last_char_is_separator).then(|| ()).ok_or(ParseAccountError::Invalid)
```
EgorKulikov, line 77: last part should be at least 2 characters long I think, because it is root account that should be valid by itself

miraclx: I think it's worth noting that the previous behaviour didn't ensure this constraint. Rather, we check that the entire account ID is at least 2 characters *(this behaviour was retained)*.
https://github.com/near/nearcore/blob/f453895831ed423ec41628c034d5aae5e1164896/runtime/near-runtime-utils/src/lib.rs#L14-L15
And there's a chance that account IDs on, say testnet for instance, could currently violate this. While it's a check we should have, doing that here could block this PR.

EgorKulikov: also I think all account ids that are 64 characters long have to be implicit (i. e. consist only of hex characters), see https://nomicon.io/DataStructures/Account.html

runtime/near-vm-logic/src/logic.rs
frol: @bowenwang1996 @olonho I don't understand whether an account id read by this method can be invalid or it somehow maintains the invariant of being valid?

frol: If the account id is always valid, we may as well use `.expect` here or abuse `BadUTF8` error variant (that would be unfortunate).

If the account id is not always valid, we may either abuse `BadUTF8` (again, unfortunate) or use `String` instead of `AccountId`.

bowenwang1996: > If the account id is not always valid, we may either abuse BadUTF8 (again, unfortunate) or use String instead of AccountId.

That is not the problem. The problem here is that this function could successfully parse an account id that is 65 characters long today and with this change it won't be able to and will return some error (does not matter which one it returns), which caused the behavior of the protocol to change.

frol: @bowenwang1996 I understand the problem of a protocol-breaking change very well; sorry for not being explicit about that. I am trying to understand if we want to break it or it is fine to just return `String` here and no breaking change will occur.

bowenwang1996: > I am trying to understand if we want to break it or it is fine to just return String here and no breaking change will occur.

Good question. Ideally it should return a valid account id, but it is not urgent, so maybe let's just return a `String` for now

miraclx:
@chefsale, @bowenwang1996 please review.

http://nayduck.eastus.cloudapp.azure.com:3000/#/run/1878

frol:
* [ ] Document new_unchecked with all the warnings and provide the context there.
* [ ] Document the use of new_unchecked in logic.rs, so once someone reads the code, they have a clear understanding why that is safe.
* [ ] I suggest to rename `is_64_len_hex` to `is_implicit_account_id`

runtime/near-vm-logic/src/logic.rs
frol: We (@miraclx and I) reviewed the code and arrived to a non-breaking solution.

1. We decided to introduce an "unsafe" constructor to `AccountId` type, which lets to instantiate it from any string. This is brutal and does not exactly holds the contract of `AccountId` instance always holding a valid account id; however, after cosidering the tradeoffs, we decided to go this path.
2. We manually reviewed the code path that is triggered and learned that the constructed value only lives for a certain time before it gets finally validated (the "unsafe" behaviour is completely safe today with the String type alias, and remains to be safe with this PR)
3. We hid the "unsafe" constructor befind a feature flag, so hopefully nobody else would need it
4. There should be no breaking change

bowenwang1996:
@mina86 some nayduck tests fail in a very suspicious way:
* http://nayduck.eastus.cloudapp.azure.com:3000/#/test/191674
* http://nayduck.eastus.cloudapp.azure.com:3000/#/test/191852

And it is not clear what happened

frol (approved):
This is huge!

I went through the changes myself and together with @miraclx a few times, and the implementation looks good to me.

* [x] CI tests pass
* [ ] nayduck tests pass (still have some issues; @miraclx we probably want to invite the code owners to help with those that failed)

core/account-id/Cargo.toml
```diff
@@ -0,0 +1,28 @@
+[package]
```
bowenwang1996, line 1: @frol why do we decide to make this its own crate?

core/account-id/src/lib.rs
```diff
@@ -0,0 +1,544 @@
...
+    }
+}
+
+impl TryFrom<&[u8]> for AccountId {
+    type Error = ParseAccountError;
+
+    fn try_from(account_id: &[u8]) -> Result<Self, Self::Error> {
+        std::str::from_utf8(account_id).map_err(|_| Self::Error::Invalid)?.parse()
+    }
+}
```
matklad, line 179: This impl is not needed, better let the caller call `from_utf8`

core/account-id/src/lib.rs
```diff
@@ -0,0 +1,544 @@
...
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    #[cfg(feature = "borsh")]
+    use borsh::{BorshDeserialize, BorshSerialize};
+    #[cfg(feature = "serde")]
+    use serde_json::json;
```
matklad, line 218: unreleated to the PR, but this really wants fuzz tests, a-la the ones in semver: https://github.com/dtolnay/semver/tree/master/fuzz

runtime/near-vm-logic/src/logic.rs
```diff
@@ -2162,7 +2162,16 @@ impl<'a> VMLogic<'a> {
         let buf = self.get_vec_from_memory_or_register(ptr, len)?;
         self.gas_counter.pay_base(utf8_decoding_base)?;
         self.gas_counter.pay_per(utf8_decoding_byte, buf.len() as u64)?;
-        let account_id = AccountId::from_utf8(buf).map_err(|_| HostError::BadUTF8)?;
+
+        // We return an unsafely constructed AccountId here for the sole purpose of
```
matklad, line 2166: Please only use safe/unsafe terminology for memory safety in the Rust sense.

runtime/near-vm-logic/src/logic.rs
```diff
@@ -2162,7 +2162,16 @@ impl<'a> VMLogic<'a> {
...
+        // ensuring backwards compatibility, we go ahead to validate this further down the line
+        // in node-runtime/verifier.rs#fn(validate_receipt)
```
matklad, line 2168: This is a bit of a misleading comment, as the `validator_stake` code path uses this function as well, and doesn't do any validation.

matklad (commented):
Left a bunch of comments, they call can be tackled separately, except for unsafe/unchecked terminology. I feel that it's important to stick to established terminology. 

Not sure how I feel about unvalidated accounts. Do we have any further plans here? Ie, do we want so split receipts into validated and unvalidated on the type level?Might make sense to document our expectations about `AccountId` validity somewhere...

khorolets (approved):

miraclx:
Nayduck: http://nayduck.eastus.cloudapp.azure.com:3000/#/run/1915

runtime/runtime/src/lib.rs
```diff
@@ -2099,88 +2101,6 @@ mod tests {
...
-        let initial_balance = to_yocto(1_000_000);
-        let initial_locked = to_yocto(500_000);
-        let small_transfer = to_yocto(10_000);
-        let gas_limit = 1;
-        let (runtime, tries, root, apply_state, _, epoch_info_provider) =
-            setup_runtime(initial_balance, initial_locked, gas_limit);
-
-        let n = 1;
-        let mut receipts = generate_receipts(small_transfer, n);
-        let invalid_account_id = "Invalid".to_string();
```
ailisp, line 2113: Cool these errors are statically prevented

runtime/runtime/src/verifier.rs
```diff
@@ -231,22 +221,21 @@ pub fn verify_and_charge_transaction(
...
-        });
-    }
-    if !is_valid_account_id(&receipt.receiver_id) {
-        return Err(ReceiptValidationError::InvalidReceiverId {
-            account_id: receipt.receiver_id.clone(),
-        });
-    }
+    // We retain these checks here as to maintain backwards compatibility
+    // with AccountId validation since we illegally parse an AccountId
+    // in near-vm-logic/logic.rs#fn(VMLogic::read_and_parse_account_id)
```
ailisp, line 231: :+1:

ailisp (approved):
Looks good. Nice hard and meticulous work!

bowenwang1996 (approved):

runtime/near-vm-logic/src/logic.rs
bowenwang1996: Agree

chefsale (approved):
Good luck with resolving conflicts 🗡️

runtime/near-vm-logic/src/logic.rs
miraclx: Yeah, but the previous behavior in `validator_stake` didn't require validation either so it's practically fine. Just updated the comment and we're good.

mfornet (approved):

matklad (approved):

PR #4442, opened 2021-06-30
Update to rocksdb dependency to 0.16.0

mina86:


mina86:
I’ve given cursory look at the commits between 0.15 and 0.16 and didn’t notice anything breaking. I’ve run `cargo test` and run the node for a while but don’t really know if that’s sufficient testing for a storage library upgrade.

ailisp:
@mina86 The failing is about cargo-deny. We disallow any crate in Cargo.lock have two different versions. When there's third party dependency cause this (in this case, rocksdb 0.16), add the exception to `deny.toml` and make sure `cargo-deny check bans` pass locally

bowenwang1996 (approved):

ailisp (approved):

bowenwang1996:
@olonho @frol @khorolets please review this PR

khorolets (approved):

mina86:
Haha. The number of reviewers is too damn high. ;) @olonho, could you also approve?

core/store/src/db.rs
```diff
@@ -777,10 +777,10 @@ mod tests {
     impl RocksDB {
         #[cfg(not(feature = "single_thread_rocksdb"))]
         fn compact(&self, col: DBCol) {
-            self.db.compact_range_cf::<&[u8], &[u8]>(
+            self.db.compact_range_cf(
                 unsafe { &*self.cfs[col as usize] },
-                None,
-                None,
+                Option::<&[u8]>::None,
```
olonho, line 782: Why this change?

deny.toml
```diff
@@ -129,4 +126,9 @@ skip = [
 
     # hashbrown uses an older version
     { name = "ahash", version = "=0.4.7" },
+
+    # rocksdb (transitively through clang-sys) uses this as a build
+    # dependency which conflicts with 0.6.7 that wasmer-engine-native
```
olonho, line 131: Actually do not think that we need `wasmer-engine-native` at all.

core/store/src/db.rs
mina86: The type of the argument changed from `&[u8]` to `Option<AsRef<[u8]>>` and with that the compiler could no longer deduce type.

deny.toml
mina86: It’s used in `runtime/near-vm-runner/src/wasmer1_runner.rs`. It’s also a dependency on `wasmer` 1.0.2.

olonho: OK, that's probably a mistake, but will fix it.

olonho: cc @ailisp

olonho (approved):

deny.toml
ailisp: @olonho we provided configuration option for future switching to native engine here: https://github.com/near/nearcore/blob/f453895831ed423ec41628c034d5aae5e1164896/runtime/near-vm-runner/src/wasmer1_runner.rs#L319

We haven't tried wasmer2 has native engine + singlepass support yet, if it's supported that's still a better option than JIT engine. Note the wasmer 2 renaming: native engine -> DylibEngine; JIT engine -> UniversalEngine

matklad: Yeah, we should remove support for dylib engine, filed https://github.com/near/nearcore/issues/4520 for that.

PR #4443, opened 2021-06-30
feat: prepare for 1.20.0 release

bowenwang1996:
Stabilize multiple protocol features

Test plan
---------
Nayduck http://nayduck.eastus.cloudapp.azure.com:3000/#/run/1711


runtime/near-vm-errors/src/lib.rs
```diff
@@ -213,7 +213,6 @@ pub enum HostError {
     #[cfg(feature = "protocol_feature_alt_bn128")]
     AltBn128SerializationError { msg: String },
     /// General errors for ECDSA recover.
-    #[cfg(feature = "protocol_feature_math_extension")]
     ECRecoverError { msg: String },
```
frol, line 216: I believe this should be moved before the nightly features to ensure the correct union tag

core/primitives/src/version.rs
```diff
@@ -133,30 +126,23 @@ impl ProtocolFeature {
             ProtocolFeature::ForwardChunkParts => 45,
             ProtocolFeature::RectifyInflation => 45,
             ProtocolFeature::AccessKeyNonceRange => 45,
+            ProtocolFeature::AccountVersions => 46,
+            ProtocolFeature::TransactionSizeLimit => 46,
+            ProtocolFeature::AllowCreateAccountOnDelete => 46,
+            ProtocolFeature::FixStorageUsage => 46,
+            ProtocolFeature::CapMaxGasPrice => 46,
+            ProtocolFeature::CountRefundReceiptsInGasLimit => 46,
+            ProtocolFeature::MathExtension => 46,
```
frol, line 135: Are these features documented?

@khorolets @telezhnaya please, keep in mind the `AllowCreateAccountOnDelete` feature.

https://github.com/near/nearcore/blob/5a81fa01ba9402b02251e835699c75f9bb98b516/runtime/runtime/src/actions.rs#L519-L559

This seem to use a regular transfer action receipt, so Indexer for Explorer should be fine.

frol:
@bowenwang1996 I suggest we stabilize features the week before we cut a release.

Issue #4337, opened 2021-05-28
Store restored recepts in the storage


frol:
#4248 implemented a protocol upgrade that restores receipts in the first block of an epoch when the protocol will become effective, but the current implementation only cared about the state consistency while we also want to maintain storage consistency for JSON RPC and Indexer clients. By storage consistency, I mean that we used to be able to query all the non-local receipts from the storage ([view-client](https://github.com/near/nearcore/blob/928f15df25ca4128d88b98e930d580ed2023d576/chain/client/src/view_client.rs#L879-L890)).

After brief sync with @bowenwang1996 we came to conclusion that it is possible to implement this support as a part of the protocol change and keep all the layers built on top happy.

bowenwang1996:
To be specific, we only need to store those receipts to `ColReceipts`. Execution outcomes will be populated once we apply those receipts.

bowenwang1996:
@Longarithm what is the status of this issue?

Longarithm:
@bowenwang1996 I didn't start it yet, while working on precompilation. Will start today

Longarithm:
@bowenwang1996 I think this is the best place to add restored receipts: https://github.com/near/nearcore/blob/7330f37380057e45e02cac9dc4a04c72e4338242/chain/client/src/client.rs#L561
These outgoing receipts are saved into `EncodedShardChunk` which is passed to `distribute_encoded_chunk` and then to `decode_and_persist_encoded_chunk` where the chunk itself is saved into `store_update`.
The issue is that `Client` doesn't know `NightshadeRuntime`, where `migration_data` is stored. In the past we decided to avoid extending `RuntimeAdapter` with method `get_migration_data`, but now I don't see a better option. What do you think?

bowenwang1996:
@Longarithm what is the issue with saving them to database when we load them the first time?

Longarithm:
@bowenwang1996 oh, that would be simpler. Then adding field like `restored_receipts` to `ApplyTransactionResult` should work.

PR #4444, opened 2021-06-30
Store restored receipts in the storage

Longarithm:
**Idea**
~~Return `restored_receipts` from `apply_transactions` to indicate that some receipts were added in runtime due to migration, so they will not be lost in chain and could be accessed by Indexer.~~

Create new database migration to add receipts missing in store directly into it.

**Testing**
~~Extend existing `protocol_feature_restore_receipts_after_fix_tests` by running `get_receipt` method.~~

- setup archival mainnet node
- run `./target/release/neard --home ~/.near init`, observe migration:
```
$ ./target/debug/neard --home ~/.near run
Jul 09 12:57:59.635  INFO near: Version: 1.19.1, Build: c8e49bb5, Latest Protocol: 45
Jul 09 12:57:59.760  INFO near: Opening store database at "/home/Aleksandr1/.near/data"
apply_store_migrations
Jul 09 12:57:59.781  INFO near: Migrate DB from version 22 to 23
```
- stop node, access store using separate script and ensure that receipts are saved in it
```
for receipt in receipts {
    eprintln!("{}", receipt.get_hash());
    chain_store.get_receipt(&receipt.get_hash()).unwrap().unwrap();
}
```


runtime/runtime/src/lib.rs
```diff
@@ -122,6 +122,8 @@ pub struct ApplyResult {
     pub state_changes: Vec<RawStateChangesWithTrieKey>,
     pub stats: ApplyStats,
     pub proof: Option<PartialStorage>,
+    /// Receipts restored during runtime apply after migration.
+    pub receipts_to_restore: Vec<Receipt>,
```
bowenwang1996, line 126: Why do we actually need this field? I thought we have receipts to restore as a global variable

chain/chain/src/chain.rs
```diff
@@ -3709,6 +3712,9 @@ impl<'a> ChainUpdate<'a> {
                 receipt_proof_response.1,
             );
         }
+        if !apply_result.receipts_to_restore.is_empty() {
+            self.chain_store_update.save_receipts(&apply_result.receipts_to_restore);
+        }
```
bowenwang1996, line 3717: I don't think we need this here. If we state sync past that specific block, we do not have the block in the first place and therefore it does not make sense for us to restore receipts here.

runtime/runtime/src/lib.rs
Longarithm: We do, but we need `chain_id` to decide to load it or not, which is stored on `Runtime` side.
And also at some moment in the past we decided that `Chain` should not know about `chain_id`, which discourages adding method like `get_chain_id` to `RuntimeAdapter`. So I leave `receipts_to_restore` computation to `Runtime`.

bowenwang1996: This is not a great solution. If possible, we should avoid introducing some field to `ApplyResult` only to use it once.

> also at some moment in the past we decided that Chain should not know about chain_id

I think it is more like we didn't see a good reason to do that. It does not mean it should never be done. Alternatively we can update it in `process_state_update` by writing directly to the database, which is also an ugly solution.

bowenwang1996 (commented):
Another option is that we can do a database migration to add those receipts

nearcore/src/migrations.rs
```diff
@@ -349,6 +349,20 @@ pub fn migrate_22_to_23(path: &String, near_config: &NearConfig) {
...
+/// Put receipts restored in scope of issue https://github.com/near/nearcore/pull/4248 to storage.
+pub fn migrate_23_to_24(path: &String, near_config: &NearConfig) {
+    let store = create_store(path);
+    if &near_config.genesis.config.chain_id == "mainnet" {
+        let genesis_height = near_config.genesis.config.genesis_height;
+        let mut chain_store = ChainStore::new(store.clone(), genesis_height);
+        let restored_receipts = serde_json::from_slice(&MAINNET_RESTORED_RECEIPTS)
+            .expect("File with receipts restored after apply_chunks fix have to be correct").get(&0u64);
+        let mut chain_store_update = ChainStoreUpdate::new(&mut chain_store);
+        chain_store_update.save_receipts(restored_receipts);
```
bowenwang1996, line 361: it is not committed

bowenwang1996 (changes requested):

Longarithm:
Changed implementation to DB migration.

Note: added receipts will be stored in the store forever for each mainnet node, because they don't belong to any chunk and their refcount always equals 1. This is not ideal, but as long as file with these receipts takes only 300 KB of memory, it doesn't look like a big issue.

nearcore/src/migrations.rs
```diff
@@ -358,7 +360,12 @@ pub fn migrate_23_to_24(path: &String, near_config: &NearConfig) {
         let restored_receipts = serde_json::from_slice(&MAINNET_RESTORED_RECEIPTS)
             .expect("File with receipts restored after apply_chunks fix have to be correct").get(&0u64);
         let mut chain_store_update = ChainStoreUpdate::new(&mut chain_store);
-        chain_store_update.save_receipts(restored_receipts);
+        let mut store_update = chain_store_update.store().store_update();
+        for receipt in receipts.iter() {
+            let bytes = receipt.try_to_vec().expect("Borsh cannot fail");
+            store_update.update_refcount(ColReceipts, receipt.get_hash().as_ref(), &bytes, 1);
+        }
+        store_update.commit().expect("");
```
bowenwang1996:
```suggestion
        store_update.commit().unwrap();
```

bowenwang1996: Also committing per value is a bit excessive. You can commit at the end since we know that there are not that many receipts ti begin with.

Longarithm: But I don't commit per value, commit is outside the loop.

nearcore/src/migrations.rs
```diff
@@ -349,6 +349,32 @@ pub fn migrate_22_to_23(path: &String, near_config: &NearConfig) {
...
+/// Put receipts restored in scope of issue https://github.com/near/nearcore/pull/4248 to storage.
+pub fn migrate_23_to_24(path: &String, near_config: &NearConfig) {
+    let store = create_store(path);
+    if &near_config.genesis.config.chain_id == "mainnet" {
+        let genesis_height = near_config.genesis.config.genesis_height;
+        let mut chain_store = ChainStore::new(store.clone(), genesis_height);
+        let restored_receipts: ReceiptResult = serde_json::from_slice(&MAINNET_RESTORED_RECEIPTS)
+            .expect("File with receipts restored after apply_chunks fix have to be correct");
+        let chain_store_update = ChainStoreUpdate::new(&mut chain_store);
+        let mut store_update = chain_store_update.store().store_update();
```
bowenwang1996, line 368: Why not construct a `StoreUpdate` directly from `store`?

Longarithm:
BTW it's a bit weird that `cargo fmt` changed files from runtime/runtime-params-estimator/src, because we have a test for it. 
cc @matklad?

bowenwang1996 (approved):

matklad (approved):

matklad:
filed https://github.com/near/nearcore/issues/4505 for me to fix the formatter.

frol:
@bowenwang1996 we should have not stabilize `protocol_feature_restore_receipts_after_fix` before fixing #4337! Backport this PR to 1.20.x since otherwise we will break all indexers on mainnet!

cc @khorolets

PR #4449, opened 2021-07-01
neard: handle SIGINT by stopping the node

mina86:
Stop the system once a SIGINT is received.  This should allow for
graceful termination since System::stop will stop all the arbiters
and that in turn will stop all the actors (leading them through
stopping and stopped states thus allowing all the necessary clean-ups).

To achieve this, also update uptade wasmer-runtime-core dependency to
0.17.4.  Among other things, the new version no longer catches the INT
signal making it available for tokio to handle.


neard/src/cli.rs
```diff
@@ -250,6 +250,9 @@ impl RunCmd {
         let sys = actix::System::new();
         sys.block_on(async move {
             nearcore::start_with_config(home_dir, near_config);
+            tokio::signal::ctrl_c().await.unwrap();
+            info!("Got Ctrl+C, stopping");
+            actix::System::current().stop();
```
matklad, line 255: do we have some existing ctrlc handling somewhere else? I *think* I saw related hacks somewhere else...

mina86: Only in tests as far as I can tell. Specifically `test-utils/actix-test-utils/src/lib.rs` and `pytest/lib/cluster.py`. I am looking into removing those. Cc: @miraclx

mina86: I’ve looked at `actix-test-utils` and it could be because my async-fu is insufficient but I couldn't figure out how to simplify code there. Since it’s test it’s probably fine even if the additional thread isn’t actually necessary.

bowenwang1996 (approved):

ailisp (approved):

neard/src/cli.rs
```diff
@@ -258,6 +258,9 @@ impl RunCmd {
...
         });
```
matklad, line 264: For posterity, while this is undoubtedly a big improvement, I think that this is not the right end-game logic. 

Proper shutdown should be a three-phase process.

* first, you *request* the system to cancel. 
* second, you *wait* for the system to actually stop.
* finally, if the system didn't stop within a time limit, you abruptly terminate it.

For networked services, such graceful shutdown means that all currently-pending requests are finished, but new connections and new requests are not accepted.

The `actix::System::current().stop();` I think goes directly to abrupt shutdown. So, if the node is in the middle of sending an HTTP response, it won't be send in full.

OTOH, there's an argument that, in a P2P byzantine network, graceful shutdown is a misfeature, as it adds a happy path, while the code should always be robust for all sad paths.

Issue #4450, opened 2021-07-01
Migration to introduce metadata in execution outcome fails


bowenwang1996:
Migration introduced in #4376 fails:
```
thread 'main' panicked at 'BorshDeserialize should not fail: Custom { kind: InvalidInput, error: "Unexpected variant index: 125" }', core/store/src/migrations.rs:846:14
```

To reproduce, remove the 22 -> 23 migration and apply this migration directly on top of archival data for mainnet. Feel free to sync with @chefsale on how to get the archival data for mainnet.

ailisp:
@bowenwang1996 my network is very unstable today, just have 1.3T mainnet state downloaded on a instance, almost unable to finish typing one command before ssh stucked. Will try next Monday

ailisp:
@bowenwang1996 Looks like the error is from an invalid ExecutionOutcome stored in mainnet. I'm not sure how to handle it, maybe **simply drop it** given it cannot be borsh deserialized with current definition of ExecutionOutcomeWithIdAndProof so it's junk?

Here is the problematic `Vec<ExecutionOutcomeWithIdAndProof>` by dumped by std::fs::write: 

https://github.com/ailisp/migration-fail-repro/blob/master/h6QEnvsiX5HcgEqKt5PNLKkm4ztZHM9BxYFmEhw6pxf

Not sure how many of these ExecutionOutcomes are corrupted. Guess it's not many because run for more than 20 minutes, got 1.5M execution outcomes, this is the first fail.

bowenwang1996:
@ailisp one way to fix it is to re-apply the block that generates the execution outcome. Please sync with @khorolets on how to address this so that indexer is not affected by the migration.

ailisp:
@khorolets in short, some execution outcomes were not migrated properly after a (still unknown) history migration and those execution outcomes cannot be borsh deserialized. I'm going to change these execution outcomes, and let you know which are them, and I think you can do a manually update on these execution outcomes on indexer side. Do you have a better plan?

khorolets:
> @khorolets in short, some execution outcomes were not migrated properly after a (still unknown) history migration and those execution outcomes cannot be borsh deserialized. I'm going to change these execution outcomes, and let you know which are them, and I think you can do a manually update on these execution outcomes on indexer side. Do you have a better plan?

Sounds legit, thank you!

bowenwang1996:
I believe this is already fixed

PR #4451, opened 2021-07-01
Revert "feat(primitive): prepare for adding profile data to execution…

bowenwang1996:
… outcome (#4376)"

This reverts commit 793178eddc29727f88ac61f6241d713e2748154d.

Revert this commit due to #4450 to unblock the release.


EgorKulikov (approved):

bowenwang1996:
Only db migration test fails and the test is expected to fail since the db version in the branch is older than that in master

PR #4452, opened 2021-07-01
tests: remove `near_actix_test_utils::run_actix_until_panic`

mina86:
The way `near_actix_test_utils::run_actix_until` was written,
the `expect_panic` flag didn’t actually matter:

    SET_PANIC_HOOK.call_once(|| {
        let default_hook = std::panic::take_hook();
        std::panic::set_hook(Box::new(move |info| {
            if !expect_panic {
                default_hook(info);
            }
            // ...
        }));
    });

Since `SET_PANIC_HOOK.call_once` invokes the closure only once, the
value of `expect_panic` when that calls happen is the only one that
matters.  In other words, the first run of `run_actix_until` function
decides what the value of `expect_panic` in the panic handler is.

Fortunately this didn’t actually matter.  The only test which set the
flag to true – `chunks_recovered_from_full_timeout_too_short` – was
marked `#[should_panic]` and running the default panic hook didn’t
negatively influence the test.

As such, get rid of `run_actix_until_panic` and rename
`run_actix_until_stop` to simply be `run_actix`.


mina86:
@miraclx

test-utils/actix-test-utils/src/lib.rs
```diff
@@ -64,21 +64,19 @@ pub fn spawn_interruptible<F: std::future::Future + 'static>(
...
     static SET_PANIC_HOOK: std::sync::Once = std::sync::Once::new();
 
     // This is a workaround to make actix/tokio runtime stop when a task panics.
     // See: https://github.com/actix/actix-net/issues/80
     SET_PANIC_HOOK.call_once(|| {
         let default_hook = std::panic::take_hook();
         std::panic::set_hook(Box::new(move |info| {
-            if !expect_panic {
-                default_hook(info);
-            }
```
miraclx, line 77: 😅 nice catch!

bowenwang1996 (approved):

khorolets (approved):

chefsale (approved):
LGTM updated to rebase onto master, as github automerge doesn't do it. @mina86 I would recommend to use the `S-automerge` label instead, as that bot also does automatocally rebase you PR.

mina86:
> I would recommend to use the S-automerge label instead, as that bot also does automatocally rebase you PR.

I suspected that is the case. Has been tested this with a few recent PRs. Thanks for confirming.

PR #4456, opened 2021-07-02
chore: modify stale bot configuration

bowenwang1996:
Change days until stale to be longer (sorry for the opposite suggestion earlier). Also @chefsale my experience so far is that it seems that we don't actually need the bot to automatically close the issue. What we want is for the bot to ping when an issue becomes stale so that we can do some manual triage. Is it possible to configure the bot that way?


chefsale:
I have added an additional commit which disables automatic closing of both issues and PRs. I hope this works, will land it.

chefsale (approved):

PR #4459, opened 2021-07-05
doc: add more documentation on jsonrpc in architecture.md

bowenwang1996:
Document on the high-level how rpc requests are processed.


docs/architecture.md
```diff
@@ -102,6 +102,9 @@ Note: `EpochManager` is constructed in `NightshadeRuntime` rather than in `Chain
 ### `chain/jsonrpc`
 
 This crate implements [JSON-RPC](https://www.jsonrpc.org/) API server to enable submission of new transactions and inspection of the blockchain data, the network state, and the node status.
+When a request is processed, it generates a message to either `Client` or `ViewClient` to interact with the blockchain.
+For queries of blockchain data, such as block, chunk, account, etc, the request usually generates a message to `ViewClient`.
+Transactions, on the other hand, are sent to `Client` for further processing.
```
matklad, line 107: This can benefit a bit from explaining what runs where and why we need to generate messages (as opposed to just calling a function). I think what we have is:

* an actix actor system, which runs `ClientActor` and `ViewClientActor`.
* an `actix::HttpServer` server, which is run on top of the actix actor system, but isn't an actor itself (IIRC, the `Addr` of the server isn't exposed; `ClientActor` doesn't send messages back to `HttpServer`, rather the server directly calls the client and gets back a result).

It probably also makes sense to write `ClientActor` rather than `Client`. It's also worth mentioning the entry points where we start all of the actors and wire them together.

matklad: A somewhat related thing, it might make sense to explain the timeout/loop/sleep pattern we use:

https://github.com/near/nearcore/blob/b7c8a3b5d2bfaeb1d689972fd867d1c10afcc440/chain/jsonrpc/src/lib.rs#L569-L590

frol (approved):
While I agree with what @matklad suggested, I feel that this PR is already a worth improvement and we can land it

bowenwang1996:
@matklad this PR requires your approval

matklad (approved):

matklad:
sorry, didn't mean to block this

PR #4463, opened 2021-07-06
cli: unix: catch SIGTERM in addition to SIGINT and terminate gracefully

mina86:
The ‘kill’ command sends SIGTERM signal by default.  Catch it in
addition ta SIGINT when running under Unix-like system.

### DIFF
--- a/neard/src/cli.rs
+++ b/neard/src/cli.rs
@@ -277,7 +277,7 @@ impl RunCmd {
                 tokio::signal::ctrl_c().await.unwrap();
                 "Ctrl+C"
             };
-            info!("Got {}, stopping", sig);
+            info!(target: "neard", "Got {}, stopping", sig);
             actix::System::current().stop();
         });
         sys.run().unwrap();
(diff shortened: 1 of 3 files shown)


neard/src/cli.rs
```diff
@@ -263,8 +264,20 @@ impl RunCmd {
...
+                let mut sigterm = signal(SignalKind::terminate()).unwrap();
+                futures::select! {
+                    _ = sigint .recv().fuse() => "SIGINT",
+                    _ = sigterm.recv().fuse() => "SIGTERM"
+                }
+            } else {
+                tokio::signal::ctrl_c().await.unwrap();
+                "Ctrl+C"
+            };
+            info!("Got {}, stopping", sig);
```
matklad, line 280:
```suggestion
            info!(target: "neard", "Got {}, stopping", sig);
```

we always specify the target everywhere else: https://github.com/near/nearcore/blob/master/docs/architecture.md#logging--observability

bowenwang1996 (approved):

Issue #4468, opened 2021-07-07
Refactor `RuntimeConfig` handling 


matklad:
We store `RuntimeConfig` in the NightShadeRuntime:

https://github.com/near/nearcore/blob/056f789e38c3e364cf1ae3db846499db518e32c5/nearcore/src/runtime/mod.rs#L132

The original idea seemed to be that this is basically a constant, but now we have two places where we actually modify config.

First, when we get config for specific protocol version:
 https://github.com/near/nearcore/blob/056f789e38c3e364cf1ae3db846499db518e32c5/core/primitives/src/runtime/config.rs#L58-L71

Second, when max_gas_burnt_view is set:

https://github.com/near/nearcore/blob/056f789e38c3e364cf1ae3db846499db518e32c5/nearcore/src/runtime/mod.rs#L157-L164

There are two problems here:

* The first code-path is subtly buggy -- it caches RuntimeConfig into a global variable, which might create dependence between different tests
* The data flow is unclear -- it's not obvious which config value is actually used when. For example, the logic [here](https://github.com/near/nearcore/blob/056f789e38c3e364cf1ae3db846499db518e32c5/nearcore/src/runtime/mod.rs#L1547-L1564) is quite subtle. 

To fix the first point, as suggested by @mina86 , we should just remove the lazy static can compute the config eagarly:

```rust
pub struct CurrentRuntimeConfig {
    inital_config: RuntimeConfig
    after_lowered_costs: RuntimeConfig
}

impl CurrentRuntimeConfig {
    pub fn  for_protocol_version(&self, pv: ProtocolVersion) -> &RuntimeConfig
}
```

Not sure what's the best solution to the second bullet point, I haven't looked super closely what happens there.

matklad:
cc https://github.com/near/nearcore/issues/1738

matklad:
@mina86 I wonder if you might be interested in picking this up -- you suggested much better solution than I had in #4263 :-)

mina86:
Yeah, I probably can take it. It shouldn’t be too much work.

Issue #4447, opened 2021-07-01
GenesisConfig should be separated from protocol config


bowenwang1996:
Currently when we calculate genesis json hash we serialize `GenesisConfig` as part of the hash computation. However, it is not really genesis config since we add new fields to `VMConfig` for example. This leads to change in genesis hash and cause the node to crash on this line [https://github.com/near/nearcore/blob/4e79467f35086b2ee58a1d4b41ceac09b7563409/nearcore/src/runtime/mod.rs#L295](https://github.com/near/nearcore/blob/4e79467f35086b2ee58a1d4b41ceac09b7563409/nearcore/src/runtime/mod.rs#L295). We may want to separate genesis config, which is fixed for a given network, from protocol config, which can be changed to address this issue.

frol:
@bowenwang1996 do you want to introduce `GenesisConfig` struct that matches the network or what? I don't believe it is feasible. I believe we should ensure that the key-values in the genesis.json file hashes to the same value as before (the order and formatting should not matter). `serde_json::Value` might be at a rescue.

frol:
Mental note: we will need to make some sort of migration once implementing this as otherwise, we will either discard the DBs with the older hash or DBs with the newer hash.

bowenwang1996:
> @bowenwang1996 do you want to introduce `GenesisConfig` struct that matches the network or what? I don't believe it is feasible. I believe we should ensure that the key-values in the genesis.json file hashes to the same value as before (the order and formatting should not matter). `serde_json::Value` might be at a rescue.

I don't have any concrete suggestion in mind and I'll let @posvyatokum figure this out. Yes we should not change how the existing genesis json is hashed, but rather we should introduce something that literally just hashes that file (without filling in new fields with default values) and check whether it matches what is in the database. And it does sound like `serde_json::Value` could be helpful here.

PR #4472, opened 2021-07-07
primitives: remove `LOWER_STORAGE_COST_CONFIG` global

mina86:
Protocol version changes may affect the runtime configuration.  To
handle this, the runtime uses a RuntimeConfig::from_protocol_version
method which returns the configuration adjusted for given version.
(Currently storage_amount_per_byte can be lowered by this mechanism).

However, it would be rather wasteful to construct a new adjusted
config object each time so the RuntimeConfig::from_protocol_version
method caches the result so that when it’s called next time for the
same (or newer) version it can return the config immediately.

The issue is, the cache has the form of a global singleton object
which is initialised once and never changed.  This means that calling
from_protocol_version on different config objects will return wrong
adjusted configs.

To fix the issue, introduce ActualRuntimeConfig (holding the original
config object as well as the cache which is now initialised eagerly)
and move the RuntimeConfig::from_protocol_version method to it.
With this the global cache is no longer used.

Fixes: https://github.com/near/nearcore/issues/4468

### DIFF
--- a/core/primitives/src/runtime/config.rs
+++ b/core/primitives/src/runtime/config.rs
@@ -5,9 +5,9 @@ use crate::checked_feature;
 use crate::config::VMConfig;
 use crate::runtime::fees::RuntimeFeesConfig;
 use crate::serialize::u128_dec_format;
-use crate::types::{AccountId, Balance};
+use crate::types::{AccountId, Balance, Gas};
 use crate::version::ProtocolVersion;
-use std::sync::{Arc, Mutex};
+use std::sync::Arc;
 
 /// The structure that holds the parameters of the runtime, mostly economics.
 #[derive(Debug, Serialize, Deserialize, Clone, PartialEq, Eq)]
@@ -38,10 +38,6 @@ impl Default for RuntimeConfig {
     }
 }
 
-lazy_static::lazy_static! {
-    static ref LOWER_STORAGE_COST_CONFIG: Mutex<Option<Arc<RuntimeConfig>>> = Mutex::new(None);
-}
-
 impl RuntimeConfig {
     pub fn free() -> Self {
         Self {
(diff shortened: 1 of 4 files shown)


nearcore/src/runtime/mod.rs
```diff
@@ -1547,19 +1536,15 @@ impl RuntimeAdapter for NightshadeRuntime {
         let protocol_version = self.get_epoch_protocol_version(epoch_id)?;
         let mut config = self.genesis_config.clone();
         config.protocol_version = protocol_version;
+        // If we were initialised with a custom max_gas_burnt_view (see
+        // ActualRuntimeConfig::new method), keep the value from genesis file.
+        // Since max_gas_burnt_view never changes as a result of protocol
+        // upgrade, we can use it when reporting protocol config.
+        let gas = config.runtime_config.wasm_config.limit_config.max_gas_burnt_view;
```
matklad, line 1543: This logic still feels like a smell to me, but I still don't see simple, short improvement. 

I guess, the question I have (going by the naming mostly): why is  `max_gas_burnt_view` is a part of `ProtocolConfig` at all? As far as I understand, view functionality exists outside of the protocol, so perhaps it shouldn't be part of ProtocolConfig?

To ask this in another way: here, we do some logic to get the `max_gas_burnt_view` from the original genesis_config. But what does that number mean? Who uses `max_gas_burnt_view`, as it can be configured separately by each node?

matklad (approved):
LGTM as is, but I am still confused about big picture with respect to max_gas_burnt_view

nearcore/src/runtime/mod.rs
matklad: Re-reading https://github.com/near/nearcore/issues/3080, it does seem that we want to remove the field from GenesisConfig (or at least make it unused). @bowenwang1996 , if that's roughly correct, could you create an issue for that refactor? I am not feeling good about having two sources of truth for `max_gas_burnt_view` -- it won't cause bugs, but it is tech debt.

mina86: > To ask this in another way: here, we do some logic to get the max_gas_burnt_view from the original genesis_config. But what does that number mean?

It means nothing at all. ;) At this point it’s essentially the default value for the limit which is used if client doesn’t override it. In fact we *probably* could return the modified value here, but having different client return different ‘protocol config’ does feel weird as well. (For some more context, this value is returned by `EXPERIMENTAL_protocol_config` RPC).

I think part of the issue is that this value is part of the genesis hash, though I might be misunderstandingn how hashing works.

core/primitives/src/runtime/config.rs
```diff
@@ -51,30 +47,54 @@ impl RuntimeConfig {
...
+    pub fn new(genesis_runtime_config: RuntimeConfig, max_gas_burnt_view: Option<Gas>) -> Self {
+        let mut config = genesis_runtime_config;
+        if let Some(gas) = max_gas_burnt_view {
+            config.wasm_config.limit_config.max_gas_burnt_view = gas;
+        }
+        let runtime_config = Arc::new(config.clone());
+
+        // Adjust as per LowerStorageCost protocol feature.
+        config.storage_amount_per_byte = 10u128.pow(19);
+        let with_lower_storage_cost = Arc::new(config);
```
bowenwang1996, line 81: We cannot do this and need to preserve the original semantics. Otherwise some old blocks won't be able to be applied

nearcore/src/runtime/mod.rs
bowenwang1996: @matklad I created https://github.com/near/nearcore/issues/4447 and @posvyatokum is working on that

bowenwang1996 (changes requested):

core/primitives/src/runtime/config.rs
mina86: The lower storage is only used in protocol version which enabled it. See `for_protocol_version` function.

mina86:
Sorry, deleted the branch by mistake.

core/primitives/src/runtime/config.rs
```diff
@@ -51,30 +47,54 @@ impl RuntimeConfig {
...
+/// but i) may have it’s `max_gas_burnt_view` limit adjusted and ii) provides
+/// a method which returns configuration with adjustments done through protocol
+/// version upgrades.
+pub struct ActualRuntimeConfig {
+    /// The runtime configuration taken from the genesis file but with possibly
+    /// modified `max_gas_burnt_view` limit.
+    runtime_config: Arc<RuntimeConfig>,
+
+    /// The runtime configuration with lower storage cost adjustment applied.
+    with_lower_storage_cost: Arc<RuntimeConfig>,
```
bowenwang1996, line 64: Does this mean that every time we change the runtime config through a protocol change we will need to introduce a new field? Is that a desirable way to do it?

mina86: > Does this mean that every time we change the runtime config through a protocol change we will need to introduce a new field?

Yes. Note that this commit doesn’t change this. This is the same behaviour as before.

> Is that a desirable way to do it?

Maybe not, but if we want to have ability to get the configuration for any possible protocol version, we either need to store the configs or build the configs on the fly each time a config is requested. If we want to change how this is handled I can certainly do it but in a separate commit since scope of this one is only to get rid of the global.

bowenwang1996 (approved):

PR #4481, opened 2021-07-08
refactor: remove needless rayon pool setup in the estimator

matklad:
Forgot to do this in #4293: the original motivation here was to force
wasmer1 to compile code in a single thread,  but we've since contributed
a "no rayon" mode upstream and upgraded to that.

In other words, wasmer1 no longer uses threads, so we don't need this.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4892,7 +4892,6 @@ dependencies = [
  "once_cell",
  "rand 0.7.3",
  "rand_xorshift 0.2.0",
- "rayon",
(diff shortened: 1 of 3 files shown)


bowenwang1996:
> the original motivation here was to force
wasmer1 to compile code in a single thread, but

@matklad what is after "but" :)

matklad:
but I got distracted apparently, fixed!

olonho (approved):

PR #4484, opened 2021-07-08
chore: delete all unused/obsolete/deprecated script in the scripts dir

chefsale:


bowenwang1996 (approved):

bowenwang1996 (changes requested):
Actually I think people still use start_localnet.py

matklad:
Bit :+1: from me. If there are some scripts that people do use, might be a good idea to document that somewhere. For example, I didn't know that `start_localnet.py` is a thing, but it sounds like something I actually might want to use sometimes?

ailisp:
>  For example, I didn't know that start_localnet.py is a thing, but it sounds like something I actually might want to use sometimes?

I didn't find start_localnet.py useful any more. If you want to start a localnet, `target/debug/neard --home near-dir init` then `neard --home near-dir run` works faster

chefsale:
@bowenwang1996 I agree with @ailisp, i don't think anymore is using start localnet anymore, as it is much simpler just to do an empty init and the localnet is setup eitherways.

mhalambek (approved):

bowenwang1996:
hmm I still use start_localnet.py sometimes, but I am also fine with removing it

chefsale:
@bowenwang1996 In that case I have just returned the start_localnet.py script, removed the other ones. Can you please approve?

matklad (approved):

bowenwang1996 (approved):

PR #4488, opened 2021-07-09
refactor: switch to pure-rust wasm parser for test

### LINKED ISSUE #4471: Switch from wabt to wat
We currently use `wabt` to parse wasm textual format into binary form. We should switch to wat instead: https://github.com/bytecodealliance/wasm-tools/tree/main/crates/wat

wabt is a binding to a C++ library, it complicates the build. Additionally, the version we currently use is buggy (can segfault or overflow the stack), and we can't upgrade that, because the new version requires newer CMake. 

wat is a pure-rust soulution by bytecodeallience, so it's high quality and dependable.

matklad:
This substitutes wat (pure Rust impl from bytecode alliance) for wabt
(bindings to C++ lib) to parse wasm s-expresisons to binary form.

`wat` doesn't do validation, only syntax checking, that's why
`.validate(false)` calls are just removed.

`wat` preserves the debug names section, that's why we chage burnt_gas
once

https://webassembly.github.io/spec/core/appendix/custom.html#name-section

See #4471 for motivation

closes #4471


bowenwang1996 (commented):
How does this affect gas costs?

ailisp (approved):
Looks good. My concern is since wat doesn't validate, should we test more syntax correct, wabt not validated contracts to wasmer? Can we find a few from fuzzing?

matklad:
@bowenwang1996  It doesn't affect gas costs: it's in the dev-deps for the vm-runner, and in the estimator it is used (and should be used) outised of the measured segments. 

@ailisp yeah, I also was surprised at the difference in validation. Thinking more about this, wat behavior (not validating) seems more reasonable and more useful for us. 

It's more reasonable because semantic validation ("type checking" of the stack discipline) is distinct from parsing, and belongs to the separate layer. It' doesn't make much sense to run validation to return just `Vec<u8>`, which erases the info that validation has been applied. 

It is more useful for us, because, *during testing* we specifically care about behavior of the runtime on invalid contracts. Note that we only use wat/wabt during tests (this is statically enforced by them being a dev dep) -- in production, it's the VM (wasmer) which enforces validation.

olonho (approved):

PR #4491, opened 2021-07-09
Add cython to pytest requirements

Longarithm:
I suggest to add `cython` because building venv for pytest from scratch always ends up with

```
    Running setup.py install for scikit-learn ... error
    ERROR: Command errored out with exit status 1:
     command: /home/Aleksandr1/nearcore/venv/bin/python -u -c 'import sys, setuptools, tokenize; sys.argv[0] = '"'"'/tmp/pip-install-cak7r1gr/scikit-learn/setup.py'"'"'; __file__='"'"'/tmp/pip-install-cak7r1gr/scikit-learn/setup.py'"'"';f=getattr(tokenize, '"'"'open'"'"', open)(__file__);code=f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' install --record /tmp/pip-record-z15d8knj/install-record.txt --single-version-externally-managed --compile --install-headers /home/Aleksandr1/nearcore/venv/include/site/python3.8/scikit-learn
         cwd: /tmp/pip-install-cak7r1gr/scikit-learn/
...
      File "sklearn/utils/setup.py", line 8, in configuration
        from Cython import Tempita
    ModuleNotFoundError: No module named 'Cython'
    ----------------------------------------
```

### DIFF
--- a/pytest/requirements.txt
+++ b/pytest/requirements.txt
@@ -11,6 +11,7 @@ python-rc==0.3.9
 tqdm
 deepdiff
 pynacl
+cython
 numpy
 scikit-learn==0.21.3
 PyGithub


matklad:
Semi related question: why do we have both requirenments.txt and Pipfile?

bowenwang1996:
Not sure why we even depend on scikit-learn. That feels very unnecessary to me

mina86:
> Semi related question: why do we have both requirenments.txt and Pipfile?

Dunno. But `requirements.txt` is what NayDuck uses. (And it does it in a very peculiar way to be honest).

> Not sure why we even depend on scikit-learn. That feels very unnecessary to me

It’s what provides `sklearn` as far as I understand:

```
pytest/lib/data.py:2:from sklearn.linear_model import LinearRegression
```

We’re using it for it’s linear regression code which is then used to measure performance of mocknet if I understand the code correctly.

mina86 (approved):
LGTM

Issue #4480, opened 2021-07-08
Non-reproducible tests due to pytest/empty-contract-rs


matklad:
We have this "stub" contract: 

https://github.com/near/nearcore/blob/3ac2b829f019372d69c9e98b29a5ccdf89bf6d5d/pytest/empty-contract-rs/

It is used primarily by this function:

https://github.com/near/nearcore/blob/c85bf403909018e24c7a83a20b9bbf33958a16c9/pytest/lib/utils.py#L192-L211

The problem with this empty contract is that it depends on master branch of the sdk repo *without lockfile*:

https://github.com/near/nearcore/blob/c85bf403909018e24c7a83a20b9bbf33958a16c9/pytest/empty-contract-rs/Cargo.toml#L11

So, every time we run a test, we get what's currently the latest commit in master, meaning that test actually run different contracts. 

I suggest to, at minimum, pin the version in Cargo.toml to a specific commit using the `rev = ` Cargo.toml key. 

Perhaps a better solution is to move the tests currently using empty-contract to use the standard testing contracts instead, a-la: https://github.com/near/nearcore/issues/4475. I don't think it makes a ton of sense that `nearcore` depends on `near-sdk-rs` for testing (nearcore ships SDK-independent ABI), and that create extra complexity (need to track sdk changes, more time spent compiling sdk-enabled contrcts, which include heavy proc-macro machinery).

matklad:
Looking at the git history of this test contract, looks like this issue suggests to revert the decision in https://github.com/near/nearcore/pull/1947. 

@nearmax, I wonder what are your thoughts here, should we, or should we not use `near-sdk-rs` for pytests?

I think your main point is that it's easier to use sdk for testing complex functionality like cross-shard calls.

That's true, but we also didn't add a lot of sdk-dependent tests in a year, so perhaps we don't need this just yet, and should re-visit once we need to add such tests? My gut feeling would be that it's still beneficial to write those tests on top of our bare-bones sdk-less test contract -- that way it's easier for nearcore devs to understand what exactly the contract is doing, as sdk adds quite a bit of abstraction and indirection.

mina86:
At this point there's only one remaining test which uses near-sdk-rs. It uses it to build a trivial busy work function. It’s easiest to just not depend on the near-sdk IMO. The only thing to consider is whether we want to test near-sdk as part of the tests but we probably don’t want to do that.

mina86:
We no longer depend on `near-sdk`.

PR #4493, opened 2021-07-09
pytest: don’t depend on near-sdk in empty-contract-rs

mina86:
There’s very little reason to make the empty-contract-rs depend on
near-sdk.  The code is trivial and only uses the log method which
can be easily called directly without the near-sdk wrapper.

This also fixes the issue where the test was not reproducible as the
empty-contract-rs would depend on the near-sdk from the git
repository rather than a released crate.

An obvious question which arises is why not use near-test-contracts
like the rest of pytests.  That would be an ideal endgoal but it
doesn’t interact well with how NayDuck runs mocknet tests.
Specifically, near-test-contracts needs to be built and NayDuck
currently skips build stage when running mocknet tests.  Long term
this could be addressed but for now this change improves the
situation so it’s what we’ve got.

Issue: https://github.com/near/nearcore/issues/4480

### DIFF
--- a/pytest/lib/utils.py
+++ b/pytest/lib/utils.py
@@ -1,15 +1,21 @@
-from transaction import sign_payment_tx
-import random, base58
+import base58
+import hashlib
+import json
+import os
+import pathlib
+import random
+import shutil
+import subprocess
+import sys
+import tempfile
+import time
+
 from retrying import retry
+from rc import gcloud
+
 from cluster import LocalNode, GCloudNode, CONFIG_ENV_VAR
 from configured_logger import logger
-import sys
-from rc import run, gcloud
-import os
-import tempfile
-import json
-import hashlib
-import time
+from transaction import sign_payment_tx
 
 
 class TxContext:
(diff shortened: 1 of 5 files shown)


pytest/lib/utils.py
```diff
@@ -190,24 +196,27 @@ def load_test_contract(filename='test_contract_rs.wasm'):
...
+        (build_dir / 'src').mkdir(parents=True, exist_ok=True)
+        with open(build_dir / 'src' / 'lib.rs', 'w') as wr:
+            wr.write(content)
+        subprocess.check_call(('cargo', 'build', '--release', '--target',
+                               'wasm32-unknown-unknown'), cwd=build_dir)
+        wasm_src = (build_dir / 'target' / 'wasm32-unknown-unknown' /
+                    'release' / 'empty_contract_rs.wasm')
+        # Yes, we are leaving this file in /tmp/near and there’s no automatic
+        # cleanup.  Yes, we probably should fix that at some point so that this
+        # temporary is deleted once the test finishes.
```
bowenwang1996, line 215: Should we create an issue for this?

bowenwang1996 (approved):

pytest/lib/utils.py
mina86: https://github.com/near/nearcore/pull/4494

PR #4495, opened 2021-07-12
Add a flag to run a separate http server for exporting Prometheus metric on a different address

### LINKED ISSUE #2693: Provide flag to expose Prometheus metrics on custom port and listening address
Currently Prometheus metrics are exposed on the same port as RPC. It would be useful to be able to start the validator with a flag that could specify a custom port and listening address to expose the metrics on. This is because it is good practice to tightly restrict the RPC port to the local host but reasonable to have the Prometheus metrics accessible remotely, so they can be ingested by a Prometheus server.

Being able to specify the listening address as well (eg. 0.0.0.0 to listen on any address) would be useful (though not as important), as some validators run their nodes on systems with multiple network interfaces and may wish to only have the metrics available on one of these

nikurt:
Tested the change manually by adding a flag `--rpc-monitoring-addr=127.0.0.1:3041` and by changing `~/.near/config.json`. Verified that /metrics exports the same values on both running http servers.
The second server only starts if `--rpc-monitoring-addr` is provided and is different from `--rpc-addr`.

Fixes #2693


chain/jsonrpc/src/lib.rs
```diff
@@ -67,6 +66,8 @@ impl Default for RpcLimitsConfig {
 #[derive(Serialize, Deserialize, Clone, Debug)]
 pub struct RpcConfig {
     pub addr: String,
+    // If provided, will start an http server exporting only Prometheus metrics on that address.
+    pub monitoring_addr: Option<String>,
```
chefsale, line 70: Can we change the variable and the flag to be `prometheus` rather than `monitoring`

neard/src/cli.rs
```diff
@@ -190,6 +190,12 @@ pub(super) struct RunCmd {
     #[cfg(feature = "json_rpc")]
     #[clap(long)]
     rpc_addr: Option<String>,
+    /// Export prometheus metrics on an additional listening address, which is useful
+    /// for having separate access restrictions for the RPC and prometheus endpoints.
+    /// Ignored if RPC http server is disabled, see 'rpc_addr'.
+    #[cfg(feature = "json_rpc")]
+    #[clap(long)]
+    rpc_monitoring_addr: Option<String>,
```
chefsale, line 198: Please change to prometheus here as well.

neard/src/cli.rs
```diff
@@ -233,8 +239,14 @@ impl RunCmd {
         #[cfg(feature = "json_rpc")]
         if self.disable_rpc {
             near_config.rpc_config = None;
-        } else if let Some(rpc_addr) = self.rpc_addr {
-            near_config.rpc_config.get_or_insert(Default::default()).addr = rpc_addr;
+        } else {
+            if let Some(rpc_addr) = self.rpc_addr {
+                near_config.rpc_config.get_or_insert(Default::default()).addr = rpc_addr;
+            }
+            if let Some(rpc_monitoring_addr) = self.rpc_monitoring_addr {
```
chefsale, line 246: Please change to prometheus here as well.

nikurt (commented):
Renamed in the second file as well.

chefsale (approved):
LGTM, will leave to @frol and @matklad to take another look.

chain/jsonrpc/src/lib.rs
```diff
@@ -1216,7 +1208,10 @@ pub fn start_http(
     client_addr: Addr<ClientActor>,
     view_client_addr: Addr<ViewClientActor>,
 ) {
-    let RpcConfig { addr, cors_allowed_origins, polling_config, limits_config } = config;
+    let RpcConfig { addr, prometheus_addr, cors_allowed_origins, polling_config, limits_config } =
+        config;
+    let prometheus_addr = prometheus_addr.filter(|it| it != &addr);
+    let cors_allowed_origins_clone = cors_allowed_origins.clone();
     info!(target:"network", "Starting http server at {}", addr);
```
bowenwang1996, line 1215: not sure why target is `network`. cc @frol

bowenwang1996 (approved):
LGTM. @frol please take a look

matklad (approved):

PR #4497, opened 2021-07-12
Runtime tester

posvyatokum:
Adding bare minimum to run tests for long block production. 
Added json Serialize/Deserialize for scenario. 
Creating Scenario from code is still cumbersome.
The next step is basic fuzzing implementation on top of this. This will require fn run to return Result with fuzzing Error, and therefore run_from_json to return custom error.


bowenwang1996 (approved):
looks good to me. @matklad please take a look

test-utils/runtime-tester/src/run_test.rs
```diff
@@ -0,0 +1,127 @@
...
+use near_primitives::transaction::{Action, SignedTransaction};
+use near_primitives::types::{BlockHeight, Nonce};
+use near_store::test_utils::create_test_store;
+use nearcore::{config::GenesisExt, NightshadeRuntime};
+
+use serde::{Deserialize, Serialize};
+
+pub fn run_from_json(path: &str) -> Result<RuntimeStats, io::Error> {
+    let reader = io::BufReader::new(File::open(path)?);
+    let scenario: Scenario = serde_json::from_reader::<io::BufReader<File>, Scenario>(reader)?;
```
matklad, line 20: I'd do just `serde_json::from_str(std::fs::read_to_string)`. Not sure if it's still the case (probaby is), but reading everything to memeory used to be significantly faster than using `from_reader`

test-utils/runtime-tester/src/run_test.rs
```diff
@@ -0,0 +1,127 @@
...
+use near_client::test_utils::TestEnv;
+use near_crypto::InMemorySigner;
+use near_primitives::transaction::{Action, SignedTransaction};
+use near_primitives::types::{BlockHeight, Nonce};
+use near_store::test_utils::create_test_store;
+use nearcore::{config::GenesisExt, NightshadeRuntime};
+
+use serde::{Deserialize, Serialize};
+
+pub fn run_from_json(path: &str) -> Result<RuntimeStats, io::Error> {
```
matklad, line 18: Nit:

```suggestion
pub fn run_from_json(path: &str) -> io::Result<Error> {
```

matklad, line 18: Slightly more than a nit: let's keep the types strong! 

```suggestion
pub fn run_from_json(path: &Path) -> Result<RuntimeStats, io::Error> {
```

test-utils/runtime-tester/Cargo.toml
```diff
@@ -0,0 +1,18 @@
+[package]
+name = "runtime-tester"
```
matklad, line 2: Seems like we need to add this as a `member` to the top-level Cargo.toml.

test-utils/runtime-tester/src/run_test.rs
```diff
@@ -0,0 +1,127 @@
...
+
+    runtime_stats
+}
+
+#[derive(Serialize, Deserialize)]
+pub struct Scenario {
+    pub network_config: NetworkConfig,
+    pub blocks: Vec<BlockConfig>,
+}
+
```
matklad, line 72: Let's use the following API:

```
impl Scenario {
    pub fn from_file(path: &Path) -> io::Resul<zc>;
    pub fn run(&self) -> RuntimeStats;
}
```

That way the user needs to import only `Scenario` and get the rest from auto-complition, instead of having two functions to import. An additional benefit is that it's now ergonomic to call from_file and run in a row:

```rust
Scenario::from_file("s.json")?
    .run()
```

That's  better than the current `run_from_json(path: &str) -> io::Result<Error>`, because it guarantees that no io errors can happen during the run itselef.

test-utils/runtime-tester/src/lib.rs
```diff
@@ -0,0 +1,2 @@
+pub mod run_test;
```
matklad, line 1: Let's add a short doc comment explaining what this crate is about. 

Right now, if an new person comes to the codbases and sees the code, it'd be difficult for them to understand what's this is for.

matklad (commented):
Looks great to me, left a few rust-style nits. 

I have two hight-level questions:

* is JSON the best format here? Should we use faster and more compact borsh from the start? I think it probably is better to start with JSON, but curious if you have other thoughts here. 
* Does making `InMemorySigner` `Serialize` makes sense? (this might be a very stupid question). If we need `InMemorySigner`, than we need to know secret keys, but seems that we can't know secret keys for other nodes? Should be perhaps serialize already signed transactions?

chefsale:
Can we please rebase this branch onto master as now we have a bunch of cherry picked commits in this branch and the files changed seem to include changes which shouldn't be there for some reason.

posvyatokum:
> * is JSON the best format here? Should we use faster and more compact borsh from the start? I think it probably is better to start with JSON, but curious if you have other thoughts here.
It is easier to debug with JSON, so I started with it. 

> * Does making `InMemorySigner` `Serialize` makes sense? (this might be a very stupid question). If we need `InMemorySigner`, than we need to know secret keys, but seems that we can't know secret keys for other nodes? Should be perhaps serialize already signed transactions?
I didn't want to include 'block_hash' in Scenario. I've just seen that in tests the hash of the last produced block is passed to this variable. In this case, we need to run transactions before creating a scenario, thus defeating the purpose of this framework. Also serialized scenario files will be invalid after protocol change.

test-utils/runtime-tester/src/lib.rs
```diff
@@ -0,0 +1,6 @@
+/// run_test is a framework for creating and executing runtime scenarios.
+/// You can create Scenario in rust code or have it in a JSON file.
+/// Scenario::run executes scenario, keeping track of different metrics.
+/// So far, the only metric is how much time block production takes.
```
matklad, line 4:
```suggestion
//! run_test is a framework for creating and executing runtime scenarios.
//! You can create Scenario in rust code or have it in a JSON file.
//! Scenario::run executes scenario, keeping track of different metrics.
//! So far, the only metric is how much time block production takes.
```

With `///`a this would be a doc-comment for the following item, the `run_test` module. With `//!`, this will be a comment for the current item, the whole crate.

matklad (approved):
`///` to `//!` would be a good change, but LGTM regardless!

chefsale (approved):
LgTm

PR #4498, opened 2021-07-12
Fix order of regression constants

Longarithm:
According to `least_squares_method` formulas, `a` stands for constant and `b` stands for multiplier. Other usages of `compute_compile_cost_vm` have correct order.


bowenwang1996 (commented):
@Longarithm what is the new cost after this fix?

Longarithm:
@bowenwang1996 it should affect only cost of contract deployment and made it same as expected, 2.5 TGas + 470 Mgas * num_bytes. It also matches the result of `test_compile_cost_icount`

olonho (approved):

Issue #4430, opened 2021-06-29
Nayduck build is broken


bowenwang1996:
All the adversarial tests fail with something like
```
AssertionError: {'jsonrpc': '2.0', 'error': {'name': 'REQUEST_VALIDATION_ERROR', 'cause': {'name': 'METHOD_NOT_FOUND', 'info': {'method_name': 'adv_disable_doomslug'}}, 'code': -32601, 'message': 'Method not found', 'data': 'adv_disable_doomslug'}, 'id': 'dontcare'}
```
which seems to indicate that the build is broken. This might be a result of the nearcore crate refactoring we did recently.

mina86:
This has been broken for a while http://nayduck.eastus.cloudapp.azure.com:3000/#/test/160516 so it must be caused by something else. In general this looks like the tests using non-adversarial build of the node.

PR #4499, opened 2021-07-12
fix(test): fix adversarial build of neard

bowenwang1996:
Adversarial build is broken because the feature does not include rpc, which causes some nightly tests to fail.

Test plan
-----------
build neard with `cargo build -p neard --features adversarial` and run some adversarial tests.

Fixes: https://github.com/near/nearcore/issues/4430


mina86 (commented):
LGTM. I thought that was an issue but then wasn’t able to reproduce the problem locally. There might be something different about NayDuck and local builds perhaps.s

frol (approved):

PR #4500, opened 2021-07-12
fix(test): fix nayduck nearcore expensive tests

bowenwang1996:
Some expensive tests are broken because 1) they are moved to `integration-tests` crate and 2) there is no `expensive_tests` feature in that crate.


mina86 (commented):
LGTM

mina86 (approved):

PR #4502, opened 2021-07-13
Report the earliest available block in the status response

### LINKED ISSUE #3750: Feature Request: Report the earliest available block in the status response
When running a non-archive node its hard to know the earliest block available in the node since old block are GCed at some point.

Here's what the status response might look like: 

```json
{
  "jsonrpc": "2.0",
  "result": {
    "version": {
      "version": "1.17.0-rc.2",
      "build": "46b02963"
    },
    "chain_id": "testnet",
    "protocol_version": 41,
    "latest_protocol_version": 41,
    "rpc_addr": "0.0.0.0:3030",
    "validators": [],
    "sync_info": {
      "first_block_hash": "...",
      "first_block_height": 11111111,
      "first_block_time": "...",

      "latest_block_hash": "BdvayetRTppNmpzuccHod7DjjwPDkPKD2fEy3miiKqY6",
      "latest_block_height": 28596082,
      "latest_state_root": "3v9HcsnaVowHun9UZYbnWgTXH2BWSkC2Vx6HCEiCa647",
      "latest_block_time": "2020-12-17T16:58:00.945712173Z",
      "syncing": false
    },
    "validator_account_id": "nearup-node10"
  },
  "id": "dontcare"
}
```

So having `first_block_*` (or `oldest_block_*`) fields inside the `sync_info` attribute will be super useful!

nikurt:
Tested on `localnet` and `betanet`.
Fixes #3750.

`sync_info` part of the `/status` response on `betanet`:
```
  "sync_info": {
    "latest_block_hash": "DnQTTEUKpyc4Pj8ym6gtSaSiWyNpQLq4N4WyzEQjTRUU",
    "latest_block_height": 30171782,
    "latest_state_root": "7rMsXJ7XVfsuAYpyZ3v45eXkeAtuYNkvPX4StB4WMDxy",
    "latest_block_time": "2021-07-13T09:54:46.381635593Z",
    "syncing": true,
    "earliest_block_hash": "9Nf7ZaNSoFDtQVEnDncyx5dgZ4aSBs6EopBfmNP4xz7t",
    "earliest_block_height": 30170194,
    "earliest_block_time": "2021-07-13T09:38:33.866737622Z"
  },
```


chain/client/src/client_actor.rs
```diff
@@ -557,8 +557,9 @@ impl Handler<Status> for ClientActor {
         self.check_triggers(ctx);
 
         let head = self.client.chain.head()?;
-        let header = self.client.chain.get_block_header(&head.last_block_hash)?;
-        let latest_block_time = header.raw_timestamp().clone();
+        let head_header = self.client.chain.get_block_header(&head.last_block_hash)?;
+        let latest_block_time = head_header.raw_timestamp().clone();
```
bowenwang1996, line 561:
```suggestion
        let latest_block_time = head_header.raw_timestamp();
```

core/primitives/src/views.rs
```diff
@@ -297,6 +297,9 @@ pub struct StatusSyncInfo {
     pub latest_state_root: CryptoHash,
     pub latest_block_time: DateTime<Utc>,
     pub syncing: bool,
+    pub earliest_block_hash: Option<CryptoHash>,
+    pub earliest_block_height: Option<BlockHeight>,
+    pub earliest_block_time: Option<DateTime<Utc>>,
```
bowenwang1996, line 302: We probably don't need to have these two because depending on whether the node is archival, the earliest block can be very different whereas latest block is likely the same across different types of nodes. There is, in some sense, no point in exposing earliest block height and earliest block time in the response. They can be queried easily if one knows the earliest block hash.

nikurt: I think exposing these two fields slightly improves the usability. If a user has more than 1 node, querying the block hashes can be frustrating.

bowenwang1996 (approved):

PR #4504, opened 2021-07-13
feat(build): Added param-estimator and genesis-populate to the Makefile for nightly and stable releases to store artifacts

chefsale:
@Longarithm has asked me if we can build the binaries for `param-estimator` and `genesis-populate` in the current pipeline as currently he is using a Docker image which recompiles them every time and this would save him time.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,12 +11,16 @@ release:
 	cargo build -p near-vm-runner-standalone --release
 	cargo build -p state-viewer --release
 	cargo build -p store-validator --release
+	cargo build -p runtime-param-estimator --release
+	cargo build -p genesis-populate --release
 
 debug:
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -11,12 +11,16 @@ release:
 	cargo build -p near-vm-runner-standalone --release
 	cargo build -p state-viewer --release
 	cargo build -p store-validator --release
+	cargo build -p param-estimator --release
```
Longarithm, line 14:
```suggestion
	cargo build -p runtime-params-estimator --release
```
Sorry I told you incorrect package name, could you replace it here and below?

chefsale: updated.

Longarithm (approved):

Issue #4403, opened 2021-06-23
Consider disabling lto in Cargo.toml


matklad:
Apparently, this resurrects https://github.com/near/near-ops/issues/169

We currently set `lto = true`, `codegen-units=1` in Cargo.toml:

https://github.com/near/nearcore/blob/cb229d06595ebf01bf9e39d43f11a7478c92664e/Cargo.toml#L57-L58

This **massively** increases compile times (and compile-time memory usage) in release mode. This in turn hurts productivity for cases where you work on some benchmark and want to recompile the code repeatedly. 

The work-around is to 

```bash
export CARGO_PROFILE_RELEASE_LTO=false
```

locally.

The reason why we set `lto=true` is that it does improve performance of release. 

I think the best practice in this area is to optimize the default configuration for developer convenience, and making the release process opt into additional optimizations. At the same time, it's important that a developer can easily reproduce the exact configuration we use for release. So I suggest:

* _First_, to move the script that prepares release binaries into this repo. It doesn't have to be the full release & deploy pipeline, but the result should be that a developer  can run `./.buildkite/build-release-binaries.sh` and get the bins compiled as for release. as a side benefit, a developer can look into `build-release-binaries.sh` and understand what we are we actually releasing (I, for example, don't know what constitutes a `nearcore` release).
* _Second_, we modify this script to explicitely enable LTO, remove lto and codegen units config from Cargo.toml. 

Checklist 
---------
- [x] Fix our existing makefile: https://github.com/near/nearcore/issues/4475
- [x] Add `export CARGO_PROFILE_RELEASE_LTO=fat`, `export CARGO_PROFILE_RELEASE_CODEGEN_UNITS=1` to the Makefiles
- [ ] Remove corresponding entires [from cargo-toml](https://github.com/near/nearcore/blob/ff01c544ce2675fadbf0284af52b356a383b27a8/Cargo.toml#L57-L58).
- [x] Verify that release is actually build with LTO,
- [x] Verify that removing LTO actually speeds up local `cargo build --release -p neard`.
- [x] Verify that the `Makefile` is the only place which produces release builds. Perhaps there are some nayduck load tests or something which also require a true release binary?

frol:
My concern is that I would prefer to have the same production build by default when I do just `cargo build --package neard --release`. I configured my local env to speed compilation up on my side, but I don't think it is sane to break defaults for release builds in order to optimize development use-case.

Still, I am fine with the proposed solution.

mina86:
[...]

PR #4506, opened 2021-07-13
Makefile: make sure LTO is enable in release builds

mina86:
Regardless of how the release profile is configured in Cargo.toml,
make sure that release binaries built through the makefile have
LTO enabled.  This will allow disabling LTO in Cargo.toml (and thus
speed up regular developement builds) without negatively affecting
officially released binaries.

To that end, add CARGO_PROFILE_RELEASE_{CODEGEN_UNITS,LTO} environment
variables to Makefile which will override Cargo.toml configuration when
cargo is run.

While at it, clean up some other environment variables such that all
of them are set at the top of the file.


chefsale (approved):
I assume this should be fine. cc: @frol @matklad

PR #4507, opened 2021-07-13
Disable LTO in release builds

mina86:
Note: This change makes release builds less optimised by default.  If
you’re building the executable yourself and use them in production to
get the same level of optimisation as before this commit make sure to
enable LTO via environment variable:

    CARGO_PROFILE_RELEASE_LTO=fat
    CARGO_PROFILE_RELEASE_CODEGEN_UNITS=1
    export CARGO_PROFILE_RELEASE_LTO CARGO_PROFILE_RELEASE_CODEGEN_UNITS

Alternatively, build the binaries via `make release` et al.

LTO is slow and memory-intensive process which is usually unnecessary
during development.  The difference isn’t small either as a build with
LTO can take an order of magnitude longer than without it.  With
enough cores an LTO build can easily consume all available memory and
start failing due to lack of memory.  (Benchmark results below).

Because of all that, disable LTO by default.  This should improve
development experience however it’s important to now explicitly enable
LTO when creating a release intended to be run in production as per
instructions above.

I’ve performed rough benchmarks to try and quantify the differences.
The tests were performed an Ryzen 9 5900X 12-core CPU with 32 GiB of
memory.  A full build from clean state and an incremental builds were
test with the following sequence of commands:

    cargo clean
    git checkout 65bd8cac6c3b45561ab3a677
    cargo build -j$jobs --release \
                --features nightly_protocol,nightly_protocol_features
    git checkout HEAD~
    cargo build -j$jobs --release \
                --features nightly_protocol,nightly_protocol_features

* Elapsed time of full builds:

      |          |   -j6 |  -j8 | -j12 | -j24 |
      | lto=off  |  5:16 |      | 3:22 | 2:58 |
      | lto=thin | 11:07 |      | 6:24 | 5:21 |
      | lto=fat  |  8:57 | 7:18 | OOM  | OOM  |

* Elapsed time of incremental builds:

      |          |  -j6 | -j12 | -j24 |
      | lto=off  | 0:10 | 0:10 | 0:10 |
      | lto=thin | 2:32 | 1:22 | 1:09 |
      | lto=fat  | 2:43 | 2:41 | OOM  |

* neard binary size:

      | --lto=off  | 50 MB |
      | --lto=thin | 43 MB |
      | --lto=fat  | 32 MB |

I did some effort to control other variables but the results should be
taken with a grain of salt.  In particular, ‘-j8 lto=fat’ being faster
than ‘-j8 lto=thin’ was likely a fluke.  Nonetheless, the numbers
represent the real experience.

Fixes: https://github.com/near/nearcore/issues/4403


chefsale:
I believe this is problematic, as we use this profile to build all our current production binaries which get deployed in the current Makefile: https://github.com/near/nearcore/blob/master/Makefile. I think if we want to introduce a change like this, that should be added as a seperate `make dev option` instead of the release one. cc: @frol @bowenwang1996 @matklad

chefsale:
I just saw the other PR which changes the actual makefile configuration: https://github.com/near/nearcore/pull/4506, so I believe this should be fine.

mina86:
Sorry, yes, this is just a draft and is blocked on <https://github.com/near/nearcore/pull/4506> and probably some documentation changes I’ve not done yet.

chefsale (approved):

PR #4510, opened 2021-07-14
Makefile: add `neard` target which just buids neard binary

mina86:
The new target will be mentioned in the ‘Compile and Run without
Container’ document† and replace `cargo build` invocations listed
there.

This is because the plan is to disable LTO in the release build
profile and thus to get a best optimised version of the binary
environment variables need to be set.  The Makefile target does
that.

The target is market as `PHONY` so that Make won’t look at the
`neard` directory in the repository as the result of the target.
And while doing that, do the same to other existing targets.

† https://docs.near.org/docs/develop/node/validator/compile-and-run-a-node


chefsale (approved):
LGTM, thanks for adding this. :smile

PR #4511, opened 2021-07-14
refactor(runtime): use wasmparser instead of always depend on wasmer 0.x

ailisp:
As shown by the TODO comment

Test Plan
---------
CI


runtime/near-vm-runner/Cargo.toml
```diff
@@ -16,8 +16,8 @@ This crate implements the specification of the interface that Near blockchain ex
 borsh = "0.8.1"
 serde = { version = "1", features = ["derive"] }
 wasmer-runtime = { version="0.17.1", features = ["default-backend-singlepass"], default-features = false, package = "wasmer-runtime-near", optional = true }
-# Always used even for wasmer 1.0 for validating wasm, will be replaced when refactor prepare.rs
-wasmer-runtime-core = {version = "0.17.4", package = "wasmer-runtime-core-near" }
+wasmer-runtime-core = {version = "0.17.4", package = "wasmer-runtime-core-near", optional = true}
+wasmparser = "0.79"
```
matklad, line 20: Note that, besides decoupling deps, this *also* upgrades the version of `wasmparser` we use for validation, and that is potentially behavior changing (ie, new version of wasmparser might accept more files as valid wasm). So, upgrading `wasmparser` should be done very carefully. 

To separate the two pieces of work (untangling deps and updating the parser), let's use exactly the same version of wasmeparser we were using before  (I think `wasmparser 0.51` ?)

matklad (changes requested):

matklad (approved):

Issue #4515, opened 2021-07-14
Remove protocol_feature_evm


matklad:
I think we want to completely remove `protocol_feature_evm` and the related functionality. Is that correct?

I don't think we currently are tracking this work, so let's create an issue for this. I wonder what's the right team to do this? Probably contract runtime at this point?

bowenwang1996:
> I think we want to completely remove protocol_feature_evm and the related functionality. Is that correct?

Not so sure about that. @artob @djsatok what do you think?

birchmd:
As far as I know the feature is not needed any more because we used a different one for the math API work, which is all that Aurora needs from nearcore since it was decided to have the engine be a contract instead of part of nearcore itself.

PR #4516, opened 2021-07-14
fix: fix nightly param estimator build

### LINKED ISSUE #4509: runtime-param-estimator doesn't build under nightly
https://buildkite.com/nearprotocol/nearcore-nightly-release/builds/451#fbf9530b-88db-4e2a-b9c2-87ec58d49488

matklad:
closes #4509


frol (approved):
I have zero context about evm support, so I can only confirm that the change itself is legit.

bowenwang1996 (approved):
I think this change makes sense, but not so sure about removing the evm feature completely. Let's wait for the Aurora team to comment on #4515

matklad:
Actually, seems just the second commit is necessary and sufficient.

chefsale (approved):
:shipit: :shipit: :shipit: As the evm stuff got removed.

Issue #4374, opened 2021-06-16
Fuzzing the entire runtime


matklad:
[OKR 2021Q4] At a high level, our runtime is a pure function which takes some state, a bunch of actions, interprets those actions and returns a new state. We also care a lot about runtime being correct even in the phase of adversary inputs. So it behooves us to implement fuzzing of the runtime. This article (and it's bibliography) give a good overview of state of the art fuzzing in Rust: https://fitzgeraldnick.com/2020/08/24/writing-a-test-case-generator.html.

The TL;DR is that the best approach is structured, coverage guided fuzzing. We use something like `libfuzzer` to generate random inputs `&[u8]`, then we use this input as a seed to generate a random sequence of valid actions, then we feed this input into the runtime. The fuzzer then observes code coverage as the runtime executes the input, and uses that info to generate better seeds to cover more of the branches, and to minimize failures for free. 

Practically, that means that we should: 

* define a data structure describing a series of inputs to the runtime
* implement https://docs.rs/arbitrary/1.0.1/arbitrary/trait.Arbitrary.html for this data structure. Care must be taken to generate reasonably well-formed sequences. Ie, maintain a pool of active accounts, using valid signatures, etc. Of course, generating invalid inputs is also something we need to do from time-to-time. The link blog post (and the [wasm-smith](https://github.com/bytecodealliance/wasm-tools/tree/main/crates/wasm-smith) crate it describes) are a good model here. We actually should directly use `wasm-smith`  for `DeployContract` action.
* implement a fuzzing target for `cargo-fuzz`
* run the fuzzing locally and test that the coverage is reasonable. That is, deliberately introduce paniking bugs into runtime, and check that fuzzing catches them
* setup some kind of CI job to run the fuzzing. Perhaps even apply for https://google.github.io/oss-fuzz/ ? 
* optional: look into "swarm testing" (see the link in the post)

bowenwang1996:
Regarding fuzz testing function calls, I think we should focus on the following:
* fuzzing function calls that generate receipts with different types of actions (transfer, delete account, create account, deploy contracts, etc). Fuzzing should be able to uncover bugs like what is fixed in #4512.
* fuzzing all host functions. Host function bugs like #4535 should be discovered by fuzzing.
* fuzzing cross-contract calls. This should ideally be combined with the two above to make fuzzing more robust and useful.

posvyatokum:
Pushed more code to #4546
Right now switching to #4550 to run current fuzzer in CI

PR #4524, opened 2021-07-15
Runtime fuzzer

posvyatokum:
Adding creating of an arbitrary Scenario with Transfer, Stake, and CreateAccount transactions. Also adding fuzzing target.
Currently, all the bounds for random numbers are also chosen randomly (by me).
Next pr will add contract calls, but it will only be for contracts from 'runtime/near-test-contracts'


bowenwang1996 (commented):
Adding a readme on how to run the fuzzer could be helpful

test-utils/runtime-tester/src/fuzzing.rs
```diff
@@ -0,0 +1,174 @@
...
+    fn arbitrary(u: &mut Unstructured<'_>) -> Result<Self> {
+        let num_accounts = u.int_in_range(2..=100)?;
+
+        let seeds: Vec<String> = (0..num_accounts).map(|i| format!("test{}", i)).collect();
+
+        let mut scope = Scope::from_seeds(&seeds);
+
+        let network_config = NetworkConfig { seeds };
+
+        let num_blocks = u.int_in_range(1500..=2500)?;
```
posvyatokum, line 22: Just realised, that input might not be long enough to create that many blocks. Need to change this to the loop, that runs while u.len() is big enough to create at least an empty block.

test-utils/runtime-tester/src/fuzzing.rs
```diff
@@ -0,0 +1,174 @@
...
+        Ok(Scenario { network_config, blocks })
+    }
+}
+
+impl BlockConfig {
+    fn arbitrary(u: &mut Unstructured, scope: &mut Scope) -> Result<BlockConfig> {
+        scope.inc_height();
+        let mut block_config = BlockConfig::at_height(scope.get_height());
+
+        let num_txs = u.int_in_range(0..=500)?;
```
posvyatokum, line 38: Same here, need to choose number of txs based on u.len() (and probably some hardcoded max number of txs in a block)

test-utils/runtime-tester/src/run_test.rs
```diff
@@ -51,30 +52,32 @@ impl Scenario {
...
             runtime_stats.blocks_stats.push(block_stats);
         }
 
-        runtime_stats
+        Ok(runtime_stats)
     }
 }
 
-#[derive(Serialize, Deserialize)]
+#[derive(Serialize, Deserialize, Debug)]
```
posvyatokum, line 69: Debug is needed to print scenario in case of error. It is better to implement Debug as json serialization, so we can run 'cargo fuzz fmt <target> <input> > scenario.json' and have scenario that we can run separately from fuzz.

test-utils/runtime-tester/src/fuzzing.rs
```diff
@@ -0,0 +1,174 @@
...
+}
+
+#[derive(Clone)]
+pub struct Account {
+    pub id: AccountId,
+    pub balance: Balance,
+}
+
+impl Scope {
+    fn from_seeds(seeds: &Vec<String>) -> Self {
```
matklad, line 138: Nit: `&[String]`

test-utils/runtime-tester/src/fuzzing.rs
```diff
@@ -0,0 +1,174 @@
...
+        let accounts = seeds.iter().map(|id| Account::from_id(id.clone())).collect();
+        Scope { accounts, nonce: 0, height: 0 }
+    }
+
+    pub fn inc_height(&mut self) {
+        self.height += 1
+    }
+
+    pub fn get_height(&self) -> BlockHeight {
```
matklad, line 147:
```suggestion
    pub fn height(&self) -> BlockHeight {
```

Rust doesn't use `get_` prefix (and it's ok to have method and field with the same name)

test-utils/runtime-tester/src/run_test.rs
```diff
@@ -119,6 +122,17 @@ impl TransactionConfig {
...
+use core::fmt::Debug;
+impl Debug for TransactionConfig {
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+        write!(
+            f,
+            "(Nonce: {}, SignerId: {}, ReceiverId: {}, Actions: {:?} )",
+            &self.nonce, &self.signer_id, &self.receiver_id, &self.actions
+        )
+    }
+}
```
matklad, line 134: It's better to `use std::fmt` once. 

```suggestion
impl fmt::Debug for TransactionConfig {
    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
        write!(
            f,
            "(Nonce: {}, SignerId: {}, ReceiverId: {}, Actions: {:?} )",
            &self.nonce, &self.signer_id, &self.receiver_id, &self.actions
        )
    }
}
```

test-utils/runtime-tester/src/run_test.rs
```diff
@@ -119,6 +122,17 @@ impl TransactionConfig {
     }
 }
 
+use core::fmt::Debug;
+impl Debug for TransactionConfig {
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+        write!(
+            f,
+            "(Nonce: {}, SignerId: {}, ReceiverId: {}, Actions: {:?} )",
```
matklad, line 130: Consider using debug_struct here:

https://doc.rust-lang.org/stable/std/fmt/struct.Formatter.html#method.debug_struct

test-utils/runtime-tester/src/run_test.rs
posvyatokum: I changed the way Scenario Debug works, but used std::fmt this time

test-utils/runtime-tester/fuzz/README.md
```diff
@@ -0,0 +1,59 @@
...
+
+
+To run fuzz test:
+```bash
+rustup run nightly cargo fuzz run runtime-fuzzer
+```
+`max_len` here is to create bigger scenarios.
+
+Fuzzing starts from small inputs (Unstructured) and slowly goes to bigger.
+To go to bigger input faster run fuxx with additional arguments:
```
matklad, line 15:
```suggestion
To go to bigger input faster run fuzz with additional arguments:
```

test-utils/runtime-tester/fuzz/README.md
```diff
@@ -0,0 +1,59 @@
+# Runtime Fuzz
+
+Currently only one target is present -- runtime-fuzzer.  
+This target will create random scenarios using Arbitrary trait
+and execute them. This will keep happening, until one scenario fails.  
+
+
+To run fuzz test:
+```bash
+rustup run nightly cargo fuzz run runtime-fuzzer
```
matklad, line 10: I usually run fuzzing as

```suggestion
RUSTC_BOOTSTRAP=1 cargo fuzz run runtime-fuzzer
```

RUSTC_BOOTSTRAP is a cheat code that unlocks nightly features on the stable compiler. It shouldn't be used for anything which requires stability, but its fine for fuzzing I think.

test-utils/runtime-tester/src/run_test.rs
```diff
@@ -89,17 +92,24 @@ pub struct TransactionConfig {
...
 
-#[derive(Serialize, Deserialize, Default)]
+#[derive(Serialize, Deserialize, Default, Debug)]
 pub struct BlockStats {
     pub height: u64,
     pub block_production_time: Duration,
 }
 
+use std::fmt::Debug;
+impl Debug for Scenario {
```
matklad, line 107: Nit:

```suggestion
impl fmt::Debug for Scenario {
```

we don't usually put `use` in the middle of the file

test-utils/runtime-tester/src/run_test.rs
```diff
@@ -51,15 +52,17 @@ impl Scenario {
 
             let start_time = Instant::now();
 
-            last_block = env.clients[0].produce_block(block.height).unwrap().unwrap();
+            last_block = env.clients[0]
+                .produce_block(block.height)?
+                .ok_or(Error::Other(String::from("No block has been produced")))?;
```
matklad, line 57: Nit
```suggestion
                .ok_or_else(|| Error::Other(String::from("No block has been produced")))?;
```

The original allocates a string every time, the new version only on the sad path.

matklad (approved):
The code looks good to me, left just a couple of small nits. 

Two comments/questions:

* should we perhaps add this to the nightly pipeline? To run the fuzzing for half-an-hour every night, or something like this? Not sure what's the best strategy? 
* did we try to estimate how well our generator works? Can we, eg, detect an overflow if we replace some of `checked_add`s with plain `+`?

posvyatokum:
> * should we perhaps add this to the nightly pipeline? To run the fuzzing for half-an-hour every night, or something like this? Not sure what's the best strategy?
> * did we try to estimate how well our generator works? Can we, eg, detect an overflow if we replace some of `checked_add`s with plain `+`?

At this point, fuzzer is very weak, so probably shouldn't add it to the nightly pipeline yet. 
Currently, I mostly create PR's to approve the basic design. Right after merging this PR, I will create a new one with contract deployment, function calls, and a couple of tricks to make scenarios more realistic. And even then overflows will not be detected in a reasonable time -- these scenarios are too simplistic. 
Bowen laid out some examples of what we want to catch https://github.com/near/nearcore/issues/4374#issuecomment-882826238. I will follow these suggestions and hopefully fuzzer will start catching at least intentional bugs quick

matklad:
>At this point, fuzzer is very weak, so probably shouldn't add it to the nightly pipeline yet.

Process-wise, I'd suggest adding stuff to CI as early as possible, and then using CI as a ratchet, to make sure that stuff can progress forward, but can not regress. That is, the benefit of CI at this stage is not that we catch bugs, but that we prevent fuzzing infra itself from regressing. 

But this is a "rule of thumb" suggestion, there's little specific benefits in this case.

bowenwang1996:
> > At this point, fuzzer is very weak, so probably shouldn't add it to the nightly pipeline yet.
> 
> Process-wise, I'd suggest adding stuff to CI as early as possible, and then using CI as a ratchet, to make sure that stuff can progress forward, but can not regress. That is, the benefit of CI at this stage is not that we catch bugs, but that we prevent fuzzing infra itself from regressing.
> 
> But this is a "rule of thumb" suggestion, there's little specific benefits in this case.

@matklad do you mind creating an issue for this?

PR #4526, opened 2021-07-15
Makefile: add a default ‘all’ target

mina86:
Introduce a default ‘all’ target to the Makefile so that running
‘make’ on its own produces a relaese build which is likely what
user meant.


chefsale (approved):
LGTM, thanks for adding this :)

PR #4528, opened 2021-07-15
fix(test): remove rpc routing tests

bowenwang1996:
Remove rpc routing tests since now even transaction status query routing is disabled (see https://github.com/near/nearcore/discussions/4393 for more details). These tests currently work out of coincidence and if validator-to-shard assignment changes, it is not guaranteed to work. Blocks #4193.

Test plan
----------
* CI


frol (approved):

frol:
cc @khorolets

PR #4529, opened 2021-07-16
Base function call computations

olonho:


runtime/runtime-params-estimator/src/function_call.rs
```diff
@@ -0,0 +1,150 @@
...
+        write!(
+            &mut methods,
+            "
+            (export \"hello{}\" (func {}))
+              (func (;{};)
+                i32.const {}
+                drop
+                return
+              )
+            ", i, i, i, i)
```
matklad, line 81:
```suggestion
            r#"
            (export "hello{i}" (func {i}))
              (func (;{i};)
                i32.const {i}
                drop
                return
              )
            "#, i=i)
```

although I guess it's better to use some wat sugar here to not repeat the i in the first place:

```rust
        write!(
            &mut methods,
            r#"
            (func (export "hello{i}")
                i32.const {i}
                drop
                return
            )
            "#,
            i = i
        )

```

runtime/runtime-params-estimator/src/function_call.rs
```diff
@@ -0,0 +1,150 @@
...
+    }
+
+    let code = format!(
+        "
+        (module
+            {}
+            )",
+        methods
+    );
+    ContractCode::new(wabt::wat2wasm(code.as_bytes()).unwrap(), None)
```
matklad, line 92: needs a rebase, we now use `wat` instead of `wabt`.

runtime/runtime-params-estimator/src/function_call.rs
```diff
@@ -0,0 +1,150 @@
...
+        vm_kind, metric,
+        ratio_to_gas_signed(metric, cost_base),
+        ratio_to_gas_signed(metric, cost_byte),
+    );
+}
+
+#[test]
+fn test_function_call_time() {
+    // Run with
+    // cargo test --release --lib function_call::test_function_call_time -- --exact --nocapture
```
matklad, line 48: Hm, shouldn't this include `--features required` as well?

I *think* as written, this'll hit the in-memory cache, masking the deserializaion costs, and desiralization should be the primary contributor to func call cost

runtime/runtime-params-estimator/src/function_call.rs
```diff
@@ -0,0 +1,150 @@
...
+    // --lib function_call::test_function_call_icount -- --exact --nocapture
+    // Where runner.sh is
+    // /host/nearcore/runtime/runtime-params-estimator/emu-cost/counter_plugin/qemu-x86_64 \
+    // -cpu Westmere-v1 -plugin file=/host/nearcore/runtime/runtime-params-estimator/emu-cost/counter_plugin/libcounter.so $@
+    test_function_call(GasMetric::ICount, VMKind::Wasmer0);
+    test_function_call(GasMetric::ICount, VMKind::Wasmer1);
+    test_function_call(GasMetric::ICount, VMKind::Wasmtime);
+}
+
+fn make_many_methods_contact(method_count: i32) -> ContractCode {
```
Longarithm, line 69:
```suggestion
fn make_many_methods_contract(method_count: i32) -> ContractCode {
```

runtime/runtime-params-estimator/src/function_call.rs
```diff
@@ -0,0 +1,150 @@
...
+use std::fmt::Write;
+use std::sync::Arc;
+
+#[allow(dead_code)]
+fn test_function_call(metric: GasMetric, vm_kind: VMKind) {
+    let mut xs = vec![];
+    let mut ys = vec![];
+    const REPEATS: i32 = 50;
+    for method_count in vec![5, 20, 30, 50, 100, 200, 1000] {
+        let contract = make_many_methods_contact(method_count);
```
Longarithm, line 22: Nit
```suggestion
        let contract = make_many_methods_contract(method_count);
```

runtime/runtime-params-estimator/src/function_call.rs
```diff
@@ -0,0 +1,150 @@
...
+use near_vm_runner::{run_vm, VMKind};
+use nearcore::get_store_path;
+use std::fmt::Write;
+use std::sync::Arc;
+
+#[allow(dead_code)]
+fn test_function_call(metric: GasMetric, vm_kind: VMKind) {
+    let mut xs = vec![];
+    let mut ys = vec![];
+    const REPEATS: i32 = 50;
```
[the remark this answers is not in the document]
olonho: Done

runtime/runtime-params-estimator/src/function_call.rs
olonho: Done, although from results it doesn't seem so.

runtime/runtime-params-estimator/src/function_call.rs
olonho: Done

matklad (approved):

PR #4530, opened 2021-07-16
Base parallelization

EgorKulikov:
This parallelizes applying chunks within one block and applying blocks for this and next epoch

### Test plan
There are existing tests for this behavior. Some additional sanity tests would be added shortly


chain/chain/src/chain.rs
```diff
@@ -2887,32 +2890,102 @@ impl<'a> ChainUpdate<'a> {
...
-                        )
-                        .map_err(|e| ErrorKind::Other(e.to_string()))?;
+                    let runtime_adapter = self.runtime_adapter.clone();
+                    let block_hash = block.hash().clone();
+                    let challenges_result = block.header().challenges_result().clone();
+                    let block_timestamp = block.header().raw_timestamp();
+                    let gas_price = prev_block.header().gas_price();
+                    let random_seed = *block.header().random_value();
+                    let height = chunk_header.height_included();
+                    let prev_block_hash = chunk_header.prev_block_hash().clone();
```
bowenwang1996, line 2900: This feels somewhat ugly, but I also don't see a better way to do it :(

bowenwang1996 (commented):
We should also parallelize applying chunks for this epoch and next epoch. Currently they are done sequentially in `process_block`

chain/chain/src/chain.rs
EgorKulikov: Tbh this is not ready for review, I just wanted to run all tests

matklad:
I see that this is a draft, but just in case: it's better not to spawn threads by hand and rather use a threadpool, which is shared by all the components which need parallelism. I know two nice patterns for that:

* Just using [rayon](https://docs.rs/rayon/1.5.1/rayon/), which gives you convenient global threadpool and a lot of utilities to schedule parallel work. I think we already use rayon in https://github.com/near/nearcore/pull/4344
* Create thread pool once, at the start on the program, and pass it explicitly into everything that needs to be parallel. I really love how sorbet type checker [does this](https://github.com/sorbet/sorbet/blob/ba614a5a69c77c89c88b8c9983185927856b6ac0/common/concurrency/WorkerPool.h). 

It would also help to document approach to parallelism in [architecture](https://github.com/near/nearcore/blob/master/docs/architecture.md) doc, to make sure we use one consistent pattern.

EgorKulikov:
> I see that this is a draft, but just in case: it's better not to spawn threads by hand and rather use a threadpool, which is shared by all the components which need parallelism. I know two nice patterns for that:
> 
> * Just using [rayon](https://docs.rs/rayon/1.5.1/rayon/), which gives you convenient global threadpool and a lot of utilities to schedule parallel work. I think we already use rayon in [Compile contracts on writing them to state #4344](https://github.com/near/nearcore/pull/4344)
> * Create thread pool once, at the start on the program, and pass it explicitly into everything that needs to be parallel. I really love how sorbet type checker [does this](https://github.com/sorbet/sorbet/blob/ba614a5a69c77c89c88b8c9983185927856b6ac0/common/concurrency/WorkerPool.h).
> 
> It would also help to document approach to parallelism in [architecture](https://github.com/near/nearcore/blob/master/docs/architecture.md) doc, to make sure we use one consistent pattern.

I am not sure we should overengineer this exact case. If later we would settle on any of this frameworks it would be relatively easy to change code here

chain/chain/src/chain.rs
```diff
@@ -2891,98 +2900,187 @@ impl<'a> ChainUpdate<'a> {
...
+                        );
+                    }
+                    Err(err) => {
+                        let msg = if let Some(msg) = err.downcast_ref::<&'static str>() {
+                            msg.to_string()
+                        } else if let Some(msg) = err.downcast_ref::<String>() {
+                            msg.clone()
+                        } else {
+                            format!("?{:?}", err)
+                        };
```
bowenwang1996, line 3077: is this part necessary? Could we print the error directly?

bowenwang1996 (commented):
Could we refactor that part that joins handles and does other updates into a separate function? Right now the code is a bit hard to read.

chain/chain/src/chain.rs
EgorKulikov: You mean leaving just the "else" part? We are not really printing here, we are propagating any panic that happened inside

chain/chain/src/chain.rs
```diff
@@ -2891,98 +2900,187 @@ impl<'a> ChainUpdate<'a> {
...
+                    let challenges_result = block.header().challenges_result().clone();
+                    let block_timestamp = block.header().raw_timestamp();
+                    let gas_price = prev_block.header().gas_price();
+                    let random_seed = *block.header().random_value();
+                    let height = chunk_header.height_included();
+                    let prev_block_hash = chunk_header.prev_block_hash().clone();
+                    #[cfg(feature = "sandbox")]
+                    let states_to_patch = self.states_to_patch.take();
+
+                    same_height_handlers.push(thread::spawn(
```
matklad, line 2915: I don't think spawning the threads directly is necessary the best approach. For example, Meyers effective modern C++ advises against that: Item 35: Prefer task-based programming to thread-based.
 
A specific problems we can face are

* threads are somewhat costly to create, for best performance, we want to cache and re-use threads
* this spawns the number of threads proportional to the amount of work, and the work is CPU bound. That means that we can easily over-subscribe the CPU and spend unnecessary time on context switches. It's better to spawn the number of threads proportional to the number of CPUs
* threads consume limited OS resources. If you spawn more than 10k threads, you can run into a situation where `thread::spawn` just panics. Note that some of those resources are global per OS. I don't think will run into the those limits (they are pretty high in practice today), but it feels more robust to not have potentially unbounded resource usage.

If we pursue this approach with imperative task spawning, I suggest using a thread pool to control the OS threads we use.

chain/chain/src/chain.rs
```diff
@@ -2765,13 +2770,15 @@ impl<'a> ChainUpdate<'a> {
...
+    // Applies chunks in separate threads. Returns callback that would actually wait till all
+    // chunks are applied
     fn apply_chunks(
         &mut self,
         me: &Option<AccountId>,
         block: &Block,
         prev_block: &Block,
         mode: ApplyChunksMode,
-    ) -> Result<(), Error> {
+    ) -> Result<impl FnOnce(&mut ChainUpdate) -> Result<(), Error>, Error> {
```
matklad, line 2782: This function is now very hard to read. The core issue is that it now scores high on three orthogonal complexity axis:

* it is just complex code (has many parameters, complex control flow)
* separately, it manages concurrency -- it spawns a concurrent thread of execution, which is not encapsulated to the function, but is returned to the caller as well. 
* separately, it involved complicated high-order language machinery -- rather than just computing stuff, it returns the closure to compute stuff later. 

I don't see a direct way to just make stuff simpler -- all three facets feel essential to how the code is written at the moment. But I wonder if we can restructure the whole computation such that we don't need this complexity. 

I think the original code was structured roughly like this:

```rust
fn apply_chunks(args) {
  for {
    let work = prepare_work();
    let result = do_work(work);
    apply_result(result)
  }
}

fn process_block() {
  if cond1 {
    apply_chunks(args1)
  }
  if cond2 {
    apply_chunks(args2)
  }
}
```

The structure in this PR looks like this:


```rust
fn apply_chunks(args) -> impl Fn() {
  let mut threads = Vec::new();
  for {
    let work = prepare_work();
    threads.push(spawn(|| do_work(work));
  }
  return || {
    for t in threads {
      let result = t.join();
      apply_result(result);
    }
  }
}

fn process_block() {
  let mut callbacks = Vec::new();
  if cond1 {
    callbacks.push(apply_chunks(args1))
  }
  if cond2 {
    callbacks.push(apply_chunks(args2))
  }
  for callback in callbacks {
    callback()
  }
}
```

I wonder if we can do some loop unswitching here, and express this as a parallel map? Something along these lines:

```rust
fn apply_chunks(args) -> Vec<Work> {
  let mut res = Vec::new()
  for {
    let work = prepare_work();
    res.push(work);
  }
  res
}

fn process_block() {
  let mut work: Vec<Work> = Vec::new();
  if cond1 {
    work.extend(apply_chunks(args1));
  }
  if cond2 {
    work.extend(apply_chunks(args2));
  }
  work
   .par_iter() // using rayon, declaratively process all work in parallel
   .map(|work| do_work(work))
   .for_each(|result| apply_result(result))
}
```

To clarify, I have a very superficial understanding of what's going on here, so I might be missing something obvious. But I do suggest trying to cast the problem as "First, prepare the work which can be done independently, then run each bit of work in parallel, then apply results".

chain/chain/src/chain.rs
```diff
@@ -2891,98 +2900,187 @@ impl<'a> ChainUpdate<'a> {
...
                 }
             }
         }
 
-        Ok(())
+        let block_hash = block.hash().clone();
+        let prev_block_hash = prev_block.hash().clone();
+        Ok(move |chain_update: &mut ChainUpdate| -> Result<(), Error> {
+            for result in same_height_handlers {
+                let res = result.join();
```
matklad: I'd just

```suggestion
                let res = result.join().unwrap();
```

here. 

The panic message is displayed in the thread that panicked, so here we can just propagate without additional logging or handling. I'd even say that it's a minor problem in the API that, by default join returns a Result, it should have paniked by default: https://matklad.github.io/2020/12/12/notes-on-lock-poisoning.html.

chain/chain/src/chain.rs
```diff
@@ -1841,7 +1842,9 @@ impl Chain {
         let prev_block = self.store.get_block(block.header().prev_hash())?.clone();
 
         let mut chain_update = self.chain_update();
-        chain_update.apply_chunks(me, &block, &prev_block, ApplyChunksMode::NextEpoch)?;
+        chain_update.apply_chunks(me, &block, &prev_block, ApplyChunksMode::NextEpoch)?(
+            &mut chain_update,
+        )?;
```
matklad, line 1848: We probably want some helper function for the "apply without parallelism" case.

EgorKulikov:
> Could we refactor that part that joins handles and does other updates into a separate function? Right now the code is a bit hard to read.

Done

chain/chain/src/chain.rs
EgorKulikov: Ok, you've persuaded me, it would be no harder to switch from rayon to something else in the future if need arises

matklad: To clarify, I am not married to rayon in particular: hand-rolled threadpool would work fine as well. In rust-analyzer, I use https://docs.rs/threadpool which is a dead simple thing. 

One problem with rayon is that its a pretty complex beast actually. https://github.com/sorbet/sorbet/blob/ba614a5a69c77c89c88b8c9983185927856b6ac0/common/concurrency/WorkerPool.h shows that this thing needn't be complex.

chain/chain/src/chain.rs
```diff
@@ -2782,18 +2782,19 @@ impl<'a> ChainUpdate<'a> {
         })
     }
 
-    // Applies chunks in separate threads, awaits finishing and processes results
+    /// Applies chunks in separate threads, awaits finishing and processes results
```
matklad, line 2785: :+1:, good idea to use doc comment format for function docs.

chain/chain/src/chain.rs
matklad: This probably isnt' applicable in this case, but one cool thing that rayon does is it allows to close over stack data without cloning it:

https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=f9af6bb5b8554343b310d8012a293edf

chain/chain/src/chain.rs
```diff
@@ -2891,97 +2942,159 @@ impl<'a> ChainUpdate<'a> {
...
-                    );
+                        ) {
+                            Ok(apply_result) => {
+                                Ok(ApplyChunkResult::SameHeight(SameHeightResult {
+                                    shard_id,
+                                    gas_limit,
+                                    apply_result,
+                                }))
+                            }
+                            Err(err) => Err(ErrorKind::Other(err.to_string()).into()),
```
matklad, line 2985: Curisous, why this isn't just `Err(err) => Err(err),`?

chain/chain/src/chain.rs
```diff
@@ -2765,13 +2782,47 @@ impl<'a> ChainUpdate<'a> {
...
+    }
+
+    /// Applies chunks and processes results
+    fn do_apply_chunks(
+        &mut self,
+        block: &Block,
+        prev_block: &Block,
+        work: Vec<Box<dyn FnOnce() -> Result<ApplyChunkResult, Error> + Send + 'static>>,
+    ) -> Result<(), Error> {
+        work.into_par_iter().map(|task| task()).collect::<Vec<_>>().into_iter().try_for_each(
```
matklad, line 2804: Nice, now we can say "*this line* is where all the || happens"!

chain/chain/src/chain.rs
```diff
@@ -2765,13 +2782,47 @@ impl<'a> ChainUpdate<'a> {
...
+    }
+
+    /// Creates jobs that would apply chunks
+    fn apply_chunks_preprocessing(
+        &mut self,
+        me: &Option<AccountId>,
+        block: &Block,
+        prev_block: &Block,
+        mode: ApplyChunksMode,
+    ) -> Result<Vec<Box<dyn FnOnce() -> Result<ApplyChunkResult, Error> + Send + 'static>>, Error>
```
matklad, line 2822: Summarising zulip discussion: it's a bit unfortunate that we are returning a closure here, instead of some specific things. The reason for that is that what we want to do here is just pack arguments for `RuntimeAdapter::apply_transactions` into a single struct, and there are a lot of arguments there ([seen worse](https://github.com/JetBrains/intellij-community/blob/9c92f6e34aa959cbd8b3452a28898a7c63052915/platform/platform-impl/src/com/intellij/ui/BalloonImpl.java#L326-L354) though :) ). 

One potential solution here is to change `fn apply_transactions` to 

```
    fn apply_transactions(&self, transacions_to_apply: TransactionsToApply)
```

but:

* it's unclear if its worth it
* this is out of scope for this PR.

chain/chain/src/chain.rs
```diff
@@ -3129,17 +3242,31 @@ impl<'a> ChainUpdate<'a> {
...
+        let mut apply_chunk_work = Vec::new();
+
         // Always apply state transition for shards in the current epoch
-        self.apply_chunks(me, block, &prev_block, ApplyChunksMode::ThisEpoch)?;
+        apply_chunk_work.extend(self.apply_chunks_preprocessing(
+            me,
+            block,
+            &prev_block,
+            ApplyChunksMode::ThisEpoch,
+        )?);
```
matklad, line 3253: Just as an alternative, one pattern I like when I need to assemble a complex sequence of things is passing an accumulator into the function, instead of retruning a vec:

```rust
fn before() -> Vec<Thing>;
fn after(acc: &mut Vec<Thing>);
```

This allows saving intermediate allocations, but is mostly unrelated to this PR.

matklad (approved):

chain/chain/src/chain.rs
EgorKulikov: I am as well, but this is to preserve behavior

chain/chain/src/chain.rs
```diff
@@ -2518,6 +2519,22 @@ pub struct ChainUpdate<'a> {
...
+struct DifferentHeightResult {
+    shard_id: ShardId,
+    apply_result: ApplyTransactionResult,
+}
+
+enum ApplyChunkResult {
+    SameHeight(SameHeightResult),
+    DifferentHeight(DifferentHeightResult),
+}
+
```
bowenwang1996, line 2537: Could we add a comment on what these structs do?

bowenwang1996 (approved):

bowenwang1996:
@EgorKulikov please run nayduck to verify that there is no regression. Now that the flaky tests are removed from nayduck, it should be more clear whether there is any regression

EgorKulikov:
> @EgorKulikov please run nayduck to verify that there is no regression. Now that the flaky tests are removed from nayduck, it should be more clear whether there is any regression

http://nayduck.eastus.cloudapp.azure.com:3000/#/run/1986

PR #4533, opened 2021-07-16
pytest: fix sandbox test

mina86:
Make sure that NayDuck builds neard binary with --features sandbox when
running sandbox tests and also make the sandbox test a bit smarter about
locating the executable.


chefsale (approved):
LgTm

Issue #2736, opened 2020-05-27
add a commit hook which compare tests in Nightly and all non-ignored tests


ailisp:
to make sure no expensive test is forgotten

bowenwang1996:
Shouldn't we just add this to CI? @ailisp what is the status of this issue?

ailisp:
@bowenwang1996 I believe this is still nice to have. Yeah it should be part of CI, perhaps some kind of bot, so that if you add an expensive test or a test under pytest dir, a script should check complete_nayduck_set.txt and remind if it's not added

mina86:
Just to confirm that I understand what this issue is about. On high level the request is to list all `pytest/tests/**/*.py` files and check if they have a corresponding line somewhere in one of the `nightly/*.txt` files?

bowenwang1996:
@mina86 yes

PR #4539, opened 2021-07-19
pytest: add scripts/check_pytest.py checking if tests are listed

mina86:
On high level, this simply lists all `pytest/tests/**/*.py` files and
checks whether they are listed somewhere in `nightly/*.txt` files.
The end goal is to have this as a required check before a commit
is accepted so that no pytest’s are forgotten.

Currently though, there are 9 files which the script complains
about.  The output of the script is:

    Found 9 pytest files (i.e. Python files in pytest/tests directory)
    which are not included in any of the nightly/*.txt files:

     -  pytest/tests/adversarial/gc_rollback.py
     -  pytest/tests/sanity/concurrent_function_calls.py
     -  pytest/tests/sanity/proxy_example.py
     -  pytest/tests/sanity/restart.py
     -  pytest/tests/sanity/rpc_finality.py
     -  pytest/tests/sanity/rpc_tx_submission.py
     -  pytest/tests/sanity/state_migration.py
     -  pytest/tests/stress/network_stress.py
     -  pytest/tests/stress/saturate_routing_table.py

    Add the tests to one of the lists in nightly/*.txt files.  For example
    as:

        pytest sanity/rpc_tx_submission.py

    If the test is temporarily disabled, add it to one of the files with an
    appropriate TODO comment.  For example:

        # TODO(#1234): Enable the test again once <some condition>
        # pytest sanity/rpc_tx_submission.py

    Note that the TODO comment must reference a GitHub issue (i.e. must
    contain a #<number> string).

    If the file is not a test but a helper library, consider moving it out
    of the pytest/tests directory to pytest/lib or add it to HELPER_SCRIPTS
    list at the top of scripts/check_pytests.py file.

Before this check is incorporated in checks pipeline, the above needs
to be resolved.

Issue: https://github.com/near/nearcore/issues/2736


bowenwang1996:
@mina86 we should certainly exclude scripts in `/stress/hundred_nodes` as they are used to orchestrate the mocknet nodes

mina86:
Done.

bowenwang1996 (approved):
I think we should use just one txt file to avoid confusions. Otherwise looks good.

mina86:
> I think we should use just one txt file to avoid confusions. Otherwise looks good.

While that may be true, I’d much rather deal with that separately along with adding all the nine files listed here. There’s 14 files which are in one of the `*.txt` files other than `complete_nayduck_set.txt`.

bowenwang1996:
@chefsale @mhalambek looks like this PR requires approval from you

chefsale (approved):

PR #4543, opened 2021-07-20
Remove `PhantomData` from `Genesis`.

### LINKED ISSUE #2459: Remove unnecessary PhantomData usage
`PhantomData` has a very specific use case -- to work around issues with the compiler checks: https://doc.rust-lang.org/nomicon/phantom-data.html

Currently, we are using it in some places to prevent initialization of the struct with curly braces `{}`, e.g.: https://github.com/nearprotocol/nearcore/blob/3cab57953dbaf1f89c5f6026aa121ed519b50c73/core/chain-configs/src/genesis_config.rs#L109
It is actually not important that we are using PhantomData in there. We could be using any type as long as borsh and serde skip them and they have default initializer.

Instead, we should prohibit usage of `{}` through the visibility, e.g. by making some or all fields of `Genesis` non-public and providing accessor-methods. We also can potentially avoid accessor methods by using `pub(self)` or similar see: https://doc.rust-lang.org/reference/visibility-and-privacy.html#pubin-path-pubcrate-pubsuper-and-pubself

nikurt:
`Genesis` attempts to ensure consistency of `genesis.genesis_config.total_supply` with `genesis.records`, but in some cases this consistency can't be enforced. On top of that, the `Genesis` struct is `Serializable`, which means it's technically possible to construct via deserialization a struct with arbitrary values. Accept that and mark all fields as public, but warn the user to use the provided functions to instantiate and de-/serialize.

Fixes #2459


bowenwang1996 (approved):

matklad (approved):
LGTM as the Rust goes, as this makes the code more straightforward. 

However, I am still confused about the desired behavior here. It  feels strange that we both store `total_supply` in JSON but also forcefully set it in `new` to be computed from the set of records. In other words, we have two sources of truth here. 

Can we remove `total_supply` from JSON?

bowenwang1996:
> Can we remove total_supply from JSON?

Good point. We should still be able to read an existing genesis json file after doing that right?

nikurt:
I don't think we can remove `total_supply`, because in some cases, the code doesn't go through all the records, and `total_supply` in JSON is used as the only source of truth.
However, it should be ok to remove the forceful assignment of `total_supply` when records are available. WDYT?

bowenwang1996:
> because in some cases the code doesn't go through all the records

Could you show an example of such behavior? I thought we always go through all the records when the genesis file is loaded

nikurt:
> > because in some cases the code doesn't go through all the records

> Could you show an example of such behavior? I thought we always go through all the records when the genesis file is loaded

The codepath I had in mind is `neard run`. It [loads the config](https://github.com/near/nearcore/blob/daeaa51a54ace03a2172ccf2b603ec2db92dd208/neard/src/cli.rs#L217) and if the records are not in `genesis.json`, then the constructor of `Genesis` struct doesn't read those records.

What I haven't realized before is that `GenesisValidator` [reads all records](https://github.com/near/nearcore/blob/daeaa51a54ace03a2172ccf2b603ec2db92dd208/nearcore/src/genesis_validate.rs#L43) and [checks `total_supply`](https://github.com/near/nearcore/blob/daeaa51a54ace03a2172ccf2b603ec2db92dd208/nearcore/src/genesis_validate.rs#L83). You are right, that all codepaths read all records. If we remove `total_supply` from json, the field will remain unset in `Genesis`, unless `GenesisValidator` sets it.

Brainstorming:
* Maybe we can read the records file in `Genesis` constructor even if it means the records will be read twice: by the constructor and by the validator.
* Maybe we always want to do the validation, and should run the validation in the constructor.
* Do you know any usecases for the `genesis_records_file`? None of `mainnet`, `testnet`, `betanet`, `localnet` have any value there in their respective `config.json` files.

bowenwang1996:
> Maybe we always want to do the validation, and should run the validation in the constructor.

Yeah that sounds reasonable. 

> Do you know any usecases for the genesis_records_file? None of mainnet, testnet, betanet, localnet have any value there in their respective config.json files.

I am not aware of any use cases.

nikurt:
Implemented validation of Genesis upon construction and deserialization. Please take another look.

bowenwang1996 (commented):
how does this PR relate to #4592

nikurt:
Superseded by https://github.com/near/nearcore/pull/4592

PR #1928, opened 2020-01-07
Diagnostics

mfornet:

PR #4545, opened 2021-07-21
refactor: drop unneeded `derive(Serialize)`

matklad:
Serialize is a big promise, so better not to make it if we don't have
to.

Additionally, structs which are both `derive(Serialize, BorshSerialize)`
are suuuper suspicious and a code smell. bors and JSON (what we use
Serialize for) have very different format compatibility guarantees,
keeping a type backwards compatible with both formats at the same time
is tough.

Historical context: this was added in #1928, but we don't use it anymore.

### DIFF
--- a/chain/chain/src/test_utils.rs
+++ b/chain/chain/src/test_utils.rs
@@ -6,7 +6,6 @@ use std::sync::{Arc, RwLock};
 use borsh::{BorshDeserialize, BorshSerialize};
 use chrono::Utc;
 use num_rational::Rational;
-use serde::Serialize;
 use tracing::debug;
 
 use near_chain_primitives::{Error, ErrorKind};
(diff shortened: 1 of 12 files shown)


mfornet (commented):
This was added in PeerMessage and all its dependencies, if we won't use them anymore, probably it makes sense to remove them all.

matklad:
Wow, that's a lot of unused `Seriliazes`, at this scale, they should even have non-trivial impact on compilation time!

chain/chain/src/test_utils.rs
```diff
@@ -55,12 +54,10 @@ use near_chain_configs::ProtocolConfig;
...
 use near_primitives::state_record::StateRecord;
 
-#[derive(
-    BorshSerialize, BorshDeserialize, Serialize, Hash, PartialEq, Eq, Ord, PartialOrd, Clone, Debug,
-)]
+#[derive(BorshSerialize, BorshDeserialize, Hash, PartialEq, Eq, Ord, PartialOrd, Clone, Debug)]
 struct AccountNonce(AccountId, Nonce);
 
-#[derive(BorshSerialize, BorshDeserialize, Serialize, Clone, Debug)]
+#[derive(BorshSerialize, BorshDeserialize, Clone, Debug)]
```
bowenwang1996, line 60: why we even derived `Serialize` for something that is only used in tests is beyond me :(

bowenwang1996 (approved):
Thanks! This is epic

matklad:
@mfornet JFYI, I think since you originally implemented diagnostics, tracing gained a first-class support for what you were emulating via serialize: https://tokio.rs/blog/2021-05-valuable.

PR #4546, opened 2021-07-21
Runtime fuzzer

posvyatokum:
Adding contract deployment and function calls.

### DIFF
--- a/test-utils/runtime-tester/src/fuzzing.rs
+++ b/test-utils/runtime-tester/src/fuzzing.rs
@@ -296,7 +296,7 @@ pub enum Function {
 }
 
 impl Scope {
-    fn from_seeds(seeds: &Vec<String>) -> Self {
+    fn from_seeds(seeds: &[String]) -> Self {
         let accounts = seeds.iter().map(|id| Account::from_id(id.clone())).collect();
         Scope {
             accounts,
--- a/test-utils/runtime-tester/src/fuzzing.rs
+++ b/test-utils/runtime-tester/src/fuzzing.rs
@@ -2,7 +2,10 @@ use crate::run_test::{BlockConfig, NetworkConfig, Scenario, TransactionConfig};
 use near_crypto::{InMemorySigner, KeyType};
 use near_primitives::{
     account::{AccessKey, AccessKeyPermission},
-    transaction::{Action, AddKeyAction, CreateAccountAction, DeployContractAction, FunctionCallAction, TransferAction},
+    transaction::{
+        Action, AddKeyAction, CreateAccountAction, DeleteAccountAction, DeployContractAction,
+        FunctionCallAction, TransferAction,
+    },
     types::{AccountId, Balance, BlockHeight, Nonce},
 };
 use nearcore::config::{NEAR_BASE, TESTING_INIT_BALANCE};
@@ -11,7 +14,10 @@ use byteorder::{ByteOrder, LittleEndian};
 use libfuzzer_sys::arbitrary::{Arbitrary, Result, Unstructured};
 
 use std::collections::HashSet;
+use std::iter::FromIterator;
 use std::mem::size_of;
+
+pub type ContractId = usize;
 
 pub const MAX_BLOCKS: usize = 2500;
 pub const MAX_TXS: usize = 500;
@@ -75,6 +81,8 @@ impl TransactionConfig {
             vec![];
 
         scope.inc_nonce();
+
+        assert!(scope.alive_accounts.contains(&0), "The only validator account got deleted.");
 
         // Transfer
         options.push(|u, scope| {
@@ -146,6 +154,30 @@ impl TransactionConfig {
             })
         });
 
+        // Delete Account
+        if scope.num_alive_accounts() > 1 {
+            options.push(|u, scope| {
+                let account_usize_id = scope.random_non_zero_alive_account_usize_id(u)?;
+                let signer_account = scope.accounts[account_usize_id].clone();
+
+                let signer = InMemorySigner::from_seed(
+                    &signer_account.id,
+                    KeyType::ED25519,
+                    &signer_account.id,
+                );
+
+                Ok(TransactionConfig {
+                    nonce: scope.nonce(),
+                    signer_id: signer_account.id.clone(),
+                    receiver_id: signer_account.id.clone(),
+                    signer,
+                    actions: vec![Action::DeleteAccount(DeleteAccountAction {
+                        beneficiary_id: signer_account.id.clone(),
+                    })],
+                })
+            });
+        }
+
         // Deploy Contract
         options.push(|u, scope| {
             let nonce = scope.nonce();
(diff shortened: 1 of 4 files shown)


bowenwang1996:
@posvyatokum it is better to request reviews from rather than assigning the PR to someone. At least for me a review request is easier to notice.

test-utils/runtime-tester/src/fuzzing.rs
```diff
@@ -2,17 +2,32 @@ use crate::run_test::{BlockConfig, NetworkConfig, Scenario, TransactionConfig};
...
+
 pub const MAX_BLOCKS: usize = 2500;
-pub const MAX_TXS: usize = 300;
+pub const MAX_TXS: usize = 500;
+pub const MAX_TX_DIFF: usize = 10;
 pub const MAX_ACCOUNTS: usize = 100;
 
+const GAS_1: u64 = 900_000_000_000_000;
+const GAS_2: u64 = GAS_1 / 3;
+// const GAS_3: u64 = GAS_2 / 3;
```
bowenwang1996, line 29: Should we remove it?

test-utils/runtime-tester/src/fuzzing.rs
```diff
@@ -141,20 +266,66 @@ impl TransactionConfig {
...
+    }
+
+    fn construct_available_contracts() -> Vec<Contract> {
+        vec![
+            /*
+            Contract {
+                code: near_test_contracts::load_contract().to_vec(),
+                functions: vec![Function::BubbleSort],
+            },
+             */
```
bowenwang1996, line 318: What happened here?

test-utils/runtime-tester/src/fuzzing.rs
```diff
@@ -140,32 +173,206 @@ impl TransactionConfig {
...
+    SignerAccountPk,
+    SignerAccountId,
+    CurrentAccountId,
+    AccountBalance,
+    AttachedDeposit,
+    ValidatorTotalStake,
+    ExtSha256,
+    UsedGas,
+    WriteKeyValue,
+    WriteBlockHeight,
```
bowenwang1996, line 338: Is it intentional that `promise_*` host functions are not included? Also I wonder whether there is a way to import the host functions we have in vm logic directly to make sure that the set of host functions we fuzz is the same as the ones we actually use. cc @matklad

posvyatokum: Yes, it is intentional in two ways. First, I don't want to add promises in this stage, because it is a bit complex and adds weight to the review. Second, I was going by implemented functions in test_contract, I'm not calling host functions directly (I'm not sure if I can, maybe someone can explain to me how to do it correctly). In test contract only one function uses promises and it is, again, way to big to be included in this review.

bowenwang1996 (approved):
@matklad please review this PR

matklad:
Curions, did we try https://rust-fuzz.github.io/book/cargo-fuzz/coverage.html ?

test-utils/runtime-tester/fuzz/fuzz_targets/runtime_fuzzer.rs
```diff
@@ -2,12 +2,24 @@
...
 fuzz_target!(|scenario: Scenario| {
     let filename = "failed_scenario.json";
     let runtime_stats = scenario.run();
     if let Err(e) = runtime_stats {
         serde_json::to_writer(&File::create(filename).unwrap(), &scenario).unwrap();
         panic!("Bad scenario: {}, Error: {}", filename, e);
+    } else {
+        let stats = runtime_stats.unwrap();
+        for block_stats in stats.blocks_stats {
+            if block_stats.block_production_time > Duration::from_secs(1) {
```
matklad, line 16: :+1: :+1: :+1: 

That's a great idea, I like this

test-utils/runtime-tester/run_scenario/src/main.rs
```diff
@@ -0,0 +1,30 @@
...
+fn main() {
+    let matches = App::new("Run scenario")
+        .arg(
+            Arg::with_name("scenario")
+                .long("scenario")
+                .required(true)
+                .takes_value(true)
+                .help("Scenario to run"),
+        )
+        .get_matches();
```
matklad, line 16: We already use clap3 with derived macro for neard, so I suggest that we use derive approach for all other tools as well:

https://github.com/near/nearcore/blob/7399e0964c2d87529206f3f59d91ddb785feafb7/neard/src/cli.rs#L16-L26

It increases consitency and makes much easiter to add and document the options.

test-utils/runtime-tester/run_scenario/src/main.rs
```diff
@@ -0,0 +1,30 @@
+use clap::{App, Arg};
+use runtime_tester::Scenario;
+
+use std::path::Path;
+use std::time::{Duration, Instant};
+
+fn main() {
```
matklad, line 7: Is this just to reproduce a failure? `cargo fuzz` has a built-in feature for this (when a case fails, it prints the command you need to run to re-run this single case). I would probably default to using that instead of our own tool, but no strong opinion!

test-utils/runtime-tester/src/fuzzing.rs
```diff
@@ -1,19 +1,32 @@
...
+use std::str::FromStr;
+
+pub type ContractId = usize;
+
 pub const MAX_BLOCKS: usize = 2500;
-pub const MAX_TXS: usize = 300;
+pub const MAX_TXS: usize = 500;
+pub const MAX_TX_DIFF: usize = 10;
 pub const MAX_ACCOUNTS: usize = 100;
+pub const MAX_ACTIONS: usize = 100;
```
matklad, line 27: I'd group these into a single `Config` struct

test-utils/runtime-tester/src/fuzzing.rs
```diff
@@ -66,12 +82,27 @@ impl TransactionConfig {
...
             let signer_account = scope.random_account(u)?;
             let receiver_account = scope.random_account(u)?;
             let amount = u.int_in_range::<u128>(0..=signer_account.balance)?;
 
+            scope.accounts[signer_account.usize_id()].balance = scope.accounts
+                [signer_account.usize_id()]
+            .balance
+            .checked_sub(amount)
+            .or(Some(0))
+            .unwrap();
```
matklad, line 98:
```suggestion
            .saturating_sub(amount);
```

test-utils/runtime-tester/src/fuzzing.rs
```diff
@@ -66,12 +82,27 @@ impl TransactionConfig {
...
+            scope.accounts[receiver_account.usize_id()].balance = scope.accounts
+                [receiver_account.usize_id()]
+            .balance
+            .checked_add(amount)
+            .or(Some(Balance::MAX))
```
matklad, line 103: ditto

test-utils/runtime-tester/src/fuzzing.rs
```diff
@@ -184,19 +391,177 @@ impl Scope {
...
+                args: vec![],
+                gas: GAS_1,
+                deposit: 0,
+            }),
+            Function::BlockIndex => Ok(FunctionCallAction {
+                method_name: "ext_block_index".to_string(),
+                args: vec![],
+                gas: GAS_1,
+                deposit: 0,
+            }),
```
matklad, line 461: One way to compress such "most cases are identical, but few are not like the others" is to create a default and modify it:

```
let mut res =  FunctionCallAction {
                method_name: String::new(),
                args: vec![],
                gas: GAS_1,
                deposit: 0,
            };
match self {
  Function::StorageUsage => res.method_name = "ext_storage_usage".to_string();
  ...
}
Ok(res)
```

matklad (approved):

PR #4547, opened 2021-07-21
rename ShardId to ShardOrd

mzhangmzz:


chain/chain-primitives/src/error.rs
```diff
@@ -180,7 +180,7 @@ pub enum ErrorKind {
     InvalidBalanceBurnt,
     /// Invalid shard id
     #[fail(display = "Shard id {} does not exist", _0)]
```
bowenwang1996, line 182: maybe we should change this error message as well

bowenwang1996 (commented):
LGTM. If the PR is ready for view, please mark it as such.

chain/chain-primitives/src/error.rs
mzhangmzz: It's failing some test and I am trying to fix that now

mzhangmzz:
Will start a new one

PR #4549, opened 2021-07-22
refactor: use clap3 in parameter estimator & standalone runner

matklad:
We'll want to tweak param estimator CLI shortly, and it's easier to do that with declarative clap3. Note that we already use calp3 for the main `neard` binary, so this change also increases overall consistency of the code base.

Test Plan
-------

Smoke-tested manually using examples from readme.


ailisp (approved):
Looks good to me!

olonho (approved):

olonho:
Just fix calp ->clap in description.

PR #4553, opened 2021-07-22
Rename ShardId to ShardOrd and shard_id to shard_ord

mzhangmzz:

To avoid breaking the API, the shard_id fields in view structs are not changed


chain/indexer/src/streamer/types.rs
```diff
@@ -39,7 +39,7 @@ pub struct IndexerExecutionOutcomeWithReceipt {
 
 #[derive(Debug, Serialize, Deserialize)]
 pub struct IndexerShard {
-    pub shard_id: types::ShardId,
+    pub shard_ord: types::ShardOrd,
```
bowenwang1996, line 42: @khorolets @frol would the change here break anything? Looks like the struct derives `Serialize`

core/primitives/src/views.rs
```diff
@@ -648,7 +648,7 @@ pub struct ChunkHeaderView {
     pub encoded_length: u64,
     pub height_created: BlockHeight,
     pub height_included: BlockHeight,
-    pub shard_id: ShardId,
+    pub shard_ord: ShardOrd,
```
bowenwang1996, line 651: Looks like you removed the old field directly here, which is not desirable.

bowenwang1996 (commented):
@frol please review the changes related to rpc to make sure that we don't break any API

core/primitives/src/views.rs
mzhangmzz: hmm this changeset looks like the changeset I committed yesterday, not sure why, I thought I deleted that one. Let me try to re-submit

mzhangmzz:
created a new one

PR #4554, opened 2021-07-23
Rename ShardId to ShardOrd and shard_id to shard_ord

mzhangmzz:

To avoid breaking the API, the shard_id fields in view structs are not changed


chain/indexer/src/streamer/types.rs
```diff
@@ -39,7 +39,7 @@ pub struct IndexerExecutionOutcomeWithReceipt {
 
 #[derive(Debug, Serialize, Deserialize)]
 pub struct IndexerShard {
-    pub shard_id: types::ShardId,
+    pub shard_ord: types::ShardOrd,
```
mzhangmzz, line 42: (From @bowenwang1996) @khorolets @frol will this change here break anything? Looks like the struct derives Serialize

core/primitives/src/views.rs
```diff
@@ -648,7 +648,7 @@ pub struct ChunkHeaderView {
     pub encoded_length: u64,
     pub height_created: BlockHeight,
     pub height_included: BlockHeight,
-    pub shard_id: ShardId,
+    pub shard_id: ShardOrd,
```
bowenwang1996, line 651: I think it is better to add a new field called `shard_ord` that is synonymous to what `shard_id` means here and keep both. Later on we can deprecate `shard_id`. @frol what do you think?

bowenwang1996 (commented):
@frol please review rpc and indexer related changes

chain/client-primitives/src/types.rs
```diff
@@ -248,8 +248,8 @@ impl From<near_chain_primitives::Error> for GetChunkError {
             near_chain_primitives::ErrorKind::DBNotFoundErr(error_message) => {
                 Self::UnknownBlock { error_message }
             }
-            near_chain_primitives::ErrorKind::InvalidShardId(shard_id) => {
-                Self::InvalidShardId { shard_id }
+            near_chain_primitives::ErrorKind::InvalidShardId(shard_ord) => {
+                Self::InvalidShardId { shard_id: shard_ord }
```
frol, line 252: Same question as above, I am just leaving some pointers for the future reference

chain/indexer/src/streamer/types.rs
frol: To answer this question I need to understand whether `shard_ord` even need to be exposed to the outside world. I feel that we should treat it as an implementation detail only useful for an efficient shard look-ups. In this case, I would only expose `shard_id`, and while we are in this migration state (we have not introduced `ShardId` yet), I suggest we keep `shard_id: ShardOrd` and later change it to `shard_id: ShardId`.

WDYT?

P.S. Maybe we want to introduce `ShardId` type now, to clearly communicate the intent here.

chain/jsonrpc-primitives/src/types/chunks.rs
```diff
@@ -6,7 +6,7 @@ use serde_json::Value;
 pub enum ChunkReference {
     BlockShardId {
         block_id: near_primitives::types::BlockId,
-        shard_id: near_primitives::types::ShardId,
+        shard_id: near_primitives::types::ShardOrd,
```
frol, line 9: Just for the reference, I would apply the same reasoning about `shard_id: ShardId` I expressed above. Thus, it looks fine to me now.

core/primitives-core/src/types.rs
```diff
@@ -19,7 +19,7 @@ pub type BlockHeight = u64;
 /// Height of the epoch.
 pub type EpochHeight = u64;
 /// Shard index, from 0 to NUM_SHARDS - 1.
```
frol, line 21: Please, reference the NEP and I also suggest to clarify that `ord` is "ordinal number", just to so any maintainer can easily get the context here.

core/primitives/src/sharding.rs
```diff
@@ -81,7 +81,7 @@ pub struct ShardChunkHeaderInner {
     pub encoded_length: u64,
     pub height_created: BlockHeight,
     /// Shard index.
```
frol, line 83: Can you also extend this documentation to make it clear why we have `shard_ord`, and not `shard_id`?

core/primitives/src/views.rs
frol: See my comment above. I assume we won't move to `shard_ord` and rather hide it, so `shard_id` is the right naming.

chain/indexer/src/streamer/types.rs
mzhangmzz: @frol thanks for the detailed review! To answer your question, `shard_ord` will be exposed to the outside world, and `shard_id` is only used inside a validator's state. `shard_ord` is an ordinal number ranging from 0 to NUM_SHARDS-1 that is used on the protocol level. `shard_id` is an internal id that validators use to identify shards. The reason why they can be different is that there can be shards from different epochs after we introduce resharding. I don't think we will ever expose `shard_id`, but we should expose `shard_ord`

chain/chain-primitives/src/error.rs
```diff
@@ -180,7 +180,7 @@ pub enum ErrorKind {
     InvalidBalanceBurnt,
     /// Invalid shard id
     #[fail(display = "Shard id {} does not exist", _0)]
-    InvalidShardId(ShardId),
+    InvalidShardId(ShardOrd),
```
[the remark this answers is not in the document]
mzhangmzz: @frol, I agree we should rename the error, but I wasn't sure if renaming it would break any public facing API. If no, I will rename it.

frol:
Just for the reference: https://github.com/near/NEPs/pull/241#issuecomment-888867327

mzhangmzz:
Will abandon this PR since it is a very invasive change to rename ShardId. Will take @frol's suggestion to create a new struct ShardUId

PR #4556, opened 2021-07-23
Break the circular dependencies caused by tests

### LINKED ISSUE #4357: Consider removing circular dependencies during tests
At the moment, nearcore consists of a large number of crates which are relatively independent. However, there's one notable exception -- the `testlib` crates depends on majority of the crates, and almost every crate has `testlib` as a dev-dependency. That means that we have circular dependencies during tests. For example, if you are fixing a bug in `near-client` and run `cargo t -p near-client`, you'll be rebuilding `nearcore` and `near-json-rpc` as well. 

I think there's a benefit in removing `dev-dendencies` on `testlib`, and instead moving all tests which need access to testlib (and nearcore) from individual crates into the top-level `integration-tests` crate we have. Having tests for individual functionalities in a common uber-test crate might not look as pretty, but it'll faithfully reflect the physical architecture of our project. We should strive to test things in isolation, but we don't actually have such isolation today for many tests.

nikurt:
Fixes #4357 
The code dependencies are all over the place.
Some source files have tests, which use local test utils, which depend on testlib.
This change changes the direction of dependencies:
* The code no longer depends on test utils or testlib
* If inline tests need to depend on test utils or testlib, they are moved to `integration-tests/tests` (not the best location, will get to it later)
* `testlib` doesn't depend on test utils.
* Merged test utils into the `testlib` crate.
* Because some tests depend on non-public functions and types defined in the local crate, I simply made those functions and types public to allow the testing code to be moved. Will get back to it later.
* Tools like `state-viewer` are allowed to depend on `testlib`.
* Github refuses to show some files as moved, even though they were actually moved with `git mv`.


matklad:
Process question: reviewing such big-bang changes is quite hard. They also risk conflicting with pretty much every other PR in the queue, and might require rebasing and re-review. 

Would it be possible to execute this refactor one crate at a time? I feel that merging N smaller PRs will overall be faster than merging one giant PR

bowenwang1996:
> Process question: reviewing such big-bang changes is quite hard. They also risk conflicting with pretty much every other PR in the queue, and might require rebasing and re-review.
> 
> Would it be possible to execute this refactor one crate at a time? I feel that merging N smaller PRs will overall be faster than merging one giant PR

We can also try to move fast and finish reviewing the PR before new PRs are created :)

nikurt:
@matklad Great suggestion. Know that I know that this refactoring can be done, let's split it into parts.
Part #1 is https://github.com/near/nearcore/pull/4559

integration-tests/tests/chain/challenges.rs
```diff
@@ -0,0 +1,117 @@
+mod tests {
```
bowenwang1996, line 1: I would call these tests unit tests. They are solely within the realm of chain and does not depend on other parts of the system. In other words, they could be in the same file as `chain/chain/src/chain.rs` for example and it's just that we chose not to do it.

PR #4557, opened 2021-07-23
nit: drop unused native evm and feature

ailisp:
Just a clean up. It's depreciated since aurora evm contract

Test Plan
------------
CI


runtime/near-vm-errors/src/lib.rs
```diff
@@ -57,7 +56,6 @@ pub enum FunctionCallErrorSer {
     WasmTrap(WasmTrap),
     WasmUnknownError,
     HostError(HostError),
-    EvmError(EvmError),
```
matklad, line 60: Wait, as this wasnt' gated behind a feature, this now changes serialization format, which is a breaking change. Let's do the standard trick and rename this variant to `Unused` to preserve borsh serialization format. Let's also add a comment that this variant was, in fact, *never* used, so that we could re-use this variant in the future. 

To bad that we didn't make this variant last to begin with. Had we done so, just dropping it would have been correct.

matklad: Could you double check that this is the single place where we need to think about compatibility?

matklad (changes requested):

runtime/near-vm-errors/src/lib.rs
bowenwang1996: how come this was not gated by a feature :(

runtime/near-vm-errors/src/lib.rs
```diff
@@ -306,8 +222,6 @@ pub enum VMLogicError {
     ExternalError(Vec<u8>),
     /// An error that is caused by an operation on an inconsistent state.
     InconsistentStateError(InconsistentStateError),
-    /// An error coming from native EVM.
-    EvmError(EvmError),
```
ailisp, line 310: This one is okay to drop as it's the last variant

runtime/near-vm-errors/src/lib.rs
ailisp: @matklad I double checked there is another enum that not place evmerror as last, which need to be replaced by `_EVMError` to keep borsh correct. Also double checked every file else don't see any compatibility issue.

@bowenwang1996 Evm error was added first then evm code & feature. I think we were sure that evm will be landed so just guard evm logic code was sufficient. We didn't consider the case that evm would be removed

matklad (approved):

chefsale (approved):
LGTM, thanks for cleaning this up.

chefsale:
@ailisp Should we also remove the evm tests here: https://github.com/near/nearcore/tree/master/pytest/tests/evm?

ailisp:
@chefsale They're removed: https://github.com/near/nearcore/pull/4557/files#diff-8c7bae33a4954ef2f3a7712bf379f9f727efc01ecbe243ed4bb5e1263a7c9d17
thanks for checking!

ailisp:
@bowenwang1996 @frol @nearmax @artob @frankbraun @EgorKulikov need a lot of approvals here :)

bowenwang1996 (commented):
Thanks!

bowenwang1996 (approved):

artob (approved):

PR #4558, opened 2021-07-23
Describe `Trie` entities related to `Runtime`

Longarithm:
**Goal**

Document Trie implementation details closely related to Runtime.
Don't dig much into details, because there are some plans to rewrite it.

Moved from https://github.com/near/NEPs/pull/242


docs/trie.md
```diff
@@ -0,0 +1,77 @@
...
+We use Merkle-Patricia Trie to store blockchain state.
+Trie is persistent, which means that insertion of new node actually leads to creation of new path to this node, and thus root of Trie after insertion will also be presented by new object.
+
+Here we describe its implementation details which are closely related to Runtime.
+
+## Main structures
+
+### Trie
+
+Base structure. Almost never called directly from `Runtime`, modifications are made using `TrieUpdate`.
```
matklad, line 12: This misses "big picture" info about how the data is actually stored, while it may be obvious to maintaners, it's a crucial missing knowledge for new contributors. 

Let's say that we store trie in the rocksdb database, which is persistet across node restarts, and metion specific columns which we use for the storage. 

That is, the document focuses primarily on the in-code abstraction we use in the code to represent the trie, but it is somewhat more useful, in these sort of documents, do describe the underlying thing in the real world that the abstraction models. In this case, that would be a bit of persistent state in the rocksdb.

docs/trie.md
```diff
@@ -0,0 +1,77 @@
...
+Provides a way to access Storage and record changes to commit in the future. Update is prepared as follows:
+
+- changes are made using `set` and `remove` methods, which are added to `prospective` field,
+- call `commit` method which moves `prospective` changes to `committed`,
+- call `finalize` method which prepares `TrieChanges` and state changes based on `committed` field.
+
+Note that `finalize` does not modify `Trie` itself, it only prepares changes to be passed to `ChainStore`.
+`Trie::insert` and `Trie::update` also does not modify `Trie`. The former one puts new nodes to temporary memory object, and the latter one prepares `TrieChanges` object.
+
+`Trie` is actually updated when `ShardTries::apply_insertions` is called, which puts new values to `ColState` part of key-value database.
```
matklad, line 25: Here, the docs would also benefit from explaining *why* before *how*. So it makes sense to start with something along the lines of "we want to make changes to the trie with the transactional semantics. That is, we want to be able to start transaction, accumulate a bunch of changes, and either apply them all or rollback them all. To achieve that, we buffer all the changes in memory, then we compute the actual diff we need to apply to the data on disk, and finally we write the data out" (note that this is how I'd imagine this to work, might be completely off base). While we are at it, it will also be helpful to say *when* we start transaction. Are we running a trie transaction per block? Or are we running a trie transaction per block transaction? Can we have more than one transaction open?

Longarithm: In fact, I also don't currently know all the reasons of current implementation.
My objective is to explain some details I was closely working with and which were not easily understandable from the code, though I certainly will spend some time on improving explanations.

docs/trie.md
```diff
@@ -0,0 +1,90 @@
...
+
+Here we describe its implementation details which are closely related to Runtime.
+
+## Main structures
+
+### Trie
+
+Base structure. 
+It is stored in the RocksDB, which is persistent across node restarts. Trie communicates with database using `TrieStorage`.
+Trie data is stored in key-value format in `ColState` column, wherein nodes are stored altogether with `Vec<u8>` values.
```
bowenwang1996, line 14: Not sure what you mean by "stored altogether with `Vec<u8>` values`

bowenwang1996:
@Longarithm what is the status of this PR?

Longarithm:
@bowenwang1996 
Didn't have time to look at it, will address comments this week.

docs/trie.md
Longarithm: The point is that on the DB level, we store two kinds of DB records - trie nodes and values. I added more details about it now.

bowenwang1996 (approved):

PR #4559, opened 2021-07-23
Move `nearcore` test utils into `testlib`

nikurt:
Issue #4357

### DIFF
--- a/neard/Cargo.toml
+++ b/neard/Cargo.toml
@@ -28,6 +28,7 @@ futures = "0.3"
 nearcore = { path = "../nearcore" }
 near-primitives = { path = "../core/primitives" }
 near-performance-metrics = { path = "../utils/near-performance-metrics" }
+testlib = { path = "../test-utils/testlib" }
 
 [features]
 default = ["json_rpc"]
--- a/neard/src/cli.rs
+++ b/neard/src/cli.rs
@@ -314,7 +314,7 @@ pub(super) struct TestnetCmd {
 
 impl TestnetCmd {
     pub(super) fn run(self, home_dir: &Path) {
-        nearcore::config::init_testnet_configs(
+        testlib::nearcore_test_utils::init_testnet_configs(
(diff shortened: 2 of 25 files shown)


neard/Cargo.toml
```diff
@@ -28,6 +28,7 @@ futures = "0.3"
 nearcore = { path = "../nearcore" }
 near-primitives = { path = "../core/primitives" }
 near-performance-metrics = { path = "../utils/near-performance-metrics" }
+testlib = { path = "../test-utils/testlib" }
```
matklad, line 31: This feels like something we should not do.

In my mind, `testlib` is something we can use solely for testing, so it generally can depend on random stuff, and needn't be developed as carefully as production code. At the same time, this dependency means that we are going to link this testing code into the actual binary we ship to users.

neard/src/cli.rs
```diff
@@ -314,7 +314,7 @@ pub(super) struct TestnetCmd {
 
 impl TestnetCmd {
     pub(super) fn run(self, home_dir: &Path) {
-        nearcore::config::init_testnet_configs(
+        testlib::nearcore_test_utils::init_testnet_configs(
```
matklad, line 317: So this I think is why we have to add `testlib` dependency -- we want to be able to run testnet via `neard`. 

It means that this functionality, isn't, strictly speaking, about nearcore internal testing -- this is a public API available to the users. So it seems to me that the current placement of the `GenesisExt` makes more sense. 

Why do we want to move it to the testlib?

matklad (commented):
I feel like I don't fully understand the motivation here.

bowenwang1996 (commented):
Same concern as what @matklad mentioned. The purpose of this PR is a bit confusing to me: I thought we want to refactor tests such that `near-client` does not need to depend on `nearcore` as a dev dependency. That means we likely should move some tests in `near-client` to `nearcore`

PR #4562, opened 2021-07-23
feat: one button "run estimator in docker \w qemu" 

matklad:
Today, to get precise gas estimations via instruction counts, one needs
to follow a series of manual steps from emu-cost/README.md.

This PR optimizes part of that. It adds `--docker` argument to the
parameter estimator, which takes care of docker and qemu.

With this argument, the estimator on the host machine launches a docker
container, compiles a copy of itself in docker, and runs that copy via
qemu.

There are a couple of manual steps left:

* running neard init
* running genesis-populate
* picking the specific parameter values we use for estimation

Those will be automated in subsequent PR. The desired end goal is that

```console
$ cargo run -- --docker
```

is the *single* command you need to run the whole estimation process.
At that point, all various bash and python scripts will be removed.

### DIFF
--- a/runtime/runtime-params-estimator/README.md
+++ b/runtime/runtime-params-estimator/README.md
@@ -14,7 +14,7 @@ Use this tool to measure the running time of elementary runtime operations that 
     ```
 
     With the given parameters above estimator will run relatively fast.
-    Note the `--metric time` flag: it instructs the estimator to use wal-clock time for estimation, which is quick, but highly variable between runs and physical machines.
+    Note the `--metric time` flag: it instructs the estimator to use wall-clock time for estimation, which is quick, but highly variable between runs and physical machines.
     To get more robust estimates, use these arguments:
 
     ```bash
--- a/runtime/runtime-params-estimator/README.md
+++ b/runtime/runtime-params-estimator/README.md
@@ -13,6 +13,17 @@ Use this tool to measure the running time of elementary runtime operations that 
     cargo run --release --package runtime-params-estimator --features required --bin runtime-params-estimator -- --home /tmp/data --accounts-num 20000 --iters 1 --warmup-iters 1 --metric time
     ```
 
-    With the given parameters above estimator will run relatively fast. We will be using different parameters to do the actual parameter estimation. Also note that the default metric is `icount`, instruction count, but requires using QEMU to emulate the processor. So this example provides a way to get a quick way to test out the estimator based on time, but the instructions in [`emu-cost/README.md`](./emu-cost/README.md) should be followed to get the real data.
+    With the given parameters above estimator will run relatively fast.
+    Note the `--metric time` flag: it instructs the estimator to use wal-clock time for estimation, which is quick, but highly variable between runs and physical machines.
+    To get more robust estimates, use these arguments:
+
+    ```bash
+    --docker --home /tmp/data --accounts-num 20000 --iters 1 --warmup-iters 1 --metric icount
+    ```
+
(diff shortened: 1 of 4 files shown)


runtime/runtime-params-estimator/src/main.rs
```diff
@@ -79,5 +93,100 @@ fn main() {
...
+
+    let init = {
+        // Build a bash script to run inside the container. Concatenating a bash
+        // script from strings is fragile, but I don't know a better way.
+
+        let mut buf = String::new();
+        buf.push_str("set -ex;\n");
+        buf.push_str(
+            "\
+cargo build --manifest-path /nearcore/Cargo.toml \
```
Longarithm, line 123: Can we mount directory to `/host/nearcore` instead of `/nearcore`?
The minor issue with it is that we use `/host/nearcore` in other places, and it would be convenient to use the same folder everywhere:
- https://github.com/near/nearcore/blob/master/runtime/runtime-params-estimator/emu-cost/README.md
- https://github.com/near/nearcore/blob/master/runtime/runtime-params-estimator/src/function_call.rs#L61

runtime/runtime-params-estimator/README.md
```diff
@@ -13,6 +13,17 @@ Use this tool to measure the running time of elementary runtime operations that
     cargo run --release --package runtime-params-estimator --features required --bin runtime-params-estimator -- --home /tmp/data --accounts-num 20000 --iters 1 --warmup-iters 1 --metric time
     ```
 
-    With the given parameters above estimator will run relatively fast. We will be using different parameters to do the actual parameter estimation. Also note that the default metric is `icount`, instruction count, but requires using QEMU to emulate the processor. So this example provides a way to get a quick way to test out the estimator based on time, but the instructions in [`emu-cost/README.md`](./emu-cost/README.md) should be followed to get the real data.
+    With the given parameters above estimator will run relatively fast.
+    Note the `--metric time` flag: it instructs the estimator to use wal-clock time for estimation, which is quick, but highly variable between runs and physical machines.
```
olonho, line 17: wall

olonho (approved):

runtime/runtime-params-estimator/src/main.rs
```diff
@@ -79,5 +93,100 @@ fn main() {
...
+        .context("Failed to write runtime config to file")?;
+    Ok(())
+}
+
+fn main_docker(state_dump_path: &Path) -> anyhow::Result<()> {
+    exec("docker --version").context("please install `docker`")?;
+
+    let project_root = project_root();
+    if exec("docker images -q rust-emu")?.is_empty() {
+        // Build docked image if there isn't one already.
```
Longarithm, line 105:
```suggestion
        // Build docker image if there isn't one already.
```

Longarithm:
Can we add an option to run docker container in background and attach to it?
It saved me a lot of build time:
```
sudo docker run -d      --rm --mount type=bind,source=/home/Aleksandr1,target=/host      --cap-add=SYS_PTRACE --security-opt seccomp=unconfined      -i -t rust-emu      /usr/bin/env bash
sudo docker ps
sudo docker attach xyz
```

Longarithm:
Regarding https://github.com/near/nearcore/pull/4562#issuecomment-894334544 - discussed with @matklad offline.
The new approach reuses build data:
```
.args(&["--mount", "source=rust-emu-target-dir,target=/nearcore/target"])
.args(&["--mount", "source=rust-emu-cargo-dir,target=/usr/local/cargo"])
```
So it works great even when we run a new container instance.

PR #4567, opened 2021-07-26
Remove the dependency of `chain/client` on `nearcore` by moving its tests into `integraton_tests`

nikurt:
Issue #4357


integration-tests/tests/tests_client.rs
```diff
@@ -0,0 +1,6 @@
+mod client {
+    mod challenges;
+    mod process_blocks;
+    mod runtimes;
+    mod sandbox;
+}
```
matklad, line 6: A better layout would be 

```diff
diff --git a/integration-tests/tests/client/main.rs b/integration-tests/tests/client/main.rs
new file mode 100644
index 000000000..581e87ad1
--- /dev/null
+++ b/integration-tests/tests/client/main.rs
@@ -0,0 +1,4 @@
+mod challenges;
+mod process_blocks;
+mod runtimes;
+mod sandbox;
diff --git a/integration-tests/tests/tests_client.rs b/integration-tests/tests/tests_client.rs
deleted file mode 100644
index 9a8dd52bd..000000000
--- a/integration-tests/tests/tests_client.rs
+++ /dev/null
@@ -1,6 +0,0 @@
-mod client {
-    mod challenges;
-    mod process_blocks;
-    mod runtimes;
-    mod sandbox;
-}
```

chain/client/src/client.rs
```diff
@@ -1662,197 +1662,3 @@ impl Client {
...
-    use near_chunks::test_utils::ChunkForwardingTestFixture;
-    use near_chunks::ProcessPartialEncodedChunkResult;
-    use near_crypto::KeyType;
-    use near_primitives::block::{Approval, ApprovalInner};
-    use near_primitives::hash::hash;
-    use near_primitives::validator_signer::InMemoryValidatorSigner;
-    use near_primitives::version::ProtocolFeature;
-    use near_primitives::version::PROTOCOL_VERSION;
-    use near_store::test_utils::create_test_store;
-    use nearcore::config::GenesisExt;
```
matklad, line 1686: Hm, so this is actually the only import which needs dependency on `nearcore` -- to get a `GenesisExt`. The rest of the code here works without `nearcore` just fine. 

I wonder if the problem here is that `GenesisExt` is a part of `nearcore`? 

My understanding is that:

* The `Genesis` struct is defined in `near-chain-configs`, which is one of the leaf crates (it only depends on the primitives).
* The `GenesisExt` is defined in `nearcore`, which is *the* root crate, it depends on everything. At the same time, `GenesisExt` just adds a bunch of additional constructors to `Genesis`, it doesn't seem to fundamentally need nearcore.


So, I am wondering, can we just define the `GenesisExt` in `near-chain-configs` instead? @bowenwang1996 is there some reason why that might *not* work?

matklad: That is, I am wondering if we should explore the following alternative:

* leave these tests as they are
* move genesis ext to the near-chain-config
* remove dependency on `nearcore` without moving the tests.

integration-tests/tests/tests_client.rs
nikurt: Done

chain/client/src/client.rs
nikurt: Let me look into it.

nikurt: While combining `GenesisExt` and `Genesis` in the same module make sense and should be done eventually, `near-client` also depends on `nearcore::NightshadeRuntime`: https://github.com/near/nearcore/blob/master/chain/client/src/client.rs#L1701

matklad: Ah, right, this means that we have move these tests!

integration-tests/tests/client/main.rs
```diff
@@ -0,0 +1,4 @@
+mod challenges;
+mod process_blocks;
+mod runtimes;
+mod sandbox;
```
matklad, line 4:
```suggestion
#[cfg(feature = "sandbox")]
mod sandbox;
```

so that we don't have to gate individual tests.

integration-tests/Cargo.toml
```diff
@@ -26,3 +43,9 @@ protocol_feature_alt_bn128 = ["nearcore/protocol_feature_alt_bn128"]
 nightly_protocol_features = ["nearcore/nightly_protocol_features", "testlib/nightly_protocol_features", "protocol_feature_alt_bn128", "protocol_feature_restore_receipts_after_fix"]
 nightly_protocol = ["nearcore/nightly_protocol", "testlib/nightly_protocol"]
 protocol_feature_restore_receipts_after_fix = ["nearcore/protocol_feature_restore_receipts_after_fix"]
+sandbox = ["near-network/sandbox", "near-chain/sandbox", "node-runtime/sandbox", "near-client/sandbox"]
+
+[[test]]
+name = "sandbox"
+required-features = ["sandbox"]
+path = "tests/client/sandbox.rs"
```
matklad, line 51: I think this doesn't do what you want this todo. As written, the `tests/client/sandbox.rs` file is part of **two** test crates.
One has the root at `tests/client/sandbox.rs`, the other at `tests/client/main.rs`. 

I think you can completely remove the `[[test]]` section if you gate `mod test;` by the cfg as suggested in other ccomment.

bowenwang1996 (commented):
@matklad I don't like the fact that we make many fields in `Client` public just to be able to write tests. Do you have good suggestions on what we could do here?

integration-tests/tests/client/main.rs
nikurt: Done

integration-tests/Cargo.toml
nikurt: Done

chain/client/Cargo.toml
```diff
@@ -50,7 +50,6 @@ delay-detector = { path = "../../tools/delay_detector", optional = true }
 near-logger-utils = { path = "../../test-utils/logger" }
 near-actix-test-utils = { path = "../../test-utils/actix-test-utils" }
 testlib = { path = "../../test-utils/testlib" }
```
matklad, line 52: In the next PR, I suggest tackling this dependency, so that `chain/client` becomes free of cyclic deps.

matklad:
@bowenwang1996 I wouldn't worry about `pub` in this case. For internal code which is easy to change (that is, not public API), what matters is not specific `pub`s, but where the abstraction boundaries are. The current type already has half of its fields public, so it's pretty clear that it isn't a boundary, and adding or removing a couple of `pub`s won't change much. That is, what *would* be problematic is adding *the first* `pub` to a set of fields.

What worries me more is just the fact that tests need to care about the specific fields the type has. That means that refactoring internal representation also needs to change the tests themselves. This is a problem regardless of whether the tests live in the same crate (and can access private fields) or in another crate (and need `pub`). 

My preferred solution here is to come up with some sort of abstraction boundary, and test at this boundary, while adding observability to check for internal states of the system as well. For example, to test what `test_process_partial_encoded_chunk_with_missing_block` (the one needing access to `rs` field) is doing, I'd do the following.

* devise a an input format which depends on the protocol, but is independent of the actual implementation 
* write a universal test harness which feeds this input into the system
* write a specific instance of input which triggers the desired behaviors (partial chunks with unknown prev block hash)
* add logging call to say `need block` when we process chunk with a missing block
* use the input with the test harness and verify that the logs contain `need block`

A couple of long reads on the topic:

* https://www.tedinski.com/2019/03/19/testing-at-the-boundaries.html
* https://matklad.github.io/2021/05/31/how-to-test.html

To take of the architecture astronaut suit and get back to earth: let's tackle problems one at a time! cyclic test dependencies create a real pain, and are an easy to understand problem (if not an easy to fix one). Let's fix them first, and then come back to overall test suite design. In particcular, this PR is good to go imo!

nikurt:
Thanks @matklad for the thoughtful comment.
I totally agree that relying on internal data representation in tests is a technical debt. As you noted, fixing this is out of scope of this PR.

bowenwang1996:
Agree with what @matklad said. In fact I think `Client` should not have public fields to begin with, but that certainly requires more work to refactor it.

bowenwang1996 (approved):

PR #4573, opened 2021-07-26
Move 'standard_test_cases' to 'integration-tests'

### LINKED ISSUE #4470: Move `standard_test_cases` out of `test_utils`
https://github.com/near/nearcore/blob/master/test-utils/testlib/src/standard_test_cases.rs this file is part of testlib and gets complied & linked for every crate that uses testlib.

However, it's only needed for https://github.com/near/nearcore/blob/master/integration-tests/tests/test_cases_testnet_rpc.rs and https://github.com/near/nearcore/blob/master/integration-tests/tests/test_cases_runtime.rs. 

Those tests should probably be moved to be just two modules of a single test crate, where the root module contains the shared code. Something like

```
integration_tests/
  tests/
    standard_cases/
      main.rs # <- contains what standard_test_cases.rs now contains
      rpc.rs
      runtime.rs
```

While we are at it, we might move integration tests to a single crate as well (cc #4468).

nikurt:
Fixes #4470

Would be great to make `testlib` more light-weight with fewer dependencies. This change is the first step in that direction.


bowenwang1996 (approved):

integration-tests/tests/standard_cases/rpc.rs
```diff
@@ -1,17 +1,16 @@
 //! Runs standard test cases against TestNet with several nodes running in separate threads.
 //! The communication is performed through `RPCUser` that uses the standard RPC API to communicate.
 #[cfg(feature = "expensive_tests")]
-#[cfg(test)]
 mod test {
```
matklad, line 4: @nikurt this `mod test` here and in the other file is not necessary now. Let's move `#[cfg(feature = "expensive_tests")]` to the module declaration in the `main.rs` remove an extra level of nesting / indentation here.

PR #4575, opened 2021-07-26
pytest: respect $TEMP variable when creating /tmp/near directory

mina86:
Respect TMPDIR, TEMP and TMP environment variables when creating
a temporary /tmp/near directory.  This helps clean up any artefacts
remaining after a test.

There are other places where the scripts hardcode /tmp as the temporary
directory but for now those are left alone just to keep this change
smaller.

Issue: https://github.com/near/nayduck/issues/9

### DIFF
--- a/pytest/lib/utils.py
+++ b/pytest/lib/utils.py
@@ -181,6 +181,16 @@ def chain_query(node, block_handler, *, block_hash=None, max_blocks=-1):
                 break
 
 
+def get_near_tempdir(subdir=None, *, clean=False):
+    tempdir = pathlib.Path(tempfile.gettempdir()) / 'near'
+    if subdir:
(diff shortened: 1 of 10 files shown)


chefsale (approved):
LGTM, what is funny that we still use gcloud (https://github.com/near/nearcore/blob/master/pytest/lib/utils.py#L17) to spin up machines here, so one more reason to migrate everything to gcloud instead of azure.

As well the prometheus code is not working at all, I'm not sure what it was supposed to test or do, but this machine doesn't exists anymore `prometheus-grafana` (https://github.com/near/nearcore/blob/master/pytest/tests/stress/hundred_nodes/create_prometheus_targets.py#L17), I believe this is some obsolete code which we can delete. The way we are adding the machines there instead of using service discovery is not great either.

PR #4579, opened 2021-07-26
cleanup(version): move master version to 0.0.0 instead of 1.2.0

chefsale:
Currently our master version is 1.2.0 and we do not update it due to using a trunk based release process.

I have looked into some options and the most sensible is this (https://softwareengineering.stackexchange.com/questions/333680/how-to-version-when-using-trunk-based-development).

Use 0.0.0 for master so it doesn't create confusion as we already include the build commit in the versioning.

Before:
```
sandi@sandi-ThinkPad-X1-Carbon-7th ~/w/n/nearcore (use-trunk) [2]> ./target/release/neard --version
neard 1.2.0 (build 12fd620c) (protocol 46) (db 26)
```

Now:
```
sandi@sandi-ThinkPad-X1-Carbon-7th ~/w/n/nearcore (use-trunk)> ./target/release/neard --version
neard (version trunk) (build 5ba32a5f-modified) (protocol 46) (db 26)
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3728,7 +3728,7 @@ dependencies = [
 
 [[package]]
 name = "nearcore"
-version = "1.2.0"
+version = "0.0.0"
 dependencies = [
  "actix",
  "actix-rt",
@@ -3777,7 +3777,7 @@ dependencies = [
 
 [[package]]
 name = "neard"
-version = "1.2.0"
+version = "0.0.0"
 dependencies = [
  "actix",
  "clap 3.0.0-beta.2",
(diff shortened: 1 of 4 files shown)


mina86 (approved):
LGTM

mina86:
Our tags branch off `master` so `git describe --tags` sadly doesn’t work. Other than updating the version on each release to be something like `1.20.0-git` setting it to zero seems like the best option.

chefsale:
Landed tests fix: https://github.com/near/near-api-js/commit/26b3dd2f55ef97107c429fdfa704de566617c9b3

chefsale:
Changed also to print `(version trunk)` or `(version X.X.X)` instead of `X.X.X` only.

frol (approved):

Issue #4560, opened 2021-07-23
Create ProfileData per receipt 


matklad:
The issue was uncovered in https://near.zulipchat.com/#narrow/stream/295306-dev-contract-runtime/topic/data.20receipt.20creation.20fee/near/246980086.

Today, we store `ProfileData` (gas profiling record) in `ApplyState`:

https://github.com/near/nearcore/blob/6cc390944660354fdd28e1f52071ea03a5d5634f/core/primitives/src/runtime/apply_state.rs#L43-L45

This means that we collect a cumulative gas profile per block, which doesn't make sense: we need profile per ExecutionOutccome. 

I think the following refactor is required:

* rather than passing `ProfileData` everywhere via `ApplyAction`, create a fresh `GasProfile` in [`GasCounter::new`](https://github.com/near/nearcore/blob/6cc390944660354fdd28e1f52071ea03a5d5634f/runtime/near-vm-logic/src/gas_counter.rs#L72) -- this is the only place that uses it. 
* add `gas_profile` field to [VMOutcome](https://github.com/near/nearcore/blob/6cc390944660354fdd28e1f52071ea03a5d5634f/runtime/near-vm-logic/src/logic.rs#L2536-L2546)
* Add `gas_profile` field to `ActionResult` [here](https://github.com/near/nearcore/blob/6cc390944660354fdd28e1f52071ea03a5d5634f/runtime/runtime/src/actions.rs#L259-L276) and in [merge](https://github.com/near/nearcore/blob/6cc390944660354fdd28e1f52071ea03a5d5634f/runtime/runtime/src/lib.rs#L139). 

Before we do this, it makes sense to take a look at git history and understand, why this wasn't done in the first place -- the current design of ProfileData, which uses `Rc` and is threaded via `ApplyState` feels very roundabout to me (as opposed to creating a fresh profile in gas counter), I might be missing some rationale for that.

matklad:
@ailisp I wonder if we want to do this refactor before #4438?

ailisp:
[...]

PR #4582, opened 2021-07-27
feat(runtime): profile per receipt.

ailisp:
Very draft. Basically it does:
> 
-  rather than passing ProfileData everywhere via ApplyAction, create a fresh GasProfile in GasCounter::new -- this is the only place that uses it.
-   add gas_profile field to VMOutcome
-   Add gas_profile field to ActionResult here and in merge.

As mentioned in https://github.com/near/nearcore/issues/4560. Need more examination, consider about compatibility, fix and add tests etc.

### DIFF
--- a/core/primitives-core/src/profile.rs
+++ b/core/primitives-core/src/profile.rs
@@ -8,7 +8,7 @@ use crate::types::Gas;
 type DataArray = [Cell<u64>; ProfileData::LEN];
 
 /// Profile of gas consumption.
-#[derive(Clone)]
+#[derive(Clone, PartialEq, Eq)]
 pub struct ProfileData {
     data: Rc<DataArray>,
 }
@@ -31,6 +31,17 @@ impl ProfileData {
 
         let data = Rc::new([ZERO; ProfileData::LEN]);
         ProfileData { data }
+    }
+
+    #[inline]
+    pub fn merge(&self, other: &ProfileData) {
+        for i in 0..ProfileData::LEN {
+            let slot = &self.data[i];
+            let slot_other = &other.data[i];
+            let value = slot.get();
+            let value_other = slot_other.get();
+            slot.set(value.saturating_add(value_other));
+        }
     }
 
     #[inline]
--- a/runtime/near-vm-logic/src/logic.rs
+++ b/runtime/near-vm-logic/src/logic.rs
@@ -110,7 +110,6 @@ impl<'a> VMLogic<'a> {
         fees_config: &'a RuntimeFeesConfig,
         promise_results: &'a [PromiseResult],
         memory: &'a mut dyn MemoryLike,
-        profile: ProfileData,
         current_protocol_version: ProtocolVersion,
     ) -> Self {
         ext.reset_touched_nodes_counter();
@@ -128,7 +127,6 @@ impl<'a> VMLogic<'a> {
             max_gas_burnt,
             context.prepaid_gas,
             context.is_view,
-            profile,
         );
         Self {
             ext,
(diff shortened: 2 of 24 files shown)


ailisp:
Sorry after turn this to draft, cannot remove code owners on github. Feel free to ignore the PR until I ping you.

ailisp:
TODO: add a test apply two blocks, each have more than one txns and txn have more than one action, test profile data is properly reset

core/primitives-core/src/profile.rs
```diff
@@ -33,6 +33,18 @@ impl ProfileData {
         ProfileData { data }
     }
 
+    #[inline]
+    pub fn merge(&self, other: &ProfileData) {
+        println!("merge");
```
matklad, line 38: stray debug print

core/primitives-core/src/profile.rs
```diff
@@ -33,6 +33,18 @@ impl ProfileData {
...
+        for i in 0..ProfileData::LEN {
+            let slot = &self.data[i];
+            let slot_other = &other.data[i];
```
matklad, line 41: I'd leverage zip here:

```
for (dst, src) in self.data.iter().zip(other.data.iter()) {
    let updated = dst.get().saturating_add(src.get());
    dst.set(updated);
}
```

runtime/near-vm-logic/src/logic.rs
```diff
@@ -2533,7 +2531,7 @@ impl<'a> VMLogic<'a> {
     }
 }
 
-#[derive(PartialEq, Serialize, Deserialize, Clone)]
+#[derive(Serialize, Deserialize, Clone)]
```
matklad, line 2534: Why do we need to remove `PartialEq`? Seems proper to compare outcomes comparing profiling as well. Is this because all tests compare vm outcomes directly? I wonder if there's a better solution here...

If the isn't, I am fine with the current approach, but we should leave a comment explaining why we do it this way. 

I think we should clean this up, and the Serialize as well, but that should be left for a follow up.

runtime/near-vm-runner/src/runner.rs
```diff
@@ -115,15 +108,14 @@ pub fn run_vm(
...
             cache,
         ),
         #[cfg(not(feature = "wasmer1_vm"))]
         VMKind::Wasmer1 => panic!("Wasmer1 is not supported, compile with '--features wasmer1_vm'"),
     };
-    if let Some(VMOutcome { burnt_gas, .. }) = &outcome {
-        profile.set_burnt_gas(*burnt_gas)
+    if let Some(outcome) = &outcome {
+        outcome.profile.set_burnt_gas(outcome.burnt_gas)
     }
```
matklad, line 119: One more thing to fix in a follow up -- it doesn't make sense that we track the same amount in two places.

matklad (commented):
LGTM, not threading ProfileData everywhere feels so much better. 

Could you, after this is merge, create an issue to track further cleanups we want to do here? I think we want at leat:

* remove serialize/deserialize from vm outcome
* re-install partial eq on vm outcome
* don't track burn gas in two places
* get rid of the cells inside ProfileData

Not sure if I got everything, feel free to add more. 

But we should land this without cleanups -- this PR *enables* them in the first place.

matklad (approved):

bowenwang1996 (approved):

runtime/near-vm-logic/src/logic.rs
ailisp: In practice it's impossible that total used gas and burnt gas equals but profile data doesn't matches. Unless profile data is not working at all, but that would be caught by another test. So just comparing fields except profile data makes current tests that compare VMOutcome cleaner.

runtime/near-vm-runner/src/runner.rs
ailisp: Sounds good, profile data doesn't need track burnt gas now

runtime/near-vm-logic/src/logic.rs
ailisp: I'll leave a code comment, since no matter we clean up it or not it'll be in a follow up.

Issue #4552, opened 2021-07-22
Enable missed pytest tests


mina86:
- [ ] As part of working on https://github.com/near/nearcore/issues/2736 nine pytest tests were identified which aren’t currently enabled (see <https://github.com/near/nearcore/pull/4539>). Those tests need to be included in the test collections or documented why they aren’t.
- [x] Some tests are not present in the`nightly/complete_nayduck_set.txt` file which is currently what NayDuck is using to run on each new commit. We probably want all tests to be run on NayDuck so it may be that we need a single list file which includes all the tests. (Note that `scripts/check_nightly.py` reads `nightly/nightly.txt` only but that file isn’t used by default by NayDuck).

PR #4584, opened 2021-07-27
pytest: add a pre-commit check testing if all pytests are enabled

mina86:
Add the scripts/check_pytest.py to the ‘sanity checks’ pre-commit
tests to make sure that all pytest tests are enabled in one of the
nightly/*.txt files.

Because at the moment there are nine files which do not meet that
criteria, add them to nightly/complete_nayduck_set.txt with a comment
documenting the state.

Fixes: https://github.com/near/nearcore/issues/2736


chefsale:
Seems like there is a problem in finding the new script in the tests, please take a lot into it and should be fine.

Issue #4548, opened 2021-07-21
ecrecover and ripemd160 fees are highly inconsistent


nearmax:
We observed is that there are two benchmarks that are used to estimate the fees: one that is currently committed into nearcore and another one that was implemented in this [commit](https://github.com/near/nearcore/commit/ac53dcba). Unfortunately, these two benchmarks produce results that are 3-4 orders of magnitude different which is a big discrepancy since we can only tolerate 5-15% volatility when it comes to the fees. In order for the host function to be enabled on Mainnet we need to be sure that its cost is within 5-15% volatility (CC @bowenwang1996 ) and unfortunately, currently we have an evidence of the opposite.

Contract Runtime team will try to help figuring out what is the issue (CC @matklad ), however given a very large number of other urgent [issues with the fees](https://gov.near.org/t/list-of-issues-with-the-fees/3710), it is a good chance that this is going to be stalled for a long time, unless Aurora team steps in.

What Aurora team could do to speed up the process is to run both benchmarks on [the target hardware](https://docs.near.org/docs/develop/node/validator/hardware#recommended-hardware-specifications) and confirm that they produce the same fees. Please post results in this channel. (Note that if in the future it will turn out that ecrecover or ripemd160 are underestimated we will have to increase their fees potentially breaking already deployed EVM contracts). If benchmarks would return different fees then I would ask Aurora team to help NEAR Inc investigate. It is useful to think of param estimator as a unit test framework, because it is just a collection of benchmarks. Any issue with the fees can be an issue, with the host function, with the benchmark, or with the param estimator; similarly to how a failing unit test can indicate an issue with the code being tested, the test, or the testing framework.

Please make sure to follow the instructions here: https://github.com/near/nearcore/tree/master/runtime/runtime-params-estimator/emu-cost and check with @matklad if in doubt.

Q: Has param estimator changed between running the above two benchmarks?
A: Not that I know of, you can verify it by diffing the most recent code and the linked commit;

Q: When I run it on a AWS node, my computer, or my server, I get different results with everything.
A: IO fees are expected to be different because SSD in not emulated, unlike CPU. All other fees should stay within 5-15% accuracy. Please make sure you are running correct commands on all platforms. E.g. last time we found out that compilation with `-z` optimization level produces unpredictable results.

Q: Should we run cryptographic benchmarks from Wasm or from Rust?
[...]

PR #4587, opened 2021-07-28
Update to correct costs for new host functions

joshuajbouw:
As per https://github.com/near/nearcore/issues/4548, I ran 3 estimations for the new host function costs and have taken the average of each run then updated the values.


bowenwang1996 (commented):
It's too late. Now we need a protocol upgrade to change the cost.

matklad:
Obsolete, see action item in https://github.com/near/nearcore/issues/4778#issuecomment-925187102

PR #4589, opened 2021-07-28
feat(build) Added set -eux to all steps of the CI pipeline

chefsale:
The approach of the pre-command hook doesn't work: https://github.com/near/nearcore/pull/4589#issuecomment-888492079

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -1,7 +1,7 @@
 steps:
   - label: "cargo test"
     command: |
-      source ~/.cargo/env
+      source ~/.cargo/env && set -eux
       RUSTFLAGS='-D warnings' cargo test --locked --workspace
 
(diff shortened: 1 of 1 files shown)


matklad:
that's not pre-**commit** hook, right? That is ,there's typo in the title?

chefsale:
Wow, good catch. I don't know how did I came up with the precommand name (wuuuut)😄

chefsale:
Oh, no it is actually `precommand`based on the docs: https://buildkite.com/docs/agent/v3/hooks. That's why I forgot in the meantime @matklad

matklad (approved):

chefsale:
It doesn't seem to work, as it runs in a seperate shell and has no affect on other steps. It seems like we will need to run this in every step. I will update the PR.

PR #4590, opened 2021-07-28
nayduck: make nayduck.py a library; support `./` in scripts/check_*.py

mina86:
nightly/nayduck.py script supports a `./<path>` syntax for including
another file when reading contents of particular file.  (This is akin
to shell’s `source` command or C’s `#include` directive).  Sadly,
neither scripts/check_pytests.py nor scripts/check_nightly.py support
that syntax.

Fix this by making the nayduck.py script easier to import as a library
and use it in the aforementioned scripts.

While at it, do minor cleanups to scripts/check_nightly.py script and
move nightly.py to scripts directory.

Issue: https://github.com/near/nearcore/issues/2736


chefsale (approved):
LgTM

PR #4592, opened 2021-07-29
Remove `PhantomData` from `Genesis` and always validate `Genesis` object on creation and deserialization.

nikurt:
Issue #2459


bowenwang1996 (approved):
@posvyatokum @nikurt it seems that we have two ways of loading genesis: either directly through serde or through the stream reader. It feels a bit unnecessary to have both ways. Could we consolidate them into one? That is maybe outside the scope of this PR, but just want to bring it to your attention

integration-tests/tests/client/process_blocks.rs
```diff
@@ -3255,13 +3255,14 @@ mod storage_usage_fix_tests {
...
     use std::rc::Rc;
+    use testlib::test_helpers::wait;
 
     fn process_blocks_with_storage_usage_fix(
         chain_id: String,
         check_storage_usage: fn(AccountId, u64, u64),
     ) {
         let epoch_length = 5;
-        let mut genesis = Genesis::test(vec!["test0", "near"], 1);
+        let mut genesis = Genesis::test(vec!["test0", "near1"], 1);
```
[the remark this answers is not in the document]
nikurt: These tests used to create an account "near" which conflicted with the master account "near" and failed the validation of genesis.
AFAIK, in these tests "near" isn't used a special account, and renaming it to "test1" avoids duplicate account ids.

core/chain-configs/src/lib.rs
```diff
@@ -1,7 +1,10 @@
...
 mod genesis_config;
+pub mod genesis_validate;
 
 pub use client_config::{ClientConfig, LogSummaryStyle, TEST_STATE_SYNC_TIMEOUT};
 pub use genesis_config::{
-    Genesis, GenesisConfig, GenesisRecords, ProtocolConfig, ProtocolConfigView,
+    get_initial_supply, Genesis, GenesisConfig, GenesisRecords, ProtocolConfig, ProtocolConfigView,
 };
+#[cfg(feature = "protocol_feature_evm")]
+pub use genesis_config::{BETANET_EVM_CHAIN_ID, MAINNET_EVM_CHAIN_ID, TESTNET_EVM_CHAIN_ID};
```
matklad, line 10:
```suggestion
```

we removed `evm` feature from master branch

neard/src/cli.rs
```diff
@@ -215,7 +215,6 @@ impl RunCmd {
     pub(super) fn run(self, home_dir: &Path) {
         // Load configs from home.
         let mut near_config = nearcore::config::load_config_without_genesis_records(home_dir);
-        nearcore::genesis_validate::validate_genesis(&near_config.genesis);
```
matklad, line 218: Great removal! The conceptual distance between this call and the place where the fn is defined is large!

core/chain-configs/Cargo.toml
```diff
@@ -7,11 +7,12 @@ edition = "2018"
...
 chrono = { version = "0.4.4", features = ["serde"] }
 derive_more = "0.99.3"
+num-rational = { version = "0.3", features = ["serde"] }
 serde = { version = "1", features = ["derive"] }
 serde_json = "1"
-smart-default = "0.6"
-num-rational = { version = "0.3", features = ["serde"] }
 sha2 = "0.9"
+smart-default = "0.6"
+tracing = "0.1.13"
```
matklad, line 15: why do we need `tracing` here?

core/chain-configs/src/genesis_config.rs
```diff
@@ -164,6 +164,9 @@ impl From<&GenesisConfig> for EpochConfig {
 )]
 pub struct GenesisRecords(pub Vec<StateRecord>);
 
+/// `Genesis` has an invariant that we can't enforce due to an optimization for saving memory.
```
matklad, line 167: Would be useful to spell *which* invariant this is.

matklad (approved):
That movement of validate call is a great improvement! 

From the API design point of view, there's additional refactor we can do here make the invariant actually enforced via types, but, even if we do that, that's for a separate PR

runtime/runtime/tests/runtime_group_tools/mod.rs
```diff
@@ -205,10 +219,15 @@ impl RuntimeGroup {
                 });
                 state_records
                     .push(StateRecord::Contract { account_id, code: contract_code.to_vec() });
+                validators.push(AccountInfo {
+                    account_id: signer.account_id.clone(),
+                    public_key: signer.public_key.clone(),
+                    amount: if (i as u64) < num_existing_accounts { TESTING_INIT_STAKE } else { 0 },
```
frol, line 225: The check is already done above, so there is no need for this `if`:

```suggestion
                    amount: TESTING_INIT_STAKE,
```

frol (approved):
Thanks for refactoring the code base for the greater good!

This PR looks good to me once all the review comments are addressed

core/chain-configs/src/lib.rs
nikurt: Done

core/chain-configs/Cargo.toml
nikurt: Done.

core/chain-configs/src/genesis_config.rs
nikurt: Done

runtime/runtime/tests/runtime_group_tools/mod.rs
nikurt: Done

PR #4593, opened 2021-07-29
nayduck: add argument to `read_tests_from_file` defining how files are read

mina86:
Add a `reader` argument to `nayduck.read_tests_from_file` function such
that it’s possible to customise how the file is read.  By default, it’s
a simple read from the file system, but NayDuck needs a custom code which
will verify that paths don’t try to escape from the repository.

Issue: https://github.com/near/nayduck/issues/8


chefsale (approved):
LGTM

PR #4594, opened 2021-07-29
nightly: rework nightly/*.txt test lists; include all in nightly.txt

mina86:
Take advantage of the `./<path>` syntax in the NayDuck test lists and group
different tests that belong to the same category into separate files.
With the include syntax they can all be referenced from `nightly.txt` which
is now the file that contains all the tests NayDuck runs every 24 hours.

Issue: https://github.com/near/nayduck/issues/8


chefsale (approved):
LGTM

Issue #4595, opened 2021-07-29
1.20.0 fails to load /nearcore/nearcore/res/mainnet_genesis.json


dtp263:
**The bug**
neard fails to run with the supplied `mainnet_genesis.json`. The error message looks like this and is likely related to json parsing.
```
thread 'main' panicked at 'error while streaming records: Custom { kind: InvalidData, error: Error("trailing comma", line: 934, column: 4) }', /near/core/chain-configs/src/genesis_config.rs:435:18
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

**To Reproduce**
```
# start neard container
docker run -it nearprotocol/nearcore:1.20.0 bash

# basic init
neard init

# install some extra tools
apt-get update
apt-get install curl vim

# get mainnet json from repo
curl https://raw.githubusercontent.com/near/nearcore/1.20.0/nearcore/res/mainnet_genesis.json > /root/.near/genesis.json

# confirm that the genesis file is from mainnet
vim /root/.near/genesis.json

# try to run neard
neard run
```

**Expected behavior**
The neard software should be able to startup from the supplied json genesis file.

**Screenshots**
<img width="1393" alt="Screen Shot 2021-07-29 at 9 12 10 AM" src="https://user-images.githubusercontent.com/3021938/127517732-70fb9a65-9f5a-41af-9e1a-e5f7c2ae6d9d.png">

**Version (please complete the following information):**
- nearcore: `1.20.0`
- rust version: `using docker`
- docker: `20.10.7`
- mainnet

**Additional context**
This is blocking the 1.20.0 upgarde

posvyatokum:
The way genesis deserialisation is currently done requires GenesisRecords to be last in json. So, the reader does not expect any fields after records. And this is exactly where it fails -- there are config fields after records.
Quick fix -- rewrite genesis in this format and try again. https://github.com/near/nearcore/pull/4596

Also, the whole situation is very weird. Why would serde serialisation write genesis in such an order? It’s not alphabetical, it’s not in order of the fields. Globally we should address that on either side (ser/deser), but it may be not quick/easy enough.

bowenwang1996:
@dtp263 does it work if you run neard init on some tmp directory and copy the resulting genesis? I think the order of fields is changed in that process.

dtp263:
@bowenwang1996 the default genesis file created by running `neard init` is able to successfully startup.

[...]

PR #4596, opened 2021-07-29
Rearrange genesis fields

posvyatokum:
Quick fix for #4595 
GenesisRecords should go last in the file


chefsale (changes requested):
I dont think this should be a solution for the bug if the order is a problem, it should work with any valid json genesis file, as this is just a temporary patch. We should come up which works uniformly.

bowenwang1996:
We don't need this PR any more

PR #4598, opened 2021-07-29
pytest: prefer ‘neard’ binary name

mina86:


pytest/lib/branches.py
```diff
@@ -68,17 +68,17 @@ def compile_current():
     branch = current_branch()
     try:
         # Accommodate rename from near to neard
-        subprocess.check_output(['cargo', 'build', '-p', 'neard'])
-
-        subprocess.check_output(['cargo', 'build', '-p', 'near-test-contracts'])
```
chefsale, line 73: I don't think the --bin part is needed.

pytest/lib/branches.py
```diff
@@ -68,17 +68,17 @@ def compile_current():
...
+        subprocess.check_call(['cargo', 'build', '-p', 'neard',
+                               '--bin', 'neard'])
```
chefsale, line 72: I believe --bin neard is not needed, we can delete that part.

chefsale (changes requested):

pytest/lib/branches.py
mina86: That’s true but from my testing it does speed up the build by a couple of seconds. (We currently have `neard` and `near` binaries defined so `--bin neard` saves on linking the latter). Not a huge improvement but it’s trivially achievable so why not? Thoughts?

chefsale: I mean hard to say if that is really gonna be such an improvement, but lets keep it than. I'm fine with either.

chefsale (approved):

PR #4599, opened 2021-07-29
fix: remove potential panic with profile data fmt

austinabell:
Looking into https://github.com/near/near-sdk-rs/pull/339 I noticed that panic was fixed, but to me still seems like this line can still panic unless I'm missing some assumption about the host gas.

Also opening this to request a release at some point with this change so the fix can be applied to the SDK simulator :)


bowenwang1996 (commented):
if host_gas is 0, should we log something different? cc @ailisp

ailisp (approved):

matklad (approved):

PR #4604, opened 2021-07-30
ignore fields after records too

posvyatokum:
Read records from genesis.json properly, even if fields are in arbitrary order.
 #4595


matklad (approved):
LGTM. Should we add a test here?

core/chain-configs/src/genesis_config.rs
```diff
@@ -293,18 +293,26 @@ impl<'de, F: FnMut(StateRecord)> Visitor<'de> for RecordsProcessor<&'_ mut F> {
...
+        let mut has_records_field = false;
         while let Some(key) = map.next_key::<String>()? {
             match key.as_str() {
                 "records" => {
-                    map.next_value_seed(self)?;
-                    return Ok(());
+                    map.next_value_seed(
+                        s.take().ok_or_else(|| de::Error::custom("duplicate field: records"))?,
+                    )?;
+                    has_records_field = true;
```
matklad, line 304: (sorry if the mesasge is duplicated, there was some glitch when submitting the comment)

nit: this is not as readable as it could be. The specific issue is that the inner expression, `s.take().ok_or_else(|| de::Error::custom("duplicate field: records"))?,`, has non-trivial control-flow (exiting the function on duplicates), and it is hidden in an outre function call.

It's better if control flow has a dediccated statemen. So, I'd written this as 

```
let s = s.take().ok_or_else(|| de::Error::custom("duplicate field: records"))?;
map.next_value_seed(s);
```

I also like using names `me` or `this` for self, but variable.

core/chain-configs/src/genesis_config.rs
```diff
@@ -293,18 +293,26 @@ impl<'de, F: FnMut(StateRecord)> Visitor<'de> for RecordsProcessor<&'_ mut F> {
     where
         A: MapAccess<'de>,
     {
+        let mut me = Some(self);
```
bowenwang1996, line 296: Do we actually need this? It seems to me that we use this to check whether we have already processed `records`, but that could be achieved by checking whether `has_records_field` is true

bowenwang1996 (approved):

core/chain-configs/src/genesis_config.rs
posvyatokum: has_records_field is needed to check whether the file has records field at all. Proper config should have exactly on "records" field

posvyatokum: It's a workaround for rust ownership principles. I know, that "map.next_value_seed" will be called exactly once, but compiler doesn't.

PR #4612, opened 2021-07-30
Add tests for ecrecover

abacabadabacaba:


abacabadabacaba:
What is the proper fix for `cargo-deny` error?

joshuajbouw:
Looks to be failing because of the double entry for `darling` crate. Possibly just run `cargo update`.

joshuajbouw (approved):
Looks good. Obviously, there is that Cargo failure that needs fixing.

bowenwang1996:
> What is the proper fix for `cargo-deny` error?

Check to see if this can be avoided. It not, then add the offending crate to deny.toml

bowenwang1996 (approved):
Please fix CI

abacabadabacaba:
Now it's failing because of the test `test_external_call_indirect`. I didn't change anything that should have affected it. How to fix it?

ailisp:
@abacabadabacaba I think it's related to you have cargo.lock updated, which updates a few wasm manipulation libs, which cause a minor injected-wasm difference. I had your branch locally and test passed after doing this:

```
10053  git checkout master
10054  git pull
10055  cp ~/Cargo.lock /tmp
10056  git checkout ecrecover-tests
10057  mv /tmp/Cargo.lock .
10058  cargo t -p near-vm-runner --tests tests::error_cases::test_external_call_indirect -- --nocapture
```

ailisp (approved):
Looks good to me!

runtime/near-vm-logic/Cargo.toml
```diff
@@ -31,7 +31,8 @@ near-vm-errors = { path = "../near-vm-errors", version = "3.0.0" }
 bn = { package = "zeropool-bn", version = "0.5.9", features = [], optional = true }
 
 [dev-dependencies]
-serde_json = {version= "1", features= ["preserve_order"]}
+serde_json = { version = "1", features = ["preserve_order"] }
+serde_with = { version = "^1.9", features = ["hex"] }
```
matklad, line 35: I personally would avoid adding this dependency. I don't feel strongly about this (-0), so I am fine with having it if you prefer. 

Here's the rational (wordy not because I hate this, but because I want to make sure we are on the same page regarding the tradeoffs):

* this is a new depency for the project, it isn't already present in Cargo.lock (new deps increase compile time, number of maintainers we need to trust, risks that `cargo update` fails due to someone messing up semver)
* this dependency is used only in tests, and in a single test (the benefit we get out of it is small in scope)
* this is "convenience" dependency -- it doesn't contain any meaningful computational logic, and just helps to make the syntax nicer (it tradeoffs amount of code you need to write for API surface you need to understand, which is zero-sumish). 
* this is a "proc macro on proc macro" dependency, and proc macros are inherently brittly and poorly composable part of the Rust language (so, its hard to understand how this thing actually works). 

What I'd do instead is define 

```rust
#[derive(Deserialize)]
struct EcrecoverTest {
    m: String,
    v: u8,
    sig: String,
    mc: bool,
    res: Option<String>,
}
```

and then use the `hex` library (which is already present in Cargo.lock, and which encapsulate the actual decode hex computation logic) in the test to manually decode the free fields. Alternatively, I'd use `https://serde.rs/field-attrs.html#deserialize_with` directly.

matklad, line 35:
```suggestion
serde_with = { version = "1.9", features = ["hex"] }
```

Don't use `^` in Cargo.tomls, it is implied by default, and spelling it explicitly is non-idiomatic.

runtime/near-vm-logic/tests/test_miscs.rs
```diff
@@ -559,51 +563,54 @@ fn test_ripemd160() {
...
+    #[serde_as(as = "Option<Hex>")]
+    res: Option<[u8; 64]>,
+}
+
 #[test]
 fn test_ecrecover() {
-    let mut logic_builder = VMLogicBuilder::default();
-    let mut logic = logic_builder.build(get_context(vec![], false));
+    for EcrecoverTest { m, v, sig, mc, res } in
+        from_slice::<'_, Vec<_>>(read(Path::new("tests/ecrecover-tests.json")).unwrap().as_slice())
```
matklad, line 582: Could you add a comment explaining what's the source for this test file, if there's any?

matklad, line 582:
```suggestion
        from_slice::<'_, Vec<_>>(fs::read("tests/ecrecover-tests.json").unwrap().as_slice())
```

* no need to explicitly cast to Path
* it's idiomatic to qualify things from fs with fs

matklad, line 582: Given that the `for` header is split by formatter into two lines, I think this code would benefit from an introduction of a couple of temporary variables :)

matklad (approved):
:+1: for the overall change `for` looping over array of data is a great way to write the tests, we need more of that!

-0 for Cargo.toml/Cargo.lock changes, but no strong opinion here.

Left couple of comments on Rust idioms, nothing blocking.

PR #4613, opened 2021-07-30
tests: remove write_random_value smart contract function

mina86:
There’s really no point to it other than reducing amount of code in
the test contract and we know that less code is always better.


bowenwang1996 (changes requested):
No this function was specifically written for a test that depends on a function call that gives a different output based on the block in which it is executed. Please do not remove this function

mina86:
There’s also `write_block_height` and I thought that was serving that purposes. Though now that I look at `write_random_data` it might be just testing `random_seed`.

PR #4626, opened 2021-08-03
nit(runtime): get rid of cells inside profile data

ailisp:
As we don't pass profiledata everywhere now, we do not need mutate immutable reference every where, so cell can be dropped

Test Plan
----------
CI


bowenwang1996 (approved):

matklad:
Sooo much better now, thanks!

I think we no longer need the `DataArray` type alias: if you think the same, might make sense to remove it.

matklad (approved):

PR #4631, opened 2021-08-04
pytest: enable `rpc_light_client_execution_outcome_proof.py` test

mina86:
Run the test when the file is executed via `python` (rather than
`pytest`) so that the test is actually run when the file is processed
by NayDuck.


bowenwang1996 (commented):
how is this specific test different from others? I believe we don't do `if __name__ == '__main__'` for most tests

mina86:
In all tests we either do `if __name__ == '__main__'` or we don’t wrap the code in functions and just execute it without the guard. This test wrapped the test body in a test function but then never executed it. This works if the test is executed via `pytest` but it doesn’t when it’s called via `python` as is the case with NayDuck.

bowenwang1996 (approved):

PR #4635, opened 2021-08-04
pytest: refactor cluster.start_cluster

mina86:
Use os.listdir rather than unnecessarily forking and calling find.
Furthermore, remove polling from the threads spinning up the nodes.

### DIFF
--- a/pytest/lib/cluster.py
+++ b/pytest/lib/cluster.py
@@ -3,6 +3,7 @@ import threading
 import subprocess
 import json
 import os
+import pathlib
 import sys
 import signal
 import atexit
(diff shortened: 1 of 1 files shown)


pytest/lib/cluster.py
```diff
@@ -836,28 +837,29 @@ def start_cluster(num_nodes,
     if not config:
         config = load_config()
 
-    if not os.path.exists(os.path.expanduser("~/.near/test0")):
+    dot_near = pathlib.Path.home() / '.near'
```
bowenwang1996, line 840: What is this magic :)

bowenwang1996 (approved):

pytest/lib/cluster.py
mina86: `pathlib.Path` overrides the `/` operator. For `path` being a `pathlib.Path`, `path / string` is morally equivalent to `pathlib.Path(os.path.join(path, string))`.

PR #4636, opened 2021-08-04
refactor: restructure estimator to allow computing a subset of costs

matklad:
The main idea here is to *not* return `RuntimeConfig` from `run` (which
requires us to compute all the costs, and fake out a bunch of unrelated
fields). Instead, we return a flat BTreeMap<String, Gas>, wrapped in a
nice API (`CostTable`).

We serialize the map into costs.txt file in a hand-rolled format
optimized for human readability and diffability. The actual config is
then restored out of this costs.txt file.

This is not yet implemented, but the idea is to make costs.txt a source
of truth for costs (we can just include_str! it into the binary).

While we do not implement computation of subset of costs in this PR
either, this now becomes possible -- a cost table can contain only a
subset of costs.

Finally, as a side bonus, this PR moves all actual logic for computing
the fees into the `run` method. While we do use `Measurements` for most
of the costs, we compute compilation, wasm instruction and trie node
touching costs in a special way.

The `costs.txt` file included in this PR is reconstructed from the
current RuntimeConfig.

Test Plan
------------

Manual run of the estimator to make sure that results "make sense"


runtime/runtime-params-estimator/src/cases.rs
```diff
@@ -616,16 +610,21 @@ pub fn run(mut config: Config, only_compile: bool) -> RuntimeConfig {
...
     }
 
-    get_runtime_config(&m, &config)
+    let contract_compile_costs = compute_compile_cost_vm(config.metric, config.vm_kind, false);
+    let wasm_instr_cost = ratio_to_gas(config.metric, cost_per_op(config.metric, &CODE_1M));
+
+    let mut cost_table = measurements_to_costs(m, contract_compile_costs, wasm_instr_cost);
+
+    if let Some(gas) = cost_table.get(Cost::StorageReadBase) {
+        // TODO: Actually compute it once our storage is complete.
```
olonho, line 619: Not sure how touching cost can be several time less than read operation.

runtime/runtime-params-estimator/src/cost.rs
```diff
@@ -0,0 +1,140 @@
+use std::convert::TryFrom;
+use std::fmt;
+use std::str::FromStr;
+
+/// Kinds of things we measure in parameter estimator and charge for in runtime.
+///
+/// TODO: Deduplicate this enum with `ExtCosts` and `ActionCosts`.
+#[derive(Copy, Clone, PartialEq, Eq, Debug, PartialOrd, Ord)]
+#[repr(u8)]
+pub(crate) enum Cost {
```
olonho, line 10: Guess we shall have stable string name for each cost.

runtime/runtime-params-estimator/src/cost.rs
```diff
@@ -0,0 +1,140 @@
...
+    StorageIterNextKeyByte,
+    StorageIterNextValueByte,
+    TouchingTrieNode,
+    PromiseAndBase,
+    PromiseAndPerPromise,
+    PromiseReturn,
+    ValidatorStakeBase,
+    ValidatorTotalStakeBase,
+
+    #[cfg(feature = "protocol_feature_alt_bn128")]
```
olonho, line 84: Do not think that any element in this table shall be under feature.

runtime/runtime-params-estimator/src/cost_table.rs
```diff
@@ -0,0 +1,78 @@
...
+            if let Some(gas) = self.get(cost) {
+                let gas = separate_thousands(gas);
+                writeln!(f, "{:<35} {:>25}", cost.to_string(), gas)?
+            }
+        }
+        Ok(())
+    }
+}
+
+fn separate_thousands(mut n: u64) -> String {
```
olonho, line 61: Cool, thanks

runtime/runtime-params-estimator/src/costs_to_runtime_config.rs
```diff
@@ -0,0 +1,152 @@
...
+        ..RuntimeConfig::default()
+    };
+    Ok(res)
+}
+
+fn runtime_fees_config(cost_table: &CostTable) -> anyhow::Result<RuntimeFeesConfig> {
+    let fee = |cost: Cost| -> anyhow::Result<Fee> {
+        let total_gas =
+            cost_table.get(cost).with_context(|| format!("undefined cost: {}", cost))?;
+        // Split the total cost evenly between send and execution fee.
```
olonho, line 43: I was always curious if this logic is correct.

runtime/runtime-params-estimator/src/main.rs
```diff
@@ -36,12 +43,41 @@ struct CliArgs {
...
+
     let cli_args = CliArgs::parse();
 
     let state_dump_path = cli_args.home.unwrap_or_else(|| get_default_home().into());
+
+    if let Some(path) = cli_args.costs_file {
+        let cost_table = fs::read_to_string(&path)
+            .ok()
+            .and_then(|it| it.parse::<CostTable>().ok())
+            .with_context(|| format!("Failed to pares {}", path.display()))?;
```
olonho, line 63: to parse

runtime/runtime-params-estimator/README.md
```diff
@@ -15,4 +15,11 @@ Use this tool to measure the running time of elementary runtime operations that
 
     With the given parameters above estimator will run relatively fast. We will be using different parameters to do the actual parameter estimation. Also note that the default metric is `icount`, instruction count, but requires using QEMU to emulate the processor. So this example provides a way to get a quick way to test out the estimator based on time, but the instructions in [`emu-cost/README.md`](./emu-cost/README.md) should be followed to get the real data.
 
+3. The result of the estimator run is the `costs-$timestamp$.txt` file, which contains human-readable representation of the costs.
```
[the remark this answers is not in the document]
matklad: Yeah, I am torn between using a standard format, like JSON or TOML or roll-your-own format. In the end, I've picked the later approach -- we are going to look at the costs a lot, so a benefit of nicely formatted an aligned table is substantial. 
To clarify, the current code fully supports serializing and deserializing the `CostTable`, via `Display` and `FromStr` impls. 

I suggest sticking to current representation for the time being. Switching to JSON won't be hard:

Adding JSON support would be pretty easy:

```diff
diff --git a/runtime/runtime-params-estimator/Cargo.toml b/runtime/runtime-params-estimator/Cargo.toml
index 446607747..0ff8417d7 100644
--- a/runtime/runtime-params-estimator/Cargo.toml
+++ b/runtime/runtime-params-estimator/Cargo.toml
@@ -21,6 +21,7 @@ borsh = "0.8.1"
 num-rational = "0.3"
 anyhow = "1"
 chrono = "0.4"
+serde = { version = "1", features = ["derive"] }
 
 genesis-populate = { path = "../../genesis-tools/genesis-populate"}
 near-chain-configs = { path = "../../core/chain-configs" }
diff --git a/runtime/runtime-params-estimator/src/cost.rs b/runtime/runtime-params-estimator/src/cost.rs
index 7d19a72b6..eea949c65 100644
--- a/runtime/runtime-params-estimator/src/cost.rs
+++ b/runtime/runtime-params-estimator/src/cost.rs
@@ -5,7 +5,7 @@ use std::str::FromStr;
 /// Kinds of things we measure in parameter estimator and charge for in runtime.
 ///
 /// TODO: Deduplicate this enum with `ExtCosts` and `ActionCosts`.
-#[derive(Copy, Clone, PartialEq, Eq, Debug, PartialOrd, Ord)]
+#[derive(Copy, Clone, PartialEq, Eq, Debug, PartialOrd, Ord, serde::Serialize, serde::Deserialize)]
 #[repr(u8)]
 pub(crate) enum Cost {
     ActionReceiptCreation,
diff --git a/runtime/runtime-params-estimator/src/cost_table.rs b/runtime/runtime-params-estimator/src/cost_table.rs
index 828aaddcd..e3f6b4826 100644
--- a/runtime/runtime-params-estimator/src/cost_table.rs
+++ b/runtime/runtime-params-estimator/src/cost_table.rs
@@ -3,14 +3,16 @@ use std::fmt;
 use std::str::FromStr;
 
 use near_primitives::types::Gas;
+use serde::{Deserialize, Serialize};
 
 use crate::cost::Cost;
 
 /// For each [`Cost`], the price of a single unit in [`Gas`].
 ///
 /// This is the ultimate result of a parameter estimator run.
-#[derive(Default)]
+#[derive(Default, Serialize, Deserialize)]
 pub struct CostTable {
+    #[serde(flatten)]
     map: BTreeMap<Cost, Gas>,
 }
 
```

runtime/runtime-params-estimator/src/costs_to_runtime_config.rs
matklad: The current formula seems correct to me: we split the cost in half, one for execution, one for sending (and sending can be cross-shard or same shard). What is not cleart to me is that why we have two different send fees. It seems that we want `send_sir + execution = total_gas` and `send_not_sir + execution = total_gas` to hold, which means that both sends should be equal.

runtime/runtime-params-estimator/src/cases.rs
matklad: Yeah, the correctness here is questionable, especially given that we didn't enable IO costs. But the logic seems fine to me -- a single storage_read operation needes to touch several trie nodes.

runtime/runtime-params-estimator/src/cost.rs
matklad: I don't think we need a super-stable name (we can change the name later), so just using enum name seems fine.

runtime/runtime-params-estimator/src/main.rs
```diff
@@ -141,7 +180,7 @@ cargo build --manifest-path /host/nearcore/Cargo.toml \
         buf
     };
 
-    let nearcore = format!("type=bind,source={},target=/nearcore", project_root.to_str().unwrap());
+    let nearcore = format!("type=bind,source={},target=/host/nearcore", project_root.to_str().unwrap());
```
matklad, line 183: forgot to update this one in https://github.com/near/nearcore/pull/4562

runtime/runtime-params-estimator/src/cases.rs
olonho: So what's the meaning of "touching" then? Anyway, not to relevant for this PR.

matklad: My understanding (guess, really) is that `read` operation reads a key from a trie search tree. To read a key from search tree, you need to traverse several interior nodes. Touch describes reading a single node from rocks.

olonho (approved):

PR #4637, opened 2021-08-04
pytest: remove support for preexisting cluster

mina86:
The preexisting cluster is not running and there is no plan to make it
so.  Remove the convoluted code which tries to connect to non-existent
NayDuck endpoints in hope of accessing the preexisting cluster.

Issue: https://github.com/near/nayduck/issues/1


bowenwang1996 (commented):
> there is no plan to make it
so

Is that the case? I was under the impression that we wanted to do this

mina86:
That would be news to me.

Note that there are a few different types of nodes:
* local nodes are ones test starts on the host the test is running (all but mocknet tests use this type);
* mocket nodes are ones running 24/7 on GCP (only mocknet nodes use those);
* remote nodes are ones where test spins up new spot virtual machines on GCP and starts nodes there (no tests do this and honestly I have no idea if this feature actually works);
* preexisting nodes are ones where test grabs some of the NayDuck worker machines and starts nodes there (this is not implemented).

It’s not clear what problem preexisting cluster tries to solve. For vast majority of tests running a local cluster is perfectly reasonable approach. If that’s not feasible for some reason spinning up virtual machines is also an option.

With NayDuck moving to GCP and spot instances adding support for preexisting cluster introduces complexity in NayDuck for gain I don’t really see.

chefsale (approved):
I think @mina86 is right here, I think we can just nuke this as we only run on "local machines" or however we want to call them, but essentially nayduck workers we currently run we run tests on the same hosts. Maybw we will want remote hosts in the future, but not in the way it is done currently.

PR #4455, opened 2021-07-02
WIP: Recompute gas costs considering precompilation

Longarithm:
**Plan**

Run `runtime-param-estimator` and update `RuntimeConfig` fees.

Issue #4410, opened 2021-06-24
Experiment with wasm-global based gas metering


matklad:
One of the things contract runtime dose is measuring how many wasm instructions a given contract executed, and stopping the contract early if it exceeds the budget. The current implementation for this logic lives here: 

https://github.com/near/nearcore/blob/8ebff13a53db01a121b5b729c582a5450bbabb98/runtime/near-vm-runner/src/prepare.rs#L60-L66

The way it works is by using `pwasm` to inject calls to the `gas(u32)` host function at every basic block entry. 

Calling a host function is a relatively heavy operation, so this style of gas metering slows down execution a lot (>order-of-magnitude). See the rough benchmarks here:

https://gov.near.org/t/on-running-some-wasm-contracts-with-llvm-specifically-for-evm/2869/5?u=matklad

There is alternative, faster way to account for gas, which bypasses host functions. The idea is to add a gas allowance mutable global to the wasm module after validation, and instrument basic blocks to immediately deduct allowance. This is how wasmer metering middleware works:

https://github.com/wasmerio/wasmer/blob/6773109fe0cf2fea78376939e084c097d37a9629/lib/middlewares/src/metering.rs

This approach is safe -- because the module is validated before instrumentation, it can not access the global directly. Wasm doesn't have instructions for indirect access global. The fact that wasm globals & locals are managed is a part of wasm security model, so this is unlikely to change and can be relied on (see gov.near.org discusion for additional links). 

We should see if we can use this approach to significantly speed up our metering. The rough plan:

* double check that an address of global in wasm can't be escaped
* implement wasmer-style instrumentation as a function transforming wasm to wasm. Note that we don't want to use wasmer middleware directly, to not depend on a specific engine. Not sure what's the right tool here -- I suggest looking into the bytecodealliance wasm-tools ecosystem and also looking for existing alternative implementations. 
* integrate this style of gas metering with our host function mechanism -- host functions also use gas, so we should deduct from wasm global upon returning from a host function. 
* using various sepecific and randomly generated contracts, verify that both style of instrumentations yield exactly same gas costs.

EDIT: also we should take a closer look at the approach from https://github.com/near/nearcore/issues/1057

nearmax:
[...]

PR #4639, opened 2021-08-05
refactor: make it clear that regular_op_cost uses the multiplier

matklad:
The numeric value of the cost stays exactly the same (3856371 is evenly
divisible by 3).

Looking at the recent measurements in

https://github.com/near/nearcore/pull/4455#issuecomment-874790365

it's clear that the measured cost is 3x lower.

Test Plan:
--------

Just visual inspection that the constant stays the same.


bowenwang1996 (approved):
Do we actually want this multiplier for `regular_op_cost` though?

matklad:
I think we want to keep it until at least #4410 is concluded.

PR #4640, opened 2021-08-05
feat: allow computing a subset of costs in the estimator

matklad:
You can now pass something like

  --metrics-to-measure Receipt,data_receipt_10b_1000,data_receipt_base_10b_1000

to calculate only those metrics. Note that metrics != costs. The end
goal is to do this for actual costs, as outlined in

  near#4633 (comment)

The current approach is an intermediate step in that direction. In
particular, it allows to debug estimator refactors much faster.

Test Plan
---------

Manually run estimator to smoke test that this works, use a tool
([SSR](https://rust-analyzer.github.io/manual.html#structural-search-and-replace))
to change extraction logic without introducing typos.


runtime/runtime-params-estimator/src/cases.rs
```diff
@@ -294,7 +294,7 @@ pub fn run(mut config: Config, only_compile: bool) -> CostTable {
     let mut f = || {
         let account_idx = loop {
             let x = rand::thread_rng().gen::<usize>() % config.active_accounts;
-            if !deleted_accounts.contains(&x) & &!beneficiaries.contains(&x) {
+            if !deleted_accounts.contains(&x) && !beneficiaries.contains(&x) {
```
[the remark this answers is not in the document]
olonho: Yeah, nice find.

runtime/runtime-params-estimator/src/ext_costs_generator.rs
```diff
@@ -15,16 +15,25 @@ impl ExtCostsGenerator {
...
     fn extract(&mut self, metric: Metric, ext_cost: ExtCosts) {
-        let gas = self.extract_value(metric, ext_cost);
-        self.result.insert(ext_cost, gas);
+        if let Some(value) = self.extract_value(metric, ext_cost) {
+            self.result.insert(ext_cost, value);
+        }
+    }
+
+    fn extract_max(&mut self, metric1: Metric, metric2: Metric, ext_cost: ExtCosts) {
+        if let Some(value1) = self.extract_value(metric1, ext_cost) {
```
olonho, line 32: Shalln't we report error if one is present and another is not?

matklad: I think it's fine to not report the error, as just we won't compute the cost then (that is, we won't return a wrong cost). That's essentially the same behavior we have for other complex costs computed out of several metrics -- if you omit the metric, the whole cost is omitted. 

The right behavior here is to just figure out what you need to compute to get the cost, and compute just that. That's a larger refactor that is being done #4733. The goal here is primaraly to be able to quickly test if #4733 returns roughly the same results.

runtime/runtime-params-estimator/src/testbed_runners.rs
```diff
@@ -200,6 +215,12 @@ where
             Arc::new(Mutex::new(RuntimeTestbed::from_state_dump(&config.state_dump_path)))
         }
     };
+
+    if config.should_skip(metric) {
+        println!("skipped");
```
olonho, line 220: Maybe remove print?

olonho (approved):
LGTM, few nits.

PR #4642, opened 2021-08-05
deny: remove unnecessary deny skip rules

mina86:


chefsale (approved):
LGTM, but adding @frol as a sanity checker.

Issue #4618, opened 2021-08-02
Need to fix broken NayDuck tests


mina86:
The following NayDuck tests are currently broken and disabled:

*   [x] [https://github.com/near/nearcore/issues/5027](https://github.com/near/nearcore/issues/5027)
*   [x] `expensive integration-tests client test_gc_after_state_sync`
*   [x] `expensive integration-tests test_rejoin test::test_4_20_kill2`
*   [x] `expensive near-client catching_up tests::test_all_chunks_accepted_1000_slow`
*   [x] `expensive near-client catching_up tests::test_chunk_grieving`
*   [ ] `expensive near-client near_client tests::cross_shard_tx::test_cross_shard_tx` et al
*   [x] `expensive nearcore test_tps_regression test::test_highload`
*   [ ] `pytest adversarial/gc_rollback.py`
*   [ ] `pytest sanity/garbage_collection_sharding_upgrade.py` (<[https://github.com/near/nearcore/issues/5095](https://github.com/near/nearcore/issues/5095)>)
*   [x] `pytest sanity/restaked.py` (<[https://github.com/near/nearcore/issues/5503](https://github.com/near/nearcore/issues/5503)>)
*   [x] `pytest sanity/restart.py`
*   [x] `pytest sanity/repro_2916.py`
*   [x] `pytest sanity/rpc_finality.py`
*   [ ] `pytest sanity/state_migration.py`
*   [ ] `pytest sanity/state_sync2.py`
*   [x] `pytest sanity/state_sync_fail.py` (<[https://github.com/near/nearcore/issues/5182](https://github.com/near/nearcore/issues/5182)>)
*   [ ] `pytest sanity/state_sync_late.py` (keeps timing out)
*   [x] `pytest sanity/state_sync_massive.py`
*   [x] `pytest sanity/sync_chunks_from_archival.py`
*   [ ] `pytest sanity/sync_ban.py`
*   [x] `pytest sanity/validator_switch_key.py` (<[https://github.com/near/nearcore/issues/2949](https://github.com/near/nearcore/issues/2949)>)
*   [ ] `pytest stress/network_stress.py`

They need to be fixed or deleted if they are no longer needed.

PR #4644, opened 2021-08-05
nightly: fix `test_gc_after_state_sync` test

mina86:
The test name in the `expensive` nightly tests must match exactly
the path of the path otherwise it won’t match any known tests.
The `test_gc_after_state_sync` is in `process_blocks` module thus
it requires `process_blocks::test_gc_after_state_sync` to be used.


mina86:
http://nayduck.eastus.cloudapp.azure.com:3000/#/run/1951

chefsale (approved):

PR #4651, opened 2021-08-06
refactor(chain): expunge `rocksdb` from `near-jsonrpc-primitives` dependency tree

miraclx:
Under the initiative to build `near-api-rs`, we need to ensure its dependencies from within `nearcore` are as lightweight as possible. On initial inspection, `near-jsonrpc-primitives`, which should be lightweight enough to be used across the ecosystem, actually isn't, as it's dependency tree currently includes `rocksdb`. To remedy this, we extract useful structures from `near-chunks` and `near-network` into their own `-primitives` counterpart crates and depend on those instead.

### New Crates

- `near-chunks-primitives` from `near-chunks`
- `near-network-primitives` from `near-network`

### Analysis

Dump of `cargo tree -p near-jsonrpc-primitives | wc -l`:

```console
   624 master (before this refactor)
   610 after introducing `near-chunks-primitives` (`rocksdb` still present)
   494 after introducing `near-network-primitives` (bye, bye, rocksdb)
   459 after refining `near-network{,-primitives}`
   457 after moving metrics back to `near-network` (thanks, @bowenwang1996)
   417 after removing `metric_recorder` and moving it's dependent back to `near-network` 
```

#### Test Plan

- [X] Ensure unit tests pass

<sub> <i> An unfortunate consequence of this kind of code movearound is that git looses track of blame, but for reference, modules in either of the new crates that git thinks of as a "new file", was extracted from their original counterparts. </i> </sub>


chain/network/src/routing.rs
```diff
@@ -22,237 +20,22 @@ use near_store::{
...
 use crate::{
-    cache::RouteBackCache,
     types::{PeerIdOrHash, Ping, Pong},
     utils::cache_to_hashmap,
 };
 use conqueue::{QueueReceiver, QueueSender};
 #[cfg(feature = "delay_detector")]
 use delay_detector::DelayDetector;
 
+pub use near_network_primitives::routing::*;
```
frol, line 30: Thinking about this `*`, I think the owners of near-network should revisit the structure and draw a better line between `near-network-primitives` and `near-network` since currently, having two `routing.rs` files it is not immediately clear which one should be extended with the new types (I am talking about hypothetical future need to extend `near-netwok`)

chain/network/src/types.rs
```diff
@@ -1,1050 +1,10 @@
...
-
-#[derive(Message, Clone, Debug)]
-#[rtype(result = "()")]
-pub struct SendMessage {
-    pub message: PeerMessage,
-}
+pub use near_network_primitives::types::*;
 
 /// Actor message to consolidate potential new peer.
 /// Returns if connection should be kept or dropped.
```
frol, line 10: I feel that all the Actor message types should be part of -primitives. Does `Consolidate` use something that blocks us from moving it to the primitives?

frol (approved):

chain/network/src/types.rs
miraclx: Yeah, it does...

1. [`near_network::types::Consolidate`](https://github.com/near/nearcore/blob/a44be206018b547f53cb3b704f3a822e0d9d5db0/chain/network/src/types.rs#L1051-L1052)

  ```rust
  pub struct Consolidate { // `Consolidate` wasn't extracted because 👇🏽
      pub actor: Addr<Peer>, // <-- `Peer` wasn't extracted, because 👇🏽
  ...
  }
  ```

2. [`near_network::peer::Peer`](https://github.com/near/nearcore/blob/a44be206018b547f53cb3b704f3a822e0d9d5db0/chain/network/src/peer.rs#L173)

  ```rust
  pub struct Peer {
  ...
      peer_manager_addr: Addr<PeerManagerActor>, // <-- `PeerManagerActor` wasn't extracted, because 👇🏽
  ...
  }
  ```

3. [`near_network::peer_manager::PeerManagerActor`](https://github.com/near/nearcore/blob/a44be206018b547f53cb3b704f3a822e0d9d5db0/chain/network/src/peer_manager.rs#L158)

  ```rust
  pub struct PeerManagerActor {
  ...
      peer_store: PeerStore, // `PeerStore` wasn't extracted, because 👇🏽
  ...
  }
  ```

4. [`near_network::peer_store::PeerStore`](https://github.com/near/nearcore/blob/a44be206018b547f53cb3b704f3a822e0d9d5db0/chain/network/src/peer_store.rs#L48-L49)

  ```rust
  pub struct PeerStore {
      store: Arc<Store>, // `Store` wasn't extracted, because 👇🏽
  ...
  }
  ```

5. [`near_store::PeerStore`](https://github.com/near/nearcore/blob/a44be206018b547f53cb3b704f3a822e0d9d5db0/core/store/src/lib.rs#L48-L49)

  ```rust
  pub struct PeerStore {
      storage: Pin<Arc<dyn Database>>, // 👈🏽 rocksdb
  }
  ```

chefsale (approved):

chain/network/src/types.rs
frol: I see. Sounds reasonable to me. Thanks for the detailed explanation :+1:

frol:
This PR improved the compilation time and other metrics of `near-cli-rs` significantly! (see the before / after figures)

debug clean build time: 7m 44s / 2m 7s
debug incremental build time: 33m 89s / 17m 27s
debug binary size: 55.9 MB / 38.2 MB
dependencies: 442 / 390

@miraclx Great job!

chain/network/src/routing.rs
bowenwang1996: @pmnoxx please take a look

chain/network-primitives/src/metrics.rs
```diff
@@ -6,6 +6,8 @@ use near_metrics::{
 use std::collections::HashMap;
 use strum::VariantNames;
 
+use lazy_static::lazy_static;
```
bowenwang1996, line 9: What's the reason for moving this file to network-primitives? It feels to me that those counters should be defined in the network crate itself

bowenwang1996 (commented):
What are the criteria for deciding what goes into network-primitives?

chain/network-primitives/src/metrics.rs
miraclx: You're right. Nice catch!

bowenwang1996: Where does the dependency on `Edge` come from? I don't think it is used outside of the network crate

miraclx: So, while `near-client-primitives` only required;

```rust
near_network::{recorder::MetricRecorder, types::{AccountOrPeerIdOrHash, KnownProducer}, PeerInfo}
```

In the case of `near_network::types`, I believed it was safe to extract as much as was possible since that module name typically encapsulates reusable primitives. And that worked fine with the only exception being `Consolidate` which is retained in `near_network::types`.

Following that logic, `routing::Edge` has a number of dependents in `near_network_primitives::types`. So, naturally, it's extracted alongside it.

bowenwang1996: > I believed it was safe to extract as much as was possible since that module name typically encapsulates reusable primitives

Well that is exactly my concern. `Edge` should not be used outside of the network crate. Why do we need `MetricRecorder` in `near-client-primitives`?

frol: @bowenwang1996 What would be your suggestion here? I would love to merge this PR first since it already gives net positive result for us. The public interfaces needs more careful thought anyway, it is not like we promoted it to be public here

bowenwang1996: @mfornet is my understanding correct that we don't actually use metric recorder? Can we just get rid of it?

mfornet: > Can we just get rid of it?
Yes 

I was using metric recorder while optimising the network more than a year ago, but it is not used anymore.

bowenwang1996: I see. @miraclx please remove metric_recorder from `NetworkInfoResponse`

chain/network/src/lib.rs
```diff
@@ -1,12 +1,12 @@
...
-pub use types::{
+pub use near_network_primitives::types::{
     FullPeerInfo, NetworkAdapter, NetworkClientMessages, NetworkClientResponses, NetworkConfig,
     NetworkRecipient, NetworkRequests, NetworkResponses, PeerInfo,
 };
+pub use peer::{EPOCH_SYNC_PEER_TIMEOUT_MS, EPOCH_SYNC_REQUEST_TIMEOUT_MS};
+pub use peer_manager::PeerManagerActor;
+
+#[cfg(feature = "metric_recorder")]
+pub use near_network_primitives::recorder;
```
frol, line 9: This seems to also need to be cleaned up

chain/network-primitives/src/metrics.rs
frol: @mfornet @bowenwang1996 should the whole feature be removed altogether or just drop the field from the `NetworkInfoResponse`? Is the feature still useful without the exposed data?

chain/network/src/lib.rs
frol: Or maybe we want to keep the feature in the Cargo.toml file since there are other places where `metric_recorder` feature is used

chain/network-primitives/src/metrics.rs
bowenwang1996: > @mfornet @bowenwang1996 should the whole feature be removed altogether or just drop the field from the NetworkInfoResponse? Is the feature still useful without the exposed data?

I think the feature should be removed altogether, but my point is that we can now revert the change that extracts network primitives since I don't think they should be in a separate crate to begin with @miraclx.

miraclx: `MetricRecorder` was removed, which freed up the dependency on `PeerMessage` which, in-turn, freed up the dependency on `Edge`. This means `::routing` and a number of structures in `::types` we're returned to `near-network`.
But we (`near_client_primitives`) still need `near_network_primitives::types::{AccountOrPeerIdOrHash,  KnownProducer, PeerInfo}`

frol: @bowenwang1996 @mfornet Could you review the PR after the update?

bowenwang1996 (approved):

PR #4656, opened 2021-08-09
scripts: nayduck: switch to new authentication mode

mina86:
To reduce complexity of the authentication flow, NayDuck is switching
to a new process for requesting and managing authentication tokens.
This deprecates the old codes used when scheduling a NayDuck run.
Switch `nayduck.py` script to use the new tokens.

Issue: https://github.com/near/nayduck/issues/17

### DIFF
--- a/scripts/nayduck.py
+++ b/scripts/nayduck.py
@@ -20,11 +20,13 @@
 #    https://github.com/utka/nayduck
 
 import json
+import os
 import pathlib
 import subprocess
 
 
 DEFAULT_TEST_FILE = 'nightly/nightly.txt'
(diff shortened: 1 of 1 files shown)


chefsale (approved):
LgTm

chefsale:
Can we seitch prints to some proper logging in nayduck? 😁

mina86:
`nayduck.py` is a CLI tool. Honestly I don’t think using logging there makes anything better.

chefsale:
> `nayduck.py` is a CLI tool. Honestly I don’t think using logging there makes anything better.

I see my bad didn't realize that.

PR #4657, opened 2021-08-09
Address outstanding comments from PR #4592

### LINKED ISSUE #2459: Remove unnecessary PhantomData usage

nikurt:
Fixes #2459


nikurt:
Sorry about auto-merging https://github.com/near/nearcore/pull/4592 without resolving the comments. This PR resolves those comments, please take a look.
@matklad @frol

frol (approved):

bowenwang1996 (approved):

Issue #4569, opened 2021-07-26
Regression in `neard testnet` behavior


frol:
**Describe the bug**

There is a regression in the default behavior of `testnet` neard subcommand between 1.19.2 and 1.20.0, which caused nayduck tests to fail: https://github.com/near/nearcore/issues/4464#issuecomment-885903776. Initializing a new cluster of localnet nodes for testing now has the following values:

```diff
   "num_block_producer_seats_per_shard": [
-    4
+    1,
+    1,
+    1,
+    1
   ],
   "avg_hidden_validator_seats_per_shard": [
+    0,
+    0,
+    0,
     0
   ],
```

Full genesis.json diffs:

```diff
--- /tmp/near-1.19.2/node0/genesis.json	2021-07-26 15:35:35.583899833 +0200
+++ /tmp/near-1.20.0/node0/genesis.json	2021-07-26 15:35:45.384005185 +0200
@@ -1,13 +1,19 @@
 {
-  "protocol_version": 45,
-  "genesis_time": "2021-07-26T13:35:35.595752612Z",
-  "chain_id": "test-chain-44x3g",
+  "protocol_version": 46,
+  "genesis_time": "2021-07-26T13:35:45.392565119Z",
+  "chain_id": "test-chain-kBWkp",
   "genesis_height": 0,
   "num_block_producer_seats": 4,
   "num_block_producer_seats_per_shard": [
-    4
+    1,
+    1,
+    1,
+    1
   ],
   "avg_hidden_validator_seats_per_shard": [
+    0,
+    0,
+    0,
     0
   ],
   "dynamic_resharding": false,
@@ -154,6 +160,9 @@
         "keccak256_byte": 21471105,
         "keccak512_base": 5811388236,
         "keccak512_byte": 36649701,
+        "ripemd160_base": 853675086,
+        "ripemd160_block": 680107584,
+        "ecrecover_base": 3365369625000,
         "log_base": 3543313050,
         "log_byte": 13198791,
         "storage_write_base": 64196736000,
@@ -203,6 +212,7 @@
         "max_arguments_length": 4194304,
         "max_length_returned_data": 4194304,
         "max_contract_size": 4194304,
+        "max_transaction_size": 4194304,
         "max_length_storage_key": 4194304,
         "max_length_storage_value": 4194304,
         "max_promises_per_function_call_action": 1024,
@@ -258,7 +268,8 @@
           "amount": "950000000000000000000000000000000",
           "locked": "50000000000000000000000000000000",
[...]

PR #4659, opened 2021-08-09
Implement ShardUId

mzhangmzz:
This PR implements ShardUId as specified in https://github.com/near/NEPs/pull/241/ so that trie states are indexed by 'shard_uid` instead of `shard_id`. 

The main change is to change the interface around trie storage to use `shard_uid` instead of `shard_id`. The database columns `ColState` and `ColTrieChanges` are now indexed by `shard_uid` instead of `shard_id`.

Changes that this PR does not include:
- change `account_id_to_shard_id` to account for changing shard layout
- gc

Test Plan: 
The main source for errors is where `shard_id` and `shard_uid` is mixed up or the wrong `shard_version` is used for `shard_uid`, which may result in `DB not found error`. To test for the first type, we will change a wide range of existing unit and integration tests to use a shard layout with shard_version = 1 so `shard_id` and `shard_uid` cannot be mixed up.


chain/chain/src/chain.rs
```diff
@@ -1452,8 +1452,11 @@ impl Chain {
             root_proofs.push(root_proofs_cur);
         }
 
-        let state_root_node =
-            self.runtime_adapter.get_state_root_node(shard_id, &chunk_header.prev_state_root())?;
+        let state_root_node = self.runtime_adapter.get_state_root_node(
+            shard_id,
+            &sync_hash,
```
bowenwang1996, line 1457: I am afraid that `sync_hash` is not what you want here because `sync_hash` is the first block of the new block and I believe here the chunk could be included in a block in the previous epoch

chain/chain/src/types.rs
```diff
@@ -244,10 +244,10 @@ pub trait RuntimeAdapter: Send + Sync {
     fn get_tries(&self) -> ShardTries;
 
     /// Returns trie.
-    fn get_trie_for_shard(&self, shard_id: ShardId) -> Trie;
+    fn get_trie_for_shard(&self, shard_id: ShardId, block_hash: &CryptoHash) -> Trie;
```
bowenwang1996, line 247: Could we add a comment on what `block_hash` is for?

core/primitives/src/shard_layout.rs
```diff
@@ -90,11 +108,62 @@ fn is_top_level_account(top_account: &AccountId, account: &AccountId) -> bool {
...
+pub struct ShardUId {
+    pub version: ShardVersion,
+    pub shard_id: u32,
+}
+
+impl ShardUId {
+    pub fn default() -> Self {
+        Self { version: 0, shard_id: 0 }
+    }
+    pub fn to_vec(&self) -> [u8; 8] {
```
bowenwang1996, line 122: hmm it doesn't actually return a vector?

core/primitives/src/shard_layout.rs
```diff
@@ -90,11 +108,62 @@ fn is_top_level_account(top_account: &AccountId, account: &AccountId) -> bool {
...
+        let mut res = [0; 8];
+        res[0..4].copy_from_slice(&u32::to_le_bytes(self.version));
+        res[4..].copy_from_slice(&u32::to_le_bytes(self.shard_id));
+        res
+    }
+
+    pub fn from_vec(bytes: &[u8]) -> Result<Self, Box<dyn std::error::Error>> {
```
bowenwang1996, line 129: Might be better to implement `TryFrom<&[u8]>`

core/primitives/src/shard_layout.rs
```diff
@@ -90,11 +108,62 @@ fn is_top_level_account(top_account: &AccountId, account: &AccountId) -> bool {
...
+    let mut res = Vec::with_capacity(40);
+    res.extend_from_slice(block_hash.as_ref());
+    res.extend_from_slice(&shard_uid.to_vec());
+    res
+}
+
+#[allow(unused)]
+pub fn get_block_shard_uid_rev(
+    key: &[u8],
+) -> Result<(CryptoHash, ShardUId), Box<dyn std::error::Error>> {
```
bowenwang1996, line 149: We generally avoid generic error types such as `std::error::Error` because it creates trouble when we have to convert error types between each other. It is better to create a specific error type

core/store/src/trie/shard_tries.rs
```diff
@@ -1,104 +1,122 @@
...
+            .collect::<HashMap<_, _>>();
         for op in &transaction.ops {
             match op {
                 DBOp::UpdateRefcount { col, ref key, ref value } if *col == DBCol::ColState => {
-                    let (shard_id, hash) = TrieCachingStorage::get_shard_id_and_hash_from_key(key)?;
-                    shards[shard_id as usize].push((hash, Some(value.clone())));
+                    let (shard_uid, hash) =
+                        TrieCachingStorage::get_shard_uid_and_hash_from_key(key)?;
+                    shards.get_mut(&shard_uid).unwrap().push((hash, Some(value.clone())));
+                    // [shard_uid as usize].push((hash, Some(value.clone())));
```
bowenwang1996, line 92: What happened here?

chain/chain/src/chain.rs
mzhangmzz: Yeah, but we will disable state sync for epoch changes where resharding happens. I will add a comment here noting that.

core/store/src/trie/shard_tries.rs
mzhangmzz: This is just some artifacts from me copy-pasting some code. Removed

core/primitives/src/shard_layout.rs
mzhangmzz: renamed it to `to_bytes`

core/primitives/src/shard_layout.rs
mzhangmzz: The signature of `get_block_shard_id_rev` also uses `Box<dyn std::error::Error>` and this function will only be used to replace `get_block_shard_id_rev` in selected places, so I think it should be fine to just use `std::error:Error`

bowenwang1996:
looks like CI is still failing

chain/chain/src/chain.rs
```diff
@@ -1527,19 +1530,21 @@ impl Chain {
         if shard_id as usize >= sync_prev_block.chunks().len() {
             return Err(ErrorKind::InvalidStateRequest("shard_id out of bounds".into()).into());
         }
+        // TODO: why is this prev_state_root here?
```
bowenwang1996, line 1533: what is the action item here?

chain/chain/src/chain.rs
```diff
@@ -1527,19 +1530,21 @@ impl Chain {
...
             .runtime_adapter
-            .get_state_root_node(shard_id, &state_root)
+            .get_state_root_node(shard_id, &sync_hash, &state_root)
             .log_storage_error("get_state_root_node fail")?;
         let num_parts = get_num_state_parts(state_root_node.memory_usage);
 
         if part_id >= num_parts {
             return Err(ErrorKind::InvalidStateRequest("part_id out of bound".to_string()).into());
         }
+        // TODO: this part of logic may change when shards may change?
```
bowenwang1996, line 1544: is this still an issue?

chain/chain/src/store.rs
```diff
@@ -2007,15 +2008,17 @@ impl<'a> ChainStoreUpdate<'a> {
             GCMode::Fork(tries) => {
                 // If the block is on a fork, we delete the state that's the result of applying this block
                 for shard_id in 0..header.chunk_mask().len() as ShardId {
+                    // TODO: pass in the actual shard version that this block uses
```
bowenwang1996, line 2011: is this already done?

chain/chain/src/store.rs
```diff
@@ -2048,8 +2053,10 @@ impl<'a> ChainStoreUpdate<'a> {
             }
             GCMode::StateSync { .. } => {
                 // Not apply the data from ColTrieChanges
+                // TODO: pass in the actual shard version that this block uses
```
bowenwang1996, line 2056: same here

core/primitives/src/shard_layout.rs
```diff
@@ -16,6 +23,8 @@ pub enum ShardLayout {
 pub struct ShardLayoutV0 {
```
bowenwang1996, line 23: do these have to be public? Is there some place where we use `ShardLayoutV0` directly instead of `ShardLayout`?

bowenwang1996:
Given how invasive this change is, could you run nayduck to see that no test regresses?

chain/chain/src/store.rs
mzhangmzz: No, this is not. I will change this later.

chain/chain/src/store.rs
mzhangmzz: Here, it just uses shard_version = 0

chain/chain/src/chain.rs
mzhangmzz: sorry, that was me commenting when I read code. Will delete this

core/primitives/src/shard_layout.rs
mzhangmzz: yeah, I think they do, because they can be accessed from ShardLayout directly

mzhangmzz: error[E0446]: private type `ShardLayoutV1` in public interface

mzhangmzz:
> looks like CI is still failing

@bowenwang1996 
I found why the CI failed but not sure what's the best way to fix it.

The reason why it failed is because the genesis file generated by the stable branch for the backward compatible test does not work for the new binary. The reason is because of the newly added `shard_layout` struct in the genesis file. By default, that field is initiated as `ShardLayoutV0{num_shards:1, version: 0}`. That is fine if the config file does specify `num_shards` to be 1 through `num_block_producer_seats_per_shard.len()`, but the backward compatible test is actually testing for 4 shards. 

Is there a special reason why the backward compatible test test 4 shards?

frol:
@mzhangmzz @bowenwang1996 

>  That is fine if the config file does specify num_shards to be 1 through num_block_producer_seats_per_shard.len(), but the backward compatible test is actually testing for 4 shards.

#4569

mzhangmzz:
> @mzhangmzz @bowenwang1996
> 
> > That is fine if the config file does specify num_shards to be 1 through num_block_producer_seats_per_shard.len(), but the backward compatible test is actually testing for 4 shards.
> 
> #4569

Thanks @frol! Do you know if anyone is actively working on fixing it now?

chain/chain/src/store.rs
bowenwang1996: Could you create an issue for that and add the issue number in the comment?

bowenwang1996:
> @mzhangmzz @bowenwang1996
> 
> > That is fine if the config file does specify num_shards to be 1 through num_block_producer_seats_per_shard.len(), but the backward compatible test is actually testing for 4 shards.
> 
> #4569

@frol I saw that you assigned it to Miraculous. Should someone from our side take this over since it is blocking? @mzhangmzz do you mind looking into the issue?

matklad (approved):
Estimator changes looks good to me, leaving approval to unblock the PR. I haven't looked at the change as a whole.

frol:
@bowenwang1996 Yeah, @miraclx is currently busy with other items on his list, so this was not our priority. Feel free to take it

mzhangmzz:
> @bowenwang1996 Yeah, @miraclx is currently busy with other items on his list, so this was not our priority. Feel free to take it

Thanks! I will take it over then. Thank you for linking the issue!

chain/chain/src/store.rs
mzhangmzz: https://github.com/near/nearcore/issues/4710

chain/chain/src/chain.rs
mzhangmzz: I created an issue https://github.com/near/nearcore/issues/4711 noting this

mzhangmzz:
@frol @bowenwang1996 hmm I realized that https://github.com/near/nearcore/pull/4709 will not fix the backward_compatible test until that commit is included the in the latest release branch, i.e., the branch the test is using to make comparison. 

I'm not sure when is the next time we will change the release branch. Is there any way to bypass that?

bowenwang1996:
> @frol @bowenwang1996 hmm I realized that #4709 will not fix the `backward_compatible` test until that commit is included the in the latest release branch, i.e., the branch the test is using to make comparison.
> 
> I'm not sure when is the next time we will make the next release. Is there any way to bypass that?

Do you need just the branch or do you need a specific commit? If you just need it to be included in the branch, feel free to cherry-pick the commit on top of 1.21.0 branch.

bowenwang1996 (approved):

ailisp (approved):
I mostly checked runtime part, and looks good!

PR #4661, opened 2021-08-10
nayduck: add support for suffix in ./<path> include directive

mina86:
Extend the `./<path>` directive so that it accepts arbitrary suffix
after the path.  If that suffix is present, all lines read from the
file are augmented by adding the prefix.  This makes it possible to do
things like:

    ./pytest.txt
    ./pytest.txt --features nightly_protocol,nightly_protocol_features

and avoid having to list every test twice in `pytest.txt` file.

Furthermore, when check_pytest.py and check_nightly.py read the files
also handle `#./<path>` so that a whole include can be commented out
with a TODO and the check scripts will treat files listed in those
files as being mentioned.

### DIFF
--- a/nightly/TODO-disabled-4552.txt
+++ b/nightly/TODO-disabled-4552.txt
@@ -1,14 +1,4 @@
-# TODO(#4552): Those tests were identified as missing from any of the test list
-# files.  It’s not entirely clear why so some investigation is necessary to
-# figure that out.  The tests need to either be enabled or removed completely if
-# they are no longer needed/valid.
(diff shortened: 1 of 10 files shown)


mina86:
Actually the more I think about it the less convinced I am this is a good feature. It’s kinda neat and it shrinks the txt files but then again with the txt files shrunk it’s now harder to just run `pytest-spec.txt` tests for normal and nightly builds for example. It’s also additional code that’ll need to be maintained. Gonna close it for now (though I might still keep `#./`).

PR #3869, opened 2021-01-29
refactor: Upgraded all the dependencies

frol:
The main goal is to get the latest actix/actix-web along with tokio 1.x. Fixes #3868

Extra bonuses:

1. Resolved compilation errors due to use of private modules of `syn` and `serde` crates in our code-base
2. Hopefully it will address a long-standing Telemetry issue "not an error"
3. It may also address a memory leak in Telemetry

PR #4663, opened 2021-08-10
chore: Update `primitive-types` to `0.10`, and remove old EVM dependencies

joshuajbouw:
On Aurora, some of our dependencies updated to the latest `ethereum-types` which is preventing us from progressing. This simply just updates this dependency. No issues or extra code is needed from what I could see.

This also removes unused `ethabi` or other `eth` related crates.


frol:
@joshuajbouw Thanks for bringing it up!

> One question, what is the recommended way to resolve the `check bans`?

Review the duplicates and consider upgrading the other dependencies to arrive to a single version, otherwise, whitelist it in `deny.toml`

joshuajbouw:
Hmm, looks to be like a nightmare to get everyone on the same page. Think this will be a fit for `deny.toml` @frol ?

frol:
Yeah, that is fine. Also, you may consider patching 0.11 with 0.12 version if they are compatible for our usage. I used to do that: https://github.com/near/nearcore/pull/3869/files#diff-2e9d962a08321605940b5a657135052fbcef87b5e360662bb527c96d9a615542R73-R74

bowenwang1996 (approved):

joshuajbouw:
Actually, I see a bit of cleanup that I can target with this and get what we wanted in.

joshuajbouw:
@bowenwang1996 is there something wrong with the test suite currently? Pretty confused as to why it is failing where it is when that isn't an area that I had modified.

joshuajbouw:
Ok, looks like it was re-run and was successful. This is ready then.

frol:
@joshuajbouw CI still fails.

joshuajbouw:
@frol thats funny. Was added into `master` branch as I was updating it. Anyways, lets see how this run goes.

frol:
@joshuajbouw the PR name seems to be outdated :smile:

joshuajbouw:
@frol Haha, sure. It's good to go.

frol (approved):

PR #4621, opened 2021-08-02
refactor: Reintroduce strictly typed AccountId

miraclx:
Resolves #4600

- <details>
	<summary> Reintroduces #4440 (reverted in #4601) </summary>

    > This PR introduces a strictly typed structure - `AccountId` that enforces account ID validation.
    > This way, we can ensure validity and thereby fail fast before an invalid account name is used.
    > 
    > This also, optionally, ensures validation on {,de}serialization so, yeah..
    > 
    > The structure is defined within its own self-contained crate `near-account-id` to allow for easy publishing.
    > The extra dependencies - `borsh`, `serde` are made optional to allow the crate to be lightweight enough to be depended upon.
    > 
    > Within `nearcore`, the structure is reexported via `near_primitives::AccountId` and the crate via `near_primitives{,_core}::account::id` with default features (`borsh`, `serde`) but you can opt-out of those by depending on it directly.
    > 
    > Resolves #2074, supersedes #2831.
    > 
    > # Testing Plan
    > 
    > * [x] Existing unit tests pass with minor updates
    > * [x] Nayduck tests pass without modifications
    > * [x] Manually ensured that the protocol is not changed

    </details>
-  Fixes issue of loading genesis with records of invalid AcountId-s in AccessKey-s

### Nayduck

http://nayduck.eastus.cloudapp.azure.com:3000/#/run/1943

### Testing Plan

- [x] Manually booted a testnet node to ensure genesis loads properly and that no further AccountId-related issues occur.
- [x] Ensured unit tests pass with no further updates
- [x] Nayduck tests pass without modifications

PR #4666, opened 2021-08-10
Update `total_supply` when dumping state for the new genesis

### LINKED ISSUE #4665: `state-viewer` fails to dump new genesis state due to mismatching supply
**Describe the bug**
`cargo build --package state-viewer --features nightly_protocol,nightly_protocol_features && ./target/debug/state-viewer --home=~/.near/betanet dump_state`
```
Aug 10 16:24:45.084  INFO near: Opening store database at "/home/user/.near/betanet/data"
Aug 10 16:24:48.135  INFO runtime: Tracking shards: {0}
Generating genesis from state data of #34125348 / Dve5dnycfdsn7F9gWfVUQEwwCD3xqw4SpJ7f9bmqo9Cn
thread 'main' panicked at 'assertion failed: `(left == right)`
  left: `1371902924464758940086228587317865`,
 right: `1331261183170775131069856846142903`: wrong total supply', core/chain-configs/src/genesis_validate.rs:85:9
```

This is caused by a recent change forcing validation of total_supply upon each creation of `Genesis`.
What happens is that every epoch the total supply is burned and/or minted. Which leads to a discrepancy between the supply in `records` and in `total_supply` in `genesis.json`.

nikurt:
Fixes #4665


chain/client/tests/chunks_management.rs
```diff
@@ -100,8 +100,7 @@ fn update_chunk_height_created(
 fn store_partial_encoded_chunk_sanity() {
     init_test_logger();
     let mut env = TestEnv::new(ChainGenesis::test(), 1, 1);
-    let signer =
-        InMemoryValidatorSigner::from_seed("test0".parse().unwrap(), KeyType::ED25519, "test0");
+    let signer = InMemoryValidatorSigner::from_seed("test0", KeyType::ED25519, "test0");
```
frol, line 103: Revert this change, it is expected to have `.parse().unwrap()` since we introduced strictly typed AccountId #4621

frol (approved):

bowenwang1996 (approved):

chain/client/tests/chunks_management.rs
nikurt: Done.
It took github a few hours to show it as reverted due to a github incident yesterday.

PR #4668, opened 2021-08-10
Change ShardTracker interface

mzhangmzz:
Now a client can either track a set of accounts or track all shards. They no longer has the option of tracking an individual shard. The reason is that an outside client should not care about a particular shard at all. This also makes the interface more robust to sharding changes.

### DIFF
--- a/chain/client/src/view_client.rs
+++ b/chain/client/src/view_client.rs
@@ -1074,7 +1074,11 @@ impl Handler<NetworkViewClientMessages> for ViewClientActor {
                             hash: *self.chain.genesis().hash(),
                         },
                         height,
-                        tracked_shards: self.config.tracked_shards.clone(),
+                        tracked_shards: if self.config.track_all_shards {
+                            (0..self.runtime_adapter.num_shards() - 1).collect()
+                        } else {
+                            vec![]
+                        },
                         archival: self.config.archive,
                     }
                 }
--- a/nearcore/src/runtime/mod.rs
+++ b/nearcore/src/runtime/mod.rs
@@ -140,12 +140,16 @@ pub struct NightshadeRuntime {
 }
 
 impl NightshadeRuntime {
+    pub fn default(home_dir: &Path, store: Arc<Store>, genesis: &Genesis) -> Self {
+        Self::new(home_dir, store, genesis, vec![], false, None, None)
+    }
+
     pub fn new(
         home_dir: &Path,
         store: Arc<Store>,
         genesis: &Genesis,
-        initial_tracking_accounts: Vec<AccountId>,
-        initial_tracking_shards: Vec<ShardId>,
+        tracking_accounts: Vec<AccountId>,
+        track_all_shards: bool,
         trie_viewer_state_size_limit: Option<u64>,
         max_gas_burnt_view: Option<Gas>,
     ) -> Self {
@@ -174,8 +178,8 @@ impl NightshadeRuntime {
             .expect("Failed to start Epoch Manager"),
         ));
         let shard_tracker = ShardTracker::new(
-            initial_tracking_accounts,
-            initial_tracking_shards,
+            tracking_accounts,
+            track_all_shards,
             epoch_manager.clone(),
             num_shards,
         );
@@ -1741,8 +1745,17 @@ mod test {
             prefix: &str,
             validators: Vec<Vec<AccountId>>,
             epoch_length: BlockHeightDelta,
-            initial_tracked_accounts: Vec<AccountId>,
-            initial_tracked_shards: Vec<ShardId>,
+            has_reward: bool,
+        ) -> Self {
+            Self::new_with_tracking(prefix, validators, epoch_length, vec![], false, has_reward)
+        }
+
+        pub fn new_with_tracking(
+            prefix: &str,
+            validators: Vec<Vec<AccountId>>,
+            epoch_length: BlockHeightDelta,
+            tracked_accounts: Vec<AccountId>,
+            track_all_shards: bool,
             has_reward: bool,
         ) -> Self {
             let dir = tempfile::Builder::new().prefix(prefix).tempdir().unwrap();
@@ -1770,8 +1783,8 @@ mod test {
                 dir.path(),
                 store,
                 &genesis,
-                initial_tracked_accounts,
-                initial_tracked_shards,
+                tracked_accounts,
+                track_all_shards,
                 None,
                 None,
             );
@@ -1933,14 +1946,7 @@ mod test {
         let validators = (0..num_nodes)
             .map(|i| AccountId::try_from(format!("test{}", i + 1)).unwrap())
             .collect::<Vec<_>>();
-        let mut env = TestEnv::new(
-            "test_validator_rotation",
-            vec![validators.clone()],
-            2,
-            vec![],
-            vec![],
-            false,
-        );
+        let mut env = TestEnv::new("test_validator_rotation", vec![validators.clone()], 2, false);
         let block_producers: Vec<_> = validators
             .iter()
             .map(|id| InMemoryValidatorSigner::from_seed(id.clone(), KeyType::ED25519, id.as_ref()))
@@ -2025,14 +2031,8 @@ mod test {
         let validators = (0..num_nodes)
             .map(|i| AccountId::try_from(format!("test{}", i + 1)).unwrap())
             .collect::<Vec<_>>();
-        let mut env = TestEnv::new(
-            "test_validator_stake_change",
-            vec![validators.clone()],
-            2,
-            vec![],
-            vec![],
-            false,
-        );
+        let mut env =
+            TestEnv::new("test_validator_stake_change", vec![validators.clone()], 2, false);
         let block_producers: Vec<_> = validators
             .iter()
             .map(|id| InMemoryValidatorSigner::from_seed(id.clone(), KeyType::ED25519, id.as_ref()))
@@ -2077,8 +2077,6 @@ mod test {
             "test_validator_stake_change_multiple_times",
             vec![validators.clone()],
             4,
-            vec![],
-            vec![],
             false,
         );
         let block_producers: Vec<_> = validators
@@ -2188,8 +2186,6 @@ mod test {
             "test_validator_stake_change_multiple_times",
             vec![validators.clone()],
             5,
-            vec![],
-            vec![],
             false,
         );
         let block_producers: Vec<_> = validators
@@ -2230,14 +2226,8 @@ mod test {
         let validators = (0..2)
             .map(|i| AccountId::try_from(format!("test{}", i + 1)).unwrap())
             .collect::<Vec<_>>();
-        let env = TestEnv::new(
-            "verify_validator_signature_failure",
-            vec![validators.clone()],
-            2,
-            vec![],
-            vec![],
-            true,
-        );
+        let env =
+            TestEnv::new("verify_validator_signature_failure", vec![validators.clone()], 2, true);
         let data = [0; 32];
         let signer = InMemorySigner::from_seed(
             validators[0].clone(),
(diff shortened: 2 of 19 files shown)


mzhangmzz:
@bowenwang1996 @frol this PR implements what we discussed in https://github.com/near/NEPs/pull/241 about how to handle  tracking shards when shard may change across epochs.

bowenwang1996 (commented):
I think this change is not backward compatible, i.e, after this change a node cannot read an old config since the field name has changed, which is problematic. This will likely lead to no node being able to restart after adopting this change. We should make sure that this won't happen.

mzhangmzz:
> I think this change is not backward compatible, i.e, after this change a node cannot read an old config since the field name has changed, which is problematic. This will likely lead to no node being able to restart after adopting this change. We should make sure that this won't happen.

Are you suggesting we should not do this change at all, because to fully adopt it we need to change the config format, or do you think we should adopt the change but implementing it in a backward compatible way?

mzhangmzz:
We can do it in a backward compatible way for now by still allowing them to specifying `tracked_shards` in the config. And if that is not empty, set `track_all_shards` to true. That will work because right now we only have one shard. 

But eventually when we migrate to 8 shards, we should deprecate the old `tracked_shards` parameter

bowenwang1996:
> I think it's good to fail hard instead of hiding it and running different behavior than what the users intend to do. It's an easy fix for them, just update one line in the config file

But *everyone* who runs a node has to update their config and if this is not done properly, it can potentially lead to a network stall for a short period of time. Please sync with @janewang on what is the best thing to do to make it easy for node operators to update their nodes.

frol:
Consider implementing an automatic config migration or auto-fallback with a warning/info message about the new configuration option. I would not expect the manual upgrade to go smooth for the network of over 300 decentralized nodes (250+ mainnet and 80+ testnet).

chain/indexer/src/lib.rs
```diff
@@ -95,9 +95,9 @@ impl Indexer {
         let near_config = nearcore::load_config(&indexer_config.home_dir);
         near_chain_configs::genesis_validate::validate_genesis(&near_config.genesis);
         assert!(
-            !&near_config.client_config.tracked_shards.is_empty(),
-            "Indexer should track at least one shard. \n\
-            Tip: You may want to update {} with `\"tracked_shards\": [0]`
+            near_config.client_config.track_all_shards,
+            "Indexer should track all shards. \n\
+            Tip: You may want to update {} with `\"track_all_shards\": true`
```
frol, line 100: @khorolets We should keep an eye on these changes and potentially update some of our docs

janewang:
Synced with @mzhangmzz. We will communicate the required config change in the sharding announcement to all validators. We tentatively circled the Oct 4 testnet release date for the sharding announcements and Oct 18 for mainnet release.

@mzhangmzz plan to implement an automatic config migration, and plan to update the default `config.json` during for a new node's working directory initialization with neard.

cc @frol @bowenwang1996 @chefsale @mhalambek

mzhangmzz:
I synced with @janewang today and we agreed on the following points:

- It is better that this change is rolled out on the same release together with sharding. It makes sense to communicate to validators and other nodes that they will need to make some changes to support the sharding release. 
- We should provide a binary that updates existing configs. This way, existing nodes can update their config files locally. 
- We should also change `neard init` to update the config it generates.

bowenwang1996:
@mzhangmzz is this PR ready to be reviewed?

nearcore/src/shard_tracker.rs
```diff
@@ -1,95 +1,69 @@
...
-    pub fn track_accounts(&mut self, account_ids: &[AccountId]) {
-        for account_id in account_ids.iter() {
-            self.track_account(account_id);
+    fn tracks_shard_at_epoch(
+        &self,
+        shard_id: ShardId,
+        epoch_id: &EpochId,
+    ) -> Result<bool, EpochError> {
+        let mut tracking_shards = self.tracking_shards.write().expect(POISONED_LOCK_ERR);
+        if !tracking_shards.contains_key(epoch_id) {
```
matklad, line 48: This looks like a standard once_cell-like caching idiom, so we might want to not write the `RwLock` at the call-site, but rather encapsulate it behind some API.

```
struct AppendOnlyMap {}

impl AppendOnlyMap {
    fn get_or_insert(&/*no mut*/ self, key: K, value: FnOnce() -> V) -> &V
}
```

We have something like that in the SDK: https://github.com/near/near-sdk-rs/blob/fc56a7913c1538e42a438d7373add6d620e5a462/near-sdk/src/utils/stable_map.rs

Also, you might find reading the docs/implementaton of OnceCell useful: https://docs.rs/once_cell/1.8.0/once_cell/. It should give a good intuition about making caches in Rust work.

nearcore/src/runtime/mod.rs
```diff
@@ -145,12 +146,35 @@ pub struct NightshadeRuntime {
 }
 
 impl NightshadeRuntime {
+    pub fn default(home_dir: &Path, store: Arc<Store>, genesis: &Genesis) -> Self {
+        Self::new(home_dir, store, genesis, vec![], false, None, None, RuntimeConfigStore::test())
+    }
```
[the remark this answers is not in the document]
mzhangmzz: Thanks for the suggestion! Yeah I felt a little weird about the `default` name as well but all the fields are initialized with the default value, that's why I decided to go with it. `test` is a much better name!

nearcore/src/runtime/mod.rs
```diff
@@ -145,12 +146,35 @@ pub struct NightshadeRuntime {
...
+
+    pub fn new_with_config(
+        home_dir: &Path,
+        store: Arc<Store>,
+        config: &NearConfig,
+        trie_viewer_state_size_limit: Option<u64>,
```
mzhangmzz, line 157: @Longarithm I had some questions here about `trie_viewer_state_size_limit` and `max_gas_burnt_view`, both of them are part of `config`, so I initially wanted this function to take these two values from `config` as well. But some call sites, almost all tests, specified these values separately. I am not too familiar with the code there and I thought you might have an idea why

nearcore/src/shard_tracker.rs
mzhangmzz: I moved the RWLock inside the API, but not sure if that is what you had in mind @matklad let me know!

In general, I think there are many cases in our code that violates interior mutability and causes quite some inconvenience. For example, `EpochManager` https://github.com/near/nearcore/blob/master/chain/epoch_manager/src/lib.rs#L62 stores caches for some epoch info. Mutability leaked from inside `SizedCache` to outside `EpochManager`, so the callers of `EpochManager` have to maintain a RwLock (https://github.com/near/nearcore/blob/master/nearcore/src/runtime/mod.rs#L1771) on `EpochManager` for any access to it. Same things happen for `ChainStore` or almost any other cache like struct because they all use `SizedCache`.

nearcore/src/shard_tracker.rs
```diff
@@ -1,95 +1,69 @@
...
+use near_primitives::errors::EpochError;
 use near_primitives::hash::CryptoHash;
-use near_primitives::shard_layout::{account_id_to_shard_id, ShardLayout};
+use near_primitives::shard_layout::account_id_to_shard_id;
 use near_primitives::types::{AccountId, EpochId, ShardId};
 
 const POISONED_LOCK_ERR: &str = "The lock was poisoned.";
 
+// bit mask for which shard to track
+type ShardBitMask = Vec<bool>;
```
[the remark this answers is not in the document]
mzhangmzz: so should I keep it the same since "bitvec" adds more dependencies and this is just a simple use case?

chain/client/src/view_client.rs
```diff
@@ -1095,16 +1095,30 @@ impl Handler<NetworkViewClientMessages> for ViewClientActor {
...
-                            hash: *self.chain.genesis().hash(),
-                        },
-                        height,
-                        tracked_shards: self.config.tracked_shards.clone(),
-                        archival: self.config.archive,
+                    match self.runtime_adapter.num_shards(&head.epoch_id) {
+                        Ok(num_shards) => {
+                            // convert config tracked shards
+                            // runtime will track all shards if config tracked shards is not empty
+                            // https://github.com/near/nearcore/issues/4930
```
bowenwang1996, line 1102: Very helpful context 👍

nearcore/src/shard_tracker.rs
```diff
@@ -99,6 +94,7 @@ impl ShardTracker {
         shard_id: ShardId,
         is_me: bool,
     ) -> bool {
+        // TODO: fix these unwrap_or here and handle error correctly. The current behavior masks potential errors and bugs
```
bowenwang1996, line 97: Do we have an issue for this?

nearcore/src/shard_tracker.rs
```diff
@@ -1,95 +1,90 @@
...
 use near_primitives::hash::CryptoHash;
-use near_primitives::shard_layout::{account_id_to_shard_id, ShardLayout};
+use near_primitives::shard_layout::account_id_to_shard_id;
 use near_primitives::types::{AccountId, EpochId, ShardId};
+use std::hash::Hash;
 
 const POISONED_LOCK_ERR: &str = "The lock was poisoned.";
 
+struct AppendOnlyMap<K, V> {
+    map: RwLock<HashMap<K, Arc<V>>>,
```
bowenwang1996, line 14: why do we need a lock here?

nearcore/src/shard_tracker.rs
bowenwang1996: I agree. We have some quite substandard code there :(

nearcore/src/shard_tracker.rs
bowenwang1996: nvm found it

bowenwang1996 (approved):

chain/client/src/view_client.rs
```diff
@@ -1095,16 +1095,30 @@ impl Handler<NetworkViewClientMessages> for ViewClientActor {
...
+                            } else {
+                                (0..num_shards).collect()
+                            };
+                            NetworkViewClientResponses::ChainInfo {
+                                genesis_id: GenesisId {
+                                    chain_id: self.config.chain_id.clone(),
+                                    hash: *self.chain.genesis().hash(),
+                                },
+                                height: self.get_height(&head),
+                                tracked_shards,
```
frol, line 1114: Is there a plan to deprecate `tracked_shards`?

frol (approved):
LGTM, though I only reviewed it on a high-level since I am a bit out of the loop with the specifics of the implementation

nearcore/src/shard_tracker.rs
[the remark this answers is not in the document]
matklad: I think just leave the code as is is fine! Might make sense to expand the comment to say that we are not bothering to make this an actual *bit* set.

:thinking: I wonder if just `type Mask = Vec<bool>` would be best here? "Mask" is the main meaning here, not, "bit". And "Shared" isn't really a part of the type. 

But that's very minor, really, fine either way, I trust your judgement!

nearcore/src/shard_tracker.rs
```diff
@@ -1,95 +1,90 @@
...
+    }
+
+    fn get_or_insert<F: FnOnce() -> V>(&self, key: &K, value: F) -> Arc<V> {
+        let mut map = self.map.write().expect(POISONED_LOCK_ERR);
+        if !map.contains_key(key) {
+            map.insert(key.clone(), Arc::new(value()));
+        }
+        map.get(key).unwrap().clone()
+    }
+}
```
matklad, line 32: I'd move this into a `mod append_only_map;` -- this is a nice data structure which isn't related to anything else in nearcore, so there's beauty in having just a small tiny isolated file for it.

chain/client/src/view_client.rs
mzhangmzz: Yes, I just probably don't have time to do it before sharding launch since I still have a lot other to finish. 

The plan is to make neard automatically update the client config so that the validators won't need to worry about it at all

nearcore/src/shard_tracker.rs
mzhangmzz: https://github.com/near/nearcore/issues/4936

chain/client/src/view_client.rs
mzhangmzz: https://github.com/near/nearcore/issues/4930

bowenwang1996:
@ailisp please review this PR

bowenwang1996:
@ailisp is on vacation, so I'll merge this one directly

PR #4673, opened 2021-08-12
feat: prepare for 1.21.0 release

bowenwang1996:
Stabilize protocol feature `RestoreReceiptsAfterFix` for release 1.21.0


matklad (approved):

core/primitives/src/version.rs
```diff
@@ -97,21 +97,21 @@ pub enum ProtocolFeature {
     CountRefundReceiptsInGasLimit,
     /// Add `ripemd60` and `ecrecover` host function
     MathExtension,
+    /// Restore receipts that were previously stuck because of https://github.com/near/nearcore/pull/4228
+    RestoreReceiptsAfterFix,
```
frol, line 101: I just want to mention that this naming is too context heavy, so in the future we might have a confusion if we need to restore some receipts again. Can we think of a better name here?

frol (approved):
Looks good to me.

I am glad we addressed the issue of the dropped receipts that brings our state in a consistent state! Kudos to @Longarithm for implementing it in https://github.com/near/nearcore/pull/4248 and https://github.com/near/nearcore/issues/4444

EgorKulikov (approved):

core/primitives/src/version.rs
Longarithm: I was thinking about `RestoreReceiptsAfterFixApplyChunks` or `RestoreReceiptsLostInApplyChunks`, but this feels too long.
What do you think about `RestoreReceiptsAfterFix4228` using the number of PR https://github.com/near/nearcore/pull/4228?

frol: I feel that the root cause is not the receipts lost during the apply chunks, but rather receipts lost due to execution outcomes not being saved when they should have been. I forgot the specifics of the https://github.com/near/nearcore/pull/4226, but I feel the name should be something like `RestoreLostReceiptsAfterMissingExecutionOutcomes` (there was a more specific reason why the execution outcomes were lost, but I cannot recall it)

P.S. I would also prefer a shorter name, yet I prefer a descriptive name over a short one.

Issue #4362, opened 2021-06-14
Remove tests that often fail from nayduck


bowenwang1996:
There are some nightly tests that fail most of the time and we should exclude them from the nayduck result to make it easier to see whether a regression has happened.

PR #4674, opened 2021-08-12
nightly: comment out tests which are currently consistently failing

mina86:
With NayDuck nightly runs always including some failing tests it’s hard to
uses it for checking whether a change introduced a regression.  Comment out
all the tests which consistently fail so that the nightly runs become
green.

Fixes: https://github.com/near/nearcore/issues/4362
Issue: https://github.com/near/nearcore/issues/4618


bowenwang1996 (approved):
epic!

chefsale (approved):
Just restarted the nearlib tests as they failed.

Issue #4489, opened 2021-07-09
Implement a lame duck mode for RPC services


mina86:
With #3266 done, the node handles SIGINT and SIGTERM somewhat gracefully, i.e. all the cleanup code is executed so the result is no longer effectively a crash. However, all processing is interrupted. Most notably, if a client sent an RPC request, if the node gets SIGINT it will stop the processing and close connection to the client.

This should be improved by adding a lame duck mode to RPC endpoints. That is, once SIGINT or SIGTERM is received, the node should stop accepting new requests on RPC endpoints, finish whatever it’s doing and only then exit.

Note that there’s similar situation with nodes communicating with each other when the node may drop a connection to another peer while working on some request. This is less of an issue since peers should expect and handle correctly unexpected network interruptions and peer crashes. However, RPC clients are usually less fault tolerant so it would be nice if the node handled all incoming requests properly.

PR #4676, opened 2021-08-12
nearcore: return struct from start_with_config rather than a tuple

mina86:
Rather than relying on positions in a tuple, change start_with_config
function to return its data in a structure so that the elements can be
named.  This will also make it easier to add new return values and
thanks to the use of ‘..’ callers won’t be affected.

Issue: https://github.com/near/nearcore/issues/4489


nearcore/src/lib.rs
```diff
@@ -261,10 +261,13 @@ pub fn init_and_migrate_store(home_dir: &Path, near_config: &NearConfig) -> Arc<
     store
 }
 
-pub fn start_with_config(
-    home_dir: &Path,
-    config: NearConfig,
-) -> (Addr<ClientActor>, Addr<ViewClientActor>, Vec<ArbiterHandle>) {
+pub struct StartResult {
```
frol, line 264: I would consider refactoring it to `NearNode` type with some API scope (we may start with a single method `start_with_config`, but it might be an opportunity to consider splitting the function into smaller components). I believe [`NodeCluster`](https://github.com/near/nearcore/blob/942e02daf7d75a94db51ef1c4ea9ef27aed459a4/integration-tests/tests/nearcore/node_cluster.rs#L74-L93) could be a great source for inspiration.

@miraclx What do you think?

At the very least, I would rename this struct to `NearNode` since `StartResult` looks odd to me from the public API point of view.

frol (commented):
It is already a net positive improvement, but I have some suggestions to discuss

frol (approved):

Issue #4678, opened 2021-08-13
Regular op cost estimation from different binaries differs by 1.42


Longarithm:
Running function for estimating `regular_op_cost` from different binaries shows difference by 1.3.
This shouldn't be the case, because estimation should not depend on exact binary where it was launched.

Reproduction:
Look at commit in test branch https://github.com/near/nearcore/compare/master...93339de

1. Launch in a test

- launch docker, cd `/host/nearcore` 
- run `CARGO_TARGET_X86_64_UNKNOWN_LINUX_GNU_RUNNER=./runner.sh cargo test --release --features no_cpu_compatibility_checks,required --lib cases::test_cost_per_op -- --exact --nocapture`
Result:
```
     Running target/release/deps/runtime_params_estimator-5636b750d3ecdf04

running 1 test
test cases::test_cost_per_op ... test cases::test_cost_per_op has been running for over 60 seconds
1734559
test cases::test_cost_per_op ... ok
```

It is the same, if we run testing binary itself by command `./emu-cost/counter_plugin/qemu-x86_64 -cpu Westmere-v1 -plugin file=./emu-cost/counter_plugin/libcounter.so ../../target/release/deps/runtime_params_estimator-5636b750d3ecdf04 --test cases::test_cost_per_op --exact --nocapture` 

2. Launch in a params-estimator

- launch docker, build package `runtime-params-estimator`
- `touch /host/tmp/data/profile.txt` for debugging
- run `./emu-cost/counter_plugin/qemu-x86_64 -cpu Westmere-v1 -plugin file=./emu-cost/counter_plugin/libcounter.so ../../target/release/runtime-params-estimator --home /host/tmp/data --accounts-num 20000 --iters 1 --warmup-iters 1`
- `$ cat ~/tmp/data/profile.txt` gives:
```
REGULAR_OP_COST: 2289348 state: main
REGULAR_OP_COST: 2289374 state: pre-run
REGULAR_OP_COST: 2289438 state: before
```

`2289348` and `1734559` differ by 1.3.
1.42 was observed here: https://github.com/near/nearcore/pull/4647#issuecomment-896984076

Historically regular_op_cost was close to `1734559` in params-estimator, so it may help to find the first commit where the difference occurred.

**Note**: we should also keep it in mind for function call fee estimations.

cc @matklad @olonho

Longarithm:
Related: https://github.com/near/nearcore/pull/4679

Longarithm:
[...]

PR #4679, opened 2021-08-13
refactor: keep bench and release profile in sync to avoid surprises

matklad:
Cargo's smaller profiles, like `test` and `bench`, work in a rather
curious fashion. When you run `cargo test`, `cargo test --release` or
`cargo bench`, the tset/bench profiles apply only to a subset of code
being compiled. Most things are compiled with dev/release profile
instead.

To avoid surprises where things compile differently, it's advisable to
keep bench in sync with release.

cc https://github.com/near/nearcore/pull/4507

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -57,8 +57,6 @@ wasmer-compiler-singlepass = { git = "https://github.com/near/wasmer", branch = 
 overflow-checks = true
 
 [profile.bench]
-lto = true
-codegen-units = 1 # Use only 1 codegen-unit to enable full optimizations.
 overflow-checks = true
 
 [profile.dev.package.hex]


bowenwang1996 (commented):
@matklad but we do want lto to be enabled. The fact that it is not enabled by default was because we did not want to increase compilation time during development. In production it should be enabled through an environment variable

matklad:
Exactly: `cargo test --release` / `cargo build --release` is something you do during development, I don't think we use `bench` profile in production? 

https://github.com/near/nearcore/issues/4678 is an example of a problem which is caused by the mismatch.

matklad:
Interesting bits of profile triva:

* the `cogen-units = 1` option in the `bench` profile doesn't actually affect most of the code, only the test crate itself (so, bench profile doesn't actually match production profile).
* the `lto` does get enabled globally (surprised to me, to be honest: I'd expected lto in bench to be a no-op without LTO in the release profile as well)

mina86:
LTO would be kinda useless if it didn’t enable it transitively through all dependencies.

#4678 is about params estimator which wouldn’t be affected by the `bench` profile, no? I don’t see how this commit solves the cited issue.

matklad:
>LTO would be kinda useless if it didn’t enable it transitively through all dependencies.

That's more or less true for every other thing which is controlled by profile: it is not too useful to compile the tests themselves with codegen-units=1, and yet compile everything else with default value. Yet this is exactly how bench/test profiles work, they don't affect dependencies. This is not a useful behavior: bench&test profiles are a design mistake in cargo, which can't be removed due to backwrads compatibility sadly :(

The difference for LTO is that it simply *can't* be enabled for subset of the crates. I originally thought that Cargo only ltos if all involved profile tell it to lto, but apparently it's not the case. For example, if you specify lto in release, then `cargo test --release` won't lto. 

> I don’t see how this commit solves / affects the cited issue.

the problem there is that we try to run the same code in the estimator (via `cargo run`) and in an isolated fashion (ve `cargo test`), and get different results. That is confusing. That is, the problem is not that something is broken, but that, when you start investigating things, you get confused because behavior of `--release` cargo flag differs between `cargo test` and `cargo run`

matklad:
`cargo test --release` uses `bench`, and that's what used in #4678

mina86:
My only worry here is that benchmarks should probably run code which is close to production code. Maybe instead we should add `lto = true` to runtime-params-estimator instead? Either way it looks sensible to me.

bowenwang1996 (approved):

matklad:
>Maybe instead we should add lto = true to runtime-params-estimator instead?

Yeah, that's what I do separately, as a part of more general "make it easy to run the 'real' run of the estimator" -- estimatro has a bunch of toggles, lto's just one of them.

PR #4681, opened 2021-08-13
feat(neard): stop RPC servers gracefully after getting a signal

mina86:
With the default signal handlers actix_web HTTP servers uses, sending
a SIGINT results in a forceful shutdown of the servers even if there
are requests in flight.  This is unfriendly to the clients which may
as a result get a partial response or a broken connection.

Disable the signal handlers and instead explicitly call `stop(true)`
on the HTTP servers.  This gives the workers time to finish their
processing.  As a result, if a client successfully connected to an RPC
endpoint, they will get their response.  (That is, of course, unless
processing the request would take more than five seconds but that
should’t be the case with any of the supported RPC calls).

Fixes: https://github.com/near/nearcore/issues/4489


chain/jsonrpc/src/lib.rs
```diff
@@ -1207,13 +1207,14 @@ pub fn start_http(
     genesis_config: GenesisConfig,
     client_addr: Addr<ClientActor>,
     view_client_addr: Addr<ViewClientActor>,
-) {
+) -> Vec<(&'static str, actix_web::dev::Server)> {
```
bowenwang1996, line 1210: is it better to make this a struct? Or at least we can add a comment on what the return value is

bowenwang1996 (approved):

frol (approved):

PR #4684, opened 2021-08-14
Explain differences caused by LTO in params-estimator doc

Longarithm:
We recently observed tangible differences in estimation with enabled/disabled LTO, so it may worth explicit noting in the params-estimator readme.

See https://near.zulipchat.com/#narrow/stream/295306-dev-contract-runtime/topic/regular.20op.20cost for more details.


matklad (approved):
LGTM!

This all will be folded into `param-estimator --docker` command medium term.

PR #4690, opened 2021-08-16
pytest: download `neard` binary i/o `near`

mina86:
For now update the download URL to refer to the `neard` file rather
than `near` (they are the same binary; at some point `near` got renamed
to `neard`) only.  The file is still downloaded as `near-<branch>` and
there are other places in the code which expect it to be called that
way.


chefsale:
Noop, already fixed on master.

PR #4692, opened 2021-08-16
Rename `key_hash` from `TrieRefcountChange` to `trie_node_or_value_hash`

Longarithm:
As it was mentioned [here](https://github.com/near/nearcore/pull/4558#discussion_r676181684), `key_hash` is not a hash of trie key, but I missed some places while fixing it. Renamed it to ~~`value_hash`~~ `trie_node_or_value_hash`, because it is actually hash of the value.

Also, add some clarity on how the reference count is actually stored and updated.


core/store/src/trie/mod.rs
```diff
@@ -410,11 +410,12 @@ pub struct Trie {
...
 
-/// Stores reference count change for some Trie key and value.
+/// Stores reference count change for some key-value pair in DB.
 #[derive(BorshSerialize, BorshDeserialize, Clone, PartialEq, Eq, PartialOrd, Ord, Debug)]
 pub struct TrieRefcountChange {
-    /// Hash of TrieKey.
-    key_hash: CryptoHash,
+    /// Hash of the value and part of the DB key.
+    /// Used for uniting with shard id to get actual DB key.
+    value_hash: CryptoHash,
```
bowenwang1996, line 418: `value` is also ambiguous. This is actually the hash of a trie node right?

Longarithm: Not necessarily. I mention [here](https://github.com/near/nearcore/blob/master/docs/trie.md#trie-1) that it can be either serialized RawTrieNodeWithSize or value corresponding to some TrieKey. Also fixed comment below, because it was also incorrect.

Actually, I ran into this ambiguity so much times that I even think about splitting `ColState` to `ColStateNode` and `ColStateRecord`. Do you know any cases which would make that impossible? 

Though, I don't feel we will ever have time for doing this. The simplest option could be renaming `value` to `node_or_trie_value`.

bowenwang1996: > it can be either serialized RawTrieNodeWithSize or value

Ah yes that is what I meant. Somehow in my mind value is also a trie node...

bowenwang1996: > The simplest option could be renaming value to node_or_trie_value.

Do you mean `trie_node_or_value_hash`?

Longarithm: >Do you mean `trie_node_or_value_hash`?

Yes

Longarithm:
@bowenwang1996 renamed `key_hash` and `value` to `trie_node_or_value_hash` and `trie_node_or_value`, could you take a look?

bowenwang1996 (approved):

core/store/src/trie/mod.rs
```diff
@@ -410,13 +410,15 @@ pub struct Trie {
...
+/// Stores reference count change for some key-value pair in DB.
 #[derive(BorshSerialize, BorshDeserialize, Clone, PartialEq, Eq, PartialOrd, Ord, Debug)]
 pub struct TrieRefcountChange {
-    /// Hash of TrieKey.
-    key_hash: CryptoHash,
-    /// Value corresponding to the TrieKey stored in Trie.
-    value: Vec<u8>,
+    /// Hash of trie_node_or_value and part of the DB key.
+    /// Used for uniting with shard id to get actual DB key.
+    trie_node_or_value_hash: CryptoHash,
```
matklad, line 418: Nothing actionable, just a drive by comment:

Not a huge fan of `x_or_y` names, they are bulky. But also I can't suggest a better alternative. In rust-analyzer, we have exactly the same problem where we often need to deal with `NodeOrToken` types, By convention, we use `element` term there (element = a interior node or a leat token), but that only works if the name is sufficiently freqently used.

matklad (approved):

PR #4697, opened 2021-08-17
make `account_id_to_shard_id` and `num_shards` to take shard layout from the current epoch

mzhangmzz:
Now the signature becomes `account_id_to_shard_id(account_id: &AccountId, epoch_id: &EpochId)` and `num_shards(epoch_id: &EpochId)`

Test Plan: passes all existing tests. 

It is hard to fully test this change before we implement state transitioning for resharding. Before that, all epochs still use the same shard layout, so we can't actually test that `account_id_to_shard_id` and `num_shards` are using the correct shard layout from the current epoch. I will add tests later once we implement state transition.

### DIFF
--- a/chain/client/src/client_actor.rs
+++ b/chain/client/src/client_actor.rs
@@ -745,8 +745,8 @@ impl ClientActor {
                 == Some(&next_block_producer_account)
             {
                 let num_chunks = self.client.shards_mgr.num_chunks_for_block(&head.last_block_hash);
-                let have_all_chunks =
-                    head.height == 0 || num_chunks == self.client.runtime_adapter.num_shards();
+                let have_all_chunks = head.height == 0
+                    || num_chunks == self.client.runtime_adapter.num_shards(&epoch_id);
 
                 if self.client.doomslug.ready_to_produce_block(
                     Instant::now(),
@@ -1328,7 +1328,8 @@ impl ClientActor {
                 let block_header =
                     unwrap_or_run_later!(self.client.chain.get_block_header(&sync_hash));
                 let prev_hash = block_header.prev_hash().clone();
-                let shards_to_sync = (0..self.client.runtime_adapter.num_shards())
+                let epoch_id = self.client.chain.get_block_header(&sync_hash).unwrap().epoch_id();
+                let shards_to_sync = (0..self.client.runtime_adapter.num_shards(&epoch_id))
                     .filter(|x| {
                         self.client.shards_mgr.cares_about_shard_this_or_next_epoch(
                             me.as_ref(),
--- a/chain/client/src/client_actor.rs
+++ b/chain/client/src/client_actor.rs
@@ -746,7 +746,7 @@ impl ClientActor {
             {
                 let num_chunks = self.client.shards_mgr.num_chunks_for_block(&head.last_block_hash);
                 let have_all_chunks = head.height == 0
-                    || num_chunks == self.client.runtime_adapter.num_shards(&epoch_id);
+                    || num_chunks == self.client.runtime_adapter.num_shards(&epoch_id).unwrap();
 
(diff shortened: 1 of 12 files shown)


chain/client/src/client_actor.rs
```diff
@@ -1331,8 +1331,10 @@ impl ClientActor {
                     };
 
                 let me = self.client.validator_signer.as_ref().map(|x| x.validator_id().clone());
-                let shards_to_sync = (0..self.client.runtime_adapter.num_shards())
+                let epoch_id = self.client.chain.get_block_header(&sync_hash).unwrap().epoch_id();
+                let shards_to_sync = (0..self.client.runtime_adapter.num_shards(&epoch_id))
                     .filter(|x| {
+                        // TODO: this seems like it would break if epoch only has one epoch?
```
bowenwang1996, line 1337: Do you mean one block?

chain/client/src/view_client.rs
```diff
@@ -358,7 +358,10 @@ impl ViewClientActor {
         }
 
         let head = self.chain.head().map_err(|e| TxStatusError::ChainError(e))?;
-        let target_shard_id = self.runtime_adapter.account_id_to_shard_id(&signer_account_id);
+        // TODO: get_tx may not work properly after re-sharding for transactions processed
+        //       by previous shards, fix this
```
bowenwang1996, line 362: Please create an issue if you do not plan to fix it in the same PR

nearcore/src/runtime/mod.rs
```diff
@@ -358,9 +349,12 @@ impl NightshadeRuntime {
         ShardUId { version: shard_version, shard_id: shard_id as u32 }
     }
 
-    pub fn get_shard_layout(&self, epoch_id: &EpochId) -> ShardLayout {
+    fn account_id_to_shard_uid(&self, account_id: &AccountId, epoch_id: &EpochId) -> ShardUId {
         let mut epoch_manager = self.epoch_manager.as_ref().write().expect(POISONED_LOCK_ERR);
-        epoch_manager.get_shard_layout(epoch_id).unwrap().clone()
+        let shard_layout = epoch_manager.get_shard_layout(epoch_id).unwrap();
```
bowenwang1996, line 354: is it actually reasonable to unwrap here? How do we guarantee that we know the epoch?

bowenwang1996 (commented):
We need to make sure that `ShardLayout` is upgraded through a protocol upgrade properly. Is my understanding correct that the switch is entirely encapsulated within the implementation of `ShardLayout`?

chain/client/src/client_actor.rs
mzhangmzz: Yes, I fixed this in https://github.com/near/nearcore/pull/4705

nearcore/src/runtime/mod.rs
mzhangmzz: That's a good question. I didn't think very carefully about errors at all when I was designing the interface. I will go through these functions again and fix that

chain/client/src/view_client.rs
mzhangmzz: actually this is not an issue any more, I will remove the comment

mzhangmzz:
> We need to make sure that `ShardLayout` is upgraded through a protocol upgrade properly. Is my understanding correct that the switch is entirely encapsulated within the implementation of `ShardLayout`?

Yes, I already added this field in `GenesisConfig` https://github.com/near/nearcore/blob/master/core/chain-configs/src/genesis_config.rs#L135. Once we are ready, we just simply replace that field with the ShardLayout we want for simple nightshade

bowenwang1996 (approved):

chefsale (approved):

ailisp (approved):
Looks good!

Issue #4649, opened 2021-08-06
Move towards a better way to update RuntimeConfig


bowenwang1996:
Currently updating runtime config is quite painful and usually involves adding more structs and code to the database. While there are discussions about long-terms plans to potentially [move all protocol configs to a contract](https://gov.near.org/t/store-protocol-config-in-a-contract/2105), it is likely not going to be implemented any time soon. As a result, we should figure out a better way to do this in the meantime. After a discussion with @matklad, we agreed that it is better to move most of the config to data, i.e, some file that we can read from and write to. More specifically, for each change of config, we generate a new config file with all the parameters and their values. Globally we can maintain a `BTreeMap<ProtocolVersion, RuntimeConfig>` to determine at runtime which config to use. 

Some considerations here:
* We eventually want to do this for the entirety of protocol config (what is `GenesisConfig` now), but we can start with runtime config to make the change more incremental and easier to implement.
* The format of the config files needs to be decided. For now I think json is fine. @matklad has some other ideas though.
* When new fields are added in a protocol change, we need to be able to handle that.

bowenwang1996:
This is also somewhat related to #4447. cc @posvyatokum

matklad:
I suggest prioritizing removal of 

https://github.com/near/nearcore/blob/721a581d85b15813c3356f0a694002f1dac599d3/core/primitives/src/runtime/config.rs#L52-L98

It fully duplicates `RuntimeConfigStore` so it's pure technical dept. As we've introduced both `RuntimeConfigStore` and `ActualRuntimeConfig` relatively recently, fixing this *now* would be easier, as the relevant code is still in mental L1 cache for some people.

matklad:
@Longarithm let's close the issue? I think this is mostly done? There's probably some specific follow-ups, but they are worth opening separate issues for.

bowenwang1996:
@Longarithm any update on this?

Longarithm:
The issue is actually resolved, I forgot to close it.

PR #4698, opened 2021-08-17
feat: introduce structure for storing `RuntimeConfig`s

Longarithm:
First step towards https://github.com/near/nearcore/issues/4649.

**Idea**

Instead of modifying runtime for genesis, store configs for each protocol version where the config was updated.
Expected way of updating config:
- update nightly version (say, to 140) and put config `140.json` to `nearcore/res/runtime_configs` folder.
- after some time, update stable version (say, to 50) and rename `140.json` to `50.json`.

We can also add new fields with default values and remove old fields.
Overall, this should simplify the process of updating configs.

**Testing**

- Same as for `ActualRuntimeConfig`
- Check that `RuntimeConfigStore` and `ActualRuntimeConfig` give the same configs set

**Future plan**
- Simplify code, e.g. by getting rid of `RuntimeConfig::default()` which is not corresponding to any protocol version and has unclear definition;
- Submit new config with updated data receipt fees;
- Submit new config with updated function call and deploy contract fees, when they will also be estabilished.


core/primitives/Cargo.toml
```diff
@@ -18,6 +18,7 @@ byteorder = "1.3"
 chrono = { version = "0.4.4", features = ["serde"] }
 derive_more = "0.99.3"
 easy-ext = "0.2"
+include_dir = "0.6.1"
```
bowenwang1996, line 21: Is it possible to avoid this? Ideally we don't want to introduce new dependencies if possible

core/primitives/src/runtime/config_store.rs
```diff
@@ -0,0 +1,103 @@
...
+        Self {
+            store: BTreeMap::from_iter(runtime_configs_dir.files().iter().map(|file| {
+                let mut config: RuntimeConfig = serde_json::from_slice(file.contents()).unwrap();
+                if let Some(gas) = max_gas_burnt_view {
+                    config.wasm_config.limit_config.max_gas_burnt_view = gas;
+                }
+                (
+                    file.path().file_stem().unwrap().to_str().unwrap().parse().unwrap(),
+                    Arc::new(config),
+                )
```
bowenwang1996, line 34: This is somewhat hard to read. Could you introduce intermediate variables to make this more readable? Also I feel like there are too many `unwrap`s here and could potentially be simplified. cc @matklad

core/primitives/src/runtime/config_store.rs
```diff
@@ -0,0 +1,103 @@
...
+    /// Returns a `RuntimeConfig` for the corresponding protocol version.
+    ///
+    /// Note that even if some old version is given as the argument, this may
+    /// still return configuration which differs from configuration found in
+    /// genesis file by the `max_gas_burnt_view` limit.
+    pub fn get_config(&self, protocol_version: ProtocolVersion) -> &Arc<RuntimeConfig> {
+        self.store
+            .range((Bound::Unbounded, Bound::Included(protocol_version)))
+            .next_back()
+            .unwrap_or_else(|| {
```
bowenwang1996, line 48: I wonder whether it is better to return an option here.

bowenwang1996 (commented):
It is hard to review this PR in isolation without knowing where and how `RuntimeConfigStore` is going to be used

core/primitives/src/runtime/config_store.rs
matklad: Good question! As a rule of thumb, I usually do like pushing `.unwraps` to the caller, but I think in this case it's better to unwrap in the callee. 

My understanding is that we have an invariant that we only care about protocol versions below the the current max, so it's fine to assume that we know config for the current protocol version. In other words, if we return an option here, there's no way any current or future caller would want to not unwrap.

But thinking about this more, perhaps it's better to make this case statically impossible? We can require in the constructor that we pass config for the zero'th version, and then there's always a version which we can return.

Longarithm: I think it is fine to rename `29.json` to `0.json` and avoid panic, this should not break anything.

core/primitives/src/runtime/config_store.rs
Longarithm: I agree that current implementation of `RuntimeConfigStore::new` can be confusing, because it has internal logic unrelated to storage creation.

I think that making `RuntimeConfigStore::new` trivial and having a function `RuntimeConfigStore::from_max_gas_burnt_view` resolves it, because we need some function for building configs using custom `max_gas_burnt_view` anyway.

"Apply view-related parameters outside of `NightshadeRuntime`" looks like a good separate goal. Though, at first I would put `RuntimeConfigStore::from_max_gas_burnt_view(max_gas_burnt_view)` to `NightshadeRuntime` and then change the signature of `NightshadeRuntime` and put `RuntimeConfigStore` creation outside, because the latter change would affect lots of places.

core/primitives/src/runtime/config_store.rs
```diff
@@ -0,0 +1,103 @@
...
+    /// Maps protocol version to the config with possibly modified `max_gas_burnt_view` limit.
+    store: BTreeMap<ProtocolVersion, Arc<RuntimeConfig>>,
+}
+
+impl RuntimeConfigStore {
+    /// Constructs a new store.
+    ///
+    /// If `max_gas_burnt_view` is provided, the property in wasm limit
+    /// configuration will be adjusted to given value.
+    pub fn new(max_gas_burnt_view: Option<Gas>) -> Self {
```
mzhangmzz, line 23: What is special about `max_gas_burnt_view` that it is not stored as part of the file?

nearcore/res/runtime_configs/29.json
```diff
@@ -0,0 +1,180 @@
+{
```
mzhangmzz, line 1: Should we store the diff for each version instead of the whole file?

Currently, only `with_lower_storage_cost` changes between different version. Storing the whole file makes it hard to see what changed between different versions and can be error prone because of that. 

We should also add regression tests to make sure existing runtime_configs do not get changed once they are added.

core/primitives/src/runtime/config_store.rs
mzhangmzz: I agree with what @matklad suggested here. But I am also wondering, why are we setting `max_gas_burnt_view` separately at all? Why can't we just use the value set in `config` and just set `config.max_gas_burnt_view` to be the correct value at the callers of `NightshadeRuntime::new`. Setting this field separately seems very hacky.

And that should be done in a separate PR.

nearcore/res/runtime_configs/29.json
Longarithm: That's a good question.

I would vote for storing the whole files, because it improves readability - in the future, when we may have > 10 configs, it will be much easier to find the exact value for some protocol version, comparing with looking at all previous versions.

Also, I feel that applying diffs one-by-one is error prone in different sense - we never see the config which is actually used in production. So we would need the whole configs at least for testing.

For comparing configs, we already have a tool:

```
python3 ./runtime/runtime-params-estimator/emu-cost/compare_costs.py ./nearcore/res/runtime_configs/29.json ./nearcore/res/runtime_configs/42.json --diff
```

Before adding new config, I suggest running this tool and attaching the result to corresponding PR. Though at first I need to improve its usability a bit. 

What do others think? @matklad @bowenwang1996 

> We should also add regression tests to make sure existing runtime_configs do not get changed accidentally once they are added.

Yes, that's right.

core/primitives/src/runtime/config_store.rs
Longarithm: Yep, I agree with it. From what I see, `max_gas_burnt_view` is taken from `RuntimeConfig` only due to historical reasons.

nearcore/res/runtime_configs/29.json
bowenwang1996: I discussed the idea of storing diffs with @matklad before. We agreed that it felt like overengineering at this point. Also it is very likely we change many config values in the same PR (we will likely do this once we are ready to update the costs), which would make storing diffs less useful.

core/primitives/src/runtime/config_store.rs
Longarithm: Removing `max_gas_burnt_view` here: https://github.com/near/nearcore/pull/4716

Longarithm:
Addressed current changes and switched to `RuntimeConfigStore` in runtime.
Not fully ready to review yet because I need to refactor tests logic which also assume that `RuntimeConfig` is taken from `Genesis` and thus define everything in `Genesis`.

ailisp (approved):
Looks good to me!

Issue #4574, opened 2021-07-26
Flaky behavior in nayduck


bowenwang1996:
It appears that sometimes tests fail for reasons unrelated to the tests themselves. Examples:
* http://nayduck.eastus.cloudapp.azure.com:3000/#/test/187252
* http://nayduck.eastus.cloudapp.azure.com:3000/#/test/187386
* http://nayduck.eastus.cloudapp.azure.com:3000/#/test/187368
* http://nayduck.eastus.cloudapp.azure.com:3000/#/test/187378

And it appears that there is some weird cascading effects here where such failures often appear together in multiple tests.

mina86:
I’m certainly not opposed to that.

PR #4700, opened 2021-08-17
tests: make panicks in near-test-contracts more verbose

mina86:
Include the full path of the file that the test tried to read in
panic message.

Issue: https://github.com/near/nearcore/issues/4574

### DIFF
--- a/runtime/near-test-contracts/src/lib.rs
+++ b/runtime/near-test-contracts/src/lib.rs
@@ -1,32 +1,35 @@
-use std::io;
 use std::path::Path;
 
 use once_cell::sync::OnceCell;
 
 pub fn rs_contract() -> &'static [u8] {
     static CONTRACT: OnceCell<Vec<u8>> = OnceCell::new();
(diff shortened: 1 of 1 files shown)


chefsale:
Seems like the same test like in https://github.com/near/nearcore/pull/4701 is failing due to flakiness (here on nightly), created an issue for this: https://github.com/near/nearcore/issues/4702. cc: @bowenwang1996.

Restarted the failing tests, otherwise looks good.

chefsale (approved):

bowenwang1996 (approved):

matklad (approved):
LGTM!

I don't like panicking in the callee, it's generally is better to let the caller unwrap -- panicking is fine "in the small", but code often evolves to be bigger than originally envisioned, and, if you don't return errors from the start, it might be difficult to change to that later. 

So, one alternative is to use `anyhow` crate and `context` method -- while this is an overkill for this specifc case, if someone tries to copy-pasted or extend the code later, they'll start with a more solid base. 

I am +0 on using anyhow here.

PR #4701, opened 2021-08-17
pytest: verify that `genesis-populate` run succeeded

mina86:
Popen.wait does not check whether the executed command failed or not and
there was nothing in genesis_populate that would test that.  As a result,
the test would ignore any failures in the genesis-populate tool.  Change
the code to use subprocess.check_call instead which raises on failure.

Issue: https://github.com/near/nearcore/issues/4574

### DIFF
--- a/pytest/lib/populate.py
+++ b/pytest/lib/populate.py
@@ -1,17 +1,16 @@
-from subprocess import Popen
+import subprocess
 from os.path import join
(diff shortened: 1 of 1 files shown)


chefsale:
Retrying `cargo test` which seems flaky. Otherwise seems good.

chefsale (approved):

PR #4705, opened 2021-08-17
fix: two places in the code where we should be prev_hash instead of hash

mzhangmzz:
Fix two places in the code where it should use hash of the previous block instead of the current block. One is at a test, and another one in state_sync.


chain/client/src/client_actor.rs
```diff
@@ -1331,11 +1331,14 @@ impl ClientActor {
...
                 let me = self.client.validator_signer.as_ref().map(|x| x.validator_id().clone());
+                let block_header =
+                    unwrap_or_run_later!(self.client.chain.get_block_header(&sync_hash));
+                let prev_hash = block_header.prev_hash();
                 let shards_to_sync = (0..self.client.runtime_adapter.num_shards())
                     .filter(|x| {
                         self.client.shards_mgr.cares_about_shard_this_or_next_epoch(
                             me.as_ref(),
-                            &sync_hash,
+                            prev_hash,
```
bowenwang1996, line 1341: Why should we use prev hash here?

bowenwang1996: hmm...not sure how this code worked before....

bowenwang1996:
@mzhangmzz could you make sure that CI passes and also run nayduck tests? I wonder whether this fix reveals some unexpected issues

chain/client/src/client_actor.rs
mzhangmzz: I think this will only cause a bug if epochs only have one block. Maybe the state sync tests never tested that case

mzhangmzz:
> @mzhangmzz could you make sure that CI passes and also run nayduck tests? I wonder whether this fix reveals some unexpected issues

Nayduck tests passed http://nayduck.eastus.cloudapp.azure.com:3000/#/run/2004
There is one failing but I think it fails on master too

bowenwang1996 (approved):

PR #4707, opened 2021-08-18
refactor: remove work-around for cargo features

matklad:
refactor: remove work-around for cargo features

The `node-runtime` dependency in `client` creates a cyclic dependency:

near_client -> node_runtime -> testlib -> near_client

Looks like we can remove it just now, because the dep is used to
work-around cargo features, and is not otherwise used.

I think this was enabled by the recent work on

https://github.com/near/nearcore/issues/4357

Test Plan
--------

```
cargo check --all-targets && \
cargo check --all-targets --features sandbox && \
cargo check --all-targets -p near-client && \
cargo check --all-targets -p near-client --features sandbox
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2801,7 +2801,6 @@ dependencies = [
  "near-telemetry",
  "near-test-contracts",
  "near-vm-runner",
- "node-runtime",
  "num-rational",
  "rand 0.7.3",
  "reed-solomon-erasure",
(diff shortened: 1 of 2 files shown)


matklad:
cc @nikurt looks like your recent work let us to remove one hack! cc @ailisp

bowenwang1996 (commented):
Epic 🚀

bowenwang1996 (approved):

PR #4712, opened 2021-08-19
Move network tests to integration tests to avoid circular dependencies of network->near_client

nikurt:
Issue #4357


integration-tests/tests/network/ban_peers.rs
```diff
@@ -1,8 +1,6 @@
-pub use runner::*;
+pub use crate::runner::*;
 use std::time::Duration;
 
-mod runner;
```
matklad, line 4: :+1: we no longer need to recompile the `runner` module for each test over and over again!

matklad (approved):

matklad:
Confirmed that this closes #4357, thank you so much for working on this @nikurt !

bowenwang1996 (approved):

PR #4714, opened 2021-08-19
Apply state parts in separate actor

EgorKulikov:
Instead of waiting for state parts to be applied we now do this in separate thread


chain/chain/src/chain.rs
```diff
@@ -1767,26 +1771,79 @@ impl Chain {
...
+                    epoch_id,
+                    source,
+                });
+                self.currently_applying_state_part.insert(map_key, part_id);
+                return Ok(SetStateFinalizeResult::ScheduledNext);
+            } else {
+                self.currently_applying_state_part.remove(&map_key);
+            }
+        } else {
+            if let Some(_) = current_task {
```
bowenwang1996, line 1822:
```suggestion
            if current_task.is_some() {
```

bowenwang1996 (commented):
What is the test plan? One simple approach is to sync a node to testnet and make sure it can sync without issues, but this likely will take some time. It may be a good idea to revive sync_massive.py

chain/chain/src/chain.rs
EgorKulikov: I guess I've got too addicted to if let and match to use old plain if

chain/client/src/client_actor.rs
```diff
@@ -121,8 +123,18 @@ impl ClientActor {
         validator_signer: Option<Arc<dyn ValidatorSigner>>,
         telemetry_actor: Addr<TelemetryActor>,
         enable_doomslug: bool,
+        ctx: &Context<ClientActor>,
         #[cfg(feature = "adversarial")] adv: Arc<RwLock<AdversarialControls>>,
     ) -> Result<Self, Error> {
+        let state_parts_arbiter = Arbiter::new();
```
matklad, line 129: An `Arbiter` is basically an `std::thread`, so the two usual questions apply: 

* how do we make sure that the thread eventually stops? 
* how do we make sure that, when the thread panics due to a bug, we eventually notice this and tear-down the process? 

For the first, I think what we might want to do is to store the arbiter as a field of `ClientActor`, and call `arbiter.stop(); arbiter.join()` in the `Drop` impl. I think we should do that, otherwise writing tests becomes cumbersome, as `StatePartsActor` from old test might somehow interferer with newer tests. 

Not sure what we need to do with the second case. What would be behavior when `apply_state_part` panics? I guess the client actor will be just infinitely stuck in the old epoch. Seems that, if the background actor dies, the `do_send` will just drop messages on the floor. Is this an OK behavior? Can we link the ClientActor and the StatePartsActor such that the client is guaranteed to die after `StatePartsActor`?

chain/chain/src/chain.rs
```diff
@@ -1767,26 +1771,79 @@ impl Chain {
...
+                    epoch_id,
+                    source,
+                });
+                self.currently_applying_state_part.insert(map_key, part_id);
+                return Ok(SetStateFinalizeResult::ScheduledNext);
+            }
         }
 
+        // We've done with applying parts
+
```
matklad, line 1846: This probably wants

`assert!(num_parts == 0 && self.currently_applying_state_part.is_empty())`

assert (not sure if the condition I've got is right).

chain/chain/src/chain.rs
```diff
@@ -1767,26 +1771,79 @@ impl Chain {
...
+                    if *scheduled_part_id != result.part_id {
+                        // We had probably had to abort process last time
+                        return Ok(SetStateFinalizeResult::NoAction);
+                    }
+                }
+                None => {
+                    // We had probably had to abort process last time
+                    return Ok(SetStateFinalizeResult::NoAction);
+                }
+            }
```
matklad, line 1801: These cases seem operationally interesting, makes sense to `tracing::info` them.

matklad (commented):
Left a bunch of comments! 

This looks broadly reasonable to me, but I'll probably take another look later. 


Here's a bunch of "idiom" style suggestions. Not sending them inline to not distract from correctness aspects of the code :)

* We might want to define `pub struct PartId(pub u64)` instead of using raw `u64`, it makes it clearer what we should compare to what.
* A bunch of functions take `arg: &Box<dyn Fn(...)>` arguments. They should be changed to `arg: &dyn Fn(...)`, as `&Box` is a useless double indirection.
* Similarly, `arg: &Option<StatePartsResponse>` -> `arg: Option<&StatePartsResponse>`.
  That would be "more idiomatic", where idiomatic means:
    * more flexible for the caller: you can always get an `Option<&T>` out of `&Option<T>` using `Option::as_ref`, but going in the opposite direction is not possible
    * nicer semantics/generated code: `Option<&T>` is just a pointer to T, which is a null pointer in the case of None. `&Option<T>` is a pointer to a pair of bool and T.

chain/client/src/client_actor.rs
```diff
@@ -161,6 +173,9 @@ impl ClientActor {
             doomslug_timer_next_attempt: now,
             chunk_request_retry_next_attempt: now,
             sync_started: false,
+            state_parts_task_scheduler: Box::new(move |msg: StatePartsMessage| {
```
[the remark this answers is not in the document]
EgorKulikov: Tbh first I just wanted to avoid making chain dependent on actix. Then it became unavoidable, but some tests can live with empty closures, so I left it as is

chain/client/src/client_actor.rs
bowenwang1996: > Can we link the ClientActor and the StatePartsActor such that the client is guaranteed to die after StatePartsActor?

That is an interesting question. I am not sure whether that is necessarily something we want to do. It sort of depends on why exactly `apply_state_part` panicked and we could potentially still recover by retrying.

mzhangmzz:
I have some questions on why we decide to implement it this way.

1. Why not let StatePartActor process all parts for one shard at one time together, instead of sending one request for each part one by one? 
2. It's quite confusing to pass `prev_execution_result` to `sync` and then propagate that all the way to `sync_shards_status` to be used for case `ShardSyncStatus::StateDownloadFinalize`. Why not move that part of logic to the handler of `StatePartsResponse` directly?
3. This probably should be done in a different PR, but we should think about parallelizing state sync for shards too, now that chunks processing are parallelized.

chain/chain/src/chain.rs
```diff
@@ -1799,32 +1800,47 @@ impl Chain {
...
 
-            self.runtime_adapter.apply_state_part(
-                shard_id,
-                &state_root,
-                part_id,
-                num_parts,
-                &part,
-                &epoch_id,
-            )?;
+            parts.push(part);
```
mzhangmzz, line 1820: I think you cannot read `parts` here and send them over through `StatePartsMessage` because this will trigger OOM if the state is too large. The reason why the code needed to divide state to parts and apply it one by one is to limit memory usage.

To solve the issue, I think you will need to store a `Arc<Store>` in `StatePartsActor` and read the parts there.

EgorKulikov: Right, I thought for some reason this have to do with downloading. Fixed

EgorKulikov:
It is a bit weird that Chain::new do something even if we disregard result

EgorKulikov:
http://nayduck.near.org/#/run/2068 Seems like 4 failing tests are serial offenders. I should be able to test state catch up on mainnet by tomorrow

chain/chain/src/chain.rs
```diff
@@ -4030,3 +4038,21 @@ pub fn collect_receipts_from_response(
         receipt_proof_response.iter().flat_map(|ReceiptProofResponse(_, proofs)| proofs),
     )
 }
+
+#[derive(Message)]
+#[rtype(result = "()")]
+pub struct StatePartsMessage {
```
mzhangmzz, line 4044: Could you rename these two structs to `ApplyStatePartsRequest` and `ApplyStatePartsResponse`? I think that explains what these messages imply better.

chain/client/src/client_actor.rs
```diff
@@ -1484,6 +1509,67 @@ impl ClientActor {
...
+    runtime: Arc<dyn RuntimeAdapter>,
+    client_addr: Addr<ClientActor>,
+}
+
+impl StatePartsActor {
+    fn apply_parts(
+        &mut self,
+        msg: &StatePartsMessage,
+    ) -> Result<(), near_chain_primitives::error::Error> {
+        let store = self.runtime.get_store();
```
mzhangmzz, line 1522: nit: since store is used at every request, it is better to make it a field in `StatePartsActor`, instead of getting it every time

chain/chain/src/chain.rs
```diff
@@ -1799,32 +1800,38 @@ impl Chain {
         Ok(())
     }
 
-    pub fn set_state_finalize(
+    pub fn start_set_state_finalize(
```
mzhangmzz, line 1803: maybe rename this function to `schedule_apply_state_parts`, I think that describes better what this function does

chain/chain/src/chain.rs
```diff
@@ -1799,32 +1800,38 @@ impl Chain {
...
-                part_id,
-                num_parts,
-                &part,
-                &epoch_id,
-            )?;
-        }
+        Ok(())
+    }
+
+    pub fn end_set_state_finalize(
```
mzhangmzz, line 1825: Then you can just call this function `set_state_finalize`

chain/client/src/sync.rs
```diff
@@ -744,20 +750,25 @@ impl StateSync {
                         update_sync_status = true;
                         *shard_sync_download = ShardSyncDownload {
                             downloads: vec![],
-                            status: ShardSyncStatus::StateDownloadFinalize,
+                            status: ShardSyncStatus::StateDownloadDownloaded,
                         };
                     }
                 }
-                ShardSyncStatus::StateDownloadFinalize => {
+                ShardSyncStatus::StateDownloadDownloaded => {
```
mzhangmzz, line 757: Maybe this stage can be called `StateDownloadApplyParts`? I think the names of all other stages describe what it will do in this stage, not what has been done. Just to keep the naming consistent

chain/client/src/sync.rs
```diff
@@ -770,6 +781,37 @@ impl StateSync {
                         }
                     }
                 }
+                ShardSyncStatus::StateDownloadApplying => {
```
mzhangmzz, line 784: Following the same logic, this stage can be called `StateDownloadFinalizeState`.

Or, I think it can just be merged to the next stage 'StateDownloadComplete`. Both seem fine to me.

chain/client/src/client_actor.rs
```diff
@@ -1484,6 +1509,67 @@ impl ClientActor {
     }
 }
 
+struct StatePartsActor {
```
mzhangmzz, line 1512: I think we might want to change the name to be more generic to include other jobs we want to separate from `ClientActor`, such as splitting states and block catchup. I am not super good at making names, but maybe something like `BackgroundClientActor` could work (probably not the best name, but hopefully the message I want to convey gets through). The idea is that we can put all job that might block `ClientActor` to there.

Similarly, the messages between should also be an enum. It can be something like

`
pub enum BackgroundJobRequest {
    ApplyStatePartsRequest(ApplyStatePartsRequest),
    SplitStateRequest(SplitStateRequest),
    BlockCatchupRequest(BlockCatchupRequest),
}
`

chain/client/src/sync.rs
EgorKulikov: But we are in this state only for very short time - when we've downloaded parts but not yet scheduled them to be applied

chain/client/src/sync.rs
EgorKulikov: While we are in this state while we apply them in other thread. So I think my naming is more in line with convention

chain/client/src/client_actor.rs
EgorKulikov: I am not sure we will use same actor for those jobs

chain/client/src/client_actor.rs
EgorKulikov: This is pretty heavy job, I do not see why we should optimize one Arc::clone out of it. But if for some reason in the future runtime.store will point somewhere else than it points during ClientActor initialization we will save someone a lot of debugging

chain/client/src/client_actor.rs
EgorKulikov: as an aside, we can just implement Handler trait for different messages without need for enum

chain/client/src/sync.rs
mzhangmzz: I see what you mean. Sorry for being nit picky, but maybe the following way is clearer. We can move the waiting logic to this stage 

```
let result = self.state_parts_apply_results.remove(&shard_id);
                    if let Some(result) = result {
```
and only proceed to the next stage once the result is ready. This way, it is clear that at one stage, we are applying the state parts, the next one we are doing the post processing after state parts being applied

chain/client/src/client_actor.rs
mzhangmzz: I am not sure about block catchup, but I think it makes sense to put split state into the same actor. This way, we can have fewer actors in the code to maintain. But maybe it can be done at another PR.

frol (approved):
LGTM modulo the concerns raised by @matklad

chain/client/src/sync.rs
EgorKulikov: I am not sure what you mean. Waiting logic is already in the stage code of which you cited

EgorKulikov: Maybe we can rename this one to StateDownloadScheduling

EgorKulikov: And I am not consider this not picking, I am all for better naming

chain/client/src/client_actor.rs
```diff
@@ -161,6 +176,9 @@ impl ClientActor {
             doomslug_timer_next_attempt: now,
             chunk_request_retry_next_attempt: now,
             sync_started: false,
+            state_parts_task_scheduler: Box::new(move |msg: ApplyStatePartsRequest| {
+                state_parts_actor_addr.do_send(msg);
+            }),
```
bowenwang1996, line 181: What is the advantage of doing this versus having state_parts_actor_addr directly?

chain/client/src/client_actor.rs
bowenwang1996: Good point. I agree that it probably makes sense to use the same actor for splitting state and applying state parts

nearcore/src/runtime/mod.rs
```diff
@@ -629,6 +629,10 @@ impl RuntimeAdapter for NightshadeRuntime {
         (self.store.clone(), self.genesis_state_roots.clone())
     }
 
+    fn get_store(&self) -> Arc<Store> {
+        self.store.clone()
+    }
```
bowenwang1996, line 634: This is a bit weird. I think it is intentional that we do not want to expose `Store` from `NightshadeRuntime` to avoid breaking the abstraction.

chain/chain/src/chain.rs
```diff
@@ -1799,32 +1800,38 @@ impl Chain {
...
+    }
 
-        // Applying the chunk starts here
+    pub fn set_state_finalize(
+        &mut self,
+        shard_id: ShardId,
+        sync_hash: CryptoHash,
+        apply_result: Result<(), near_chain_primitives::Error>,
+    ) -> Result<(), Error> {
+        apply_result?;
```
bowenwang1996, line 1831: This code feels a bit strange to me. If the sole purpose is to return error if an error is passed in, then the caller should probably handle the error instead.

bowenwang1996 (commented):
@matklad could you review this PR please 🙏

EgorKulikov:
Seems to be working on mainnet:
`Sep 15 04:18:48.453  INFO stats: #46199317 Downloading headers 99.99% (4420)  30/28/40 peers ⬇ 654.0kiB/s ⬆ 43.6kiB/s 0.00 bps 0 gas/s CPU: 65%, Mem: 2.5 GiB
Sep 15 04:18:58.455  INFO stats: #46199317 Downloading headers 99.99% (2381)  30/28/40 peers ⬇ 683.8kiB/s ⬆ 43.8kiB/s 0.00 bps 0 gas/s CPU: 84%, Mem: 2.5 GiB
Sep 15 04:19:09.150  INFO stats: #46199317 Downloading headers 100.00% (854)  30/28/40 peers ⬇ 712.8kiB/s ⬆ 44.2kiB/s 0.00 bps 0 gas/s CPU: 81%, Mem: 2.5 GiB
Sep 15 04:26:00.703  INFO stats: State [0: header]  30/29/40 peers ⬇ 152.4kiB/s ⬆ 51.4kiB/s 0.00 bps 0 gas/s CPU: 90%, Mem: 3.8 GiB
Sep 15 04:26:10.705  INFO stats: State [0: parts]  30/6/40 peers ⬇ 2.7MiB/s ⬆ 59.8kiB/s 0.00 bps 0 gas/s CPU: 92%, Mem: 3.9 GiB
Sep 15 04:26:15.705  INFO telemetry: Telemetry data could not be sent due to: Failed to connect to host: Timeout while establishing connection
Sep 15 04:26:20.712  INFO stats: State [0: parts]  30/6/40 peers ⬇ 3.0MiB/s ⬆ 59.8kiB/s 0.00 bps 0 gas/s CPU: 39%, Mem: 3.9 GiB
Sep 15 04:26:30.714  INFO stats: State [0: parts]  30/3/40 peers ⬇ 4.2MiB/s ⬆ 52.5kiB/s 0.00 bps 0 gas/s CPU: 138%, Mem: 3.9 GiB
Sep 15 04:26:40.717  INFO stats: State [0: parts]  30/1/40 peers ⬇ 5.4MiB/s ⬆ 46.5kiB/s 0.00 bps 0 gas/s CPU: 96%, Mem: 3.9 GiB
Sep 15 04:26:50.720  INFO stats: State [0: parts]  30/3/40 peers ⬇ 6.6MiB/s ⬆ 26.9kiB/s 0.00 bps 0 gas/s CPU: 92%, Mem: 3.9 GiB
Sep 15 04:27:00.725  INFO stats: State [0: parts]  30/3/40 peers ⬇ 7.9MiB/s ⬆ 22.8kiB/s 0.00 bps 0 gas/s CPU: 80%, Mem: 3.9 GiB
Sep 15 04:27:00.735  WARN sync: State sync didn't download the state for shard 0 in 60 seconds, sending StateRequest again
Sep 15 04:27:10.728  INFO stats: State [0: parts]  30/7/40 peers ⬇ 7.4MiB/s ⬆ 16.7kiB/s 0.00 bps 0 gas/s CPU: 96%, Mem: 3.9 GiB
Sep 15 04:27:20.730  INFO stats: State [0: parts]  30/1/40 peers ⬇ 7.9MiB/s ⬆ 23.2kiB/s 0.00 bps 0 gas/s CPU: 77%, Mem: 3.9 GiB
Sep 15 04:27:30.734  INFO stats: State [0: parts]  30/3/40 peers ⬇ 8.1MiB/s ⬆ 27.1kiB/s 0.00 bps 0 gas/s CPU: 68%, Mem: 4.0 GiB
Sep 15 04:27:40.736  INFO stats: State [0: parts]  30/1/40 peers ⬇ 8.2MiB/s ⬆ 28.4kiB/s 0.00 bps 0 gas/s CPU: 58%, Mem: 4.0 GiB
Sep 15 04:27:50.738  INFO stats: State [0: parts]  30/3/40 peers ⬇ 8.4MiB/s ⬆ 30.1kiB/s 0.00 bps 0 gas/s CPU: 95%, Mem: 4.0 GiB
Sep 15 04:28:00.741  INFO stats: State [0: parts]  30/2/40 peers ⬇ 8.4MiB/s ⬆ 28.3kiB/s 0.00 bps 0 gas/s CPU: 48%, Mem: 4.0 GiB
Sep 15 04:28:00.743  WARN sync: State sync didn't download the state for shard 0 in 60 seconds, sending StateRequest again
Sep 15 04:28:10.744  INFO stats: State [0: parts]  30/3/40 peers ⬇ 8.5MiB/s ⬆ 22.2kiB/s 0.00 bps 0 gas/s CPU: 98%, Mem: 3.8 GiB
Sep 15 04:28:20.746  INFO stats: State [0: parts]  30/4/40 peers ⬇ 8.5MiB/s ⬆ 20.7kiB/s 0.00 bps 0 gas/s CPU: 99%, Mem: 3.8 GiB
Sep 15 04:28:30.749  INFO stats: State [0: parts]  30/1/40 peers ⬇ 8.5MiB/s ⬆ 18.7kiB/s 0.00 bps 0 gas/s CPU: 87%, Mem: 3.8 GiB
Sep 15 04:28:40.751  INFO stats: State [0: parts]  30/1/40 peers ⬇ 8.6MiB/s ⬆ 19.4kiB/s 0.00 bps 0 gas/s CPU: 64%, Mem: 3.8 GiB
Sep 15 04:28:50.753  INFO stats: State [0: parts]  30/1/40 peers ⬇ 8.4MiB/s ⬆ 23.9kiB/s 0.00 bps 0 gas/s CPU: 49%, Mem: 3.8 GiB
Sep 15 04:29:00.755  WARN sync: State sync didn't download the state for shard 0 in 60 seconds, sending StateRequest again
Sep 15 04:29:00.769  INFO stats: State [0: parts]  30/1/40 peers ⬇ 8.6MiB/s ⬆ 32.1kiB/s 0.00 bps 0 gas/s CPU: 62%, Mem: 3.8 GiB
Sep 15 04:29:06.303  INFO telemetry: Telemetry data could not be sent due to: Failed to connect to host: Timeout while establishing connection
Sep 15 04:29:10.772  INFO stats: State [0: parts]  30/2/40 peers ⬇ 8.6MiB/s ⬆ 30.6kiB/s 0.00 bps 0 gas/s CPU: 44%, Mem: 3.8 GiB
Sep 15 04:29:20.774  INFO stats: State [0: parts]  30/1/40 peers ⬇ 8.5MiB/s ⬆ 26.9kiB/s 0.00 bps 0 gas/s CPU: 33%, Mem: 3.8 GiB
Sep 15 04:29:30.777  INFO stats: State [0: parts]  30/2/40 peers ⬇ 7.8MiB/s ⬆ 23.9kiB/s 0.00 bps 0 gas/s CPU: 70%, Mem: 3.8 GiB
Sep 15 04:29:40.780  INFO stats: State [0: parts]  30/5/40 peers ⬇ 9.2MiB/s ⬆ 23.3kiB/s 0.00 bps 0 gas/s CPU: 140%, Mem: 3.8 GiB
Sep 15 04:29:50.783  INFO stats: State [0: parts]  30/4/40 peers ⬇ 10.7MiB/s ⬆ 30.5kiB/s 0.00 bps 0 gas/s CPU: 123%, Mem: 3.8 GiB
Sep 15 04:30:00.769  WARN sync: State sync didn't download the state for shard 0 in 60 seconds, sending StateRequest again
Sep 15 04:30:00.787  INFO stats: State [0: parts]  30/2/40 peers ⬇ 12.2MiB/s ⬆ 33.9kiB/s 0.00 bps 0 gas/s CPU: 88%, Mem: 3.8 GiB
Sep 15 04:30:10.796  INFO stats: State [0: parts]  30/3/40 peers ⬇ 14.5MiB/s ⬆ 45.6kiB/s 0.00 bps 0 gas/s CPU: 144%, Mem: 3.8 GiB
Sep 15 04:30:20.797  INFO stats: State [0: parts]  30/2/40 peers ⬇ 16.7MiB/s ⬆ 55.9kiB/s 0.00 bps 0 gas/s CPU: 104%, Mem: 3.8 GiB
Sep 15 04:30:30.801  INFO stats: State [0: parts]  30/2/40 peers ⬇ 19.0MiB/s ⬆ 66.8kiB/s 0.00 bps 0 gas/s CPU: 110%, Mem: 3.8 GiB
Sep 15 04:30:40.806  INFO stats: State [0: parts]  30/3/40 peers ⬇ 19.1MiB/s ⬆ 85.0kiB/s 0.00 bps 0 gas/s CPU: 99%, Mem: 3.8 GiB
Sep 15 04:30:50.809  INFO stats: State [0: parts]  30/4/40 peers ⬇ 19.0MiB/s ⬆ 81.8kiB/s 0.00 bps 0 gas/s CPU: 146%, Mem: 3.8 GiB
Sep 15 04:31:00.779  WARN sync: State sync didn't download the state for shard 0 in 60 seconds, sending StateRequest again
Sep 15 04:31:00.812  INFO stats: State [0: parts]  30/11/40 peers ⬇ 18.8MiB/s ⬆ 85.9kiB/s 0.00 bps 0 gas/s CPU: 98%, Mem: 3.8 GiB
Sep 15 04:31:10.815  INFO stats: State [0: parts]  30/8/40 peers ⬇ 18.7MiB/s ⬆ 84.1kiB/s 0.00 bps 0 gas/s CPU: 143%, Mem: 3.8 GiB
Sep 15 04:31:20.818  INFO stats: State [0: parts]  30/9/40 peers ⬇ 18.7MiB/s ⬆ 90.7kiB/s 0.00 bps 0 gas/s CPU: 168%, Mem: 3.8 GiB
Sep 15 04:31:30.821  INFO stats: State [0: parts]  30/11/40 peers ⬇ 18.7MiB/s ⬆ 92.7kiB/s 0.00 bps 0 gas/s CPU: 144%, Mem: 3.8 GiB
Sep 15 04:31:40.846  INFO stats: State [0: parts]  30/11/40 peers ⬇ 18.6MiB/s ⬆ 82.8kiB/s 0.00 bps 0 gas/s CPU: 131%, Mem: 3.8 GiB
Sep 15 04:31:50.848  INFO stats: State [0: parts]  30/13/40 peers ⬇ 18.7MiB/s ⬆ 82.7kiB/s 0.00 bps 0 gas/s CPU: 124%, Mem: 3.8 GiB
Sep 15 04:32:00.788  WARN sync: State sync didn't download the state for shard 0 in 60 seconds, sending StateRequest again
Sep 15 04:32:00.855  INFO stats: State [0: parts]  30/16/40 peers ⬇ 18.5MiB/s ⬆ 76.1kiB/s 0.00 bps 0 gas/s CPU: 94%, Mem: 3.8 GiB
Sep 15 04:32:10.859  INFO stats: State [0: parts]  30/18/40 peers ⬇ 18.5MiB/s ⬆ 74.0kiB/s 0.00 bps 0 gas/s CPU: 133%, Mem: 3.8 GiB
Sep 15 04:32:20.860  INFO stats: State [0: parts]  30/14/40 peers ⬇ 18.4MiB/s ⬆ 63.5kiB/s 0.00 bps 0 gas/s CPU: 76%, Mem: 3.8 GiB
Sep 15 04:32:30.863  INFO stats: State [0: parts]  30/11/40 peers ⬇ 17.7MiB/s ⬆ 56.3kiB/s 0.00 bps 0 gas/s CPU: 42%, Mem: 3.8 GiB
Sep 15 04:32:40.864  INFO stats: State [0: parts]  30/12/40 peers ⬇ 16.0MiB/s ⬆ 48.7kiB/s 0.00 bps 0 gas/s CPU: 66%, Mem: 3.8 GiB
Sep 15 04:32:50.868  INFO stats: State [0: parts]  30/21/40 peers ⬇ 15.3MiB/s ⬆ 63.4kiB/s 0.00 bps 0 gas/s CPU: 63%, Mem: 3.8 GiB
Sep 15 04:33:00.789  WARN sync: State sync didn't download the state for shard 0 in 60 seconds, sending StateRequest again
Sep 15 04:33:00.872  INFO stats: State [0: parts]  30/23/40 peers ⬇ 14.6MiB/s ⬆ 62.6kiB/s 0.00 bps 0 gas/s CPU: 55%, Mem: 3.8 GiB
Sep 15 04:33:10.876  INFO stats: State [0: parts]  30/24/40 peers ⬇ 14.0MiB/s ⬆ 62.1kiB/s 0.00 bps 0 gas/s CPU: 54%, Mem: 3.8 GiB
Sep 15 04:33:20.878  INFO stats: State [0: parts]  30/27/40 peers ⬇ 12.9MiB/s ⬆ 64.5kiB/s 0.00 bps 0 gas/s CPU: 48%, Mem: 3.8 GiB
Sep 15 04:33:30.882  INFO stats: State [0: parts]  30/21/40 peers ⬇ 11.7MiB/s ⬆ 64.5kiB/s 0.00 bps 0 gas/s CPU: 27%, Mem: 3.8 GiB
Sep 15 04:33:40.884  INFO stats: State [0: parts]  30/23/40 peers ⬇ 10.3MiB/s ⬆ 67.8kiB/s 0.00 bps 0 gas/s CPU: 41%, Mem: 3.8 GiB
Sep 15 04:33:50.887  INFO stats: State [0: applying parts]  30/25/40 peers ⬇ 8.7MiB/s ⬆ 52.1kiB/s 0.00 bps 0 gas/s CPU: 150%, Mem: 4.3 GiB
Sep 15 04:34:00.890  INFO stats: State [0: applying parts]  30/26/40 peers ⬇ 7.2MiB/s ⬆ 54.5kiB/s 0.00 bps 0 gas/s CPU: 194%, Mem: 4.3 GiB
Sep 15 04:34:10.891  INFO stats: State [0: applying parts]  30/26/40 peers ⬇ 5.0MiB/s ⬆ 56.4kiB/s 0.00 bps 0 gas/s CPU: 192%, Mem: 4.3 GiB
Sep 15 04:34:20.894  INFO stats: State [0: applying parts]  30/25/40 peers ⬇ 3.4MiB/s ⬆ 52.4kiB/s 0.00 bps 0 gas/s CPU: 227%, Mem: 4.4 GiB
Sep 15 04:34:30.897  INFO stats: State [0: applying parts]  30/26/40 peers ⬇ 3.0MiB/s ⬆ 52.9kiB/s 0.00 bps 0 gas/s CPU: 187%, Mem: 4.5 GiB
Sep 15 04:34:40.899  INFO stats: State [0: applying parts]  30/25/40 peers ⬇ 2.2MiB/s ⬆ 50.2kiB/s 0.00 bps 0 gas/s CPU: 182%, Mem: 4.6 GiB
Sep 15 04:34:50.901  INFO stats: State [0: applying parts]  30/26/40 peers ⬇ 1.3MiB/s ⬆ 46.0kiB/s 0.00 bps 0 gas/s CPU: 174%, Mem: 4.6 GiB
Sep 15 04:35:00.903  INFO stats: State [0: applying parts]  30/26/40 peers ⬇ 829.7kiB/s ⬆ 42.1kiB/s 0.00 bps 0 gas/s CPU: 194%, Mem: 4.6 GiB
Sep 15 04:35:10.906  INFO stats: State [0: applying parts]  30/26/40 peers ⬇ 646.7kiB/s ⬆ 37.8kiB/s 0.00 bps 0 gas/s CPU: 217%, Mem: 4.6 GiB
Sep 15 04:35:20.908  INFO stats: State [0: applying parts]  30/26/40 peers ⬇ 490.3kiB/s ⬆ 39.2kiB/s 0.00 bps 0 gas/s CPU: 229%, Mem: 4.6 GiB
Sep 15 04:35:30.911  INFO stats: State [0: download complete]  30/26/40 peers ⬇ 490.7kiB/s ⬆ 39.2kiB/s 0.00 bps 0 gas/s CPU: 260%, Mem: 4.6 GiB
Sep 15 04:35:40.913  INFO stats: #47749864 Downloading blocks 99.91% (34301)  30/26/40 peers ⬇ 542.1kiB/s ⬆ 38.5kiB/s 0.20 bps 14.34 Tgas/s CPU: 170%, Mem: 4.6 GiB
Sep 15 04:35:50.982  INFO stats: #47749890 Downloading blocks 99.91% (34285)  30/26/40 peers ⬇ 573.0kiB/s ⬆ 43.5kiB/s 2.58 bps 139.02 Tgas/s CPU: 97%, Mem: 4.6 GiB
Sep 15 04:36:00.983  INFO stats: #47749915 Downloading blocks 99.91% (34268)  30/26/40 peers ⬇ 607.4kiB/s ⬆ 50.8kiB/s 2.50 bps 146.11 Tgas/s CPU: 39%, Mem: 4.6 GiB
Sep 15 04:36:10.986  INFO stats: #47749941 Downloading blocks 99.91% (34252)  30/26/40 peers ⬇ 641.2kiB/s ⬆ 62.4kiB/s 2.60 bps 133.44 Tgas/s CPU: 41%, Mem: 4.6 GiB`

chain/chain/src/chain.rs
EgorKulikov: But error handling for this and for the rest of this function is the same, this would lead to duplicate code

nearcore/src/runtime/mod.rs
EgorKulikov: But we already expose store through genesis_state method (which returns (store, state_roots))

chain/client/src/client_actor.rs
EgorKulikov: Among other things quite a few tests call some intermediate methods without benefit of having ClientActor

chain/client/src/client_actor.rs
EgorKulikov: Yep, I think any renaming should be done when this actor will acquire new functions

nearcore/src/runtime/mod.rs
bowenwang1996: We should probably fix that instead of taking it for granted

EgorKulikov: The method we use from runtime would not make sense if runtime do not have access to state

bowenwang1996 (approved):
I'll approve to unblock, given that there does not seem to be concerns regarding correctness, but I suggest we also get a review from @mzhangmzz and @matklad

mzhangmzz (approved):

PR #4716, opened 2021-08-19
refactor: introduce `ViewConfig` for storing `max_gas_burnt_view` 

Longarithm:
One step towards https://github.com/near/nearcore/issues/4649.

`max_gas_burnt_view` should not be a part of `RuntimeConfig`, see https://github.com/near/nearcore/issues/3080 for more details. While it is the case, we have to modify `RuntimeConfig` inside runtime in a hacky way.

Here we introduce `ViewConfig` to store view parameters and remove usages of `max_gas_burnt_view` from `RuntimeConfig`. In the following protocol upgrade, this field will be removed.

Test plan
---------
`python3 tests/sanity/rpc_max_gas_burnt.py`
Nayduck run: http://nayduck.eastus.cloudapp.azure.com:3000/#/run/2013


runtime/near-vm-logic/src/logic.rs
```diff
@@ -115,8 +113,11 @@ impl<'a> VMLogic<'a> {
         // Overflow should be checked before calling VMLogic.
         let current_account_balance = context.account_balance + context.attached_deposit;
         let current_storage_usage = context.storage_usage;
-        let max_gas_burnt = if context.is_view {
-            config.limit_config.max_gas_burnt_view
+        let max_gas_burnt = if context.view_config.is_view {
+            match context.view_config.max_gas_burnt_view {
+                Some(max_gas_burnt_view) => max_gas_burnt_view,
+                None => config.limit_config.max_gas_burnt,
```
matklad, line 119: Hm, am I correct that this is essentially dead code? Every time we view_config.is_view is true, view_cconfig.max_gas_burnt is Some?

If that's the case, I suggest making this code staticall unreachable, as follows: 

```
impl VmContext {
  view_config: Option<ViewConfig>
}

struct ViewConfig {
  max_gas_burnt: gas
}
```

than, we'll be passing `Option<VIewConfig>` everywhere, with the semantics that we are in a view if the config is Some.

core/primitives-core/src/config.rs
```diff
@@ -27,6 +27,7 @@ pub struct VMLimitConfig {
     /// Max amount of gas that can be used, excluding gas attached to promises.
     pub max_gas_burnt: Gas,
     /// Max burnt gas per view method.
+    /// TODO: remove this field because it does not affect the protocol
     pub max_gas_burnt_view: Gas,
```
matklad, line 31: How hard is it to fix this? I'd like to prioritize this follow up, if possible, as this is a significant chunk of debt.

runtime/runtime/src/state_viewer/mod.rs
```diff
@@ -228,9 +221,11 @@ impl TrieViewer {
         let function_call = FunctionCallAction {
             method_name: method_name.to_string(),
             args: args.to_vec(),
-            gas: config.wasm_config.limit_config.max_gas_burnt_view,
+            gas: self.max_gas_burnt_view,
             deposit: 0,
         };
+        let view_config =
+            ViewConfig { is_view: true, max_gas_burnt_view: Some(self.max_gas_burnt_view) };
```
matklad, line 228: This looks so much nicer now, thanks!

matklad (commented):
`

core/primitives-core/src/config.rs
matklad: ah, I didn't read the PR description properly, you *are* tackling this in the next change, thanks!

core/primitives-core/src/config.rs
```diff
@@ -165,6 +166,13 @@ impl Default for VMLimitConfig {
     }
 }
 
+/// Configuration of view methods execution, during which no costs should be charged.
+#[derive(Default, Clone, Serialize, Deserialize, Debug, Hash, PartialEq, Eq)]
+pub struct ViewConfig {
+    /// If specified, defines max burnt gas per view method.
+    pub max_gas_burnt_view: Gas,
```
matklad, line 173:
```suggestion
    pub max_gas_burnt: Gas,
```

"If sepecified" bit is wrong -- this is no longer an option. The `_vew` suffix is reduntant, we are in a `ViewConfig` (rule-of-thumb -- each part of a compound name should add info). The comment somewhat repeats the name,  and feels like it can be removed.

runtime/near-vm-logic/src/context.rs
```diff
@@ -53,8 +54,8 @@ pub struct VMContext {
     #[serde(with = "crate::serde_with::bytes_as_base58")]
     /// Initial seed for randomness
     pub random_seed: Vec<u8>,
-    /// Whether the execution should not charge any costs.
-    pub is_view: bool,
+    /// If present, defines that view execution is enabled and configures it.
```
matklad, line 57: That's a good comment -- it explains the meaning of the `None / Some`. Might make sense to unpack what view is: if non-None, we are in a view mode, which means that only read-only operations are allowed.

runtime/near-vm-logic/src/logic.rs
```diff
@@ -115,17 +113,16 @@ impl<'a> VMLogic<'a> {
...
+            Some(ViewConfig { max_gas_burnt_view }) => max_gas_burnt_view,
+            None => config.limit_config.max_gas_burnt,
         };
         let current_account_locked_balance = context.account_locked_balance;
         let gas_counter = GasCounter::new(
             config.ext_costs.clone(),
             max_gas_burnt,
             context.prepaid_gas,
-            context.is_view,
+            context.view_config.is_some(),
```
matklad, line 125: This can benefit a bit from giving a name to the expresson:

```
let is_view = context.view_config.is_some();
```

I guess, it's even better to define an `is_view` method, and use it here  and in every `if self.context.view_config.is_some()` check.

matklad (approved):

core/primitives-core/src/config.rs
```diff
@@ -27,6 +27,7 @@ pub struct VMLimitConfig {
     /// Max amount of gas that can be used, excluding gas attached to promises.
     pub max_gas_burnt: Gas,
     /// Max burnt gas per view method.
+    /// TODO: remove this field because it does not affect the protocol
```
bowenwang1996, line 30: Let's add the issue number

runtime/near-vm-logic/src/context.rs
```diff
@@ -53,9 +54,17 @@ pub struct VMContext {
     #[serde(with = "crate::serde_with::bytes_as_base58")]
     /// Initial seed for randomness
     pub random_seed: Vec<u8>,
-    /// Whether the execution should not charge any costs.
-    pub is_view: bool,
+    /// If non-None, means that execution is made in a view mode and defines its configuration.
```
bowenwang1996, line 57:
```suggestion
    /// If Some, it means that execution is made in a view mode and defines its configuration.
```

runtime/near-vm-logic/src/context.rs
```diff
@@ -53,9 +54,17 @@ pub struct VMContext {
...
+    /// View mode means that only read-only operations are allowed.
+    /// See https://nomicon.io/Proposals/0018-view-change-method.html for more details.
```
bowenwang1996, line 59: 👍

bowenwang1996 (approved):

PR #4723, opened 2021-08-23
feat(indexer): Extend Indexer Framework with logs on start

khorolets:
Add a couple of `info!` logs on the indexer side to see the starting process more clearly as it takes quite a while to start the indexer on `testnet` because of the big genesis file. Additional logs should help to understand that the process is ongoing and nothing has stuck.


chain/indexer/src/lib.rs
```diff
@@ -92,7 +95,11 @@ pub struct Indexer {
 impl Indexer {
     /// Initialize Indexer by configuring `nearcore`
     pub fn new(indexer_config: IndexerConfig) -> Self {
+        tracing::info!(target: INDEXER, "Load config from {:?}...", &indexer_config.home_dir);
+
         let near_config = nearcore::load_config(&indexer_config.home_dir);
+
+        tracing::info!(target: INDEXER, "Validate genesis...",);
         near_chain_configs::genesis_validate::validate_genesis(&near_config.genesis);
```
frol, line 103: We don't need to explicitly call `validate_genesis` since #4592

chain/indexer/src/lib.rs
```diff
@@ -92,7 +95,11 @@ pub struct Indexer {
 impl Indexer {
     /// Initialize Indexer by configuring `nearcore`
     pub fn new(indexer_config: IndexerConfig) -> Self {
+        tracing::info!(target: INDEXER, "Load config from {:?}...", &indexer_config.home_dir);
+
         let near_config = nearcore::load_config(&indexer_config.home_dir);
```
frol, line 100: `load_config` loads the whole nearcore "config" which is the `config.json` and `genesis.json` (don't even mention how the naming is misleading here). I would consider adding the logs to the `nearcore::load_config` function. Also, consider using `nearcore::load_config_without_genesis_records`.

frol (changes requested):

chain/indexer/src/lib.rs
khorolets: Why `nearcore::load_config_without_genesis_records` if we later use genesis records in indexers?

khorolets: OK, I'll consider but it will completely ruin the current way of handling genesis records on indexer for explorer side. I don't know if anyone else is storing genesis records on their indexers, so I assume it'll be a breaking change. Investigating at the moment will let you know.

chain/indexer/src/lib.rs
```diff
@@ -92,7 +95,11 @@ pub struct Indexer {
 impl Indexer {
     /// Initialize Indexer by configuring `nearcore`
     pub fn new(indexer_config: IndexerConfig) -> Self {
+        tracing::info!(target: INDEXER, "Load config from {:?}...", &indexer_config.home_dir);
```
matklad, line 98:
```suggestion
        tracing::info!(target: INDEXER, "Load config from {}...", indexer_config.home_dir.display());
```

matklad: If we use `...` to signify the process here, then I suggest to use spans instead (https://docs.rs/tracing/0.1.26/tracing/#spans) and adjust our reporting infra to display spans in a useful way (they may be hidden by default)

Issue #4724, opened 2021-08-23
Get rid of similar `VMContext` creations in tests


Longarithm:
We have a bunch of `VMContext` creations for testing, which are very similar. Some of fields are filled with random values, and currently it's not very clear which of them are important:

 VMContext 

 { current_account_id: CURRENT_ACCOUNT_ID.parse().unwrap(), signer_account_id: SIGNER_ACCOUNT_ID.parse().unwrap(), signer_account_pk: Vec::from(&SIGNER_ACCOUNT_PK[..]), predecessor_account_id: PREDECESSOR_ACCOUNT_ID.parse().unwrap(), input, block_index: 10, block_timestamp: 42, epoch_height: 1, account_balance: 2u128, account_locked_balance: 0, storage_usage: 12, attached_deposit: 2u128, prepaid_gas: 10_u64.pow(14), random_seed: vec![0, 1, 2], view_config: None, output_data_receivers: vec![], } 

It seems we can unite most of such creations and simplify our code.

bowenwang1996:
@Longarithm is this fixed?

Longarithm:
No, it's far in the backlog.

pmnoxx:
@Longarithm I think it's a good idea to refactor at the code. This data structure has already 17 fields, it would be more readable, if we should group code of those fields into logical components. 

I see we have here bunch of accounts:
```
    current_account_id: CURRENT_ACCOUNT_ID.parse().unwrap(),
    signer_account_id: SIGNER_ACCOUNT_ID.parse().unwrap(),
    signer_account_pk: Vec::from(&SIGNER_ACCOUNT_PK[..]),
    predecessor_account_id: PREDECESSOR_ACCOUNT_ID.parse().unwrap(),
```

Something related to blocks:
```
   block_index: 10,
    block_timestamp: 42,
    epoch_height: 1,
```

Balance
```
    account_balance: 2u128,
    account_locked_balance: 0,
```

This code could be more readable by organizing it better, and hiding some information behind structs. Though that may be not needed if it's just for testing. It's up to you.

Though I see we construct this a couple of times. Maybe adding a common constructor would be a good idea.
@Longarithm What do you think?

Longarithm:
@pmnoxx I agree on that!

We could create sub-structures like `AccountContext`, `ChainContext`, `BalanceContext` and then simplify `VMContext` creations by invoking just `AccountContext::test()`, `AccountContext::default()`, etc.

PR #4725, opened 2021-08-23
refactor: get rid of a single duplicate of `VMContext`

Longarithm:
Get rid of a single duplicate for `VMContext`.
Advance a little to solution of https://github.com/near/nearcore/issues/4724.
For other usages, it's still unclear how to get rid of duplicates properly.

Test plan
---------
`cargo check`

### DIFF
--- a/runtime/near-vm-logic/src/context.rs
+++ b/runtime/near-vm-logic/src/context.rs
@@ -68,3 +68,26 @@ impl VMContext {
         self.view_config.is_some()
     }
 }
+
+impl Default for VMContext {
+    fn default() -> Self {
+        VMContext {
+            current_account_id: "alice".parse().unwrap(),
+            signer_account_id: "bob".parse().unwrap(),
+            signer_account_pk: vec![0, 1, 2],
+            predecessor_account_id: "carol".parse().unwrap(),
+            input: vec![],
+            block_index: 1,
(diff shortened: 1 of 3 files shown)


runtime/near-vm-logic/src/context.rs
```diff
@@ -68,3 +68,26 @@ impl VMContext {
         self.view_config.is_some()
     }
 }
+
+impl Default for VMContext {
+    fn default() -> Self {
+        VMContext {
```
bowenwang1996, line 74: I am not sure whether implementing default is a good idea. It looks like this implementation is used solely for testing. In that case we should probably implement a standalone function `VMContext::test`

matklad: +1

In addition to what Bowen said, this is also the case where the duplication is superficial -- nothing breaks if the two usages would use different contracts. The problem with deduping here is that, if in the future it turns out that you *do* want to have different contexts, untangling that would be difficult. 

It's easier to notice duplication when to notice that some things are merged while they need to be separate. 

I think it's better not to extract this at all (but no strong opinion -- adding an `put fn default_test_context` looks fine by me!)

Longarithm: Okay, that seems to be bigger issue than dealing with different `VMContext`. Thank you for reviewing!

PR #4728, opened 2021-08-24
Add a metric for the average amount of Tgas burned in chunks

nikurt:
…t run mainnet validators, this metric can still be exported by the RPC nodes and can be used in our dashboard. This is a useful indicator of block fullness for loadtests.

### DIFF
--- a/chain/client/src/info.rs
+++ b/chain/client/src/info.rs
@@ -157,6 +157,8 @@ impl InfoHelper {
         set_gauge(&metrics::BLOCKS_PER_MINUTE, (avg_bls * (60 as f64)) as i64);
         set_gauge(&metrics::CPU_USAGE, cpu_usage as i64);
         set_gauge(&metrics::MEMORY_USAGE, (memory_usage * 1024) as i64);
+        let teragas = 1_000_000_000_000u64;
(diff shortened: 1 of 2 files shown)


chefsale (approved):
LGTM

chain/client/src/info.rs
```diff
@@ -157,6 +157,8 @@ impl InfoHelper {
         set_gauge(&metrics::BLOCKS_PER_MINUTE, (avg_bls * (60 as f64)) as i64);
         set_gauge(&metrics::CPU_USAGE, cpu_usage as i64);
         set_gauge(&metrics::MEMORY_USAGE, (memory_usage * 1024) as i64);
+        let teragas = 1_000_000_000_000u64;
+        set_gauge(&metrics::AVG_TGAS_USAGE, (avg_gas_used / teragas) as i64);
```
bowenwang1996, line 161: Should we round it to the nearest integer to be more accurate?

bowenwang1996 (approved):

chain/client/src/info.rs
nikurt: Done

PR #4729, opened 2021-08-24
feat: support cost tables in script for comparing costs 

Longarithm:
Should be useful to find significant differences in `CostTables` as we do for `RuntimeConfig`s.

Test plan
---------
Check that old run `python3 ./runtime/runtime-params-estimator/emu-cost/compare_costs.py ~/tmp/runtime_config_nochange.json ~/tmp/runtime_config_no_reuse.json --diff` don't fail

### DIFF
--- a/runtime/runtime-params-estimator/emu-cost/compare_costs.py
+++ b/runtime/runtime-params-estimator/emu-cost/compare_costs.py
@@ -41,12 +41,12 @@ def parse_debug_print(path):
                     prefix = m.group(1) + ": "
     return result
 
-def read_costs(path):
+def read_config(path):
     result = OrderedDict()
     try:
         with open(path) as f:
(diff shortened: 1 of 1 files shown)


matklad:
I think this is something we want to have, but I don't think we should modify python script for that. One problem with the estimator is that we have a load of scripts everywhere, and it's hard to learn how to use them. 

I suggest adding this functionality directly to the param estimator, such that 

```console
# compare with canonical costs txt
$ cargo run -- compare costs-file.txt

# compare two cost files direclty
$ cargo run -- compare costs-file1.txt costs-file2.txt
```

implementation wise, you'll need clap subcommand, and you'll need to define `struct CostTableDiff { ... } impl Display for CostTableDiff`.

Longarithm:
Closeing because @matklad planned to implement this in Rust inside param-estimator.

PR #4732, opened 2021-08-25
chore: Add MIT and Apache licenses and update crate licenses

austinabell:
I just updated the dependency crates used from the SDK

Open questions before opening PR:
- Does the copyright year and holder need to be filled out for these?
  - Note: it's not filled out in the gpl3 license
- Is it fine for the other license files to be in the root directory?
- Should the licenses be symlinked to each crate specifically that uses it?
  - If this is the case, should the licenses be moved out of root to not apply to whatever gpl3 needs to be applied to? I'm not sure what parts need to be gpl3


chefsale:
This looks reasonable to me. Let's just use little or instead of OR in the string.

chefsale (approved):

austinabell:
> This looks reasonable to me. Let's just use little or instead of OR in the string. Migrating from Apache to Apache or MIT will not create any new issue except the ones which already exists there will still be there, so this can only be an improvement. We just need to take a look as a follow up, is there and GPL based licence which we include in those crates if yes all the crates should be GPL licenced.

Every rust license declaration uses `OR` over `or`, for example Rust itself https://github.com/rust-lang/rust/blob/master/library/std/Cargo.toml#L4. As for checking if any GPL, why not just do this before this comes in, to avoid any versions being cut with an invalid license? Is there anything I can do to help with this? I'm a bit unfamiliar with this process

chefsale:
Feel free to land it with OR in that case, if it is a rust licence style. @bowenwang1996 and I will have to take a deeper look if in case of an violation, but we already use FOSSA for automated scanning of compliance issues like this. If anything pops out there it will be flagged. I will run the scan on this PR and post the results

chefsale:
The scan results can be found here for this PR once done: https://app.fossa.com/projects/custom+21311%2Fgit@github.com:near%2Fnearcore.git/refs/branch/austin%2Flicenses/dd11010a22c65447890a5673653b8811ffa7b2c7

This is for the whole nearcore repo, some of them are GPL/LGPL issues on master  which aren't resolved. @bowenwang1996 and @matklad will probably be able to say more about specific packages and dependencies.

chefsale:
Update on the previous comment once the scan has finished, the dependencies which are in a violation are:
- gnuplot
- librocksdb-sys
- jemalloc-sys
- openssl-sys

If none of these are used in these crates transitively, we are good to proceed.

matklad:
jemalloc is mit/apache: https://github.com/gnzlbg/jemallocator/blob/master/jemalloc-sys/Cargo.toml#L10

bowenwang1996:
> If none of these are used in these crates transitively, we are good to proceed.

@austinabell please confirm this and move this PR forward

austinabell:
> Update on the previous comment once the scan has finished, the dependencies which are in a violation are:
> 
> * gnuplot
> * librocksdb-sys
> * jemalloc-sys
> * openssl-sys
> 
> If none of these are used in these crates transitively, we are good to proceed.

As well as `jemalloc-sys` being MIT/apache2 as Aleksey points out, [openssl-sys](https://github.com/sfackler/rust-openssl/blob/master/openssl-sys/Cargo.toml#L6) and [librocksdb-sys](https://github.com/rust-rocksdb/rust-rocksdb/blob/master/librocksdb-sys/Cargo.toml#L6) (although includes bsd3, is this an issue?) are not gpl3, so fine to just ensure that gnuplot is not used with any of these?

matklad (approved):
not a lawyer, but LGTM!

After this PR, it'd be cool to add a CI check that we don't introduce gpl dependencies inadvertently. Here's how this check looks in rust-analyzer: 

https://github.com/rust-analyzer/rust-analyzer/blob/master/crates/rust-analyzer/tests/slow-tests/tidy.rs#L213-L267

chefsale:
@matklad we already use FOSSA and there could be such a policy implemented there once we have resolved all the current compliance issues.

frol:
We already use [`cargo-deny`](https://github.com/EmbarkStudios/cargo-deny) to check duplicate dependencies, and it also can check the licenses

chefsale:
> We already use [`cargo-deny`](https://github.com/EmbarkStudios/cargo-deny) to check duplicate dependencies, and it also can check the licenses
> 
> https://github.com/near/nearcore/blob/e5315e520b87662e58ddb08cfd65524946f6e61f/.buildkite/pipeline.yml#L22-L28

Even better, didn't know this is in the CI :P

frol (approved):
I am not a lawyer either, it looks good to me

bowenwang1996:
> I am not a lawyer

Is this the new pick up phrase these days :)

PR #4733, opened 2021-08-25
make it easier to understand how each cost is computed

matklad:
Ready to review!

This is a huge change which overhauls how parameter estimator works. Previously, we were running a bunch of estimations in order, which created interference between estimations (as they sometimes were re-using the same state), made it hard to run only subset of estimations (existing `--metrics-to-measure` flag partially alleviates the problem), and forced using a specific way to estimate all the costs (ie, if we wanted to use something else than "run a block with transactions" for estimation, that required ad-hoc code).

In the new setup, each cost is estimated by a single rust function. That function can do whatever, but there are helpers for common estimation methods, like the "run a block of transactions"  one. So, the end result looks like this: 

```rust
fn action_stake(ctx: &mut Ctx) -> GasCost {
    let total_cost = {
        let test_bed = ctx.test_bed();

        let mut make_transaction = |tb: &mut TransactionBuilder| -> SignedTransaction {
            let sender = tb.random_unused_account();
            let receiver = sender.clone();

            let actions = vec![Action::Stake(StakeAction {
                stake: 1,
                public_key: "22skMptHjFWNyuEWY22ftn2AbLPSYpmYwGJRGwpNHbTV".parse().unwrap(),
            })];
            tb.transaction_from_actions(sender, receiver, actions)
        };
        transaction_cost(test_bed, &mut make_transaction)
    };

    let base_cost = action_sir_receipt_creation(ctx);

    total_cost - base_cost
}
```

Notable changes in behavior: 

* we now prepare testbed with estimator contracts deployed. Previously, we prepared testbed with some trivial contract, and had a separate logic to populate it with estimator contracts. 
* we no-longer use `mean + 4*stddev` as the estimation. Instead, we just use `mean` but flag the cost as suspicious if we observe deviations of more than 15% during estimation. The two problems with stddev is that:
  *  it gives you *some* number instead of saying "hey, your measures are highly inconsistent and I can't give a sensible estimate"
  * it's not entirely statistically sound, as we are calculating stdev over blocks, and block already aggregates many transactions. 

But overall I've tried to preserve existing behavior as much as possible. I don't have a guarantee that I didn't make some stupid mistake for every specific cost, but I generally cross-referenced new results with old estimator and with current costs, and didn't find anything super weird except for #4836. After this is merged, I want to go through each cost and make sure it actually gives a reasonable, reproducible result which accounts for IO costs and matches our currently reployed cost table.


bowenwang1996:
@matklad we know that costs have drifted from what we have on mainnet. Maybe it makes sense to focus on resolving old vs. new first?

runtime/runtime-params-estimator/src/cost_table.rs
```diff
@@ -50,15 +50,15 @@ impl fmt::Display for CostTable {
...
+                let gas = format_gas(gas);
                 writeln!(f, "{:<35} {:>25}", cost.to_string(), gas)?
             }
         }
         Ok(())
     }
 }
 
-fn separate_thousands(mut n: u64) -> String {
+pub(crate) fn format_gas(mut n: Gas) -> String {
```
olonho, line 61: Seems worth putting somewhere in higher upstream crates.

runtime/runtime-params-estimator/src/estimator_context.rs
```diff
@@ -0,0 +1,90 @@
...
+    pub(crate) compile_cost_base_per_byte: Option<(GasCost, GasCost)>,
+}
+
+impl<'c> EstimatorContext<'c> {
+    pub(crate) fn new(config: &'c Config) -> Self {
+        let cached = CachedCosts::default();
+        Self { cached, config }
+    }
+
+    pub(crate) fn test_bed(&mut self) -> TestBed<'_> {
```
olonho, line 36: Isn't it a single word, i.e. `testbed`?

matklad: Good catch!

olonho (approved):

matklad:
Discussed with olonho and figured out that the best way forward is to merge this. 

It would be *nice* if we can demonstate that the behavior is the same as with the old estimator, but we know that the behavior of the old estimator is not always explainable, and not always correct. What we actually care about is not the diff with the old estimator, but

* the diff with mainnet costs
* the diff with reality

The old estimator isn't great across this dimetions. The new estimator isnt' better, but it is at least possible to debug each cost in isolation.

PR #4748, opened 2021-08-27
high level changes for how split state will be triggered in sync and catch up

mzhangmzz:
This PR implements the high level logic of how state splitting will be triggered in sync and catch up. The specific logic of splitting states will be added in a following PR.

Test Plan:
- no existing test breaks
- I added an integration test in rust to test for the case when shard layout changes. Because the logic of splitting states is not implemented yet, the test is not complete. I simply verified the logs to make sure the added logic in this PR is triggered correctly.


chain/client/src/client.rs
```diff
@@ -1576,10 +1576,44 @@ impl Client {
             assert_eq!(sync_hash, state_sync_info.epoch_tail_hash);
             let network_adapter1 = self.network_adapter.clone();
 
+            let new_shard_sync = {
+                let prev_hash =
+                    self.chain.get_block(&sync_hash).unwrap().header().prev_hash().clone();
```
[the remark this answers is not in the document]
mzhangmzz: I think here is very safe to assume the no error should occur, but I get your point. I will use error propagation as much as possible in the future.

chain/client/src/sync.rs
```diff
@@ -768,11 +771,24 @@ impl StateSync {
...
+                        *shard_sync_download = ShardSyncDownload {
+                            downloads: vec![],
+                            status: ShardSyncStatus::StateSplit,
+                        }
+                    } else {
+                        this_done = true;
+                    }
+                }
+                ShardSyncStatus::StateSplit => {
+                    assert!(split_states);
```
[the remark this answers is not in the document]
mzhangmzz: It is a local check that checks the implementation of the code that triggers StateSplit is correct. I'll change it to debug_assert so that this assert will not be triggered at release code

chain/epoch_manager/src/lib.rs
```diff
@@ -747,13 +747,29 @@ impl EpochManager {
...
         &mut self,
         parent_hash: &CryptoHash,
         account_id: &AccountId,
         shard_id: ShardId,
     ) -> Result<bool, EpochError> {
         let next_epoch_id = self.get_next_epoch_id_from_prev_block(parent_hash)?;
+        if let Some(split_shards) =
+            self.get_split_shards_if_shards_will_change(parent_hash, vec![shard_id])?
+        {
+            for next_shard_id in split_shards.get(&shard_id).unwrap() {
```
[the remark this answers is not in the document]
mzhangmzz: removed

chain/chain/src/chain.rs
```diff
@@ -1013,6 +1013,18 @@ impl Chain {
...
+            // Currently, the state sync algorithm assumes that the number of chunks do not change
+            // between the epoch being synced to and the last epoch.
+            // Fortunately, since all validators track all shards for now, this error will not be
+            // triggered in live yet
+            let str = format!(
+                "Cannot download states for epoch {:?} because sharding just changed. I'm {:?}",
+                block.header().epoch_id(),
+                me
+            );
+            return Err(Error::from(ErrorKind::Other(str)));
```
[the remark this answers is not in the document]
mzhangmzz: I thought about it more. Instead of erroring here, I decided to simply log an error message. This way, the failure of this method will not affect the processing the the block for this epoch. So if this error is ever triggered, the validator will keep processing blocks until the end of this epoch and either halt or fail when the next epoch starts and it doesn't have the states ready for that.

chain/chain/src/chain.rs
```diff
@@ -1013,6 +1013,23 @@ impl Chain {
...
+            // triggered in live yet
+            // Instead of propagating the error, we simply log the error here because the error
+            // do not affect processing blocks for this epoch. However, when the next epoch comes,
+            // the validator will not have the states ready so it will halt.
+            error!(
+                "Cannot download states for epoch {:?} because sharding just changed. I'm {:?}",
+                block.header().epoch_id(),
+                me
+            );
+            debug_assert!(false);
```
bowenwang1996, line 1031: Why do we need this debug assert?

chain/chain/src/chain.rs
```diff
@@ -1175,29 +1192,41 @@ impl Chain {
...
+    ///    in the current epoch that will be split into a future shard that `me` will track.
+    /// 2) Shard layout will be the same. In this case, the method returns all shards that `me` will
+    ///    track in the next epoch but not this epoch
     fn get_shards_to_dl_state(
         &self,
         me: &Option<AccountId>,
         parent_hash: &CryptoHash,
     ) -> Vec<ShardId> {
-        let epoch_id = self.runtime_adapter().get_epoch_id_from_prev_block(parent_hash).unwrap();
+        let epoch_id = self.runtime_adapter.get_epoch_id_from_prev_block(parent_hash).unwrap();
```
bowenwang1996, line 1206: is it safe to unwrap here?

chain/chain/src/chain.rs
```diff
@@ -1175,29 +1192,41 @@ impl Chain {
...
     }
 
+    fn should_catch_up_shard(
+        runtime_adapter: Arc<dyn RuntimeAdapter>,
+        me: &Option<AccountId>,
+        parent_hash: &CryptoHash,
+        shard_id: ShardId,
+    ) -> bool {
+        let will_shard_layout_change =
+            runtime_adapter.will_shard_layout_change(parent_hash).unwrap();
```
bowenwang1996, line 1221: Same here

chain/client/src/sync.rs
```diff
@@ -674,6 +674,9 @@ impl StateSync {
             status: ShardSyncStatus::StateDownloadHeader,
         };
 
+        let prev_hash = chain.get_block_header(&sync_hash)?.prev_hash().clone();
+        let split_states = runtime_adapter.will_shard_layout_change(&prev_hash).unwrap();
```
bowenwang1996, line 678: Same here

chain/epoch_manager/src/lib.rs
```diff
@@ -354,7 +354,7 @@ impl EpochManager {
         block_info: &BlockInfo,
         last_block_hash: &CryptoHash,
         rng_seed: RngSeed,
-    ) -> Result<EpochId, EpochError> {
+    ) -> Result<(), EpochError> {
```
bowenwang1996, line 357: What is this change for?

chain/epoch_manager/src/lib.rs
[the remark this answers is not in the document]
bowenwang1996: it is not removed :(

chain/epoch_manager/src/lib.rs
```diff
@@ -1201,6 +1220,34 @@ impl EpochManager {
...
+            let split_shards: Result<Vec<Vec<ShardId>>, String> = shards
+                .iter()
+                .map(|shard_id| {
+                    next_shard_layout.get_split_shards(*shard_id).cloned().ok_or(format!(
+                        "cannot find parent shard for shard {} in layout {:?}",
+                        *shard_id, next_shard_layout
+                    ))
+                })
+                .collect();
+            let split_shards = split_shards.map_err(|str| EpochError::ShardingError(str))?;
```
bowenwang1996, line 1244: It is better to use a variable name does not coincide with a rust keyword

chain/epoch_manager/src/lib.rs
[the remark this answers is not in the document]
mzhangmzz: Oh sorry, I replied you incorrectly. This unwrap is actually guaranteed to succeed because the returned dictionary will always have `shard_id` in there

chain/epoch_manager/src/lib.rs
mzhangmzz: I don't see the return value used anywhere

chain/epoch_manager/src/lib.rs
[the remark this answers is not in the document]
bowenwang1996: ah I see. Let's add a comment here

chain/chain/src/chain.rs
mzhangmzz: yeah technically the error will never be triggered because this function is called after `chain_update.process_block` returns `Ok` and `process_block` checked that epoch exists

mzhangmzz: I changed the code so that this function propagates the error though

chain/chain/src/chain.rs
mzhangmzz: Technically no because of the same reason I mentioned above, but I changed the code so that it doesn't just unwrap

chain/chain/src/chain.rs
mzhangmzz: I wanted this to fail in tests to ensure that this is not triggered without our attention.

core/primitives/src/shard_layout.rs
```diff
@@ -34,9 +34,11 @@ pub struct ShardLayoutV1 {
     fixed_shards: Vec<AccountId>,
     /// The rest are divided by boundary_accounts to ranges, each range is mapped to a shard
     boundary_accounts: Vec<AccountId>,
-    /// Parent shards for the shards, useful for constructing states for the shards.
+    /// Maps shards from the last shard layout to shards that it splits to in this shard layout,
+    /// Instead of using map, we just use a vec here because shard_id ranges from 0 to num_shards-1
+    /// Useful for constructing states for the shards.
     /// None for the genesis shard layout
-    parent_shards: Option<Vec<ShardId>>,
+    shards_split_map: Option<Vec<Vec<ShardId>>>,
```
frol, line 41: Consider having a separate type since `Vec<Vec<>>` looks suspicious, and I had to read the comment to understand the data layout (the meaning of the indices)

core/primitives/src/errors.rs
```diff
@@ -749,6 +749,8 @@ pub enum EpochError {
     IOErr(String),
     /// Given account ID is not a validator in the given epoch ID.
     NotAValidator(AccountId, EpochId),
+    /// Error getting information for a shard
+    ShardingError(String),
```
frol, line 753: Should this be hidden behind a feature-flag and protocol upgrade?

nearcore/src/runtime/mod.rs
```diff
@@ -1646,6 +1651,14 @@ impl RuntimeAdapter for NightshadeRuntime {
             epoch_manager.get_prev_epoch_id(prev_block_hash).map_err(Error::from)
         }
     }
+
+    fn will_shard_layout_change(&self, parent_hash: &CryptoHash) -> Result<bool, Error> {
+        let mut epoch_manager = self.epoch_manager.as_ref().write().expect(POISONED_LOCK_ERR);
+        Ok(epoch_manager
+            .get_split_shards_if_shards_will_change(parent_hash, vec![])
+            .map_err(Error::from)?
```
frol, line 1659: This should be enough since `?` operator uses Into trait anyway:

```suggestion
            .get_split_shards_if_shards_will_change(parent_hash, vec![])?
```

frol (approved):

core/primitives/src/errors.rs
mzhangmzz: Good idea. However, I think it is not necessary to put it in a feature-flag for the following reasons.
- I think `EpochError` is not used on the protocol level, they are all converted to `chain::Error` at some point (@bowenwang1996 is that true?). 
- The logic that triggers this Error also belongs to internal implementation details instead of protocol changes, so I don't want to put that logic in a feature flag
- The error will not be triggered right now with our current implementation because Simple Nightshade is enabled.

bowenwang1996 (approved):

Issue #4738, opened 2021-08-26
Make ContractCode getters actually useful


matklad:
Change this:

https://github.com/near/nearcore/blob/dabf861c38dcc1c11c2e31bf898b0ca08de95e15/core/primitives-core/src/contract.rs#L3-L21

to this:

```rust
pub struct ContractCode {
    code: Vec<u8>,
    hash: CryptoHash,
}

impl ContractCode {
    pub fn new(code: Vec<u8>, hash: Option<CryptoHash>) -> ContractCode {
        let hash = hash.unwrap_or_else(|| sha256(&code));
        ContractCode { code, hash }
    }

    pub fn code(&self) -> &[u8] {
        self.code.as_slice()
    }

    pub fn hash(&self) -> &CryptoHash {
        &self.hash
    }
}
```

* `get_` prefixes are not idiomatic
* having a getter for a public field doesn't make sense 
* ContratCode has an invariant that hash and code match, so the fields should be private

While we are at it, might add a debug assertion that hash passed in in constructor is the right one.

PR #4749, opened 2021-08-29
refactor(core, primitives): ContractCode getters

shekhirin:
Resolves https://github.com/near/nearcore/issues/4738


core/primitives/src/views.rs
```diff
@@ -127,13 +127,13 @@ impl From<AccountView> for Account {
 
 impl From<ContractCode> for ContractCodeView {
     fn from(contract_code: ContractCode) -> Self {
-        ContractCodeView { code: contract_code.code, hash: contract_code.hash }
+        ContractCodeView { code: contract_code.code().to_vec(), hash: contract_code.hash().clone() }
```
bowenwang1996, line 130: looks like there is some unnecessary clone given that we take ownership of contract_code here.

shekhirin: hmm in fact I think all the `contract_code.hash().clone()` calls could be replaced with `*contract_code.hash()` since `CryptoHash` implements the `Copy` trait. Am I right or you'd prefer explicit cloning?

ailisp: I think since it's Copy, `*contract_code.hash()` is fine since don't clone. Bowen probably means contract_code is `ContractCode` instead of `&ContractCode` so that there should not need a clone.

ailisp (approved):

core/primitives/src/views.rs
shekhirin: got it, fixed

core/primitives/src/views.rs
```diff
@@ -127,13 +127,13 @@ impl From<AccountView> for Account {
 
 impl From<ContractCode> for ContractCodeView {
     fn from(contract_code: ContractCode) -> Self {
-        ContractCodeView { code: contract_code.code, hash: contract_code.hash }
+        ContractCodeView { code: contract_code.code().to_vec(), hash: *contract_code.hash() }
```
matklad, line 130: Yeah, this `.to_vec()` is unfortunate, let's add `impl ContractCode { pub fn into_code(self) -> Vec<u8> { self.code } }` to avoid it.

shekhirin: then here we'll have
```rust
fn from(contract_code: ContractCode) -> Self {
    let hash = *contract_code.hash();
    ContractCodeView { code: contract_code.into_code(), hash }
}
```
and in `core/store/src/lib.rs`
``` rust
pub fn set_code(state_update: &mut TrieUpdate, account_id: AccountId, code: &ContractCode) {
    state_update.set(TrieKey::ContractCode { account_id }, code.clone().into_code());
}
```

which doesn't look good imo.

matklad: I'd write the first as

```rust
    let hash = *contract_code.hash();
    let code = contract_code.into_code()
    ContractCodeView { code, hash }
```

That's a bit boilerplate-y, but boilerplate is expected in this layer. What we are doing here is converting from our internal, private types to public JSON representation exposed via HTTP API. Such code just always look like "lets convert a struct to exactly the same struct", but it provides great value -- you can change private stuff without affecting the API. That it does look boilerplaty doesn't meant that we should drop perf on the floor ;)

For the second case, you shouldn't `code.clone().into_code()` -- this first clones the whole `ContractCode`, and then extracts the vector. It's more minimal to clone just the vector with `codo.code().to_vec()`. Although perhaps the function can be changed to take `code` by value? (if that's a big change, better to keep it out of this PR)

shekhirin: got it, fixed

runtime/near-vm-runner/src/cache.rs
```diff
@@ -56,7 +56,7 @@ pub fn get_contract_cache_key(
 ) -> CryptoHash {
     let _span = tracing::debug_span!(target: "vm", "get_key").entered();
     let key = ContractCacheKey::Version3 {
-        code_hash: code.hash,
+        code_hash: code.hash().clone(),
```
matklad, line 59: Crypto hash is copy, so here and in all other places it's better to spell `*code.hash()`. This makes it clear to the reader that no expensive allocationg clone is happening.

matklad (approved):
It's better to change `.hash().clone()` to `*.hash()`, but this already is an improvement, so I won't block on that.

runtime/near-vm-runner/src/cache.rs
shekhirin: fixed!

frol (approved):
@shekhirin Thanks for the code quality improvements! :heart:

bowenwang1996 (approved):

matklad (approved):

Issue #4550, opened 2021-07-22
Run runtime fuzzer on CI


matklad:
in #4524 we introduced the basic fuzzing infra. We should actually run in on CI to pursue two goals:

* prevent the infra itself from bitrotting 
* actually capturing regression with fuzzing

PR #4755, opened 2021-08-31
pytest: fuzz: enable parallelism

mina86:
Issue: https://github.com/near/nearcore/issues/4550

### DIFF
--- a/pytest/tests/runtime/fuzz.py
+++ b/pytest/tests/runtime/fuzz.py
@@ -6,11 +6,13 @@ import sys
 def run_fuzz():
     env = os.environ.copy()
     env["RUSTC_BOOTSTRAP"] = "1"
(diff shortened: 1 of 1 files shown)


pytest/tests/runtime/fuzz.py
```diff
@@ -6,7 +6,9 @@
 def run_fuzz():
     env = os.environ.copy()
     env["RUSTC_BOOTSTRAP"] = "1"
-    fuzzing = subprocess.Popen(('cargo', 'fuzz', 'run', 'runtime-fuzzer',
+    cpus = len(os.sched_getaffinity(0))
+    fuzzing = subprocess.Popen(('cargo', 'fuzz', 'run', '--jobs', cpus,
```
posvyatokum, line 10: Are you sure that passing cpus as integer here would work? It fails on gcp machine

posvyatokum (approved):
Thanks for the improvement!

PR #4756, opened 2021-08-31
pytest: remove unnecessary companion.py script

mina86:
The tests/companion.py script is not used for anything (and as far as
I can tell it was never really used).  Delete it.


chefsale (approved):
LGTM

PR #4759, opened 2021-08-31
pytest: fuzz: don’t capture output

mina86:
Standard error output of the fuzzing is passed through directly
and standard output generated by the `fuzz.py` test only duplicates
information already present in standard error.  And the data can get
quite big† so it’s a waste to include it in both streams.

Stop capturing output of the `cargo fuzz` command and instead
convert the `fuzz.py` test to simply exec the command setting up
the environment beforehand.

† For example http://nayduck.eastus.cloudapp.azure.com:3000/#/run/2047

Issue: https://github.com/near/nearcore/issues/4550

### DIFF
--- a/pytest/tests/runtime/fuzz.py
+++ b/pytest/tests/runtime/fuzz.py
@@ -1,33 +1,16 @@
-import subprocess
 import os
-import re
+import shutil
 import sys
-
-def run_fuzz():
-    env = os.environ.copy()
-    env["RUSTC_BOOTSTRAP"] = "1"
-    cpus = len(os.sched_getaffinity(0))
-    fuzzing = subprocess.Popen(('cargo', 'fuzz', 'run', '--jobs', str(cpus),
-                                'runtime-fuzzer', '--', '-len_control=0'
-                                '-prefer_small=0', '-max_len=4000000'),
-                               env=env, cwd='../test-utils/runtime-tester/fuzz',
-                               stderr=subprocess.PIPE)
-    _, err = fuzzing.communicate()
-
-    sys.stderr.buffer.write(err)
-
-    re_expr = re.compile(rb'Output of `std::fmt::Debug`:(.*)Reproduce', re.DOTALL)
-    res = re_expr.search(err)
-
-    if res:
-        sys.stdout.buffer.write(res.group(1))
-
-    if fuzzing.returncode != 0:
-        sys.exit(f'Invalid result: { fuzzing.returncode }')
 
 
 def main():
-    run_fuzz()
+    cpus = len(os.sched_getaffinity(0))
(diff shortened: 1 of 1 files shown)


posvyatokum:
The point of scenario being written in stdout is that you can download stdout and reproduce the fail locally. We can get rid of full stderr, but we need the scenario part, otherwise fuzzing is useless.

posvyatokum:
We can also reduce output size in a different way (while still keeping some failing scenario in stdout) #4760

mina86:
The scenario is included in stderr which you can also download and it looks like the scenario constitutes vast majority of the output. From the example given in the commit message, stdout and stderr are the same size (up to a megabyte) so excluding stderr but outputting the scenario to stdout doesn’t save any space. Furthermore, note that I’m working on getting the crash files saved along other log files.

posvyatokum:
I inspected the output and it is not what it's supposed to be. The are multiple scenarios as well as some extra stderr output. 
It probably is the result of parallelism. 
stdout is supposed to be just one scenario json and nothing else.
If it were it would be around 50mb

posvyatokum:
Also I suspect that all of fails in the example run were due to slow block production and they don't reproduce locally (I've tried several). Can this slowdown be the result of the parallel execution?
If so, we have one more reason to revert the parallelism commit (and this is the big one from the fuzzing perspective).

mina86:
Even still, this does not change the fact that there’s no reason to include the stderr output in stdout. In other runs, e.g. <http://nayduck.eastus.cloudapp.azure.com:3000/#/test/217189>, the size of stderr and stdout is the same as well.

posvyatokum:
1. If we see a slowdown with parallelism, then we should revert that commit.
2. We don't need stderr as a whole, we just need the scenario part of it (which is a majority of stderr). We need scenario, so we can reproduce the bug, just knowing that fuzzing fails on specific commit wouldn't help us. We can obtain scenario not by parsing stderr, but by running one extra command, but it seems needles at this point. Anyway, we need output with which we can get full scenario json.

So, we can reduce memory in half just removing `sys.stderr.buffer.write(err)`.
And we can reduce memory 10 times by returning to 1 job.

We need to do at least both if these steps (and maybe something extra), because err will always contain full scenarios of all failed jobs, and because we want to also catch core slowdown in fuzzing (and we so much false positives it becomes impossible).

Additionally, I can add one extra call to output not scenario json, but unstructured input, from which this scenario was created. This will be far smaller, but also it will become harder to reproduce errors.

mina86:
Parallelism is unrelated issue. If it is an issue we should revert it.

Right, so if we need majority of stderr why do we bother doing the capture and matching? Just let the command run and NayDuck stores the stderr which has all the necessary information. The test script is simpler that way.

posvyatokum (approved):

Issue #4710, opened 2021-08-18
Pass `shard_layout` to garbage collection 


mzhangmzz:
After https://github.com/near/nearcore/pull/4659, Trie nodes are stored in the database with `shard_version` as part of their keys. However, the pull request didn't finish changing garbage collection to make it take the correct `shard_version` to garbage collect. It is left as a TODO there. 

This issue needs to be fixed before sharding is enabled, otherwise gc will not work correctly

PR #4762, opened 2021-08-31
make chunk_extra referenced by shard_uid instead of shard_id

mzhangmzz:
To support re-sharding, at the epoch where splitting states happens, the validators need to keep chunk extra for the shards for this epoch and the shards in the coming epoch. This PR changes chunk_extra to be stored and referenced by `shard_uid` instead of `shard_id`


chain/chain/src/store.rs
```diff
@@ -2077,7 +2077,11 @@ impl<'a> ChainStoreUpdate<'a> {
             self.gc_col(ColIncomingReceipts, &block_shard_id);
             self.gc_col(ColChunkPerHeightShard, &block_shard_id);
             self.gc_col(ColNextBlockWithNewChunk, &block_shard_id);
-            self.gc_col(ColChunkExtra, &block_shard_id);
+
+            // TODO: use the real shard version
```
bowenwang1996, line 2081: is there an issue for this?

bowenwang1996 (approved):

chain/chain/src/store.rs
mzhangmzz: What do you mean?

I created an issue for this TODO https://github.com/near/nearcore/issues/4710. I will implement that in a later PR.

One issue is that state validator will fail if these columns are not garbage collected correctly. One way to hack that is to garbage collect both `shard_version` 0 and 1. The most correct way is pass the correct shard layout version to the gc function.

bowenwang1996: Could we add the issue number here?

chefsale (approved):

ailisp (approved):
Looks good to me!

bowenwang1996:
@nikurt @Longarithm please review

nikurt (approved):

Longarithm (approved):

PR #4763, opened 2021-09-01
Fix(test): gas usage changed due to wat parse slightly different

ailisp:
After cargo update, one test (tests::error_cases::test_external_call_indirect) failed of gas usage difference. The root cause is wat 1.0.38 and 1.0.39 parse the wast form of the test snippet slightly different. I checked when given the same wasm bytes, gas usage is same before and after cargo update. So the gas usage changed is really due to the test itself changed and not a vulnerability of contract-runtime.

Test Plan
---------
CI

Notes on how exactly the gas changed: wat 1.0.39 parse the test wast snippet to:
```
[0, 97, 115, 109, 1, 0, 0, 0, 1, 8, 2, 96, 0, 1, 126, 96, 0, 0, 2, 19, 1, 3, 101, 110, 118, 11, 112, 114, 101, 112, 97, 105, 100, 95, 103, 97, 115, 0, 0, 3, 2, 1, 1, 4, 4, 1, 112, 0, 1, 7, 8, 1, 4, 109, 97, 105, 110, 0, 1, 9, 7, 1, 0, 65, 0, 11, 1, 0, 10, 10, 1, 8, 0, 65, 0, 17, 0, 0, 26, 11, 0, 23, 4, 110, 97, 109, 101, 1, 6, 1, 0, 3, 108, 111, 108, 4, 8, 1, 0, 5, 108, 111, 108, 95, 116]
```

While wat 1.0.38:
```
[0, 97, 115, 109, 1, 0, 0, 0, 1, 8, 2, 96, 0, 1, 126, 96, 0, 0, 2, 19, 1, 3, 101, 110, 118, 11, 112, 114, 101, 112, 97, 105, 100, 95, 103, 97, 115, 0, 0, 3, 2, 1, 1, 4, 4, 1, 112, 0, 1, 7, 8, 1, 4, 109, 97, 105, 110, 0, 1, 9, 7, 1, 0, 65, 0, 11, 1, 0, 10, 10, 1, 8, 0, 65, 0, 17, 0, 0, 26, 11, 0, 13, 4, 110, 97, 109, 101, 1, 6, 1, 0, 3, 108, 111, 108]
```

The differenced sized bytes result in the slight different compile cost per byte. And the difference is the trailing 10 bytes: `[4,8,...,116]` And the gas difference exactly matches: 2167500 = 10 bytes * contract_compile_bytes (216750).  And this trailing bytes doesn't cause a functional difference of the parsed wasm. When turn wasm to wast again, I observed they're identical.


runtime/near-vm-runner/Cargo.toml
```diff
@@ -38,7 +38,7 @@ threadpool = "1.8.1"
 [dev-dependencies]
 near-test-contracts = { path = "../near-test-contracts" }
 assert_matches = "1.3"
-wat = "1.0"
+wat = "<=1.0.38"
```
matklad, line 41: Pining to old wat is backwards, it's better to upgrade it instead. 

I think the difference boils down to wat generating (or not generating) a custom section with debug names for functions.

ailisp: > I think the difference boils down to wat generating (or not generating) a custom section with debug names for functions.

I believe you're correct, and that slight size changes compilation cost which is charged per byte. I'll pin it to latest wat, in case new version would change debug names again.

matklad: I'd rather not pin -- pining creates an instant technical dept. If the change it again, we should raise an issue upstream asking about adding a mode which doesn't emit custom sections.

matklad (approved):
Won't block this, but I'd still perfer if we said `wat = "1.0.40"` and just kept an eye on further changes.

PR #4773, opened 2021-09-01
chore(near-primitives): document why FunctionCallPermission::receiver_id != AccountId

### LINKED ISSUE #4739: Explain why FunctionCallPersmission doesn't use AccountId
This looks suspect to me:

https://github.com/near/nearcore/blob/dabf861c38dcc1c11c2e31bf898b0ca08de95e15/core/primitives-core/src/account.rs#L202-L203

I'd expect this to be `AccountId`. This was introduced in https://github.com/near/nearcore/commit/2d7b3c0eb1823f4f96748cac9e561874dad10af9#diff-e15377fa8f9e6deba5e3549c4b497cb0c96b0d799982a30ec5d6ef442729d32aR203, so I suspect the reason is that "it'd be a protocol change" is the reason for that. 

This needs a comment in the source code I think, cc @miraclx

miraclx:
Resolves #4739: documents why the `receiver_id` field in `FunctionCallPermission` isn't an `AccountId` despite recently refactoring the whole workspace to using strict `AccountId`-s.

See https://github.com/near/nearcore/pull/4621#issuecomment-892099860 for more context


frol (approved):

frol:
@miraclx  Thanks for landing these code quality improvements!

bowenwang1996 (approved):

core/primitives-core/src/account.rs
```diff
@@ -199,6 +199,9 @@ pub struct FunctionCallPermission {
     #[serde(with = "option_u128_dec_format")]
     pub allowance: Option<Balance>,
 
+    // this isn't an AccountId because already existing records in testnet genesis have invalid
```
matklad, line 202:
```suggestion
    // This isn't an AccountId because already existing records in testnet genesis have invalid
```

I've found that the rule "make each comment a proper sentence/paragrapth, with correct capitalization and punctuation" improved quality of my comments. It's not that capital letters matter per se, its' just that when you set up to write a proper English sentence, and not a "mere comment", you tend to express yourself more clearly. 

Obviously not important :)

matklad (approved):

PR #4779, opened 2021-09-02
refactor: reintroduce structure for storing `RuntimeConfig`s

Longarithm:
Reintroduce https://github.com/near/nearcore/pull/4698, revert its revert https://github.com/near/nearcore/pull/4752.

Update is that now we take into account previous way to get correct runtime config, where we took config from genesis and modified it sequentially. This implementation takes into account that for testnet currently `min_allowed_top_level_account_length` is zero.

Future plan is to make a protocol upgrade and make configs on all chains the same. 

~~Also we will have to hardfork betanet where runtime config is very far from actual.~~
 
Difference between old and new PR is smaller, you can take a look at https://github.com/near/nearcore/compare/3474a46...updatable-config for reviewing.

Test plan
---------
* Same tests as in https://github.com/near/nearcore/pull/4698
* New test to check `genesis_runtime_config` override


core/primitives/src/runtime/config_store.rs
```diff
@@ -0,0 +1,147 @@
...
+}
+
+impl RuntimeConfigStore {
+    /// Constructs a new store.
+    ///
+    /// If min_allowed_top_level_account_length is Some, overrides the value in
+    /// account_creation_config in all runtime configs in resulting store.
+    /// TODO #4775: disable this override for future protocol versions, because it is needed only
+    /// for old testnet versions.
+    pub fn new(min_allowed_top_level_account_length: Option<u8>) -> Self {
```
bowenwang1996, line 35: I am not sure whether this is the best approach. I remember that on betanet we also have registrar account id set to something different from mainnet.

I think we can always load the latest runtime config to override the existing one based on protocol version here. A more general approach could be to expand this to a "protocol config" and override what is currently `GenesisConfig` (most of it) based on the protocol version, but that likely requires more refactoring and could be done later.

core/primitives/src/runtime/config.rs
```diff
@@ -55,6 +55,7 @@ impl RuntimeConfig {
 /// but i) may have it’s `max_gas_burnt_view` limit adjusted and ii) provides
 /// a method which returns configuration with adjustments done through protocol
 /// version upgrades.
+/// TODO #4649: deprecated after RuntimeConfigStore creation, remove it
```
matklad, line 58: One trick to refactoring large code-bases is to mark deprecated functionality as `#[deprecated]`, and mark all call-sites as `#[allow(deprecated)]`. You can do this in a mechanical fashion, and it makes sure that people won't introduce new usages. 

You can then fix call-sites one by one in the follow up PRs. 

NOt saying that we should do this in this PR, it's just a general observation that incomplete refactors often lead to code debt, and that's something to keep an eye (or tools) on!

core/primitives/src/runtime/config_store.rs
```diff
@@ -0,0 +1,147 @@
...
+        include_bytes!(concat!("../../../../nearcore/res/runtime_configs/", $file))
+    };
+}
+
+/// Stores pairs of protocol versions for which runtime config was updated and
+/// the new runtime config in bytes.
+/// Protocol versions are given in increasing order. First one is always 0, so that each version is
+/// mapped to some config.
+static CONFIGS: [(ProtocolVersion, &[u8]); 2] =
+    [(0, include_config!("29.json")), (42, include_config!("42.json"))];
```
matklad, line 19: Yeah, this is ineed much, much better than pulling some random crate.io crate for this!

core/primitives/src/runtime/config_store.rs
matklad: Yeah, plus one to the idea that, if you pass one config parameter, you might instead allow the caller to do free-form modifications. That way, the callee isn't coupled to the specifics of what needs to be overridden. 

I usually write such API as:

```rust
impl RuntimeConfigStore {
    fn override_genesis_config(self, f: impl FnOnce(&mut RuntimeConfig)) -> RuntimeConfigStore
}
```

nearcore/res/README.md
```diff
@@ -0,0 +1,24 @@
...
+Genesis used for testing.
+
+### `mainnet_restored_receipts.json`
+
+Stores receipts restored after the fix of applying chunks. See [#4248]((see https://github.com/near/nearcore/pull/4248/)) for more details.
+
+### `storage_usage_delta.json`
+
+Stores difference of storage usage applied to mainnet after observed bug related to delete key action. See [#3824](https://github.com/near/nearcore/issues/3824) for more details.
+
```
matklad, line 24: :+1: for documenting the parts which are important to document!

nearcore/tests/load_configs.rs
```diff
@@ -0,0 +1,19 @@
...
+    let genesis_runtime_config = genesis.config.runtime_config;
+    let actual_runtime_config = ActualRuntimeConfig::new(genesis_runtime_config);
+
+    let store = RuntimeConfigStore::new(None);
+    for protocol_version in [29u32, 34u32, 42u32, 50u32].iter() {
+        let old_config = actual_runtime_config.for_protocol_version(protocol_version.clone());
+        let new_config = store.get_config(protocol_version.clone());
+        assert_eq!(old_config, new_config);
+    }
+}
```
matklad, line 19: This adds a new test binary, which non-trivially increases the build time:

```
# make sure deps are fully compiled
λ cargo test -q --no-run --test economics

# compile the test in isolation
λ cargo test --no-run --test load_configs
   Compiling nearcore v0.0.0 (/home/matklad/projects/nearcore/nearcore)
    Finished test [unoptimized + debuginfo] target(s) in 1.81s
```

That is, just this one test adds 1.81 seconds of compile time. 

Better to put this as a unit-test into some existing crate:

https://matklad.github.io/2021/02/27/delete-cargo-integration-tests.html

matklad (approved):

Longarithm:
Now we take genesis runtime config again, and override configs from store by this config for lower protocol versions.

core/primitives/src/runtime/config_store.rs
```diff
@@ -0,0 +1,153 @@
...
+                (protocol_version, Arc::new(serde_json::from_slice(config_bytes).unwrap()))
+            }));
+
+        if let Some(runtime_config) = genesis_runtime_config {
+            let mut config = runtime_config.clone();
+            store.insert(0, Arc::new(config.clone()));
+
+            config.storage_amount_per_byte = 10u128.pow(19);
+            store.insert(42, Arc::new(config.clone()));
+        }
```
bowenwang1996, line 48: This piece of code is a bit confusing and hard to read, but I also don't see a much better way to do it. I guess we dug a hole for ourselves :(

bowenwang1996 (approved):

nearcore/res/README.md
```diff
@@ -0,0 +1,24 @@
+## Core Resource Files
+
+Stores resource data which is part of the protocol stable enough to be moved outside of the code.
+
+### `runtime_configs`
+
+For each protocol version `V` where runtime config was updated, we store corresponding config in this folder in `N.json` file.
```
ailisp, line 7: Is this V and N same?

ailisp (approved):
Looks good to me!

nikurt (approved):

nearcore/res/README.md
Longarithm: Yes, good catch :)

posvyatokum (approved):

Issue #4771, opened 2021-09-01
Runtime hits unknown perfomance cliff in the estimator


matklad:
See 

[https://github.com/matklad/nearcore/commit/80b44a877a1d2de76b4adcf63567984d615e7c36](https://github.com/matklad/nearcore/commit/80b44a877a1d2de76b4adcf63567984d615e7c36)

In that commit, I add a benchmark which sends 150 blocks, where each
 block contains 100 deploy code transaction.

EDIT: [https://github.com/near/nearcore/issues/4771#issuecomment-933347496](https://github.com/near/nearcore/issues/4771#issuecomment-933347496) is modernized version.

The benchmark is implemented in two ways:

*   on top of parameter estimator testbed
*   on top of runtime tester

In the first case, I observe a very stage behavior, where at some point
 block processsing times severely degrades.

To reproduce, checkout this commit and run

```console
 $ cargo test --release --package runtime-params-estimator --lib --features required – v2::fast --exact --nocapture
 $ cargo test --release --package runtime-params-estimator --lib --features required – v2::slow --exact --nocapture
 ```

Look for `Runtime::apply` lines in the output, they they tell you how
 long it takes to process one apply.

In the `fast` case, every apply takes about 16ms for me. In the `slow`
 case, apply start at about 30ms, but then abruptly go to 600ms.

Would be great to understand why the time suddenly balloons!

mina86:
Quick look indicates that the slowdown comes from the storage.  I’ve
added some more tracing and here’s output with a fast
`process_transaction` invocation followed by a slow one (they are
within the same `apply` call):

    147.38µs         Runtime::process_transaction
      145.70µs         verify_and_charge_transaction
        48.12µs          validate_transaction
        46.70µs          get_account
          46.15µs          TrieUpdate::get
        49.21µs          get_access_key
          48.65µs          TrieUpdate::get
    …
    2.64ms           Runtime::process_transaction
      2.64ms           verify_and_charge_transaction
        49.03µs          validate_transaction
        1.30ms           get_account
          1.30ms           TrieUpdate::get
        1.29ms           get_access_key
          1.29ms           TrieUpdate::get

`TrieUpdate::get` is suddenly 25 times slower and it’s called quite
a bit so it adds up.

bowenwang1996:
@mina86 is it because the value is not cached and you have to do a trie traversal?

mina86:
It doesn’t look like cache issues.  I haven’t measured cache
[...]

PR #4781, opened 2021-09-02
refactor: use external crate for displaying tracing spans

matklad:
When debugging #4771, I had to copy-paste tracing_timings module to get
a useful profiling output, which isn't ideal. So I improved and
published that code to crates.io as `tracing-span-tree`.

In particular, this new implementation prints the trace in the correct
order, and it also has an `.aggregate(true)` mode for cases where you
run the profiled code in a loop and want a summary.

Example without aggregation:

```
  328.25ms         vm_script_smoke_test
    306.34ms         run_vm
      306.32ms         run_wasmer
        37.16µs          get_key
        301.50ms         compile_and_serialize_wasmer
        4.44ms           run_method
          2.53ms           run_method/instantiate
          128.17µs         run_method/call
          1.74ms           run_method/drop_instance
    4.50ms           run_vm
      4.49ms           run_wasmer
        31.81µs          get_key
        4.25ms           run_method
          2.45ms           run_method/instantiate
          45.46µs          run_method/call
          1.72ms           run_method/drop_instance
    4.43ms           run_vm
      4.42ms           run_wasmer
        22.43µs          get_key
        4.21ms           run_method
          2.41ms           run_method/instantiate
          42.84µs          run_method/call
          1.72ms           run_method/drop_instance
    4.81ms           run_vm
      4.80ms           run_wasmer
        22.09µs          get_key
        4.58ms           run_method
          2.40ms           run_method/instantiate
          419.57µs         run_method/call
          1.73ms           run_method/drop_instance
```

Example with aggregation (unit-less number is the number of aggregated calls):

```
  333.98ms         vm_script_smoke_test
    325.54ms  4      run_vm
      325.48ms  4      run_wasmer
        307.07ms         compile_and_serialize_wasmer
        129.19µs  4      get_key
        17.32ms   4      run_method
          612.46µs  4      run_method/call
          6.63ms    4      run_method/drop_instance
          9.94ms    4      run_method/instantiate
```

Test Plan
---------

This is a debugging utility which is not enabled in production, so I just manually verified that the output makes sense.


matklad:
Good call!

Issue #4561, opened 2021-07-23
Understand the hardware requirements for simple nightshade block producers


bowenwang1996:
When we transition from one shard to eight shards in simple nightshade, block producers need to track all shards. As a result, they may need to have better hardware (more cpus more example) than what is currently specified in [the docs](https://docs.near.org/docs/develop/node/validator/hardware#docsNav). After #3825 is done, we should run some experiments to understand what changes to hardware are needed for block producers. We could then potentially batch this change in hardware requirements with #4461 to update the fees.

bowenwang1996:
To be more specific, we need:
* A cluster with 100 nodes running and loadtest it with transactions. For this part we can take/modify the existing mocknet scripts
* We also need to change the genesis configuration to make sure that every validator tracks all 8 shards. This can be done by modifying `num_block_producer_seats_per_shard`.

cc @nikurt

PR #4785, opened 2021-09-03
Update mocknet loadtest scripts

nikurt:
Tested this by running load tests with 4 and 100 instances brought up by the terraform scripts.
Results of loadtests can be seen in Grafana: https://grafana.near.org/d/jHbiNgSnz/mocknet?orgId=1&from=now-1h&to=now&refresh=10s

Issue #4561


pytest/Pipfile
```diff
@@ -0,0 +1,24 @@
+[[source]]
```
mina86, line 1: Bad merge? I’ve deleted Pipfile a few commits back since we already have requirements.txt and it’s best to have one source where dependencies are listed.

pytest/lib/cluster.py
```diff
@@ -386,14 +382,13 @@ def resume_network(self):
...
             # Get existing instance assume it's ready to run
-            name = args[0]
-            self.instance_name = name
+            self.instance_name = instance_name
             self.port = 24567
             self.rpc_port = 3030
-            self.machine = gcloud.get(name)
+            self.machine = gcloud.get(instance_name, username=username, project=project, ssh_key_path=ssh_key_path)
             self.ip = self.machine.ip
         elif len(args) == 4:
```
mina86, line 393: This branch looks weird now. Should this be `3` now? Should the
prototype of `__init__` be:

```
def __init__(self, instance_name=None, *args, username=None, project=None, ssh_key_path=None, rpc_port=3030):
    ...
```

I think that’s valid in Python 3.7 (which we seem to be targeting sinc
that’s what’s on NayDuck).

pytest/lib/mocknet.py
```diff
@@ -1,36 +1,41 @@
...
 while tmux has-session -t near; do
 tmux kill-session -t near || true
 done
+sudo mv /home/ubuntu/near.log /home/ubuntu/near.log.1
+sudo rm -rf /home/ubuntu/.near/data
 '''
 
 TMUX_START_SCRIPT = '''
-sudo rm -rf /tmp/near.log
+sudo rm -rf /home/ubuntu/near.log
```
mina86, line 36: Should this be the `mv` instead of having the move in stop script?

pytest/lib/mocknet.py
```diff
@@ -1,36 +1,41 @@
...
+RPC_PORT = 3030
+PROJECT = 'near-mocknet'
+PUBLIC_KEY = "ed25519:76NVkDErhbP1LGrSAf5Db6BsFJ6LBw6YVA4BsfTBohmN"
 TX_OUT_FILE = '/home/ubuntu/tx_events'
 
 TMUX_STOP_SCRIPT = '''
 while tmux has-session -t near; do
 tmux kill-session -t near || true
 done
+sudo mv /home/ubuntu/near.log /home/ubuntu/near.log.1
```
mina86, line 31: `2>/dev/null` in case file doesn’t exist?

pytest/lib/mocknet.py
```diff
@@ -327,6 +336,103 @@ def stop_node(node):
         m.run('sudo -u ubuntu -i', input=TMUX_STOP_SCRIPT)
 
 
+def create_and_upload_genesis(nodes, genesis_template_filename):
+    mocknet_genesis_filename = '/tmp/mocknet_genesis.json'
```
mina86, line 340: Use `tempfile.NamedTemporaryFile`. You can change `create_genesis_file` to take output file handler as argument and that’ll work nicely.

pytest/lib/mocknet.py
```diff
@@ -327,6 +336,103 @@ def stop_node(node):
...
+                                                                      "code_hash": "11111111111111111111111111111111",
+                                                                      "storage_usage": 0, "version": "V1"}}})
+            genesis_config['records'].append({"AccessKey": {"account_id": node_account_name(node.instance_name),
+                                                            "public_key": PUBLIC_KEY,
+                                                            "access_key": {"nonce": 0, "permission": "FullAccess"}}})
+            genesis_config['validators'].append(
+                {"account_id": node_account_name(node.instance_name), "public_key": PUBLIC_KEY,
+                 "amount": STAKED_BALANCE})
+    with open(mocknet_genesis_filename, 'w') as f:
+        f.write(json.dumps(genesis_config, indent=2))
```
mina86, line 392: json.dump(genesis_config, f, indent=2)

pytest/lib/mocknet.py
```diff
@@ -327,6 +336,103 @@ def stop_node(node):
...
+def create_genesis_file(nodes, genesis_template_filename, mocknet_genesis_filename):
+    with open(genesis_template_filename) as f:
+        MASTER_BALANCE = "1000000000000000000000000000000000"
+        TREASURY_BALANCE = MASTER_BALANCE
+        VALIDATOR_BALANCE = "20000000000000000000000000000000"
+        STAKED_BALANCE = "15000000000000000000000000000000"
+        MASTER_ACCOUNT = "near"
+        TREASURY_ACCOUNT = "test.near"
+
+        genesis_config = json.loads(f.read())
```
mina86, line 361: with open(genisis_template_filename) as rd:
        genesis_config = json.load(rd)

and then dedent the rest of the code.

pytest/lib/mocknet.py
```diff
@@ -327,6 +336,103 @@ def stop_node(node):
...
+                 "amount": STAKED_BALANCE})
+    with open(mocknet_genesis_filename, 'w') as f:
+        f.write(json.dumps(genesis_config, indent=2))
+
+
+def update_boot_nodes(nodes):
+    assert (len(nodes) > 0)
+    first_node = nodes[0]
+    mocknet_config_filename = '/tmp/mocknet_config.json'
+    first_node_key_filename = '/tmp/mocknet_node_key.json'
```
mina86, line 399: Again, prefer `NamedTemporaryFile` since it respects `TEMPDIR` environment variable and cleans up the files.

pytest/lib/mocknet.py
```diff
@@ -327,6 +336,103 @@ def stop_node(node):
...
+def apply_genesis_changes(node_dir, genesis_config_changes):
+    # apply genesis.json changes
+    fname = os.path.join(node_dir, 'genesis.json')
+    with open(fname) as f:
+        genesis_config = json.loads(f.read())
+    for change in genesis_config_changes:
+        cur = genesis_config
+        for s in change[:-2]:
+            cur = cur[s]
+        assert change[-2] in cur
```
mina86, line 426: This assert isn’t really necessary and inconsistent with the rest of the code where you don’t check that explicitly. You’ll get KeyError anyway.

mina86 (commented):
Looks sensible to me though to be honest I’m not that familiar with the code.

pytest/requirements.txt
```diff
@@ -1,5 +1,6 @@
 PyGithub
 base58
+cython
```
mina86, line 3: Are we actually using this for anything? I couldn’t find any usage.

pytest/tests/mocknet/load_testing_helper.py
```diff
@@ -52,7 +47,9 @@ def get_nonce_for_pk(account_id, pk, finality='optimistic'):
             "account_id": account_id,
             "finality": finality
         })
+    assert len(access_keys['result']['keys']) > 0
     for k in access_keys['result']['keys']:
+        assert 'public_key' in k
```
mina86, line 52: KeyError will be thrown before if this is not true so perhaps this is unnecessary?

pytest/requirements.txt
mina86: So scikit-learn requires cypthon. Very nice of them to include that in their dependencies.

pytest/lib/mocknet.py
```diff
@@ -337,96 +335,108 @@ def stop_node(node):
 
 
 def create_and_upload_genesis(nodes, genesis_template_filename):
-    mocknet_genesis_filename = '/tmp/mocknet_genesis.json'
-    create_genesis_file(nodes, genesis_template_filename, mocknet_genesis_filename)
+    mocknet_genesis_file = tempfile.NamedTemporaryFile(mode='r+', delete=False, suffix='.mocknet.genesis')
```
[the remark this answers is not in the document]
nikurt: Your suggestions how to avoid closing the file make sense to me, and I tried them out, but I got a truncated `genesis.json`. Given that I close and reopen the files, `delete=False` is actually needed. No harm in keeping these temp files in the temp directory, it's about 6MB of files per loadtest run.

pytest/lib/mocknet.py
```diff
@@ -50,30 +65,51 @@
...
+                           username=NODE_USERNAME,
+                           ssh_key_path=NODE_SSH_KEY_PATH)
+    nodes = []
+    pmap(
+        lambda machine: nodes.append(
+            GCloudNode(machine.name,
+                       username=NODE_USERNAME,
+                       project=PROJECT,
+                       ssh_key_path=NODE_SSH_KEY_PATH)), machines)
+    return nodes
```
mina86, line 88: Why not `return pmap(lambda machine: GCloudNode(...), machines)`?

pytest/lib/mocknet.py
```diff
@@ -50,30 +65,51 @@
...
-    m = node.machine
-    target_dir = create_target_dir(m)
-    m.download(f'/home/ubuntu/.near/validator_key.json', target_dir)
-    return Key.from_json_file(f'{target_dir}/validator_key.json')
+    validator_key_file = tempfile.NamedTemporaryFile(
+        mode='r+', delete=False, suffix=f'.mocknet.loadtest.validator_key')
+    validator_key_file.close()
+    node.machine.download(f'/home/ubuntu/.near/validator_key.json',
+                          validator_key_file.name)
+    return Key.from_json_file(validator_key_file.name)
```
mina86, line 112: Even if you don’t use the file returned by `NamedTemporaryFile` you
should still be able to do:

    validator_key_file = tempfile.NamedTemporaryFile(
        mode='r+', suffix=f'.mocknet.loadtest.validator_key')
        node.machine.download(f'/home/ubuntu/.near/validator_key.json',
                              validator_key_file.name)
        return Key.from_json_file(validator_key_file.name)

which will properly clean up the temporary file at the end.  Same in
other places.

pytest/lib/mocknet.py
```diff
@@ -101,45 +139,43 @@ def setup_python_environments(nodes, wasm_contract):
     pmap(lambda n: setup_python_environment(n, wasm_contract), nodes)
 
 
-def start_load_test_helper_script(script, index, pk, sk):
+def start_load_test_helper_script(script, node_account_id, pk, sk):
     return f'''
         cd {PYTHON_DIR}
-        nohup ./venv/bin/python {script} {index} "{pk}" "{sk}" > load_test.out 2> load_test.err < /dev/null &
+        nohup ./venv/bin/python {script} {node_account_id} "{pk}" "{sk}" > load_test.out 2> load_test.err < /dev/null &
```
mina86, line 145: I’d sleep less worried if you’d use `shlex` to quote the arguments:

    import shlex

    return '''
        cd {dir}
        nohup ./venv/bin/python {script} {node_account_id} {pk} {sk} > load_test.out 2> load_test.err < /dev/null &
    '''.format(dir=shlex.quote(PYTHON_DIR),
               script=shlex.quote(script),
               node_account_id=shlex.quote(str(node_account_id)),
               pk=shlex.quote(pk),
               sk=shlex.quote(sk))

On the other hand that might be beyond the scope of this commit.

pytest/lib/mocknet.py
```diff
@@ -167,43 +212,59 @@ def chain_measure_bps_and_tps(archival_node,
...
             block_times.append(curr_time)
-            chunk_hash = curr_block['chunks'][0]['chunk_hash']
-            chunk = archival_node.get_chunk(chunk_hash)['result']
-            tx_count.append(len(chunk['transactions']))
+            txs = 0
+            tx_per_chunk = [None] * len(curr_block['chunks'])
+            pmap(
+                lambda i: get_chunk_txn(i, curr_block['chunks'], archival_node,
+                                        tx_per_chunk),
+                range(len(curr_block['chunks'])))
```
mina86, line 231: Isn’t this just:

    def get_chunk_txn(chunk, archival_node):
        chunk_hash = chunk['chunk_hash']
        chunk = archival_node.get_chunk(chunk_hash)['result']
        return len(chunk['transactions'])

    tx_per_chunk = pmap(
        lambda chunk: get_chunk_txn(chunk, archival_node),
        curr_block['chunks'])

pytest/lib/mocknet.py
```diff
@@ -327,13 +388,244 @@ def stop_node(node):
...
 
 
+def upload_and_extract(node, compressed_filename, dst_filename, src_filename):
+    node.machine.upload(compressed_filename,
+                        '/home/ubuntu',
+                        switch_user='ubuntu')
+    node.machine.run(
+        'bash',
+        input=
+        f'unzip /home/ubuntu/$(basename {compressed_filename}); mv /home/ubuntu/$(basename {src_filename}) {dst_filename}'
```
mina86, line 398: Doesn’t `node.machine.run` already run command through shell? So this should just be: `node.machine.run('unzip ...; mv ...')`, no?  Also, why call `basename` on remote side? Just use `os.path.basename`? And like before perhaps use `shlex`. Lastly, it feels like it would be easier to use `gzip` since you wouldn’t need to do `mv` afterwards; just upload directly as `{dst_filename}.gz`.

pytest/lib/mocknet.py
```diff
@@ -327,13 +388,244 @@ def stop_node(node):
...
         start_process = m.run('sudo -u ubuntu -i', input=TMUX_START_SCRIPT)
-        assert start_process.returncode == 0, m.name + '\n' + start_process.stderr
+        if start_process.returncode == 0:
+            break
+        else:
+            print(
+                f"Failed to start process, returncode: {returncode}\n{m.name}\n{start_process.stderr}"
+            )
+            attempt += 1
+            time.sleep(1)
```
mina86, line 629: Shouldn’t this throw some kind of exception if all attempts fail?

pytest/lib/mocknet.py
```diff
@@ -327,13 +388,244 @@ def stop_node(node):
...
+        if pid != '':
+            break
         start_process = m.run('sudo -u ubuntu -i', input=TMUX_START_SCRIPT)
-        assert start_process.returncode == 0, m.name + '\n' + start_process.stderr
+        if start_process.returncode == 0:
+            break
+        else:
+            print(
+                f"Failed to start process, returncode: {returncode}\n{m.name}\n{start_process.stderr}"
+            )
```
mina86, line 626: `logger.warn` or some such.

pytest/lib/mocknet.py
```diff
@@ -327,13 +388,244 @@ def stop_node(node):
...
+    assert res.returncode == 0
+    pmap(
+        lambda node: upload_and_extract(node, compressed_genesis_filename,
+                                        dst_filename, src_filename), nodes)
+
+
+# We assume that the nodes already have the .near directory with the files
+# node_key.json, validator_key.json and config.json.
+def create_and_upload_genesis(nodes, genesis_template_filename):
+    print('Uploading genesis and config files')
```
mina86, line 414: `logger.info`

pytest/lib/mocknet.py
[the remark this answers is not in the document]
mina86: You probably just need to call `flush` on the objected returned by `NamedTemporaryFile` before you pass the path to the file to external command (such as zip).

pytest/lib/mocknet.py
```diff
@@ -167,43 +212,60 @@ def chain_measure_bps_and_tps(archival_node,
...
-            chunk = archival_node.get_chunk(chunk_hash)['result']
-            tx_count.append(len(chunk['transactions']))
+            txs = 0
+            tx_per_chunk = [None] * len(curr_block['chunks'])
+            pmap(
+                lambda i: get_chunk_txn(i, curr_block['chunks'], archival_node,
+                                        tx_per_chunk),
+                range(len(curr_block['chunks'])))
+            for stat in tx_per_chunk:
+                txs += stat
```
mina86, line 233: `txs = sum(tx_per_chunk)`

pytest/lib/mocknet.py
```diff
@@ -167,43 +212,60 @@ def chain_measure_bps_and_tps(archival_node,
...
-        target_file = f'./logs/{m.name}_txs'
-        m.download(TX_OUT_FILE, target_file)
-        with open(target_file) as f:
+        target_file = f'./logs/{node.instance_name}_txs'
+        node.machine.download(tx_filename, target_file)
+        with open(target_file, 'r') as f:
             return [float(line.strip()) for line in f.readlines()]
-    except:
+    except Exception as e:
+        logger.info(f'Getting tx_events from {node.instance_name} failed: {e}')
```
mina86, line 260: Perhaps `logger.exception` instead? You won’t need to pass the exception explicitly then.

pytest/tests/mocknet/load_testing.py
```diff
@@ -100,14 +99,31 @@ def check_slow_blocks(initial_metrics, final_metrics):
 if __name__ == '__main__':
     logger.info('Starting Load test.')
     nodes = mocknet.get_nodes()
-    initial_validator_accounts = mocknet.list_validators(nodes[0])
+    logger.info(f'Starting Load test using {len(nodes)} nodes.')
+
+    if not '--skip_restart' in sys.argv:
```
mina86, line 104: `'--skip_restart' not in sys.argv` is more idiomatic.

pytest/tests/mocknet/load_testing_helper.py
```diff
@@ -62,119 +60,144 @@ def get_latest_block_hash():
...
+def send_random_transactions(node_account_id, test_accounts):
+    pmap(
+        lambda account_and_index: random_transaction(account_and_index[
+            0], account_and_index[1], node_account_id), test_accounts)
 
-def send_random_transactions(i0):
-    pmap(lambda x: random_transaction(x, i0), test_accounts)
 
+def random_delay():
+    random_delay = random.randint(0, 10)
```
mina86, line 96: Any reason to limit the sleep to integer number of seconds? Perhaps `random_delay = random.random() * 10` instead?

mina86 (approved):
I didn’t analyse the logic of the test too closely but overall it all looks good to me.

pytest/lib/mocknet.py
nikurt: Yes and no. You are right that I forgot that `pmap` returns a list. But I also want the results to be in the order of the chunks. To inspect if a single shard gets over- or under-loaded.

pytest/lib/mocknet.py
nikurt: Thanks for the `gzip` suggestion.

nikurt:
Thank you @mina86 for the thorough review and great suggestions.

pytest/lib/mocknet.py
mina86: `pmap` does return values in order.

PR #4793, opened 2021-09-06
fix(chain): fix block ordinal to block hash map

bowenwang1996:
There is a subtle bug in how we populate `ColBlockOrdinal`. If there is a fork, then depending on when the fork block arrives, it is possible for the fork block to override the block ordinal (since it has the same ordinal as the block on the canonical chain). This PR fixes this issue by updating block ordinal mapping when the canonical chain changes.

Test plan
-----------
* unit tests
* migrated an archival node and confirmed with @mfornet that the previous issue with bridge would indeed be resolved with this PR.


chain/chain/src/store.rs
```diff
@@ -1616,6 +1616,10 @@ impl<'a> ChainStoreUpdate<'a> {
             for height in (header_height + 1)..prev_height {
                 self.chain_store_cache_update.height_to_hashes.insert(height, None);
             }
+            // Override block ordinal to hash mapping for blocks in between.
+            // At this point block_merkle_tree for header is already saved.
+            let block_ordinal = self.get_block_merkle_tree(&header_hash)?.size();
+            self.chain_store_cache_update.block_ordinal_to_hash.insert(block_ordinal, header_hash);
```
matklad, line 1622: Not sure if that's applicable in this case, but, just a general tip, I like using `.insert` like this: 

```rust
let prev = hash_map.insert(key, value);
assert!(prev.is_none()); // or prev.is_some(), depending on the use case
```

integration-tests/tests/client/process_blocks.rs
```diff
@@ -1917,10 +1918,27 @@ fn test_block_merkle_proof_with_len(n: NumBlocks) {
...
     let genesis_block = env.clients[0].chain.get_block_by_height(0).unwrap().clone();
     let mut blocks = vec![genesis_block.clone()];
-    for i in 1..n {
-        let block = env.clients[0].produce_block(i).unwrap().unwrap();
-        blocks.push(block.clone());
-        env.process_block(0, block, Provenance::PRODUCED);
+    let mut rng = rand::thread_rng();
+    let mut cur_height = genesis_block.header().height() + 1;
+    while cur_height < n {
+        let should_fork = rng.gen_bool(0.5);
```
matklad, line 1924: FWIW, I am not a fan of "just pull `thread_rng()` out of thin air when you need" for large projects. I usually prefer to explicitly thread rng as a parameter. This helps when you figure out that the stuff does break, and you want to fix a seed to debug it. 

As Longarithm recently pointed out to me, `rand::thread_rng` doesn't have reseeding API by design. So debugging tests becomes more cumbersome, as you first need to switch the rng. 

But, that's a minor point, not that it make big practical difference for a single case!

matklad (approved):

frol (approved):

PR #4709, opened 2021-08-18
fix: change default behavior of neard testnet 

mzhangmzz:
fixes #4464 
fixes #4569

Test plan: verify the genesis config generated is the same as 1.19.2

``` 
$ nearcore-1 git:(ebc1e0a26) diff /Users/minzhang/.near-1.19.2/test0/genesis.json /Users/minzhang/.near-now/test0/genesis.json
2,4c2,4
<   "protocol_version": 45,
<   "genesis_time": "2021-08-18T15:45:26.235826Z",
<   "chain_id": "test-chain-WLJhS",
---
>   "protocol_version": 47,
>   "genesis_time": "2021-08-18T15:40:31.252129Z",
>   "chain_id": "test-chain-b25OT",
156a157,159
>         "ripemd160_base": 853675086,
>         "ripemd160_block": 680107584,
>         "ecrecover_base": 3365369625000,
205a209
>         "max_transaction_size": 4194304,
242a247
>   "simple_nightshade_shard_config": null,
251c256,257
<           "storage_usage": 0
---
>           "storage_usage": 0,
>           "version": "V1"
272c278,279
<           "storage_usage": 0
---
>           "storage_usage": 0,
>           "version": "V1"
293c300,301
<           "storage_usage": 0
---
>           "storage_usage": 0,
>           "version": "V1"```

PR #4794, opened 2021-09-06
refactor: reword error msg for reporting transactions failing for requiring full-access keys

### LINKED ISSUE #4746: Misleading error when not full access key is used
**Describe the bug**
If developers tries to send money using function access key, the error is confusing.
Please provide a short description of the bug.

**To Reproduce**
1. Add function access key to the account
2. Sign transaction with single transfer action
Observer error:
```
            InvalidAccessKeyError::RequiresFullAccess => write!(
                f,
                "The transaction contains more then one action, but it was signed \
                 with an access key which allows transaction to apply only one specific action. \
                 To apply more then one actions TX must be signed with a full access key"
            ),
```

**Expected behavior**
Error that says you can not send money with non full access key.

**Screenshots**
image

- nearcore commit/branch
- rust version (if local)
- docker (if using docker)
- mainnet/testnet/betanet/local

**Additional context**
The error is coming from here: https://github.com/near/nearcore/blob/2d7b3c0eb1823f4f96748cac9e561874dad10af9/core/primitives/src/errors.rs#L523

miraclx:
Resolves #4746, hopes to be clearer with the error message when using a non-full-access key for a transaction that either has multiple actions or is not a function-call action.

Before:

```
"The transaction contains more then one action, but it was signed with an access key which allows transaction to apply only one specific action. To apply more then one actions TX must be signed with a full access key
```

After:

```
Invalid access key type. Full-access keys are required for transactions that have multiple or non-function-call actions
```

---

Would've been nice if we could resolve this issue with a message like `Cannot send money without a full access key` but that would require the error variant to encode a cause. But we can at least ease understanding what went wrong with a better error message that properly describes why this variant would be returned.


frol (approved):

frol:
@bowenwang1996 I feel that https://github.com/near/nearcore/pull/4709 did not land to 1.21.x and now we have a discrepancy in the backward compatibility tests.

matklad (approved):

PR #4795, opened 2021-09-06
feat: lower data receipt cost

Longarithm:
Main goal of this PR is to reduce data receipt cost, due to new estimation.

**Cost is reduced by 4.7 Tgas and 4.2 Ggas per KB.**
Note: the latter one is not going to have big impact, because maximal length of data receipt in mainnet is ~0.5 KB:
```
nearcore % psql postgres://public_readonly:nearprotocol@104.199.89.51/mainnet_explorer
mainnet_explorer=> select max(length(data)) as len from data_receipts;
 len
-----
 567
```
 
Secondary goal is to make protocol configs of all chains the same by default. Doesn't seem significant to make separate PR for this.

Test plan
---------
- Check that cost is lowered
- Check that genesis config is overridden for new protocol version


Longarithm:
Steps to reproduce: https://hackmd.io/w6ODyKjUReuuofXTuqdyFQ

core/primitives/src/runtime/config_store.rs
```diff
@@ -15,8 +15,11 @@ macro_rules! include_config {
 /// the new runtime config in bytes.
 /// Protocol versions are given in increasing order. First one is always 0, so that each version is
 /// mapped to some config.
-static CONFIGS: [(ProtocolVersion, &[u8]); 2] =
-    [(0, include_config!("29.json")), (42, include_config!("42.json"))];
+static CONFIGS: [(ProtocolVersion, &[u8]); 3] = [
+    (0, include_config!("29.json")),
+    (42, include_config!("42.json")),
+    (115, include_config!("115.json")),
```
bowenwang1996, line 21: I think we should consider guarding this with nightly protocol feature as well so that we don't forget to change it when it is ready for release.

bowenwang1996 (approved):

matklad (approved):
Confirmed that I can reproduce the costs on my machine,  that they are not sensitive to the number of accounts (I got the same costs with 20_000 and with 1_000_000 accounts), and that they also reproduce on the new estimator. 

What I haven't confirmed *yet* is that the logic to compute the cost itself is indeed reasonable (that is, that `data_receipt_10b_1000` and `data_receipt_base_10b_1000` indeed measure what they are supposed to measure, in terms of runtime behavior). I would love to take a somewhat deeper look into that, but I don't want to block here, so LGTM if @Longarithm thinks that there aren't *logical* errors in the cost computation.

PR #4796, opened 2021-09-06
feat: pass environment variables to docker

Longarithm:
And also build test contract, which is used in some metrics computation. Corresponding command was included into README.

We could also consider adding this to each metric separately.

Test plan
---------
- cargo check
- make two separate runs and check difference: https://hackmd.io/glLIGVM7RjKSKSe8p9h6rw

### DIFF
--- a/runtime/runtime-params-estimator/src/main.rs
+++ b/runtime/runtime-params-estimator/src/main.rs
@@ -196,7 +196,6 @@ fn main_docker(state_dump_path: &Path, full: bool) -> anyhow::Result<()> {
 
         let mut buf = String::new();
         buf.push_str("set -ex;\n");
-        buf.push_str("export CARGO_PROFILE_RELEASE_LTO CARGO_PROFILE_RELEASE_CODEGEN_UNITS;\n");
         buf.push_str("cd /host/nearcore;\n");
         buf.push_str(
             "\
--- a/runtime/runtime-params-estimator/src/main.rs
+++ b/runtime/runtime-params-estimator/src/main.rs
@@ -242,8 +242,8 @@ cargo build --manifest-path /host/nearcore/Cargo.toml \
         .args(&["--mount", "source=rust-emu-target-dir,target=/host/nearcore/target"])
         .args(&["--mount", "source=rust-emu-cargo-dir,target=/usr/local/cargo"])
         .args(&["--interactive", "--tty"])
-        .arg(&["--env", "CARGO_PROFILE_RELEASE_LTO"])
-        .arg(&["--env", "CARGO_PROFILE_RELEASE_CODEGEN_UNITS"])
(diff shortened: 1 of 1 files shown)


runtime/runtime-params-estimator/src/main.rs
```diff
@@ -176,6 +189,9 @@ fn main_docker(state_dump_path: &Path) -> anyhow::Result<()> {
 
         let mut buf = String::new();
         buf.push_str("set -ex;\n");
+        buf.push_str("export CARGO_PROFILE_RELEASE_LTO CARGO_PROFILE_RELEASE_CODEGEN_UNITS;\n");
+        buf.push_str("echo $CARGO_PROFILE_RELEASE_LTO;\n");
+        buf.push_str("echo $CARGO_PROFILE_RELEASE_CODEGEN_UNITS;\n");
```
matklad: This solves the right problem, but I think there's a better way to do this.

The problem with this setup is that pass-thru requires the person who uses the stuff to a) know about LTO b) to manually set it up. A better meta approach is to encode such tribal as LTO as code. So here I suggest to do the following:

Adding a `--full` flag to the estimator's CLI with the semantics of "run the estimation in the fully production setting to get usable cost table". 

If the `--full` is passed, this script just *sets* lto, instead of passing it through. 

PS: don't really like the `full` name, but, having spend 2 minutes looking for alternatives, I don't know a better one. SUggestions welcome!

runtime/runtime-params-estimator/src/main.rs
```diff
@@ -176,6 +196,7 @@ fn main_docker(state_dump_path: &Path) -> anyhow::Result<()> {
 
         let mut buf = String::new();
         buf.push_str("set -ex;\n");
+        buf.push_str("export CARGO_PROFILE_RELEASE_LTO CARGO_PROFILE_RELEASE_CODEGEN_UNITS;\n");
```
matklad, line 199: Do we actually need this `export`? I'd think that it works without it as well, could you check this?

matklad (approved):

Longarithm:
Comparison of costs with/without LTO (in full/default mode): https://hackmd.io/glLIGVM7RjKSKSe8p9h6rw

Almost everywhere first cost is lower than second cost, and in remaining cases difference fits in 15% deviation. And also for `ActionFunctionCallPerByte` where difference is maximal, it seems to be compensated with `ActionFunctionCallBase`.

Some outliers:
```
DataReceiptCreationPerByte: first=34424023 second=51989896 second/first=1.51
HostFunctionCall: first=104001330 second=160444307 second/first=1.54
PromiseReturn: first=153044323 second=230794857 second/first=1.51
```

runtime/runtime-params-estimator/src/main.rs
Longarithm: Checked, that's correct.

PR #4797, opened 2021-09-07
implement split states

mzhangmzz:
This PR implement the specific logic of how we will split states for re-sharding during block catchup and syncing. 

The entry function is `Chain::build_state_for_split_shards`


chain/chain/src/chain.rs
```diff
@@ -1843,6 +1844,35 @@ impl Chain {
...
+            (block_header.epoch_id().clone(), block_header.next_epoch_id().clone())
+        };
+        let state_root = {
+            let block = self.get_block(&sync_hash)?;
+            block
+                .chunks()
+                .get(shard_id as usize)
+                .ok_or(ErrorKind::InvalidShardId(shard_id))?
+                .prev_state_root()
+        };
```
bowenwang1996, line 1863: What happens when the first block of the epoch (sync hash) does not contain a new chunk for some shard? According to what is done here, we will ignore the state transition happened in the last chunk for that shard, which is not acceptable.

core/store/src/trie/iterator.rs
```diff
@@ -216,6 +216,27 @@ impl<'a> TrieIterator<'a> {
         prefix
     }
 
+    /// Visits all nodes belonging to the interval [path_begin, path_end) in depth-first search
+    /// order and return TrieItem for each visited node with value stored
+    /// Used to generate split states for re-sharding
+    pub(crate) fn get_trie_items(
```
bowenwang1996, line 222: I thought we already have a similar function? cc @Longarithm

nearcore/src/runtime/mod.rs
```diff
@@ -1542,6 +1580,73 @@ impl RuntimeAdapter for NightshadeRuntime {
...
+                .map(|id| {
+                    (
+                        ShardUId::from_shard_id_and_layout(*id, &next_epoch_shard_layout),
+                        StateRoot::default(),
+                    )
+                })
+                .collect()
+        };
+
+        let state_root_node = trie.retrieve_root_node(state_root).map_err(|err| err.to_string())?;
```
bowenwang1996, line 1606: Why do we map err to string here?

nearcore/src/runtime/mod.rs
```diff
@@ -1542,6 +1580,73 @@ impl RuntimeAdapter for NightshadeRuntime {
...
+                .tries
+                .apply_changes_to_new_states(
+                    state_roots,
+                    trie_items.into_iter().map(|(key, value)| (key, Some(value.clone()))).collect(),
+                    Box::new(|raw_key| {
+                        // Here changes on DelayedReceipts or DelayedReceiptsIndices will be excluded
+                        // This is because we cannot migrate delayed receipts part by part. They have to be
+                        // reconstructed in the new states after all DelayedReceipts are ready in the original
+                        // shard.
+                        if let Some(account_id) = parse_account_id_from_raw_key(&raw_key).unwrap() {
```
bowenwang1996, line 1627: is the unwrap safe?

chain/chain/src/chain.rs
mzhangmzz: That's a very good point. I forgot about that. I will get state root from chunk extra

core/store/src/trie/shard_tries.rs
```diff
@@ -47,6 +47,15 @@ impl ShardTries {
         }
     }
 
+    pub fn add_new_shards(&mut self, shards: &Vec<ShardUId>) {
+        Arc::get_mut(&mut self.caches)
+            .unwrap()
+            .extend(Self::get_new_cache(&shards).as_ref().clone());
```
matklad, line 53: Let's change `get_new_cache(shards: &Vec<ShardUId>)` to return an `HashMap<ShardUId, TrieCache>` without an `Arc` -- this cloning of data is superfluous. 

Wow, `get_new_cache` is written in a somewhat more elaborate way then it needs to be, here's a simpler version: 

```rust
    fn get_new_cache(shards: &Vec<ShardUId>) -> HashMap<ShardUId, TrieCache> {
        shards.iter().map(|&shard_id| (shard_id, TrieCache::new())).collect()
    }
```

core/store/src/trie/shard_tries.rs
```diff
@@ -47,6 +47,15 @@ impl ShardTries {
         }
     }
 
+    pub fn add_new_shards(&mut self, shards: &Vec<ShardUId>) {
+        Arc::get_mut(&mut self.caches)
+            .unwrap()
```
matklad, line 52: This `unwrap` feels very worrisome to me -- it isn't locally obvious that we guarantee that there's a single reference here. 

Looking at the broader context, the whole of 

```rust
#[derive(Clone)]
pub struct ShardTries {
    pub(crate) store: Arc<Store>,
    /// Cache reserved for client actor to use
    pub(crate) caches: Arc<HashMap<ShardUId, TrieCache>>,
    /// Cache for readers.
    pub(crate) view_caches: Arc<HashMap<ShardUId, TrieCache>>,
}
```

feels a bit suspect:

* why we arc each field separatelly instead of Arcing the whole thing once? 
* `Arc<Store>` doesn't make sense, because `Store` is already an `Arc`. 


:thinking: yeah, I think we should return to this bit after the PR with an eye towards refactring, for the present case, I think we can take advantage of the fact that the map is small and shallow (the `TrieCache` is another arc), and just re-build the map:

```rust
    pub fn add_new_shards(&mut self, shards: &[ShardUId]) {
        let add_empty_caches = |old_caches: &HashMap<ShardUId, TrieCache>| {
            old_caches
                .iter()
                .map(|(k, v)| (*k, v.clone()))
                .zip(shards.iter().map(|&k| (k, TrieCache::new())))
                .collect::<HashMap<_>>()
        };
        self.caches = add_empty_caches(&*self.caches);
        self.view_caches = add_empty_caches(&*self.caches);
    }
```

core/store/src/trie/split_state.rs
```diff
@@ -0,0 +1,372 @@
...
+    /// # Panics
+    /// storage must be a TrieCachingStorage
+    /// part_id must be in [0..num_parts)
+    ///
+    /// # Errors
+    /// StorageError if the storage is corrupted
+    pub fn get_trie_items_for_part(
+        &self,
+        part_id: u64,
+        num_parts: u64,
```
matklad, line 25: We might consider defining 

```rust
pub struct PartId {
  idx: u64,
  total: u64,
}
```

core/store/src/trie/split_state.rs
```diff
@@ -0,0 +1,372 @@
...
+    /// Apply `changes` to build states for new shards
+    /// `state_roots` contains state roots for the new shards
+    /// Ignore changes on DelayedReceipts or DelayedReceiptsIndices
+    /// Update `store_update` and return new state_roots
+    /// used for building states for new shards in resharding
+    pub fn apply_changes_to_new_states<'a>(
+        &self,
+        state_roots: HashMap<ShardUId, StateRoot>,
+        changes: Vec<(Vec<u8>, Option<Vec<u8>>)>,
+        key_to_shard_id: Box<dyn Fn(&Vec<u8>) -> Option<ShardUId> + 'a>,
```
matklad, line 47:
```suggestion
        key_to_shard_id: &dyn Fn(&[u8]) -> Option<ShardUId> + 'a,
```

core/store/src/trie/split_state.rs
```diff
@@ -0,0 +1,372 @@
...
+                assert!(
+                    state_roots.contains_key(&new_shard_uid),
+                    "Cannot find state roots for shard {:?} for key {}. Only has state roots for shards {:?}",
+                    new_shard_uid,
+                    from_utf8(&raw_key).unwrap(),
+                    state_roots.keys(),
+                );
+                changes_by_shard
+                    .entry(new_shard_uid)
+                    .or_insert_with(Vec::new)
```
matklad, line 61:
```suggestion
                    .or_default()
```

core/store/src/trie/split_state.rs
```diff
@@ -0,0 +1,372 @@
...
+        Ok((store_update, new_state_roots))
+    }
+
+    pub fn apply_delayed_receipts_to_split_states<'a>(
+        &self,
+        state_roots: &HashMap<ShardUId, StateRoot>,
+        receipts: &Vec<Receipt>,
+        account_id_to_shard_id: Box<dyn Fn(&AccountId) -> ShardUId + 'a>,
+    ) -> Result<(StoreUpdate, HashMap<ShardUId, StateRoot>), StorageError> {
+        let mut trie_updates: HashMap<_, _> = HashMap::new();
```
matklad, line 83:
```suggestion
        let mut trie_updates = HashMap::new();
```

the type signature doesn't add any extra info

core/store/src/trie/iterator.rs
Longarithm: The difference is that here we extract only key-value pairs, and in `visit_nodes_interval` below we extract all nodes, including intermediate ones (like branches).
I guess that's intended, because when we split a trie the new trie structure may be different.
Let's state this explicitly by renaming these functions to `get_key_value_pairs_in_interval` and `get_nodes_in_interval`?

core/store/src/trie/shard_tries.rs
mzhangmzz: That's a very good point. The `Arc` seems confusing. 

I actually moved this part of change to another PR https://github.com/near/nearcore/pull/4804 because I realized I need to add a RwLock to shard_tries, since now it can be modified. I'll fix your comments there.

core/store/src/trie/split_state.rs
mzhangmzz: Yeah that's a good suggestion. `PartId` is used in multiple other places too, so I would think it's better to do the refactoring in another PR. I'll create an issue for this.

mzhangmzz: https://github.com/near/nearcore/issues/4806

core/store/src/trie/iterator.rs
bowenwang1996: I see. Yeah probably makes sense to emphasize that we only care about key value pairs here and not trie nodes

core/store/src/trie/shard_tries.rs
mzhangmzz: I checked the code and it seems that `Arc<Store>` is used in many other places in the code too. I don't have enough context to tell why that is the case, but I assume that is why `Arc<Store>` is used here. 

And the function `get_store` needs to return `Arc<Store>`, I guess that's why the field needs to be arc'ed individually

core/store/src/trie/iterator.rs
mzhangmzz: Updated the comment

nearcore/src/runtime/mod.rs
mzhangmzz: I changed it so that the error is propagated

nearcore/src/runtime/mod.rs
mzhangmzz: hmm I think I saw somewhere in the code that `StorageError` is converted to a string and I just followed that. I implemented the trait `From<StorageError>` for `Error` so now `StorageError` is converted to `Error` correctly

bowenwang1996 (approved):

core/store/src/trie/shard_tries.rs
matklad: might make sense to create an issue to get rid of `Arc<Store>` everywhere then -- seems like a minor technical debt. The main problem here is that `Arc`s and `RwLock`s are important large pieces of overall application architecture, so having fewer of those can make the code more understandable.

core/store/src/trie/split_state.rs
```diff
@@ -0,0 +1,373 @@
...
+        key_to_shard_id: &(dyn Fn(&[u8]) -> Result<Option<ShardUId>, StorageError> + 'a),
+    ) -> Result<(StoreUpdate, HashMap<ShardUId, StateRoot>), StorageError> {
+        let mut changes_by_shard: HashMap<_, Vec<_>> = HashMap::new();
+        for (raw_key, value) in changes.into_iter() {
+            if let Some(new_shard_uid) = key_to_shard_id(&raw_key)? {
+                assert!(
+                    state_roots.contains_key(&new_shard_uid),
+                    "Cannot find state roots for shard {:?} for key {}. Only has state roots for shards {:?}",
+                    new_shard_uid,
+                    from_utf8(&raw_key).unwrap(),
```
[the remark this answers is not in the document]
mzhangmzz: Good point. I am going to move this assert outside of this function for a clearer error message. This error will happen if `ShardLayout` is not specified correctly. Since we specify which shards will be split to which shards in shard layout, if that information is not consistent with the result of `account_id_to_shard_id`, this assert will fail.

core/store/src/trie/split_state.rs
```diff
@@ -0,0 +1,373 @@
...
+                    state_roots.keys(),
+                );
+                changes_by_shard.entry(new_shard_uid).or_default().push((raw_key, value));
+            }
+        }
+        let mut new_state_roots = state_roots.clone();
+        let mut store_update = StoreUpdate::new_with_tries(self.clone());
+        for (shard_uid, changes) in changes_by_shard.into_iter() {
```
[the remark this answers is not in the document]
mzhangmzz: The `into_iter` here is intentional so I don't have the dereference later

Longarithm (approved):

frol (approved):
I did not do a heavy review of the internal logic, only checked the modules I am familiar with, they look good to me :+1:

core/primitives/src/shard_layout.rs
```diff
@@ -155,6 +155,10 @@ impl ShardUId {
         assert!(shard_id < shard_layout.num_shards());
         Self { shard_id: shard_id as u32, version: shard_layout.version() }
     }
+
+    pub fn shard_id(&self) -> ShardId {
+        self.shard_id as ShardId
```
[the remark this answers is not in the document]
mzhangmzz: That's a good question. The reason is that `ShardId` right now is part of the database key for state. I need to change that to `ShardUId`, but I don't want to trigger a database migration, so I want to keep the serialization of `ShardId` and `ShardUId` the same. To do that, I made both `version` and `shard_id` in ShardUId u32, so combined they take 8 bytes, which is the same as `ShardId` originally.

PR #4801, opened 2021-09-08
docs: drop obsolete link

matklad:
The link points to the issues from 2020, so it seems that it doesn't
have a place in README?


bowenwang1996 (commented):
Can you remove the paragraph above as well? We no longer use zenhub

bowenwang1996 (approved):

PR #4804, opened 2021-09-08
Add a RwLock to tries in NightshadeRuntime

mzhangmzz:
This PR makes two changes
- Add a new function `add_new_shards` in ShardTries that adds new shards to it. This function will need to be called for building states for new shards in re-sharding
- Locks `tries` in NightshadeRuntime since now it can be modified through `add_new_shards`

Test Plan:
Existing CI tests

### DIFF
--- a/core/store/src/trie/shard_tries.rs
+++ b/core/store/src/trie/shard_tries.rs
@@ -20,9 +20,9 @@ use crate::{StorageError, Store, StoreUpdate, Trie, TrieChanges, TrieUpdate};
 pub struct ShardTries {
     pub(crate) store: Arc<Store>,
     /// Cache reserved for client actor to use
-    pub(crate) caches: HashMap<ShardUId, TrieCache>,
+    pub(crate) caches: Arc<HashMap<ShardUId, TrieCache>>,
     /// Cache for readers.
-    pub(crate) view_caches: HashMap<ShardUId, TrieCache>,
+    pub(crate) view_caches: Arc<HashMap<ShardUId, TrieCache>>,
 }
 
 impl ShardTries {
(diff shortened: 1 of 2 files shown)


Longarithm (approved):

bowenwang1996 (commented):
Given the way that `ShardTries` works right now, I wonder whether it makes more sense to put the lock inside, i.e, changing the types of `caches` and `view_caches`. cc @matklad

bowenwang1996 (approved):

matklad:
Hm, yeah, the current state of the code doesn't makes sense to me, but maybe I am missing the 
end state we have in mind. Currently the code hah `stuff: RwLock<Arc<Stuff>>`, and we use it as

```rust
let stuff = self.stuff.lock().unwrap();.
stuff.compute()
```

This feels weird, because we hold the lock during the whole process of `.compute`. The more natural approach would be to do 

```rust
let stuff = Arc::clone(*&self.stuff.lock().unwrap());
stuff.compute()
```

such that we don't hold the lock longer then we should. 

The problem here is not performance, but the round-about semantics. 

What `RwLock` gives you is ability to go from `&T` to `&mut T`. But for `Arc`s, `&Arc<T>` and `&mut Arc<T>` are *almost* equivalent -- `Arc` is inherently shared, so it neuters `&mut`.

If we do really need something RCU-like here where we can swap the whole datastructure, the arc_swap would be the more precise type to use https://docs.rs/arc-swap/1.3.2/arc_swap/. 

I wonder though, do we actually need that? The ownership pattern we have here is cache. That's tricky in Rust -- it generally means that you only have `&Cache`, but you actually do want to mutate the data stored inside, so you need some kind of interior mutability. A specific problem here is that our cache exposes `&mut` method, and that infects caller's interface -- it's the caller who has to have a lock. 

The reason for why we need that `&mut` is because growing hash map via `&self` is hard. In `get_trie_for_shard_internal`, if the `shard_uid` is known, we can do an unsynchronized lookup in the map and clone the value's  `Arc`. If the `shard_uid` is *not* known, we sort-of want to insert a new, empty cache in the map, but that's not possible. 

It feels like the intended, important property of the code as written that getting `TrieCache` is lockless. We kinda loss that by wrapping the whole thing in `RwLock` :-)

Ok, having written all that (sorry for the stream of consciousness style), I think that's what would be a better approach here.

We should treat `ShardTries` as immutable, read-only thing, all caching should be encapsulated inside. That means that we either need to change the signature of `add_new_shards` to `fn add_new_shards(&self, shards: &[ShardUId])` or (potentially better) just lazily create an empty shard cache on the first access and remove `add_new_shards` altogether. 

To do that, we need to add interior mutability *inside* `ShardTries`, so somethin like 

```rust
pub(crate) caches: ArcSwap<HashMap<ShardUId, TrieCache>>,
// OR
pub(crate) caches: Arc<RwLock<HashMap<ShardUId, TrieCache>>>,
```


I feel this overall thing is relatively important -- Rust is easy, because it has single ownership, & and `&mut`. `RwLock` is an explicit opt-in into "yeah, I *love* debugging iterator invalidations". This `RwLock` in `NightshadeRuntime` would be the first field to directly expose interior mutability at this level, so it feels like a big change.

Disclaimer -- this is all 90% based on my intuition, I don't have a full understanding of the actual problem we are solving here, so it might be that practicality beats purity :)

matklad:
I guess, it would also be useful to take a look at the call-site for `add_new_shards` -- that little `&mut` there is the whole reason for the above comment, so it would be useful to understand if we perhaps can just have `&mut NightshadeRuntime` there?

bowenwang1996:
@mzhangmzz please take a look at Alex's comments

mzhangmzz:
@bowenwang1996 yes I met with Alex yesterday to discuss these comments,  I will fix them in another PR

PR #4805, opened 2021-09-08
pytest: format Python source files

mina86:
Add YAPF style configuration file and format all files using it.  I’ve
skimmed over the changes and there were no outrageous formatting changes
that I’ve noticed.

Formatting can be done with `python3 -m yapf -pir .`.

The YAPF configuration is the one used by NayDuck.  At this point I don’t
recall where I got it from or whether I wrote it myself.

### DIFF
--- a/pytest/.style.yapf
+++ b/pytest/.style.yapf
@@ -1,0 +1,4 @@
+[style]
+based_on_style = google
+indent_width = 4
+column_limit = 80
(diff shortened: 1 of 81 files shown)


mina86:
@nikurt, if this gets merged it will mess with #4785 so I can wait with this PR or you can also try with `git merge -s ours` to just pick up your changes when merging.

nikurt:
Thanks for considering #4785, but it's in a bad state because of CLion formatting. Please proceed, I'll be happy to merge :)

bowenwang1996 (approved):

Issue #2411, opened 2020-04-07
Pytest should have a readme


bowenwang1996:
There is currently no readme as to how pytests should be executed. This makes it difficult for external contributors to start writing python tests. We should document the necessary steps it takes (including installing requirements & compiling nearcore) to run pytests.

bowenwang1996:
@mina86 Could you also address this while overhauling nayduck?

bowenwang1996:
Done in #4680

PR #4808, opened 2021-09-08
pytest: add more to README.md

mina86:
Mention temporary directory (and the tempfile module), describe how
in general NayDuck test environment looks like and add instructions
for invoking YAPF.  Furthermore, change `python` to `python3` since
the latter is the official way to invoke Python 3.x interpreter.

Issue: https://github.com/near/nearcore/issues/2411


bowenwang1996 (approved):
Thanks!

PR #4809, opened 2021-09-09
pytest: add cython back to requirements

mina86:
While we’re not using Cython directly, scikit-learn does and
for some reason that dependency is not picked up automatically.
Add Cython back to requirements.txt.


nikurt (approved):
I also added it back in my draft PR :)

Issue #4696, opened 2021-08-16
Add an integration test for creating invalid subaccount


bowenwang1996:
Add a test that asserts an account `alice` cannot create accounts of type `x.y.alice` but can create accounts of type `x.alice`. This should catch bugs like what this fixed in https://github.com/near/nearcore/pull/4694

miraclx:
I suppose this covers it, right?

https://github.com/near/nearcore/blob/6db93dab90dbdaf238b627c03d0af5b855961718/core/account-id/src/lib.rs#L361

bowenwang1996:
@miraclx that is not an integration test

PR #4810, opened 2021-09-09
Add test for creating invalid subaccount

yordanmadzhunkov:
Add an integration test that asserts an account alice cannot create
accounts of type x.y.alice but can create accounts of type x.alice.

Fixes https://github.com/near/nearcore/issues/4696


bowenwang1996 (approved):
Great. Thanks!

PR #4812, opened 2021-09-09
Adding ScenarioBuilder

posvyatokum:
Adding an easier way to create basic scenarios.


test-utils/runtime-tester/src/scenario_builder.rs
```diff
@@ -0,0 +1,70 @@
...
+}
+
+impl ScenarioBuilder {
+    pub fn new(num_accounts: usize, use_in_memory_store: bool) -> Self {
+        let network_config =
+            NetworkConfig { seeds: (0..num_accounts).map(|x| id_to_seed(x)).collect() };
+
+        ScenarioBuilder {
+            height: 1,
+            nonce: 1,
```
bowenwang1996, line 23: Should we start from 0 instead?

bowenwang1996 (approved):
It's great that we have documentation for how this thing works!

test-utils/runtime-tester/src/scenario_builder.rs
posvyatokum: No, I forgot why, but block at height 0 could not be produced

test-utils/runtime-tester/README.md
```diff
@@ -0,0 +1,105 @@
...
+
+    let runtime_stats = builder.scenario().run().unwrap();
+
+    for block_stats in runtime_stats.blocks_stats {
+        assert!(block_stats.block_production_time < Duration::from_secs(1),
+                "Block at height {} was produced in {:?}",
+                block_stats.height, block_stats.block_production_time);
+    }
+}
+```
```
matklad, line 105: I am not sure what's the best practice here -- putting the stuff in README or into the `lib.rs`. 

Benefits of README:

* easier to notice if you look at the directory structure 
* somewhat easier to edit an external file

Benefits of `lib.rs`

* easier to notice if you look at the code directly
* renders in `cargo doc` output / reader mode in an IDE
* code examples could be run or at least checked for compilation errors.

I think, my gut feel is that we probably should try to keep the bulk of docs as close to code as possible (so, lib.rs). If we want to boost the signal for something, it probably is better to mention that from architecture.md document.

test-utils/runtime-tester/README.md
```diff
@@ -0,0 +1,105 @@
...
+#[derive(Serialize, Deserialize, Default, Debug)]
+pub struct BlockStats {
+    pub height: u64,
+    pub block_production_time: Duration,
+}
+```
+
+Be careful to remember, that block height should be positive and ascending.
+
+# Scenario Builder
```
matklad, line 58: Extreme html pedantry: `#` generates a top-level heading `h1`. It is invalid to have more than one `h1`. The correct way is to use `# Moby Dick` for the title of the document, and `## Loomings` for the title of top-level chapter.

test-utils/runtime-tester/README.md
```diff
@@ -0,0 +1,105 @@
...
+
+    pub fn add_transaction(&mut self, signer_index: usize, receiver_index: usize, actions: Vec<Action>)
+
+    pub fn scenario(&self) -> &Scenario
+}
+```
+`new` -- Creates builder with an empty scenario with `num_accounts` accounts.
+`add_block` -- Adds empty block to the scenario with the next height (starting from 1).
+`add_transaction` -- Adds transaction to the last block in the scenario.
+`scenario` -- Returns a reference to the built scenario.
```
matklad, line 75: These are API-level docs, they definitely should be specified as doc-comments in code:

```rust
impl ScenarioBuilder {
    /// Adds transaction to the last block in the scenario.
    pub fn add_transaction(
        &mut self,
        signer_index: usize,
        receiver_index: usize,
        actions: Vec<Action>,
    ) {
}
```

The reasoning is that docs and code gets desynchronized in no-time. As this is an internal tool, and we don't have a dedicated tech-writer to make sure it is up-to-date, it is almost guaranteed that this form of the docs becomes misleading very soon. 

The trick is to document as little as you can, and keep the docs as close to the thing they describe as possible.

test-utils/runtime-tester/src/lib.rs
```diff
@@ -5,8 +5,10 @@
 //! fuzzing provides Arbitrary trait for Scenario, thus enabling creating random scenarios.
 pub mod fuzzing;
 pub mod run_test;
+pub mod scenario_builder;
 
 pub use crate::run_test::{BlockConfig, NetworkConfig, Scenario, TransactionConfig};
+pub use crate::scenario_builder::ScenarioBuilder;
```
matklad, line 11: This is a minor code smell -- now the caller can use `ScenarioBuilder` as either `runtime_tested::ScenarioBuilder` or `runtime_tester::scenario_builer::ScenarioBuilder`, and it's not immediately clear from the code which variant is preferred. 

I usually prefer to make API with only one possilbe import path. For smaller crates, it's usually best for the caller if all the API is available at root, and all the internals are private. 

But that's a minor point, the code looks ok as is!

test-utils/runtime-tester/src/scenario_builder.rs
```diff
@@ -0,0 +1,70 @@
...
+use std::str::FromStr;
+
+pub struct ScenarioBuilder {
+    height: BlockHeight,
+    nonce: Nonce,
+    scenario: Scenario,
+}
+
+impl ScenarioBuilder {
+    pub fn new(num_accounts: usize, use_in_memory_store: bool) -> Self {
```
matklad, line 17: I'd suggest the following AP:

```rust
let builder = ScenarioBuilder::new() // some default number of accounts here
  .number_of_accounts(92) // in case we want to change it
  .in_memory_store(false); // or to use a different store
```

This is a tool for testing and investigation, and here it pays off to make API frictionless.

matklad (commented):
LGTM overall!

I like the docs, but I think they might not survive the merciless grind of time in the current form :-) I suggest pushing most of them into the API comments and [doc tests](https://doc.rust-lang.org/rustdoc/documentation-tests.html)

PR #4817, opened 2021-09-14
refactor ShardTries

mzhangmzz:
This PR refactors the implementation of ShardTries to make it more comply with the general rust design guideline such as interior mutability. It makes two changes
- remove the Arc of individual fields in ShardTries and Arc the whole ShardTries struct instead
- hide the RwLock inside ShardTries so the callers of ShardTries do not need to worry about locking (interior mutability)

### DIFF
--- a/core/store/src/trie/shard_tries.rs
+++ b/core/store/src/trie/shard_tries.rs
@@ -1,5 +1,5 @@
 use std::rc::Rc;
-use std::sync::Arc;
+use std::sync::{Arc, RwLock};
 
 use borsh::BorshSerialize;
 use near_primitives::borsh::maybestd::collections::HashMap;
@@ -13,17 +13,27 @@ use near_primitives::types::{
 
 use crate::db::{DBCol, DBOp, DBTransaction};
 use crate::trie::trie_storage::{TrieCache, TrieCachingStorage};
-use crate::trie::TrieRefcountChange;
+use crate::trie::{TrieRefcountChange, POISONED_LOCK_ERR};
 use crate::{StorageError, Store, StoreUpdate, Trie, TrieChanges, TrieUpdate};
 
-#[derive(Clone)]
-pub struct ShardTries {
-    pub(crate) store: Arc<Store>,
+struct ShardTriesInner {
+    store: Arc<Store>,
     /// Cache reserved for client actor to use
-    pub(crate) caches: Arc<HashMap<ShardUId, TrieCache>>,
+    caches: RwLock<HashMap<ShardUId, TrieCache>>,
     /// Cache for readers.
-    pub(crate) view_caches: Arc<HashMap<ShardUId, TrieCache>>,
+    view_caches: RwLock<HashMap<ShardUId, TrieCache>>,
 }
+
+#[derive(Clone)]
+pub struct ShardTries(Arc<ShardTriesInner>);
+
+impl PartialEq for ShardTries {
+    fn eq(&self, other: &Self) -> bool {
+        Arc::ptr_eq(&self.0, &other.0)
+    }
+}
+
+impl Eq for ShardTries {}
 
 impl ShardTries {
     fn get_new_cache(shards: &[ShardUId]) -> HashMap<ShardUId, TrieCache> {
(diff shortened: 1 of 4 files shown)


nikurt:
Can you please update the PR description to clarify why this refactoring is needed and what it does?

core/store/src/trie/shard_tries.rs
```diff
@@ -81,12 +91,12 @@ impl ShardTries {
     }
 
     pub fn get_store(&self) -> Arc<Store> {
```
matklad, line 93: I'd chage this to 

```rust
pub(crate) fn store(&self) -> &Store {
```

It seems wrong that we have access to the whole `Store` from just `ShardTries`. It seems all call-sites outside of this crate can get access to store from elsewhere.

core/store/src/trie/shard_tries.rs
```diff
@@ -13,18 +13,28 @@ use near_primitives::types::{
...
-#[derive(Clone)]
-pub struct ShardTries {
-    pub(crate) store: Arc<Store>,
+struct ShardTriesInner {
+    store: Arc<Store>,
     /// Cache reserved for client actor to use
-    pub(crate) caches: Arc<HashMap<ShardUId, TrieCache>>,
+    caches: RwLock<HashMap<ShardUId, TrieCache>>,
     /// Cache for readers.
-    pub(crate) view_caches: Arc<HashMap<ShardUId, TrieCache>>,
```
matklad, line 25: :+1: for reducing visibility! It's so much easier to understand the code when you know that random things don't poke internals of structs!

core/store/src/trie/shard_tries.rs
```diff
@@ -13,18 +13,28 @@ use near_primitives::types::{
...
+pub struct ShardTries(Arc<ShardTriesInner>);
+
+impl PartialEq for ShardTries {
+    fn eq(&self, other: &Self) -> bool {
+        Arc::ptr_eq(&self.0, &other.0)
+    }
 }
 
+impl Eq for ShardTries {}
+
```
matklad, line 37: This looks a bit odd. Identity semantics is at times very useful in Rust, but it usually indicates that something really important is going on, and that the caller is supposed to actively use the identity by, eg, storing things in a hash map. As there's only one call-site which makes use of this, I suggest starting with special casing instead of `impl Eq`:

```rust
impl ShardTries {
  pub fn is_same(&self, other: &ShardTries) -> bool
}
```

core/store/src/trie/shard_tries.rs
```diff
@@ -81,12 +91,12 @@ impl ShardTries {
...
     }
 
     pub fn update_cache(&self, transaction: &DBTransaction) -> std::io::Result<()> {
-        let mut shards = self
-            .caches
+        let caches = self.0.caches.read().expect(POISONED_LOCK_ERR);
+        let mut shards = caches
             .iter()
             .map(|(shard_uid, ..)| (*shard_uid, Vec::new()))
             .collect::<HashMap<_, _>>();
```
matklad, line 102:
```suggestion
        let caches = self.0.caches.read().expect(POISONED_LOCK_ERR);
        let mut shards = caches
            .keys()
            .map(|&shard_uid| (shard_uid, Vec::new()))
            .collect::<HashMap<_, _>>();
```

core/store/src/trie/shard_tries.rs
matklad: But maybe that's best left to a follow up PR!

matklad (approved):
LGTM, but the PR would indeed from describing why we are doing the refactor

core/store/src/trie/shard_tries.rs
mzhangmzz: Agreed

Longarithm (approved):

core/store/src/trie/shard_tries.rs
```diff
@@ -13,18 +13,20 @@ use near_primitives::types::{
...
 use crate::trie::trie_storage::{TrieCache, TrieCachingStorage};
-use crate::trie::TrieRefcountChange;
+use crate::trie::{TrieRefcountChange, POISONED_LOCK_ERR};
 use crate::{StorageError, Store, StoreUpdate, Trie, TrieChanges, TrieUpdate};
 
-#[derive(Clone)]
-pub struct ShardTries {
-    pub(crate) store: Arc<Store>,
+struct ShardTriesInner {
+    store: Arc<Store>,
```
bowenwang1996, line 20: Do we need this Arc here?

core/store/src/trie/shard_tries.rs
```diff
@@ -35,23 +37,26 @@ impl ShardTries {
...
-            caches: Arc::new(Self::get_new_cache(&shards)),
-            view_caches: Arc::new(Self::get_new_cache(&shards)),
-        }
+            caches: RwLock::new(Self::get_new_cache(&shards)),
+            view_caches: RwLock::new(Self::get_new_cache(&shards)),
+        }))
+    }
+
+    pub fn is_same(&self, other: &Self) -> bool {
+        Arc::ptr_eq(&self.0, &other.0)
```
bowenwang1996, line 48: Probably better to implement `PartialEq` for `ShardTries`?

bowenwang1996 (approved):

core/store/src/trie/shard_tries.rs
matklad: I'd rather not: https://github.com/near/nearcore/pull/4817#discussion_r708279480

`PartialEq` signals -- "hey, you are supposed to `==` this relatively frequently". For our use-case, we have only a single debug-assertion that verifies identity -- it's better if that use-case stands out.

In a similar vein, if you have only a *single* place where you convert `Foo` to `Bar` it might be better to add an inherent method for that, instead of implementing a blanket `From<Foo> for Bar`. Single use-case is probably contextual (it makes sense to Foo -> Bar in *this specific* context), but the impl is global (under any circumstances, it's ok to turn Foo into Bar).

core/store/src/trie/shard_tries.rs
matklad: Going to be removed in https://github.com/near/nearcore/pull/4817#discussion_r708279480 I suppose.

core/store/src/trie/shard_tries.rs
bowenwang1996: I see. Thanks for the explanation!

PR #4818, opened 2021-09-15
chore: move MIT/Apache from root and link to relevant folders

### LINKED ISSUE #4798: Move MIT and Apache licenses to their crates folders
We need to avoid any confusion of MIT or APACHE license files being attributed to the entire nearcore. It is safer if we move them to the corresponding crates, even if it means we need to copy the files multiple times. 

This is regarding https://github.com/near/nearcore/pull/4732

austinabell:
Closes #4798 

Going to assume this solves this issue, but if there is any preference to actually copy files to each instead of symlinking them, I'm happy to switch this


bowenwang1996 (commented):
> if there is any preference to actually copy files to each instead of symlinking them, I'm happy to switch this

This is probably what we want. cc @nearmax

matklad:
cargo also symlinks the licenses, so it should be fine I belive: https://github.com/rust-lang/cargo/tree/master/crates/crates-io. When you publish a crate, cargo correctly follows symlinks (you can check this by running `cargo package`, and then extracting the resulting traball in the traget directory)

EgorKulikov (approved):

matklad (approved):

bowenwang1996 (approved):

frol (approved):

Issue #4159, opened 2021-03-24
View structs should not be used on the protocol level


bowenwang1996:
All the view structs `core/primitives/src/views.rs` should only be used for rpc purposes and should not be used on the protocol level for other things. However, today `EpochSyncResponse` uses `LightClientBlockView`.

PR #4819, opened 2021-09-15
Make PeerMessage:Routed hashable in preparation for next changes

pmnoxx:
This change will allow to compute hash of `PeerMessage:Routed`. In the next change we will add dropping duplicates of messages with the same hash.


bowenwang1996 (approved):
It is really unfortunate that we still include view structs in network messages, but sadly that is a known issue https://github.com/near/nearcore/issues/4159

bowenwang1996:
Do we still need this PR @pmnoxx ? If not let's close it

PR #4820, opened 2021-09-15
Drop duplicated routed messages

pmnoxx:
We should drop duplicated routed messages.


pmnoxx:
I think computing hash of message body is not necessary, but I can add it. What do you think @bowenwang1996

chain/network/src/peer.rs
```diff
@@ -731,6 +742,18 @@ impl StreamHandler<Result<Vec<u8>, ReasonForBan>> for Peer {
                 return;
             }
         };
+        // Drop duplicated messages routed within 1s
```
bowenwang1996, line 745: one second is too much. For example, I think we re-request chunk parts every 200ms. I suggest we use a smaller number like 100ms for example

chain/network/src/peer.rs
```diff
@@ -731,6 +742,18 @@ impl StreamHandler<Result<Vec<u8>, ReasonForBan>> for Peer {
...
+        if let PeerMessage::Routed(msg) = &peer_msg {
+            let key = (msg.author.clone(), msg.target.clone(), msg.signature.clone());
+            let now = Utc::now();
+            if let Some(time) = self.routed_message_cache.cache_get(&key) {
+                if time.signed_duration_since(now).num_seconds() >= 1 {
```
bowenwang1996, line 750: should this be `<=` instead of `>=`? Also we should make this number a constant

bowenwang1996 (commented):
Please add a test

chain/network/src/peer.rs
pmnoxx: I'm ok changing it to 100ms, we can adjust later based on real world data.

chain/network/src/peer.rs
pmnoxx: fixed

chain/network/src/peer.rs
```diff
@@ -193,6 +202,8 @@ pub struct Peer {
     peer_counter: Arc<AtomicUsize>,
     /// The last time a Epoch Sync request was received from this peer
     last_time_received_epoch_sync_request: Instant,
+    /// Cache of recently routed messages, this allows us to drop duplicates
+    routed_message_cache: SizedCache<(PeerId, PeerIdOrHash, Signature), DateTime<Utc>>,
```
matklad, line 206: suggestion: use `std::time::Instant` here rather than `DateTime<Utc>`. Or is there some reason why we need calendar (as opposed to monotonic) time here? Calendar time is a pain -- it can go backwards, and you can be tricked into thinking wrong about calendar time...

chain/network/src/peer.rs
```diff
@@ -731,6 +743,20 @@ impl StreamHandler<Result<Vec<u8>, ReasonForBan>> for Peer {
                 return;
             }
         };
+        // Drop duplicated messages routed within 1s
+        if let PeerMessage::Routed(msg) = &peer_msg {
+            let key = (msg.author.clone(), msg.target.clone(), msg.signature.clone());
+            let now = Utc::now();
```
matklad, line 749: Separately from the calendar-clock vs monotonic clock issue, it feels a bit odd that we just reach out for *current* time in the middle of computation. This makes such code difficult to test, as you can't trivially change the time. I would probably expect the timestamp of receiving to be just a part of the message, and used throughout the code. 

But this is just a chache, so it's also not a big deal if we are not strict about time.

core/crypto/src/signature.rs
```diff
@@ -668,6 +668,15 @@ pub enum Signature {
     SECP256K1(Secp256K1Signature),
 }
 
+impl Hash for Signature {
+    fn hash<H: Hasher>(&self, state: &mut H) {
+        match self {
+            Signature::ED25519(sig) => sig.to_bytes().hash(state),
```
matklad, line 674: Unfortuate that we need to by hand this, but seems like that's just the current state: https://github.com/dalek-cryptography/ed25519-dalek/issues/52

chain/network/src/peer.rs
```diff
@@ -687,7 +699,7 @@ impl StreamHandler<Result<Vec<u8>, ReasonForBan>> for Peer {
             Ok(msg) => msg,
             Err(ban_reason) => {
                 self.ban_peer(ctx, ban_reason);
-                return ();
+                return;
```
matklad, line 702: :+1:

chain/network/src/peer.rs
```diff
@@ -731,6 +743,20 @@ impl StreamHandler<Result<Vec<u8>, ReasonForBan>> for Peer {
...
+            let now = Utc::now();
+            if let Some(time) = self.routed_message_cache.cache_get(&key) {
+                if time.signed_duration_since(now).num_milliseconds() as i128
+                    <= DROP_DUPLICATED_MESSAGES_PERIOD.as_millis() as i128
+                {
+                    debug!(target: "network", "Dropping duplicated message from {} to {:?}", msg.author, msg.target);
+                    return;
+                }
+            }
+            self.routed_message_cache.cache_set(key, now);
```
matklad, line 758: alternative:

```rust
            if let Some(prev) = self.routed_message_cache.cache_set(key, now) {
                if prev.signed_duration_since(now).num_milliseconds() as i128
                    <= DROP_DUPLICATED_MESSAGES_PERIOD.as_millis() as i128
                {
                    debug!(target: "network", "Dropping duplicated message from {} to {:?}", msg.author, msg.target);
                    return;
                }
            }
```

that way we always refresh the timestamp of the message. Not sure if that's better or worse. If it is worse (ie, we do want to send duped messages eventually), it probably is worth a comment.

matklad (changes requested):
I think we should be careful with `DateTime<Utc>` -- knowing which calendar year is outside is a rather big dependency (leap seconds, ntpd, etc), better avoid it if we can.

pmnoxx:
@bowenwang1996 I added a test for dropping duplicated routed messages.

chain/network/src/peer.rs
pmnoxx: @matklad We use DataTime<Utc> a lot of in our code. I'll switch to use Instant

pmnoxx: Thanks, fixed.

chain/network/src/peer.rs
pmnoxx: @matklad Consider a situation when peer `A` sends a message to `B` every 10 ms. 
In my current implementation, we would drop 9 out 10 messages, and send one message every 100-100ms.

In the `alternative` you proposed, we would update the counter every 10ms and the message would never get send for the 2nd time. I think the original approach is more intuitive. @bowenwang1996 what do you think?

chain/network/src/peer.rs
pmnoxx: @matklad Currently, the timestamp is not a part of the message. We can't change the API.

However, we can change our code to use a function to get time, instead of `Instant::now()`. 
For unit tests, we can have api that mocks this function, and changes it's behavior.

integration-tests/tests/network/runner.rs
```diff
@@ -320,25 +356,35 @@ impl StateMachine {
...
+                                        let ping_ok = pings.len() == pings_expected.len()
+                                            && pings_expected.into_iter().all(
+                                                |(nonce, source, count)| {
+                                                    pings.get(&nonce).map_or(false, |ping| {
+                                                        ping.0.source == source
+                                                            && (count.is_none()
+                                                                || count.unwrap() == ping.1)
+                                                    })
+                                                },
+                                            );
```
matklad, line 368: That's a bussy snippet of code, though I don't see a clear way to write this better. Probably OK for tests.

matklad (approved):

pmnoxx:
@frol @bowenwang1996 @mfornet @chefsale @khorolets Can you take a look?

integration-tests/tests/network/runner.rs
```diff
@@ -113,11 +116,16 @@ pub enum Action {
...
+    CheckPingPong(usize, Vec<(usize, usize, Option<usize>)>, Vec<(usize, usize, Option<usize>)>),
     // Send stop signal to some node.
     Stop(usize),
     // Wait time in milliseconds
     Wait(usize),
+    #[cfg(feature = "adversarial")]
+    SetOptions {
+        target: usize,
+        max_num_peers: Option<u64>,
+    },
```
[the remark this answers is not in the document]
pmnoxx: @bowenwang1996 @matklad 
I'm trying to write a code that is supposed to compile if either we are running test or adversarial flag is not set.
The following doesn't compile
```

#[cfg(any(test, feature = "adversarial"))]
pub struct SetAdvOptions {
    pub disable_edge_signature_verification: Option<bool>,
    pub disable_edge_propagation: Option<bool>,
    pub disable_edge_pruning: Option<bool>,
    pub set_max_peers: Option<u64>,
}
```

when running
```
cargo test --all-targets
```

```
no `SetAdvOptions` in `types`
```
Any ideas?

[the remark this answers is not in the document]
pmnoxx: @bowenwang1996 @matklad I see that we run `cargo test` without `--feature adversarial`. My tests depend on having `adversarial` flag enables, which adds extra functionality, like setting flags, to make the test work.

Due to the way `[cfg(test)]` works, it can't be used to mark function or struct, that should be compiled only while running a test. The existing tests we have in `integration-test/network/*.rs` really on using methods from `PeerManager` that are only used for for testing. That's not great, Ideally I would prefer to put all those methods behind a feature flag.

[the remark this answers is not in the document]
pmnoxx: I think that the best solution would be to add `adversarial` flag to CI. Currently `adversarial` provides additional test functionality, but it is only used by python tests.

Of course if we were to simply add `adversarial` to existing 2 test pipelines, you could have a code that work with, but doesn't with adversarial. So the best solution I see so far would be to add 2 extra pipelines:
```
cargo test --features adversarial,nightly_protocol,nightly_protocol_features
cargo test --features adversarial
```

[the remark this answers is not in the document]
pmnoxx: @bowenwang1996 @matklad What do you think?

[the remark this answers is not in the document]
matklad: If we want to have an ability to write adversarial integrated tests in Rust, then we indeed need to add an extra CI pipeline here -- `#[cfg(test)]` only works within a single crate.

However, I want to ask *why* we didn't have a use for such tests before, and why we need to introduce them here. I have two concerns with this:

* First, adding an extra CI pipeline is sort-of a big deal. Granted, we can just dedicate more machines to CI and this shouldn't affect actual merge time, but, still, it increases the overall fragility of the system.
* Second, I think the original idea behind adversarial testing is that you want to spawn a network of nodes, *some* of which are adversarial and some are not, and you want to test that normal nodes are functioning correctly in the presence of adversaries. That is, it's important for adversarial tests that there are several neards running, and that we are testing the one with adversarial features disabled. 

  What we are doing here feels different -- this is a rust integrated test, so there's only one binary, and we are testing an adversarial nodes against itself. That is, to me it makes sense (abstractly), that we should add a new feature flag here, like "extra-test-knobs", and use that. Re-using existing "adversarial" feature for this seems accidental, adversarial feels like it should have a different semantics. 

>The existing tests we have in integration-test/network/*.rs realy on using methods from PeerManager that are only used for for testing. That's not great

I think this approach is actually fine. Sure, exposing extra knobs just for testing *feels* wrong, but I don't think there are actual problems there? I don't think we need to worry about someone accidentally using such a function wrongly, besides clearly labeling them as "for testing", preferably by hiding them in some `mod test_utils` or `trait TestExt`. Though, it might be good to try to minimize the amount of such tweaks we need for tests. 

Another plausible alternative is to use `pub(crate)` for those extra knobs, and move all tests to be unit-tests. Note that this requires that tests don't actually need `nearcore` or other such crates which are not already a dependency of the network.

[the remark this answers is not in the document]
bowenwang1996: > I think the original idea behind adversarial testing is that you want to spawn a network of nodes, some of which are adversarial and some are not

I don't think so. I believe the idea is that some nodes "could be" adversarial, but it doesn't mean that we _necessarily_ have to have normal nodes in a test when adversarial ones are present.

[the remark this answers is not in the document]
bowenwang1996: Also agree with what @matklad said. It makes more sense at this point to go with the existing approach -- exposing test-only structs. While this is not optimal, it is an existing issue and it is better to not pile new problems on top of that.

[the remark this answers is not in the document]
pmnoxx: -  I partially agree with @matklad. However, we currently use `adversarial` already in our python tests that run through `Nayduck`. So by adding extra pipeline for running tests will not make use system more fragile than it already is. It's possible to make code changes that will make `nayduck` fail, but pass CI. I would rather caught issues early while running CI, then spend hours a few days later, trying to figure out which PR broke them.

Cost of additional machines, shouldn't be an issue.
- I don't like the idea of having test methods in production. For purpose of `integration-test/network/*.rs` we introduced new data structures, that are only used by test, but are not read in production. I would prefer to not worry, that a node may crash, or that test data structures may grow in memory usage infinitely large.
There are also other reason why separation of test code is a good idea, it's much easier to read the code when doing PR, when you can guarantee that some branches of the code don't even get executed in producition.
My plan was to add 1-2 more data structures, that would allow us to make testing much easier, but I would rather not have them execute in production.
- Moving to `mod test_utils` would be ideal, but the existing code can't be easily moved there.

FYI tests in ` integration-test/network/*.rs ` spawn multiple nodes. We are actually testing interaction between them.
My intention was to move some of the tests from python to CI. That would make development easier. Test would also run faster. However, we don't have access to all `adversarial` features in CI.

Though, I agree with the naming of `adversarial`. Renaming `adversarial` to `test-features` seems appropriate, we use it for testing only. It's not indented for production.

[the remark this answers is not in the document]
pmnoxx: My proposal is to move all `integration-test/network/*.rs` to be under `adversarial` flag, and all methods they use in `networking`. @bowenwang1996 If that's something that would be accepted I prepare PR.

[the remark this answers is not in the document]
pmnoxx: I created an issue for that: https://github.com/near/nearcore/issues/4893

integration-tests/tests/network/routing.rs
```diff
@@ -274,3 +274,39 @@ fn archival_node() {
 
     start_test(runner);
 }
+
+/// Spawn 3 nodes, add edges to form 0---1---2 line. Then send 2 pings from 0 to 2, one should be dropped.
+#[test]
+#[cfg(feature = "adversarial")]
```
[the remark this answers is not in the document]
pmnoxx: Those tests rely on code, which is under `adversarial`. I either need to make this code always present or rewrite those tests in python.

integration-tests/tests/network/runner.rs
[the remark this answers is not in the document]
bowenwang1996: I see. So you need what is already behind adversarial feature to run the tests. In that case, we should just make sure we can run the tests you write in CI. Otherwise they are even worse than tests run by nayduck -- they would not be run at all.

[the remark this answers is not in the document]
pmnoxx: I created separate PR for renaming `adversarial` to `test_features`. I see that we changed `pipeline.xml` to use `--features adversarial`. Is that enough?

chain/network/src/peer.rs
```diff
@@ -68,6 +71,10 @@ pub const EPOCH_SYNC_REQUEST_TIMEOUT_MS: u64 = 1_000;
 /// How frequently a Epoch Sync response can be sent to a particular peer
 // TODO #3488 set 60_000
 pub const EPOCH_SYNC_PEER_TIMEOUT_MS: u64 = 10;
+/// Limit cache size of 1000 messages
+pub const ROUTED_MESSAGE_CACHE_SIZE: usize = 1000;
+/// Duplicated messages will be dropped if routed through the same peer multiple times.
+pub const DROP_DUPLICATED_MESSAGES_PERIOD: Duration = Duration::from_millis(100);
```
bowenwang1996, line 77: Let's change this to 50ms to err on the side of caution

integration-tests/tests/network/runner.rs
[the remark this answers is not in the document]
bowenwang1996: This code still looks very strange to me. You have an enum that is purely used for testing but you want to gate one specific variant of it instead of the whole thing. Could you explain what is the rationale for that?

chain/network/src/peer.rs
pmnoxx: Done

pmnoxx:
I created a separate PR for moving `runner.rs` under a feature flag. Currently in the master branch this code is still compiler by default. However, `SetOption` uses code, which is under `adversarial`.

pmnoxx:
> This code still looks very strange to me. You have an enum that is purely used for testing but you want to gate one specific variant of it instead of the whole thing. Could you explain what is the rationale for that?

@bowenwang1996 Yes, currently we use `SetAdvOptions` for testing in `PeerManager` for python tests. That code is behind `adversarial`/`test_features` flag. At the same time, `integration-test/network/*.rs` uses test code in `PeerManager`, which is not hidden behind any flags.
That's why I created the following issue: https://github.com/near/nearcore/issues/4893

bowenwang1996:
@pmnoxx what is the issue with not gating that variant for now?

pmnoxx:
> @pmnoxx what is the issue with not gating that variant for now?

What do you mean? We are still gating it. Once we push https://github.com/near/nearcore/pull/4894, everything will be behind the `test_features` flag.

bowenwang1996 (approved):

bowenwang1996:
> @bowenwang1996 What do you mean? We are still gating it. Once we push #4894, everything will be behind the test_features flag.

I looked at #4894 and it does not look like we are doing anything with `Action`?

PR #4823, opened 2021-09-15
chore: update rust version

### LINKED ISSUE #4588: Consider upgrading to the most stable Rust
At the moment, we still use Rust 1.51.0. The latest stable is 1.53.0, and 1.54.0 release is imminent, so we might consider upgrading. 

Some interesting considerations:  

* 1.53.0 includes https://github.com/rust-lang/cargo/pull/9298, which should significantly improve compile times for devs using macs. 
* 1.51 (current version) has broken incremental compilation, 1.52&1.53 have incremental compilation disabled, I *think* 1.54 fixes and re-enabled it. 

I suggest that we wait for 1.54 and switch to that. From then on, it perhaps makes sense to always upgrade to the latest rust. but we might want to do that couple of weeks *after* the release, in case a minor release happens.

matklad:
closes #4588

Test Plan
---------

Existing CI. 1.55 has been stable for some time, there are no known
outstanding issues with this rust version.


olonho (approved):

runtime/near-test-contracts/test-contract-rs/src/lib.rs
```diff
@@ -484,10 +484,8 @@ pub unsafe fn sum_n() {
     for i in 0..n {
         // LLVM optimizes sum += i into O(1) computation, use volatile to thwart
         // that.
-        unsafe {
```
matklad, line 487: this now (correctly) fires a warning on a  new compiler.

bowenwang1996 (approved):

chefsale (approved):

PR #4827, opened 2021-09-16
fix(indexer): Handle restored receipts by attaching them to the chunk

khorolets:


chain/indexer/src/streamer/mod.rs
```diff
@@ -153,7 +154,33 @@ async fn build_streamer_message(
...
+        // protocol version on mainnet. In this block ExecutionOutcomes for restored Receipts appear.
+        // However the Receipts are not included in any Chunk. Indexer Framework needs to include them,
+        // so it was decided to artificially include the Receipts into the Chunk of the Block where
+        // ExecutionOutcomes appear.
+        // ref: https://github.com/near/nearcore/pull/4248
+        if &block.header.hash.to_string() == "ErdT2vLmiMjkRoSUfgowFYXvhGaLJZUWrgimHRkousrK" {
+            let protocol_config =
+                fetchers::fetch_protocol_config(&client, block.header.hash).await?;
+
+            if &protocol_config.chain_id == "mainnet" {
```
bowenwang1996, line 169: Do we need this check? The hash we use is cryptographic

khorolets: We had hash collisions previously, so it's some sort of insurance from the indexer side. Do you insist on dropping this check?

frol: I am not so concerned about the hash collisions (if that happens, we are really doomed), but I still prefer to explicitly communicate the fact that we care about mainnet state here, so in the future we may consider just dropping it.

chain/indexer/src/streamer/mod.rs
```diff
@@ -153,7 +154,33 @@ async fn build_streamer_message(
...
+                    chunk_receipts.iter().map(|receipt| receipt.receipt_id).collect();
+
+                for outcome in receipt_execution_outcomes.into_iter() {
+                    if receipt_ids_included.get(&outcome.receipt.receipt_id).is_none() {
+                        restored_receipts.push(outcome.receipt.clone());
+                    }
+                }
+
+                restored_receipts.extend(chunk_receipts);
+                chunk_receipts = restored_receipts;
```
frol, line 181: @khorolets I have just checked the log that you shared with me, and I see that the order of the execution outcomes (and thus I want to preserve that order for receipts) is the following:

1. local receipts (there are 3 local receipts there)
2. restored receipts (there has to be 383 receipts, but only 329 have execution outcomes!)
3. delayed receipts (there are 9 of them)

Thus, we have 2 problems:
1. Where are the execution outcomes for 54 more restored receipts? Are they in the next block?
2. The execution outcomes store the values sequentially, so there is no need to use hashmap here. If you just slice execution outcomes as by skipping 3 local receipts at the beginning and 9 delayed receipts at the end, you get the list of the execution outcomes produced by the restored receipts (yet, we miss 54 restored receipts anyway, so we should first investigate what is going on there)

Notes:

[restored-receipts-outcomes.txt](https://github.com/near/nearcore/files/7185198/restored-receipts-outcomes.txt)

```
$ grep '"receipt_id":' nearcore/res/mainnet_restored_receipts.json | grep -oP ': ".*' | grep -oP '[^: ",]*' > expected-restored-receipt-ids.txt
$ grep receipt_id: restored-receipts-outcomes.txt | grep -oP ': `.*' | grep -oP '[^: `,]*' > restored-receipt-outcome-ids.txt
```

[restored-receipt-outcome-ids.txt](https://github.com/near/nearcore/files/7185196/restored-receipt-outcome-ids.txt)
[expected-restored-receipt-ids.txt](https://github.com/near/nearcore/files/7185197/expected-restored-receipt-ids.txt)

```python
with open('restored-receipt-outcome-ids.txt') as f:
    execution_outcomes_for_receipt_ids = f.read().split()

with open('expected-restored-receipt-ids.txt') as f:
    expected_to_be_restored = f.read().split()

print(f'Number of restored receipts: {len(expected_to_be_restored)}')
print(f'Number of execution outcomes in the block: {len(execution_outcomes_for_receipt_ids)}')
missing_receipts = set(expected_to_be_restored) - set(execution_outcomes_for_receipt_ids)
print(f'There are {len(missing_receipts)} missing execution outcomes for restored receipts: {missing_receipts}')
```

Output:

```
Number of restored receipts: 383
Number of execution outcomes in the block: 341
There are 54 missing execution outcomes for restored receipts: {'6Pbv7vTfnWc8hZHwXKPv3m2uGQyCK1qYKKUg6MrvCtgq', '93PNwuWnLQSAfw5zx5YQUnbS4vddsJuBwNBSj5FJrq9g', '8MFfyUY4YVzERq69UY5vV8NzaEugBG25LaTJNpye8HyL', '2G3geRWVnRV9eAGgk5xssPBXXRyKvYgvGdghJRJhRTwV', '2CXM88aiSZgaMT6ikP9fTYj6fxxGX3uUwW5kgncwTdeF', 'BcHYZhXM1QY4gcEkPN9bKywSPpSiNi1eU8hkTHRuoSKY', 'B53ZWDd84EMRASB53FRFrUT3R3Z7ucZkBLwfzts7myxz', '8imDisv6DPXa8YWNztgYWXM4M3F9kkHBRipUui1t2zrz', '7eCy6LsyTn46Zw6zHgC4AEgD6KsuJFVPy7pGET1kgrvE', '24sdwPdfYLuWwCNEA7B3wkgZGPxtFH6gpSFAnSQHqsEe', '3eLrWheRxChogkdUVbBsQfWrt3hGop3p4YEfpBnkXdY7', '4WAW8jWocqJdgjgNHQkvBmtmU5u4MSjDyMd3HvgvhPUJ', 'GSXVxk2HtAPeR4xumXW26Jcy5MmedxKrg5dWAyo1ytaK', 'DAHX7gRMSyX7Ni9RQYARdfZegnCYXqGhHCKJRymC8v1F', '4bPpcLBgjt9iNwYAyWpSCFTgNH5CHLMvqJ1sLmKmi5mC', 'FEi3W4tG6MfAVPnFWnEj9WDcTUJH5tDAurLgM8UG64K5', '6zNuWyDqmYkaRTktT8iCWpbgUWjXcSvDcJbxq9N4ctX9', 'Fs4TWKMVNoWZ5roWAM4TC1bgX7NQrKqJ7k8qFNTfQedf', 'HaDECLKDQYvtjnGkbAJjU1Mgrdvv141yRE6CGzDDi5B2', '9rTpUwYqQyWQg6BXDAUabwuZqFTYoRY18gEMtL87AHJ', '8dxnRTnXaY8SCZ4BDHEyMzLgKYYRpAZRrnnohyD2bds5', 'k8AWZp6v3oy1MB2iAYWcXoUHKrdivaZTPdJnP7gt5YY', 'G8YWiNSJcQX7t3SYtqAWTZzWy5RmhkbNoKzv1FigwRU6', '3RgTMGUKLxZF6BJ6zhwTVgPM3MvykAoSiS9MPKWwEgeR', 'E4BzkPXBLLvzCGD1DVGpzwjTUoXQSuRqcTZqCvAuz73m', '7U4xv5UuKc4YR8GEStsYH3yHYPwfwe6xDVLb7d32GRM', '44PSDS2kSMyHRLpgQfMR8uNTEA8DrKPyH7LMCUdef5oK', 'FRcUo5JcqWRrUH6wfCE5KzJzyNyCtK1Rd9YXQrtu7MjN', 'HJ8v7EdZ5QVxpsr2j3B1hJrS6fu5e93aGcCVfbHQb7Sk', '9A9j7Zc5rCaUfRUz9uiiyz1Eqhd2LTAWfsxYPxw1NxEr', '2gBBVqMZQE3Q2UnbXXPwicUCt8XvzzzpFvwvVk7CLCAM', '9WFY4HnxnNYFntLc4eGsDDzoPsHqUaKFeJmjJvBK2Ncr', '6jXkb2hcgFh9kF1a9eFBrkyfvW3uHXC7rTVFuNWgPz5m', 'BQViHrVcV7YA9jLzLtwFakF2iGSRkDsAxHYiX5SpwWeQ', '8fcSTmLEj1XKpWAZfX9KiMTvjy1qYV74PDLUsXiJj77j', '3KQJRDQn6s7mw4dv9fEWsyKUeXYZmLw3xHvub2EF3kXV', 'Bt67WMrbX1FmTX7hF7RVzRr2cXmqLcB4MtL9b1ELNksB', '3zUPNxLXpUSt9EBBdNSKe8UfYcqgtQoaqtvr7E1tNywR', '8hWafkKaMig2CFqfvEheLShSsKx471ZAjiB1VGJscGbQ', '4ZYG8PZUvp4ciwUJHvefshUpPcJeKkRwRyjWwjCtqxGr', '7qCe9WqqSv4nuqzMnn3HteapyCQ2hcgcxydtGbqCTtez', 'ChJvLzqmGmtj6Y7SVPAMJyF37npQegySBmpDjahpnbV2', '7u5jVvybKFPvP6BJMD5ZctQNKqMkamt8VZVEGvJCdVVB', 'AV1FeCTf3NgBAYeqk4SR35DaudMbdYG8ktMFPycmi9FE', 'A55KHCTQyFNe6rZhEnK6V6H7k7c5XwHXvLNw9oKxTzHc', 'A4PpbYGxs2wAvXHjmDaQ55JfoBw6zUobMS8S2mJkWCxD', 'CLfnEkZZ3nafgg26Pr7XYKa2seZKG7Q9KvHKSdsNkyFb', '49RbmebsToNvfHxwBbfjESwyJaRwivz1HMVdCGBHkUtg', 'BDDm9G9hqWqpZyn6o3Nhvjz7uAnhcGtosFjAB9414LUG', 'AsGiozZga53PsZ7PLDMjEcfCsNL1sjPxZpwK6K1tzyvh', '7FD8fcXzsvoxPwfkLFHrLEiPLUKLoERCpdoPD61Va9rD', '7qTxqgWY7EcNGsNdFAAY5xbwLks4UrfqgCDWtSbw3dm', 'GkKaXeNhGHoB3bmMB6KxrWhhUmxPxRnmpajdRQSEVdMU', '3SCbhx5E5hiESF5wJ5Nvo7gUba2ghxzHHkxvnHPH7SWT'}
```

frol (changes requested):

chain/indexer/src/streamer/mod.rs
Longarithm: @frol did you get this data using `state-viewer` or some other tool?

frol: @Longarithm I got the initial data from the Indexer Framework which uses view-client calls to `GetChunk` and `GetExecutionOutcomesForBlock`. Get in touch with @khorolets on more details.

Longarithm: @frol The reason is found. I expected that all restored receipts will be included in one block, but they didn't fit, so they were applied in two sequential blocks.

Will it be OK to include block 47317864 with hash `2Fr7dVAZGoPYgpwj6dfASSde6Za34GNUJb4CkZ8NSQqw` here?

I've checked that this is enough for commit https://github.com/near/nearcore/compare/88a9e74...bc3256a by running `./target/debug/state-viewer apply_range --start_index 47317863 --end_index 47317864`

khorolets: [63-64.log](https://github.com/near/nearcore/files/7283546/63-64.log)

Here's the `StreamerMessage` for 47317863 and 47317864 blocks. I've checked with the @frol script above and I have the output:

```
Number of restored receipts: 383
Number of execution outcomes in the block: 518
There are 0 missing execution outcomes for restored receipts: set()
```

frol (approved):

PR #4840, opened 2021-09-17
near-test-contracts: make sure RUSTFLAGS env. variable takes effect

mina86:
If CARGO_ENCODED_RUSTFLAGS environment variable is set Cargo uses its
value in preference to RUSTFLAGS variable.  As such, to make sure the
latter takes effect the former needs to be unset.


matklad:
TIL about CARGO_ENCODED_RUSTFLAGS!

matklad (approved):

bowenwang1996 (approved):

PR #4843, opened 2021-09-18
fix(script): build sandbox when dir not exist

### LINKED ISSUE #4799: `make sandbox` fails at the final step due to missing `target/debug` directory
**Describe the bug**
Following the steps at https://docs.near.org/docs/develop/contracts/sandbox fails because the `target` directory doesn't exist in `nearcore` so the last step of `make sandbox` fails like so:
```
mv sandbox/debug/neard target/debug/near-sandbox
mv: rename sandbox/debug/neard to target/debug/near-sandbox: No such file or directory
make: *** [sandbox] Error 1
```
Doing `mkdir -p target/debug` and rerunning `make sandbox` fixes the issue.

**To Reproduce**
1. Clone a clean `nearcore` repo
2. Run `make sandbox`

**Expected behavior**
Everything works and the binary shows up in `target/debug`

Nearcore on commit hash `b84b3fcd`

ailisp:
Fix #4799

Test Plan
------------
make sandbox success when target, target/debug not exist and exist


Makefile
```diff
@@ -66,10 +66,12 @@ nightly-debug:
 
 sandbox:
 	CARGO_TARGET_DIR=sandbox cargo build -p neard --features sandbox
+	mkdir -p target/debug
 	mv sandbox/debug/neard target/debug/near-sandbox
```
bowenwang1996, line 70: Why do we want to move this back and forth? Is it not possible to specify the name of the output binary directly? cc @matklad

matklad: Not really: there's a bunch of hacks, but none is strictly better than an extra mv (we've discussed that a while back with chefsale).

chefsale (approved):

Issue #4745, opened 2021-08-27
Refactor code duplication when creating NightshadeRuntime::new


matklad:
In #4698, there are shotgun changes all other the place to add a new argument to NightshadeRuntime::new:

!<span class="error">[image]</span>([https://user-images.githubusercontent.com/1711539/131116783-fe809221-f28a-4f0c-9577-c259917afba6.png](https://user-images.githubusercontent.com/1711539/131116783-fe809221-f28a-4f0c-9577-c259917afba6.png))

I think this is about the third time I notice this effect. Seems it's rather high priority to fix it, as it is a persistent productivity drain. Many duplicates originate in tests, so fixing those seems like an easy win. Fo example, the following three tests contain identical setup:

[https://github.com/near/nearcore/blob/13141b4b78b402734a83df41fc4b7feb5b7ced01/integration-tests/tests/client/process_blocks.rs#L3475-L3729](https://github.com/near/nearcore/blob/13141b4b78b402734a83df41fc4b7feb5b7ced01/integration-tests/tests/client/process_blocks.rs#L3475-L3729)

The general approach here is to move from "each tests calls the APIs directly" to "there's a single driver function which takes an input and uses it to drive the API, and test differ only in the input specified". 

 See previous discussion <span class="error">[here]</span>([https://github.com/near/nearcore/pull/4287#discussion_r630982585](https://github.com/near/nearcore/pull/4287#discussion_r630982585)), and the data driven testing section of <span class="error">[this post]</span>([https://matklad.github.io/2021/05/31/how-to-test.html#data-driven-testing](https://matklad.github.io/2021/05/31/how-to-test.html#data-driven-testing)). 

Specific steps:

*   look at #4698 closely and identify the biggest offenders in terms of code duplication
*   if they are indeed tests, consider the best way to remove duplication, which might include (but is not limited to):
*   mechanical extraction of common code into helper function
*   larger scale refactoring of a set of tests into a "driver + test cases" mode
*   now that many of our tests are in the single integration-tests crate, take a broader look at them and try to come up with universal testing framework

matklad:
`apply_transaction` seems to suffer from a similar issue: https://github.com/near/nearcore/pull/4847#discussion_r712962554

PR #4847, opened 2021-09-20
Apply block updates to split shards

mzhangmzz:
This PR makes block updates and catchups also update split states for the next epoch.

The algorithm works like follows. 
There are two possibilities:
1) States for next epoch are not ready when a block is processed. In this case, `apply_chunks_preprocessing` will be called first with `ApplyChunksMode::NotCaughtUp` and then `ApplyChunksMode::CatchingUp`. With `NotCaughtUp`, if a shard will be split
into multiple shards in the next epoch and the validator cares about one of the split shards, the validator stores the state changes in the database through a new column `ConsolidatedStateChanges`. Later, when catching up blocks, `apply_chunks_preprocessing` with `CatchingUp` will read the stored state changes and process them.
Note: we cannot use the existing stored `state_changes` or `trie_changes` for updating split states. `trie_changes` are updates on trie nodes and trie structure of the old and new states are different.  The existing `state_changes` do not include updates on internal states such as postponed receipts, delayed receipts, etc..
2) States for next epoch are ready. In this case, `apply_chunks_preprocessing` is only called once with `ApplyChunksMode::CaughtUp`. `apply_transactions` can update the states for shard in this epoch and the split shards in next epoch together


frol (commented):
I read through the PR and realized that there is nothing that I can efficiently review here without spending too much time on it. I defer my review vote to other reviewers who understand the problem space and the code base around those places much better than I do. Ping me if my vote would block you from landing this PR.

mzhangmzz:
> I read through the PR and realized that there is nothing that I can efficiently review here without spending too much time on it. I defer my review vote to other reviewers who understand the problem space and the code base around those places much better than I do. Ping me if my vote blocks you from landing this PR.

Thanks @frol!

chain/chain/src/chain.rs
```diff
@@ -2877,6 +2887,28 @@ impl<'a> ChainUpdate<'a> {
...
+            shard_id
+        ));
+        let state_roots: Result<Vec<_>, _> = new_shards
+            .iter()
+            .map(|shard_uid| {
+                self.chain_store_update
+                    .get_chunk_extra(block.header().prev_hash(), shard_uid)
+                    .map(|chunk_extra| chunk_extra.state_root().clone())
+            })
+            .collect();
```
bowenwang1996, line 2908: Do we store chunk extra only to get the state roots? Feels hacky to me :(

chain/chain/src/chain.rs
```diff
@@ -3088,19 +3144,79 @@ impl<'a> ChainUpdate<'a> {
...
+        apply_split_state_results: Option<Vec<ApplySplitStateResult>>,
+        state_changes_to_add: Option<ConsolidatedStateChanges>,
+    ) {
+        if let Some(results) = apply_split_state_results {
+            for result in results {
+                // Save chunk extras for the split states
+                self.chain_store_update.save_chunk_extra(
+                    &block_hash,
+                    &result.trie_changes.shard_uid,
+                    ChunkExtra::new(&result.new_root, CryptoHash::default(), vec![], 0, 0, 0),
```
bowenwang1996, line 3204: It probably makes sense to at least make this a separate function

chain/chain/src/chain.rs
```diff
@@ -3088,19 +3144,79 @@ impl<'a> ChainUpdate<'a> {
...
+                );
+                self.chain_store_update.save_trie_changes(result.trie_changes);
+            }
+            // should not store state_changes after state results have already been processed
+            assert!(!state_changes_to_add.is_some());
```
bowenwang1996, line 3209:
```suggestion
            assert!(state_changes_to_add.is_none());
```

chain/chain/src/chain.rs
```diff
@@ -2898,33 +2930,56 @@ impl<'a> ChainUpdate<'a> {
...
+                // catching up next epoch's shard states, do not update this epoch's shard state
+                // since it has already been updated through ApplyChunksMode::NotCaughtUp
+                ApplyChunksMode::CatchingUp => {
+                    !cares_about_shard_this_epoch && cares_about_shard_next_epoch
+                }
             };
+            // For splitting states, there are also two possibilities
+            // 1) states for next epoch are not ready when the block is processed
+            //    In this case, apply_chunks is first called with ApplyChunksMode::NotCaughtUp and
+            //    then ApplyChunksMode::CatchingUp. With NotCatchUp, state changes will be stored
```
bowenwang1996, line 2959:
```suggestion
            //    then ApplyChunksMode::CatchingUp. With NotCaughtUp, state changes will be stored
```

chain/chain/src/chain.rs
```diff
@@ -2898,33 +2930,56 @@ impl<'a> ChainUpdate<'a> {
...
+                ApplyChunksMode::NotCaughtUp => cares_about_shard_this_epoch,
+                // update both this epoch and next epoch
+                ApplyChunksMode::IsCaughtUp => {
+                    cares_about_shard_this_epoch || cares_about_shard_next_epoch
+                }
+                // catching up next epoch's shard states, do not update this epoch's shard state
+                // since it has already been updated through ApplyChunksMode::NotCaughtUp
+                ApplyChunksMode::CatchingUp => {
+                    !cares_about_shard_this_epoch && cares_about_shard_next_epoch
+                }
```
bowenwang1996, line 2954: I am a bit confused by this. My understanding is that even if a node tracks the shard in the epoch before the split, it will still go through this phase due to the time it takes to split the state. However, `cares_about_shard_this_epoch` is true in this case.

chain/chain/src/chain.rs
```diff
@@ -3109,13 +3225,10 @@ impl<'a> ChainUpdate<'a> {
...
         match result {
-            ApplyChunkResult::SameHeight(SameHeightResult {
-                shard_id,
-                gas_limit,
-                apply_result,
-            }) => {
+            ApplyChunkResult::SameHeight(SameHeightResult { gas_limit, apply_result }) => {
                 let (outcome_root, outcome_paths) =
                     ApplyTransactionResult::compute_outcomes_proof(&apply_result.outcomes);
+                let shard_id = apply_result.trie_changes.shard_uid.shard_id();
```
bowenwang1996, line 3231: this way of getting shard id is a bit strange. What is the reason for this?

nearcore/src/runtime/mod.rs
```diff
@@ -528,19 +534,45 @@ impl NightshadeRuntime {
...
+                        consolidated_state_changes,
+                    )?);
+                    (None, split_state_results)
+                } else {
+                    // split states are not ready yet, store state changes in consolidated_state_changes
+                    (Some(consolidated_state_changes), None)
+                }
+            } else {
+                (None, None)
+            };
```
bowenwang1996, line 567: It might make sense to refactor this out into its own function

bowenwang1996 (commented):
Generally looks good! Though I still could not fully follow all the logic and it would be better if there is some sanity test :)

chain/chain/src/chain.rs
```diff
@@ -2877,6 +2887,28 @@ impl<'a> ChainUpdate<'a> {
...
+        &mut self,
+        block: &Block,
+        shard_id: ShardId,
+    ) -> Result<HashMap<ShardUId, StateRoot>, Error> {
+        let next_shard_layout =
+            self.runtime_adapter.get_shard_layout(block.header().next_epoch_id())?;
+        let new_shards = next_shard_layout.get_split_shards(shard_id).expect(&format!(
+            "shard layout must contain maps of all shards to its split shards {}",
+            shard_id
+        ));
```
matklad, line 2900: nitpick: this will allocate an error string message even on the happy path. The more frugal way to write this would be

```rust
        let new_shards = next_shard_layout.get_split_shards(shard_id).unwrap_or_else( || panic!(
            "shard layout must contain maps of all shards to its split shards {}",
            shard_id
        ));
```

chain/chain/src/chain.rs
```diff
@@ -2877,6 +2887,28 @@ impl<'a> ChainUpdate<'a> {
...
+        let state_roots: Result<Vec<_>, _> = new_shards
+            .iter()
+            .map(|shard_uid| {
+                self.chain_store_update
+                    .get_chunk_extra(block.header().prev_hash(), shard_uid)
+                    .map(|chunk_extra| chunk_extra.state_root().clone())
+            })
+            .collect();
+        Ok(new_shards.into_iter().zip(state_roots?.into_iter()).collect())
```
matklad, line 2909: I have a knee-jerk reaction to `.zip` -- it often hides bugs when the two iterators are of different lengths. In this case, I think it's better to express same length-ness at compile time. This also saves intermediate vec allocation: 

```suggestion
        new_shards
            .into_iter()
            .map(|shard_uid| {
                let chunk_extra = self
                    .chain_store_update
                    .get_chunk_extra(block.header().prev_hash(), &shard_uid)?;
                Ok((shard_uid, *chunk_extra.state_root()))
            })
            .collect()
```

chain/chain/src/types.rs
```diff
@@ -88,6 +93,8 @@ pub struct ApplyTransactionResult {
     pub total_gas_burnt: Gas,
     pub total_balance_burnt: Balance,
     pub proof: Option<PartialStorage>,
+    pub split_state_apply_results: Option<Vec<ApplySplitStateResult>>,
+    pub consolidated_state_changes: Option<ConsolidatedStateChanges>,
```
matklad, line 97: Looking at the assert in `process_split_state_result`, is it true that at most one of these fields can be non-None? If thats the case, I suggest strongly-typing this: 

```rust
pub struct ApplyTransactionResult {
    pub trie_changes: WrappedTrieChanges,
    pub new_root: StateRoot,
    pub outcomes: Vec<ExecutionOutcomeWithId>,
    pub receipt_result: ReceiptResult,
    pub validator_proposals: Vec<ValidatorStake>,
    pub total_gas_burnt: Gas,
    pub total_balance_burnt: Balance,
    pub proof: Option<PartialStorage>,
    pub results: Option<Results>
}
enum Results {
  ApplySplitStateResult(Vec<ApplySplitStateResult>),
  ConsolidatedStateChanges(ConsolidatedStateChanges),
}
```

don't have enough context to suggest a good type name though!

chain/chain/src/chain.rs
```diff
@@ -3308,28 +3438,15 @@ impl<'a> ChainUpdate<'a> {
...
+        // If we have the state for shards in the next epoch already downloaded, apply the state transition
+        // for these states as well
+        // otherwise put the block into the permanent storage, waiting for be caught up
+        let apply_chunk_work = if is_caught_up {
+            self.apply_chunks_preprocessing(me, block, &prev_block, ApplyChunksMode::IsCaughtUp)?
         } else {
             self.chain_store_update.add_block_to_catchup(prev_hash, *block.hash());
-        }
+            self.apply_chunks_preprocessing(me, block, &prev_block, ApplyChunksMode::NotCaughtUp)?
+        };
```
matklad, line 3449: Alternative way to write this:

```rust
        let mode = if is_caught_up {
            ApplyChunksMode::IsCaughtUp
        } else {
            self.chain_store_update.add_block_to_catchup(prev_hash, *block.hash());
            ApplyChunksMode::NotCaughtUp
        };
        let apply_chunk_work = self.apply_chunks_preprocessing(me, block, &prev_block, mode)?;
```

Not sure if it's actually better -- I personally default to deduping such things just as a rule of thumb.

chain/chain/src/store.rs
```diff
@@ -1853,6 +1880,19 @@ impl<'a> ChainStoreUpdate<'a> {
         self.trie_changes.push(trie_changes);
     }
 
+    pub fn remove_consolidated_state_changes(&mut self, block_hash: CryptoHash, shard_id: ShardId) {
+        self.remove_consolidated_state_changes.push((block_hash, shard_id));
+    }
+
+    pub fn add_consolidated_state_changes(
```
matklad, line 1887: nitpick: elswhere we are using `add, remove` ordering.

chain/chain/src/store.rs
```diff
@@ -1853,6 +1880,19 @@ impl<'a> ChainStoreUpdate<'a> {
...
+        &mut self,
+        block_hash: CryptoHash,
+        shard_id: ShardId,
+        state_changes: ConsolidatedStateChanges,
+    ) {
+        self.add_consolidated_state_changes.insert((block_hash, shard_id), state_changes);
```
matklad, line 1893: This is an opportunity for an assertion (if, of course, it would be a valid one): 

```suggestion
        let prev = self.add_consolidated_state_changes.insert((block_hash, shard_id), state_changes);
        assert!(prev.is_none());
```

core/primitives/src/trie_key.rs
```diff
@@ -128,6 +128,20 @@ impl TrieKey {
         }
     }
 
+    pub fn get_account(&self) -> Option<&AccountId> {
```
matklad, line 131: `fn account_id` would be a more idiomatic name: https://rust-lang.github.io/api-guidelines/naming.html#getter-names-follow-rust-convention-c-getter

core/primitives/src/types.rs
```diff
@@ -173,6 +173,28 @@ pub struct RawStateChangesWithTrieKey {
...
+/// Consolidate state change of trie_key and the final value the trie key will be changed to
+#[derive(Debug, Clone, BorshSerialize, BorshDeserialize)]
+pub struct ConsolidatedStateChange {
+    pub trie_key: TrieKey,
+    pub value: Option<Vec<u8>>,
+}
+
+pub type ConsolidatedStateChanges = Vec<ConsolidatedStateChange>;
+
+#[easy_ext::ext(ConsolidatedStateChangesExt)]
```
matklad, line 185: tbh, not a big fan of `easy_ext`: it uses proc-macros *and* traits (both are rather advanced Rust features) to implement essentially a syntactical `foo(x) -> x.foo()` transformation. 

I'd just 

```rust
impl ConsolidatedStateChange {
  pub fn from_raw(changes: &[RawStateChangesWithTrieKey]) -> Vec<Self>
}
```

core/primitives/src/types.rs
```diff
@@ -173,6 +173,28 @@ pub struct RawStateChangesWithTrieKey {
...
 }
 
+/// Consolidate state change of trie_key and the final value the trie key will be changed to
+#[derive(Debug, Clone, BorshSerialize, BorshDeserialize)]
+pub struct ConsolidatedStateChange {
+    pub trie_key: TrieKey,
+    pub value: Option<Vec<u8>>,
+}
+
+pub type ConsolidatedStateChanges = Vec<ConsolidatedStateChange>;
```
matklad, line 183: nitpick: `ConsolidatedStateChange` and `ConsolidatedStateChanges` are indistinguishable at a glance. I'd just use the raw vec type.

core/store/src/trie/shard_tries.rs
```diff
@@ -72,7 +74,10 @@ impl ShardTries {
         let caches = caches_to_use.read().expect(POISONED_LOCK_ERR);
         let store = Box::new(TrieCachingStorage::new(
             self.0.store.clone(),
-            caches[&shard_uid].clone(),
+            caches
+                .get(&shard_uid)
+                .expect(&format!("cache for shard {:?} must exsit", shard_uid))
```
matklad, line 79: consider `.unwrap_or_else`

core/store/src/trie/split_state.rs
```diff
@@ -35,21 +39,96 @@ impl Trie {
...
+        let trie = self.new_trie_update_view(shard_uid, state_root);
+        let mut trie_updates: HashMap<_, _> = self.get_trie_updates(state_roots);
+        let mut insert_receipts = Vec::new();
+        let mut delete_receipts = Vec::new();
+        for ConsolidatedStateChange { trie_key, value } in changes {
+            match trie_key {
+                TrieKey::DelayedReceiptIndices => {}
+                TrieKey::DelayedReceipt { index } => match value {
+                    Some(value) => {
+                        let receipt: Receipt = <_>::try_from_slice(&value).map_err(|_| {
```
matklad, line 59: That's cute syntax, but I'd suggest simpler 
```suggestion
                        let receipt = Receipt::try_from_slice(&value).map_err(|_| {
```

core/store/src/trie/split_state.rs
```diff
@@ -35,21 +39,96 @@ impl Trie {
...
+                            StorageError::StorageInconsistentState(format!(
+                                "invalid delayed receipt {:?}",
```
matklad, line 61: Should be include the underlying error message instead of just `_` it?

core/store/src/trie/split_state.rs
```diff
@@ -35,21 +39,96 @@ impl Trie {
...
+                    None => {
+                        let receipt = get::<Receipt>(&trie, &TrieKey::DelayedReceipt { index })?
+                            .expect("deleted receipt does not exist");
+                        delete_receipts.push((index, receipt));
+                    }
+                },
+                _ => {
+                    // we can safely unwrap here because all other types of trie keys have account ids
+                    let account_id = trie_key.get_account().unwrap();
+                    let new_shard_uid = account_id_to_shard_id(account_id);
```
matklad, line 76:
```suggestion
                TrieKey::Account { account_id }
                | TrieKey::ContractCode { account_id }
                | TrieKey::AccessKey { account_id, .. }
                | TrieKey::ReceivedData { receiver_id, .. }
                | TrieKey::PostponedReceiptId { receiver_id, .. }
                | TrieKey::PendingDataCount { receiver_id, .. }
                | TrieKey::PostponedReceipt { receiver_id, .. }
                | TrieKey::ContractData { account_id, .. } => {
                    let new_shard_uid = account_id_to_shard_id(&account_id);
```

While the comment is true at this point in time, nothing prevents it from becoming false in the future, if we add a new `TrieKey`.

core/store/src/trie/split_state.rs
```diff
@@ -35,21 +39,96 @@ impl Trie {
...
+                    // we can safely unwrap here because the caller of this function guarantees trie_updates contains all shard_uids for the new shards
+                    let trie_update = trie_updates.get_mut(&new_shard_uid).unwrap();
+                    match value {
+                        Some(value) => {
+                            trie_update.set(trie_key, value);
+                        }
+                        None => {
+                            trie_update.remove(trie_key);
+                        }
+                    }
```
matklad, line 86:
```suggestion
                    let trie_update = trie_updates.get_mut(&new_shard_uid).unwrap();
                    match value {
                        Some(value) => trie_update.set(trie_key, value),
                        None => trie_update.remove(trie_key),
                    }
```

core/store/src/trie/split_state.rs
```diff
@@ -35,21 +39,96 @@ impl Trie {
...
+                }
+            }
+        }
+        for (_, update) in trie_updates.iter_mut() {
+            update.commit(StateChangeCause::Resharding);
+        }
+
+        insert_receipts.sort_by(|a, b| a.0.cmp(&b.0));
+        delete_receipts.sort_by(|a, b| a.0.cmp(&b.0));
```
matklad, line 95:
```suggestion
        insert_receipts.sort_by_key(|it| it.0);
        delete_receipts.sort_by_key(|it| it.0);
```

core/store/src/trie/split_state.rs
```diff
@@ -67,68 +146,125 @@ impl ShardTries {
...
+                (*shard_uid, self.new_trie_update(*shard_uid, state_root.clone()))
+            })
+            .collect()
+    }
+
     pub fn apply_delayed_receipts_to_split_states<'a>(
         &self,
         state_roots: &HashMap<ShardUId, StateRoot>,
-        receipts: &[Receipt],
+        receipts: Vec<Receipt>,
```
matklad, line 164: Hm, why do we need a Vec here, seems [] is enough?

tools/restored-receipts-verifier/src/main.rs
```diff
@@ -93,6 +93,7 @@ fn main() -> Result<()> {
             .apply_transactions(
                 shard_id,
                 chunk_extra.state_root(),
+                None,
```
matklad, line 96: It's not great that we have to change a whole bunch of tests to add an extra `None`, feels similar to #4745

chain/chain/src/chain.rs
EgorKulikov, line 2909: consider returning tuple (shard_uid, state_root) from closure and collecting to HashMap to avoid ziping

chain/epoch_manager/src/lib.rs
```diff
@@ -3849,31 +3832,11 @@ mod tests {
...
-            );
-            assert_eq!(
-                epoch_manager
-                    .get_split_shards_if_shards_will_change(&h[i], vec![0])
-                    .unwrap()
-                    .unwrap(),
-                vec![(0, vec![0, 1, 2, 3])].into_iter().collect::<HashMap<_, _>>(),
-            );
-            assert!(epoch_manager.get_split_shards_if_shards_will_change(&h[i], vec![1]).is_err());
+            assert_eq!(epoch_manager.will_shard_layout_change(&h[i]).unwrap(), true,);
```
nikurt, line 3837: The trailing comma looks out of place.

nearcore/src/migrations.rs
```diff
@@ -68,6 +68,7 @@ fn apply_block_at_height(
         .apply_transactions(
             shard_id,
             &chunk_header.prev_state_root(),
+            None,
```
nikurt, line 71: This `None` is hard to read without referring to the function definition.

Don't know what's the best way to handle it in Rust. Leaving it as it is now is not a big deal.
In C++ I'd suggest writing `/*split_state_roots=*/None,`.

nikurt (approved):

chain/chain/src/chain.rs
EgorKulikov: We are taking care of this outside of branch that is gated by update_shard_state

nearcore/src/migrations.rs
nikurt: Ignore this.
Clion already annotates function arguments at the callsite.

chain/chain/src/store.rs
mzhangmzz: Good point!

tools/restored-receipts-verifier/src/main.rs
mzhangmzz: I agree!

chain/chain/src/chain.rs
mzhangmzz: yep, that's right

core/store/src/trie/split_state.rs
mzhangmzz: Yeah, I just saw that everywhere this function is called, `Vec<Receipt>` is used, so I changed it. I just changed it back

nearcore/src/runtime/mod.rs
mzhangmzz: I'm debating whether to move this part outside of runtime and to chain, I'll leave it as it is right now and do the refactoring at the end

chain/chain/src/chain.rs
mzhangmzz: I had some problem figuring out how to make `?` work in a closure, but I think I just figured it out

chain/chain/src/chain.rs
```diff
@@ -90,9 +90,16 @@ pub const NUM_EPOCHS_TO_KEEP_STORE_DATA: u64 = 5;
 /// Maximum number of height to go through at each step when cleaning forks during garbage collection.
 const GC_FORK_CLEAN_STEP: u64 = 1000;
 
+/// apply_chunks may be called in two code paths, through process_block or through catchup_blocks
+/// When it is called through process_block, it is possible that the shard state for the next epoch
+/// has not been caught up yet, thus the two modes IsCaughtUp and NotCaughtUp.
+/// CatchingUp is for when apply_chunks is called through catchup_blocks, this is to catch up the
+/// shard states for the next epoch
```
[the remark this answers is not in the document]
mzhangmzz: Do you have any suggestions? I am not sure which area you want it to be more explicit

chain/chain/src/chain.rs
```diff
@@ -1857,18 +1864,22 @@ impl Chain {
...
             shard_uid,
             &state_root,
             &next_shard_layout,
         )?;
-        Ok(())
+
+        let mut chain_update = self.chain_update();
+        for (shard_uid, state_root) in state_roots {
+            let chunk_extra = ChunkExtra::new(&state_root, CryptoHash::default(), vec![], 0, 0, 0);
+            chain_update.chain_store_update.save_chunk_extra(&prev_hash, &shard_uid, chunk_extra);
```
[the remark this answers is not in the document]
mzhangmzz: yeah right now I'm just using ChunkExtra to store the state_root and not using any other field in chunk_extra.

I might change it in a following PR where I need to deal with the epoch boundary where number of shards changes. There I need to figure out where to get information such as `validator_proposals` and `incoming/outgoing_receipts` for chunks, and I might need to change this struct here, although I haven't thought very carefully about exactly how I need to do that.

I will add a comment here explaining that and deal with this in the next PR.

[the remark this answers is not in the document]
mzhangmzz: Actually, sorry here it is intended because this is the initial state, there is no need for any other field to be stored

chain/chain/src/chain.rs
mzhangmzz: I agree it looks a bit hacky, I changed it

chain/chain/src/chain.rs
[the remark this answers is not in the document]
bowenwang1996: Something like `StateCatchingUp`? I am also fine as is

chain/chain/src/chain.rs
```diff
@@ -2898,33 +2938,63 @@ impl<'a> ChainUpdate<'a> {
...
-                ),
+            let cares_about_shard_this_epoch =
+                self.runtime_adapter.cares_about_shard(me.as_ref(), prev_hash, shard_id, true);
+            let cares_about_shard_next_epoch =
+                self.runtime_adapter.will_care_about_shard(me.as_ref(), prev_hash, shard_id, true);
+            // We want to guarantee that transactions are only applied once for each shard, even
+            // though apply_chunks may be called twice, once with ApplyChunksMode::NotCaughtUp
+            // once with ApplyChunksMode::CatchingUp
+            // Note that this variable does not guard whether we split states or not, see the comments
+            // before `need_to_split_state`
```
bowenwang1996, line 2955: This comment is very informative!

chain/chain/src/chain.rs
```diff
@@ -3088,19 +3160,77 @@ impl<'a> ChainUpdate<'a> {
...
+                                    apply_result,
                                 }))
                             }
                             Err(err) => Err(ErrorKind::Other(err.to_string()).into()),
                         }
                     }));
                 }
+            } else if mode == ApplyChunksMode::CatchingUp
+                && need_to_split_state
+                && cares_about_shard_this_epoch
```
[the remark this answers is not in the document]
mzhangmzz: Yes, that will be `mode = ApplyChunksMode::CatchingUp && cares_about_shard_next_epoch && !cares_about_shard_this_epoch`, and you can see that `should_apply_transactions` will be true in this case, which means transactions will be applied from scratch and split states will be done within `apply_transactions`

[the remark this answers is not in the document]
mzhangmzz: So basically there are two ways that we apply transactions and split states, 1) at the same time through `apply_transactions` if the split states are ready or 2) `apply_transactions` applies transactions but does not update split states and split states are updated later when `split_state_roots` are ready.

Case 1) can be further split to two cases a) mode == IsCaughtUp, that means split states are ready for all shards in the next epoch b) mode == CatchingUp, in this case, we still need to call `apply_transactions` for shards that `!cares_about_shard_this_epoch && cares_about_shard_next_epoch` and split states are updated in `apply_transactions` for these shards too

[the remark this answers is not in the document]
bowenwang1996: > which means transactions will be applied from scratch and split states will be done within apply_transactions

I see. That makes sense. I feel like the state machine here is quite complicated and we may need a more detailed writeup to document the logic. In addition, it behooves us to consider refactoring this piece of code to make the logic more clear and potentially break it down into smaller functions to improve readability. We don't need to do these now, but probably good to create some issues to ensure that we don't forget about them :)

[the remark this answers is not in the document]
mzhangmzz: I refactored this part of code and add more comments. Could you take a look again?

nearcore/src/runtime/mod.rs
```diff
@@ -609,7 +645,15 @@ fn apply_delayed_receipts(
             &receipts,
             &|account_id| {
                 let shard_id = account_id_to_shard_id(&account_id, next_shard_layout);
-                ShardUId::from_shard_id_and_layout(shard_id, next_shard_layout)
+                let shard_uid = ShardUId::from_shard_id_and_layout(shard_id, next_shard_layout);
+                assert!(
```
EgorKulikov, line 649: We do this check twice, maybe introduce function to do this check?

core/primitives/src/shard_layout.rs
```diff
@@ -128,6 +136,13 @@ pub fn account_id_to_shard_id(account_id: &AccountId, shard_layout: &ShardLayout
     }
 }
 
+pub fn account_id_to_shard_uid(account_id: &AccountId, shard_layout: &ShardLayout) -> ShardUId {
```
EgorKulikov, line 139: I wonder why this is not a method on ShardLayout, seems natural

chain/chain/src/store.rs
```diff
@@ -1150,6 +1164,9 @@ pub struct ChainStoreUpdate<'a> {
     final_head: Option<Tip>,
     largest_target_height: Option<BlockHeight>,
     trie_changes: Vec<WrappedTrieChanges>,
+    // All state changes made by a chunk, this is only used for splitting states
+    add_state_changes_for_split_states: HashMap<(CryptoHash, ShardId), StateChangesForSplitStates>,
```
EgorKulikov, line 1168: Seems like we only use HashMap here to assert there are no duplicates. Maybe we should use HashSet for removals for the same reason?

EgorKulikov (approved):
LGTM
My comments are mostly nits, but please make sure there is no problem with db as per @bowenwang1996 comment
Great work

matklad (approved):
(adding :heavy_check_mark: review to unblock: Rust code looks great to me, but I can't vouch for correctness :)

core/primitives/src/shard_layout.rs
mzhangmzz: That's a very good point. I think I just extended from the original `account_id_to_shard_id` function and didn't think too much.

mzhangmzz:
> LGTM
> My comments are mostly nits, but please make sure there is no problem with db as per @bowenwang1996 comment
> Great work

Thanks @EgorKulikov

mzhangmzz:
> (adding ✔️ review to unblock: Rust code looks great to me, but I can't vouch for correctness :)

Thanks!

Longarithm (approved):

core/store/src/trie/split_state.rs
```diff
@@ -35,21 +40,110 @@ impl Trie {
...
+        Ok(trie_changes_map)
+    }
+
     /// Apply `changes` to build states for new shards
     /// `state_roots` contains state roots for the new shards
     /// The caller must guarantee that `state_roots` contains all shard_ids
     /// that `key_to_shard_id` that may return
     /// Ignore changes on DelayedReceipts or DelayedReceiptsIndices
     /// Update `store_update` and return new state_roots
     /// used for building states for new shards in resharding
```
Longarithm, line 115: This comment is helpful, but I feel that it is not up to date:

```
/// Apply `changes` to build states for new shards
```
Does it refer to `apply_state_changes_to_split_states` above?

```
/// Update `store_update` and return new state_roots
/// used for building states for new shards in resharding
```
Should it be "Return `store_update` and new state_roots"?

core/store/src/trie/split_state.rs
```diff
@@ -67,68 +161,125 @@ impl ShardTries {
...
-        for (shard_uid, mut trie_update) in trie_updates {
-            set(
-                &mut trie_update,
-                TrieKey::DelayedReceiptIndices,
-                delayed_receipts_indices_by_shard.get(&shard_uid).unwrap(),
+    for receipt in delete_receipts {
+        let new_shard_uid: ShardUId = account_id_to_shard_id(&receipt.receiver_id);
+        if !trie_updates.contains_key(&new_shard_uid) {
+            let err = format!(
+                "Account {} is in new shard {:?} but state_roots only contains {:?}",
```
Longarithm, line 252:
```suggestion
                "Account {:?} is in new shard {:?} but state_roots only contains {:?}",
```

It's interesting that package compiles successfully but CLion claims that `AccountId` doesn't implement Display. The same applies to line 223 above.

Maybe cc @matklad

Longarithm (approved):
Also some nits

chain/chain/src/store.rs
```diff
@@ -1853,6 +1880,27 @@ impl<'a> ChainStoreUpdate<'a> {
...
+        assert!(prev.is_none());
+    }
+
+    pub fn remove_state_changes_for_split_states(
+        &mut self,
+        block_hash: CryptoHash,
+        shard_id: ShardId,
+    ) {
+        // We should not remove state changes for the same chunk twice
+        assert!(self.remove_state_changes_for_split_states.insert((block_hash, shard_id)));
```
bowenwang1996, line 1901: The style here is not consistent with the function above :) I personally prefer assertions to not have side effects

bowenwang1996 (approved):
Epic! I particularly like that there are ample comments where the logic is complex and this makes reviewing the code much easier. Great work 🚀

core/store/src/trie/split_state.rs
mzhangmzz: Good catch, I have updated the comments.

Thanks for pointing out! I think it is confusing that `apply_state_changes_to_split_states` and `add_values_to_split_states` are kind of similar. I have added comments for `apply_state_changes_to_split_states` as well. Hopefully that makes it clearer

chain/chain/src/store.rs
mzhangmzz: Updated

core/store/src/trie/split_state.rs
mzhangmzz: hmm I'm not sure why, it works in my CLion

Longarithm: Oh, I had to update my version of CLion and then it worked.

chain/chain/src/store.rs
matklad: +1 while `assert!` is guaranteed to be always on in Rust, and it's OK to, eg, rely on that in unsafe code, splitting side-effectful part of it into a separate statement makes the code more obviously clear.

PR #4848, opened 2021-09-20
fix(near-primitives): `EpochReference` enum can be flattened into a struct

miraclx:
`EpochReference` is defined as such:

https://github.com/near/nearcore/blob/06d05bce3fab578a91b210d7354fb17427fcc9d6/core/primitives/src/types.rs#L1001-L1007

And it's serialization, as such:

| % | `EpochReference::BlockId(...)` | `EpochReference::Latest` |
| - | ------------------------------ | ------------------------ |
| expectation | `{"block_id": ...}` | `{"latest": null}`
| reality | `{"block_id": ...}` | `"latest"` |

As you can already see, `serde`'s default method for serializing unit variants is to use a string 
unlike how it uses an object for the rest of the variants.

This PR implements custom serialization for `EpochReference`, specifically for the `Latest` variant. So it serializes to `{"latest": null}`.

> Note: this retains the default deserialization so we have no breaking changes on the deserialization side.
>
> So, the following values properly deserialize into that variant:
> - `"latest"`
> - `{"latest": null}`
> - `{"latest": []}`

_props to @volovyk-s 🎉  for discovering this while using the new [`near-jsonrpc-client`](https://github.com/near/near-jsonrpc-client-rs)_

### DIFF
--- a/chain/jsonrpc-primitives/src/types/status.rs
+++ b/chain/jsonrpc-primitives/src/types/status.rs
@@ -6,7 +6,7 @@ pub struct RpcStatusResponse {
     pub status_response: near_primitives::views::StatusResponse,
 }
 
-#[derive(Debug, Serialize)]
+#[derive(Debug, Serialize, Deserialize)]
 pub struct RpcHealthResponse;
 
 #[derive(thiserror::Error, Debug, Serialize, Deserialize)]
--- a/core/primitives/src/types.rs
+++ b/core/primitives/src/types.rs
@@ -998,12 +998,31 @@ pub struct BlockChunkValidatorStats {
     pub chunk_stats: ValidatorStats,
 }
 
-#[derive(Serialize, Deserialize, Debug)]
+#[derive(Deserialize, Debug)]
 #[serde(rename_all = "snake_case")]
 pub enum EpochReference {
     EpochId(EpochId),
     BlockId(BlockId),
     Latest,
+}
+
+impl Serialize for EpochReference {
+    fn serialize<S>(&self, s: S) -> Result<S::Ok, S::Error>
+    where
+        S: serde::Serializer,
+    {
+        match self {
+            EpochReference::EpochId(epoch_id) => {
+                s.serialize_newtype_variant("EpochReference", 0, "epoch_id", epoch_id)
+            }
+            EpochReference::BlockId(block_id) => {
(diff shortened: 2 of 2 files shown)


miraclx:
More context:

In the case where it's embedded in a struct whose serialization requires flattening `EpochReference`, for example:

https://github.com/near/nearcore/blob/06d05bce3fab578a91b210d7354fb17427fcc9d6/chain/jsonrpc-primitives/src/types/validator.rs#L18-L22

All is well until you use the `Latest` variant of `EpochReference`.

<table role="table">
  <thead>
    <tr>
      <th>code</th>
      <th>expectation</th>
      <th>reality</th>
    </tr>
  </thead>
  <tbody>
    <tr>
      <td>
        <pre lang="rust">
RpcValidatorRequest {
    epoch_reference: EpochReference::BlockId(...)
};
        </pre>
	  </td>
      <td colspan=2>
        <pre lang="json">
{
  "block_id": ...
}
        </pre>
      </td>
    </tr>
<tr></tr>
    <tr>
      <td>
        <pre lang="rust">
RpcValidatorRequest {
    epoch_reference: EpochReference::Latest
};
        </pre>
	  </td>
      <td>
        <pre lang="json">
{
  "latest": null
}
        </pre>
      </td>
      <td>
        <pre lang="console"> Error: can only flatten structs and maps (got an enum) </pre>
      </td>
    </tr>
  </tbody>
</table>

bowenwang1996 (approved):

frol (approved):
Meh, this somehow slipped through my review, but the API is quite odd. I am fine with this hotfix, though I want to share more context here.

```
validators({block_id: 123})
validators({block_id: "deadbeef"})
validators({latest: null})
```

We have not even updated the documentation: https://docs.near.org/docs/api/rpc/network#validation-status, so we may as well refactor the API altogether.

```
validators({epoch_reference: {block_id: 123}})
validators({epoch_reference: {block_id: "deadbeef"}})
validators({epoch_reference: "latest"})
```

Maybe you have a better idea, I am currently not at my highest with naming :disappointed:

chefsale (approved):

matklad:
It's not immediately obvious that derive serialization can deserialize our custom one, so it might make sense to add a unit-test for this.

PR #4851, opened 2021-09-21
fix(primitives): backwards compatible ExecutionOutcomeView

miraclx:
Since #4438, the `metadata` field was introduced on `ExecutionOutcomeView`.
Older nodes _(e.g the ones deployed on `[archival-]rpc.<network>.near.org` as of today)_ do not include that field. This means deserializing that data into the new `ExecutionOutcomeView` fails.

This PR introduces a default value for that field. Specifically, from `ExecutionMetadata::V1`.

### DIFF
--- a/core/primitives/src/views.rs
+++ b/core/primitives/src/views.rs
@@ -1031,6 +1031,12 @@ pub struct ExecutionMetadataView {
     gas_profile: Option<Vec<CostGasUsed>>,
 }
 
+impl Default for ExecutionMetadataView {
+    fn default() -> Self {
+        ExecutionMetadata::V1.into()
+    }
+}
+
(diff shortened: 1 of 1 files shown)


frol (approved):
@matklad I feel like V1 is equal to None, so I don't have any hard feelings about deserializing missing value to V1. WDYT?

matklad:
Don't have strong opinion: metadata indeed has a  reasonable empty, default state, and we can just use that. So let's do that, but, to make this reasonig more clear in code, I suggest using just `#[serde(default)]` and implementing `Default` for `ExecutionMetadataView::view`.

matklad (approved):
Fine as is, though `impl Default for ExecutionMetadataView` would be a touch nicer imo!

PR #4852, opened 2021-09-21
chore(sandbox): add more records to test sandbox

willemneal:


willemneal:
@ailisp I can't seem to trigger the sandbox tests with nayduck?  Does this happen nightly?

As for the tests, I found locally that when adding more records I would get a mailbox full error.  This test is also a bit on the heavy side, which I hope triggers this same error.  We can scale it back if you think this is beyond the limit of the number of records passed.

ailisp:
@willemneal seems this is your branch on nayduck and sandbox test is triggered: http://nayduck.near.org/#/run/2075

http://nayduck.near.org/#/test/222951

ailisp:
> As for the tests, I found locally that when adding more records I would get a mailbox full error. 

I haven't stress tested it to see how big state it can be patched. Logically, a patch state of same size of a deploy contract action should be okay without mailbox error. More than that may require user to break patch state into multiple rpc calls.

ailisp (approved):

willemneal:
@ailisp I have narrowed down my issue with sandbox.  It seems this line is failing
https://github.com/near/nearcore/blob/d95687564b89a8fea481813d8e1b50c76ce7ee19/runtime/runtime/src/lib.rs#L1435-L1440

Why is it the case that the new code hash should equal the previous?

Issue #4835, opened 2021-09-17
runtime-tester is sensitive to account names for some reason


matklad:
If I change this line

https://github.com/near/nearcore/blob/3080cf7d305c8e1d25d09d378c7f0ab1f841ec27/test-utils/runtime-tester/src/lib.rs#L23

to use any other account name, like `format!("best_{}", i)`, the test starts to fail. This seems odd -- it should be possible to use different account names.

matklad:
cc https://github.com/near/nearcore/issues/4771#issuecomment-921375049

mina86:
The issue is with `TestEnv::new_with_runtime_and_network_adapter` assuming the format.

bowenwang1996:
@mina86 do you want to fix this?

mina86:
I can take it, sure.

PR #4855, opened 2021-09-21
Introduce TestEnvBuilder structure

mina86:
Replace TestEnv constructors with a TestEnvBuilder structure instead.
Motivation for this change are twofold.  First of all, with Rust’s
lack of default or keyword arguments constructors of TestEnv would
just continue to multiply with more and more arguments.  Second of
all, with the current design whoever uses TestEnv structure must
adhere to `test{}` AccountId format which may not always be desired.

With this commit also change three instances where
TestEnv::new_with_runtime_and_network_adapter was used by the use of
the builder.  Other places where TestEnv constructors are used are
left for now and will be dealt in another commit.

### DIFF
--- a/chain/client/src/test_utils.rs
+++ b/chain/client/src/test_utils.rs
@@ -1102,7 +1102,137 @@ pub struct TestEnv {
     pub clients: Vec<Client>,
 }
 
+/// A builder for the TestEnv structure.
+pub struct TestEnvBuilder {
+    chain_genesis: ChainGenesis,
+    clients: Vec<AccountId>,
+    validators: Vec<AccountId>,
+    runtime_adapters: Option<Vec<Arc<dyn RuntimeAdapter>>>,
+    network_adapters: Option<Vec<Arc<MockNetworkAdapter>>>,
+}
+
+impl TestEnvBuilder {
+    pub fn new(chain_genesis: ChainGenesis) -> Self {
+        Self {
+            chain_genesis,
+            clients: Self::make_accounts(1, Self::default_formatter),
+            validators: Self::make_accounts(1, Self::default_formatter),
+            runtime_adapters: None,
+            network_adapters: None,
+        }
+    }
+
+    pub fn clients(mut self, clients: Vec<AccountId>) -> Self {
+        assert!(!clients.is_empty());
+        self.clients = clients;
+        self
+    }
+
+    pub fn clients_count(mut self, num_clients: usize) -> Self {
+        self.clients_with_formatter(num_clients, Self::default_formatter)
+    }
+
+    pub fn clients_with_formatter<F>(mut self, num_clients: usize, formatter: F) -> Self
+    where
+        F: Fn(usize) -> std::string::String,
+    {
+        self.clients(Self::make_accounts(num_clients, formatter))
+    }
+
+    pub fn validators(mut self, validators: Vec<AccountId>) -> Self {
+        assert!(!validators.is_empty());
+        self.validators = validators;
+        self
+    }
+
+    pub fn validators_count(mut self, num_validators: usize) -> Self {
+        self.validators_with_formatter(num_validators, Self::default_formatter)
+    }
+
+    pub fn validators_with_formatter<F>(mut self, num_validators: usize, formatter: F) -> Self
+    where
+        F: Fn(usize) -> std::string::String,
+    {
+        self.validators(Self::make_accounts(num_validators, formatter))
+    }
+
+    pub fn runtime_adapters(mut self, adapters: Vec<Arc<dyn RuntimeAdapter>>) -> Self {
+        self.runtime_adapters = Some(adapters);
+        self
+    }
+
+    pub fn network_adapters(mut self, adapters: Vec<Arc<MockNetworkAdapter>>) -> Self {
+        self.network_adapters = Some(adapters);
+        self
+    }
+
+    pub fn build(self) -> TestEnv {
+        let chain_genesis = self.chain_genesis;
+        let clients = self.clients;
+        let num_clients = clients.len();
+        let validators = self.validators;
+        let num_validators = validators.len();
+        let network_adapters = self
+            .network_adapters
+            .unwrap_or_else(|| (0..num_clients).map(|_| Arc::new(Default::default())).collect());
+        assert!(clients.len() == network_adapters.len());
+        let clients = match self.runtime_adapters {
+            None => clients
+                .into_iter()
+                .zip(network_adapters.iter())
+                .map(|(account_id, network_adapter)| {
+                    setup_client(
+                        create_test_store(),
+                        vec![validators.clone()],
+                        1,
+                        1,
+                        Some(account_id),
+                        false,
+                        network_adapter.clone(),
+                        chain_genesis.clone(),
+                    )
+                })
+                .collect(),
+            Some(runtime_adapters) => {
+                assert!(clients.len() == runtime_adapters.len());
+                clients
+                    .into_iter()
+                    .zip((&network_adapters).iter())
+                    .zip(runtime_adapters.into_iter())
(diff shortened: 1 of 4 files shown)


bowenwang1996 (approved):

chain/client/src/test_utils.rs
```diff
@@ -1102,7 +1102,151 @@ pub struct TestEnv {
...
+/// A builder for the TestEnv structure.
+pub struct TestEnvBuilder {
+    chain_genesis: ChainGenesis,
+    clients: Vec<AccountId>,
+    validators: Vec<AccountId>,
+    runtime_adapters: Option<Vec<Arc<dyn RuntimeAdapter>>>,
+    network_adapters: Option<Vec<Arc<MockNetworkAdapter>>>,
+}
+
+/// Builder for the `TestEnv` structure.
```
matklad, line 1114: nitpick: we might start using `` [`TestEnv`] `` syntax to get proper links in rustdoc and goto definition in IDE.

chain/client/src/test_utils.rs
```diff
@@ -1102,7 +1102,151 @@ pub struct TestEnv {
...
+                        )
+                    })
+                    .collect()
+            }
+        };
+
+        TestEnv { chain_genesis, validators, network_adapters, clients }
+    }
+
+    fn default_formatter(id: usize) -> std::string::String {
```
matklad, line 1233: Just `String`?

chain/client/src/test_utils.rs
```diff
@@ -1102,7 +1102,151 @@ pub struct TestEnv {
...
+        format!("test{}", id)
+    }
+
+    fn make_accounts<F>(count: usize, formatter: F) -> Vec<AccountId>
```
matklad, line 1237: We always use `default_formatter`, so I'd just YAGNI this. 

If the user wants to customize it, i'd let them provide the actual names, rather than a formatter.

matklad (commented):
Delightful!

chain/client/src/test_utils.rs
mina86: In early version of this code there was a `clients_with_formatter` function which took number of clients and a formatter function. While it’s true that with `clients` method where the vector can be provided directly support for custom formatters is unnecessary but it seems that the most common case will be to to simply use `clients_count(number)` and the second most common will be to use `clients_with_formatter(number, |id| format("blah{}", id)`. Perhaps it’s `clients` method that we’re not gonna need?

chain/client/src/test_utils.rs
mina86: Yeah. Admittedly this is a vestige of the earlier version of the code. The idea is to let users provide custom formatters (though currently that feature is not available) and in that case it seems that the typical use case will be to call `format!` which returns a `String`. I therefore moved converting the string to `AccountId` inside of the builder.  This could also by anything implementing `TryInto<AccountId>` but that adds more generic types which I didn’t care to define.

chain/client/src/test_utils.rs
mina86: Yeah, good point. I was actually write that but then I forgot whether that’s actually rustdoc syntax and then forgot to check it and correct it… I’ll fix that (and potentially other issues) in separate commit.

chain/client/src/test_utils.rs
matklad: > the second most common will be to use clients_with_formatter(number, |id| format("blah{}", id)

Why would this be common? It seems that there shouldn't be a need to customize the names, outside of the special cases like #4771, where overriding the whole thing should be enough. 

I feel `clients` is needed because it's the most general -- even if we call it zero times, it's important that we can customize this behavior when we suddenly need this for some future investigation. This avoids [midlayer mistake](https://lwn.net/Articles/336262/) -- providing convenience API without exposing the full capabilities of the underlying platform. 

Playing around this ideas, I got the following: 

```diff
diff --git a/chain/client/src/test_utils.rs b/chain/client/src/test_utils.rs
index 723584675..9d97d3ece 100644
--- a/chain/client/src/test_utils.rs
+++ b/chain/client/src/test_utils.rs
@@ -1117,11 +1117,13 @@ impl TestEnvBuilder {
     fn new(chain_genesis: ChainGenesis) -> Self {
         Self {
             chain_genesis,
-            clients: Self::make_accounts(1, Self::default_formatter),
-            validators: Self::make_accounts(1, Self::default_formatter),
+            clients: Vec::new(),
+            validators: Vec::new(),
             runtime_adapters: None,
             network_adapters: None,
         }
+        .clients_count(1)
+        .validator_seats(1)
     }
 
     /// Sets list of client `AccountId`s to the one provided.  Panics if the
@@ -1132,10 +1134,11 @@ impl TestEnvBuilder {
         self
     }
 
-    /// Sets number of clients to given one.  Each client will use `AccountId`
-    /// in the form `test{}` where `{}` will count from zero.
+    /// Sets number of clients to given one.
     pub fn clients_count(self, num: usize) -> Self {
-        self.clients(Self::make_accounts(num, Self::default_formatter))
+        let clients =
+            (0..num).map(|i| AccountId::try_from(format!("test{}", i)).unwrap()).collect();
+        self.clients(clients)
     }
 
     /// Sets list of validator `AccountId`s to the one provided.  Panics if the
@@ -1146,10 +1149,10 @@ impl TestEnvBuilder {
         self
     }
 
-    /// Sets number of validator seats to given one.  Each validator will use
-    /// `AccountId` in the form `test{}` where `{}` will count from zero.
+    /// Sets the first `num` accounts as validators.
     pub fn validator_seats(self, num: usize) -> Self {
-        self.validators(Self::make_accounts(num, Self::default_formatter))
+        let validators = self.clients[..num].to_vec();
+        self.validators(validators)
     }
 
     /// Specifies custom runtime adaptors for each client.  This allows us to
@@ -1229,17 +1232,6 @@ impl TestEnvBuilder {
 
         TestEnv { chain_genesis, validators, network_adapters, clients }
     }
-
-    fn default_formatter(id: usize) -> std::string::String {
-        format!("test{}", id)
-    }
-
-    fn make_accounts<F>(count: usize, formatter: F) -> Vec<AccountId>
-    where
-        F: Fn(usize) -> std::string::String,
-    {
-        (0..count).map(|i| AccountId::try_from(formatter(i)).unwrap()).collect()
-    }
 }
 
 impl TestEnv {

```

I've almo made a couple observations while playing with it:

* there's slight naming inconsistensy where we use count, num, and seats
* not sure if we should document the exact format of test account naming. Seems better to provide accessor to resulting `AccountId` structs. Reliance on specific names is exactly what caused the original issues with different names not working.

mina86: You’re not wrong and my assertion isn’t really supported by hard data. On the other hand, we have no examples of tests where we use account ids which don’t follow simple pattern while we have an example of test which wants to change the pattern. ;)

I’m also not exactly convinced that we should be worried about the midlayer mistake. We can have both `clients` and `clients_with_formatter`. While that ‘bloats’ the builder, I think it’s acceptable for a helper test code whose whole purpose is to provide convenience.

> there's slight naming inconsistensy where we use count, num, and seats

I was going back and forth between `validator_seats` and `validators_count`. I don’t really have any good reasons for either to be honest. I first went with count but then saw bunch of code use seats. Don’t really have an opinion.

> Seems better to provide accessor to resulting AccountId structs.

Yes, I've been looking into that as well though I think such accessor belongs to `TestEnv` rather than the builder.

PS. Note that there are tests where number of validator seats is greater than number of clients, see `test-utils/state-viewer/src/state_dump.rs` so `validators = self.clients[..num].to_vec();` is sadly not valid.

chain/client/src/test_utils.rs
mina86: Ah, wait, you just meant to use `String` i/o `std::string::String`. Will do in a follow up commit.

PR #4856, opened 2021-09-22
Remove TestEnv::new_with_runtime constructor in favour of a builder

mina86:
Replace all invocations of TestEnv::new_with_runtime constructor with
uses of the TestEnvBuilder.  This (hopefully) makes the code more
readable by naming the arguments and the use of the builder provides
a way for future expansion of the ‘constructor’ arguments.

This leaves TestEnv::new as the remaining constructor.  It will be
dealt with in a follow-up commit.


integration-tests/tests/client/process_blocks.rs
```diff
@@ -104,6 +104,19 @@ pub fn create_nightshade_runtimes(genesis: &Genesis, n: usize) -> Vec<Arc<dyn Ru
         .collect()
 }
 
+pub fn create_test_env(
```
matklad, line 107: suggestion(non-blocking): we might inline/remove this function and call `TestEnv::builder` at the call-site? This function mostly converts named arguments to positional arguments, which hurts readability. Otoh, even if we do want to do that, might be better left to a follow-up.

matklad (approved):

nikurt (approved):

bowenwang1996 (approved):

PR #4858, opened 2021-09-22
refactor: ensure that estimator's test contract is always build

matklad:
Replaces build.sh which has to be invoked manually with build.rs which
is run automatically when needed.

This duplicates https://github.com/near/nearcore/blob/master/runtime/near-test-contracts/build.rs a bit, but I think that's fine -- the build process for estimator is slightly different, and trying to dedupe this will just create extra complexity.


runtime/runtime-params-estimator/build.rs
```diff
@@ -0,0 +1,100 @@
...
+    for mode in ["stable", "nightly"] {
+        let make_payload = {
+            let features = if mode == "nightly" { "nightly_protocol_features" } else { "" };
+            let baseline = res.join(format!("{}_baseline_contract.wasm", mode));
+            build_contract(features, &baseline)?;
+            let size = std::fs::metadata(&baseline)?.len();
+
+            move |desired_size: u64| -> io::Result<()> {
+                let payload_size = desired_size.saturating_sub(size);
+                let urandom = fs::File::open("/dev/urandom")?;
```
olonho, line 33: Will only work on Linux. Also makes tests somewhat non-deterministic.

matklad: Yeah, I've just directly ported whatever was originally in bash: 
https://github.com/near/nearcore/blob/8f90b9267e6b672bab4ee6b483369e66c281b349/runtime/runtime-params-estimator/test-contract/build.sh#L42

And, to be fair, another big (but bad) motivation for me was to use `io::copy + .take` combo, which is quite nice, but indeed completely unnecessary here :D Will replace with in-process generation while I am at it!

matklad: fixed in https://github.com/near/nearcore/pull/4858/commits/9877c249cc1d9e49ad7ce4c23ab0839ed1f073d4, PTAL

matklad:
Going to put this on pause -- working around https://github.com/rust-lang/rustup/issues/988 turned out to be non-trivial, need to think about some creative solution here )

chefsale (approved):

PR #4864, opened 2021-09-23
near_chain: don’t hard-code account ids in TestEnv

mina86:
Don’t assume account ids of the clients use test{} format.  Instead,
introduce get_client_id method which allows getting account id used
when creating given client and use that to fetch the actual ids.  The
method will allow other users of TestEnv to query that information as
well without having to rely on any particular format TestEnv might use.


bowenwang1996 (approved):
This is awesome!

PR #4865, opened 2021-09-23
Feat: decrease ecrecover base cost

### LINKED ISSUE #4861: EcRecover Base update
As requested per https://github.com/near/nearcore/issues/4778#issuecomment-925187102, we need to update the EcRecover costs to what @matklad has observed.

This should follow along of the lines of the work done here: https://github.com/near/nearcore/pull/4795.

birchmd:
Closes #4861

See the issue for details.

Test plan: check cost is lowered

### DIFF
--- a/core/primitives/src/runtime/config_store.rs
+++ b/core/primitives/src/runtime/config_store.rs
@@ -20,6 +20,8 @@ static CONFIGS: &[(ProtocolVersion, &[u8])] = &[
     (42, include_config!("42.json")),
     #[cfg(feature = "protocol_feature_lower_data_receipt_cost")]
     (116, include_config!("116.json")),
+    #[cfg(feature = "protocol_feature_lower_ecrecover_base_cost")]
+    (117, include_config!("117.json")),
 ];
 
 /// Stores runtime config for each protocol version where it was updated.
@@ -84,6 +86,8 @@ mod tests {
     use crate::serialize::to_base;
     #[cfg(feature = "protocol_feature_lower_data_receipt_cost")]
     use crate::version::ProtocolFeature::LowerDataReceiptCost;
+    #[cfg(feature = "protocol_feature_lower_ecrecover_base_cost")]
+    use crate::version::ProtocolFeature::LowerEcrecoverBaseCost;
     use crate::version::ProtocolFeature::LowerStorageCost;
     use near_primitives_core::hash::hash;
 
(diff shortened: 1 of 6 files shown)


core/primitives/src/runtime/config_store.rs
```diff
@@ -198,4 +203,20 @@ mod tests {
...
+        let new_cfg = store.get_config(LowerEcrecoverBaseCost.protocol_version());
+        assert!(
+            base_cfg.transaction_costs.data_receipt_creation_config.base_cost.send_sir
+                > new_cfg.transaction_costs.data_receipt_creation_config.base_cost.send_sir
+        );
+        assert!(
+            base_cfg.transaction_costs.data_receipt_creation_config.cost_per_byte.send_sir
+                > new_cfg.transaction_costs.data_receipt_creation_config.cost_per_byte.send_sir
+        );
+    }
```
matklad, line 221: This test feel a bit like busy work, to be honest, I am not sure it adds much value. 

Thinking about this more.... Why didn't this change failed any of the existing tests? It seems that we should have a test that ecrecover behaves as expected at the latest version of the protocol, to make sure we don't change it by accident.

birchmd: I also missed changing the config value I wanted to test when I was copying the test from #4795 🤦‍♂️ Obviously this test was meant to test the ecrecover base cost value. Fixed in https://github.com/near/nearcore/pull/4865/commits/d4668b0361991b6f134a8a59bf0c324a3d50db1c

>  to make sure we don't change it by accident.

I think it would fail the [test checking the config hashes](https://github.com/near/nearcore/blob/master/core/primitives/src/runtime/config_store.rs#L111) if I were to change an existing config as opposed to adding a new one.

Longarithm (approved):
Side note: this helped to found that ripemd and ecrecover fees are incorrectly missing in 29.json, 42.json and 116.json.
But this will have no effect if we drop 116.json and leave 117.json while stabilizing nightly features.

Longarithm:
@birchmd will it be fine if I remove 116.json in scope of this PR?

bowenwang1996 (approved):

birchmd:
> @birchmd will it be fine if I replace existing 116.json with 117.json in scope of this PR?

Sure, if that makes sense to do. I don't know what the flow here is supposed to be (does a later version supersede a previous one or is there always a file for all nightly versions?)

Longarithm:
>I don't know what the flow here is supposed to be (does a later version supersede a previous one or is there always a file for all nightly versions?)

We just add configs with new version on top of existing ones. For stabilizing nightly features into one stable feature, we leave only last nightly runtime config and give new stable feature id to it.

core/primitives/src/runtime/config_store.rs
matklad: >I think it would fail the test checking the config hashes if I were to change an existing config as opposed to adding a new one.

Yeah, that's what I mean -- we have the test for config, but we don't have the test that verifies that the final cost is what we'd expect. 

https://github.com/near/nearcore/blob/43dc0d15190d25f8ec01dbd330bb7eb956826e6f/runtime/near-vm-logic/tests/test_miscs.rs#L589-L624 sort of does that, but it looks at what costs were used, not at the final gas cost. I think that's maybe though.

matklad (approved):
Oups, sorry, didn't realise I forgot to :heavy_check_mark: this, LGTM!

PR #4866, opened 2021-09-23
Catchup blocks parallel

EgorKulikov:
Block catch up is now done in a separate actor to avoid blocking client actor while we are applying chunks during catch up

### Test plan
As this is not new functionality there are already quite some tests (see chain/client/tests/catching_up.rs and integration-tests/tests/test_catchup.rs). There is also a plan to try this with new validator on testnet


chain/chain/src/chain.rs
```diff
@@ -2875,34 +2983,30 @@ impl<'a> ChainUpdate<'a> {
...
-    /// Applies chunks in separate threads, awaits finishing and processes results
-    fn apply_chunks(
+    /// Applies chunks and processes results
+    fn apply_chunks_and_process_results(
         &mut self,
-        me: &Option<AccountId>,
         block: &Block,
         prev_block: &Block,
-        mode: ApplyChunksMode,
+        work: Vec<Box<dyn FnOnce() -> Result<ApplyChunkResult, Error> + Send + 'static>>,
```
matklad, line 2991: The `FnOnce() -> Result` signature worries me. It means that we close over some state to do something side-effectful later. That has two problems:

* The state is effectively invisible -- looking at this, it's clear that *something* is going to be modified, but I can't understand what that would be without looking at the call-site which creates the `work`
* The asynchronosy here means that we might have potential issue with capturegin state at one point in time and looking at it later. 

I think the part that carries mutability here is runtime_adapter, so maybe change this to 

```rust
FnOnce(&dyn RuntimeAdapter) -> Result<ApplyChunkResult, Error> + Send + 'static>
```

?

chain/chain/src/store.rs
```diff
@@ -2933,6 +2933,78 @@ impl<'a> ChainStoreUpdate<'a> {
...
+            add_state_dl_infos: self.add_state_dl_infos,
+            remove_state_dl_infos: self.remove_state_dl_infos,
+            challenged_blocks: self.challenged_blocks,
+        }
+    }
+}
+
+/// Saves changes from ChainStoreUpdate without link to ChainStore. Needed to preserve changes while
+/// applying block in other thread
+pub struct SavedStoreUpdate {
```
matklad, line 2963: Couple of alternatives to avoid copy-paste:

a) Change `ChainStoreUpdate` to be 

```
ChainStoreUpdate<'a> {
  chain_store: &'a mut ChainStore,
  data: SavedStoreUpdate // but with some different name
}
```

b) change it to be 

```
type ChainStoreUpdate<'a>  = ChainStoreUpdateImpl<&'a mut ChainStore> ;
type SavedChainStoreUpdate<'a>  = ChainStoreUpdateImpl<()> ;
pub struct ChainStoreUpdateImpl<T> {
    chain_store: T,
    store_updates: Vec<StoreUpdate>,
```

chain/chain/src/chain.rs
EgorKulikov: The only real mutability inside is on store, not on RuntimeAdapter. I'll think about your suggestion though

chain/chain/src/store.rs
EgorKulikov: b look nice, will try to do it

chain/client/src/client_actor.rs
```diff
@@ -179,8 +185,27 @@ impl ClientActor {
...
+            }),
+            block_catch_up_scheduler: Box::new(move |msg: BlockCatchUpRequest| {
+                if let Err(err) = sync_jobs_actor_addr2.try_send(msg) {
+                    match err {
+                        SendError::Full(request) => {
+                            sync_jobs_actor_addr2.do_send(request);
+                        }
+                        SendError::Closed(_) => {
+                            panic!("Can't send message to SyncJobsActor, mailbox is closed");
+                        }
```
bowenwang1996, line 207: this code could probably be refactored to avoid duplication?

chain/client/src/client_actor.rs
```diff
@@ -1227,23 +1252,23 @@ impl ClientActor {
...
+            Err(err) => self.handle_catchup_error(err),
         }
 
         near_performance_metrics::actix::run_later(
             ctx,
             file!(),
             line!(),
-            self.client.config.catchup_step_period,
+            // self.client.config.catchup_step_period,
+            Duration::from_millis(1),
```
bowenwang1996, line 1271: Please fix this

chain/client/src/client_actor.rs
```diff
@@ -1512,6 +1537,10 @@ impl ClientActor {
             },
         );
     }
+
+    fn handle_catchup_error(&self, err: Error) {
+        error!(target: "client", "{:?} Error occurred during catchup for the next epoch: {:?}", self.client.validator_signer.as_ref().map(|vs| vs.validator_id()), err);
+    }
```
bowenwang1996, line 1543: is this intended? If so, the name of the function is not optimal -- there is no handling here since it simply logs the error

chain/client/src/sync.rs
```diff
@@ -823,13 +823,24 @@ impl StateSync {
                             status: ShardSyncStatus::StateSplit,
                         }
                     } else {
+                        *shard_sync_download = ShardSyncDownload {
+                            downloads: vec![],
+                            status: ShardSyncStatus::BlocksCatchingUp,
+                        };
```
bowenwang1996, line 829: Ideally we should separate the two since catching up block is functionally not part of state sync. Mixing them together also makes state sync even more complex and harder to reason about

EgorKulikov: We need a way to skip through state downloading part of sync without doing anything, this seems the easiest way to do that

bowenwang1996: > We need a way to skip through state downloading part of sync without doing anything

What is the reason for that?

chain/chain/src/chain.rs
```diff
@@ -1903,81 +1903,111 @@ impl Chain {
...
+                        );
+                    }
+                    blocks_catch_up_state.done_blocks.push(queued_block);
+                }
+                Err(_) => {
+                    error!("Error processing block during catch up, retrying");
+                    blocks_catch_up_state.pending_blocks.push(queued_block);
+                }
+            }
+        }
```
bowenwang1996, line 1937: Is it possible that a malicious attacker can exploit this to cause a node to run out of memory?

chain/client/src/client_actor.rs
```diff
@@ -179,8 +185,27 @@ impl ClientActor {
             chunk_request_retry_next_attempt: now,
             sync_started: false,
             state_parts_task_scheduler: Box::new(move |msg: ApplyStatePartsRequest| {
-                if let Err(_) = state_parts_actor_addr.try_send(msg) {
-                    panic!("Can't send message to StatePartsActor");
+                if let Err(err) = sync_jobs_actor_addr.try_send(msg) {
```
mzhangmzz, line 188: I think you don't need the `move` here, capturing by reference seems fine to me. Then you don't need to clone the address.

chain/client/src/client_actor.rs
mzhangmzz: Agreed, especially we will add another type in the future for state split jobs.

chain/client/src/sync.rs
mzhangmzz: Yeah I'm not sure how the `BlocksCatchingUp` logic here actually does anything. I don't see it used anywhere else (maybe I am missing something). So it seems to just add a stage after either `StateSplit` or `StateDownloadComplete`, in which nothing happens?

Anyway, I agree with Bowen that this logic here is quite confusing.

chain/chain/src/chain.rs
```diff
@@ -1903,81 +1903,111 @@ impl Chain {
...
+            let work = chain_update.apply_chunks_preprocessing(
+                me,
+                &block,
+                &prev_block,
+                ApplyChunksMode::CatchingUp,
+            )?;
+            blocks_catch_up_state
+                .scheduled_blocks
+                .insert(pending_block, chain_update.save_store_update());
+            block_catch_up_scheduler(BlockCatchUpRequest {
```
mzhangmzz, line 1953: I'm not familiar with the Arbiter interface. What kind of guarantee does it make for the ordering of sending and receiving messages? It's very important here that the ordering by which `pending_blocks` get processed are correct since blocks must be processed sequentially.

mzhangmzz: Actually I think the for loop here is confusing and may be incorrect. I don't think more than one blocks should be send to `SyncJobActor` at the same time, the request will simply fail because the information needed for the later block is not ready in the database yet. Processed block information will only be stored to the database in `block_catch_up_postprocess`, and a block must be post-processed before the next block can be processed. 


A more correct and simpler logic would work like this. There should be a queue for pending blocks, and an `Option` for processed block since there will be at most one processed block at a time. At every step, this function checks if there is an processed block, if so, it postprocesses that block and sends the next block from the pending queue to be processed.  At any given point, there should be at most one block being processed

mzhangmzz (commented):
I have a high level comment.

I understand that a lot of the separation of the preprocess and postprocess logic is because the `SyncJobsArbiter` doesn't have access to chain information. I think it would significantly simplifies the logic if `SyncJobsArbitor` holds its own copy of `ChainStore`, which is just store plus some cache of some chain information, so it's fine to have two copies of it. `SyncJobsArbitor` can then create its own `ChainStoreUpdate` object for each request it receives and returns the `ChainStoreUpdate` as response. Then `ClientActor` can commit this `ChainStoreUpdate`. This way, all updates to the database are still only done by `ClientActor`, but there is a lot more we can move to `SyncJobsArbiter`

chain/chain/src/chain.rs
EgorKulikov: We only send more than one block per sync_hash if there is a fork. We do not send any block for pre-processing unless we post-processed its predecessor. So we do not care about any guarantees on order of message processing

chain/client/src/client_actor.rs
EgorKulikov: But I think this closure outlive any address object because it will be destroyed at the end of new method? I may be wrong here

chain/chain/src/chain.rs
EgorKulikov: I am not sure how exactly it can do it. Anyway, I just preserved our current behavior, although I am not sure if it is correct

chain/client/src/sync.rs
EgorKulikov: Actually our current implementation is buggy if there are more than 2 shards. If we need different number of sync steps for different shards (due to download speed, for example), at some point one for the shards would get into StateDownloadComplete state earlier then others, and on next step it would execute code that is executed under this state again. Although I agree that maybe we should name new enum variant some other way to indicate we are just done with this shard and nothing need to be done here

chain/chain/src/chain.rs
```diff
@@ -4165,6 +4269,12 @@ impl<'a> ChainUpdate<'a> {
     }
 }
 
+pub fn do_apply_chunks(
+    work: Vec<Box<dyn FnOnce() -> Result<ApplyChunkResult, Error> + Send>>,
+) -> Vec<Result<ApplyChunkResult, Error>> {
+    work.into_par_iter().map(|task| task()).collect::<Vec<_>>()
```
matklad, line 4275: :thinking: this `par_iter` in chain is inconsistent with a similar construct in runtime:

https://github.com/matklad/nearcore/blob/64e4ee198866135863975cb5b4daf0438e9adbbc/nearcore/src/runtime/mod.rs#L609-L616

Here, we use all the CPU cores to run the tasks.

There, we use half of the cores to run the tasks. 

If the two activities can actually happen concurrently, we will be using 1.5x threads total, which is clearly suboptimal. 

I wonder if we should change the runtime code to also just use global threadpool, but just limit the number of threads there at the startup.

EgorKulikov: This is probably out of scope of this PR as this part is just refactoring. Can you create issue for this?

matklad: https://github.com/near/nearcore/issues/4882

chain/client/src/client_actor.rs
mzhangmzz: Ah that is true

chain/chain/src/chain.rs
mzhangmzz: Ah I forgot about forks, yeah now the implementation makes sense to me, sorry for the confusion

chain/client/src/sync.rs
mzhangmzz: Ah that's a good catch! 

Maybe we should name the new stage to `StateSyncDone`? That signifies that it's the end of state sync for this shard.

EgorKulikov: I like this name

chain/chain/src/chain.rs
```diff
@@ -4197,3 +4307,66 @@ pub struct ApplyStatePartsResponse {
     pub shard_id: ShardId,
     pub sync_hash: CryptoHash,
 }
+
+#[derive(Message)]
+#[rtype(result = "()")]
```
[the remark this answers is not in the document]
EgorKulikov: I am not sure that is what we need as this is actually used to get future and then await on it

EgorKulikov:
> I have a high level comment.
> 
> I understand that a lot of the separation of the preprocess and postprocess logic is because the `SyncJobsArbiter` doesn't have access to chain information. I think it would significantly simplifies the logic if `SyncJobsArbitor` holds its own copy of `ChainStore`, which is just store plus some cache of some chain information, so it's fine to have two copies of it. `SyncJobsArbitor` can then create its own `ChainStoreUpdate` object for each request it receives and returns the `ChainStoreUpdate` as response. Then `ClientActor` can commit this `ChainStoreUpdate`. This way, all updates to the database are still only done by `ClientActor`, but there is a lot more we can move to `SyncJobsArbiter`

I am not sure. For me it is much easier understand this way, everything is isolated, we only have chain level structures in one thread. What other think? @bowenwang1996 @matklad

bowenwang1996:
> we only have chain level structures in one thread

That is unfortunately already not the case. Both `Client` and `ViewClient` hold their own copy of `Chain`

chain/chain/src/chain.rs
bowenwang1996: @matklad I don't understand rayon very well, but is it correct that this code uses `work.len()` number of threads?

chain/client/src/client_actor.rs
```diff
@@ -1583,6 +1612,40 @@ impl Handler<ApplyStatePartsResponse> for ClientActor {
...
+            let saved_store_update = blocks_catch_up_state
+                .scheduled_blocks
+                .remove(&msg.block_hash)
+                .expect("block caught up, but is not in processing");
+            blocks_catch_up_state
+                .processed_blocks
+                .insert(msg.block_hash, (saved_store_update, msg.results));
+        } else {
+            panic!("block catch up processing result from unknown sync hash");
+        }
```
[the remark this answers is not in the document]
EgorKulikov: We can just log error, I think. Though this returns us to the topic of how we want to process errors

EgorKulikov:
> I would like to see a test where we run catch up with a relatively large state. We could do this in a separate PR

#4888

chain/chain/src/chain.rs
EgorKulikov: I think it uses min(work.len(), number of cores) threads

matklad: To be more precise:

Rayon has a global work-stealing thread pool. It's main primitive is scheduling a task for concurrent execution. The task then gets picked up by the most convenient thread from the thread pool. The threads dynamically redistribute the tasks among themselves.

So, what *this* code is doing is just slicing up the vector into a bunch of independent tasks. 

How exactly the tasks are mapped to threads depends on the details of the global threadpool. If we use default theradpool, and it is idle, then we'll see n_cpus threads churning through the tasks (and the threads won't be spawned just for this specific task, they will be pre-existing). In, in contrast, the thread pool is busy by executing tasks from some unrelated subsistem, this code might end up being executed by the single thread.

bowenwang1996 (approved):

bowenwang1996:
@chefsale @frol @khorolets please review the PR

chain/chain/src/chain.rs
EgorKulikov: I actually think it is best to remove Runtime from SyncJonsActor at all and pass it when needed, this seems more clean to me. What do you think? (although this probably out of scope for this PR)

chefsale (approved):

chefsale:
Near-testlibs are failing, please take a look into that. I restarted them, but seems consistent.

EgorKulikov:
> Near-testlibs are failing, please take a look into that. I restarted them, but seems consistent.

They are failing for all PRs actually, I am investigating

chefsale:
I believe this change is the cause of the failing tests: https://github.com/near/near-api-js/commit/9cbc119b0fcf189b4b831d3c05c1b8299d85d960. cc: @bowenwang1996 @EgorKulikov @volovyk-s

chefsale:
Created a comment: https://github.com/near/sdk-docs/pull/49#issuecomment-930227241 and a revert  PR: https://github.com/near/near-api-js/pull/696

chefsale:
Reverted: https://github.com/near/near-api-js/pull/696 @RNHTTR @bowenwang1996 @EgorKulikov

chain/chain/src/chain.rs
matklad: Sounds good to me! As far as I understand, `Runtime` is the main state-holding struct, and it's better if there's a single root of all the state. Ie, two actors, each of which owns an Arc to runtime dosen't feel right, it's better if there's only one actor that owns the thing, and just passes references to (snaphots of) it to other things whch want to do stuff in background.

EgorKulikov: #4898

chain/chain/src/store.rs
```diff
@@ -2988,6 +2990,62 @@ impl<'a> ChainStoreUpdate<'a> {
     }
 }
 
+impl Into<SavedStoreUpdate> for ChainStoreUpdate<'_> {
```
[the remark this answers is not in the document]
EgorKulikov: No specific reason, no, just my general ignorance of some rust features

chain/client/src/client.rs
```diff
@@ -73,8 +75,9 @@ pub struct Client {
     /// Approvals for which we do not have the block yet
     pub pending_approvals: SizedCache<ApprovalInner, HashMap<AccountId, (Approval, ApprovalType)>>,
     /// A mapping from a block for which a state sync is underway for the next epoch, and the object
-    /// storing the current status of the state sync
-    pub catchup_state_syncs: HashMap<CryptoHash, (StateSync, HashMap<u64, ShardSyncDownload>)>,
+    /// storing the current status of the state sync and blocks catch up
+    pub catchup_state_syncs:
+        HashMap<CryptoHash, (StateSync, HashMap<u64, ShardSyncDownload>, BlocksCatchUpState)>,
```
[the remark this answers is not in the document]
EgorKulikov: #4904

PR #4867, opened 2021-09-23
Compare chunk extra with the existing chunk extra, when replaying the…

nikurt:
… chain history.

### DIFF
--- a/test-utils/state-viewer/src/main.rs
+++ b/test-utils/state-viewer/src/main.rs
@@ -7,14 +7,14 @@ use std::path::{Path, PathBuf};
 use std::sync::Arc;
 
 use ansi_term::Color::Red;
+use borsh::BorshSerialize;
 use clap::{App, AppSettings, Arg, SubCommand};
 use tracing::info;
 
-use borsh::BorshSerialize;
+use near_chain::{ChainStore, ChainStoreAccess, ChainStoreUpdate, RuntimeAdapter};
 use near_chain::chain::collect_receipts_from_response;
 use near_chain::migrations::check_if_block_is_first_with_chunk_of_version;
 use near_chain::types::{ApplyTransactionResult, BlockHeaderInfo};
-use near_chain::{ChainStore, ChainStoreAccess, ChainStoreUpdate, RuntimeAdapter};
 use near_epoch_manager::EpochManager;
 use near_logger_utils::init_integration_logger;
 use near_network::peer_store::PeerStore;
@@ -26,10 +26,10 @@ use near_primitives::serialize::to_base;
 use near_primitives::shard_layout::ShardUId;
 use near_primitives::state_record::StateRecord;
 use near_primitives::trie_key::TrieKey;
-use near_primitives::types::chunk_extra::ChunkExtra;
 use near_primitives::types::{BlockHeight, ShardId, StateRoot};
-use near_store::test_utils::create_test_store;
+use near_primitives::types::chunk_extra::ChunkExtra;
 use near_store::{create_store, Store, TrieIterator};
+use near_store::test_utils::create_test_store;
 use nearcore::{get_default_home, get_store_path, load_config, NearConfig, NightshadeRuntime};
 use node_runtime::adapter::ViewRuntimeAdapter;
 use state_dump::state_dump;
@@ -307,6 +307,8 @@ fn apply_chain_range(
     }
     let mut applied_blocks = 0;
     let mut skipped_blocks = 0;
+    let mut different_outcomes_cnt = 0;
+    let mut same_outcomes_cnt = 0;
     for height in start_height..=end_height {
         let block_hash = if let Ok(block_hash) = chain_store.get_block_hash_by_height(height) {
             block_hash
@@ -329,7 +331,7 @@ fn apply_chain_range(
                 .clone();
 
             let prev_block = if let Ok(prev_block) =
-                chain_store.get_block(&block.header().prev_hash())
+            chain_store.get_block(&block.header().prev_hash())
             {
                 prev_block.clone()
             } else {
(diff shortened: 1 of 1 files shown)


test-utils/state-viewer/src/apply_chain_range.rs
```diff
@@ -0,0 +1,178 @@
...
+        let (outcome_root, _) =
+            ApplyTransactionResult::compute_outcomes_proof(&apply_result.outcomes);
+        let chunk_extra = ChunkExtra::new(
+            &apply_result.new_root,
+            outcome_root,
+            apply_result.validator_proposals,
+            apply_result.total_gas_burnt,
+            near_config.genesis.config.gas_limit,
+            apply_result.total_balance_burnt,
+        );
```
bowenwang1996, line 164: This is where the problem is. When the block contains no new chunk, we copy the previous chunk extra and update state_root

bowenwang1996 (commented):
Please add a test :)

test-utils/state-viewer/src/apply_chain_range.rs
nikurt: Fixed, thanks!

test-utils/state-viewer/src/apply_chain_range.rs
```diff
@@ -0,0 +1,288 @@
...
+    fn safe_produce_blocks(
+        env: &mut TestEnv,
+        initial_height: BlockHeight,
+        num_blocks: BlockHeightDelta,
+    ) {
+        let mut h = initial_height;
+        for _ in 1..=num_blocks {
+            let mut block = None;
+            // `env.clients[0]` may not be the block producer at `h`,
+            // loop until we find a height env.clients[0] should produce.
```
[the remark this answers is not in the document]
nikurt: Will look into it.
I'm also struggling to make a test that panics.

test-utils/state-viewer/Cargo.toml
```diff
@@ -9,6 +9,8 @@ ansi_term = "0.12"
 borsh = "0.9"
 clap = "2.33"
 tracing = "0.1"
+indicatif = {version = "0.16.2", features = ["rayon"]}
```
matklad, line 12: TBH, not a huge fun indicatif for *this* use case. 

* It can be slow if used naively. For me, `genesis-populate`  takes 7 seconds with a progress bar and 6 without. That's unreasonably high cost for a UI-nicety. 
* It has (had?) some funky buffering inside, which eat printfs and panic messages: https://github.com/mitsuhiko/indicatif/issues/121. 
* for non-polished dev-tools, to much output is a significantly bigger problem than too little output. Its best to stick to few log lines if possible. 


Where indicatiff shines is for building delightful tools for the end user, where you think very carefully about the overall output and judiciously place just the right amount of prints. For internal dev tooling, less is more.

matklad: To clarify, that's my personal opinion, and not a very strong one! I am fine with keeping it!

test-utils/state-viewer/src/apply_chain_range.rs
```diff
@@ -0,0 +1,288 @@
...
+
+    debug!("============================");
+    debug!("Printing results including outcomes of applying receipts");
+
+    let progress_bar = ProgressBar::new(end_height - start_height + 1);
+    progress_bar.set_style(
+        ProgressStyle::default_bar().template("[{elapsed_precise}] {bar:80} {pos:>8}/{len:8}"),
+    );
+    let processed_blocks_cnt = AtomicU64::new(0);
+    // For some reason I can't call `par_iter` directly on the range.
```
matklad, line 52: You want `into_par_iter` probably

test-utils/state-viewer/src/apply_chain_range.rs
```diff
@@ -0,0 +1,288 @@
...
+
+            let prev_block = if let Ok(prev_block) =
+            chain_store.get_block(&block.header().prev_hash())
+            {
+                prev_block.clone()
+            } else {
+                info!(target:"state-viewer", "Skipping applying block #{} because the previous block is unavailable and I can't determine the gas_price to use.", height);
+                inc_and_report_progress(&processed_blocks_cnt);
+                return;
+            };
```
matklad, line 90: very weak presonal opinion: `if let {} else {}` most of the time is less readable than the equivalent match 

https://github.com/rust-analyzer/rust-analyzer/blob/master/docs/dev/style.md#if-let

test-utils/state-viewer/src/apply_chain_range.rs
```diff
@@ -0,0 +1,288 @@
...
+                    &block.header().challenges_result(),
+                    *block.header().random_value(),
+                    false,
+                    false,
+                    None,
+                )
+                .unwrap()
+        };
+
+        println!("outcomes: {:#?}", apply_result.outcomes);
```
matklad, line 159: It feels a bit odd that we use all three of `tracing`, `indicatiff`, `println` to print stuff to the terminal. I'd probaby try to stick to prints for all output (but, again, not that it matters a lot)

Cargo.lock
```diff
@@ -2135,7 +2135,20 @@ checksum = "7baab56125e25686df467fe470785512329883aab42696d661247aca2a2896e4"
 dependencies = [
  "console",
  "lazy_static",
- "number_prefix",
+ "number_prefix 0.3.0",
+ "regex",
+]
+
+[[package]]
+name = "indicatif"
```
matklad, line 2143: suggestion(blocking): Regardless of whether we use or not use indicatiff, we shouldn't be adding a second copy of it. See indicatiff 0.13 above. So, we should either use 0.13 here and then send a follow-up to update  both cases, or we should first send an update PR for another usage, and then rebase this work on top.

PR #4868, opened 2021-09-23
runtime-tester: add support for inspecting the home directory

mina86:
Currently, scenarios using on-disk storage use a temporary directory
which is deleted once the run finishes.  This is correct behaviour but
it makes it impossible to investigate the storage of a failing (or
otherwise misbehaving) test.

Change Scenario::run such that it returns the temporary directory
where the storage is held.  This way, caller can choose to hold onto
the directory and thus prevent it from being deleted.

Issue: https://github.com/near/nearcore/issues/4771


test-utils/runtime-tester/src/run_test.rs
```diff
@@ -86,6 +90,26 @@ pub struct Scenario {
...
+    /// This is useful when wanting to save the storage on-disk and investigate
+    /// it after the test runs and it’s also why the directory is not cleared.
+    ///
+    /// Note that this field is not serialised thus a scenario read from file
+    /// won’t have this path set.  This is motivated partially by security
+    /// considerations since if this could be stored in JSON file than user
+    /// might be tricked to run a test against `/home/foo/.near/mainnet`
+    /// directory.
+    #[serde(skip_serializing)]
+    pub home_dir: Option<std::path::PathBuf>,
```
matklad, line 112: `#[serde(skip_serializing)]`  feels like a smell. What if we had two `run` methods instead: `run(self)`, `run_with_home_dir(self, home: &Path)`?.

test-utils/runtime-tester/src/scenario_builder.rs
```diff
@@ -68,6 +73,21 @@ impl ScenarioBuilder {
...
+    /// `in_memory_store(false)`.  Setting home directory means that the caller
+    /// is responsible for cleaning the directory before and after the test is
+    /// run.  This setting is intended only for one-off cases where someone
+    /// wants to manually inspect the store and thus does not want the home
+    /// directory to be deleted after the test finishes.
+    pub fn home_dir(mut self, home_dir: impl Into<std::path::PathBuf>) -> Self {
+        self.scenario.use_in_memory_store = false;
+        self.scenario.home_dir = Some(home_dir.into());
+        self
+    }
```
matklad, line 89: An alternative would be to add just `fn keep_temp_dir(self, yes: bool)` and print the path at the end of the test. I usually do that approach, as that way I don't have to think about where to put it, and just let the computer decide

https://docs.rs/tempfile/3.2.0/tempfile/struct.TempDir.html#method.into_path

test-utils/runtime-tester/src/run_test.rs
mina86: I don’t really like that ’cuz then we end up with combinatoric explosion of different `run` methods. How about just always returning the temporary directory and letting the caller decide whether they want to use it or ignore.

matklad: > How about just always returning the temporary directory and letting the caller decide whether they want to use it or ignore.

Genius idea!

matklad (approved):

PR #4869, opened 2021-09-24
change how outgoing receipts are stored and processed 

mzhangmzz:
This PR changes how outgoing receipts are stored and processed in preparation of sharding layout change

There are two changes involved:
- get_outgoing_receipts_for_shard now can handle the case when shard layout changes from epoch to epoch
- outgoing_receipts and receipt_id_to_shard_id are stored separately. outgoing_receipts are stored at the block when receipts are generated while receipt_id_to_shard_id are stored at the chunk where they are included as incoming receipts. This is because to sort receipts by shard, we need the shard layout for the next block, which is hard to get at the time when the receipts are generated


nearcore/src/runtime/mod.rs
```diff
@@ -530,30 +530,23 @@ impl NightshadeRuntime {
                 ErrorKind::Other("Integer overflow during burnt balance summation".to_string())
             })?;
 
-        // Sort the receipts into appropriate outgoing shards.
-        let mut receipt_result = HashMap::default();
-        // Outgoing receipts should be sorted by shards of the epoch of the next block
-        let next_shard_layout = {
```
mzhangmzz, line 536: This part of logic was wrong. Here we are getting the shard layout of the next block of "prev_block_hash", which is the shard layout of the current block. However, we need to get the shard layout of the next block after the current block. Since the current block info has not been stored at this point, it is very hard to get that info here

mzhangmzz: It is also probably better in terms of code modularity to move the part where outgoing receipts are sorted by destination shard outside of runtime code since runtime should not care about shard layout at all, which should be handled on the chain level. (I know following the same logic, the code for splitting states here should also be moved to runtime, I just haven't have time to do so)

chain/chain/src/validate.rs
```diff
@@ -65,29 +68,28 @@ pub fn validate_chunk_proofs(chunk: &ShardChunk, runtime_adapter: &dyn RuntimeAd
...
-        return receipts.len() == 0 && outgoing_receipts_root == CryptoHash::default();
+        return Ok(receipts.len() == 0 && outgoing_receipts_root == CryptoHash::default());
     } else {
-        let epoch_id = {
+        let shard_layout = {
             let prev_block_hash = match chunk {
                 ShardChunk::V1(chunk) => chunk.header.inner.prev_block_hash,
                 ShardChunk::V2(chunk) => chunk.header.prev_block_hash(),
             };
-            runtime_adapter.get_epoch_id_from_prev_block(&prev_block_hash).unwrap()
```
mzhangmzz, line 79: Not really relevant to this PR, but I spot a few unwraps here added by me before and decided to correct them here

matklad: I am actually thinking if perhaps such unwraps are fine? 

That is, I *think* our general strategy is "node should shut-down itself in the face of catastrophic failures". If that's indeed the case, I see an argument for initiating shutdown at the place where the error originates, not at the top-level where we handle the errors. 

There's a certain beauty if, in the API level, the database is reliable and never returns an `io::Error`. 


Though, this is utterly unrelate to the PR, and we already generally just `?` errors

core/primitives/src/shard_layout.rs
```diff
@@ -63,7 +67,29 @@ impl ShardLayout {
...
+        let to_parent_shard_map = if let Some(shards_split_map) = &shards_split_map {
+            let mut to_parent_shard_map = HashMap::new();
+            let num_shards = (fixed_shards.len() + boundary_accounts.len() + 1) as NumShards;
+            for (parent_shard_id, shard_ids) in shards_split_map.iter().enumerate() {
+                for &shard_id in shard_ids {
+                    let prev = to_parent_shard_map.insert(shard_id, parent_shard_id as ShardId);
+                    // check that no shard appears in shards_split_map twice
+                    assert!(prev.is_none());
+                    // check shard id is value
+                    assert!(shard_id < num_shards);
```
matklad, line 79: Might be worthwhile to move these comments into assert messages.

core/primitives/src/shard_layout.rs
```diff
@@ -85,6 +111,17 @@ impl ShardLayout {
         }
     }
 
+    #[inline]
+    pub fn get_parent_shard_id(&self, shard_id: ShardId) -> Option<ShardId> {
+        match self {
+            Self::V0(_) => None,
+            Self::V1(v1) => match &v1.to_parent_shard_map {
+                Some(to_parent_shard_map) => Some(to_parent_shard_map[shard_id as usize]),
```
matklad, line 119: Question: should we perhaps `to_parent_shard_map.get(shrd_id)?` here? As in (way out of my depth here), can someone submit us a challenge which would cause us to run this fn with large `shard_id` which is out of bounds?

core/primitives/src/sharding.rs
```diff
@@ -919,7 +919,7 @@ impl EncodedShardChunkBody {
 }
 
 #[derive(BorshSerialize, Debug, Clone)]
-pub struct ReceiptList<'a>(pub ShardId, pub &'a Vec<Receipt>);
+pub struct ReceiptList<'a>(pub ShardId, pub &'a [Receipt]);
```
matklad, line 922: :+1:

nearcore/src/runtime/mod.rs
matklad: Do we have a test which we can add to the old codebase which would surface this bug? Is such test added in this PR?

chain/chain/src/store.rs
```diff
@@ -440,25 +440,70 @@ impl ChainStore {
...
+                    runtime_adapter.get_shard_layout(block_header.epoch_id())?;
+
+                // get the shard from which the outgoing receipt were generated
+                let receipts_shard_id = if shard_layout != receipts_shard_layout {
+                    // we unwrap here because otherwise it means the shard layout is invalid
+                    shard_layout.get_parent_shard_id(shard_id).unwrap()
+                } else {
+                    shard_id
+                };
                 let receipts = if let Ok(cur_receipts) =
```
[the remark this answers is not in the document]
matklad: http://sled.rs/errors.html

```rust
enum Error {
  Local(LocalError),
  Fatal(FatalError),
}
```

it feels exactly like what we are doing here :D

nearcore/src/runtime/mod.rs
mzhangmzz: This bug will only surface when shard layout changes. I haven't finished the full implementation of splitting shards, so we don't have end to end integration test for sharding upgrade now, but I will add that next week after I'm done with implementation

chain/chain/src/store.rs
```diff
@@ -440,25 +440,70 @@ impl ChainStore {
...
+    // in the future
     pub fn get_outgoing_receipts_for_shard(
         &mut self,
+        runtime_adapter: &dyn RuntimeAdapter,
         prev_block_hash: CryptoHash,
         shard_id: ShardId,
         last_included_height: BlockHeight,
-    ) -> Result<ReceiptResponse, Error> {
+    ) -> Result<Vec<Receipt>, Error> {
+        let shard_layout = runtime_adapter.get_shard_layout_from_prev_block(&prev_block_hash)?;
```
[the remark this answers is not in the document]
mzhangmzz: I totally agree! I think the ultimate way to solve this is to move this function (and other functions related to epoch information) outside of runtime. Logically, runtime should not care about epoch information at all. The current design of putting EpochManager inside runtime causes a lot of awkward code pieces like this. Code outside of runtime need to get access to epoch info through runtime, code within runtime can get exposure to epoch or shard layout info where they shouldn't

chain/chain/src/store.rs
```diff
@@ -964,17 +1009,17 @@ impl ChainStoreAccess for ChainStore {
 
     fn get_outgoing_receipts(
         &mut self,
-        block_hash: &CryptoHash,
+        prev_hash: &CryptoHash,
```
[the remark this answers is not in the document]
mzhangmzz: I'll add a comment here

core/primitives/src/shard_layout.rs
mzhangmzz: Good point, I think such it shouldn't be possible for such challenges to reach code places like this because we would check number of shards before calling this function. But yeah that's not obvious to see here. I will change it.

chain/chain/src/store.rs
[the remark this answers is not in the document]
mzhangmzz: Yeah I also found it weird. I think the error should be propagated through. I don't want to change it in this PR though since I want to preserve the behavior. I will change it in another PR.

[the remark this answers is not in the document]
mzhangmzz: Yeah personally I have always struggled to find a good way to understand how errors should be handled. Maybe this deserves a larger scope discussion so that everyone working on the code is on the same page for error handling. @bowenwang1996

chain/chain/src/validate.rs
mzhangmzz: I think the struggle here is that it's hard to distinguish from within the function whether the error is catastrophic. I always struggled to find a good way to deal with this, and I am glad you brought this up. 

Focusing on this specific case, this error will occur if by the time this function is called, `prev_block_hash` is not processed yet. This almost should never happen because we would probably only validate chunks after the previous block is processed. But it's hard to guarantee that. Unwrapping here seems risky because all callers of this function then need to know that this function may panic in the case of `prev_block` not processed. What if someone uses this function incorrectly in the future, or maybe simply I just understand this function wrong. On the other hand, passing along the the error also doesn't seem ideal because it could be a critical failure (maybe the database is corrupt) and it is easy for errors to get lost and not handled correctly if we keep propagating them.

mzhangmzz:
@frol I would like your opinion on my change on `receipt_id_to_shard_id`, to my understanding, that only affects functions like `get_tx_status`. So this PR changes when `receipt_id_to_shard_id` is stored. Let's say some outgoing receipts are generated at block h, before this PR, `receipt_id_to_shard_id` is stored at block h, this PR changes that to happen at block h+1. I thought about it and I think it should be fine, since these receipts won't be processed until block h+1 and I think `receipt_id_to_shard_id` is used to get execution outcomes (please correct me if I am wrong), so the execution outcomes won't be ready anyways at block h.

chain/chain/src/validate.rs
matklad: For general applications, I really love sqlite approach for this(https://www.sqlite.org/assert.html): crash in debug, in release log error and return a "neutral" value (ie, error object). 

For blockchain, not really sure whats the best apporach. 

OTOH, sqlite is aviation grade software, so if thinks it ok to return newtral value, we can do the same.

:thinking: I've just realized that a lot of tradeoffs we face feel similar to those of sqlite, and that is a brilliant piece of engineering. I need to re-read its docs to figure out what we might want to apply in nearcore.

mzhangmzz: Yeah I think that's a great point! With better test coverage, we can afford to be more assertive in error handling. 

At my previous company (a trading firm) we did that too since we don't want the trading program just crash in live but we also don't want miss any catastrophic error and lose a lot of money in live. One thing we did was aggressively assert, which panic in debug but log error in release. We then run lots of simulations (regression tests) using the debug binary to make sure nothing crashes. In live we would log the errors and ops will get sound notifications (something like "oh no we are all doomed" lol) every time when an error is logged. 

Maybe we can apply that to betanet vs testnet/mainnet. In betanet we use the debug binary which will fail aggressively in case of error. 

I think we should also have more "simulation" kind of tests in near which can use real historic data and cover more complicated scenarios (or maybe we already have load test that is similar to that. I am not very familiar with our test framework). But that's more complicated .

bowenwang1996 (commented):
Great work! one question: how does outgoing receipt root verification work when the shard layout changes? Is there a test for this?

bowenwang1996:
> @bowenwang1996 @frol I would need your opinion on my change on receipt_id_to_shard_id, to my understanding, that only affects functions like get_tx_status. This PR changes when receipt_id_to_shard_id is stored. Let's say some outgoing receipts are generated at block h, before this PR, receipt_id_to_shard_id is stored at block h, this PR changes that to happen at block h+1. I thought about it and I think it should be fine, since these receipts won't be processed until block h+1 and I think receipt_id_to_shard_id is used to get execution outcomes (please correct me if I am wrong) and the execution outcomes won't be ready anyways until block h+1.

That sounds fine to me

chain/chain/src/store.rs
```diff
@@ -1824,26 +1861,41 @@ impl<'a> ChainStoreUpdate<'a> {
...
-                self.chain_store_cache_update
-                    .receipt_id_to_shard_id
-                    .insert(receipt.receipt_id, receipt_shard_id);
-                outgoing_receipts.push(receipt);
-            }
-        }
         self.chain_store_cache_update
             .outgoing_receipts
             .insert((*hash, shard_id), outgoing_receipts);
     }
```
bowenwang1996, line 1874: This function does not make sense any more. Why do we only update the cache?

chain/chain/src/validate.rs
```diff
@@ -160,15 +162,15 @@ pub fn validate_chunk_with_chunk_extra(
...
+        prev_chunk_height_included,
     )?;
     let outgoing_receipts_hashes = {
-        let epoch_id = runtime_adapter.get_epoch_id_from_prev_block(prev_block_hash)?;
-        let shard_layout = runtime_adapter.get_shard_layout(&epoch_id)?;
-        runtime_adapter.build_receipts_hashes(&receipt_response.1, &shard_layout)
+        let shard_layout = runtime_adapter.get_shard_layout_from_prev_block(prev_block_hash)?;
+        runtime_adapter.build_receipts_hashes(&outgoing_receipts, &shard_layout)
     };
     let (outgoing_receipts_root, _) = merklize(&outgoing_receipts_hashes);
```
bowenwang1996, line 176: I think this validation is not fully correct. If I understand it correctly, for every split shard it does the same thing -- it tries to get the receipts from before the split and validate that against the outgoing receipt from last epoch before the layout changes. However, there seems to be no validation that the split itself is correct

chain/chain/src/store.rs
mzhangmzz: The cache is a cache for update, it gets committed to the database when you call ChainStoreUpdate.commit

I do think there are too many layers of caches and updates in ChainStore that it does get confusing

chain/chain/src/validate.rs
mzhangmzz: I'm not sure that I understand your question. The `get_outgoing_receipts_for_shard` function will return receipts after the split. Say shard changes after block `prev_hash` and block `hash`, then get_outgoing_receitpts_for_shard` function returns the outgoing receipts for shard `shard_id` in block `hash`

mzhangmzz:
> Great work! one question: how does outgoing receipt root verification work when the shard layout changes? Is there a test for this?

Good question. This PR does not complete the implementation for transition when the shard layout changes, so the part for outgoing receipt root verification is not implemented here. The plan for that is to change how chunk extra of the last block in the last epoch is stored. If you remember, in my last PR, the chunk extra for the split states only stores state root, nothing else. This needs to be changed for the last block, because that chunk extra for split states will be used for chunk header verifications for the first block in the new epoch. I will implement that in a following PR.

chain/chain/src/store.rs
```diff
@@ -440,25 +440,60 @@ impl ChainStore {
...
+    /// to other shards in the block after `prev_block_hash`
+    /// The difference of one block is important because shard layout may change between the previous
+    /// block and the current block and the meaning of `shard_id` will change.
+    ///
+    /// Note, the current way of implementation assumes that at least one chunk is generated before
+    /// shard layout are changed twice. This is not a problem right now because we are changing shard
+    /// layout for the first time for simple nightshade and generally not a problem if shard layout
+    /// changes very rarely.
+    /// But we need to implement a more theoretically correct algorithm if shard layouts will change
+    /// more often in the future
```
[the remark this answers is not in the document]
mzhangmzz: https://github.com/near/nearcore/issues/4877

core/primitives/src/shard_layout.rs
```diff
@@ -63,7 +67,27 @@ impl ShardLayout {
...
+            let mut to_parent_shard_map = HashMap::new();
+            let num_shards = (fixed_shards.len() + boundary_accounts.len() + 1) as NumShards;
+            for (parent_shard_id, shard_ids) in shards_split_map.iter().enumerate() {
+                for &shard_id in shard_ids {
+                    let prev = to_parent_shard_map.insert(shard_id, parent_shard_id as ShardId);
+                    assert!(prev.is_none(), "no shard should appear in the map twice");
+                    assert!(shard_id < num_shards, "shard id should be valid");
+                }
+            }
+            Some((0..num_shards).map(|shard_id| to_parent_shard_map[&shard_id]).collect())
```
EgorKulikov, line 80: Maybe we should also have some informative expect here and not just panic if info for some shard is missing?

EgorKulikov (approved):
I have some reservations about our error handling as a whole, but this is probably a separate topic for discussion, and it seems @matklad already is interested in this. Otherwise great job

bowenwang1996:
> > Great work! one question: how does outgoing receipt root verification work when the shard layout changes? Is there a test for this?
> 
> Good question. This PR does not complete the implementation for transition when the shard layout changes, so the part for outgoing receipt root verification is not implemented here. The plan for that is to change how chunk extra of the last block in the last epoch is stored. If you remember, in my last PR, the chunk extra for the split states only stores state root, nothing else. This needs to be changed for the last block, because that chunk extra for split states will be used for chunk header verifications for the first block in the new epoch. I will implement that in a following PR.

I see. Thanks for the explanation!

bowenwang1996 (approved):

PR #4873, opened 2021-09-24
runtime-tester: use provided account ids when creating test environment

mina86:

### DIFF
--- a/test-utils/runtime-tester/src/run_test.rs
+++ b/test-utils/runtime-tester/src/run_test.rs
@@ -31,10 +31,10 @@ impl Scenario {
     }
 
     pub fn run(&self) -> ScenarioResult<RuntimeStats, Error> {
-        let genesis = Genesis::test(
-            self.network_config.seeds.iter().map(|x| x.parse().unwrap()).collect(),
-            1,
-        );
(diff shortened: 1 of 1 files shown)


test-utils/runtime-tester/src/run_test.rs
```diff
@@ -31,10 +31,10 @@ impl Scenario {
     }
 
     pub fn run(&self) -> ScenarioResult<RuntimeStats, Error> {
-        let genesis = Genesis::test(
-            self.network_config.seeds.iter().map(|x| x.parse().unwrap()).collect(),
-            1,
-        );
+        let accounts: Vec<AccountId> =
+            self.network_config.seeds.iter().map(|x| x.parse().unwrap()).collect();
+        let clients = vec![accounts[0].clone()];
```
matklad, line 36: Is it intentional that we only take one client here?

mina86: Yes. This is what the old code was doing (i.e. creating just one client and setting up one validator seat).

matklad: Ah, inded, I misread the next line!

matklad (approved):

PR #4874, opened 2021-09-24
feat: upgrade to wasmer2 via a protocol feature

matklad:
This feature switch our WASM engine implementation from wasmer 0.* to
wasmer 2.*, brining better performance and reliability.

The implementations should be sufficiently similar for this to not be a
protocol upgrade, but we concervatively do a protocol upgrade to be on
the safe side.

Although wasmer2 is faster, we don't change fees with this protocol
version -- we can safely do that in a separate step.

Test Plan
------------

Add an upgrade test, verify that it passes with and without feature. Verify additionally that it fails with the features if the impl doesn't actually enable wasmer2.


core/primitives/src/version.rs
```diff
@@ -111,6 +111,17 @@ pub enum ProtocolFeature {
...
     LowerDataReceiptCost,
+    /// This feature switch our WASM engine implementation from wasmer 0.* to
+    /// wasmer 2.*, brining better performance and reliability.
+    ///
+    /// The implementations should be sufficiently similar for this to not be a
+    /// protocol upgrade, but we concervatively do a protocol upgrade to be on
+    /// the safe side.
+    ///
+    /// Although wasmer2 is faster, we don't change fees with this protocol
+    /// version -- we can safely do that in a separate step.
```
bowenwang1996, line 122: Yay!

bowenwang1996 (approved):
Masterful!

nearcore/src/runtime/mod.rs
```diff
@@ -614,9 +614,12 @@ impl NightshadeRuntime {
             .unwrap()
             .install(|| {
                 contract_codes.par_iter().for_each(|code| {
+                    // FIXME: fetch the correct protocol version somehow?
+                    let current_protocol_version = ProtocolVersion::MAX;
```
matklad, line 618: Note to self: this is wrong in the absolute terms, but maybe fine in practice? Need to think more about this and write propper comment. here.

bowenwang1996: hmm why don't we simply use `PROTOCOL_VERSION`?

matklad: :facepalm: because I thought that `ProtocolVersion` is a proper struct, and so `ProtocolVersion::MAX` (which I saw used elsewhere) is the correct way to refer to current protocol version, while `PROTOCOL_VERSION` is some temporary constant we use in a specific test. 

But of course that's the other way around -- ProtocolVersion is just u32, and we have `PROTOCOL_VERSION` top-level const because we can't say `ProtocolVersion::latest()`.

matklad: :facepalm: :facepalm: :facepalm: 

There's `let protocol_version =` at the start of this function that gets the right protocol version.

core/primitives/src/version.rs
olonho: Also note that on the weekend we completed full history playback with Wasmer2, so it seems to work fine.

integration-tests/tests/client/process_blocks.rs
```diff
@@ -3825,14 +3924,15 @@ mod contract_precompilation_tests {
             .iter()
             .map(|s| Arc::new(StoreCompiledContractCache { store: s.clone() }))
             .collect();
+        let vm_kind = VMKind::for_protocol_version(PROTOCOL_VERSION);
```
olonho, line 3927: Cool!

runtime/runtime/src/lib.rs
```diff
@@ -2407,11 +2408,8 @@ mod tests {
         store_update.commit().unwrap();
 
         let contract_code = ContractCode::new(wasm_code, None);
-        let key = get_contract_cache_key(
-            &contract_code,
-            VMKind::default(),
-            &apply_state.config.wasm_config,
-        );
+        let vm_kind = VMKind::for_protocol_version(apply_state.current_protocol_version);
+        let key = get_contract_cache_key(&contract_code, vm_kind, &apply_state.config.wasm_config);
```
olonho, line 2412: Maybe pass `apply_state.current_protocol_version` to `get_contract_cache_key` to make selection logic to belong there?

Issue #3838, opened 2021-01-21
Reconsider how we exchange routing table information


bowenwang1996:
Currently we always exchange the entire routing table with peers when new connections are established and broadcast newly received routing table information. This is not efficient and leads to an increase of network traffic. We should consider optimizing this process with better strategies to synchronize routing table between nodes in the network.

*   <span class="error">[x]</span> propose new algorithm - Details of the new algorithm are in #3934
*   <span class="error">[x]</span> Investigate if we verify edges multiple times Should be fixed by: [https://github.com/near/nearcore/pull/4098](https://github.com/near/nearcore/pull/4098)
*   <span class="error">[x]</span> Investigate how edges are stored
*   <span class="error">[x]</span> Make a prototype [https://github.com/near/nearcore/pull/4112](https://github.com/near/nearcore/pull/4112)
*   [ ] Finish optimizing routing table computation
*   [ ] Finish optimizing Validation of edges

bowenwang1996:
@pmnoxx should we close this one in favor of #3934 ?

pmnoxx:
@bowenwang1996 I decided to split issues into two to make it more cleaner. #3934 focuses on theoretical work
This one on the actual implementation.

pmnoxx:
I wrote a prototype: https://github.com/near/nearcore/pull/4103
I'm going to test it and see if it works.

pmnoxx:
There are still a few optimizations we can do to exchange routing tables:
- Improve routing table exchange computation by factor of 4x.
It used to take 1s to compute routing table computation for `100_100` test, now it takes `1ms`, we can finish optimizing it, and make it `0.25ms`.

PR #4878, opened 2021-09-27
ibf: fix [usize; 3] not being an iterator compilation error

mina86:
Fixes the following compiler error:

    $ cargo build --release
    error[E0277]: `[usize; 3]` is not an iterator
       --> chain/network/src/ibf.rs:215:20
    215 |         for pos in pos_list {
        |                    ^^^^^^^^ borrow the array with `&` or call `.iter()` on it to iterate over it

Issue: https://github.com/near/nearcore/issues/3838

### DIFF
--- a/chain/network/src/ibf.rs
+++ b/chain/network/src/ibf.rs
@@ -212,9 +212,9 @@ impl Ibf {
     ) {
         let pos_list = self.generate_idx(elem_hash);
 
-        for pos in pos_list {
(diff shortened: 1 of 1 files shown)


pmnoxx (approved):
LGTM.
@bowenwang1996 Don't we check this in CI?

mina86:
Good point. `cargo test` step should have caught it. I think I might be running ‘ancient’ rustc.

PR #4880, opened 2021-09-27
nightly: remove unnecessary ‘adevrsarial’ feature

mina86:
NayDuck always enables the ‘adversarial’ feature when building ‘neard’.
There’s no need to explicitly enable it.  Doing so forces NayDuck to
unnecessarily create a separate build.

Issue: https://github.com/near/nearcore/issues/3838


pmnoxx (approved):
LGTM.
That makes sense. Thanks for updating README.md

PR #4886, opened 2021-09-28
Move routing table exchange to another thread

pmnoxx:
In this PR we will creating copy of routing table to another thread.


pmnoxx:
@bowenwang1996 This is the first part of the change I'm planning. `get_edges` will be removed, and we will start updating 2nd routing table in another thread by default. Once we ship this PR, I'll submit 2nd PR for fully moving routing table computation to another thread.

pmnoxx:
@bowenwang1996 @mfornet @chefsale @frol @khorolets I added tests, everything works. Can you take a look?

nearcore/src/lib.rs
```diff
@@ -332,8 +333,18 @@ pub fn start_with_config(home_dir: &Path, config: NearConfig) -> NearNode {
...
+    #[cfg(all(feature = "json_rpc", feature = "adversarial"))]
+    let ibf_routing_pool2 = ibf_routing_pool.clone();
     let network_actor = PeerManagerActor::start_in_arbiter(&arbiter.handle(), move |_ctx| {
-        PeerManagerActor::new(store, network_config, client_actor1, view_client1).unwrap()
+        PeerManagerActor::new(
+            store,
+            network_config,
+            client_actor1,
+            view_client1,
+            ibf_routing_pool.clone(),
```
frol, line 345: I feel you don't need to clone it here and below since you already have two instances of `ibf_routing_pool`

frol (approved):

chain/jsonrpc/Cargo.toml
```diff
@@ -41,3 +41,4 @@ dump_errors_schema = ["near-rpc-error-macro/dump_errors_schema"]
 adversarial = ["near-jsonrpc-primitives/adversarial"]
 nightly_protocol = ["near-primitives/nightly_protocol"]
 sandbox = []
+protocol_feature_routing_exchange_algorithm = []
```
bowenwang1996, line 44: Why do we need this?

chain/network/src/peer_manager.rs
```diff
@@ -150,32 +148,44 @@ impl IbfRoutingTableExchangeActor {
...
     AddPeerIfMissing(PeerId, Option<u64>),
+    #[cfg(feature = "protocol_feature_routing_exchange_algorithm")]
     RemovePeer(PeerId),
-    ProcessIbfMessage { peer_id: PeerId, ibf_msg: RoutingVersion2 },
+    #[cfg(feature = "protocol_feature_routing_exchange_algorithm")]
+    ProcessIbfMessage {
+        peer_id: PeerId,
+        ibf_msg: RoutingVersion2,
+    },
+    RequestRoutingTable,
```
bowenwang1996, line 168: I am confused by this change. Shouldn't the entire struct be behind some feature? What is the point of making some of the variants feature gated but not others?

bowenwang1996 (commented):
The feature gating feels arbitrary and inconsistent to me. We should come up with some guidelines on where and how they should be used. cc @matklad

chain/jsonrpc/Cargo.toml
pmnoxx: There are two methods that can't be called unless you enable `protocol_feature_routing_exchange_algorithm`.

chain/network/src/peer_manager.rs
pmnoxx: `IbfRoutingTableExchangeActor` holds information about `IbfSet` data structures, as well as holds `edges_info` with list of all edges. 
`IbfSet` related data structures are behind ` #[cfg(feature = "protocol_feature_routing_exchange_algorithm")]`.
However, now we will always populate copy of routing table ( `edges_info`) on another thread. That will offload work related to copying routing table for purpose of doing exchange to another thread.

I'll rename `IbfRoutingTableExchangeActor`to `RoutingTableActor` to be more descriptive.

pmnoxx: Renaming is done

nearcore/src/lib.rs
pmnoxx: Thanks, fixed.

chain/network/src/peer_manager.rs
```diff
@@ -131,16 +129,16 @@ struct ActivePeer {
     peer_type: PeerType,
 }
 
-#[cfg(feature = "protocol_feature_routing_exchange_algorithm")]
 #[derive(Default)]
-struct IbfRoutingTableExchangeActor {
+pub struct RoutingTableActor {
```
bowenwang1996, line 133: Please add a comment on what the actor is responsible for

chain/network/src/routing.rs
```diff
@@ -797,16 +795,6 @@ impl RoutingTable {
...
-        if force_pruning {
-            let mut edges_to_remove2 = vec![];
-            for (k, edge) in self.edges_info.iter() {
-                if !self.peer_last_time_reachable.contains_key(&k.0)
-                    || !self.peer_last_time_reachable.contains_key(&k.1)
-                {
-                    edges_to_remove2.push(edge.clone());
-                }
-            }
-        }
```
bowenwang1996, line 809: why is this piece of code removed?

pmnoxx: This code is only executed during tests, and it isn't needed any more.

chefsale (approved):

chain/network/src/test_utils.rs
```diff
@@ -281,12 +282,17 @@ impl MockNetworkAdapter {
     }
 }
 
+pub fn make_ibf_routing_pool() -> Addr<RoutingTableActor> {
+    SyncArbiter::start(1, move || RoutingTableActor::default())
```
matklad, line 286: How do we stop the thread we spawn here?

chain/jsonrpc/Cargo.toml
matklad, line 44: Super high level question, but why is it a protocol feature? Intuitively, the way peers communicate with each other feels like it shouldn't affect protocol per-se? We don't actually look at the runtime protocol version here at all. 

I guess here we just piggy-back onto the protocol featue infrastructure, just to avoig introducing ad-hoc feature for this? Migh make sense to explain this in a comment then.

chain/jsonrpc/src/lib.rs
```diff
@@ -1295,6 +1327,7 @@ pub fn start_http(
     client_addr: Addr<ClientActor>,
     view_client_addr: Addr<ViewClientActor>,
     #[cfg(feature = "adversarial")] peer_manager_addr: Addr<PeerManagerActor>,
+    #[cfg(feature = "adversarial")] ibf_routing_pool: Addr<RoutingTableActor>,
```
matklad, line 1330: Why do we need to pass `ibf_routing_pool` here? We already have peer_manager's address, and peer_manager has ibf pool already. Can we communicate with peer manager for adversarial needs?

It doesn't feel right that we need to thread ibf_pool, which is a private impl detail, up to top-level start_http call. #[cfg()] function parameters are also odd to work with. 


I wonder...  should we perhaps instead make existing `AdversarialControls` a singleton instead, accessible everywhere? I think for such kind of code globals are better than threading impl details everywhere. Eg, tikvs failpoints are global-based: https://docs.rs/fail/0.4.0/fail/. Although it probably still better not to make it a literal static variable (which doesn't jive well with threading), but just pass an `Arc<AdversarialControlls>` everywhere we need them?

pmnoxx: This is needed for tests. After this change we will have copy of routing table in both `RoutingTableActor` and `PeerManagerActor`. I modified a python test, that uses jsonrpc, to request routing table  from both `RoutingTableActor`, `PeerManagerActor`, there was actually a bug, where those two wouldn't be in sync.

Unfortunately, actors `handle` api is synchronous, so there is no way for `PeerManager` to return result from `RoutingTableActor` in `handle` method. The only way is to modify `jsonrpc` server to ask `RoutingTableActor` directly.

chain/jsonrpc/Cargo.toml
pmnoxx: `RoutingTableActor` was previously only enabled under `adversarial` flag. It's going to be enabled by default now. However, some of it's methods are only available when `protocol_feature_routing_exchange_algorithm` is enabled. Otherwise `jsonrpc` wouldn't be able to call them.

chain/network/src/test_utils.rs
pmnoxx: I created it the same way as `
    edge_verifier_pool: Addr<EdgeVerifier>,
`.
We don't stop `EdgeVerifier` either. Maybe we should stop both?

chain/network/src/peer_manager.rs
pmnoxx: Ok

pmnoxx: Done. I also moved it to another file.

chain/network/src/test_utils.rs
bowenwang1996: Yes

chain/network/src/edge_verifier.rs
```diff
@@ -0,0 +1,48 @@
...
+            if !msg.adv_disable_edge_signature_verification && !edge.verify() {
+                return false;
+            }
+
+            #[cfg(not(feature = "adversarial"))]
+            if !edge.verify() {
+                return false;
+            }
+            {
+                let mut guard = msg.edges_info_shared.lock().unwrap();
```
bowenwang1996, line 38: why do we need a mutex here?

pmnoxx: Yes, this is a data structure, that's shared between thread poll that `EdgeVerifier` uses. It should be safe.
I didn't change this code, just moved to another file.

chain/network/src/test_utils.rs
pmnoxx: Done
@bowenwang1996 Can you check that the logic for stopping is correct?

chain/network/src/routing_table_actor.rs
```diff
@@ -0,0 +1,293 @@
...
+use crate::ibf_peer_set::IbfPeerSet;
+#[cfg(feature = "protocol_feature_routing_exchange_algorithm")]
+use crate::ibf_set::IbfSet;
+use crate::routing::Edge;
+#[cfg(feature = "protocol_feature_routing_exchange_algorithm")]
+use crate::routing::SimpleEdge;
+#[cfg(feature = "protocol_feature_routing_exchange_algorithm")]
+use crate::routing::ValidIBFLevel;
+#[cfg(feature = "protocol_feature_routing_exchange_algorithm")]
+use crate::routing::MIN_IBF_LEVEL;
```
bowenwang1996, line 23: Can we group these together?

chain/network/src/routing_table_actor.rs
```diff
@@ -0,0 +1,293 @@
...
+use crate::types::StopMsg;
+#[cfg(feature = "protocol_feature_routing_exchange_algorithm")]
+use crate::types::{PartialSync, RoutingState, RoutingVersion2};
+
+/// Actor that maintains routing table information.
+/// TODO (PIOTR, #4859) Finish moving routing table computation to new thread.
```
bowenwang1996, line 29: Should this be removed?

chain/network/src/routing_table_actor.rs
```diff
@@ -0,0 +1,293 @@
...
+    RemoveEdges(Vec<Edge>),
+    #[cfg(feature = "protocol_feature_routing_exchange_algorithm")]
+    AddPeerIfMissing(PeerId, Option<u64>),
+    #[cfg(feature = "protocol_feature_routing_exchange_algorithm")]
+    RemovePeer(PeerId),
+    #[cfg(feature = "protocol_feature_routing_exchange_algorithm")]
+    ProcessIbfMessage {
+        peer_id: PeerId,
+        ibf_msg: RoutingVersion2,
+    },
```
bowenwang1996, line 73: It is better to move feature gated variants to the end of all variants so that it is easier to see which ones are gated

bowenwang1996 (approved):

chain/network/src/routing_table_actor.rs
pmnoxx: Done

chain/network/src/routing_table_actor.rs
pmnoxx: Done

chain/network/src/routing_table_actor.rs
pmnoxx: @bowenwang1996 This is part 1 of the full change. In the part 1 we offload work of getting routing table to another thread.
In part 2 we will move `RoutingTable -  pub raw_graph: Graph,` to another thread.

So, this shouldn't be marked as done yet.

pmnoxx:
@mfornet
@khorolets
@mzhangmzz
Can you take a look at this PR?

bowenwang1996:
@pmnoxx you already have all the approvals you needed

pmnoxx:
@bowenwang1996 I updated the description of this PR as you requested.

PR #4890, opened 2021-09-28
refactor: remove ActualRuntimeConfig

Longarithm:
Remove deprecated `ActualRuntimeConfig` which has no uses now.
Helper for https://github.com/near/nearcore/issues/4649.

Test plan
---------
Existing tests. Removed tests were migrated to `RuntimeConfigStore`.


matklad:
Oh, that was this easy, nice!

Note: the reason why rustc didn't flag this as unused is because this is a `pub` type which could have been used from another crate. That's one of the reasons why it's important to keep things `pub(crate)` if possible.

matklad (approved):

PR #4891, opened 2021-09-28
refactor: separate default and test runtime configs

Longarithm:
Having default impls for `RuntimeConfig` and `RuntimeFeesConfig` is confusing, because one may think that it is the actual protocol sub-configs, whereas they are used mostly for testing.
Here I separate cases where we want to take:
- actual protocol config
- some test config

Helps to solve https://github.com/near/nearcore/issues/4649.

Test plan
---------
Existing tests. 

~~Note that `test_data_roundtrip_is_more_expensive` was moved to `config_store` because I don't see a reason for testing test config.~~
Note that `test_data_roundtrip_is_more_expensive` is removed because its logic seems to be incorrect (see comments).

### DIFF
--- a/core/primitives-core/src/runtime/fees.rs
+++ b/core/primitives-core/src/runtime/fees.rs
@@ -38,7 +38,7 @@ impl Fee {
     }
 
     /// The minimum fee to send and execute.
-    fn min_send_and_exec_fee(&self) -> Gas {
+    pub fn min_send_and_exec_fee(&self) -> Gas {
         std::cmp::min(self.send_sir, self.send_not_sir) + self.execution
     }
 }
@@ -136,8 +136,8 @@ pub struct StorageUsageConfig {
     pub num_extra_bytes_record: u64,
 }
 
-impl Default for RuntimeFeesConfig {
-    fn default() -> Self {
+impl RuntimeFeesConfig {
+    pub fn test() -> Self {
         #[allow(clippy::unreadable_literal)]
         Self {
             action_receipt_creation_config: Fee {
@@ -231,9 +231,7 @@ impl Default for RuntimeFeesConfig {
             pessimistic_gas_price_inflation_ratio: Rational::new(103, 100),
         }
     }
-}
 
-impl RuntimeFeesConfig {
     pub fn free() -> Self {
         let free = Fee { send_sir: 0, send_not_sir: 0, execution: 0 };
         RuntimeFeesConfig {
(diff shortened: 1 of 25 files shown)


core/primitives/src/runtime/config_store.rs
```diff
@@ -134,6 +132,26 @@ mod tests {
...
 
+    #[test]
+    fn test_data_roundtrip_is_more_expensive() {
+        // We have an assumption that the deepest receipts we can create is by creating recursive
+        // function call promises (calling function call from a function call).
+        // If the cost of a data receipt is cheaper than the cost of a function call, then it's
+        // possible to create a promise with a dependency which will be executed in two blocks that
+        // is cheaper than just two recursive function calls.
+        // That's why we need to enforce that the cost of the data receipt is not less than a
+        // function call. Otherwise we'd have to modify the way we compute the maximum depth.
```
bowenwang1996, line 143: Interesting. With the new change it looks like this test will fail. @Longarithm

bowenwang1996 (approved):

core/primitives/src/runtime/config_store.rs
Longarithm: I took a closer look and I think this test should be removed:
> If the cost of a data receipt is cheaper than the cost of a function call, then it's
> possible to create a promise with a dependency which will be executed in two blocks that
> is cheaper than just two recursive function calls.

If that was **really** the case, recursive function calls would not make sense because we could always create promises with dependencies. And we actually charge function call fees on promise creation here:
 
https://github.com/near/nearcore/blob/a20c12ec02d7f22550139da38838efd1fba75ede/runtime/near-vm-logic/src/logic.rs#L1530

ailisp (approved):

core/primitives/src/runtime/config_store.rs
```diff
@@ -134,6 +132,26 @@ mod tests {
...
+        let store = RuntimeConfigStore::new(None);
+        for config in store.store.values() {
```
matklad, line 145: Good catch! That's a good example of specific issue you can get from a rather amorphous problem of "having default doesn't feel right".

runtime/runtime-params-estimator/src/testbed.rs
```diff
@@ -38,21 +39,22 @@ impl RuntimeTestbed {
...
-            max_register_size: u64::max_value(),
-            max_number_logs: u64::max_value(),
-
-            max_actions_per_receipt: u64::max_value(),
-            max_promises_per_function_call_action: u64::max_value(),
-            max_number_input_data_dependencies: u64::max_value(),
-
-            max_total_prepaid_gas: u64::max_value(),
-            max_number_bytes_method_names: u64::max_value(),
+            max_total_log_length: u64::MAX,
```
matklad, line 46: :+1: `max_value` is indeed soft-deprecated.

core/primitives-core/src/runtime/fees.rs
```diff
@@ -136,8 +135,8 @@ pub struct StorageUsageConfig {
     pub num_extra_bytes_record: u64,
 }
 
-impl Default for RuntimeFeesConfig {
-    fn default() -> Self {
+impl RuntimeFeesConfig {
+    pub fn test() -> Self {
```
matklad, line 139: Rather than hard-coding a bunch of constants here, can we re-define this as

```
fn test() {
   RuntimeStoreConfig::new(None).for_protocol_version(PROTOCL_VERSION)
}
```

?

matklad (approved):
LGTM as is, but would be cool if we can replace hard-coded consts with just using RuntimeConfigStore

EgorKulikov (approved):
Good job

core/primitives-core/src/runtime/fees.rs
Longarithm: That's a good point, thank you!

I need to check if this doesn't break tests, and this PR is going to have some meaningful logic already, so I prefer to try it in the following PR if you don't mind.

bowenwang1996 (approved):

frol (approved):

PR #4894, opened 2021-09-29
Separate test methods from production code in PeerManager.

pmnoxx:
Puts test methods under `adversarial` flag.

### DIFF
--- a/chain/client/src/test_utils.rs
+++ b/chain/client/src/test_utils.rs
@@ -965,7 +965,7 @@ pub fn setup_mock_all_validators(
                         | NetworkRequests::RequestUpdateNonce(_, _)
                         | NetworkRequests::ResponseUpdateNonce(_)
                         | NetworkRequests::ReceiptOutComeRequest(_, _)=> {}
-                        #[cfg(feature = "adversarial")]
+                        #[cfg(feature = "test_features")]
                         | NetworkRequests::PingTo(_, _)
                         | NetworkRequests::FetchPingPongInfo
                         | NetworkRequests::FetchRoutingTable  => {}
--- a/chain/client/src/test_utils.rs
+++ b/chain/client/src/test_utils.rs
@@ -958,9 +958,6 @@ pub fn setup_mock_all_validators(
                         }
                         NetworkRequests::ForwardTx(_, _)
                         | NetworkRequests::Sync { .. }
-                        | NetworkRequests::FetchRoutingTable
-                        | NetworkRequests::PingTo(_, _)
-                        | NetworkRequests::FetchPingPongInfo
                         | NetworkRequests::BanPeer { .. }
                         | NetworkRequests::TxStatus(_, _, _)
                         | NetworkRequests::Query { .. }
@@ -968,6 +965,10 @@ pub fn setup_mock_all_validators(
                         | NetworkRequests::RequestUpdateNonce(_, _)
                         | NetworkRequests::ResponseUpdateNonce(_)
                         | NetworkRequests::ReceiptOutComeRequest(_, _) => {}
+                        #[cfg(feature = "adversarial")]
+                        | NetworkRequests::PingTo(_, _)
+                        | NetworkRequests::FetchPingPongInfo
+                        | NetworkRequests::FetchRoutingTable  => {}
                         #[cfg(feature = "protocol_feature_routing_exchange_algorithm")]
                         | NetworkRequests::IbfMessage { .. } => {}
                     };
--- a/chain/client/src/test_utils.rs
+++ b/chain/client/src/test_utils.rs
@@ -964,7 +964,8 @@ pub fn setup_mock_all_validators(
                         | NetworkRequests::Challenge(_)
                         | NetworkRequests::RequestUpdateNonce(_, _)
                         | NetworkRequests::ResponseUpdateNonce(_)
-                        | NetworkRequests::ReceiptOutComeRequest(_, _)
+                        | NetworkRequests::ReceiptOutComeRequest(_, _)=> {}
+                        #[cfg(feature = "adversarial")]
                         | NetworkRequests::PingTo(_, _)
                         | NetworkRequests::FetchPingPongInfo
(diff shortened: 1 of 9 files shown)


chain/network/src/routing.rs
```diff
@@ -442,9 +448,13 @@ impl RoutingTable {
             store,
             raw_graph: Graph::new(peer_id),
             route_nonce: SizedCache::with_size(ROUND_ROBIN_NONCE_CACHE_SIZE),
+            #[cfg(feature = "adversarial")]
             ping_info: SizedCache::with_size(PING_PONG_CACHE_SIZE),
+            #[cfg(feature = "adversarial")]
             pong_info: SizedCache::with_size(PING_PONG_CACHE_SIZE),
+            #[cfg(feature = "adversarial")]
             waiting_pong: SizedCache::with_size(PING_PONG_CACHE_SIZE),
+            #[cfg(feature = "adversarial")]
```
bowenwang1996, line 457: It might be better if we create a single struct to hold all of these fields for testing

chain/network/src/peer_manager.rs
```diff
@@ -2416,12 +2420,18 @@ impl Handler<RoutedMessageFrom> for PeerManagerActor {
...
+                #[cfg(feature = "adversarial")]
+                RoutedMessageBody::Ping(ping) => {
+                    self.handle_ping(ctx, ping.clone(), msg.hash());
+                    false
+                }
+                #[cfg(feature = "adversarial")]
+                RoutedMessageBody::Pong(pong) => {
+                    self.handle_pong(ctx, pong.clone());
+                    false
+                }
```
bowenwang1996, line 2432: Does this actually make sense? Since we did not change `RoutedMessageBody` people can still send `Ping` and `Pong` message and shouldn't we return `false` for those?

bowenwang1996 (commented):
Should we change `adversarial` to some other name?

chain/network/src/peer_manager.rs
pmnoxx: @bowenwang1996 That is a good question.  There is no code path to send `Ping` in production. Though, RoutedMessageBody::Ping / RoutedMessageBody::Pong are already in production, we have code which processes them, so we can't simply remove those messages. Only a peer that runs a modified code could use `ping`/`pong` messages. I don't think we care about those.

pmnoxx: I changed PR, unfortunately we have to keep following test messages, because they are part of protocol now, and we can't  remove test fields anymore, because they aren't at the end of enum.
```
                        | NetworkRequests::PingTo(_, _)
                        | NetworkRequests::FetchPingPongInfo
                        | NetworkRequests::FetchRoutingTable 
```

chain/network/src/routing.rs
pmnoxx: I agree

chain/client/src/test_utils.rs
```diff
@@ -953,16 +953,16 @@ pub fn setup_mock_all_validators(
...
                         | NetworkRequests::TxStatus(_, _, _)
                         | NetworkRequests::Query { .. }
                         | NetworkRequests::Challenge(_)
                         | NetworkRequests::RequestUpdateNonce(_, _)
                         | NetworkRequests::ResponseUpdateNonce(_)
-                        | NetworkRequests::ReceiptOutComeRequest(_, _) => {}
+                        | NetworkRequests::ReceiptOutComeRequest(_, _)
+                        | NetworkRequests::PingTo(_, _)
+                        | NetworkRequests::FetchPingPongInfo
+                        | NetworkRequests::FetchRoutingTable  => {}
```
bowenwang1996, line 965: `NetworkRequests` is not part of the protocol. We can still gate them

chain/network/src/peer_manager.rs
bowenwang1996: hmm the code here still remains unchanged it seems?

pmnoxx: I wasn't sure whenever making `NetworkRequests` enums is fine. I checked the code, you are right, we don't use it as part of protocol. I put this put under `adversarial`.

chain/client/src/test_utils.rs
pmnoxx: Ok, done.

chain/network/src/routing.rs
pmnoxx: Done

chain/network/src/routing.rs
```diff
@@ -382,6 +384,29 @@ impl Default for EdgeVerifierHelper {
     }
 }
 
+#[cfg(feature = "adversarial")]
+pub struct TestHelper {
```
bowenwang1996, line 388: This name feels too general. `NetworkTestHelper` might be better

bowenwang1996 (approved):

chain/network/src/routing.rs
pmnoxx: Thanks, that seems like a good idea to rename it. Done

pmnoxx:
@mfornet
@mzhangmzz
@chefsale
@frol
@khorolets
Please take a look at this PR.

pmnoxx:
I'll close this PR for now. We will decide what do to later.

pmnoxx:
@mzhangmzz
@chefsale
@frol
@khorolets
@mm-near
Can you take a look?

chain/network/src/types.rs
```diff
@@ -639,6 +639,8 @@ pub enum NetworkRequests {
 
     /// The following types of requests are used to trigger actions in the Peer Manager for testing.
     /// Fetch current routing table.
+    /// Used for testing purposes, unfortunately, this is now part of protocol.
+    #[cfg(feature = "test_features")]
```
mm-near, line 643: is it safe to 'remove' such entries from the enums? wouldn't it break the Borsch serialization of things below (like Sync or RequestUpdateNonce)

mm-near (changes requested):

PR #4897, opened 2021-09-29
producing/processing last block before shard layout changes

mzhangmzz:
This PR makes the transition at the last block before sharding changes correct.
1. chunk_extra for splitting states are stored correctly. These chunk_extra will have all information that is needed for chunk header validation when the new epoch with the new shard layout begins. Note that technically, only the last block of the epoch before sharding changes need to store the full extra information, but here we implement storing that for all blocks in the transitioning epoch.
2. `produce_blocks`, `process_blocks` now can handle the case where number of chunks in the last block is different from number of chunks in this block

Test Plan:
1. `test_shard_layout_upgrade` tests that blocks can still get produced after shard layout changes. However, currently it only tests for a simple state. It doesn't test the chunk producing/receipts handling logic because the produced chunks are not distributed. I will modify that in a later PR. 
2. I will add more integration tests

### DIFF
--- a/chain/chain/src/chain.rs
+++ b/chain/chain/src/chain.rs
@@ -67,7 +67,7 @@ use crate::{metrics, DoomslugThresholdMode};
 use actix::Message;
 #[cfg(feature = "delay_detector")]
 use delay_detector::DelayDetector;
-use near_primitives::shard_layout::{account_id_to_shard_uid, ShardLayoutError, ShardUId};
+use near_primitives::shard_layout::{account_id_to_shard_uid, ShardUId};
 use rayon::iter::{IntoParallelIterator, ParallelIterator};
 
 /// Maximum number of orphans chain can store.
@@ -2639,11 +2639,8 @@ impl Chain {
     ) -> Result<Vec<ShardChunkHeader>, Error> {
         let epoch_id = runtime_adapter.get_epoch_id_from_prev_block(&prev_block.hash())?;
         let num_shards = runtime_adapter.num_shards(&epoch_id)?;
-        let prev_shard_ids = Self::get_prev_chunk_headers_impl(
-            runtime_adapter,
-            prev_block,
-            (0..num_shards).collect(),
-        )?;
+        let prev_shard_ids =
+            runtime_adapter.get_prev_shard_ids(prev_block.hash(), (0..num_shards).collect())?;
         let chunks = prev_block.chunks();
         Ok(prev_shard_ids
             .into_iter()
@@ -2657,37 +2654,8 @@ impl Chain {
         shard_id: ShardId,
     ) -> Result<ShardChunkHeader, Error> {
         let prev_shard_id =
-            Self::get_prev_chunk_headers_impl(runtime_adapter, prev_block, vec![shard_id])?[0];
+            runtime_adapter.get_prev_shard_ids(prev_block.hash(), vec![shard_id])?[0];
         Ok(prev_block.chunks().get(prev_shard_id as usize).unwrap().clone())
-    }
-
-    fn get_prev_chunk_headers_impl(
-        runtime_adapter: &dyn RuntimeAdapter,
-        prev_block: &Block,
-        shard_ids: Vec<ShardId>,
-    ) -> Result<Vec<ShardId>, Error> {
-        if runtime_adapter.is_next_block_epoch_start(prev_block.hash())? {
-            let shard_layout =
-                runtime_adapter.get_shard_layout_from_prev_block(prev_block.hash())?;
-            let prev_shard_layout =
-                runtime_adapter.get_shard_layout(prev_block.header().epoch_id())?;
-            if prev_shard_layout != shard_layout {
-                return Ok(shard_ids
-                    .into_iter()
-                    .map(|shard_id| {
-                        shard_layout.get_parent_shard_id(shard_id).map(|parent_shard_id|{
-                            assert!(parent_shard_id < prev_shard_layout.num_shards(),
-                                    "invalid shard layout {:?}: parent shard {} does not exist in last shard layout",
-                                    shard_layout,
-                                    parent_shard_id
-                            );
-                            parent_shard_id
-                        })
-                    })
-                    .collect::<Result<_, ShardLayoutError>>()?);
-            }
-        }
-        Ok(shard_ids)
     }
 }
 
--- a/chain/chain/src/chain.rs
+++ b/chain/chain/src/chain.rs
@@ -3014,14 +3014,14 @@ impl<'a> ChainUpdate<'a> {
             };
             let need_to_split_states = will_shard_layout_change && cares_about_shard_next_epoch;
             // We can only split states when states are ready, i.e., mode != ApplyChunksMode::NotCaughtUp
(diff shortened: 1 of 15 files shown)


nearcore/src/runtime/mod.rs
```diff
@@ -531,36 +529,6 @@ impl NightshadeRuntime {
             })?;
 
         let shard_uid = self.get_shard_uid_from_prev_hash(shard_id, prev_block_hash)?;
-        let apply_split_state_result_or_state_changes =
```
mzhangmzz, line 534: I decided to refactor the code and move this logic to chain

chain/chain/src/chain.rs
```diff
@@ -703,6 +703,7 @@ impl Chain {
             if chunk_header.height_created() == genesis_block.header().height() {
                 // Special case: genesis chunks can be in non-genesis blocks and don't have a signature
                 // We must verify that content matches and signature is empty.
+                // TODO: this code will not work when genesis block has different number of chunks as the current block
```
bowenwang1996, line 706: could we create an issue for this?

chain/chain/src/chain.rs
```diff
@@ -3222,34 +3295,150 @@ impl<'a> ChainUpdate<'a> {
...
-                        ChunkExtra::new_with_only_state_root(&result.new_root),
-                    );
-                    self.chain_store_update.save_trie_changes(result.trie_changes);
+                if is_new_chunk {
+                    // Split validator_proposals, gas_burnt, balance_burnt to each split shard
+                    // and store the chunk extra for split shards
+                    // Note that here we do not split outcomes by the new shard layout, we simply store
+                    // the outcome_root from the parent shard. This is because outcome proofs are
+                    // generated per shard using the old shard layout and stored in the database.
+                    // For these proofs to work, we must store the outcome root using per shard
```
bowenwang1996, line 3354:
```suggestion
                    // For these proofs to work, we must store the outcome root per shard
```

chain/chain/src/chain.rs
```diff
@@ -3222,34 +3295,150 @@ impl<'a> ChainUpdate<'a> {
...
+                        .unwrap_or_else(|| {
+                            panic!("invalid shard layout {:?}", next_epoch_shard_layout)
+                        })
+                        .len() as NumShards;
+                    let total_gas_used = chunk_extra.gas_used();
+                    let total_balance_burnt = chunk_extra.balance_burnt();
+                    let gas_res = total_gas_used % num_split_shards;
+                    let gas_split = total_gas_used / num_split_shards;
+                    let balance_res = (total_balance_burnt % num_split_shards as u128) as NumShards;
+                    let balance_split = total_balance_burnt / (num_split_shards as u128);
```
bowenwang1996, line 3388: is this actually the right way to split? What is the rationale behind it?

bowenwang1996 (commented):
I still do not understand how outgoing receipts verification works after this change :( could you enlighten me on that 😃

chain/chain/src/chain.rs
mzhangmzz: I think gas_used and balance burnt are only used at the block level to calculate gas price, so how they are divided here actually does not matter as long as the sum remains the same

chain/chain/src/chain.rs
mzhangmzz: Yes, I probably won't fix this before the testnet launch though. This won't affect the simple nightshade launch because chunks get produced between the genesis block and when sharding changes

Sorry that I always forget to create issues for TODOs. Sometimes I just added TODOs as I write code and I wasn't sure if they are actually real TODOs and forget about them later .

mzhangmzz:
> I still do not understand how outgoing receipts verification works after this change :( could you enlighten me on that 😃

Yes. I didn't actually make any change on outgoing receipts verification in this PR, so I guess it did just work before the last PR. 

Say sharding is about to change at epoch T. When producing chunks for the first block of epoch T, `produce_chunk` uses `get_outgoing_receipts_for_shard` to compute outgoing receipts, which handles splitting shards and will split the outgoing receipts according the new shard layout. The new chunk header will contain the root of these outgoing receipts and the chunk producer send these outgoing receipts out. For outgoing receipts verification, 
- in `process_partial_encoded_chunk`, the validator verifies that the outgoing receipts matches of chunk_header.outgoing_receipts_root(). That will succeed because the root are calculated from the same set of outgoing receipts
- when processing a chunk, the validator verifies the outgoing receipts root of the chunk header is valid. This happens in `validate_chunk_with_chunk_extra`, which also uses `get_outgoing_receipts_for_shard` to get outgoing receipts, so they will be from the shard in the new shard layout.

chain/chain/src/chain.rs
mzhangmzz: https://github.com/near/nearcore/issues/4908

chain/chain/src/chain.rs
bowenwang1996: hmm why wouldn't `validate_chunk_with_chunk_extra` break?

bowenwang1996 (approved):

bowenwang1996:
@nikurt @Longarithm please review

Longarithm (approved):

nikurt (approved):

chain/chain/src/chain.rs
matklad: Yeah, I'd probably expect to see something like 

```
fn split(total: u64, n_parts: usize) -> Vec<u64> {}
```

where the split is as even as possible (ie, the first rem items get +1)

chain/chain/src/chain.rs
```diff
@@ -3326,34 +3397,150 @@ impl<'a> ChainUpdate<'a> {
...
+
+                        self.chain_store_update.save_chunk_extra(
+                            &block_hash,
+                            &result.shard_uid,
+                            new_chunk_extra,
+                        );
+                        self.chain_store_update.save_trie_changes(result.trie_changes);
+                    }
+                    assert_eq!(sum_gas_used, total_gas_used);
+                    assert_eq!(sum_balance_burnt, total_balance_burnt);
```
matklad, line 3521: :+1:

integration-tests/tests/client/process_blocks.rs
```diff
@@ -2614,36 +2626,55 @@ fn test_shard_layout_upgrade() {
...
             .runtime_adapter
             .get_epoch_id_from_prev_block(&head.last_block_hash)
             .unwrap();
-        let block_producer =
-            env.clients[0].runtime_adapter.get_block_producer(&epoch_id, i).unwrap();
-        let index = if block_producer.as_ref() == "test0" { 0 } else { 1 };
-        let (encoded_chunk, merkle_paths, receipts) =
-            create_chunk_on_height(&mut env.clients[index], i);
+        let shard_layout = env.clients[0].runtime_adapter.get_shard_layout(&epoch_id).unwrap();
+        println!(
```
matklad, line 2642: tbh, not a huge fan of just leaving `println`s in tests. They make the test harder to debug if you add debug prints elsewhere.

mzhangmzz: Ah I'm still working on this test (in another PR), after I am done, I will remove this line!

PR #4906, opened 2021-09-30
debug_spans for block processings

nikurt:


chain/chain/src/chain.rs
```diff
@@ -3451,6 +3451,7 @@ impl<'a> ChainUpdate<'a> {
     where
         F: FnMut(ChallengeBody) -> (),
     {
+        let _span = tracing::debug_span!(target: "runtime", "process_block").entered();
```
matklad, line 3454:
```suggestion
        let _span = tracing::debug_span!(target: "chain", "process_block").entered();
```

?

matklad: It feels that this span now duplicates the subsequent log line, so I suggest merging the two. And we probably should be using structured logging capability of tracing to log block.hash, height and such: 

https://docs.rs/tracing/0.1.28/tracing/#recording-fields

(nodex team probably wants some general guidelines with respcet to how we use structured logs)

nikurt: Having general guidelines would certainly be great. And it makes sense for nodeX to own them.
However, the guidelines are rather far out of scope of this PR :)
Trying a few options, I think that including block height in the span is the best tradeoff. It keeps the logs short, but makes them easily parsable.

bowenwang1996 (approved):

chain/chain/src/chain.rs
```diff
@@ -3451,7 +3451,10 @@ impl<'a> ChainUpdate<'a> {
     where
         F: FnMut(ChallengeBody) -> (),
     {
-        debug!(target: "chain", "Process block {} at {}, approvals: {}, me: {:?}", block.hash(), block.header().height(), block.header().num_approvals(), me);
+        let _span =
+            tracing::debug_span!(target: "chain", "Process block", "#{}", block.header().height())
```
frol, line 3455: @matklad Was there any decision about the use of spans in regards to the fact that we don't see them on segfaults?

frol (approved):
What is up with the number of commits here?

chain/chain/src/chain.rs
matklad: Not really, easier to just implement the thing than decide: https://github.com/near/nearcore/pull/4916 )

PR #4909, opened 2021-09-30
chore: mark all but a few crates private (unpublishable)

miraclx:
This marks all the crates in the nearcore workspace `publish = false` except for the small number of crates required to be public for the purpose of publishing `near-jsonrpc-client`.

All but the following crates are marked non-publishable:

- `near-account-id`
- `near-crypto`
- `near-rpc-error-core`
- `near-rpc-error-macro`
- `near-vm-errors`
- `near-primitives-core`
- `near-primitives`
- `near-chain-configs`
- `near-chain-primitives`
- `near-chunks-primitives`
- `near-network-primitives`
- `near-metrics`

Future pipelines for automating publishing would respect that flag and skip flagged crates meaning we only ever publish a subset of all the crates, instead of all by default.

The plan for CI publishing is to have it run on every merge on master. Non-private crates whose versions get bumped in a PR, once merged get updated or just quits otherwise.

### DIFF
--- a/chain/chain-primitives/Cargo.toml
+++ b/chain/chain-primitives/Cargo.toml
@@ -2,7 +2,6 @@
 name = "near-chain-primitives"
 version = "0.1.0"
 authors = ["Near Inc <hello@nearprotocol.com>"]
-publish = false
 edition = "2018"
 
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
--- a/chain/chain-primitives/Cargo.toml
+++ b/chain/chain-primitives/Cargo.toml
@@ -2,6 +2,7 @@
 name = "near-chain-primitives"
 version = "0.1.0"
 authors = ["Near Inc <hello@nearprotocol.com>"]
+publish = false
 edition = "2018"
 
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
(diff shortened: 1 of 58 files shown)


bowenwang1996 (approved):

pmnoxx (approved):

frol (approved):

matklad (approved):

nikurt (approved):

matklad:
I am surprised to see `near-vm-errors` on the list. The way we structure specific error inside runtime shouldn't have any bearing on jsonrpc-client. 

In particular, I think it should be fine for contract runtime to use precise wm specific errors (like `wasmer::Error` or `wasmtime::Error`), and to be able to switch between them without breaking changes to `jsonrpc-clinet` or making it to depend on wasm runtimes. 

I think the culprit here is this enum variant:

https://github.com/near/nearcore/blob/f0765fcbbda45e9cfe709ac99680626a84737992/core/primitives/src/errors.rs#L425

We probably shouldn't be directly using a vm-errors type there, but just spell out the error structure in the primitives crate, like we do for the rest of the errors

posvyatokum (approved):

chefsale (approved):

frol:
@matklad Good point, we definitely will need to revisit the whole story around `ActionErrorKind` since I don't want it to be part of the public interfaces as is (the new structured errors for JSON RPC did not expose the details since we did not want to commit to that old way of doing things, but we were out of capacity addressing this issue).

I feel that this is not a blocker for now and we can proceed.

matklad:
>the new structured errors for JSON RPC did not expose the details since

Oh, that's excellent to hear, :+1:

PR #4112, opened 2021-03-15
new routing exchange prototype using IBF 

pmnoxx:
Implement new algorithm for exchanging routing tables using Inverse Bloom Filters.
#3838

Reposting #4108

PR #4912, opened 2021-09-30
refactor(jsonrpc-primitives): move adversarial to own specialized crate

miraclx:
#4651 successfully removed `near-jsonrpc-primitives`-s heavy dependency on `near-network` which in turn depended on `rocksdb`. #4112 reintroduced that dependency on `near-network`, though optionally, this hinders the publishing process.

This PR extracts the adversarial features that depend on `near-network` into a private, specialized crate: `near-jsonrpc-adversarial-primitives` and sets us right on track for publishing `jsonrpc-primitives`.


frol (approved):

pmnoxx (approved):

miraclx:
@pmnoxx, please could you take a look at CI.. I tried mapping out the relationship but everything seems to checkout in my head. For some reason, when compiling `near-client` without adversarial features, it's dependency `near-network` is built with the feature regardless.

chain/jsonrpc-adversarial-primitives/Cargo.toml
```diff
@@ -0,0 +1,13 @@
...
+version = "0.2.0"
+authors = ["Near Inc <hello@nearprotocol.com>"]
+publish = false
+edition = "2018"
+
+[dependencies]
+serde = { version = "1", features = ["derive"] }
+serde_json = "1"
+near-primitives = { path = "../../core/primitives" }
+near-jsonrpc-primitives = { path = "../jsonrpc-primitives", features = ["adversarial"] }
```
pmnoxx, line 12: Please remove, `features = ["adversarial"]`.
This causes `near-jsonrpc-primitives ` and `near-network` to be build with `adversarial` even if it's not specified.

At the end of the file please add
```
[features]
adversarial = ["near-jsonrpc-primitives/adversarial", "near-network/adversarial"]
```
This will propagate `adversarial` to those 2 crates.

pmnoxx: > @pmnoxx, please could you take a look at CI.. I tried mapping out the relationship but everything seems to checkout in my head. For some reason, when compiling `near-client` without adversarial features, it's dependency `near-network` is built with the feature regardless.

See me comment about usage of `adversarial`. That should help.

miraclx: Yeah, I already experimented with that locally, didn't seem to work either.

pmnoxx: @miraclx It works just fine: https://github.com/near/nearcore/pull/4915/files

miraclx: Thanks!

miraclx: Still have a small nit here:

The issue, though tough to explain, is likely a bug in how cargo resolves dep features.

For a crate A with an optional dependency B depending on C/feat where you only directly depend on C.
Interaction between A and C ends up in conflict, despite B not being selected.

```text
A
├── B { optional = true }
│   └── C { features = ["feat"] }
└── C
```

Cargo somehow infers that `A > C { features = ["feat"] }` despite that only being the case if `A { features = ["B"] }`

Now, this patch explicitly tells cargo to include `C/feat` if and only if `B` selected

It's unintuitive and perhaps even a mouthful

---

I started going this route on my own but stopped halfway because it didn't make much sense to me

The crate name is `near-jsonrpc-adversarial-primitives` .. It's expected to have `adversarial` features by default. But now, we have to do

```toml
near-jsonrpc-adversarial-features = { path = .., features = ["adversarial"] }
```

which isn't neat.. at all

chain/jsonrpc/src/lib.rs
```diff
@@ -18,17 +18,15 @@ use near_client::{
...
-use near_jsonrpc_primitives::types::adversarial::SetAdvOptionsRequest;
 #[cfg(feature = "adversarial")]
+use near_jsonrpc_adversarial_primitives::SetAdvOptionsRequest;
 #[cfg(feature = "protocol_feature_routing_exchange_algorithm")]
-use near_jsonrpc_primitives::types::adversarial::SetRoutingTableRequest;
-#[cfg(feature = "adversarial")]
+use near_jsonrpc_adversarial_primitives::SetRoutingTableRequest;
 #[cfg(feature = "protocol_feature_routing_exchange_algorithm")]
-use near_jsonrpc_primitives::types::adversarial::StartRoutingTableSyncRequest;
+use near_jsonrpc_adversarial_primitives::StartRoutingTableSyncRequest;
```
bowenwang1996, line 26: should this be under `adversarial` as well @pmnoxx ?

miraclx: It's covered here https://github.com/near/nearcore/pull/4912/files#diff-2b29afffc8afa7eed1365ace3137ccec9802b75eb3b5dd35dc4fd4d388a4c099R45

bowenwang1996: This dependency is strange. I suggest that we decouple them

pmnoxx: Those functions are used for testing `protocol_feature_routing_exchange_algorithm`.
Therefore you need `adversarial` for testing, and `protocol_feature_routing_exchange_algorithm` for the feature itself.

chain/jsonrpc-adversarial-primitives/Cargo.toml
```diff
@@ -0,0 +1,23 @@
...
+near-primitives = { path = "../../core/primitives" }
+
+near-jsonrpc-primitives = { path = "../jsonrpc-primitives", optional = true }
+near-network = { path = "../network" }
+
+
+[features]
+ser_de = [
+  "serde_json",
+  "serde",
```
frol, line 20: I feel this might work:

```suggestion
serde = [
  "serde_json",
```

frol (approved):

chain/jsonrpc-adversarial-primitives/Cargo.toml
miraclx: Unfortunately, cargo doesn't support that yet. You have to explicitly include the dependency in the feature list.

The closest alternative would be to rename the crate to free up the feature name.

Example: https://github.com/miraclx/nearcore/commit/e499e195f7dfaf57ca5d1871c9b5ee22f8e6c7a0

bowenwang1996 (approved):

PR #4914, opened 2021-10-01
Rename adversarial to test features

pmnoxx:
We've been using `adversarial` feature for testing. We should rename it to `test_features`, that will allow us to use this feature for all kinds of tests.


pmnoxx:
What else do we need to change?

chefsale (approved):
Please make sure that you take into account that this PR is approved and landing: 
https://github.com/near/nearcore/pull/4920 cc: @mina86

bowenwang1996 (commented):
Please check with @mina86 that this will not break CI or nayduck

mina86:
So the idea is that we would run all the tests with `cargo test --features test_features`? To be honest I don’t really like that but at the same time I cannot think of a better solution.

pmnoxx:
> Please make sure that you take into account that this PR is approved and landing: #4920 cc: @mina86

Ok, I'll rebase and take this into account.

pmnoxx:
> So the idea is that we would run all the tests with `cargo test --features test_features`? To be honest I don’t really like that but at the same time I cannot think of a better solution.

Yes, I was hoping to use  "#[cfg(test)]"  for that purpose. However, `test` only works within the same crate. This seems to be `Rust` limitation. Ideally `Rust` would support having test function shared between multiple crates, that looks to me like a good feature to request.

pmnoxx:
> Please make sure that you take into account that this PR is approved and landing: #4920 cc: @mina86

Ok, I updated `pipeline.yml`

pmnoxx:
@bowenwang1996 @frol @khorolets @matklad @mfornet@mzhangmzz @mhalambek Can you take a look?

mina86:
So I’ve been thinking a lot about this and I really dislike it. Integration tests should only use public interfaces of crates they test. After all, why are integration tests testing behaviour which is not observable through normal calls? Test methods should only be used by unit tests which test individual modules. `adversarial` feature, I always imagined, was to simulate clients which misbehave and maliciously break the protocol.

Having said all that, I still can’t think of a better quick solution so perhaps this is the way forward.

bowenwang1996:
> So I’ve been thinking a lot about this and I really dislike it. Integration tests should only use public interfaces of crates they test. After all, why are integration tests testing behaviour which is not observable through normal calls? Test methods should only be used by unit tests which test individual modules. adversarial feature, I always imagined, was to simulate clients which misbehave and maliciously break the protocol.

That is true originally. However, later on we added some rpcs to fetch data that help us understand some of the internal state of the node (network mostly). While we could have exposed those rpcs as regular rpc endpoints, it makes no sense since they are only used by people who work on nearcore to test or debug issues. We could keep the name `adversarial` though. I don't have a strong opinion here @mina86

mina86:
If the state is something we might be using when debugging it sounds like it should be extracted to Prometheus? Or through logging?

We want to go this route, changing the name is better than staying with `adversarial`. It will cause some short-term pain (as we’ll have to update all places which might still refer to the old name), but `test_features` is more descriptive.

bowenwang1996:
@mina86 could you advise @pmnoxx on where else we need to make changes?

bowenwang1996:
@mina86 could you advise @pmnoxx on where else we need to make changes?

mina86:
I think this is it. NayDuck will need to be updated as well but I’ll take care of it once this gets merged.

matklad:
>If the state is something we might be using when debugging it sounds like it should be extracted to Prometheus? Or through logging?

Strong plus one here: it's better much to add extra observability to make test-only code not necessary. Though, this is not a mechanical solution and very much requires understanding the details of each specific case.

matklad (approved):

pmnoxx:
> > If the state is something we might be using when debugging it sounds like it should be extracted to Prometheus? Or through logging?
> 
> Strong plus one here: it's better much to add extra observability to make test-only code not necessary. Though, this is not a mechanical solution and very much requires understanding the details of each specific case.

We are using `adversarial` flag to enable a few rps, it's not only used for reading data, but also modifying behavior of code.
For example:
- force edge pruning, which runs every 3 hours be default
- disable edge verification - to make tests run faster
- disable edge propagation - to introduce states, where peers have different view of routing table, etc. 
- ping/pong messages, they are used to test routing, though they are not needed for main network
etc.

For purpose of debugging / tests, we need to read data from node. For example, read routing table, that may be too much to log to Prometheus.

bowenwang1996 (approved):

PR #4918, opened 2021-10-01
runtime-tester: update README.md after changes to Scenario

mina86:
Issue: https://github.com/near/nearcore/issues/4771

### DIFF
--- a/test-utils/runtime-tester/README.md
+++ b/test-utils/runtime-tester/README.md
@@ -8,12 +8,16 @@ Scenario can be loaded from a json file or constructed in rust code.
 pub fn from_file(path: &Path) -> io::Result<Scenario>
 ```
 
-Scenario::run tries to produce all the described blocks and if succeeded returns RuntimeStats.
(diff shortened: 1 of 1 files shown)


matklad (approved):

matklad:
I wonder it this readme could be moved to doc comments on the smoke-test function? 

https://github.com/near/nearcore/blob/72798f20c6d0d60b36c3d3c600c35daa37c27fe1/test-utils/runtime-tester/src/lib.rs#L15

Keeping docs in sync is non-trivial work.

mina86:
It probably could work as a documentation of the crate (i.e. top level documentation in `lib.rs` you’ve pointed at) but really there are arguments in favour of either direction. Having it in README.md makes it show up in GitHub; having it in `lib.rs` makes it show up when someone browses source files.

matklad:
:thinking: 

I wonder if that's an argument for doing `#![doc = include_str!("../README.md")]` trick?

That way it's still visible at github, but at the same time the code inside is doctested

PR #4919, opened 2021-10-01
Integration test for sharding upgrade

mzhangmzz:
This PR adds an integration test for sharding upgrade. It also fixes function `check_if_block_is_first_with_chunk_of_version` to make it work when number of shards changes.

The integration test tests logic on the client level (no logic in ClientActor is included) and checks that transactions/receipts are processed correctly during sharding upgrade. It also checks that the final after sharding upgrade is correct. 

This PR is a beginning. I will add more tests in the same file using the framework.

### DIFF
--- a/chain/chain/src/store.rs
+++ b/chain/chain/src/store.rs
@@ -286,16 +286,19 @@ pub trait ChainStoreAccess {
     /// Get epoch id of the last block with existing chunk for the given shard id.
     fn get_epoch_id_of_last_block_with_chunk(
         &mut self,
+        runtime_adapter: &dyn RuntimeAdapter,
         hash: &CryptoHash,
         shard_id: ShardId,
     ) -> Result<EpochId, Error> {
         let mut candidate_hash = *hash;
+        let mut shard_id = shard_id;
         loop {
             let block_header = self.get_block_header(&candidate_hash)?;
             if block_header.chunk_mask()[shard_id as usize] {
                 break Ok(block_header.epoch_id().clone());
             }
             candidate_hash = *block_header.prev_hash();
+            shard_id = runtime_adapter.get_prev_shard_ids(&candidate_hash, vec![shard_id])?[0];
         }
     }
 }
--- a/chain/client/src/test_utils.rs
+++ b/chain/client/src/test_utils.rs
@@ -54,6 +54,7 @@ use crate::{start_view_client, Client, ClientActor, SyncStatus, ViewClientActor}
 use near_chain::chain::{do_apply_chunks, BlockCatchUpRequest};
 use near_chain::types::AcceptedBlock;
 use near_client_primitives::types::Error;
+use near_primitives::utils::MaybeValidated;
 
 pub type NetworkMock = Mocker<PeerManagerActor>;
 
@@ -1105,6 +1106,7 @@ pub struct TestEnv {
     pub validators: Vec<AccountId>,
     pub network_adapters: Vec<Arc<MockNetworkAdapter>>,
     pub clients: Vec<Client>,
+    account_to_client_index: HashMap<AccountId, usize>,
 }
 
 /// A builder for the TestEnv structure.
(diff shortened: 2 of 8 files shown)


bowenwang1996:
Could you add a description to the PR to document what it does?

chain/client/src/test_utils.rs
```diff
@@ -1268,6 +1281,36 @@ impl TestEnv {
...
+                                .unwrap();
+                        }
+                        _ => {}
+                    }
+                } else {
+                    break;
+                }
+            }
+        }
+    }
```
bowenwang1996, line 1312: @yordanmadzhunkov this might be relevant for the work that you are doing

bowenwang1996 (approved):
Could we test the following are handled properly during the split?
* Postponed receipts
* Delayed receipts
* account deletion

chain/chain/src/store.rs
```diff
@@ -286,16 +286,19 @@ pub trait ChainStoreAccess {
...
     ) -> Result<EpochId, Error> {
         let mut candidate_hash = *hash;
+        let mut shard_id = shard_id;
         loop {
             let block_header = self.get_block_header(&candidate_hash)?;
             if block_header.chunk_mask()[shard_id as usize] {
                 break Ok(block_header.epoch_id().clone());
             }
             candidate_hash = *block_header.prev_hash();
+            shard_id = runtime_adapter.get_prev_shard_ids(&candidate_hash, vec![shard_id])?[0];
```
bowenwang1996, line 301: What does this do? Could you add a comment?

mzhangmzz: This converts the shard id to the corresponding shard id in the previous block. 

Otherwise, number of shards may change in this loop and the function will break

mzhangmzz:
> Could we test the following are handled properly during the split?
> 
> * Postponed receipts
> * Delayed receipts
> * account deletion

Yes I do plan to add more tests, this PR is just a beginning

chain/client/src/test_utils.rs
```diff
@@ -1270,6 +1283,36 @@ impl TestEnv {
...
+                                    partial_encoded_chunk.into(),
+                                ))
+                                .unwrap();
+                        }
+                        _ => {}
+                    }
+                } else {
+                    break;
+                }
+            }
```
matklad, line 1312: Isn't this just 

```
while let Some(request) = network_adapter.pop() {

}
```

chain/client/src/test_utils.rs
```diff
@@ -1270,6 +1283,36 @@ impl TestEnv {
...
+        &mut self.clients[self.account_to_client_index[account_id]]
+    }
+
+    pub fn process_partial_encoded_chunks(&mut self) {
+        let network_adapters = self.network_adapters.clone();
+        for network_adapter in network_adapters {
+            // process partial encoded chunks
+            loop {
+                if let Some(request) = network_adapter.pop() {
+                    match request {
```
matklad, line 1296: Heh, I'd write the `if let` as `match`, as it has the `else` branch, and `if let esle` is almost always noisier than the equivalent match. 

On the other hand, this `match` I'd write as `if let`, as it doesn't have an else branch.

core/store/src/test_utils.rs
```diff
@@ -49,7 +49,7 @@ pub fn test_populate_trie(
 }
 
 pub fn gen_accounts(rng: &mut impl Rng, max_size: usize) -> Vec<AccountId> {
-    let alphabet = &b"abcdefgh"[0..rng.gen_range(4, 8)];
```
matklad, line 52: Is this intentional? In general, there are benefits in randomizing the size of the alphabet -- if the alphabet is large, it's hard to get repeates and dupes.

integration-tests/tests/client/sharding_upgrade.rs
```diff
@@ -0,0 +1,274 @@
...
+use nearcore::config::GenesisExt;
+use nearcore::NEAR_BASE;
+
+const SIMPLE_NIGHTSHADE_PROTOCOL_VERSION: ProtocolVersion =
+    ProtocolFeature::SimpleNightshade.protocol_version();
+
+// Checks that account exists in the state after `block` is processed
+// This function checks both state_root from chunk extra and state root from chunk header, if
+// the corresponding chunk is included in the block
+fn check_account(env: &mut TestEnv, account_id: &AccountId, block: &Block) {
```
matklad, line 27: Probably better to default to `///` for item-level comments (they become semantic and processed by `cargo doc`), but no strong feeling here.

core/store/src/test_utils.rs
mzhangmzz: Yes, I actually added this function and I think all of use cases of this function right now actually do not want repeats. I will make the name more obvious for that though.

mzhangmzz:
@matklad since this PR has already landed, I address your comments in https://github.com/near/nearcore/pull/4935

PR #4922, opened 2021-10-01
Fix various minor documentation warnings

mina86:
Mostly wrapping URLs in angle brackets and code fragments in backticks
to make `cargo doc` happy.


matklad (approved):
Yay!

The obvious question is "should we put `cargo doc -Dwarning`" on CI to prevent this from regressing? I'd say no -- for internal crates, it creates more busy work than benefit. But fixing various thinghs once in a while is good!

bowenwang1996 (approved):

PR #4923, opened 2021-10-01
refactor: cleanup handling of compilation and caching errors in vm_runner

### LINKED ISSUE #4217: Get rid of serialization annotations on most runtime structures.

### LINKED ISSUE #4375: Process `VMError::CacheError` more safely
Ideologically, when there is some trouble loading the compiled module, `precompile_contract` and `action_function_call` should not result in `panic`. This is because compilation error are normal on the blockchain side when user submitted the uncompilable contract, thus it should be treated as a specific result, not an error.

OTOH, currently `FunctionCall` execution may result in `VMError::CacheError` which may be transferred to `StorageError::StorageInconsistentState` which lead to panic in `apply_transactions_with_optional_storage_proof`.

cc @bowenwang1996 @olonho

matklad:
builds up on top of #4913

Before we run a wasm contract, we need to compile it to machine code.
There are two categories of errors that can happen during this process.

_First_, the contract might be just an invalid wasm. In that case, we
should record that as an error into the state.

_Second_, we might get an IO error when trying to use compiled contract
cache due to the database being corrupt. In this case, we should emit a
critical error and exit.

We used to conflate the two paths into returning just a
`Result<VMError>`, which additionally contains error which *can not*
happen at this phase (that is, function runtime errors).

So in this PR we change `Result<T, VMError>` into more precise and
reliable `Result<Result<T, CompilationError>, CacheError>`.

Concretely, this brings the following benefits:

* we no longer need to mark the whole of `VMError` as `Serialize`, as we
    only store compilation errors in the cache.
* we can not accidentally mix logical and fatal errors when working with
    compilation

The following I think is still an issue with code, to be fixed in some
future refactor:

* VM Error (logic error) contains a CacheError (fatal error) as a
    variant.

closes #4217
closes #4375

Test Plan
----------

Preserves current behavior, so existing tests.


runtime/near-vm-runner/src/cache.rs
```diff
@@ -228,9 +254,29 @@ pub mod wasmer2_cache {
...
+            wasmer::CompileError::Validate(_) => {
+                CompilationError::WasmerCompileError { msg: err.to_string() }
+            }
+            wasmer::CompileError::UnsupportedFeature(_) => {
+                CompilationError::WasmerCompileError { msg: err.to_string() }
+            }
+            wasmer::CompileError::UnsupportedTarget(_) => {
+                CompilationError::WasmerCompileError { msg: err.to_string() }
+            }
+            wasmer::CompileError::Resource(_) => {
```
matklad, line 276: @olonho this preserves the existing behavior, but just makes it explicit what's going on. And, looking at this, it seems we might not actually treat some of these variants as deterministic errors?

runtime/near-vm-runner/src/cache.rs
```diff
@@ -275,25 +327,22 @@ pub mod wasmer2_cache {
...
+                    None => compile_and_serialize_wasmer2(wasm_code, &key, config, cache, store),
+                }
+            }
         }
     }
 
     #[cfg(not(feature = "no_cache"))]
     cached_key! {
-        MODULES: SizedCache<CryptoHash, Result<wasmer::Module, VMError>>
+        MODULES: SizedCache<CryptoHash, Result<Result<wasmer::Module, CompilationError>, CacheError>>
```
matklad, line 345: This (both old and new code) I think is wrong -- we shouldn't be storing fatal cache errors in memory cache, we should just always propagate this. This is the reason is why we drop the CompiledContractCache's io::Error from CacheError -- CacheError needs to be clone to store it in in-memory cache. 

Don't want to fix this in this PR, as figuring out how to implement correct caching with this macro is non-trivial.

runtime/near-vm-runner/src/cache.rs
```diff
@@ -304,7 +353,8 @@ pub mod wasmer2_cache {
             wasm_code: &[u8],
             config: &VMConfig,
             cache: Option<&dyn CompiledContractCache>,
-            store: &wasmer::Store) -> Result<wasmer::Module, VMError> = {
+            store: &wasmer::Store
```
matklad, line 356: @olonho not sure it's correct to pass `store` into this function -- it might be the case that the cached contract mixes two stores that way. From my reading of the code, this should be fine. if it is, we probably should add an assert with [`Store::same`](https://docs.rs/wasmer/2.0.0/wasmer/struct.Store.html#method.same).

bowenwang1996 (approved):

runtime/near-vm-runner/src/cache.rs
matklad: Discussed with olonho, this behaves correctly today, so no immediate action is required. But would be nice to straighten this up some day!

runtime/near-vm-runner/src/cache.rs
olonho: Checked that, it seems this is mostly deterministic error, although can happen in memory limited situations with resource constraints in some OSes (i.e. Linux). Guess we could keep it as it.

olonho (approved):

PR #4927, opened 2021-10-02
A prototype of socket read limiter

pmnoxx:
`ThrottleRateLimiter` will allow us to stop/resume reading from TcpSocket from given peer under specific conditions.

Current implementation:
- [x] stop reading after we detect that we have more than 20 unprocessed messages from peer
- [x] stop reading after we detect that the total size of all messages for peer is more than 512mb.
- [x] add a unit test for `RateLimiterHelper`

In the future we can extend this implementation to have more features. Please refer to design doc:  (https://docs.google.com/document/d/1FV3oyMRo52EQAgYroVd6Ws-5-ylkLSTdMOstY-IdL-M/edit?usp=sharing)

Related to https://github.com/near/nearcore/issues/5155


bowenwang1996 (commented):
Any specific reason for copying the entire tokio utils crate?

pmnoxx:
> Any specific reason for copying the entire tokio utils crate?
I removed most of tokio utils from that crate. I only files from `tokio-util`, which are included under `codec` feature.

`FramedRead` does have a few dependencies inside `src/codec` folder. I can strip down some function we don't use if needed.
I can also get rid of `FramedWrite`, though I would prefer to keep changes to `FramedRead` to minimum in order to preserve tests. Though if we want code to be shorter I can rewrite `FramedRead` to be much shorter, and l less generic than the one that was in `tokio-util.`

matklad:
Right now it's pretty hard to understand the motivation for copying huge swaths of code from tokio util. In the future, once this code gets to the master branch, it would be even more hard to figure out what we've changed and why. It would be helpful to see some doc comment/PR descriptions describing what is exactly the diff with the original tokio version

It's also not clear to me *why* we need this copying? https://docs.rs/tokio-util/0.6.8/tokio_util/codec/struct.FramedRead.html is parametrized over `T: AsyncRead`, I would expect that we'd rate-limit the `AsyncRead` instead of doing framing and rate-limiting together? 

```rust
            let (read, write) = tokio::io::split(stream);
            let read = ThrottledReader::with_bytes_per_duration(
                1_000_000,
                Duration::from_millis(100),
                read,
            );

            // TODO: check if peer is banned or known based on IP address and port.
            Peer::add_stream(
                // limit reading to 10 mb/s
                FramedRead::new(read, Codec::new())
```

pmnoxx:
> Right now it's pretty hard to understand the motivation for copying huge swaths of code from tokio util. In the future, once this code gets to the master branch, it would be even more hard to figure out what we've changed and why. It would be helpful to see some doc comment/PR descriptions describing what is exactly the diff with the original tokio version
> 
> It's also not clear to me _why_ we need this copying? https://docs.rs/tokio-util/0.6.8/tokio_util/codec/struct.FramedRead.html is parametrized over `T: AsyncRead`, I would expect that we'd rate-limit the `AsyncRead` instead of doing framing and rate-limiting together?
> 
> ```rust
>             let (read, write) = tokio::io::split(stream);
>             let read = ThrottledReader::with_bytes_per_duration(
>                 1_000_000,
>                 Duration::from_millis(100),
>                 read,
>             );
> 
>             // TODO: check if peer is banned or known based on IP address and port.
>             Peer::add_stream(
>                 // limit reading to 10 mb/s
>                 FramedRead::new(read, Codec::new())
> ```

Thanks, that's a really good question. I agree, if our goal was only to limit the amount of data we read from the socket, it would make sense to create a feature between FramedRead and TCPSocket. That approach would be simpler, and cleaner.

However, we want to be able to stop reading from socket based on a few factors:
- the amount of bandwidth used - This is already in this PR.
- the total number of messages from `Peer` - If a peer send us a few messages, they may be expensive to process, so it makes sense to limit the number of messages that are in progress. Therefore we need to be able to limit reading messages from socket on `per message level` instead of `per byte level`.
- total amount of memory used - to do this, both size of messages will need to be known, as well as, the number of messages that are being processed by peer.

In order to be able to do throttling on 3 factors mentioned above, we need to do changes to `FramedRead`

bowenwang1996:
> However, we want to be able to stop reading from socket based on a few factors:

Do you need to accomplish all of them at once? Also is changing `FrameRead` the only way to do it? I believe for number of messages we could do this in the implementation of `Codec`.

pmnoxx:
> > However, we want to be able to stop reading from socket based on a few factors:
> 
> Do you need to accomplish all of them at once? Also is changing `FrameRead` the only way to do it? I believe for number of messages we could do this in the implementation of `Codec`.

Yes, that's a good question. 

> Do you need to accomplish all of them at once?

We don't need to accomplish everything at once, doing this separately will be more complicated. I'm going to rewrite `FramedRead` to be shorter and easier to read.

> Also is changing `FrameRead` the only way to do it? 
No, it's not the only way to do it. Though I think this is the cleanest way. Theoretically, you could create a wrapper around `TcpReader` and appropriate changes in `Codec`, and have them communicate together through `atomics`. I would rather not do that. Though if you are interested I can describe that solution in detail.

> I believe for number of messages we could do this in the implementation of `Codec`.
Well, yes and no. Imagine a situation when someone sends you over a network 1000 messages, but you only decode 20 through `Codec`. Once you are done processing those 20 messages, the remaining 980 are stuck inside `FramedBuffer`, you have no access to them. In order to process more messages, you need to use some mechanism of waking up the future, in order to trigger `decode` method. That usually happens only on socket read, and you can't really on that.

pmnoxx:
My plan is to create following data structure, that will control reading from the socket
```rust
struct PeerStats {
// number of messages from Peer that we haven't finished processing
    messages_in_progress: AtomicUsize,
// total size of memory used by all messages in peer.
    total_size_f_messages: AtomicUsize,
};
```

I see two possible approaches
1) Used modified `FramedReader` to maintain this data structure. Whenever we process messages anywhere in the code, we will update those fields, and `FramedReader` will check every `20`ms whenever it's safe to resume reading.

2) Create a wrapper in between `TcpRead` and `FramedReader`. Let's call it `Waker`. This wrapper will be a future, that checks `PeerStats` whenever it's ok to read more data from tcp socket or not. It can be woken at will by using `AtomicWaker`. 
And we will have to modify `Codec`, codec will update stats inside `PeerStats`, and let know `Waker` whenever to stop reading or not. In other places in the code, after we finish processing messages from peer, we will update `PeerStats` and wake `Waker` through use of `AtomicWaker`.
In the second approach we can avoid modifying `FramedRead`, by splitting logic into two places.

matklad:
> is to create following data structure,

:thinking: the problem with this setup is that raw Atomics don't have notifications mechanism. You can `messages_in_progress.fetch_sub(1)`, but you can't say "and now whatever was blocked on this can be unblocked", we have to use busy-polling. I am not a huge fan of polling in general personally -- it adds latency (especially in tests), burns CPU, and makes stuff more difficult to test. I don't know about polling in the network layer specifically, maybe it's OK. 

>In order to be able to do throttling on 3 factors mentioned above, we need to do changes to FramedRead

:thinking: is it really the best way to do this? My understanding is that one of the ideas behind Rust async with streams, codecs, AsyncRead and such is that of the composability. Rather than building a single thing that takes a TcpSocket, frames it, applies backpressure, decrypts and then uncompresses the contents, you just chain together various utilities for this. 

In this case, it seems that for limiting based on messages we should just let the caller signal when it can process the next message. Something like this could work:

```rust
fn throttle_stream<I, S: Stream<Item = I>>(s: S) -> (impl Stream<Item = I>, Arc<Semaphore>) {
    let sema = Arc::new(Semaphore::new(0));
    let permits = futures_lite::stream::unfold((), {
        let sema = Arc::clone(&sema);
        move |()| {
            let sema = Arc::clone(&sema);
            async move {
                let permit = sema.acquire().await;
                match permit {
                    Ok(permit) => {
                        permit.forget();
                        Some(((), ()))
                    }
                    Err(_) => None,
                }
            }
        }
    });

    let throttled = permits.zip(s).map(|((), msg)| msg);
    (throttled, sema)
}
```

The caller gets a semaphore, and they can add permits to it to receive next message. *Maybe* we want to write our fancy multi-dimensional semaphor which separately tracks the size and the number of messges, but it doesn't seem like it needs to be built-in into the FrameRead. 

Similarly, we can write semaphorish-thing to throttle an `AsyncRead`

```rust
fn throttle_read<R: AsyncRead + Unpin>(
    read: R,
    n_bytes: usize,
    duration: Duration,
) -> impl AsyncRead {
    struct ThrottledRead<R> {
        sema: PollSemaphore,
        read: R,
    }

    impl<R: AsyncRead + Unpin> AsyncRead for ThrottledRead<R> {
        fn poll_read(
            mut self: std::pin::Pin<&mut Self>,
            cx: &mut std::task::Context<'_>,
            buf: &mut tokio::io::ReadBuf<'_>,
        ) -> std::task::Poll<std::io::Result<()>> {
            // Should acquire many permits
            // https://github.com/tokio-rs/tokio/issues/3209#issuecomment-934634720
            let permit = match ready!(self.sema.poll_acquire(cx)) {
                Some(permit) => permit,
                None => {
                    return Poll::Ready(Err(std::io::Error::new(
                        std::io::ErrorKind::Other,
                        "closed semaphore",
                    )))
                }
            };
            permit.forget();
            Pin::new(&mut self.read).poll_read(cx, buf)
        }
    }

    let read_sema = Arc::new(Semaphore::new(n_bytes));
    let read = ThrottledRead {
        sema: PollSemaphore::new(Arc::clone(&read_sema)),
        read,
    };
    tokio::spawn(async move {
        loop {
            tokio::time::sleep(duration).await;
            // FIXME: this is racy, gotta write our own semaphore it seems.
            let available = read_sema.available_permits();
            read_sema.add_permits(n_bytes.sub(available));
        }
    });
    read
}
```

Here, again the built-in semaphore isn't quite enough, we can do something smarter (and correct for starters), but it seems that we can do this in a composable fashion? 

Then, we can put the two together to throttle our actual peer messages: 

```rust
fn limit(
    read: ReadHalf<TcpStream>,
    n_bytes: usize,
    duration: Duration,
) -> (
    impl Stream<Item = Result<Vec<u8>, std::io::Error>>,
    Arc<Semaphore>,
) {
    let read = throttle_read(read, n_bytes, duration);
    throttle_stream(FramedRead::new(read, Codec))
}
```

utils/near-framed-read/src/framed_read.rs
```diff
@@ -0,0 +1,72 @@
...
+use std::task::{Context, Poll};
+
+pin_project! {
+    pub struct NearFramedRead<T, D> {
+        #[pin]
+        inner: FramedImpl<T, D, ReadFrame>,
+    }
+}
+
+impl<T, D> NearFramedRead<T, D>
```
matklad, line 22: I don't think there's anything specific to near in the way this thing operates, a better name would be `ThrottledFrameRead`.

utils/near-framed-read/src/framed_impl.rs
```diff
@@ -0,0 +1,144 @@
...
+        // Ensure the pointer does not change from under us
+        assert_eq!(ptr, buf.filled().as_ptr());
+        buf.filled().len()
+    };
+
+    // Safety: This is guaranteed to be the number of initialized (and read)
+    // bytes due to the invariants provided by `ReadBuf::filled`.
+    unsafe {
+        buf.advance_mut(n);
+    }
```
matklad, line 141: Seems like here we can use the original implementation?

https://docs.rs/tokio-util/0.6.9/tokio_util/io/fn.poll_read_buf.html

Reading into uninitialized buffers is notoriously tricky in Rust, as the rules of what is UB are not super specified yet. I wouldn't want us to have a copy of this code, if that is possible.

utils/near-framed-read/src/framed_impl.rs
```diff
@@ -0,0 +1,144 @@
...
+                        pinned.delay.reset(*pinned.duration);
+                    }
+
+                    Poll::Pending => {
+                        return Poll::Pending;
+                    }
+                }
+                // We may wait more than 1 duration, to make sure we don't go over read limit.
+                if *pinned.bytes_left < 0 {
+                    return Poll::Pending;
```
matklad, line 72: Hm, this feels like a bug -- we return a `Pending` here, but we don't seem to arrange for `waker` to be called. So if we hit this code path, the executor might never poll this future. Not entirely sure what's the proper fix is, but maybe just `continue;` instead of `return`?

matklad (commented):
With +341 this looks much more reasonable! This makes sense to me overall! As this is a draft, I didn't do a super thorough review for correctness, but I've left a couple of comments about things to address!

utils/near-framed-read/.cargo_vcs_info.json
```diff
@@ -0,0 +1,5 @@
+{
+  "git": {
+    "sha1": "60d88840f4f3ea6b7d1ced6118a7ddef1290e542"
+  }
+}
```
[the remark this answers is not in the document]
pmnoxx: Thanks, removed.

utils/near-framed-read/src/framed_read.rs
pmnoxx: That's a good idea.

Cargo.lock
```diff
@@ -1714,6 +1714,12 @@ version = "0.3.13"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "fa189ef211c15ee602667a6fcfe1c1fd9e07d42250d2156382820fba33c9df80"
 
+[[package]]
+name = "futures-timer"
```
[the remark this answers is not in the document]
pmnoxx: Ok, I removed it.

bowenwang1996:
@pmnoxx if you think the PR is ready for review, please mark it as such

pmnoxx:
@bowenwang1996 @matklad Ok, I marked it as ready to review.

pmnoxx:
@mm-near @matklad @bowenwang1996 I added a design doc https://docs.google.com/document/d/1FV3oyMRo52EQAgYroVd6Ws-5-ylkLSTdMOstY-IdL-M/edit

mm-near:
Thanks - can you also give us commenting rights ?

chain/network/Cargo.toml
```diff
@@ -35,8 +35,10 @@ near-metrics = { path = "../../core/metrics" }
 near-performance-metrics = { path = "../../utils/near-performance-metrics" }
 near-performance-metrics-macros = { path = "../../utils/near-performance-metrics-macros" }
 near-stable-hasher = { path = "../../utils/near-stable-hasher" }
+near-framed-read = {path = "../../utils/near-framed-read" }
 
 delay-detector = { path = "../../tools/delay_detector", optional = true}
+pin-project-lite = "0.2.7"
```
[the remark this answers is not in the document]
pmnoxx: Removed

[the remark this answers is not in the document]
pmnoxx: Initially I intended this PR's code to be inside network, but I moved it out.

chain/network/src/peer.rs
```diff
@@ -695,6 +700,9 @@ impl StreamHandler<Result<Vec<u8>, ReasonForBan>> for Peer {
                 return;
             }
         };
+        // TODO We should change our code to track size of messages received from Peer
+        // as long as it travels to PeerManager, etc.
```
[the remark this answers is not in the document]
pmnoxx: https://github.com/near/nearcore/issues/5155

chain/network/src/peer_manager.rs
```diff
@@ -644,8 +645,12 @@ impl PeerManagerActor {
             let (read, write) = tokio::io::split(stream);
 
             // TODO: check if peer is banned or known based on IP address and port.
+
+            let (tx, rx) = unbounded();
+            let rate_limiter = RateLimiterHelper::new(tx);
             Peer::add_stream(
-                FramedRead::new(read, Codec::new())
+                // limit reading to 10 mb/s
```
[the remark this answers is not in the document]
pmnoxx: To make this change simpler. I added adding bandwidth limits to be the #5155 This can be done as separate PR.

utils/near-framed-read/src/framed_impl.rs
pmnoxx: Done

utils/near-framed-read/src/framed_read.rs
```diff
@@ -0,0 +1,290 @@
...
+    pub tx: UnboundedSender<()>,
+}
+
+impl RateLimiterHelper {
+    pub fn new(tx: UnboundedSender<()>) -> Self {
+        Self {
+            num_messages_in_progress: Arc::new(AtomicUsize::new(0)),
+            max_messages: 20,
+            sizeof_messages_in_progress: Arc::new(AtomicUsize::new(0)),
+            max_sizeof_messages: 500_000_000,
```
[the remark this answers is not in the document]
pmnoxx: I moved those values into constants for now. We could move later them to `new` if we find out that we are actually changing those constants often.

chain/network/src/peer.rs
```diff
@@ -1004,6 +1012,11 @@ impl StreamHandler<Result<Vec<u8>, ReasonForBan>> for Peer {
             }
             #[cfg(feature = "protocol_feature_routing_exchange_algorithm")]
             (_, _state, PeerMessage::RoutingTableSyncV2(ibf_message)) => {
+                // TODO Add wrapper to be something like this for all messages.
+                // self.peer_manager_addr.do_send(ActixMessageWrapper<NetworkRequests>::new(
```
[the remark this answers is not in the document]
pmnoxx: Fixed.

pmnoxx:
I added limiting size of read buffer. If someone were to send us msg of size more than 8GB, a node would crash:
Take a look at:
```
            // If at this point we have more elements in the buffer than `MAX_MESSAGES_TOTAL_SIZE`,
            // that means we received a message that was longer than `MAX_MESSAGES_TOTAL_SIZE`.
            if state.buffer.len() as u64 > MAX_MESSAGES_TOTAL_SIZE {
                state.eof = true;
                state.is_readable = true;
                continue;
            }
```

chain/network/src/peer.rs
```diff
@@ -695,6 +700,9 @@ impl StreamHandler<Result<Vec<u8>, ReasonForBan>> for Peer {
                 return;
             }
         };
+        // TODO(5155) We should change our code to track size of messages received from Peer
```
bowenwang1996, line 703:
```suggestion
        // TODO(#5155) We should change our code to track size of messages received from Peer
```

utils/near-framed-read/Cargo.toml
```diff
@@ -0,0 +1,28 @@
...
+#
+# If you believe there's an error in this file please file an
+# issue against the rust-lang/cargo repository. If you're
+# editing this file be aware that the upstream Cargo.toml
+# will likely look very different (and much more reasonable)
+
+[package]
+edition = "2018"
+name = "near-framed-read"
+version = "0.6.3"
```
bowenwang1996, line 16: Why do we start versioning it at 0.6.3? cc @matklad

utils/near-framed-read/src/framed_read.rs
```diff
@@ -0,0 +1,276 @@
...
+        pub(crate) receiver: UnboundedReceiver<()>,
+    }
+}
+
+// Wrapper around Actix messages, used to track size of all messages sent to PeerManager.
+// TODO(#5155) Finish implementation of this.
+
+#[allow(unused)]
+pub struct ActixMessageWrapper<T> {
+    msg: T, // TODO
```
bowenwang1996, line 49: What is left to be done? Please be more clear on that

bowenwang1996 (approved):

utils/near-framed-read/Cargo.toml
pmnoxx: Thanks, I changed it to 0.1.

chain/network/src/peer.rs
pmnoxx: Fixes, thanks.

pmnoxx:
@matklad I finished `remove_msg` method. It's not implemented properly. The main thing that's stopping me from marking this PR as done is not having a unit test.

utils/near-framed-read/src/framed_read.rs
```diff
@@ -0,0 +1,296 @@
+use std::borrow::BorrowMut;
+use std::io;
+use std::pin::Pin;
+use std::sync::atomic::{AtomicU64, AtomicUsize, Ordering};
+use std::sync::Arc;
+use std::task::{Context, Poll};
+
+use bytes::BytesMut;
+use futures::channel::mpsc::{UnboundedReceiver, UnboundedSender};
+use futures::SinkExt;
```
matklad, line 10: Ok, I think this is actually the core problem of why `remove_msg` creates problems. 

The TL;DR here is that we shouldn't be using `futures` crate, and instead use channels from tokio. They have an explicit, documented solution for bridging sync and async worlds: 

https://docs.rs/tokio/1.13.0/tokio/sync/mpsc/index.html#communicating-between-sync-and-async-code

More generally, we shouldn't be using the `futures` crate anywhere, and instead stick to pure `tokio`. Some historical context: 

The `futures` crate was was historically the first approach to non-blocking Rust, designed well in advance of `async/await` language support. `async/await` was retrofitted onto the `futures`, but not perfectly -- by that time, `futures` was already a widely used crate, so it had a lot of constraints on API evolution. 

After async-await shipped, community spend a lot of iteration on designing tokio and alternative runtimes (smol, async-std). At that time, the advice was "try to stick to futures crate, as that is runtime agnostic". 

Today, tokio actually shipped a stable 1.0 version. Now, it esstially fulfils async-std vision -- an almost drop-in replacement for types from std, compatible with epoll event loop. Everything that the `futures`  crate used to do is now available in tokio, and generally has better design. Channels are just an instance of this: tokio version takes lessons learned from futures, and makes "send from blocking to async world" an explicitly supported pattern.

pmnoxx: @matklad Thanks, thanks for the update. Yes, in the past using `futures` crate was the recommended solution.

Anyway, I stopped using the `futures` crate in `framed_read.rs`. And I switched to using `tokio::sync::mpsc::unbounded_channel::<()>();` instead.
Please take a look.

pmnoxx:
I'm working on adding one unit test for testing `ThrottledFrameRead`. After that I'll finish `PeerManagerActor` part of the code.

pmnoxx:
@matklad I fixed the code to prevent the memory leak. Maybe we should use bounded queue instead of `unbounded`?

Take a look at:

```
                    while let Some(_) = ready!(pinned.receiver.poll_recv(cx)) {
                        // pop all elements from queue, to prevent memory leak
                    }
```

pmnoxx:
@matklad I simplified the design. There is no longer need to call `add_msg` inside `ThrottledFrameRead`.

We will only call `add_msg`, `remove_msg` while the message is in the process of being transferred between `Actors`.

pmnoxx:
@bowenwang1996 @mm-near I think this PR is ready to be merged. I'll wait for @matklad to take a look at it again.

utils/near-rate-limiter/src/framed_read.rs
```diff
@@ -0,0 +1,380 @@
...
+        let state: &mut ReadFrame = pinned.state.borrow_mut();
+        loop {
+            while let task::Poll::Ready(_) = pinned.receiver.poll_recv(cx) {
+                // pop all elements from queue, to prevent memory leak
+            }
+            while !pinned.rate_limiter.is_ready() {
+                // This will cause us to subscribe to notifier when something gets pushed to
+                // `pinned.receiver`. If there is an element in the queue, we will check again.
+                ready!(pinned.receiver.poll_recv(cx));
+            }
```
matklad, line 219: Hm, so basically what we are doing is simulating level triggering (there's some message in channel) via edge triggerring (we receieved a new message). Intuitively, it seems like there should be a more direct synchronization primitive for this case. The right way place to shop for them is the `tokio::sync` module 

https://docs.rs/tokio/1.13.0/tokio/sync/index.html#structs

Looking at the list, it seems that `Notify` is what we need here: 

https://docs.rs/tokio/1.13.0/tokio/sync/struct.Notify.html

utils/near-rate-limiter/src/framed_read.rs
```diff
@@ -0,0 +1,380 @@
...
+            // get a spurious 0 that looks like EOF
+            state.buffer.reserve(1);
+
+            let bytect = match poll_read_buf(pinned.inner.as_mut(), cx, &mut state.buffer)? {
+                Poll::Ready(ct) => ct,
+                // We are waiting for more data on TCP socket.
+                Poll::Pending => {
+                    return Poll::Pending;
+                }
+            };
```
matklad, line 254: Let's use the `ready!` macro here. Explicit `return Pending` makes the reader think about potential lost wakeup (returning Pending without arranging for the waker to be notified). With `ready!`, this problem is impossible by construction.

utils/near-rate-limiter/src/framed_read.rs
```diff
@@ -0,0 +1,380 @@
...
+            if bytect == 0 {
+                state.eof = true;
+            }
+
+            state.is_readable = true;
+        }
+    }
+}
+
+pub trait Decoder {
```
matklad, line 265: Introducing our own `Decoder` trait feels quite suboptimal (as all the `impl tokio_utils::Decoder for X` become void), let's re-use the trait from tokio_utils.

utils/near-rate-limiter/src/framed_read.rs
```diff
@@ -0,0 +1,380 @@
...
+#[derive(Clone, Debug)]
+pub struct RateLimiterHelper {
+    // Total count of all messages that are tracked by `RateLimiterHelper`
+    pub num_messages_in_progress: Arc<AtomicUsize>,
+    // Total size of all messages that are tracked by `RateLimiterHelper`
+    pub total_sizeof_messages_in_progress: Arc<AtomicUsize>,
+    // We have an unbounded queue of messages, that we use to wake `ThrottledRateLimiter`.
+    // This is the sender part, which is used to notify `ThrottledRateLimiter` to try to
+    // read again from queue.
+    pub tx: UnboundedSender<()>,
```
matklad, line 91: Let's make the fields private: this is again from a perspective of "let's try to keep this a self-contained utility".

utils/near-rate-limiter/src/framed_read.rs
```diff
@@ -0,0 +1,380 @@
...
+    pub fn new(tx: UnboundedSender<()>) -> Self {
+        Self {
+            num_messages_in_progress: Arc::new(AtomicUsize::new(0)),
+            total_sizeof_messages_in_progress: Arc::new(AtomicUsize::new(0)),
+            tx,
+        }
+    }
+
+    // Check whenever
+    pub fn is_ready(&self) -> bool {
```
matklad, line 110: can be private

utils/near-rate-limiter/src/framed_read.rs
```diff
@@ -0,0 +1,380 @@
...
+    // This is the sender part, which is used to notify `ThrottledRateLimiter` to try to
+    // read again from queue.
+    pub tx: UnboundedSender<()>,
+}
+
+/// `RateLimiterHelper` is a helper data structure that stores message count/total memory
+/// limits per Peer. It controls whenever associated `ThrottleFramedReader` is able to read
+/// from TcpSocket. It accepts `tx` as argument, which is used to notify `ThrottleFramedReader`
+/// to wake up and consider reading again. That happens when a message is removed, and limits
+/// are increased.
```
matklad, line 98: Let's place the docs onto the struct definition.

utils/near-rate-limiter/src/framed_read.rs
```diff
@@ -0,0 +1,380 @@
...
+}
+
+/// Each RateLimiterHelper is associated with exactly one `Peer`. It keeps track total size of all
+/// messages / their count that were received from given peer. We are going to keep tracking them
+/// as long as their exist in transit in inside `Actix` queues. For example from `Peer` to
+/// `PeerManager` and back to `Peer`.
+///
+/// TODO (#5155) Add throttling by bandwidth.
+#[derive(Clone, Debug)]
+pub struct RateLimiterHelper {
```
matklad, line 83: Not a huge fan of the naming:

* RateLimiter is inconsistent with throttling terminology we use elsewhere 
* `Helper` suffix adds zero bits of information

Don't know a good, catchy name here, but something unimaginative like `ThrottleController` might work.

utils/near-rate-limiter/src/framed_read.rs
pmnoxx: Thanks, fixed.

utils/near-rate-limiter/src/framed_read.rs
pmnoxx: That's a good idea. I implemented this the same way as `FrameReader` is implemented in `tokio_utils`. Maybe we should update it there as well.

utils/near-rate-limiter/Cargo.toml
```diff
@@ -0,0 +1,15 @@
...
+
+[dependencies]
+bytes = "1.0.0"
+futures-core = "0.3.0"
+log = "0.4"
+pin-project-lite = "0.2.0"
+slab = "0.4.1"
+tokio = { version = "1.0.0", features = ["sync"]}
+tokio-util = { version = "0.6.9", features = ["io"]}
+actix = "=0.11.0-beta.2"
```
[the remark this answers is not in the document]
pmnoxx: I'm using it for tests
```

        actix::System::new().block_on(async {
            for _ in 0..16 {
                rx.recv().await.unwrap();
            }
            assert_eq!(rx.try_recv().is_err(), true);
        });
```

utils/near-rate-limiter/src/framed_read.rs
pmnoxx: Ok, `ThrottleController` is fine with me.

utils/near-rate-limiter/src/framed_read.rs
pmnoxx: I tried to use `Notify`, but I couldn't get it to work. I'm having trouble writing code, which pools from it using `notify2.notified()`

Consider the following code:
```
    let notify = Arc::new(Notify::new());

    tokio::spawn(async move {
        notify2.notified().await;
        println!("received notification");
    });
```
 notify2.notified() returns `Notified<'_>`. In order to pool on a future, I need to pin it.
However, `Notified` keep a reference to `Notify`, that makes things more complicated to pool.

utils/near-rate-limiter/src/framed_read.rs
```diff
@@ -0,0 +1,380 @@
...
+use tokio::sync::mpsc::{UnboundedReceiver, UnboundedSender};
+use tokio_util::io::poll_read_buf;
+
+// Initial capacity of read buffer.
+const INITIAL_CAPACITY: usize = 8 * 1024;
+
+// Max number of messages we received from peer, and they are in progress, before we start throttling.
+const MAX_MESSAGES_COUNT: usize = 20;
+// Max total size of all messages that are in progress, before we start throttling.
+const MAX_MESSAGES_TOTAL_SIZE: usize = 500_000_000;
```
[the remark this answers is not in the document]
pmnoxx: Done.

utils/near-rate-limiter/Cargo.toml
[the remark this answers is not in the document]
pmnoxx: I removed actix, and this part of test for now.

utils/near-rate-limiter/src/framed_read.rs
pmnoxx: Ok, I added TODO.

I see that both `Notify` and `Semaphore` have a similar issue.

pmnoxx:
@matklad I had to bring back `actix` package, but this dependency is used for tests only.

utils/near-rate-limiter/src/framed_read.rs
matklad: Aha, indeed missed that. Yeah, I believe there aren't super-nice solutions here, so probably a comment is all there is needed. 

For semaphore, tokio hash https://docs.rs/tokio-util/0.6.9/tokio_util/sync/struct.PollSemaphore.html specifically for this case. So, I guess a long-term soln here is to send a PR to tokio-utils adding `PollNotify`?

chain/network/Cargo.toml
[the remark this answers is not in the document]
matklad: `pin-project-lite` is a good dependency, in two senses: 

* it is small, well maintained crate
* it is a dependency of tokio itself, so we can't not use it

Meta heuristic: for dependencies, look at the changes in Cargo.**lock**. If there are new packages in lockfile, we need to think twice. If it just uses a package we use otherwise, it is probably OK (but still needs an "is this the right level of abstraction" sanity check.)

utils/near-rate-limiter/Cargo.toml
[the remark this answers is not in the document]
matklad: Hm, it seems we still have an actix-dev dependency? This doesn't make sense to me -- this layer is logically strictly below actix, we don't need actors here even for testing. Here's how I'd write the test using only tokio: 

```diff
diff --git a/Cargo.lock b/Cargo.lock
index d5fafab9d..453a0473e 100644
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2883,7 +2883,6 @@ dependencies = [
 name = "near-rate-limiter"
 version = "0.1.0"
 dependencies = [
- "actix",
  "bytes",
  "futures-core",
  "log",
diff --git a/utils/near-rate-limiter/Cargo.toml b/utils/near-rate-limiter/Cargo.toml
index d5ce1b25a..77e5adf68 100644
--- a/utils/near-rate-limiter/Cargo.toml
+++ b/utils/near-rate-limiter/Cargo.toml
@@ -2,7 +2,6 @@
 edition = "2018"
 name = "near-rate-limiter"
 version = "0.1.0"
-tokio-util = { version = "0.6", features = ["codec"]}
 
 [dependencies]
 bytes = "1.0.0"
@@ -11,7 +10,7 @@ log = "0.4"
 pin-project-lite = "0.2.0"
 slab = "0.4.1"
 tokio = { version = "1.0.0", features = ["sync"]}
-tokio-util = { version = "0.6.9", features = ["io"]}
+tokio-util = { version = "0.6.9", features = ["io", "codec"]}
 
 [dev-dependencies]
-actix = "=0.11.0-beta.2"
\ No newline at end of file
+tokio = { version = "1.0.0", features = ["sync", "rt-multi-thread"]}
diff --git a/utils/near-rate-limiter/src/framed_read.rs b/utils/near-rate-limiter/src/framed_read.rs
index aa4ecf386..3e87cdae6 100644
--- a/utils/near-rate-limiter/src/framed_read.rs
+++ b/utils/near-rate-limiter/src/framed_read.rs
@@ -307,12 +307,12 @@ mod tests {
         assert_eq!(throttle_controller.num_messages_in_progress.load(SeqCst), 0);
         assert_eq!(throttle_controller.total_sizeof_messages_in_progress.load(SeqCst), 0);
 
-        actix::System::new().block_on(async {
+        tokio::runtime::Runtime::new().unwrap().block_on(async {
             for _ in 0..40 {
                 rx.recv().await.unwrap();
             }
             assert_eq!(rx.try_recv().is_err(), true);
-        });
+        })
     }
 
     #[test]
@@ -352,7 +352,7 @@ mod tests {
         assert_eq!(throttle_controller.num_messages_in_progress.load(SeqCst), 0);
         assert_eq!(throttle_controller.total_sizeof_messages_in_progress.load(SeqCst), 0);
 
-        actix::System::new().block_on(async {
+        tokio::runtime::Runtime::new().unwrap().block_on(async {
             for _ in 0..16 {
                 rx.recv().await.unwrap();
             }
```

Or an alternative, using macros: 

```diff
diff --git a/Cargo.lock b/Cargo.lock
index d5fafab9d..453a0473e 100644
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2883,7 +2883,6 @@ dependencies = [
 name = "near-rate-limiter"
 version = "0.1.0"
 dependencies = [
- "actix",
  "bytes",
  "futures-core",
  "log",
diff --git a/utils/near-rate-limiter/Cargo.toml b/utils/near-rate-limiter/Cargo.toml
index d5ce1b25a..380f3de7b 100644
--- a/utils/near-rate-limiter/Cargo.toml
+++ b/utils/near-rate-limiter/Cargo.toml
@@ -2,7 +2,6 @@
 edition = "2018"
 name = "near-rate-limiter"
 version = "0.1.0"
-tokio-util = { version = "0.6", features = ["codec"]}
 
 [dependencies]
 bytes = "1.0.0"
@@ -11,7 +10,7 @@ log = "0.4"
 pin-project-lite = "0.2.0"
 slab = "0.4.1"
 tokio = { version = "1.0.0", features = ["sync"]}
-tokio-util = { version = "0.6.9", features = ["io"]}
+tokio-util = { version = "0.6.9", features = ["io", "codec"]}
 
 [dev-dependencies]
-actix = "=0.11.0-beta.2"
\ No newline at end of file
+tokio = { version = "1.0.0", features = ["sync", "rt", "macros"]}
diff --git a/utils/near-rate-limiter/src/framed_read.rs b/utils/near-rate-limiter/src/framed_read.rs
index aa4ecf386..d2b615740 100644
--- a/utils/near-rate-limiter/src/framed_read.rs
+++ b/utils/near-rate-limiter/src/framed_read.rs
@@ -268,8 +268,8 @@ mod tests {
     use std::sync::atomic::Ordering::SeqCst;
     use tokio::sync::mpsc;
 
-    #[test]
-    fn test_rate_limiter_helper_by_count() {
+    #[tokio::test]
+    async fn test_rate_limiter_helper_by_count() {
         let (tx, mut rx) = mpsc::unbounded_channel::<()>();
         let max_messages_count = 20;
         let mut throttle_controller = ThrottleController::new(tx, max_messages_count, 2000000);
@@ -307,16 +307,14 @@ mod tests {
         assert_eq!(throttle_controller.num_messages_in_progress.load(SeqCst), 0);
         assert_eq!(throttle_controller.total_sizeof_messages_in_progress.load(SeqCst), 0);
 
-        actix::System::new().block_on(async {
-            for _ in 0..40 {
-                rx.recv().await.unwrap();
-            }
-            assert_eq!(rx.try_recv().is_err(), true);
-        });
+        for _ in 0..40 {
+            rx.recv().await.unwrap();
+        }
+        assert_eq!(rx.try_recv().is_err(), true);
     }
 
-    #[test]
-    fn test_rate_limiter_helper_by_size() {
+    #[tokio::test]
+    async fn test_rate_limiter_helper_by_size() {
         let max_messages_total_size = 500_000_000;
         let (tx, mut rx) = mpsc::unbounded_channel::<()>();
         let mut throttle_controller = ThrottleController::new(tx, 1000, max_messages_total_size);
@@ -352,11 +350,9 @@ mod tests {
         assert_eq!(throttle_controller.num_messages_in_progress.load(SeqCst), 0);
         assert_eq!(throttle_controller.total_sizeof_messages_in_progress.load(SeqCst), 0);
 
-        actix::System::new().block_on(async {
-            for _ in 0..16 {
-                rx.recv().await.unwrap();
-            }
-            assert_eq!(rx.try_recv().is_err(), true);
-        });
+        for _ in 0..16 {
+            rx.recv().await.unwrap();
+        }
+        assert_eq!(rx.try_recv().is_err(), true);
     }
 }
```

utils/near-rate-limiter/Cargo.toml
```diff
@@ -0,0 +1,17 @@
+[package]
+edition = "2018"
+name = "near-rate-limiter"
+version = "0.1.0"
+tokio-util = { version = "0.6", features = ["codec"]}
```
matklad, line 5: misplaced dep

utils/near-rate-limiter/Cargo.toml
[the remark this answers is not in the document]
pmnoxx: Thanks, I removed actix dependency.

utils/near-rate-limiter/src/framed_read.rs
pmnoxx: Ok, I started using `PoolSemaphore`, the code is much cleaner, thanks.

Yes, adding `PollNotify` PR seems like a good idea.

pmnoxx: Anyway, I'm using now the `PoolSemaphore` as I would `PollNotify`. So it's not an urgent thing.

utils/near-rate-limiter/src/framed_read.rs
```diff
@@ -0,0 +1,354 @@
...
+        let state: &mut ReadFrame = pinned.state.borrow_mut();
+        loop {
+            if !pinned.throttle_controller.is_ready() {
+                // This will cause us to subscribe to notifier when something gets pushed to
+                // `pinned.receiver`. If there is an element in the queue, we will check again.
+                ready!(PollSemaphore::poll_next(Pin::new(pinned.semaphore), cx));
+            }
+            if !pinned.throttle_controller.is_ready() {
+                return Poll::Pending;
+            }
```
matklad, line 223: Hm, I don't understand how this branch works. It is checked on the same event loop iteration as the one before, with identical condition, so, logically, it's just redundant? If we do end up hitting this `return Pendng`, than we don't have anything scheduled to wake us up, no?

matklad (approved):
LGTM overall, although I am still irked a tiny bit by the incosistency between ThrottleController type name and `rate-limiter` crate name, and the fact that we define ActixMessageWrapper in this crate.

pmnoxx:
> LGTM overall, although I am still irked a tiny bit by the incosistency between ThrottleController type name and `rate-limiter` crate name, and the fact that we define ActixMessageWrapper in this crate.

We don't use `Actix` in this crate. So I removed `ActixMessageWrapper` from here. Besides, in the next PR we will need to implement some features from `Actix` to make it work anyway.

Besides, I changed sending data to semaphore, to do it only if it's empty.
```
        if self.semaphore.available_permits() == 0 {
            self.semaphore.add_permits(1);
        }
```
The issue was that if we didn't reach throttle limit ever, that is `is_ready()` would be always true, then we could have accumulated a lot of permits. And once the reader becomes not ready anymore, we could be spinning in `pool` method until all permits are gone. This will permit not accumulating permits.

pmnoxx:
> LGTM overall, although I am still irked a tiny bit by the incosistency between ThrottleController type name and `rate-limiter` crate name, and the fact that we define ActixMessageWrapper in this crate.

I'm open to name changes of thiscreate. Though, I'm not sure what the right one would be.

PR #4931, opened 2021-10-04
feat: add tar files and sandbox to macos release

willemneal:
@chefsale thanks for your help! Here is the same for macos.


chefsale (approved):
Amazing stuff, thanks a lot for landing these. <3

PR #4938, opened 2021-10-06
fix: correct feature handling

matklad:
This is based on syntax-level understanding of the code, so please
tripple-check. It seems that the old code flipped the order of
if-branches: if the feature is not active at runtime, we should be doing
the same thing as if the features is not enabled at compile time.

Fix the issue by not repeating the code twice and leveraging the
feature-checking macro instead.


bowenwang1996 (approved):
Thanks!

PR #4939, opened 2021-10-06
fix: correct #4938

matklad:
#4938 we noticed the buggy behavior *if* the feature is enabled at compile time and changed that to be *always* buggy. Let's fix it to be always correct instead :)  

Side note: I am a bit unnerved that this isn't covered by tests.

### DIFF
--- a/chain/chain/src/chain.rs
+++ b/chain/chain/src/chain.rs
@@ -391,11 +391,11 @@ impl Chain {
             BlockHeaderV3,
             protocol_version,
             {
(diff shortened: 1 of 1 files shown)


matklad:
`into_v1` exists, it just does nothing: https://github.com/near/nearcore/blob/3113b02ad05304b12d66c420a6c0b5847f98d26a/core/primitives/src/types.rs#L724-L728

matklad:
Ok, given that we don't actually understand what's going on here, let's just revert 

https://github.com/near/nearcore/pull/4949

PR #4940, opened 2021-10-06
Stream dumping of state

nikurt:
Streaming genesis records to a separate file makes the dumping process work in O(1) memory, regardless of the size of the state.


Longarithm (approved):
LGTM, thank you!

Issue #4778, opened 2021-09-01
Re-evaluate EVM host functions using new param estimator


nearmax:
@matklad is working on new param estimator and he has an intermediate result. It is much more intuitive and diagnosable. We should re-evaluate EVM host functions using this param estimator and see if the fees are getting reduced.

joshuajbouw:
As per discussion with @matklad a week ago (I should've updated this, thats on me), he ran into an issue and will let me know when this is ready for us to re-run the param estimator.

matklad:
The costs we use today are:

```
Ripemd160Base                                     284_558_362
Ripemd160Block                                    226_702_528
EcrecoverBase                               1_121_789_875_000
```

My understanding is that they were estimating using [this code](https://github.com/joshuajbouw/nearcore/blob/57e0ca7333cd6a0e2114f2dbfb2a8a7986fcbaa1/runtime/runtime-params-estimator/src/math_ops_cost.rs) (the numbers line up with the ones from #4548).

The costs I get locally with the new estimator using our standard approach to estimation (via  [wasm contract](https://github.com/matklad/nearcore/blob/13faa28d4fce661ec674ab8cb458fc7a711c52fe/runtime/runtime-params-estimator/test-contract/src/lib.rs#L437-L483)) are:

```
Ripemd160Base                                     690_977_200
Ripemd160Block                                    309_200_758
EcrecoverBase                                  92_940_662_819
```

Command to reproduce this numbers locally:

```console
# switch to in-progress estimator branch 
$ git clone git@github.com:matklad/nearcore.git && cd nearcore
$ git switch --detach 13faa28d4fce661ec674ab8cb458fc7a711c52fe
$ cargo run --release -p runtime-params-estimator --features required -- --docker --v2 --full \
  --metrics-to-measure Ripemd160Base,Ripemd160Block,EcrecoverBase
```

These "new estimator" costs roughly match the  "old estimator" costs from #4548. 

The ecrecover cost is substantially lower in my estimates (10x) than what we currently use. This is explained in https://github.com/near/nearcore/issues/4548#issuecomment-888323690, the new cost is correct, the old cost is too big due to erroneous benchmark. So, action item for aurora team: lower EcrecoverBase cost to ~93*10**9 (per multiplier). See https://github.com/near/nearcore/pull/4795 for cost lowering process (this requires a protocol upgrade).

[...]

Issue #4482, opened 2021-07-08
Investigate unrealistic base cost of `DataReceiptCreationConfig`


Longarithm:
In recent `runtime-params-estimator` launches, `DataReceiptCreationConfig` base cost is 50 TGas, which is unrealistically high comparing to current cost of 4.5 TGas.

Examples:
https://github.com/near/nearcore/pull/4231
https://github.com/near/nearcore/pull/4455

```
data_receipt_creation_config: DataReceiptCreationConfig {
  base_cost: Fee {
      send_sir: 50745691155500,
      send_not_sir: 50745691155500,
      execution: 50745691155500,
  }
  ...
}
```

cc @bowenwang1996 @olonho

bowenwang1996:
To be clear, the current cost of 4.5Tgas is also unrealistically high. See https://github.com/near/nearcore/pull/3279 for a discussion on this.

Longarithm:
From the first sight, difference is explained by taking IO costs into account:
https://github.com/near/nearcore/pull/3771#issuecomment-750358010 - 50 TGas
https://github.com/near/nearcore/pull/3771#issuecomment-750357731 - 300 Ggas
Local computation in docker confirm this.

But there is conflicting info here:
https://github.com/near/nearcore/pull/3771#issuecomment-751226986 - 40 Ggas **in both cases**. 
Need to investigate further

cc https://github.com/near/nearcore/issues/4474 @matklad

Longarithm:
Data receipt fee estimation depends on its place in the vector of metrics: https://github.com/near/nearcore/blob/master/runtime/runtime-params-estimator/src/cases.rs#L536

Consider an example:
```
    let v = calls_helper! {
        data_receipt_base_10b_1000_1 => data_receipt_base_10b_1000,
        data_receipt_10b_1000_1 => data_receipt_10b_1000,
        data_receipt_100kib_1000_1 => data_receipt_100kib_1000
        cpu_ram_soak_test => cpu_ram_soak_test,
        base_1M => base_1M,
        read_memory_10b_10k => read_memory_10b_10k,
        ...
        data_receipt_base_10b_1000_2 => data_receipt_base_10b_1000,
        data_receipt_10b_1000_2 => data_receipt_10b_1000,
        data_receipt_100kib_1000_2 => data_receipt_100kib_1000
    };
```
[...]

PR #4948, opened 2021-10-07
feat: stabilize features lowering costs

Longarithm:
Stabilize features lowering costs for new release:
* https://github.com/near/nearcore/pull/4795
* https://github.com/near/nearcore/pull/4865

Quality control:

* We run param estimator several times and got consistent results: 
  * beginning of Sep 2021, my GCP instance https://hackmd.io/w6ODyKjUReuuofXTuqdyFQ
  * end of Sep 2021, @matklad instance https://github.com/near/nearcore/issues/4778#issuecomment-925187102

* The current fee values are explained:
  * Data receipt costs - investigated here https://github.com/near/nearcore/issues/4482, the reason was relatively explained and the observed issue was fixed. Note that we don't know the exact root cause, but we assume that it is related to a separate fee (touching_trie_node) for taking store size into account. This fee is problematic because it assumes the constant height of a trie, but we treat it as a separate problem.
  * Ecrecover cost - follow links from this one: https://github.com/near/nearcore/issues/4778#issuecomment-925187102

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,19 @@
 * Upgrade from Wasmer 0 to Wasmer 2, bringing better performance and reliability. 
 * Lower data receipt cost and base cost of `ecrecover` host function
 
+## `1.21.0` [09-06-2021]
+
+### Protocol Changes
+
+* Fix some receipts that were stuck previously due to #4228. [#4248](https://github.com/near/nearcore/pull/4248)
+
+### Non-protocol Changes
+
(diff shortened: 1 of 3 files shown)


matklad:
Might be good idea to explain what "exactly" is being stabilized. My rule-of-thumb for stabilizations is that a person who doesn't have a relevant context in their head should to a reasonable degree understand what's being proposed. I *know* what is stabilized, but that's just becaues I was closely following the issue. 

In this specific case, I'd probably list the three costs which are changed, and their values before/after the change.

Issue #2348, opened 2020-03-31
Introduce business model - database layer


maxzaver:
Currently we have the following issues:

*   The lifetime of the business objects is managed by whatever entity they live in. E.g. block, state node, etc have their lifetime managed by high-level components of the system. Some business objects have a co-dependent lifetime, e.g. state and state changes are supposed to be computed simultaneously. We don't have a central place that enforces these assumptions on the dependencies of the lifetimes, instead this logic is scattered around our codebase, e.g. GC makes sure when it deletes blocks, states, and state changes that they are removed consistently; similarly runtime makes sure there is not state without state changes and NightshadeRuntime makes sure there is no state without the block. However, because this logic is scattered across our entire codebase we make mistakes. ***We already have an issue***, currently due to state sync it is possible for state to exist without the corresponding state changes and it is not clear how RPC server will behave if requested these state changes.
*   There is complex mapping between columns in kv-database and business objects, i.e. a single type of business object needs to use multiple columns, e.g. `ColBlock`, `ColBlockHeight`, `ColBlockMisc`. Also we many columns are used to created DB indices, like `ColNextBlockWithNewChunk`, `ColReceiptIdToShardId`, etc. Currently we have 38 columns and this mapping is not specified anywhere which makes it really hard to maintain;
*   In many places in our codebase we clone business objects to avoid complexities with Rust lifetimes if we try avoiding cloning then we might end-up with resources shared across threads, i.e. Mutexes which are notoriously hard to maintain.

The proposed solution is to have an explicit layer that every component in our system would use to interface with the storage. The advantages:

*   The main goal of the layer would be making sure the objects with co-dependent lifetimes are maintained correctly;
*   It can be used to keep mapping of columns to business objects in one place, this is especially useful for columns that are used for indices;
*   It will own all non-temporary business objects and the rest of our codebase will refer to them through some form of linking (e.g. `Arc` or `Weak`. Most of the objects living in ORM will be immutable);
*   It can use arena-like allocators to significantly speed up our codebase (we have experimented with arena allocators in Rust in the past for TxFlow, and the difference is very noticeable).

For the record, this was already proposed one year ago: [https://github.com/nearprotocol/nearcore/issues/575#issuecomment-463841872](https://github.com/nearprotocol/nearcore/issues/575#issuecomment-463841872) .

1.  1.  1.  Patterns to consider

[...]

PR #4952, opened 2021-10-07
fix(test): fix a python test failure caused by storage validation

bowenwang1996:
We changed how `ColChunkExtra` is keyed but did not update it in one place in store validator and caused some python tests to fail as a result. This PR fixes the issue.

Test plan
----------
http://nayduck.near.org/#/run/2108


bowenwang1996:
> I am afraid that these kind of issues will happen again. I think the problem is that the `get_ser` and `set_ser` functions are not type safe. We should definitely think of reimplementing them to avoid these kind of issues happening again.

Yeah it is a known issue https://github.com/near/nearcore/issues/2348 :(

PR #4953, opened 2021-10-08
Move SimpleNightshade from nightly to stable

mzhangmzz:
Move SimpleNIghtshade from nightly to stable.

This feature enables the transition from one shard to four shards. For more details, see https://near.org/blog/near-launches-simple-nightshade-the-first-step-towards-a-sharded-blockchain/ and https://github.com/near/NEPs/pull/241

### DIFF
--- a/core/primitives/src/version.rs
+++ b/core/primitives/src/version.rs
@@ -167,8 +167,6 @@ impl ProtocolFeature {
             ProtocolFeature::ChunkOnlyProducers => 115,
             #[cfg(feature = "protocol_feature_routing_exchange_algorithm")]
             ProtocolFeature::RoutingExchangeAlgorithm => 117,
-            #[cfg(feature = "protocol_feature_simple_nightshade")]
-            ProtocolFeature::SimpleNightshade => 122,
         }
     }
 }


matklad:
Couple of process-level suggestions for stabilization PRs which I think are good ideas, but by no means are established practice today:

* Add entry to the https://github.com/near/nearcore/blob/master/CHANGELOG.md#unreleased. We gotta write changelog someday anyway, and stabilization PR seems like the most effective place for that. 
* In the PR description, write a stabilization report, which describes what is stabilized and why, without assuming prior knowledge of context. Stabilization is the only step in the lifitime of the feature that "really matters", so we can use this point to double check that we are agreeing on what actually is being stabilized.

matklad:
> remove log

In general, just removing a large file in a subsequent commit is less than ideal. If you add and remove a large file it will still be in the git history, and will still increase repo's size. So in such cases it's better to rewrite history/rebase to not let the file into the history at all.

We, however, automatically squash all commits when merging PRs, so that will be done automatically.

mzhangmzz:
@bowenwang1996 could you take a look at the CI failure here? It has something to do with `protocol_feature_chunk_only_producers` and I don't know enough context there to know how to fix this

mzhangmzz:
@bowenwang1996 thanks for the fix, we can merge this PR once betanet's protocol version is updated to 122. It hasn't yet because node1 has majority of the stake and its binary is not updated

mzhangmzz:
> @mzhangmzz I partially reverted the change you made in #4990 since it actually causes near-api-js tests to fail because a local node cannot be started if the protocol version is the same as simple nightshade version. This will not cause betanet to break once the PR is merged since the protocol version of simple nightshade at that point will be much smaller than the protocol version of betanet itself.

Actually I'm not sure if that will work. That depends on whether state dump (or whatever process I'm actually not sure, but the hard fork process) knows to dump the correct shard layout after the state upgrade to the new genesis file. I can check that tomorrow.

bowenwang1996:
> That depends on whether state dump (or whatever process I'm actually not sure, but the hard fork process) knows to dump the correct shard layout after the state upgrade to the new genesis file.

Ah yes that is a good point. Another related problem is that the hard fork will result again in a genesis config with one shard even after the upgrade. We should change state-viewer to dump with the current protocol config to fix this issue.

Issue #4826, opened 2021-09-15
Improve function call fee estimation


Longarithm:
The current estimation function call fee must be improved, because it doesn't take into account the contract deserialization and thus complexity of the contract. This already leads to significant undercharging of huge contracts like Aurora.

Intention of the fee is to charge users for all operations not accounted by wasm gas counter, so estimation is made by counting instructions spent on no-op function call.

## Simple approach
 
We can't charge constant fee, because gas spent varies on contract' complexity. But the algorithm of parsing raw wasm code should be linear, so the fee should be bounded by `C * code.len()`.

The problem here is that when there are too many functions in contract, fee becomes very inaccurate, see https://github.com/near/nearcore/issues/4660. We absolutely can't afford to charge 302 Tgas for each Aurora engine call, because 1) it is ~10 times overestimation 2) we could fit only 3 such tx in a block.

## New approach

Let's make number of functions a separate factor to make estimation more accurate.
Take it as `func_assoc` here: https://github.com/near/wasmer/blob/0.18.1/lib/runtime-core/src/module.rs#L48
Experiment show that the effect shows up on all functions, not only exported ones.

So, from the above the following constants should exist:

```
gas_spent <= base_cost + cost_per_code_byte * code.len() + cost_per_function * module_info.func_assoc 
``` 

Then we estimate these coefficients by LSM here using function `least_squares_method_2`: https://github.com/near/nearcore/compare/master...2c1e4fe

Current results are: 
```
base_cost ~= 10 Ggas
cost_per_code_byte ~= 311 Mgas per KB
cost_per_function ~= 2 Ggas
```

It gives upper estimation on a simple test set if we multply coefficients by 2. Aurora fee should be ~4 Tgas after this.

## Plan

- Check that `args.len()` is an independent factor
- Decide if we need safety multiplier
- Validate results on real contracts
- Check this effect on other items included to wasm modules:
- - globals
- - tables / elems
- - memory
- - exports / imports?

 
## Other ideas

- Set upper limit on methods number - worsens devx, can be an unnecessary bottleneck for developers
- https://gov.near.org/t/boosting-top-1000-contracts-through-fee-reduction/4132/4 - makes fees very unstable
- avoid multi-linear LSM by estimating coefficients independently. E.g. compare contracts with same lengths but different number of functions

Longarithm:
Validated `cost_per_function` as follows:

[...]

PR #4954, opened 2021-10-08
feat: limit contract functions number

Longarithm:
Limit max contract functions number to ~2000~ 10000.
See corresponding Zulip thread for more details.

Test plan
---------
- [x] add new tests on config and vm side
- [x] check that this doesn't break existing contracts

Notes:
* This is not a final solution. We are going to work on better validation mechanism and fees estimation.
* Link to the Initial discussion https://github.com/near/nearcore/issues/4826

### DIFF
--- a/chain/client/src/view_client.rs
+++ b/chain/client/src/view_client.rs
@@ -1117,13 +1117,29 @@ impl Handler<NetworkViewClientMessages> for ViewClientActor {
                         }
                         Err(err) => {
                             error!(target: "view_client", "Cannot retrieve num shards: {}", err);
-                            NetworkViewClientResponses::NoResponse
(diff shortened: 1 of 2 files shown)


Longarithm:
There was a problem with reusing `make_simple_contract_call_with_gas_vm`.
Tests like `test_stack_overflow` check exact number of gas spent, so it seems better to use test runtime configs there, to avoid recomputing numbers each time.
And new test `test_limit_contract_functions_number` need real vm configs to check enabling of functions number limit.
So now we have `make_simple_contract_call_with_protocol_version_vm`, `make_simple_contract_call_with_gas_vm`, `make_cached_contract_call_vm` and I think they can be united later.

PR #4957, opened 2021-10-08
fix: stable protocol version should be updated only once

Longarithm:
Decrement stable protocol version, because they must not be skipped for upgrade process: https://github.com/near/nearcore/pull/4948#discussion_r725022966

Test plan
---------
Existing tests.


bowenwang1996:
> @bowenwang1996 can we merge it without db_migration test passed? I guess it's problematic only because we decrease protocol version, but if we merge this PR shortly, it shouldn't really affect anything.

Yeah I'll do it manually

Issue #4577, opened 2021-07-26
Add a linter and CI check for pytests folder


chefsale:
We currently use yapf and pylint in nearup which is a good reference (https://github.com/near/nearup/blob/master/tox.ini). The root directory in nearcore has the yapf style file (https://github.com/near/nearcore/blob/master/.style.yapf) but we do not enforce it, but should in the CI. 

cc: @mina86 @mhalambek @bowenwang1996

chefsale:
This will also take a lot of work to get that part of the codebase up to standards and make the CI eventually pass.

PR #4958, opened 2021-10-08
Format all Python scripts with YAPF

mina86:
Run YAPF on the `scripts` and `pytest` directories (i.e. `python3 -m yapf
-pir scripts pytest) to format all the Python files there with a consistent
style.  Furthermore, since there’s already a `.style.yapf` file at the root
of the repository, remove the one in `pytest` directory.

Issue: https://github.com/near/nearcore/issues/4577


matklad:
>Just looking at repositories one can ses that YAPF was released in 2015 while Black started in 2018. 

Yeah, I didn't mean to imply any historical or casual relationship, rather the difference in context: supporting Google's Python code style vs supporting PEP8 code style.

PR #4960, opened 2021-10-08
Add CI for checking if Python code is properly formatted

mina86:
Fixes: https://github.com/near/nearcore/issues/4577


mina86:
@chefsale, @mhalambek, what’s the deal with BuildKite? Are the agents run on our infrastructure we can change or is this not that easy to affect? Specifically, I need `yapf` installed on the agent and it seems faster in the long run if we just get that installed or BuildKite agents rather than adding `python3 -m pip install yapf` to the pipeline.

PR #4971, opened 2021-10-09
Template for feature stabilization PR

bowenwang1996:
Add a PR template for stabilizing protocol features as suggested by @matklad. However, I am not sure whether I can even test this without merging to master first...


matklad:
Hm, seems like to make it work you'd have to use special URL for creating a PR: https://stackoverflow.com/a/62613198/1936422, https://github.com/near/nearcore/pull/4979#issuecomment-940229737

PR #4934, opened 2021-10-05
feat: stabilize wasmer2 protocol feature

matklad:

PR #4979, opened 2021-10-11
feat: stabilize lower_regular_op_cost feature

matklad:
# Feature to stabilize

This PR stabilizes new, lower regular_op_cost. It is lowered from `3_856_371` to `2_207_874`. regular_op_cost is roughly the cost of primitive wasm operation, and is a major cost factor for computation-intensive contracts. The underlying change which allowed us to lower the cost is the switch from `wasmer0` to `wasmer2` implementation, which improved real wasmer perfomance across the board

# Testing and QA

Cost measurement was reproduced by several people using both old and new estimator. The cost measurement itself is rather simple (probably the simplest of all our costs), so the relative risk here is low. Additionally, the reduction "makes sense" in terms of various ad-hoc comparisons of wall-clock times I did when comparing wasm runtimes. 

Note that we are stabilizing this feature on a very short notice. While this in general is risky, I think it's ok for this particular case, as it is just cost reduction, and there's strong evidence that new wasmer is indeed faster. In a sense, *most* risk here is actually carried by #4934 , but that was on nightly for a significantly longed time. 

# Checklist
- [ ] No nightly test regression
- [x] Update CHANGELOG.md to include this protocol feature in the `Unreleased` section.

### DIFF
--- a/chain/chain/src/chain.rs
+++ b/chain/chain/src/chain.rs
@@ -2723,7 +2723,6 @@ pub struct DifferentHeightResult {
 pub struct SplitStateResult {
     // parent shard of the split states
     shard_uid: ShardUId,
-    is_new_chunk: bool,
     results: Vec<ApplySplitStateResult>,
 }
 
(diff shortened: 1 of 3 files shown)


matklad:
@bowenwang1996 bad news: I didn't see PR template suggested anywhere, and I had to copy-paste it manually. Good news -- the  idea of the template works, going to fill the changelog now as I've forgot to:)

PR #4981, opened 2021-10-11
Add test for missing chunks and fix a bug related to missing chunks

mzhangmzz:
This PR adds integration tests for sharding upgrade where there might be missing chunks in the block. The test also uncovers a bug in our implementation regarding missing chunks. If the last block before the sharding switch does not have a chunk, the gas limit of the chunk extra will be set to 0, which causes the new blocks in the new epoch also have gas limit zero, which is incorrect


mzhangmzz:
@bowenwang1996 this PR contains both the test and the fix. If you think it's better, I can also separate them into two PRs so we can commit the fix quickly

PR #4983, opened 2021-10-12
Add integration test for sharding upgrade when some chunks are missing

mzhangmzz:
This PR adds integration test for sharding upgrade when some chunks are missing

### DIFF
--- a/chain/client/src/client.rs
+++ b/chain/client/src/client.rs
@@ -964,9 +964,10 @@ impl Client {
     }
 
     /// Gets called when block got accepted.
-    /// Only produce chunk if `produce_chunk` is true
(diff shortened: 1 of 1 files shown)


mzhangmzz:
Test is failing because feature simple nightshade is not compatible with ChunkOnlyPoducers, in another PR, to enabled simple nightshade in betanet, I had to change the SimpleNightshade version to be larger than ChunkOnlyProducers, which causes the test to fail. To fix this, I need to bump nightly version again and move ChunkOnlyProducers to be after SimpleNightshade

mzhangmzz:
Or we can just merge this PR after the feature is stabilized

PR #4986, opened 2021-10-12
feat(state-viewer): show first outcome difference in replay

ailisp:
It show the first replay difference in outcome and exit (metadata not count since old metadata is always v1 but replay return v2). 

Test Plan
------------
Follow https://github.com/near/nearcore/blob/master/test-utils/state-viewer/README.md and use this branch to replay mainnet history


bowenwang1996:
@nikurt please review this PR. Thanks!

PR #4997, opened 2021-10-13
fix(runtime): new ext functions should only available after its protocol version

### LINKED ISSUE #4713: Newly introduced extension functions are available regardless of the protocol version
Today, nothing in https://github.com/near/nearcore/blob/master/runtime/near-vm-runner/src/imports.rs mentions when a particular host function became part of the protocol. This is wong -- that means that new functions like `ecrecover` become available *at all protocol version*. So, depending on the version of the neard binary, the contract that calls `ecrecover` at the old protocol version can either fail with linking error, or run successfully. 

It seems that, when stabilizing extension functions, we need to preserve protocol features they correspond to. So, in https://github.com/near/nearcore/pull/4443/files#diff-e9b653ec008c4aeece3871e171113ef743d50a1d8229bddf14d1ea4c51391675L224, we probably wanted to say something like 

```rust
#["stable", MathExtension] ecrecover<[hash_len: u64, hash_ptr: u64, sign_len: u64, sig_ptr: u64, v: u64, malleability_flag: u64, register_id: u64] -> [u64]>,
```

rather then removing `#[]` part altogether. 

Note that this won't just work -- we need to change the import.rs infra a bit to allow for that.

ailisp:
Fix #4713. Basically there're three kinds of import:
1) unconditionally import, for old imports before we introduce protocol version
2) conditional import after specific protocol version and certain nightly protocol feature
3) conditional import after specific protocol version, on stable node

1) and 2) are properly handled. 3) was not. To handle 3) but still keep 1) and 2) correct, we need to guard ext function impl under given features and also `checked_feature!` to ensure current protocol version is greater. We cannot drop feature guards in `imports.rs` because in 2) ext function implementation is guard under feature. We cannot keep feature guard to do 3) because "stable" isn't a real feature. Therefore Introduce a special `stable` feature in near-vm-runner is the most clean way to fix 3) while ensure correctness of 1) and 2).

Test Plan
------------
CI and
```
cargo test --package near-vm-logic --test test_miscs --features costs_counting -- test_ripemd160 --exact --nocapture
```


matklad:
:thinking: @bowenwang1996 could you take a quick look here to make sure the overall logic is correct? 

This retroactively gates math extension functions by a protocol version (with the main value is that this adds infra for gating future extension as well). 

So we have the following outcomes for tx that calls ecrecover at various protocol versions and neard versions

| neard version \ protocol version| 45 | 46+
| --- | --- | ---|
| old neard | err | N/A
| current neard| ok (but wrong) | ok
|this PR| err | ok

I think this is correct to rollout that change (presumably, we don't have anything in the "wrong ok" cell).

I guess, let's actually hold this off until release is branched just in case?

ailisp:
@matklad Sounds a better approach! ptal

PR #4998, opened 2021-10-13
docs: clarify feature stabilization checklist

matklad:


matklad:
Note: we didn't do nayduck runs for wasmer2 rollout and cost reductions. I assume we'll have a pre-release nayduck run?

PR #5000, opened 2021-10-13
up time limit for block production Closes #4982

### LINKED ISSUE #4982: Investigate fuzzing failure on nayduck
Fuzzing almost always fails on nayduck http://nayduck.near.org/#/test_history/231155 and we should look into the failures.

posvyatokum:
Time limit nayduck fuzzing failures are not reproducible, it takes third of the time to run these scenarios on gcp. Thus I would like to up time limit for block production from 1s to 3s.


mina86:
The test is still failing though http://nayduck.near.org/#/test/232969
And at this point I’m wondering if running it on NayDuck the way it’s done now is a good solution. The test now takes 6h to finish (it actually times out with current nightly configuration).

bowenwang1996:
@posvyatokum please take a look. Also should we limit how long it runs on nayduck? For example 30 minutes per nightly run?

PR #5007, opened 2021-10-14
fix(test): prevent a race condition on neard start

### LINKED ISSUE #4928: Investigate rust expensive test failure
In this run http://nayduck.near.org/#/run/2102, many rust expensive tests failed and when I try to run those failed tests locally, they pass without issues.

bowenwang1996:
There is a potential race condition between when the peer manager actor starts and when the client actor starts. The issue is that if a peer establishes connection before client is initialized, `fetch_client_chain_info` won't get anything and as a result peers cannot be properly initialized and this caused some tests to fail. It is still not clear to me what change in nearcore caused this as it was not observed before, but this PR fixes the issue by always return *some* `ChainInfo` in `ViewClient`.

Fixes #4928 

Test plan
----------
http://nayduck.near.org/#/test/232592 @mina86 those scp failures :(


bowenwang1996:
@chefsale @frol @khorolets please review the PR. Thanks!

bowenwang1996:
> @bowenwang1996 I feel it is not a proper solution since there are other conditions that could still lead to `NoResponse`, and what do we do then? Should we always return something then? Wouldn't it better to retry after some delay on the receiving side?

Good point. I basically considered the other case impossible since the commit operation should be atomic, but you are right that we should not ignore it. Indeed a better solution may be refactoring on the requesting side, but I looked at that code and decided it would require a lot more effort to do and we really need to fix these tests :(

PR #5025, opened 2021-10-18
fix nearcore get_protocol_config to account for changing shard layout

mzhangmzz:
This fixes `get_protocol_config` to account for changing shard layout too. Indexer relies on this function to get current number of shards, so indexer will break if it returns wrong number of shards


frol:
@mzhangmzz This PR is the evidence of the fragility of the system since changing it in one place did not force you to change anything else in the system, engineers should KNOW that they should also update this place rather than being forced by the system to do that. I don't have a proposal on how to improve this particular area around protocol config, but I wanted to raise your awareness that next time someone else will forget to update it as well (and who knows, maybe we already forgot to update something there).

PR #5039, opened 2021-10-19
clean up NibbleSlice

mzhangmzz:


mzhangmzz:
@matklad this is the error I am getting 

```
error[E0597]: `encoded` does not live long enough
  --> core/store/src/trie/nibble_slice.rs:83:28
   |
70 | impl<'a> NibbleSlice<'a> {
   |      -- lifetime `'a` defined here
...
83 |         Self::from_encoded(&encoded).0
   |         -------------------^^^^^^^^-
   |         |                  |
   |         |                  borrowed value does not live long enough
   |         argument requires that `encoded` is borrowed for `'a`
84 |     }
   |     - `encoded` dropped here while still borrowed
```

I'm not sure if there is any other way to solve that except for making `data` in `NibbleSlice` not a reference

PR #5040, opened 2021-10-19
fix gc for sharding upgrade

mzhangmzz:
resolve https://github.com/near/nearcore/issues/4710

Also added a python test for gc with sharding upgrade


mzhangmzz:
> The test is a bit confusing to me. It does not seem to test that anything is actually garbage collected. We should probably check that after the state split is done, data for new shards can also be garbage collected

Good point, I just copied the code from an existing test `garbage_collection.py` and didn't think too much. I'll add that

PR #5042, opened 2021-10-19
chore(jsonrpc): updated rpc_errors_schema

miraclx:
Update `rpc_errors_schema.json` after the introduction of `TooManyFunctions` in https://github.com/near/nearcore/pull/4954


matklad:
>Why that error variant is not behind a feature-flag then?

The code that produces error is gated, so gating the error variant itself isn't strictly necessary. Doing that is possible, it's just requires a bunch of extra cfgs everywhere. I and @Longarithm looked at that and figured that there's little benefit in that. No strong opinion on my side here though, could add a feature-gate there.

miraclx:
Superseded by https://github.com/near/nearcore/pull/5052

PR #5044, opened 2021-10-20
chore(lockfile): build all targets in workspace with all features

miraclx:
Fixes `cargo check --workspace --all-targets --all-features` which fails on master.

Identifying the root cause of this made no sense, but removing the lockfile and re-initializing it appropriately fixes the issue. A side effect of this is; a number of dependencies have been updated. We've already cut a release, so we can observe until the next release if this breaks anything and resolve those issues.


miraclx:
🤦🏽 removed the `target` folder and now it works.. 😪 ..thanks for checking

PR #5046, opened 2021-10-20
fix a bug in produce_block regarding shard_id

mzhangmzz:
This PR fixes bug in `produce_block`. This bug will only be triggered during sharding upgrade at the first block of the new epoch that uses the new shard layout, if the new block has empty chunks.

I checked that this bug was not triggered in the testnet launch because that block does not have empty chunks (#68081211)


mzhangmzz:
Actually this is not the correct way to fix this bug, since if we modify `shard_id`, the chunk signature won't work anymore.

PR #5071, opened 2021-10-22
chore: rebuild QEMU on top of current Rust docker image 

matklad:


matklad:
@jakmeier did this in a better way in #6213

PR #5083, opened 2021-10-26
fix: bring costs.txt inline with the actual protocol costs

matklad:
forgot to do this in https://github.com/near/nearcore/pull/4979

This doesn't affect any logic anywhere, just makes it easier to figure
out what our measured costs are.


mzhangmzz:
sorry I just saw this. @matklad that test was flaky but I fixed it. It should work if you merge with the latest master

matklad:
Hopelessly obsolete, but let me tag @jakmeier here :) 

The idea was that we need an easy reference for what the costs are today.

PR #5087, opened 2021-10-26
feat: add param-estimtor CLI for comparing cost files

matklad:
Example output:

    $ cargo run --release -p runtime-params-estimator -- --costs-file rust-1.55.0.txt --compare-to rust-1.56.0.txt
    Cost                                                    First                    Second  Second/First
    WasmInstruction                                       820_758                   704_775          0.86
    HostFunctionCall                                   68_433_428                60_104_942          0.88
    ActionReceiptCreation                         240_033_972_875           236_978_285_750          0.99

Note that CLI interface is not as good as it can be.
I aim for the minimal diff, and overhaul the overall CLI once #4733 is merged

Test Plan
=======

This is a helper CLI util, so just manual verification


jakmeier (approved):

runtime/runtime-params-estimator/src/main.rs
```diff
@@ -137,11 +140,17 @@ fn main() -> anyhow::Result<()> {
...
+        println!("{}", baseline.diff(&compare_to));
+        return Ok(());
+    }
+
     if let Some(path) = cli_args.costs_file {
-        let cost_table = fs::read_to_string(&path)
-            .ok()
-            .and_then(|it| it.parse::<CostTable>().ok())
-            .with_context(|| format!("Failed to parse {}", path.display()))?;
+        let cost_table = read_costs_table(&path);
```
jakmeier, line 153: I think this Result needs to be unpacked
```suggestion
        let cost_table = read_costs_table(&path)?;
```

PR #5091, opened 2021-10-27
scripts: hide authorisation code when entering it to nayduck.py

mina86:
The NayDuck authorisation code is essentially a password so hide it when
user is entering it.


mina86:
@chefsale, @mhalambek, friendly ping.

chefsale (approved):

PR #5093, opened 2021-10-28
ci: don’t print echo commands

mina86:
The pipeline commands are executed with -x flag active which means
that all commands are printed before being executed.  While this is
helpful in majority of cases (letting the user see what commands are
being run in the CI) it only adds noise when echo commands are
executed.  For this reason, disable -x flag before a series of echo
commands explaining a failure.


mina86:
@chefsale, @mhalambek, friendly ping.

PR #5097, opened 2021-10-28
Replace chrono::now() with Instant::now()

### LINKED ISSUE #5096: Stop using non monotous chrono::DateTime
As @matklad noted, we are using non-monotonic clock ` chrono::Utc::now()`. Let's fix that.

pmnoxx:
We shouldn't be using non monotonous time with chrono::Now()

Closes #5096 

Moving changes out of https://github.com/near/nearcore/pull/5089

### DIFF
--- a/chain/network/src/cache.rs
+++ b/chain/network/src/cache.rs
@@ -1,12 +1,11 @@
 use std::collections::btree_map;
 use std::collections::{BTreeMap, BTreeSet, HashMap};
-
-use chrono::Utc;
+use std::ops::Sub;
+use std::time::{Duration, Instant};
 
 use near_primitives::hash::CryptoHash;
 use near_primitives::network::PeerId;
 
-type Time = u64;
 type Size = u64;
 
 /// Cache to store route back messages.
@@ -48,12 +47,12 @@ pub struct RouteBackCache {
     /// Maximum number of records allowed in the cache.
     capacity: u64,
     /// Maximum time allowed before removing a record from the cache.
-    evict_timeout: u64,
+    evict_timeout: Duration,
     /// Minimum number of records to delete from offending peer when the cache is full.
     remove_frequent_min_size: u64,
     /// Main map from message hash to time where it was created + target peer
     /// Size: O(capacity)
-    main: HashMap<CryptoHash, (Time, PeerId)>,
+    main: HashMap<CryptoHash, (Instant, PeerId)>,
     /// Number of records allocated by each PeerId.
     /// The size is stored with negative sign, to order in PeerId in decreasing order.
     /// To avoid handling with negative number all sizes are added by capacity.
@@ -62,7 +61,7 @@ pub struct RouteBackCache {
     /// List of all hashes associated with each PeerId. Hashes within each PeerId
     /// are sorted by the time they arrived from older to newer.
     /// Size: O(capacity)
-    record_per_target: BTreeMap<PeerId, BTreeSet<(Time, CryptoHash)>>,
+    record_per_target: BTreeMap<PeerId, BTreeSet<(Instant, CryptoHash)>>,
 }
 
 impl RouteBackCache {
@@ -117,8 +116,8 @@ impl RouteBackCache {
         if self.is_full() {
             self.remove_frequent();
 
-            let now = Utc::now().timestamp_millis() as Time;
-            let remove_until = now.saturating_sub(self.evict_timeout);
+            let now = Instant::now();
+            let remove_until = now.sub(self.evict_timeout);
 
             let mut remove_empty = vec![];
 
@@ -153,7 +152,7 @@ impl RouteBackCache {
         }
     }
 
-    pub fn new(capacity: u64, evict_timeout: u64, remove_frequent_min_size: u64) -> Self {
+    pub fn new(capacity: u64, evict_timeout: Duration, remove_frequent_min_size: u64) -> Self {
         assert!(capacity > 0);
 
         Self {
@@ -209,7 +208,7 @@ impl RouteBackCache {
 
         self.remove_evicted();
 
-        let now = Utc::now().timestamp_millis() as Time;
+        let now = Instant::now();
 
         self.main.insert(hash, (now, target.clone()));
 
@@ -265,7 +264,7 @@ mod test {
 
     #[test]
     fn simple() {
-        let mut cache = RouteBackCache::new(100, 1000000000, 1);
+        let mut cache = RouteBackCache::new(100, Duration::from_millis(1000000000), 1);
         let (peer0, hash0) = create_message(0);
 
         check_consistency(&cache);
@@ -281,7 +280,7 @@ mod test {
     /// Check record is removed after some timeout.
     #[test]
     fn evicted() {
-        let mut cache = RouteBackCache::new(1, 1, 1);
+        let mut cache = RouteBackCache::new(1, Duration::from_millis(1), 1);
         let (peer0, hash0) = create_message(0);
 
         cache.insert(hash0, peer0.clone());
@@ -296,7 +295,7 @@ mod test {
     /// Check element is removed after timeout triggered by insert at max capacity.
     #[test]
     fn insert_evicted() {
-        let mut cache = RouteBackCache::new(1, 1, 1);
+        let mut cache = RouteBackCache::new(1, Duration::from_millis(1), 1);
         let (peer0, hash0) = create_message(0);
         let (peer1, hash1) = create_message(1);
 
@@ -313,7 +312,7 @@ mod test {
     /// Check element is removed after insert because cache is at max capacity.
     #[test]
     fn insert_override() {
-        let mut cache = RouteBackCache::new(1, 1000000000, 1);
+        let mut cache = RouteBackCache::new(1, Duration::from_millis(1000000000), 1);
         let (peer0, hash0) = create_message(0);
         let (peer1, hash1) = create_message(1);
 
@@ -331,7 +330,7 @@ mod test {
     /// Check that old element from peer0 is removed, even while peer1 has more elements.
     #[test]
     fn prefer_evict() {
-        let mut cache = RouteBackCache::new(3, 1000, 1);
+        let mut cache = RouteBackCache::new(3, Duration::from_millis(100), 1);
         let (peer0, hash0) = create_message(0);
         let (peer1, hash1) = create_message(1);
         let (_, hash2) = create_message(2);
@@ -354,7 +353,7 @@ mod test {
     /// Check that older element from peer1 is removed, since evict timeout haven't passed yet.
     #[test]
     fn prefer_full() {
-        let mut cache = RouteBackCache::new(3, 100000, 1);
+        let mut cache = RouteBackCache::new(3, Duration::from_millis(100000), 1);
         let (peer0, hash0) = create_message(0);
         let (peer1, hash1) = create_message(1);
         let (_, hash2) = create_message(2);
@@ -377,7 +376,7 @@ mod test {
     /// Check that older element from peer1 is removed, since evict timeout haven't passed yet.
     #[test]
     fn remove_all_frequent() {
-        let mut cache = RouteBackCache::new(3, 100000, 2);
+        let mut cache = RouteBackCache::new(3, Duration::from_millis(100000), 2);
         let (peer0, hash0) = create_message(0);
         let (peer1, hash1) = create_message(1);
         let (_, hash2) = create_message(2);
@@ -403,7 +402,7 @@ mod test {
     /// initial hashes should be present in the cache after the attack.
     #[test]
     fn poison_attack() {
-        let mut cache = RouteBackCache::new(17, 1000000, 1);
+        let mut cache = RouteBackCache::new(17, Duration::from_millis(1000000), 1);
         let mut ix = 0;
 
         let mut peers = vec![];
--- a/chain/network/src/cache.rs
+++ b/chain/network/src/cache.rs
@@ -1,6 +1,5 @@
 use std::collections::btree_map;
 use std::collections::{BTreeMap, BTreeSet, HashMap};
-use std::ops::Sub;
 use std::time::{Duration, Instant};
 
 use near_primitives::hash::CryptoHash;
@@ -117,7 +116,7 @@ impl RouteBackCache {
             self.remove_frequent();
 
             let now = Instant::now();
-            let remove_until = now.sub(self.evict_timeout);
+            let remove_until = now - self.evict_timeout;
 
             let mut remove_empty = vec![];
 
--- a/chain/network/src/peer_manager.rs
+++ b/chain/network/src/peer_manager.rs
@@ -62,13 +62,13 @@ use crate::types::{GetPeerId, GetPeerIdResult, SetAdvOptions};
 use crate::types::{RoutingState, RoutingSyncV2, RoutingVersion2};
 
 /// How often to request peers from active peers.
-const REQUEST_PEERS_SECS: u64 = 60;
+const REQUEST_PEERS_SECS: Duration = Duration::from_millis(60_000);
 /// How much time to wait (in milliseconds) after we send update nonce request before disconnecting.
 /// This number should be large to handle pair of nodes with high latency.
-const WAIT_ON_TRY_UPDATE_NONCE: u64 = 6_000;
+const WAIT_ON_TRY_UPDATE_NONCE: Duration = Duration::from_millis(6_000);
 /// If we see an edge between us and other peer, but this peer is not a current connection, wait this
 /// timeout and in case it didn't become an active peer, broadcast edge removal update.
-const WAIT_PEER_BEFORE_REMOVE: u64 = 6_000;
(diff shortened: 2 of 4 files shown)


pmnoxx:
> https://passcod.name/technical/no-time-for-chrono.html also feels relevant.
> 
> Can we go as far (in subsequent PRs) as removing `chrono` from `Cargo.toml` altogether? That is, as a system, do we care about calendar time at all? If not, I feel we should try to expunge chrono from Cargo.tomls, and, ideally, from Cargo.lock as well. Not caring about system time is nice!

Yes, it'a a good idea to get rid of chrono. I checked we always use `DateTime` as `DateTime<Utc>`. So, the time is actually monotonic. Though, the code with `Instant` is cleaner.

pmnoxx:
If you look at the documentation of `Instant`: https://doc.rust-lang.org/src/std/time.rs.html#241
It's mentioned that in practice instant is not-monotonic, and that can cause panics.

Maybe, we should switch back to `chrono:Datatime<Utc>`?

matklad:
Yeah, that's a good point. 

To clarify, the doc say that the underlying OS functionality doesn't always live up to its promise, and so, for that reason, on certain platforms Rust's stdlib tries to paper over this in user-space.

The problem with that strategy though is that, with virtualization, you don't actually know which platform you are running on top. Eg, x86 Linux implements monotonic time correctly, but if it is run in a VM in windows host, the windows can break things. This was observed in https://github.com/tokio-rs/tokio/issues/3619 and is being fixed in https://github.com/rust-lang/rust/pull/89926. Technically, that is a bug in Linux kernel and/or hypervisor implementations, but this indeed is observed in practice .

I don't think using `chrono::DateTime` is a solution here -- that is decidedly non-monotonic (as it changes when you change system time, or when ntpd makes adjustments), and is generally a wrong kind of time. We shouldn't be caring about timezones, etc. If we want to work-around it, I suggest to avoid using `Instance::now/elapsed`, and instead define our own wrappers which force the thing to be monotonic

```rust
pub fn now() -> Instance {
    // furious hand-waving
    static LAST: Mutex<Instance> = ...
    let now = Instance::now();
    let now = now.max(*LAST.lock());
    *LAST.lock() = now;
    now
}
```

Practically though, it seems that this problem shouldn't be something that we care about. That is, the amount of setups where time goes backwards despite stdlib's mitigations should be petty small, and, if time is unreliable in your setup, you should fix the setup, as we do rely on time significantly. 

@bowenwang1996 do you remember more details about https://near.zulipchat.com/#narrow/stream/295302-dev-general/topic/general? Was it some production environment, or just some weird ad-hoc setup?

pmnoxx:
> Yeah, that's a good point.
> 
> To clarify, the doc say that the underlying OS functionality doesn't always live up to its promise, and so, for that reason, on certain platforms Rust's stdlib tries to paper over this in user-space.
> 
> The problem with that strategy though is that, with virtualization, you don't actually know which platform you are running on top. Eg, x86 Linux implements monotonic time correctly, but if it is run in a VM in windows host, the windows can break things. This was observed in [tokio-rs/tokio#3619](https://github.com/tokio-rs/tokio/issues/3619) and is being fixed in [rust-lang/rust#89926](https://github.com/rust-lang/rust/pull/89926). Technically, that is a bug in Linux kernel and/or hypervisor implementations, but this indeed is observed in practice .
> 
> I don't think using `chrono::DateTime` is a solution here -- that is decidedly non-monotonic (as it changes when you change system time, or when ntpd makes adjustments), and is generally a wrong kind of time. We shouldn't be caring about timezones, etc. If we want to work-around it, I suggest to avoid using `Instance::now/elapsed`, and instead define our own wrappers which force the thing to be monotonic
> 
> ```rust
> pub fn now() -> Instance {
>     // furious hand-waving
>     static LAST: Mutex<Instance> = ...
>     let now = Instance::now();
>     let now = now.max(*LAST.lock());
>     *LAST.lock() = now;
>     now
> }
> ```
> 
> Practically though, it seems that this problem shouldn't be something that we care about. That is, the amount of setups where time goes backwards despite stdlib's mitigations should be petty small, and, if time is unreliable in your setup, you should fix the setup, as we do rely on time significantly.
> 
> @bowenwang1996 do you remember more details about https://near.zulipchat.com/#narrow/stream/295302-dev-general/topic/general? Was it some production environment, or just some weird ad-hoc setup?

@matklad The main reason we use `DateTime` is because it has nice functions to convert time to unix timestamp, and back from it. I don't see any such function inside the same crate as `Instant`. 

For some reason `UNIX_EPOCH` is available as `SystemTime`, but not as `Instant`.  It would be easy to write down such conversion function if we had some way to convert `UNIX_EPOCH` to `Instant`, or do something equivalent like initializing `Instant` with value of `UNIX_EPOCH`.

Any ideas?

`
#[stable(feature = "time2", since = "1.8.0")]
pub const UNIX_EPOCH: SystemTime = SystemTime(time::UNIX_EPOCH);
`

pmnoxx:
> Yeah, that's a good point.
> 
> To clarify, the doc say that the underlying OS functionality doesn't always live up to its promise, and so, for that reason, on certain platforms Rust's stdlib tries to paper over this in user-space.
> 
> The problem with that strategy though is that, with virtualization, you don't actually know which platform you are running on top. Eg, x86 Linux implements monotonic time correctly, but if it is run in a VM in windows host, the windows can break things. This was observed in [tokio-rs/tokio#3619](https://github.com/tokio-rs/tokio/issues/3619) and is being fixed in [rust-lang/rust#89926](https://github.com/rust-lang/rust/pull/89926). Technically, that is a bug in Linux kernel and/or hypervisor implementations, but this indeed is observed in practice .
> 
> I don't think using `chrono::DateTime` is a solution here -- that is decidedly non-monotonic (as it changes when you change system time, or when ntpd makes adjustments), and is generally a wrong kind of time. We shouldn't be caring about timezones, etc. If we want to work-around it, I suggest to avoid using `Instance::now/elapsed`, and instead define our own wrappers which force the thing to be monotonic
> 
> ```rust
> pub fn now() -> Instance {
>     // furious hand-waving
>     static LAST: Mutex<Instance> = ...
>     let now = Instance::now();
>     let now = now.max(*LAST.lock());
>     *LAST.lock() = now;
>     now
> }
> ```
> 
> Practically though, it seems that this problem shouldn't be something that we care about. That is, the amount of setups where time goes backwards despite stdlib's mitigations should be petty small, and, if time is unreliable in your setup, you should fix the setup, as we do rely on time significantly.
> 
> @bowenwang1996 do you remember more details about https://near.zulipchat.com/#narrow/stream/295302-dev-general/topic/general? Was it some production environment, or just some weird ad-hoc setup?

Here is a part of documentation from `Instant`:
```
/// https://doc.rust-lang.org/src/std/time.rs.html#60

/// Instants are opaque types that can only be compared to one another. There is
/// no method to get "the number of seconds" from an instant. Instead, it only
/// allows measuring the duration between two instants (or comparing two
/// instants).
```
It looks like you are meant to use instant just to compare a relative time, which is completely fine, but we need to be able to convert time, for example to unix timestamp and back. That's not what `Instant` is meant to do.

I think we can't simply replace `Datetime` with `Instant` for that reason. I'm not suggesting to keep `DateTime` either, but maybe there is some other way to represent time that doesn't have drawback of either of those. One valid solution would be do define our own type that wraps around the UNIX timestamp, or find some other library.

@matklad What do you think?

bowenwang1996:
> Yeah, that's a good point.
> 
> To clarify, the doc say that the underlying OS functionality doesn't always live up to its promise, and so, for that reason, on certain platforms Rust's stdlib tries to paper over this in user-space.
> 
> The problem with that strategy though is that, with virtualization, you don't actually know which platform you are running on top. Eg, x86 Linux implements monotonic time correctly, but if it is run in a VM in windows host, the windows can break things. This was observed in [tokio-rs/tokio#3619](https://github.com/tokio-rs/tokio/issues/3619) and is being fixed in [rust-lang/rust#89926](https://github.com/rust-lang/rust/pull/89926). Technically, that is a bug in Linux kernel and/or hypervisor implementations, but this indeed is observed in practice .
> 
> I don't think using `chrono::DateTime` is a solution here -- that is decidedly non-monotonic (as it changes when you change system time, or when ntpd makes adjustments), and is generally a wrong kind of time. We shouldn't be caring about timezones, etc. If we want to work-around it, I suggest to avoid using `Instance::now/elapsed`, and instead define our own wrappers which force the thing to be monotonic
> 
> ```rust
> pub fn now() -> Instance {
>     // furious hand-waving
>     static LAST: Mutex<Instance> = ...
>     let now = Instance::now();
>     let now = now.max(*LAST.lock());
>     *LAST.lock() = now;
>     now
> }
> ```
> 
> Practically though, it seems that this problem shouldn't be something that we care about. That is, the amount of setups where time goes backwards despite stdlib's mitigations should be petty small, and, if time is unreliable in your setup, you should fix the setup, as we do rely on time significantly.
> 
> @bowenwang1996 do you remember more details about https://near.zulipchat.com/#narrow/stream/295302-dev-general/topic/general? Was it some production environment, or just some weird ad-hoc setup?

Are you sure the link is correct? It does not link to an issue related to time

matklad:
correct link: https://near.zulipchat.com/#narrow/stream/295302-dev-general/topic/general/near/248247028

matklad:
>For some reason UNIX_EPOCH is available as SystemTime, but not as Instant. 

Yeah, that's what I am getting at: `Instant` doesn't have have this functionality in a fundamental way, and that is a feature. Basically, there are three kinds of time in computers:

* Number of seconds since hardware timer started. This is monotonic time / `std::time::Instant`. It is reliably tracked by hardware, allows for measuring time intervals within the process precisely, but can't tell anything about time in the absolute terms. Hardware timer can't know how much time elapsed in total since the unix epoch. This time never goes backwards.
* Number of seconds since unix epoch. This is absolute time /  `std::time::SystemTime`. This can't be tracked purely by the hardware (as the clock would drift). The way it works is there's a system-level state which stores "current time the last time it was adjusted", and the hardware adds a relative time to that. The config can be changed by a human, or by other computers. Usually, when the computer boots, and then periodically, it sends network requests to time servers asking for time, and then trusts this info. This kind of time is significantly more complicated than the previous one -- the time can go backwards after upgrade, or the computer can be tricked into thinking wrong about time, if time servers are compromised. One extra bit of complexity is that this time, most of the time, is roughly the same on different computers, but this isn't reliable: there's always at least some amount of small drifts, and occasionally big diff due to misconfigurations. 
* Calendar date/wall clock time in the room where the computer is situated. This is `chrono::Instant<TimeZone>`. This time depends on the "time since epoch", plus the current physical location of the computer (to get the time zone), plus the time zone database (as the meaning of timezone itself changes over time, when, eg, government decides that daylight savings time is not a great idea, and then changes its mind back couple of years later). This time has all the complexity of `SystemTime` plus a lot of edge-cases related to human's notion of timezone plus fun things like leap seconds. 

I feel relatively strongly that we need to stick to the simplest model of time we can, that is, to `Instant` or, if we must, to `SystemTime`. 


>but we need to be able to convert time, for example to unix timestamp and back

Yeah, if we need that, we do need SystemTime in those parts that use such functionality. Could you point where we need this, though? I want to understand whether we fundamentally need to know absolute time, or whether this is something which can be replaced by `Instant`? Ie, here:

https://github.com/near/nearcore/blob/af64bfbf9860535ceb70299448d0434185550835/core/primitives/src/block_header.rs#L29-L31

we definitely want timestamp (so, `SystemTime`), because we want to coordinate this value across different machines. Though, I'd say we should be using `SystemTime`, and not `chrono`, as the source of time there.

But for, eg

https://github.com/near/nearcore/blob/af64bfbf9860535ceb70299448d0434185550835/chain/network-primitives/src/types.rs#L657-L659

it seems that we want to use `Instant`, as that is machine-local state.

matklad:
Oh, while looking at it, I found a case where we very well may panic:

https://github.com/near/nearcore/blob/af64bfbf9860535ceb70299448d0434185550835/chain/network/src/peer_store.rs#L241-L245

Here, the `?` will return an `Error` if the difference is negative. But `Utc::now()` can return time in the past, and that *can* happen during the normal operation, is system time is adjusted just before the call. Practically, I think systems try relatively hard to make system time monotonic, but that's best effort, and can definitely break in practice if you, eg, manually adjust the clock. 

If this code returns an error, we will panic at the call-site, where we unwrap the thing.

pmnoxx:
@matklad Some data structures that we store on disk have timestamps

```

#[derive(Debug, Clone)]
pub struct KnownPeerState {
    pub peer_info: PeerInfo,
    pub status: KnownPeerStatus,
    pub first_seen: u64,
    pub last_seen: u64,
}

    pub fn peer_connected(
        &mut self,
        peer_info: &PeerInfo,
    ) -> Result<(), Box<dyn std::error::Error>> {
        self.add_trusted_peer(peer_info.clone(), TrustLevel::Signed)?;
        let entry = self.peer_states.get_mut(&peer_info.id).unwrap();
        entry.last_seen = to_timestamp(Utc::now());
        entry.status = KnownPeerStatus::Connected;
        let mut store_update = self.store.store_update();
        store_update.set_ser(ColPeers, &peer_info.id.try_to_vec()?, entry)?;
        store_update.commit().map_err(|err| err.into())
    }
```

pmnoxx:
> > For some reason UNIX_EPOCH is available as SystemTime, but not as Instant.
> 
> Yeah, that's what I am getting at: `Instant` doesn't have have this functionality in a fundamental way, and that is a feature. Basically, there are three kinds of time in computers:
> 
> * Number of seconds since hardware timer started. This is monotonic time / `std::time::Instant`. It is reliably tracked by hardware, allows for measuring time intervals within the process precisely, but can't tell anything about time in the absolute terms. Hardware timer can't know how much time elapsed in total since the unix epoch. This time never goes backwards.
> * Number of seconds since unix epoch. This is absolute time /  `std::time::SystemTime`. This can't be tracked purely by the hardware (as the clock would drift). The way it works is there's a system-level state which stores "current time the last time it was adjusted", and the hardware adds a relative time to that. The config can be changed by a human, or by other computers. Usually, when the computer boots, and then periodically, it sends network requests to time servers asking for time, and then trusts this info. This kind of time is significantly more complicated than the previous one -- the time can go backwards after upgrade, or the computer can be tricked into thinking wrong about time, if time servers are compromised. One extra bit of complexity is that this time, most of the time, is roughly the same on different computers, but this isn't reliable: there's always at least some amount of small drifts, and occasionally big diff due to misconfigurations.
> * Calendar date/wall clock time in the room where the computer is situated. This is `chrono::Instant<TimeZone>`. This time depends on the "time since epoch", plus the current physical location of the computer (to get the time zone), plus the time zone database (as the meaning of timezone itself changes over time, when, eg, government decides that daylight savings time is not a great idea, and then changes its mind back couple of years later). This time has all the complexity of `SystemTime` plus a lot of edge-cases related to human's notion of timezone plus fun things like leap seconds.
> 
> I feel relatively strongly that we need to stick to the simplest model of time we can, that is, to `Instant` or, if we must, to `SystemTime`.
> 
> > but we need to be able to convert time, for example to unix timestamp and back
> 
> Yeah, if we need that, we do need SystemTime in those parts that use such functionality. Could you point where we need this, though? I want to understand whether we fundamentally need to know absolute time, or whether this is something which can be replaced by `Instant`? Ie, here:
> 
> https://github.com/near/nearcore/blob/af64bfbf9860535ceb70299448d0434185550835/core/primitives/src/block_header.rs#L29-L31
> 
> we definitely want timestamp (so, `SystemTime`), because we want to coordinate this value across different machines. Though, I'd say we should be using `SystemTime`, and not `chrono`, as the source of time there.

This struct is part of our protocol. We need to serialize it and and it over the network. `SystemTime` doesn't implement `BorshSerialize, BorshDeserialize, Serialize`; therefore it can't be used.
```

#[derive(BorshSerialize, BorshDeserialize, Serialize, Debug, Clone, Eq, PartialEq)]
pub struct BlockHeaderInnerLite {
...
    /// Timestamp at which the block was built (number of non-leap-nanoseconds since January 1, 1970 0:00:00 UTC).
    pub timestamp: u64,
...
}```

Although I agree, we shouldn't have timestamp as part of our protocol.

matklad:
>Some data structures that we store in DB have timestamps

Doh, I suspected something like this, and specifically looked for it, and somehow still missed `#[BorsSerialize]` :facepalm: Thanks for pointing this out. 

Yeah, so this is the case where we need `SystemTime`. We still don't need chrono here, as we don't care about timezones. 

>SystemTime doesn't implement BorshSerialize, BorshDeserialize, Serialize; therefore it can't be used.

SystemTime does implement `Serialize`:

https://docs.rs/serde/1.0.130/src/serde/ser/impls.rs.html#611-625

 It can implement BorshSerialize as well, although that would need to use `u128`, rather than `u64`, as Durations can hold more than u64 nanos, and panicking/truncating would be wrong for the borsh-style library. It seems that we are just using `u64` 


So my current understanding is:

* for in-process parts which don't involve serialization/IPC we should prefer monotonic time `Instant`.
* for parts which are stored, send to other processes or serialized in other way, we should prefer `SystemTime`, but be careful with monotonicity assumptions
* we don't need timezones, so we don't need chrono.

PR #5101, opened 2021-10-29
Mocknet loadtest using the skyward smartcontract

nikurt:
Mocknet scripts now support running two different tests:
* Using skyward sale contracts
* Using a simple contract for adding and deleting state
Tested by running on mocknet networks of different sizes.


pytest/lib/mocknet.py
```diff
@@ -524,146 +579,169 @@ def create_genesis_file(validator_node_names,
...
+    accounts_and_balances = [(TREASURY_ACCOUNT, TREASURY_BALANCE),
+                             (MASTER_ACCOUNT, master_balance),
+                             (SKYWARD_ACCOUNT, SKYWARD_CONTRACT_BALANCE),
+                             (TOKEN1_ACCOUNT, TOKEN1_BALANCE),
+                             (TOKEN2_ACCOUNT, TOKEN2_BALANCE),
+                             (TOKEN2_OWNER_ACCOUNT, TOKEN2_OWNER_BALANCE),
+                             (ACCOUNT1_ACCOUNT, ACCOUNT1_BALANCE)]
+    accounts = {}
+    for a in accounts_and_balances:
+        accounts[a[0]] = a[1]
```
mina86, line 595:
```
accounts = {
    TREASURY_ACCOUNT: TREASURY_BALANCE,
    MASTER_ACCOUNT: master_balance,
    ...
}
```

There’s really no need for the intermediate `accounts_and_balances`.

pytest/lib/mocknet.py
```diff
@@ -524,146 +579,169 @@ def create_genesis_file(validator_node_names,
...
+    seen_accounts = set()
+    for record in genesis_config['records']:
+        if 'Account' in record and 'account_id' in record['Account'] and record[
+                'Account']['account_id'] in accounts:
+            seen_accounts.add(record['Account']['account_id'])
+            if 'account' in record['Account'] and 'amount' in record['Account'][
+                    'account']:
```
mina86, line 602: So if `amount` is specified we want to override it? But if it’s not specified we want to leave it unspecified? Shouldn’t it be `not in`?

And what if `account` is not set? Should we set it instead of skipping the `account` for given `Account`?

pytest/lib/mocknet.py
```diff
@@ -524,146 +579,169 @@ def create_genesis_file(validator_node_names,
...
+                record['Account']['account']['amount'] = str(
+                    accounts[record['Account']['account_id']])
+
+    for (account_id, balance) in [(TREASURY_ACCOUNT, TREASURY_BALANCE),
+                                  (MASTER_ACCOUNT, master_balance),
+                                  (SKYWARD_ACCOUNT, SKYWARD_CONTRACT_BALANCE),
+                                  (TOKEN1_ACCOUNT, TOKEN1_BALANCE),
+                                  (TOKEN2_ACCOUNT, TOKEN2_BALANCE),
+                                  (TOKEN2_OWNER_ACCOUNT, TOKEN2_OWNER_BALANCE),
+                                  (ACCOUNT1_ACCOUNT, ACCOUNT1_BALANCE)]:
```
mina86, line 612:
```
for account_id, balance in accounts.items():
```

pytest/lib/mocknet.py
```diff
@@ -524,146 +579,169 @@ def create_genesis_file(validator_node_names,
...
         })
 
+    total_supply = 0
+    for record in genesis_config['records']:
+        if 'Account' in record:
+            if 'account' in record['Account']:
+                if 'locked' in record['Account']['account']:
+                    total_supply += int(record['Account']['account']['locked'])
+                if 'amount' in record['Account']['account']:
+                    total_supply += int(record['Account']['account']['amount'])
```
mina86, line 730: Again, `dict.get` is your friend:

```
        acc = record.get('Account', {}).get('account', {})
        total_supply += int(acc.get('locked', 0))
        total_supply += int(acc.get('amount, 0))
```

pytest/lib/mocknet.py
```diff
@@ -679,19 +757,34 @@ def get_node_addr(node, port, tmp_dir):
...
 
-    mocknet_config_filename = os.path.join(tmp_dir, "config.json")
+
+def get_node_keys(node, tmp_dir):
+    logger.info(f'get_node_keys from {node.instance_name}')
+    node_key_filename = os.path.join(tmp_dir,
+                                     f'node_key.{node.instance_name}.json')
+    node.machine.download('/home/ubuntu/.near/node_key.json', node_key_filename)
+    with open(node_key_filename, 'r') as f:
+        node_key_json = json.load(f)
```
mina86, line 767: We seem to be doing this a lot. Download JSON file from the machine and then load it. Perhaps this should be a separate function? If that function would use `tempfile.NamedTeporaryFile` with `delete=True` than this one wouldn’t even need `tmp_dir` simplifying the call site.

pytest/tests/mocknet/helpers/load_testing_add_and_delete_helper.py
```diff
@@ -0,0 +1,360 @@
...
+    r = requests.post(f'http://{LOCAL_ADDR}:{RPC_PORT}', json=j, timeout=10)
+    return r.json()
+
+
+def get_nonce_for_pk(account_id, pk, finality='optimistic'):
+    access_keys = json_rpc(
+        'query', {
+            "request_type": "view_access_key_list",
+            "account_id": account_id,
+            "finality": finality
```
mina86, line 46: Single quotes here as well then?

pytest/tests/mocknet/helpers/load_testing_add_and_delete_helper.py
```diff
@@ -0,0 +1,360 @@
...
+    for (account, _) in accounts_and_indices:
+        all_tx_events += account.tx_timestamps
+    all_tx_events.sort()
+    with open(filename, 'w') as output:
+        for t in all_tx_events:
+            output.write(f'{t}\n')
+
+
+def get_skyward_account():
+    global special_accounts
```
mina86, line 129: No need for `global` here and below. You need `global` only if you bind the global variable to a different value.

pytest/tests/mocknet/load_test_spoon.py
```diff
@@ -131,21 +139,30 @@ def check_slow_blocks(initial_metrics, final_metrics):
     logger.info(f'initial_validator_accounts: {initial_validator_accounts}')
     test_passed = True
 
+    script, deploy_time, test_timeout = (None, None, None)
+    if args.script == 'skyward':
+        script, deploy_time, test_timeout = 'load_testing_add_and_delete_helper.py', load_testing_add_and_delete_helper.CONTRACT_DEPLOY_TIME, load_testing_add_and_delete_helper.TEST_TIMEOUT
```
mina86, line 144: Nah, just split it (and the assignment below) into three separate lines. Easier to read that way.

mina86 (commented):
I haven’t looked too closely at the logic of the test but overall looks reasonable.

pytest/lib/mocknet.py
```diff
@@ -722,12 +691,10 @@ def create_genesis_file(validator_node_names,
...
-        if 'Account' in record:
-            if 'account' in record['Account']:
-                if 'locked' in record['Account']['account']:
-                    total_supply += int(record['Account']['account']['locked'])
-                if 'amount' in record['Account']['account']:
-                    total_supply += int(record['Account']['account']['amount'])
+        total_supply += int(
+            record.get('Account', {}).get('account', {}).get('locked', str(0)))
+        total_supply += int(
+            record.get('Account', {}).get('account', {}).get('amount', str(0)))
```
mina86, line 697: This would read better with a `account = record.get('Account', {}).get('account', {})` temporary.

mina86 (approved):

pytest/lib/mocknet.py
```diff
@@ -745,27 +712,23 @@ def create_genesis_file(validator_node_names,
...
-    node_key_filename = os.path.join(tmp_dir,
-                                     f'node_key.{node.instance_name}.json')
-    node.machine.download('/home/ubuntu/.near/node_key.json', node_key_filename)
-    with open(node_key_filename, 'r') as f:
-        node_key_json = json.load(f)
-    return f'{node_key_json["public_key"]}@{node.ip}:{port}'
+def download_and_read_json(node, filename):
+    tmp_file = tempfile.NamedTemporaryFile(mode='r+')
+    node.machine.download(filename, tmp_file.name)
+    return json.load(tmp_file)
```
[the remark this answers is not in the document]
nikurt: I've just tried it and the file was empty.
Reverted to the old way of downloading and re-opening.

PR #5108, opened 2021-11-02
nit(runtime): drop serde Serialize and Deserialize for VMOutcome

ailisp:
VMOutcome is not serialized in rpc, only used in standalone as print result, which can be replaced by pretty print. Therefore, drop serialize/deserialize for VMOutcome

Test Plan
------------
CI


ailisp:
Looks like master "cargo check nearcore library" is fail on building openssl-sys v0.9.70

matklad:
Uhu, filed https://github.com/near/nearcore/pull/5111 for that.

matklad (approved):

bowenwang1996 (approved):

Issue #2834, opened 2020-06-11
[Enhancement] Additional RPC and Prometheus Data Points


htafolla:
The current challenge on Stakewars is to setup monitoring with email alerts and as I was working through it I noticed a couple of enhancements that would improve things. The idea would be to monitor and alert from a single dashboard or service.

Additional Prometheus Data points - http://localhost:3030/metrics 
1. Blocks Expected
2. Blocks Produced
3. Node Version and Build
4. Seat Price
5. Is Validator
6. Is Kicked
7. Has Enough Stake

Additional Remote RPC Data points - These would enable the ability for true remote monitoring
1. Node Version and Build
2. Peers

gaia:
Version/Build is already available via `echo -n "Remote " && curl -s https://rpc.testnet.near.org/status | jq -r .version.build && echo -n "Local " && curl -s http://localhost:3030/status | jq -r .version.build`? What else would you like to see?

bowenwang1996:
> Node Version and Build

This is provided through `/status` endpoint.

> Peers

You can use `/network_info` for this

janewang:
In addition to `num_expected_blocks` and `num_produced_blocks`, consider add a graph of delta of those two, so num_expected_blocks - num_produced_blocks. Closing https://github.com/near/nearcore/issues/3580 in favor of this story, as this is the superset.

nikurt:
@htafolla Thank you for posting these suggestions. As most of them are currently available, let me close the issue. If you have a concrete use case for one of the metrics please let me know by re-opening the issue.


## Metrics

> Blocks Expected

Available as metrics `near_validators_blocks_expected` and `near_validators_chunks_expected`

> Blocks Produced

Available as metrics `near_validators_blocks_produced` and `near_validators_chunks_produced`

> Node Version and Build

Available, as mentioned above by @gaia and @bowenwang1996 

> Seat Price

What's your use case for making this a metric? This is provided by `near-cli` as `near validators current`.

> Is Validator

Available as `near_is_validator`

> Is Kicked

Seems that "Is Validator" should be sufficient in most cases.

> Has Enough Stake

Seems that "Is Validator" should be sufficient in most cases.


## Additional Remote RPC Data points

> Node Version and Build

Available as mentioned above by @gaia and @bowenwang1996 

> Peers

Available as mentioned above by @gaia and @bowenwang1996

Issue #3580, opened 2020-11-07
num_expected_blocks and num_produced_blocks in /metrics


kucharskim:
Please add `num_expected_blocks` and `num_produced_blocks` to `/metrics` endpoint, so it can be graphed easily out of the box. Maybe also another graph of delta of those two, so `num_expected_blocks - num_produced_blocks`.

janewang:
Closing this story in favor of https://github.com/near/nearcore/issues/2834.

PR #5113, opened 2021-11-02
feat(metrics): Add a near_block_expected_total prometheus metric

marcelo-gonzalez:
This requires a bit of refactoring in order to increment the metrics in the same place where this info is already accounted (in the EpochInfoAggregator). We add a `me: &Option<AccountId>` parameter to runtime.add_validator_proposals() and rename that function to make this parameter less confusing.

Test plan
------------
Any of the long-lived pytests (like contracts/gibberish.py for exaple) can be used to check that the metric seems to be working.

Issue: https://github.com/near/nearcore/issues/2834


marcelo-gonzalez:
trie::split_state::tests::test_get_trie_items_for_part fails for some reason :(. trying to figure out why...

bowenwang1996 (changes requested):
Is this metric supposed to measure the block produced by the node itself? Even if that is the case we can read it in client without having to pass `me` to epoch manager. Also I am a bit skeptical about the metric itself -- it is something that we can read directly from the validator rpc output. What is the point of adding this metric to the node itself?

marcelo-gonzalez:
@bowenwang1996
> Is this metric supposed to measure the block produced by the node itself?

Yeah actually exposing it for all validators including a metric label that gives the validator ID (so you could filter/group by that) would make things much easier. Then this whole PR could look basically like the last commit (Move near_block_produced_total to chain/epoch_manager) where we just increment the metrics in EpochInfoAggregator without having to pass `me` everywhere. Reason I didn't do that for this first proposal is that the existing near_block_produced_total is exposed only for the current node. Maybe it would be ok to just add a new version of that metric (maybe near_blocks_produced_total or near_block_produced_total2) so we could do that.

> Also I am a bit skeptical about the metric itself -- it is something that we can read directly from the validator rpc output. What is the point of adding this metric to the node itself?

I figured https://github.com/near/nearcore/issues/3580 asked for this to be exposed in /metrics so it can be looked at/plotted over time with Prometheus/Grafana without having to do anything special. Then stuff like https://github.com/near/near-ops/issues/260 could be implemented more easily probably. For now I'll just close this since either way we'll want to do it differently than in this PR

Issue #5105, opened 2021-11-01
Don't rely on default features when validating wasm


matklad:
Wasm is a living spec and new features are introduced from time to time. For us, it's important to know exactly which features are introduced when. Today, we just rely on defaults from upstream libraries:

* here, we pass `None` as the config: 
  https://github.com/near/nearcore/blob/1cb995fa77664da5fb22c1ad6b0b81a4d8deecca/runtime/near-vm-runner/src/prepare.rs#L18
* and here, we are not calling `.features`, which again make us rely on defaults:
  https://github.com/near/nearcore/blob/1cb995fa77664da5fb22c1ad6b0b81a4d8deecca/runtime/near-vm-runner/src/wasmer2_runner.rs#L321

We should change both call-sites to explicitly list the features we use. 

I believe this is the config we happen to use today:

```rust
    OperatorValidatorConfig {
        enable_threads: false,
        enable_reference_types: false,
        enable_simd: false,
        enable_bulk_memory: false,
        enable_multi_value: false,
 };
```

Ie, we happen to correctly not to enable any features. 

While we are at it, we should add explicit tests to make sure that attempts to use such features break.

nagisa:
So, after some investigation, the configurations that we do use are definitely not consistent. First of all, both `near-vm-runner` and `wasmer` use `wasmparser`. However the versions of `wasmparser` used are different (`near-vm-runner` uses `0.51` and `wasmer` utilizes the `0.78` release instead.

I'm not sure quite yet what's the reason behind the double verification (maybe we could remove one of those?) but for now it is pretty apparent to me that the defaults are different between the two validators.

The `wasmparser::validate(original_code, None)` does indeed validate with the following configuration, 

```rust
OperatorValidatorConfig {
        enable_threads: false,
        enable_reference_types: false,
        enable_simd: false,
        enable_bulk_memory: false,
        enable_multi_value: false,
        #[cfg(feature = "deterministic")]
        deterministic_only: true,
}
```

And then wasmer validates with the following configuration:

```rust
WasmFeatures {
    bulk_memory: true,
    threads: false,
    reference_types: true,
    multi_value: false, // This is, I believe, required to be false for the singlepass codegen
    simd: true,
    tail_call: false,
    module_linking: false,
    multi_memory: false,
    memory64: false,
    exceptions: false,
    deterministic_only: false,
}
```

[...]

PR #5115, opened 2021-11-02
fix: memory leak in wasmer 2

matklad:
Pulls in https://github.com/near/wasmer/pull/69

Test Plan
---------

* check that wasmer own tests unrelated to reference types/module linking pass
* check that we don't use refernce/types module linking https://github.com/near/nearcore/issues/5105
* fuzz test wasmer with patch vs wasmer without the patch overnight
* confirm using a soak test that the leak goes away

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5716,9 +5716,9 @@ dependencies = [
 
 [[package]]
 name = "wasmer-compiler-near"
-version = "2.0.1"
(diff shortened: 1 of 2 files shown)


olonho (approved):

bowenwang1996:
We don't need this one. Will cherry pick to the branch

PR #5116, opened 2021-11-02
fix test_get_trie_items_for_part to be less flaky

mzhangmzz:


marcelo-gonzalez:
Don't really know this code well enough to review this, but can confirm this fixes the failure I ran into :D

bowenwang1996 (approved):
Does this completely fix the flakiness?

marcelo-gonzalez:
> Does this completely fix the flakiness?

What I ran into was this

[2021-11-02T14:22:11Z] thread 'trie::split_state::tests::test_get_trie_items_for_part' panicked at 'part length 1 avg length 0', core/store/src/trie/split_state.rs:451:13

So at least the total_trie_items.len() == 0 type of failure won't happen I guess

mzhangmzz:
> Does this completely fix the flakiness?

I think the test fails when the trie does not have many items in there. I changed the max_size to 100 (so the edge case will happen more often) and ran the test a few times and all of them passed. So I think the test should be quite stable after the fix.

PR #5118, opened 2021-11-02
chain: make debug panic a bit more verbose

mina86:

### DIFF
--- a/chain/network/src/peer.rs
+++ b/chain/network/src/peer.rs
@@ -371,8 +371,9 @@ impl Peer {
             self.receive_view_client_message(ctx, msg);
         } else if msg.is_client_message() {
             self.receive_client_message(ctx, msg);
-        } else {
-            debug_assert!(false);
+        } else if cfg!(debug_assertions) {
+            panic!("expected (view) client message, got: {}",
+                   msg.msg_variant());
         }
     }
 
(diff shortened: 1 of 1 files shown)


chain/network/src/peer.rs
```diff
@@ -371,8 +371,9 @@ impl Peer {
             self.receive_view_client_message(ctx, msg);
         } else if msg.is_client_message() {
             self.receive_client_message(ctx, msg);
-        } else {
-            debug_assert!(false);
+        } else if cfg!(debug_assertions) {
+            panic!("expected (view) client message, got: {}",
+                   msg.msg_variant());
         }
```
pmnoxx, line 377: Should we add 
```
else {
     debug_assert!(false);
}
```

To preserve code behavior in case `debug_assertions` is not set?

bowenwang1996 (commented):
Couldn't we do the following?
```rust
debug_assert!(false, "expected (view) client message, got: {}", msg);
```

chain/network/src/peer.rs
mina86: `debug_assert!` is a no-op if `debug_assertions` is not set.

mina86:
> Couldn't we do the following?

Done.

pmnoxx (approved):

PR #5120, opened 2021-11-03
nightly: disable currently broken tests

mina86:
Those tests are consistently failing. Disable them and add to
the tracking Issue.

Issue: https://github.com/near/nearcore/issues/5095

### DIFF
--- a/nightly/pytest-sanity.txt
+++ b/nightly/pytest-sanity.txt
@@ -63,7 +63,10 @@ pytest --timeout=360 sanity/block_sync_archival.py --features nightly_protocol,n
 pytest --timeout=240 sanity/validator_switch.py
 pytest --timeout=240 sanity/validator_switch.py --features nightly_protocol,nightly_protocol_features
 pytest --timeout=240 sanity/restaked.py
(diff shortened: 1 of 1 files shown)


bowenwang1996 (approved):
`sanity/repro_2916.py` should not fail. @mzhangmzz could you take a look at this?

mzhangmzz:
I fixed it in https://github.com/near/nearcore/pull/5154. 

I also noticed that the python version that nayduck uses is 3.9. Is that intentional? The reason why I asked is because the python version on my local machine was 3.7 and the test only failed with 3.9 not 3.7.

bowenwang1996:
> I also noticed that the python version that nayduck uses is 3.9. Is that intentional? The reason why I asked is because the python version on my local machine was 3.7 and the test only failed with 3.9 not 3.7.

Interesting question. @mina86 what do you think?

mina86:
> I also noticed that the python version that nayduck uses is 3.9. Is that intentional? The reason why I asked is because the python version on my local machine was 3.7 and the test only failed with 3.9 not 3.7.

Yes, for the most part. All of the code should run on 3.7 but it’s not unreasonable to expect that tests should pass on 3.9. My dev machine runs 3.9 and eventually people will update their Python version as well.

PR #5121, opened 2021-11-03
Simplify gas metering

olonho:
Rework the logic of gas counting to ensure that there's a single limit to be compared with.
It allows to speed up gas metering significantly and opens possibility of intrinsified gas counting.

### DIFF
--- a/runtime/near-vm-logic/src/gas_counter.rs
+++ b/runtime/near-vm-logic/src/gas_counter.rs
@@ -141,11 +141,11 @@ impl GasCounter {
             // See https://github.com/near/nearcore/issues/5148.
             // TODO: consider making this change!
             assert!(self.prepaid_gas >= self.fast_counter.burnt_gas);
-            self.promises_gas = self.prepaid_gas - self.fast_counter.burnt_gas;
+            self.promises_gas = min(self.prepaid_gas - self.fast_counter.burnt_gas, hard_burnt_limit - self.fast_counter.burnt_gas);
             HostError::GasExceeded.into()
         } else {
             self.fast_counter.burnt_gas = min(new_burnt_gas, self.max_gas_burnt);
-            self.promises_gas = 0;
+            self.promises_gas = min(new_used_gas - self.fast_counter.burnt_gas, self.prepaid_gas - self.fast_counter.burnt_gas);
             HostError::GasLimitExceeded.into()
         }
     }
--- a/runtime/near-vm-logic/src/gas_counter.rs
+++ b/runtime/near-vm-logic/src/gas_counter.rs
@@ -24,15 +24,35 @@ pub fn with_ext_cost_counter(f: impl FnOnce(&mut HashMap<ExtCosts, u64>)) {
 
 type Result<T> = ::std::result::Result<T, VMLogicError>;
 
+/// Fast gas counter with very simple structure, could be exposed to compiled code in the VM.
+#[repr(C)]
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct FastGasCounter {
+    /// The following three fields must be put next to another to make sure
+    /// generated gas counting code can use and adjust them.
+    /// We will share counter to ensure we never miss synchronization.
+    /// This could change and in such a case synchronization required between compiled WASM code
+    /// and the host code.
+
+    /// The amount of gas that was irreversibly used for contract execution.
+    burnt_gas: u64,
+    /// Hard gas limit for execution
+    gas_limit: u64,
+    /// Single WASM opcode cost
+    opcode_cost: u64,
+}
+
 /// Gas counter (a part of VMlogic)
 pub struct GasCounter {
-    /// The amount of gas that was irreversibly used for contract execution.
-    burnt_gas: Gas,
-    /// `burnt_gas` + gas that was attached to the promises.
-    used_gas: Gas,
-    /// Gas limit for execution
+    /// Shared gas counter data.
+    fast_counter: FastGasCounter,
+    /// Gas that was attached to the promises.
+    promises_gas: Gas,
+    /// Hard gas limit for execution
     max_gas_burnt: Gas,
+    /// Amount of prepaid gas, we can never burn more than prepaid amount
     prepaid_gas: Gas,
+    /// If this is a view-only call.
     is_view: bool,
     ext_costs_config: ExtCostsConfig,
     /// Where to store profile data, if needed.
@@ -49,14 +69,20 @@ impl GasCounter {
     pub fn new(
         ext_costs_config: ExtCostsConfig,
         max_gas_burnt: Gas,
+        opcode_cost: u32,
         prepaid_gas: Gas,
         is_view: bool,
     ) -> Self {
+        use std::cmp::min;
         Self {
             ext_costs_config,
-            burnt_gas: 0,
-            used_gas: 0,
-            max_gas_burnt,
+            fast_counter: FastGasCounter {
+                burnt_gas: 0,
+                gas_limit: min(max_gas_burnt, prepaid_gas),
+                opcode_cost: Gas::from(opcode_cost),
+            },
+            max_gas_burnt: max_gas_burnt,
+            promises_gas: 0,
             prepaid_gas,
             is_view,
             profile: Default::default(),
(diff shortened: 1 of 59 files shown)


runtime/near-vm-logic/src/gas_counter.rs
```diff
@@ -49,48 +69,94 @@ impl GasCounter {
...
+            self.fast_counter.burnt_gas.checked_add(burn_gas).ok_or(HostError::IntegerOverflow)?;
+        let new_used_gas =
+            new_burnt_gas.checked_add(new_promises_gas).ok_or(HostError::IntegerOverflow)?;
         if new_burnt_gas <= self.max_gas_burnt && (self.is_view || new_used_gas <= self.prepaid_gas)
         {
-            self.burnt_gas = new_burnt_gas;
-            self.used_gas = new_used_gas;
+            if promise_gas != 0 {
+                self.fast_counter.gas_limit =
+                    min(self.max_gas_burnt, self.prepaid_gas - new_promises_gas);
```
matklad, line 106: This subtraction can underflow if `self.is_view` is true. I guess all places in the logic which do `gas_limit: min(max_gas_burnt, prepaid_gas)` should special-case `is_view` case? It seems that the correct logic is 

```
gas_limit: if self.is_view { 
  // completely ignore prepaied gas limit and promises
  max_gas_burnt
} else {
    min(max_gas_burnt, prepaid_gas - promises_gas)
}
```

runtime/near-vm-runner/src/tests.rs
```diff
@@ -63,7 +63,8 @@ fn make_simple_contract_call_with_gas_vm(
     let mut fake_external = MockedExternal::new();
     let mut context = create_context(vec![]);
     context.prepaid_gas = prepaid_gas;
-    let config = VMConfig::default();
+    let mut config = VMConfig::default();
+    config.limit_config.max_gas_burnt = prepaid_gas;
```
matklad, line 67: It seems that this makes `near-vm-runner` tests less useful, as they now can only hit the cases where the two limits are identical. Can we keep the original version?

runtime/near-vm-logic/src/gas_counter.rs
```diff
@@ -49,48 +69,94 @@ impl GasCounter {
...
+            self.fast_counter.burnt_gas = new_burnt_gas;
+            self.promises_gas = new_promises_gas;
             Ok(())
         } else {
-            use std::cmp::min;
-            let res = if new_burnt_gas > self.max_gas_burnt {
+            // if max_gas_burnt == prepaid_gas we must use GasExceeded error.
+            if new_burnt_gas > self.max_gas_burnt && self.max_gas_burnt != self.prepaid_gas {
+                self.fast_counter.burnt_gas = self.max_gas_burnt;
+                self.promises_gas = 0;
```
matklad, line 115: If I comment-out this line, the tests seem to pass. That's suspicious. Can we add a test that shows that this line is needed?

runtime/near-vm-logic/src/gas_counter.rs
olonho: Yeah, makes sense.

runtime/near-vm-runner/src/tests.rs
olonho: Ok.

runtime/near-vm-logic/src/gas_counter.rs
olonho: We can add this line when making protocol change.

runtime/near-vm-logic/src/gas_counter.rs
```diff
@@ -49,48 +69,98 @@ impl GasCounter {
...
+            }
+        }
+    }
 
-            res
+    // Optimized version of above function for cases where no promises involved.
+    pub fn burn_gas(&mut self, value: Gas) -> Result<()> {
+        let new_burnt_gas =
+            self.fast_counter.burnt_gas.checked_add(value).ok_or(HostError::IntegerOverflow)?;
+        if new_burnt_gas <= self.fast_counter.gas_limit {
```
EgorKulikov, line 137: Theoretically in view mode we could have promised something close to u64::max() gas by now and with new burnt gas that could have overflown integer in deduct_gas. Not sure if we care enough about this case

EgorKulikov (approved):
Good work

matklad:
Results for me

```
Before:
WasmInstruction                                       709_273
After:
WasmInstruction                                       459_905
```

That's sort-of the motivation here -- we can measurably speed the stuff up by just writing simpler code for the common case.

runtime/near-vm-logic/src/gas_counter.rs
olonho: Don't think that (checked) overflow in view mode is critical/important.

runtime/near-vm-logic/src/gas_counter.rs
```diff
@@ -49,48 +69,98 @@ impl GasCounter {
...
-                unreachable!()
-            };
-
-            let max_burnt_gas = min(self.max_gas_burnt, self.prepaid_gas);
-            self.burnt_gas = min(new_burnt_gas, max_burnt_gas);
-            self.used_gas = min(new_used_gas, self.prepaid_gas);
+                self.fast_counter.burnt_gas = min(new_burnt_gas, self.prepaid_gas);
+                // Technically we shall do `self.promises_gas = 0;` or error paths, as in this case
+                // no promises will be kept, but that would mean protocol change.
+                // TODO: consider making this change!
```
bowenwang1996, line 125: @olonho could you create an issue for this?

runtime/near-vm-logic/src/gas_counter.rs
```diff
@@ -49,48 +69,98 @@ impl GasCounter {
...
+            } else {
+                use std::cmp::min;
+                // Now we limit burnt gas with prepaid amount.
+                self.fast_counter.burnt_gas = min(new_burnt_gas, self.prepaid_gas);
+                // Technically we shall do `self.promises_gas = 0;` or error paths, as in this case
+                // no promises will be kept.
+                // TODO: consider making this change and fix tests/protocol!
+                assert!(self.prepaid_gas >= self.fast_counter.burnt_gas);
+                self.promises_gas = self.prepaid_gas - self.fast_counter.burnt_gas;
+                Err(HostError::GasExceeded.into())
```
bowenwang1996, line 154: @olonho it seems that this part of the code is duplicated. Could we refactor it?

olonho: Done

runtime/near-vm-logic/src/gas_counter.rs
olonho: Done

runtime/near-vm-logic/src/gas_counter.rs
olonho: Anyway, fixed it.

olonho:
Latest version gives `WasmInstruction                                       442_952`.

runtime/near-vm-logic/src/gas_counter.rs
```diff
@@ -65,32 +96,65 @@ impl GasCounter {
...
+            // Technically we shall do `self.promises_gas = 0;` or error paths, as in this case
+            // no promises will be kept, but that would mean protocol change.
+            // See https://github.com/near/nearcore/issues/5148.
+            // TODO: consider making this change!
+            assert!(self.prepaid_gas >= self.fast_counter.burnt_gas);
+            self.promises_gas = self.prepaid_gas - self.fast_counter.burnt_gas;
+            HostError::GasExceeded.into()
+        } else {
+            self.fast_counter.burnt_gas = min(new_burnt_gas, self.max_gas_burnt);
+            self.promises_gas = 0;
```
matklad, line 148: Hm, if I replace `self.promises_gas = 0;` with `panic!()`, `cargo t -p near-vm-runner -p near-vm-logic` passes successfully for me. Ie, we don't hit this code path in test *at all*. This doesn't inspire confidence, let me see if I can add one more test here.

bowenwang1996:
> Results for me
> 
> ```
> Before:
> WasmInstruction                                       709_273
> After:
> WasmInstruction                                       459_905
> ```
> 
> That's sort-of the motivation here -- we can measurably speed the stuff up by just writing simpler code for the common case.

@matklad this seems to be higher than what we projected. Do we know why that is the case?

runtime/near-vm-logic/src/gas_counter.rs
olonho: Fixed, although TBH do not like the code I have to write there. IMO, promises gas shall be zero'ed out instead.

olonho:
> > Results for me
> > ```
> > Before:
> > WasmInstruction                                       709_273
> > After:
> > WasmInstruction                                       459_905
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > That's sort-of the motivation here -- we can measurably speed the stuff up by just writing simpler code for the common case.
> 
> @matklad this seems to be higher than what we projected. Do we know why that is the case?

It's only part of the optimization - second one is fully intrinsified counter on Wasmer side.

matklad (approved):
I think this now preserves behavior, so that's good to go!

I am still a little bit on the fence about adding `#[repr(C)]` bit in this PR and not in the one which actually enables wasmer change, but we can always remove that later if it proves to be unnecessary. 

Could you confirm that the this final version with all bugs fixed still gives us an improvement?

olonho:
Sure, we haven't changed fast path since last measurement.

runtime/near-vm-logic/src/gas_counter.rs
```diff
@@ -65,32 +96,67 @@ impl GasCounter {
...
-            res
+            Err(self.process_gas_limit(new_burnt_gas, new_used_gas))
         }
     }
 
+    // Optimized version of above function for cases where no promises involved.
+    pub fn burn_gas(&mut self, value: Gas) -> Result<()> {
+        let new_burnt_gas =
+            self.fast_counter.burnt_gas.checked_add(value).ok_or(HostError::IntegerOverflow)?;
+        if new_burnt_gas <= self.fast_counter.gas_limit {
```
mina86, line 125: This may be silly question, but why aren’t we checking used gas here? For example, if burnt_gas=5, promises_gas=5 and prepaid_gas=10 burning 1 gas won’t trigger this check but used gas will become greater than prepaid gas.

matklad: In this function, we only burn gas, new promises gas is always zero. 

When we attach some promises gas via generic deduct_gas, we update `fast_counter.gas_limit` here: 

https://github.com/near/nearcore/pull/5121/files#diff-6f4c80473731cc44faa4c8b6e4e25c047e56e4bae235ac1e5138b92a75ffde1aR110

PR #5123, opened 2021-11-03
Mocking the system clock

yordanmadzhunkov:
I accidentally invoke everybody on the review. Feel free to leave, if you feel there are enough colleagues already invoked. For those who want to participate, I will explain the motivation:

We want to test our code using fuzz tests, because it provides benefits, using like random input 
 * to maximize code coverage
 * identify bugs in the code and recording the randomized input, so later a developer can replay/inspect that specific test
 
 In order to get those benefits, we need repeatable tests with require high level of determinism. Putting the `Utc::now` in blockchain header and computing hash is not deterministic. There are other source of not-deterministic behaviour like threads, random seed which will not be addressed in this review.
 
The goal of this PR is to make it possible to inject values into system clock before invoking the production code. The actual code in production should have no idea that we are mocking the system clock. 

Mocking the system clock.

### DIFF
--- a/core/primitives/src/time.rs
+++ b/core/primitives/src/time.rs
@@ -1,0 +1,155 @@
+use std::default::Default;
+
+use chrono;
+
+pub use chrono::Utc;
+pub use std::time::{Duration, Instant};
+
+use chrono::DateTime;
+use std::cell::RefCell;
+use std::collections::VecDeque;
+
+struct MockClockPerThread {
+    utc: VecDeque<DateTime<Utc>>,
+    durations: VecDeque<Duration>,
+    utc_call_count: u64,
+    instant_call_count: u64,
+    instant: Instant,
+    is_mock: bool,
+}
+
+pub struct Clock {}
+
+impl MockClockPerThread {
+    pub fn reset(&mut self) {
+        self.utc.clear();
+        self.durations.clear();
+        self.utc_call_count = 0;
+        self.instant_call_count = 0;
+        self.instant = Instant::now();
+        self.is_mock = false;
+    }
+
+    fn with<F, T>(f: F) -> T
+    where
+        F: FnOnce(&mut MockClockPerThread) -> T,
+    {
+        thread_local! {
+            static INSTANCE: RefCell<MockClockPerThread> = RefCell::default()
+        }
+        INSTANCE.with(|it| f(&mut *it.borrow_mut()))
+    }
+
+    fn pop_utc(&mut self) -> Option<DateTime<chrono::Utc>> {
+        self.utc_call_count += 1;
+        self.utc.pop_front()
+    }
+    fn pop_instant(&mut self) -> Option<Instant> {
+        self.instant_call_count += 1;
+        let x = self.durations.pop_front();
+        match x {
+            Some(t) => self.instant.checked_add(t),
+            None => None,
+        }
+    }
+}
+
+impl Default for MockClockPerThread {
+    fn default() -> Self {
+        Self {
+            utc: VecDeque::with_capacity(16),
+            durations: VecDeque::with_capacity(16),
+            utc_call_count: 0,
+            instant_call_count: 0,
+            instant: Instant::now(),
+            is_mock: false,
+        }
+    }
+}
+
+pub struct MockClockGuard {}
+
+impl Default for MockClockGuard {
+    fn default() -> Self {
+        Clock::set_mock();
+        Self {}
+    }
+}
+
+impl Drop for MockClockGuard {
+    fn drop(&mut self) {
+        Clock::reset();
+    }
+}
+
+impl Clock {
+    pub fn set_mock() {
+        MockClockPerThread::with(|clock| {
+            clock.is_mock = true;
+        });
+    }
+    pub fn reset() {
+        MockClockPerThread::with(|clock| {
+            clock.reset();
+        });
+    }
+    pub fn add_utc(mock_date: DateTime<chrono::Utc>) {
+        MockClockPerThread::with(|clock| {
+            if clock.is_mock {
+                clock.utc.push_back(mock_date);
+            } else {
+                panic!("Use MockClockGuard in your test");
+            }
+        });
+    }
+
+    pub fn add_instant(mock_instant: Duration) {
+        MockClockPerThread::with(|clock| {
+            if clock.is_mock {
+                clock.durations.push_back(mock_instant);
+            } else {
+                panic!("Use MockClockGuard in your test");
+            }
+        });
+    }
+
+    pub fn utc() -> DateTime<chrono::Utc> {
+        MockClockPerThread::with(|clock| {
+            if clock.is_mock {
+                let x = clock.pop_utc();
+                match x {
+                    Some(t) => t,
+                    None => {
+                        panic!("Mock clock run out of samples");
+                    }
+                }
+            } else {
+                chrono::Utc::now()
+            }
+        })
+    }
+
+    pub fn instant() -> Instant {
+        MockClockPerThread::with(|clock| {
+            if clock.is_mock {
+                let x = clock.pop_instant();
+                match x {
+                    Some(t) => t,
+                    None => {
+                        panic!("Mock clock run out of samples");
+                    }
+                }
+            } else {
+                Instant::now()
+            }
+        })
+    }
+
+    pub fn instant_call_count() -> u64 {
+        MockClockPerThread::with(|clock| clock.instant_call_count)
+    }
+
+    pub fn utc_call_count() -> u64 {
+        MockClockPerThread::with(|clock| clock.utc_call_count)
+    }
+}
--- a/chain/chain-primitives/src/error.rs
+++ b/chain/chain-primitives/src/error.rs
@@ -1,7 +1,9 @@
 use std::fmt::{self, Display};
 use std::io;
 
-use chrono::{DateTime, Utc};
+use chrono::DateTime;
+use near_primitives::time::Utc;
+
 use failure::{Backtrace, Context, Fail};
 use log::error;
 
--- a/chain/chain/src/chain.rs
+++ b/chain/chain/src/chain.rs
@@ -4,8 +4,8 @@ use std::time::{Duration as TimeDuration, Instant};
 
 use borsh::BorshSerialize;
 use chrono::Duration;
-use chrono::Utc;
 use itertools::Itertools;
+use near_primitives::time::{Clock, Utc};
 use rand::rngs::StdRng;
 use rand::seq::SliceRandom;
 use rand::SeedableRng;
@@ -476,7 +476,7 @@ impl Chain {
         self.orphans.add(Orphan {
             block: block.clone(),
             provenance: Provenance::NONE,
-            added: Instant::now(),
+            added: Clock::instant(),
         });
         Ok(())
     }
@@ -1145,7 +1145,7 @@ impl Chain {
                         // we only add blocks that couldn't have been gc'ed to the orphan pool.
                         if block_height >= tail_height {
                             let block_hash = *block.hash();
-                            let orphan = Orphan { block, provenance, added: Instant::now() };
+                            let orphan = Orphan { block, provenance, added: Clock::instant() };
 
                             self.orphans.add(orphan);
 
@@ -1165,7 +1165,7 @@ impl Chain {
                     ErrorKind::ChunksMissing(missing_chunks) => {
                         let block_hash = *block.hash();
                         block_misses_chunks(missing_chunks.clone());
-                        let orphan = Orphan { block, provenance, added: Instant::now() };
+                        let orphan = Orphan { block, provenance, added: Clock::instant() };
 
                         self.blocks_with_missing_chunks.add_block_with_missing_chunks(
                             orphan,
--- a/chain/chain/src/doomslug.rs
+++ b/chain/chain/src/doomslug.rs
@@ -5,6 +5,7 @@ use std::time::{Duration, Instant};
 use near_crypto::Signature;
 use near_primitives::block::{Approval, ApprovalInner};
 use near_primitives::hash::CryptoHash;
+use near_primitives::time::Clock;
 use near_primitives::types::{AccountId, ApprovalStake, Balance, BlockHeight, BlockHeightDelta};
 use near_primitives::validator_signer::ValidatorSigner;
 
@@ -296,8 +297,8 @@ impl Doomslug {
             tip: DoomslugTip { block_hash: CryptoHash::default(), height: 0 },
             endorsement_pending: false,
             timer: DoomslugTimer {
-                started: Instant::now(),
-                last_endorsement_sent: Instant::now(),
+                started: Clock::instant(),
+                last_endorsement_sent: Clock::instant(),
                 height: 0,
                 endorsement_delay,
                 min_delay,
@@ -596,11 +597,12 @@ impl Doomslug {
 #[cfg(test)]
 mod tests {
     use std::sync::Arc;
-    use std::time::{Duration, Instant};
+    use std::time::Duration;
 
     use near_crypto::{KeyType, SecretKey};
     use near_primitives::block::{Approval, ApprovalInner};
     use near_primitives::hash::hash;
+    use near_primitives::time::Clock;
     use near_primitives::types::{AccountId, ApprovalStake};
     use near_primitives::validator_signer::InMemoryValidatorSigner;
 
@@ -625,7 +627,7 @@ mod tests {
             DoomslugThresholdMode::TwoThirds,
         );
 
-        let mut now = Instant::now(); // For the test purposes the absolute value of the initial instant doesn't matter
+        let mut now = Clock::instant(); // For the test purposes the absolute value of the initial instant doesn't matter
 
         // Set a new tip, must produce an endorsement
         ds.set_tip(now, hash(&[1]), 1, 1);
@@ -781,7 +783,7 @@ mod tests {
             DoomslugThresholdMode::TwoThirds,
         );
 
-        let mut now = Instant::now();
+        let mut now = Clock::instant();
 
         // In the comments below the format is
         // account, height -> approved stake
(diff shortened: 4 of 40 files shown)


matklad:
Oh, fascinating, we are just having an interesting discussion of what time we should use in the first place at https://github.com/near/nearcore/pull/5097 cc @pmnoxx

core/primitives/src/time.rs
```diff
@@ -0,0 +1,171 @@
...
+    pub fn current_mut(&mut self) -> Option<&mut MockTimeSingletonPerThread> {
+        let handle = std::thread::current();
+        let id = handle.id();
+        if !self.threads.contains_key(&id) {
+            self.threads.insert(id, MockTimeSingletonPerThread::default());
+        }
+        self.threads.get_mut(&id)
+    }
+
+    pub fn current(&self) -> Option<&MockTimeSingletonPerThread> {
```
matklad, line 100: Feels like significant amount of stuff in this file doesn't need to be `pub`. Right now, it's pretty hard to see what is the public interface I should use from other crates, and what is a private impl detail.

core/primitives/src/time.rs
```diff
@@ -0,0 +1,171 @@
...
+        self.current().unwrap().durations.len()
+    }
+}
+
+pub trait MockTime {
+    type Value;
+
+    fn now_or_mock() -> Self::Value;
+
+    fn system_time() -> Self::Value;
```
matklad, line 136: This sholdn't be called `system_time`, as https://doc.rust-lang.org/stable/std/time/struct.SystemTime.html is a very specific, different thing.

core/primitives/src/time.rs
```diff
@@ -0,0 +1,171 @@
...
+use once_cell::sync::Lazy;
+
+pub use chrono::Utc;
+pub use std::time::Instant;
+
+use std::collections::{HashMap, VecDeque};
+use std::thread::ThreadId;
+
+pub struct MockTimeSingletonPerThread {
+    utc: VecDeque<chrono::DateTime<chrono::Utc>>,
```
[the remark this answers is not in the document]
yordanmadzhunkov: Done, thanks for the suggestion

bowenwang1996 (commented):
Please change the title of your PR to a descriptive one. Thanks!

core/primitives/src/time.rs
```diff
@@ -0,0 +1,171 @@
...
+    }
+
+    pub fn current_mut(&mut self) -> Option<&mut MockTimeSingletonPerThread> {
+        let handle = std::thread::current();
+        let id = handle.id();
+        if !self.threads.contains_key(&id) {
+            self.threads.insert(id, MockTimeSingletonPerThread::default());
+        }
+        self.threads.get_mut(&id)
+    }
```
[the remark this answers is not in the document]
yordanmadzhunkov: Awesome suggestion. Everything becomes much simpler.

core/primitives/src/time.rs
```diff
@@ -0,0 +1,171 @@
...
+
+    pub fn reset(&mut self) {
+        self.current_mut().unwrap().reset();
+    }
+
+    pub fn add_instant(&mut self, mock_instant: time::Duration) {
+        self.current_mut().unwrap().durations.push_back(mock_instant);
+    }
+
+    pub fn get_instant_call_count(&mut self) -> u64 {
```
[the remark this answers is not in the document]
yordanmadzhunkov: Thanks

yordanmadzhunkov:
@matklad 
I also have mixed feelings. 
Specific design issues with the current solution:

1. You understood it correctly. Refactoring the example code to call now once, I can break the tests, (only if you check number of times timer is called) 

2. I will implement
`"if time is mocked { get mocked instance or panic } else { return real time}"`

3. Good point. Looks like the real clock ::now method also has build in mutex. At this point I will not be worried about performance issues

4. I am also not sure, what is the right approach

5. The motivation to make time mocking local to specific thread is the fact that test run in parallel in separate threads. If time mocking is not thread specific, different tests interfere between each other.

yordanmadzhunkov:
@matklad 
Actually, on point 4. I think the correct approach is to create a clock interface and use dependency injection. This way, we can mock tests and in production we can simply inject real system clock. This will require a lot of work to pass the instance to the clock between functions, classes and etc.

pmnoxx:
I like the idea of adding mocking to time. While writing tests I had to add a bit of code to support testing code, which gets executed rarely, having mocks for time would be useful.

However:
- A test usually spawns multiple threads. Most likely you will have to change code that starts all actors in the code, in such a way, that child thread uses the same time mock as it's parent.
I would use something like:
```
thread_local! {
    pub static Singletor: Optional<Arc<RefCell<MockClockPerThread>>>> = ...;
};
```
Everytime you start new actor, yo would make it point to the same Arc<RefCell<MockClockPerThread>>>` as it parent.

- We also don't mock time for Actix, timeouts, etc. It would be great to be able to add an option to increase speed of progression of the time to test timeouts, etc.

- I'm starting to think that it may be a good idea refactor our code to use one library for time. That would make code much cleaner, we wouldn't have to mock different type of times. 
I would be in favor of having our own clock.

```
struct Clock { 
    // number of ns since 1970 / utc time
    ns: u64
}
struct Duration { 
    // difference in ns
    ns: i64
}
```
function:
- now()
- duration_since(...)
- sub(...) // singled difference
- we can add `BorshSerialize` / `BorshDeserialize` / this can be used to replace `u64` as time storage.

This would allow us to use single source of time for all code, and also it would be easier to mock.
Sometimes we use `u64`, sometimes `Instant::now`, `Datetime<Utc>::new()`, etc.

bowenwang1996 (commented):
@yordanmadzhunkov after a discussion with @matklad we agree that the design in this implementation is not optimal. However, to not get blocked on this for fuzzing, I suggest that you figure out the minimal amount of change we need to make the fuzzing possible (deterministic). There are likely a few places where we can refactor the functions to have time passed in as a parameter so that we can avoid doing a big change like this at once.

core/primitives/src/time.rs
```diff
@@ -0,0 +1,171 @@
...
+        match x {
+            Some(t) => t,
+            None => time::Instant::now(),
+        }
+    }
+
+    fn system_time() -> time::Instant {
+        time::Instant::now()
+    }
+}
```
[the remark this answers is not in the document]
yordanmadzhunkov: Done

runtime/runtime-params-estimator/src/main.rs
```diff
@@ -188,7 +189,7 @@ fn main() -> anyhow::Result<()> {
     );
 
     let output_path = {
-        let timestamp = chrono::Utc::now().to_rfc3339_opts(chrono::SecondsFormat::Secs, true);
+        let timestamp = Clock::utc().to_rfc3339_opts(chrono::SecondsFormat::Secs, true);
```
matklad, line 192: Here we always want to use the real clock.

runtime/runtime-params-estimator/src/testbed_runners.rs
```diff
@@ -4,6 +4,7 @@ use crate::testbed::RuntimeTestbed;
 use indicatif::{ProgressBar, ProgressStyle};
 use near_crypto::{InMemorySigner, KeyType};
 use near_primitives::hash::CryptoHash;
+use near_primitives::time::Instant;
```
matklad, line 7: ditto here: basically, all the code in `runtime-params-estimator` should be exempt from time mocking. This is essentially benchmarking code, and it has its' own requrenments for time, unrelated to the rest of the code.

core/primitives/src/time.rs
```diff
@@ -0,0 +1,156 @@
...
+static SINGLETON: Lazy<Arc<Mutex<Clock>>> = Lazy::new(|| Arc::from(Mutex::new(Clock::new())));
+
+impl Clock {
+    pub fn new() -> Self {
+        Self::default()
+    }
+
+    pub fn get() -> &'static Arc<Mutex<Clock>> {
+        SINGLETON.deref()
+    }
```
matklad, line 67:
```suggestion
    pub fn get() -> MutexGuard<'static, Clock> {
        SINGLETON.lock().unwrap()
    }
```

would simplify call-sites.

core/primitives/src/time.rs
```diff
@@ -0,0 +1,156 @@
...
+}
+
+impl Default for Clock {
+    fn default() -> Self {
+        Self { clocks_per_thread: HashMap::default() }
+    }
+}
+
+static SINGLETON: Lazy<Arc<Mutex<Clock>>> = Lazy::new(|| Arc::from(Mutex::new(Clock::new())));
```
matklad, line 58: Rather than maintaining per-thread map, Arc and lazy ourselves, we can leverage std::therad_local:

```rust
struct MockClockPerThread {
    utc: VecDeque<DateTime<Utc>>,
    durations: VecDeque<Duration>,
    utc_call_count: u64,
    instant_call_count: u64,
    instant: Instant,
    is_mock: bool,
}

impl MockClockPerThread {
    fn with<F, T>(f: F) -> T
    where
        F: FnOnce(&mut MockClockPerThread) -> T,
    {
        thread_local! {
            static INSTANCE: RefCell<MockClockPerThread> = RefCell::default()
        }
        INSTANCE.with(|it| f(&mut *it.borrow_mut()))
    }

    fn pop_utc(&mut self) -> Option<DateTime<chrono::Utc>> {
        self.utc_call_count += 1;
        self.utc.pop_front()
    }
}

pub struct Clock {}

impl Clock {
    pub fn add_utc(mock_date: DateTime<chrono::Utc>) {
        MockClockPerThread::with(|clock| {
            clock.utc.push_back(mock_date);
            clock.is_mock = true;
        });
    }
    pub fn utc() -> DateTime<chrono::Utc> {
        MockClockPerThread::with(|clock| {
            let x = clock.pop_utc();
            match x {
                Some(t) => t,
                None => {
                    if clock.is_mock {
                        panic!("Mock clock run out of samples");
                    }
                    chrono::Utc::now()
                }
            }
        })
    }
}
```

core/primitives/src/time.rs
```diff
@@ -0,0 +1,156 @@
...
+    pub fn get() -> &'static Arc<Mutex<Clock>> {
+        SINGLETON.deref()
+    }
+    pub fn reset(&mut self) {
+        self.current_mut().reset();
+    }
+    pub fn add_utc(&mut self, mock_date: DateTime<chrono::Utc>) {
+        let instance = self.current_mut();
+        instance.utc.push_back(mock_date);
+        instance.is_mock = true;
```
matklad, line 74: :thinking: I am surprised that this passes the tests. If two tests are run on the same thread, the first test uses mocking, but the second doesn't, the second should fail at runtime, as it'll hit `is_mock = true` from the first test. For the "thread-local singleton" design, I'd probably expect some kind of RAII API here: 

```rust
#[test] 
fn test_with_mock_time() {
  let mut tm = Clock::time_machine(); // sets `is_mock` to  true
  tm.add_utc("July 3, 1985");

  // implimit `drop(tm)` sets `is_mock` to false;
}
```

core/primitives/Cargo.toml
```diff
@@ -29,6 +29,8 @@ reed-solomon-erasure = "4"
 hex = "0.4"
 num-rational = { version = "0.3", features = ["serde"] }
 primitive-types = "0.10"
+once_cell = "1"
+arc-swap = "1.2"
```
matklad, line 33: We no longer need this dependency.

core/primitives/src/time.rs
```diff
@@ -0,0 +1,156 @@
...
+
+pub use chrono::Utc;
+pub use std::time::{Duration, Instant};
+
+use chrono::DateTime;
+use std::collections::{HashMap, VecDeque};
+use std::ops::Deref;
+use std::thread::ThreadId;
+
+pub struct MockClockPerThread {
```
matklad, line 15: No need to keep this `pub`. It's relatively important to not make things `pub`, as knowing what API exends past the current compilation unit makes reading the code easier.

matklad (approved):
Approving to unblock, but there are three fixes we need to do here: 

* remove unneded cargo dependency 
* remove changes to the estimator
* make sure that `is_mock` can't leak between tests

The `thread_local` suggestion is optional -- I do like to minimize the amount of machinery we use, but as this is impl details, this doesnt' really matter. 

Two other nice-to-haves:

* are we sure that this change is enough to enable the fuzzing we want? Given that we are not mocking tokio's time or `thread::sleep`s I wouldn't be surprised if this isn't actually enough.
* are we sure that this is the minimal change? From my conversation with @bowenwang1996, I recall the idea was to just refactor some specific part of the code-base to take time as an argument, without introducing thread-locals, but also without fixing *every* usage of time. If that turns out to be infeasible, I think it's OK to do what we are doing here, but, still, let's try to minimize the blast radius of the design which we agree is not the long-term one.

test-utils/loadtester/src/stats.rs
```diff
@@ -0,0 +1,91 @@
...
+    pub fn collect_transactions(&mut self, node: &RemoteNode) {
+        let mut curr_height = self.from_height.unwrap() + 1;
+        let mut total_tx = 0u64;
+        loop {
+            total_tx += get_result(|| node.get_transactions(curr_height));
+            curr_height += 1;
+            if curr_height > self.to_height.unwrap() {
+                break;
+            }
+        }
```
bowenwang1996, line 88: It seems that we can rewrite this as 
```rust
while curr_height <= self.to_height.unwrap() {
   ...
}
```

test-utils/loadtester/src/stats.rs
```diff
@@ -0,0 +1,91 @@
+use crate::remote_node::{get_result, RemoteNode};
```
bowenwang1996, line 1: I believe `test-utils/loadtester` is removed in master. What is the reason for restoring it?

bowenwang1996 (approved):
Approve to unblock, but please address Aleksey's comments.

core/primitives/src/time.rs
matklad: To clarify -- the current design also works I think, no need to rewrite if you don't like the new version better.

matklad: :thinking: that being said, I think there's a potential problem with the hash map -- we never remove entries form it. If we do maintain hashmap in the prod code, and we do create new threads from time to time, than this would leak memory. If that is indeed the problem (which I dont' know), we should switch to `thread_local`, as that includes the necessary code to release resources when a thread exists.

chain/network/src/routing.rs
```diff
@@ -824,8 +826,8 @@ impl RoutingTable {
         trace!(target: "network", "Update routing table.");
 
         self.peer_forwarding = self.raw_graph.calculate_distance();
-
-        let now = Clock::instant();
+        let clock = &mut Clock {};
+        let now = clock.instant();
```
matklad, line 830: These `&mut Clock {}` feel odd, perhaps we should keep `::instant` to be an associated fn (without self)?

pmnoxx:
I created my own implementation Clock for purpose of mocking time. Please, consider an alternative:
https://github.com/near/nearcore/pull/5175

core/primitives/src/time.rs
```diff
@@ -0,0 +1,122 @@
...
+    }
+
+    pub fn add_instant(&mut self, mock_instant: Duration) {
+        MockClockPerThread::with(|clock| {
+            clock.durations.push_back(mock_instant);
+            clock.is_mock = true;
+        });
+    }
+
+    pub fn utc(&mut self) -> DateTime<chrono::Utc> {
```
pmnoxx, line 85: How about we change signature to `pub fn utc_now() -> DateTime<chrono::Utc>`

- self is not used - This will allow us to use `Clock::utc_now()` without having to create `Clock {}`
- changing `utc` to `utc_now` should make code more readable. Without prior knowledge we will know that we are getting current time.

pmnoxx (approved):
LGTM

pmnoxx:
@yordanmadzhunkov I pushed a few huge changes to `chain/network`, you may have to resolve merge conflicts.

Good job on the PR.
I look forward to seeing the next PR, where you add propagation of `Clock` to throughout our code base.
After that we can work on cleaning up the code, to make sure we only have one structure representing time instead of 4.

pmnoxx:
@chefsale @frol Can you take a look?

core/primitives/src/time.rs
yordanmadzhunkov: Done

core/primitives/src/time.rs
yordanmadzhunkov: Done :)

chain/network/src/routing.rs
yordanmadzhunkov: Done

test-utils/loadtester/src/stats.rs
yordanmadzhunkov: Removed, it was added by accident

core/primitives/src/time.rs
```diff
@@ -0,0 +1,171 @@
...
+
+pub use chrono::Utc;
+pub use std::time::Instant;
+
+use std::collections::{HashMap, VecDeque};
+use std::thread::ThreadId;
+
+pub struct MockTimeSingletonPerThread {
+    utc: VecDeque<chrono::DateTime<chrono::Utc>>,
+    durations: VecDeque<time::Duration>,
```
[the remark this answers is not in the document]
yordanmadzhunkov: This will be reworked later.

core/primitives/src/time.rs
yordanmadzhunkov: yes, you are right :)

test-utils/loadtester/src/stats.rs
yordanmadzhunkov: no reason, it looks like it happened by accident

matklad (approved):
:+1: 

Couple of more nitpicly coments:

* some `pub` can still be removed
* Rather than making the user import `MockClockGuard`, the API could look like this:

```rust
let _guard: MockClockGuard = Clock::mock();
```

:thinking: thinking more about this, I think it even makes sense to make `add_utc` and such methods of the guard object, rather than of the `Clock` itself. 

Anyway, I don't think we should try to polish Rust API in this PR (we must polish it sometime later though, time mocking is a fundamental fascility used thourgouht the codebase, its importatn that the API is just right, otherwise changing it later would be a pain).

mzhangmzz (approved):

PR #5125, opened 2021-11-03
make view client use a different runtime object to avoid contention

mzhangmzz:
The idea is to use a separate runtime for view client to prevent it from blocking transaction processing. Right now, they share the same `RuntimeAdapter`, which means they share the same `EpochManager` and `ShardTries`, both of which are mostly just an in-memory cache around storage access. The cache is read-blocking, so even though view_client only need to read from the caches, it blocks `client` from accessing or writing to the cache. 
 
This fix makes view_client and client used different `RuntimeAdapter` objects. It works because all the uses in `view_client` on `runtime_adapter` are wrappers around `EpochManager` except for `RuntimeAdapter::query`. And the access to `EpochManager` on access the caches. In other words, `view_client` access do not depend on the internal state of `RuntimeAdapter` or `EpochManager`.

Test Plan:
- CI
- Nayduck


matklad (commented):
Very cool finding indeed!

Seems like this'll essentially 2x our memory requirenment? Do we have any estimates on the impact of this change in terms of resource usage?

mzhangmzz:
There are a few more items we should do because this solution is quite error prone (what if someone in the future adds a use in view_client that does depend on the internal state of `RuntimeAdapter` or `EpochManager`.

1. We use restrict `ViewClient` to use `ViewRuntimeActor` instead of `RuntimeAdapter`. This way, we can finely control `ViewClient`'s access to runtime. The method in `ViewRuntimeActor` should be read-only and does not depend on runtime's internal state. This is doable because as I said, most call sites for `RuntimeAdapter` in `ViewClient` is just a wrapper around `EpochManager` except for `RuntimeAdapter::query`. We can move the `query` implementation to `ViewRuntimeAdapter` and made `ViewClient` calls `EpochManager` and `ViewRuntimeAdapter`. This will also be a first step to separating `EpochManager` from `RuntimeAdapter`
2. We should make the cache implementation in `EpochManager` not read-blocking. Right now, it's read blocking because we use a LRU-cache for cache eviction policy. Since most of the access to `EpochManager` will be read-only, we should change the cache implementation. This will also help the performance of `ClientActor` since it calls on `EpochManager` extensively.

mzhangmzz:
> Very cool finding indeed!
> 
> Seems like this'll essentially 2x our memory requirenment? Do we have any estimates on the impact of this change in terms of resource usage?

That's a good point. I hope the memory increase would not be 2x because most of the data is stored on the chain level instead of runtime level. I think the only memory intensive objects that are duplicate will be `TrieCache` and `EpochManager`. `TrieCache` can hold up to 4000 * 4000 (TRIE_MAX_CACHE_SIZE * TRIE_MAX_CACHE_SIZE) bytes = 16 MB and `EpochManager` can hold up to 1000 `BlockInfo`s and 50 `EpochInfo`s. I am not sure how large that is, but I don't think it will be too bad.

bowenwang1996:
> There are a few more items we should do because this solution is quite error prone (what if someone in the future adds a use in view_client that does depend on the internal state of `RuntimeAdapter` or `EpochManager`.
> 
> 1. We should restrict `ViewClient` to use `ViewRuntimeActor` instead of `RuntimeAdapter`. This way, we can finely control `ViewClient`'s access to runtime. The method in `ViewRuntimeActor` should be read-only and does not depend on runtime's internal state. This is doable because as I said, most call sites for `RuntimeAdapter` in `ViewClient` is just a wrapper around `EpochManager` except for `RuntimeAdapter::query`. We can move the `query` implementation to `ViewRuntimeAdapter` and made `ViewClient` calls `EpochManager` and `ViewRuntimeAdapter`. This will also be a first step to separate `EpochManager` from `RuntimeAdapter`
> 2. We should make the cache implementation in `EpochManager` not read-blocking. Right now, it's read blocking because we use a LRU-cache for cache eviction policy. Since most of the access to `EpochManager` will be read-only, we should change the cache implementation. This will also help the performance of `ClientActor` since it calls on `EpochManager` extensively.

Very good points. Could you create separate issues for those? Thanks!

bowenwang1996 (approved):

matklad:
>We should make the cache implementation in EpochManager

FWIW, we use `cached` crate a lot at the moment, and I am not a huge fan of it:

* the API is overly-complicated for what it does: there's `Cached` trait, cached_key` declarative macro, and procedural macros as well. Cache is a kind of hash map, it doesn't really need these advanced features of the Rust type-system. 
* the API is inconvenient: it requires `&mut` access, but for caches you almost always need `&`. Solving `&mut` vs `&` via interior mutability is actually the primary thing I'd expect from a cache crate,  as that's the most interesting problem, but, in this case, that's just punted to the user. 
* the crate is at `0.26` version, so it is still unsure about it's own API

Somewhat orthogonally, we are a database, so we use caches a lot and they are our core competency. So, over the long run, it generally makes sense to build such things ourselves.

mzhangmzz:
> > We should make the cache implementation in EpochManager
> 
> FWIW, we use `cached` crate a lot at the moment, and I am not a huge fan of it:
> 
> * the API is overly-complicated for what it does: there's `Cached` trait, `cached_key` declarative macro, and procedural macros as well. Cache is a kind of hash map, it doesn't really need these advanced features of the Rust type-system.
> * the API is inconvenient: it requires `&mut` access, but for caches you almost always need `&`. Solving `&mut` vs `&` via interior mutability is actually the primary thing I'd expect from a cache crate,  as that's the most interesting problem, but, in this case, that's just punted to the user.
> * the crate is at `0.26` version, so it is still unsure about it's own API
> 
> Somewhat orthogonally, we are a database, so we use caches a lot and they are our core competency. So, over the long run, it generally makes sense to build such things ourselves.

I totally agree. Another thought I have is that if we are going to rewrite our cache implementation, we should also think about cache management. Right now, different structs in our code keep their own cache, and use a function called `read_with_cache` to read from the database and cache and they need to implement write to cache and database separately. Nothing prevents two structs holding cache/read/write to the same column. As a result, we need to be really careful algorithmically about the ownership of caches on different columns in the database.

PR #5126, opened 2021-11-03
refactor: drop dead code

matklad:
This impl of `Hash` is inconsistent with our derived Eq, which violates
trait's contract:

https://doc.rust-lang.org/stable/std/hash/trait.Hash.html#hash-and-eq

Seems like we don't actually use this, so let's just remove.


bowenwang1996 (approved):

mm-near (approved):
Thank you !

PR #5100, opened 2021-10-29
chore: ready workspace for publishing

miraclx:
Patch-bump versions of public crates across the workspace to latest.

- [near-account-id](https://crates.io/crates/near-account-id/versions) `v0.1.0` => `v0.2.0`
- [near-chain-configs](https://crates.io/crates/near-chain-configs/versions) `v0.1.0` => `v0.2.0`
- [near-chain-primitives](https://crates.io/crates/near-chain-primitives/versions) `v0.1.0` => `v0.2.0`
- [near-chunks-primitives](https://crates.io/crates/near-chunks-primitives/versions) `v0.1.0` => `v0.2.0`
- [near-client-primitives](https://crates.io/crates/near-client-primitives/versions) `v0.1.0` => `v0.2.0`
- [near-crypto](https://crates.io/crates/near-crypto/versions) `v0.1.0` => `v0.2.0`
- [near-jsonrpc-primitives](https://crates.io/crates/near-jsonrpc-primitives/versions) `v0.2.0` => `v0.3.0`
- [near-metrics](https://crates.io/crates/near-metrics/versions) `v0.1.0` => `v0.2.0`
- [near-network-primitives](https://crates.io/crates/near-network-primitives/versions) `v0.1.0` => `v0.2.0`
- [near-primitives-core](https://crates.io/crates/near-primitives-core/versions) `v0.4.0` => `v0.5.0`
- [near-primitives](https://crates.io/crates/near-primitives/versions) `v0.1.0-pre.1` => `v0.1.0`
- [near-rpc-error-core](https://crates.io/crates/near-rpc-error-core/versions) `v0.1.0` => `v0.2.0`
- [near-rpc-error-macro](https://crates.io/crates/near-rpc-error-macro/versions) `v0.1.0` => `v0.2.0`
- [near-vm-errors](https://crates.io/crates/near-vm-errors/versions) `v3.0.0` & `v4.0.0-pre.1` => `v3.1.0`

Note: Since https://github.com/near/nearcore/pull/4909, all but these few crates marked private. Meaning, while there are many more crates in the workspace that are currently published, they're excluded from this list, this is to make sure we consciously, explicitly and hopefully, meticulously manage what gets published and what doesn't.

It's also worth noting that we're currently automating the publishing workflow. Getting a new version of a crate in the workspace published should be as simple as; first, ensuring there's no `publish = false` specification in it's `Cargo.toml` and finally, simply updating the version number as with this PR.

CI should check and auto-publish new versions of public crates..

[...]

PR #5130, opened 2021-11-03
re: chore: ready workspace for publishing

miraclx:
### The Plan

I closed #5100 because manually bumping versions is understandably tasking, and automating this requires special tooling or at least patching existing tooling to support our use case. Otherwise, we'd run into issues from having certain specific dep relationships. Luckily `cargo-workspaces` had already done most of the work, I just needed to patch it to better suit our workflow.

Here's what we do want:

- Organized and automated publishing. Ideally, nearcore as an entity as maintained by all its core contributors should publish its crates. Not individual contributors.
- All crates in the workspace should share the same version.
- The versions of published crates on crates.io should always match their versions on GitHub.
- Unversioned crates like `nearcore`, `neard`, `integration-tests` that are on `0.0.0` should be excluded from version bumping.
- Runtime crates that are already on `3.0.0` (e.g [`crates.io:near-vm-logic`](https://crates.io/crates/near-vm-logic/versions)) should share a common version but be distinct from the rest of the workspace.
- All crates should, by default, be marked private. But to publish, it should be as easy as marking it publishable. (removing `publish = false`).

My fork of [`miraclx/cargo-workspaces`](https://github.com/pksunkara/cargo-workspaces/compare/master...miraclx:grouping-and-exclusion) is now better suited for our use case.

#### How?

 - Unversioned crates that should, in no case, be versioned, are excluded in the workspace manifest.
   https://github.com/near/nearcore/blob/31763be351db59ac6f667d1fca21d91da0229536/Cargo.toml#L49-L58
 - Groups of related crates whose versions need to differ from the rest of the workspace can be grouped in the workspace manifest. (This ensures that while everything is versioned to `0.5.0`, the runtime group is ticked to `3.1.0`)
   https://github.com/near/nearcore/blob/31763be351db59ac6f667d1fca21d91da0229536/Cargo.toml#L60-L69

#### Versioning

The current common version across the workspace as at the current master (00138f22aed6a877ac86bc609a015e7ce6071e21) is `0.4.0` and the current common version across the `[runtime]` crates is `3.0.0`

With this, we bump the versions of crates across the entire workspace to the next minor - `0.5.0` and in the `[runtime]` group to `3.1.0` and only publish however many are marked publishable:

- [near-account-id](https://crates.io/crates/near-account-id/versions) `0.1.0` => `0.5.0`
- [near-chain-configs](https://crates.io/crates/near-chain-configs/versions) `0.1.0` => `0.5.0`
- [near-chain-primitives](https://crates.io/crates/near-chain-primitives/versions) `0.1.0` => `0.5.0`
- [near-chunks-primitives](https://crates.io/crates/near-chunks-primitives/versions) `0.1.0` => `0.5.0`
- [near-client-primitives](https://crates.io/crates/near-client-primitives/versions) `0.1.0` => `0.5.0`
- [near-crypto](https://crates.io/crates/near-crypto/versions) `0.1.0` => `0.5.0`
- [near-jsonrpc-primitives](https://crates.io/crates/near-jsonrpc-primitives/versions) `0.2.0` => `0.5.0`
- [near-metrics](https://crates.io/crates/near-metrics/versions) `0.1.0` => `0.5.0`
- [near-network-primitives](https://crates.io/crates/near-network-primitives/versions) `0.1.0` => `0.5.0`
- [near-primitives-core](https://crates.io/crates/near-primitives-core/versions) `0.4.0` => `0.5.0`
- [near-primitives](https://crates.io/crates/near-primitives/versions) `0.1.0-pre.1` => `0.5.0`
- [near-rpc-error-core](https://crates.io/crates/near-rpc-error-core/versions) `0.1.0` => `0.5.0`
- [near-rpc-error-macro](https://crates.io/crates/near-rpc-error-macro/versions) `0.1.0` => `0.5.0`
- `[runtime]` [near-vm-errors](https://crates.io/crates/near-vm-errors/versions) `3.0.0` => `3.1.0`

#### Publishing Automation

I propose two systems;
 - The first one, let's call this Hermes (the messenger), runs on every push to master, checking for any version changes, publishing any unpublished versions.
 - The second, let's call this Chronos (for time), runs once at the end of the week. Bumping the versions across the entire workspace by one minor tick. Which then opens a PR with all the changes to be reviewed, approved and merged. Once merged, Hermes (the first system) takes over to publish them. This way, we ensure to have consistent versioning across the entire workspace.


matklad:
>[runtime] near-vm-errors 3.0.0 => 3.1.0

Sticks out like a sore thumb. I'll try to merge the crate into primitives to remove one thing we need to think about here.

matklad:
>I'll try to merge the crate into primitives to remove one thing we need to think about here.

I've reconsidered that. The list above misses `near-vm-logic`, but that is the crate we need to publish some version of, as the near-sdk-rs team needs it. cc @austinabell. The publishing is very much on the "just last one more time" basis though, we don't plant to support `near-vm-logic` to be usable outside of `nearcore`, current usages should migrate to sandbox instead. 

near-vm-logic is also past 1.0, so solving the problem just for `vm-errors` doesn't help with keeping error numbers unified. While removing `vm-errors` crate and merging it into primitives makes sense by itself, I don't want to do that until we publish near-vm-logic version for the sdk team to use, to avoid shaking crate structure sdk uses today too much. 

So, I suggest to do the following, to keep the versioning conssitent: publish *all* crates, including `near-vm-errors` and `near-vm-logic`, with version `0.10.0`. After sdk no longer uses past-1.0 versioned crates, we yank all those versions. 

It's very weird to have `0.10` which is newer than "3.0.0", but this is the weirdness only the nearcore team is exposed to (users shouldn't directly use any version of runtime crates), and it, especially after yanking, it is less weird than having some crates at `0.92` and some crates at `92.0` for obscure historical reason.

matklad:
>Bumping the versions across the entire workspace by one minor tick. Which then opens a PR with all the changes to be reviewed, approved and merged.

What value is brought in by this process? My gut instinct is that it's pure overhead. 

To take a step back, do we need "The versions of published crates on crates.io should always match their versions on GitHub." if we are also striving for "All crates in the workspace should share the same version."? A simpler solution is to avoid any meaningful versions in Cargo.tomls at all, and instead just fill them in at the point of publishing. I think that's what rustc and rust-analyzer are doing.

An alternative, used by Cargo, is to release the crates only after the binary release is cut (so, every six weeks), and have `0.xx.0` as a version in source Cargo.tomls for the `1.xx.0` release of the binary. 

Between the two, I lean towards the first option -- publish arbitrary commit every week, fill-in versions on publish.

miraclx:
> publish all crates, including near-vm-errors and near-vm-logic, with version 0.10.0. After sdk no longer uses past-1.0 versioned crates, we yank all those versions.

Love this idea. Thank you so much for your contributions, I wasn't sure if we could safely down-version those runtime crates already published past `1.0.0` but seeing as the major thing is `near-vm-logic` and that it's only used in `near-sdk` _(please confirm)_, here's what we can do:

I'll version the `[runtime]` group to `0.5.0`, yank the `>=1.0.0` versions from crates.io, I'll then dissolve that group into the rest of the workspace. Thereafter, I'll re-version everything to `0.0.0` _(stay with me)_.

> A simpler solution is to avoid any meaningful versions in Cargo.tomls at all, and instead just fill them in at the point of publishing

On the tooling side, I'll introduce a `version` field to `[workspace.metadata.workspaces]` that predefines a version for the entire workspace and set this to `0.10.0` as suggested.

So, on every push to master, Hermes (CI) would apply that version across the entire workspace, and if there's a version change, all non-private crates in the workspace gets published to crates.io.

> do we need "The versions of published crates on crates.io should always match their versions on GitHub."

Since we're dropping individual crate versions from the repo, Hermes (CI) can employ tagging instead. Any commit on master that changes the version field would be tagged. Something like `lib-v0.10.0` so you can, at least, easily find the git revision from which a non-private crate's version was published from. We could discuss this, because you could also just git blame on the version field, step through changes until you find the version, enter the tree. But tagging is easier, of course.

This would ensure we have consistent versioning across the workspace without needing Chronos (CI) to automate bumping versions across the workspace, nor do we need the system creating PRs with those changes. As, it would be as easy as bumping the `version` field in the workspace's manifest.

> What value is brought in by this process? My gut instinct is that it's pure overhead.

With the process describes in this comment, we won't need that anymore.

matklad:
Yeah, really like the updated proposal! Just one small thing, I wouldn't do "yank the >=1.0.0 versions from crates.io" right away. For that, I'd wait until the new version of the SDK is released, so that contract developers don't have yanked crates in the lockfiles. 

+1 to tags. Implementation wise, you want to first `cargo publish`, and *then* `git push --tags`. As a general tidbit of Rust knowledge, `cargo publish` now actually includes current commit into the published crate automatically: 

```
λ cat ~/.cargo/registry/src/github.com-1ecc6299db9ec823/once_cell-1.8.0/.cargo_vcs_info.json 
{
  "git": {
    "sha1": "8671ffc69cb78e345b7647802e5a5fd0c53f8c07"
  }
}
```

So, tagging isn's strictly necessary, but would be nice.

matklad:
>and if there's a version change

For my project, I use a stateless implementation of the idea -- new version is published if there isn't a git tag for the version currently specified in the manifest. This allows for workflows where you upgrade the version, publishing fails for some reason, you push a fix to publishing, and the publishing of the same version is retried. 


https://github.com/matklad/xshell/blob/4e5090e9f79baeed1037bc0925a6a84174566eb1/examples/ci.rs#L42-L90

miraclx:
> I wouldn't do "yank the >=1.0.0 versions from crates.io" right away. For that, I'd wait until the new version of the SDK is released, so that contract developers don't have yanked crates in the lockfiles.

Got it. Looking at the stable [`near-sdk:3.1.0`](https://crates.io/crates/near-sdk/3.1.0/dependencies) on crates.io, it uses `near-vm-logic:4.0.0-pre.1` so that's the version I'll preserve until the `4.0.0` release, yank everything else. Although the pre-version should, ideally, depend on the new `0.10.0` version when it's up.

> new version is published if there isn't a git tag for the version currently specified in the manifest. This allows for workflows where you upgrade the version, publishing fails for some reason, you push a fix to publishing, and the publishing of the same version is retried.

Yeah, that makes a lot of sense. The equivalent of this in `cargo-workspaces` is https://github.com/pksunkara/cargo-workspaces/blob/0cbcbccf537b834cc8116da0a892f3890ccb9d0d/cargo-workspaces/src/utils/cargo.rs#L396-L398 which is half-done. The logic is there, but tagging currently happens before publishing. I'll touch on that because it's definitely useful.

> `cargo publish` now actually includes current commit into the published crate automatically

Good to know. Would be nice for crates.io to actually link directly to the specific git hash instead of the homepage when you click repository. Then we wouldn't need tagging on our end. But, we'll go with it, for now.

miraclx:
I also think we should retain the `group` and `independent` features. So that crates can, if needed, opt-out from the general workspace versioning. In the case that certain crates absolutely need to deviate from the versioning scheme or needs a different publishing schedule. They can use these features.

Here's what I mean:

#### Workspace versioning

```toml
# Cargo.toml

[workspace.metadata.workspaces]
version = "0.10.0"
```


```toml
# Cargo.toml

[[workspace.metadata.workspaces.group]]
name = "chain"
version = "1.0.3"
members = [
  "./core/foo",
  "./core/bar"
]
```

**What this does?**
All crates belonging to this group are versioned to `1.0.3`, disregarding of the workspace version.

[[workspace.metadata.workspaces.group]]
name = "core"
members = [
  "./core/foo",
  "./core/bar"
]
```

**What this does?**
All crates belonging to this group are versioned to `1.0.3`, disregarding of the workspace version.

#### Package manifest
```toml
[package]
name = "foobar"
version = "2.2.5"

[package.metadata.workspaces]
independent = true
```

**What this does?**
This crate is versioned 

You can think of grouped crates as independent crates. But instead of each independent crate having its own version, grouping lets you say “these crates all have the same version”.

This means, while crates in a versioned group can depend on crates in the general workspace. The grouped crates won't be published until their version is bumped, at which point, they'll be using the latest version of the workspace crates.

Bumping an independent crate like `foobar` could be as simple `cargo ws version -p foobar minor` to go to `2.3.0`
Bumping all crates in a versioned group would be as simple as changing the version in the group spec.
Bumping all crates in an unversioned group would be as simple as `cargo ws version --groups chain`.

miraclx:
I also think we should retain the `group` and `independent` features. So that crates can, if needed, opt-out from the general workspace versioning. In the case where certain crates absolutely need to deviate from the general versioning scheme or needs a different publishing schedule, they can use these features.

<details>
<summary>Click to see full explanation</summary>

_(Just documenting some thoughts, you can skip over this)_

Here's what I mean:

#### Workspace versioning

```toml
# Cargo.toml

[workspace.metadata.workspaces]
version = "0.10.0"
```

```toml
# pkg/Cargo.toml

[package]
version = "0.0.0" # <-- this would be overriden by the workspace version
```

**What this does?**
All ungrouped crates are versioned to `0.10.0`.

**Updating the version?**
Update the version in the `Cargo.toml` manually.

#### Versioned Group

```toml
# Cargo.toml

[[workspace.metadata.workspaces.group]]
name = "core"
version = "1.0.3" # <-- 
members = [
  "./core/foo",
  "./core/bar"
]
```

```toml
# ./core/foo/Cargo.toml

[package]
version = "0.0.0" # <-- this would be overriden by the group version
```

**What this does?**
All crates belonging to this group are versioned to `1.0.3`, irrespective of the workspace version.

**Updating the version?**
Update the version in the `Cargo.toml` manually.

#### Unversioned Grouping

```toml
[[workspace.metadata.workspaces.group]]
name = "lorem"
members = [
  "./lorem",
  "./ipsum"
]
```

```toml
# ./lorem/Cargo.toml

[package]
version = "2.1.0" # <-- minor bump becomes 2.2.0
```

```toml
# ./lorem/Cargo.toml

[package]
version = "1.2.0" # <-- minor bump becomes 2.2.0
```

**What this does?**
All crates belonging to this group are versioned from a bump of the collective common version of members in the group, irrespective of the workspace version.

**Updating the version?**
`cargo ws version --groups lorem minor`

#### Independent Package Versioning

```toml
[package]
name = "foobar"
version = "2.2.5"

[package.metadata.workspaces]
independent = true
```

**What this does?**
This crate is explicitly independently versioned. Irrespective of the workspace version.

**Updating the version?**
`cargo ws version -p foobar minor`

### Relationship between groups

Crates can inter-depend on themselves across groups, but crates whose versions aren't bumped don't get their dependencies updated.

This means, while crates in a versioned group can depend on crates in the general workspace. The grouped crates won't be published until their version is explicitly bumped, at which point, they'll be using the latest version of the workspace crates.
</details>

matklad:
>I also think we should retain the group and independent features. So that crates can, if needed, opt-out from the general workspace versioning.

I'd default to YAGNI, and only implement this functionality when we actually need it, just in case we need something different. 

As an example, at some point in the future we might want to publish a generally useful `near-account-id 1.0.0` with manually maintained semantic version, which is intended to stay at 1.0.0 forever. In that world, a plausible alternative would be (this is something [we are considering for rust-analyzer](https://github.com/rust-analyzer/rust-analyzer/issues/10307)) to move the crate outside of the workspace and make sure that all internal crates depend on crates.io version of near-account-id. 

No strong opinion here, if group is easy to support, we might as well just do it.

austinabell:
> > I'll try to merge the crate into primitives to remove one thing we need to think about here.
> 
> I've reconsidered that. The list above misses `near-vm-logic`, but that is the crate we need to publish some version of, as the near-sdk-rs team needs it. cc @austinabell. The publishing is very much on the "just last one more time" basis though, we don't plant to support `near-vm-logic` to be usable outside of `nearcore`, current usages should migrate to sandbox instead.

We can't use sandbox for unit tests, though. We need `VMLogic` and associated types to be able to have fully functioning syscalls that match as close to actually running the wasm compiled code on chain. Are you suggesting that we copy all of this logic into the SDK and maintain it to match any protocol changes?

matklad:
I don't know what our plans for unit-testing. My current understanding is that we generally want to move **all** testing to sandbox, as that allows for much better fidelity, and is independent of a particular language the contract is implemented in. As such, I am acting under the assumption that we do want to support current unit-tests by publishing a frozen-in-time update of vm-logic, but that we aren't going to support it going forward.

nearmax:
We should consider phasing out unit testing. Blockchains like Ethereum don't have that so it does not seem necessary. Also, it will create a maintenance overhead in the future given that we are building RS testing framework. Without unit tests we might also be able to make Rust SDK simpler. For now we can separate tests into a separate published crate for which we stop doing updates and generate a warning when used, which should unblock the Contract Runtime team. When RS testing framework is released we can replace all unit tests in examples with RS framework tests. Eventually it would be great if unit tests crate will be evolving independently from Rust SDK and RS testing framework, which could even be something that is done by the community. WDYT?

chefsale (approved):

miraclx:
This is ready for review.

What's changed? TLDR:
- All crates are now directly unversioned (`v0.0.0`)
- There's a version spec in the workspace manifest that CI applies to all non-private crates at the time of publishing
  https://github.com/near/nearcore/blob/5553a4987dcf72cca7fae2556c8608f300062a57/Cargo.toml#L50-L51
- `near-vm-logic` is now marked publishable.
- All the functionality described here https://github.com/near/nearcore/pull/5130#issuecomment-961243685 is implemented, and available if needed.

matklad (approved):

nikurt (approved):

austinabell:
> We should consider phasing out unit testing. Blockchains like Ethereum don't have that so it does not seem necessary. Also, it will create a maintenance overhead in the future given that we are building RS testing framework. Without unit tests we might also be able to make Rust SDK simpler. For now we can separate tests into a separate published crate for which we stop doing updates and generate a warning when used, which should unblock the Contract Runtime team. When RS testing framework is released we can replace all unit tests in examples with RS framework tests. Eventually it would be great if unit tests crate will be evolving independently from Rust SDK and RS testing framework, which could even be something that is done by the community. WDYT?

I'm not sure I completely agree with this. Unit tests can be a great way to quickly test and avoid the pattern of `compile to wasm -> start up sandbox (if not using existing network) -> deploy contract -> send transactions through RPC` when it can be done all in a Rust test and potentially include many other architectures we currently don't support.

As for not having unit tests in solidity, there are some pseudo solutions like https://remix-ide.readthedocs.io/en/latest/unittesting.html, but I think it's important to compare against other Rust/wasm based VM in the ecosystem, and the ones I know about all support unit tests.

Maybe we can consider moving these test utils to a separate crate to that of the SDK, but it currently has no overhead for actual compilation because none are enabled for `wasm32-unknown-unknown`

core/primitives/res/runtime_configs
```diff
@@ -0,0 +1 @@
+../../../nearcore/res/runtime_configs
```
frol, line 1: Link might not work well on Windows :man_shrugging:

frol (approved):
I love it! :heart: 

@matklad Thanks for the fruitful discussion!

@miraclx Thanks for adding all that metadata to the Cargo.toml files and writing down the section to CONTRIBUTING.md!

core/primitives/res/runtime_configs
miraclx: Hm.. Can someone with Windows test this out. Previously we were referencing outside the package and while the local build worked, `cargo package` didn't bundle those artifacts. This symlink aids `cargo` in knowing to copy over those files when packaging, but I'm now unsure if the build works on Windows.

matklad: cc @Longarithm 

I'd maybe try to keep it simple and just move files themselves here.

Longarithm: Are we doing that to make `primitives-core` independent? If so, I agree with @matklad that `core/primitives/res/runtime_configs` is a better place for configs.

@miraclx could you move the files in this PR? If yes, could you also please split `nearcore/res/README.md` correspondingly?

frol: I tried that on Windows, and indeed it did not work well. I feel this primitives/runtime screams for being extracted into a separate crate since including all these json blobs does not feel right

bowenwang1996 (approved):

khorolets (approved):

pmnoxx:
A few questions?
- Who is going to be responsible to increasing the version number?
- Who will publish those creates?
- How often will they be publishes ?
- Is that going tto imply that we will have a requirement that packages api will not to change between versions?
-

miraclx:
> Who is going to be responsible to increasing the version number?

Anyone really, but since it goes through our familiar PR review process, we can ensure to only justifiably bump the version and not arbitrarily.

> Who will publish those creates?

Once this PR is merged, CI automation kicks in to auto-publish crates once it detects a change to the version number on master. So, for every PR that updates the version, when pushed to master, all non-private crates would in turn, get published with that version number.

> How often will they be publishes?

This is variable, there's no requirement to constantly turn out new versions, so we only need to publish when we actually need to. I suppose reasonable times are when we have protocol changes, the resolution of a critical bug involving the published crates or a security flaw is addressed.

> Is that going tto imply that we will have a requirement that packages api will not to change between versions?

Could you elaborate? I'm not sure I understood the question.

pmnoxx:
> > Is that going tto imply that we will have a requirement that packages api will not to change between versions?
> 
> Could you elaborate? I'm not sure I understood the question.

There is a general guideline in `create.io` that packages shouldn't break api if there is a minor version change. We are going to be publishing multiple crates at the same time. Strictly applying this rule to us, would hinder the development process.
My assumption is that we will not.

pmnoxx:
What is the review process of bumping version going to be. Are we going to require that for simple change such as bumping the version number up, we will require everyone from the company to approve?

@bowenwang1996 @mm-near @matklad @nearmax @miraclx Maybe we should give someone ownership to all code to make version changes easier?

pmnoxx:
Consider using a bot, that has permissions to approve if there is just a version change. Or maybe a similar solution @chefsale @mhalambek What can we do?

miraclx:
> > > Is that going tto imply that we will have a requirement that packages api will not to change between versions?
> > 
> > 
> > Could you elaborate? I'm not sure I understood the question.
> 
> There is a general guideline in `create.io` that packages shouldn't break api if there is a minor version change. We are going to be publishing multiple crates at the same time. Strictly applying this rule to us, would hinder the development process. My assumption is that we will not.

Thanks for clarifying, well because we have a unified version across the workspace, we can't maintain that rule on a per-crate basis. We're keeping things on `v0.x` indicating that we don't necessarily have a stable API. So, we arbitrarily publish with minor version bumps. Regardless of whether we have breaking changes. Anything that should have proper versioning can either opt-in to having an "independent" version from the workspace or be moved outside the workspace and managed independently.

pmnoxx:
I see a few possible solutions:
- Proposal A - every code owners has to approve
- Proposal B - give someone permissions to be code owners of everything  or make a GitHub bot for it
- Proposal C -  Don't bump up the version, but the script that published the change, bumps up the version temporarily, without committing, and it does cargo publish

miraclx:
> What is the review process of bumping version going to be. Are we going to require that for simple change such as bumping the version number up, we will require everyone from the company to approve?

Currently, with no codeowner for the workspace manifest, GitHub, I believe should only require approval from at least, one maintainer.

> Maybe we should give someone ownership to all code to make version changes easier?

Given the process, one would only need to own the workspace's Cargo.toml, but seeing as that spec dictates the versions of a number of non-private crates owned by sets of people, notifying those owners might be in order, but not as necessary since versioning and publishing is rather trivial.

miraclx:
> Proposal B - give someone permissions to be code owners of everything or make a GitHub bot for it

if a bot's job is only to detect and auto-approve changes to the version, it's no different from the usual approval from a maintainer. But if we stick to the manual approval, at least we get eyes on it.

> Proposal C - Don't bump up the version, but the script that published the change, bumps up the version temporarily, without committing, and it does cargo publish

I think this defeats the purpose of publishing automation, though.

pmnoxx:
> > Proposal B - give someone permissions to be code owners of everything or make a GitHub bot for it
> 
> if a bot's job is only to detect and auto-approve changes to the version, it's no different from the usual approval from a maintainer. But if we stick to the manual approval, at least we get eyes on it.
> 
> > Proposal C - Don't bump up the version, but the script that published the change, bumps up the version temporarily, without committing, and it does cargo publish
> 
> I think this defeats the purpose of publishing automation, though.

Ideally we should do:
Proposal D:
- script that goes through all files and bumps up versions
- one person approves the PR
- we merge, and then publish it

miraclx:
> script that goes through all files and bumps up versions

We have this already: a custom fork of `cargo-workspaces` https://github.com/pksunkara/cargo-workspaces/compare/master...miraclx:grouping-versioning-and-exclusion#files_bucket, but this only needs to happen on CI because it's only relevant to the publishing step.

In the repo, we keep all crates on `v0.0.0`. Relegating the versions to all be governed by the what's specified in the workspace manifest.

It's equivalent to what `rust-analyzer` [does](https://github.com/rust-analyzer/rust-analyzer/blob/14dff25107cec4473fde8c999256a3484a7ef1d0/crates/ide/Cargo.toml?#L3).

pmnoxx:
@miraclx My suggestion in general. Not including this PR is to split such PR into parts.
For example in this case:
- only the version change as 1 PR
- other changes as second PR.

If you don't mind I can do it for you. If you want to do it yourself, I can show you some good git commands to do it.

pmnoxx:
@miraclx Pt 1 https://github.com/near/nearcore/pull/5394

pmnoxx:
@miraclx @matklad  Pt 1 https://github.com/near/nearcore/pull/5394
Pt 2 https://github.com/near/nearcore/pull/5396
Pt 3 https://github.com/near/nearcore/pull/5395

This is just an example split. You don't have to use it. I just wanted to show how we can split code related changes from not code related changes. Feel free to use the orginal PR, or the split PR whatever you see fit.

I think in general it's a good practice we should split code related changes, from not code related changes.

pmnoxx:
@miraclx @matklad Let me know what you think about it.

pmnoxx (changes requested):
I recommend splitting code related changes from changes that aren't code related.

pmnoxx:
The reasons for split are:
- While having large PR, it's really easy to miss important code changes; That may result in changes not being reviewed properly.
- If an issue happens, it may be hard to pin point where the bug is.
- Issue will take a lot of more time for reviewers. It's hard to find out where the exact changes are. Some change may look like refactoring, but there is an actual behavior change, that the reviewer doesn't expect.
- If other PRs get merged in between. PR may require rewrite, and take a lot of time to fix due to merge conflicts.
- In issues involving multiple teams.
That can result in either one of a few cases:
a) each reviewer only reviews changes to his own part of code base, without fully reading the review. - There is no guarantee that a single reviewer will fully understand the change.
b) or each reviewer reads the review multiple times, even the code he is no familiar with. That requires much more time and afford than if he were to read the fully functional part of he is responsible for.
- It's difficult to hold a discussion about such PR. For example, if you have a few changes mixed into one PR, we will be using one discussion for all changes at once, that makes asking questions harder

Although, it's true that some changes may be hard to split apart. A last few large PRs I have seen, were written in such way that it was possible to divide those changes in logical components, such that each of them made a sense on their own.

In this PR, I see a few changes that seem independent:

- moving json files to a different folder - that change seems independent. If It was split, we would review it already, and for example, the merge conflicts we see in this PR wouldn't have happened. - In the future, when I look at git history, if that was a separate change, I would just look at the description of the PR, and I would know immediately, what it was about. It would be trivial to verify correctness of that change.
- workspace.metadata.workspaces changes. Same problem here. If this was a separate change. I would be able to ask questions about it. And I would see conversation history, and see what it was about. 

- adding license/repository/homepage/description 
- version change
- Adding docs
- Just version change.

Splitting those changed separately would be optional. It would be much easier to read separately, but It would be fine to ship them together. Those aren't changes that affect the code. This would be an easy change.

Each of those changes separately, could be something we ship separately, and would be easy to review. 

Anyway, those are my thought, I would like to hear other people opinions. Of course we need to keep common sense. The general rule of thumb would be to try to divide change into logical components. And then ask if they make sense of their own.
There are cases where over splitting is bad, and doing splitting just for sake of splitting is not a good idea. However, right now I get the feeling we are at the opposite end of the spectrum, and we should find some balance.

Also, if those changes were to be shipped individually, They would separately, involve smaller group of people, without having to get people from many teams involved. By the this PR was made, we could have already merged half of refactoring into the code base, and this change would trivial to review.

pmnoxx:
Disclaimer: This is not about making splitting mandatory, or forcing people to write small changes, but help developers be more productive, and to create a plan on how changes can be shipped safely, and in timely matter..

Some people like writing big changes, like I do. If that makes it easier for purpose of development that's acceptable. However, changes should be split for purpose of shipping. Arguments, like "not splitting, because it's hard or will create merge conflicts" are valid arguments. However, it's not hard to do it once you learn how to do it. And, we should offer people assistance with learning on how to use git commands; provide useful scripts that can make rebasing; splitting change easy; Strategies on how to write PRs in such a way that merge conflicts are not a real issue anymore.  In most cases it's possible to avoid merge conflicts, and if they appear easily fix it within a minute or so.

frol:
@pmnoxx Thanks for your inputs! I feel we are on the same page, and as @miraclx pointed out, the current implementation almost follows your proposal D (the order of version bump and PR is flitted to simplify the flow (it is easier to edit a single line in the project root Cargo.toml rather than needing to have a script / cargo-workspaces on every developer machine willing to bump the version)

> What is the review process of bumping version going to be. Are we going to require that for simple change such as bumping the version number up, to require everyone from the company to approve?

I proposed this strategy of publishing crates through a regular PR process + CI automation, and I believe we indeed don't need a heavy review process here, but at the same time, it allows the process to be controlled and fairly straightforward, i.e. one person proposes the version bump based on some need (via PR), and at least one other person approves it, thus there are at least two people in the loop.

bowenwang1996:
@pmnoxx I think this PR is fine as is. We don't always need to split a PR for the sake of making it easier to approve. Especially when it is about some renaming or version updates, I actually think it is easier to review it this way.

pmnoxx:
@mm-near Can you take a look?

pmnoxx (approved):

ailisp (approved):
Nice work and clear explanation in doc!

miraclx:
Happy to announce: the automation succeeded! We have everything published, and we have it tagged as a reference point!

https://github.com/near/nearcore/releases/tag/crates-0.10.0

matklad:
@miraclx thank you so much for persevering here!

PR #5132, opened 2021-11-03
runtime fuzz: fuzz epoch_length and gas_limit

posvyatokum:


test-utils/runtime-tester/src/fuzzing.rs
```diff
@@ -36,7 +36,11 @@ impl Arbitrary<'_> for Scenario {
         let mut scope = Scope::from_seeds(&seeds);
 
         let network_config = NetworkConfig { seeds };
-        let runtime_config = RuntimeConfig { max_total_prepaid_gas: GAS_1 * 100 };
+        let runtime_config = RuntimeConfig {
+            max_total_prepaid_gas: GAS_1 * 100,
+            gas_limit: (GAS_1 as f64 * *u.choose(&[0.01, 0.1, 1., 10., 100.])?) as u64,
+            epoch_length: *u.choose(&[5, 10, 100, 500])? as u64,
```
bowenwang1996, line 42: how are these numbers chosen?

bowenwang1996 (approved):

test-utils/runtime-tester/src/fuzzing.rs
posvyatokum: Intuitively, because I didn't have any ideas on how to do it properly. Open to any suggestions.

matklad (approved):
Looks reasonable to me!

But I don't know how to verify that the changes here actually improve our fuzzing coverage. Do we have some results here like "it now takes 2x less time to find this injected bug"?

posvyatokum:
With gas_limit changes I want to emulate congestion and with epoch_length I want to gas price to change. I've tested that both these scenarios now do occur during fuzzing.

PR #5134, opened 2021-11-04
component_nonce variable not updated properly on node restart.

### LINKED ISSUE #5133: Component nonce is not updated properly on node restart.
In our current implementation we store all edges removed to DB. Each set of edges is called `component` and has unique id `nonce`.

There is currently a bug, where `nonce` that's saved in database doesn't get updated properly.

If nonce is currently let's say `self.component_nonce = 10`.  In function `try_save_edges`,
which creates new component, we will increase this value by 1; save to database `component_nonce = 10` ;
assign new component to have id `10`.

If restart happens, we will load value `10` from database to `self.component_nonce`.
And whenever a new nonce gets created, we will repeat the procedure above. And we will assign another component to the same id.

pmnoxx:
Closes #5133 

After we remove edges we store them inside components, each with unique `component_nonce`. 
When that happens, we store that `component_nonce` on disk, increase it by 1 in memory, and then create new component with `component_nonce`.
However, if the node restarts we will load `component_nonce` value from disk, and when we remove edges, we will create another component with `component_nonce`.

That's a problem, because ids of components are no longer unique, and we will override previously stored component on disk. 

To solve this problem, we will change the logic to;
- use `component_nonce` for current component
- store `component_nonce+1` on disk
- increase `component_nonce` for future

### DIFF
--- a/chain/network/src/routing.rs
+++ b/chain/network/src/routing.rs
@@ -769,23 +769,23 @@ impl RoutingTable {
         }
         debug!(target: "network", "try_save_edges: We are going to remove {} peers", to_save.len());
 
-        let component_nonce = self.component_nonce;
+        let old_component_nonce = self.component_nonce;
         self.component_nonce += 1;
 
         let mut update = self.store.store_update();
-        let _ = update.set_ser(ColLastComponentNonce, &[], &component_nonce);
+        let _ = update.set_ser(ColLastComponentNonce, &[], &self.component_nonce);
 
         for peer_id in to_save.iter() {
             let _ = update.set_ser(
                 ColPeerComponent,
                 Vec::from(peer_id.clone()).as_ref(),
-                &component_nonce,
+                &old_component_nonce,
             );
 
             self.peer_last_time_reachable.remove(peer_id);
         }
 
-        let component_nonce = index_to_bytes(component_nonce);
+        let component_nonce = index_to_bytes(old_component_nonce);
         let mut edges_to_remove = vec![];
 
         self.edges_info.retain(|(peer0, peer1), edge| {
--- a/chain/network/tests/cache_edges.rs
+++ b/chain/network/tests/cache_edges.rs
@@ -206,7 +206,7 @@ fn load_component_nonce_on_start() {
     test.set_times(vec![(1, 2)]);
     test.update();
     let routing_table = RoutingTable::new(random_peer_id(), test.store.clone());
-    assert_eq!(routing_table.component_nonce, 1);
+    assert_eq!(routing_table.component_nonce, 2);
 }
 
 #[test]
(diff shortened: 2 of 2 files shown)


chain/network/src/routing.rs
```diff
@@ -769,23 +769,23 @@ impl RoutingTable {
         }
         debug!(target: "network", "try_save_edges: We are going to remove {} peers", to_save.len());
 
-        let component_nonce = self.component_nonce;
+        let old_component_nonce = self.component_nonce;
         self.component_nonce += 1;
 
         let mut update = self.store.store_update();
-        let _ = update.set_ser(ColLastComponentNonce, &[], &component_nonce);
+        let _ = update.set_ser(ColLastComponentNonce, &[], &self.component_nonce);
```
bowenwang1996, line 776: It seems that we are not actually storing last component nonce in the database, since that would be `old_component_nonce`. Should we rename the column?

bowenwang1996 (commented):
It would be beneficial to add in your PR description how this change fixes the issue you mentioned.

chain/network/src/routing.rs
pmnoxx: That's a good observation. Please take a look at https://github.com/near/nearcore/pull/5140/files and https://github.com/near/nearcore/pull/5089
I renamed this variable there to have the proper meaning.

pmnoxx:
> It would be beneficial to add in your PR description how this change fixes the issue you mentioned.

@bowenwang1996 I updated the description of the fchange.

pmnoxx:
@mm-near Can you take a look?

chain/network/src/routing.rs
```diff
@@ -769,23 +769,23 @@ impl RoutingTable {
         }
         debug!(target: "network", "try_save_edges: We are going to remove {} peers", to_save.len());
 
-        let component_nonce = self.component_nonce;
+        let old_component_nonce = self.component_nonce;
```
mm-near, line 772: I think that the original code didn't have the bug, but was very confusing..

If you look at the line 428:
self.component_nonce = 
.map_or(0, |nonce| nonce + 1);

So basically the approach was:
- the "nonce" that is stored in the database, is the last row/nonce that was used to store something there.
Then - on startup (like 428) - we read it from DB and increment.

which means that self.component_nonce is the next nonce that can be used.

mm-near (changes requested):
TL;DR: from what I see this is not a bug, but the code is a little bit confusing - and we might want to update the comments instead.

chain/network/src/routing.rs
pmnoxx: Ok, then we can close this PR.
Renaming `component_nonce` to `next_unused_component_nonce` should be enough.

PR #5137, opened 2021-11-04
Hotfix removing edges from `Graph` inside `RoutingTable`

### LINKED ISSUE #5135: Edges not removed properly during routing table pruning.
`RoutingTable::update` method is not removing edges from `Graph` while graph pruning.

```
        self.edges_info.retain(|(peer0, peer1), edge| {
            if to_save.contains(peer0) || to_save.contains(peer1) {
                edges_to_remove.push(edge.clone());
                false
            } else {
                true
            }
        });
```

pmnoxx:
Closes #5135

`RoutingTable::update` method is not removing edges from `Graph` while graph pruning.
We will use `remove_edges` method, which has proper implementation of removing edges.

### DIFF
--- a/chain/network/src/routing.rs
+++ b/chain/network/src/routing.rs
@@ -785,7 +785,7 @@ impl RoutingTable {
         }
 
         let component_nonce = index_to_bytes(component_nonce);
-        let mut edges_to_remove = self
+        let edges_to_remove = self
             .edges_info
             .iter()
             .filter_map(|((peer0, peer1), edge)| {
--- a/chain/network/src/routing.rs
+++ b/chain/network/src/routing.rs
@@ -598,7 +598,6 @@ impl RoutingTable {
         }
     }
 
-    #[cfg(feature = "test_features")]
     pub fn remove_edges(&mut self, edges: &Vec<Edge>) {
         for edge in edges.iter() {
             let key = (edge.peer0.clone(), edge.peer1.clone());
--- a/chain/network/src/routing.rs
+++ b/chain/network/src/routing.rs
@@ -786,16 +786,18 @@ impl RoutingTable {
         }
 
         let component_nonce = index_to_bytes(component_nonce);
-        let mut edges_to_remove = vec![];
-
-        self.edges_info.retain(|(peer0, peer1), edge| {
-            if to_save.contains(peer0) || to_save.contains(peer1) {
-                edges_to_remove.push(edge.clone());
-                false
-            } else {
-                true
-            }
-        });
+        let mut edges_to_remove = self
+            .edges_info
+            .iter()
+            .filter_map(|((peer0, peer1), edge)| {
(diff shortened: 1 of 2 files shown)


pmnoxx:

chain/network/src/routing.rs
```diff
@@ -598,7 +598,6 @@ impl RoutingTable {
         }
     }
 
-    #[cfg(feature = "test_features")]
```
bowenwang1996, line 601: Could you explain what happened here? Is `remove_edges` supposed to be a function that is only used for testing or not?

pmnoxx: @bowenwang1996 In the `try_save_edges(...)` method we have been removing edges from `self.edges_info` using `filter_map`. That is not correct, because we also need to remove edges from `self.raw_graph`.
Now, we remove those edges using `remove_edges` method. It was used previously for testing only, now it's used in production.

chain/network/src/routing.rs
```diff
@@ -786,16 +785,18 @@ impl RoutingTable {
...
-
-        self.edges_info.retain(|(peer0, peer1), edge| {
-            if to_save.contains(peer0) || to_save.contains(peer1) {
-                edges_to_remove.push(edge.clone());
-                false
-            } else {
-                true
-            }
-        });
+        let edges_to_remove = self
```
mm-near, line 788: the fact that this bug existed, means that we didn't have any test covering that.

When you do a hotfix like that (unless it is super-emergency etc) - you should also add a test case, that would be "failing" before this fix and then passing afterwards.


in our case, raw_graph is used to calculate distance, so a test that removes the edge and makes sure that the calculated distance is changed accordingly would be good (at the same time, it is strange that no other test has detected this problem -- do we know why ?

pmnoxx: @mm-near I already added check for that issue in the tests. Take a look at:
```
        let active_edges = on_memory
            .iter()
            .filter_map(|x| if x.2 == EdgeType::Added { Some(1) } else { None })
            .count();

        assert_eq!(active_edges, self.routing_table.raw_graph.total_active_edges as usize);**
```

mm-near: ah, right. (I have to get used having "tests" in the same file as production code).

Having said that -- do we know why no other test has caught it ? Do we have a one that removes the edge and recalculates the distances ?

mm-near (approved):

chain/network/src/routing.rs
pmnoxx: Yes, we have tests for that. We didn't check whenever ` self.routing_table.raw_grap` has correct number of edges.

mm-near: checking whether edges in graph (which is internal structure) are matching edge_list (another structure) is testing the "implementation" not the feature.

We should add (in separate PR then) - a test that does the edge removal & distance recalculation -- as this is the one that would have failed without this PR.

PR #5140, opened 2021-11-04
Refactor network code

pmnoxx:
We should refactor network code, before https://github.com/near/nearcore/pull/5089

- This refactoring doesn't contain any code changes, only methods/attributes are renamed to better names
- All changes are extracted from #5089, this was done to keep that PR smaller. This will cause merge conflicts, but it's better to extract syntax only changes to make #5089 easier to review.

Here is design doc for changes planned (https://docs.google.com/document/d/1FV3oyMRo52EQAgYroVd6Ws-5-ylkLSTdMOstY-IdL-M/edit?usp=sharing)

### DIFF
--- a/chain/network/src/peer_manager.rs
+++ b/chain/network/src/peer_manager.rs
@@ -405,8 +405,8 @@ impl PeerManagerActor {
         let target_peer_id = full_peer_info.peer_info.id.clone();
 
         let new_edge = Edge::new(
-            self.my_peer_id.clone(), // source
-            target_peer_id.clone(),  // target
+            self.my_peer_id.clone(),
+            target_peer_id.clone(),
             edge_info.nonce,
             edge_info.signature,
             full_peer_info.edge_info.signature.clone(),
--- a/chain/network/src/peer_manager.rs
+++ b/chain/network/src/peer_manager.rs
@@ -43,7 +43,7 @@ use crate::routing::{
     ProcessEdgeResult, RoutingTable, SimpleEdge, MAX_NUM_PEERS, SAVE_PEERS_AFTER_TIME,
 };
 
-use crate::edge_verifier::EdgeVerifier;
+use crate::edge_verifier::EdgeVerifierActor;
 #[cfg(feature = "test_features")]
 use crate::types::SetAdvOptions;
 use crate::types::{
@@ -62,7 +62,7 @@ use crate::types::{GetPeerId, GetPeerIdResult};
 use crate::types::{RoutingState, RoutingSyncV2, RoutingVersion2};
 
 /// How often to request peers from active peers.
-const REQUEST_PEERS_SECS: Duration = Duration::from_millis(60_000);
+const REQUEST_PEERS_INTERVAL: Duration = Duration::from_millis(60_000);
 /// How much time to wait (in milliseconds) after we send update nonce request before disconnecting.
 /// This number should be large to handle pair of nodes with high latency.
 const WAIT_ON_TRY_UPDATE_NONCE: Duration = Duration::from_millis(6_000);
@@ -124,7 +124,7 @@ pub struct PeerManagerActor {
     /// Networking configuration.
     config: NetworkConfig,
     /// Peer information for this node.
-    peer_id: PeerId,
+    my_peer_id: PeerId,
     /// Address of the client actor.
     client_addr: Recipient<NetworkClientMessages>,
     /// Address of the view client actor.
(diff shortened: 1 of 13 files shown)


bowenwang1996 (commented):
This PR description is not very informative. It does not answer the following:
* What exactly do we refactor here?
* Why do we need to do the refactoring?
* how is the refactor done on the code level?

pmnoxx:
> This PR description is not very informative. It does not answer the following:
> 
> * What exactly do we refactor here?

We only rename attributes / methods, those changes were extracted from #5089 to make it smaller.

> * Why do we need to do the refactoring?

- To keep #5089 smaller ; those methods / attributes had confusing meaning and we talked about renaming them.

> * how is the refactor done on the code level?

- I used automatic tools from CLion to do the renaming; to keep everything safe.

pmnoxx:
@mm-near Can you take a look?

chain/network/src/routing.rs
```diff
@@ -408,7 +408,7 @@ pub struct RoutingTable {
     /// Last nonce sent to each peer through pings.
     last_ping_nonce: SizedCache<PeerId, usize>,
     /// Last nonce used to store edges on disk.
-    pub component_nonce: u64,
+    pub next_available_component_nonce: u64,
```
mm-near, line 411: update the comment above

chain/network/src/routing.rs
```diff
@@ -444,11 +444,11 @@ impl RoutingTable {
             pong_info: SizedCache::with_size(PING_PONG_CACHE_SIZE),
             waiting_pong: SizedCache::with_size(PING_PONG_CACHE_SIZE),
             last_ping_nonce: SizedCache::with_size(PING_PONG_CACHE_SIZE),
-            component_nonce,
+            next_available_component_nonce: component_nonce,
```
mm-near, line 447: update the name in line 425 too

mm-near (approved):
Thanks a lot for splitting this into separate PR

chain/network/src/peer_manager.rs
```diff
@@ -347,14 +348,14 @@ impl PeerManagerActor {
...
-                act.routing_table_pool
+                act2.routing_table_pool
                     .send(RoutingTableMessages::AddPeerIfMissing(peer_id, None))
-                    .into_actor(act)
-                    .map(move |response, act2, _ctx| match response {
+                    .into_actor(act2)
+                    .map(move |response, act3, _ctx3| match response {
                         Ok(RoutingTableMessagesResponse::AddPeerResponse { seed }) => {
-                            act2.start_routing_table_syncv2(addr, seed)
+                            act3.start_routing_table_syncv2(addr, seed)
```
matklad, line 358: I'd personally just use `act` and `ctx` everywhere, as shadowing is idiomatic Rust, but no strong opinion.

chain/network/src/peer_manager.rs
```diff
@@ -404,8 +405,8 @@ impl PeerManagerActor {
         let target_peer_id = full_peer_info.peer_info.id.clone();
 
         let new_edge = Edge::new(
-            self.peer_id.clone(),   // source
-            target_peer_id.clone(), // target
```
matklad, line 408: Unrelated to the PR, but, rather than using comments, I'd extracted `let source` and `let = target` variables.

chain/network/src/peer_manager.rs
```diff
@@ -1022,7 +1030,7 @@ impl PeerManagerActor {
     ///  - bootstrap outbound connections from known peers,
     ///  - unban peers that have been banned for awhile,
     ///  - remove expired peers,
-    fn monitor_peers(&mut self, ctx: &mut Context<Self>) {
+    fn monitor_peers_trigger(&mut self, ctx: &mut Context<Self>) {
```
matklad, line 1033: :+1: adding a common suffix to "do thing in a loop" is very helpful, thanks!

matklad (approved):

chain/network/src/peer_manager.rs
pmnoxx: Thanks, that's a good idea.

pmnoxx:

bowenwang1996 (approved):

bowenwang1996:
@frol @khorolets looks like this PR requires your approval

pmnoxx:
@frol @khorolets @chefsale Can you take a look?

khorolets (approved):

pmnoxx:
@bowenwang1996 Build fails after rebase

PR #5141, opened 2021-11-04
Refactor PeerManager to group messages under one enum

### LINKED ISSUE #5142: Refactor PeerManager to move all messages under one enum
We are in progress of changing all actors in such a way, that all their messages will be under one enum.
This makes it easier to write unit tests, and is needed for https://github.com/near/nearcore/pull/4927

pmnoxx:
Closes #5142

We are going to group all messages in `PeerManager` under one enum called `PeerManagerMessageRequest`. `PeerManager` will have one single `Handler<...>` for messages. We will be able to extend this enum with extra paypoad that track sizes of all messages not processed yet, but coming from w given peer, which is needed for (https://github.com/near/nearcore/pull/4927), and make mocking PeerManager easier for test in the future.

### DIFF
--- a/chain/network/src/types.rs
+++ b/chain/network/src/types.rs
@@ -32,8 +32,6 @@ use near_primitives::views::QueryRequest;
 
 use crate::ibf::IbfBox;
 use crate::peer::Peer;
-#[cfg(feature = "test_features")]
-use crate::routing::SetAdvOptionsResult;
 use crate::routing::{
     Edge, EdgeInfo, GetRoutingTableResult, PeerRequestResult, RoutingTableInfo, SimpleEdge,
     ValidIBFLevel,
@@ -465,17 +463,20 @@ pub struct GetPeerIdResult {
     pub peer_id: PeerId,
 }
 
+#[derive(Debug)]
 pub struct GetRoutingTable {}
 
 impl Message for GetRoutingTable {
     type Result = GetRoutingTableResult;
 }
 
+#[derive(Clone, Debug)]
 #[cfg(feature = "test_features")]
 pub struct StartRoutingTableSync {
     pub peer_id: PeerId,
 }
 
+#[derive(Clone, Debug)]
 #[cfg(feature = "test_features")]
 pub struct SetAdvOptions {
     pub disable_edge_signature_verification: Option<bool>,
@@ -486,7 +487,7 @@ pub struct SetAdvOptions {
 
 #[cfg(feature = "test_features")]
 impl Message for SetAdvOptions {
-    type Result = SetAdvOptionsResult;
+    type Result = ();
 }
 
 #[cfg(feature = "test_features")]
@@ -502,7 +503,7 @@ pub enum ConsolidateResponse {
 }
 
 /// Unregister message from Peer to PeerManager.
-#[derive(Message)]
+#[derive(Message, Debug)]
 #[rtype(result = "()")]
 pub struct Unregister {
     pub peer_id: PeerId,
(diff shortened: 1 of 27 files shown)


mm-near:
Are there any "behavior" changes in this PR ? or just renaming things/moving things around ?

pmnoxx:
@mm-near No, there are no behavior changes.

bowenwang1996 (commented):
Could you explain why this is needed for #4927 ?

pmnoxx:
> Could you explain why this is needed for #4927 ?

@bowenwang1996 In #4927 we want to be able to suspend/resume reading from Peer's Tcp socket based on the number of messages that are in progress. 
1) To achieve this, we want to keep track of sizes of all messages that are in progress of being send from `PeerActor` or to `PeerActor`. The easiest way to achieve this is to add an extra information with each message. It will be way easier to add extra field to one enum, than having to modify all `Handler<...>` methods separately.
2) This will allow us to write unit tests for `Peer` if we want to. Currently we have no way to mock `PeerManager`. That's a separate goal that was discussed with @matklad

pmnoxx:
@mm-near Can you take a look?

chain/network/src/types.rs
```diff
@@ -525,27 +528,155 @@ impl Message for PeerRequest {
...
 /// Received new peers from another peer.
-#[derive(Message)]
+#[derive(Message, Debug, Clone)]
 #[rtype(result = "()")]
 pub struct PeersResponse {
     pub peers: Vec<PeerInfo>,
 }
 
+#[derive(Debug)]
+pub enum PeerMessageRequest {
```
mm-near, line 553: Can you add comment describing this enum (what it is used for etc)

chain/network/src/types.rs
```diff
@@ -525,27 +528,155 @@ impl Message for PeerRequest {
...
+        } else {
+            panic!("expected PeerMessageRequest::NetworkRequests(");
+        }
+    }
+
+    pub fn as_consolidate_response(self) -> ConsolidateResponse {
+        if let PeerMessageResponse::ConsolidateResponse(item) = self {
+            item
+        } else {
+            panic!("expected PeerMessageRequest::NetworkRequests(");
```
mm-near, line 643: btw - you didn't update the contents of these panic (as this one should be about ConsolidateResponse)

mm-near:
The code looks ok, but I'd like to understand more the "an extra information with each message".

Does this mean, that in the next step you'll change Handler<PeerMessageRequest> to something else ? (something else that also contains that additional message metadata ?)

I guess what I'm asking for, is a half-page google doc with some "draft" design of how the final thing is going to look.

pmnoxx:
> The code looks ok, but I'd like to understand more the "an extra information with each message".
> 
> Does this mean, that in the next step you'll change Handler to something else ? (something else that also contains that additional message metadata ?)
> 
> I guess what I'm asking for, is a half-page google doc with some "draft" design of how the final thing is going to look.
https://docs.google.com/document/d/1FV3oyMRo52EQAgYroVd6Ws-5-ylkLSTdMOstY-IdL-M/edit

chain/network/src/types.rs
pmnoxx: Fixed

chain/network/src/types.rs
pmnoxx: Done

bowenwang1996 (approved):

pmnoxx:
@frol @khorolets @chefsale Can you take a look?

khorolets (approved):

pmnoxx:
@mm-near Can you take a look?

pmnoxx:
@bowenwang1996 Code fails after rebase

PR #5151, opened 2021-11-05
refactor: simplify

matklad:
Just remove some extra characters :)


matklad:
Note: abstractly, it'd be better to use `&mut dyn io::Write` rather than `&mut File` as well (so that, eg, we can wrap File into BufWriter easily), but it doesn't really matter.

nikurt (approved):

PR #5153, opened 2021-11-05
Fix ViewClient::GetExecutionOutcome to work with sharding upgrade

### LINKED ISSUE #4977: Fix ViewClient::GetExecutionOutcome
Currently, the handler of `GetExecutionOutcome` returns outcome of an execution and the root proof that this outcome is confirmed by a following block. Each chunk header in a block includes the outcome root of outcomes produced in the previous chunk and the block header includes the root hash of all outcome roots from all chunks included in this block. Therefore, after a block process a receipt or transaction, its outcome is confirmed by the first block that includes a new chunk for the shard that originally processes the transaction/receipt, because a merkle proof can be generated to prove that the outcome is included in the outcome_root of that block. 

This is exactly what `GetExecutionOutcome` needs to do. It needs to find the first block with a new chunk for the shard that the outcome is generated. There are two issues with the current implementation

1) it does not take into account the fact that the number of shards may change between the block where the execution is performed and the block that outcome_root is included.
2) It looks for the next block with new chunk through two database columns `ColNextBlockWithNewChunk` and `ColLastBlockWithNewChunk `. These two columns assume a linear relationship between blocks, but they are updated when all blocks are processed, so they will not work when there are forks.

mzhangmzz:
fixes #4977 

Test plan:
- Modified sharding upgrade test to test `get_next_block_hash_with_new_chunk`, which is used by GetExecutionOutcome
- Existing CI and nayduck tests on GetExeuctionOutcome


chain/chain/src/chain.rs
```diff
@@ -2542,8 +2542,46 @@ impl Chain {
...
+                            next_shard_layout.get_split_shards(id).unwrap_or_else(|| {
+                                panic!("invalid shard layout {:?} because it does not contain split shards for parent shard {}", next_shard_layout, id)
+                            })
+                        })
+                        .collect::<Vec<_>>()
+                        .into_iter()
+                        .map(|shard_uid| shard_uid.shard_id())
+                        .collect();
+
+                    shard_layout = next_shard_layout;
```
bowenwang1996, line 2569: It seems that we have similar logic in some other places. Could we consider refactor this part into its own function?

chain/chain/src/chain.rs
```diff
@@ -2542,8 +2542,46 @@ impl Chain {
         &mut self,
         block_hash: &CryptoHash,
         shard_id: ShardId,
-    ) -> Result<Option<&CryptoHash>, Error> {
-        self.store.get_next_block_hash_with_new_chunk(block_hash, shard_id)
+    ) -> Result<Option<(CryptoHash, ShardId)>, Error> {
```
bowenwang1996, line 2545: Could we update the comment for this function to include what it does when the number of shards changes?

bowenwang1996 (commented):
It might be helpful to include an end-to-end test that checks transaction status from rpc.

chain/chain/src/chain.rs
```diff
@@ -2542,8 +2542,46 @@ impl Chain {
...
+                let next_shard_layout = self.runtime_adapter.get_shard_layout(&next_epoch_id)?;
+                if next_shard_layout != shard_layout {
+                    shard_ids = shard_ids
+                        .into_iter()
+                        .flat_map(|id| {
+                            next_shard_layout.get_split_shards(id).unwrap_or_else(|| {
+                                panic!("invalid shard layout {:?} because it does not contain split shards for parent shard {}", next_shard_layout, id)
+                            })
+                        })
+                        .collect::<Vec<_>>()
```
mm-near, line 2564: nit: can this cause duplicates in the vector ? 

(only if we allow shards to be merged - not sure if we do).

integration-tests/tests/client/sharding_upgrade.rs
```diff
@@ -179,6 +179,55 @@ impl TestShardUpgradeEnv {
         }
     }
 
+    fn check_next_block_with_new_chunk(&mut self, height: BlockHeight) {
```
mm-near, line 182: add comment

integration-tests/tests/client/sharding_upgrade.rs
```diff
@@ -179,6 +179,55 @@ impl TestShardUpgradeEnv {
...
+        let block = self.env.clients[0].chain.get_block_by_height(height).unwrap().clone();
+        let block_hash = block.hash();
+        let num_shards = block.chunks().len();
+        for shard_id in 0..num_shards {
+            let mut last_block_hash_with_empty_chunk = match self.env.clients[0]
```
mm-near, line 187: nit: this name (last_block_hash_with_empty_chunk) is not necesarily true, as it might be equal to block_hash..

integration-tests/tests/client/sharding_upgrade.rs
```diff
@@ -179,6 +179,55 @@ impl TestShardUpgradeEnv {
...
+                    .unwrap()
+                    .clone();
+                let chunks = last_block.chunks();
+                if chunks.len() == num_shards {
+                    assert_ne!(
+                        chunks.get(shard_id).unwrap().height_included(),
+                        last_block.header().height()
+                    );
+                } else {
+                    for chunk in chunks.iter() {
```
mm-near, line 222: this is a little bit unexpected (as you assume here that the block has no chunks at all, right?)

maybe add a comment

chain/chain/src/chain.rs
mzhangmzz: Yes, in the case of shards merging, this won't work. In fact, I didn't consider shard merging when I designed the shard layout API (you can tell from the fact that the function is called `split_shards` instead of something else.

integration-tests/tests/client/sharding_upgrade.rs
mzhangmzz: A block actually always contain chunks for all shards in the chunks vector. When there is no new chunk for a shard, it simply copies the chunk for this shard from the last block. That's why I'm checking chunk.height_included() != last_block.header().height() here, that means the height that this chunk was included is not equal to the block height, which means the chunk is not included in this block.

integration-tests/tests/client/sharding_upgrade.rs
mzhangmzz: I added more comments, hopefully that will make the meaning of this variable more clear.

core/primitives/src/shard_layout.rs
```diff
@@ -104,17 +104,19 @@ impl ShardLayout {
...
     #[inline]
-    pub fn get_split_shards(&self, parent_shard_id: ShardId) -> Option<Vec<ShardUId>> {
+    pub fn get_split_shard_uids(&self, parent_shard_id: ShardId) -> Option<Vec<ShardUId>> {
+        self.get_split_shard_ids(parent_shard_id).map(|shards| {
+            shards.into_iter().map(|id| ShardUId::from_shard_id_and_layout(id, &self)).collect()
+        })
+    }
+
+    #[inline]
+    pub fn get_split_shard_ids(&self, parent_shard_id: ShardId) -> Option<Vec<ShardId>> {
```
matklad, line 114: Feels like some methods here don't need to be `#[inline]`.

core/primitives/src/shard_layout.rs
```diff
@@ -104,17 +104,19 @@ impl ShardLayout {
...
         match self {
             Self::V0(_) => None,
             Self::V1(v1) => match &v1.shards_split_map {
                 Some(shards_split_map) => match shards_split_map.get(parent_shard_id as usize) {
```
matklad, line 118: no need for this match now: `=> hards_split_map.get(parent_shard_id as usize).cloneds()`

core/store/src/db.rs
```diff
@@ -83,8 +83,8 @@ pub enum DBCol {
     /// `LightClientBlock`s corresponding to the last final block of each completed epoch
     ColEpochLightClientBlocks = 26,
     ColReceiptIdToShardId = 27,
-    ColNextBlockWithNewChunk = 28,
-    ColLastBlockWithNewChunk = 29,
```
matklad, line 87: I'd expect to see a comment explaining `_`

core/primitives/src/shard_layout.rs
mzhangmzz: I'll clean those up

core/store/src/db.rs
mzhangmzz: Hmm actually just took the existing code `_ColTransactionRefCount` as an example. I am not sure what's the right way to delete/deprecate a column from the database

bowenwang1996: Don't think we have a process for that 😅 Feel free to propose something!

chain/chain/src/chain.rs
mzhangmzz: Also, right now, in the constructor, we check that there is no duplicates in the map. https://github.com/near/nearcore/blob/master/core/primitives/src/shard_layout.rs#L68

core/store/src/db.rs
mzhangmzz: I did a few things:
1) moved all the deprecated columns to the end of enum so it's easier to see all of them
2) added a database migration to remove data from deprecated columns

I think for now I am just going to keep the deprecated columns still in the database and in the codebase. We may want to drop the column family entirely, but I think that should be in another PR because it's relatively involved and I don't have time right now. I can open an issue for that if you think it's necessary.

matklad (approved):

core/store/src/db.rs
mzhangmzz: I actually revert the change on 2) because some test were broken and I couldn't quickly figure out why. I don't want to spend too much time on this PR right now so I'll just create an issue to track this

mzhangmzz:
> It might be helpful to include an end-to-end test that checks transaction status from rpc.

Agreed. I'll try to see if I can do that quickly. If not, I might just skip that. I think it should be fine because I didn't make much change to the handler of `GetExecutionOutcome,` the main change is on function `get_block_with_new_chunk`

bowenwang1996 (approved):

bowenwang1996:
@frol @khorolets this PR requires your approval

mm-near (approved):

khorolets (approved):

chefsale (approved):

PR #5154, opened 2021-11-05
Fix pytest/sanity/repro_2916.py

mzhangmzz:


mina86:
Sorry, I’m not seeing fix for the test. This only re-enables it and it’ll just start failing on NayDuck again.

mzhangmzz:
Not an urgent issue, but I think it's probably the best for us to set up a consistent python env in the future, instead of using whichever python on the machine. Otherwise, tests may break between version upgrades, such like this, and it could introduce some debugging confusion

mzhangmzz:
> Sorry, I’m not seeing fix for the test. This only re-enables it and it’ll just start failing on NayDuck again.

oops sorry don't know why it wasn't included, just updated it

mina86 (approved):

PR #5160, opened 2021-11-08
Add new features to RoutingTableActor

pmnoxx:
Add new features to routing table actor.

- do edge pruning (not used yet)
- store all edges in `edges_info`
-  `StartRoutingTableSync` (handling first message of protocol_feature_routing_exchange_algorithm)  (not yet used)
-  `RoutingTableUpdate` (updating routing table and maybe pruning edges) (not yet used)
-  `AddVerifiedEdges` - adds verified edges and updates stats (Similar to `AddEdges`, but does more) (not yet used)

This is done to split https://github.com/near/nearcore/pull/5089 into smaller pieces.
This should be reviewed after https://github.com/near/nearcore/pull/5159, as this PR also includes changes in that request.

### DIFF
--- a/chain/network/src/routing_table_actor.rs
+++ b/chain/network/src/routing_table_actor.rs
@@ -188,15 +188,22 @@ impl RoutingTableActor {
                             continue;
                         }
 
-                        // We remove mapping from `peer_id` to component_nonce else if
-                        // the mapping from `peer_id`, that we store in DB, points to the current nonce.
+                        // `edge = (peer_id, other_peer_id)` belongs to component that we loaded from database.
                         if let Ok(cur_nonce) = self.component_nonce_from_peer(peer_id.clone()) {
+                            // If `peer_id` belongs to current component
                             if cur_nonce == component_nonce {
+                                // Mark it as reachable and delete from database.
                                 self.peer_last_time_reachable
                                     .insert(peer_id.clone(), Instant::now() - SAVE_PEERS_MAX_TIME);
                                 update
                                     .delete(ColPeerComponent, Vec::from(peer_id.clone()).as_ref());
+                            } else {
+                                warn!("We expected `peer_id` to belong to component {}, but it belongs to {}",
+                                       component_nonce, cur_nonce);
                             }
+                        } else {
+                            warn!("We expected `peer_id` to belong to a component {}, but it doesn't belong anywhere",
+                                       component_nonce);
                         }
                     }
                     self.add_verified_edge(edge);
@@ -524,6 +531,7 @@ impl Handler<RoutingTableMessages> for RoutingTableActor {
                     edges_to_remove: edges_removed
                         .iter()
                         .filter(|p| p.contains_peer(&self.my_peer_id()))
+                        .cloned()
                         .collect(),
                     peer_forwarding: self.peer_forwarding.clone(),
                 }
--- a/chain/network/src/routing_table_actor.rs
+++ b/chain/network/src/routing_table_actor.rs
@@ -159,7 +159,7 @@ impl RoutingTableActor {
 
         // Update metrics after edge update
         near_metrics::inc_counter_by(&metrics::EDGE_UPDATES, total as u64);
-        near_metrics::set_gauge(&metrics::EDGE_ACTIVE, self.raw_graph.total_active_edges as i64);
+        near_metrics::set_gauge(&metrics::EDGE_ACTIVE, self.raw_graph.total_active_edges() as i64);
 
         edges
     }
@@ -219,7 +219,7 @@ impl RoutingTableActor {
     }
 
     fn my_peer_id(&self) -> &PeerId {
-        &self.raw_graph.source
+        &self.raw_graph.my_peer_id()
     }
 
     /// Recalculate routing table and update list of reachable peers.
--- a/chain/network/src/routing_table_actor.rs
+++ b/chain/network/src/routing_table_actor.rs
@@ -54,7 +54,8 @@ pub struct RoutingTableActor {
     /// Data structure used for exchanging routing tables.
     #[cfg(feature = "protocol_feature_routing_exchange_algorithm")]
     pub peer_ibf_set: IbfPeerSet,
-    /// Current view of the network. Nodes are Peers and edges are active connections.
+    /// Current view of the network represented by undirected graph.
+    /// Nodes are Peers and edges are active connections.
     pub raw_graph: Graph,
     /// Active PeerId that are part of the shortest path to each PeerId.
     pub peer_forwarding: Arc<HashMap<PeerId, Vec<PeerId>>>,
--- a/chain/network/src/routing_table_actor.rs
+++ b/chain/network/src/routing_table_actor.rs
@@ -115,7 +115,8 @@ impl RoutingTableActor {
         }
     }
 
-    /// `add_verified_edge` add edges that have already been verified.
+    /// `add_verified_edge` adds edges, for which we already that theirs signatures
+    /// are valid (`signature0`, `signature`).
     fn add_verified_edge(&mut self, edge: Edge) -> bool {
         let key = edge.get_pair();
         if !self.is_edge_newer(&key, edge.nonce) {
--- a/chain/network/src/routing_table_actor.rs
+++ b/chain/network/src/routing_table_actor.rs
@@ -279,7 +279,7 @@ impl RoutingTableActor {
         }
         debug!(target: "network", "try_save_edges: We are going to remove {} peers", peers_to_remove.len());
 
-        let old_component_nonce = self.next_available_component_nonce;
+        let current_component_nonce = self.next_available_component_nonce;
         self.next_available_component_nonce += 1;
 
         let mut update = self.store.store_update();
@@ -292,13 +292,13 @@ impl RoutingTableActor {
             let _ = update.set_ser(
                 ColPeerComponent,
                 Vec::from(peer_id.clone()).as_ref(),
-                &old_component_nonce,
+                &current_component_nonce,
             );
 
             self.peer_last_time_reachable.remove(peer_id);
         }
 
-        let component_nonce = index_to_bytes(old_component_nonce);
+        let component_nonce = index_to_bytes(current_component_nonce);
         let edges_to_remove = self
             .edges_info
             .iter()
--- a/chain/network/src/routing_table_actor.rs
+++ b/chain/network/src/routing_table_actor.rs
@@ -62,7 +62,7 @@ pub struct RoutingTableActor {
     pub peer_last_time_reachable: HashMap<PeerId, Instant>,
     /// Everytime a group of peers becomes unreachable at the same time; We store edges belonging to
     /// them in components. We remove all of those edges from memory, and save them to database,
-    /// If any of them were to be reachable again, we would re-add them.
+    /// If any of them become reachable again, we re-add whole component.
     ///
     /// To store components, we have following column in the DB.
     /// ColLastComponentNonce -> stores component_nonce: u64, which is the lowest nonce that
--- a/chain/network/src/routing_table_actor.rs
+++ b/chain/network/src/routing_table_actor.rs
@@ -221,7 +221,7 @@ impl RoutingTableActor {
     pub fn recalculate_routing_table_and_maybe_prune_edges(
         &mut self,
         prune: Prune,
-        prune_edges_after: Duration,
+        prune_edges_not_reachable_for: Duration,
     ) -> Vec<Edge> {
         #[cfg(feature = "delay_detector")]
         let _d = DelayDetector::new("routing table update".into());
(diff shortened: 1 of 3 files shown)


chain/network/src/routing_table_actor.rs
```diff
@@ -14,22 +16,355 @@ use crate::ibf::{Ibf, IbfBox};
...
+use near_store::{Store, StoreUpdate};
+use std::sync::Arc;
+use std::time::{Duration, Instant};
 
-/// Actor that maintains routing table information.
-/// TODO (PIOTR, #4859) Finish moving routing table computation to new thread.
-#[derive(Default)]
+#[derive(Debug, Eq, PartialEq)]
+pub enum Prune {
+    Allow,
```
mm-near, line 35: can you add comment explaining the difference between Allow and Force

chain/network/src/routing_table_actor.rs
```diff
@@ -14,22 +16,355 @@ use crate::ibf::{Ibf, IbfBox};
...
+    /// Data structure with all edges.
+    pub edges_info: HashMap<(PeerId, PeerId), Edge>,
     /// Data structure used for exchanging routing tables.
     #[cfg(feature = "protocol_feature_routing_exchange_algorithm")]
     pub peer_ibf_set: IbfPeerSet,
+    /// Current view of the network. Nodes are Peers and edges are active connections.
+    pub raw_graph: Graph,
+    /// Active PeerId that are part of the shortest path to each PeerId.
+    pub peer_forwarding: Arc<HashMap<PeerId, Vec<PeerId>>>,
+    /// Last time a peer was reachable through with active edges.
```
mm-near, line 61: remove 'with'

chain/network/src/routing_table_actor.rs
```diff
@@ -14,22 +16,355 @@ use crate::ibf::{Ibf, IbfBox};
...
+    pub peer_last_time_reachable: HashMap<PeerId, Instant>,
+    /// Everytime a group of peers becomes unreachable at the same time; We store edges belonging to
+    /// them in components. We remove all of those edges from memory, and save them do database,
```
mm-near, line 64: s/do/to

chain/network/src/routing_table_actor.rs
```diff
@@ -14,22 +16,355 @@ use crate::ibf::{Ibf, IbfBox};
...
+    /// If any of them were to be reachable again, we would re-add them.
```
mm-near, line 65: If any of them become reachable again, we re-add whole component.

chain/network/src/routing_table_actor.rs
```diff
@@ -14,22 +16,355 @@ use crate::ibf::{Ibf, IbfBox};
...
 pub struct RoutingTableActor {
-    /// Data structures with all edges.
-    edges_info: HashMap<(PeerId, PeerId), Edge>,
+    /// Data structure with all edges.
+    pub edges_info: HashMap<(PeerId, PeerId), Edge>,
     /// Data structure used for exchanging routing tables.
     #[cfg(feature = "protocol_feature_routing_exchange_algorithm")]
     pub peer_ibf_set: IbfPeerSet,
+    /// Current view of the network. Nodes are Peers and edges are active connections.
+    pub raw_graph: Graph,
```
mm-near, line 58: this is the undirected graph, right?

chain/network/src/routing_table_actor.rs
```diff
@@ -14,22 +16,355 @@ use crate::ibf::{Ibf, IbfBox};
...
+///   - routing information (where a message should be send to reach given peer)
+///  
+/// We use store for following reasons:
+///   - store removed edges to disk
+///   - we currently don't store active edges to disk
 pub struct RoutingTableActor {
-    /// Data structures with all edges.
-    edges_info: HashMap<(PeerId, PeerId), Edge>,
+    /// Data structure with all edges.
+    pub edges_info: HashMap<(PeerId, PeerId), Edge>,
```
mm-near, line 53: and here we guarantee that first peer < second peer, right?

chain/network/src/routing_table_actor.rs
```diff
@@ -14,22 +16,355 @@ use crate::ibf::{Ibf, IbfBox};
...
+        self.peer_ibf_set.remove_edge(&edge.to_simple_edge());
+
+        let key = (edge.peer0.clone(), edge.peer1.clone());
+        if self.edges_info.remove(&key).is_some() {
+            self.raw_graph.remove_edge(&edge.peer0, &edge.peer1);
+            self.needs_routing_table_recalculation = true;
+        }
+    }
+
+    /// `add_verified_edge` add edges that have already been verified.
```
mm-near, line 117: what does 'verified' mean?

mm-near: also mention that it returns true if edge was added succesfully.

you can  also remove the `add_verified_edge` from the beginning of the comment

chain/network/src/routing_table_actor.rs
```diff
@@ -14,22 +16,355 @@ use crate::ibf::{Ibf, IbfBox};
...
+            if let Ok(edges) = self.get_and_remove_component_edges(component_nonce, &mut update) {
+                for edge in edges {
+                    for &peer_id in vec![&edge.peer0, &edge.peer1].iter() {
+                        if peer_id == &my_peer_id
+                            || self.peer_last_time_reachable.contains_key(peer_id)
+                        {
+                            continue;
+                        }
+
+                        // We remove mapping from `peer_id` to component_nonce else if
```
mm-near, line 189: this comment is a little bit confusing.

I'd write something like this:

(other_peer_id, peer_id) is the edge that belongs to the component that we've recovered from disk. Now we're checking what happened to peer_id. If it was also the part of the same component - great, let's mark it as reachable and remove the reference in DB.
If it was a part of another component - let's not touch it.

chain/network/src/routing_table_actor.rs
```diff
@@ -14,22 +16,355 @@ use crate::ibf::{Ibf, IbfBox};
...
+
+    /// Recalculate routing table and update list of reachable peers.
+    /// If pruning is enabled we will remove unused edges and store them to disk.
+    ///
+    /// # Returns
+    /// List of edges removed.
+    pub fn recalculate_routing_table(
+        &mut self,
+        prune: Prune,
+        prune_edges_after: Duration,
```
mm-near, line 224: could we add this prune_edges_after to 'Allow' enum from Prune?

chain/network/src/routing_table_actor.rs
```diff
@@ -14,22 +16,355 @@ use crate::ibf::{Ibf, IbfBox};
...
+            .collect::<HashSet<_>>();
+
+        // Save nodes on disk and remove from memory only if elapsed time from oldest peer
+        // is greater than `SAVE_PEERS_MAX_TIME`
+        if !force_pruning && now.duration_since(oldest_time) < SAVE_PEERS_MAX_TIME {
+            return Vec::new();
+        }
+        debug!(target: "network", "try_save_edges: We are going to remove {} peers", peers_to_remove.len());
+
+        let old_component_nonce = self.next_available_component_nonce;
```
mm-near, line 282: nit: I'd call it 'current_component_nonce'

chain/network/src/routing_table_actor.rs
```diff
@@ -50,28 +385,44 @@ impl Handler<StopMsg> for RoutingTableActor {
...
-}
-
+// Messages for RoutingTableActor
 #[derive(Debug)]
 pub enum RoutingTableMessages {
+    // Add verified edges to routing table actor (Will be removed soon).
     AddEdges(Vec<Edge>),
+    // Remove edges. (Will be removed soon).
     RemoveEdges(Vec<Edge>),
+    // Add verified edges to routing table actor and update stats.
```
mm-near, line 395: explain what 'verified' means

chain/network/src/routing_table_actor.rs
```diff
@@ -129,25 +486,32 @@ impl Handler<RoutingTableMessages> for RoutingTableActor {
...
+            }
+            RoutingTableMessages::RoutingTableUpdate { prune, prune_edges_after } => {
+                let edges_to_remove = if self.needs_routing_table_recalculation {
+                    self.recalculate_routing_table(prune, prune_edges_after)
+                } else {
+                    Vec::new()
+                };
+                self.needs_routing_table_recalculation = false;
+                RoutingTableMessagesResponse::RoutingTableUpdateResponse {
+                    edges_to_remove,
```
mm-near, line 511: as these are 'edges_TO_remove', who is doing the 'actual removing'?

as I see that we're leaving only AdvRemoveEdges as a way to remove stuff. What am I missing?

mm-near (changes requested):

chain/network/src/routing_table_actor.rs
pmnoxx: // Add verified edges to routing table actor and update stats.
    // Each edge contains signature of both peers to which given edge belongs to.
    // We say that the edge is "verified" if and only if we checked that the `signature0` and
    // `signature1` is valid.

pmnoxx: // Add verified edges to routing table actor and update stats.
    // Each edge contains signature of both peers.
    // We say that the edge is "verified" if and only if we checked that the `signature0` and
    // `signature1` is valid.

chain/network/src/routing_table_actor.rs
pmnoxx: I renamed the above function name to `recalculate_routing_table_and_maybe_prune_edges`.

The only way to remove edges from routing table is by pruning, which is only done while doing routing table update.
Edge pruning used to be done by `PeerManager`, so it was telling `RoutingTableActor` to remove edges. Now, the situation is reversed.

chain/network/src/routing_table_actor.rs
pmnoxx: Well, the name is confusing. We prune edges that were not rechable for `prune_edges_after` duration.
`Force` is used for testing, it runs pruning every time.
`Allow` only does pruning at most 1 per hour.

I renamed `prune_edges_after` to `prune_edges_not_reachable_for`.

chain/network/src/routing_table_actor.rs
pmnoxx: Yes.

    /// Current view of the network represented by undirected graph.
    /// Nodes are Peers and edges are active connections.

chain/network/src/routing_table_actor.rs
pmnoxx: Yes. I updated comment to
```
    /// Data structure with all edges. It's guaranteed that `peer.0` < `peer.1`.
```

chain/network/src/routing_table_actor.rs
pmnoxx: I renamed variants to become self explanatory:
```
pub enum Prune {
    PruneOncePerHour,
    PruneNow,
    Disable,
}
```

chain/network/src/routing_table_actor.rs
pmnoxx: I meant that we checked their signatures. I updated comment to be 
```rust
    /// `add_verified_edge` adds edges, for which we already that theirs signatures
    /// are valid (`signature0`, `signature`).
```

chain/network/src/routing_table_actor.rs
pmnoxx: Ok, I updated the code. I also added warnings, because other cases shouldn't happen unless node crashes in middle of update.

mm-near (approved):

pmnoxx:
@chefsale @frol @khorolets Can you take a look?

chefsale (approved):

chain/network/src/routing_table_actor.rs
```diff
@@ -14,22 +16,367 @@ use crate::ibf::{Ibf, IbfBox};
...
+pub enum Prune {
+    PruneOncePerHour,
+    PruneNow,
+    Disable,
+}
+
+/// RoutingTableActor that maintains routing table information. We currently have only one
+/// instance of this actor.
+///
+/// We store following information
```
bowenwang1996, line 43:
```suggestion
/// We store the following information:
```

chain/network/src/routing_table_actor.rs
```diff
@@ -14,22 +16,367 @@ use crate::ibf::{Ibf, IbfBox};
...
+use near_store::db::DBCol::{ColComponentEdges, ColLastComponentNonce, ColPeerComponent};
+use near_store::{Store, StoreUpdate};
+use std::sync::Arc;
+use std::time::{Duration, Instant};
 
-/// Actor that maintains routing table information.
-/// TODO (PIOTR, #4859) Finish moving routing table computation to new thread.
-#[derive(Default)]
+#[derive(Debug, Eq, PartialEq)]
+pub enum Prune {
```
bowenwang1996, line 34: could we add a comment here on what we use this enum for?

bowenwang1996 (approved):

PR #5162, opened 2021-11-08
Add extra checks to `Graph` and more comments

pmnoxx:
We will add extra checks to make sure `RoutingTable` updates `Graph` properly. In addition I added documentation for `Graph`.

### DIFF
--- a/chain/network/src/routing.rs
+++ b/chain/network/src/routing.rs
@@ -437,7 +437,7 @@ impl RoutingTable {
     }
 
     fn my_peer_id(&self) -> &PeerId {
-        &self.raw_graph.source
+        &self.raw_graph.my_peer_id
     }
 
     pub fn reachable_peers(&self) -> impl Iterator<Item = &PeerId> {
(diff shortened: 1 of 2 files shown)


chain/network/src/routing.rs
```diff
@@ -876,23 +876,34 @@ pub struct RoutingTableInfo {
     pub peer_forwarding: HashMap<PeerId, Vec<PeerId>>,
 }
 
+/// `Graph` is used to compute `peer_routing`, which contains information how to route messages to
```
mm-near, line 879: Thank you for adding documentation, it helps a lot.

chain/network/src/routing.rs
```diff
@@ -876,23 +876,34 @@ pub struct RoutingTableInfo {
...
+/// all known peers. That is, for each `peer`, we get a sub-set of peers to which we are connected
+/// to that are on the shortest path between us as destination `peer`.
 #[derive(Clone)]
 pub struct Graph {
-    pub source: PeerId,
+    /// peer_id of current peer
+    my_peer_id: PeerId,
+    /// `id` as integer corresponding to `my_peer_id`
```
mm-near, line 886: can you also mention why do we need the u32 ids at all? is it for performance reasons?

chain/network/src/routing.rs
```diff
@@ -876,23 +876,34 @@ pub struct RoutingTableInfo {
...
     p2id: HashMap<PeerId, u32>,
+    /// List of existing `PeerId`s
     id2p: Vec<PeerId>,
-    pub used: Vec<bool>,
-    pub unused: Vec<u32>,
+    /// Which ids are currently in use
+    used: Vec<bool>,
+    /// List of unused peer ids
+    unused: Vec<u32>,
+    /// Compressed adjacency table, we use 32 bit integer as ids instead of using full `PeerId`.
```
mm-near, line 896: btw - is our graph directed or undirected?
do we store edges in both directions?

chain/network/src/routing.rs
```diff
@@ -909,6 +920,14 @@ impl Graph {
         res
     }
 
+    pub fn total_active_edges(&self) -> u64 {
+        self.total_active_edges
+    }
+
+    pub fn compute_total_active_edges(&self) -> u64 {
```
mm-near, line 927: naming is a little bit confusing -- as 'total_active_edges' returns X, while 'compute_total_active_edges' returns 2*X..

mm-near (approved):

chain/network/src/routing.rs
pmnoxx: Yes, I rewrote this code in order to use u32 to improve performance.

chain/network/src/routing.rs
pmnoxx: Yes, that's an undirected graph, we store edges in both directions.

chain/network/src/routing.rs
pmnoxx: Ok, I simplified the code to return the same value.

Issue #5138, opened 2021-11-04
Move EdgeVerifier from `PeerManager` to `RoutingTableActor


pmnoxx:
We should finish cleanup, and move `EdgeVerifier` to `RoutingTableActor`.

pmnoxx:
After we do this, it will be easy to fix https://github.com/near/nearcore/issues/5136

PR #5163, opened 2021-11-08
Move code for starting new routing table sync to RoutingTableActor

pmnoxx:
In preparation for moving routing table computation to `RoutingTableActor`, we should move  handling `start_routing_table_syncv2` from `PeerManager` to `RoutingTableActor` first.

Related to https://github.com/near/nearcore/issues/5138
Change extracted from https://github.com/near/nearcore/pull/5089, to make that PR shorter.

### DIFF
--- a/chain/network/src/routing_table_actor.rs
+++ b/chain/network/src/routing_table_actor.rs
@@ -19,7 +19,7 @@ use crate::routing::Edge;
 use crate::routing::{SimpleEdge, ValidIBFLevel, MIN_IBF_LEVEL};
 use crate::types::StopMsg;
 #[cfg(feature = "protocol_feature_routing_exchange_algorithm")]
-use crate::types::{PartialSync, RoutingState, RoutingVersion2};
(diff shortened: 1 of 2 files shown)


pmnoxx:

chain/network/src/routing_table_actor.rs
```diff
@@ -68,6 +68,10 @@ pub enum RoutingTableMessages {
         peer_id: PeerId,
         ibf_msg: RoutingVersion2,
     },
+    #[cfg(feature = "protocol_feature_routing_exchange_algorithm")]
+    StartRoutingTableSync {
```
mm-near, line 72: let's start improving the documentation:

can you add the description of what will happen when this message is received?

mm-near (approved):

chain/network/src/routing_table_actor.rs
pmnoxx: Yes, once we merge other PR changes I'll add documentation.

PR #5167, opened 2021-11-08
fix: require that wasm target is installed via toolchain file

### LINKED ISSUE #5082: Check if adding wasm32-unknown-unknown target to rust-toolchain.toml can simplify build
See https://github.com/near/nearcore/pull/5081#issuecomment-951141224

We get weird CI failure during build sometimes. I *think* this might be due to my clever use of `rustup` to install wasm target in near-test-contracts (although need to check this). If that's the case, modern rustup includes a feature to require a target being pre-installed, so we can try to use that.

matklad:
This commit replaces self-installation of wasm toolchain with just
asking rustup to have it installed (which is a relatively new rustup
feature). I think this should fix mysterious CI failures we've seen
recently, which I think due to rustup not being concurrently-safe.

closes #5082


matklad:
@chefsale please take a look

chefsale (approved):
LGTM

ailisp (approved):
Nice! TIL rust-toolchain.toml

Issue #5174, opened 2021-11-08
Implement near-time - a near create for time


pmnoxx:
As discussed with @matklad, we discussed implementing our of crate for time. 

Currently we use 4 methods to represent time `chrono` (`Utc::now()`) (`Datetime<Utc>`) , `SystemTime`, `Instant::now`, `u64`, they all have their issues.

- We don't want to use external `chrono` library, mainly because we don't really care about timezones
- `Instant` is written in such a way, that it's not possible to convert it to Unix timestamp. 
- `SystemTime` works, but it's not monitonic. It's possible that if we measure time at time `a` then at time `b`; the duration `b-a` will be negative, causing our code to fail.
- We want to use a saturating `Duration`, so any duration such that `a < b`, created by `a - b`, will default to `0`, if the difference were to become nagative. This will reduce chances of any crashed happening, when negative time is not expected.

We are going to create our own wrapper around `SystemTime` and `Duration`. This will allow us to do the mocking properly.

- [x] pt 1 - https://github.com/near/nearcore/pull/5177
- [ ] stop using `pub use` for external crates like UTC - that's a bad practice. 
- [ ] PR for near-network
- [ ] following PRs

pmnoxx:
@matklad @bowenwang1996 @mm-near @yordanmadzhunkov

bowenwang1996:
Please sync with @yordanmadzhunkov to avoid duplication of work

pmnoxx:
@bowenwang1996 I'll discuss it tomorrow with @yordanmadzhunkov during a meeting.

pmnoxx:
@bowenwang1996 This is an independent issue. I'll chose the issue for now. First we need to finish implementing mocking for time.

PR #5175, opened 2021-11-08
Replace chrono with near-create inside near-network

pmnoxx:
As discussed with @matklad, we discussed implementing our of crate for time. 

Currently we use 4 methods to represent time `chrono` (`Utc::now()`) (`Datetime<Utc>`) , `SystemTime`, `Instant::now`, `u64`, they all have their issues.

- We don't want to use external `chrono` library, it contains timezone information, we only care about `Utc` time
- `Instant` is written in such a way, that it's not possible to convert it to unix timestamp. However, it's monotonic, which is good.
- `SystemTime` works, but it's not monitonic. It's possible that if we measure time at time `a` then at time `b`; the duration `b-a` will be negative, causing our code to fail.
- We want to have a way to mock time. We can't use global instance, because tests run in parallel.
- We want to have a single source of `XXX::now()`, which we can mock. Ideally, we would be able to ensure that ::now() is correct, by grepping code and checking that we see `NearClock::now()` in all places that time is used. This will make it easy to check that we use correct time anywhere.
- We want to use a saturating `Duration`, so any duration such that `a < b`, created by `a - b`, will default to `0`, if the difference were to become nagative. This will reduce chances of any crashed happening, when negative time is not expected.

We are going to create our own wrapper around `SystemTime` and `Duration`. This will allow us to do the mocking properly.

Note:
If we decide to ship it, I'll split this PR into two parts:
- Adding NearClock (only changes in /utils/near-clock) - pt 1 (https://github.com/near/nearcore/pull/5177)
- Making changes for near-network to use NearClock

See https://github.com/near/nearcore/issues/5174


utils/near-clock/src/lib.rs
```diff
@@ -0,0 +1,118 @@
+use std::cell::RefCell;
+use std::ops::{Add, Mul, Sub};
+use std::sync::atomic::{AtomicU64, Ordering};
+use std::sync::Arc;
+use std::time::{Duration, SystemTime};
+
+thread_local! {
+    // A shared pointer. We should propagate it across multiple threads.
+    // Replace by AtomicU128 by 2262-04-12, after that date 64 bits won't be enough to
```
matklad, line 9: :rofl:

utils/near-clock/src/lib.rs
```diff
@@ -0,0 +1,118 @@
...
+    pub const fn from_system_time(system_time: SystemTime) -> NearClock {
+        NearClock { time: system_time }
+    }
+
+    pub fn now() -> NearClock {
+        now()
+    }
+
+    pub fn duration_since(&self, rhs: &NearClock) -> Duration {
+        self.time.duration_since(rhs.time).unwrap_or(Duration::from_millis(0))
```
matklad, line 64:
```suggestion
        self.time.duration_since(rhs.time).unwrap_or_default()
```

A bit less explicit, but more idiomatic and minimal. For pure Rust code-bases, I think "use the most direct solution within the language" is a useful consistency guideline.

matklad (commented):
Hm, given that what we want to do here is a "proper" mocked clock, I think it makes sense to try to avoid thread-locals, and use proper dependency injection. That is, to have two types: a clock (which returns time) and a time (which doesn't have a global `now` associated function): 


```rust
#[derive(Clone)]
pub struct Clock {
    repr: ClockRepr,
}

#[derive(Clone)]
enum ClockRepr {
    Real,
    Mocked { now: Arc<AtomicU64> },
}

#[derive(Clone, Copy)]
pub struct Time {
    inner: SystemTime,
}

impl Clock {
    pub fn real_clock() -> Clock {
        Clock { repr: ClockRepr::Real }
    }

    pub fn mocked_clock() -> (Clock, Arc<AtomicU64>) {
        let now = Arc::new(AtomicU64::new(0));
        let clock = Clock { repr: ClockRepr::Mocked { now: Arc::clone(&now) } };
        (clock, now)
    }

    pub fn now(&self) -> Time {
        let inner = match &self.repr {
            ClockRepr::Real => SystemTime::now(),
            ClockRepr::Mocked { now } => {
                std::time::UNIX_EPOCH + Duration::from_nanos(now.load(Ordering::Relaxed))
            }
        };
        Time { inner }
    }
}

impl Time {
    #[deprecated(
        reason = "deliberately not implementing, use `now` method on the appropriate `Clock` instance."
    )]
    pub fn now() -> Time {
        unimplemented!(
            "deliberately not implementing, use `now` method on the appropriate `Clock` instance."
        )
    }

    pub fn duration_since(&self, start: Time) -> Duration {
        self.inner.duration_since(self.inner).unwrap_or_default()
    }
}
```

I think this might be a way to slice the work between this PR and #5123. #5123 is the simplest, stupidest thing we can do to unblock fuzzing of whatever specific feature we want to fuzz, and this PR is the proper, long-term solution which would serve us until at least April 2262.

pmnoxx:
> Hm, given that what we want to do here is a "proper" mocked clock, I think it makes sense to try to avoid thread-locals, and use proper dependency injection. That is, to have two types: a clock (which returns time) and a time (which doesn't have a global `now` associated function):

It depends on how we want to use time function. I see 2 options:
- Use one instance of time, which can be mocked or not, and propagate it everywhere throughout the code
- Use `thread_local` to exchange information between threads. On the main thread we will have a pointer to a helper data structure used for mocking time. Whenever we start a new thread, we would update that pointer, to point to the same data structure as used by the main thread (for example when starting new Actor)

pmnoxx:
@matklad I see. I talked with @yordanmadzhunkov I'll close this PR for now. We need to first finish how we mock time, we will have to do this some other way later.

PR #5177, opened 2021-11-08
feat: Create `near-clock` crate for mocking time - pt 1

pmnoxx:
Discussed with matklad, we decided to implement `Time` crate.

Currently we use 4 methods to represent time `chrono` (`Utc::now()`) (`Datetime<Utc>`) , `SystemTime`, `Instant::now`, `u64`, they all have their issues.

- We don't want to use external `chrono` library, mainly because we don't really care about timezones
- `Instant` is written in such a way, that it's not possible to convert it to Unix timestamp. 
- `SystemTime` works, but it's not monitonic. It's possible that if we measure time at time `a` then at time `b`; the duration `b-a` will be negative, causing our code to fail.
- We want to use a saturating `Duration`, so any duration such that `a < b`, created by `a - b`, will default to `0`, if the difference were to become nagative. This will reduce chances of any crashed happening, when negative time is not expected.
- We should handle UTC `Leap seconds` correctly, see https://en.wikipedia.org/wiki/Leap_second

We are going to create our own wrapper around `SystemTime` and `Duration`. This will allow us to do the mocking properly.

TODO:
- [x] add BorshSerialize / BorshDeserialize to u64
- [x] move to time.rs
- [x] create `time` and `clock` submodules in `time.rs`
- [x] Add `pub use`
- [x] Add `TODO`

See https://github.com/near/nearcore/issues/5174

Waiting for #5362


matklad:
Argh, I left the review on the wrong PR, please see https://github.com/near/nearcore/pull/5175#pullrequestreview-801085360

pmnoxx (commented):
The solution is fine. However, it may be easy for anyone to forget and use Instant::now(), Utc::now() accidentally. I

pmnoxx:
@matklad What do you think?

pmnoxx:
@yordanmadzhunkov Can you take a look?  What do you think about the overall design.

I was thinking about renaming the `Clock` struct you created to `ClockMaker`. And putting `NearClock` as `Clock` inside `near-primitives`. You probably had something similar in mind. If you have any other design in mind we can go with that.

Feel free to take the change and/or modify it if you want to work on it. Or I can work on finishing this struct as it is then incorporating in in `near-network` before we use it in other places. It's up to you. 

Finishing this change will take a while. To be safe, we will have to go through every crate one-by-one and carefully be careful when changing the time library.

pmnoxx:
> As for naming, I'd like to use just `Time` for the POD with number of seconds since epoch, and `Clock` for the thing which can return `Time` in the `.now()` method. Given that we want to use single time throughout, I think using simple, prefx-less names woudl be the best (of course, doc comments on the corresponding structs should explain the guarantees we have)

Ok, I renamed struct to `Time` and moved it to the same file as `Clock`.

pmnoxx:
@matklad @mm-near What else do we need to finish this PR?

matklad (commented):
Seems good to go!

One thing which might be helpful though is documenting the semantics of this time (ie, why we don't just use SystemTime or Instance).

matklad (approved):
misclicekd

pmnoxx:
@matklad I refactored the code. I created two submodules `time` for `Time` and `clock` for `Clock`.
The code looks more readable now. By using `pub use`, we can split the api from the implementation. If we ever want to refactor the code, we can do that separately from changing the api. Which is great!

pmnoxx:
@mm-near Can you take a look?

pmnoxx:
@matklad I think that the PR is good as it is right now. Please, take a look after the refactor. I update the issue with what needs to be done.

core/primitives/src/time.rs
```diff
@@ -1,155 +1,295 @@
...
+        pub fn set_mock() {
+            MockClockPerThread::with(|clock| {
+                clock.is_mock = true;
+            });
+        }
+        pub fn reset() {
+            MockClockPerThread::with(|clock| {
+                clock.reset();
+            });
+        }
```
matklad, line 100: Seems like these two methonds should be private, as one always must use the guard to properly restore the clock.

core/primitives/src/time.rs
```diff
@@ -1,155 +1,295 @@
...
-            if clock.is_mock {
-                clock.utc.push_back(mock_date);
-            } else {
-                panic!("Use MockClockGuard in your test");
-            }
-        });
+
+    impl BorshSerialize for Time {
+        fn serialize<W: std::io::Write>(&self, writer: &mut W) -> Result<(), std::io::Error> {
+            let nanos = self.duration_since(&TIME_UNIX_EPOCH).as_nanos() as u64;
```
matklad, line 199: thought: this is subtle. `duration_since` is saturating, so using it for serialization might break borsh consistency/canonicity guarantees and give us two times which serialize to the same bytes. However, because we use UNIX_EPOCH which is zero, we actually guaranteed to never hit overflow behavior.

core/primitives/src/time.rs
```diff
@@ -1,155 +1,295 @@
...
-            clock.reset();
-        });
+mod time {
+    use crate::time::{Clock, Duration, Utc};
+    use borsh::{BorshDeserialize, BorshSerialize};
+    use chrono::DateTime;
+    use std::ops::{Add, Sub};
+    use std::time::SystemTime;
+
+    pub const TIME_UNIX_EPOCH: Time = Time::from_system_time(SystemTime::UNIX_EPOCH);
```
matklad, line 190: Actually, instead of defining top-level constant, let's do

```
impl Time {
  const UNIX_EPOCH: Time = Time::from_system_time(SystemTime::UNIX_EPOCH);
}
```

The `std::time::UNIX_EPOCH` global constant is a historical artifact -- rust didn't have associated costs originally.

core/primitives/src/time.rs
```diff
@@ -1,155 +1,295 @@
...
+    impl Time {
+        pub const fn from_system_time(system_time: SystemTime) -> Time {
+            Time { time: system_time }
+        }
+
+        pub fn from_utc(utc: DateTime<Utc>) -> Time {
+            let nanos = utc.timestamp_nanos() as u64;
+
+            TIME_UNIX_EPOCH + Duration::from_nanos(nanos)
+        }
```
matklad, line 222: These two can alternatively be `From` impls.

matklad (approved):
Looks good to me!

I'd personally wouldn't group time and clock into separate modules, as they feel intimately interrelated, but no strong opinion here.

core/primitives/src/time.rs
pmnoxx: https://github.com/near/nearcore/pull/5362

pmnoxx:
@matklad I added tests for Clock and some refactoring. You may want to take a look: https://github.com/near/nearcore/pull/5362

core/primitives/src/time.rs
pmnoxx: @matklad I would like to implement `to_unix_timestampt_nanos` without using `duration_since`. Any suggestions?
```
        pub fn to_unix_timestampt_nanos(&self) -> u64 {
            self.duration_since(&Time::UNIX_EPOCH).as_nanos() as u64
        }
```

core/primitives/src/time.rs
pmnoxx: Thanks, done.

pmnoxx:
I changed
```

        pub fn from_unix_timestamp(duration: Duration) -> Self {
            Self::UNIX_EPOCH + duration
        }

        pub fn to_unix_timestamp_nanos(&self) -> Duration {
            self.duration_since(&Self::UNIX_EPOCH)
        }
``` to use Duration rather than u64

File: core/primitives/src/time.rs [hunk 5]
```diff
@@ -1,155 +1,300 @@
...
+
+        pub fn elapsed(&self) -> Duration {
+            Self::now().duration_since(self)
+        }
+
+        pub fn from_unix_timestamp(duration: Duration) -> Self {
+            Self::UNIX_EPOCH + duration
+        }
+
+        pub fn to_unix_timestamp_nanos(&self) -> Duration {
```
frol, line 244: This "nanos" in the naming does not match my expectations given the `Duration` return value. To be honest, "unix_timestamp" as Duration is also a bit misleading (should we create a wrapper type or at least a type-alias?)

frol (commented):
I am fine with the implementation, but I would keep this in a PR state until we actually add some usage to it, so we can see if we need anything else added to it.

core/primitives/src/time.rs
pmnoxx: > This "nanos" in the naming does not match my expectations given the `Duration` return value. 

You are right, I missed that.

> To be honest, "unix_timestamp" as Duration is also a bit misleading

That's a good point. I'll rename it

>  (should we create a wrapper type or at least a type-alias?)

Yes, I think we should. I'll do it

pmnoxx:
> I am fine with the implementation, but I would keep this in a PR state until we actually add some usage to it, so we can see if we need anything else added to it.

I already wrote a PR for using it in `near-network`: https://github.com/near/nearcore/pull/5175
I need to rebase.

core/primitives/src/time.rs
pmnoxx: @frol @matklad Should we use 
```
type UnixTimeStamp = Duration;
```
or is it better to use keep u64 in those methods for compatibility reasons?

pmnoxx: > I am fine with the implementation, but I would keep this in a PR state until we actually add some usage to it, so we can see if we need anything else added to it.

Ok, I'll wait until I rewrite the other PR which adds support of this to `near-network`.

pmnoxx: After thinking about it:
```
type UnixTimeStamp = u64
```
Makes most sense. We only use nanos as UnixTimeStamp, there is no point in using duration.

matklad (approved):

pmnoxx:
@mm-near Can you take a look?

matklad:
Yeah! To clarify, that's the reason why `DateTime<Utc>`  in parciular may go backwards. Leap seconds shouldn't affect `SystemTime`, as that just counts the number of seconds since epoch. Even with leap seconds, the number total of seconds stays the same. What changes is whether a second number 123456789 is considered part of 2021 calendar year or 2022 calendar year.

pmnoxx:
@matklad @mm-near  I added pt2 of this PR, where we use `Time` in `near-network`: https://github.com/near/nearcore/pull/5556

PR #5184, opened 2021-11-09
runtime: simplify prepaid gas accounting in view calls

mina86:
Rather than special casing view calls each time prepaid gas is
checked, simply assume that prepaid gas is effectively infinite.  The
max gas burnt limit will still have an effect.

This fixes `sanity/rpc_max_gas_burnt.py` pytest which has been broken
by the recent ‘Simplify gas metering’ commit.  The commit changed
GasLimitExceeded failure into GasExceeded (by changing order in which
error conditions were checked).  This commit changes this back by
effectively disabling the first if branch in `process_gas_limit`
function when performing a view call.

### DIFF
--- a/runtime/near-vm-logic/src/gas_counter.rs
+++ b/runtime/near-vm-logic/src/gas_counter.rs
@@ -262,15 +262,16 @@ impl GasCounter {
 
 #[cfg(test)]
 mod tests {
+    use near_primitives_core::types::Gas;
     use crate::HostError;
 
-    fn test_counter(is_view: bool) -> super::GasCounter {
-        super::GasCounter::new(Default::default(), 10, 1, 12, is_view)
+    fn test_counter(max_burnt: Gas, prepaid: Gas, is_view: bool) -> super::GasCounter {
+        super::GasCounter::new(Default::default(), max_burnt, 1, prepaid, is_view)
     }
 
     #[test]
     fn test_deduct_gas() {
-        let mut counter = test_counter(false);
+        let mut counter = test_counter(10, 10, false);
         counter.deduct_gas(5, 10).expect("deduct_gas should work");
         assert_eq!(counter.burnt_gas(), 5);
         assert_eq!(counter.used_gas(), 10);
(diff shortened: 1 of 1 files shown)


bowenwang1996 (commented):
> The commit changed
GasLimitExceeded failure into GasExceeded (by changing order in which
error conditions were checked)

This is not expected. @olonho @matklad please take a look

matklad:
Good catch, missed this one. Makes me wonder, how many *more* unit-tests are we missing here.

I don't think this is quite the full fix though -- the order is still flipped on non-view calls.

mina86:
True. To be strictly correct we need to check which of the limits – gas limit or prepaid gas – is lower and prefer that. In the edge case where they are equal we just need to pick one or the other. I can prepare change to that effect.

mina86:
<https://github.com/near/nearcore/pull/5187> has the other change.

If general consensus is that this PR is a valid improvement let’s get it merged and we can discuss further on the other PR.

runtime/near-vm-logic/src/gas_counter.rs
```diff
@@ -103,8 +100,7 @@ impl GasCounter {
             self.fast_counter.burnt_gas.checked_add(burn_gas).ok_or(HostError::IntegerOverflow)?;
         let new_used_gas =
             new_burnt_gas.checked_add(new_promises_gas).ok_or(HostError::IntegerOverflow)?;
-        if new_burnt_gas <= self.max_gas_burnt && (self.is_view || new_used_gas <= self.prepaid_gas)
-        {
+        if new_burnt_gas <= self.max_gas_burnt && new_used_gas <= self.prepaid_gas {
```
mm-near, line 103: can you add the comment in line 94 - to explain what this method does and what are the arguments?
(what is the difference between 'burn' gas and 'use' gas?)

runtime/near-vm-logic/src/gas_counter.rs
```diff
@@ -266,21 +262,62 @@ impl GasCounter {
...
     #[should_panic]
-    fn test_prepaid_gas_min() {
-        let mut counter = GasCounter::new(ExtCostsConfig::default(), 100, 1, 10, false);
-        counter.deduct_gas(10, 5).unwrap();
+    fn test_burn_gas_must_be_lt_use_gas() {
+        let _ = test_counter(10, 10, false).deduct_gas(5, 2);
+        let _ = test_counter(10, 10, true).deduct_gas(5, 2);
+    }
+
+    #[test]
```
mm-near, line 287: thank you for adding the tests !

mm-near (approved):

matklad (approved):
Yeah, with tests this is definitely a significant improvement, let's merge.

runtime/near-vm-logic/src/gas_counter.rs
mina86: Added some short docs. Problem is that I’m not completely clear on how the gas accounting work.

Issue #5181, opened 2021-11-09
Investigate the time it takes to receive chunks of a block


nikurt:
https://github.com/near/nearcore/issues/5112

Things to measure:
* Time between receiving a block and receiving all of its chunks
* Time between requesting a chunk and receiving that chunk
Ideally that should be a prometheus metric to make it obvious whether the delays are decreasing between releases.

nikurt:
We see that some RPC nodes get chunks as much as 8 seconds after receiving the block, as described in #1689.

PR #5185, opened 2021-11-09
Chunk delay monitoring

nikurt:
Issue #5181
Tested on a testnet rpc node, nothing crashed, the metrics showed positive values. (rebased on origin/1.22.0: https://github.com/near/nearcore/compare/nikurt-chunk-monitoring-modern)

This change adds a new object to `Client`, which keeps track of timestamps of receiving blocks and chunks, keyed by block height. Timestamps of blocks older than N=10 get deleted. No restriction for the number of timestamps held which are newer than the current head.

Results from a single testnet rpc node:
image


chain/client/src/client.rs
```diff
@@ -1661,4 +1669,119 @@ impl Client {
...
+}
+
+#[derive(Debug, Default)]
+struct HeightInfo {
+    block_received: Option<Instant>,
+    chunks_received: HashMap<ShardId, Instant>,
+}
+
+#[derive(Debug, Default)]
+struct ChunksDelayTracker {
```
matklad, line 1692: it looks like a nice, self-contained data structure, let's move it to a dedicated file?

chain/client/src/client.rs
```diff
@@ -1661,4 +1669,119 @@ impl Client {
...
+const CHUNKS_DELAY_TRACKER_HORIZON: u64 = 10;
+
+impl ChunksDelayTracker {
+    fn lowest_height(head_height: BlockHeight) -> BlockHeight {
+        head_height.saturating_sub(CHUNKS_DELAY_TRACKER_HORIZON)
+    }
+    fn remove_old_entries(&mut self, head_height: BlockHeight) {
+        let heights_to_drop: Vec<BlockHeight> = self
+            .heights
+            .iter()
```
matklad, line 1706: You want `BTreeMap::range` here: https://doc.rust-lang.org/stable/std/collections/struct.BTreeMap.html#method.range

chain/client/src/client.rs
```diff
@@ -1661,4 +1669,119 @@ impl Client {
...
+        for (&k, v) in &self.heights {
+            if k > head_height {
+                break;
+            }
+            if let Some(block_received) = v.block_received {
+                if let Some(latest_chunk) = v.chunks_received.values().max() {
+                    if latest_chunk > &block_received {
+                        cnt += 1;
+                        sum_delays +=
+                            latest_chunk.duration_since(block_received).as_micros() as f64;
```
matklad, line 1730: Seems like we can keep all computations in integers?

chain/client/src/client.rs
```diff
@@ -1661,4 +1669,119 @@ impl Client {
...
+            }
+        }
+        if cnt == 0 {
+            near_metrics::set_gauge(&metrics::CHUNKS_RECEIVING_DELAY_US, 0);
+        } else {
+            near_metrics::set_gauge(
+                &metrics::CHUNKS_RECEIVING_DELAY_US,
+                (sum_delays / (cnt as f64)).round() as i64,
+            );
+        }
```
matklad, line 1742: I'd factor out:

```
let value = if  cnt == 0 { ... };
near_metrics::set_gauge(DELAY_US, value);

File: chain/client/src/client.rs [hunk 5]
```diff
@@ -1661,4 +1669,119 @@ impl Client {
...
+        self.heights
+            .entry(height)
+            .and_modify(|info| {
+                info.chunks_received.entry(shard_id).or_insert_with(|| Instant::now());
+            })
+            .or_insert_with(|| {
+                let mut chunks_received = HashMap::new();
+                chunks_received.insert(shard_id, Instant::now());
+                HeightInfo { block_received: None, chunks_received }
+            });
```
matklad, line 1784: Seems like you can replace and_modify/or_insert_with pair with a single `.or_deafult().chuns_received.entry(shrad_id).or_insert_with(Instant::now)`

neard/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "neard"
-version = "0.0.0"
+version = "1.21.1"
```
matklad, line 3: Seems unrelated?

bowenwang1996 (commented):
Please rebase your change on top of master. It is very hard to review the change this way

chain/client/src/client.rs
```diff
@@ -1661,4 +1669,119 @@ impl Client {
...
+    fn update_metrics(&mut self, head_height: BlockHeight) {
+        if head_height == self.prev_head_height {
+            return;
+        }
+        let mut cnt = 0;
+        let mut sum_delays = 0.;
+        for (&k, v) in &self.heights {
+            if k > head_height {
+                break;
+            }
```
mzhangmzz, line 1724: If I understand correctly, every time when `update_metrics` is called, it prints stats on block with height 
 between [head_height - CHUNKS_DELAY_TRACKER_HORIZON, head_height]. Does that mean delay on the same block can be printed more than twice?

Additionally, in the case of forks, a block at `height+1` can be processed before a block at `height`, if they are on different forks of the chain. So blocks with height <= head_height are not necessarily processed.

Therefore, I think it's the best to update metrics on a specific block by specifically checking that it has finished processing and remove it from the map. One way to check if a block is processed successfully is check `Chain::get_block_by_height` returns successfully.

chain/client/src/chunks_delay_tracker.rs
```diff
@@ -0,0 +1,200 @@
...
+        head_height.saturating_sub(CHUNKS_DELAY_TRACKER_HORIZON)
+    }
+
+    fn remove_old_entries(&mut self, head_height: BlockHeight) {
+        let mut heights_to_drop = vec![];
+        for (&k, _) in
+            self.heights.range(RangeTo { end: ChunksDelayTracker::lowest_height(head_height) })
+        {
+            heights_to_drop.push(k);
+        }
```
matklad, line 45:
```suggestion
        let heights_to_drop =
            self.heights.range(.. ChunksDelayTracker::lowest_height(head_height))
                .map(|(&k, _v)| k)
                .collect();
```

chain/client/src/chunks_delay_tracker.rs
```diff
@@ -0,0 +1,200 @@
...
+            self.heights.remove(&h);
+        }
+    }
+
+    fn update_chunks_metric(&mut self, head_height: BlockHeight) {
+        if head_height == self.prev_head_height {
+            return;
+        }
+        let mut max_delay = 0;
+        for (_, v) in self.heights.range(RangeToInclusive { end: head_height }) {
```
matklad, line 56:
```suggestion
        for (_, v) in self.heights.range(..=head_height) {
```

chain/client/src/chunks_delay_tracker.rs
```diff
@@ -0,0 +1,200 @@
...
+                    }
+                }
+            }
+        }
+        near_metrics::set_gauge(&metrics::CHUNKS_RECEIVING_DELAY_US, max_delay as i64);
+        self.prev_head_height = head_height;
+    }
+
+    fn update_blocks_missing_chunks_metric(&mut self, head_height: BlockHeight) {
+        let value = if let Some((&latest, _)) = self.heights.iter().rev().next() {
```
matklad, line 71:
```suggestion
        let value = if let Some((&latest, _)) = self.heights.iter().next_back() {
```

chain/client/src/chunks_delay_tracker.rs
```diff
@@ -0,0 +1,200 @@
...
+        &mut self,
+        height: BlockHeight,
+        head_height: BlockHeight,
+        timestamp: Instant,
+    ) {
+        self.remove_old_entries(head_height);
+        if height >= head_height {
+            self.heights
+                .entry(height)
+                .and_modify(|info| {
```
matklad, line 94: Feels like these two could be or_defaulted as well.

chain/client/src/chunks_delay_tracker.rs
```diff
@@ -0,0 +1,200 @@
...
+        }
+        self.update_metrics(head_height);
+    }
+
+    pub fn add_chunk_timestamp(
+        &mut self,
+        height: BlockHeight,
+        shard_id: ShardId,
+        head_height: BlockHeight,
+        timestamp: Instant,
```
matklad, line 112: :+1: for passing time in as a parameter!

chain/client/src/client.rs
nikurt: I agree that from the correctness point of view your approach is much better.
I want to build a prometheus metric, it's limited to being polled every 5-10 seconds. To provide a clear enough picture with that sampling rate I suggest to average metrics of the previous few blocks.
Forks is an important problem, but it should be ok for a monitoring utility not to worry about it.

chain/client/src/chunks_delay_tracker.rs
```diff
@@ -0,0 +1,188 @@
...
+            self.heights.remove(&h);
+        }
+    }
+
+    fn update_chunks_metric(&mut self, head_height: BlockHeight) {
+        if head_height == self.prev_head_height {
+            return;
+        }
+        let mut max_delay = 0;
+        for (_, v) in self.heights.range(..=head_height) {
```
mzhangmzz, line 54: Could you add a comment for what this function tries to do and the definition of this metric?

chain/client/src/chunks_delay_tracker.rs
```diff
@@ -0,0 +1,188 @@
...
+        self.prev_head_height = head_height;
+    }
+
+    fn update_blocks_missing_chunks_metric(&mut self, head_height: BlockHeight) {
+        let value = if let Some((&latest, _)) = self.heights.iter().next_back() {
+            latest.saturating_sub(head_height)
+        } else {
+            0
+        };
+        near_metrics::set_gauge(&metrics::BLOCKS_AHEAD_OF_HEAD, value as i64);
```
mzhangmzz, line 74: I think this metric is not that different from the block difference metric we already have right now? Also, it's not necessarily true that a block is received but not processed always has missing chunks.

chain/client/src/client.rs
```diff
@@ -848,6 +853,10 @@ impl Client {
         &mut self,
         partial_encoded_chunk: MaybeValidated<PartialEncodedChunk>,
     ) -> Result<Vec<AcceptedBlock>, Error> {
+        self.record_receive_chunk_timestamp(
```
mzhangmzz, line 856: It's possible that for a block and a shard, a node needs to receive multiple PartialEncodedChunk before the chunk is complete. So it's the best to call `record_receive_chunk_timestamp` when the chunk is actually ready, here https://github.com/near/nearcore/blob/master/chain/client/src/client.rs#L880

chain/client/src/chunks_delay_tracker.rs
nikurt: If I understand correctly, the block difference metric is computed as the difference of `max(head)` across all nodes and `head` of the current node.
`BLOCKS_AHEAD_OF_HEAD` metric tries to do that on the node itself.

> it's not necessarily true that a block is received but not processed always has missing chunks.
Renamed the function and cleaned up a few comments. Thanks.

mzhangmzz:
Looks good!

mzhangmzz (approved):

bowenwang1996 (approved):

PR #5186, opened 2021-11-09
Hotfix routing table recalculation

pmnoxx:
`tests/sanity/routing_table_sync.py` was failing, indicating that we were not running edge pruning nor routing table recalculation.

Here is a hotfix.
The following code was missing:
```
    self.needs_routing_table_recalculation = true;
```

### DIFF
--- a/chain/network/src/routing/routing_table_actor.rs
+++ b/chain/network/src/routing/routing_table_actor.rs
@@ -151,6 +151,10 @@ impl RoutingTableActor {
     /// If new edge comes comes that is adjacent to a peer that has been previously removed,
     /// we will try to re-add edges previously removed from disk.
     pub fn add_verified_edges_to_routing_table(&mut self, mut edges: Vec<Edge>) -> Vec<Edge> {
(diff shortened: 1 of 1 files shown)


pmnoxx:
@mm-near Can you take a look?

chain/network/src/routing/routing_table_actor.rs
```diff
@@ -151,6 +151,10 @@ impl RoutingTableActor {
     /// If new edge comes comes that is adjacent to a peer that has been previously removed,
     /// we will try to re-add edges previously removed from disk.
     pub fn add_verified_edges_to_routing_table(&mut self, mut edges: Vec<Edge>) -> Vec<Edge> {
+        if edges.len() == 0 {
```
bowenwang1996, line 154:
```suggestion
        if edges.empty() {
```

pmnoxx: Thanks, fixed.

mm-near (approved):

PR #5187, opened 2021-11-09
runtime: prefer reporting GasLimitExceeded

mina86:
If during execution of a contract both prepaid gas and max burnt gas
limits were exceeded, the `GasCounter` would report `GasExceeded`
error.  This was purely based on which condition was checked first
and entirely accidental based on how the code was written.

Unfortunately, that’s exactly the opposite preference to what we did
in old code which preferred reporting `GasLimitExceeded` instead.

Change the code to match the old behaviour and thus avoid protocol
changes.

### DIFF
--- a/runtime/near-vm-logic/src/gas_counter.rs
+++ b/runtime/near-vm-logic/src/gas_counter.rs
@@ -148,15 +148,11 @@ impl GasCounter {
         assert!(used_gas_limit >= self.fast_counter.burnt_gas);
         self.promises_gas = used_gas_limit - self.fast_counter.burnt_gas;
 
-        // If we crossed both limits prefer reporting the one which is lower.
-        // If both limits are the same arbitrarily prefer reporting burnt limit.
-        let report_burnt = if over_burnt != over_used {
-            over_burnt
-        } else {
-            debug_assert!(over_burnt && over_used);
-            self.max_gas_burnt <= self.prepaid_gas
-        };
-        if report_burnt { HostError::GasLimitExceeded } else { HostError::GasExceeded }.into()
+        // If we crossed both limits prefer reporting GasLimitExceeded.
+        // Alternative would be to prefer reporting limit that is lower (or
+        // perhaps even some other heuristic) but old code preferred
+        // GasLimitExceeded and we’re keeping compatibility with that.
+        if over_burnt { HostError::GasLimitExceeded } else { HostError::GasExceeded }.into()
     }
 
     pub fn pay_wasm_gas(&mut self, opcodes: u32) -> Result<()> {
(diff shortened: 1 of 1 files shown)


mina86:
This is meant to be on top of <https://github.com/near/nearcore/pull/5184>.  Only last commit here consists of the changes described in PR.

To be honest I don’t fully comprehend the accounting for burnt, used and promises gas but all the tests are passing with this change.

EgorKulikov:
Isn't this a protocol change?

mina86:
Here’s comparison of results of the `test_deduct_too_much` test on
various versions of the code:

| test        | 1.21.0           | Pre #5121        | HEAD (post #5121) | This PR v1       |
|-------------|------------------|------------------|-------------------|------------------|
| 5, 7, false | GasLimitExceeded | GasExceeded      | GasExceeded       | GasLimitExceeded |
| 5, 7, true  | GasLimitExceeded | GasExceeded      | GasLimitExceeded  | GasLimitExceeded |
| 5, 5, false | GasLimitExceeded | GasExceeded      | GasExceeded       | GasLimitExceeded |
| 5, 5, true  | GasLimitExceeded | GasExceeded      | GasLimitExceeded  | GasLimitExceeded |
| 7, 5, false | GasLimitExceeded | GasExceeded      | GasExceeded       | GasExceeded      |
| 7, 5, true  | GasLimitExceeded | GasExceeded      | GasLimitExceeded  | GasLimitExceeded |
| 5, 8, false | GasLimitExceeded | GasLimitExceeded | GasLimitExceeded  | GasLimitExceeded |
| 5, 8, true  | GasLimitExceeded | GasLimitExceeded | GasLimitExceeded  | GasLimitExceeded |
| 8, 5, false | GasExceeded      | GasExceeded      | GasExceeded       | GasExceeded      |
| 8, 5, true  | OK               | OK               | OK                | OK               |

With this change the (max_gas_burnt=7, prepaid_gas=5, is_view=false)
test produces a different result compared to what we had in 1.21.0.
In this test, we’re burning and using 8 gas which is above both
limits.

Changing the method to prefer reporting GasLimitExceeded brings all
the tests to the state they were at 1.21.0.

matklad:
This is not a protocol change, as we don't record which particular error occurred: https://github.com/near/nearcore/blob/b4531a00f4b77bfcce7126c29c9744d65762fdf4/core/primitives/src/transaction.rs#L299.

This is an unexpected behavior change which we missed in #5121 though, so it's still something we need to fix.

matklad (approved):
Thanks for fixing this!

bowenwang1996 (approved):

PR #5190, opened 2021-11-10
Fix `tests/sanity/routing_table_sync.py`test

pmnoxx:
`tests/sanity/routing_table_sync.py` test wasn't fully fixed.

We were not pruning while `prune == Prune::PruneNow`, because no routing table calculation was done.

This PR fixed the test properly.

The bug was in not having ```prune == Prune::PruneNow ``` check. In this PR we will improve code quality as well, and split `recalculate_routing_table_and_maybe_prune_edges` into to separate methods.

### DIFF
--- a/chain/network/src/routing/routing_table_actor.rs
+++ b/chain/network/src/routing/routing_table_actor.rs
@@ -516,10 +516,13 @@ impl Handler<RoutingTableMessages> for RoutingTableActor {
                 )
             }
             RoutingTableMessages::RoutingTableUpdate { prune, prune_edges_not_reachable_for } => {
-                if self.needs_routing_table_recalculation {
-                    self.recalculate_routing_table();
-                }
-                let edges_removed = self.prune_edges(prune, prune_edges_not_reachable_for);
+                let edges_removed =
+                    if self.needs_routing_table_recalculation || prune == Prune::PruneNow {
+                        self.recalculate_routing_table();
+                        self.prune_edges(prune, prune_edges_not_reachable_for)
+                    } else {
+                        Vec::new()
+                    };
                 self.needs_routing_table_recalculation = false;
                 RoutingTableMessagesResponse::RoutingTableUpdateResponse {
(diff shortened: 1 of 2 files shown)


pmnoxx:

chain/network/src/routing/routing_table_actor.rs
```diff
@@ -511,23 +516,21 @@ impl Handler<RoutingTableMessages> for RoutingTableActor {
...
-                    self.recalculate_routing_table_and_maybe_prune_edges(
-                        prune,
-                        prune_edges_not_reachable_for,
-                    )
-                } else {
-                    Vec::new()
-                };
-                self.needs_routing_table_recalculation = false;
+                let mut edges_removed =
+                    if self.needs_routing_table_recalculation || prune == Prune::PruneNow {
```
mm-near, line 520: so this is the place where the bug was, right?

and the refactoring/splitting of the methods was just to improve the readability?
(if yes - please mention it in PR description - as I was looking for the bug in those methods instead)

mm-near (approved):

chain/network/src/routing/routing_table_actor.rs
pmnoxx: Thanks. Yes, that was the bug. I updated the PR.

bowenwang1996:
Can we actually attach a link to a nayduck run that runs the test 30 times to verify the test is indeed fixed?

pmnoxx:
> Can we actually attach a link to a nayduck run that runs the test 30 times to verify the test is indeed fixed?

How can I schedule a single test to run on nayduck?

pmnoxx:
@bowenwang1996 Ok, I'll investigate on how to run single test on nayduck.

pmnoxx:
> Can we actually attach a link to a nayduck run that runs the test 30 times to verify the test is indeed fixed?

Thanks, that a good suggestion.

Issue #5155, opened 2021-11-05
Track memory usage for each Peer


pmnoxx:
Currently we are not tracing the amount of data that each peer sends to us. It's possible we will run out of memory because we get too many messages for each peer.

We should do the following
- [x] Add TrottledFramedRead - a data structure that tracks the size of all messages, and their count that come from `Peer` and are being processed. (https://github.com/near/nearcore/pull/4927)
- [ ] Add limiting bandwidth usage to `TrottledFramedRead` - we need to investigate first how much bandwidth is reasonable
- [ ] Add bandwidth stats to `TrottledFramedRead` - it's good to know how much bandwidth each peer is using to figure out proper throttling limits
- [x] Refactor `PeerManager` to keep all messages in `Handler` under one enum. https://github.com/near/nearcore/pull/5140
- [x] Add decorator to measure size of all messages that we send to `PeerManager` or `Peer` similar to https://crates.io/crates/deepsize_derive (https://github.com/near/nearcore/pull/5191)
- [ ] Keep tracking all messages that generated from `PeerActor` as they travel to `PeerManager` (https://github.com/near/nearcore/pull/5198)
- [ ] Keep tracking all messages that generated from `PeerActor` as they travel from `PeerManager` to another `Peer`(https://github.com/near/nearcore/pull/5198)

DeepSizeOf improvements - external repo
- [ ] add packages from std https://github.com/Aeledfyr/deepsize/pull/19
- [ ] add .gitignore https://github.com/Aeledfyr/deepsize/pull/20
- [ ] add packages from tokio/actix  https://github.com/Aeledfyr/deepsize/pull/18

pmnoxx:
@bowenwang1996 @mm-near

bowenwang1996:
>  Add decorator to measure size of all messages that we send to PeerManager or Peer similar to https://crates.io/crates/deepsize_derive

What do we plan to achieve with this? I believe quite a few messages are dynamically sized.

pmnoxx:
@bowenwang1996 `DeepSize` trait works by recursively computing size of each message. The overhead of doing a memory allocations to create given data structure than measuring it's size. So the performance overhead will be negligible.

pmnoxx:
> > Add decorator to measure size of all messages that we send to PeerManager or Peer similar to https://crates.io/crates/deepsize_derive
> 
> What do we plan to achieve with this? I believe quite a few messages are dynamically sized.

[...]

PR #5191, opened 2021-11-10
feat: Add `DeepSizeOf` decorator to `PeerManagerMessageRequest`

pmnoxx:
Hey, we are adding a big feature to around database. We will add a functionality to help investigate sizes of all data structures we use. This will require use to add `DeepSizeOf` to code that uses `BorshSerialize`. 

However, this has a lot of benefits:
- rate limiting - this will allow us to control memory usage - make sure our node doesn't ran out of memory
- debugging - it will be easy to check how big given data structure is, and optimize it by size
- network stats - we will be able to generate network stats. How big messages are, what's taking up memory / bandwidth. This will help us analyze, for example why nodes are out of sync sometimes.

This PR only adds `DeepSizeOf` decorator; this shouldn't effect current code execution in anyway.

This PR is  required to implement rate limiting, we want to control maximum memory usage in our node, per each connected peer. The goal is to stop reading from network, when we reach memory limits. https://github.com/near/nearcore/issues/5155
In the future, we will add more use cases this feature as well.

Split: 
- [x] part 1 - make `DeepSizeOf` decoration optional #5373

It will become optional in:
- near-network
- jsonrpc-adversarial-primitives
- near-network-primitives

### DIFF
--- a/chain/jsonrpc-adversarial-primitives/Cargo.toml
+++ b/chain/jsonrpc-adversarial-primitives/Cargo.toml
@@ -12,7 +12,7 @@ near-primitives = { path = "../../core/primitives" }
 
 near-jsonrpc-primitives = { path = "../jsonrpc-primitives", optional = true }
 near-network = { path = "../network" }
-
+deepsize = { version = "0.2.0", optional = true }
 
 [features]
 ser_de = [
@@ -21,3 +21,4 @@ ser_de = [
   "near-jsonrpc-primitives",
   "near-network/test_features"
 ]
+deepsize_feature = ["deepsize", "near-primitives/deepsize_feature"]
--- a/chain/jsonrpc-adversarial-primitives/src/lib.rs
+++ b/chain/jsonrpc-adversarial-primitives/src/lib.rs
@@ -1,3 +1,5 @@
+#[cfg(feature = "deepsize_feature")]
+use deepsize::DeepSizeOf;
 #[cfg(feature = "ser_de")]
 use near_jsonrpc_primitives::errors::RpcError;
 use near_network::routing::{Edge, SimpleEdge};
@@ -8,6 +10,7 @@ use serde::Deserialize;
 use serde_json::Value;
 
 #[cfg_attr(feature = "ser_de", derive(Deserialize))]
+#[cfg_attr(feature = "deepsize_feature", derive(DeepSizeOf))]
 pub struct SetRoutingTableRequest {
     pub add_edges: Option<Vec<Edge>>,
     pub remove_edges: Option<Vec<SimpleEdge>>,
--- a/chain/network-primitives/Cargo.toml
+++ b/chain/network-primitives/Cargo.toml
@@ -16,10 +16,11 @@ borsh = "0.9"
 serde = { version = "1", features = ["derive", "rc", "alloc"] }
 strum = { version = "0.20", features = ["derive"] }
 tracing = "0.1.13"
-
+deepsize = { version = "0.2.0", optional = true }
 near-crypto = { path = "../../core/crypto" }
 near-primitives = { path = "../../core/primitives" }
 
 [features]
 test_features = []
 sandbox = []
+deepsize_feature = ["deepsize", "near-primitives/deepsize_feature"]
--- a/chain/network-primitives/src/types.rs
+++ b/chain/network-primitives/src/types.rs
@@ -2,6 +2,8 @@ use actix::dev::{MessageResponse, ResponseChannel};
 use actix::{Actor, Message};
 use borsh::{BorshDeserialize, BorshSerialize};
 use chrono::DateTime;
+#[cfg(feature = "deepsize_feature")]
+use deepsize::DeepSizeOf;
 use near_crypto::{KeyType, PublicKey, SecretKey, Signature};
 use near_primitives::block::{Approval, Block, BlockHeader, GenesisId};
 use near_primitives::hash::CryptoHash;
@@ -49,6 +51,13 @@ pub struct PeerInfo {
     pub id: PeerId,
     pub addr: Option<SocketAddr>,
     pub account_id: Option<AccountId>,
+}
+
+#[cfg(feature = "deepsize_feature")]
+impl deepsize::DeepSizeOf for PeerInfo {
+    fn deep_size_of_children(&self, context: &mut deepsize::Context) -> usize {
+        self.id.deep_size_of_children(context) + self.account_id.deep_size_of_children(context)
+    }
 }
 
 impl PeerInfo {
@@ -121,6 +130,7 @@ impl TryFrom<&str> for PeerInfo {
 
 /// Peer chain information.
 /// TODO: Remove in next version
+#[cfg_attr(feature = "deepsize_feature", derive(DeepSizeOf))]
 #[derive(BorshSerialize, BorshDeserialize, Clone, Debug, Eq, PartialEq, Default)]
 pub struct PeerChainInfo {
     /// Chain Id and hash of genesis block.
@@ -132,6 +142,7 @@ pub struct PeerChainInfo {
 }
 
 /// Peer chain information.
+#[cfg_attr(feature = "deepsize_feature", derive(DeepSizeOf))]
 #[derive(BorshSerialize, BorshDeserialize, Clone, Debug, Eq, PartialEq, Default)]
 pub struct PeerChainInfoV2 {
     /// Chain Id and hash of genesis block.
@@ -156,6 +167,7 @@ impl From<PeerChainInfo> for PeerChainInfoV2 {
 }
 
 /// Peer type.
+#[cfg_attr(feature = "deepsize_feature", derive(DeepSizeOf))]
 #[derive(Copy, Clone, Debug, Eq, PartialEq, Serialize)]
 pub enum PeerType {
     /// Inbound session
@@ -165,6 +177,7 @@ pub enum PeerType {
 }
 
 /// Peer status.
+#[cfg_attr(feature = "deepsize_feature", derive(DeepSizeOf))]
 #[derive(Copy, Clone, Debug, Eq, PartialEq)]
 pub enum PeerStatus {
     /// Waiting for handshake.
@@ -176,6 +189,7 @@ pub enum PeerStatus {
 }
 
 /// Account route description
+#[cfg_attr(feature = "deepsize_feature", derive(DeepSizeOf))]
 #[derive(BorshSerialize, BorshDeserialize, PartialEq, Eq, Clone, Debug)]
 pub struct AnnounceAccountRoute {
     pub peer_id: PeerId,
@@ -183,12 +197,14 @@ pub struct AnnounceAccountRoute {
     pub signature: Signature,
 }
 
+#[cfg_attr(feature = "deepsize_feature", derive(DeepSizeOf))]
 #[derive(BorshSerialize, BorshDeserialize, PartialEq, Eq, Clone, Debug)]
 pub struct Ping {
     pub nonce: u64,
     pub source: PeerId,
 }
 
+#[cfg_attr(feature = "deepsize_feature", derive(DeepSizeOf))]
 #[derive(BorshSerialize, BorshDeserialize, PartialEq, Eq, Clone, Debug)]
 pub struct Pong {
     pub nonce: u64,
@@ -196,6 +212,7 @@ pub struct Pong {
 }
 
 // TODO(#1313): Use Box
+#[cfg_attr(feature = "deepsize_feature", derive(DeepSizeOf))]
 #[derive(
     BorshSerialize,
     BorshDeserialize,
@@ -313,12 +330,14 @@ impl Debug for RoutedMessageBody {
     }
 }
 
+#[cfg_attr(feature = "deepsize_feature", derive(DeepSizeOf))]
 #[derive(BorshSerialize, BorshDeserialize, PartialEq, Eq, Clone, Debug, Hash)]
 pub enum PeerIdOrHash {
     PeerId(PeerId),
     Hash(CryptoHash),
 }
 
+#[cfg_attr(feature = "deepsize_feature", derive(DeepSizeOf))]
 #[derive(Debug, Clone, PartialEq, Eq, BorshSerialize, BorshDeserialize, Hash)]
 // Defines the destination for a network request.
 // The request should be sent either to the `account_id` as a routed message, or directly to
@@ -339,6 +358,7 @@ impl AccountIdOrPeerTrackingShard {
     }
 }
 
+#[cfg_attr(feature = "deepsize_feature", derive(DeepSizeOf))]
 #[derive(Debug, Clone, PartialEq, Eq, BorshSerialize, BorshDeserialize, Hash)]
 pub enum AccountOrPeerIdOrHash {
     AccountId(AccountId),
@@ -393,6 +413,7 @@ pub struct RoutedMessageNoSignature<'a> {
 /// sender of the package should be banned instead.
 /// If target is hash, it is a message that should be routed back using the same path used to route
 /// the request in first place. It is the hash of the request message.
+#[cfg_attr(feature = "deepsize_feature", derive(DeepSizeOf))]
 #[derive(BorshSerialize, BorshDeserialize, PartialEq, Eq, Clone, Debug)]
 pub struct RoutedMessage {
     /// Peer id which is directed this message.
@@ -448,6 +469,7 @@ impl RoutedMessage {
 }
 
 /// Routed Message wrapped with previous sender of the message.
+#[cfg_attr(feature = "deepsize_feature", derive(DeepSizeOf))]
 #[derive(Clone, Debug)]
 pub struct RoutedMessageFrom {
     /// Routed messages.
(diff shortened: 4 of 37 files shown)


pmnoxx:
@bowenwang1996 @mm-near @matklad Can you take a look?

pmnoxx:
@bowenwang1996 Unfortunately, I can't make this PR smaller.

pmnoxx:
Check fails, because `deepsize_derive`  depends on `proc-macro2 0.4.30 `
While we use `proc-macro2 1.0.30`.

We may have to copy  `deepsize_derive`  or submit `PR` to them, but who knows how long will that take.

```
[2021-11-10T07:19:18Z] error[B004]: found 2 duplicate entries for crate 'proc-macro2'
--
  | [2021-11-10T07:19:18Z]     ┌─ /var/lib/buildkite-agent/builds/buildkite-i-0a8ad557d7118ea25-1/nearprotocol/nearcore/Cargo.lock:314:1
  | [2021-11-10T07:19:18Z]     │
  | [2021-11-10T07:19:18Z] 314 │ ╭ proc-macro2 0.4.30 registry+https://github.com/rust-lang/crates.io-index
  | [2021-11-10T07:19:18Z] 315 │ │ proc-macro2 1.0.30 registry+https://github.com/rust-lang/crates.io-index
  | [2021-11-10T07:19:18Z]     │ ```╰────────────────────────────────────────────────────────────────────────^ lock en

**Comment by matklad** [comment]
Note: we should be more careful when selecting dependencies. https://crates.io/crates/deepsize was updated last time a year ago and it is used by just two other crates. At the same time, it is of the "conquer the world" style of crates which requires annotating every type. So, this is a pretty risky dependency to add. 

Not saying that we shouldn't be adding it, but the PR with the addition should probably contain an explicit callout like "hey, we are adding this big capability to our codebase, it's somewhat experimental crate, but we feel its worth it". 

We should probably also review some alternatives as well. In particular, https://github.com/wasmerio/loupe is a more fresh alternative. There's also MallocSizeOf in the guts of servo. 

More generally, "getting sizes of things" in Rust is't a satisfactory solved problem at the moment. I wanted to have a solution for this for ages (fun fact, the deepsize crate was created upon my request to solve this problem https://github.com/dtolnay/request-for-implementation/issues/22), but, eg, in rust-analyzer we still don't use that, as it isn't worth it, and instead do [horrible tricks](https://rust-analyzer.github.io/blog/2020/12/04/measuring-memory-usage-in-rust.html). I do think that, if we want exactly this, than taking on deepsize dependency is probably the way to go, it's just that this'll add non-zero long-term maintenance.

I am not sure we want exactly this though. My understanding is that the end consumer here would be rate limiting, and for that, we want to estimate a size of message. My gut feeling is that just writing a `fn messge_size(msg: &Msg) -> usize` by hand would probably be a more robust approach: 

* we can write it in one place, without changes to the codebase elsewhere
* with "one place", it will be much easier to tweak/remove/replace
* we care about approximate answer, and, ideally, we'd take not just the size of the message, but the amount of memory it takes to process. With hand-written function, it would be much easier to do tweaks like "multiply the size of this particular kind of message by two, as they are costly to process".

**Comment by matklad** [comment]
Note: if we do take on this depenrency, we should send a PR upstream to upgrade its `syn` version. Compiling two versions of `syn` is not fun.

**Comment by pmnoxx** [comment]
I checked dependencies of `deepsize_derive`. In https://github.com/Aeledfyr/deepsize/blob/master/deepsize_derive/Cargo.toml
they specified to use stable version of libraries:
```
[package]
name = "deepsize_derive"
version = "0.1.1"
...

[dependencies]
proc-macro2 = "^1.0"
syn = "^1.0"
quote = "^1.0"
```

However, in our `Cargo.lock`, I see that the old ones are used

```

[[package]]
name = "deepsize_derive"
version = "0.1.1"
source = "registry+https://github.com/rust-lang/crates.io-index"
checksum = "eefac8b754bebccb018f15a8ff0015a3f58fb298825f5bd0dcf3e7e2c017e992"
dependencies = [
 "proc-macro2 0.4.30",
 "quote 0.6.13",
 "syn 0.15.44",
]
```

@matklad Any ideas?


I checked out `https://github.com/wasmerio/loupe`. It was updated 8 months ago version 13 months ago for `DeepSizeOf`. So it's not much different. I think that `DeepSizeOf` is better maintained, it supports more types. Anyway, I couldn't find `MallocSizeOf` that you mentioned.

**Comment by pmnoxx** [comment]
> Note: we should be more careful when selecting dependencies. https://crates.io/crates/deepsize was updated last time a year ago and it is used by just two other crates. At the same time, it is of the "conquer the world" style of crates which requires annotating every type. So, this is a pretty risky dependency to add.

Yes, I agree. Though, I think that the benefits outweigh the drawbacks. In the near future, we should implement adding network stats. I would like to see how big are all messages that we send over the network. That can help us explain which data structures are big, and consume a lot of bandwidth, use a lot of memory etc.

> Not saying that we shouldn't be adding it, but the PR with the addition should probably contain an explicit callout like "hey, we are adding this big capability to our codebase, it's somewhat experimental crate, but we feel its worth it".

Good idea, I updated the PR.

> We should probably also review some alternatives as well. In particular, https://github.com/wasmerio/loupe is a more fresh alternative. There's also MallocSizeOf in the guts of servo.

I looked it, but currently it offer less features than `DeepSizeOf`. However, I'm open to suggestions I can replace `DeepSizeOf` with any other library if needed.
 
> More generally, "getting sizes of things" in Rust is't a satisfactory solved problem at the moment. I wanted to have a solution for this for ages (fun fact, the deepsize crate was created upon my request to solve this problem [dtolnay/request-for-implementation#22](https://github.com/dtolnay/request-for-implementation/issues/22)), but, eg, in rust-analyzer we still don't use that, as it isn't worth it, and instead do [horrible tricks](https://rust-analyzer.github.io/blog/2020/12/04/measuring-memory-usage-in-rust.html). I do think that, if we want exactly this, than taking on deepsize dependency is probably the way to go, it's just that this'll add non-zero long-term maintenance.

Yes, we should investigate if there is a better way to do it. However, I think this is a good enough, if we find a better way we will replace DeepSizeOf with something equivalent.
 
> I am not sure we want exactly this though. My understanding is that the end consumer here would be rate limiting, and for that, we want to estimate a size of message. My gut feeling is that just writing a `fn messge_size(msg: &Msg) -> usize` by hand would probably be a more robust approach:

This is not the only use case. Previously, when I was investigating memory leaks, I used `DeepSizeOf` to print sizes of data structures to see how big they are. Right now, it's hard to estimate how big our data structures are by just looking at the code.
Sometimes, you want to check whenever a particular data structure is big enough, that it should be optimized or not.
 
> * we can write it in one place, without changes to the codebase elsewhere
> * with "one place", it will be much easier to tweak/remove/replace
> * we care about approximate answer, and, ideally, we'd take not just the size of the message, but the amount of memory it takes to process. With hand-written function, it would be much easier to do tweaks like "multiply the size of this particular kind of message by two, as they are costly to process".

Yes, that's possible. However, writing manually code for over 500 data structures would be both time consuming, and hard to maintain. Changing any data structure, would require changes in two places. Hand writing code with any estimator my by really error-prone, requiring extra time. I strongly feel having to write manual method each time would add a lot of maintenance.

**Comment by matklad** [comment]
>I couldn't find MallocSizeOf that you mentioned.

Oups, sorry, should have added a link: https://github.com/servo/servo/tree/master/components/malloc_size_of. It implements things slightly differently -- rather than using `size_of`, it asks the allocator about the size of the underlying allocation, so:

* it is more precise, as it accounts for extra padding when the object doesn't fit exactly into allocator's size classes
* it is more fragile, as it requires runtime-support from a particular allocator. 

>Previously, when I was investigating memory leaks, I used DeepSizeOf to print sizes of data structures to see how big they are. 

Yeah, that' very convincing argument for having this thing. 

> Any ideas?

The code on github wasn't published yet, I've asked the author to publish: https://github.com/Aeledfyr/deepsize/commit/ffc79ef6277d3e14a6b5ced49a6f45e825c6c36f#r60105545

File: chain/jsonrpc-adversarial-primitives/src/lib.rs [hunk 1]
```diff
@@ -7,7 +7,10 @@ use serde::Deserialize;
 #[cfg(feature = "ser_de")]
 use serde_json::Value;
 
+use deepsize::DeepSizeOf;
+
 #[cfg_attr(feature = "ser_de", derive(Deserialize))]
```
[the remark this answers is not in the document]
pmnoxx: `json-adversarial-primitives` is only enabled when code is compiled with `test_features`.

core/crypto/src/signature.rs
```diff
@@ -76,6 +78,12 @@ fn split_key_type_data(value: &str) -> Result<(KeyType, &str), crate::errors::Pa
 #[derive(Copy, Clone)]
 pub struct Secp256K1PublicKey([u8; 64]);
 
+impl DeepSizeOf for Secp256K1PublicKey {
+    fn deep_size_of_children(&self, _context: &mut Context) -> usize {
+        64
```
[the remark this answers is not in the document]
pmnoxx: Fixed. You are right, this array will not cause an allocation, and the size should be 0.

pmnoxx:
> * it is more fragile, as it requires runtime-support from a particular allocator.

Yes, it may be worth considering it. However, doing c calls may have some performance implications, because that code can't be fully optimized by the compiler. We would have to compare the performance.

It's probably fine for us to use less precise measurement. 

I'm also a little bit worried that this may not work on all platforms or with some memory allocators.

chain/network-primitives/src/types.rs
```diff
@@ -796,21 +812,27 @@ pub struct KnownProducer {
     pub peer_id: PeerId,
 }
 
-#[derive(PartialEq, Eq, Clone, Debug, BorshSerialize, BorshDeserialize)]
+impl DeepSizeOf for KnownProducer {
+    fn deep_size_of_children(&self, context: &mut Context) -> usize {
+        self.account_id.deep_size_of_children(context) + self.peer_id.deep_size_of_children(context)
+    }
+}
```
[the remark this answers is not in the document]
pmnoxx: Yeah, I was thinking about doing that.

chain/network/src/types.rs
```diff
@@ -446,11 +448,24 @@ pub struct Consolidate {
     pub peer_protocol_version: ProtocolVersion,
 }
 
+// we can't implement this for Consolidate, because that would violate rust rule not to
+// implement foreign traits for foreign types.
+impl DeepSizeOf for Consolidate {
```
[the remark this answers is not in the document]
pmnoxx: Yeah, I was thinking about doing that.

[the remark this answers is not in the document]
pmnoxx: https://github.com/Aeledfyr/deepsize/pull/18

chain/network-primitives/src/types.rs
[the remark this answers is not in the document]
pmnoxx: https://github.com/Aeledfyr/deepsize/pull/18

pmnoxx:
@matklad We may have to create a work for `DeepSizeOf`. We don't know how much time it will take for the author to publish new create with updated dependency, and that's blocking us.

matklad:
deepsize-derive with syn 1.0 was published just yesterday, that's great: https://crates.io/crates/deepsize-derive/0.1.2

pmnoxx:
@matklad Yes, but  we have to wait for `deepsize`, right?

matklad:
Yeah, I mean that in general the maintainer is responsive, so we can hope for deep size publishing.

frol:
This change is taxing all the crates. I hate to be this guy, but at the very least we need to make this dependency optional.

pmnoxx:
> This change is taxing all the crates. I hate to be this guy, but at the very least we need to make this dependency optional.

@frol No problem. I can make it so `DeepSize` trait is noop as long as you don't compile `near-network.`

That's actually a good idea. I can implement the change #5198
without usage of DeepSizeOf. Even without `DeepSizeOf` we will be able to get most of needed functionality. That will allow us to unblock #5198

Later, we will be able to decide whenever we want it always enabled. We can do it just for the purpose of `near-network`.

pmnoxx:
> This change is taxing all the crates. I hate to be this guy, but at the very least we need to make this dependency optional.

Btw, what are your main concerns?
- compilation time
- extra memory - deepsize doesn't increase size of structs
- maintenance - having to add `DeepSize` to all structs
- crate size - this crate has very little code, it doesn't be an issue.
- adding extra  dependencies for other binaries - like indexer - I agree, `indexer` and other binaries don't need it.

pmnoxx:
@frol Here is how we can do it without adding extra import.

We could do it for other derives, and/or traits as well if we want to.

pmnoxx:
@frol @matklad @mm-near Can you take a look?

We merged https://github.com/near/nearcore/pull/5373 where we make `DeepSizeOf` optional for most crates.

This PR makes `DeepSizeOf` required for 
- `jsonrpc-adversarial-primitives`
- `near-network`
- `near-network-primitives`

This PR is ready. The only last question is to decide whenever we want to make `jsonrpc-adversarial-primitives` and ` `near-network-primitives` use `deepsize` as an optional dependency.

chain/jsonrpc-adversarial-primitives/Cargo.toml
```diff
@@ -21,3 +21,4 @@ ser_de = [
   "near-jsonrpc-primitives",
   "near-network/test_features"
 ]
+deepsize_feature = ["deepsize", "near-primitives/deepsize_feature"]
```
matklad, line 24: Sadly, I don't know a way to make `deepsize` name here "just work".

matklad (approved):

chain/jsonrpc-adversarial-primitives/Cargo.toml
pmnoxx: @matklad Yes, that's unfortunate.

chain/network/src/types.rs
```diff
@@ -458,10 +469,25 @@ pub struct Consolidate {
     pub(crate) peer_protocol_version: ProtocolVersion,
 }
 
+// we can't implement this for Consolidate, because that would violate rust rule not to
```
mm-near, line 472: What is *this* that you're referring to in 'we can't implement this'?

mm-near (approved):

chain/network-primitives/Cargo.toml
```diff
@@ -16,10 +16,11 @@ borsh = "0.9"
 serde = { version = "1", features = ["derive", "rc", "alloc"] }
 strum = { version = "0.20", features = ["derive"] }
 tracing = "0.1.13"
-
+deepsize = { version = "0.2.0", optional = true }
 near-crypto = { path = "../../core/crypto" }
-near-primitives = { path = "../../core/primitives" }
+near-primitives = { path = "../../core/primitives", features = ["deepsize_feature"] }
```
frol, line 21: I feel we should not enable the feature unconditionally:

```suggestion
near-primitives = { path = "../../core/primitives" }
```

I see the same pattern in other Cargo.toml files, so we should address it there as well

chain/network/src/routing/ibf_peer_set.rs
```diff
@@ -1,12 +1,14 @@
 use crate::routing::edge::{Edge, SimpleEdge};
 use crate::routing::ibf_set::IbfSet;
 use borsh::{BorshDeserialize, BorshSerialize};
+use deepsize::DeepSizeOf;
```
frol, line 4: This should be behind a feature-flag.

P.S. I feel it would be easier to just use a fully-qualified path inside the `derive()` to avoid a need to have a conditional use.

frol (approved):
LGTM! Make your own judgement whether it makes sense to address my review comments

chain/network/src/types.rs
pmnoxx: I replaced this comment with a more descriptive one
```
/// Addr<PeerActor> doesn't implement `DeepSizeOf` waiting for `deepsize` > 0.2.0.
```

chain/network-primitives/Cargo.toml
pmnoxx: Thanks. Fixed.

chain/network/src/routing/ibf_peer_set.rs
pmnoxx: > This should be behind a feature-flag.

@frol `near-network` requires `deepsize::DeepSizeOf` to work. That's why it's not under feature flag. However, we may make `near-network` optionally require `deepsize`, but I think this should be addressed by another PR.

> P.S. I feel it would be easier to just use a fully-qualified path inside the `derive()` to avoid a need to have a conditional use.

I agree, this was suggested by @matklad

pmnoxx:
I'll prepare another PR, which deals with resolves issue of conditional imports.

chain/network/src/routing/ibf_peer_set.rs
pmnoxx: @frol Oh, I realized that I made all packages use `deepsize_feature` optionally. 
Thanks I fixed it.

I also did a PR, where I do it for whole `near-network`: https://github.com/near/nearcore/pull/5436

Issue #5112, opened 2021-11-02
RPC nodes falling out of sync


mhalambek:
RPC nodes on mainnet are lagging behind latest block. This is causing issues for users of wallet and other services which use RPC.

User reported errors from wallet
![image](https://user-images.githubusercontent.com/7011689/139862378-730f7e88-a3a2-4612-8865-5a7d5642e348.png)
![image (1)](https://user-images.githubusercontent.com/7011689/139862386-b65ed061-332b-48f8-8936-f4a5dfbcc865.png)

Grafana graph showing time it takes the node to reach latest block.
![2021-11-02-140632_1920x1080_scrot](https://user-images.githubusercontent.com/7011689/139863093-944e8b72-9d54-41bd-8cb1-4c8afbc25fc3.png)

Nearcore version 1.21.0

evgenykuzyakov:
I've also got errors from `near-cli` that marks transaction as `Expired`:

E.g. `ServerError: Transaction has expired`

I think the reason is the node is out of sync and rejects the transaction with the blockhash that is recent, but not synced yet.

I suggest to reply from such nodes with a different error. E.g. `OutOfSync` RPC error, so a client can retry the request to a different RPC node.

nikurt:
Based on the issue #4937 , dumped column `ColStateDlInfos` and it was empty.
The code: https://github.com/near/nearcore/compare/nikurt-col-dump?expand=1
Couldn't use `ldb` or `sst_dump` due to `too many files open` error.

bowenwang1996:
I looked a bit more into this issue. It appears to me that the issue is caused by nodes waiting sometimes too long for a chunk to arrive. I think we may need to prioritize #3863, #1689, and potentially #3890. cc @mzhangmzz

nikurt:
Extracted a few lines of log during an archival node switching back and forth between syncing and working:
[archive_rpc_sync_log.txt](https://github.com/near/nearcore/files/7495419/archive_rpc_sync_log.txt)

nikurt:
All cases of an Asian archival rpc node starting and finishing syncing:
[Uploading archive_rpc_sync_log.txt…]()

nikurt:
The node managed to fall 10 blocks behind merely 10 seconds after finishing the last sync: #52158507 -> #52158517
[big_gap.txt](https://github.com/near/nearcore/files/7496063/big_gap.txt)

pmnoxx:
I'll take a look

[...]

PR #5192, opened 2021-11-10
Add extra logs to investigate issues

pmnoxx:
We have seen cases where nodes would get stack syncing. (https://github.com/near/nearcore/issues/5112)

It looks like we are trying to request headers from a route to which we don't have a route:
```
Nov 08 08:05:14 public-v1-archive-mainnet-rpc-node-p3x5 sh[1942]: Nov 08 08:05:14.108 DEBUG network: None Drop signed message to Hash(`GC72RjNQpJuKMierihJ4bYaF7WMFQCYfPgadbRyGE3fA`) Reason RouteBackNotFound. Num known peers: 411 Message PartialChunkResponse(ChunkHash(`6Y2YYNUpp62uNQxAvo3a7LxfC4eLMDeXeoZisKHcPANx`), [84])
Nov 08 08:05:14 public-v1-archive-mainnet-rpc-node-p3x5 sh[1942]: Nov 08 08:05:14.131 DEBUG network: None Drop signed message to Hash(`C6Q52jRSJxnfuSn7TuoSbVXcEVM8DtpB2xikFDamzLVT`) Reason RouteBackNotFound. Num known peers: 411 Message PartialChunkResponse(ChunkHash(`DNEWNwDysCsAHg14vy6UdVQHvdSQHU1i6yxiDcanhieA`), [84])
```

I noticed a code path, where it's possible we will return to `PeerActor` list of peers to which we have no route to `highest_height_peers`.
I added printing an error when that happens.

### DIFF
--- a/chain/network/src/peer_manager/peer_manager_actor.rs
+++ b/chain/network/src/peer_manager/peer_manager_actor.rs
@@ -855,7 +855,8 @@ impl PeerManagerActor {
             None => return vec![],
         };
         // Find all peers whose height is within `highest_peer_horizon` from max height peer(s).
-        self.active_peers
+        let result = self
+            .active_peers
             .values()
             .filter_map(|active_peer| {
                 if active_peer.full_peer_info.chain_info.height + self.config.highest_peer_horizon
@@ -866,7 +867,23 @@ impl PeerManagerActor {
                     None
                 }
             })
-            .collect::<Vec<_>>()
+            .collect::<Vec<_>>();
+
+        for peer in result.iter() {
+            if self
+                .routing_table_view
+                .peer_forwarding
+                .get(&peer.peer_info.id)
+                .unwrap_or(&Vec::new())
+                .is_empty()
+            {
+                // We suspect that there is an active peer to which we don't have a route to
+                // and that's causing a desync. More research is required.That shouldn't have happened!.
+                error!(target: "network", "returning peer to ClientActor to which we have no route: {}", &peer.peer_info.id);
+                break;
+            }
+        }
+        result
     }
 
     /// Returns bytes sent/received across all peers.
@@ -1328,7 +1345,10 @@ impl PeerManagerActor {
                 // Remember if we expect a response for this message.
                 if msg.author == self.my_peer_id && msg.expect_response() {
                     trace!(target: "network", "initiate route back {:?}", msg);
-                    self.routing_table_view.add_route_back(msg.hash(), self.my_peer_id.clone());
+                    if !self.routing_table_view.add_route_back(msg.hash(), self.my_peer_id.clone())
+                    {
+                        warn!(target: "network", "We couldn't add message to add_route_back: {}", PeerMessage::Routed(msg.clone()));
+                    }
                 }
 
                 self.send_message(ctx, peer_id, PeerMessage::Routed(msg))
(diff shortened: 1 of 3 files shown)


chain/network/src/peer_manager/peer_manager_actor.rs
```diff
@@ -852,7 +853,26 @@ impl PeerManagerActor {
...
+        // remove peers to which we have no route from the list
+        result.retain(|peer| {
+             if self
+                .routing_table_view
+                .peer_forwarding
+                .get(&peer.peer_info.id)
+                .unwrap_or(&Vec::new())
+                .is_empty()
+            {
+                error!(target: "network", "returning peer to ClientActor to which we have no route: {}", &peer.peer_info.id);
```
mm-near, line 867: how can we make sure that this doesn't spam our logs?

chain/network/src/peer_manager/peer_manager_actor.rs
```diff
@@ -852,7 +853,26 @@ impl PeerManagerActor {
                     None
                 }
             })
-            .collect::<Vec<_>>()
+            .collect::<Vec<_>>();
+
+        // remove peers to which we have no route from the list
```
mm-near, line 858: what is the scenario when it happens? 

can we add a test for this case?

chain/network/src/peer_manager/peer_manager_actor.rs
```diff
@@ -852,7 +853,26 @@ impl PeerManagerActor {
...
+                .get(&peer.peer_info.id)
+                .unwrap_or(&Vec::new())
+                .is_empty()
+            {
+                error!(target: "network", "returning peer to ClientActor to which we have no route: {}", &peer.peer_info.id);
+                false
+            } else {true}
+        });
+        if result.is_empty() {
+            error!(target: "network", "result of highest_height_peers became empty");
```
mm-near, line 872: maybe also explain in this log statement what could be the possible reason?

pmnoxx: @mm-near I suspect that there is an issue. We return in this function a peer, to which we don't have a route.
My plan is to do two things
- add a print, and see if that's really the same. once we confirm that this is the case. 
a)Once we confirm that, we can dedicate some time to add extra tests to find out where is the bug in logic that causes us to choose peers out of sync.
b) if that's not the case, we will have to look at a bug, in another code path
- There is no need to send list of peers to which we don't have path to, they would be dropped anyway. That's why I added `retain`.


Right now I'm working on other changes. This is a simple change, that should let us know whenever the issue is somewhere in this code.

chain/network/src/peer_manager/peer_manager_actor.rs
pmnoxx: Ok, I can add something that limits number of prints.

chain/network/src/peer_manager/peer_manager_actor.rs
pmnoxx: I don't know exactly why this happens yet. This will require some time to investigate the logic. I think it's a good idea to add a print to alert if that happens again, while we work on finding the issue.

For example:
- there may be an issue with routing table not being updated properly. Or for whatever other reason why `active_peers` has peers to which we don't have a route to. 

Anyway, if the issue is in this function, that would explain the error we are seeing,

matklad (approved):

pmnoxx:
@matklad @mm-near I changed this PR to only add prints. That should be safe to ship.

bowenwang1996 (approved):

Issue #5188, opened 2021-11-09
`pytest sanity/routing_table_sync.py` test is broken


mina86:
The test currently fails on NayDuck.  To reproduce:

    ./scripts/nayduck -b master -s master -i <<<EOF
    pytest --timeout=480 sanity/routing_table_sync.py --features nightly_protocol,nightly_protocol_features
    EOF

pmnoxx:
@mina86 Thanks, the hotfix has been merged.

pmnoxx:
@mina86 Sorry, I didn't notice that the previous fix didn't fix issue.
https://github.com/near/nearcore/pull/5190/files

pmnoxx:
I reopened this pr.
@bowenwang1996 @mm-near Can you take a look?

mina86:
Looks like it’s still failing: http://nayduck.near.org/#/test/241592

pmnoxx:
> Looks like it’s still failing: http://nayduck.near.org/#/test/241592

I wrote a fix: https://github.com/near/nearcore/pull/5199

I ran in 30 times, it works fine: http://nayduck.near.org/#/run/2200

PR #5199, opened 2021-11-10
Fix timeout in routing_table_sync.py

pmnoxx:
We saw that nayduck test failed, due to having 2s timeout in `nodes[0].get_status(timeout=30)`

Closes https://github.com/near/nearcore/issues/5188#event-5600015413


pmnoxx:
@bowenwang1996 http://nayduck.near.org/#/run/2196

pmnoxx:
@bowenwang1996 All tests are passing. I had to lower test limits. Those tests are failing, because apparently our CI runs on a machines a few times slower than my own.

mina86 (approved):

PR #5200, opened 2021-11-11
Implement fast gas counting using intrinsics

olonho:
To speed up execution and lower gas cost we need to implement optimised gas counting algorithm.
It involves intrinsified code generation for `gas()` external function invocations and mechanism to pass
an external gas counter pointer.

 In current form we refer to git branch in Wasmer repo matching https://github.com/near/wasmer/pull/64/, as we need to make certain change in Wasmer codebase.

 After this change we can reduce gas cost for single Wasm instruction to 312_544 gas.


bowenwang1996 (commented):
@matklad please review this PR. Thanks!

matklad:
I think this formally needs to be a draft, as we didn't fully finish the wasmer side of the work, but the changes here look good!

bowenwang1996 (approved):

matklad (approved):

Issue #1689, opened 2019-11-14
Request parts for multiple blocks simultaneously.


SkidanovAlex:
Today if we receive a block at height `k`, we only accept it if
a) It's parent is *accepted*
b) We have all the chunks / partial chunks.

If we receive a block at height `k`, and it doesn't have all the chunks / partial chunks, we put it into `blocks_with_missing_chunks` and request chunks. If at this time a block arrives for height `k+1`, we put it into `orphans`.

The problem is that the block at height `k+1` likely also misses chunks / partial chunks. Such chunks will not be requested until all the chunks for block at `k` arrive, block at `k` gets accepted, we check orphans, identify `k+1` there, process it again, see that it misses chunks, put it into `blocks_with_missing_chunks` and request chunks.

A better approach would be as we put the block into orphans check if the following condition is satisfied:
1. Among the `z` immediate parents of the block at least one is accepted (is in `store`)
2. All the `z` immediately parents of the block are either accepted (are in `store`), or are orphans, or are in `blocks_with_missing_chunks`
And if both conditions are satisfied, still request chunks, while still putting the block into orphans.

Kouprin:
PR is closed. See details at comment: https://github.com/nearprotocol/nearcore/pull/1709#issuecomment-567252591.

mzhangmzz:
Implemented in https://github.com/near/nearcore/pull/5224

PR #5201, opened 2021-11-11
store multiple PartialEncodedChunks before processing them instead of only storing one

mzhangmzz:
Upon receiving a PartialEncodedChunk, we store them if the block is not accepted and only process them once the block is accepted. However, the way they are stored right now only allows one PartialEncodedChunk to be stored for each shard and each block. This is not ideal because almost for all cases, the node needs to receive multiple PartialEncodedChunk from different part owners to be able to reconstruct the whole chunk. 

This PR also fixes another bug in `store_partial_encoded_chunk` about verifying chunk header signature. It was wrong before because it uses block producer instead of chunk producer. 

Test Plan:
- existing CI and nayduck test

I will add test later after I implement https://github.com/near/nearcore/issues/1689. They are related and I might change how store_partial_encoded_chunk there again.


mzhangmzz:
I still need to add tests for this PR, but please review this @bowenwang1996 to confirm that my understanding of our current way of requesting and processing PartialEncodedChunks is correct

chain/chunks/src/lib.rs
```diff
@@ -730,34 +732,28 @@ impl ShardsManager {
...
-                            header.signature(),
-                        )
-                        .unwrap_or(false)
-                    {
+                    if runtime_adapter.verify_chunk_header_signature(&header, &epoch_id, known_header.prev_hash()).unwrap_or(false) {
                         // We prove that this one is valid for `epoch_id`.
                         // We won't store it by design if epoch is changed.
-                        *stored_chunk = partial_encoded_chunk.clone();
+                        if stored_chunk.len() < MAX_STORED_PARTIAL_CHUNK_SIZE {
+                            stored_chunk.push(partial_encoded_chunk.clone());
```
bowenwang1996, line 739: Is this attackable?

chain/chain/src/chain.rs
```diff
@@ -717,8 +717,12 @@ impl Chain {
...
                     return Err(ErrorKind::InvalidChunk.into());
                 }
-            } else {
-                if !runtime_adapter.verify_chunk_header_signature(&chunk_header.clone())? {
+            } else if chunk_header.height_created() == block.header().height() {
+                if !runtime_adapter.verify_chunk_header_signature(
+                    &chunk_header.clone(),
+                    block.header().epoch_id(),
+                    block.header().prev_hash(),
+                )? {
```
bowenwang1996, line 725: Could we add a test for this?

bowenwang1996 (commented):
Epic finding! My only concern is that I didn't ask you to look at this part of the code earlier 😅

chain/chain/src/chain.rs
mzhangmzz: What kind of test do you mean? I can also just keep the code behavior here the same if you want.

chain/chunks/src/lib.rs
mzhangmzz: Yeah I think this is not optimal because the validation here only validate the header for the PartialEncodedChunk, so it's possible that someone sends a lot of invalid PartialEncodedChunk messages and saturate this temporary storage. However, these PartialEncodedChunks will be found invalid when the block is ready and the node calls `process_partial_encoded_chunks`. It will not cause any real problem, but delays of block processing. 

I think I can move more validation logic in `process_partial_encoded_chunk` to here, or even just calls `process_parital_encoded_chunk` here instead of simply storing it, but I want to do it in another PR.

chain/chain/src/chain.rs
bowenwang1996: > This PR also fixes another bug in store_partial_encoded_chunk about verifying chunk header signature. It was wrong before because it uses block producer instead of chunk producer.

I am not sure what the bug is actually. Could you add a test for that?

chain/chunks/src/lib.rs
```diff
@@ -57,6 +57,7 @@ const CHUNK_REQUEST_RETRY_MAX_MS: u64 = 1_000_000;
 const CHUNK_FORWARD_CACHE_SIZE: usize = 1000;
 const ACCEPTING_SEAL_PERIOD_MS: i64 = 30_000;
 const NUM_PARTS_REQUESTED_IN_SEAL: usize = 3;
+pub const MAX_STORED_PARTIAL_CHUNK_SIZE: usize = 100;
```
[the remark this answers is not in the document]
bowenwang1996: @mzhangmzz this value is per height right? If so, should we set the value by num_total_parts * num_shards?

chain/chain/src/chain.rs
mzhangmzz: Ah i c, the bug was not related to this line. I just refactored the code on `verify_chunk_header_signature`, which was related to the bug.

chain/chunks/src/lib.rs
```diff
@@ -730,34 +732,28 @@ impl ShardsManager {
                     let epoch_id = unwrap_or_return!(
                         runtime_adapter.get_epoch_id_from_prev_block(known_header.prev_hash())
                     );
-                    let block_producer =
-                        unwrap_or_return!(runtime_adapter.get_block_producer(&epoch_id, height));
```
mzhangmzz, line 734: @bowenwang1996 here is where the bug regarding chunk producer signature was, I believe this line should be getting chunk producer instead of block producer.

The test on this function didn't catch this because the test only had one validators. I'll change that.

chain/chunks/src/lib.rs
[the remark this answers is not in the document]
mzhangmzz: The value is per height per shard.

nearcore/src/runtime/mod.rs
```diff
@@ -903,16 +903,16 @@ impl RuntimeAdapter for NightshadeRuntime {
         &self,
         chunk_hash: &ChunkHash,
         signature: &Signature,
-        prev_block_hash: &CryptoHash,
+        epoch_id: &EpochId,
+        last_known_hash: &CryptoHash,
```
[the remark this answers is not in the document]
mzhangmzz: I added comments in `verify_chunk_header_signature`

chain/chunks/src/lib.rs
bowenwang1996: Interesting. I wonder why we never caught this in any tests

bowenwang1996 (approved):

chain/chunks/src/lib.rs
mzhangmzz: I think because
1) block producer and chunk producers are the same because of how we assign block producer and chunk producers. 
2) there is not much test for testing the code regarding store_partial_encoded_chunks itself. And even if this part of logic is not exactly correct, it won't affect block processing that much because the partial chunks can be requested again when the block is ready. This will only affect block processing delay, and not many tests in our code check that.

I know I'm not adding any test here for this logic right now. Because
1) As I said, even if the logic is not entirely correct, it won't affect block processing correctness
2) I'm thinking getting rid of this part of logic of `store_partial_encoded_chunks` logic entirely, because they could be just processed instead of stored. I want to make that change after I finish https://github.com/near/nearcore/pull/5224
3) I'll add tests in https://github.com/near/nearcore/pull/5224 to make sure for orphans, chunks are requested and stored so they don't need to be requested again when the orphan is ready to be processed.

PR #5204, opened 2021-11-11
refactor: remove default impl of `VMConfig`

Longarithm:
Test plan:

Existing tests.


bowenwang1996 (approved):

core/primitives-core/src/config.rs
```diff
@@ -86,18 +86,16 @@ pub struct VMLimitConfig {
     pub max_functions_number_per_contract: Option<u64>,
 }
 
-impl Default for VMConfig {
-    fn default() -> VMConfig {
+impl VMConfig {
+    pub fn test() -> VMConfig {
```
mm-near, line 90: shouldn't we mark it as #[cfg(feature = "test_features")] ?

runtime/runtime-params-estimator/src/gas_metering.rs
```diff
@@ -175,10 +175,13 @@ pub fn compute_gas_metering_cost(
     let store = create_store(&get_store_path(workdir.path()));
     let cache_store = Arc::new(StoreCompiledContractCache { store });
     let cache: Option<&dyn CompiledContractCache> = Some(cache_store.as_ref());
-    let vm_config_gas = VMConfig::default();
+    let protocol_version = ProtocolVersion::MAX;
```
mm-near, line 178: add a comment why we're using protocolversion::MAX here

runtime/runtime-params-estimator/src/gas_metering.rs
```diff
@@ -79,7 +79,7 @@ fn test_gas_metering_cost_icount() {
 
 /*
 fn dump_prepared(code: &String) {
-    let config = VMConfig::default();
+    let config = VMConfig::test();
```
mm-near, line 82: does it make sense to keep this commented-out code?

runtime/runtime-params-estimator/src/vm_estimator.rs
```diff
@@ -89,26 +96,23 @@ pub fn cost_per_op(gas_metric: GasMetric, code: &ContractCode) -> Ratio<u64> {
...
+    Ratio::new(measured * (vm_config.regular_op_cost as u64), NUM_ITERATIONS * outcome.burnt_gas)
 }
 
 type CompileCost = (u64, Ratio<u64>);
 
 fn compile(code: &[u8], gas_metric: GasMetric, vm_kind: VMKind) -> Option<CompileCost> {
     let start = start_count(gas_metric);
     for _ in 0..NUM_ITERATIONS {
-        let prepared_code = prepare::prepare_contract(code, &VMConfig::default()).unwrap();
+        let prepared_code = prepare::prepare_contract(code, &VMConfig::test()).unwrap();
```
mm-near, line 115: should you still use the 'test' version of the config here? (and not the one loaded from config_store)

runtime/runtime-params-estimator/src/gas_metering.rs
matklad: Yeah, this actually should be `PROTOCOL_VERSION` I believe. `ProtocolVersion::MAX` sadly looks like something reasonable, but it is not. `ProtocolVersion` is just `u64`, so that is just `!0` bit pattern, and not some logically meaningful protocol version. `PROTOCOL_VERSION` is the constant defined  by us, which means the latest protocol version supported.

core/primitives-core/src/config.rs
matklad: I think not, at least not initially. Currently, we use `test_features` only for things which are actively harmful to keep in the production binary (adversarial features, extra fields in data structures, that stuff). "append only" test code is generally not gated. 

We *could* gate all test code, but then means that `cargo test` would be useless, as it won't actually run any tests, and you'd have to pass `cargo test --features test_features` to run them. What's worse, the test would run *only* against the code with extra test functionality compiled in, which would be different from production code.

runtime/runtime-params-estimator/src/vm_estimator.rs
matklad: Strictly speaking, both are wrong. What we ideally should use in the estimator is unitary config -- a config where all costs are equal to `1` and just count the number of operations. Will hopefully do that after https://github.com/near/nearcore/pull/4733.

Longarithm: @matklad Isn't it better to use runtime config with **last** protocol version to check that after protocol upgrades estimation still works and fees don't change unexpectedly?

Longarithm: @matklad Isn't it better to use runtime config with **last** protocol version to check that estimation still works and fees don't change unexpectedly after protocol upgrades?

ailisp (approved):
Looks good to me!

runtime/runtime-params-estimator/src/vm_estimator.rs
matklad: :thinking: I think I am becoming convinced that that's actually might be indeed better. What I originally overlooked is that `VMConfig` contains not only fees, but also variou other runtime constants like `max_contract_size` and the like. While specific fees should be irrelevant for estimation, those other params might still be relevant.

No super strong opinion either way though.

matklad (approved):
LGTM, with or without replacing `::test` with `::for_protocol_version(PROTOCOL_VERSION)`.

PR #5210, opened 2021-11-12
We shouldn't be cloning PeerId so often

pmnoxx:
We've been cloning `Edge` in order to convert it to `Vec<u8>`. That seems wasteful.

- change `impl From<PeerId> for Vec<u8> {` to `impl From<&PeerId> for Vec<u8> {`

### DIFF
--- a/chain/network/benches/graph.rs
+++ b/chain/network/benches/graph.rs
@@ -13,16 +13,13 @@ fn build_graph(depth: usize, size: usize) -> Graph {
     let mut graph = Graph::new(source.clone());
 
     for i in 0..size {
-        graph.add_edge(source.clone(), nodes[i].clone());
+        graph.add_edge(&source, &nodes[i]);
     }
 
     for layer in 0..depth - 1 {
         for u in 0..size {
             for v in 0..size {
-                graph.add_edge(
-                    nodes[layer * size + u].clone(),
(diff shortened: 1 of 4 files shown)


chain/network/benches/graph.rs
```diff
@@ -70,6 +67,6 @@ benchmark_group!(
 benchmark_main!(benches);
 
 // running 3 tests
-// test calculate_distance_10_10  ... bench:   1,503,222 ns/iter (+/- 126,811)
-// test calculate_distance_10_100 ... bench: 988,705,595 ns/iter (+/- 118,318,208)
-// test calculate_distance_3_3    ... bench:      18,928 ns/iter (+/- 3,174)
+// test calculate_distance_10_10  ... bench:      12,442 ns/iter (+/- 307)
```
mm-near, line 70: wow.. such a difference just due to the reference change? or we simply didn't update the test for a while

mm-near (approved):

pmnoxx:

chain/network/benches/graph.rs
pmnoxx: We didn't update it for a while.  There were a lot of changes.

bowenwang1996 (approved):

PR #5211, opened 2021-11-12
Replace Edge with Arc<Edge> to improve performance

pmnoxx:
Add bench to measure time it takes to copy routing table.

- We should compare what the performance would be if we were using
` HashSet<(PeerId, PeerId), Edge>` vs ` HashSet<(PeerId, PeerId), Arc<Edge>>`

### DIFF
--- a/chain/network/benches/routing_table_actor.rs
+++ b/chain/network/benches/routing_table_actor.rs
@@ -58,11 +58,11 @@ fn get_all_edges_bench_new(bench: &mut Bencher) {
     let all_edges = routing_table_actor.get_all_edges();
     let mut new_edges_info = HashMap::new();
     for edge in all_edges {
-        new_edges_info.insert((edge.key().0.clone(), edge.key().1.clone()), Arc::new(edge.clone()));
+        new_edges_info.insert((edge.key().0.clone(), edge.key().1.clone()), edge.clone());
     }
 
     bench.iter(|| {
-        let result: Vec<Arc<Edge>> = new_edges_info.iter().map(|x| x.1.clone()).collect();
+        let result: Vec<Edge> = new_edges_info.iter().map(|x| x.1.clone()).collect();
         black_box(result);
     });
 }
--- a/chain/network/benches/routing_table_actor.rs
+++ b/chain/network/benches/routing_table_actor.rs
@@ -50,24 +50,6 @@ fn get_all_edges_bench_old(bench: &mut Bencher) {
 }
 
 #[allow(dead_code)]
-fn get_all_edges_bench_new(bench: &mut Bencher) {
-    // this is how we efficient we could make get_all_edges by using Arc
-
-    // 1000 nodes, 10m edges
-    let routing_table_actor = build_graph(10, 100);
-    let all_edges = routing_table_actor.get_all_edges();
-    let mut new_edges_info = HashMap::new();
-    for edge in all_edges {
-        new_edges_info.insert((edge.key().0.clone(), edge.key().1.clone()), edge.clone());
-    }
-
-    bench.iter(|| {
-        let result: Vec<Edge> = new_edges_info.iter().map(|x| x.1.clone()).collect();
-        black_box(result);
-    });
-}
-
-#[allow(dead_code)]
 fn get_all_edges_bench_new2(bench: &mut Bencher) {
     // this is how we efficient we could make get_all_edges by using Arc
 
@@ -93,12 +75,7 @@ fn get_all_edges_bench_new2(bench: &mut Bencher) {
     });
 }
 
-benchmark_group!(
-    benches,
-    get_all_edges_bench_old,
-    get_all_edges_bench_new,
-    get_all_edges_bench_new2
-);
+benchmark_group!(benches, get_all_edges_bench_old, get_all_edges_bench_new2);
 
 benchmark_main!(benches);
 
--- a/chain/network/benches/routing_table_actor.rs
+++ b/chain/network/benches/routing_table_actor.rs
@@ -47,9 +47,6 @@ fn get_all_edges_bench_old(bench: &mut Bencher) {
         let result = routing_table_actor.get_all_edges();
         black_box(result);
     });
-    // println!("took {}ns", bench.ns_per_iter());
-    // This code gets executed multiple tests, so assert will not work.
-    // assert!(((bench.ns_per_iter() as f64) < (5_045_404.0 * TEST_COEF)));
 }
 
 #[allow(dead_code)]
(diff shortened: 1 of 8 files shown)


chain/network/benches/routing_table_actor.rs
```diff
@@ -0,0 +1,74 @@
...
+        black_box(result);
+    });
+}
+
+benchmark_group!(benches, get_all_edges_bench_old, get_all_edges_bench_new);
+
+benchmark_main!(benches);
+
+// running 2 tests
+// test get_all_edges_bench_new ...bench:   1,231,155 ns/iter (+/- 95,776)
```
mm-near, line 73: wow - that's a nice improvement.

looks like we should be using Arc as a the value inside the HashMap
@matklad - PTAL

chain/network/src/routing/routing_table_actor.rs
```diff
@@ -385,6 +385,10 @@ impl RoutingTableActor {
 
         res
     }
+
+    pub fn get_all_edges(&self) -> Vec<Edge> {
+        self.edges_info.iter().map(|x| x.1.clone()).collect()
```
mm-near, line 390: if @matklad  agrees - I'd go ahead and change edges_info to have this Arc within this PR

mm-near:
nit: in the PR description it should be HashMap and not HashSet, right?
(HashSet<(PeerId, PeerId), Edge> --> HashMap<...>)

mm-near (changes requested):

chain/network/src/routing/routing_table_actor.rs
matklad: I think it's fine! Perhaps a bit more rustic design would be to represent both peers and edges as just integer indices into a table which contains the actual data, but managing such a table between two actors would be a pain. 

Given that size_of edge is a whopping `344` bytes, I wonder if we should even move the Arc inside the edge itself? 

```rust
pub struct Edge {
    repr: Arc<EdgeRepr>
}

struct EdgeRepr {
    peer0: PeerId,
    peer1: PeerId,
    ...
}
```

Keeping such large objects on the stack seems inefficient. 

Anyway, just sticking an Arc there by itself is :+1:

matklad (approved):

chain/network/src/routing/routing_table_actor.rs
pmnoxx: @matklad I was able to further optimize the code changing `edge_info` to be `HashMap<Arc<PeerId, PeerId>, Arc<Edge>>`
and `Edge` to be
```rust

pub struct EdgeNew {.
    pub key: Arc<(PeerId, PeerId)>,
    ...
}
```

There are a few reasons to use this structure:
- performance of cloning -This helped reduce the time to clone Edges from `1.25s to 0.95s`. It is really inefficient to have large structures as keys / values in hashmap.

- memory usage per item - `(PeerId, PeerId)` is around 132 bytes, and edge is 344 bytes. That given total of around 500 bytes. Given that in hashmap your occupancy has to  be between 25% to 50%, in the worst case you will have to use 4x as much memory. So you end up using around 2kb per Edge.

Assuming we have `n` edges.
Each element of hashmap would be 16 bytes.
With the new approach: in the worst case we will be using `16*4*n +  (344 + 8) * n =  (64 + 344 + 8) * n = (416) * n`. That is 416 bytes per item.
With old one it would be: `(132+344)*4*n = 1904 * n`. So `1904` bytes per each item.

- In addition we are using `(PeerId, PeerId)` copy twice for each edge  in `SlotMap` inside `ibf_peer_set.rs`. So we can reduce the number of allocations needed.

- We can go even further and reduce duplicates of `PeerId`. We can achieve this by creating `HashSet<Arc<PeerId>>`. And then replacing `PeerId` with a global copy of `Arc<PeerId>`,

Anyway, to be careful, we could make 3 separate changes:
- replace Edge with Arc<Edge>
- replace `key: <(PeerId, PeerId)>,` with `key: Arc<(PeerId, PeerId)>,`
- replace `PeerId` with `Arc<PeerId>`

pmnoxx: > Given that size_of edge is a whopping `344` bytes, I wonder if we should even move the Arc inside the edge itself?

Let's do changes step by step. That will allow us to measure benefits vs having to add extra `Arc<...>`.
I think it makes sense to use `Arc<...>` on all structures that are at least 64, maybe even 32 bytes, that we use as values or keys inside `HashMap` / `HashSet`.

chain/network/benches/routing_table_actor.rs
pmnoxx: We should be using `Arc` for both value and key. I tested it, it made another 30% improvement. Even though technically we don't need keys to make a close, the internal implementation of `HashMap` stores both keys and values together. Having a shared pointer for key will certainly improve performance, as well as we will reduce memory usage.

bowenwang1996 (approved):

pmnoxx:
@matklad Can you look at this PR again?

chain/network/src/routing/routing_table_actor.rs
matklad: >Given that in hashmap your occupancy has to be between 25% to 50%,

I think Rust's hashmap is build for much higher occupancy: 

https://github.com/rust-lang/hashbrown/blob/462666a0bc6738793912c133878e533dbf478b0d/src/raw/mod.rs#L203

>`HashMap<Arc<PeerId, PeerId>, Arc<Edge>>`

I still don't like that we duplicate the key twice here (as peer pair is already stored as part of an Edge), Id try to refactor that to be represented as a has set internally. THough just arcing the key is ok as well probably.

pmnoxx: > > Given that in hashmap your occupancy has to be between 25% to 50%,
> 
> I think Rust's hashmap is build for much higher occupancy:
> 
> https://github.com/rust-lang/hashbrown/blob/462666a0bc6738793912c133878e533dbf478b0d/src/raw/mod.rs#L203
> 
> > `HashMap<Arc<PeerId, PeerId>, Arc<Edge>>`
> 
> I still don't like that we duplicate the key twice here (as peer pair is already stored as part of an Edge), Id try to refactor that to be represented as a has set internally. THough just arcing the key is ok as well probably.

We can make a separate PR for replacing `edges_info` with `HashMap<Arc<PeerId>, Arc<PeerId>, Arc<Edge>>`.
However, first we should do other refactoring. I checked that there wasn't much performance difference between those two.

Btw, `Arc` is adding extra `16` bytes to each allocation. Maybe, we should work `Arc` and only keep `Strong` as `AtomicUsize32`. That would reduce overhead to only 4 bytes.

Issue #5203, opened 2021-11-11
Consider adding LRU cache to RocksDB impl of Database


nearmax:
We haven't had an opportunity to optimize the performance of the RPC server and the ViewClient inside nearcore. Which leads to individual RPC nodes not being able to sustain the needed throughput, which in turn leads to high cloud expenses for running multiple nodes and the need to use throttling. 

Anecdotally, some people were able to optimize RPC server to achieve higher QPS by using Redis in nearcore: https://twitter.com/vgrichina/status/1458289037525934084 . We might be able to achieve QPS increase by simpler means as a temporary measure until we revisit RPC server and ViewClient implementations. I suggest we consider adding simple LRU cache from https://docs.rs/cached/0.26.2/cached/ directly into `get` and `write` methods of `impl Database for RocksDB` here:
https://github.com/near/nearcore/blob/734978677a7491743b5a2b4519b619f263e747ce/core/store/src/db.rs#L469-L563

`write` method would be responsible for invalidating LRU cache entries.

Unfortunately, AFAIR there is a suboptimality in `near-api-js` that forces it re-request all access from an account before signing a transaction. This will likely touch `iter_*` methods (likely `iter_prefix` method) of `impl Database for RocksDB`, which we might also want to wrap in cache. CC @MaximusHaximus , @frol 

We most likely would want to have this cache enabled only for the RPC nodes, and we should let the Contract Runtime team know about its existence, since it might affect the measurements of the param estimator.

bowenwang1996:
Rocksdb has in-memory cache internally and we have it enabled.

nearmax:
> Rocksdb has in-memory cache internally and we have it enabled.

Given the complexity of Rocksdb and that we revisited its configuration only once, I don't think we can be sure that we configured it correctly or that it is optimized for the view calls.

bowenwang1996:
Indeed, we currently set the cache size for all columns to be 32mb https://github.com/near/nearcore/blob/6d8eee6ffbf6dedbae5022bd3d8acd2c2870a6b3/core/store/src/db.rs#L664, but not all columns are equal. I think we can and should allocate more for `ColState`. For example, we can do 256mb or even 512 for `ColState`. @mina86 @pmnoxx given that you worked with Rocksdb before, I wonder whether this is something you could help with.

pmnoxx:
@bowenwang1996 Hmm, we could write some PR like this: https://github.com/near/nearcore/pull/5212

[...]

PR #5212, opened 2021-11-12
Increase cache for RockDB by column

pmnoxx:
We recently found out that our rockdb is not efficient enough.
Therefore we would like to increase cache size of specific columns.
In this PR we will increase cache size for `DBCol::ColState` to 128mb.

We can adjust that cache size later, but we try to be conservative in memory usage. 
Some testing should be done to determine optimize cache sizes. 

Closes https://github.com/near/nearcore/issues/5203

### DIFF
--- a/core/store/src/db.rs
+++ b/core/store/src/db.rs
@@ -672,7 +672,7 @@ fn rocksdb_block_based_options(cache_size: usize) -> BlockBasedOptions {
 
 fn choose_cache_size_mb(col: DBCol) -> usize {
     match col {
-        DBCol::ColState => 256 * 1024 * 1024,
+        DBCol::ColState => 128 * 1024 * 1024,
         _ => 32 * 1024 * 1024,
     }
 }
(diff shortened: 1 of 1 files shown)


core/store/src/db.rs
```diff
@@ -658,22 +658,30 @@ fn rocksdb_read_options() -> ReadOptions {
...
     block_opts.set_pin_l0_filter_and_index_blocks_in_cache(true);
     block_opts.set_cache_index_and_filter_blocks(true);
     block_opts.set_bloom_filter(10, true);
     block_opts
 }
 
+// TODO(#5213) Use ByteSize package to represent sizes.
+fn choose_cache_size(col: DBCol) -> usize {
+    match col {
+        DBCol::ColState => 128 * 1024 * 1024,
```
bowenwang1996, line 675: let's do 512mb for now. It should not be a big issue

bowenwang1996 (commented):
Thanks!

pmnoxx:
@bowenwang1996 Ok, I updated the cache size.

bowenwang1996 (approved):

PR #5215, opened 2021-11-12
Use bytesize for size constants in db.rs

pmnoxx:
Use `ByteSize` for size constants in db.rs. We are currently using `ByteSize` in other places. Let's use it here as well.


matklad (commented):
Given that here we immediately pass the thing into the u64-typed API, I would probably lean towards not using `ByteSize` in this particular case. Although no strong opinion!

pmnoxx:
@matklad I rewrote the change to use u64 constants from that crate.

bowenwang1996 (approved):
Thanks!

Issue #5156, opened 2021-11-06
Write reference level documentation for the network code


pmnoxx:
To make network code easier to understand we should write a reference level documentation. 
That will make it easier for new developers to understand different parts of the networking code.

- [x] https://github.com/near/nearcore/pull/5580
- [x] https://github.com/near/nearcore/pull/5621
- [x] https://github.com/near/nearcore/pull/5622
- [x] https://github.com/near/nearcore/pull/5623
- [x] https://github.com/near/nearcore/pull/5625
- [x] https://github.com/near/nearcore/pull/5626
- [x] https://github.com/near/nearcore/pull/5627

Add in-code documentation
- [x] - `Graph` - https://github.com/near/nearcore/pull/5162
- [x] Add comments to `utils.rs` (https://github.com/near/nearcore/pull/5216)
- [x] Add comments to `test_utils.rs` (https://github.com/near/nearcore/pull/5217)
- [x] Add comments/refactor `rate_counter.rs` (https://github.com/near/nearcore/pull/5218)

Dead code
- [x] remove dead code (https://github.com/near/nearcore/pull/5207)

Improve code quality:
- [x]  Remove not needed clones - https://github.com/near/nearcore/pull/5209
- [x]  Remove 'pub' modifier from `Edge` - https://github.com/near/nearcore/pull/5223
- [x] Use constants for time in db.rs https://github.com/near/nearcore/pull/5215
- [x] Refactor Arc<Edge> to a new struct https://github.com/near/nearcore/pull/5228/files

pmnoxx:
@bowenwang1996 @mm-near @matklad

matklad:
Great idea! I've just spend five minutes reading the doc in the wip PR, and I feel I've already learned more about near networking more than I knew before.

I have one high-level comment here: we might want to ensure strict separation between protocol specification and code-level documentation. Ideally, I'd love to see two documents: "NEAR P2P protocol" and "Implementation of networking in nearcore". The first documents abstract protocol without references to a particular implementation. It doesn't mention Rust, tokio, nearcore or actors, but it does mention TCP, borsh format, persistent nonces etc. So if you want to implement a nearcore node in Go, you can use that document as a spec. The second document describes how we implement protocol in nearcore (eg, "to maintain persistent nonces, we store removed edges into the database using this column"). 

Don't think that we should immediately create that, but it might make sense to keep the separation between proper protocol and specific implementation in mind while writing the docs.

matklad:
[...]

PR #5216, opened 2021-11-12
Add comments to utils.rs

pmnoxx:
Add missing comments to `utils.rs`

Related to https://github.com/near/nearcore/issues/5156

### DIFF
--- a/chain/network/src/utils.rs
+++ b/chain/network/src/utils.rs
@@ -5,6 +5,12 @@ use cached::SizedCache;
 
 use crate::types::{BlockedPorts, PatternAddr};
 
+// `blacklist_from_iter` - converts list of addresses represented by strings to <IpAddr, BlockedPorts> HashMap
+//
+// Arguments:
+// - `blacklist`- list of strings in following formats:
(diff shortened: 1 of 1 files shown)


chain/network/src/utils.rs
```diff
@@ -5,6 +5,12 @@ use cached::SizedCache;
 
 use crate::types::{BlockedPorts, PatternAddr};
 
+// `blacklist_from_iter` - converts list of addresses represented by strings to <IpAddr, BlockedPorts> HashMap
```
mm-near, line 8: nit: you might want to use '///' -- this way this is marked as 'documentation' rather than just a comment (and some of the rust tools can pick it up)

chain/network/src/utils.rs
```diff
@@ -5,6 +5,12 @@ use cached::SizedCache;
...
+//
+// Arguments:
+// - `blacklist`- list of strings in following formats:
+//    - "IP" - for example 127.0.0.1
```
mm-near, line 12: I'd also mention that if only ip is provided, we'll block all the ports

mm-near (approved):

PR #5218, opened 2021-11-12
Add comments/refactor rate_counter.rs

pmnoxx:
Add comments/refactor rate_counter.rs

We also need to fix a bug in RateCounter https://github.com/near/nearcore/issues/5226

Mentioned in https://github.com/near/nearcore/issues/5156

### DIFF
--- a/chain/network/src/peer/rate_counter.rs
+++ b/chain/network/src/peer/rate_counter.rs
@@ -13,13 +13,16 @@
 // limitations under the License.
 
 use std::collections::VecDeque;
-use std::time::SystemTime;
+use std::time::{Duration, SystemTime};
 
-const MINUTE_IN_MILLIS: u128 = 60_000;
+const MINUTE: Duration = Duration::from_secs(60);
 
+/// Stores list of entries for `RateCounter`
 struct Entry {
+    /// bytes since last reset
     bytes: u64,
-    expiration_timestamp: u128,
+    /// Time we created the entry.
+    recorded: SystemTime,
 }
 
 /// A rate counter tracks number of transfers, the amount of data exchanged and the rate of transfer
@@ -36,8 +39,8 @@ impl RateCounter {
 
     /// Increment number of bytes transferred, updating counts and rates.
     pub fn increment(&mut self, bytes: u64) {
-        let now = millis_since_epoch();
-        self.entries.push_back(Entry { bytes, expiration_timestamp: now + MINUTE_IN_MILLIS });
+        let now = SystemTime::now();
+        self.entries.push_back(Entry { bytes, recorded: now });
         self.bytes_sum += bytes;
         self.truncate(now);
     }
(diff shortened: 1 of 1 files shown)


chain/network/src/stats/rate_counter.rs
```diff
@@ -15,10 +15,13 @@
 use std::collections::VecDeque;
 use std::time::{Duration, SystemTime};
 
-const MINUTE_IN_MILLIS: u128 = 60_000;
+const MINUTE_IN_MILLIS: Duration = Duration::from_secs(60);
```
mm-near, line 18: I think I'd actually leave it as u128 -- as we only add it to the other integer (and I'm not sure how much calling '.to_millis' add to compute cost)

or - we could 'migrate' all the things to be a duration: expiration_timestamp and return values from millis_since_epoch

chain/network/src/stats/rate_counter.rs
```diff
@@ -15,10 +15,13 @@
...
 
+/// Stores list of entries for `RateCounter`
 struct Entry {
+    /// bytes since last reset
     bytes: u64,
+    /// Time when we measured num `bytes` plus 1 minute.
```
mm-near, line 24: time when this entry should no longer be counted (which is currently set to time when we measured it plus 1 minute)

chain/network/src/stats/rate_counter.rs
```diff
@@ -37,7 +40,8 @@ impl RateCounter {
     /// Increment number of bytes transferred, updating counts and rates.
     pub fn increment(&mut self, bytes: u64) {
         let now = millis_since_epoch();
-        self.entries.push_back(Entry { bytes, expiration_timestamp: now + MINUTE_IN_MILLIS });
+        self.entries
+            .push_back(Entry { bytes, expiration_timestamp: now + MINUTE_IN_MILLIS.as_millis() });
```
mm-near, line 44: meta comment:

seems that this class (RateCounter) might not work properly if there is no data coming in (as then truncate is not called and bytes_per_min / count_per_min show results from the past).

we should add a clear warning about it to the class comment (and check if this can cause us trouble).

Alternatively, we could fix it, by having a bytes_per_min/count_per_min actually call the truncate - if truncate wasn't called recently.

mm-near (changes requested):

chain/network/src/stats/rate_counter.rs
pmnoxx: Ok, I changed while `rate_counter.rs` to use `SystemTime`/`Duration`

chain/network/src/stats/rate_counter.rs
pmnoxx: I rewrote code to store current time. This will be much easier to understand.

pmnoxx:
@mm-near Can you take a look again?

bowenwang1996 (commented):
> Resolves #5156

I believe this will close the issue, which is not what you want

chain/network/src/stats/rate_counter.rs
pmnoxx: I think it's correct. I changed how `expiration_timestamp` is used to represent current time, and make the logic less confusing.

pmnoxx: Fixed in https://github.com/near/nearcore/pull/5225/files

chain/network/src/stats/rate_counter.rs
```diff
@@ -50,21 +53,14 @@ impl RateCounter {
         self.entries.len() as u64
     }
 
-    fn truncate(&mut self, now: u128) {
-        while !self.entries.is_empty() && self.entries.front().unwrap().expiration_timestamp < now {
+    fn truncate(&mut self, now: SystemTime) {
+        // Remove entries longer than 1m.
```
mm-near, line 57: s/longer/older

mm-near (approved):

chain/network/src/stats/rate_counter.rs
pmnoxx: Fixed

pmnoxx:
Rebasing due to 
```


[2021-11-15T22:37:20Z] Mon, 15 Nov 2021 22:37:20 GMT JsonRpcProvider.experimental_protocolConfig() deprecated use `experimental_protocolConfig({ sync_checkpoint: 'genesis' })` to fetch the up-to-date or genesis protocol config explicitly at test/providers.test.js:289:37
--
  | [2021-11-15T22:37:21Z]  FAIL  test/providers.test.js (40.981 s)
  | [2021-11-15T22:37:21Z]   ● Console


```

PR #5223, opened 2021-11-12
REFACTOR Remove unused method from Edge struct

pmnoxx:
`key` and `nonce` attributes of `Edge` have `pub` modifier. We are going to remove `pub` modifier to make internal structure of `Edge` private. This will be used in the future, to change `Edge` internal representation to make `PeerManager`, `RoutingTableActor` more efficient.

For example, we could make Edge use `Arc<PeerId>` to remove duplicates of signatures, thus reducing size of `Edge` and improving efficiency.
Or change key to use `Arc<PeerId, PeerId>`, etc. This would remove having to store duplicates of key, in a few hashmaps like `edges_info`, and ones used for `IBF` algorithm.

### DIFF
--- a/chain/network/src/peer_manager/peer_manager_actor.rs
+++ b/chain/network/src/peer_manager/peer_manager_actor.rs
@@ -300,7 +300,7 @@ impl PeerManagerActor {
             if !edge.contains_peer(&self.my_peer_id) {
                 continue;
             }
-            let key = edge.get_pair();
+            let key = edge.key();
             if !self.routing_table_view.is_local_edge_newer(&key, edge.nonce()) {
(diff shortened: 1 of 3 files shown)


matklad (approved):
LGTM!

I'd also consider moving Edge to a dedicated file. With private fields, it's useful to understand that they are inaccessable outside of a specific file.

mm-near (approved):

pmnoxx:
> LGTM!
> 
> I'd also consider moving Edge to a dedicated file. With private fields, it's useful to understand that they are inaccessable outside of a specific file.

That's a good idea. I already wrote and merged PR for that.

PR #5224, opened 2021-11-12
Request chunks for multiple blocks simultaneously

mzhangmzz:
implements #1689

This PR include the following changes:
- implement `ShardsManager::request_chunks_for_orphan`. The existing function `request_chunks` does not work for orphan because it requires the previous block has been accepted. `request_chunks_for_orphan` gets around that by using an ancestor block of the orphan block for requesting chunks. The ancestor block must satisfies: 1) get_epoch_id_from_prev_block(ancestor_hash) has the same epoch_id as the current block and 2) it is accepted.
- implement logic in chain regarding when to request missing chunks for orphans. We check whether we can request missing chunks for a orphan block when 
   1) it is first added to the orphan pool
   2) a new block has been accepted, check the descendants of this block to see if any of them satisfy the requirements for requesting missing chunks


chain/chain/src/chain.rs
```diff
@@ -76,6 +76,12 @@ pub const MAX_ORPHAN_SIZE: usize = 1024;
 /// Maximum age of orhpan to store in the chain.
 const MAX_ORPHAN_AGE_SECS: u64 = 300;
 
+// Number of orphan ancestors should be checked to request chunks
+const NUM_ORPHAN_ANCESTORS_CHECK: u64 = 5;
+
+// Maximum number of orphans that we can request missing chunks
+const MAX_ORPHAN_MISSING_CHUNKS: usize = 100;
```
bowenwang1996, line 83: how are these number set?

chain/chain/src/chain.rs
```diff
@@ -119,6 +125,12 @@ impl BlockLike for Orphan {
     }
 }
 
+impl Orphan {
+    fn prev_hash(&self) -> CryptoHash {
+        *self.block.header().prev_hash()
```
bowenwang1996, line 130: I think it is usually better to return a reference and let caller decide whether they want to clone/deference

chain/chain/src/chain.rs
```diff
@@ -200,6 +216,37 @@ impl OrphanBlockPool {
...
+        let mut queue = vec![(prev_hash, 0)];
+        while let Some((prev_hash, depth)) = queue.pop() {
+            if depth == target_depth {
+                continue;
+            }
+            if let Some(block_hashes) = self.prev_hash_idx.get(&prev_hash) {
+                for hash in block_hashes {
+                    // there should be no loop, but just in case
+                    if visited.insert(*hash) {
+                        queue.push((*hash, depth + 1));
```
bowenwang1996, line 235: Actually since this property must hold (otherwise something is seriously wrong), I suggest that we avoid this checking and add some `debug_assert` to further improve performance. This should also allow us to change visited to `Vec` (which also slightly improves performance)

chain/chain/src/chain.rs
```diff
@@ -200,6 +216,37 @@ impl OrphanBlockPool {
...
+    pub fn get_orphans_within_depth(
+        &self,
+        prev_hash: CryptoHash,
+        target_depth: u64,
+    ) -> Vec<CryptoHash> {
+        let mut visited = HashSet::new();
+        let mut queue = vec![(prev_hash, 0)];
+        while let Some((prev_hash, depth)) = queue.pop() {
+            if depth == target_depth {
+                continue;
```
bowenwang1996, line 229: noob question: isn't it better to break here?

chain/chain/src/chain.rs
```diff
@@ -208,6 +255,7 @@ pub struct Chain {
     store: ChainStore,
     pub runtime_adapter: Arc<dyn RuntimeAdapter>,
     orphans: OrphanBlockPool,
+    orphans_requested_missing_chunks: HashSet<CryptoHash>,
```
bowenwang1996, line 258: I wonder whether it make sense to make this part of `OrphanBlockPool`

chain/chain/src/chain.rs
```diff
@@ -1206,6 +1277,63 @@ impl Chain {
...
+    /// Check if we can request chunks for this orphan. Conditions are
+    /// 1) Among the `NUM_ORPHAN_ANCESTORS_CHECK` immediate parents of the block at least one is
+    ///    accepted(is in store), call it `ancestor`
+    /// 2) The next block of `ancestor` has the same epoch_id as the orphan block
+    ///    (This is because when requesting chunks, we will use `ancestor` hash instead of the
+    ///     previous block hash of the orphan to decide epoch id
+    /// 3) All the immediate parents of the block after `ancestor` are either orphans or in
+    ///     `blocks_with_missing_chunks`
+    /// 4) Orphans that with outstanding missing chunks request has not exceed `MAX_ORPHAN_MISSING_CHUNKS`
+    /// 5) Orphans have missing chunks
```
bowenwang1996, line 1289: This is great! I think it could be even better if you annotate in the function body the boundary of each step

chain/chain/src/chain.rs
```diff
@@ -1206,6 +1277,63 @@ impl Chain {
...
+        orphan: &Block,
+    ) -> Option<OrphanMissingChunksInfo> {
+        if self.orphans_requested_missing_chunks.len() >= MAX_ORPHAN_MISSING_CHUNKS {
+            return None;
+        }
+        if self.orphans_requested_missing_chunks.contains(orphan.hash()) {
+            return None;
+        }
+        let mut block_hash = *orphan.header().prev_hash();
+        for _ in 1..NUM_ORPHAN_ANCESTORS_CHECK {
```
bowenwang1996, line 1302: is it intentional that this starts from 1? So we only loop `NUM_ORPHAN_ANCESTORS_CHECK - 1` times?

chain/chain/src/chain.rs
```diff
@@ -1206,6 +1277,63 @@ impl Chain {
...
+                                        epoch_id,
+                                        ancestor_hash: block_hash,
+                                    })
+                                }
+                                _ => None,
+                            };
+                        }
+                    }
+                }
+                return None;
```
bowenwang1996, line 1322: I looked at where this function is called. It seems to me that this part of the code may be called multiple times for a block if it is deep in the orphan pool, since we call it every time we call `check_orphan`. We may want to consider caching some intermediate results

chain/client/src/client.rs
```diff
@@ -707,6 +737,7 @@ impl Client {
         // TODO: replace to channels or cross beams here? we don't have multi-threading here so it's mostly to get around borrow checker.
         let accepted_blocks = Arc::new(RwLock::new(vec![]));
         let blocks_missing_chunks = Arc::new(RwLock::new(vec![]));
+        let orphans_missing_chunks = Arc::new(RwLock::new(vec![]));
```
bowenwang1996, line 740: fyi I believe the use of `Arc<RwLock<..>>` here is completely unnecessary and can be simply replaced by `Rc<RefCell<..>>`. It was written at a time when Rust was still foreign to us...

chain/chain/src/chain.rs
mzhangmzz: That's legit, I changed it to break here

chain/chain/src/chain.rs
mzhangmzz: The second number is somewhat random, just to set a max number and that should almost never be hit, unless the chain has many forks at the same time. 

The first number was set because both ChunkCache(stores processed PartialEncodedChunks) and stored_partial_encoded_chunks(stores unprocessed PartialEncodedChunks) only stores chunks within 5 blocks of the current head, so requesting something larger than that doesn't make sense.

Looking at the chunks delay metric provided by Nikolay, it may make sense to change both of these numbers to 10, because that's usually how much a node is behind. But I think we can try 5 now and if that doesn't solve the issue, we can make the number higher.

chain/chain/src/chain.rs
mzhangmzz: Yeah it was intentional, but I was overthinking, iterating from 0 was correct

chain/client/src/client.rs
mzhangmzz: I'll create an issue that tracks this

chain/chain/src/chain.rs
mzhangmzz: The function checks if this orphan is already in `orphans_requested_missing_chunks` and skip the rest of the checks if it is. This check is also only called for a new orphan or children of a block that is just accepted that are within depth 5. The only case where I think this function can be called multiple times on the an orphan is if the orphan does not have missing chunks when it was first added, which is actually impossible now, since a node can't process chunks for orphans.

chain/chunks/src/lib.rs
```diff
@@ -637,24 +643,25 @@ impl ShardsManager {
...
 
-            // If the protocol version is later than the one where chunk forwarding was introduced,
+            // If chunk forwarding is enabled,
             // we purposely do not send chunk request messages right away for new blocks. Such requests
             // will eventually be sent because of the `resend_chunk_requests` loop. However,
             // we want to give some time for any `PartialEncodedChunkForward` messages to arrive
             // before we send requests.
             let is_chunk_forwarding_enabled =
                 checked_feature!("stable", ForwardChunkParts, protocol_version);
+            let is_chunk_forwarding_enabled = false;
```
mzhangmzz, line 660: It's a temporary hack here. The issue is that chunk parts won't be forwarded to non-validators, so the code here should check whether the node is a validator or not and only delay the request if it's a validator.

nearcore/src/runtime/mod.rs
```diff
@@ -224,7 +224,7 @@ impl NightshadeRuntime {
             home_dir,
             store,
             genesis,
-            TrackedConfig::new_empty(),
+            TrackedConfig::AllShards,
```
mzhangmzz, line 227: Also a hack here, will correct it soon

chain/chunks/src/lib.rs
mzhangmzz: waiting for https://github.com/near/nearcore/pull/5360 to be merged

chain/chain/src/chain.rs
mzhangmzz: done

mzhangmzz:
@bowenwang1996 @mm-near could you take a look at this PR again? I tried to add more comments, but let me know if you think it needs more

chain/client/src/client.rs
```diff
@@ -879,6 +883,7 @@ impl Client {
                             pec_v2.header.height_created(),
                             pec_v2.header.shard_id(),
                         );
+                        debug!("accept chunk");
```
bowenwang1996, line 886: is this intended?

chain/client/src/client_actor.rs
```diff
@@ -387,7 +387,9 @@ impl Handler<NetworkClientMessages> for ClientActor {
                                 }
                                 return NetworkClientResponses::NoResponse;
                             } else if block.hash() == sync_hash {
-                                if let Err(e) = self.client.chain.save_orphan(block) {
+                                // This is the immediate block after a state sync
+                                // We can afford to delay requesting missing chunks for this one block
+                                if let Err(e) = self.client.chain.save_orphan(block, false) {
```
bowenwang1996, line 392: Thanks for adding comments here! I also want to note that this is a ridiculous hackery and there is no guarantee that the block is still in the orphan pool when the chunks arrive

nearcore/src/runtime/mod.rs
bowenwang1996: Is it corrected :)

bowenwang1996 (commented):
I'll leave it to @mm-near to review this PR

mm-near (approved):

nearcore/src/runtime/mod.rs
mzhangmzz: now it is! :)

matklad (approved):

chain/chain/src/tests/mod.rs
```diff
@@ -12,6 +12,6 @@ impl Chain {
         me: &Option<AccountId>,
         block: Block,
     ) -> Result<Option<Tip>, Error> {
-        self.process_block(me, block, Provenance::PRODUCED, |_| {}, |_| {}, |_| {})
+        self.process_block(me, block, Provenance::PRODUCED, |_| {}, |_| {}, |_| {}, |_| {})
```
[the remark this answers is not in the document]
mzhangmzz: That's a good idea. I'll do that in another PR

Issue #5221, opened 2021-11-12
Improve Routing Table exchange by fasctor of 5x


pmnoxx:
There are a lot of inefficiencies in network code, we should fix them:

- [x] Replace Edge with Arc<Edge> - This reduced the amount of time `RoutingTableActor` spends on creating copy of routing table from `5s` to `1.3s` (https://github.com/near/nearcore/pull/5211)
- [x] Replace PeerId with Arc<PeerId>- This reduced the amount of time `RoutingTableActor` spends on creating copy of routing table from `1.3s` to `1.0s` (https://github.com/near/nearcore/pull/5227)
- [x] Remove cloning of `Edge` that's not needed (https://github.com/near/nearcore/pull/5209)
- [x] Reduce cloning of `PeerId` (https://github.com/near/nearcore/pull/5210)

PR #5227, opened 2021-11-13
Replace PeerId with Arc<PeerId>

pmnoxx:
We ran a test, and we found that that we can replace old routing table exchange by 30% if we replace `PeerId` with `Arc<PeerId>`.

Mentioned in https://github.com/near/nearcore/issues/5221

Before
```
running 3 tests
test calculate_distance_10_10  ... bench:      12,261 ns/iter (+/- 329)
test calculate_distance_10_100 ... bench:   1,333,230 ns/iter (+/- 20,893)
test calculate_distance_3_3    ... bench:         732 ns/iter (+/- 16)
```

After
```
test calculate_distance_10_10  ... bench:      11,488 ns/iter (+/- 220)
test calculate_distance_10_100 ... bench:     672,570 ns/iter (+/- 58,342)
test calculate_distance_3_3    ... bench:         616 ns/iter (+/- 5)
```

Before
```
test get_all_edges_bench_old  ... bench:   1,001,797 ns/iter (+/- 383,813)
```
After:
```                                                                                                                                                                                               
test get_all_edges_bench_old  ... bench:     761,119 ns/iter (+/- 463,381)            
```

### DIFF
--- a/chain/network/Cargo.toml
+++ b/chain/network/Cargo.toml
@@ -21,6 +21,7 @@ near-rust-allocator-proxy = "0.3.0"
 bytesize = "1.0.1"
 conqueue = "0.4.0"
 serde = { version = "1", features = ["derive", "rc", "alloc"], optional=true }
+once_cell = "1.5.2"
 
 borsh = { version = "0.9", features = ["rc"]}
 cached = "0.23"
--- a/chain/network/Cargo.toml
+++ b/chain/network/Cargo.toml
@@ -20,8 +20,7 @@ strum = { version = "0.20", features = ["derive"] }
 near-rust-allocator-proxy = "0.3.0"
 bytesize = "1.0.1"
 conqueue = "0.4.0"
-serde = { version = "1", features = ["derive"], optional=true }
-once_cell = "1.5.2"
+serde = { version = "1", features = ["derive", "rc", "alloc"], optional=true }
 
 borsh = { version = "0.9", features = ["rc"]}
 cached = "0.23"
--- a/core/primitives/src/network.rs
+++ b/core/primitives/src/network.rs
@@ -1,5 +1,6 @@
 use std::fmt;
 use std::hash::Hash;
+use std::sync::Arc;
 
 use borsh::{BorshDeserialize, BorshSerialize};
 use serde::{Deserialize, Serialize};
@@ -11,11 +12,27 @@ use crate::types::{AccountId, EpochId};
 
 /// Peer id is the public key.
 #[derive(
+    BorshSerialize, BorshDeserialize, Clone, Eq, PartialOrd, Ord, Hash, Serialize, Deserialize,
+)]
+pub struct PeerId(Arc<PeerIdInner>);
+
+/// Peer id is the public key.
+#[derive(
     BorshSerialize, BorshDeserialize, Clone, Eq, PartialOrd, Ord, Serialize, Deserialize, Hash,
 )]
-pub struct PeerId(PublicKey);
+pub struct PeerIdInner(PublicKey);
 
 impl PeerId {
+    pub fn new(key: PublicKey) -> Self {
+        Self(Arc::new(PeerIdInner(key)))
+    }
+
+    pub fn public_key(&self) -> &PublicKey {
+        &self.0 .0
+    }
+}
+
+impl PeerIdInner {
     pub fn new(key: PublicKey) -> Self {
         Self(key)
     }
@@ -39,7 +56,7 @@ impl From<&PeerId> for Vec<u8> {
 
 impl From<PublicKey> for PeerId {
     fn from(public_key: PublicKey) -> PeerId {
-        PeerId(public_key)
+        PeerId::new(public_key)
     }
 }
 
@@ -47,11 +64,17 @@ impl TryFrom<Vec<u8>> for PeerId {
     type Error = Box<dyn std::error::Error>;
 
     fn try_from(bytes: Vec<u8>) -> Result<PeerId, Self::Error> {
-        Ok(PeerId(PublicKey::try_from_slice(&bytes)?))
+        Ok(PeerId::new(PublicKey::try_from_slice(&bytes)?))
     }
 }
 
 impl PartialEq for PeerId {
+    fn eq(&self, other: &Self) -> bool {
+        self.0 .0 == other.0 .0
+    }
+}
+
+impl PartialEq for PeerIdInner {
     fn eq(&self, other: &Self) -> bool {
         self.0 == other.0
     }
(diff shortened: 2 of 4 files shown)


core/primitives/src/network.rs
```diff
@@ -9,13 +10,29 @@ use near_crypto::{KeyType, PublicKey, SecretKey, Signature};
...
-pub struct PeerId(pub PublicKey);
+pub struct PeerIdInner(pub PublicKey);
 
 impl PeerId {
+    pub fn new(key: PublicKey) -> Self {
+        Self(Arc::new(PeerIdInner(key)))
+    }
+
+    pub fn public_key(&self) -> PublicKey {
+        self.0 .0.clone()
```
mm-near, line 31: space before a dot (or is it on purpose?)

core/primitives/src/network.rs
```diff
@@ -9,13 +10,29 @@ use near_crypto::{KeyType, PublicKey, SecretKey, Signature};
 use crate::hash::{hash, CryptoHash};
 use crate::types::{AccountId, EpochId};
 
+/// Peer id is the public key.
+#[derive(
+    BorshSerialize, BorshDeserialize, Clone, Eq, PartialOrd, Ord, Hash, Serialize, Deserialize,
+)]
+pub struct PeerId(Arc<PeerIdInner>);
```
mm-near, line 17: @matklad - PTAL - I'm surprised that the performance difference is so large.

chain/network/Cargo.toml
```diff
@@ -21,7 +21,7 @@ strum = { version = "0.20", features = ["derive"] }
 near-rust-allocator-proxy = "0.3.0"
 bytesize = "1.0.1"
 conqueue = "0.4.0"
-serde = { version = "1", features = ["derive"], optional=true }
+serde = { version = "1", features = ["derive", "rc", "alloc"], optional=true }
```
mm-near, line 24: why do we need the rc/alloc? do we (somehow) change how this is serialized?

core/primitives/src/network.rs
matklad: In older rusts, `self.0.0.clone()` is be lexed as `self. 0.0 .clone`, that is, `0.0` would be treated as a floating point literal by the compiler. So adding a space used to be an idiom for this. Modern rust has a hack in the compiler to handle this in intuitive way, but I guess rustfmt still adds a space to be on a safe side.

core/primitives/src/network.rs
```diff
@@ -9,13 +10,29 @@ use near_crypto::{KeyType, PublicKey, SecretKey, Signature};
...
+
 /// Peer id is the public key.
 #[derive(
     BorshSerialize, BorshDeserialize, Clone, Eq, PartialOrd, Ord, Serialize, Deserialize, Hash,
 )]
-pub struct PeerId(pub PublicKey);
+pub struct PeerIdInner(pub PublicKey);
```
matklad, line 23: Doesn't need to be `pub`

core/primitives/src/network.rs
matklad: This feeld roughly right to me, tbh. The sizeof goes from 65 to just 8. Given that we use, eg, `HashMap<PeerId, _>` or `Vec<PeerId>`, the savings here seem like they can be substantial. 

What is surprising though, is that we see perf change for `calculate_distance`. That code doesn't operate with `PeerId`s at all, it's all just indices and bitmasks. We only need PeerId when we convert the result of the algorithm back to the answer. That is, the runtime of the algorithm itelf is englible in comparison to operations with PeerId to return the answer of the right shape.

This makes me think that maybe we should reconsider the repr here, and try to use a rustic impl with indices? Basically, we'd use `PeerId(u32)` internally everywhere, and use the full PublicKey repr only when sending a message? That means lifting `p2id` and `id2p` hasmaps somwhere to PeerManager (?) layer.

core/primitives/src/network.rs
```diff
@@ -39,38 +56,44 @@ impl From<&PeerId> for Vec<u8> {
 
 impl From<PublicKey> for PeerId {
```
matklad, line 57: Unrelated, but this impl seems semantically wrong. It means that any public key is a PeerId, but that's not true. Some PKs identify things which are not PeerId. We only should have `From` impl in the other direction.

core/primitives/src/network.rs
pmnoxx: > This feeld roughly right to me, tbh. The sizeof goes from 65 to just 8. Given that we use, eg, `HashMap<PeerId, _>` or `Vec<PeerId>`, the savings here seem like they can be substantial.
> 
> What is surprising though, is that we see perf change for `calculate_distance`. That code doesn't operate with `PeerId`s at all, it's all just indices and bitmasks. We only need PeerId when we convert the result of the algorithm back to the answer. That is, the runtime of the algorithm itelf is englible in comparison to operations with PeerId to return the answer of the right shape.

@matklad `Calculate_distance` calls ` self.compute_result(&mut routes, &distance)` to compute the result.
We spend most of the time converting the result to  `HashMap<PeerId, Vec<PeerId>>`

> This makes me think that maybe we should reconsider the repr here, and try to use a rustic impl with indices? Basically, we'd use `PeerId(u32)` internally everywhere, and use the full PublicKey repr only when sending a message? That means lifting `p2id` and `id2p` hasmaps somwhere to PeerManager (?) layer.

@matklad The best solution would be to keep a global structure representing list of all created `PublicKeys`. We would reduce data usage a bit, but I need to think more about it. It would be best to do this while decoding data from the network to completely avoid doing the memory allocation.

Honestly, using Arc<PublicKey> is not that but, we could save a lot of memory, by simply removing the duplicates. Though having `PeerId(u32)` , would have it easier to compute hash functions. However, that would require us to keep track when PeerId is no longer used, or risk leaking memory.

The easier way to do this, would be to override `PeerId` -> `BorshSerialize` to do a call global structure, which keep track of all duplicates. However, that has it's own issues. We probably don't want static structure for unit tests.

bowenwang1996 (approved):

chain/network/Cargo.toml
pmnoxx: @mm-near Yes. `Edge` uses `BorshSerialize` decorator, which doesn't work on `Arc` by default, unless you enabled both "rc", and "alloc". `BorshSerialize` is designed to not require `std` to work. They have chosen not to include `std` by default.

core/primitives/src/network.rs
pmnoxx: Btw. `p2id` and `id2p` are not inside `RoutingTableActor`. 
Anyway, let's do other changes first, then we can talk about maybe replacing it. 

Keep in mind that using ` PeerId(u32)` by itself will not be enough. At some point you need to serialize it and convert it to `PeerId(PublicKey)`.`Will RoutingTableActor do it?` Or maybe should that be done by `PeerActor`. Though that requires inner process synchronization, which may not be a bad idea.

Other way would be to have a global cache of `PeerId<Arc<...>>`, we would use it while decoding to remove duplicates.

We have a few design choices. First, I want to do changes, that are good, and obvious how to do it.

core/primitives/src/network.rs
pmnoxx: @matklad Yeah, also there is no real benefit to doing that. Why not just use `PeerId::new(public_key)`. Having a few ways to do the same thing just makes the code harder to read. If I see somewhere in the code `public_key.into()`. I wouldn't expect it to convert itself to `PeerId`.Those are essentially different types. There is a reason why we don't have `Impl From<u64> for SystemTime`, and other weird usage of `From` like that.

pmnoxx: I'll do it in another PR.

core/primitives/src/network.rs
pmnoxx: @pmnoxx I'm thinking about rewriting `PeerId` inner to have same format at `Edge`. That is that most methods will be in `PeerId`.

core/primitives/src/network.rs
pmnoxx: Fixed

PR #5228, opened 2021-11-13
Replace Arc<Edge> with a structure

pmnoxx:
As @matklad suggested, it's a good idea to replace `Arc<Edge>` with a structure, to make code more readable.

### DIFF
--- a/chain/network/src/routing/routing.rs
+++ b/chain/network/src/routing/routing.rs
@@ -132,8 +132,11 @@ impl Edge {
     /// to verify the signature.
     pub fn partial_verify(peer0: PeerId, peer1: PeerId, edge_info: &EdgeInfo) -> bool {
         let pk = peer1.public_key();
-        let (peer0, peer1) = Edge::make_key(peer0, peer1);
-        let data = Edge::build_hash(&peer0, &peer1, edge_info.nonce);
+        let data = if peer0 < peer1 {
+            Edge::build_hash(&peer0, &peer1, edge_info.nonce)
+        } else {
+            Edge::build_hash(&peer1, &peer0, edge_info.nonce)
+        };
         edge_info.signature.verify(data.as_ref(), &pk)
     }
 
--- a/chain/network/src/routing/routing.rs
+++ b/chain/network/src/routing/routing.rs
@@ -51,8 +51,8 @@ pub struct EdgeInfo {
 
 impl EdgeInfo {
     pub fn new(peer0: PeerId, peer1: PeerId, nonce: u64, secret_key: &SecretKey) -> Self {
-        let (peer0, peer1) = EdgeInner::make_key(peer0, peer1);
-        let data = EdgeInner::build_hash(&peer0, &peer1, nonce);
+        let (peer0, peer1) = Edge::make_key(peer0, peer1);
+        let data = Edge::build_hash(&peer0, &peer1, nonce);
         let signature = secret_key.sign(data.as_ref());
         Self { nonce, signature }
     }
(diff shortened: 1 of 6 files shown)


chain/network/src/routing/routing.rs
```diff
@@ -65,7 +65,95 @@ pub enum EdgeType {
...
+    }
+}
+
+impl std::ops::Deref for Edge {
+    type Target = EdgeInner;
+
+    fn deref(&self) -> &EdgeInner {
+        &self.0
+    }
+}
```
mm-near, line 156: comment to line 177 below -- add comment on what is EdgeInner (and what is the 'split' between Edge and EdgeInner)

chain/network/src/routing/routing.rs
```diff
@@ -65,7 +65,95 @@ pub enum EdgeType {
     Removed,
 }
 
-pub type Edge = Arc<EdgeInner>;
+#[derive(BorshSerialize, BorshDeserialize, Clone, Debug, PartialEq, Eq)]
+#[cfg_attr(feature = "test_features", derive(Serialize, Deserialize))]
+pub struct Edge(pub Arc<EdgeInner>);
+
+impl Edge {
```
mm-near, line 72: add comment to explain what is an edge (and that this is an undirected connection)

chain/network/src/routing/routing.rs
```diff
@@ -65,7 +65,95 @@ pub enum EdgeType {
...
+        peer0: PeerId,
+        peer1: PeerId,
+        nonce: u64,
+        signature0: Signature,
+        signature1: Signature,
+    ) -> Self {
+        Edge(Arc::new(EdgeInner::new(peer0, peer1, nonce, signature0, signature1)))
+    }
+
+    pub fn make_fake_edge(peer0: PeerId, peer1: PeerId, nonce: u64) -> Self {
```
mm-near, line 84: comment - also is it only used for testing or do we use it in production?

chain/network/src/routing/routing.rs
```diff
@@ -65,7 +65,95 @@ pub enum EdgeType {
...
+            key: (peer0, peer1),
+            nonce,
+            signature0: Signature::empty(KeyType::ED25519),
+            signature1: Signature::empty(KeyType::ED25519),
+            removal_info: None,
+        }))
+    }
+
+    /// Build a new edge with given information from the other party.
+    pub fn build_with_secret_key(
```
mm-near, line 95: nit: the function below is called 'make' this one is called 'build' -> maybe align the names?

chain/network/src/routing/routing.rs
```diff
@@ -65,7 +65,95 @@ pub enum EdgeType {
...
+            Self::build_hash(&peer0, &peer1, nonce)
+        } else {
+            Self::build_hash(&peer1, &peer0, nonce)
+        };
+        let signature0 = secret_key.sign(hash.as_ref());
+        Self::new(peer0, peer1, nonce, signature0, signature1)
+    }
+
+    /// Build the hash of the edge given its content.
+    /// It is important that peer0 < peer1 at this point.
```
mm-near, line 112: if that's true we should check this condition inside (or allow both and do the flip inside)

chain/network/src/routing/routing.rs
```diff
@@ -65,7 +65,95 @@ pub enum EdgeType {
...
+    fn build_hash(peer0: &PeerId, peer1: &PeerId, nonce: u64) -> CryptoHash {
+        let mut buffer = Vec::<u8>::new();
```
mm-near, line 114: you can mention that the hash is simply a hash of the concat of peer0, peer1 and nonce

chain/network/src/routing/routing.rs
```diff
@@ -65,7 +65,95 @@ pub enum EdgeType {
...
+impl Edge {
+    /// Create an addition edge.
+    pub fn new(
+        peer0: PeerId,
+        peer1: PeerId,
+        nonce: u64,
+        signature0: Signature,
+        signature1: Signature,
+    ) -> Self {
+        Edge(Arc::new(EdgeInner::new(peer0, peer1, nonce, signature0, signature1)))
```
mm-near, line 81: if this is the 'addition edge' - should we check that nonce is even?

mm-near (approved):

chain/network/src/routing/routing.rs
matklad: And there's also `new`. Without understanding the context, `new`, `new_with_secret_key`, `new_fake` is what I'd expect.

chain/network/src/routing/routing.rs
```diff
@@ -65,7 +65,95 @@ pub enum EdgeType {
...
+        if nonce % 2 == 1 {
+            nonce + 2
+        } else {
+            nonce + 1
+        }
+    }
+}
+
+impl std::ops::Deref for Edge {
+    type Target = EdgeInner;
```
matklad, line 151: It feels stange that both `Edge` and `EdgeInner` are public, and this operates via deref. 

Id probably expect Edge to be public, `EdgeInner` to be private, and `Edge` to have a bunch of accessor functions.

matklad (commented):
If we start to have Edge and `EdgeInner`, ther's an argument for  moving Edge to a separte file, to clearly separate Edge public inteface (a bunch of accessor functions) from internal impl details (an arc-allocated data)

chain/network/src/routing/routing.rs
pmnoxx: Ok, I'll submit a separate PR for commenting this.

chain/network/src/routing/routing.rs
pmnoxx: Ok, I'll submit a separate PR for commenting this.

chain/network/src/routing/routing.rs
pmnoxx: Ok, I created another PR for that. https://github.com/near/nearcore/pull/5267

pmnoxx:
> And there's also new. Without understanding the context, new, new_with_secret_key, new_fake is what I'd expect.

@matklad Yes, we should refactor that code and rename methods to something that makes more sense.

chain/network/src/routing/routing.rs
pmnoxx: Done in https://github.com/near/nearcore/pull/5267

chain/network/src/routing/routing.rs
pmnoxx: @matklad I agree, I'll do it in a separate PR.

chain/network/src/routing/routing.rs
pmnoxx: That's going to change in https://github.com/near/nearcore/pull/5267. I added a comment there.

chain/network/src/routing/routing.rs
pmnoxx: I'll add comments in one PR after https://github.com/near/nearcore/pull/5267

chain/network/src/routing/routing.rs
pmnoxx: I'll add comments in one PR after https://github.com/near/nearcore/pull/5267

chain/network/src/routing/routing.rs
pmnoxx: I'll add comments in one PR after https://github.com/near/nearcore/pull/5267

chain/network/src/routing/routing.rs
pmnoxx: I'll add comments in one PR after https://github.com/near/nearcore/pull/5267

PR #5229, opened 2021-11-13
Dont clone PublicKey

pmnoxx:
We shouldn't be cloning `PublicKey`.


mm-near (approved):

core/primitives/src/network.rs
```diff
@@ -20,8 +20,8 @@ impl PeerId {
         Self(key)
     }
 
-    pub fn public_key(&self) -> PublicKey {
-        self.0.clone()
+    pub fn public_key(&self) -> &PublicKey {
```
matklad, line 23: Feels a bit strange that `peer_id.0` and `public_key()` are both public synonyms for the same thing, I'd probably expect the field to be private and to have `From<PeerId> for PublicKey` if we need it.

matklad (approved):

frol (approved):

PR #5230, opened 2021-11-13
feat: Move `EdgeVerifierActor` inside `RoutingTableActor`

pmnoxx:
Currently, `PeerManager` is responsible to sending edges to check their signatures `EdgeVerifierActor`.
Then `PeerManager` runs `broadcast_edges_trigger` trigger, which is time consuming.
And then `PeerManager` sends edges to `RoutingTableActor`.

This has a few issues:
- `PeerManager` is spends a bit of time running  `broadcast_edges_trigger`.
- There is an extra structure in `EdgeVerifierActor` which leaks memory and can be removed
- `PeerManager` doesn't care about the edges by itself. This adds extra complexity in the logic.

In the current design:
We create `RoutingTableActor` first then `PeerManagerActor` We pass `Addr<RoutingTableActor>` to `PeerManagerActor`, and we keep references to for both.
- `RoutingTableActor` create `EdgeVerifierActor` and also talks to ` RoutingTableActor`. 
- When new edges are received from network. `PeerManagerActor` sends edges to `EdgeVerifierActors`, reads reply from a `ConcurrentQueue`, and sends them to `RoutingTableActor` for `RoutingTableActor` to update it's storage of `Edges` and updating `RoutingTable`.

That design is a bit compileted:
The final design in 3-4 PRs.
- We create `PeerManagerActor`, `PeerManagerActor` creates `RoutingTableActor`. Then then `RoutingTableActor` created `EdgeVerifierActor`.
- When edges are received from Network. `PeerManagerActor` will send a message with edges to `RoutingTableActor`.
`RoutingTableActor` actor will send edges to `EdgeVerifierActor`, read the response, and update the routing table.

In conclusion:
- `PeerManager` will be less loaded
- We will fix memory leak
- `Code` structure will be easier to understand.

Part 1: `RoutingTableActor` goes inside `RouringTableActor`
- [x] implement ban peer logic
- [x] filter out edges before sending them to `EdgeVerifier`
- [x] Fix tests

Part 2: Offload work from `PeerManagerActor`
- [ ] Move `trigger` from `PeerManager` to `RoutingTableActor`. 

Part 3: Fix memory leak
- [ ] Rewrite how we use `EdgeVerifierActor`

Part 4: Move `RoutingTableActor` fully back to `PeerManagerActor
- [ ] Rewrite how `JsonRpcServer` communicates with `PeerManagerActor`
- [ ] Initially this was done to add support for `JsonRpcServer`, that can be done in a better way.

### DIFF
--- a/chain/network/src/peer_manager/peer_manager_actor.rs
+++ b/chain/network/src/peer_manager/peer_manager_actor.rs
@@ -8,7 +8,7 @@ use crate::peer_manager::peer_store::{PeerStore, TrustLevel};
 ))]
 use crate::routing::edge::SimpleEdge;
 use crate::routing::edge::{Edge, EdgeInfo, EdgeType};
-use crate::routing::edge_verifier_actor::{EdgeVerifierActor, EdgeVerifierHelper};
+use crate::routing::edge_verifier_actor::EdgeVerifierHelper;
 use crate::routing::routing::{
     PeerRequestResult, RoutingTableView, DELETE_PEERS_AFTER_TIME, MAX_NUM_PEERS,
 };
@@ -28,7 +28,7 @@ use crate::types::{RoutingSyncV2, RoutingVersion2};
 use crate::{PeerInfo, RoutingTableActor, RoutingTableMessages, RoutingTableMessagesResponse};
 use actix::{
     Actor, ActorFuture, Addr, Arbiter, AsyncContext, Context, ContextFutureSpawner, Handler,
-    Recipient, Running, StreamHandler, SyncArbiter, WrapFuture,
+    Recipient, Running, StreamHandler, WrapFuture,
 };
 #[cfg(feature = "delay_detector")]
 use delay_detector::DelayDetector;
@@ -164,8 +164,6 @@ pub struct PeerManagerActor {
     local_peer_pending_update_nonce_request: HashMap<PeerId, u64>,
     /// Dynamic Prometheus metrics
     network_metrics: NetworkMetrics,
-    /// EdgeVerifierActor, which is responsible for veryfing edges.
-    edge_verifier_pool: Addr<EdgeVerifierActor>,
     /// RoutingTableActor, responsible for computing routing table, routing table exchange, etc.
     routing_table_addr: Addr<RoutingTableActor>,
     /// Shared counter across all PeerActors, which counts number of `RoutedMessageBody::ForwardTx`
@@ -175,9 +173,6 @@ pub struct PeerManagerActor {
     pending_incoming_connections_counter: Arc<AtomicUsize>,
     /// Number of active peers, used for rate limiting.
     peer_counter: Arc<AtomicUsize>,
-    /// Number of edge verifications in progress; We will not update routing table as long as
-    /// this number is non zero.
-    edge_verifier_requests_in_progress: u64,
     /// Used for testing, for disabling features.
     adv_helper: AdvHelper,
 }
@@ -230,8 +225,6 @@ impl PeerManagerActor {
         debug!(target: "network", "Found known peers: {} (boot nodes={})", peer_store.len(), config.boot_nodes.len());
         debug!(target: "network", "Blacklist: {:?}", config.blacklist);
 
-        let edge_verifier_pool = SyncArbiter::start(4, || EdgeVerifierActor {});
-
         let my_peer_id: PeerId = PeerId::new(config.public_key.clone());
         let routing_table = RoutingTableView::new(my_peer_id.clone(), store);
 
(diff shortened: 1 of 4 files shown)


pmnoxx:
@matklad I'm not sure where those crashes are coming from. Any ideas?

```


[2021-11-13T09:02:55Z] ---- inactive_recent_edge stdout ----
--
  | [2021-11-13T09:02:55Z] thread 'inactive_recent_edge' panicked at 'System is not running', /var/lib/buildkite-agent/.cargo/registry/src/github.com-1ecc6299db9ec823/actix-rt-2.3.0/src/system.rs:99:21
  | [2021-11-13T09:02:55Z]


```

chain/network/src/routing/routing_table_actor.rs
```diff
@@ -89,6 +101,22 @@ impl RoutingTableActor {
...
+                   at ./tests/cache_edges.rs:44:28
+         8: cache_edges::empty
+                   at ./tests/cache_edges.rs:183:20
+         9: cache_edges::empty::{{closure}}
+                   at ./tests/cache_edges.rs:182:1
+        10: core::ops::function::FnOnce::call_once
+
+
+               */
+        let edge_verifier_pool = SyncArbiter::start(4, || EdgeVerifierActor {});
```
bowenwang1996, line 119: What is this?

pmnoxx: @bowenwang1996 That has been fixed. Some tests were not working due to now starting system with `System::new()`

pmnoxx:
@mm-near Can you take a look?

mm-near:
Can you explain a little bit more in the PR description on why it would improve code cleanness ?

pmnoxx:
@mm-near I updated the PR with a better explanation.

chain/network/src/routing/routing_table_actor.rs
```diff
@@ -392,6 +408,11 @@ impl Actor for RoutingTableActor {
     type Context = Context<Self>;
 
     fn started(&mut self, _ctx: &mut Self::Context) {}
+
+    /// Try to gracefully disconnect from active peers.
```
mm-near, line 412: what active peers?

chain/network/src/routing/routing_table_actor.rs
```diff
@@ -408,10 +429,38 @@ impl RoutingTableActor {
 impl Handler<StopMsg> for RoutingTableActor {
     type Result = ();
     fn handle(&mut self, _: StopMsg, _ctx: &mut Self::Context) -> Self::Result {
+        self.edge_verifier_pool.do_send(StopMsg {});
         System::current().stop();
     }
 }
 
+impl Handler<EdgeList> for RoutingTableActor {
```
mm-near, line 437: add comment on what this handler does

chain/network/src/peer_manager/peer_manager_actor.rs
```diff
@@ -1212,33 +1206,22 @@ impl PeerManagerActor {
         );
     }
 
-    fn verify_edges(&mut self, ctx: &mut Context<Self>, peer_id: PeerId, edges: Vec<Edge>) {
+    fn verify_edges(&mut self, _ctx: &mut Context<Self>, peer_id: PeerId, edges: Vec<Edge>) {
```
mm-near, line 1209: can you mention what this method does?

as it seems to be banning peers that send the invalid edges - but I don't know what happens to the invalid edges themselves. Are we removing them somewhere from the list?

chain/network/src/routing/routing_table_actor.rs
```diff
@@ -408,10 +429,38 @@ impl RoutingTableActor {
...
+        self.edge_verifier_pool
+            .send(msg)
+            .into_actor(self)
+            .map(move |res, act, _| {
+                act.edge_verifier_requests_in_progress -= 1;
+
+                if let Ok(false) = res {
+                    act.peers_to_ban.push(peer_id);
+                }
+                // TODO missing ban peer logic
```
mm-near, line 455: what is the 'ban peer logic' that is missing - can you explain more?

mm-near (approved):

chain/network/src/routing/routing_table_actor.rs
pmnoxx: "active peers" in this case means peers we are connected to that are in "active_peers" struct.

We should rename it.
https://github.com/near/nearcore/issues/5424

chain/network/src/peer_manager/peer_manager_actor.rs
pmnoxx: https://github.com/near/nearcore/pull/5429

chain/network/src/routing/routing_table_actor.rs
pmnoxx: See updated comments in https://github.com/near/nearcore/pull/5429

chain/network/src/routing/routing_table_actor.rs
pmnoxx: See updated comments in https://github.com/near/nearcore/pull/5429

chain/network/src/routing/routing_table_actor.rs
pmnoxx: Sorry, that's an old comment which was there while I was rewriting the code.
Right now `RoutingTableActor` sends to `PeerManagerActor` periodically, list of peers to ban, for sending us invalid edges.

PR #5231, opened 2021-11-13
Replace constants with Duration

pmnoxx:
There are still some places where we should be using `Duration` in `PeerActor`.


bowenwang1996 (approved):

chain/network/src/peer/peer_actor.rs
```diff
@@ -68,10 +68,10 @@ const MAX_TXNS_PER_BLOCK_MESSAGE: usize = 1000;
 
 /// The time we wait for the response to a Epoch Sync request before retrying
 // TODO #3488 set 30_000
-pub const EPOCH_SYNC_REQUEST_TIMEOUT_MS: u64 = 1_000;
+pub const EPOCH_SYNC_REQUEST_TIMEOUT_MS: Duration = Duration::from_millis(1_000);
```
bowenwang1996, line 71: Probably we don't need the `MS` suffix after the change

yordanmadzhunkov (approved):
Good job

PR #5232, opened 2021-11-13
Remove deprecated code std::u64::MAX

pmnoxx:
Usage of `std::u64::MAX` is deprecated in Rust. Let's use `u64::MAX` instead


matklad (approved):
:+1: 

Sometimes I get lazy and spell this just as `!0`, but that's definitely a bad advice :)

bowenwang1996 (approved):

EgorKulikov (approved):

PR #5234, opened 2021-11-13
Use saturating_duration_since

pmnoxx:
We have seen crashes related to using `duration_since`. Instead we should start using `saturating_durection_since` in `near-network`.


chain/network/src/stats/rate_counter.rs
```diff
@@ -59,9 +59,7 @@ impl RateCounter {
 
 /// Returns timestamp in milliseconds.
 fn millis_since_epoch() -> u128 {
-    let since_epoch = SystemTime::now()
-        .duration_since(SystemTime::UNIX_EPOCH)
-        .unwrap_or_else(|_| Duration::new(0, 0));
+    let since_epoch = SystemTime::now().duration_since(SystemTime::UNIX_EPOCH).unwrap_or_default();
```
mm-near, line 62: you're removing this method in a different PR anyway (so you can ignore it here)

mm-near (approved):

chain/network/src/stats/rate_counter.rs
pmnoxx: Ok, I can fix merge requests later.

PR #5235, opened 2021-11-13
Mark some of near-network crates as private

pmnoxx:
Mark some of near-network crates as private.

Also move 
```

/// The time we wait for the response to a Epoch Sync request before retrying
// TODO #3488 set 30_000
pub const EPOCH_SYNC_REQUEST_TIMEOUT_MS: u64 = 1_000;
/// How frequently a Epoch Sync response can be sent to a particular peer
// TODO #3488 set 60_000
pub const EPOCH_SYNC_PEER_TIMEOUT_MS: u64 = 10;

```

to `ClientActor`.

### DIFF
--- a/chain/client/src/client.rs
+++ b/chain/client/src/client.rs
@@ -23,10 +23,7 @@ use near_chain::{
 use near_chain_configs::ClientConfig;
 use near_chunks::{ProcessPartialEncodedChunkResult, ShardsManager};
 use near_network::types::{PartialEncodedChunkResponseMsg, PeerManagerMessageRequest};
-use near_network::{
-    FullPeerInfo, NetworkClientResponses, NetworkRequests, PeerManagerAdapter,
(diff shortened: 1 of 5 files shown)


chain/client/src/client.rs
```diff
@@ -57,6 +54,13 @@ use near_network::types::PartialEncodedChunkForwardMsg;
 
 const NUM_REBROADCAST_BLOCKS: usize = 30;
 
+/// The time we wait for the response to a Epoch Sync request before retrying
+// TODO #3488 set 30_000
+pub const EPOCH_SYNC_REQUEST_TIMEOUT_MS: u64 = 1_000;
```
bowenwang1996, line 59: shouldn't they be `Duration`?

bowenwang1996 (approved):

chain/client/src/client.rs
pmnoxx: @bowenwang1996 Yes, I split changing type and moving them out to make it easier to review.

PR #5237, opened 2021-11-13
refactor: use near_network_primitives directly

pmnoxx:
Hey, this is going to be an epic change, but it will improve code quality.

We currently have a package called `near-network-utils`. For some reason we are important all types through `near-network`.
From not on we will import all packages directly.

Also `blacklist_from_iter` got moved out of `near-network`. It's not even used there.

The final goal is to remove dependency on `near-network` from packages that don't need it.


nearcore/src/config.rs
```diff
@@ -36,6 +37,45 @@ use near_primitives::version::PROTOCOL_VERSION;
...
 use near_telemetry::TelemetryConfig;
 
+/// `blacklist_from_iter` - converts list of addresses represented by strings to <IpAddr, BlockedPorts> HashMap
+///
+/// Arguments:
+/// - `blacklist`- list of strings in following formats:
+///    - "IP" - for example 127.0.0.1 - if only IP is provided we will block all ports
+///    - "IP:PORT - for example 127.0.0.1:2134
+pub fn blacklist_from_iter<T>(blacklist: T) -> HashMap<IpAddr, BlockedPorts>
+where
```
bowenwang1996, line 47: I am not sure whether it makes sense for this function to be in this file. It feels like a function specific to network

pmnoxx: @bowenwang1996 It doesn't. I removed it out and moved it out to another package in another PR.

pmnoxx: @bowenwang1996 This function is not used in network at all. It's used only for parsing config. Though we can put it inside near-network-primitives if you think that's better.

bowenwang1996: yeah that might be better

chain/network-primitives/src/utils.rs
```diff
@@ -0,0 +1,42 @@
+use crate::types::{BlockedPorts, PatternAddr};
+use std::collections::{HashMap, HashSet};
+use std::net::IpAddr;
+
+/// `blacklist_from_iter` - converts list of addresses represented by strings to <IpAddr, BlockedPorts> HashMap
```
mm-near, line 5: nit: you don't need to repeat the name of the function ('blacklist_from_iter')

mm-near (approved):

pmnoxx:
@frol @chefsale @bowenwang1996 Can you take a look?

chefsale (approved):

frol:
I delegate my vote to @miraclx since he has more context around -primitive crates

miraclx (approved):
LGTM

PR #5240, opened 2021-11-13
Stop using lazy macro

pmnoxx:
In general it's best to avoid macros if possible.

We will replace `lazy_static! {} ` with `Lazy<...>`.


chain/network/src/stats/metrics.rs
```diff
@@ -10,73 +10,77 @@ use near_metrics::{
...
-        try_create_int_gauge("near_peer_connections_total", "Number of connected peers");
-    pub static ref PEER_DATA_RECEIVED_BYTES: near_metrics::Result<IntCounter> =
-        try_create_int_counter("near_peer_data_received_bytes", "Total data received from peers");
-    pub static ref PEER_MESSAGE_RECEIVED_TOTAL: near_metrics::Result<IntCounter> =
-        try_create_int_counter(
-            "near_peer_message_received_total",
-            "Number of messages received from peers"
-        );
-    pub static ref PEER_CLIENT_MESSAGE_RECEIVED_TOTAL: near_metrics::Result<IntCounter> =
+pub static PEER_CONNECTIONS_TOTAL: Lazy<near_metrics::Result<IntGauge>> =
```
matklad, line 13: Unrelated to the PR, but it always felt odd to me that we expose laziness and fallibility in the interface here, and that the call-site needs to call `inc_gauge` with a result. Feels like a rather awkward wrapping of the underlying nice API.

I'd expect one of the two here: 

* TiKV style usage which just unwraps on error during creation: https://github.com/tikv/tikv/blob/0b11c480475e910f3e626793040aa2d225e93aed/components/resolved_ts/src/metrics.rs#L7-L11
* Our own wrapper which does proper error reporting and fallbacks to no-op on erros: 

Sketch:

```rust
static PEER_CONNECTIONS_TOTAL = NearIncGauge::new("near_peer_connections_total", "Number of connected peers");

pub struct NearIntGauge {
  name: &'static str,
  desc: &'static str
  inner: OnceCell<prometheus::IntGauge>,
}

impl NearIntGauge {
  pub const fn new(name, desc) -> NearIntGauge { ... }
  pub fn inc(&'static self) {
    let inner = self.inner.get_or_try_init(|| try_create_int_gauge(..).map_err(|err| tracing::error!("failed to create a gauge: {}", err)));
    if let Ok(inner) = inner { inner.inc() }
  }
}
```

matklad (approved):

chain/network/src/stats/metrics.rs
pmnoxx: @matklad How about we do this instead: https://github.com/near/nearcore/pull/5265

bowenwang1996 (approved):

PR #5241, opened 2021-11-13
Reorganize imports in near-network

pmnoxx:
Imports in many `near-network` files are not organized properly.

We usually follow order `std`, external packages, our packages, `crate::`

### DIFF
--- a/chain/network/src/peer_manager/peer_manager_actor.rs
+++ b/chain/network/src/peer_manager/peer_manager_actor.rs
@@ -15,13 +15,10 @@ use actix::{
 };
 use futures::task::Poll;
 use futures::{future, Stream, StreamExt};
-use near_primitives::time::Clock;
 use tokio::net::{TcpListener, TcpStream};
 use tokio::sync::Semaphore;
 use tracing::{debug, error, info, trace, warn};
 
-use crate::stats::metrics;
-use crate::stats::metrics::NetworkMetrics;
 #[cfg(feature = "delay_detector")]
 use delay_detector::DelayDetector;
 use near_performance_metrics::framed_write::FramedWrite;
@@ -29,6 +26,7 @@ use near_performance_metrics_macros::perf;
 use near_primitives::checked_feature;
 use near_primitives::hash::CryptoHash;
 use near_primitives::network::{AnnounceAccount, PeerId};
+use near_primitives::time::Clock;
 use near_primitives::types::{AccountId, ProtocolVersion};
 use near_primitives::utils::from_timestamp;
 use near_store::Store;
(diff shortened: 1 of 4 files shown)


chain/network/src/peer_manager/peer_manager_actor.rs
```diff
@@ -38,6 +36,8 @@ use tokio_util::sync::PollSemaphore;
 use crate::peer::peer_actor::PeerActor;
 use crate::peer_manager::peer_store::{PeerStore, TrustLevel};
 use crate::routing::codec::Codec;
+use crate::stats::metrics;
+use crate::stats::metrics::NetworkMetrics;
```
mm-near, line 40: @matklad - as this is rust code style question.

But from what I've seen in Zulip, we should aim at having 'one single block of imports' and let cargo fmt handle it

matklad: >we should aim at having 'one single block of imports' and let cargo fmt handle it

Not so sure -- I know that at least nodeX team expects several blocks of imports (cc @khorolets), as in the title of the PR. 

I think the process here should be:

* take a look at what our current practices are, and pick some rule which is the closest to them
* submit a PR with `docs/style.md` document that documents those practices
* start aligning the code. 

Let me actually do that...

matklad: https://github.com/near/nearcore/pull/5262 -- didn't go as far as figuring out whether we need import grouping, but that's a start.

pmnoxx: @matklad I have a mixed feeling about this. I would rather put all imports together and let cargofmt figure it out. Unfortunately, splitting imports into 4 groups adds a bit of extra maintenance, that's not really bringing us any value.
The only reason why I reformatted network code to be like this, was because, I saw we had a code with imports at the beggining (std, empty line, libraries, empty line, near, empty like create) and then some random imports at the end.

My proposal would be to keep a single block of imports and let rustfmt reformat it.

The reason is that, rustfmt will already split imports in 4 groups.
- packages with `crate::` will be on top
- packages that start with `near-` will be together
- packages with std will be kept together

matklad: Yeah, at this point I am leaning towards not grouping the imports, as that's easier to make consistent.

pmnoxx: There are some bad things about `cargo fmt` is that it doesn't separate libraries from `std` or any other standard packages from imported packages from github.

bowenwang1996 (approved):

PR #5246, opened 2021-11-13
Add RUST_BACKTRACE=1 to CI to debug test failures

pmnoxx:
I've seen test failures that are hard to explain what exactly happened. Can we run CI with `RUST_BACKTRACE=1` to get more debug info?

For example, there is a failure with `unwrap()`. The alternative would be to replace all `.unwrap()` with `.expect("...")`

https://buildkite.com/nearprotocol/nearcore/builds/6673#6086b04c-077a-4ffa-97c0-0dcf5dbd8b7b


pmnoxx:
@bowenwang1996 @mm-near

matklad (approved):

matklad:
We might also consider force-enabling backtraces for `neard` itselfs as well, like this: 

https://github.com/rust-analyzer/rust-analyzer/blob/73668334f05c3446b04116ccc3156240d2d8ab19/crates/rust-analyzer/src/bin/main.rs#L96-L98

chefsale (approved):

PR #5249, opened 2021-11-14
feat: Add database cache settings to `config.json`

pmnoxx:
Hey, this is another epic change. We will add cache configuration for our store. to `config.json`
However, some big changes are needed.

This will allow us to specify cache, for db columns, etc. 

Resolves https://github.com/near/nearcore/issues/5255

TODO
- [x] Test on a node

### DIFF
--- a/core/store/src/db.rs
+++ b/core/store/src/db.rs
@@ -49,6 +49,8 @@ impl Into<io::Error> for DBError {
 /// This enum holds the information about the columns that we use within the RocksDB storage.
 /// You can think about our storage as 2-dimensional table (with key and column as indexes/coordinates).
 // TODO(mm-near): add info about the RC in the columns.
+pub const NUM_DB_COLS: usize = 50;
+
 #[derive(PartialEq, Debug, Copy, Clone, EnumIter, BorshDeserialize, BorshSerialize, Hash, Eq)]
 pub enum DBCol {
     /// Column to indicate which version of database this is.
@@ -201,9 +203,6 @@ pub enum DBCol {
     ColStateChangesForSplitStates = 49,
 }
 
-// Do not move this line from enum DBCol
-pub const NUM_COLS: usize = 50;
-
 impl std::fmt::Display for DBCol {
     fn fmt(&self, formatter: &mut std::fmt::Formatter<'_>) -> Result<(), std::fmt::Error> {
         let desc = match self {
@@ -272,8 +271,8 @@ impl DBCol {
 
 // List of columns for which GC should be implemented
 
-pub static SHOULD_COL_GC: [bool; NUM_COLS] = {
-    let mut col_gc = [true; NUM_COLS];
+pub static SHOULD_COL_GC: [bool; NUM_DB_COLS] = {
+    let mut col_gc = [true; NUM_DB_COLS];
     col_gc[DBCol::ColDbVersion as usize] = false; // DB version is unrelated to GC
     col_gc[DBCol::ColBlockMisc as usize] = false;
     // TODO #3488 remove
@@ -297,8 +296,8 @@ pub static SHOULD_COL_GC: [bool; NUM_COLS] = {
 
 // List of columns for which GC may not be executed even in fully operational node
 
-pub static SKIP_COL_GC: [bool; NUM_COLS] = {
-    let mut col_gc = [false; NUM_COLS];
+pub static SKIP_COL_GC: [bool; NUM_DB_COLS] = {
+    let mut col_gc = [false; NUM_DB_COLS];
     // A node may never restarted
     col_gc[DBCol::ColStateHeaders as usize] = true;
     // True until #2515
@@ -308,8 +307,8 @@ pub static SKIP_COL_GC: [bool; NUM_COLS] = {
 
 // List of reference counted columns
 
-pub static IS_COL_RC: [bool; NUM_COLS] = {
-    let mut col_rc = [false; NUM_COLS];
+pub static IS_COL_RC: [bool; NUM_DB_COLS] = {
+    let mut col_rc = [false; NUM_DB_COLS];
     col_rc[DBCol::ColState as usize] = true;
     col_rc[DBCol::ColTransactions as usize] = true;
     col_rc[DBCol::ColReceipts as usize] = true;
@@ -475,7 +474,11 @@ impl RocksDBOptions {
     }
 
     /// Opens the database in read/write mode.
-    pub fn read_write<P: AsRef<std::path::Path>>(self, path: P) -> Result<RocksDB, DBError> {
+    pub fn read_write<P: AsRef<std::path::Path>>(
+        self,
+        path: P,
+        extra_db_memory: usize,
+    ) -> Result<RocksDB, DBError> {
         use strum::IntoEnumIterator;
         let options = self.rocksdb_options.unwrap_or_else(|| rocksdb_options());
         let cf_names = self
@@ -486,7 +489,7 @@ impl RocksDBOptions {
                 .map(|col| {
                     ColumnFamilyDescriptor::new(
                         format!("col{}", col as usize),
-                        rocksdb_column_options(col),
+                        rocksdb_column_options(col, extra_db_memory),
                     )
                 })
                 .collect()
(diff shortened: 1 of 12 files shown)


pmnoxx:
@ilblackdragon @bowenwang1996 @mm-near We could implement something like this.

bowenwang1996 (commented):
Have you tested that this works on a local node?

pmnoxx:
> Have you tested that this works on a local node?

It's on my TODO list.

pmnoxx:
@bowenwang1996 Yes, I tested it works.

Do we have a default template of `config.json` somewhere in the code. I would be good to update the template at the same time.
Also, who manages how to which version of `config.json` to choose? Is it the team maintaining `nearup`?

pmnoxx:
@bowenwang1996 

Here is how example config looks like
```
{
        ...
	"db_settings": {
		"default_col_settings": {
			"new_lru_cache": 64000000
		},
		"col_settings":{
			"ColState": {
				"new_lru_cache": 64000005
			}
		}
	}
}

```

pmnoxx:
@bowenwang1996 I started a node, it works fine. I printed col settings to make sure.

pmnoxx:
@chefsale @mhalambek We are making changes to config can you take a look?

core/store/src/lib.rs
```diff
@@ -45,6 +47,30 @@ pub mod migrations;
...
 mod trie;
 
+/// Describe database settings for column.
+///
+/// If needed we can add more settings.
+#[derive(Serialize, Deserialize, Clone, Debug)]
+pub struct DbColSettings {
+    /// Lru cache for column
+    #[serde(skip_serializing_if = "Option::is_none")]
+    new_lru_cache: Option<u64>,
```
frol, line 57: Let's see how it looks like in the config file:

```json
"db_settings": {
  "col_settings": {
    "ColState": {
      "new_lru_cache": 10
     }
  },
  "default_col_settings": {
    "new_lru_cache": 10
  }
}
```

I know that we already have a config file that is hard to reason about, but this does not help us...

core/store/src/migrations.rs
```diff
@@ -138,7 +138,7 @@ pub fn migrate_6_to_7(path: &Path) {
 }
 
 pub fn migrate_7_to_8(path: &Path) {
-    let store = create_store(path);
+    let store = create_store(path, None);
```
frol, line 141: Instead of adding a ton of `, None`, I would introduce a new `create_store_with_settings` method in addition to `create_store` and leave the old code as is, and the new code can pass db_settings explicitly without a need to wrap it into `Some`

core/store/src/lib.rs
```diff
@@ -45,6 +47,30 @@ pub mod migrations;
...
+
+/// Describe database settings, for example cache_sizes.
+///
+/// If needed we can add implement settings.
+#[derive(Serialize, Deserialize, Clone, Debug)]
+pub struct DbSettings {
+    /// Mapping from col id to size of cache in bytes.
+    /// See `enum DBCol` in `core/store/src/db.rs`
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub col_settings: Option<HashMap<DBCol, DbColSettings>>,
```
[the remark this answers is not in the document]
pmnoxx: @frol That was suggestion by @ilblackdragon
Now that I think about it. I makes sense not to add it, unless someone wants to use it. That person would have to be an engineer though.

core/store/src/migrations.rs
pmnoxx: @frol Yeah, after I wrote the change I realized I don't like it either.

pmnoxx:
@frol I would be in favor of abandoning this PR. Unless there are some people who want to find tune settings from with the config.
If there are then we should probably expose more settings. Besides, how can someone fine tune settings, if we don't have benchmarks? Or good DB performance stats.

bowenwang1996:
> @frol I would be in favor of abandoning this PR. Unless there are some people who want to find tune settings from with the config. If there are then we should probably expose more settings. Besides, how can someone fine tune settings, if we don't have benchmarks? Or good DB performance stats.

I suggest we only allow people to chance the cache size. For that there is no need to fine tune -- some people run more expensive hardware and they can have larger cache size if they want

pmnoxx:
> > @frol I would be in favor of abandoning this PR. Unless there are some people who want to find tune settings from with the config. If there are then we should probably expose more settings. Besides, how can someone fine tune settings, if we don't have benchmarks? Or good DB performance stats.
> 
> I suggest we only allow people to chance the cache size. For that there is no need to fine tune -- some people run more expensive hardware and they can have larger cache size if they want

Ok, I'll change the API to be simpler. Then we can take another look.

pmnoxx:
@frol @bowenwang1996 @mm-near I rewrote the change to use a single db setting:
- `extra_db_memory_bytes` - specifies extra mount of memory that will be used to improve database performance. In the first iteration, we can spend the cache evenly across all columns.

core/store/src/db.rs
```diff
@@ -46,6 +46,9 @@ impl Into<io::Error> for DBError {
     }
 }
 
+/// Number of DBCol columns.
```
pmnoxx, line 49: @matklad This probably can be written in a better way.

core/store/src/db.rs
```diff
@@ -46,6 +46,9 @@ impl Into<io::Error> for DBError {
...
+static DB_NUM_COLS: u64 = 50;
```
pmnoxx, line 50: @matklad This can probably be written in a better way.

bowenwang1996: wait I think `NUM_COLS` already exists in this file

pmnoxx: @bowenwang1996 It is, but it was in a wrong place.

core/store/src/db.rs
```diff
@@ -46,6 +46,10 @@ impl Into<io::Error> for DBError {
     }
 }
 
+/// Number of DBCol columns.
+/// Do not move this line from enum DBCol.
```
frol, line 50: I feel the warning is outdated

frol (approved):
LGTM

core/store/src/db.rs
pmnoxx: @frol I feel the same way. Regardless, of the warning, the variable got moved anyway.

I removed the warnings, as it serves no purpose.

bowenwang1996 (approved):

nikurt (approved):

chefsale (approved):

pmnoxx:
@matklad @olonho Can you take a look?

pmnoxx:
```
@mm-near: /core/ /chain/ /nearcore/
@matklad: /runtime/runtime-params-estimator/
```
@mm-near @matklad Can you take a look?

core/store/src/db.rs
```diff
@@ -49,6 +49,8 @@ impl Into<io::Error> for DBError {
 /// This enum holds the information about the columns that we use within the RocksDB storage.
 /// You can think about our storage as 2-dimensional table (with key and column as indexes/coordinates).
 // TODO(mm-near): add info about the RC in the columns.
```
matklad, line 51: The above comment is now misplaced

core/store/src/db.rs
```diff
@@ -201,9 +203,6 @@ pub enum DBCol {
     ColStateChangesForSplitStates = 49,
 }
 
-// Do not move this line from enum DBCol
-pub const NUM_COLS: usize = 50;
```
matklad, line 205: seems like a footgun to move this to top? I think it's intended that that's next to 49?

core/store/src/db.rs
```diff
@@ -475,7 +474,11 @@ impl RocksDBOptions {
     }
 
     /// Opens the database in read/write mode.
-    pub fn read_write<P: AsRef<std::path::Path>>(self, path: P) -> Result<RocksDB, DBError> {
+    pub fn read_write<P: AsRef<std::path::Path>>(
+        self,
+        path: P,
+        extra_db_memory: usize,
+    ) -> Result<RocksDB, DBError> {
```
matklad, line 481: This is a builder object, please add  `fn extra_db_memory(self, amount: bytesize::ByteSize) -> Self` rather than a separate parameter.

core/store/src/db.rs
```diff
@@ -783,8 +790,11 @@ impl RocksDB {
...
     }
 
-    pub fn new<P: AsRef<std::path::Path>>(path: P) -> Result<Self, DBError> {
-        RocksDBOptions::default().read_write(path)
+    pub fn new<P: AsRef<std::path::Path>>(
+        path: P,
+        extra_db_memory: usize,
+    ) -> Result<Self, DBError> {
+        RocksDBOptions::default().read_write(path, extra_db_memory)
     }
```
matklad, line 798: No need to change this -- if the user wants to create RocskDB with non-default options, they should use `RocksDBOptions`

runtime/runtime-params-estimator/src/main.rs
```diff
@@ -114,7 +114,10 @@ fn main() -> anyhow::Result<()> {
             );
 
             let near_config = load_config(&state_dump_path);
-            let store = create_store(&get_store_path(&state_dump_path));
+            let store = create_store_with_extra_cache(
+                &get_store_path(&state_dump_path),
+                near_config.config.extra_db_memory_bytes,
```
matklad, line 119: This should be `0` -- we are interested in estimating worst-case, when the cache is not used. Though, I guess we'll get `0` here as that's the default for config anyway? 

I guess, it's fine as is, though I'd slightly perfer explicit `0`.

matklad (commented):
.

matklad:
The change to estimator looks ok, but I am hesitant to :heavy_check_mark: as I feel moderately strongly about https://github.com/near/nearcore/pull/5249#discussion_r773050783

pmnoxx:
Closing this PR, as it's replaced by #5998

PR #5247, opened 2021-11-14
Fix routing::ping_jump test

pmnoxx:
Add fix to `routing::ping_jump` test
As seen in `https://buildkite.com/nearprotocol/nearcore/builds/6677#5ccd6608-35c5-474b-9cb9-819a4033eccb`

This happened, because updating routing table is not done in `PeerManager` anymore.

PR #5250, opened 2021-11-14
Fix routing_table_sync after changing  internal representation

pmnoxx:
We recently changed internal representation of `Edge` to use `key` attribute instead of `peer0`, `peer1`.

This will fix the test. Maybe we should use different structure for  `Edge`, that's independent from the one used in `jsonrpc`.


pmnoxx:
https://nayduck.near.org/#/run/2211

mina86 (approved):

pmnoxx:
@mina86 Thanks for noticing the issue.

pmnoxx:
https://github.com/near/nearcore/pull/5247

PR #5252, opened 2021-11-14
Optimize build_hash function by 50%

pmnoxx:
`build_hash` can be optimized by 50% from 450ns per loop to around 300 ns per loop

### DIFF
--- a/chain/network-primitives/src/types.rs
+++ b/chain/network-primitives/src/types.rs
@@ -18,7 +18,7 @@ use tracing::{error, warn};
 
 use near_crypto::{KeyType, PublicKey, SecretKey, Signature};
 use near_primitives::block::{Approval, Block, BlockHeader, GenesisId};
-use near_primitives::hash::{hash, CryptoHash};
+use near_primitives::hash::CryptoHash;
 use near_primitives::merkle::combine_hash;
 use near_primitives::network::{AnnounceAccount, PeerId};
 use near_primitives::sharding::{
@@ -418,11 +418,7 @@ impl RoutedMessage {
         source: &PeerId,
         body: &RoutedMessageBody,
     ) -> CryptoHash {
-        hash(
-            &RoutedMessageNoSignature { target, author: source, body }
-                .try_to_vec()
-                .expect("Failed to serialize"),
-        )
+        CryptoHash::hash_borsh(&RoutedMessageNoSignature { target, author: source, body })
     }
 
     pub fn hash(&self) -> CryptoHash {
--- a/chain/network/src/routing/routing.rs
+++ b/chain/network/src/routing/routing.rs
@@ -114,7 +114,7 @@ impl Edge {
 
     /// Build the hash of the edge given its content.
     /// It is important that peer0 < peer1 at this point.
-    fn build_hash(peer0: &PeerId, peer1: &PeerId, nonce: u64) -> CryptoHash {
+    pub fn build_hash(peer0: &PeerId, peer1: &PeerId, nonce: u64) -> CryptoHash {
         CryptoHash::hash_borsh(&(peer0, peer1, nonce))
     }
 
--- a/chain/network/src/routing/routing.rs
+++ b/chain/network/src/routing/routing.rs
@@ -1,5 +1,6 @@
 use near_primitives::time::Clock;
 use std::collections::{hash_map::Entry, HashMap, VecDeque};
+use std::hash::Hash;
 use std::sync::{Arc, Mutex};
 use std::time::{Duration, Instant};
 
@@ -9,7 +10,7 @@ use conqueue::{QueueReceiver, QueueSender};
 use serde::{Deserialize, Serialize};
 use tracing::warn;
 
-use near_primitives::hash::{hash, CryptoHash};
+use near_primitives::hash::CryptoHash;
 use near_primitives::network::{AnnounceAccount, PeerId};
 use near_primitives::types::AccountId;
 use near_store::{ColAccountAnnouncements, Store};
@@ -23,7 +24,6 @@ use crate::{
 use actix::dev::{MessageResponse, ResponseChannel};
 use actix::{Actor, Message};
 use borsh::{BorshDeserialize, BorshSerialize};
-use byteorder::{LittleEndian, WriteBytesExt};
 use near_crypto::{KeyType, SecretKey, Signature};
 
 const ANNOUNCE_ACCOUNT_CACHE_SIZE: usize = 10_000;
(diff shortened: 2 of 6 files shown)


pmnoxx:
@matklad Do we really need this allocation to compute hash 
```
  let mut buffer = Vec::<u8>::with_capacity(65 * 2 + 8);
```
inside ``` pub fn build_hash(peer0: &PeerId, peer1: &PeerId, nonce: u64) -> CryptoHash {```?

chain/network-primitives/src/types.rs
```diff
@@ -418,11 +419,13 @@ impl RoutedMessage {
         source: &PeerId,
         body: &RoutedMessageBody,
     ) -> CryptoHash {
-        hash(
-            &RoutedMessageNoSignature { target, author: source, body }
-                .try_to_vec()
-                .expect("Failed to serialize"),
```
matklad, line 424: I feel like there might be a ton of places where we might want something like this, so let's extract this into function. 

```rust
pub fn hash(data: &[u8]) -> CryptoHash { }
pub fn hash_borsh<T: BorshSerialize>(value: &T) -> CryptoHash { }
```

In general, I feel that 99% of usages of `.try_to_vec` are suboptimal -- you almost never want to get a freshly allocated vec, you almost always want to serialized to some writer.

chain/network/src/routing/routing.rs
```diff
@@ -153,14 +158,12 @@ impl EdgeInner {
...
-        let peer1: Vec<u8> = peer1.into();
-        buffer.extend_from_slice(peer1.as_slice());
-        buffer.write_u64::<LittleEndian>(nonce).unwrap();
-        hash(buffer.as_slice())
+    pub fn build_hash(peer0: &PeerId, peer1: &PeerId, nonce: u64) -> CryptoHash {
+        let mut hasher = sha2::Sha256::default();
+        BorshSerialize::serialize(peer0, &mut hasher).unwrap();
+        BorshSerialize::serialize(peer1, &mut hasher).unwrap();
+        BorshSerialize::serialize(&nonce, &mut hasher).unwrap();
+        CryptoHash(hasher.finalize().into())
```
matklad, line 166:
```suggestion
        hash_borsh((peer0, peer1, nonce))
```

chain/network/src/routing/routing.rs
```diff
@@ -153,14 +158,12 @@ impl EdgeInner {
...
     /// It is important that peer0 < peer1 at this point.
-    fn build_hash(peer0: &PeerId, peer1: &PeerId, nonce: u64) -> CryptoHash {
-        let mut buffer = Vec::<u8>::new();
-        let peer0: Vec<u8> = peer0.into();
-        buffer.extend_from_slice(peer0.as_slice());
-        let peer1: Vec<u8> = peer1.into();
-        buffer.extend_from_slice(peer1.as_slice());
-        buffer.write_u64::<LittleEndian>(nonce).unwrap();
-        hash(buffer.as_slice())
+    pub fn build_hash(peer0: &PeerId, peer1: &PeerId, nonce: u64) -> CryptoHash {
```
matklad, line 161: "It is important that peer0 < peer1 at this point"

Not feeling good about making a function with caller-checked precondition `pub`. Can we keep it private? Or at least add an assertion. I guess, at least `debug_assert(peer0 < peer1)` would be helpful regardless.

chain/network-primitives/src/types.rs
matklad: Aside: I'd probably even argue for 

```
impl CryptoHash {
  pub fn hash_read<R: Read>(rdr: R) -> io::Result<CryptoHash> { }
  pub fn hash_bytes(bytes: &[u8]) -> CryptoHash {}
  pub fn hash_borsh<T: BorshSerialize>(value: &T) -> CryptoHash { }
}
```

API if designing the thing from scratch, but let's not change the cowpaths we already have.

core/primitives-core/src/hash.rs
```diff
@@ -134,9 +150,9 @@ impl Hash for CryptoHash {
 /// let data = [1, 2, 3];
 /// let hash = near_primitives_core::hash::hash(&data);
 /// ```
+#[deprecated(since = "#5252", note = "Use CryptoHash::hash_bytes or ::hash_borsh")]
```
matklad: I think I was confusing here. I do think that the APIs in this PR with inherent functions on `CrytpoHash` are abstractly better and more idiomatic. I don't think they are significantly better enough to justify the change. Even if we do change them, I feel that's better done in a follow up PR.

Just marking the method as deprecated I think *adds* to technical debt, instead of removing it, as it moves codebase from "consistent, but not ideal" to "inconsistent, mostly not ideal with small islands which are better".

chain/network-primitives/src/types.rs
pmnoxx: >   
> In general, I feel that 99% of usages of .try_to_vec are suboptimal -- you almost never want to get a freshly allocated vec, you almost always want to serialized to some writer.
> ```


Yes, that's why i'm doing this change. I'm removing it from `near-network`. We can ship a separate PRs to remove it from other places

>   pub fn hash_bytes(bytes: &[u8]) -> CryptoHash {}
>   pub fn hash_borsh<T: BorshSerialize>(value: &T) -> CryptoHash { }

I added those two above. I can can ```
  pub fn hash_read<R: Read>(rdr: R) -> io::Result<CryptoHash> { }
```
as well, but I don't see any usages of it at the moment.

File: chain/network/src/routing/routing.rs (hunk 3 above)
Comment (pmnoxx) [reply chain/network/src/routing/routing.rs:160]: Ok, I added assert.

File: chain/network/src/routing/routing.rs (hunk 2 above)
Comment (pmnoxx) [reply chain/network/src/routing/routing.rs]: Thanks, that's a good idea.

File: core/primitives-core/src/hash.rs (hunk 4 above)
Comment (pmnoxx) [reply core/primitives-core/src/hash.rs]: My plan is to ship PR's to make the change. At the very least do in it places where where using `try_to_vec`, or some other memory allocation is not optimal.

I didn't do this in this PR. To keep things simple, and I kind of expected that we may change the api anyway.

I'm fine with removing the `#[deprecated(since = "#5252", note = "Use CryptoHash::hash_bytes or ::hash_borsh")]`

Comment (pmnoxx) [reply core/primitives-core/src/hash.rs]: Ok, I removed deprecated.

My goal right now is to clean up `near-network` code. We probably should clean up code in other packages as well, but that can be done later.

**Comment by pmnoxx** [comment]
@matklad Once this gets merged. I'll create follow up PR's to fix places, where is probably matters.

File: chain/network-primitives/Cargo.toml [hunk 5]
```diff
@@ -10,6 +10,7 @@ tokio = { version = "1.1", features = ["full"] }
 chrono = { version = "0.4.4", features = ["serde"] }
 borsh = "0.9"
 serde = { version = "1", features = [ "derive" ] }
+sha2 = "0.9.8"
```
matklad, line 13: I think now that there's `borsh_hash` helper, changes to Cargo.toml are not needed?

matklad (approved):

chain/network-primitives/Cargo.toml
pmnoxx: Thanks, fixed.

bowenwang1996 (approved):

PR #5253, opened 2021-11-14
Make <&PeerId> to Vec<u8> conversions explicit

pmnoxx:
Remove `<&PeerId>` to `Vec<u8>` conversion.  In Rust we generally avoid having implicit memory allocations.

It's better to call `try_to_vec` directly.


matklad (approved):
Yeah, very much this! 

I'd say that even `try_to_vec` is less than ideal, as it forces reallocation. Ideally, we want something of this shape:

```rust
fn append_borsh<T: BorshSerialize>(value: &T, buf: &mut Vec<u8>)
```

Such that we can re-use the same `Vec` across loop iterations

chain/network/src/routing/routing_table_actor.rs
```diff
@@ -316,7 +320,7 @@ impl RoutingTableActor {
         for peer_id in peers_to_remove.iter() {
             let _ = update.set_ser(
                 ColPeerComponent,
-                Vec::from(peer_id).as_ref(),
```
matklad, line 319: Eg here, looking at all those allocated and then immediately freed vectors is painful :)

bowenwang1996 (approved):

PR #5257, opened 2021-11-15
Add documentation to .gitignore about '.idea' folder

pmnoxx:
We didn't describe properly what `.idea` folder is inside `.gitignore`. 
Let's fix that.


matklad:
LGTM!

In theory, there is some subset of `.idea` which we might actually want to share (see, eg, [Intellij Rust](https://github.com/intellij-rust/intellij-rust/tree/master/.idea)) but sadly that's not super trivial to define.

matklad (approved):

matklad:
Actually, I see we have "editor configs" section, let's move `.idea` there.

matklad:
Noticed another strange bit: https://github.com/near/nearcore/pull/5260. Feel free to fold into this PR if that's convenient.

PR #5258, opened 2021-11-15
Make `download_file` not store the intermediary bytes in RAM

### LINKED ISSUE #4627: `neard init --download-genesis` consumes too much of RAM
**Describe the bug**

This implementation is inefficient:

https://github.com/near/nearcore/blob/61584bba043d1fd62048a23c14e1aef1b6b39036/nearcore/src/config.rs#L1072-L1087

This downloads the whole blob into RAM and only then stores it on disk. We should download it straight to the disk (to avoid partially downloaded results, I suggest we download it to a file with `.swp` extension and once that is complete, rename it to the original file name)

This presents an issue only on testnet since other networks do not have such huge genesis JSON file **yet**.

**To Reproduce**

```
$ neard --home /tmp/neard-tmp init --download-genesis --chain-id testnet
```

Observe in the resource monitoring that the process takes 5+GB of RAM.

**Expected behavior**

RAM consumption does not exceed 1GB of RAM.

- nearcore commit/branch: master
- testnet

itegulov:
Fixes #4627

Unfortunately, it does not seem like [Actix HTTP client library](https://docs.rs/awc/2.0.3/awc/) provides a way to handle HTTP responses in a chunked fashion, so I had to use [hyper](https://hyper.rs/) instead.

To avoid partially downloaded results, we create a temporary file with `.swp` extension that is atomically renamed to the target destination once the download is over. I made sure that the temporary file gets deleted even if something goes wrong during the download, that is why there is an internal async block inside the function. I tried to replace this logic with [tempfile](https://docs.rs/tempfile) library, but it only gives you cleanup guarantees if you allocate temporary files in `/tmp`, which we want to avoid since we need the temporary and target file to be on the same filesystem in order to have access to atomic rename.

I have tested this manually with different scenarios:
1. Downloaded mainnet genesis
2. Downloaded testnet genesis
3. Tried to download mainnet/testnet genesis to a directory without write permission
4. Tried to download mainnet/testnet genesis when such a file already exists and is not overwritable (in which case the temporary file gets deleted)

Furthermore, I tried to come up with a unit tests for `download_file`, but did not want to depend on any external resource. My experience with platforms different from Rust told me to try mocking the HTTP client/response, so I tried to use [mockall](https://docs.rs/mockall/0.10.2/mockall/). It quickly started getting out of hand as I started to realize I am writing my own light wrapper around hyper. This might still be viable to do, but due to my lack of experience with Rust ecosystem I wanted to get others input on this before I am committed to this idea. Alternatively, I could just add an integration test that downloads the 5GB tesnet genesis (and maybe even play with jemalloc via [tikv-jemallocator](https://docs.rs/tikv-jemallocator/0.4.1/tikv_jemallocator/) to measure RAM usage). Let me know what you think.


nearcore/src/config.rs
```diff
@@ -1098,32 +1099,75 @@ pub fn get_config_url(chain_id: &String) -> String {
...
+        let mut tmp_file = File::create(&tmp_path)?;
+
+        let process_tmp_file = async {
+            let mut downloaded_bytes = 0usize;
+            while let Some(next_chunk_result) = resp.data().await {
+                let next_chunk = next_chunk_result?;
+                tmp_file.write_all(next_chunk.as_ref())?;
+
+                downloaded_bytes += next_chunk.len();
+                if downloaded_bytes > limit {
```
chefsale, line 1142: I believe we can drop the limit, that was expected by actix and causing problems when genesis went over the limit.

chefsale (changes requested):

nearcore/src/config.rs
```diff
@@ -1098,32 +1099,75 @@ pub fn get_config_url(chain_id: &String) -> String {
...
-            .await
-            .expect("File is bigger than specified limit. Please make the limit higher.");
-
-        std::fs::write(&path, &body).expect("Failed to create / write a file.");
+#[derive(thiserror::Error, Debug)]
+pub enum FileDownloadError {
+    #[error("Failed to download the file: {0}")]
+    HttpError(#[from] hyper::Error),
+    #[error("Failed to create/write to file: {0}")]
+    IoError(#[from] std::io::Error),
```
nagisa, line 1107: Nit: this error variant conflates three different failure modes (open, write, move/rename). These operations fail with a same `io::Error` but have fairly distinct underlying reason. I would recommend splitting this up into an `OpenError`, `WriteError` and a `RenameError` without a `#[from]`.

nearcore/src/config.rs
itegulov: Makes sense, fixed

nearcore/src/config.rs
itegulov: That seems like a better approach indeed. Do you happen to know if there is a less verbose way to wrap an error than:
```rust
std::fs::rename(&tmp_path, path).map_err(|e| {
    FileDownloadError::RenameError(e)
})?;
```

nearcore/src/config.rs
```diff
@@ -1098,32 +1099,71 @@ pub fn get_config_url(chain_id: &String) -> String {
...
+
+        let uri = url.parse()?;
+        let mut resp = client.get(uri).await?;
+
+        // To avoid partially downloaded files, we first download the file to a temporary *.swp file
+        // and rename it once the download is finished.
+        let tmp_path = path.with_extension("swp");
+        let mut tmp_file = File::create(&tmp_path).map_err(|e| {
+            FileDownloadError::OpenError(e)
+        })?;
```
ChaoticTempest, line 1131: We don't need to create closures for these. We can directly reference the enum variant itself:
```suggestion
        let mut tmp_file = File::create(&tmp_path).map_err(FileDownloadError::OpenError)?;
```

nearcore/src/config.rs
```diff
@@ -1098,32 +1099,71 @@ pub fn get_config_url(chain_id: &String) -> String {
...
+
+        let process_tmp_file = async {
+            while let Some(next_chunk_result) = resp.data().await {
+                let next_chunk = next_chunk_result?;
+                tmp_file.write_all(next_chunk.as_ref()).map_err(|e| {
+                    FileDownloadError::WriteError(e)
+                })?;
```
ChaoticTempest, line 1138:
```suggestion
                tmp_file.write_all(next_chunk.as_ref()).map_err(FileDownloadError::WriteError)?;
```

nearcore/src/config.rs
```diff
@@ -1098,32 +1099,71 @@ pub fn get_config_url(chain_id: &String) -> String {
...
+            }
+            std::fs::rename(&tmp_path, path).map_err(|e| {
+                FileDownloadError::RenameError(e)
+            })?;
```
ChaoticTempest, line 1142:
```suggestion
            std::fs::rename(&tmp_path, path).map_err(FileDownloadError::RenameError)?;
```

nearcore/src/config.rs
```diff
@@ -1098,32 +1099,71 @@ pub fn get_config_url(chain_id: &String) -> String {
...
+
+        // We run everything that interacts with the temporary *.swp file in a separate async block
+        // so that we can clean up and safely delete it if something went wrong.
+        process_tmp_file.await.map_err(|e| match std::fs::remove_file(tmp_path) {
+            Ok(_) => e,
+            Err(remove_file_error) => {
+                FileDownloadError::RemoveTemporaryFileError(remove_file_error, Box::new(e))
+            }
+        })?;
+        Ok(())
```
ChaoticTempest, line 1154: nit: already returning `Ok(())` in the `process_tmp_file` block, so we can just forward it at the end instead
```suggestion
        process_tmp_file.await.map_err(|e| match std::fs::remove_file(tmp_path) {
            Ok(_) => e,
            Err(remove_file_error) => {
                FileDownloadError::RemoveTemporaryFileError(remove_file_error, Box::new(e))
            }
        })
```

nearcore/src/config.rs
itegulov: Ah, as per @ChaoticTempest's comment you can directly reference the enum variant like this:
```rust
std::fs::rename(&tmp_path, path).map_err(FileDownloadError::RenameError)?;
```

Thanks for the pointer

chefsale (approved):

nearcore/src/config.rs
```diff
@@ -1098,32 +1099,64 @@ pub fn get_config_url(chain_id: &String) -> String {
...
+
+        // To avoid partially downloaded files, we first download the file to a temporary *.swp file
+        // and rename it once the download is finished.
+        let tmp_path = path.with_extension("swp");
+        let mut tmp_file = File::create(&tmp_path).map_err(FileDownloadError::OpenError)?;
+
+        let process_tmp_file = async {
+            while let Some(next_chunk_result) = resp.data().await {
+                let next_chunk = next_chunk_result?;
+                tmp_file.write_all(next_chunk.as_ref()).map_err(FileDownloadError::WriteError)?;
```
frol, line 1134: While we are at it, we might also make it async-ready since `std::fs` is actually blocking.

P.S. I am surprised that there is no good crate dealing with downloading files well; I have found the following crates:

* https://crates.io/crates/siwi-download - async `write_all` (:heavy_check_mark:), support for resuming downloads (:heavy_check_mark:), and progress-bars (:heavy_check_mark:), but code-wise I feel it is quite a hack project rather than a well-supported library
* https://crates.io/crates/download-async - it uses blocking `write_all` operation, which is a red flag to me
* https://crates.io/crates/downloader - it uses blocking `write_all` operation, which is a red flag to me

nearcore/src/config.rs
```diff
@@ -1098,32 +1099,64 @@ pub fn get_config_url(chain_id: &String) -> String {
...
+            }
+            std::fs::rename(&tmp_path, path).map_err(FileDownloadError::RenameError)?;
+            Ok(())
+        };
+
+        // We run everything that interacts with the temporary *.swp file in a separate async block
+        // so that we can clean up and safely delete it if something went wrong.
```
frol, line 1141: > separate async block so that we can clean up and safely delete it if something went wrong.

This statement does not make sense to me. It is not async block, but proper error handling that helps us to properly handle the error and clean up broken downloads.

frol (approved):
Overall, it is already an improvement over what we had before, so I approve this PR.

Regarding testing it, I feel your pain, and that is why I mentioned that I would rather expect to use some existing crate that is well-tested, but I did not find one that I would be happy to pull in. I think since we don't want to invest a whole lot of resources into this problem, I would even consider writing some minimal integration test by running a web-server and at least confirming that the transferred data is correctly stored on disk. Let's leave the memory usage test for the future.

nearcore/src/config.rs
ChaoticTempest: @frol why not https://docs.rs/async-std/1.10.0/async_std/fs/struct.File.html? It's a drop in replacement for `std::fs::File`

itegulov: @frol 
> While we are at it, we might also make it async-ready since std::fs is actually blocking.

Makes sense, done! Used async-std as per @ChaoticTempest's suggestion.

@ChaoticTempest 
> why not https://docs.rs/async-std/1.10.0/async_std/fs/struct.File.html? It's a drop in replacement for std::fs::File

I think what @frol actually wanted is a decent e2e library that would **download a file**, not a low-level async library/http client. But since such a thing does not exist, I used `async-std` as per your suggestion.

nearcore/src/config.rs
itegulov: "async block" here refers to `process_tmp_file` which is an async block ([ref](https://doc.rust-lang.org/reference/expressions/block-expr.html#async-blocks)). I moved the comment upwards towards its declaration, hopefully that is clearer.

itegulov:
> I think since we don't want to invest a whole lot of resources into this problem, I would even consider writing some minimal integration test by running a web-server and at least confirming that the transferred data is correctly stored on disk.

I wrote an integration test that uses nginx Docker container (via [testcontainers-rs](https://github.com/testcontainers/testcontainers-rs)). Hopefully that is what you were looking for. Let me know if it is and I will merge the PR if so!

nearcore/src/config.rs
frol: Async-std is another runtime, so we should use tokio::fs::File: https://docs.rs/tokio/1.14.0/tokio/fs/struct.File.html

integration-tests/tests/test_download_file.rs
```diff
@@ -0,0 +1,101 @@
...
+    let mut tmp_static_file = File::create(tmp_static_file_path).unwrap();
+    let zeros: [u8; 1024 * 1024] = [0; 1024 * 1024];
+    tmp_static_file.write_all(&zeros).unwrap();
+
+    let mut nginx = Nginx::default();
+    nginx.version = "1.21".to_string();
+    nginx.static_content_path = Some(tmp_static_content_dir.path());
+
+    let docker = clients::Cli::default();
+    let container = docker.run(nginx);
```
frol, line 83: Well, this adds a new host system dependency on having Docker and being able to run it (e.g. it is problematic if you run tests inside a Docker container)

nearcore/src/config.rs
itegulov: Done

integration-tests/tests/test_download_file.rs
itegulov: Hmm, I didn't think using testcontainer-rs would be introducing a new host system dependency since Docker seems to be actively used in Python tests (like [here](https://github.com/near/nearcore/blob/master/pytest/tests/stress/hundred_nodes/start_100_nodes.py)). That being said, setting up a hyper server that just serves a byte array is actually simpler code-wise, so I went with that instead. PTAL.

frol (approved):
Nice!

nearcore/src/config.rs
```diff
@@ -1097,32 +1099,68 @@ pub fn get_config_url(chain_id: &String) -> String {
...
+    #[error("Failed to rename file: {0}")]
+    RenameError(std::io::Error),
+    #[error("Invalid URI: {0}")]
+    UriError(#[from] hyper::http::uri::InvalidUri),
+    #[error("Failed to remove the temporary file after failure: {0}, {1}")]
+    RemoveTemporaryFileError(std::io::Error, Box<FileDownloadError>),
+}
+
+pub fn download_file(url: &String, path: &Path) -> Result<(), FileDownloadError> {
+    return actix::System::new().block_on(async move {
```
austinabell, line 1119: Not necessarily something that needs to change for the specific usage, but seems like this function should be unopinionated about the runtime and just be async.

I don't think anything needs to change here though, probably fine, just thinking out loud.

integration-tests/tests/test_download_file.rs
```diff
@@ -0,0 +1,42 @@
...
+            if let Err(e) = server.await {
+                eprintln!("server error: {}", e);
+            }
+        });
+    });
+
+    let tmp_downloaded_file = tempfile::NamedTempFile::new().unwrap();
+    let tmp_downloaded_file_path = tmp_downloaded_file.path();
+    nearcore::config::download_file(
+        &format!("http://localhost:{}", port).into(),
```
austinabell, line 32:
```suggestion
        &format!("http://localhost:{}", port),
```

nearcore/Cargo.toml
```diff
@@ -15,13 +15,16 @@ byteorder = "1.2"
 easy-ext = "0.2"
 chrono = { version = "0.4.4", features = ["serde"] }
 futures = "0.3"
+hyper = { version = "0.14", features = ["full"] }
```
austinabell, line 18: The downside is requiring compiling hyper and tokio on top of actix-web and actix and I'm not certain mixing both runtimes, even if just for tests, will always work together? I guess there isn't a better solution here if you can't use actix for the requests as you mention in the description, though.

nearcore/src/config.rs
```diff
@@ -6,9 +6,11 @@ use std::sync::Arc;
 use std::time::Duration;
 
 use actix;
+use hyper::body::HttpBody as _;
```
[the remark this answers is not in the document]
itegulov: Good point, I think actix had a trait named like that and this is just something I forgot to fix after I was done prototyping

nearcore/src/config.rs
```diff
@@ -1097,32 +1099,68 @@ pub fn get_config_url(chain_id: &String) -> String {
...
+    WriteError(std::io::Error),
+    #[error("Failed to rename file: {0}")]
+    RenameError(std::io::Error),
+    #[error("Invalid URI: {0}")]
+    UriError(#[from] hyper::http::uri::InvalidUri),
+    #[error("Failed to remove the temporary file after failure: {0}, {1}")]
+    RemoveTemporaryFileError(std::io::Error, Box<FileDownloadError>),
+}
+
+pub fn download_file(url: &String, path: &Path) -> Result<(), FileDownloadError> {
```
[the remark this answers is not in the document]
itegulov: Haha, I was following [this](https://github.com/rust-analyzer/rust-analyzer/blob/master/docs/dev/style.md#useless-types) style guide and was confused because I was sure that `&String` was the preferred method but apparently I misinterpeted the table :slightly_smiling_face:

nearcore/src/config.rs
itegulov: I actually encountered this exact problem while trying to write a tokio test (i.e. `#[tokio:test]`) that instantiates a server. I had to do it in a roundabout way to avoid tokio and hyper competing for `block_on`. I am going to do a follow-up to this PR anyway that is going to add a progression bar, so I guess I will see if I can fix this then.

nearcore/Cargo.toml
itegulov: As I said in the other thread, I think there are definitely some friction points while trying to use both, but I don't really see what I can do about it either...

bowenwang1996 (approved):

Issue #5145, opened 2021-11-04
New cache implementation


mzhangmzz:
Right now, we use `cached` crate for caches for database access. It has a few issues as @matklad mentioned https://github.com/near/nearcore/pull/5125#issuecomment-960527199

* the API is overly-complicated for what it does: there's `Cached` trait, `cached_key` declarative macro, and procedural macros as well. Cache is a kind of hash map, it doesn't really need these advanced features of the Rust type-system. 
* the API is inconvenient: it requires `&mut` access, but for caches you almost always need `&`. Solving `&mut` vs `&` via interior mutability is actually the primary thing I'd expect from a cache crate,  as that's the most interesting problem, but, in this case, that's just punted to the user. 
* the crate is at `0.26` version, so it is still unsure about it's own API

The second point also creates a performance bottleneck as we need to create a lock to guard every access to the cache, including read access. This is unnecessary and costly because our load is read intensive and read could be non-blocking. For example, `EpochManager` holds cache for multiple database columns and there is one lock that guards `EpochManager`. Since `EpochManager` interface is called extensively throughout our codebase, this could potentially be a performance bottleneck.

pmnoxx:
@mzhangmzz is there any reason why we can't use something simple list this https://github.com/near/nearcore/pull/5259

bowenwang1996:
@mzhangmzz with #5259 merged, should we close this issue?

PR #5259, opened 2021-11-15
feat: Implement `SyncLruCache`

pmnoxx:
This is just a proof of concept.

Implement a cache that works from a "&" context.

- get_from_cache_or_insert - gets a value or inserts a computed value
- get_from_cache - tries to get value 

TODO:
- [x] rename methods
- [x] add new methods
- [x] benchmarks
- [x] comments
- [x] add more tests.
- [x] one example of replacing real code
- [x] implement example with `lru` (done in another PR)
- [x] Use `once_cell` instead of `lazy_static!` (done in another PR). 

See https://github.com/near/nearcore/issues/5145

Was blocked by: https://github.com/near/nearcore/pull/5512


matklad:
:+1: 

I think one important thing to support is sizing the cache -- IIRC, all our cache are bounded to have at most `X` entries under LRU eviction policy.

pmnoxx:
> +1
> 
> I think one important thing to support is sizing the cache -- IIRC, all our cache are bounded to have at most `X` entries under LRU eviction policy.

@matklad I switch to use SizedCache instead of HashSet. This way we will keep last `X` entries.

pmnoxx:
@matklad The actual implementation of cache depends on what we want to achieve, what kind of features we want, etc.

- First we should decide what interface we want. Is this interface enough?

- Is there any reason we want to rewrite whole `SizedCache`? If the interface is the only reason, then we can make adjustments to this wrapper. 

- After we decide the interface, we can make a more efficient implementation if needed. Run benchmark, decide whenever use Arc or not, etc.

pmnoxx:
Rebasing due to a flunky test `https://buildkite.com/nearprotocol/nearcore/builds/6774`

utils/near-cache/src/lib.rs
```diff
@@ -0,0 +1,57 @@
...
+}
+
+// MyCache<CryptoHash, Result<wasmer_runtime::Module, CompilationError>>
+impl<K, V> MyCache<K, V>
+where
+    K: Hash + Eq + PartialEq + Clone,
+    V: Clone,
+{
+    #[allow(unused)]
+    pub fn new(capacity: usize) -> Self {
```
matklad, line 17: rustc complains because `MyCache` is private, so the compiler knows that there can't be any usges. A more natural way to fix this would be to make `MyCache` a `pub` struct.

utils/near-cache/src/lib.rs
```diff
@@ -0,0 +1,57 @@
...
+        if let Some(result) = self.get_from_cache(&key) {
+            return result.clone();
+        }
+        let val = f(&key);
+        self.inner.lock().unwrap().cache_set(key, val.clone());
+        val
+    }
+
+    #[allow(unused)]
+    pub fn get_from_cache(&self, key: &K) -> Option<V> {
```
matklad, line 36: I'd call this just `get`. At the call site, the field will probably already have `cache` in the name, and `self.foo_cache.get_from_cache` would be an unnecessary repetition.

utils/near-cache/src/lib.rs
```diff
@@ -0,0 +1,57 @@
...
+    K: Hash + Eq + PartialEq + Clone,
+    V: Clone,
+{
+    #[allow(unused)]
+    pub fn new(capacity: usize) -> Self {
+        Self { inner: Mutex::new(SizedCache::<K, V>::with_size(capacity)) }
+    }
+
+    #[allow(unused)]
+    pub fn get_from_cache_or_insert<F>(&self, key: K, f: F) -> V
```
matklad, line 22: ditto here, just `get_or_insert` seems unambiguous in context.

utils/near-cache/src/lib.rs
```diff
@@ -0,0 +1,57 @@
...
+    where
+        F: FnOnce(&K) -> V,
+        V: Clone,
+    {
+        if let Some(result) = self.get_from_cache(&key) {
+            return result.clone();
+        }
+        let val = f(&key);
+        self.inner.lock().unwrap().cache_set(key, val.clone());
```
matklad, line 31: Micro-optimizatin: let's clone the value before locking the lock, to avoid allocations under lock: 

```
let val = f(&key);
let val_clone = val.clone();
self.inner.lock().unwrap().cache_set(key, val_clone);
val
```

matklad (commented):
LGTM!

I guess @EgorKulikov or @mzhangmzz should make the overall value judgement call here.

One thing we can do here impl-wise is to replace a mutex with concurrent hash map. Not sure we should do this though: 

* I don't think there's a goto concurrent hash map for Rust which you can just use, there are several competing impls without a clear winner
* Concurrent hash map with LRU sounds complicated, if possible at all
* We hold the lock only transitively, we do not invoke user-controlled call under the lock. So, I don't think we'll see a lof of unwanted blocking from just using a lock. 

The second thing is that this sounds like the case where we can get a meaningful per win by using `parking_lot` rather than std Mutex.

pmnoxx:
@matklad Can you take a look again?

pmnoxx:
@mzhangmzz Can you take a look?

pmnoxx:
> parking_lot



> LGTM!
> 
> I guess @EgorKulikov or @mzhangmzz should make the overall value judgement call here.
> 
> One thing we can do here impl-wise is to replace a mutex with concurrent hash map. Not sure we should do this though:
> 
> * I don't think there's a goto concurrent hash map for Rust which you can just use, there are several competing impls without a clear winner
> * Concurrent hash map with LRU sounds complicated, if possible at all
> * We hold the lock only transitively, we do not invoke user-controlled call under the lock. So, I don't think we'll see a lof of unwanted blocking from just using a lock.
> 
> The second thing is that this sounds like the case where we can get a meaningful per win by using `parking_lot` rather than std Mutex.

We can try to do a benchmark and see what works best. On linux I've seen that `RwLock` is actually slow than `Mutex`.

utils/near-cache/src/lib.rs
```diff
@@ -0,0 +1,51 @@
+use cached::{Cached, SizedCache};
+use std::hash::Hash;
+use std::sync::Mutex;
+
+pub struct MyCache<K, V> {
+    inner: Mutex<SizedCache<K, V>>,
+}
+
+// MyCache<CryptoHash, Result<wasmer_runtime::Module, CompilationError>>
```
matklad, line 9: Needs removal?

matklad (approved):

pmnoxx:
@mzhangmzz @EgorKulikov Ok, I think that the package is ready. Can you take a look at the API? Is that what you need. I can add one example use, to make it clearer what's going on.

mzhangmzz:
> > parking_lot
> 
> > LGTM!
> > I guess @EgorKulikov or @mzhangmzz should make the overall value judgement call here.
> > One thing we can do here impl-wise is to replace a mutex with concurrent hash map. Not sure we should do this though:
> > 
> > * I don't think there's a goto concurrent hash map for Rust which you can just use, there are several competing impls without a clear winner
> > * Concurrent hash map with LRU sounds complicated, if possible at all
> > * We hold the lock only transitively, we do not invoke user-controlled call under the lock. So, I don't think we'll see a lof of unwanted blocking from just using a lock.
> > 
> > The second thing is that this sounds like the case where we can get a meaningful per win by using `parking_lot` rather than std Mutex.
> 
> We should benchmark this code and see what works best. From my testing `RwLock` is generally slower than `Mutex`. We can try different libraries and see what works the best.

The interface looks fine to me. I do agree with @matklad that we may want to move away from `SizedCache`, but we can do some benchmark before that, since implementing a new type of cache is quite involved. My main complaint on `SizedCache` is that it's read-blocking, but in our use case, most cache access will be reads.

utils/near-cache/src/lib.rs
```diff
@@ -0,0 +1,51 @@
+use cached::{Cached, SizedCache};
+use std::hash::Hash;
+use std::sync::Mutex;
+
+pub struct MyCache<K, V> {
```
bowenwang1996, line 5: This name is not optimal. Could to change to `Cache` instead (or some more descriptive name)?

pmnoxx: @bowenwant1996
Maybe `NearLRUCache`

matklad: I'd just `LRUCache` -- there's no near specific thing here. And Rust is not C, so there's no need to add perfixes to names just to avoid clashes.

mzhangmzz (approved):

pmnoxx:
Let's first wait for https://github.com/near/nearcore/pull/5512, and then we can start using `lru`. I did benchmarks, `lru` is 50% faster than `cached`, has better api, is more popular, better maintained. 

@matklad `lru` uses name `LRUCache`. We should avoid that name, to not have name conflict. Maybe `ThreadSafeLRUCache`?

matklad:
As a preface, as a general note, it is `LruCache` rather than `LRUCache`. That's the official Rust style for acronyms, and something I myself mess up all the time :) 

https://rust-lang.github.io/api-guidelines/naming.html

---

I think this might actually be fine to call `LruCache`. There isn't a risk of some language-level confusion here, Rust is designed to not have ODR or obscure linker failure modes. The only risk here is confusion, but because nothing will depend `lru` crate directly, I don't think that's what'll happen in practice. That is, any crate will have only one `LruCache` in scope by virtue of having the correct crate in Cargo.toml.

If we do want to have a different name, I suggest `SyncLruCache`. I don't think there's a definitive convention of how to name thread-safe variation of a data-structure, but `Sync` is one of the least questionable options for the ongoing "`once_cell` in std" naming debate (https://github.com/rust-lang/rust/issues/74465).

pmnoxx:
> As a preface, as a general note, it is `LruCache` rather than `LRUCache`. That's the official Rust style for acronyms, and something I myself mess up all the time :)
> 
> https://rust-lang.github.io/api-guidelines/naming.html
> 
> I think this might actually be fine to call `LruCache`. There isn't a risk of some language-level confusion here, Rust is designed to not have ODR or obscure linker failure modes. The only risk here is confusion, but because nothing will depend `lru` crate directly, I don't think that's what'll happen in practice. That is, any crate will have only one `LruCache` in scope by virtue of having the correct crate in Cargo.toml, so auto-import wold do the right thing, etc.
> 
> If we do want to have a different name, I suggest `SyncLruCache`. I don't think there's a definitive convention of how to name thread-safe variation of a data-structure, but `Sync` is one of the least questionable options for the ongoing "`once_cell` in std" naming debate ([rust-lang/rust#74465](https://github.com/rust-lang/rust/issues/74465)).

Actually, we are going to depend on `LruCache` from `lru` directly. Please, take a look at https://github.com/near/nearcore/pull/5512.
I understand that this isn't an issue for Rust. However I want to avoid a confusion. Just by looking at the type name, I should be able to tell which of those two we are using. Indeed, if we didn't import `lru` directly, then it would make sense to use the same name. I would prefer to avoid using the wrapper with `Mutex`, for use cases, where single threaded access is guaranteed.

pmnoxx:
@mm-near @matklad I'll merge the change as it is.

Issue #5193, opened 2021-11-10
Package `state-viewer` with `neard`


nikurt:
Please consider including `state-viewer` functionality into the `neard` binary.
The two binaries share >90% of the code, and the code size and compilation time increase should be insignificant.

My motivating use case is investigating issues or incidents. I need `state-viewer` to inspect the state of the DB. Currently it requires me to compile `state-viewer` separately and the version of `neard` and `state-viewer` must be the same, otherwise differences in protocol version or DB version can prevent `state-viewer` from reading the DB. The worst possible outcome is `state-viewer` silently upgrading the DB version.

@matklad

matklad:
>compilation time increase should be insignificant.

I'd expect somewhat significant *decrease* in compilation time, as we'd have to link one fewer binary.

PR #5261, opened 2021-11-15
Migrate `state-viewer` to `clap-derive`

nikurt:
`neard` uses `clap-derive` aka `clap3-beta`. This PR makes `state-viewer` CLI definition consistent with `neard` CLI definition.
Manually tested that every command still works. Note that names of some parameter names changed: `--start-index` instead of `--start_index`.

This PR is the first step to embedding `state-viewer` into `neard`. Issue #5193


test-utils/state-viewer/src/cli.rs
```diff
@@ -0,0 +1,244 @@
...
+use near_primitives::types::{BlockHeight, ShardId};
+use near_primitives::version::{DB_VERSION, PROTOCOL_VERSION};
+use near_store::{create_store, Store};
+use nearcore::{get_default_home, get_store_path, load_config, NearConfig};
+
+use crate::commands::*;
+
+lazy_static! {
+    static ref DEFAULT_HOME: PathBuf = get_default_home();
+}
```
matklad, line 17: For newer code, I'd weakly recommend using `once_cell` over `lazy_static`, but that's not super important.

matklad (approved):

PR #5262, opened 2021-11-15
doc: start code style document

matklad:
This documents code style rules beyond rustfmt which we want to hold for
nearcore. This PR intentionally adds basically a placeholder. If we
agree that we need a style document in the first place, specific rules
will be bikeshedded separately.


matklad:
cc @frol @khorolets , I think node interfaces parts of nearcore have at least some style to them :)

docs/style.md
```diff
@@ -0,0 +1,75 @@
...
+    types::NetworkViewClientMessages,
+};
+
+// BAD
+use near_network::test_utils::convert_boot_nodes;
+use near_network::test_utils::open_port;
+use near_network::types::NetworkViewClientMessages;
+```
+
+This corresponds to `"rust-analyzer.assist.importGranularity": "module"` setting
```
nagisa, line 71: This is not the default, right? IIRC R-A uses the

```
use near_network::{
    test_utils::{convert_boot_nodes, open_port},
    types::NetworkViewClientMessages,
};
```

by default.

(.oO maybe it would make sense to implement enforcing/formatting of this in rustfmt…?)

matklad: Yeah, that's not default. Without explicit config ra actually tries to just guess based on the current file, which works most of the time, but yeah, if there are no imports, it defaults to more aggressive merging. 

It is implemented in rustfmt, but is currently unstable: https://rust-lang.github.io/rustfmt/?version=v1.4.38&search=#imports_granularity

docs/style.md
```diff
@@ -0,0 +1,75 @@
...
+`rustfmt`.
+
+### Import Granularity
+
+Group import by module, but not deeper:
+
+```rust
+// GOOD
+use near_network::test_utils::{convert_boot_nodes, open_port};
+use near_network::types::NetworkViewClientMessages;
```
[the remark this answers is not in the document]
matklad: Yeah, good point. Added that and also switched to using `std` rather than a random use cribbled from nearcore, as std is more familiar.

nagisa (approved):

docs/style.md
```diff
@@ -0,0 +1,75 @@
...
+`rustfmt`.
+
+### Import Granularity
+
+Group import by module, but not deeper:
+
+```rust
+// GOOD
+use std::collections::{hash_map, BTreeSet};
+use std::sync::Arc;
```
austinabell, line 57: Do we want to make a note about separating `std`, `crate`/`super`, and external imports? Or would this be considered a separate PR?

I've just noticed you have a preference on this and the ordering and would be nice to have in some doc to use as reference or reason to change something

matklad: Yeah, that's the motivation change, and that's going to be a separate PR. It's not immediately clear which of the two we should pick:

* std / external / nearcore / crate:: imports
* just a single block of imports

I think we use both to some degree. 

This rule about grouping I think is an uncontroversial one, as this is pretty consistent across the codebase (although I think I saw a couple of dusty corners with nested imports I think). 

context: https://github.com/near/nearcore/pull/5241#discussion_r749372928

bowenwang1996 (approved):
Thanks!

docs/style.md
```diff
@@ -0,0 +1,75 @@
...
+
+// BAD
+use std::{
+    collections::{hash_map, BTreeSet},
+    sync::Arc,
+};
+
+// BAD
+use std::collections::BTreeSet;
```
pmnoxx, line 66: It may be good to say why it's bad.
```
// BAD - not grouped together
```

docs/style.md
```diff
@@ -0,0 +1,75 @@
...
+### Import Granularity
+
+Group import by module, but not deeper:
+
+```rust
+// GOOD
+use std::collections::{hash_map, BTreeSet};
+use std::sync::Arc;
+
+// BAD
```
pmnoxx, line 59:
```
// BAD - imports from multiple crates at once.
```

docs/style.md
```diff
@@ -0,0 +1,75 @@
...
+This section documents all micro-rules which are not otherwise enforced by
+`rustfmt`.
+
+### Import Granularity
+
+Group import by module, but not deeper:
+
+```rust
+// GOOD
+use std::collections::{hash_map, BTreeSet};
```
pmnoxx, line 56: We should write down on how we sort imports. Do we group them manually, or use formatted for that.
I'm in favor of using a formatted in one group. Though I would like to heard other people thoughts.

matklad: Agree that we should, but I want to agree on that specific point in a separate PR, as it is not clear-cut.

pmnoxx:
@matklad That's not directly related to this PR, but Rust plugin for intellij is sometimes bad at managing imports.

Clicking on "add imports", when hovering over red variable. Adds it to imports, but the plugin doesn't pay attention, that the given variable is under something like `#[cfg(feature = "performance_stats")]`.

That also leads to having to do a bit of working later manually adding `#[cfg(feature = "performance_stats")]`, for some of the imports that need it.

docs/style.md
matklad: separate PR: https://github.com/near/nearcore/pull/5290

PR #5265, opened 2021-11-15
refactor: Stop using near_metrics::Result<IntCounter>

pmnoxx:
Current implementation of metrics in `chain/network/src/stats/metrics.rs` can be done better.

As discussed in https://github.com/near/nearcore/pull/5240
I replaced usage of `Lazy<near_metrics::Result<IntCounter>> ` with simply `Lazy<IntCounter>`.

This will make code much cleaner, and we will notice whenever there is an issue with name conflict between two different metrics in Prometheus while running CI, rather than finding out later that something doesn't work as it should.

### DIFF
--- a/chain/network/src/routing/codec.rs
+++ b/chain/network/src/routing/codec.rs
@@ -61,7 +61,7 @@ impl Encoder<Vec<u8>> for Codec {
             {
                 error!(target: "network", "{} throwing away message, because buffer is full item.len(): {} buf.capacity: {}", get_tid(), item.len(), buf.capacity());
 
-                near_metrics::inc_counter_by(&metrics::DROPPED_MESSAGES_COUNT, 1);
+                metrics::DROPPED_MESSAGES_COUNT.inc_by(1);
                 return Err(Error::new(ErrorKind::Other, "Buf max capacity exceeded"));
             }
             // First four bytes is the length of the buffer.
--- a/chain/network/src/routing/routing_table_actor.rs
+++ b/chain/network/src/routing/routing_table_actor.rs
@@ -171,8 +171,8 @@ impl RoutingTableActor {
         self.needs_routing_table_recalculation = true;
 
         // Update metrics after edge update
-        near_metrics::inc_counter_by(&metrics::EDGE_UPDATES, total as u64);
-        near_metrics::set_gauge(&metrics::EDGE_ACTIVE, self.raw_graph.total_active_edges() as i64);
+        metrics::EDGE_UPDATES.inc_by(total as u64);
+        metrics::EDGE_ACTIVE.set(self.raw_graph.total_active_edges() as i64);
 
         edges
     }
@@ -242,7 +242,7 @@ impl RoutingTableActor {
         #[cfg(feature = "delay_detector")]
         let _d = DelayDetector::new("routing table update".into());
         let _routing_table_recalculation =
-            near_metrics::start_timer(&metrics::ROUTING_TABLE_RECALCULATION_HISTOGRAM);
+            metrics::ROUTING_TABLE_RECALCULATION_HISTOGRAM.start_timer();
 
         trace!(target: "network", "Update routing table.");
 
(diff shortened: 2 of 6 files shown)


chain/network/src/routing/codec.rs
```diff
@@ -60,7 +60,7 @@ impl Encoder<Vec<u8>> for Codec {
             {
                 error!(target: "network", "{} throwing away message, because buffer is full item.len(): {} buf.capacity: {}", get_tid(), item.len(), buf.capacity());
 
-                near_metrics::inc_counter_by(&metrics::DROPPED_MESSAGES_COUNT, 1);
+                metrics::DROPPED_MESSAGES_COUNT.inc_by(1);
```
bowenwang1996, line 63:
```suggestion
                metrics::DROPPED_MESSAGES_COUNT.inc();
```

chain/network/src/routing/routing_table_actor.rs
```diff
@@ -245,9 +245,8 @@ impl RoutingTableActor {
         for peer in self.peer_forwarding.keys() {
             self.peer_last_time_reachable.insert(peer.clone(), now);
         }
-
-        near_metrics::inc_counter_by(&metrics::ROUTING_TABLE_RECALCULATIONS, 1);
-        near_metrics::set_gauge(&metrics::PEER_REACHABLE, self.peer_forwarding.len() as i64);
+        metrics::ROUTING_TABLE_RECALCULATIONS.inc_by(1);
```
bowenwang1996, line 248:
```suggestion
        metrics::ROUTING_TABLE_RECALCULATIONS.inc();
```

chain/network/src/routing/codec.rs
pmnoxx: Fixed.

chain/network/src/routing/routing_table_actor.rs
pmnoxx: Fixed.

core/metrics/src/near_gauge.rs
```diff
@@ -0,0 +1,70 @@
...
+
+pub struct NearIntGauge {
+    gauge: GenericGauge<AtomicI64>,
+}
+
+impl NearIntGauge {
+    pub fn new(name: &'static str, help: &'static str) -> Self {
+        let opts = Opts::new(name, help);
+        let gauge = IntGauge::with_opts(opts).unwrap();
+        prometheus::register(Box::new(gauge.clone())).unwrap();
```
matklad, line 13: :thinking: Do we actually want to panic if we can't create a gauge? If that's true, I don't think we need our own wrapper, and can write something like this instead: 

```
    pub static ref PEER_CONNECTIONS_TOTAL: IntGauge =
        try_create_int_gauge("near_peer_connections_total", "Number of connected peers").unwrap();
```

pmnoxx: @matklad According to the documentation it should fail only if there is a name conflict. I think it's better to panic rather than have those metrics not work silently.

pmnoxx: @matklad Can you take a look again?

pmnoxx: @matklad We can do this instead; https://github.com/near/nearcore/pull/5303

bowenwang1996 (approved):

chain/network/src/routing/routing_table_actor.rs
```diff
@@ -253,8 +253,8 @@ impl RoutingTableActor {
             self.peer_last_time_reachable.insert(peer.clone(), now);
         }
 
-        near_metrics::inc_counter_by(&metrics::ROUTING_TABLE_RECALCULATIONS, 1);
-        near_metrics::set_gauge(&metrics::PEER_REACHABLE, self.peer_forwarding.len() as i64);
```
matklad, line 257: I guess we now should drop these helpers? That is, just remove `near_metrics::set_gauge`, `near_metrics::inc_counter_by` and the like?

pmnoxx: @matklad They are used in a lot of places. More PR's are needed.

matklad: Yeah, that's what I mean: let's be consistent and panic on init for *every* gauge, and not just for a subset.

PR #5267, opened 2021-11-15
refactor: Move EdgeInner methods to Edge

pmnoxx:
As @matklad suggested, the code will be much cleaner if we move `EdgeInner` methods to `Edge`. (https://github.com/near/nearcore/pull/5228#discussion_r749509847) 

After this change is done. I'l work add comments to Edge and move it to separate file.


pmnoxx:
I'm splitting this change, https://github.com/near/nearcore/pull/5276
Otherwise it's taking too much time to debug test failures.

chain/network/src/routing/routing.rs
```diff
@@ -261,40 +229,76 @@ impl EdgeInner {
...
+    signature0: Signature,
+    signature1: Signature,
+    /// Info necessary to declare an edge as removed.
+    /// The bool says which party is removing the edge: false for Peer0, true for Peer1
+    /// The signature from the party removing the edge.
+    removal_info: Option<(bool, Signature)>,
+}
+
+impl EdgeInner {
+    pub fn new(
```
matklad, line 286: Doesn't need to be pub

chain/network/src/routing/routing.rs
```diff
@@ -216,42 +184,42 @@ impl EdgeInner {
...
-                    || !self.signature1.verify(add_hash.as_ref(), &self.key.1.public_key())
+                if !self.signature0().verify(add_hash.as_ref(), &self.key().0.public_key())
+                    || !self.signature1().verify(add_hash.as_ref(), &self.key().1.public_key())
                 {
                     return false;
                 }
 
-                if let Some((party, signature)) = &self.removal_info {
-                    let peer = if *party { &self.key.0 } else { &self.key.1 };
+                if let Some((party, signature)) = &self.removal_info() {
```
matklad, line 221:
```suggestion
                if let Some((party, signature)) = self.removal_info() {
```

chain/network/src/routing/routing.rs
```diff
@@ -81,6 +82,23 @@ impl Edge {
...
+        &self.0.signature0
+    }
+
+    pub fn signature1(&self) -> &Signature {
+        &self.0.signature1
+    }
+
+    pub fn removal_info(&self) -> &Option<(bool, Signature)> {
+        &self.0.removal_info
+    }
```
matklad, line 99: `&Option<T>` isn't an idiomatic type to return from the getter, `Option<&T>` is. 

```suggestion
    pub fn removal_info(&self) -> Option<&(bool, Signature)> {
        self.0.removal_info.as_ref()
    }
```

matklad (approved):

pmnoxx:
@mm-near Can you take a look?

mm-near (approved):

Issue #3863, opened 2021-01-28
Don't verify orphans twice


mikhailOK:
We verify orphans as much as possible before adding them to the orphan pool, and fully verify upon when they are ready to be processed. We should verify only once, since it's expensive.

PR #5270, opened 2021-11-15
chain: avoid validating orphan blocks multiple times

mina86:
Track whether a block has been validated as it travels through
Chain::process_block, Chain::process_block_single and
ChainUpdate::process_block methods.  This avoids validating orphan
blocks multiple times.

Fixes: https://github.com/near/nearcore/issues/3863


mina86:
To be perfectly honest, I’m kinda perplexed by the chain code and how blocks end up moving to and from the orphan state. In general I’m not super certain about this change being the best way of achieving the goal.

bowenwang1996 (commented):
@mzhangmzz could you review this PR?

chain/chain/src/chain.rs
```diff
@@ -202,6 +227,82 @@ impl OrphanBlockPool {
     }
 }
 
+/// A wrapper for a block being processed via process_block methods.  Besides
+/// block itself and provenance it specifies whether the block has been
+/// validated.  To validate use [`BlockToProcess::check_block_validity`] method.
+struct BlockToProcess {
```
mzhangmzz, line 233: There is actually already a data structure for this use case, called `MaybeValidated`. https://github.com/near/nearcore/blob/master/core/primitives/src/utils.rs#L34, and you can search the usage of `validate_with` to see how it is used right now.

I think it's probably better to use `MaybeValidated<Block>` instead of creating a new data structure. Additionally, `provenance` and `added` don't need to be included here so that code regarding them don't need to be modified.

chain/chain/src/chain.rs
```diff
@@ -464,20 +565,13 @@ impl Chain {
     }
 
     pub fn save_orphan(&mut self, block: &Block) -> Result<(), Error> {
```
mzhangmzz, line 567: For example, you could change function to `save_orphan(&mut self, block: &MaybeValidated<Block>)` and calls 
```block.validate_with(|b|{Chain::check_block_validity(self.runtime_adapter.as_ref(), &self.genesis, b)})``` 
for validation

File: chain/chain/src/chain.rs [hunk 3]
```diff
@@ -1145,11 +1241,10 @@ impl Chain {
                         let tail_height = self.store.tail()?;
                         // we only add blocks that couldn't have been gc'ed to the orphan pool.
                         if block_height >= tail_height {
-                            let block_hash = *block.hash();
-                            let orphan = Orphan { block, provenance, added: Clock::instant() };
-
-                            self.orphans.add(orphan);
-
+                            let block_hash = block.hash();
+                            self.orphans.add(
```
mzhangmzz, line 1245: I know the code was not written by you, but I think you can call `self.save_orphan` here. This way, it's clear that `self.save_orphan` is the function to call when to add orphan to the orphan pool, and because block is validated there, we guarantee that all blocks in the orphan pool have been validated.

mzhangmzz (commented):
I think the main logic of this function is right! I added comments on how to make it simpler.

chain/chain/src/chain.rs
mina86: I ended up leaving it as is for now to minimise diff.

chain/chain/src/chain.rs
```diff
@@ -693,14 +690,22 @@ impl Chain {
...
+    pub fn validate_block(&mut self, block: &MaybeValidated<Block>) -> Result<(), Error> {
+        block
+            .validate_with(|block| {
+                Chain::validate_block_impl(
+                    self.runtime_adapter.as_ref(),
+                    &self.genesis_block(),
+                    block,
+                )
+                .map(|_| true)
+            })
```
matklad, line 704: Hm, looking at the call site, it seems what we actually want here is 

```rust
    pub fn validate_with<VE, F>(&self, validator: F) -> Result<(), VE>
    where
        F: FnOnce(&T) -> Result<(), VE>,
    {
        match self.try_validate_with(|it| Ok::<_, std::convert::Infallible>(validator(it))) {
            Ok(it) => it,
            Err(never) => match never {},
        }
    }

    pub fn try_validate_with<VE, E, F>(&self, validator: F) -> Result<Result<(), VE>, E>
    where
        F: FnOnce(&T) -> Result<Result<(), VE>, E>,
    {
        if self.validated.get() {
            Ok(Ok(()))
        } else {
            let res = validator(&self.payload)?;
            self.validated.set(res.is_ok());
            Ok(res)
        }
    }
```

That is, validation is a `-> Result<(), VE>` computation that either asserts that the thing is valid, or returns validation error. Sometimes, the validation process itself is failable -- we fail for unrelated reason before we can conclude if the stuff is valid or not.


That is somewhat slightly-more-advanced-than-usual type-fu, so the current approach is also ok.

chain/client/src/client_actor.rs
```diff
@@ -1019,10 +1022,16 @@ impl ClientActor {
         // before sending it out.
         if provenance == Provenance::PRODUCED {
             self.network_adapter.do_send(PeerManagerMessageRequest::NetworkRequests(
-                NetworkRequests::Block { block: block.clone() },
+                NetworkRequests::Block { block: block.as_ref().into_inner().clone() },
             ));
+            // If we produced it, we don’t need to validate it.  Mark the block
+            // as validated.
+            let _ = block.validate_with::<(), _>(|_| Ok(true));
```
matklad, line 1029: This looks a bit funky. Perhaps add a helper like `block.mark_as_valid()`?

chain/chain/src/chain.rs
mina86: Yes, this is something I’ve contemplated. The other use case does take advantage of the bool flag so there’s no one simple solution. For now adding `map` felt reasonable enough and least intrusive solution. But if we start using MaybeValidated more we might want to revisit the design.

matklad: To clarify, I think the other use-case abuses the bool, and encodes `Result<Result<(),()>, Error>` as `Result<bool, Error>`. That is, the semantics of `Err` actually differs.

mina86: That kinda depends how you think about the operation. Is it a ‘validate and error out if not valid’ operation or a ‘is this payload valid’ question. Though I agree that the naming ‘validate_with’ suggests the first interpretation.

integration-tests/tests/client/process_blocks.rs
```diff
@@ -3601,7 +3605,7 @@ mod storage_usage_fix_tests {
                 ProtocolFeature::FixStorageUsage.protocol_version(),
             );
 
-            let (_, res) = env.clients[0].process_block(block.clone(), Provenance::NONE);
+            let (_, res) = env.clients[0].process_block(block.clone().into(), Provenance::NONE);
             assert!(res.is_ok());
             run_catchup(&mut env.clients[0], &vec![]).unwrap();
```
matklad, line 3611: Completely unrelated to the PR, but this is a very clear indication that we want some better way to specify testing scenarios, which doesn't need to change every time we do optimization.

matklad (approved):

mzhangmzz:
@mina86 sorry I may have given some wrong suggestion before on using `MaybeValidated`. I am actually confused now why we need to call `check_block_validity` so many times in the code now. It seems to me that when `check_block_validity` returns errors, the errors always means the block is invalid (I'm not 100% sure about that but I tried to follow all the error propagation paths within the function and I think it's true) and the block should be discarded immediately and the node should send challenges when challenges are enabled. Therefore, I don't understand why we need to call this function multiple times. @bowenwang1996 any comments on that?

@mina86 I want to a deeper look into this today, sorry for making you wait. I think your PR is good overall in terms of reducing the number of times to call `check_block_validity`, but I'm now questioning whether it should be called multiple times at all. However, if we want to change the block validation code, I want to be really careful as this touches the most basic element of the protocol security.

mina86:
One path where it happens twice:

    ClientActor::process_block
        calls self.chain.validate_block
        which calls Chain::check_block_validity
    → Client::process_block
      → Client::process_block
        → Chain::process_block
          → Chain::process_block_single
            → ChainUpdate::process_block
                calls Chain::check_block_validity

This is kinda why I was puzzled by how the blocks flow through the code.

mzhangmzz:
> One path where it happens twice:
> 
> ```
> ClientActor::process_block
>     calls self.chain.validate_block
>     which calls Chain::check_block_validity
> → Client::process_block
>   → Chain::process_block
>     → Chain::process_block_single
>       → ChainUpdate::process_block
>           calls Chain::check_block_validity
> ```
> 
> This is kinda why I was puzzled by how the blocks flow through the code.

I think calling it once in `ClientActor::process_block` should be enough. The rest of the `process_block` calls should assume that the block is already validated by `check_block_validity`

mzhangmzz:
Also, I really don't like how the validation errors are mixed up with other types of chain errors. It is very error prone. 

The correctness and security of our protocol depends on the precise handling of errors propagated from within `process_block`, because different errors affects the reaction of the node later. For example, depending on the error type, the node either needs to store the block to be processed later or discard the block and send challenges if the block is invalid. And there are other types of error that indicates there is something wrong with the state of the chain. Right now, all these different types of errors are put into the same enum, `near_chain::Error` and imo this this enum type is abused in many functions. 

For example, many validation functions returns `near_chain::Error`. It is very hard to tell what it means when it returns an error because it can either mean the block is invalid, or the validation process itself failed because of insufficient information (for example, the block is an orphan). These two different scenarios require different handling at the call site, but it's undistinguishable from the function signature. And there is no consistency between different functions. For example, errors returned by `check_block_validity` means the block is invalid, while errors returned by `validate_block` either means the block is known or invalid. This creates an extremely high barrier when working with the protocol, because the correctness of our code then requires the comprehensive knowledge of possible error types each function may return. Lack of documentation also doesn't help.

I think we should reorganize chain errors. The error type should be hierarchical instead of flat. `near_chain::Error` should contain only a few error types at the top level, for example, `InvalidData`, `MissingChunks`, `Orphan`, 'InternalChainError`, which guides the general direction for how these errors will be handled. Individual functions should not be allowed to return `near_chain::Error`. They should return more specific error types.

@matklad @bowenwang1996

mzhangmzz:
As a first step, we can reorganize all the validation related errors in the code

bowenwang1996:
> @matklad @bowenwang1996 I wonder what your thoughts are

@mzhangmzz Yeah I totally agree with that. I attempted to do it once, but quickly realized that it is a deep rabbit hole so didn't dig into it :( If you are happy to embark on this long and arduous journey, I am all for it!

mina86:
It seems we’re now talking about three different things:
1. Avoiding duplicate validity checks at run time.
2. Restructuring code so there are no/fewer paths where validation happens.
3. Refactoring validation handling and error reporting.

This commit covers the first point (and thus improves the performance of the node) so we should rather discuss that and get this commit merged rather than speculating on potential future improvements.

mzhangmzz:
> It seems we’re now talking about three different things:
> 
> 1. Avoiding duplicate validity checks at run time.
> 2. Restructuring code so there are no/fewer paths where validation happens.
> 3. Refactoring validation handling and error reporting.
> 
> This commit covers the first point (and thus improves the performance of the node) so we should rather discuss that and get this commit merged rather than speculating on potential future improvements.

Yes you are right. Sorry for blocking your PR. I have some comments on this PR regarding 1) itself. I'll take another look today and post the comments. For 2) and 3) I'll take an attempt at 3) and then try to do 2).

mzhangmzz:
> > @matklad @bowenwang1996 I wonder what your thoughts are
> 
> @mzhangmzz Yeah I totally agree with that. I attempted to do it once, but quickly realized that it is a deep rabbit hole so didn't dig into it :( If you are happy to embark on this long and arduous journey, I am all for it!

Great! I'll try.

chain/client/src/client_actor.rs
```diff
@@ -1019,10 +1022,16 @@ impl ClientActor {
...
         if provenance == Provenance::PRODUCED {
             self.network_adapter.do_send(PeerManagerMessageRequest::NetworkRequests(
-                NetworkRequests::Block { block: block.clone() },
+                NetworkRequests::Block { block: block.as_ref().into_inner().clone() },
             ));
+            // If we produced it, we don’t need to validate it.  Mark the block
+            // as valid.
+            block.mark_as_valid();
         } else {
-            match self.client.chain.validate_block(&block) {
```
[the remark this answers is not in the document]
mina86: I’m not sure what you mean. Note that `chain.validate_block` will update the `is_validated` flag.

mzhangmzz (approved):

PR #5276, opened 2021-11-16
REFACTOR Make all attributes in `EdgeInfo` private

pmnoxx:
`EdgeInfo` is used in a lot of places in our code. We should remove `pub` modifier from all it's attributes.

Splitting https://github.com/near/nearcore/pull/5267 due to some unknown bug.

### DIFF
--- a/chain/network/src/peer_manager/peer_manager_actor.rs
+++ b/chain/network/src/peer_manager/peer_manager_actor.rs
@@ -301,12 +301,12 @@ impl PeerManagerActor {
                 continue;
             }
             let key = edge.get_pair();
-            if !self.routing_table_view.is_local_edge_newer(&key, edge.nonce) {
+            if !self.routing_table_view.is_local_edge_newer(&key, edge.nonce()) {
                 continue;
             }
             self.routing_table_view
                 .local_edges_info
-                .insert((edge.key.0.clone(), edge.key.1.clone()), edge.clone());
+                .insert((edge.key().0.clone(), edge.key().1.clone()), edge.clone());
         }
 
         self.routing_table_addr
@@ -401,7 +401,7 @@ impl PeerManagerActor {
                     continue;
                 }
                 let key = edge.get_pair();
-                if !self.routing_table_view.is_local_edge_newer(&key, edge.nonce) {
+                if !self.routing_table_view.is_local_edge_newer(&key, edge.nonce()) {
                     continue;
                 }
                 // Check whenever peer needs to be removed when edge is removed.
@@ -1907,7 +1907,7 @@ impl PeerManagerActor {
                         self.routing_table_view.get_edge(self.my_peer_id.clone(), peer_id.clone())
                     {
                         if cur_edge.edge_type() == EdgeType::Added
-                            && cur_edge.nonce >= edge_info.nonce
+                            && cur_edge.nonce() >= edge_info.nonce
                         {
                             return NetworkResponses::EdgeUpdate(Box::new(cur_edge.clone()));
                         }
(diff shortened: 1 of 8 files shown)


chain/network/src/routing/routing.rs
```diff
@@ -85,6 +85,22 @@ impl Edge {
         Edge(Arc::new(EdgeInner::new(peer0, peer1, nonce, signature0, signature1)))
     }
 
+    pub fn nonce(&self) -> u64 {
```
mm-near, line 88: @matklad - is it the right way to do "getter" accessors ? or does Rust offer some other option ?

mm-near (approved):
looks good - but let's wait for @matklad feedback.

chain/network/src/routing/routing.rs
pmnoxx: @mm-near Yes, @mm-near said in another PR that's how you do it in Rust.

chain/network/src/peer_manager/peer_manager_actor.rs
```diff
@@ -301,12 +301,12 @@ impl PeerManagerActor {
...
             }
             let key = edge.get_pair();
-            if !self.routing_table_view.is_local_edge_newer(&key, edge.nonce) {
+            if !self.routing_table_view.is_local_edge_newer(&key, edge.nonce()) {
                 continue;
             }
             self.routing_table_view
                 .local_edges_info
-                .insert((edge.key.0.clone(), edge.key.1.clone()), edge.clone());
+                .insert((edge.key().0.clone(), edge.key().1.clone()), edge.clone());
```
matklad, line 309: Isn't this just 

```suggestion
                .insert(edge.key().clone(), edge.clone());
```

? If this is corret, there are a couple more places where we want this. 

Also, looking at this, maybe it makes sense to define (in a separte PR, if at all)

```rust
struct EdgeMap {
  raw: HashMap<(PeerId, PeerId), Edge>
}
```

? This'll make it easer to get rid of key duplication.

chain/network/src/routing/routing.rs
```diff
@@ -260,8 +260,8 @@ impl RoutingTableView {
 
     pub fn remove_edges(&mut self, edges: &Vec<Edge>) {
         for edge in edges.iter() {
-            assert!(edge.key.0 == self.my_peer_id || edge.key.1 == self.my_peer_id);
-            let key = (edge.key.0.clone(), edge.key.1.clone());
+            assert!(edge.key().0 == self.my_peer_id || edge.key().1 == self.my_peer_id);
+            let key = (edge.key().0.clone(), edge.key().1.clone());
```
matklad, line 264: same, can just use `edge.key()`

matklad (approved):

chain/network/src/routing/edge.rs
```diff
@@ -44,6 +44,22 @@ impl Edge {
...
+
+    pub fn signature0(&self) -> &Signature {
+        &self.0.signature0
+    }
+
+    pub fn signature1(&self) -> &Signature {
+        &self.0.signature1
+    }
+
+    pub fn removal_info(&self) -> &Option<(bool, Signature)> {
```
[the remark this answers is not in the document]
pmnoxx: This won't work, because `removal_info` attribute is defined as
```
   removal_info: Option<(bool, Signature)>,
```

chain/network/src/routing/routing.rs
pmnoxx: Thanks, fixed. I found a few other places like that as well.

chain/network/src/routing/edge.rs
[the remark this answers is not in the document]
matklad: and you use `.as_ref` in the gettre to get from refernce to an option to an optional reference.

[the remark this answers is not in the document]
pmnoxx: It makes sense thanks.

PR #5282, opened 2021-11-16
Dont set/get global variables in a unit test.

### LINKED ISSUE #5244: Random test failure `chunks_delay_tracker::test::test_blocks_ahead_of_head`
I saw some random test failure
```<html>
<body>
<!--StartFragment-->

[2021-11-13T20:41:39Z] running 7 tests
--
  | [2021-11-13T20:41:39Z] test sync::test::test_get_locator_heights ... ok
  | [2021-11-13T20:41:39Z] test chunks_delay_tracker::test::test_timestamps ... ok
  | [2021-11-13T20:41:39Z] test chunks_delay_tracker::test::test_blocks_ahead_of_head ... FAILED
  | [2021-11-13T20:41:39Z] test sync::test::test_sync_headers_fork ... ok
  | [2021-11-13T20:41:39Z] test sync::test::test_block_sync ... ok
  | [2021-11-13T20:41:39Z] test sync::test::test_block_sync_archival ... ok
  | [2021-11-13T20:41:51Z] test sync::test::test_slow_header_sync ... ok
  | [2021-11-13T20:41:51Z]
  | [2021-11-13T20:41:51Z] failures:
  | [2021-11-13T20:41:51Z]
  | [2021-11-13T20:41:51Z] ---- chunks_delay_tracker::test::test_blocks_ahead_of_head stdout ----
  | [2021-11-13T20:41:51Z] thread 'chunks_delay_tracker::test::test_blocks_ahead_of_head' panicked at 'assertion failed: `(left == right)`
  | [2021-11-13T20:41:51Z]   left: `Ok(2)`,
  | [2021-11-13T20:41:51Z]  right: `Ok(0)`', chain/client/src/chunks_delay_tracker.rs:126:9
  | [2021-11-13T20:41:51Z] note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
  | [2021-11-13T20:41:51Z]
  | [2021-11-13T20:41:51Z]
  | [2021-11-13T20:41:51Z] failures:
  | [2021-11-13T20:41:51Z]     chunks_delay_tracker::test::test_blocks_ahead_of_head
  | [2021-11-13T20:41:51Z]
  | [2021-11-13T20:41:51Z] test result: FAILED. 6 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 12.19s
  | [2021-11-13T20:41:51Z]
  | [2021-11-13T20:41:51Z] error: test failed, to rerun pass '-p near-client --lib'
  | [2021-11-13T20:41:51Z] 🚨 Error: The command exited with status 101

<!--EndFragment-->
</body>
</html>
```

nikurt:
Fixes #5244


chain/client/src/chunks_delay_tracker.rs
```diff
@@ -46,7 +46,7 @@ impl ChunksDelayTracker {
 
     // Computes the delay between receiving a block and receiving the latest of its chunks.
     // Updates the metric to the max delay across the most recently processed blocks.
-    fn update_chunks_metric(&mut self, head_height: BlockHeight) {
+    fn get_max_delay_us(&mut self, head_height: BlockHeight) -> u128 {
```
pmnoxx, line 49: It may be a good idea to use `Duration`.

chain/client/src/chunks_delay_tracker.rs
```diff
@@ -58,17 +58,30 @@ impl ChunksDelayTracker {
...
             latest.saturating_sub(head_height)
         } else {
             0
-        };
-        near_metrics::set_gauge(&metrics::BLOCKS_AHEAD_OF_HEAD, value as i64);
+        }
+    }
+
+    fn update_blocks_ahead_metric(&mut self, head_height: BlockHeight) {
+        near_metrics::set_gauge(
```
pmnoxx, line 81: This is not related to this PR. Probably we should refactor the code to have ability to check whenever metrics are updater properly.

chain/client/src/chunks_delay_tracker.rs
pmnoxx: I see later we actually compare this to `Duration`
Instead of using:
```
  assert_eq!(tracker.get_max_delay_us(head_height), Duration::from_secs(5).as_micros());
```
we could make it
```
```
  assert_eq!(tracker.get_max_delay_us(head_height), Duration::from_secs(5));
```

pmnoxx (approved):
LGTM

mm-near (approved):

chain/client/src/chunks_delay_tracker.rs
nikurt: I don't see how to make it cleanly.
Either:
* I need a test-only API, which is prone to testing not production code but test-only code.
* Complicate the interface by adding some `Option<>` parameter which in prod is always `None`. This exposes implementation details to the interface.
* Maybe something in-between: let us register a callback in `Tracker` object. Then we can set expectations on the callback. However that needs a few more structs defined.

pmnoxx: Yes, if we want to do this, this will require creating `Metrics` structs and passing it down. So either `Tracker` will have to take it in constructor, or in the method that use metrics.

I'm not saying it's worth doing, but if we really want to test those metrics in unit tests, that's the only way I see right now,

PR #5283, opened 2021-11-16
refactor: Split `test_utils.rs` to avoid having code with `test_features` and without in one file.

pmnoxx:
`test_utils.rs` currently has code code that's under `test_features` feature. It's better to split the code into two parts.

### DIFF
--- a/chain/network/src/test_utils.rs
+++ b/chain/network/src/test_utils.rs
@@ -1,14 +1,8 @@
 use std::collections::{HashMap, HashSet, VecDeque};
 use std::net::TcpListener;
-#[cfg(feature = "test_features")]
-use std::sync::atomic::Ordering;
 use std::sync::{Arc, Mutex, RwLock};
 use std::time::Duration;
 
-#[cfg(feature = "test_features")]
-use actix::actors::mocker::Mocker;
-#[cfg(feature = "test_features")]
-use actix::Addr;
 use actix::{Actor, ActorContext, Context, Handler, MailboxError, Message};
 use futures::future::BoxFuture;
 use futures::{future, FutureExt};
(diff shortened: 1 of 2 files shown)


chefsale (approved):

chain/network/src/test_utils_with_test_features.rs
```diff
@@ -0,0 +1,104 @@
...
+};
+use actix::actors::mocker::Mocker;
+use actix::{Actor, Addr};
+use near_primitives::block::GenesisId;
+use near_primitives::network::PeerId;
+use near_store::test_utils::create_test_store;
+use near_store::Store;
+use std::sync::atomic::{AtomicUsize, Ordering};
+use std::sync::Arc;
+
```
mm-near, line 16: and then this file would become 'test_utils"

chain/network/src/test_utils.rs
```diff
@@ -306,89 +276,3 @@ impl MockPeerManagerAdapter {
         self.requests.write().unwrap().pop_front()
```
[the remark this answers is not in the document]
pmnoxx: @mm-near That code is used by tests, just by tests that aren't under feature flag,

pmnoxx:
@mm-near Can you take a look?

pmnoxx:
@mm-near Can you take a look?

chain/network/src/test_utils.rs
```diff
@@ -307,88 +277,110 @@ impl MockPeerManagerAdapter {
...
-        net_config,
-        vec![("test1", open_port())],
-        10,
-        routing_table_addr.clone(),
-    )
-    .0
-    .start();
-    (peer_manager_addr, routing_table_addr)
-}
+pub(crate) mod test_features {
```
matklad, line 281: :+1: this looks much better. In general, I'd say that conditional imports are a smell, and that we should try to minimize those!

matklad (approved):

pmnoxx:
@mm-near Can you take a look?

mm-near (approved):

PR #5285, opened 2021-11-16
REFACTOR Move blacklist_from_iter to near-network-primitives

pmnoxx:
`blacklist_from_iter` is currently in `near-network`, but it's not used there.
We should move it out to `near-network-primitives`.


chain/network-primitives/src/lib.rs
```diff
@@ -1 +1,2 @@
 pub mod types;
+pub mod utils;
```
mm-near, line 2: please also add the README.md file to network-primitives to describe what kind of functions should end up in this crate

mm-near (approved):

pmnoxx:

pmnoxx:

bowenwang1996 (approved):

PR #5286, opened 2021-11-16
REFACTOR Remove direct conversion from PublicKey to PeerId

pmnoxx:
As discussed with @matklad it doesn't make sense to have converted function from `PublicKey` to `PeerId`. Not every key, is a PeerId. It's cleaner to call the constructor instead.


chain/network/src/peer_manager/peer_store.rs
```diff
@@ -400,7 +400,7 @@ mod test {
 
     fn gen_peer_info(port: u8) -> PeerInfo {
         PeerInfo {
-            id: PeerId::from(SecretKey::from_random(KeyType::ED25519).public_key()),
+            id: PeerId::from(PeerId::new(SecretKey::from_random(KeyType::ED25519).public_key())),
```
matklad, line 403:
```suggestion
            id: PeerId::new(SecretKey::from_random(KeyType::ED25519).public_key()),
```

matklad (approved):

mm-near (approved):

pmnoxx:
@bowenwang1996 @frol It seems you approval is required.

pmnoxx:

PR #5288, opened 2021-11-16
Fuzz actions

posvyatokum:
Adding key actions. In addition every transaction now uses random available key. Had to increase base nonce for this to work.
Also adding implicit account creation on transfer therefor adding creation of account with implicit account id. Changed all the places that assumed that account id is test{i}.


test-utils/runtime-tester/src/fuzzing.rs
```diff
@@ -344,14 +403,22 @@ pub enum Function {
...
     fn from_seeds(seeds: &[String]) -> Self {
-        let accounts = seeds.iter().map(|id| Account::from_id(id.clone())).collect();
+        let accounts: Vec<Account> = seeds.iter().map(|id| Account::from_id(id.clone())).collect();
+        let account_id_to_idx = {
+            let mut result = HashMap::new();
+            for i in 0..accounts.len() {
+                result.insert(accounts[i].id.clone(), i);
+            }
+            result
+        };
```
matklad, line 413: Seems this can be shorteneed with the help of accounts.iter().enumerate().map(...).collect()

test-utils/runtime-tester/src/fuzzing.rs
```diff
@@ -423,34 +488,188 @@ impl Scope {
...
+    }
+
+    fn new_test_account(&mut self) -> Account {
         let new_id = format!("test{}", self.accounts.len());
         self.alive_accounts.insert(self.accounts.len());
         self.accounts.push(Account::from_id(new_id));
         self.accounts[self.accounts.len() - 1].clone()
     }
 
+    fn new_64_account(&mut self, u: &mut Unstructured) -> Result<Account> {
```
bowenwang1996, line 502:
```suggestion
    fn new_implicit_account(&mut self, u: &mut Unstructured) -> Result<Account> {
```

test-utils/runtime-tester/src/fuzzing.rs
```diff
@@ -423,34 +488,188 @@ impl Scope {
...
+        } else {
+            Ok(u.choose(&possible_signers)?.clone())
+        }
+    }
+
+    pub fn delete_random_key(
+        &mut self,
+        u: &mut Unstructured,
+        account: &Account,
+    ) -> Result<PublicKey> {
```
bowenwang1996, line 614: how do we handle the situation where all keys on an account are deleted? Is that expected or no?

posvyatokum: This function is only called once and before that call we specifically check if account has any keys.

bowenwang1996 (approved):

matklad (approved):

PR #5290, opened 2021-11-16
docs: document consistent import style

matklad:
Document whether we should do

```rust
use near_store::{ColPeers, Store};
use std::net::SocketAddr;
```

or 

```rust
use std::net::SocketAddr;

use near_store::{ColPeers, Store};
```

This PR currently documents both suggested guidelines, and we'll use this PR to decide which one we pick. 

Decision procedure: if you don't care about the style, ignore this PR, and just note the final decision :) Otherwise, react with :+1: for proposal A or proposal B (do read the actual proposals in the changset). Feel free to leave a comment explaining your choice.

If there's a clear consensus around one option, we'll just pick it. Otherwise, @bowenwang1996 makes the final call or delegates the decision to make the final call.

### DIFF
--- a/docs/style.md
+++ b/docs/style.md
@@ -74,6 +74,81 @@ in rust-analyzer
 
 **Rationale:** Consistency, matches existing practice.
 
+### Import Blocks
+
+**Proposal A**
+
+Separate the group of `use` items with blank lines:
+
+1. Imports from the standard library.
+2. Imports from crates.io crates.
+3. Imports from other crates in nearcore repository.
+4. Imports from the current crate
+
+Within each group, imports are sorted automatically by `rustfmt`
+
+```rust
+// GOOD
+use std::collections::HashMap;
+use std::net::SocketAddr;
+
+use borsh::BorshSerialize;
+use rand::seq::SliceRandom;
+
+use near_primitives::utils::to_timestamp;
+use near_store::{ColPeers, Store};
+
+use crate::types::KnownPeerState;
+
+// BAD -- no blank lines between the groups, groups are in a wrong order
+use borsh::BorshSerialize;
+use crate::types::KnownPeerState;
+use near_primitives::utils::to_timestamp;
+use near_store::{ColPeers, Store};
+use rand::seq::SliceRandom;
+use std::collections::HashMap;
+use std::net::SocketAddr;
+```
+
+**Rationale:** Consistency, improved readability. Specifically, correct
+dependencies and non-dependencies between modules and crates is what makes the
+overall architecture easy or hard to work with. More readable imports allow to
+spot problems like layering violations more easily.
+
+<hr/>
+
+**Proposal B**
+
+Do not separate imports into groups with blank lines. Write a single block of
+imports and rely on `rustfmt` to sort them.
+
+```rust
+// GOOD
+use borsh::BorshSerialize;
+use crate::types::KnownPeerState;
+use near_primitives::utils::to_timestamp;
+use near_store::{ColPeers, Store};
+use rand::seq::SliceRandom;
+use std::collections::HashMap;
+use std::net::SocketAddr;
+
+// BAD -- several groups of imports
+use std::collections::HashMap;
+use std::net::SocketAddr;
+
+use borsh::BorshSerialize;
+use rand::seq::SliceRandom;
+
+use near_primitives::utils::to_timestamp;
+use near_store::{ColPeers, Store};
+
+use crate::types::KnownPeerState;
+```
+
+**Rationale:** Consistency, ease of automatic enforcement. Today stable rustfmt
+can't split imports into groups automatically, and doing that manually
+consistently is a chore.
+
 ## Documentation
 
 When writing documentation in `.md` files, wrap lines at approximately 80
--- a/docs/style.md
+++ b/docs/style.md
@@ -76,49 +76,6 @@ in rust-analyzer
 
 ### Import Blocks
 
-**Proposal A**
-
-Separate the group of `use` items with blank lines:
-
-1. Imports from the standard library.
-2. Imports from crates.io crates.
-3. Imports from other crates in nearcore repository.
-4. Imports from the current crate
-
-Within each group, imports are sorted automatically by `rustfmt`
-
-```rust
-// GOOD
-use std::collections::HashMap;
-use std::net::SocketAddr;
-
-use borsh::BorshSerialize;
-use rand::seq::SliceRandom;
-
-use near_primitives::utils::to_timestamp;
-use near_store::{ColPeers, Store};
-
-use crate::types::KnownPeerState;
-
-// BAD -- no blank lines between the groups, groups are in a wrong order
-use borsh::BorshSerialize;
-use crate::types::KnownPeerState;
-use near_primitives::utils::to_timestamp;
-use near_store::{ColPeers, Store};
-use rand::seq::SliceRandom;
-use std::collections::HashMap;
-use std::net::SocketAddr;
-```
-
-**Rationale:** Consistency, improved readability. Specifically, correct
-dependencies and non-dependencies between modules and crates is what makes the
-overall architecture easy or hard to work with. More readable imports allow to
-spot problems like layering violations more easily.
-
-<hr/>
-
-**Proposal B**
-
 Do not separate imports into groups with blank lines. Write a single block of
 imports and rely on `rustfmt` to sort them.
 


matklad:
**Proposal A** separate std/crates.io/nearcore import groups with blank lines.

matklad:
**Proposal B** just write a single block of imports and let rustfmt sort it in arbitrary consistent order (alphabetically).

matklad:
I vote for plan B. While I am personally a huge fan of proposal A, and that's what we use in `rust-analyzer`, I think plan B makes more sense as it is easier to implement. In particular, today significant parts of nearcore are written in slightly-regressed variant A. That is, we have evidence that we can't actually maintain A:

https://github.com/matklad/nearcore/blob/49554f6b08f0d06aa7c343c67ebe05ad8b5a6c21/runtime/runtime/src/verifier.rs#L1-L27

When rustfmt stabilizes option to auto-group the imports, then I think we can switch back to nice imports, but for now it seems that simplicity wins.

nikurt:
Voting for plan B.
Implementation of plan A has some non-intuitive corners cases. It's obvious that `std::` and `near_chain::` should belong to different groups. What about `borsh` or `state-viewer`? Should `crate::` imports be separate from `near_chain::`? Other than implementing policy to figure out all corner cases, let's go for a much simpler solution.
Also, I'm unaware of any Cons of Proposal A.

pmnoxx:
> I vote for plan B. While I am personally a huge fan of proposal A, and that's what we use in `rust-analyzer`, I think plan B makes more sense as it is easier to implement in our context. In particular, today significant parts of nearcore are written in slightly-regressed variant A. That is, we have evidence that we can't actually maintain A:
> 
> https://github.com/matklad/nearcore/blob/49554f6b08f0d06aa7c343c67ebe05ad8b5a6c21/runtime/runtime/src/verifier.rs#L1-L27
> 
> When rustfmt stabilizes option to auto-group the imports, then I think we can switch back to nice imports, but for now it seems that simplicity wins.

In the code you mentioned. I see someone tried initially to use `B`. https://github.com/matklad/nearcore/blob/49554f6b08f0d06aa7c343c67ebe05ad8b5a6c21/runtime/runtime/src/verifier.rs#L1-L25 then afterwards someone must have used rust plugin to auto add imports at the end https://github.com/matklad/nearcore/blob/49554f6b08f0d06aa7c343c67ebe05ad8b5a6c21/runtime/runtime/src/verifier.rs#L26-L27

That's why I think plan A is better for us. If we try to stick to `B`, then anytime you add a new import, it will get added automatically to the last group. And then you have to remember to move rearrange manually all the imports yourself.

Personally I like how `B` looks, but in every change I do, I have to manually rearrange all imports. I would rather use `A`, that would make changes much faster. If someday rustfmt gets updated and groups then together, then sure we can use `B`. 

My vote is for `A` for now, as it requires less maintanance.

In the example above, if we were using `A` from the beginning, rust plugin for intellij would add imports properly, and we would still be in `A`, not something in between.

bowenwang1996 (approved):
Looks like most people support plan B

pmnoxx:
> **Proposal B** just write a single block of imports and let rustfmt sort it in arbitrary consistent order (alphabetically).

My vote is for B. Btw. In the first message you gave example for B first then for A.

pmnoxx (approved):

frol:
I am huge fan of proposal A, and I want to point out that there are a few modules in nearcore that actually follow this style:

* jsonrpc ([example](https://github.com/near/nearcore/blob/3a730f38f124a1855fb4cb5ea7662db89bbfe5dd/chain/jsonrpc/src/lib.rs) - I am not proud of the amount of those imports, but that is a different story)
* rosetta-rpc ([example](https://github.com/near/nearcore/blob/3a730f38f124a1855fb4cb5ea7662db89bbfe5dd/chain/rosetta-rpc/src/lib.rs))

[indexer-for-explorer](https://github.com/near/near-indexer-for-explorer/blob/f459ca7586aba91cdc31980bdbc2914bcd9391ec/src/aggregated/circulating_supply/lockup.rs) also follows proposal A, and I even [defined it as a code style](https://github.com/near/near-indexer-for-explorer/blob/master/CONTRIBUTING.md#imports-ordering-and-grouping)

I feel going with proposal B is kind of pointless (it is as good as no rules since we basically rely on rustfmt, which is implied, and according to the core code style document, the current file code style should take prevalence).

I also hate discussions about code style, so I would consider just set expectations that we want to reach proposal A, but before there is support in rustfmt, we ignore import style issues unless it breaks the core code-style rule (do not break the existing code style of a file that you edit).

matklad:
>it is as good as no rules since we basically rely on rustfmt

I think there's significant meaningful difference here. https://github.com/matklad/nearcore/blob/49554f6b08f0d06aa7c343c67ebe05ad8b5a6c21/runtime/runtime/src/verifier.rs#L1-L27 again is a good example. It's clearly self-inconsistent, and, provided that we care about style, it *should* be refactored to either A or B. Similarly, when a new contributor crates new file, they can make a choice between A or B (or something else).

matklad:
My personal opinion shifted a bit more towards A, for two reasons:

* pmnoxx in #5311 shows a neat trick of using rustfmt nightly *once* to achieve consistency, so we can actually arrive at A (*) relatively simply 
* indexer has a style guide, which picks option A: https://github.com/near/near-indexer-for-explorer/blob/master/CONTRIBUTING.md#imports-ordering-and-grouping. It'd be great if we can say that both nearcore and indexer follow the same style guide.

That being said, I still think we should put B into our styleguide. The deal breaking argument for me is tooling enforceability. While nightly `rustfmt` *can* do `std/extern-crate/crate::`, it can't distinguish between crates from crates.io and crates from the workspace. As nikurt points out, that's a fundamental problem, as the answer isn't always clear cut. rust-analyzer supports the same grouping options as rustfmt. IntelliJ doesn't support configurable imports at all (pinged folks at https://github.com/intellij-rust/intellij-rust/issues/6094). So rustfmt supported `A'` can be a reasonable choice with partial tooling support, but than it is different from `A`. On the other hand, `B` is fully support by all tools.

matklad:
Let's go with B then!

pmnoxx:
a

> My personal opinion shifted a bit more towards A, for two reasons:
> 
> * pmnoxx in [refactor: Reorganize imports in near-network #5311](https://github.com/near/nearcore/pull/5311) shows a neat trick of using rustfmt nightly _once_ to achieve consistency, so we can actually arrive at A (*) relatively simply
> * indexer has a style guide, which picks option A: https://github.com/near/near-indexer-for-explorer/blob/master/CONTRIBUTING.md#imports-ordering-and-grouping. It'd be great if we can say that both nearcore and indexer follow the same style guide.
> 
> That being said, I still think we should put B into our styleguide. The deal breaking argument for me is tooling enforceability. While nightly `rustfmt` _can_ do `std/extern-crate/crate::`, it can't distinguish between crates from crates.io and crates from the workspace. As nikurt points out, that's a fundamental problem, as the answer isn't always clear cut. rust-analyzer supports the same grouping options as rustfmt. IntelliJ doesn't support configurable imports at all (pinged folks at [intellij-rust/intellij-rust#6094](https://github.com/intellij-rust/intellij-rust/issues/6094)). So rustfmt supported `A'` can be a reasonable choice with partial tooling support, but than it is different from `A`. On the other hand, `B` is fully support by all tools.
> 
> Curious if other folks want to change there +1 based on new input!

I think we should revisit this subject in the future. Once `imports_granularity` feature of `rustfmt` becomes stable we will have more choices when it comes to deciding enforceable code style.

matklad:
Yeah, in general the decision here is easily reversible. We don't even have to wait until rustfmt is stable -- if we want to change *all* code in some way, it's would be relatively easy for me to write one-of code in rust-analyzer to apply the transform. What is hard is maintaining consistency over time, that's the bit where it's better to stick to simplest tools.

pmnoxx:
> Yeah, in general the decision here is easily reversible. We don't even have to wait until rustfmt is stable -- if we want to change _all_ code in some way, it's would be relatively easy for me to write one-of code in rust-analyzer to apply the transform. What is hard is maintaining consistency over time, that's the bit where it's better to stick to simplest tools.

Yes, changing the code to use still is not hard.

Maintaining consistency is not hard, we can write a simple script that runs in CI, and rejects all code, which is not formatted properly. It even shows you how formatting is done. I wrote couple a couple of tools, to do simple code refactor, which are not done by `rustfmt`.

However, the hard part is how to do it without adding extra burned on developers. Sure, we could run an extra script before every commit, but that would soon become an annoyance. The cleanest way would be to sponsor such feature to `rustfmt` or wait for features to come to `rustfmt`. I like `rustfmt` as it add zero overhead on developers.

PR #5292, opened 2021-11-16
refactor: remove default impls inside VMConfig

Longarithm:
Test plan:

Existing tests.


matklad (approved):

runtime/runtime-params-estimator/src/testbed.rs
```diff
@@ -55,7 +55,7 @@ impl RuntimeTestbed {
             max_total_prepaid_gas: u64::MAX,
             max_number_bytes_method_names: u64::MAX,
 
-            ..Default::default()
+            ..VMLimitConfig::test()
```
mm-near, line 58: is feels strange that we're using a 'test' method here  - please comment why

core/primitives-core/src/config.rs
```diff
@@ -113,15 +117,15 @@ impl VMConfig {
...
+                ..VMLimitConfig::test()
             },
         }
     }
 }
 
 // TODO #4649: remove default impl and create impls with explicitly stated purposes
-impl Default for VMLimitConfig {
-    fn default() -> Self {
+impl VMLimitConfig {
```
mm-near, line 127: should we update the comment above?

core/primitives-core/src/config.rs
```diff
@@ -19,10 +19,13 @@ pub struct VMConfig {
     pub limit_config: VMLimitConfig,
 }
 
+fn default_max_transaction_size() -> u64 {
+    4 * 2u64.pow(20) // 4 MiB
```
mm-near, line 23: optional: or maybe just write: 4194304 ? (I'm not sure what's better in rust)

core/primitives-core/src/config.rs
```diff
@@ -72,6 +75,7 @@ pub struct VMLimitConfig {
     /// Max contract size
     pub max_contract_size: u64,
     /// Max transaction size
+    #[serde(default = "default_max_transaction_size")]
```
mm-near, line 78: is it strange that this is the only field that has a default set - please explain in the comments why

mm-near (approved):

core/primitives-core/src/config.rs
Longarithm: This field was introduced in protocol version 46, but was not explicitly included into runtime configs.
This was not problematic because it didn't affect any of past txs, but actually I think it's more meaningful to remove default value and include this field into all past configs - what I did now.

runtime/runtime-params-estimator/src/testbed.rs
Longarithm: Done. Overall intuition is that it is a test structure, it's hard to say something more.

core/primitives-core/src/config.rs
```diff
@@ -321,34 +318,75 @@ pub struct ExtCostsConfig {
...
+
+#[cfg(feature = "protocol_feature_alt_bn128")]
+fn default_alt_bn128_g1_sum_base() -> Gas {
+    SAFETY_MULTIPLIER * 1058438125
+}
+
+#[cfg(feature = "protocol_feature_alt_bn128")]
+fn default_alt_bn128_g1_sum_byte() -> Gas {
+    SAFETY_MULTIPLIER * 25406181
+}
```
matklad, line 386: That feels suboptimal. Can we just include these in json configs? They should be ignored by stable code. 

Also, @bowenwang1996 what's up with alt_bn128 thing at all? Do we plan to eventually stabilze it? Do we want to remove it?

bowenwang1996: Yes we are right now going through an audit of its implementation

runtime/runtime-params-estimator/src/testbed.rs
mm-near: sorry - I misread the context.

matklad (approved):

PR #5296, opened 2021-11-16
chore: unbreak builds on master

matklad:
cargo t -p near-vm-runner currently fails because we don't properly
specify the feature we use.

### DIFF
--- a/core/primitives/Cargo.toml
+++ b/core/primitives/Cargo.toml
@@ -20,7 +20,7 @@ chrono = { version = "0.4.4", features = ["serde"] }
 derive_more = "0.99.3"
 easy-ext = "0.2"
 sha2 = "0.9"
-serde = { version = "1", features = ["derive"] }
+serde = { version = "1", features = ["derive", "rc"] }
 serde_json = "1"
 smart-default = "0.6"
 validator = "0.12"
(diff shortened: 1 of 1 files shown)


mina86 (approved):
For what it’s worth, LGTM. This indeed fixes build failure.

bowenwang1996 (approved):

pmnoxx:
@matklad Thanks. Why did the build break? I thought we do checks before merging.

matklad:
One thing we check on master, but not on PR, is building all the release artifacts (building in `--relese` takes ages). One think we include in release is `near-vm-runner-standalone`, and `cargo build -p near-vm-runner-standalone` got broke. The non-release tests were building only `neard`, and that worked becaues the feature was enabled by some unrelated dependency. 

The fact that  we build standalone runner as a part of release process is accidental though, fixed that as well in https://github.com/near/nearcore/pull/5299.

matklad:
As to what actually got broken: on the previous, broken `fdb377f92` commit

```
$ cargo check -p near-vm-runner-standalone
```

fails, but 

```
$ cargo check -p near-vm-runner-standalone -p near-network
```

succeeds. That's because cargo featrue we enable in network unbreaks the build of standalone runner. In other words, conditional compilation in Rust is as big a source of accidental complexity as ever :)

PR #5297, opened 2021-11-16
refactor: remove unused `max_gas_burnt_view` field

Longarithm:
Test plan:

Existing tests.


core/primitives-core/src/config.rs
```diff
@@ -26,9 +26,6 @@ pub struct VMConfig {
 pub struct VMLimitConfig {
     /// Max amount of gas that can be used, excluding gas attached to promises.
     pub max_gas_burnt: Gas,
-    /// Max burnt gas per view method.
-    /// TODO #4719: remove this field because it does not affect the protocol
-    pub max_gas_burnt_view: Gas,
```
matklad, line 31: Question: this struct is `Deserialize`, does anything break in serialization due to field removal?

bowenwang1996 (approved):

core/primitives-core/src/config.rs
matklad: Should we remove 

https://github.com/near/nearcore/blob/14f08a72b4e61f44f44d10ebf86318120847ccbb/nearcore/res/mainnet_genesis.json#L184

line then?

Longarithm: Yes, it's better.
We can also remove it from JSON runtime configs, but let's do it later because we have other open PRs changing runtime configs.

mm-near (approved):

core/primitives-core/src/config.rs
matklad: >but let's do it later because we have other open PRs changing runtime configs.

FWIW, my heuristic is to not think too much about potential conflicts when submitting PRs. It's actually hard to predict which work would end up being merged first, and rebasing usually isn't a problem, as you have a very concrete issue to resolve.

PR #5299, opened 2021-11-16
chore: don't include near-vm-runner-standalone in release artifacts

matklad:
We used to use near-vm-runner-standalone as a way to run contracts
locally, but that never worked properly, as the standalone runner can't
perfectly recreate proper blockchain environment. Today, the task of
running a contract "locally" is solved by a sandbox node

https://github.com/near/sandbox

We still keep the code in the repo, as it is sometimes useful for ad-hoc
investigations by contract runtime team. We just forgot to remove it
from releases, which we should have done ages ago.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,7 +16,6 @@ docker-nearcore-nightly:
 
 release:
 	cargo build -p neard --release
-	cargo build -p near-vm-runner-standalone --release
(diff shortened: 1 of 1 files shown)


matklad:
cc @nagisa for some bits of historical context.

chefsale (approved):

PR #5303, opened 2021-11-16
FEATURE Simplify usage of metrics - panic if we fail to create prometheus metric

pmnoxx:
As discussed with @matklad I think we should refactor code for using Prometheus metrics. 

Creating a propethius metric can fail if there is a naming error. We should panic when application starts, and find out that there is an issue, rather than trying to silently hide errors.

If we think this change is good, we should consider doing that in other places as well.


matklad:
Sorry, thought this was an issue, and not a PR. The issue is not completely closed due to https://github.com/near/nearcore/pull/5265/files#r751135003

PR #5307, opened 2021-11-17
re: stop exporting  `PeerInfo` from `near-network-primitives` in `near-network`

pmnoxx:
It's generally a bad practice to export types defined in a different crate. This leads to code inconsistencies, where it's possible to import given type in two ways.

In a separate PR, we will address the place where each variable should belong.


bowenwang1996 (commented):
> Though PeerInfo belongs to near-network-primitives

What is the reason for that? In general what are the criteria for putting a type in near-network vs. near-network-primitives?

pmnoxx:
> > Though PeerInfo belongs to near-network-primitives
> 
> What is the reason for that? In general what are the criteria for putting a type in near-network vs. near-network-primitives?

`PeerInfo` is used in definition of `NetworkInfoResponse`, which is used in message transmitted data between `PeerManager` and `ClientActor`. 

If `PeerInfo` was only used in tests, even for modules outside `near-network`. 
I would put it in `near-network`, but export it publicly through `test_utils.rs` from `near-network`. 
So anyone one using it would know that, this type is meant to be used for tests only.

That's mainly because of limiting of how Rust tests works. Tests from other crate, can't use private values inside the crate, even for integration testing.

pmnoxx:
In this `PR` we fix the ambiguity of being able to import types from two places due to usage of `pub use ...`.  

It's better to have only way way to do imports.

pmnoxx:
@bowenwang1996 I'll create a doc defining what's where and why. What do you think?

mm-near:
| @bowenwang1996 I'll create a doc defining what's where and why. What do you think?

That would be great !

mzhangmzz:
> > > Though PeerInfo belongs to near-network-primitives
> > 
> > 
> > What is the reason for that? In general what are the criteria for putting a type in near-network vs. near-network-primitives?
> 
> `PeerInfo` is used in definition of `NetworkInfoResponse`, which is used in message transmitted data between `PeerManager` and `ClientActor`. And that's outside of near-network.
> 
> If `PeerInfo` was only used in tests, even for modules outside `near-network`. I would put it in `near-network`, but export it publicly through `test_utils.rs` from `near-network`. So anyone one using it would know that, this type is meant to be used for tests only.
> 
> That's mainly because of limiting of how Rust tests works. Tests from other crate, can't use private values inside the crate, even for integration testing.

Seems that you saw the issue that I created. Does it make sense to move these tests to within the crate? https://github.com/near/nearcore/issues/5371

pmnoxx:
> > > > Though PeerInfo belongs to near-network-primitives
> > > 
> > > 
> > > What is the reason for that? In general what are the criteria for putting a type in near-network vs. near-network-primitives?
> > 
> > 
> > `PeerInfo` is used in definition of `NetworkInfoResponse`, which is used in message transmitted data between `PeerManager` and `ClientActor`. And that's outside of near-network.
> > If `PeerInfo` was only used in tests, even for modules outside `near-network`. I would put it in `near-network`, but export it publicly through `test_utils.rs` from `near-network`. So anyone one using it would know that, this type is meant to be used for tests only.
> > That's mainly because of limiting of how Rust tests works. Tests from other crate, can't use private values inside the crate, even for integration testing.
> 
> Seems that you saw the issue that I created. Does it make sense to move these tests to within the crate? #5371

Yes, it's a good change. See my reply in that post. I provided complete reasoning and some stats.

pmnoxx:
@mm-near Is there anything stopping us from merging?

pmnoxx:
@mm-near @matklad Can you take a look?

pmnoxx:
In this PR I want to stop exporting external code outside of the package. We should ship this by itself. In separate PR, we will address where each type should belong.

mzhangmzz (approved):

mzhangmzz:
This PR only addresses PeerInfo, am I understanding that right? Should we add that to the PR title?

pmnoxx:
> This PR only addresses PeerInfo, am I understanding that right? Should we add that to the PR title?

That's a good suggestion. I changed the 'PR" to be more.

mm-near (approved):

pmnoxx:
@ailisp Hey, can you take a look? Your review is needed.

ailisp (approved):
Go through the conversation, and looks reasonable to me!

ailisp:
@nikurt looks need an approval from you

nikurt (approved):

pmnoxx:
@nikurt Thanks!

PR #5308, opened 2021-11-17
refactor: Reduce number of ways you can import types from near-network

pmnoxx:
There are currently two ways to import types from `near-network` one from `types.rs` another from `lib.rs` through
```
pub use types::{
    FullPeerInfo, NetworkClientMessages, NetworkClientResponses, NetworkRecipient, NetworkRequests,
    NetworkResponses, PeerManagerAdapter,
};
```

It's better to have one way of doing it. Later we should decide where all types should be, but it should be easier to do cleanup.


chefsale (approved):

pmnoxx:
@chefsale Thanks!

pmnoxx:
@mm-near Can you take a look?

mm-near (approved):

pmnoxx:
@frol Can you take a look, your approval is needed.

frol (approved):

PR #5312, opened 2021-11-17
feat: Add benchmark for `store` read/write

pmnoxx:
Add benchmark for reading / writing from store.

See https://github.com/near/nearcore/issues/4771

```
cd chain/network/benches
rm -rf /tmp/xxxxx ; cargo bench benchmark_db
```

TODO:
- [x] Add tempdir
- [x] Pregenerate keys
- [x] Add stats to see number of entries read successfully
- [x] Why we can only read 48.5% entries back when we write them to DB?
- [x] Move code to a better place
- [x] Document what the benchmark does


chain/network/benches/ibf.rs
```diff
@@ -16,6 +21,48 @@ fn test_measure_adding_edges_to_ibf(bench: &mut Bencher) {
     });
 }
 
-benchmark_group!(benches, test_measure_adding_edges_to_ibf);
+#[allow(dead_code)]
+fn benchmark_db(bench: &mut Bencher) {
+    let path = Path::new("/tmp/xxxxx");
```
matklad, line 26: This probably should use `tmp_file`?

chain/network/benches/ibf.rs
```diff
@@ -16,6 +21,48 @@ fn test_measure_adding_edges_to_ibf(bench: &mut Bencher) {
...
+    store_update.commit().unwrap();
+
+    // .. let keys = Arc::new(keys);
+
+    bench.iter(move || {
+        let start = Instant::now();
+        // ..let keys = keys.clone();
+        for _x in 0..num_keys {
+            // let key = keys[x].clone();
+            let key: Vec<u8> = (0..40).map(|_| rand::random::<u8>()).collect();
```
matklad, line 52: I guess ideally we'd generate random numbers outside of the loop we are timing?

chain/network/benches/ibf.rs
pmnoxx: It's a dir. So `tempdir`.

pmnoxx:
@matklad For some reason I'm only able to get back 48.5% of entries I put in the database. That seems wrong.

pmnoxx:
@matklad `use tempfile::TempDir` is too old. It hasn't been updated for 3 years and it uses old random. We will have to find another package.

matklad:
https://crates.io/crates/tempfile is what we should use here. https://crates.io/crates/tempdir is indeed the old deprecated thing.

pmnoxx:
@matklad I'm not sure this belongs to ` chain/network/benches/ibf.rs`. What would be the best place to put it?

pmnoxx:
@mm-near I refactored the benchmark, can you take another look?

pmnoxx:
@djsatok Figured out the reason. It turns out that, we set flags for this column for reference couting. That's the reason why not all columns get written to database. Thanks for looking into this.

pmnoxx:
@matklad Ok, the benchmark is ready.

Longarithm (approved):
That's great!
Can we also make separate benchmark for `ColState` [writing values like this one](https://github.com/near/nearcore/blob/810294fc45c3d31d0f8c14da2ba41e4baaa525b2/core/store/benches/store_bench.rs#L75), maybe in other PR?
I think ideally we need to check performance for each column with separately set options.

pmnoxx:
@mm-near Can you take a look?

pmnoxx:
> That's great! Can we also make separate benchmark for `ColState` [writing values like this one](https://github.com/near/nearcore/blob/810294fc45c3d31d0f8c14da2ba41e4baaa525b2/core/store/benches/store_bench.rs#L75), maybe in other PR? I think ideally we need to check performance for each column with separately set options.

Good idea, we can do that in another PR. It should be easy to modify this code as an example.

core/store/benches/store_bench.rs
```diff
@@ -0,0 +1,83 @@
...
+
+        if let Ok(Some(x)) = val {
+            black_box(x);
+            read += 1;
+        }
+    }
+    read
+}
+
+// write a value of random size between 0..333 for each key to db.
```
bowenwang1996, line 70: could you explain why 333 is chosen?

core/store/benches/store_bench.rs
```diff
@@ -0,0 +1,83 @@
...
+
+// read from DB keys in random order
+fn read_from_db(store: &Arc<Store>, keys: &Vec<Vec<u8>>) -> usize {
+    let mut read = 0;
+    for _k in 0..keys.len() {
+        let r = rand::random::<u32>() % (keys.len() as u32);
+        let key = &keys[r as usize];
+
+        let val = store
+            .get(ColBlockMerkleTree, key.as_ref())
```
bowenwang1996, line 59: Could you explain why this specific column is chosen?

core/store/benches/store_bench.rs
```diff
@@ -0,0 +1,83 @@
...
+
+use bencher::{black_box, Bencher};
+use near_primitives::borsh::maybestd::sync::Arc;
+use near_primitives::errors::StorageError;
+use near_store::db::DBCol::ColBlockMerkleTree;
+use near_store::{create_store, Store};
+use std::time::{Duration, Instant};
+
+// try to write to db `10m` keys and then read all of them in random order
+fn benchmark_write_then_read_successful(bench: &mut Bencher) {
```
bowenwang1996, line 12: it might be helpful to make the number of keys an argument

core/store/benches/store_bench.rs
```diff
@@ -0,0 +1,83 @@
...
+    for _k in 0..count {
+        let key: Vec<u8> = (0..40).map(|_| rand::random::<u8>()).collect();
+
+        res.push(key)
+    }
+    res
+}
+
+// read from DB keys in random order
+fn read_from_db(store: &Arc<Store>, keys: &Vec<Vec<u8>>) -> usize {
```
mm-near, line 52: nit: let's pass as an argument, number of entries to read (currently you're reading keys.len - but it could be a separate arg)

mm-near (approved):
What were the results of your benchmark? are they stable between runs?

core/store/benches/store_bench.rs
pmnoxx: Thanks, I'll updated the explanation in code.

I updated comments. When I added logs, I saw a lot of disk writes for value sizes between 10 ... 330. I added a benchmark to reflect that situation.

The example benchmark can be modified as needed.

core/store/benches/store_bench.rs
pmnoxx: Thanks, I updated explanation in the code.

This column is one where `.is_rc() == false`. Other, columns do reference counting, which makes it so `.set` by itself won't work correctly.

For column `.is_rc() == true`, an updated benchmark needs to be written.

core/store/benches/store_bench.rs
pmnoxx: That's a great idea. I added more keys as well.

core/store/benches/store_bench.rs
pmnoxx: @mm-near I added options as arguments. That should make it easier to use.

pmnoxx:
@bowenwang1996 Please, take a look.

core/store/benches/store_bench.rs
pmnoxx: That's a good question. This benchmark only works with such column for which we don't do reference counting.

It may be worth testing performance for columns that we do.

core/store/benches/store_bench.rs
pmnoxx: Value 333 was chosen from empirical data. After adding a log, I saw  lot of writes, where values written had sizes between `0` and `325`

PR #5313, opened 2021-11-17
refactor: Add Default and move constants to route_back_cache.rs

pmnoxx:
We should make core more readable by moving `RouteBackCache` related constants inside `route_back_cache.rs`.

### DIFF
--- a/chain/network/src/routing/route_back_cache.rs
+++ b/chain/network/src/routing/route_back_cache.rs
@@ -5,6 +5,13 @@ use near_primitives::hash::CryptoHash;
 use near_primitives::network::PeerId;
 use near_primitives::time::Clock;
 
+/// default value for `capacity`
+const DEFAULT_CAPACITY: usize = 100_000;
+/// default value for `evict_timeout`
+const DEFAULT_CACHE_EVICT_TIMEOUT: Duration = Duration::from_millis(120_000);
+/// default value for `remove_frequent_min_size`
+const DEFAULT_REMOVE_BATCH_SIZE: usize = 100;
+
 /// Cache to store route back messages.
 ///
 /// The interface of the cache is similar to a regular HashMap:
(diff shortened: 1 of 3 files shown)


chain/network/src/routing/route_back_cache.rs
```diff
@@ -7,6 +7,13 @@ use near_primitives::time::Clock;
 
 type Size = u64;
 
+/// default value for `capacity`
+const ROUTE_BACK_CACHE_SIZE: u64 = 100_000;
```
mm-near, line 11: maybe we should rename it, so that the name contains 'CAPACITY'?

chain/network/src/routing/route_back_cache.rs
```diff
@@ -7,6 +7,13 @@ use near_primitives::time::Clock;
...
+/// default value for `evict_timeout`
+const ROUTE_BACK_CACHE_EVICT_TIMEOUT: Duration = Duration::from_millis(120_000);
+/// default value for `remove_frequent_min_size`
+const ROUTE_BACK_CACHE_REMOVE_BATCH: u64 = 100;
```
mm-near, line 15: same here with REMOVE_FREQUENT_MIN_SIZE

mm-near (approved):

chain/network/src/routing/route_back_cache.rs
pmnoxx: I changed it to be:
```
/// default value for `capacity`
const DEFAULT_CAPACITY: u64 = 100_000;
/// default value for `evict_timeout`
const DEFAULT_CACHE_EVICT_TIMEOUT: Duration = Duration::from_millis(120_000);
/// default value for `remove_frequent_min_size`
const DEFAULT_REMOVE_BATCH_SIZE: u64 = 100;
```

PR #5319, opened 2021-11-17
chain: avoid cloning block in save_block and save_orphan methods

mina86:
Change Chain::save_block and Chain::save_orphan methods to take the
block by value rather than by reference.  The one place where those
methods are called does not need to keep the block after calling the
methods.


mzhangmzz (approved):
Great find! Thanks!

bowenwang1996 (approved):
Epic

PR #5318, opened 2021-11-17
refactor: Remove usage of `lazy_static!` from chain

pmnoxx:
This is continuation of https://github.com/near/nearcore/issues/5317
This time we will format metrics in `chain/chain/src/metrics.rs` file.

Part of PR: https://github.com/near/nearcore/pull/5321/files

Issue #5317, opened 2021-11-17
Stop using  lazy_static with near_metrics::Result<IntCounter>>


pmnoxx:
We should continue https://github.com/near/nearcore/pull/5265
and stop using  ` Lazy<near_metrics::Result<IntCounter>>` and replace with ` Lazy<IntCounter>` everywhere.

@matklad 

Progress
- [x] near-network https://github.com/near/nearcore/pull/5265
- [x] chain/chain https://github.com/near/nearcore/pull/5318
- [x] chain/json* #5323
- [ ] Split full PR https://github.com/near/nearcore/pull/5321

PR #5321, opened 2021-11-17
refactor: Remove usage of lazy_static! not neard files

pmnoxx:
Full change for removing `Lazy` and usage of `metrics::Result`.

See https://github.com/near/nearcore/issues/5317

I'm splitting this PR into smaller ones. I'll do more split as more RPs get merged.
- [x] 19 files chain/jsonrpc https://github.com/near/nearcore/pull/5323
- [x] 11 files chain #5318 
- [x] 10 files core only https://github.com/near/nearcore/pull/5366
- [x] 3 files runtime https://github.com/near/nearcore/pull/5367
- [x] 8 files other places #5321
- [x] Remove unused methods in metrics https://github.com/near/nearcore/pull/5374


core/store/src/db.rs
```diff
@@ -193,54 +194,51 @@ impl DBCol {
...
-}
+
+pub static IS_COL_RC: Lazy<Vec<bool>> = Lazy::new(|| {
+    let mut col_rc = vec![false; NUM_COLS];
+    col_rc[DBCol::ColState as usize] = true;
+    col_rc[DBCol::ColTransactions as usize] = true;
+    col_rc[DBCol::ColReceipts as usize] = true;
+    col_rc[DBCol::ColReceiptIdToShardId as usize] = true;
+    col_rc
+});
```
matklad, line 241: Unrelated to the PR, but I think all these don't need to be lazy, and can be just constexprs

matklad (approved):

core/store/src/db.rs
pmnoxx: We should make a separate PR to fix that.

pmnoxx:
@matklad I'm in favor in general of splitting large changes into smaller into groups by code owners.

Big changes are good to be split into 5 groups, though maybe a better division can be found.
- chain/jsonrc
- chain/* not jsorc
- core
- runtime
- everything else


What do you think:

matklad:
Yeah, smaller changes are always better!

frol (approved):

nikurt (approved):

bowenwang1996 (approved):

PR #5324, opened 2021-11-17
utils: use internally-mutable validated flag in MaybeValidated

mina86:
Change MaybeValidated type from an enum with two variants (indicating
validated and not-validated payload) into a struct which holds
validated flag and payload in separate fields.  Furthermore, hold the
flag inside of a `Cell` to allow internal mutability.

This allows for an `validate_inplace_with` method which validates the
payload if necessary and updates the flag so that the next time the
method is called no validation needs to happen.

The downside is that MaybeValidated is no longer Sync or Send.


core/primitives/src/utils.rs
```diff
@@ -25,50 +25,105 @@ pub mod min_heap;
...
     pub fn as_ref(&self) -> MaybeValidated<&T> {
-        match &self {
-            Self::Validated(ref t) => MaybeValidated::Validated(t),
-            Self::NotValidated(ref t) => MaybeValidated::NotValidated(t),
-        }
+        MaybeValidated { validated: self.validated.clone(), payload: &self.payload }
     }
 
+    /// Extracts the payload whether or not it’s been validated.
     pub fn extract(self) -> T {
```
matklad, line 110: Unrelated, but `into_inner` would be more standard name. extract actually sounds like something that should run validation.

matklad (approved):
The design of the data structure looks right to me! I don't think `!Send` is an issue, we can replace `Cell` with `AtomicBool` to regain that. 

It feels a bit iffy that you can erroneously call `validate` twice with different functions. I'd love to assert that that doesn't happen, but I don't know of a good way to do this (hacky way: make the fn `#[track_caller]`, and compare location infos). 

Luckily, we have only *one* place in the codebase where we call `validate_with`, so we don't have that problem yet. This however raises a question -- if there's only one place where we use this high-order function, do we actually need this abstraction? Perhaps we can do something simpler, like splitting some functions into `pub do_x` and `priv do_x_assert_vaid`? (Don't have a lot of context here, probably there's some reason we actually need this).

mina86:
This came about from my work on https://github.com/near/nearcore/pull/5270 so there will be a second user of the method. In that change, MaybeValidated object ends up carried over a few function calls so having two versions of a function isn’t really feasible.

integration-tests/tests/client/challenges.rs
```diff
@@ -495,7 +495,7 @@ fn test_receive_invalid_chunk_as_chunk_producer() {
         &vec![merkle_paths[0].clone()],
     );
     assert!(env.clients[1]
-        .process_partial_encoded_chunk(MaybeValidated::NotValidated(partial_encoded_chunk))
+        .process_partial_encoded_chunk(MaybeValidated::from(partial_encoded_chunk))
```
mzhangmzz, line 498: nit: does it make sense to change this function to be from_non_validated, just to be extra clear?

core/primitives/src/utils.rs
```diff
@@ -25,50 +25,108 @@ pub mod min_heap;
...
+        MaybeValidated { validated: self.validated.clone(), payload: &self.payload }
     }
 
-    pub fn extract(self) -> T {
-        match self {
-            Self::Validated(t) => t,
-            Self::NotValidated(t) => t,
-        }
+    /// Extracts the payload whether or not it’s been validated.
+    pub fn into_inner(self) -> T {
```
mzhangmzz, line 113: Does it make sense to extract only when it's validated? I think the existing use case for PartialEncodedChunk validation satisfies that we only want to extra the inner struct only if it's validated. Obviously you know better for the use cases in `process_block`, but I think it makes sense that we only want to process a block if the block has already been validated.

mzhangmzz (commented):
Thanks for the detailed comments! This looks great!

core/primitives/src/utils.rs
mina86: I don’t think so. The use in `ShardsManager::process_partial_encoded_chunk` could be replaced by one which requires payload to be validated, but uses in `Client::process_partial_encoded_chunk` need to just get the payload out regardless of its state.

integration-tests/tests/client/challenges.rs
mina86: I guess this is a bit philosophical. `from` comes from the `From` trait and gives us `into` as well so using it seems more idiomatic?

core/primitives/src/utils.rs
mzhangmzz: I see

mzhangmzz (approved):

bowenwang1996 (approved):

integration-tests/tests/client/challenges.rs
matklad: Yeah, this is bikesheddable. An appropriate std example is that for a long time there weren't `From<T> from Option<T>`, but then it was added at some point. So I'd say it's idiomatic, but not super clear-cut. 

In my personal code style, I tend to prefer inherent factory methods more in general, for a number of weak reasons: 

* uses less of the type-system machinery 
* forces somewhat more readable code. `foo.into()` in particular often is a readability issue, as you need to run the type checker in your head to figure out into *what*. 
* more robust to changes -- in the future, you can add extra parameter to an inherent method

This specific case is clearly borderline though.

PR #5325, opened 2021-11-17
bug: fix `network/routing.rs` tests

### LINKED ISSUE #5322: `routing::test_dropping_routing_messages` test is broken
```sh
cargo test --workspace --features test_features \
    --project integration-tests --test network \
    -- --exact routing::test_dropping_routing_messages
```

This hasn’t been caught because the test is guarded by a `cfg(feature = "test_features")` and since we don’t enable that feature on BuildKite the test has not been executed.

pmnoxx:
In `PeerActor` we have a logic which is used to drop duplicated messages if they happen within `50`ms. That constant used to be higher while the change was originally implemented.
```
/// Duplicated messages will be dropped if routed through the same peer multiple times.
pub const DROP_DUPLICATED_MESSAGES_PERIOD: Duration = Duration::from_millis(50);
```

In `network/routing.rs` tests we use structs `StateMachine` to test PR. The check coincidentally runs every `50ms`, which is sometimes too fast, sometimes too slow for the logic to drop duplicated messages to work.

The fix is to change constant inside testes from `50ms` to `1ms`.

Closes #5322

### DIFF
--- a/integration-tests/tests/network/runner.rs
+++ b/integration-tests/tests/network/runner.rs
@@ -374,8 +374,6 @@ impl StateMachine {
                             })
                             .collect();
 
-                        let time = Mutex::new(Instant::now());
-
                         actix::spawn(
                             info.read()
                                 .unwrap()
--- a/integration-tests/tests/network/runner.rs
+++ b/integration-tests/tests/network/runner.rs
@@ -1,13 +1,13 @@
 use std::collections::HashSet;
 use std::iter::Iterator;
 use std::sync::atomic::{AtomicBool, Ordering};
-use std::sync::{Arc, RwLock};
+use std::sync::{Arc, Mutex, RwLock};
 use std::time::Duration;
 
 use actix::{Actor, Addr, AsyncContext, Context, Handler, Message, System};
 use chrono::DateTime;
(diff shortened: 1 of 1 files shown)


integration-tests/tests/network/runner.rs
```diff
@@ -408,14 +412,19 @@ impl StateMachine {
...
+                                        debug!(target: "network", "{}: ping, pong check : {} {:?} {:?} expected {:?} {:?}",
+                                            num_prev_actions, source, pings, pongs, pings_expected, pongs_expected);
                                         if ping_ok && pong_ok {
                                             flag.store(true, Ordering::Relaxed);
                                         } else {
-                                            panic!(
-                                                "ping, pong check failed got: {:?} {:?}",
-                                                pings, pongs
-                                            );
+                                            if time.lock().unwrap().elapsed()
```
mina86, line 420: I’d prefer `} else if {` to reduce amount of indentation.

mina86 (approved):

integration-tests/tests/network/runner.rs
pmnoxx: @mina86 I removed that else branch completely.

PR #5326, opened 2021-11-17
Inject random seed used randomize transaction pool

yordanmadzhunkov:
We need deterministic tests environment to use fuzzing framework. 

TransactionPool uses pseudo random numbers to set transaction order. In this PR, I add the explicit setting of the random number generator seed to be injected in the Transaction pool. I need this change to make the test environment deterministic ( like with time mocking ).


chain/chunks/src/lib.rs
```diff
@@ -397,13 +398,15 @@ pub struct ShardsManager {
     chunk_forwards_cache: SizedCache<ChunkHash, HashMap<u64, PartialEncodedChunkPart>>,
 
     seals_mgr: SealsManager,
+    pub rng_seed: RngSeed,
```
matklad, line 401: All other fields are private, it feels odd to break encapsulation and add a pub one.

chain/chunks/src/lib.rs
```diff
@@ -775,7 +779,10 @@ impl ShardsManager {
 
     /// Returns true if transaction is not in the pool before call
     pub fn insert_transaction(&mut self, shard_id: ShardId, tx: SignedTransaction) -> bool {
-        self.tx_pools.entry(shard_id).or_insert_with(TransactionPool::new).insert_transaction(tx)
+        self.tx_pools
+            .entry(shard_id)
+            .or_insert_with(|| TransactionPool::new(self.rng_seed))
```
matklad, line 784: Is this actually correct? It seems that all `TransactioPool`s will share the same `.rng_seed`, which seems like a change in behavior? Naively, I'd probably expect most things to store an instance of `rng`, rather than a seed.

chain/client/src/client_actor.rs
```diff
@@ -1695,6 +1699,12 @@ pub fn start_client(
     #[cfg(feature = "test_features")] adv: Arc<RwLock<AdversarialControls>>,
 ) -> (Addr<ClientActor>, ArbiterHandle) {
     let client_arbiter_handle = Arbiter::current();
+    let mut rng_seed = [4; 32];
+    let thread_rng_seed = rand::thread_rng().next_u64().to_le_bytes().to_vec();
+    for i in 0..8 {
+        rng_seed[i] = thread_rng_seed[i];
+    }
```
bowenwang1996, line 1706: This piece of code looks very suspicious to me. What do you plan to achieve here? Why is `rng_seed` initialized as `[4;32]`?

chain/epoch_manager/src/test_utils.rs
```diff
@@ -51,6 +52,7 @@ pub fn epoch_info(
     validator_kickout: Vec<(AccountId, ValidatorKickoutReason)>,
     validator_reward: HashMap<AccountId, Balance>,
     minted_amount: Balance,
+    rnd_seed: RngSeed,
```
bowenwang1996, line 55: I am quite confused. Why do we need a random seed here?

chain/chunks/src/lib.rs
```diff
@@ -775,7 +779,11 @@ impl ShardsManager {
 
     /// Returns true if transaction is not in the pool before call
     pub fn insert_transaction(&mut self, shard_id: ShardId, tx: SignedTransaction) -> bool {
-        self.tx_pools.entry(shard_id).or_insert_with(TransactionPool::new).insert_transaction(tx)
+        let seed = self.random_seed(shard_id);
+        self.tx_pools
+            .entry(shard_id)
+            .or_insert_with(|| TransactionPool::new(seed))
+            .insert_transaction(tx)
```
frol, line 786: This whole PR is full of "why"s to me.

Why do we need to generate a random seed on every insertion of a transaction?

Why `[3; 32]` and `[4; 32]`? (All magical consts should have names to understand their nature)

Why `random_seed` below is implemented that way? (why 17 and 23? why `shard_id / 256` - I feel it is going to be 0 most of the time)

chain/chunks/src/lib.rs
```diff
@@ -1730,6 +1732,12 @@ impl ShardsManager {
 
         Ok(())
     }
+    pub fn random_seed(&self, shard_id: ShardId) -> RngSeed {
+        let mut res = self.rng_seed.clone();
+        res[17] = shard_id as u8;
+        res[23] = (shard_id / 256) as u8;
+        res
+    }
```
bowenwang1996, line 1740: Please add some comments on what this function does

chain/chunks/src/lib.rs
bowenwang1996: yeah same question. Also this may have nontrivial impact on the performance of `insert_transaction`.

chain/chunks/src/lib.rs
```diff
@@ -397,13 +399,15 @@ pub struct ShardsManager {
     chunk_forwards_cache: SizedCache<ChunkHash, HashMap<u64, PartialEncodedChunkPart>>,
 
     seals_mgr: SealsManager,
+    rng_seed: RngSeed,
```
mm-near, line 402: add comment - mention what we're using the seed for.

chain/chunks/src/lib.rs
```diff
@@ -1723,6 +1734,17 @@ impl ShardsManager {
...
         Ok(())
     }
+
+    /// Computes a deterministic random seed for given `shard_id`.
+    /// This seed is used to randomize the transaction pool.
+    /// For better security we want the seed to different in each shard.
+    /// For testing purposes we want it to be the reproducible and derived from the `self.rng_seed` and `shard_id`
+    fn random_seed(&self, shard_id: ShardId) -> RngSeed {
+        let mut res = self.rng_seed.clone();
+        res[17] = (shard_id + 777) as u8;
```
mm-near, line 1744: what are these 'magic' constants?

chain/chunks/src/lib.rs
```diff
@@ -1757,6 +1779,7 @@ mod test {
             Some("test".parse().unwrap()),
             runtime_adapter,
             network_adapter.clone(),
+            [3; 32],
```
mm-near, line 1782: maybe create a constant  'TEST_SEED' - and use it in all these tests (rather than passing a 'cryptic' numeric value)

chain/client/src/client_actor.rs
```diff
@@ -1683,6 +1687,18 @@ impl Handler<StateSplitResponse> for ClientActor {
     }
 }
 
+/// Returns random seed sampled from the current thread
+pub fn random_seed_from_thread() -> RngSeed {
+    let mut rng_seed: RngSeed = [0; 32];
+    for j in 0..3 {
```
mm-near, line 1693: can you explain why do you want to create a seed from concatenating multiple values? 

RngSeed doesn't offer any other constructor?

integration-tests/tests/client/process_blocks.rs
```diff
@@ -1001,6 +1001,7 @@ fn test_process_invalid_tx() {
         false,
         network_adapter,
         chain_genesis,
+        [3; 32],
```
mm-near, line 1004: same comment as in other places - let's use constant here (TEST_SEED)

chain/client/src/test_utils.rs
```diff
@@ -1137,7 +1145,16 @@ impl TestEnvBuilder {
     fn new(chain_genesis: ChainGenesis) -> Self {
         let clients = Self::make_accounts(1);
         let validators = clients.clone();
-        Self { chain_genesis, clients, validators, runtime_adapters: None, network_adapters: None }
+        let mut seeds: HashMap<AccountId, RngSeed> = HashMap::with_capacity(16);
```
mm-near, line 1148: where does 16 come from?

chain/client/src/test_utils.rs
```diff
@@ -1137,7 +1145,16 @@ impl TestEnvBuilder {
...
+        seeds.insert(clients[0].clone(), [4; 32]);
```
mm-near, line 1149: nit: maybe set this seed for all the clients? (I know that in this context we have only one)

chain/client/src/test_utils.rs
```diff
@@ -1120,6 +1126,7 @@ pub struct TestEnv {
     pub network_adapters: Vec<Arc<MockPeerManagerAdapter>>,
     pub clients: Vec<Client>,
     account_to_client_index: HashMap<AccountId, usize>,
+    seeds: HashMap<AccountId, RngSeed>,
```
mm-near, line 1129: comment

chain/client/src/test_utils.rs
mm-near: and in theory this is not needed, as [4;32] is the default seed if one is missing from this hashmap

chain/client/src/test_utils.rs
mm-near: and mention what happens if an entry for account_id is not present in this map

mm-near (dismissed):

chain/chunks/src/lib.rs
```diff
@@ -795,17 +792,22 @@ impl ShardsManager {
             pool.remove_transactions(transactions)
         }
     }
+    fn pool_for_shard(&mut self, shard_id: ShardId) -> &mut TransactionPool {
+        match self.tx_pools.entry(shard_id) {
+            Vacant(_) => {
+                self.tx_pools.insert(shard_id, TransactionPool::new(self.random_seed(shard_id)));
```
bowenwang1996, line 800: What's the point of using the entry API if you call insert of the hashmap itself? You can do `e.insert...` where e is the value in the `Vacant` variant

chain/chunks/src/lib.rs
```diff
@@ -1732,10 +1734,15 @@ impl ShardsManager {
...
+    /// Computes a deterministic random seed for given `shard_id`.
+    /// This seed is used to randomize the transaction pool.
+    /// For better security we want the seed to different in each shard.
+    /// For testing purposes we want it to be the reproducible and derived from the `self.rng_seed` and `shard_id`
+    fn random_seed(&self, shard_id: ShardId) -> RngSeed {
         let mut res = self.rng_seed.clone();
-        res[17] = shard_id as u8;
-        res[23] = (shard_id / 256) as u8;
+        res[17] = (shard_id + 777) as u8;
+        res[23] = ((shard_id + 681) / 256) as u8;
```
bowenwang1996, line 1750: this really hurts. Even if there is a need to introduce "magic" constants they should be well documented. Please explain why these numbers are chosen in the first place. If there is no good reason for them, then `0` is usually a better choice.

bowenwang1996 (commented):
Usually if we need to choose a constant in a monoid I believe it is better to use its identity. Otherwise the question would be "why not use the identity?"

chain/chunks/src/lib.rs
matklad: FWIW, I am somewhat confused by the fact that we need to mess up with seeds at all, instead of passing an RNG instance.

chain/chunks/src/lib.rs
```diff
@@ -397,13 +399,17 @@ pub struct ShardsManager {
     chunk_forwards_cache: SizedCache<ChunkHash, HashMap<u64, PartialEncodedChunkPart>>,
 
     seals_mgr: SealsManager,
+    // Useful to make tests deterministic and reproducible,
+    // while keeping the security of randomization of transactions in pool
```
bowenwang1996, line 403:
```suggestion
    /// Useful to make tests deterministic and reproducible,
    /// while keeping the security of randomization of transactions in pool
```
also it is worth explaining where exactly this `RngSeed` is used

chain/chunks/src/lib.rs
bowenwang1996: This comment was not addressed

chain/chunks/src/lib.rs
```diff
@@ -1723,6 +1736,17 @@ impl ShardsManager {
...
         Ok(())
     }
+
+    /// Computes a deterministic random seed for given `shard_id`.
+    /// This seed is used to randomize the transaction pool.
+    /// For better security we want the seed to different in each shard.
+    /// For testing purposes we want it to be the reproducible and derived from the `self.rng_seed` and `shard_id`
+    fn random_seed(&self, shard_id: ShardId) -> RngSeed {
+        let mut res = self.rng_seed.clone();
+        res[0] = shard_id as u8;
```
bowenwang1996, line 1746: I believe this will panic if shard_id is >=256

chain/pool/src/lib.rs
```diff
@@ -20,15 +20,15 @@ pub struct TransactionPool {
...
     /// The key after which the pool iterator starts. Doesn't have to be present in the pool.
     last_used_key: PoolKey,
 }
 
 impl TransactionPool {
-    pub fn new() -> Self {
+    pub fn new(key_seed: RngSeed) -> Self {
         Self {
-            key_seed: rand::thread_rng().next_u64().to_le_bytes().to_vec(),
+            key_seed: key_seed, //rand::thread_rng().next_u64().to_le_bytes().to_vec(),
```
bowenwang1996, line 31: remove the old code

bowenwang1996 (commented):
Looks like CI failed

matklad:
To re-surface a high level question: 

Why do we need to store an rng *seed* instead of storing the whole *rng*? 

Ie, I am thinking about the following alternative: 

```diff
diff --git a/chain/chunks/src/lib.rs b/chain/chunks/src/lib.rs
index 267cca505..b82da7cd9 100644
--- a/chain/chunks/src/lib.rs
+++ b/chain/chunks/src/lib.rs
@@ -399,9 +399,7 @@ pub struct ShardsManager {
     chunk_forwards_cache: SizedCache<ChunkHash, HashMap<u64, PartialEncodedChunkPart>>,
 
     seals_mgr: SealsManager,
-    // Useful to make tests deterministic and reproducible,
-    // while keeping the security of randomization of transactions in pool
-    rng_seed: RngSeed,
+    rng: rand::rngs::StdRng,
 }
 
 impl ShardsManager {
```

Why we chose one other the other? (without diving into details, my prior is that using rng rather than seed is a more reliable solution, as it makes erroneous seed re-use impossible by construction).

yordanmadzhunkov:
> To re-surface a high level question:
> 
> Why do we need to store an rng _seed_ instead of storing the whole _rng_?
> 
> Ie, I am thinking about the following alternative:
> 
> ```diff
> diff --git a/chain/chunks/src/lib.rs b/chain/chunks/src/lib.rs
> index 267cca505..b82da7cd9 100644
> --- a/chain/chunks/src/lib.rs
> +++ b/chain/chunks/src/lib.rs
> @@ -399,9 +399,7 @@ pub struct ShardsManager {
>      chunk_forwards_cache: SizedCache<ChunkHash, HashMap<u64, PartialEncodedChunkPart>>,
>  
>      seals_mgr: SealsManager,
> -    // Useful to make tests deterministic and reproducible,
> -    // while keeping the security of randomization of transactions in pool
> -    rng_seed: RngSeed,
> +    rng: rand::rngs::StdRng,
>  }
>  
>  impl ShardsManager {
> ```
> 
> Why we chose one other the other? (without diving into details, my prior is that using rng rather than seed is a more reliable solution, as it makes erroneous seed re-use impossible by construction).

I see some weak benefits of using seed to the entire Random Generator:
1. I can write good test for setting/comparing random seed per shard id ( it's included in the review)
2. There is a opportunity to change how transaction pool is shuffled in the future without changing anything about the seed

chefsale (approved):

chain/chunks/src/lib.rs
```diff
@@ -879,15 +886,31 @@ impl ShardsManager {
...
     }
 
+    /// Computes a deterministic random seed for given `shard_id`.
+    /// This seed is used to randomize the transaction pool.
+    /// For better security we want the seed to different in each shard.
+    /// For testing purposes we want it to be the reproducible and derived from the `self.rng_seed` and `shard_id`
+    fn random_seed(base_seed: &RngSeed, shard_id: ShardId) -> RngSeed {
+        let mut res = base_seed.clone();
+        res[0] = shard_id as u8;
+        res[1] = (shard_id / 256) as u8;
```
bowenwang1996, line 896: This was still not fixed?

bowenwang1996 (approved):

matklad (approved):
I guess this is OK! I am still not 100% convinced that using seeds this way is correct, but I am not a domain expert. 

Some things which prevent me from being totally convinced: 

* We do not in fact use RngSeed as a seed for rng
* If you have a seed, there's a risk that you'll erroneously re-use it. The way the code currently written this can't happen (as `TransactionPool` map is append-only), but this is a potential bug to look out for in the future.
* We derive derived seeds in a way which is neither secure nor even random: derived seeds are essentially sequential numbers. I think this is OK as we feed them into crypto hash function later (eg, we don't need randomness, noncess seem enough), but I am not crypto expert, so I can't judge if `shard_id as u8` construct is good. 
* From my cursory reading of code, the type `RngSeed` was used for a very different thing previously. Specifically, it was used to refer to blockchain-related randomness ([eg](https://github.com/near/nearcore/blob/12f4d43f43897441e69bd9e9f262d0967ac449c3/nearcore/src/runtime/mod.rs#L1338)). In this PR we overload the type to mean *either* blockchain randmoness *or* the usual OS randomness.

yordanmadzhunkov (commented):
test

chain/chunks/src/lib.rs
```diff
@@ -879,15 +886,31 @@ impl ShardsManager {
...
+        res
+    }
+
+    fn pool_for_shard(&mut self, shard_id: ShardId) -> &mut TransactionPool {
+        match self.tx_pools.entry(shard_id) {
+            Vacant(entry) => entry
+                .insert(TransactionPool::new(ShardsManager::random_seed(&self.rng_seed, shard_id))),
+            Occupied(entry) => entry.into_mut(),
+        }
```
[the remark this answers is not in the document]
matklad: Yeah, that's subtle. The thing here is that Rust is smart enough to track borrowing on field-by-field basis, but that it intentionally stops at the function boundary. Consider these three examples: 

```rust
struct Example {
    a: i32,
    b: i32,
}

fn inc_a(ex: &mut Example) {
    ex.a += 1;
}

fn inc_a_bad(ex: &mut Example) {
    ex.a += 1;
    ex.b += 1;
}


#[test]
fn field_borrowing() {
    // OK.
    {
        let mut ex = Example { a: 0, b: 0 };
        let ar = &mut ex.a;
        let br = &mut ex.b;

        *ar += 1;
        *br += 1;
    }
    // Fails, but could work.
    {
        let mut ex = Example { a: 0, b: 0 };
        let br = &mut ex.b;

        inc_a(&mut ex);
        *br += 1;
    }
    // Fails, and should not work
    {
        let mut ex = Example { a: 0, b: 0 };
        let br = &mut ex.b;

        inc_a_bad(&mut ex);
        *br += 1;
    }
}
```

The first two do the equivalent thing -- exclusively borrowing disjoint fields, and then mutating things. The first works, as the compiler analyses borrowing within a function exactly. The second example is a compilation error. While `inc_a` indeed touches only `a`, compiler doesn't look into function body to verify that. It looks only at the signature, `fn inc_a(ex: &mut Example)`, and conservatively assumes that all fields are modified. 

That is, from the compiler's POV, second and third example look the same, and the third one definitely should be forbidden, as it unsoundly aliases `b`.


It *is* possible to make the compiler smarter, and to make it actually look into functions bodies. But that creates implications for programming, as now changes in one function can break other ones. This recent post by niko describes possible future language-level solutions: https://smallcultfollowing.com/babysteps//blog/2021/11/05/view-types/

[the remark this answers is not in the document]
yordanmadzhunkov: Well, Thank you for clarifying. Now I can sleep without having nightmares.

PR #5330, opened 2021-11-17
fix(pytest): respect deposit parameter in send_call_contract_raw_tx()

marcelo-gonzalez:
This param is currently not passed through to sign_function_call_tx(),
so we get 0 deposit no matter what

### DIFF
--- a/pytest/lib/account.py
+++ b/pytest/lib/account.py
@@ -66,7 +66,7 @@ class Account:
                                   deposit):
         self.prep_tx()
         tx = sign_function_call_tx(self.key, self.key.account_id, method_name,
-                                   args, 3 * 10**14, 0, self.nonce,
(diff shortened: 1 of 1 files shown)


pmnoxx:
@mina86 FYI I just found out that `S-automerge` tag doesn't fully work with external forks. 
Only merge works, but autorebase, and deletion of branches after merge is disabled.

mina86:
@nikurt this only affects mocknet, perhaps you’d want to take a look? To me this looks good but it does change behaviour of some of the mocknet code.

> @mina86 FYI I just found out that `S-automerge` tag doesn't fully work with external forks. Only merge works, but autorebase, and deletion of branches after merge is disabled.

I stopped trying to understand how `S-automerge` works months ago. ;)

mina86 (commented):
LGTM but let’s see what Nikolay has to say

PR #5332, opened 2021-11-17
feat: nightly protocol feature to lower regular op cost some more

matklad:
This adds a nightly feature to lower wasm opcode cost from 2_207_874 to
822_756.

To reproduce the cost measurement, checkout matklad/estimator-v2 branch

```
$ git rev-parse --short HEAD
07ea7157f

$ cargo run --release -p runtime-params-estimator --features required -- --docker --full --metrics-to-measure WasmInstruction
```

Note that that'll print the cost *without* safety multiplier, you need
to multiply by 3 to get the actual cost.

Three factors contribute to the cost reduction:

* First, we just wrote faster code in logic.rs to deduct gas in the case
  where burnt_gas = used_gas (common hot path).

* Second, we switched gas counter implementation to use gas counting
  intrinsic. That is, we speed up exactly the same gas counter impl we
  do today, by making the compiler specifically recognize our
  instrumentation. The primary motivation here is the ability to
  preserve existing behavior.

* Third, we now don't count time to compile the contract itself when
  estimating the cost. This matches what actually happens in practice,
  as we now compile contracts on deployment. See
  https://github.com/near/nearcore/pull/4733/commits/07ea7157fe333808e49687a9b000ce9ce539e101
  for this change. This is a somewhat last-minute thing, but I am fairly
  confident that that's the right thing to do.


bowenwang1996 (approved):

Longarithm (approved):

matklad:
```
λ diff nearcore/res/runtime_configs/124.json  nearcore/res/runtime_configs/123.json
154c154
<     "regular_op_cost": 822756,
---
>     "regular_op_cost": 2207874,
```

PR #5333, opened 2021-11-17
Embed `state-viewer` into `neard` binary

### LINKED ISSUE #5193: Package `state-viewer` with `neard`

nikurt:
Fix #5193


neard/src/cli.rs
```diff
@@ -57,6 +58,10 @@ impl NeardCmd {
                 info!(target: "neard", "Removing all data and config from {}", home_dir.to_string_lossy());
                 fs::remove_dir_all(home_dir).expect("Removing data and config failed.");
             }
+            NeardSubCommand::StateViewer(cmd) => {
+                info!(target: "neard", "Invoking state-viewer");
```
matklad, line 62: Do we need this log line? It doesn't add any new information to what the user typed themselves. The previous two lines print the path which is not neccessary passed in as a CLI argument, and they signify a dangerous information.

Meta: I feel that "overly chatty" tools are a big problem, as irrelevant printouts often make it hard to see the real error. To paraphrase Dijkstra, we should regard every line of output as a line spent.

neard/src/cli.rs
```diff
@@ -98,6 +103,9 @@ pub(super) enum NeardSubCommand {
     /// config)
     #[clap(name = "unsafe_reset_data")]
     UnsafeResetData,
+    /// View DB state by invoking `state-viewer`.
```
matklad, line 106:
```suggestion
    /// View DB state.
```

After this PR, `state-viewer` ceases to be a concept user needs to be aware of, it's just an impl detail.

neard/Cargo.toml
```diff
@@ -30,6 +30,7 @@ tikv-jemallocator = { version = "0.4.0", optional = true }
 nearcore = { path = "../nearcore" }
 near-primitives = { path = "../core/primitives" }
 near-performance-metrics = { path = "../utils/near-performance-metrics" }
+near-state-viewer = { path = "../test-utils/state-viewer", package = "state-viewer" }
```
matklad, line 33: I wonder if, as a seprate step, we should move the code itelf somewhere? `test-utils` is just an odd thing to depend on from prod. It's also OK as is -- I don't think our directory organization is super-great anyway.

matklad (approved):

neard/Cargo.toml
nikurt: `state-viewer` isn't a test util by util, moving it somewhere sounds good, let me follow up.

bowenwang1996 (approved):

test-utils/state-viewer/Cargo.toml
```diff
@@ -14,7 +14,6 @@ rayon = "1.5"
 serde = "1"
 serde_json = "1"
 tempfile = "3"
-tracing = "0.1"
```
mm-near, line 17: we don't need this anymore? (seems unrelated to this PR)

mm-near (approved):

PR #5334, opened 2021-11-17
refactor: Move EdgeVerifierHelper from routing.rs to edge_verifier_actor.rs

pmnoxx:
For historical reasons `EdgeVerifierHelper` was in `routing.rs`.

It makes more sense for that struct to be in `crate::routing::edge_verifier_actor`


chain/network/src/routing/edge_verifier_actor.rs
```diff
@@ -1,10 +1,12 @@
...
 use actix::{Actor, Handler, SyncContext, System};
-
+use conqueue::{QueueReceiver, QueueSender};
 use near_performance_metrics_macros::perf;
-
-use crate::types::{EdgeList, StopMsg};
+use near_primitives::borsh::maybestd::collections::HashMap;
+use near_primitives::borsh::maybestd::sync::{Arc, Mutex};
+use near_primitives::network::PeerId;
+use std::cmp::max;
```
mm-near, line 9: maybe add a comment on what EdgeVerifier does (and why we even have it in the first place - is it because of multithreading?)

mm-near (approved):

chain/network/src/routing/edge_verifier_actor.rs
pmnoxx: @mm-near I agree. My plan is following
- first fix obvious code quality issues - that's mostly done
- move structs in the correct place
- add crate docs
- add comments to all structs, functions. - that will become easier to documents, once everything is in the right place

pmnoxx: @mm-near I'm also in the middle of EdgeVerifier rewrite 
- pt1 - https://github.com/near/nearcore/pull/5230

In part 2 I'll rewrite it's logic to be simpler, it will be easier to document simple version of the code.
`EdgeVerifierHelper` will be going away, as it's to complicated.

PR #5337, opened 2021-11-17
feature: Run CI with nightly features + test_features

### LINKED ISSUE #5336: CI doesn't cover tests under `test_features`.
There was a test failing in `network/routing.rs` (https://github.com/near/nearcore/issues/5322). We didn't see that test fail because we don't run CI with those features enabled.

We should address that.

@chefsale

pmnoxx:
In this PR, we will be running CI with nightly feature, together with test features.

Closes #5336

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -9,10 +9,10 @@ steps:
     - "distro=amazonlinux"
     branches: "!master"
 
-  - label: "cargo test nightly"
+  - label: "cargo test nightly + test_features"
     command: |
       source ~/.cargo/env && set -eux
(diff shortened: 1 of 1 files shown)


pmnoxx:
@chefsale What do you think about this idea?

The alternative would be to create 3rd pipeline, with only those modules that have tests that use `test_features` flags.

```
  - label: "cargo test nightly + test_features"
    command: |
      RUST_BACKTRACE=1 RUSTFLAGS='-D warnings' cargo test --workspace --features nightly_protocol,nightly_protocol_features,test_features  -project integration-tests
```

bowenwang1996:
Is this the same as https://github.com/near/nearcore/pull/5320 ?

pmnoxx:
@bowenwang1996 Thanks, that's seems to be a duplicate. I'll close this PR.+

PR #5341, opened 2021-11-17
Remove a superfluous BorshDeserialize derive

nagisa:
This derive cannot work for a type containing references, because
there are no implementations of `BorshDeserialize` for a reference of
`T: BorshDeserialize`. However `rustc` for some reason fails to verify
this in typeck, leading to an ICE in codegen when the `-Clink-dead-code`
flag actually forces the codegen of this dead code and the obligation
mismatch is discovered.


bowenwang1996 (approved):
Thanks!

pmnoxx:
LGTM, I added `automerge` and merged the change.

PR #5342, opened 2021-11-17
refactor: Mark attributes as private in `near-network::types`

pmnoxx:
In effort to better define what the api of near-network is. I'm going to mark attributes which are only used in `near-network` as `pub(crate)`  instead of `pub`.

This will make it clearer to see what's only used inside `near-network` and what's not. More PRs will be needed to finish this cleanup.


pmnoxx:
@mm-near Can you take a look?

mm-near (approved):

Issue #5149, opened 2021-11-05
Move vm_logic tests into a single crate


matklad:
Currently, each file in [https://github.com/near/nearcore/tree/master/runtime/near-vm-logic/tests](https://github.com/near/nearcore/tree/master/runtime/near-vm-logic/tests) is a separately compiled crate, which hurts compile times. Additionally, the vm_logic_builder.rs code is effectively compiled several times, once for each test.

Let's move the tests from `near-vm-logic/tests` to `near-vm-logic/src/tests`, as per [https://matklad.github.io/2021/02/27/delete-cargo-integration-tests.html](https://matklad.github.io/2021/02/27/delete-cargo-integration-tests.html)

matklad:
While we are at it, might tweak VMLogicBuilder as well to remove `context` argument from the `build` methond, and instead `.with_context()`, `.is_view`, etc, methods on the builder itself.

matklad:
cc @akhi3030

PR #5343, opened 2021-11-18
Adds process_block_test to be used in tests

mzhangmzz:
to reduce code repetition. Also helps me in https://github.com/near/nearcore/pull/5224 where I changed the signature of process_block


yordanmadzhunkov (approved):
Cool

matklad (approved):
LGTM! 

Although, as an alternative, I suggest the following:

* move chain tests from `chain/tests` to `chain/src/tests`, a-la https://github.com/near/nearcore/issues/5149, which should reduce our overall compile time
* after that, we'll be able to mark `fn process_block_test` as `#[test]`, and also make it `pub(crate)` rather than `pub`.

mzhangmzz:
> LGTM!
> 
> Although, as an alternative, I suggest the following:
> 
> * move chain tests from `chain/tests` to `chain/src/tests`, a-la #5149, which should reduce our overall compile time
> * after that, we'll be able to mark `fn process_block_test` as `#[test]`, and also make it `pub(crate)` rather than `pub`.

@matklad I did your suggestion but only moved simple_chain.rs and challenges.rs, which are the tests that this PR touched, to chain/src/tests. I didn't move the rest of the tests because some of them look rather complicated and I'm not sure if I should leave them as integration tests. 

I read your post here about how to organize tests https://matklad.github.io/2021/02/27/delete-cargo-integration-tests.html and I think there are many other places in the code that don't follow this rule right now. It should take a separate PR to fix them all. I'll create an issue to track that.

matklad:
>look rather complicated and I'm not sure if I should leave them as integration tests.

I don't think test complexity should define whether the test goes to `tests/` or `src/tests` -- cargo's integration/unit terminology is just confusing. (in general, if I were a king of the world, I'd just ban both terms and asked people to use "software engineering at google"'s terminology of test size (one thread/one process/one machine/one distributed system)). It's rather "test in the same compilation unit" vs "tests in a separate compilation unit", with the difference between the two being access to `pub(crate)` and `#[cfg(test)]` APIs.

Issue #4490, opened 2021-07-09
Make `cargo test` usable for nearcore developers


matklad:
I've noticed that I never run `cargo test --workspace` locally. I do want to have an ability to quickly check if I broke something somewhere, but the usability of our current `cargo test` is abysmal. 

Specifically: 

* Tests output *a lot* to the terminal, making terminal scrollback buffer useless. 
* Test suite is prohibitively slow. It takes 15 wall-clock minutes / 12 cpu minutes on a 24-core machine.

I suggest immediate AIs:

* fixing the tests such that the output of `cargo test` is never polluted unless the test actually fails. I think using [`TestWriter`](https://docs.rs/tracing-subscriber/0.2.19/tracing_subscriber/fmt/struct.TestWriter.html) is one possible way to do this
* moving all obviously slow tests to `expensive_tests`
* revisiting this issue after this easy wins are achieved.

Longer term, I think it might sense to do the following: 

* replace `expensive_tests` with run-time check (see the example at the end of "Make Tests Fast" section of [this post](https://matklad.github.io/2021/05/31/how-to-test.html#make-tests-fast)).
* add two tiers of `expensive_tests` -- `nightly_only_tests` (> 5 min) and `ci_only_tests` (>10s, <5min). Local `cargo t` skips `ci_only_tests` by default, but they are run by CI for every commit that gets into master
* document the expectations about tests run-time and actively fix regressions. 
* aim for really short test turn-around time. To give an aspiration goal, in rust-analyzer `cargo t` takes 5s wall-clock/35s cpu by default and 30s/100s when including ci only tests.

bowenwang1996:
@matklad this is fixed right?

matklad:
Not at all:

* running `cargo test` still outputs megabytes of DEBUG-level logging to the terminal, 
* it still takes 15 minutes to complete on a 12-core machine

pmnoxx:
@matklad The solution for having many prints is to run the code with
`RUST_LOG=warn cargo test --workspace`.

By setting proper environment flags you eliminate the issue.

matklad:
Seems reasonable to flip the default? Run with `RUST_LOG=INFO` if you need the logs, but get no unnecessary verbose output by default. Generally you only want to see logs for failing tests.

pmnoxx:
> Seems reasonable to flip the default? Run with `RUST_LOG=INFO` if you need the logs, but get no unnecessary verbose output by default. Generally you only want to see logs for failing tests.

[...]

PR #5346, opened 2021-11-18
utils: add a few more examples and tests to MaybeValidated

mina86:
Extend existing and introduce new examples of MaybeValidated usage
which test the type a bit more extensively.


core/primitives/src/utils.rs
```diff
@@ -98,17 +112,43 @@ impl<T> MaybeValidated<T> {
...
+    ///
+    /// ```
+    /// use near_primitives::utils::MaybeValidated;
+    ///
+    /// let value = MaybeValidated::from(42);
+    /// let value_as_ref = value.as_ref();
+    /// assert_eq!(Ok(true), value_as_ref.validate_with::<(), _>(|&&v| Ok(v == 42)));
+    /// assert!(value_as_ref.is_validated());
+    /// assert!(!value.is_validated());
+    /// ```
```
matklad, line 142: I tend to think that doctests are not worth it for non-library crates in general. The problem with doctests is that they are notoriously slow to compile and run, as each doctest is compiled as a separate binary. For example, in rust-analyzer, removing `doctest = false` from Cargo.tomls increases the wall-clock time for the testsuite from 5 seconds to 10 seconds, and ra doesn't even have doctets.

OTOH, our tests are already pretty slow, and we dont' add `doctest=false` to cargo.tomls, so this isn't a regression for nearcore. So, no strong opinion here!

matklad (approved):

core/primitives/src/utils.rs
mina86: I don’t have an opinion either way. Wasn’t really aware that there was noticeable compile-time cost. I’m happy to rewrite it as `[test]`.

matklad: Nah, I guess let's leave it as is -- we do seem to have some doctests scattered elsewhere. Although, once #4490 is fixed, it makes sense to quantify how bad are doctests for our case. Maybe by that time https://github.com/rust-lang/rust/issues/75341 is fixed, which would be ideal :)

mm-near (approved):

Issue #5347, opened 2021-11-18
Allow specifying the number of epochs for data retention in config


bowenwang1996:
Currently for a regular node the number of epochs for which we retain data is fixed (5 epochs) and we have received requests to make that flexible, since some partners have requirements on data retention.

pmnoxx:
@bowenwang1996 It may be a good idea to mention where that constant is stored in the code. That would allow people to judge how easy it is.

pmnoxx:
@bowenwang1996 Is this for `neard` executable?

PR #5348, opened 2021-11-18
feat: Add `num_epoch_to_store` to `config.json`

pmnoxx:
Currently for a regular node the number of epochs for which we retain data is fixed (5 epochs) and we have received requests to make that flexible, since some partners have requirements on data retention.

We are going to add `num_epoch_to_store` to `Config` with default value of `DEFAULT_NUM_EPOCHS_TO_KEEP_STORE_DATA  = 5`.

Resolves https://github.com/near/nearcore/issues/5347

### DIFF
--- a/chain/chain/src/chain.rs
+++ b/chain/chain/src/chain.rs
@@ -97,7 +97,8 @@ pub const TX_ROUTING_HEIGHT_HORIZON: BlockHeightDelta = 4;
 /// Private constant for 1 NEAR (copy from near/config.rs) used for reporting.
 const NEAR_BASE: Balance = 1_000_000_000_000_000_000_000_000;
 
-/// Number of epochs for which we keep store data
+/// Default number of epochs for which we keep store data.
+/// This value may be overridden by `Config.additional_epochs_to_keep`.
 pub const NUM_EPOCHS_TO_KEEP_STORE_DATA: u64 = 5;
 
 /// Maximum number of height to go through at each step when cleaning forks during garbage collection.
--- a/chain/epoch_manager/src/lib.rs
+++ b/chain/epoch_manager/src/lib.rs
@@ -70,12 +70,15 @@ pub struct EpochManager {
     epoch_info_aggregator: Option<EpochInfoAggregator>,
     /// Largest final height. Monotonically increasing.
     largest_final_height: BlockHeight,
+    /// Number of additional epochs to keep in memory.
+    pub additional_epochs_to_keep: u64,
 }
 
 impl EpochManager {
     pub fn new_from_genesis_config(
         store: Store,
         genesis_config: &GenesisConfig,
+        additional_epochs_to_keep: u64,
     ) -> Result<Self, EpochError> {
         let reward_calculator = RewardCalculator::new(genesis_config);
         let all_epoch_config = AllEpochConfig::from(genesis_config);
@@ -85,6 +88,7 @@ impl EpochManager {
             genesis_config.protocol_version,
             reward_calculator,
             genesis_config.validators(),
+            additional_epochs_to_keep,
         )
     }
 
@@ -94,6 +98,7 @@ impl EpochManager {
         genesis_protocol_version: ProtocolVersion,
         reward_calculator: RewardCalculator,
         validators: Vec<ValidatorStake>,
+        additional_epochs_to_keep: u64,
     ) -> Result<Self, EpochError> {
         let validator_reward = vec![(reward_calculator.protocol_treasury_account.clone(), 0u128)]
             .into_iter()
@@ -110,6 +115,7 @@ impl EpochManager {
             epoch_validators_ordered_unique: lru::LruCache::new(EPOCH_CACHE_SIZE),
             epoch_info_aggregator: None,
             largest_final_height: 0,
+            additional_epochs_to_keep,
         };
         let genesis_epoch_id = EpochId::default();
         if !epoch_manager.has_epoch_info(&genesis_epoch_id)? {
@@ -1626,6 +1632,7 @@ mod tests2 {
                 .iter()
                 .map(|(account_id, balance)| stake(account_id.clone(), *balance))
                 .collect(),
+            0,
         )
         .unwrap();
         assert!(compare_epoch_infos(epoch_manager2.get_epoch_info(&epoch3).unwrap(), &expected3));
@@ -1894,6 +1901,7 @@ mod tests2 {
             PROTOCOL_VERSION,
             default_reward_calculator(),
             validators,
+            0,
         )
         .unwrap();
         let h = hash_range(8);
@@ -1964,6 +1972,7 @@ mod tests2 {
             PROTOCOL_VERSION,
             default_reward_calculator(),
             validators,
+            0,
         )
         .unwrap();
 
@@ -2036,6 +2045,7 @@ mod tests2 {
             PROTOCOL_VERSION,
             default_reward_calculator(),
             validators,
+            0,
         )
         .unwrap();
 
@@ -3706,7 +3716,8 @@ mod tests2 {
             stake("test2".parse().unwrap(), amount_staked),
         ];
         let mut epoch_manager =
-            EpochManager::new(store, config, 0, default_reward_calculator(), validators).unwrap();
+            EpochManager::new(store, config, 0, default_reward_calculator(), validators, 0)
+                .unwrap();
         let h = hash_range(8);
         record_block(&mut epoch_manager, CryptoHash::default(), h[0], 0, vec![]);
         let mut block_info1 =
@@ -3750,6 +3761,7 @@ mod tests2 {
             new_protocol_version - 1,
             default_reward_calculator(),
             validators,
+            0,
         )
         .unwrap();
         let h = hash_range(8);
@@ -3837,7 +3849,8 @@ mod tests2 {
             stake("test2".parse().unwrap(), amount_staked / 5),
         ];
         let mut epoch_manager =
-            EpochManager::new(store, config, 0, default_reward_calculator(), validators).unwrap();
+            EpochManager::new(store, config, 0, default_reward_calculator(), validators, 0)
+                .unwrap();
         let h = hash_range(50);
         record_block(&mut epoch_manager, CryptoHash::default(), h[0], 0, vec![]);
         let mut block_info1 =
@@ -3873,6 +3886,7 @@ mod tests2 {
             UPGRADABILITY_FIX_PROTOCOL_VERSION,
             default_reward_calculator(),
             validators,
+            0,
         )
         .unwrap();
         let h = hash_range(5 * epoch_length);
--- a/integration-tests/src/tests/client/process_blocks.rs
+++ b/integration-tests/src/tests/client/process_blocks.rs
@@ -95,6 +95,23 @@ pub fn create_nightshade_runtimes(genesis: &Genesis, n: usize) -> Vec<Arc<dyn Ru
         .collect()
 }
 
+pub fn create_nightshade_runtimes_with_num_epochs(
+    genesis: &Genesis,
+    n: usize,
+    additional_epochs_to_keep: u64,
+) -> Vec<Arc<dyn RuntimeAdapter>> {
+    (0..n)
+        .map(|_| {
+            Arc::new(nearcore::NightshadeRuntime::test_with_num_epochs(
+                Path::new("."),
+                create_test_store(),
+                genesis,
+                additional_epochs_to_keep,
+            )) as Arc<dyn RuntimeAdapter>
+        })
+        .collect()
+}
+
 /// Produce `blocks_number` block in the given environment, starting from the given height.
 /// Returns the first unoccupied height in the chain after this operation.
 fn produce_blocks_from_height(
@@ -1556,6 +1573,42 @@ fn test_gc_after_state_sync() {
     assert!(env.clients[1].runtime_adapter.get_epoch_id_from_prev_block(&prev_block_hash).is_ok());
     let tries = env.clients[1].runtime_adapter.get_tries();
     assert!(env.clients[1].chain.clear_data(tries, 2).is_ok());
+}
+
+/// This tests the `additional_epochs_to_keep` setting, which specifies the number of epochs to keep.
+///
+/// For this test, we set epoch length to `128`, and `additional_epochs_to_keep` to `5`.
+/// This tests shows, we keep `9` full epochs, and one partial epoch with `epoch_length - 100` blocks.
+#[test]
+fn test_num_blocks_in_storage_config_setting() {
+    let epoch_length = 128;
+    let epochs_to_keep = 10;
+
+    let mut genesis = Genesis::test(vec!["test0".parse().unwrap(), "test1".parse().unwrap()], 1);
+    genesis.config.epoch_length = epoch_length;
+    let mut chain_genesis = ChainGenesis::test();
+    chain_genesis.epoch_length = epoch_length;
+    let mut env = TestEnv::builder(chain_genesis)
+        .clients_count(2)
+        .runtime_adapters(create_nightshade_runtimes_with_num_epochs(
+            &genesis,
+            2,
+            epochs_to_keep - 5,
+        ))
+        .build();
+    for i in 1..epoch_length * (epochs_to_keep + 1) + 2 {
+        let block = env.clients[0].produce_block(i).unwrap().unwrap();
+        env.process_block(0, block.clone(), Provenance::PRODUCED);
+        env.process_block(1, block, Provenance::NONE);
+    }
+    // Check for GC'ed blocks.
+    for i in 1..100 + epoch_length {
+        assert!(env.clients[0].chain.get_block_by_height(i).is_err());
+    }
+    // Check whenever we still keep blocks.
+    for i in epoch_length + 101..epoch_length * (epochs_to_keep) + 2 {
+        env.clients[0].chain.get_block_by_height(i).unwrap();
+    }
 }
 
 #[test]
@@ -3352,6 +3405,7 @@ fn test_limit_contract_functions_number_upgrade() {
                 &genesis,
                 TrackedConfig::new_empty(),
                 RuntimeConfigStore::new(None),
+                0,
             ))];
         let mut env = TestEnv::builder(chain_genesis).runtime_adapters(runtimes).build();
 
@@ -3730,6 +3784,7 @@ mod access_key_nonce_range_tests {
                     &genesis,
                     TrackedConfig::AllShards,
                     RuntimeConfigStore::test(),
+                    0,
                 )) as Arc<dyn RuntimeAdapter>
             })
             .collect();
@@ -3876,6 +3931,7 @@ mod access_key_nonce_range_tests {
                     &genesis,
                     TrackedConfig::AllShards,
                     RuntimeConfigStore::test(),
+                    0,
                 )) as Arc<dyn RuntimeAdapter>
             })
             .collect();
--- a/nearcore/src/config.rs
+++ b/nearcore/src/config.rs
@@ -311,6 +311,10 @@ fn default_gc_blocks_limit() -> NumBlocks {
     2
 }
 
+fn default_additional_epochs_to_keep() -> u64 {
+    0
+}
+
 fn default_view_client_threads() -> usize {
     4
 }
@@ -453,6 +457,8 @@ pub struct Config {
     /// For example, setting "use_db_migration_snapshot" to "/tmp/" will create a directory "/tmp/db_migration_snapshot" and populate it with the database files.
     #[serde(skip_serializing_if = "Option::is_none")]
     pub db_migration_snapshot_path: Option<PathBuf>,
+    #[serde(default = "default_additional_epochs_to_keep")]
+    pub additional_epochs_to_keep: u64,
 }
 
 impl Default for Config {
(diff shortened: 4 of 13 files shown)


pmnoxx:
@chefsale @mhalambek We are making changes to config can you take a look?

pmnoxx:
@bowenwang1996 What's the best way to test it besides starting a node with new config?

mzhangmzz:
@pmnoxx I think you can test by creating a `TestEnv` to produce and process blocks for a few epochs and then that an old block is no longer in storage. For example, you can see `test_gc_after_state_sync`

nearcore/src/config.rs
```diff
@@ -431,6 +432,7 @@ pub struct Config {
     /// If set, overrides value in genesis configuration.
     #[serde(skip_serializing_if = "Option::is_none")]
     pub max_gas_burnt_view: Option<Gas>,
+    pub num_epochs_to_keep_store_data: u64,
```
bowenwang1996, line 435: Please add a default value for serde, similar to what we did for other fields

nearcore/src/runtime/mod.rs
```diff
@@ -1224,7 +1229,8 @@ impl RuntimeAdapter for NightshadeRuntime {
             // maintain pointers to avoid cloning.
             let mut last_block_in_prev_epoch = *epoch_first_block_info.prev_hash();
             let mut epoch_start_height = *epoch_first_block_info.height();
-            for _ in 0..NUM_EPOCHS_TO_KEEP_STORE_DATA - 1 {
+
+            for _ in 0..self.num_epochs_to_keep_store_data - 1 {
```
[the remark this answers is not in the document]
pmnoxx: I changed the value to be minimum 5.

nearcore/src/runtime/mod.rs
```diff
@@ -172,6 +174,7 @@ impl NightshadeRuntime {
         trie_viewer_state_size_limit: Option<u64>,
         max_gas_burnt_view: Option<Gas>,
         runtime_config_store: Option<RuntimeConfigStore>,
+        num_epochs_to_keep_store_data: u64,
```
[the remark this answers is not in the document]
pmnoxx: @frol Yes, that's a good idea. I renamed it and made it at least 5.

nearcore/src/config.rs
pmnoxx: Ok, I figured it out and updated the PR.

pmnoxx:
> @pmnoxx I think you can test by creating a `TestEnv` to produce and process blocks for a few epochs and then that an old block is no longer in storage. For example, you can see `test_gc_after_state_sync`, but minus the state sync part

Thanks, I added a test:
```

#[test]
fn test_num_epochs_to_keep_store_data() {
    let epoch_length = 1024;
    let mut genesis = Genesis::test(vec!["test0".parse().unwrap(), "test1".parse().unwrap()], 1);
    genesis.config.epoch_length = epoch_length;
    let mut chain_genesis = ChainGenesis::test();
    chain_genesis.epoch_length = epoch_length;
    let mut env = TestEnv::builder(chain_genesis)
        .clients_count(2)
        .runtime_adapters(create_nightshade_runtimes(&genesis, 2))
        .build();
    // TODO figured out how to change `config.num_epochs_to_keep_store_data`
    let num_epochs_to_keep_store_data = 10;
    for i in 1..epoch_length * 4 + 2 {
        let block = env.clients[0].produce_block(i).unwrap().unwrap();
        env.process_block(0, block.clone(), Provenance::PRODUCED);
        env.process_block(1, block, Provenance::NONE);
    }
    for i in 0..num_epochs_to_keep_store_data {
        let sync_height = epoch_length * 4 + 1 - i;
        env.clients[0].chain.get_block_by_height(sync_height).unwrap();
    }
}
```
All that's left is to figure out how to set `num_epochs_to_keep_store_data` in the config.

pmnoxx:
@frol @mzhangmzz 
Ok, it works. Should I ship it with test or not? The test changes look a bit ugly, maybe there is a cleaner way to pass the value.

nearcore/src/config.rs
```diff
@@ -458,11 +466,16 @@ impl Default for Config {
...
             trie_viewer_state_size_limit: default_trie_viewer_state_size_limit(),
             max_gas_burnt_view: None,
+            num_epochs_to_keep_store_data: default_num_epochs_to_keep_store_data(),
         }
     }
 }
 
 impl Config {
+    pub fn num_epochs_to_keep_store_data(&self) -> u64 {
+        max(MIN_NUM_EPOCHS_TO_KEEP_STORE_DATA, self.num_epochs_to_keep_store_data)
```
frol, line 476: I would validate and fail starting the node instead of silently ignoring the value from the config. WDYT?

frol (approved):

nearcore/src/config.rs
pmnoxx: @frol Yes, I agree. Where should the validation be done, maybe something like this:

    pub fn num_epochs_to_keep_store_data(&self) -> u64 {
        if self.num_epochs_to_keep_store_data < MIN_NUM_EPOCHS_TO_KEEP_STORE_DATA {
              panic!("num_epochs_to_keep_store_data: {} is below minimum: {}, 
                        self.num_epochs_to_keep_store_data,
                        MIN_NUM_EPOCHS_TO_KEEP_STORE_DATA)
        }
        self.num_epochs_to_keep_store_data
   }

frol: @pmnoxx this is a weird place to do the validation. It should be done on deserialization/read-from-file/constructor stage

pmnoxx: @frol How about inside:
```

    pub fn from_file(path: &Path) -> Self {
        let mut file = File::open(path)
            .unwrap_or_else(|_| panic!("Could not open config file: `{}`", path.display()));
        file.read_to_string(&mut content).expect("Could not read from config file.");
        Config::from(content.as_str())
    }
```

pmnoxx: @frol I updated the PR to do it inside `from_file`

bowenwang1996:
@pmnoxx please consolidate this PR with #5559. It is hard to understand which PR should be reviewed.

pmnoxx:
> @pmnoxx please consolidate this PR with #5559. It is hard to understand which PR should be reviewed.

Everything for the review is already in this PR. I closed #5559 to avoid confusion.

integration-tests/tests/client/process_blocks.rs
```diff
@@ -1558,6 +1558,30 @@ fn test_gc_after_state_sync() {
     assert!(env.clients[1].chain.clear_data(tries, 2).is_ok());
 }
 
+#[test]
+fn test_blocks_in_storage() {
+    let epoch_length = 1024;
+    let mut genesis = Genesis::test(vec!["test0".parse().unwrap(), "test1".parse().unwrap()], 1);
+    genesis.config.epoch_length = epoch_length;
+    let mut chain_genesis = ChainGenesis::test();
```
mzhangmzz, line 1566: shouldn't you change the `num_epoch_to_store` here in the test to a non-default value and test that?

integration-tests/tests/client/process_blocks.rs
```diff
@@ -1558,6 +1558,30 @@ fn test_gc_after_state_sync() {
...
+    for i in 1..epoch_length * 4 + 2 {
+        let block = env.clients[0].produce_block(i).unwrap().unwrap();
+        env.process_block(0, block.clone(), Provenance::PRODUCED);
+        env.process_block(1, block, Provenance::NONE);
+    }
+    // Check whenever we got all 5 blocks
+    for i in 0..5 {
+        let sync_height = epoch_length * 4 + 1 - i;
+        env.clients[0].chain.get_block_by_height(sync_height).unwrap();
+    }
```
mzhangmzz, line 1582: Maybe you should keep producing and processing blocks for a few more blocks to verify the blocks not exist after `num_epoch_to_store` epochs.

integration-tests/tests/client/process_blocks.rs
pmnoxx: @mzhangmzz Oh, sorry. I just noticed that the test function wasn't updated yet. Please, take a look again.

integration-tests/tests/client/process_blocks.rs
pmnoxx: It's here after the update:
```
// Check for GC'ed blocks.
    for i in 1..100 {
        assert!(env.clients[0].chain.get_block_by_height(i).is_err());
    }
    // Check whenever we still keep blocks.
    for i in 101..epoch_length * (num_epochs_to_keep_store_data) + 2 {
        env.clients[0].chain.get_block_by_height(i).unwrap();
    }
```

integration-tests/tests/client/process_blocks.rs
```diff
@@ -1558,6 +1575,38 @@ fn test_gc_after_state_sync() {
     assert!(env.clients[1].chain.clear_data(tries, 2).is_ok());
 }
 
+#[test]
+fn test_num_blocks_in_storage_config_setting() {
+    let epoch_length = 128;
+    // TODO - figured out how to change `num_epochs_to_keep_store_data = 10`
```
mzhangmzz, line 1581: This TODO seems to be done

integration-tests/tests/client/process_blocks.rs
```diff
@@ -1558,6 +1575,38 @@ fn test_gc_after_state_sync() {
...
+            2,
+            Some(num_epochs_to_keep_store_data),
+        ))
+        .build();
+    for i in 1..epoch_length * (num_epochs_to_keep_store_data) + 2 {
+        let block = env.clients[0].produce_block(i).unwrap().unwrap();
+        env.process_block(0, block.clone(), Provenance::PRODUCED);
+        env.process_block(1, block, Provenance::NONE);
+    }
+    // Check for GC'ed blocks.
```
mzhangmzz, line 1600: I'm confused why the number is 100 here. Shouldn't it be a number close to 128?

pmnoxx: @mzhangmzz I'm confused as well. Blocks from 1..100 are unavailable, wile blocks 101... are available to read.

integration-tests/tests/client/process_blocks.rs
pmnoxx: Removed

integration-tests/tests/client/process_blocks.rs
pmnoxx: @mzhangmzz I added comments to the test.

```

/// This field tests `num_epochs_to_keep_store_data` setting.
/// We test, we keep the number of epochs proportional to the data.
///
/// For this test, we set epoch length to `128`, and `num_epochs_to_keep_store_data` to 10.
/// This tests shows, we keep 9 full epochs, and one partial epoch with `128-100` blocks.
/// I know it's weird, right?
```

pmnoxx:
@bowenwang1996 @mzhangmzz @matklad 

I think it may be a good idea to rename `num_epochs_to_keep_store_data` to `num_epochs_to_store`. This will make the field easier to read.

Also we need to decide on the meaning of `num_epochs_to_keep_store_data`.
We have two options:
A) current epoch + `num_epochs_to_keep_store_data - 1` full epochs
B) current epoch + `num_epochs_to_keep_store_data` full epochs

Currently we use option `A`, the default value of 5, which indicates we will store at least 4 full epochs.

Although, option `B` may be more intuitive.

integration-tests/tests/client/process_blocks.rs
```diff
@@ -1558,6 +1575,44 @@ fn test_gc_after_state_sync() {
     assert!(env.clients[1].chain.clear_data(tries, 2).is_ok());
 }
 
+/// This tests the `num_epochs_to_keep_store_data` setting, which specifies the number of epochs
+/// to keep.
+///
+/// For this test, we set epoch length to `128`, and `num_epochs_to_keep_store_data` to 10.
+/// This tests shows, we keep 9 full epochs, and one partial epoch with `epoch_length`-`100` blocks.
```
mm-near, line 1582: where does the 'magic' 100 come from?

if we have num_epochs_to_keep set to 10, I'd have expected all 10 epochs to be stored (and not 9.5) - unless I misunderstand this comment

mm-near (approved):

integration-tests/tests/client/process_blocks.rs
pmnoxx: @mm-near That's a good question. I tried adjusting the value of epoch_length, let's say to 1024, and we still keep 9 full epoch + one `(epoch_length - 100)`.

chain/chain/src/chain.rs
```diff
@@ -84,8 +84,9 @@ pub const TX_ROUTING_HEIGHT_HORIZON: BlockHeightDelta = 4;
 /// Private constant for 1 NEAR (copy from near/config.rs) used for reporting.
 const NEAR_BASE: Balance = 1_000_000_000_000_000_000_000_000;
 
-/// Number of epochs for which we keep store data
-pub const NUM_EPOCHS_TO_KEEP_STORE_DATA: u64 = 5;
+/// Default number of epochs for which we keep store data.
+/// This value may be overridden by value from settings.
```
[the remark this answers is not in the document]
pmnoxx: from `Config.num_epoch_to_store`

pmnoxx:
```
@matklad: /test-utils/runtime-tester
@ailisp: /genesis-tools/
```
@matklad @ailisp Can you take a look?

nikurt (approved):

nearcore/src/config.rs
```diff
@@ -468,7 +476,14 @@ impl Config {
...
         let mut content = String::new();
         file.read_to_string(&mut content).expect("Could not read from config file.");
-        Config::from(content.as_str())
+        let config = Config::from(content.as_str());
+        if config.num_epochs_to_keep_store_data < MIN_NUM_EPOCHS_TO_KEEP_STORE_DATA {
+            panic!(
+                "num_epochs_to_keep_store_data: {} is below: {}",
+                config.num_epochs_to_keep_store_data, MIN_NUM_EPOCHS_TO_KEEP_STORE_DATA
+            );
+        }
```
matklad, line 485: Is this the right place to put this validation? Seems rather easy to dodge by, eg, parsing from string directly.

matklad: My gut feeling is that it makes sense to spend effort to make such validation tight. In this case, I'd say that the most robust and simplest thing would be to just add 

```rust
let num_epochs_to_keep_store_data = num_epochs_to_keep_store_data.max(MIN_NUM_EPOCHS_TO_KEEP_STORE_DATA)
```

to the NightshadeRuntime::new

matklad (changes requested):
If we enforce `>= 5` condition on some paths, we should make sure we enforce it on all paths.

pmnoxx:
```
@matklad @posvyatokum: test-utils/runtime-tester
```
@matklad @posvyatokum Can you take a look?

matklad (approved):
I don't like the `.verify()` and `.max(MIN_NUM_EPOCHS_TO_KEEP_STORE_DATA),` APIs. There's nothing very specific wrong with them, as written, but they make it possible to, eg, *forget* to `.verify()`. I would probably try harder to *statically* guarantee that all the checks are made. 

One relatively easy way to do that is to use "additional_epochs_to_keep" as a config, with a default of zero. Then no validation is necessary. 


Approving to unblock, but overall it semes that the current impl is not as clean as it could reasonably be.

nearcore/src/runtime/mod.rs
```diff
@@ -141,6 +141,7 @@ pub struct NightshadeRuntime {
     shard_tracker: ShardTracker,
     genesis_state_roots: Vec<StateRoot>,
     migration_data: Arc<MigrationData>,
+    num_epochs_to_keep_store_data: u64,
```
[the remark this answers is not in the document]
bowenwang1996: > @mzhangmzz does it make sense that NightshadeRuntime knows when we start GCing epochs? Would something like EpochManager or Client be a better place for this info?

Yeah I think client is probably a better place for this. cc @pmnoxx

chefsale (approved):

bowenwang1996:
@pmnoxx could you resolve the merge conflicts? Thanks!

nearcore/src/runtime/mod.rs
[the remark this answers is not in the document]
pmnoxx: @matklad We only use `num_epochs_to_keep_store_data` inside `NightshadeRuntime `.

We can put it inside `EpochManager`, I see that nightshade accesses it through `Arc<RwLock<EpochManager>>`. This isn't much of an issue, because we take as look anyway.

pmnoxx:
> I don't like the `.verify()` and `.max(MIN_NUM_EPOCHS_TO_KEEP_STORE_DATA),` APIs. There's nothing very specific wrong with them, as written, but they make it possible to, eg, _forget_ to `.verify()`. I would probably try harder to _statically_ guarantee that all the checks are made.
> 
> One relatively easy way to do that is to use "additional_epochs_to_keep" as a config, with a default of zero. Then no validation is necessary.
> 
> Approving to unblock, but overall it semes that the current impl is not as clean as it could reasonably be.

Sure, we can change it to be `additional_epochs_to_keep` to keep, that will remove the need to use `verify`.

pmnoxx:
@bowenwang1996 @matklad Can you take a look again?

chain/epoch_manager/src/lib.rs
```diff
@@ -71,12 +71,15 @@ pub struct EpochManager {
     epoch_info_aggregator: Option<EpochInfoAggregator>,
     /// Largest final height. Monotonically increasing.
     largest_final_height: BlockHeight,
+    /// Number of additional epochs to keep in memory.
+    pub additional_epochs_to_keep: u64,
```
matklad, line 75: Not a fan of this being a *single* `pub` field. As a rule-of-thumb, types where all fields are private are good -- they are an abstraction boundary. Types where all fields are public are also good -- they are typically just PODs. Types where *some* fields are public are often a smell, as the abstraction is inconsistently enforced. So, adding the first pub field to a type which has a bunch of non-pub fields is not something which should be done lightly, seems better to hide this behind a getter just in case.

chain/epoch_manager/src/lib.rs
```diff
@@ -95,6 +99,7 @@ impl EpochManager {
         genesis_protocol_version: ProtocolVersion,
         reward_calculator: RewardCalculator,
         validators: Vec<ValidatorStake>,
+        additional_epochs_to_keep: u64,
```
matklad, line 102: An allternative to adding an extra parameter to both construtors is to add a single builted-lite method: 

```rust
impl EpochManager {
    pub with_additional_epochs_to_keep(mut self, additional_epochs_to_keep: u64) -> EpochManager {
        self.additional_epochs_to_keep = additional_epochs_to_keep;
        self
    }
}
```

integration-tests/src/tests/client/process_blocks.rs
```diff
@@ -1558,6 +1575,44 @@ fn test_gc_after_state_sync() {
     assert!(env.clients[1].chain.clear_data(tries, 2).is_ok());
 }
 
+/// This tests the `additional_epochs_to_keep` setting, which specifies the number of epochs
+/// to keep.
+///
+/// For this test, we set epoch length to `128`, and `additional_epochs_to_keep` to 10.
```
matklad, line 1581:
```suggestion
/// For this test, we set epoch length to `128`, and `additional_epochs_to_keep` to 5.
```

matklad (approved):

matklad:
FWIW, while reviewing it I realised that I misunderstand what `NightshadeRuntime` is supposed to do, see https://near.zulipchat.com/#narrow/stream/295558-nearinc.2Fprotocol-core/topic/Runtime.20vs.20NightshadeRuntime

chain/epoch_manager/src/lib.rs
pmnoxx: @matklad Yes, I see, we can add it. Though, I think the the design would be signed to maybe only keep it in NightShade.

chain/epoch_manager/src/lib.rs
```diff
@@ -71,12 +71,15 @@ pub struct EpochManager {
     epoch_info_aggregator: Option<EpochInfoAggregator>,
     /// Largest final height. Monotonically increasing.
     largest_final_height: BlockHeight,
+    /// Number of additional epochs to keep in memory.
```
mzhangmzz, line 74: Can you add a comment saying that it is additional to `NUM_EPOCHS_TO_KEEP_STORE_DATA` epochs?

integration-tests/src/tests/client/process_blocks.rs
```diff
@@ -1558,6 +1575,42 @@ fn test_gc_after_state_sync() {
     assert!(env.clients[1].chain.clear_data(tries, 2).is_ok());
 }
 
+/// This tests the `additional_epochs_to_keep` setting, which specifies the number of epochs to keep.
+///
+/// For this test, we set epoch length to `128`, and `additional_epochs_to_keep` to `5`.
+/// This tests shows, we keep `9` full epochs, and one partial epoch with `epoch_length - 100` blocks.
```
mzhangmzz, line 1581: Where does the number 100 come from?

mzhangmzz (approved):

chain/epoch_manager/src/lib.rs
bowenwang1996: Actually, could we allow a lower number? I think it is technically possible to only the past 2 epochs of data

PR #5349, opened 2021-11-18
refactor:   simplify get_gc_stop_height to not use macro

pmnoxx:
Let's simplify `get_gc_stop_height` and not use macro for not reason.

This is just what I saw by randomly looking at the code doing another change.

@matklad What do you think?


matklad (approved):
Clear win, :+1:

pmnoxx:

bowenwang1996 (approved):

PR #5353, opened 2021-11-18
Feature: Improve error output for indexer in case `--download-genesis` flag is missing

pmnoxx:
Indexer doesn't have a good output, in case input `--download-genesis` argument is missing.
Let's improve it.

### DIFF
--- a/chain/indexer/src/lib.rs
+++ b/chain/indexer/src/lib.rs
@@ -2,6 +2,7 @@
 
 use tokio::sync::mpsc;
 
+use anyhow::Result;
 pub use near_primitives;
 use near_primitives::types::Gas;
 pub use nearcore::{get_default_home, init_configs, NearConfig};
@@ -136,7 +137,7 @@ impl Indexer {
 
 /// Function that initializes configs for the node which
 /// accepts `InitConfigWrapper` and calls original `init_configs` from `neard`
-pub fn indexer_init_configs(dir: &std::path::PathBuf, params: InitConfigArgs) {
+pub fn indexer_init_configs(dir: &std::path::PathBuf, params: InitConfigArgs) -> Result<()> {
     init_configs(
         dir,
         params.chain_id.as_deref(),
--- a/core/chain-configs/src/genesis_config.rs
+++ b/core/chain-configs/src/genesis_config.rs
@@ -8,6 +8,7 @@ use std::io::{BufReader, Read};
 use std::path::{Path, PathBuf};
 use std::{fmt, io};
 
+use anyhow::Context;
 use chrono::{DateTime, Utc};
 use num_rational::Rational;
 use serde::de::{self, DeserializeSeed, IgnoredAny, MapAccess, SeqAccess, Visitor};
@@ -297,11 +298,12 @@ impl GenesisConfig {
     ///
     /// It panics if file cannot be open or read, or the contents cannot be parsed from JSON to the
     /// GenesisConfig structure.
-    pub fn from_file<P: AsRef<Path>>(path: P) -> Self {
-        let reader = BufReader::new(File::open(path).expect("Could not open genesis config file."));
-        let genesis_config: GenesisConfig =
-            serde_json::from_reader(reader).expect("Failed to deserialize the genesis records.");
-        genesis_config
+    pub fn from_file<P: AsRef<Path>>(path: P) -> anyhow::Result<Self> {
+        let file = File::open(path).with_context(|| "Could not open genesis config file.")?;
+        let reader = BufReader::new(file);
+        let genesis_config: GenesisConfig = serde_json::from_reader(reader)
+            .with_context(|| "Failed to deserialize the genesis records.")?;
+        Ok(genesis_config)
     }
 
     /// Writes GenesisConfig to the file.
@@ -496,7 +498,7 @@ impl Genesis {
         P1: AsRef<Path>,
         P2: AsRef<Path>,
     {
-        let config = GenesisConfig::from_file(config_path);
+        let config = GenesisConfig::from_file(config_path).unwrap();
         let records = GenesisRecords::from_file(records_path);
         Self::new(config, records)
     }
--- a/nearcore/src/config.rs
+++ b/nearcore/src/config.rs
@@ -5,6 +5,7 @@ use std::path::Path;
 use std::sync::Arc;
 use std::time::Duration;
 
+use anyhow::{anyhow, bail, Context};
 use hyper::body::HttpBody;
 use near_primitives::time::Clock;
 use num_rational::Rational;
@@ -463,12 +464,12 @@ impl Default for Config {
 }
 
 impl Config {
-    pub fn from_file(path: &Path) -> Self {
-        let mut file = File::open(path)
-            .unwrap_or_else(|_| panic!("Could not open config file: `{}`", path.display()));
-        let mut content = String::new();
-        file.read_to_string(&mut content).expect("Could not read from config file.");
-        serde_json::from_str::<Config>(&content).expect("Deserializing config failed")
+    pub fn from_file(path: &Path) -> anyhow::Result<Self> {
+        let s = std::fs::read_to_string(path)
+            .with_context(|| format!("Failed to read config from {}", path.display()))?;
+        let config = serde_json::from_str(&s)
+            .with_context(|| format!("Failed to deserialize config from {}", path.display()))?;
+        Ok(config)
     }
 
     pub fn write_to_file(&self, path: &Path) {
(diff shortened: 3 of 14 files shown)


khorolets (approved):

matklad (approved):
Ideally, code like this should use `anyhow`, but this is already worlds better!

nearcore/src/config.rs
```diff
@@ -837,11 +842,11 @@ pub fn init_configs(
 
     if let Some(url) = download_config_url {
         download_config(&url.to_string(), &dir.join(CONFIG_FILENAME));
-        config = Config::from_file(&dir.join(CONFIG_FILENAME));
+        config = Config::from_file(&dir.join(CONFIG_FILENAME)).unwrap();
```
matklad, line 845: I want to call-out this general pattern of pushing `.unwrap`s to the call site. It's a good approach, we should use it more:

* for the call-site, you still get the convenient "I don't care about failures" API
* but the failure points are more visible, and, if you decide in the future that after all you *do* care about this, refactoring would be easier. 
* implementation actually becomes easier, as you can use `?`, and you don't have to spend mental cycles to decide between `.unwrap` and `?`.

pmnoxx:
> anyhow

@matklad Do you mean, `use anyhow::Result`. ?

frol (approved):

pmnoxx:
@matklad I added usage of `anyhow:Result`. Code looks much better now. I think that now the output looks good. 
I got the indexer to work. I'll work on figuring out where the memory leak is in the indexer.

core/chain-configs/src/genesis_config.rs
```diff
@@ -3,11 +3,13 @@
 //! NOTE: chain-configs is not the best place for `GenesisConfig` since it
 //! contains `RuntimeConfig`, but we keep it here for now until we figure
 //! out the better place.
+use anyhow::Context;
 use std::fs::File;
 use std::io::{BufReader, Read};
 use std::path::{Path, PathBuf};
 use std::{fmt, io};
 
+use anyhow::Result;
```
frol, line 12: To be honest, I don't like overloading commonly used types (Result, Option, etc), so I would consider explicitly stating that the return value is `anyhow::Result<...>`

core/chain-configs/src/genesis_config.rs
```diff
@@ -3,11 +3,13 @@
 //! NOTE: chain-configs is not the best place for `GenesisConfig` since it
 //! contains `RuntimeConfig`, but we keep it here for now until we figure
 //! out the better place.
+use anyhow::Context;
```
frol, line 6: Following the already existing style, this import should be in the block below :neutral_face:

chain/indexer/src/lib.rs
```diff
@@ -2,6 +2,7 @@
 
 use tokio::sync::mpsc;
 
+use anyhow::Result;
```
frol, line 5: Following the existing style, this use statement should belong to the above block.

frol (approved):

core/chain-configs/src/genesis_config.rs
pmnoxx: @matklad What is the recommended codestyle for using `Result`? `anyhow::Result` or `Result`?

pmnoxx: @frol Yes, it feels weird. I tried using `anyhow::Result`, but I reversed it, because I stopped using regular `Result<>` completely. Though, I think `anyhow::Result` should be the default, the std `Result`, is hard to use, so we ended up using `.unwrap()` everything in the code, which is much worse.

chain/indexer/src/lib.rs
pmnoxx: Thanks for noticing, fixed.

tools/indexer/example/src/configs.rs
```diff
@@ -66,7 +66,7 @@ pub(crate) struct InitConfigArgs {
 
 pub(crate) fn init_logging() {
     let env_filter = EnvFilter::new(
-        "indexer-example=info,tokio_reactor=info,near=info,near=error,stats=info,telemetry=info,indexer_example=info,indexer=info,near-performance-metrics=info",
+        "nearcore=info,indexer-example=info,tokio_reactor=info,near=info,near=error,stats=info,telemetry=info,indexer_example=info,indexer=info,near-performance-metrics=info",
```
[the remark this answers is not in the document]
pmnoxx: I could make a separate PR for that, but ok  I'll do it here.

nearcore/src/config.rs
```diff
@@ -822,13 +822,21 @@ pub fn init_configs(
...
-        panic!("Found existing config in {} with chain-id = {}. Use unsafe_reset_all to clear the folder.", dir.display(), genesis_config.chain_id);
+        let config = Config::from_file(&dir.join(CONFIG_FILENAME))
+            .with_context(|| anyhow!("Failed to read config {:?}", &dir))?;
+        let file_path = dir.join(&config.genesis_file);
+        let genesis = GenesisConfig::from_file(&file_path).with_context(move || {
+            anyhow!("Failed to read genesis config {:?}/{:?}", dir, config.genesis_file)
+        })?;
+        return Err(anyhow!("Config is already downloaded: {} with chain-id = {}. Use 'cargo run -p neard -- unsafe_reset_all' to clear the folder.", 
+                file_path.display(), genesis.chain_id),
+        );
```
matklad, line 839:
```suggestion
        bail!("Config is already downloaded: {} with chain-id = {}. Use 'cargo run -p neard -- unsafe_reset_all' to clear the folder.", 
                file_path.display(), genesis.chain_id)
```

nearcore/src/config.rs
```diff
@@ -856,7 +864,7 @@ pub fn init_configs(
     match chain_id.as_ref() {
         "mainnet" => {
             if test_seed.is_some() {
-                panic!("Test seed is not supported for MainNet");
+                return Err(anyhow!("Test seed is not supported for MainNet"));
```
matklad, line 867:
```suggestion
                bail!("Test seed is not supported for MainNet")
```

nearcore/src/config.rs
```diff
@@ -866,16 +874,15 @@ pub fn init_configs(
...
+            let network_signer = InMemorySigner::from_random("node".parse()?, KeyType::ED25519);
             network_signer.write_to_file(&dir.join(config.node_key_file));
 
             genesis.to_file(&dir.join(config.genesis_file));
             info!(target: "near", "Generated MainNet genesis file in {}", dir.display());
         }
         "testnet" | "betanet" => {
             if test_seed.is_some() {
-                panic!("Test seed is not supported for official TestNet");
+                return Err(anyhow!("Test seed is not supported for official TestNet"));
```
matklad, line 885:
```suggestion
                bail!("Test seed is not supported for official TestNet")
```

nearcore/src/config.rs
```diff
@@ -884,23 +891,27 @@ pub fn init_configs(
...
-            let network_signer =
-                InMemorySigner::from_random("node".parse().unwrap(), KeyType::ED25519);
+            let network_signer = InMemorySigner::from_random("node".parse()?, KeyType::ED25519);
             network_signer.write_to_file(&dir.join(config.node_key_file));
 
             // download genesis from s3
             let genesis_path = dir.join("genesis.json");
             let mut genesis_path_str =
-                genesis_path.to_str().expect("Genesis path must be initialized");
+                genesis_path.to_str().with_context(|| "Genesis path must be initialized")?;
```
matklad, line 900: Seems  like we should be able to get rid of this `to_str` logic altogether and keep everything in `Path`.

nearcore/src/config.rs
```diff
@@ -822,13 +822,21 @@ pub fn init_configs(
     download_config_url: Option<&str>,
     boot_nodes: Option<&str>,
     max_gas_burnt_view: Option<Gas>,
-) {
+) -> Result<()> {
+    info!("creating directory at {:?}", dir);
```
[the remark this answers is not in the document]
pmnoxx: @matklad Well, not anymore. Now, we actually print the folder name in case of failure, and that's what it was needed for.

```
    fs::create_dir_all(dir).with_context(|| anyhow!("Failed to create directory {:?}", dir))?;
```

nearcore/src/config.rs
```diff
@@ -822,13 +822,21 @@ pub fn init_configs(
...
+
     if dir.join(CONFIG_FILENAME).exists() {
-        let config = Config::from_file(&dir.join(CONFIG_FILENAME));
-        let genesis_config = GenesisConfig::from_file(&dir.join(config.genesis_file));
-        panic!("Found existing config in {} with chain-id = {}. Use unsafe_reset_all to clear the folder.", dir.display(), genesis_config.chain_id);
+        let config = Config::from_file(&dir.join(CONFIG_FILENAME))
+            .with_context(|| anyhow!("Failed to read config {:?}", &dir))?;
+        let file_path = dir.join(&config.genesis_file);
+        let genesis = GenesisConfig::from_file(&file_path).with_context(move || {
+            anyhow!("Failed to read genesis config {:?}/{:?}", dir, config.genesis_file)
```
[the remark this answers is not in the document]
pmnoxx: @matklad Is usage of `display()` preferred over `{:?}`?

nearcore/src/config.rs
pmnoxx: @matklad Thanks, that's new!

nearcore/src/config.rs
```diff
@@ -463,12 +464,11 @@ impl Default for Config {
...
 impl Config {
-    pub fn from_file(path: &Path) -> Self {
-        let mut file = File::open(path)
-            .unwrap_or_else(|_| panic!("Could not open config file: `{}`", path.display()));
+    pub fn from_file(path: &Path) -> Result<Self> {
+        let mut file = File::open(path)?;
         let mut content = String::new();
-        file.read_to_string(&mut content).expect("Could not read from config file.");
-        Config::from(content.as_str())
+        file.read_to_string(&mut content)?;
```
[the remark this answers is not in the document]
pmnoxx: @matklad Thanks. Yes, we should do that in more places.

pmnoxx:
```
@bowenwang1996 @mm-near: /nearcore/ /neard/
```

mm-near (approved):

PR #5355, opened 2021-11-18
Specify explicitly the WASM features we want to enable for WASM parsing/validation

### LINKED ISSUE #5105: Don't rely on default features when validating wasm

nagisa:
This way we ensure that we don't accidentally allow use of features we
don't intend to allow as part of e.g. bumping a version of a dependency.

Fixes #5105


runtime/near-vm-runner/src/prepare.rs
```diff
@@ -8,14 +8,32 @@ use pwasm_utils::{self, rules};
...
+    bulk_memory: false,
+    module_linking: false,
+    simd: false,
+    threads: false,
+    tail_call: false,
+    deterministic_only: false,
+    multi_memory: false,
+    exceptions: false,
+    memory64: false,
+};
```
matklad, line 24: Nice to have an explicit source of truth here!

runtime/near-vm-runner/src/wasmer2_runner.rs
```diff
@@ -24,6 +25,21 @@ use wasmer_compiler_singlepass::Singlepass;
...
+    simd: WASM_FEATURES.simd,
+    bulk_memory: WASM_FEATURES.bulk_memory,
+    multi_value: WASM_FEATURES.multi_value,
+    tail_call: WASM_FEATURES.tail_call,
+    module_linking: WASM_FEATURES.module_linking,
+    multi_memory: WASM_FEATURES.multi_memory,
+    memory64: WASM_FEATURES.memory64,
+    exceptions: WASM_FEATURES.exceptions,
+    // singlepass does not support signals.
+    signal_less: true,
```
matklad, line 40: unreated to PR: looking at this, I wonder if maybe we should place the `signal_less` toggle into some other config? signals are not a wasm feature, they are an impl detal of our version of wasmer. 

Basically, by default wasmer handles traps by relying on signals, and that was observed to cause troubles (https://github.com/near/nearcore/blob/3ed678195025c643ecb6711c577545bc138b37ea/runtime/near-vm-runner/src/wasmer_runner.rs#L86-L118 beinnd manifestation of that). 

So @olonho changed singlepass to *not* use signals, and that requried a flag to tell wasmer runtime (which is shared between backends) to not install signal handlers.

matklad (approved):

PR #5358, opened 2021-11-18
Conventionalize documentation for runtime entrypoints

nagisa:
Making the documentation more approachable to somebody already used to
reading Rust-style docs.

### DIFF
--- a/runtime/near-vm-runner/src/runner.rs
+++ b/runtime/near-vm-runner/src/runner.rs
@@ -9,18 +9,19 @@ use near_vm_logic::{External, VMContext, VMOutcome};
 use crate::cache::into_vm_result;
 use crate::VMKind;
 
-/// `run` does the following:
-/// - deserializes and validate the `code` binary (see `prepare::prepare_contract`)
-/// - injects gas counting into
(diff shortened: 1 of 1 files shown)


runtime/near-vm-runner/src/runner.rs
```diff
@@ -47,6 +48,21 @@ pub fn run<'a>(
...
+/// linked with the externs specified via the `ext` argument.
+///
+/// [`VMContext::input`] will be passed to the contract entrypoint as an argument.
+///
+/// The gas cost for contract preparation will be subtracted by the VM implementation.
+///
+/// # Panics
+///
+/// Calling this function with a `VMKind` that has not been enabled at the compile time will cause
+/// this function to panic.
```
matklad, line 65: Forgot to submit comment, will expand tommorrow, but TL;DR is that we should say that `run_vm` isn't really a public API, unlike `run`.

matklad: So, yeah, let's change this to something ``Internal variant of [`run_vm`] which allows customizing the wasm VM being used for testing purposes.``

PR #5359, opened 2021-11-18
Adjust the documentation for `External` methods

nagisa:
It seems like some things have changed since the documentation has been
written originally and therefore the documentation no longer reflects
the actual implementation. This changeset attempts to bring the wording
to more closely reflect the actual implementation.


nagisa:
Some other observations: it seems to me that we probably want to adjust the signatures of some of the methods here as well – e.g. `storage_get` is no longer fallible and it returning a `Result` may be misinterpreted by the user of the API as e.g. this function potentially returning an `Err` if the key does not exist.

bowenwang1996 (approved):
Thanks!

PR #5360, opened 2021-11-18
Make non-validators not wait for forwarded chunks

mzhangmzz:
Right now, when a node requests for chunks, these requests won't be sent immediately because of the ChunkForwardParts feature. The feature makes validators forward chunk parts to each other, so validators can wait for some time before they send the requests. However, currently, non-validators also wait, but no one will forward the parts to them, so they are simply waiting in vain. This PR fixes the issue.

Right now, the retry period for chunk requests are 400 ms, meaning a non-validator node will wait for 400 ms before they actually send a chunk request. This PR will remove that delay

Test plan:
- Existing CI and nayduck tests
- Added a new unit test test_chunk_forward_non_validator


chain/chain/src/chain.rs
```diff
@@ -1261,7 +1261,7 @@ impl Chain {
         on_challenge: F3,
     ) where
         F: Copy + FnMut(AcceptedBlock) -> (),
-        F2: Copy + FnMut(Vec<ShardChunkHeader>) -> (),
+        F2: Copy + FnMut((CryptoHash, Vec<ShardChunkHeader>)) -> (),
```
bowenwang1996, line 1264: Might be helpful to create a separate type for more clarity. For example it is not clear to me what this hash is

chain/chunks/src/lib.rs
```diff
@@ -607,11 +607,46 @@ impl ShardsManager {
...
+
+    /// Only marks this chunk as being requested
+    /// Note no requests are actually sent at this point.
+    fn request_chunk_single_mark_only(&mut self, chunk_header: &ShardChunkHeader) {
+        self.request_chunk_single(chunk_header, None, false)
+    }
+
+    /// send partial chunk requests for one chunk
+    /// `chunk_header`: the chunk being requested
+    /// `ancestor_hash`: hash of an ancestor block of the block the chunk belongs to. It must satisfy
```
bowenwang1996, line 638: but it is not an argument to this function...

chain/chunks/src/lib.rs
```diff
@@ -668,16 +701,29 @@ impl ShardsManager {
         }
     }
 
+    /// send chunk requests for some chunks in a block
+    /// `chunks_to_reqeust`: chunks to request
```
bowenwang1996, line 705:
```suggestion
    /// `chunks_to_request`: chunks to request
```

bowenwang1996 (commented):
@mm-near please review this PR

chain/chunks/src/lib.rs
mzhangmzz: whoops this was from the other PR I was working on. Must have been carried over when I was splitting them

chain/chain/src/chain.rs
```diff
@@ -119,6 +119,13 @@ impl BlockLike for Orphan {
     }
 }
 
+/// Contains information for missing chunks in a block
+pub struct BlockMissingChunks {
```
bowenwang1996, line 123: 👍

bowenwang1996 (approved):

PR #5361, opened 2021-11-18
feat: stabilize limiting functions number in a contract

Longarithm:
Stabilize limiting functions number in a contract to 10_000: https://github.com/near/nearcore/pull/4954

While discussing https://github.com/near/nearcore/issues/4826, we found that both contract compilation and function call require a lot of time, if number of wasm functions in it is huge. Even if we fix contract size, function number still makes a big difference.

To mitigate this issue, we limit number of functions in a contract. This shouldn't affect current mainnet contracts because maximal number of functions is ~5k.

## Test plan

* http://nayduck.near.org/#/run/2223
* existing tests


matklad:
Note: we have stabilization template here: https://github.com/near/nearcore/blob/master/.github/PULL_REQUEST_TEMPLATE/feature_stabilization.md

matklad (approved):

bowenwang1996 (approved):

khorolets (approved):

PR #5362, opened 2021-11-18
refactor: Rewrite `Clock` / add tests

pmnoxx:
As @matklad suggested, we will make `reset` and `mock` methods private. I also removed dead code.

Done:
- removed `add_interval` - this function was unused , and it was taking `Duration` as argument.
- removed unused `instant_call_count`
- unused `instant: Instant,`
- renamed `is_mock` to `is_mocked`
- made `reset()`, `mock()` private
- replaced `new` with `#[derive(Default)]` to make code simpler
- added comments
- removed unused `self.durations`
- added tests
- removed ` fn with<F, T>(f: F) -> T` it was a method which had `  static INSTANCE: `  hidden inside. Replaced with direct calls to `INSIDE`. This made it easier to document the code

Blocks https://github.com/near/nearcore/pull/5177


pmnoxx:
@yordanmadzhunkov Can you take a look?

matklad (commented):
LGTM!

Couple of more suggestions:

* we can replace `is_mock: bool,` state flag with storing `Option<MockClockPerThread>`. That way, we statically guarantee that we only have mocking-related state when we do in fact mock the clock
* move `add_utc` and `utc_call_count` methods from `Clock` to `MockClockGuard`. This again statically guarantees that we only can call these methods when mocking is in fact active.

core/primitives/src/time.rs
```diff
@@ -9,62 +9,22 @@ use chrono::DateTime;
...
-            utc: VecDeque::with_capacity(16),
-            durations: VecDeque::with_capacity(16),
-            utc_call_count: 0,
-            instant_call_count: 0,
-            instant: Instant::now(),
-            is_mock: false,
-        }
-    }
+thread_local! {
+    static INSTANCE: RefCell<MockClockPerThread> = RefCell::default()
```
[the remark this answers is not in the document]
pmnoxx: I see your point. In that case let's use `Type::with`. I agree that `thread_local` is a bit weird.

pmnoxx:
@matklad @mm-near Can you take a look again?

matklad (approved):

core/primitives/src/time.rs
```diff
@@ -82,74 +74,107 @@ impl Drop for MockClockGuard {
...
-            if clock.is_mock {
-                clock.durations.push_back(mock_instant);
-            } else {
-                panic!("Use MockClockGuard in your test");
-            }
-        });
+    /// Gets mocked instant.
+    pub fn instant() -> Instant {
+        MockClockPerThread::with(|clock| match &mut clock.mock {
+            Some(_clock) => panic!("Mock clock run out of samples"),
```
mm-near, line 93: I'm confused - shouldn't this say 'not implemented'?

core/primitives/src/time.rs
```diff
@@ -82,74 +74,107 @@ impl Drop for MockClockGuard {
...
+            None => Instant::now(),
+        })
     }
 
+    /// Returns time pushed by `Self::push_utc()`
```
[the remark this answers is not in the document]
mm-near: and actually we should update this comment:

as this method returns current utc (if mocking is disabled) or time pushed by 'add_utc' if mocking is enabled

core/primitives/src/time.rs
```diff
@@ -82,74 +74,107 @@ impl Drop for MockClockGuard {
     }
 }
 
+pub struct Clock {}
```
mm-near, line 77: please comment that the 'Clock' is the class that you're supposed to pass to your 'production' code - and use its methods to fetch time.

core/primitives/src/time.rs
```diff
@@ -82,74 +74,107 @@ impl Drop for MockClockGuard {
...
+            );
+            assert_eq!(
+                Clock::utc(),
+                utc_now
+                    .checked_add_signed(chrono::Duration::from_std(Duration::from_secs(3)).unwrap())
+                    .unwrap(),
+            );
+
+            assert_eq!(mock_clock_guard.utc_call_count(), 3);
+            drop(mock_clock_guard);
```
mm-near, line 175: do you have to call it explicitly?

core/primitives/src/time.rs
```diff
@@ -82,74 +74,107 @@ impl Drop for MockClockGuard {
...
+        }
+        let mock_clock_guard = MockClockGuard::default();
+        assert_eq!(mock_clock_guard.utc_call_count(), 0);
     }
```
mm-near, line 179: can you also add a simple test that checks threading behavior?

mm-near (dismissed):

core/primitives/src/time.rs
```diff
@@ -9,27 +9,21 @@ use chrono::DateTime;
...
 use std::collections::VecDeque;
 
-struct MockClockPerThread {
+#[derive(Default)]
+struct MockClockPerState {
+    /// List of timestamp, we will return one timestamp to each call.
     utc: VecDeque<DateTime<Utc>>,
-    durations: VecDeque<Duration>,
+    /// Number of times `utc()` method was called since we started mocking.
     utc_call_count: u64,
```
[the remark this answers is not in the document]
pmnoxx: Ok, I brought it back.

core/primitives/src/time.rs
pmnoxx: No, this was done, so we can create another mock_clock_guard. You can have only one active at a time.

core/primitives/src/time.rs
pmnoxx: I added a test.

core/primitives/src/time.rs
```diff
@@ -39,36 +33,34 @@ impl MockClockPerThread {
...
-    }
-    fn pop_instant(&mut self) -> Option<Instant> {
-        self.instant_call_count += 1;
-        let x = self.durations.pop_front();
-        match x {
-            Some(t) => self.instant.checked_add(t),
-            None => None,
-        }
+pub struct MockClockGuard {}
+
```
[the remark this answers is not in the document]
pmnoxx: Ok, I added usage example.

core/primitives/src/time.rs
```diff
@@ -1,35 +1,51 @@
...
 use std::collections::VecDeque;
+use std::default::Default;
+pub use std::time::{Duration, Instant};
 
-struct MockClockPerThread {
-    utc: VecDeque<DateTime<Utc>>,
-    durations: VecDeque<Duration>,
+#[derive(Default)]
+struct MockClockPerState {
+    /// List of timestamps, we will return one timestamp to each call.
```
mm-near, line 32: /to each call/ for each call to Clock::utc()

core/primitives/src/time.rs
```diff
@@ -1,35 +1,51 @@
...
+    utc_list: VecDeque<DateTime<Utc>>,
+    /// List of timestamps, we will return one timestamp to each call.
```
mm-near, line 34: for Clock::now

core/primitives/src/time.rs
```diff
@@ -1,35 +1,51 @@
-use std::default::Default;
-
+/// Provides structs used for getting time.
+///
```
mm-near, line 2: I'd also mention that people should be using Clock whenever possible - and write some guideline on when they should use now() vs instant()

mm-near (approved):

core/primitives/src/time.rs
pmnoxx: @mm-near My personal opinion on this is that we should add a large disclaimer, that the usage of `Clock` is incomplete.
And, it should be discouraged. 

```
   let i = Clock::instant();
   /// This get's current time, not mocked time
   i.elapsed().
```

pmnoxx:
```
/// WARNING WARNING WARNING
/// WARNING WARNING WARNING
/// Use at your own risk. The implementation is not complete, we have a places in code not mocked properly.
/// For example, it's possible to call `::elapsed()` on `Instant` returned by `Instant::now()`.
/// We use that that function, throughout the code, and the current mocking of `Instant` is not done properly.
```

yordanmadzhunkov (approved):
Awesome refactoring.

core/primitives/src/time.rs
```diff
@@ -82,74 +123,165 @@ impl Drop for MockClockGuard {
...
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use std::ops::Add;
+    use std::thread;
+    use std::thread::sleep;
+
+    #[test]
+    #[should_panic]
```
yordanmadzhunkov, line 189: I just want to mantion, that this test is awesome :) Good job

pmnoxx:
> Awesome refactoring.

Thanks, for introducing the time mocking into our repo. That helps us a lot!

PR #5363, opened 2021-11-18
fix(near-crypto): get near-crypto compilable on Windows

frol:


bowenwang1996 (approved):

matklad:
Do we actually need to keep stuff compatible with windows? If that's the case, we probably want to check this on CI. 

I am also a bit unsure about "if windows, just don't care if everybody can read the file" logic. I feel that, rather than just not doing that extra security bit, we should do the equivalent on windows (but I am not familiar with the right windows way to do this). But I guess if @bowenwang1996 :heavy_check_mark: than we consider that secure enough?

matklad (approved):

matklad:
googled https://superuser.com/questions/1296024/windows-ssh-permissions-for-private-key-are-too-open, it seems that there is some stuff you are supposed to do on windows.

matklad:
What would be clearly ok is to just mark the whole `write_to_file` function as `#[cfg(unix)]`.

frol:
We use some of the crates in near-cli, near-jsonrpc-client-rs, so at the very least we need to check those. I agree that CI is desirable

bowenwang1996:
> Do we actually need to keep stuff compatible with windows

I think it is aspirational for us at this point

PR #5364, opened 2021-11-18
chore: Reduced @frol's code ownership

frol:
I delegate my owner-reviews to those who are more in touch with the specific subsystems than I do today.


bowenwang1996 (approved):
One day I'll get there as well....

PR #5366, opened 2021-11-18
 refactor: Remove usage of `lazy_static!` in `core` and `once_cell`

pmnoxx:
Let's remove usage of `lazy_static!` for `/core`. This will make it easier to review.

Part of #5321


core/store/src/db.rs
```diff
@@ -193,54 +194,51 @@ impl DBCol {
...
-}
+
+pub static IS_COL_RC: Lazy<Vec<bool>> = Lazy::new(|| {
+    let mut col_rc = vec![false; NUM_COLS];
+    col_rc[DBCol::ColState as usize] = true;
+    col_rc[DBCol::ColTransactions as usize] = true;
+    col_rc[DBCol::ColReceipts as usize] = true;
+    col_rc[DBCol::ColReceiptIdToShardId as usize] = true;
+    col_rc
+});
```
matklad, line 241: I'd maybe go directly to const array -- seems easier to change the code once rather than twice.

pmnoxx: @matklad I don't mind doing it now. 
However, I need to investigate on what's the cleanest way to use const array. I haven't seen them used much in our codebase. If you can provide me an example then I would do it right now.

pmnoxx: @matklad What do you mean by const array, do you mean something like

```
static NUMBERS: &'static [i32] = &[0,0,0,0,0,1,1,1,0,0,1,1,1,1,0,0];
```
I guess you had something else in mind.

pmnoxx: @matklad Any ideas?

matklad: THe following seems to work: 

```rust
pub const SKIP_COL_GC: [bool; NUM_COLS] = {
    let mut col_gc = [false; NUM_COLS];
    // A node may never restarted
    col_gc[DBCol::ColStateHeaders as usize] = true;
    // True until #2515
    col_gc[DBCol::ColStateParts as usize] = true;
    col_gc
};
```

pmnoxx:
@matklad Can you take a look again?

pmnoxx:
@matklad I removed once_cell usage from `store`.

bowenwang1996 (approved):

PR #531, opened 2019-02-07
fix nonsense transaction protobuf serde issue with send money

azban:

PR #5367, opened 2021-11-18
re: Remove usage of `lazy_static!` in `runtime`

pmnoxx:
Let's remove usage of `lazy_static!` in runtime.

Part of #531


pmnoxx:
@mm-near Can you take a look?

runtime/runtime/Cargo.toml
```diff
@@ -13,7 +13,7 @@ serde_json = "1.0"
 log = "0.4"
 tracing = "0.1"
 rand = "0.7"
-lazy_static = "1.4"
+once_cell = "1.5.2"
```
mm-near, line 16: out of curiosity - why did we update ? (just to have the newest version or was there something wrong with the old one?)

runtime/runtime/src/lib.rs
```diff
@@ -439,7 +439,7 @@ impl Runtime {
                 )?;
             }
             Action::DeleteAccount(delete_account) => {
-                near_metrics::inc_counter(&metrics::ACTION_DELETE_ACCOUNT_TOTAL);
+                (metrics::ACTION_DELETE_ACCOUNT_TOTAL.inc());
```
mm-near, line 442: do we need the surrounding brackets?

mm-near (approved):

runtime/runtime/Cargo.toml
pmnoxx: @mm-near No, we are still using the old library. The newest version is 1.8.0

chefsale (approved):

PR #5198, opened 2021-11-10
feat: Add `ActixMessageWrapper` to rate limit `PeerActor`

pmnoxx:
We want to track of sizes of messages sent from `PeerActor` to `PeerManagerActor`.
We will stop reading from given `PeerActor` once we reach the limit.

To achieve this a new type `ActixMessageWrapper` was introduced. It wraps around messages sent to peer, and contains metadata needed for stat counting.

This is a draft:
- [x] Add a wrapper for sending messages from`PeerActor` to `PeerManagerActor`
- [x] Modify `PeerActor` to use 
- [x] Add a wrapper for reply's to the above messages (wating for `DeepSizeOf`)
- [ ] Wait for `DeepSizeOf` (another PR)
- [ ] Add a wrapper for messages to Peer - for example for routing (another PR)
- [ ] Add a wrapper for messages to EdgeVerifierActor - for example for routing (another PR)
- [ ] Add a wrapper for messages to RoutingTableActor - for example for routing (another PR)
- [ ] Add a wrapper for messages to ClientActor - for example for routing (another PR)
- [ ] Extensions to `performance_stats` to track per thread limits (another PR)
- [ ] Extension to track bandwidth (another PR)

This PR is based on, not yet merged, https://github.com/near/nearcore/pull/4927

Resolves https://github.com/near/nearcore/issues/5155

Blocked by https://github.com/Aeledfyr/deepsize/pull/18

PR #5373, opened 2021-11-19
feat: Make `deepsize` import and `DeepSizeOf` derive optional 

pmnoxx:
We would like to use `deepsize` crate and `DeepSizeof` derive in our code. In discussed that it's better not to add extra dependencies to our code base if possible. That's useful for external projects using `core/primitives`.

Here is a proposal on how to include `deepsize`/`DeepSizeOf` inside our repo as optional dependency.

We could do it this the same way for other dependencies like, for example:
```
```

@frol I hope this solves your concerns. @matklad Can you take a look?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1216,6 +1216,27 @@ dependencies = [
 ]
 
 [[package]]
+name = "deepsize"
+version = "0.2.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "1cdb987ec36f6bf7bfbea3f928b75590b736fc42af8e54d97592481351b2b96c"
+dependencies = [
+ "chrono",
+ "deepsize_derive",
+]
+
+[[package]]
+name = "deepsize_derive"
+version = "0.1.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "990101d41f3bc8c1a45641024377ee284ecc338e5ecf3ea0f0e236d897c72796"
+dependencies = [
+ "proc-macro2",
+ "quote",
+ "syn",
+]
+
+[[package]]
 name = "delay-detector"
 version = "0.0.0"
 dependencies = [
@@ -2387,6 +2408,7 @@ name = "near-account-id"
 version = "0.0.0"
 dependencies = [
  "borsh 0.9.1",
+ "deepsize",
  "serde",
  "serde_json",
 ]
@@ -2559,6 +2581,7 @@ dependencies = [
  "bs58",
  "c2-chacha",
  "curve25519-dalek",
+ "deepsize",
  "derive_more",
  "ed25519-dalek",
  "hex-literal",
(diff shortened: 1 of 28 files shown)


matklad:
This is not how such things are usually done, the pattern is 

```rust 
#[cfg_attr(feature = "deepsize", derive(DeepSizeOf))]
struct MyStruct {
}
```

We also should do this only for crates which are intended to be re-usable (I don't know what is that set of crates), all internal-only crates should just derive unconditionally, to *reduce* compile times (if you make something optional, you generally pay for compilation time twice: once with, and once without the feature).

pmnoxx:
@frol @matklad Ok, let's make DeepSize optional only for `*-primites` creates. Other crates will require `DeepSize`.

pmnoxx:
@frol @matklad Any objections?

pmnoxx:
@mm-near @frol @matklad Ok, I made `DeepSizeOf` an optimal reference.
This is a blocking PR for https://github.com/near/nearcore/pull/5198.
Is there anything that prevents us from shipping?

I updated the full change as well: https://github.com/near/nearcore/pull/5191

mm-near (approved):
LGTM from me - but I'd like @matklad / @frol to PTAL

matklad:
LGTM, but I'd prefer to avoid conditional imports in the future 

```
#[cfg(feature = "deepsize")]
use deepsize::DeepSizeOf;
pub use error::{ParseAccountError, ParseErrorKind};

#[cfg_attr(feature = "deepsize", derive(DeepSizeOf))]
#[derive(Eq, Ord, Hash, Clone, Debug, PartialEq, PartialOrd)]
pub struct AccountId(Box<str>);
```

==>>

```
pub use error::{ParseAccountError, ParseErrorKind};

#[cfg_attr(feature = "deepsize", derive(deepsize::DeepSizeOf))]
#[derive(Eq, Ord, Hash, Clone, Debug, PartialEq, PartialOrd)]
pub struct AccountId(Box<str>);
```

pmnoxx:
> LGTM, but I'd prefer to avoid conditional imports in the future
> 
> ```rust
> #[cfg(feature = "deepsize")]
> use deepsize::DeepSizeOf;
> pub use error::{ParseAccountError, ParseErrorKind};
> 
> #[cfg_attr(feature = "deepsize", derive(DeepSizeOf))]
> #[derive(Eq, Ord, Hash, Clone, Debug, PartialEq, PartialOrd)]
> pub struct AccountId(Box<str>);
> ```
> 
> ==>>
> 
> ```rust
> pub use error::{ParseAccountError, ParseErrorKind};
> 
> #[cfg_attr(feature = "deepsize", derive(deepsize::DeepSizeOf))]
> #[derive(Eq, Ord, Hash, Clone, Debug, PartialEq, PartialOrd)]
> pub struct AccountId(Box<str>);
> ```

I added your comment to TODO list: https://github.com/near/nearcore/issues/5423

PR #5374, opened 2021-11-19
refactor: Remove usage of `lazy_static!` in `core/metrics`

pmnoxx:
Remove last usage of `inc_counter_opt`, and delete unused methods / comments from `chain/network/src/stats/metrics.rs`

part of https://github.com/near/nearcore/pull/5321

Edit:
I had to rewrite this change to avoid crashes. In tests we  were initializing `PeerActorManager` multiple times, and each time we would initialize `NetworkMetrics`. Only the first initialization worked, others were silently failing to register metrics in `Prometheus`. `unwrap()` was causing crashes.

- removed ` network_metrics: NetworkMetrics,` from `PeerActor` to fix crashes
- removed ` network_metrics: NetworkMetrics,` from `PeerActorManager` to fix crashes


matklad (approved):

pmnoxx:
@mm-near @matklad I had to rewrite this change to avoid crashes. In tests we  were initializing `PeerActorManager` multiple times. Only the first initialization worked, others were silently, failing to register metrics in `Prometheus.`

matklad (approved):

chain/network/src/peer_manager/peer_manager_actor.rs
```diff
@@ -1309,7 +1305,8 @@ impl PeerManagerActor {
             }
             Err(find_route_error) => {
                 // TODO(MarX, #1369): Message is dropped here. Define policy for this case.
-                self.network_metrics.inc(
+
+                NETWORK_METRICS.inc(
```
mm-near, line 1309: If we do it like this - do we still have a way to checking these values in unittests ? (especially as multiple tests can run in parallel)

pmnoxx: > If we do it like this - do we still have a way to checking these values in unittests ? (especially as multiple tests can run in parallel)

That has been an issue with Prometheus metrics for a while, including other packages. 
We have a few solutions:
- run some tests with `#[serial]`
- rewrite how we use promethius for tests

Note that, those values were simply broken for unit tests, only the first instance would get initialized properly. With this change "nothing gets more broken than it already is".

pmnoxx:
````
```

pmnoxx:
```
```

mm-near (approved):

frol:
@bowenwang1996 @mm-near @matklad Should this have been merged or was this PR superseded?

mina86:
I believe this has been superseeded. We no longer use lazy_static anywhere. Though I dunno what crashes the commit message refers to.

There are also some references to lazy_static in Cargo.toml which were left over. I’ll send PR to clean those up.

mina86:
Though it needs to be said that `NetworkMetrics` should be converted to use `IntCounterVec` with `message` label rather than having separate `IntCounter`s for each message type. That’s unrelated to this PR or cited Issue.

PR #5381, opened 2021-11-19
refactor: Reorganize imports in `chain/client`

pmnoxx:
Fix grouping dependencies in `chain/client`.

### DIFF
--- a/chain/client/src/client_actor.rs
+++ b/chain/client/src/client_actor.rs
@@ -30,9 +30,9 @@ use near_crypto::Signature;
 #[cfg(feature = "sandbox")]
 use near_network::types::SandboxResponse;
 use near_network::types::{
-    NetworkClientMessages, NetworkClientResponses, NetworkRequests, PeerManagerAdapter,
+    NetworkClientMessages, NetworkClientResponses, NetworkInfo, NetworkRequests,
+    PeerManagerAdapter, PeerManagerMessageRequest,
 };
-use near_network::types::{NetworkInfo, PeerManagerMessageRequest};
 #[cfg(feature = "test_features")]
 use near_network_primitives::types::NetworkAdversarialMessage;
 #[cfg(feature = "sandbox")]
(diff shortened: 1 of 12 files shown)


chain/client/src/client_actor.rs
```diff
@@ -9,8 +9,7 @@ use actix::dev::ToEnvelope;
 use actix::{Actor, Addr, Arbiter, AsyncContext, Context, Handler, Message};
 use actix_rt::ArbiterHandle;
 use borsh::BorshSerialize;
-use chrono::DateTime;
-use chrono::Duration as OldDuration;
+use chrono::{DateTime, Duration as OldDuration};
```
matklad, line 12: Unrelated, but I would avoid an alias, and used `chrono::Duration` instead.

matklad (approved):

mzhangmzz (approved):

bowenwang1996 (dismissed):
I suggest we wait until we finalize #5262 and change everything at once to avoid doing it piecemeal and constantly having to fix it

chain/client/src/client_actor.rs
pmnoxx: Fixed, in a different PR.

pmnoxx:
@bowenwang1996 Now that #5262 has been finalize, can we proceed?

mm-near (approved):

bowenwang1996:
Is there a way to enforce the style? Otherwise what prevents us from breaking it from time to time?

pmnoxx:
> Is there a way to enforce the style? Otherwise what prevents us from breaking it from time to time?

In terms of by only using `cargofmt`: No.

However, the most common reason why the code style becomes inconsistent is because someone.

There are some options though:
- We could write a simple language parses, which checks `*.rs` file if there is an empty line, or a few between two consecutive imports.

bowenwang1996 (approved):
I'll approve this PR, but I don't think we should spend our effort on fixing every style inconsistencies in the existing codebase, especially because we *know* that it will be broken again. I think we should try to point out style inconsistencies in new PRs and hopefully over time people will get used to what is written in the style guide.

chefsale (approved):

PR #5382, opened 2021-11-19
refactor: add a trait to abstract over a VM runtime

nagisa:
This is a non-functional-changes-intended refactoring. The primary goal
was to reduce the number of panicky branches in
`near-vm-runner::runner`. This is achieved primarily by two distinct
changes:

The runtime and the implementation of its `run` method has been moved
into the module which implements the runtime itself.

In addition to this, when creating the runtime, users are informed of
the absence of the runtime via an option which they can handle to
produce a more meaningful behaviour (e.g. tests could “simply” ignore
the VM kind) when a runtime is not available.

None of the callers should observe a change in the behaviour, as of this
change, however.

As a future cleanup, the `Runtime`s could start holding within them
certain data that's constant throughout the execution of the daemon,
such as `VMConfig`, `current_protocol_version` or
`CompiledContractCache` to reduce the amount of setup that needs happen
each time a contract execution is attempted.

Supersedes https://github.com/near/nearcore/pull/5358


matklad:
nit: we usually use `refactor:` prefix, as documented in https://github.com/near/nearcore/blob/master/CONTRIBUTING.md#submitting-the-pr

runtime/near-vm-runner/src/lib.rs
```diff
@@ -1,41 +1,34 @@
...
+pub use runner::{run, Runtime};
 
-// These two are public for the standalone runner, but are an implementation
-// detail of `near-vm-runner`. Public API like `run` should not expose VMKind.
-pub use runner::run_vm;
-pub use vm_kind::VMKind;
+// This is public for internal experimentation use only, and should otherwise be considered an
+// implementation detail of `near-vm-runner`.
+#[doc(hidden)]
+pub use vm_kind::VMKind as VMKindPublicForInternalUseOnly;
```
matklad, line 34: I think this is a bit over the top :-) We currently don't promise any kind of API stability, so, practically, I don't think we should go out of our way to discourage the use. 

What about just 

```rust
/// This is public for internal experimentation use only, and should otherwise be considered an
/// implementation detail of `near-vm-runner`.
pub mod internals {
  pub use vm_kind::VMKind;
}
```

?

matklad:
>certain data that's constant throughout the execution of the daemon,
such as VMConfig, current_protocol_version or
CompiledContractCache to reduce the amount of setup that needs happen

To clarify, VMConfig and current_protocol_version might actually change in different versions of `run`.  So, I think the design where public API is  a stateless  `run` method is good. But yeah, internall, it'd be great to bunch up all those things we need into a struct!

runtime/near-vm-runner/src/lib.rs
nagisa: I can do that. My goal here was to document this for myself and the future readers who look at this for the first time pretty much. Up until very recently I thought this was intended to be a public API and either approach helps with understanding this. Though I think there is an opportunity to actually make this private still and depend on `with_vm_variants` to iterate through the available `Runtime` implementations.

runtime/near-vm-runner/src/runner.rs
```diff
@@ -112,86 +44,80 @@ pub fn run_vm(
...
-        }
-        #[cfg(not(feature = "wasmer2_vm"))]
-        VMKind::Wasmer2 => panic!("Wasmer2 is not supported, compile with '--features wasmer2_vm'"),
-        VMKind::Wasmtime => Some(VMError::FunctionCallError(FunctionCallError::CompilationError(
-            CompilationError::UnsupportedCompiler {
-                msg: "Precompilation not supported in Wasmtime yet".to_string(),
-            },
-        ))),
-    }
+pub trait Runtime {
```
matklad, line 53: It's relatively important to pick a different name here, as we already have one `Runtime` in nearcore, and I fear having two would lead to confusion. Two of the possible options:  `Vm` or `ContractRuntime`.

runtime/near-vm-runner/src/wasmer_runner.rs
```diff
@@ -342,15 +266,113 @@ pub(crate) fn run_wasmer0_module<'a>(
...
+        current_protocol_version: ProtocolVersion,
+        cache: Option<&dyn CompiledContractCache>,
+    ) -> (Option<VMOutcome>, Option<VMError>) {
+        let _span = tracing::debug_span!(target: "vm", "run_wasmer").entered();
+
+        if !cfg!(target_arch = "x86") && !cfg!(target_arch = "x86_64") {
+            // TODO(#1940): Remove once NaN is standardized by the VM.
+            panic!(
+            "Execution of smart contracts is only supported for x86 and x86_64 CPU architectures."
+        );
```
matklad, line 295: formatting is off.

runtime/near-vm-runner/src/wasmer_runner.rs
```diff
@@ -342,15 +266,113 @@ pub(crate) fn run_wasmer0_module<'a>(
...
+        method_name: &str,
+        ext: &mut dyn External,
+        context: VMContext,
+        wasm_config: &VMConfig,
+        fees_config: &RuntimeFeesConfig,
+        promise_results: &[PromiseResult],
+        current_protocol_version: ProtocolVersion,
+        cache: Option<&dyn CompiledContractCache>,
+    ) -> (Option<VMOutcome>, Option<VMError>) {
+        let _span = tracing::debug_span!(target: "vm", "run_wasmer").entered();
```
matklad, line 289: Curious: had we used `#[tracing::instrument]` here, would the name be just `run` for all versions of `run`, or would it be unambiguous?

nagisa: `#[tracing::instrument]` only uses the function name, so all versions would be `run`. `#[tracing::instrument]` wouldn't work great here for another reason as well – the arguments to this function don't really represent themselves very nicely as span fields, forcing to `skip` most of them....

runtime/near-vm-runner/src/lib.rs
matklad: Let's go with `internal` module than. The specific think I don't like with the alias is that we rename at every call-site -- that sounds like an implicit rule for consistency, which will be broken very-very soon. :+1: to the overall idea of reflecting public API in code.

>Though I think there is an opportunity to actually make this private still and depend on with_vm_variants to iterate through the available Runtime implementations.

Not sure this'll work, it makes sense that estimator has `--vm-kind` CLI argument: 

https://github.com/near/nearcore/blob/cb20e72afeed2c4d6f6d06e87fd857c5438af8ce/runtime/runtime-params-estimator/src/main.rs#L49-L50

matklad (approved):

runtime/near-vm-runner-standalone/src/script.rs
```diff
@@ -118,22 +119,26 @@ impl Script {
...
-                    step.vm_context.clone(),
-                    &self.vm_config,
-                    runtime_fees_config,
-                    &step.promise_results,
-                    self.vm_kind,
-                    self.protocol_version,
-                    self.contract_cache.as_deref(),
-                );
-                outcomes.push(res);
+        if let Some(runtime) = self.vm_kind.runtime() {
```
EgorKulikov, line 122: Nit: As we go through both branches anyway, maybe use match?

EgorKulikov (approved):
Seems like only thing changed in runtime is import, so LGTM to unblock

runtime/near-vm-runner-standalone/src/script.rs
nagisa: Yeah, the TODO below indicates that I'm not particularly happy with the approach here. Will keep this nit in mind when I'm working on this TODO.

Issue #5328, opened 2021-11-17
Reformat protocol features in Cargo.tomls to be split across lines


matklad:
This list (and many other similar ones) is pretty hard to modify:

https://github.com/near/nearcore/blob/07dadeac0a4db8a1ff448fcf1174e706badf54eb/core/primitives/Cargo.toml#L47

we should split it across several lines.

PR #5383, opened 2021-11-19
Splitting new validator selection algorithm from chunk only producers

EgorKulikov:
In preparation to stabilize later

version.rs is currently not ready to merge as we need a way to build binary with only required features enabled


chain/epoch_manager/src/validator_selection.rs
```diff
@@ -51,11 +51,25 @@ pub fn proposals_to_epoch_info(
...
+        next_version,
+        {
+            let max_cp_selected = max_bp_selected
+                + (epoch_config.validator_selection_config.num_chunk_only_producer_seats as usize);
+            select_chunk_producers(
+                &mut chunk_producer_proposals,
+                max_cp_selected,
+                min_stake_ratio,
+                num_shards,
+            )
```
bowenwang1996, line 66: why was this code not feature gated before?

core/primitives/src/version.rs
```diff
@@ -136,12 +138,24 @@ pub enum ProtocolFeature {
...
 /// Some features (e. g. FixStorageUsage) require that there is at least one epoch with exactly
 /// the corresponding version
-#[cfg(not(feature = "nightly_protocol"))]
+#[cfg(all(
+    not(feature = "nightly_protocol"),
+    any(
+        not(feature = "protocol_feature_new_validator_selection_algorithm"),
+        feature = "protocol_feature_chunk_only_producers"
+    )
+))]
```
bowenwang1996, line 147: why?

bowenwang1996 (commented):
I am confused by the change to protocol version

chain/epoch_manager/src/validator_selection.rs
EgorKulikov: Because whole file was imported behind feature

core/primitives/src/version.rs
EgorKulikov: We need a way to test with just protocol_feature_new_validator_selection_algorithm enabled

EgorKulikov: Also see my comment about that in description

core/primitives/src/version.rs
```diff
@@ -125,6 +125,8 @@ pub enum ProtocolFeature {
     AltBn128,
     #[cfg(feature = "protocol_feature_chunk_only_producers")]
     ChunkOnlyProducers,
+    #[cfg(feature = "protocol_feature_new_validator_selection_algorithm")]
+    NewValidatorSelectionAlgorithm,
```
mm-near, line 129: please add comment describing the feature.

Also - we should come up with a better name - as 'new' will not stand the test of time ;-)

core/primitives/src/version.rs
```diff
@@ -172,7 +186,9 @@ impl ProtocolFeature {
             #[cfg(feature = "protocol_feature_block_header_v3")]
             ProtocolFeature::BlockHeaderV3 => 109,
             #[cfg(feature = "protocol_feature_chunk_only_producers")]
-            ProtocolFeature::ChunkOnlyProducers => 115,
+            ProtocolFeature::ChunkOnlyProducers => 126,
```
mm-near, line 189: can we do that (jumping feature from one protocol version to another?)

or do we have to do a fork of the nightly chain anyways?

core/primitives/src/version.rs
```diff
@@ -136,12 +138,24 @@ pub enum ProtocolFeature {
...
+    )
+))]
 pub const PROTOCOL_VERSION: ProtocolVersion = 49;
 
 /// Current latest nightly version of the protocol.
 #[cfg(feature = "nightly_protocol")]
-pub const PROTOCOL_VERSION: ProtocolVersion = 124;
+pub const PROTOCOL_VERSION: ProtocolVersion = 126;
+
+#[cfg(all(
```
mm-near, line 154: please add a proper comment here -- and mention that this should be removed in January 2022, once new_validator is in mainnet

core/primitives/src/version.rs
```diff
@@ -136,12 +138,24 @@ pub enum ProtocolFeature {
 /// Current latest stable version of the protocol.
 /// Some features (e. g. FixStorageUsage) require that there is at least one epoch with exactly
 /// the corresponding version
-#[cfg(not(feature = "nightly_protocol"))]
+#[cfg(all(
```
mm-near, line 141: please comment that this should be rolled back to 'NOT nightly'  in January 2022

chain/epoch_manager/src/validator_selection.rs
```diff
@@ -8,7 +10,9 @@ use near_primitives::types::{
 };
 use num_rational::Ratio;
 use std::cmp::{self, Ordering};
-use std::collections::{hash_map, BTreeMap, BinaryHeap, HashMap, HashSet};
+#[cfg(feature = "protocol_feature_chunk_only_producers")]
+use std::collections::hash_map;
+use std::collections::{BTreeMap, BinaryHeap, HashMap, HashSet};
 
 pub fn proposals_to_epoch_info(
```
mm-near, line 17: comment what this method does

chain/epoch_manager/src/validator_selection.rs
```diff
@@ -32,10 +36,6 @@ pub fn proposals_to_epoch_info(
         Ratio::new(*rational.numer() as u128, *rational.denom() as u128)
     };
     let max_bp_selected = epoch_config.num_block_producer_seats as usize;
-    let max_cp_selected = max_bp_selected
-        + (epoch_config.validator_selection_config.num_chunk_only_producer_seats as usize);
-    let minimum_validators_per_shard =
-        epoch_config.validator_selection_config.minimum_validators_per_shard as usize;
     let mut stake_change = BTreeMap::new();
     let mut fishermen = vec![];
     let proposals = proposals_with_rollover(
```
mm-near, line 41: what does 'proposals_with_rollover' mean? (please comment)

chain/epoch_manager/src/validator_selection.rs
```diff
@@ -51,11 +51,25 @@ pub fn proposals_to_epoch_info(
...
+                + (epoch_config.validator_selection_config.num_chunk_only_producer_seats as usize);
+            select_chunk_producers(
+                &mut chunk_producer_proposals,
+                max_cp_selected,
+                min_stake_ratio,
+                num_shards,
+            )
+        },
+        {
+            let _ = chunk_producer_proposals;
```
mm-near, line 69: if we don't use chunk_producer_proposals -> maybe we should move the line 53 into 59? (so that order_proposals is computed only when we do the chunk_only producers?

chain/epoch_manager/src/validator_selection.rs
```diff
@@ -97,39 +111,54 @@ pub fn proposals_to_epoch_info(
...
+                            let validator_id = *entry.get();
+                            shard_validator_ids.push(validator_id);
+                        }
+                    }
                 }
             }
+            chunk_producers_settlement
+        },
+        {
+            if chunk_producers.is_empty() {
```
mm-near, line 156: can you explain when this could happen?

mm-near (changes requested):

nearcore/src/lib.rs
```diff
@@ -252,7 +252,7 @@ pub fn apply_store_migrations(path: &Path, near_config: &NearConfig) {
         info!(target: "near", "Migrate DB from version 28 to 29");
         migrate_28_to_29(&path);
     }
-    #[cfg(feature = "nightly_protocol")]
+    #[cfg(any(feature = "nightly_protocol", feature = "protocol_feature_block_header_v3"))]
```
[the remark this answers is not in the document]
EgorKulikov: yes, I need this temporary to create binary with only this feature enabled. This would be reverted before merging

integration-tests/tests/network/runner.rs
```diff
@@ -714,7 +714,7 @@ impl Actor for Runner {
                 }
             }),
             1,
-            15000,
+            20000,
```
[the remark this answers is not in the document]
EgorKulikov: This was just added temporary, will revert

core/primitives/src/version.rs
EgorKulikov: We fork betanet with each release

core/primitives/src/version.rs
EgorKulikov: As mentioned in PR description, this would be removed before merge

core/primitives/src/version.rs
EgorKulikov: As mentioned in PR description, this would be removed before merge

chain/chain/Cargo.toml
```diff
@@ -44,8 +44,10 @@ test_features = []
 delay_detector = ["delay-detector"]
 no_cache = ["near-store/no_cache"]
 protocol_feature_block_header_v3 = []
-protocol_feature_chunk_only_producers = ["protocol_feature_block_header_v3", "near-primitives/protocol_feature_chunk_only_producers"]
+protocol_feature_chunk_only_producers = ["near-primitives/protocol_feature_chunk_only_producers", "protocol_feature_new_validator_selection_algorithm"]
```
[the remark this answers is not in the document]
bowenwang1996: @matklad already created this issue https://github.com/near/nearcore/issues/5328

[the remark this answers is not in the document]
EgorKulikov: I am not sure multi-line vs single-line is Marcin's concern, and yes, I agree that our feature handling is getting kinda out of hand, but I do not see better solution

EgorKulikov:
Nayduck run: http://nayduck.near.org/#/run/2227

chain/chain/Cargo.toml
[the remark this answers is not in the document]
nagisa: Alas, no, create features are predominantly associated with a single crate.

In workspaces dependent crates specifying different requested feature sets will cause feature unification to happen (i.e. if crate `a` depends on `c[feat1]` and crate `b` depends on `c[feat2]`, building the entire workspace will build `c[feat1, feat2]`, but building `a` and `b` crates separately, via `cargo build -p a/b` will still build just `c[feat1]` or `c[feat2]`)

That said, I'm not entirely sure if cargo features are the most appropriate mechanism for managing these… features. Some of these seem like they should be using a runtime mechanism instead, but that is probably way out of scope for this specific PR.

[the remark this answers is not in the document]
matklad: The relevant discussion is https://github.com/near/nearcore/discussions/4067.

nearcore/src/runtime/mod.rs
```diff
@@ -2728,11 +2766,7 @@ mod test {
             env.step(vec![vec![], vec![]], vec![true, true], ChallengesResult::default());
         }
         // which validator is kicked out depends on which validator selection is used
```
mm-near, line 2768: fix the comment

chain/epoch_manager/src/validator_selection.rs
```diff
@@ -156,6 +197,8 @@ pub fn proposals_to_epoch_info(
     ))
 }
 
+/// Generates proposals based on new proposals, last epoch validators/fishermen and validator
+/// kickouts
```
mm-near, line 201: from what I see it does a lot more (and we should consider splitting this function into separate ones in the future).

it:
- looks through new proposals and reject the ones that should be kicked out
- gives rewards to validators from the previous epoch (and adds them to the validator list of the next epoch)
- it also 'unstakes' all the fishermen that were present in the last epoch

chain/epoch_manager/src/validator_selection.rs
```diff
@@ -97,39 +110,67 @@ pub fn proposals_to_epoch_info(
...
                 }
             }
+            chunk_producers_settlement
+        },
+        {
+            if chunk_producers.is_empty() {
+                // All validators tried to unstake?
+                return Err(EpochError::NotEnoughValidators { num_validators: 0u64, num_shards });
+            }
+            let mut id = 0usize;
```
mm-near, line 159: I'd add a comment that old algorithm was doing a simple round-robin over the block_producers

chain/epoch_manager/src/validator_selection.rs
```diff
@@ -97,39 +110,67 @@ pub fn proposals_to_epoch_info(
...
+                assign_shards(chunk_producers, num_shards as usize, minimum_validators_per_shard)
+                    .map_err(|_| EpochError::NotEnoughValidators {
+                    num_validators: num_chunk_producers as u64,
+                    num_shards,
+                })?;
+
+            let mut chunk_producers_settlement: Vec<Vec<ValidatorId>> =
+                shard_assignment.iter().map(|vs| Vec::with_capacity(vs.len())).collect();
+            let mut i = all_validators.len();
+            for (shard_validators, shard_validator_ids) in
```
mm-near, line 130: can you explain what this for loop does? (in the comment)

as it seems that we're also adding more 'validators' into the all_validators and validator_to_index

mm-near (approved):

nearcore/src/runtime/mod.rs
```diff
@@ -3262,11 +3318,12 @@ mod test {
...
             .map(|id| InMemorySigner::from_seed(id.clone(), KeyType::ED25519, id.as_ref()))
             .collect();
-        let fishermen_stake = if cfg!(feature = "protocol_feature_chunk_only_producers") {
-            3200 * crate::NEAR_BASE + 1
-        } else {
-            TESTING_INIT_STAKE / 10 + 1
-        };
+        let fishermen_stake =
+            if cfg!(feature = "protocol_feature_new_validator_selection_algorithm") {
+                3200 * crate::NEAR_BASE + 1
```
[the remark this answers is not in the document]
EgorKulikov: Enough to be over fisherman threshold, not enough to be in as validator

nearcore/src/runtime/mod.rs
EgorKulikov: This comment was supposed to be in completely another place oO

core/primitives/src/version.rs
EgorKulikov: As we are ready for merge I reverted this part

core/primitives/src/version.rs
EgorKulikov: As we are ready for merge I reverted this part

nearcore/src/lib.rs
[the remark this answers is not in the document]
EgorKulikov: As we are ready for merge I reverted this part

chain/epoch_manager/src/validator_selection.rs
```diff
@@ -97,39 +110,72 @@ pub fn proposals_to_epoch_info(
...
+        {
+            if chunk_producers.is_empty() {
+                // All validators tried to unstake?
+                return Err(EpochError::NotEnoughValidators { num_validators: 0u64, num_shards });
+            }
+            let mut id = 0usize;
+            // Here we assign validators to chunks (we try to keep number of shards assigned for
+            // each validator as even as possible). Note that in prod configuration number of seats
+            // per shard is the same as maximal number of block producers, so normally all
+            // validators would be assigned to all chats
```
mm-near, line 164: s/chats/chunks

mm-near (approved):

PR #5384, opened 2021-11-19
Bump Wasmer2 VM configuration version

nagisa:
With the change of the gas measurement implementation we want to also
evict all the old cached contracts. Bumping the version here does so.

### DIFF
--- a/runtime/near-vm-runner/src/wasmer2_runner.rs
+++ b/runtime/near-vm-runner/src/wasmer2_runner.rs
@@ -355,7 +355,7 @@ enum WasmerCompiler {
 
 #[derive(Hash)]
 struct Wasmer2Config {
-    seed: i32,
+    seed: u32,
     engine: WasmerEngine,
     compiler: WasmerCompiler,
 }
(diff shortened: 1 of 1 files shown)


runtime/near-vm-runner/src/wasmer2_runner.rs
```diff
@@ -369,11 +369,11 @@ impl Wasmer2Config {
 }
 
 // We use following scheme for the bits forming seed:
-//  kind << 10, kind is 1 for Wasmer
+//  kind << 10, kind is 1 for Wasmer2
 //  major version << 6
 //  minor version
 const WASMER2_CONFIG: Wasmer2Config = Wasmer2Config {
-    seed: (1 << 10) | (2 << 6) | 0,
+    seed: (1 << 10) | (3 << 6) | 0,
```
nagisa, line 376: I'm not entirely sure if we want to bump `major` version here or the minor one. Either way this does not actually reflect the version of the VM implementation? so to me it seems like we want to just have a plain counter here just like for Wasmer0.

matklad:
Context: we should have done this in https://github.com/near/nearcore/pull/5200, to poison the caches, as we did change codgen. Interestingly, old artifacts are still compatible, they just don't get perf improvement. 

cc @bowenwang1996 similar to the latest release, this release will require all contracts to be recompiled.

matklad (approved):

runtime/near-vm-runner/src/wasmer2_runner.rs
matklad: Yeah, it's not super-clear what versioning scheme we should use here. Right now we deliberatlly don't use semver for our version of wasmer, and instead use `=` version in Cargo.toml. I guess, ideally we should mix some constant defined in the wasmer crate itself?

PR #5390, opened 2021-11-19
refactor: Add debug logs to runner.rs

pmnoxx:
Currently it's hard to debug why tests in `runner.rs` are failing. Let's print the command, so we see which one failed.

### DIFF
--- a/integration-tests/tests/network/runner.rs
+++ b/integration-tests/tests/network/runner.rs
@@ -162,6 +162,7 @@ impl StateMachine {
     pub fn push(&mut self, action: Action) {
         let num_prev_actions = self.actions.len();
         let action_clone = action.clone();
+        let can_write_log = Arc::new(AtomicBool::new(false));
         match action {
             #[cfg(feature = "test_features")]
             Action::SetOptions { target, max_num_peers } => {
@@ -170,7 +171,9 @@ impl StateMachine {
                           flag: Arc<AtomicBool>,
                           _ctx: &mut Context<WaitOrTimeoutActor>,
                           _runner| {
-                        debug!(target: "network", "runner.rs: Action {:?}", action_clone);
+                        if can_write_log.swap(false, Ordering::Relaxed) == true {
+                            debug!(target: "network", "runner.rs: Action {} {:?}", num_prev_actions, action_clone);
+                        }
                         let addr = info.read().unwrap().pm_addr[target].clone();
                         actix::spawn(
                             addr.send(PeerManagerMessageRequest::SetAdvOptions(SetAdvOptions {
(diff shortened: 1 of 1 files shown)


integration-tests/tests/network/runner.rs
```diff
@@ -861,6 +877,7 @@ pub fn ban_peer(target_peer: usize, banned_peer: usize) -> ActionFn {
               flag: Arc<AtomicBool>,
               _ctx: &mut Context<WaitOrTimeoutActor>,
               _runner| {
+            debug!(target: "network", "runner.rs ban_peer {:?}", (target_peer, banned_peer));
```
nagisa, line 880: I would recommend using the `tracing`'s built-in ability to output key-values, rather than manually formatting in the variables into the message. For example in this case I'd write

```suggestion
            debug!(target: "network", "runner.rs ban_peer", ?target_peer, ?banned_peer);
```

which will use `Debug` representation of `target_peer` and `banned_peer` separately.

pmnoxx: @nagisa I tried it, but it doesn't work. Please take a look at: https://github.com/near/nearcore/pull/5392

I haven't seen this syntax used in our code base. What is the name of this feature, is it a nightly feature?

Anyway, introducing new systax should be done in separate PR. I'm happy to do it. Let's try it out.

pmnoxx: @nagisa Oh, we are missing "message =" keyword

integration-tests/tests/network/runner.rs
```diff
@@ -374,7 +382,7 @@ impl StateMachine {
                                 (nonce, info.read().unwrap().peers_info[source].id.clone(), count)
                             })
                             .collect();
-
+let action_clone2 = action_clone.clone();
```
mm-near, line 385: is it related to this PR?

integration-tests/tests/network/runner.rs
```diff
@@ -911,6 +927,7 @@ where
               flag: Arc<AtomicBool>,
               _ctx: &mut Context<WaitOrTimeoutActor>,
               _runner: Addr<Runner>| {
+            debug!(target: "network", "runner.rs wait_for predicate");
```
mm-near, line 930: (general comment): let's make sure that these debug logs are not too spammy - otherwise we risk that one log can completely overwhelm the whole log-file

mm-near (approved):

integration-tests/tests/network/runner.rs
pmnoxx: This was used to spawn another future. This is not needed anymore.

pmnoxx:
@matklad Can you take a look?

PR #5392, opened 2021-11-19
refactor: use `debug` build-in ability to display keys

pmnoxx:
We can make improve readability of our code by using debug build-in ability to display keys.
This was suggested as improvement by @nagisa. Thanks!

https://github.com/near/nearcore/pull/5392/files#diff-0f91095b7bd60265119b46285ed31744a220414708655a2433990d1664ca2f0aR174-R176

TODO:
- [x] Find documentation of this feature. https://docs.rs/tracing/0.1.29/tracing/macro.debug.html
- [x] Read how feature works in docs.


pmnoxx:
@nagisa 

```
error: expected expression, found `?`
   --> integration-tests/tests/network/runner.rs:818:76
    |
818 |             debug!(target: "network", "runner.rs check_direct_connection", ?node_id, ?target_id);
    |                                                                            ^ expected expression

error: expected expression, found `?`
   --> integration-tests/tests/network/runner.rs:856:60
    |
856 |             debug!(target: "network", "runner.rs restart", ?node_id);
    |                                                            ^ expected expression

error: expected expression, found `?`
   --> integration-tests/tests/network/runner.rs:878:61
    |
878 |             debug!(target: "network", "runner.rs ban_peer", ?target_peer, ?banned_peer);
    |                                                             ^ expected expression

error: expected expression, found `?`
   --> integration-tests/tests/network/runner.rs:905:70
    |
905 |             debug!(target: "network", "runner.rs change_account_id", ?node_id, ?account_id.clone());

```

I tried it, but didn't work.

nagisa:
Ah, the reason this doesn't work is because the message needs to go last.

The syntax is part of the `tracing`'s macros and is documented [here](https://docs.rs/tracing/0.1.29/tracing/#recording-fields).

You will need to do either:

```
debug!(target: "network", message = "runner.rs check_direct_connection", ?node_id, ?target_id);
```

or 

```
debug!(target: "network", ?node_id, ?target_id, "runner.rs check_direct_connection");
```

A [playground with some examples](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=090a7f90ee411a6e77a5d70e0a28c0b4).

pmnoxx:
> message =



> Ah, the reason this doesn't work is because the message needs to go last.
> 
> The syntax is part of the `tracing`'s macros and is documented [here](https://docs.rs/tracing/0.1.29/tracing/#recording-fields).
> 
> You will need to write either:
> 
> ```
> debug!(target: "network", message = "runner.rs check_direct_connection", ?node_id, ?target_id);
> ```
> 
> or
> 
> ```
> debug!(target: "network", ?node_id, ?target_id, "runner.rs check_direct_connection");
> ```
> 
> A [playground with some examples](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=090a7f90ee411a6e77a5d70e0a28c0b4).

Thanks, we were missing `message =` keyword. Ok, once https://github.com/near/nearcore/pull/5390
I'll update this PR, and we can start using it in our code base. Thanks.

pmnoxx:
@matklad Can you take a looki?

integration-tests/tests/network/runner.rs
```diff
@@ -172,7 +172,7 @@ impl StateMachine {
                           _ctx: &mut Context<WaitOrTimeoutActor>,
                           _runner| {
                         if can_write_log.swap(false, Ordering::Relaxed) == true {
-                            debug!(target: "network", "runner.rs: Action {} {:?}", num_prev_actions, action_clone);
+                            debug!(target: "network", message =  "runner.rs: Action {} {}", num_prev_actions, ?action_clone);
```
nagisa, line 175: Note: the `{}` formatting specifies in the message are no longer necessary in invocations like these. When using the macro invocation with the old style, `tracing` will actually format in the fields into the message producing a log event like this (JSONish format):

```json
{ "target": "network", "message": "runner.rs: Action 42 Action { ... }" }
```

whereas with this new approach these values instead become keys on their own, and so `{}` no longer does anything (and is a part of the message string), resulting in an event like this instead:

```json
{ "target": "network", "message": "runner.rs: Action {} {}", "num_prev_actions": 42, "action_clone": "Action { ... }" }
```

integration-tests/tests/network/runner.rs
```diff
@@ -843,7 +843,7 @@ pub fn check_direct_connection(node_id: usize, target_id: usize) -> ActionFn {
             let info = info.read().unwrap();
             let target_peer_id = info.peers_info[target_id].id.clone();
             if can_write_log.swap(false, Ordering::Relaxed) == true {
-                debug!(target: "network", "runner.rs check_direct_connection {:?}", (node_id, target_id));
+                debug!(target: "network", message="runner.rs check_direct_connection {}, {}", node_id, ?target_id);
```
nagisa, line 846: nit: style around the equals sign.

integration-tests/tests/network/runner.rs
pmnoxx: @nagisa Thanks, fixed.

pmnoxx:
@mm-near @nagisa @matklad @bowenwang1996 I'll wait with merge until we decide whenever to use this code style in our code or not. If so we need to commit, to using it in other places as well.

bowenwang1996 (approved):

pmnoxx:
@nagisa Thanks, I added it to list of TODO refactoring. https://github.com/near/nearcore/issues/5423

Issue #5389, opened 2021-11-19
Random test failure `routing::archival_node`


pmnoxx:
We have seen a random test failure for `routing::archival_node`

https://buildkite.com/nearprotocol/nearcore/builds/7640#d7e5d254-c79d-443b-9a07-58df975f5130

Plan:
- [ ] Add debug logs to `runner.rs` https://github.com/near/nearcore/pull/5390
- [ ] Add fix https://github.com/near/nearcore/pull/5639

pmnoxx:
```


[2021-11-19T18:02:42Z] test routing::archival_node ... FAILED
--
  | [2021-11-19T18:02:42Z]
  | [2021-11-19T18:02:42Z] failures:
  | [2021-11-19T18:02:42Z]
  | [2021-11-19T18:02:42Z] ---- routing::archival_node stdout ----
  | [2021-11-19T18:02:42Z] BBBB Slept 15001; max_wait_ms 15000
  | [2021-11-19T18:02:42Z] thread 'routing::archival_node' panicked at 'Timed out waiting for the condition', chain/network/src/test_utils.rs:159:21
  | [2021-11-19T18:02:42Z] stack backtrace:
  | [2021-11-19T18:02:42Z]    0: rust_begin_unwind
  | [2021-11-19T18:02:42Z]              at /rustc/09c42c45858d5f3aedfa670698275303a3d19afa/library/std/src/panicking.rs:517:5
  | [2021-11-19T18:02:42Z]    1: core::panicking::panic_fmt
  | [2021-11-19T18:02:42Z]              at /rustc/09c42c45858d5f3aedfa670698275303a3d19afa/library/core/src/panicking.rs:101:14
  | [2021-11-19T18:02:42Z]    2: near_network::test_utils::WaitOrTimeoutActor::wait_or_timeout::{{closure}}
  | [2021-11-19T18:02:42Z]              at /var/lib/buildkite-agent/builds/buildkite-i-0cab78009f71c830c-1/nearprotocol/nearcore/chain/network/src/test_utils.rs:159:21
  | [2021-11-19T18:02:42Z]    3: <F as actix::utils::TimerFuncBox<A>>::call
  | [2021-11-19T18:02:42Z]              at /var/lib/buildkite-agent/.cargo/registry/src/github.com-1ecc6299db9ec823/actix-0.11.0-beta.2/src/utils.rs:117:9
[...]

PR #5399, opened 2021-11-20
bug: ignore `routing::archival_node()` test

pmnoxx:
Let's ignore this test for now. I'll work on it next week, but I don't want people from merging.

Temporary fix for #5389


pmnoxx:
@mm-near Can you take a look? This may block other people from merging.

mm-near (approved):

pmnoxx:
@mm-near Thanks!. This should unblock the production.

PR #5403, opened 2021-11-21
bug: fix sync errors in `Clion`

pmnoxx:
Clion prints lines about missing `from_bytes` due to the way we import `Signature`, This should fix it.

In this file we use ed25519_dalek::ed25519::signature::Signature struct method. Apparently, require you to import a type, if you use it's method with iit's trait. Even though it's name is not used directly. I'm sure @matklad can explain why it's done like this. Although we need to import it, it doesn't name which name we use for importing it.

So I can't just use alias Signature due to name collision, because there is already another Signature struct it in file,

As to why I'm renaming, _ to _Signature. The checked that's used in Clion has an issue, it complains about another error where there are multiple imports as _. It's a bug in that checked.

We could get rid of 'as _Signature' if we could rename the other struct from Signature to something else. Then importing signature::Signature as Signature would work.

@matklad It may be good to report to people who wrote it.


core/crypto/src/signature.rs
```diff
@@ -6,7 +6,9 @@ use std::io::{Error, ErrorKind, Write};
 use std::str::FromStr;
 
 use borsh::{BorshDeserialize, BorshSerialize};
-use ed25519_dalek::ed25519::signature::{Signature as _, Signer as _, Verifier as _};
+use ed25519_dalek::ed25519::signature::{
+    Signature as _Signature, Signer as _Signer, Verifier as _Verifier,
+};
```
mina86, line 11: Can’t we just drop the use? What is it supposed to do anyway?

pmnoxx: @mina86 That's an excellent question! 

Short answer: we can't

Long answer:
In this file we use ed25519_dalek::ed25519::signature::Signature struct method. Apparently,  require you to import a type, if you use struct's methods. Even though it's name is not used directly. I'm sure @matklad can  explain why it's done like this. Although we need to import it, it doesn't name which name we use for importing it.

So I can't just use alias `Signature` due to name collision, because there is already another `Signature` struct it in file,

As to why I'm renaming, `_` to `_Signature`. The checked that's used in `Clion` has an issue, it complains about another error where there are multiple imports `as _`. It's a bug in that checked.

@matklad It may be good to report to people who wrote it.

mina86: I see now. Those are traits in `signature` rather than structs in `ed25519`.

mina86 (approved):
Not like I’m a code owner able to approve this merge, but LGTM. ;)

core/crypto/src/signature.rs
mm-near: @pmnoxx - please put some of this explanation in the code comment.

mm-near (approved):

core/crypto/src/signature.rs
pmnoxx: @mm-near Thanks, I added a comment.

matklad: what a delightful line: it breaks both IntelliJ Rust and rust-analyzer, and in completely different ways:

* https://github.com/intellij-rust/intellij-rust/issues/8143
* https://github.com/rust-analyzer/rust-analyzer/issues/10836

PR #5407, opened 2021-11-21
Document `near-network::peer::code::Codec`  and make it's constr default

pmnoxx:
In an ongoing effort to improve code readability, I propose we
- Use `default` rather than `new` for `Coded`
- remove `max_length` from attribute list of `Codec`.  We don't use/plan to modify it through constructor. If we ever do, well, we will refactor the code, but that's unlikely.
- switch constant `MAX_CAPACITY` to proper type. It defines used to define capacity of a buffer, which is returned as `usize` - we avoid extra casting from u64 to usize.
- Add comments to constants.

### DIFF
--- a/chain/network/src/peer/codec.rs
+++ b/chain/network/src/peer/codec.rs
@@ -1,3 +1,10 @@
+/// The purpose of this crate is to encode/decode messages on the network layer.
+/// Each message contains:
+///     - 4 bytes - length of the message as u32
+///     - the message itself, which is encoded with `borsh`
+///
+/// NOTES:
+///     - Code has an extra logic to ban peers if they sent messages that are too large.
 use crate::stats::metrics;
 use bytes::{Buf, BufMut, BytesMut};
 use bytesize::{GIB, MIB};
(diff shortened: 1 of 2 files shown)


chain/network/src/peer/codec.rs
```diff
@@ -15,19 +15,14 @@ use near_rust_allocator_proxy::allocator::get_tid;
 use crate::stats::metrics;
 use crate::types::PeerMessage;
```
mm-near, line 17: can you also add a small 'file-level' comment on what this codec does?

From what I see, it is responsible for reading & writing the 'network' messages from the stream (encoded as 
- 4 bytes: length of the message
-  message itself

And maybe also add the comment to is_forward_tx -> that AFAIK is trying to figure out if transaction is forwarded without borsch-deserializing the whole context.

(btw - we should have a unittest in this file that makes sure that this function keeps working - as it looks very brittle and easily breakable)

mm-near (approved):

chain/network/src/peer/codec.rs
pmnoxx: There are units tests for `Codec` in this file. Later, we should review if any cases are missing.

pmnoxx: @mm-near Thanks, I added documentation.

pmnoxx: I added documentation about `is_forward_tx` to https://github.com/near/nearcore/pull/5409/files

PR #5409, opened 2021-11-21
Figure out what does `near-network::...::codec::is_forward_tx` do

pmnoxx:
Let's add comments to what `is_forward_tx` does.
Also refactor the code to be more readable.

### DIFF
--- a/chain/network/src/peer/codec.rs
+++ b/chain/network/src/peer/codec.rs
@@ -107,44 +107,63 @@ fn peer_id_type_field_len(enum_var: u8) -> Option<usize> {
     }
 }
 
+/// Checks `bytes` represents `PeerMessage::Routed(RoutedMessage)`,
+/// and `RoutedMessage.body` has type of `RoutedMessageBody::ForwardTx`.
+/// This is done manually as a hack, to avoid having to run expensive `PeerMessage::try_from_slice(bytes)`.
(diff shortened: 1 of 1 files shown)


matklad (approved):
LGTM!

Couple of suggestions:

* we might explicitly write the slow variant this is equivalen to: `matches! (PeerMessage::try_from_slice(bytes), Ok(PeerMessage::Routed(RoutedMessage { body: RoutedMessageBody::ForwardTx(_),..})))`
* for parsing code like this, it's usfelt to see not so much a comment for each specific it, as an example of what we are parsing: something like `tag1|PeerIdOrHash|PeerId|ttl:u8|tag2`.

pmnoxx:
> LGTM!
> 
> Couple of suggestions:
> 
> * we might explicitly write the slow variant this is equivalen to: `matches! (PeerMessage::try_from_slice(bytes), Ok(PeerMessage::Routed(RoutedMessage { body: RoutedMessageBody::ForwardTx(_),..})))`

The purpose of this method is to filter out message without decoding for performance reasons.
Otherwise we wouldn't need this method at all.

> * for parsing code like this, it's usfelt to see not so much a comment for each specific it, as an example of what we are parsing: something like `tag1|PeerIdOrHash|PeerId|ttl:u8|tag2`.

I'll  make comments shorter.

mm-near (approved):
Thanks for adding the comments !

PR #5410, opened 2021-11-21
re: First part of splitting 500 lines long method in `near-network::PeerActor`

pmnoxx:
There is a 500 lines `handle` method in `PeerActor`. We should add comments to it and try to split it.

Reasons:
- Currently too much is going on in this method. It will be much easier to read the code, by splitting parts of it into independent functions. This has to be done in parts. Analyzing what each part does it time consuming.
- As a good rule of thumb: method names alone should be descriptive of what the method does.
`handle` is too generic.
- We have a lot of weird logic here. It needs to be double checked at documented.

### DIFF
--- a/chain/network/src/peer/peer_actor.rs
+++ b/chain/network/src/peer/peer_actor.rs
@@ -28,6 +28,7 @@ use near_performance_metrics;
 use near_performance_metrics::framed_write::{FramedWrite, WriteHandler};
 use near_performance_metrics_macros::perf;
 use near_primitives::block::GenesisId;
+use near_primitives::borsh::maybestd::io::Error;
 use near_primitives::network::PeerId;
 use near_primitives::sharding::PartialEncodedChunk;
 use near_primitives::time::Clock;
@@ -56,7 +57,7 @@ const MAX_PEER_MSG_PER_MIN: u64 = u64::MAX;
 /// Maximum number of transaction messages we will accept between block messages.
 /// The purpose of this constant is to ensure we do not spend too much time deserializing and
 /// dispatching transactions when we should be focusing on consensus-related messages.
-const MAX_TXNS_PER_BLOCK_MESSAGE: usize = 1000;
+const MAX_TRANSACTIONS_PER_BLOCK_MESSAGE: usize = 1000;
 /// Limit cache size of 1000 messages
 pub const ROUTED_MESSAGE_CACHE_SIZE: usize = 1000;
 /// Duplicated messages will be dropped if routed through the same peer multiple times.
@@ -98,6 +99,7 @@ pub struct PeerActor {
     /// Dynamic Prometheus metrics
     network_metrics: NetworkMetrics,
     /// How many transactions we have received since the last block message
+    /// Note: Shared between multiple Peers.
     txns_since_last_block: Arc<AtomicUsize>,
     /// How many peer actors are created
     peer_counter: Arc<AtomicUsize>,
@@ -543,6 +545,52 @@ impl PeerActor {
             }
         }
     }
+
+    /// Update stats when receiving msg
+    fn update_stats_on_receiving_message(&mut self, msg_len: usize) {
+        metrics::PEER_DATA_RECEIVED_BYTES.inc_by(msg_len as u64);
+        metrics::PEER_MESSAGE_RECEIVED_TOTAL.inc();
+        self.tracker.increment_received(msg_len as u64);
+    }
+
+    /// Check whenever we exceeded number of transactions we got since last block.
+    /// If so, drop the transaction.
+    fn should_we_drop_msg_without_decoding(&self, msg: &Vec<u8>) -> bool {
+        if codec::is_forward_transaction(&msg).unwrap_or(false) {
+            let r = self.txns_since_last_block.load(Ordering::Acquire);
+            if r > MAX_TRANSACTIONS_PER_BLOCK_MESSAGE {
+                return true;
+            }
+        }
+        false
+    }
+
+    // Checks errors from decoding a message.
+    // We may send `HandshakeFailure` to the other peer.
+    fn handle_peer_message_decode_error(&mut self, msg: &Vec<u8>, err: Error) {
+        if let Some(version) = err
+            .get_ref()
+            .and_then(|err| err.downcast_ref::<HandshakeFailureReason>())
+            .and_then(|inner| {
+                if let HandshakeFailureReason::ProtocolVersionMismatch { version, .. } = *inner {
+                    Some(version)
+                } else {
+                    None
+                }
+            })
+        {
+            debug!(target: "network", "Received connection from node with unsupported version: {}", version);
+            self.send_message(&PeerMessage::HandshakeFailure(
+                self.node_info.clone(),
+                HandshakeFailureReason::ProtocolVersionMismatch {
+                    version: PROTOCOL_VERSION,
+                    oldest_supported_version: OLDEST_BACKWARD_COMPATIBLE_PROTOCOL_VERSION,
+                },
+            ));
+        } else {
+            info!(target: "network", "Received invalid data {:?} from {}: {}", logging::pretty_vec(&msg), self.peer_info, err);
+        }
+    }
 }
 
 impl Actor for PeerActor {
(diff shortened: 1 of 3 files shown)


pmnoxx:
@matklad I wish I could write code using '?" with Option. 
```

    fn handle(&mut self, msg: Result<Vec<u8>, ReasonForBan>, ctx: &mut Self::Context) {
        let msg = self.check_if_decoder_banned_peer(msg, ctx)?;
        // TODO(#5155) We should change our code to track size of messages received from Peer
        // as long as it travels to PeerManager, etc.

        let _ = self.update_stats_on_receiving_message(msg.len())?;

        let _ = self.should_we_drop_msg_without_decoding(&msg)?;
```

chain/network/src/peer/peer_actor.rs
```diff
@@ -608,55 +685,24 @@ impl WriteHandler<io::Error> for PeerActor {}
...
-                    info!(target: "network", "Received invalid data {:?} from {}: {}", logging::pretty_vec(&msg), self.peer_info, err);
-                }
-                return;
-            }
+
+        let mut peer_msg = if let Some(peer_msg) = self.try_to_decode_to_peer_message(&msg) {
+            peer_msg
+        } else {
+            return;
         };
```
matklad, line 706:
```suggestion
        let mut peer_msg = match self.try_to_decode_to_peer_message(&msg) {
            Some(peer_msg) => peer_msg,
            None => return,
        };
```

One line shorter and, imo, neater. My rule of thumb: every "if let else" is better written as match. Soonish we'll be able to write `let Some(mut peer_msg) = self.try_to_decode_to_peer_message(&msg)  else { return };`

chain/network/src/peer/peer_actor.rs
```diff
@@ -545,6 +546,82 @@ impl PeerActor {
             }
         }
     }
+
+    /// Check whenever `Codec::decode` method banned peer for message being too long.
+    /// If so bans peers
+    /// or returns the decoded message.
+    fn check_if_decoder_banned_peer(
```
matklad, line 553: I'd rather not have this function extracted. There are three smells currently: 

* the function is named `check`, but it takes `&mut` and manipulates state. 
* it hides control-flow -- to understand the call-site, it's useful to see an `if` there
* it doesnt' actually hide contrlo flow -- we still have a match at the call-site, it's just that the logic is more obscured. 

The original version 

```rust
        let msg = match msg {
            Ok(msg) => msg,
            Err(ban_reason) => {
                self.ban_peer(ctx, ban_reason);
                return None;
            }
        };
```

looks perfectly readable to me: "the first thing we do is we ban the peer if we have to". 

In general, if you have the code shaped like

```rust
if complex_multiline_condition {
   complex_multiline_logic
}
```

the preferred simplification is 

```rust
let cond = self.compute_condition();
if cond {
   self.do_foo()
}
```

rather than

```rust
self.do_foo_optionally()
```

Keepin control flow at the call site makes for more readable code.

chain/network/src/peer/peer_actor.rs
```diff
@@ -545,6 +546,82 @@ impl PeerActor {
...
+    /// Update stats when receiving msg
+    fn update_stats_on_receiving_message(&mut self, msg_len: usize) {
+        metrics::PEER_DATA_RECEIVED_BYTES.inc_by(msg_len as u64);
+        metrics::PEER_MESSAGE_RECEIVED_TOTAL.inc();
+        self.tracker.increment_received(msg_len as u64);
+    }
+
+    /// Check whenever we exceeded number of transactions we got since last block.
+    /// If so, drop the transaction.
+    fn should_we_drop_msg_without_decoding(&mut self, msg: &Vec<u8>) -> bool {
```
matklad, line 577: Can be `&self` I guess?

chain/network/src/peer/peer_actor.rs
```diff
@@ -545,6 +546,82 @@ impl PeerActor {
...
+            }
+        }
+        false
+    }
+
+    /// Decodes message using `BorshDeserialize
+    /// Somehow checks if there is a version mismatch. How exactly you ask, that's a great question?
+    /// TODO: Figure this out.
+    fn try_to_decode_to_peer_message(&mut self, msg: &Vec<u8>) -> Option<PeerMessage> {
+        let peer_msg = match bytes_to_peer_message(&msg) {
```
matklad, line 591: Similarly, I'd probably keep the `match` at the cal-site, and extracted only the `Err(err) =>` branch into a function.

matklad:
> wish I could write code using '?" with Option.

The way I do this today is via `mach`: 

```
let msg = match self.check_if_decoder_banned_peer(msg, ctx) {
  Some(it) => it,
  None => return,
}
```

If there are *a lot* of calls like that, I sometimes change the signature of the calling function to be `-> Option<T>`, but that's exception.

matklad:
>There is a 500 lines handle method in PeerActor. We should add comments to it and try to split it.

I have opinions here! To the first approximation, I don't think that line length is a problem in and of itself, and I think that 500 lines method might be OK, see http://number-none.com/blow/john_carmack_on_inlined_code.html. I agree with the conclusion that, when writing inherently stateful code which does things, it's better to have all state transitions as one wall of text. Factoring into functions is helpful when the code is mostly pure. 

Interestingly, the prime example of inherently stateful code is actor's main loop. For example, rust-analyzer is essentially [this giant 300 lines `handle_event` method](https://github.com/rust-analyzer/rust-analyzer/blob/b69eee64879edf9e9c063f9700e2c1f48fa0fefe/crates/rust-analyzer/src/main_loop.rs#L188-L534). This design is very intentional -- its a big method, but anything stateful that happens, happens there. It makes it *easier* to see how the state evolves under the events: it just makes the underlying complexity more apparent. 

My rules of thumb for Rust specifically: 

* do extract `&self` method
* don't generally extract `&mut self` method
* don't extract control flow -- keep `if` and `match` expressions in the caller, move conditions and stuff to the right of `=>` into helpers
* use `{}` and `let x = {}` to group logical bits within a method. 


No super strong opinion here though!

chain/network/src/peer/peer_actor.rs
pmnoxx: I agree. I wish there was an even shorted method.
 It's too bad that we can write
```
let mut peer_msg = self.try_to_decode_to_peer_message(&msg)?;
```
where `try_to_decode_to_peer_message` returns `Option<...>`

pmnoxx: Rust `?` works only with, `Result`, right?

chain/network/src/peer/peer_actor.rs
pmnoxx: @matklad 
This makes sense. However, we have methods like that, already:

```
    /// Hook called on every valid message received from this peer from the network.
    fn on_receive_message(&mut self) {
        if let Some(peer_id) = self.peer_id() {
            if self.last_time_received_message_update.elapsed()
                > UPDATE_INTERVAL_LAST_TIME_RECEIVED_MESSAGE
            {
                self.last_time_received_message_update = Clock::instant();
                self.peer_manager_addr.do_send(PeerManagerMessageRequest::PeerRequest(
                    PeerRequest::ReceivedMessage(peer_id, self.last_time_received_message_update),
                ));
            }
        }
    }```

We need to decide what to do with them.

File: chain/network/src/peer/peer_actor.rs (hunk 4 above)
Comment (pmnoxx) [reply chain/network/src/peer/peer_actor.rs]: @matklad Thanks, that's even better.

**Comment by pmnoxx** [comment]
> > There is a 500 lines handle method in PeerActor. We should add comments to it and try to split it.
> 
> I have opinions here! To the first approximation, I don't think that line length is a problem in and of itself, and I think that 500 lines method might be OK, see http://number-none.com/blow/john_carmack_on_inlined_code.html. I agree with the conclusion that, when writing inherently stateful code which does things, it's better to have all state transitions as one wall of text. Factoring into functions is helpful when the code is mostly pure.
> 
> Interestingly, the prime example of inherently stateful code is actor's main loop. For example, rust-analyzer is essentially [this giant 300 lines `handle_event` method](https://github.com/rust-analyzer/rust-analyzer/blob/b69eee64879edf9e9c063f9700e2c1f48fa0fefe/crates/rust-analyzer/src/main_loop.rs#L188-L534). This design is very intentional -- its a big method, but anything stateful that happens, happens there. It makes it _easier_ to see how the state evolves under the events: it just makes the underlying complexity more apparent.
> 
> My rules of thumb for Rust specifically:
> 
> * do extract `&self` method
> * don't generally extract `&mut self` method
> * don't extract control flow -- keep `if` and `match` expressions in the caller, move conditions and stuff to the right of `=>` into helpers
> * use `{}` and `let x = {}` to group logical bits within a method.
> 
> No super strong opinion here though!

Thanks, for good suggestions.

I agree that inlining isn't always the best option. However, as a general rule of thumb I prefer writing a code, 
such that's unnecessary for purpose of reading this function.

Rather than writing:
```

/// 5+ lines computing come_complicated_condition

let should_skip_msg = ...
if should_skip_msg {
}

```
I would rather write:
```
if self.some_complicated_msg(...) {

}
```

Another example would be:

```
/// Some code which updates some stats, has no effect on what the reset of function does.
/// Reading it's implementation can be skipped.
self.update_some_stats(...);
```

```
/// Some code which updates some stats, has no effect on what the reset of function does.
/// Reading it's implementation can be skipped.
self.update_some_stats(...);
```

```
/// Some code which updates some stats, has no effect on what the reset of function does.
/// Reading it's implementation can be skipped.
self.update_some_stats(...);
```
Or something like:

```
If not something {
    // It's better to hide error handling, as it's not that important for analyzing what the function does.
    handle_error(...)
    return;
}
```

File: chain/network/src/peer/peer_actor.rs (hunk 1 above)
Comment (matklad) [reply chain/network/src/peer/peer_actor.rs]: >Rust ? works only with, Result, right?

No, it also works with `Option` and [`ControlFlow`](https://doc.rust-lang.org/stable/std/ops/enum.ControlFlow.html), but it can't interconvert between diffrent container types:

https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=d5cfd40ecea9d548c5d21cab1b14a466

**Review by matklad — APPROVED** [review]

File: chain/network/src/peer/peer_actor.rs [hunk 5]
```diff
@@ -618,43 +664,15 @@ impl StreamHandler<Result<Vec<u8>, ReasonForBan>> for PeerActor {
...
-                        self.node_info.clone(),
-                        HandshakeFailureReason::ProtocolVersionMismatch {
-                            version: PROTOCOL_VERSION,
-                            oldest_supported_version: OLDEST_BACKWARD_COMPATIBLE_PROTOCOL_VERSION,
-                        },
-                    ));
-                } else {
-                    info!(target: "network", "Received invalid data {:?} from {}: {}", logging::pretty_vec(&msg), self.peer_info, err);
-                }
+                self.handle_peer_message_decode_error(&msg, err);
```
mm-near, line 675: here I'd also mention that this might send the HandshakeFailure to the peer if needed.

mm-near (approved):

PR #5412, opened 2021-11-22
re: Fix clippy warnings

pmnoxx:
Fix a couple of a simple issues reported by `clippy`.


utils/near-performance-metrics/src/stats_enabled.rs
```diff
@@ -172,7 +172,7 @@ impl ThreadStats {
             self.write_buf_drained = Default::default();
 
             let mut stat: Vec<_> = self.stat.iter().collect();
-            stat.sort_by(|x, y| (*x).0.cmp(&(*y).0));
+            stat.sort_by(|x, y| (*x).0.cmp(&*y.0));
```
matklad, line 175: I wonder if this can be just `.sort_by_key`?

utils/near-performance-metrics/src/stats_enabled.rs
```diff
@@ -253,7 +253,7 @@ impl Stats {
             MIN_OCCUPANCY_RATIO_THRESHOLD
         );
         let mut s: Vec<_> = self.stats.iter().collect();
-        s.sort_by(|x, y| (*x).0.cmp(&(*y).0));
+        s.sort_by(|x, y| (*x).0.cmp(&*y.0));
```
matklad, line 256: ditto

core/account-id/src/lib.rs
```diff
@@ -57,7 +57,7 @@ impl AccountId {
 
             (!last_char_is_separator)
                 .then(|| ())
-                .ok_or(ParseAccountError(ParseErrorKind::Invalid, account_id.to_string()))
```
matklad, line 60: I'd just 

```rust
if !last_char_is_separator {
    return Err(ParseAccountError(ParseErrorKind::Invalid, account_id.to_string()))
} 
Ok(())
```

The current version reads like overly-clever code to me.

matklad: Ok, I actually feel strongly that we should rewrite this. I am looking at `(!last_char_is_separator).then(|| ())` and I don't actually understand whether trailing separator is an error or not. As you see from the above suggestion, I actually got this wrong. 

The correct version is (please double check)

```rust
if last_char_is_separator {
    return Err(ParseAccountError(ParseErrorKind::Invalid, account_id.to_string()))
} 
Ok(())
```

matklad (approved):
LGTM, but I suggest rewriting the `(!last_char_is_separator).then(|| ())` bit, although it is unrelated to the PR.

pmnoxx:
@matklad Ok, I'll create a separate PR for that.

utils/near-performance-metrics/src/stats_enabled.rs
pmnoxx: @matklad Thanks, I replaced it.

PR #5413, opened 2021-11-22
chore: mark near-rate-limiter unpublishable

miraclx:
`near-rate-limiter`, introduced in https://github.com/near/nearcore/pull/4927, should be a private crate.


pmnoxx (approved):

pmnoxx:
@miraclx LGTM!

PR #5414, opened 2021-11-22
fix: near-rate-limter build

miraclx:
`near-rate-limiter`, while compilable as a dependency of `near-network`, is not compilable in isolation due to missing features.

```console
$ cargo check -p near-rate-limiter
   Compiling near-rate-limiter v0.0.0 (/git/near/nearcore/utils/near-rate-limiter)
error[E0432]: unresolved import `tokio_util::codec`
  --> utils/near-rate-limiter/src/framed_read.rs:13:17
   |
13 | use tokio_util::codec::Decoder;
   |                 ^^^^^ could not find `codec` in `tokio_util`
```


pmnoxx (approved):
LGTM

PR #5418, opened 2021-11-22
test: fixing 3 nayduck tests

mm-near:
- restart.py -- had a flaky assumption that after startup, the current head is <= 2 (probably used to work when our node creation a lot faster)
- rpc_finality - inside the code, it started 3 nodes, killed one of them and tried making the transaction, that didn't succeed. AFAIK you need now to have more than 2/3 of the stake for the chain to approve it. So I've changed number of nodes to 4.
- sync_chunks_from_archival - was checking that the data was read from the hardcoded archival node id (for example, that first 10 blocks would be read from 2, next 10 blocks from 3 etc) - AFAIK we don't enforce this in the code - so I've updated the test to verify that each block is requested from either node.

### DIFF
--- a/pytest/tests/sanity/rpc_finality.py
+++ b/pytest/tests/sanity/rpc_finality.py
@@ -14,10 +14,13 @@ sys.path.append('lib')
 from cluster import start_cluster
 from configured_logger import logger
 from transaction import sign_payment_tx
+
+
 class TestRpcFinality(unittest.TestCase):
+
     def test_finality(self):
         nodes = start_cluster(4, 1, 1, None,
-                            [["min_gas_price", 0], ["epoch_length", 100]], {})
+                              [["min_gas_price", 0], ["epoch_length", 100]], {})
 
         time.sleep(3)
         # kill one validating node so that no block can be finalized
@@ -31,7 +34,7 @@ class TestRpcFinality(unittest.TestCase):
         status = nodes[0].get_status()
         latest_block_hash = status['sync_info']['latest_block_hash']
         tx = sign_payment_tx(nodes[0].signer_key, 'test1', token_transfer, 1,
-                            base58.b58decode(latest_block_hash.encode('utf8')))
+                             base58.b58decode(latest_block_hash.encode('utf8')))
         logger.info("About to send payment")
         logger.info(nodes[0].send_tx_and_wait(tx, timeout=200))
         logger.info("Done")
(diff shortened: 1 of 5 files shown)


pytest/tests/sanity/rpc_finality.py
```diff
@@ -1,50 +1,60 @@
...
-        assert int(acc_doomslug_finality['result']
-                   ['amount']) == acc0_balance - token_transfer
-        assert int(acc_nfg_finality['result']['amount']) == acc0_balance
-    else:
-        assert int(acc_no_finality['result']
-                   ['amount']) == acc1_balance + token_transfer
-        assert int(acc_doomslug_finality['result']
-                   ['amount']) == acc1_balance + token_transfer
-        assert int(acc_nfg_finality['result']['amount']) == acc1_balance
+class TestRpcFinality(unittest.TestCase):
```
mm-near, line 17: you might want to apply 'hide whitespace' in github settings - otherwise it show a LOT of diffs.

bowenwang1996 (commented):
Could you comment on how exactly you fixed `sync_chunks_from_archival.py`? Also, for the title of the PR we usually do `test:...` instead of `[test] ...`

pytest/tests/sanity/sync_chunks_from_archival.py
```diff
@@ -62,209 +67,214 @@ async def handle(self, msg, fr, to):
...
 
-def get_validators(node):
-    return set([x['account_id'] for x in node.get_status()['validators']])
+    status = boot_node.get_status()
+    hash_ = status['sync_info']['latest_block_hash']
 
+    logging.info("Waiting for the new nodes to sync")
+    while True:
+        if not node2.get_status()['sync_info'][
+                'syncing'] and not node3.get_status()['sync_info']['syncing']:
```
mina86, line 199:
```
        if (not node2.get_status()['sync_info']['syncing'] and
            not node3.get_status()['sync_info']['syncing']):
```
reads better

mina86 (commented):
restart still looks flaky: http://nayduck.near.org/#/run/2234

pytest/tests/sanity/sync_chunks_from_archival.py
mm-near: yapf.. fixed.

mina86 (approved):

PR #5419, opened 2021-11-22
refactor: minimize work-around

matklad:
`use Trait as _` is a standard Rust idiom -- it allows calling trait's
methods without bringing the trait itself in scope. This is useful to
avoid naming conflicts, which is exactly what happens in this case
(there's a `Signature` struct defined in this file).

We *could* `as _` the other two crates as well, but we don't as:

a) we don't have an active name conflict for them
b) more than one `_` confuses CLion: https://github.com/intellij-rust/intellij-rust/issues/8143

Notes:

Rust naming convention suggests using verbs as trait names, which helps
with dodging similar conflicts.

### DIFF
--- a/core/crypto/src/signature.rs
+++ b/core/crypto/src/signature.rs
@@ -6,11 +6,9 @@ use std::io::{Error, ErrorKind, Write};
 use std::str::FromStr;
 
 use borsh::{BorshDeserialize, BorshSerialize};
-// We need to import ed25519::signature::Signature, because we use traits from those structs.
-// However, `Signature` symbol is already used to define a different data structure.
 #[cfg(feature = "deepsize_feature")]
 use deepsize::DeepSizeOf;
-use ed25519_dalek::ed25519::signature::{Signature as _Signature, Signer, Verifier};
+use ed25519_dalek::ed25519::signature::{Signature as _, Signer, Verifier};
 #[cfg(feature = "deepsize_feature")]
 use ed25519_dalek::SIGNATURE_LENGTH;
 use once_cell::sync::Lazy;


matklad:
cc @pmnoxx @mm-near 

I was a bit confused by the wording of the comment, as it implied that what we are importing are structs and not trait, so I figured I just rewrite this to be all of: 

* idiomatic Rust
* explicit (`Signature as _` makes it very clear to the reader that `Signature` is a trait)
* compatible with CLion

I also decided to remove the explicit comment. I am of two minds here -- on the one hand, `use Trait as _` is an explicit language feature, which is explained in the Rust book and the reference. On the other hand, while this isn't exactly an obscure dark corner, it's not an everyday feature. 

My reasoning was that we should generally assume that nearcore developers are proficient with Rust, and focus on teaching things once, rather than explaining them in code every time. Hence, no comment, but you can `git blame` the explanation or ask in Zulip.

pmnoxx (approved):
LGTM

pmnoxx:
> cc @pmnoxx @mm-near
> 
> I was a bit confused by the wording of the comment, as it implied that what we are importing are structs and not trait, so I figured I just rewrite this to be all of:
> 
> * idiomatic Rust
> * explicit (`Signature as _` makes it very clear to the reader that `Signature` is a trait)
> * compatible with CLion
> 
> I also decided to remove the explicit comment. I am of two minds here -- on the one hand, `use Trait as _` is an explicit language feature, which is explained in the Rust book and the reference. On the other hand, while this isn't exactly an obscure dark corner, it's not an everyday feature.
> 
> My reasoning was that we should generally assume that nearcore developers are proficient with Rust, and focus on teaching things once, rather than explaining them in code every time. Hence, no comment, but you can `git blame` the explanation or ask in Zulip.

Let's go ahead and remove the problem.

pmnoxx:
> cc @pmnoxx @mm-near
> 
> I was a bit confused by the wording of the comment, as it implied that what we are importing are structs and not trait, so I figured I just rewrite this to be all of:
> 
> * idiomatic Rust
> * explicit (`Signature as _` makes it very clear to the reader that `Signature` is a trait)
> * compatible with CLion
> 
> I also decided to remove the explicit comment. I am of two minds here -- on the one hand, `use Trait as _` is an explicit language feature, which is explained in the Rust book and the reference. On the other hand, while this isn't exactly an obscure dark corner, it's not an everyday feature.
> 
> My reasoning was that we should generally assume that nearcore developers are proficient with Rust, and focus on teaching things once, rather than explaining them in code every time. Hence, no comment, but you can `git blame` the explanation or ask in Zulip.

I agree. Let's remove the comment. We could say something like
```
// This is hotfix, due to bug in rust-analyzer (# ABCD).
```
However, this isn't a critical part of the code, I think comments are not needed.

bowenwang1996 (approved):

bowenwang1996:
@matklad what prevents us from merging this PR?

pmnoxx:
@bowenwang1996 I see that @matklad added `S-automerge` to merge this PR. However, `S-automerge` doesn't work well with external repositories, see `matklad:m/minimize-conflicts` above. 

`S-automerge` is supposed to do the rebase. This is handled by `bulldozer`, which is a program we run to support `S-automerge`. Unfortunately, `bulldozer` doesn't have permissions to do the rebase on external repos. So that failed.

We have requirement, that to merge, the branch needs to be up-to-date, and it takes 20 minutes to ran all the checks. So, if anyone merge a commit in between, you have to do this rebase manually.

There are a few ways to solve this issue, but that's a topic for another discussion.

matklad:
I was waiting for https://github.com/near/nearcore/pull/5509 to land, but I probably should heed my own advice and try to *not* intelligently schedule dependent PRs, and instead land things in greedy whatever order.

pmnoxx:
@matklad That's for consideration, there is not need to wait. Fixing merging conflicts is not an issue for me.

Issue #5371, opened 2021-11-18
Reorganize tests to save compile time


mzhangmzz:
Currently, many tests in the code are located in a separate crate, which hurts compile time. For example, `chain/client/tests`. We should rethink unit tests vs integration tests and move unit tests to within the crate, for example, move `chain/client/tests` to `chain/client/src/tests`, as per [https://matklad.github.io/2021/02/27/delete-cargo-integration-tests.html](https://matklad.github.io/2021/02/27/delete-cargo-integration-tests.html)

*   <span class="error">[x]</span> [https://github.com/near/nearcore/pull/5398](https://github.com/near/nearcore/pull/5398)
*   <span class="error">[x]</span> [https://github.com/near/nearcore/pull/5420](https://github.com/near/nearcore/pull/5420)
*   <span class="error">[x]</span> [https://github.com/near/nearcore/pull/5493](https://github.com/near/nearcore/pull/5493)
*   <span class="error">[x]</span> [https://github.com/near/nearcore/pull/5494](https://github.com/near/nearcore/pull/5494)
*   [ ] TODO /core
*   [ ] TODO /runtime
*   [ ] TODO rest

mzhangmzz:
cc @matklad

mzhangmzz:
related issue https://github.com/near/nearcore/issues/5149

pmnoxx:
> Currently, many tests in the code are located in a separate crate, which hurts compile time. For example, `chain/client/tests`. We should rethink unit tests vs integration tests and move unit tests to within the crate, for example, move `chain/client/tests` to `chain/client/src/tests`, as per https://matklad.github.io/2021/02/27/delete-cargo-integration-tests.html

@mzhangmzz I agree. There are other benefits to doing that as well. I often looks at tests within the module for examples on how the code is used. This makes it easier to understand the code.

pmnoxx:
@mzhangmzz  Based on  https://matklad.github.io/2021/02/27/delete-cargo-integration-tests.html
I see that for each crate, `(1+N)` binaries will be created, were
- 1 : binary will all tests inside `/src`
- N:  1 for each module in `/tests/`

We have 3 proposals:
A) keep as it is - (1+N) binaries needs to be build for tests. 
Those tests will only use public api.
B) move tests from `/tests` to `/tests/integration` - We need to build 2 binaries for tests.
Those tests will only use public api.
C) move tests from `/tests` to `/src/tests` - We need to build 1 binaries for tests.
[...]

PR #5420, opened 2021-11-22
Move tests from `/tests` to `/src/tests` for `/chain`

pmnoxx:
Move tests from `/tests` to `/src/tests` for `/chain`,

As part of #5371

Once this gets merged. I'll work on next steps.


mzhangmzz (approved):

mzhangmzz:
LTGM. Thanks!

matklad:
Unless I am missing something, I'd expect this to also add

```rust
#[cfg(test)]
mod tests;
```

to `chain/network/src/lib.rs` and a `tests.rs` with corresponding `mod` declaration for each file. 

Right now it seems that these tests are no longer run at all? 

Coululd you double check that if, eg, you add an explicit `panic!` to a tests you can see the failure when running `cargo test`?

pmnoxx:
@matklad Thanks, fixed.

pmnoxx:
@matklad Can you take a look?

chain/client/Cargo.toml
```diff
@@ -42,10 +42,10 @@ near-performance-metrics-macros = { path = "../../utils/near-performance-metrics
 near-vm-runner = { path = "../../runtime/near-vm-runner" }
 delay-detector = { path = "../../tools/delay_detector", optional = true }
 near-network-primitives = { path = "../network-primitives" }
+near-actix-test-utils = { path = "../../test-utils/actix-test-utils" }
+near-logger-utils = { path = "../../test-utils/logger" }
```
mzhangmzz, line 46: Do we need to move these dependencies? I think we shouldn't because dev-dependencies should work for unit tests

pmnoxx: @mzhangmzz Thanks, fixed.

matklad (approved):

pmnoxx:
@bowenwang1996 @mm-near your approval is required.

bowenwang1996 (approved):

PR #5421, opened 2021-11-22
re: Refactor  `near-performance-metrics`  based on #5412

pmnoxx:
Refactor `near-performance-metrics` based on #5412 feedback from @matklad.

### DIFF
--- a/core/account-id/src/lib.rs
+++ b/core/account-id/src/lib.rs
@@ -58,9 +58,10 @@ impl AccountId {
                 last_char_is_separator = current_char_is_separator;
             }
 
-            (!last_char_is_separator)
-                .then(|| ())
-                .ok_or_else(|| ParseAccountError(ParseErrorKind::Invalid, account_id.to_string()))
+            if last_char_is_separator {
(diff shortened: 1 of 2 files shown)


core/account-id/src/lib.rs
```diff
@@ -55,9 +55,10 @@ impl AccountId {
                 last_char_is_separator = current_char_is_separator;
             }
 
-            (!last_char_is_separator)
-                .then(|| ())
-                .ok_or_else(|| ParseAccountError(ParseErrorKind::Invalid, account_id.to_string()))
+            if last_char_is_separator {
```
matklad, line 58: Out of curioucity, does any test fail if we change this to `if !last_char_is_separator`

matklad (approved):

core/account-id/src/lib.rs
pmnoxx: @matklad I'll test it. I excluded this change from this PR, as this PR is supposed to be about `near-performance-metrics` and this change seems out of scope.

PR #5425, opened 2021-11-23
Stabilizing block header v3

EgorKulikov:


EgorKulikov:
db migration fails because this is not the first feature we are stabilizing in this release. If I bump version for this two features to 50, it passes this check (but fails upgradability, see https://buildkite.com/nearprotocol/nearcore/builds/7922)

core/store/src/migrations.rs
```diff
@@ -749,8 +743,7 @@ pub fn migrate_28_to_29(path: &Path) {
     set_store_version(&store, 29);
 }
 
-#[cfg(feature = "protocol_feature_block_header_v3")]
-pub fn migrate_18_to_new_validator_stake(store: &Store) {
+pub fn migrate_29_to_30(path: &Path) {
```
mm-near, line 746: comment what this migration does (the V2-> V3 headers)

nearcore/src/lib.rs
```diff
@@ -252,15 +251,16 @@ pub fn apply_store_migrations(path: &Path, near_config: &NearConfig) {
         info!(target: "near", "Migrate DB from version 28 to 29");
         migrate_28_to_29(&path);
     }
+    if db_version <= 29 {
+        // version 28 => 29: migrate all structures that use ValidatorStake to versionized version
```
mm-near, line 255: bad comment (we migrate 28 -> 29).

and BTW - seems that line 249 above is wrong ..

mm-near (approved):

core/store/src/migrations.rs
EgorKulikov: My sense is that we commenting such things at the place of call

nearcore/src/lib.rs
EgorKulikov: I will introduce separate pr for that

bowenwang1996:
Merging because the failure is due to the fact that we've already had a protocol version bump on master

Issue #5424, opened 2021-11-22
`PeerManagerActor` rename `active_peers` to `connected_peers`


pmnoxx:
Currently, we use term `active peer` inside `PeerManagerActor`. Based on our discussion 
https://github.com/near/nearcore/pull/5230#discussion_r754600454 the usage of this term is confusing.

In `PeerManagerActor` `active_peers` refers to connected peers, or peers that we are in process of being connected.

I propose we rename `active peer` terminology to `connected_peer`.

PR #5428, opened 2021-11-23
re: Rename active peer to connected peer terminology in `near-network`

pmnoxx:
Change terminology used in `near-network` to be more accurate.

Currently we use `active peer` terminology to mean `connected peer`.
Unfortunately, that term is not self explanatory and leads to some confussion.

See #5424

Blocked by #5515

### DIFF
--- a/chain/client-primitives/src/types.rs
+++ b/chain/client-primitives/src/types.rs
@@ -463,8 +463,8 @@ impl From<near_chain_primitives::Error> for GetGasPriceError {
 
 #[derive(Debug)]
 pub struct NetworkInfoResponse {
-    pub active_peers: Vec<PeerInfo>,
-    pub num_active_peers: usize,
+    pub connected_peers: Vec<PeerInfo>,
+    pub num_connected_peers: usize,
     pub peer_max_count: u32,
     pub sent_bytes_per_sec: u64,
     pub received_bytes_per_sec: u64,
--- a/chain/client/src/client_actor.rs
+++ b/chain/client/src/client_actor.rs
@@ -176,8 +176,8 @@ impl ClientActor {
             network_adapter,
             node_id,
             network_info: NetworkInfo {
-                active_peers: vec![],
-                num_active_peers: 0,
+                connected_peers: vec![],
+                num_connected_peers: 0,
                 peer_max_count: 0,
                 highest_height_peers: vec![],
                 received_bytes_per_sec: 0,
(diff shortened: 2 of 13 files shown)


chain/client-primitives/src/types.rs
```diff
@@ -464,8 +464,8 @@ impl From<near_chain_primitives::Error> for GetGasPriceError {
 
 #[derive(Serialize, Deserialize, Debug)]
 pub struct NetworkInfoResponse {
-    pub active_peers: Vec<PeerInfo>,
-    pub num_active_peers: usize,
+    pub connected_peers: Vec<PeerInfo>,
+    pub num_connected_peers: usize,
```
frol, line 468: This will break JSON RPC `network` backward compatibility since the old names won't be there:

```
$ http post https://rpc.testnet.near.org jsonrpc=2.0 method=network_info params:='[]' id=dontcare
```

It is suboptimal to expect this type to be serializable straight to JSON RPC response, so I suggest we go and fix it in https://github.com/near/nearcore/blob/9e6c2078dae67938a4fa29a13bb502b04819b7b8/chain/jsonrpc-primitives/src/types/network_info.rs#L5-L6, so instead of flattening `NetworkInfoResponse` there, let's have a copy of the old fields over:

```rust
    pub active_peers: Vec<PeerInfo>,
    pub num_active_peers: usize,
    pub peer_max_count: u32,
    pub sent_bytes_per_sec: u64,
    pub received_bytes_per_sec: u64,
    /// Accounts of known block and chunk producers from routing table.
    pub known_producers: Vec<KnownProducer>,
```

 and implement the `From`:

https://github.com/near/nearcore/blob/9e6c2078dae67938a4fa29a13bb502b04819b7b8/chain/jsonrpc-primitives/src/types/network_info.rs#L16-L20

frol (approved):
I am approving to unblock the PR, but it is important to address the backward compatibility comment

chain/client-primitives/src/types.rs
mm-near: That's a very good catch.. And a gap in our testing (as we should have a test that loads the JSON requests from a file - to simulate backwards compatiblity).

mm-near (dismissed):
let's make sure that backwards compatibility is fixed before proceeding with this PR

chain/client-primitives/src/types.rs
pmnoxx: We should do the code separation. I'm changing this PR to `draft` until we merge https://github.com/near/nearcore/pull/5515

I created a separate issue for `jsonrpc`  https://github.com/near/nearcore/issues/5516

miraclx (approved):

chefsale (approved):

mm-near (approved):

PR #5430, opened 2021-11-23
Fixing version required for 28->29 db migration

EgorKulikov:


EgorKulikov:
Maybe the correct way is to use = instead of <= going forward?

mm-near (approved):

mzhangmzz:
> Maybe the correct way is to use == instead of <= going forward?

I think there might be cases where a node upgrade from for example 28 to 30, then == will not work?

PR #5435, opened 2021-11-23
pytest: prefer json.load and json.dump to loads and dumps

mina86:
Prefer using json.load (json.dump) to json.loads (json.dumps) when in
addition an explicit read from (write to) a file is performed.  This
simplifies the code since there’s no longer a need for explicit read
(write) method call on the opened file.

In a few places, replace ‘naked’ open with a context manager which
closes the file when exiting the scope.


mm-near (approved):

chefsale (approved):
👯‍♂️

PR #5436, opened 2021-11-23
re: Stop using conditional imports in `near-network`

pmnoxx:
We should improve code quality by removing conditional imports.

There are a few issues with conditional imports:
-  high maintenance  cost - modifying code under conditional flags, requires manually adding #[cfg(feature=)] for each import. Rust plugin for Intellij doesn't support them well.
- readability - this makes code more readable, it's good to see what's being imported when to feature flags are used. 
- fragility CI - I've seen a lot cases, where I submit PR, only to find out, something didn't compile with some subset of features.

### DIFF
--- a/chain/network-primitives/src/types.rs
+++ b/chain/network-primitives/src/types.rs
@@ -13,8 +13,6 @@ use near_primitives::sharding::{
     ChunkHash, PartialEncodedChunk, PartialEncodedChunkPart, PartialEncodedChunkV1,
     PartialEncodedChunkWithArcReceipts, ReceiptProof, ShardChunkHeader,
 };
-#[cfg(feature = "sandbox")]
-use near_primitives::state_record::StateRecord;
 use near_primitives::syncing::{
     EpochSyncFinalizationResponse, EpochSyncResponse, ShardStateSyncResponse,
     ShardStateSyncResponseV1,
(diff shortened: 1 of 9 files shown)


pmnoxx:
@matklad What do you thing about it?

mm-near (approved):
I like it - but I'm also curious of what @matklad  thinks

matklad (approved):
As evidenced by the diff stat, much better!

One further improvement would be to pull `#[cfg(features = "delay_detoctor")]` into delay detector itself, such that we don't need to add cfg at every call-site: 

```rust
pub struct DelayDetector<'a> {
    repr: DelayDetectorRepr<'a>,
}

#[cfg(features = "performance_stats")]
struct DelayDetectorRepr<'a> {
    min_delay: Duration,
    msg: Cow<'a, str>,
    started: Instant,
    started_cpu_time: ProcessTime,
    snapshots: Vec<((String, String), Snapshot)>,
    last_snapshot: Option<(String, SnapshotInstant)>,
}

#[cfg(not(features = "performance_stats"))]
struct DelayDetectorRepr<'a> {
    ghost: PhantomData<&'a str>,
}
```

pmnoxx:
> delay_detoctor

We could do that. However, right now `delay-detector` package is optional. Do we want to make that package mandatory?

pmnoxx:
@matklad I we want to do it, here is the PR: https://github.com/near/nearcore/pull/6176

PR #5440, opened 2021-11-23
Update `bytesize` crate to latest version supporting AddAssign

mina86:
Versions of `bytesize` prior to 1.1 do not implement AddAssign trait
which leads to the following build failures:

    error[E0368]: binary assignment operation `+=` cannot be applied to type `ByteSize`
      --> /home/mpn/code/near/nearcore/utils/near-performance-metrics/src/stats_enabled.rs:85:9
       |
    85 |         self.write_buf_added += ByteSize::b(bytes as u64);
       |         --------------------^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
       |         |
       |         cannot use `+=` on type `ByteSize`

    error[E0368]: binary assignment operation `+=` cannot be applied to type `ByteSize`
      --> /home/mpn/code/near/nearcore/utils/near-performance-metrics/src/stats_enabled.rs:92:9
       |
    92 |         self.write_buf_drained += ByteSize::b(bytes as u64);
       |         ----------------------^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
       |         |
       |         cannot use `+=` on type `ByteSize`

    error[E0368]: binary assignment operation `+=` cannot be applied to type `ByteSize`
       --> /home/mpn/code/near/nearcore/utils/near-performance-metrics/src/stats_enabled.rs:268:13
        |
    268 |             other_memory_size += tmp_other_memory_size;
        |             -----------------^^^^^^^^^^^^^^^^^^^^^^^^^
        |             |
        |             cannot use `+=` on type `ByteSize`

This fixes fuzz test on NayDuck.


pmnoxx (approved):
LGTM

Good idea. I looked at the change log. Those are good improvements. We should try to migrate and use the new api.

pmnoxx:
Once, this gets merged. I'll update `near-performance-metrics` code.

bowenwang1996 (approved):

matklad:
cc @nagisa the underlying problem here is that our fuzz tests use a different Cargo.lock than the rest of the code, which is less than ideal. Is there some easy way to teach fuzz to use the workspace Cargo.lock?

nagisa:
`cargo-fuzz` was made before workspaces were a thing, but I don't see an obvious reason why adding the fuzz targets to the top-level workspace wouldn't work.

matklad:
That's insightful! I was under the impressoon that there's super deep and important reason why the things work this way! 

I'll be pushing fuzzing to near_vm_runner soon, so I guess I'll setup it there the right way, and also try to fix general runtime fuzzing in process!

mina86:
> `cargo-fuzz` was made before cargo workspaces were a thing, but I don't see an obvious reason why adding the fuzz targets to the top-level workspace wouldn't work.

So just:

```
@@ -28,6 +28,7 @@ members = [
     "chain/rosetta-rpc",
     "test-utils/actix-test-utils",
     "test-utils/runtime-tester",
+    "test-utils/runtime-tester/fuzz",
     "test-utils/state-viewer",
     "test-utils/store-validator",
     "test-utils/testlib",

```
?

nagisa:
Yeah, please try it out. I'm not 100% confident this will work, but if it doesn't, then it sounds like something I could just implement in `cargo-fuzz` for our needs.

PR #5445, opened 2021-11-24
re: Refactor `near-network` misc files

pmnoxx:
There are a few files in `near-network` that can be refactored.


chain/network/src/types.rs
```diff
@@ -386,7 +384,7 @@ pub struct RoutingVersion2 {
 
 impl fmt::Display for PeerMessage {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        f.write_str(format!("{}", self.msg_variant()).as_ref())
+        f.write_str(self.msg_variant().to_string().as_ref())
```
matklad, line 387:
```suggestion
        fmt::Display::fmt(self.msg_variant(), f)
```

This avoids allocation, and, subjectively, more directly expresses the intent.

matklad (approved):

mm-near (approved):

PR #5448, opened 2021-11-24
fix: validate functions number after initialization

Longarithm:
We've noticed that if we validate functions number just after initialization, it speeds up the contract processing.

Also we want to do this before release, because number of functions computed **before** and **after** the change is actually different. I'm working on the repro.

## Test plan

- existing tests,
- adjust `test_limit_contract_functions_number` so that it checks precise limit now.


matklad (approved):

matklad:
Kudos @Longarithm -- rather important and definitely timely finding!

Issue #2483, opened 2020-04-16
Change config in pytest


bowenwang1996:
In python tests, we have a convenient way to change node configs before the cluster is started, but there is no easy way to stop the node, change its config and restart again.

mina86:
Do we have an example of a test that needs that?

bowenwang1996:
> Do we have an example of a test that needs that?

Unfortunately I cannot remember the exact scenario :(

PR #5451, opened 2021-11-24
pytest: add LocalNode.change_config method

mina86:
Introduce cluster.LocalNode.change_config method which allows changing
the config.json file of a node after the cluster has been initialised.

Fixes: https://github.com/near/nearcore/issues/2483


pytest/lib/cluster.py
```diff
@@ -291,35 +331,33 @@ def __init__(self,
...
         self.validator_key = Key.from_json_file(
             os.path.join(node_dir, "validator_key.json"))
         self.node_key = Key.from_json_file(
             os.path.join(node_dir, "node_key.json"))
         self.signer_key = Key.from_json_file(
             os.path.join(node_dir, "validator_key.json"))
-
         self.pid = multiprocessing.Value('i', 0)
 
+        self.change_config({
```
mm-near, line 342: nice.

mm-near (approved):

PR #5455, opened 2021-11-25
feat: Add bandwidth stats to `PeerManagerActor`

pmnoxx:
We would like to track the amount of bandwidth used for all `PeerActors`.

See https://github.com/near/nearcore/issues/5155


chain/network/src/peer_manager/peer_manager_actor.rs
```diff
@@ -1534,6 +1576,9 @@ impl Actor for PeerManagerActor {
 
         // Update routing table and prune edges that are no longer reachable.
         self.update_routing_table_trigger(ctx);
+
+        // Update routing table and prune edges that are no longer reachable.
```
mm-near, line 1580: bad comment

chain/network/src/types.rs
```diff
@@ -457,6 +458,8 @@ pub struct Consolidate {
     pub(crate) other_edge_info: EdgeInfo,
     // Protocol version of new peer. May be higher than ours.
     pub(crate) peer_protocol_version: ProtocolVersion,
+    /// A helper data structure for limiting reading, reporting bandwidth stats.
```
mm-near, line 461: nit: is it a 'data structure' ? ;-) or simply a class

utils/near-rate-limiter/src/framed_read.rs
```diff
@@ -254,6 +277,8 @@ where
 
             let bytect = ready!(poll_read_buf(pinned.inner.as_mut(), cx, &mut state.buffer)?);
 
+            pinned.throttle_controller.report_bandwidth_used(bytect);
+
```
mm-near, line 281: can you add a test that checks that throttler.compute_msg_seen / consume_bandwidth_used is working.

mm-near (dismissed):

chain/network/src/types.rs
pmnoxx: @mm-near Rust doesn't support classes: https://stevedonovan.github.io/rust-gentle-intro/object-orientation.html

pmnoxx:
@mm-near Can you take a look? This change is important in order to be able to ship this feature.

bowenwang1996 (approved):

PR #5456, opened 2021-11-25
re: Refactor triggers in `PeerManagerActor`

pmnoxx:
Currently `PeerManagerActor` uses triggers various tasks. In order to make triggers more readable, let's put their intervals inside arguments.

For example:
```
        self.monitor_peer_stats_trigger(ctx, self.config.peer_stats_period);
        self.broadcast_edges_trigger(ctx, BROADCAST_EDGES_INTERVAL);
        self.update_routing_table_trigger(ctx, UPDATE_ROUTING_TABLE_INTERVAL);
```

### DIFF
--- a/chain/network/src/peer_manager/peer_manager_actor.rs
+++ b/chain/network/src/peer_manager/peer_manager_actor.rs
@@ -342,7 +342,7 @@ impl PeerManagerActor {
     /// - there are edges, that were supposed to be added, but are still in `EdgeVerifierActor,
     ///   waiting to have their signatures checked.
     /// - edge pruning may be disabled for unit testing.
-    fn update_routing_table_trigger(&mut self, ctx: &mut Context<Self>) {
+    fn update_routing_table_trigger(&mut self, ctx: &mut Context<Self>, interval: Duration) {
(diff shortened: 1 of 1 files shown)


chain/network/src/peer_manager/peer_manager_actor.rs
```diff
@@ -1521,19 +1505,19 @@ impl Actor for PeerManagerActor {
         }
 
         // Periodically push network information to client
-        self.push_network_info_trigger(ctx);
+        self.push_network_info_trigger(ctx, self.config.push_info_period);
 
         // Start peer monitoring.
```
mm-near, line 1510: mention, that for this one, the waiting time differs (as AFAIK it uses some exponential backoff inside)

mm-near (approved):

chain/network/src/peer_manager/peer_manager_actor.rs
pmnoxx: @mm-near Can you take a look again. I updated description, and i added extra argument max_interval. This argument tells us maximum interval at which the trigger will run,

PR #5459, opened 2021-11-25
Add a test for preparation of multiple memories

nagisa:
Our preparation/sanitization pass assumes a single memory only so we
better make sure modules with only 1 memory are supported.

If we did in fact enable `multiple_memories` then the preparation code for
the memories would work incorrectly.


matklad (approved):
LGTM!

One thing I am not entirely sure about is whether this is the best way to test this. In some sense, preparation is a private impl detatail, and what we actually care about is `run` failing.

nagisa:
> is [..] this is the best way to test this

The goal here is specifically ensure that the shortcomings of the private impl details do not get exposed unintentionally as a result of us deciding to e.g. enable the `multi-memory` feature in the future. This test isn't really intended to ensure anything else. For instance ensuring that we invalidate modules with multiple memories while this feature is disabled is verified by `near-vm-runner/src/tests/wasm_validation.rs`.

PR #5462, opened 2021-11-25
test: add vm-level fuzzing

matklad:
Finally got to publishing fuzzing infra we have been using in ad-hoc ways before. This just runs the contracts to make sure that VM doesn't crash. The code is setup to easily compare execution with different VM Kinds. 

Note that at the moment fuzzing *does not* generally exercise host function calls. I need to finish https://github.com/bytecodealliance/wasm-tools/pull/286 for that :sweat_smile:


runtime/near-vm-runner/fuzz/fuzz_targets/runner.rs
```diff
@@ -0,0 +1,108 @@
...
+                }
+            }
+            _ => (),
+        }
+    }
+    None
+}
+
+fn create_context(input: Vec<u8>) -> VMContext {
+    VMContext {
```
nagisa, line 70: Aren't some of these potentially controlled by the user? Would be ideal if the user-influenced parts were `Arbitrary` too. Potentially as a separate fuzz target?

nagisa (approved):
This seems reasonable to me. I'm amazed how well `arbitrary` has caught on that there's even an implementation of `Arbitrary` to generate wasm modules!

runtime/near-vm-runner/fuzz/fuzz_targets/runner.rs
matklad: Yeah, that's a good idea. Let me issue this... #5465

mina86:
pytest and nightly changes LGTM

pytest/lib/cargo_fuzz.py
```diff
@@ -1,9 +1,26 @@
...
 import os
-import subprocess
 import sys
+import subprocess
+
+
+def run(dir: str, fuzz_target: str) -> int:
+    args = ('cargo', 'fuzz', 'run', fuzz_target, '--', '-len_control=0'
+            '-prefer_small=0', '-max_len=4000000', '-rss_limit_mb=10240')
+    os.environ['RUSTC_BOOTSTRAP'] = '1'
```
olonho, line 10: Hmm, why it's needed?

runtime/near-vm-runner/fuzz/Cargo.toml
```diff
@@ -0,0 +1,26 @@
+[package]
+name = "near-vm-runner-fuzz"
+version = "0.0.0"
+authors = ["Automatically generated"]
```
olonho, line 4: Is it?

olonho (approved):

runtime/near-vm-runner/fuzz/Cargo.toml
matklad: It is. Though as `authors` field is now optional, it's better to remove it here (and upstream in the generator itself as well, cc @nagisa ).

pytest/lib/cargo_fuzz.py
matklad: To spell this implicitly, but butting fuzzing into our CI we make it our responsibility to deal with any breakage related to compiler version upgrades. This in contrast to all over Rust functionality we use -- if something stable&documented breaks with an upgrade, we can and should complain upstream about that.

Issue #5310, opened 2021-11-17
`PrepareError` occurred when call or view contract function


riversyang:
When call or view functions of a contract which written in `near-sdk-rs` v3.1.0, a `PrepareError` occurred and the contract can not work anymore. The sample error message are as the following:

```shell
Scheduling a call: wchain.dev-oct-registry.testnet.migrate_state()
Doing account.functionCall()
Receipt: D6v43eXYFZQVFbvnXgPTaKERGCeUhCCCyN797vYDfZuR
        Failure [wchain.dev-oct-registry.testnet]: Error: Error happened during instantiation
ServerTransactionError: Error happened during instantiation
    at Object.parseResultError (/usr/local/lib/node_modules/near-cli/node_modules/near-api-js/lib/utils/rpc_errors.js:31:29)
    at Account.signAndSendTransactionV2 (/usr/local/lib/node_modules/near-cli/node_modules/near-api-js/lib/account.js:160:36)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
    at async scheduleFunctionCall (/usr/local/lib/node_modules/near-cli/commands/call.js:57:38)
    at async Object.handler (/usr/local/lib/node_modules/near-cli/utils/exit-on-error.js:52:9) {
  type: 'Instantiate',
  context: undefined,
  index: 0,
  transaction_outcome: {
    block_hash: '3Jv4hShujXFwD44A2viRuma69FbAHYCfR9FnUjNnX45M',
    id: 'FzE8e11cJ91CEBv69ravUfCxyUMLdzHj6BjkhRstHi7a',
    outcome: {
      executor_id: 'dev-oct-registry.testnet',
      gas_burnt: 2427954539010,
      logs: [],
      metadata: [Object],
      receipt_ids: [Array],
      status: [Object],
      tokens_burnt: '242795453901000000000'
    },
    proof: []
  }
}
```

```shell
View call: wchain.dev-oct-registry.testnet.get_anchor_settings()
An error occured
Error: Querying [object Object] failed: wasm execution failed with error: FunctionCallError(CompilationError(PrepareError(Instantiate))).
{
  "block_hash": "7M3oTMvwEzonNx1NAtFATSWmYx2aAAmfFhCohiMLyCBM",
  "block_height": 71663269,
  "error": "wasm execution failed with error: FunctionCallError(CompilationError(PrepareError(Instantiate)))",
  "logs": []
}
[...]

PR #5466, opened 2021-11-25
refactor: Improve the error messages from prepare.rs

nagisa:
Contract preparation is the first code that will typically see the
contract code and consequentially it is also the first which will report
issues with the contract.

One such example is #5310 for which the preparation code should now
report something more meaningful.

I went ahead and implemented `std::error::Error` in some places, but it
seems to me like we probably want to move the entire file to
`thiserror`...

This is indended to be NFC in that there should not be any meaningful functionality
differences other than a change the error messages and variants seen.


nagisa:
Some observations for future work: we should not be formatting the causations into the `Display`ed error message. Instead all of the errors should have a reasonable `source` chain and the error messages should be formatted into the `: ` separated form at the presentation layer or similar, rather than baked into the code.

runtime/near-vm-errors/src/lib.rs
```diff
@@ -130,9 +154,10 @@ pub enum CompilationError {
 /// Error that can occur while preparing or executing Wasm smart-contract.
 pub enum PrepareError {
     /// Error happened while serializing the module.
-    Serialization,
+    Serialization(StringError),
     /// Error happened while deserializing the module.
-    Deserialization,
+    Deserialization(StringError),
```
matklad, line 159: >This is indended to be NFC in that there should not be any meaningful functionality
differences other than a change the error messages and variants seen.

Alas, that's not the case, and this is one example. The following code works on master, but fails with this PR:

```rust
#[test]
fn serialization_compat() {
    let _: PrepareError = BorshDeserialize::try_from_slice(&[1]).unwrap();
}
```

And this `[1]` is stored in the database for real transactions.

matklad:
>separated form at the presentation layer or similar, rather than baked into the code.

Sadly, this *is* the presentation layer at the moment. Specifically, `#[derive(Serialize)]` is what defines our HTTP API (relevant entry struct: https://github.com/near/nearcore/blob/feb0acdb8b798d77e10b2217bcaa9720f93557f8/core/primitives/src/views.rs#L1040-L1063). 

You can see how the user sees the errors here: https://explorer.testnet.near.org/transactions/45RShTh6wCnRg3PQVu9nBPXZ7PsK5hqFyEGPEAfCNX7n#GGs6dB4ktqJhVNN4gZBPNoKZiPAw4VSSqExouHsFGuNw

nagisa:
Yeah, okay, never mind then. I can see approaches we could take to make changes backwards compatibly, but I don't really see it being worth it without a more in-depth solution to decouple code errors and database/API layers.

PR #5468, opened 2021-11-25
(fix) bp hash and validator assignment for first epoch of new version

EgorKulikov:
We incorrectly calculated next_bp_hash and validator assignment for next epoch in the last epoch with previous version

I would add tests tomorrow if we'll have time, otherwise I would create ticket and separate PR afterwards


bowenwang1996 (approved):

ailisp (approved):
Looks good to me

chain/epoch_manager/src/proposals.rs
```diff
@@ -46,8 +46,9 @@ pub fn proposals_to_epoch_info(
     validator_reward: HashMap<AccountId, Balance>,
     minted_amount: Balance,
     next_version: ProtocolVersion,
+    last_epoch_version: ProtocolVersion,
```
matklad, line 49: nit: it feels odd to pass new, then old rather than old, then new (ie, I'd swapped the order of two params)

matklad (commented):
Should this include a new test?

bowenwang1996 (approved):

PR #5472, opened 2021-11-25
re: Rewrite documentation for `Edge` struct

pmnoxx:
Add in-code documentation for `Edge`, which will be useful for writing reference level documentation.


chain/network/src/routing/edge.rs
```diff
@@ -217,23 +218,34 @@ impl Edge {
     }
 }
 
-/// Edge object. Contains information relative to a new edge that is being added or removed
-/// from the network. This is the information that is required.
+/// Represents `edge` in Near Protocol P2P network.
```
matklad, line 221: Tip: when writing a summary line, it's useful to avoid words already present in the name of the thing, to make the docs less tautological. 

I'd phrase it this way:

```
/// An `Edge` represents a direct connection between two peers in Near Protocol P2P network. 
///
/// Note that edge might either active or inactive. That is, we represent removed edge explicitly. (maybe some short note about *why* we need that)
```

mm-near: +1 - especially to the part why we're keeping 'removed' edges

chain/network/src/routing/edge.rs
```diff
@@ -217,23 +218,34 @@ impl Edge {
...
 pub struct EdgeInner {
-    /// Since edges are not directed `key.0 < peer1` should hold.
+    /// Each edge consists of unordered pair of public keys of both peers.
+    /// `key.0 < key.1` holds true.
     key: (PeerId, PeerId),
-    /// Nonce to keep tracking of the last update on this edge.
-    /// It must be even
+    /// `nonce` is unique number representing state of an edge.
+    /// Odd number indicates that `edge` has been added, `even` number that it was removed.
+    /// New edge starts with value of `1`.
```
mm-near, line 232: I'd also mention that 'update' the edge only if the next nonce is larger.

mm-near (approved):

chain/network/src/routing/edge.rs
pmnoxx: I added documentation about why we keep both `Active` and `Removed` edges.

pmnoxx:
@mm-near @matklad I updated the documentation, can you take a look?

PR #5475, opened 2021-11-26
feat: Replace HashMap with HashSet

pmnoxx:
Currently `RoutingTableActor::edges_info` is using `HashMap<(PeerId, PeerId), Edge>` to store list of edges.
This if inefficient, as it requires storing both keys and value.
We can optimize this by replacing `HashMap` with `HashSet`.


pmnoxx:
@matklad Can you take a look?

matklad (commented):
LGTM!

I'd maybe considered creating a separate type here: 

```rust
// in chain/network/src/routing/edge_set.rs

pub(crate) struct EdgeSet {
  repr: HashSet<EdgeIndexedByKey>
}

struct EdgeIndexedByKey(Edge);

impl EdgeSet {
    pub(crate) fn insert(&mut self, edge: Edge) { ... }
    pub(crate) fn get(&self, key: &(PeerId, PeerId)) -> Option<&Edge> { ... }
}
```

The HashMap -> HashSet trick is a bit weird, so, if we could encapsulate it in a module, that'd be great. That way, `EdgeIndexedByKey` weirdness will be strictly contained to a single, small file. 


Also curious to see if this affects any benchmark results!

chain/network/src/routing/edge_set.rs
```diff
@@ -0,0 +1,59 @@
...
+use std::collections::HashSet;
+use std::hash::{Hash, Hasher};
+
+/// Wraps around `Edge` struct. The main feature of this struct, is that it's hashed by
+/// `(Edge::key.0, Edge::key.1)` pair instead of `(Edge::key.0, Edge::key.1, Edge::nonce)`
+/// triple.
+#[derive(Eq, PartialEq)]
+pub struct EdgeIndexedByKey {
+    pub inner: Edge,
+}
```
matklad, line 14:
```suggestion
struct EdgeIndexedByKey {
    inner: Edge,
}
```

Making this private is the main thing I'd love to achieve by moving stuff to a separate module

chain/network/src/routing/edge_set.rs
```diff
@@ -0,0 +1,59 @@
...
+        self.repr.get(key).map(|v| &v.inner)
+    }
+
+    pub(crate) fn remove(&mut self, key: &(PeerId, PeerId)) -> bool {
+        self.repr.remove(key)
+    }
+
+    pub(crate) fn iter(&self) -> Iter<'_, EdgeIndexedByKey> {
+        self.repr.iter()
+    }
```
matklad, line 53:
```suggestion
    pub(crate) fn iter(&self) -> impl Iterator<Item = &Edge> + '_ {
        self.repr.iter().map(|it| &it.repr)
    }
```

pmnoxx: @matklad Thanks, that looks much nicer!.

chain/network/src/routing/edge_set.rs
pmnoxx: @matklad Yes, I can do that thanks to usage of `impl Iterator`.

pmnoxx:
> LGTM!
> 
> I'd maybe considered creating a separate type here:
> 
> ```rust
> // in chain/network/src/routing/edge_set.rs
> 
> pub(crate) struct EdgeSet {
>   repr: HashSet<EdgeIndexedByKey>
> }
> 
> struct EdgeIndexedByKey(Edge);
> 
> impl EdgeSet {
>     pub(crate) fn insert(&mut self, edge: Edge) { ... }
>     pub(crate) fn get(&self, key: &(PeerId, PeerId)) -> Option<&Edge> { ... }
> }
> ```
> 
> The HashMap -> HashSet trick is a bit weird, so, if we could encapsulate it in a module, that'd be great. That way, `EdgeIndexedByKey` weirdness will be strictly contained to a single, small file.

@matklad Do you mean, I should put this in a separate file like `edge_set.rs`? Or did you mean something else?

> 
> Also curious to see if this affects any benchmark results!

I'll try to benchmark it again, and see how it goes. Is there anything else that should be done?

pmnoxx:
@matklad There was a small bug. `insert` function, didn't insert the value, if there already was a value. I had to add `remove` inside `insert`, which makes the code slower.

```
    pub(crate) fn insert(&mut self, edge: Edge) -> bool {
        self.remove(edge.key());
        self.repr.insert(EdgeIndexedByKey { inner: edge })
    }
```

chain/network/src/routing/edge.rs
```diff
@@ -5,6 +5,7 @@ use near_crypto::{KeyType, SecretKey, Signature};
 use near_primitives::borsh::maybestd::sync::Arc;
 use near_primitives::hash::CryptoHash;
 use near_primitives::network::PeerId;
+use std::hash::Hash;
```
mm-near, line 8: do we need this ?

chain/network/src/routing/routing_table_actor.rs
```diff
@@ -45,18 +46,20 @@ pub enum Prune {
 ///   - store removed edges to disk
 ///   - we currently don't store active edges to disk
 pub struct RoutingTableActor {
-    /// Data structure with all edges. It's guaranteed that `peer.0` < `peer.1`.
-    pub edges_info: HashMap<(PeerId, PeerId), Edge>,
+    /// Collection of edges representing P2P network.
+    /// It's indexed by `Edge::key()` key and can be search through by called `get()` function
+    /// with `(PeerId, PeerId)` as argument.
```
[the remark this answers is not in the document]
pmnoxx: @mm-near Yes, the current assumption is that keys in `edges_info` meet that condition.

pmnoxx:
@matklad I'll do benchmarks after https://github.com/near/nearcore/pull/5791

matklad:
> had to add remove inside insert, which makes the code slower.

As a meta observation, Rust standard library generally has orthogonal API, and "stupid" work-arounds like doing two hash map lookups shouldn't be necessary. In this case, what you need is replace: 

file:///home/matklad/.rustup/toolchains/beta-x86_64-unknown-linux-gnu/share/doc/rust/html/std/collections/struct.HashSet.html#method.replace

pmnoxx:
> > had to add remove inside insert, which makes the code slower.
> 
> As a meta observation, Rust standard library generally has orthogonal API, and "stupid" work-arounds like doing two hash map lookups shouldn't be necessary. In this case, what you need is `HashSet::replace`

Thanks, that's a good suggestion.
Sometimes, it's easy to miss some feature in the api.

pmnoxx:
@mm-near @matklad I'm think of closing this PR, in favor of https://github.com/near/nearcore/pull/5955

pmnoxx:
@matklad Maybe, this deserves a separate, crate, or even a repo?

pmnoxx:
```
```
 Can you take a look?

pmnoxx:
```
```
 Can you take a look?

Issue #4951, opened 2021-10-07
Compress genesis.json files offered for download


mina86:
testnet’s genesis.json available on S3 is almost 5.5GB in size. After compressing it with xz its size reduces to around 170MB. Betanet genesis.json is smaller but at 47MB it could still benefit from compression.

We should compress the files and update `nearcore::init_config` to transparently handle decompression.

PR #5477, opened 2021-11-26
nearcore: use tempfile to create temporary files when downloading

mina86:
We are already using tempfile crate in other parts of the code so
use it in download_file as well rather than having an arguably less
safe way of creating temporary files (i.e. blindly using predefined
extension).

### DIFF
--- a/nearcore/src/config.rs
+++ b/nearcore/src/config.rs
@@ -1116,40 +1116,35 @@ pub enum FileDownloadError {
 }
 
 pub fn download_file(url: &str, path: &Path) -> Result<(), FileDownloadError> {
-    return actix::System::new().block_on(async move {
+    let uri = url.parse()?;
+    let (tmp_file, tmp_path) = {
+        let tmp_dir = path.parent().unwrap_or(&Path::new("."));
+        tempfile::NamedTempFile::new_in(tmp_dir).map_err(FileDownloadError::OpenError)?.into_parts()
+    };
+
+    let result = actix::System::new().block_on(async move {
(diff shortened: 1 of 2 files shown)


nearcore/src/config.rs
```diff
@@ -1116,40 +1116,35 @@ pub enum FileDownloadError {
...
 
 pub fn download_file(url: &str, path: &Path) -> Result<(), FileDownloadError> {
-    return actix::System::new().block_on(async move {
+    let uri = url.parse()?;
+    let (tmp_file, tmp_path) = {
+        let tmp_dir = path.parent().unwrap_or(&Path::new("."));
+        tempfile::NamedTempFile::new_in(tmp_dir).map_err(FileDownloadError::OpenError)?.into_parts()
+    };
+
+    let result = actix::System::new().block_on(async move {
```
matklad, line 1125: Unrelated to the PR, but I think it doesn't make sense that we use `actix` here -- it's purely tokio stuff, it doesn't involve neither actix actors, nor actix-web http server. 

I *think* a more frugal way to do this is 

```rust
tokio::runtime::Builder::new_current_thread()
        .enable_all()
        .build()
        .unwrap()
        .block_on(async ...)
```

matklad: docs: https://tokio.rs/tokio/topics/bridging

mm-near (approved):

nearcore/src/config.rs
mina86: Hmm, let’s do that in another PR so we have one change per commit.

PR #5478, opened 2021-11-26
chain: fix nightly expensive tests after tests move to /src/tests

mina86:
First of all, add ‘gc’ to the ‘tests’ module since it was omitted and
thus completely ignored.

Second of all, since now all the tests are compiled into a single
binary, update executable name in the nightly configuration file.

Issue: https://github.com/near/nearcore/issues/5371

### DIFF
--- a/chain/chain/src/tests/gc.rs
+++ b/chain/chain/src/tests/gc.rs
@@ -2,10 +2,10 @@
 mod tests {
     use std::sync::Arc;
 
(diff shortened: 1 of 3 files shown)


mina86:
http://nayduck.near.org/#/run/2252

mm-near (approved):

pmnoxx (commented):
LGTM

PR #5479, opened 2021-11-26
re: `near-network` Rename `Consolidate` to `RegisterPeer`

pmnoxx:
In the current design. `PeerManagerActor`` is responsible for maintaining list of all connected peers.
Whenever `PeerActor` receives `Handshake` from other node. It notifies `PeerManagerActor` to add given `PeerActor` to list of all active connections.

Currently, that message is called `Consolidate`. `Consolidate` can have multiple meanings. Let's change it's name to `RegisterPeer` message. 
And also response that `PeerManagerActor` sends to `PeerActor`, about whenever registration was successful from `ConsolidateResponse` to `RegisterPeerResponse`.

This should make it easier to explain the process of `Handshake` negotiation inside https://github.com/near/nearcore/issues/5156.

### DIFF
--- a/chain/network/src/types.rs
+++ b/chain/network/src/types.rs
@@ -439,27 +439,28 @@ pub struct SendMessage {
     pub(crate) message: PeerMessage,
 }
 
-/// Actor message to consolidate potential new peer.
-/// Returns if connection should be kept or dropped.
+/// Actor message which asks `PeerManagerActor` to register peer.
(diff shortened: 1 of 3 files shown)


chain/network/src/types.rs
```diff
@@ -442,7 +442,7 @@ pub struct SendMessage {
 /// Actor message to consolidate potential new peer.
```
mm-near, line 442: update the comment

mm-near (approved):

Issue #3779, opened 2021-01-05
Tracking issue for GHSA-2v6r-g342-282f


bowenwang1996:
Tracks security advisory issue https://github.com/near/nearcore/security/advisories/GHSA-2v6r-g342-282f

PR #5482, opened 2021-11-26
fix: set access key nonce on implicit account creation

Longarithm:
Fix access key nonce for implicit accounts to fully address the issue https://github.com/near/nearcore/issues/3779.

The change is straightforward - formula for nonce should be the same as for explicit accounts, as here - https://github.com/near/nearcore/blob/7f7347fc008d592a2679101f4518cfdb8df4e617/runtime/runtime/src/actions.rs#L561-L562

## Test plan

- `test_transaction_hash_collision_for_implicit_account_[ok|fail]` - check that tx is applied or rejected depending on the protocol version.


matklad (approved):

integration-tests/tests/client/process_blocks.rs
```diff
@@ -3331,6 +3331,106 @@ mod access_key_nonce_range_tests {
...
+            env.produce_block(0, i);
+        }
+
+        // Delete implicit account.
+        env.clients[0].process_tx(delete_account_tx, false, false);
+        // Send money to implicit account again, invoking its second creation.
+        env.clients[0].process_tx(send_money_again_tx, false, false);
+        // Send money from implicit account with incorrect nonce.
+        env.clients[0].process_tx(send_money_from_implicit_account_tx, false, false)
+    }
```
bowenwang1996, line 3407: I think this is not entirely correct: what we did here will lead to all the transactions getting inserted into mempool at the same time. This means that they will end up in the same chunk and may result in `send_money_from_implicit_account_tx` getting processed first.

integration-tests/tests/client/process_blocks.rs
```diff
@@ -3331,6 +3331,106 @@ mod access_key_nonce_range_tests {
...
+    #[test]
+    fn test_transaction_hash_collision_for_implicit_account_ok() {
+        #[cfg(feature = "protocol_feature_access_key_nonce_for_implicit_accounts")]
+        let protocol_version =
+            ProtocolFeature::AccessKeyNonceForImplicitAccounts.protocol_version() - 1;
+        #[cfg(not(feature = "protocol_feature_access_key_nonce_for_implicit_accounts"))]
+        let protocol_version = PROTOCOL_VERSION;
+        let res = get_result_transaction_hash_collision_for_implicit_account(protocol_version);
+        assert!(matches!(res, NetworkClientResponses::ValidTx));
+    }
```
bowenwang1996, line 3432: Could we also make sure that transactions that are supposed to succeed actually succeed?

runtime/runtime/src/actions.rs
```diff
@@ -392,13 +394,28 @@ pub(crate) fn action_implicit_account_creation_transfer(
     actor_id: &mut AccountId,
     account_id: &AccountId,
     transfer: &TransferAction,
+    #[cfg(feature = "protocol_feature_access_key_nonce_for_implicit_accounts")]
+    current_block_height: BlockHeight,
```
bowenwang1996, line 398: Might want to make this function unconditionally take `current_block_height` as an argument to make it easier to read and use

integration-tests/tests/client/process_blocks.rs
```diff
@@ -95,13 +95,12 @@ pub fn create_nightshade_runtimes(genesis: &Genesis, n: usize) -> Vec<Arc<dyn Ru
         .collect()
 }
 
-fn produce_epochs(
+fn produce_blocks_from_height(
```
mm-near, line 98: comment

integration-tests/tests/client/process_blocks.rs
```diff
@@ -3331,6 +3346,112 @@ mod access_key_nonce_range_tests {
...
+        // Send money from implicit account with incorrect nonce.
+        let send_money_from_implicit_account_tx = SignedTransaction::send_money(
+            1,
+            implicit_account_id.clone(),
+            "test0".parse().unwrap(),
+            &implicit_account_signer,
+            100,
+            *block.hash(),
+        );
+        env.clients[0].process_tx(send_money_from_implicit_account_tx, false, false)
```
mm-near, line 3426: it'd be also good to check below, if we can still send the transaction with valid nonce.

integration-tests/tests/client/process_blocks.rs
```diff
@@ -3331,6 +3346,112 @@ mod access_key_nonce_range_tests {
...
+    }
+
+    /// Test that duplicate transactions from implicit accounts are properly rejected.
+    #[cfg(feature = "protocol_feature_access_key_nonce_for_implicit_accounts")]
+    #[test]
+    fn test_transaction_hash_collision_for_implicit_account_fail() {
+        let protocol_version =
+            ProtocolFeature::AccessKeyNonceForImplicitAccounts.protocol_version();
+        assert!(matches!(
```
mm-near, line 3435: one problem I see with such setup - is that you don't know which transaction has thrown this error..

mm-near: what you could do instead is:  have get_status_of_tx_hash_collision_for_implicit_account return the transaction - and then execute this single transaction here - and check the assert.

mm-near (approved):

chain/chain/Cargo.toml
```diff
@@ -47,5 +47,5 @@ protocol_feature_chunk_only_producers = ["near-primitives/protocol_feature_chunk
 
 protocol_feature_routing_exchange_algorithm = []
 nightly_protocol_features = ["nightly_protocol", "protocol_feature_chunk_only_producers", "protocol_feature_routing_exchange_algorithm"]
-nightly_protocol = []
+nightly_protocol = ["near-store/nightly_protocol", "near-primitives/nightly_protocol"]
```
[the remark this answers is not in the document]
Longarithm: Not strictly, but I couldn't run the `simple_chain` test without changes in this file.

integration-tests/tests/client/process_blocks.rs
Longarithm: > one problem I see with such setup - is that you don't know which transaction has thrown this error..

I know. All other txs are executed and checked using `check_tx_processing`, and for the tx being tested I return the status, which is asserted in the `test_transaction_hash_collision_for_implicit_account_fail`.

PR #5484, opened 2021-11-26
nearcore: don’t bring in Actix when Tokio is enough while dling a file

mina86:
The download_file function does not use any of the Actix features so
enabling it is unnecessary.  A less complex and heavy-handed solution
is to set up a Tokio runtime so do that.


matklad (approved):
to clarify, the main problem here is not resource usage, but just the general sanity of not using an unrelated actor framework when what you need is `epoll`.

bowenwang1996 (approved):

PR #5486, opened 2021-11-26
chain: remove nested `tests` modules

mina86:
After moving tests from `/tests` to `/src/tests` and thus bringing
them all into a single executable with a single namespace we ended up
with aesthetically unpleasing nested paths such as:

    tests::catching_up::tests::test_chunk_grieving

Flatten the structure by removing the inner `tests` module.

Issue: https://github.com/near/nearcore/issues/5371

### DIFF
--- a/chain/chain/src/tests/doomslug.rs
+++ b/chain/chain/src/tests/doomslug.rs
@@ -1,334 +1,326 @@
-#[cfg(test)]
-#[cfg(feature = "expensive_tests")]
-mod tests {
-    use near_primitives::time::Clock;
-    use rand::{thread_rng, Rng};
-    use std::collections::{HashMap, HashSet};
-    use std::sync::Arc;
-    use std::time::{Duration, Instant};
 
-    use crate::{Doomslug, DoomslugThresholdMode};
-    use near_crypto::{KeyType, SecretKey};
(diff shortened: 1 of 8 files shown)


matklad:
This'll probably get some "unused" warnings due to the way cfg work. The fix is to put `#![cfg(expensive_tests)]` at the start of files, or `#[cfg(expensive_tests)] mod gc;` at the corresponding declaration in the parent file. 

Semi-related thought: I generally think that it's best to gate slow tests at runtime, rather than at compile time, to make sure they actually compile OK, and to avoid the need to recompile everything when you want to run them. But I haven't looked closely if that'd work for us (there's no built-in mechanism to do runtime gating in Cargo, so you'd have to do something manual like `if std::env::var("SLOW_TESTS").is_err() { return }`).

mina86:
Depending on how motivated we are we could create a `#[expensive_test]` procedural macro I suppose. However, one issue would still remain that it would be a bit confusing to see a test that did not run as passing. It doesn’t look like cargo test supports ignoring tests based on run-time state. `#[ignore]` is compile-time thing only.

matklad:
Yeah, exactly this

bowenwang1996 (approved):

PR #5487, opened 2021-11-26
Use the CPU features we require across neard and the runtime

nagisa:
This changeset does a couple different things. First and foremost it
makes `neard` verify the environment it is running in. It'll do so only
for production builds by default, making regular `cargo run` or `cargo
build` workflows work as usual.

However in production artifacts (as produced by `make release`, the
`cfg(near_production)` will be set, various CPU features will be enabled
for LLVM codegen purposes and the `neard` itself will verify that the
prerequisite features are available before it does anything else.

Note however, that there is no guarantee that the code, which runs before
`main` is invoked, avoids use of these potentially unsupported features,
possibly leading to less than stellar failure modes on machines which
wouldn't be able to run `neard` either way.

This may have an effect on gas costs for host functions, and for that
reason this change must not be merged before we quantify those changes.
As a less obvious outcome, this may also cause cost estimations done on
the production build differ from cost estimations done with a regular
`cargo built` code.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,7 +22,7 @@ release: RUSTFLAGS+= -Ctarget-feature=+avx,+cmpxchg16b,+popcnt,+sse3,+ssse3,+sse
 release: CFLAGS = -mavx -mpopcnt -msse3 -mssse3 -msse4.1 -msse4.2 -msse4a -mcx16
 release: CXXFLAGS = ${CFLAGS}
 release:
-	cargo build -p neard --release
+	cargo build -p neard --release --features cpu_compatibility_checks
 	cargo build -p state-viewer --release
 	cargo build -p store-validator --release
 	cargo build -p runtime-params-estimator --release
--- a/Makefile
+++ b/Makefile
@@ -18,6 +18,9 @@ docker-nearcore-nightly:
 	docker build -t nearcore-nightly -f Dockerfile --build-arg=features=nightly_protocol,nightly_protocol_features --progress=plain .
 
 release: NEAR_RELEASE_BUILD=release
+release: RUSTFLAGS+= -Ctarget-feature=+avx,+cmpxchg16b,+popcnt,+sse3,+ssse3,+sse4.1,+sse4.2,+sse4a
+release: CFLAGS = -mavx -mpopcnt -msse3 -mssse3 -msse4.1 -msse4.2 -msse4a -mcx16
+release: CXXFLAGS = ${CFLAGS}
 release:
 	cargo build -p neard --release
 	cargo build -p state-viewer --release
--- a/core/primitives/src/cpu.rs
+++ b/core/primitives/src/cpu.rs
@@ -1,41 +1,93 @@
+#[derive(Debug)]
+pub enum Error {
+    /// Unsupported architecture
+    Architecture,
+    /// The following required features are not supported by the CPU.
+    MissingFeatures(Vec<&'static str>),
+    /// SSE is available but disabled.
+    SSEDisabled,
+    /// AVX is available but disabled.
+    AVXDisabled,
+}
+
+impl Error {
+    pub const NOTE: &'static str = "confirm hardware requirements at \
+        <https://docs.near.org/docs/develop/node/validator/hardware>.";
+
+    pub fn terminate(&self) -> ! {
+        eprintln!("error: {}\nnote: {}", self, Self::NOTE);
+        std::process::exit(1);
+    }
+}
+
+impl std::error::Error for Error {}
+
+impl std::fmt::Display for Error {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        use Error::*;
+        match self {
+            Architecture => f.write_str("unsupported CPU architecture")?,
+            SSEDisabled => f.write_str("SSE CPU feature is supported but disabled")?,
+            AVXDisabled => f.write_str("AVX CPU feature is supported but disabled")?,
+            MissingFeatures(features) => {
+                f.write_str("required CPU features are missing: ")?;
+                let mut features = features.iter().peekable();
+                while let Some(feature) = features.next() {
+                    f.write_str(feature)?;
+                    if features.peek().is_some() {
+                        f.write_str(", ")?;
+                    }
+                }
+            }
+        }
+        Ok(())
+    }
+}
+
 /// Try to ensure that the required CPU features are available and enabled for the current CPU.
+///
+/// If any of the features are not available, the returned error will specify which ones are
+/// missing.
 ///
 /// Note that this is a best-effort kind of thing. If these features are already enabled when
 /// compiling the program, then we may very well end up executing some unsupported instruction
 /// before we even get to executing this code.
-pub fn ensure_cpu_compatibility(force: bool) {
-    if cfg!(near_production) || force {
-        imp::ensure_cpu_compatibility();
-    } else {
-        // Intentionally empty: allow execution without verifying the properties of the CPU.
-    }
+pub fn verify_cpu_compatibility() -> Result<(), Error> {
+    imp::verify_cpu_compatibility()
 }
 
 mod imp {
-    const NOTE: &str = "confirm hardware requirements at \
-        <https://docs.near.org/docs/develop/node/validator/hardware>.";
-
     #[cfg(not(target_arch = "x86_64"))]
-    pub fn ensure_cpu_compatibility() {
-        eprintln!("error: processor running a x86_64 instruction set is required.\nnote: {}", NOTE);
-        std::process::exit(1);
+    pub fn verify_cpu_compatibility() -> Result<(), super::Error> {
+        Err(super::Error::Architecture)
     }
 
     #[cfg(target_arch = "x86_64")]
-    pub fn ensure_cpu_compatibility() {
-        // SAFETY: this is “just” a CPUID. Some very ancient chips don't support this instruction
-        // and may do anything else with the opcode.
-        //
-        // As an implementation note, we cannot use `is_*_feature_available!` macro from libstd
-        // here, because it will unconditionally return `true` when `cfg(target_feature="...")` is
-        // true.
-        let cpuid01 = unsafe { std::arch::x86_64::__cpuid(0x01) };
-        ensure_cpuid01(cpuid01.ecx, cpuid01.edx);
-        let cpuid0d = unsafe { std::arch::x86_64::__cpuid(0x0d) };
-        ensure_cpuid0d(cpuid0d.eax);
+    pub fn verify_cpu_compatibility() -> Result<(), super::Error> {
+        ensure_cpu_info()?;
+        ensure_cpu_state()
     }
 
-    fn ensure_cpuid01(ecx: u32, edx: u32) {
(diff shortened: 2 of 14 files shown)


runtime/near-vm-runner/src/wasmer2_runner.rs
```diff
@@ -375,16 +375,7 @@ impl crate::runner::VM for Wasmer2VM {
...
-        // So doesn't require x86. However, when it is on x86, AVX is required:
-        // https://github.com/wasmerio/wasmer/issues/1567
-        #[cfg(not(feature = "no_cpu_compatibility_checks"))]
-        if (cfg!(target_arch = "x86") || !cfg!(target_arch = "x86_64"))
-            && !is_x86_feature_detected!("avx")
-        {
-            panic!("AVX support is required in order to run Wasmer VM Singlepass backend.");
-        }
-
+        near_primitives::cpu::ensure_cpu_compatibility(cfg!(feature = "cpu_compatibility_checks"));
```
matklad, line 378:
```suggestion
        let force = cfg!(feature = "cpu_compatibility_checks");
        near_primitives::cpu::ensure_cpu_compatibility(force);
```

as much as I enjoy smart IDEs which show parameter names automagically, I feel naming them is still better :)

runtime/runtime/Cargo.toml
```diff
@@ -33,11 +33,11 @@ near-vm-runner = { path = "../../runtime/near-vm-runner" }
 near-vm-errors = { path = "../../runtime/near-vm-errors" }
 
 [features]
-default = []
+default = ["cpu_compatibility_checks"]
```
matklad, line 36: Does this do anything? we depend on `vm-runner` with default features anyway, no?

core/primitives/src/cpu.rs
```diff
@@ -0,0 +1,97 @@
+/// Try to ensure that the required CPU features are available and enabled for the current CPU.
+///
+/// Note that this is a best-effort kind of thing. If these features are already enabled when
+/// compiling the program, then we may very well end up executing some unsupported instruction
+/// before we even get to executing this code.
+pub fn ensure_cpu_compatibility(force: bool) {
+    if cfg!(near_production) || force {
+        imp::ensure_cpu_compatibility();
+    } else {
+        // Intentionally empty: allow execution without verifying the properties of the CPU.
```
matklad, line 10: Perhaps it's better to `tracing::error` here?

matklad: Or maybe return a `Result` out of this function and let the caller decide between log/process::exit/ignore ?

Makefile
```diff
@@ -1,7 +1,9 @@
 export CARGO_PROFILE_RELEASE_CODEGEN_UNITS = 1
 export CARGO_PROFILE_RELEASE_LTO = fat
 export DOCKER_BUILDKIT = 1
-export RUSTFLAGS = -D warnings
+export CFLAGS = -mavx -mpopcnt -msse3 -mssse3 -msse4.1 -msse4.2 -msse4a -mcx16
+export CXXFLAGS = ${CFLAGS}
+export RUSTFLAGS = -D warnings --cfg near_production -Ctarget-feature=+avx,+cmpxchg16b,+popcnt,+sse3,+ssse3,+sse4.1,+sse4.2,+sse4a
```
matklad, line 6: I wonder if we need the cfgs? When we control `near_production`, we also control cargo feature flags, so we *know* that `cpu_compatability_checks` is set. That is, perhaps just add `cpu_compatibility_checks` to `neard` crate?

matklad (commented):
This is pretty exciting! 

It indeed makes sense to actually fully use the CPU features we require anyway :-)

matklad:
> this may also cause cost estimations done on
> the production build differ from cost estimations done with a regular
> cargo built code.

This is already very much the case. Specifically, `lto` affects the performance **a lot** for our code base, so param estimator already includes logic to build the special build of nearcore for estimation: 

https://github.com/near/nearcore/blob/acaa632eb6a4ac80dba344e715c2fda30aa3cf43/runtime/runtime-params-estimator/src/main.rs#L280-L281

Would be much nicer with custom cargo profiles.

core/primitives/src/cpu.rs
```diff
@@ -0,0 +1,97 @@
...
+        // As an implementation note, we cannot use `is_*_feature_available!` macro from libstd
+        // here, because it will unconditionally return `true` when `cfg(target_feature="...")` is
+        // true.
+        let cpuid01 = unsafe { std::arch::x86_64::__cpuid(0x01) };
+        ensure_cpuid01(cpuid01.ecx, cpuid01.edx);
+        let cpuid0d = unsafe { std::arch::x86_64::__cpuid(0x0d) };
+        ensure_cpuid0d(cpuid0d.eax);
+    }
+
+    fn ensure_cpuid01(ecx: u32, edx: u32) {
```
pmnoxx, line 38: A link to documentation, where those bits are defined would be nice.
This would make it easier to check, whenever all bits are set correctly while doing the review.

core/primitives/src/cpu.rs
```diff
@@ -0,0 +1,97 @@
...
+
+    #[cfg(not(target_arch = "x86_64"))]
+    pub fn ensure_cpu_compatibility() {
+        eprintln!("error: processor running a x86_64 instruction set is required.\nnote: {}", NOTE);
+        std::process::exit(1);
+    }
+
+    #[cfg(target_arch = "x86_64")]
+    pub fn ensure_cpu_compatibility() {
+        // SAFETY: this is “just” a CPUID. Some very ancient chips don't support this instruction
```
olonho, line 26: Not only chips, it's known that emulators (at least qemu and macOS Rosetta) doesn't provide cpuid compatible with at least Rust AVX checks.

core/primitives/src/cpu.rs
```diff
@@ -0,0 +1,97 @@
...
+        // and may do anything else with the opcode.
+        //
+        // As an implementation note, we cannot use `is_*_feature_available!` macro from libstd
+        // here, because it will unconditionally return `true` when `cfg(target_feature="...")` is
+        // true.
+        let cpuid01 = unsafe { std::arch::x86_64::__cpuid(0x01) };
```
olonho, line 32: Looks not very readable, maybe better smth like `LEAF_PROCESSOR_INFO`.

core/primitives/src/cpu.rs
```diff
@@ -0,0 +1,97 @@
...
+        ensure_cpuid01(cpuid01.ecx, cpuid01.edx);
+        let cpuid0d = unsafe { std::arch::x86_64::__cpuid(0x0d) };
```
olonho, line 34: Per manuals (https://www.felixcloutier.com/x86/cpuid) leaf 0xd varies depending on ECX value. Better set it.
Looks not very readable, maybe better smth like `LEAF_PROCESSOR_EXTENDED_STATE_ENUMERATION`.

olonho:
Overall, this change seems better be somewhere in Rust stdlib maybe? And remaining part shall be delegated to Wasmer code I presume, as its JIT decides which instructions to use.

nagisa:
> Overall, this change seems better be somewhere in Rust stdlib maybe?

What parts do you mean by “this” exactly? The reason `is_*_feature_detected(feat)` cannot be used here is because it will unconditionally return `true` if `cfg!(target_feature = "feat")` is true. This `cfg` will get set to values enabled via `-Ctarget-feature` which would incapacitate the `is_x86_feature_detected`-based checks we had so far. I'm not sure where the standard library comes into play otherwise.

> And remaining part shall be delegated to Wasmer code I presume, as its JIT decides which instructions to use.

Wasmer making sure it does not emit instructions for unavailable instructions would indeed be awesome, but that sounds like (much) more work. And quite orthogonal to this PR, which enables use of the CPU features as per our hardware requirements across the codebase.

core/primitives/src/cpu.rs
nagisa: `__cpuid` always queries the main sub-leaf (i.e. `ecx` is set to 0)

 https://github.com/rust-lang/stdarch/blob/cfba59fccd90b3b52a614120834320f764ab08d1/crates/core_arch/src/x86/cpuid.rs#L90-L96

core/primitives/src/cpu.rs
nagisa: Do you have any more information? [Rosetta does not support AVX at all](https://developer.apple.com/documentation/apple-silicon/about-the-rosetta-translation-environment), but there doesn't seem to be any indications suggesting that `cpuid` returns incorrect information.

nagisa:
As an update, wasmer 2.1.0 is being pretty careful at validating the required features it needs at the module instantiation time, so a part of the problem is pretty much solved for us once we merge that in. I think we'll still want some cpuid checks in neard, but at the very least we no longer need to verify that for the wasmer runtime.

pmnoxx:
Specifying minimum cpu requirement is a good idea. It's better to get a message, saying that the CPU is not compatible, rather than spending a bit of time debugging, on why our don't work.

I think we should do it for convenience.

nagisa:
Go away, bot. I will get this _one day_.

core/primitives/src/cpu.rs
```diff
@@ -0,0 +1,158 @@
...
+/// missing.
+///
+/// Note that this is a best-effort kind of thing. If these features are already enabled when
+/// compiling the program, then we may very well end up executing some unsupported instruction
+/// before we even get to executing this code.
+pub fn verify_cpu_compatibility() -> Result<(), Error> {
+    imp::verify_cpu_compatibility()
+}
+
+mod imp {
```
mina86, line 59: It looks like you want:

```rust
#[cfg(not(target_arch = "x86_64"))]
mod impl {
   // ...
}

#[cfg(target_arch = "x86_64")]
mod impl {
   // ...
}
```

and then the comment about CPUID being unsupported by some
architectures is moot.

Makefile
```diff
@@ -18,8 +18,11 @@ docker-nearcore-nightly:
 	docker build -t nearcore-nightly -f Dockerfile --build-arg=features=nightly_protocol,nightly_protocol_features --progress=plain .
 
 release: NEAR_RELEASE_BUILD=release
```
mina86, line 20: There’s also this environment variable which we’re using to check if release binary is used on mainnet. We probably should convert it into use of a feature as well. So maybe `release_build` feature in `neard`?

mina86 (approved):

bowenwang1996:
@nagisa what is the status of this PR?

nagisa:
This is still largely blocked over reaching a decision on whether `cpuid` is something we should/can be using to verify feature support (there were mentions of it not working right in Rosetta, which I'm not in a position to verify.)

bowenwang1996:
@nagisa not that this is very important, but how do you think we should move forward?

nagisa:
The upstream Wasmer 2.2 singlepass has changes that obviate the requirement to have AVX available, which is the one feature that has been causing most of the concerns so far. I think it makes a lot of sense to integrate that first and then land this mostly as is, with anything AVX removed from the equation.

That way anybody running emulation (qemu,rosetta) would be able to still run the code, and for people who have AVX available, it would get detected and enabled at runtime.

akhi3030:
@nagisa: given the age of this PR and latest activity, I think it is safe to close it.  Please reopen if and when it is ready to make progress.

PR #5489, opened 2021-11-26
re: Rename `SyncData` to `RoutingTableSync` and add in code documentation

pmnoxx:
Rename `SyncData` to `RoutingTableUpdate` and add in code documentation.

Currently, we use `SyncData` struct to represent the message inside `PeerMessage::RoutingTableSync`.
To make things less confusing, let's use the same name. And also let's add documentation to `SyncData` .


chain/network/src/types.rs
```diff
@@ -263,25 +263,48 @@ impl From<HandshakeV2> for Handshake {
...
     }
 
+    /// Is empty.
     pub(crate) fn is_empty(&self) -> bool {
         self.edges.is_empty() && self.accounts.is_empty()
     }
+
+    pub(crate) fn unpack(self) -> (Vec<Edge>, Vec<AnnounceAccount>) {
+        (self.edges, self.accounts)
+    }
```
matklad, line 307: Not sure if encapsulating fields makes sense *in this case*. The struct is `derive(BorshSerialize)` so the fact that it is a pair of two vecs is exposed anyway.

chain/network/src/types.rs
```diff
@@ -263,25 +263,48 @@ impl From<HandshakeV2> for Handshake {
...
+        Self { edges, accounts }
+    }
+
+    pub(crate) fn from_edges(edges: Vec<Edge>) -> Self {
+        Self { edges, accounts: Default::default() }
+    }
+
+    pub(crate) fn from_accounts(accounts: Vec<AnnounceAccount>) -> Self {
+        Self { edges: Default::default(), accounts }
     }
```
matklad, line 288: When defining factory functions, I prefer to delegate to make it clear which constructor is primary: 

```suggestion
    pub(crate) fn from_edges(edges: Vec<Edge>) -> Self {
        Self::new(edges, Vec::new())
    }

    pub(crate) fn from_accounts(accounts: Vec<AnnounceAccount>) -> Self {
        Self::new(Vec::new(), accounts)
    }
```

useful to eg, be able to add creation debug prints in one place, and too more clearly see casese where helper factorise do genuienly different things.

chain/network/src/types.rs
pmnoxx: I make those fields `pub(crate)`

chain/network/src/types.rs
pmnoxx: Thanks, that's a good idea. I also removed `from_edge`, and `from_account` constructors to make code cleaner.

mzhangmzz (approved):

chain/network/src/types.rs
```diff
@@ -813,9 +817,9 @@ pub enum NetworkRequests {
     /// (Unit tests) Fetch current routing table.
     FetchRoutingTable,
     /// Data to sync routing table from active peer.
-    Sync {
+    SyncRoutingTable {
```
mm-near, line 820: so we have enum value and struct with the same name? that might be a little bit confusing

mm-near (approved):

chain/network/src/types.rs
pmnoxx: No, the name is different
```
pub enum PeerMessage {
    RoutingTableSync(SyncRoutingTable),
}
```

pmnoxx: @mm-near I renamed to the struct to be called `RoutingTableUpdate`.

PR #5490, opened 2021-11-27
re: Add documentation to `near-network::ValidateEdgeList`

pmnoxx:
`ValidateEdgeList` doesn't have documentation. It will be easier to explain what it does once we add it.

That's useful for https://github.com/near/nearcore/issues/5156

### DIFF
--- a/chain/network/src/types.rs
+++ b/chain/network/src/types.rs
@@ -838,10 +838,12 @@ pub enum NetworkRequests {
     },
 }
 
+/// List of `Edges`, which we received from `source_peer_id` gor purpose of validation.
+/// Those are list of edges received through `NetworkRequests::Sync` or `NetworkRequests::IbfMessage`.
 pub struct ValidateEdgeList {
     /// List of Edges, which will be sent to `EdgeValidatorActor`.
     pub(crate) edges: Vec<Edge>,
-    /// A set of edges, which have been verified. This is used to avoid doing duplicated work by
+    /// A set of edges, which have been verified. This is a cache with all verified edges.
     /// `EdgeValidatorActor`, and is a source of memory leak.
     /// TODO(#5254): Simplify this process.
     pub(crate) edges_info_shared: Arc<Mutex<HashMap<(PeerId, PeerId), u64>>>,
@@ -853,7 +855,8 @@ pub struct ValidateEdgeList {
     #[cfg(feature = "test_features")]
     /// Feature to disable edge validation for purpose of testing.
     pub(crate) adv_disable_edge_signature_verification: bool,
-    /// Peer that may be banned if any of the edges are found to be invalid.
+    /// The list of edges is provided by `source_peer_id`, that peer will be banned
+    ///if any of these edges are invalid.
     pub(crate) source_peer_id: PeerId,
 }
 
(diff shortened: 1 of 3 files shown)


chain/network/src/types.rs
```diff
@@ -838,12 +838,22 @@ pub enum NetworkRequests {
 }
 
 pub struct ValidateEdgeList {
+    /// List of Edges, which will be sent to `EdgeVerifierActor`.
```
mm-near, line 841: I think we're calling it 'validate' now, right? (update here and in other places in the comments below)

chain/network/src/types.rs
```diff
@@ -838,12 +838,22 @@ pub enum NetworkRequests {
...
     pub(crate) edges: Vec<Edge>,
+    /// A set of edges, which have been verified. This is used to avoid doing duplicated work by
+    /// `EdgeVerifierActor`, and is a source of memory leak.
+    /// TODO(#5254): Simplify this process.
     pub(crate) edges_info_shared: Arc<Mutex<HashMap<(PeerId, PeerId), u64>>>,
+    /// A concurrent queue. After edge become validated. They will be sent from `EdgeVerifierActor` back to
```
mm-near, line 847: grammar: 'after edge become*s* validated, *it* will be ...'

chain/network/src/types.rs
```diff
@@ -838,12 +838,22 @@ pub enum NetworkRequests {
 }
 
 pub struct ValidateEdgeList {
```
mm-near, line 840: I'd also mention that this list is 'per peer update', right? (basically these are the edges that we've received from a given peer during a single update).

And consider moving source_peer_id to the top - to highlight this fact (as AFAIK this is not serialized anywhere)

chain/network/src/types.rs
```diff
@@ -838,12 +838,22 @@ pub enum NetworkRequests {
...
+    /// A concurrent queue. After edge become validated. They will be sent from `EdgeVerifierActor` back to
+    /// `PeerManagetActor`, and then send to `RoutingTableActor`. And then `RoutingTableActor`
+    /// will add them.
+    /// TODO(#5254): Simplify this process.
     pub(crate) sender: QueueSender<Edge>,
     #[cfg(feature = "test_features")]
+    /// Feature to disable edge validation for purpose of testing.
     pub(crate) adv_disable_edge_signature_verification: bool,
-    pub(crate) peer_id: PeerId,
+    /// Peer that may be banned if any of the edges are found to be invalid.
```
mm-near, line 855: I'd mention that this is the peer that provides us with this list of edges (and maybe be banned if they are not correct)

chain/network/src/types.rs
```diff
@@ -838,12 +838,22 @@ pub enum NetworkRequests {
 }
 
 pub struct ValidateEdgeList {
+    /// List of Edges, which will be sent to `EdgeVerifierActor`.
     pub(crate) edges: Vec<Edge>,
+    /// A set of edges, which have been verified. This is used to avoid doing duplicated work by
```
mm-near, line 843: I'd also mention that this is 'global cache'

mm-near (approved):

chain/network/src/types.rs
pmnoxx: Yes,  fixed this after rabase.

chain/network/src/types.rs
pmnoxx: Ok, I moved `source_peer_id` to the top.
I also mentioned that those are edges received through  `NetworkRequests::Sync` or `NetworkRequests::IbfMessage`.

PR #5495, opened 2021-11-27
feat: Change nayduck to use `python3` instead of `python`

pmnoxx:
On some operating systems, the default  `python` commands points to `python2`.
For example, on `Ubuntu 20.04`.

Let's switch `nayduck` to use `python3` by default.

Instead of running:
```
python3 ./scripts/nayduck.py
```
This will work as well:
```
./scripts/nayduck.py
```


mina86 (approved):

pmnoxx:
http://nayduck.near.org/#/run/2261

mina86:
NayDuck run isn’t really necessary for this commit since the change does not affect tests.

chefsale (approved):

PR #5496, opened 2021-11-27
feat: Allow running any test in `/pytest` from any folder

pmnoxx:
Currently in order to run `pytest/tests/sanity/routing_table_sync.py` locally you need to:
- Set your local directory to `<ROOT_DIR>/pytest`
- run test with `python3.7 tests/sanity/routing_table_sync.py`

After this `PR`, it will be possible to run the test from any folder.
For example directly from root directory of the repo.
```
./pytest/tests/sanity/routing_table_sync.py
```


pytest/lib/cluster.py
```diff
@@ -777,9 +777,11 @@ def spin_up_node_and_push(i, boot_node: BootNode):
     return nodes
 
 
+ROOT_DIR = os.path.dirname(os.path.dirname(os.path.dirname(os.path.realpath(__file__))))
+
 DEFAULT_CONFIG = {
     'local': True,
-    'near_root': os.environ.get("NEAR_ROOT", '../target/debug/'),
+    'near_root': os.environ.get("NEAR_ROOT", os.path.join(ROOT_DIR, 'target/debug')),
```
mina86, line 784: `pathlib` is more readable:

    REPO_DIR = pathlib.Path(__file__).resolve().parent.parent.parent
    NEAR_ROOT = os.environ.get('NEAR_ROOT) or (REPO_DIR / 'target/debug')

    'near_root': NEAR_ROOT

pytest/tests/sanity/routing_table_sync.py
```diff
@@ -1,3 +1,4 @@
+#!/usr/bin/env python3.7
```
mina86, line 1: I’d say just `python3`. We don’t wont it to stop working on systems with Python 3.8 or newer. I only have 3.9 and 3.10 on my system for example.

pytest/tests/sanity/routing_table_sync.py
```diff
@@ -6,12 +7,20 @@
...
 import base58
 import ed25519
 
-sys.path.append('lib')
+cur_dir = os.path.dirname(os.path.realpath(__file__))
+lib_dir = os.path.dirname(os.path.dirname(cur_dir))
+
+sys.path.append(os.path.join(lib_dir, "lib"))
+
+print(lib_dir)
```
mina86, line 22: Leftover debug print.

pytest/tests/sanity/routing_table_sync.py
```diff
@@ -6,12 +7,20 @@
...
+import time
 
 import base58
 import ed25519
 
-sys.path.append('lib')
+cur_dir = os.path.dirname(os.path.realpath(__file__))
+lib_dir = os.path.dirname(os.path.dirname(cur_dir))
+
+sys.path.append(os.path.join(lib_dir, "lib"))
```
mina86, line 20: Like before, I’d go with `pathlib`, e.g.: `sys.path.append(pathlib.Path(__file__).resolve().parent.parent / 'lib')`.

The bigger problem with this is that if we want to do this we should do it in all the tests.

pytest/tests/sanity/routing_table_sync.py
pmnoxx: @mina86 Tests don't work with python 3.8.
The command below fails:
```
>>>  python3.8 pytest/tests/sanity/routing_table_sync.py
/home/pmnoxx/repos/nearcore.master/pytest
  File "pytest/tests/sanity/routing_table_sync.py", line 25, in <module>
  File "/home/pmnoxx/repos/nearcore.master/pytest/lib/cluster.py", line 8, in <module>
    import rc
  File "/home/pmnoxx/.local/lib/python3.8/site-packages/rc/__init__.py", line 11, in <module>
    from rc.cache import Cache, CacheCluster
  File "/home/pmnoxx/.local/lib/python3.8/site-packages/rc/cache.py", line 4, in <module>
    from itertools import izip
ImportError: cannot import name 'izip' from 'itertools' (unknown location)
```
This works

```
>>>  python3.7 pytest/tests/sanity/routing_table_sync.py
```

I would prefer to use `#!/usr/bin/env python3` once we fix tests.
However, tests are broken with `python `3.8` and upwards.

mina86: That’s the wrong `rc` module. Try `python3.8 -m pip install -U -r pytest/requirements.txt`.

mina86: This is a bit confusing but you have the wrong `rc` installed. Looks like you have [rc](https://pypi.org/project/rc/) while we want [python-rc](https://pypi.org/project/python-rc/). 

```
python3.8 -m pip uninstall rc
python3.8 -m pip install -U -r pytest/requirements.txt
```
will solve your issue.

pytest/lib/cluster.py
```diff
@@ -777,9 +777,11 @@ def spin_up_node_and_push(i, boot_node: BootNode):
     return nodes
 
 
+ROOT_DIR = str(pathlib.Path(__file__).resolve().parents[2])
+
 DEFAULT_CONFIG = {
     'local': True,
-    'near_root': os.environ.get("NEAR_ROOT", '../target/debug/'),
+    'near_root': os.environ.get("NEAR_ROOT", os.path.join(ROOT_DIR, 'target/debug')),
```
mina86, line 784: Better still, drop `str` from initialisation of `ROOT_DIR` and use `str(ROOT_DIR / 'target/debug')` here.

pytest/tests/sanity/routing_table_sync.py
pmnoxx: @mina86 Thanks, that fixed the issue.

pytest/lib/cluster.py
```diff
@@ -780,10 +780,15 @@ def spin_up_node_and_push(i, boot_node: BootNode):
...
 DEFAULT_CONFIG = {
-    'local': True,
-    'near_root': os.environ.get("NEAR_ROOT", str(os.path.join(ROOT_DIR, 'target/debug'))),
-    'binary_name': 'neard',
-    'release': False,
+    'local':
+        True,
+    'near_root':
+        os.environ.get("NEAR_ROOT", str(os.path.join(ROOT_DIR,
+                                                     'target/debug'))),
```
mina86, line 787: I mean, just add `NEAR_ROOT` global variable at the top so this doesn’t have to wrap. ;)

pmnoxx: Thanks, done.

pmnoxx:

mina86 (approved):

chefsale (approved):

PR #5497, opened 2021-11-27
feat: Add `--dry-run/-d` optional argument to `nayduck.py`

pmnoxx:
Let's add `--dry-run` option to `nayduck.py`. This will allow us to see which tests would run in nay-duck, without actually running them.
```
>>>  python3 ./scripts/nayduck.py --dry-run | grep integration | head
expensive integration-tests test_catchup test_catchup
expensive integration-tests test_catchup test_catchup --features nightly_protocol,nightly_protocol_features
expensive integration-tests standard_cases rpc::test::test_access_key_smart_contract_testnet
expensive integration-tests standard_cases rpc::test::test_access_key_smart_contract_testnet --features nightly_protocol,nightly_protocol_features
expensive integration-tests standard_cases rpc::test::test_access_key_smart_contract_testnet
expensive integration-tests standard_cases rpc::test::test_access_key_smart_contract_testnet --features nightly_protocol,nightly_protocol_features
expensive integration-tests standard_cases rpc::test::test_add_access_key_function_call_testnet
expensive integration-tests standard_cases rpc::test::test_add_access_key_function_call_testnet --features nightly_protocol,nightly_protocol_features
expensive integration-tests standard_cases rpc::test::test_add_access_key_with_allowance_testnet
expensive integration-tests standard_cases rpc::test::test_add_access_key_with_allowance_testnet --features nightly_protocol,nightly_protocol_features
```

### DIFF
--- a/scripts/nayduck.py
+++ b/scripts/nayduck.py
@@ -20,6 +20,7 @@ import getpass
 import json
 import os
 import pathlib
+import shlex
 import subprocess
 import sys
 import typing
@@ -56,6 +57,11 @@ def _parse_args():
                         '-l',
                         action='store_true',
                         help='Run tests locally.')
+    parser.add_argument(
+        '--dry-run',
+        '-n',
+        action='store_true',
+        help='Prints list of tests to execute, without doing anything')
     args = parser.parse_args()
 
     return args
@@ -199,7 +205,7 @@ def _parse_timeout(timeout: typing.Optional[str]) -> typing.Optional[int]:
     return int(timeout) * mul
 
 
-def run_locally(tests):
+def run_locally(args, tests):
     for test in tests:
         # See nayduck specs at https://github.com/near/nayduck/blob/master/lib/testspec.py
         fields = test.split()
@@ -218,7 +224,8 @@ def run_locally(tests):
         message = f'Running ‘{"".join(fields)}’'
         if ignored:
             message = f'{message} (ignoring flags ‘{" ".join(ignored)}`)'
-        print(message)
+        if not args.dry_run:
+            print(message)
 
         if fields[0] == 'expensive':
             # TODO --test doesn't work
@@ -242,6 +249,11 @@ def run_locally(tests):
         else:
             print(f'Unrecognised test category ‘{fields[0]}’', file=sys.stderr)
             continue
+        if args.dry_run:
+            print('( cd {} && {} )'.format(
+                shlex.quote(str(cwd)),
+                ' '.join(shlex.quote(str(arg)) for arg in cmd)))
+            continue
         print("RUNNING COMMAND cwd=%s cmd = %s", (cwd, cmd))
         subprocess.check_call(cmd, cwd=cwd, timeout=_parse_timeout(timeout))
 
@@ -264,11 +276,17 @@ def run_remotely(args, tests):
     else:
         code = github_auth(code_path)
 
+    if args.dry_run:
+        for test in tests:
+            print(test)
+        return
+
     post = {
         'branch': args.branch or get_current_branch().strip(),
         'sha': args.sha or get_curent_sha().strip(),
         'tests': list(tests)
     }
+
     while True:
         print('Sending request ...')
         res = requests.post(NAYDUCK_BASE_HREF + '/api/run/new',
@@ -278,6 +296,7 @@ def run_remotely(args, tests):
             break
         print(f'{styles[0]}Unauthorised.{styles[2]}\n')
         code = github_auth(code_path)
+
     if res.status_code == 200:
         json_res = json.loads(res.text)
         print(styles[json_res['code'] == 0] + json_res['response'] + styles[2])
@@ -298,7 +317,7 @@ def main():
         tests = list(read_tests_from_file(pathlib.Path(args.test_file)))
 
     if args.run_locally:
-        run_locally(tests)
+        run_locally(args, tests)
     else:
         run_remotely(args, tests)
 


scripts/nayduck.py
```diff
@@ -50,6 +50,10 @@ def _parse_args():
                        '-i',
                        action='store_true',
                        help='Read test set from standard input.')
+    group.add_argument('--dry-run',
```
mina86, line 53: this should be added to `parser`, not `group`, i.e. `parses.add_argument(…)`. Options within `group` are mutually exclusive.

scripts/nayduck.py
```diff
@@ -50,6 +50,10 @@ def _parse_args():
...
+                       '-d',
```
mina86, line 54: Typically, short argument for `--dry-run` is actually `-n`.

mina86 (commented):
It is indeed something I’ve been meaning to add myself. I was thinking `-o --output` but dry-run also works.

pmnoxx:
@mina86 Thanks, fixed.

scripts/nayduck.py
```diff
@@ -50,6 +50,11 @@ def _parse_args():
                        '-i',
                        action='store_true',
                        help='Read test set from standard input.')
+    parser.add_argument(
+        '--dry-run',
+        '-n',
+        action='store_true',
+        help='Prints list of tests to execute, without doing anything')
```
mina86, line 57: With https://github.com/near/nearcore/pull/5500 maybe this should be renamed to `-p --print`, `-o --output` or something similar and put in a new mutually exclusive group together with `--run-locally`?

mina86, line 57: With https://github.com/near/nearcore/pull/5500 maybe this should be renamed to `-p --print`, `-o --output` or something similar and put in a new mutually exclusive group together with `--run-locally`?

mina86, line 57: With https://github.com/near/nearcore/pull/5500 maybe this should be renamed to `-p --print`, `-o --output` or something similar and put in a new mutually exclusive group together with `--run-locally`?

pmnoxx: @mina86 I think it would make sense to be able to combine those two options

`./scripts/nayduck.py --dry-run --run-locally` should print list of local commands to run.

mina86: Right, makes sense.

pmnoxx: @mina86 Hmm, we could rename it to `--output`. However, in that case `--output` name doesn't suggest it stops the execution.

How about, we don't run anything by default, and add an option `--run`?

mina86: I think your previous suggestion to make `--dru-run --run-locally` print the commands that would be executed makes sense.

pmnoxx: @mina86 

We have a few options on how to structure the nayduck interface.

Option 1
`./scripts/nayduck.py` // runs remotely
`./scripts/nayduck.py  --run-locally` // runs commands locally
`./scripts/nayduck.py  --dry-run` // prints list of tasks
`./scripts/nayduck.py  --help` // prints help
`./scripts/nayduck.py  --run-locally --dry-run` // print list of local commands

Option 2
```
`./scripts/nayduck.py` // prints help
`./scripts/nayduck.py --run` // runs commands remotely
`./scripts/nayduck.py --run-locally` // runs commands locally
`./scripts/nayduck.py --run --dry-run` // prints list of tasks
`./scripts/nayduck.py --run-locally --dry-run` // print list of local commands
```
where `--run` and `--run-locally` are mutually exclusive

Option 3
```
`./scripts/nayduck.py` // prints help
`./scripts/nayduck.py` --run // runs commands remotely
`./scripts/nayduck.py` --run-locally // runs commands locally
`./scripts/nayduck.py` --dry-run // prints list of tasks without running them
`./scripts/nayduck.py` --dry-run-locally // print list of local commands, that you can just copy paste to run a test
```
where `--run`,  `--run-locally` ,  `--dry-run`, `--print-local-commands` are mutually exclusive

Option 1 is where we will be at once we merge #5500 #5497

I like options 2 and 3. Option 2 may be easier to use, on the other hand option 3 may be easier to explain inside `--help`.
@mina86 What do you think

mina86: I generally try to avoid changing existing behaviours which is why to me the first option seems to be the best choice.

pmnoxx: @mina86 That's an excellent point. Option 1 is backward compatible, it's a safe choice.

scripts/nayduck.py
```diff
@@ -214,6 +220,9 @@ def run_locally(tests):
         else:
             print(f'Unrecognised test category ‘{fields[0]}’', file=sys.stderr)
             continue
+        if args.dry_run:
+            print(" ".join(cmd))
```
mina86, line 224: This needs quoting through `shlex.quote`. Also, probably inclusion of `cd` as in something like `print('( cd {} && {} )'.format(shleq.quote(str(cwd)), ' '.join(shlex.quote(str(arg)) for arg in cmd)))`.

scripts/nayduck.py
```diff
@@ -188,10 +193,11 @@ def run_locally(tests):
...
         ignored = fields[1:index]
         del fields[1:index]
-        message = f'Running ‘{"".join(fields)}’'
-        if ignored:
-            message = f'{message} (ignoring flags ‘{"".join(ignored)}`)'
-        print(message)
+        if not args.dry_run:
+            message = f'Running ‘{"".join(fields)}’'
+            if ignored:
+                message = f'{message} (ignoring flags ‘{"".join(ignored)}`)'
```
mina86, line 199: Since you’re touching those, perhaps also fix `" ".join(…)` in both places.

pmnoxx: Ok, fixed.

scripts/nayduck.py
pmnoxx: @mina86 Thanks, fixed.

nikurt (approved):

Issue #5498, opened 2021-11-27
Make `nayduck.py` easier to use.


pmnoxx:
There a few improvements we can make to nayduck:
- [x] https://github.com/near/nearcore/pull/5495
- [x] https://github.com/near/nearcore/pull/5497
- [x] https://github.com/near/nearcore/pull/5500

PR #5500, opened 2021-11-27
feat: Add `--run-locally` option to `nayduck.py`

pmnoxx:
Let's add `--run-locally` option to `nayduck.py`. This will allow to run any tests locally. Ideally I would like to use this this way,

```
./scripts/nayduck.py --dry-run | grep SOME_TEST | head -1 | ./scripts/nayduck.py --stdin --run-locally
```


pmnoxx:
@mina86 This is not a full implementation. Do you know where is the script, that parses `tests` field on `nayduck`?

scripts/nayduck.py
```diff
@@ -50,6 +50,10 @@ def _parse_args():
                        '-i',
                        action='store_true',
                        help='Read test set from standard input.')
+    group.add_argument('--run-locally',
```
mina86, line 53: `parser.add_argument`

scripts/nayduck.py
```diff
@@ -203,6 +207,24 @@ def main():
         'tests': list(tests)
     }
 
+    if args.run_locally:
```
mina86, line 210: This should go on line 195, before `code_path = …`. If we’re running locally there’s no need to have the auth code or get the current branch or sha. (This is actually similar to --dry-run added in the other commit which also could be done without getting the code and post data).

scripts/nayduck.py
```diff
@@ -203,6 +207,24 @@ def main():
...
+        for test in post["tests"]:
+            # TODO: use parses similar to nayduck, where is it?
```
mina86, line 212: https://github.com/near/nayduck/blob/master/lib/testspec.py
This is not the first time I wish there was a way to share code between this repo and nayduck.

scripts/nayduck.py
```diff
@@ -203,6 +207,24 @@ def main():
...
+            fields = test.split(" ")
```
mina86, line 213: `fields = test.split()` without an argument so that it works with consecutive white space and with any white space characters (not just space).

scripts/nayduck.py
```diff
@@ -203,6 +207,24 @@ def main():
...
+                if fields[1].startswith("--timeout="):
+                    del fields[1]
+
+                package = fields[1]
+                params = test.split('::')[-1].split(" ")
+                subprocess.check_call(
+                    ['cargo', 'test', '-p', package, '--tests'] + params)
+            else:
+                print("ignoring don't know how to run locally %s" % test)
+
```
mina86, line 227: At this point it’s time to split this into multiple functions. One for local_run (with this code) and the other for remote run (with code for getting the auth code, post data, and the `while True` loop below).

scripts/nayduck.py
```diff
@@ -203,6 +207,24 @@ def main():
...
+                # timeout field not implemented
+                if fields[1].startswith("--timeout="):
+                    del fields[1]
+
+                package = fields[1]
+                params = test.split('::')[-1].split(" ")
+                subprocess.check_call(
+                    ['cargo', 'test', '-p', package, '--tests'] + params)
+            else:
+                print("ignoring don't know how to run locally %s" % test)
```
mina86, line 226:
```
        index = 1
        while len(fields) > index and fields[index].starts_with('--'):
            index += 1

        ignored = fields[1:index]
        del fields[1:index]
        message = f'Running ‘{"".join(fields)}’'
        if ignored:
            message = f'{message}  (ignoring flags ‘{"".join(ignored)}’)
        print(message)

        if fields[0] == 'expensive':
            cmd = ('cargo', 'test', '-p', fields[1], '--test', fields[2],
                   '--features', 'expensive_tests', '--', '--exact',
                   *fields[3])
            cwd = None
        elif fields[0] in ('pytest', 'mocknet'):
            fields[0] = sys.executable
            cmd = fields
            cwd = 'pytest'
        else:
            continue
        subprocess.check_call(cmd, cwd=cwd)
```

I can add better handling of the ignored flags (--timeout, --skip-build, --release and --remote) later.

scripts/nayduck.py
pmnoxx: @mina86 Yes, that's a good idea.

scripts/nayduck.py
```diff
@@ -212,18 +232,37 @@ def main():
...
         code_path.write_text(code)
 
 
+def main():
+    try:
+        import colorama
+        styles = (colorama.Fore.RED, colorama.Fore.GREEN,
+                  colorama.Style.RESET_ALL)
+    except ImportError:
+        styles = ('', '', '')
```
mina86, line 252: This could be moved to `run_remotely`.

mina86 (approved):
LGTM

scripts/nayduck.py
pmnoxx: @mina86 I had to disable parts of the code:

1) I had to remove `--test`, to make "expensive" case work
```
                                                                                                                              
subprocess.CalledProcessError: Command '('cargo', 'test', '-p', 'near-client', '--test', 'near_client', '--features', 'expensive_tests', '--', '--exact', 'tests::catching_up::test_catchup_re
ceipts_sync_third_epoch')' returned non-zero exit status 101.
```

2) +"pytest" - we run tests inside 'pytest' folder.
Tests are inside `tests/sandbox/patch_state.py'`,
however we run 'pytest /sandbox/patch_state.py' command.


  File "./scripts/nayduck.py", line 269, in <module>
    main()
  File "./scripts/nayduck.py", line 263, in main
    run_locally(tests)
  File "./scripts/nayduck.py", line 209, in run_locally
    subprocess.check_call(cmd, cwd=cwd)
  File "/usr/lib/python3.8/subprocess.py", line 364, in check_call
    raise CalledProcessError(retcode, cmd)
subprocess.CalledProcessError: Command '['/usr/bin/python3', 'sandbox/patch_state.py', '--features', 'sandbox']' returned non-zero exit status 2.

mina86: Looks like `--test` only works if the test is in `tests` not `src/tests`.

And yes, sorry, for pytest case it should be:

            fields[0] = sys.executable
            fields[1] = os.path.join('tests', fields[1])
            cmd = fields
            cwd = 'pytest'

pmnoxx:
@mina86 Thanks, the fix for `pytest` works

chefsale (approved):

PR #5501, opened 2021-11-27
Fix expensive tests not compiling

pmnoxx:
While working on https://github.com/near/nearcore/pull/5500 I found out that tests from `expensive.txt` are not compaling.
Let's fix that.

`protocol_feature_chunk_only_producers` feature was not propagated properly.


mina86 (approved):

yordanmadzhunkov (approved):

mm-near (approved):

chain/chain/Cargo.toml
```diff
@@ -43,7 +43,7 @@ expensive_tests = []
 test_features = []
 delay_detector = ["delay-detector"]
 no_cache = ["near-store/no_cache"]
-protocol_feature_chunk_only_producers = ["near-primitives/protocol_feature_chunk_only_producers"]
+protocol_feature_chunk_only_producers = ["near-primitives/protocol_feature_chunk_only_producers", "near-primitives/protocol_feature_chunk_only_producers"]
```
EgorKulikov, line 46: Why do we depend on the same feature twice here?

EgorKulikov:
I cannot understand, why this is needed. In all but three places corresponding dependency already was there, and in those three places it is easy to see that it is in the closure. @matklad can you understand why test ls are not compiling without this?

matklad:
I haven't look at every feature here, but, eg

```
 cargo t -p near-client --no-run --features expensive_tests,nightly_protocol,nightly_protocol_features
```

indeed doesn't work for me on eda132fe2. The reason for that is that we enable a feature for `near-chain` in `client/Cargo.toml`, but not for `near-chain-configs` (and `near-chain/Cargo.toml` doesn't forward the feature). Still don't see a better way to handle this, short of just writing some ad-hoc tool which would make Cargo.tomls consistent automatically.

EgorKulikov:
> I haven't look at every feature here, but, eg
> 
> ```
>  cargo t -p near-client --no-run --features expensive_tests,nightly_protocol,nightly_protocol_features
> ```
> 
> indeed doesn't work for me on [eda132f](https://github.com/near/nearcore/commit/eda132fe27aae7eb5c77c31bb7de4994d499b300). The reason for that is that we enable a feature for `near-chain` in `client/Cargo.toml`, but not for `near-chain-configs` (and `near-chain/Cargo.toml` doesn't forward the feature). Still don't see a better way to handle this, short of just writing some ad-hoc tool which would make Cargo.tomls consistent automatically.

Yep, I agree we miss some links, I am just not sure this pr add any new links, but apparently it fix the tests, otherwise @pmnoxx would not submit this pr

PR #5502, opened 2021-11-27
Add shebang to `tests/sanity/restacked.py`

pmnoxx:
Let's add shebang to `tests/sanity/restacked.py`.
See https://github.com/near/nearcore/pull/5496


mina86:
We might just as well do it for all the files:

    find -name __init__.py -o \( -name \*.py -exec sed -i -e '/python3/b' -e '1i#!/usr/bin/python3' -- {} + \)

pmnoxx:

PR #5505, opened 2021-11-27
feat: WIP: Alternative way to do imports in python - example

pmnoxx:
In Python 3, implicit relative imports are forbidden, see https://www.python.org/dev/peps/pep-0328/ and https://docs.python.org/release/3.0.1/whatsnew/3.0.html#removed-syntax:

Please, take a look at the discussion https://stackoverflow.com/questions/28400690/python3-correct-way-to-import-relative-or-absolute.

This is no longer allowed:
```
import network
```
This is the preferred way.
```
from . import network
```
should be used.

This was done in order to avoid name collusion between, for example, external module called `network` and local module, that just happens to be called `network`.

We currently use this style:
```
import sys
sys.path.append('lib')

from peer import logger
from proxy import ProxyHandler
```

We could change the style to be:
```
from ...lib.cluster import start_cluster
from ...lib.peer import logger
from ...lib.proxy import ProxyHandler
```

This is an example on how we can structure all pytests.

Tests can be run by:
```
cd .
pytest pytests/tests/sanity/routing_table_sync.py
```
or
```
cd pytests
pytest tests/sanity/routing_table_sync.py
```

Done:
- ranamed `pytest` to `pytests` to avoid name collision with `pytest` module
- Using relative imports `from ...lib.peer import logger` instead of hack with `sys.path.append`

My proposal is to migrate to the new style. Before I do the migration I would like to hear what people thing about it.


mina86:
> pytest pytests/tests/sanity/routing_table_sync.py

That’s a lie. ;) Our tests can’t really be run by `pytest` at this moment. In both cases the command needs to be `python3 …`.

mina86:
I’m generally in favour though I wonder if renaming `pytest` to `pytests` could be avoided?

pmnoxx:

PR #5506, opened 2021-11-27
scripts: add a `check_formatting.sh` script

mina86:
Introduce a new `script/check_formatting.sh` script which incorporates
Python formatting check currently coded in the BuildKite pipeline
command directly and Cargo formatting check done as an integration
test.

This has two benefits.  For the Python code, the pipeline command is
now less complex.  Call to yapf and an error message on failure are
now replaced by a `check_formatting.sh` invocation.

For the Rust code, the current situation is such that if code is not
formatted correctly ‘cargo test’ as well as ‘cargo test nightly’ steps
are failing.  With this change there’s going to be just one failure in
‘sanity checks’.


scripts/check_formatting.sh
```diff
@@ -0,0 +1,46 @@
+#!/bin/sh
+
+set -eu
+
+if [ -d .github ]; then
+    : pass
```
pmnoxx, line 6: That's a good idea. Maybe we should write a loop, to make this work from inside nested folders as well?

scripts/check_formatting.sh
```diff
@@ -0,0 +1,46 @@
+#!/bin/sh
```
pmnoxx, line 1:
```
#!/usr/bin/env sh
```

Would it be better, because it's cross platform compatible.

pmnoxx (approved):
LGTM

scripts/check_formatting.sh
mina86: `/bin/sh` is guaranteed to exist on every POSIX-like system and unless you’re running Solaris it’s going to be a fairly recent POSIX-shell.

scripts/check_formatting.sh
mina86: I dunno. I don’t think there is a strong case for running it from within a nested directory and in general I usually try to avoid looping through all parent directories in situations like this. It sometimes may lead to usability issues (e.g. if /home is autofs) but we also may end up in someone’s home directory since it’s not inconceivable they have `~/.github` directory.

pmnoxx: Yes, that's a good point.

If we only use this script to check formatting in the current directory.  It may be a good idea to check only the parent folder of `scripts/check_formatting.sh`. This way we will be able to call the script from any location.

Just my thoughts. Current design is fine as well.

chefsale (approved):

scripts/check_formatting.sh
pmnoxx: In all other scripts in that directory we use
```
#!/usr/bin/env bash
```
It may be worth using the same shebang in too .sh files

PR #5512, opened 2021-11-28
feat: use `lru` instead of `cached` in `near-network`

pmnoxx:
use `lru` instead of `cached` in `near-network`

Reasons:
- performance - 50% more efficient than `cached` in my benchmarks
- api - `LruCache` uses more standard api name `put` / `get` etc.
- less dependencies - https://crates.io/crates/lru/0.7.0/dependencies vs https://crates.io/crates/cached/0.26.2/dependencies
- actively maintained
- more popuilar Recent: 1,427,174 vs Recent: 259,687 for `LruCache` vs `cached`


pmnoxx:
```
Cargo.lock:170:1
--
  | [2021-11-28T21:00:22Z]     │
  | [2021-11-28T21:00:22Z] 170 │ ╭ hashbrown 0.11.2 registry+https://github.com/rust-lang/crates.io-index
  | [2021-11-28T21:00:22Z] 171 │ │ hashbrown 0.9.1 registry+https://github.com/rust-lang/crates.io-index
  | [2021-11-28T21:00:22Z]     │ ╰─────────────────────────────────────────────────────────────────────^ lock entries

```

pmnoxx:
This conflicts with 
```
├── cached v0.23.0     
│   │   ├── hashbrown v0.9.1 (*)
-------------------------------------------------                                                                                  
│   │   │   └── serde_json v1.0.68                                                                                                                                                            
│   │   │       ├── indexmap v1.6.1                                                                                                                                                           
│   │   │       │   ├── hashbrown v0.9.1 (*)  
------------------------
│   │   ├── borsh v0.9.1
│   │   │   └── hashbrown v0.9.1
    
```

pmnoxx:
Downgrading `lru` to `0.6.5` to try to resolve version conflict.

matklad:
>Downgrading lru to 0.6.5 to try to resolve version conflict.

Ah, that's unfortunate. But yeah, I feel it's better to not pull yet *another* copy of hashbrown. 

Fixing this properly (by updating, rather than downgrading) would require https://github.com/near/borsh-rs/issues/45, https://github.com/near/borsh-rs/issues/46, and removing `cached` altogether.

chain/network/Cargo.toml
```diff
@@ -11,11 +11,11 @@ borsh = { version = "0.9", features = ["rc"]}
 byteorder = "1.2"
 bytes = "1"
 bytesize = "1.1"
-cached = "0.23"
 chrono = { version = "0.4.4", features = ["serde"] }
 conqueue = "0.4.0"
 futures = "0.3"
 lazy_static = "1.4"
+lru = "=0.6.5"
```
matklad, line 18: No need for `=` here, pinning is done by Cargo.lock. To ask cargo to put specific verison in cargo lock, use the following command: 

```console
$ cargo update -p lru --precise 0.6.6
```

nearcore/src/config.rs
```diff
@@ -1115,24 +1114,39 @@ pub enum FileDownloadError {
...
-        }
-        Ok(())
-    });
+    let result =
+        tokio::runtime::Builder::new_current_thread().enable_all().build().unwrap().block_on(
+            async move {
+                let tmp_file = tokio::fs::File::from_std(tmp_file);
+                download_file_impl(uri, tmp_file).await
+            },
+        );
```
matklad, line 1149: unrelated changes?

matklad (approved):
Yeah, lru looks much, much more reasonable than sized cache. When this is merged, I'll update contract runtime usage of cached s well.

nearcore/src/config.rs
pmnoxx: Oh, that was a commit that's already in the master.

chain/network/Cargo.toml
pmnoxx: I think it's better to specify `0.6.5`, because the `0.6.6` is broken.

pmnoxx: I changed it to `lru = "0.6.5"`

matklad: I mean, we should say `lru = "0.6.5"` in Cargo.toml, and let Cargo.lock to take care about `=`.

matklad:
Without carefully reading every line of code, I am personally more confident in `lru` than in `cached`, as that looks like just a more tastefully design crate. lru's internal impl looks reasonable on the first glance, though it does makes sense from someone from core to give it a more comprehensive reading.

I've also noticed an opportunity for optimization there: https://github.com/jeromefroe/lru-rs/issues/119. Trying to implement that might be a fun way to audit the crate.

bowenwang1996 (approved):

pmnoxx:
@mm-near I looked at the implementation of `lru`. It looks reasonable. I can be optimized further, but it's faster than `cached` implementation.

PR #5514, opened 2021-11-28
re: Add `shebang` to all python tests in `pytest/tests`

pmnoxx:
This will allow running files directly.

for example
```
cd pytest
./tests/sanity/block_chunk_signature.py
```

See https://github.com/near/nearcore/pull/5496


mina86 (approved):

matklad:
Thanks a lot!

Q (for @mina86 I guess): is it intentional that `pytest/tests/sanity/upgradable.py` is an empty file?

pmnoxx:
> Thanks a lot!
> 
> Q (for @mina86 I guess): is it intentional that `pytest/tests/sanity/upgradable.py` is an empty file?

FYI, it's not an empty file https://github.com/near/nearcore/blob/ccbe7c37c3a763e704596d7c10f456e10314360d/pytest/tests/sanity/upgradable.py

The github UI display empty file, because there were to file changes, just permissions.

PR #5515, opened 2021-11-29
Remove serde from `near-client-primitives`.

pmnoxx:
We should remove `serde` from `near-client-primitives`. It's only needed by `jsonrpc`. 
More changes will be required, to separate `jsonrpc` api, from rest of the code base, but that's a first step.

See https://github.com/near/nearcore/issues/5516

Blocks https://github.com/near/nearcore/pull/5428#pullrequestreview-813563252


chain/jsonrpc-primitives/src/types/network_info.rs
```diff
@@ -1,9 +1,15 @@
+use near_network_primitives::types::{KnownProducer, PeerInfo};
 use serde::{Deserialize, Serialize};
 
 #[derive(Serialize, Debug)]
```
miraclx, line 4: This now needs to impl `Deserialize`.

miraclx (changes requested):
LGTM, just one little thing:

chain/jsonrpc-primitives/src/types/network_info.rs
matklad: Could you explain why? We didn't have a Deserialize impl before, so this *shouldn't* break stuff?

chain/client-primitives/Cargo.toml
```diff
@@ -13,7 +13,6 @@ description = "This crate hosts NEAR client-related error types"
 [dependencies]
 actix = "=0.11.0-beta.2"
 chrono = { version = "0.4.4", features = ["serde"] }
-serde = { version = "1", features = ["derive"] }
```
matklad, line 16: As alternative, we can use a negative import here: 

```toml
# Note: this crate intentionally doesn't depend on serde
# serde = "1"
```

I think this might be moderately valuable while we are in-transition, and most things do depend on serde.

chain/jsonrpc-primitives/src/types/network_info.rs
miraclx: Currently, `near-jsonrpc-client` uses the `near-client-primitives` version of `RpcNetworkInfoResponse` [(src)](https://github.com/near/near-jsonrpc-client-rs/blob/dd7761b51e1775350be1782370aa22c0b0fe98d7/src/methods.rs#L473). Given that the incoming https://github.com/near/nearcore/pull/5428 patch changes that interface means we have to use this new `near-jsonrpc-primitives` version to maintain compatibility. Meaning, it should be deserializable.

pmnoxx: @miraclx After we merge #5515, there will be no need to change the `jsonrpc` interface for the renaming.
I don't mind adding Deserialize. However, we don't current use it, and we won't use it after #5428

chain/client-primitives/Cargo.toml
pmnoxx: It's a good idea. However, I'm not sure we need it.

chefsale (approved):

miraclx (approved):

pmnoxx:
@miraclx I added `Deserialize` for consistency reasons.

chain/jsonrpc-primitives/src/types/network_info.rs
miraclx: @pmnoxx, Once https://github.com/near/nearcore/pull/5515 is merged, then `NetworkInfoResponse` would no longer be usable from `near-jsonrpc-client` as it would be incompatible for deserializing the server response. We would then have to use this version you just implemented `Deserialize` for. That's why it needs that spec.

chefsale (approved):

PR #5517, opened 2021-11-29
feat: Add tests to `near-stable-hasher`

pmnoxx:
While looking out our code base. I noticed we don't have tests for `near-stable-hasher`, that could protect us in case this harsher ever changes.


utils/near-stable-hasher/src/lib.rs
```diff
@@ -21,3 +21,19 @@ impl Hasher for StableHasher {
...
+mod tests {
+    use crate::StableHasher;
+    use std::hash::Hasher;
+
+    /// Make sure the stable hasher never changes
+    #[test]
+    fn test_stable_hasher() {
+        let mut sh = StableHasher::new();
+
+        sh.write(vec![1, 2, 3, 4, 5].as_slice());
```
matklad, line 35:
```suggestion
        sh.write(&[1, 2, 3, 4, 5]);
```

More generally, I feel we have a lot of "just allocate temporary vec" in our code base. This doesn't feel very rustic -- Rust APIs are very deliberately designed to be zero copy/zero allocation. I feel it makes sense to be more careful about such easy small perf (and simplicity) gains as a matter of habit.

matklad (approved):
Good catch! 

Practically, `SipHasher` in std is guaranteed to never change, but if in the future we migrate to non-std impl, it's important to have a test to check that the thing stays the same.

PR #5520, opened 2021-11-29
feat(mocknet): Progressive upgrade of `neard` during mocknet tests

nikurt:


pytest/lib/account.py
```diff
@@ -83,3 +83,10 @@ def send_stake_tx(self, stake_amount):
         tx = sign_staking_tx(self.key, self.key, stake_amount, self.nonce,
                              self.base_block_hash)
         return self.send_tx(tx)
+
+    def get_amount_yoctonear(self):
+        j = self.json_rpc('query', {
+            'request_type': 'view_account',
+            'finality': 'optimistic',
+            'account_id': self.key.account_id})
+        return int(j.get('result',{}).get('amount',0))
```
mina86, line 92: That’s essentially `get_account` from `BaseNode`.

pytest/tests/mocknet/helpers/genesis_updater.py
```diff
@@ -42,6 +44,8 @@ def main(argv):
     assert epoch_length
     assert node_pks
     assert rpc_node_names
+    assert increasing_stakes is not None
```
mina86, line 47: There is no way in which it can be `None`.

pytest/tests/mocknet/helpers/genesis_updater.py
```diff
@@ -17,7 +17,7 @@
 
 def main(argv):
     logger.info(argv)
-    assert len(argv) == 9
+    assert len(argv) >= 11
```
mina86, line 20: Why `>=`?

pytest/tests/mocknet/helpers/load_test_spoon_helper.py
```diff
@@ -292,7 +292,29 @@ def main(argv):
...
+                            f'Staking {stake_amount} for {node_account.key.account_id}'
+                        )
+                        node_account.send_stake_tx(stake_amount)
+                    last_staking = time.time()
+                    logger.info(
+                        f'Staked {stake_amount} for {node_account.key.account_id}'
+                    )
+                    break
+                except Exception as e:
+                    logger.info('Failed to stake')
```
mina86, line 317: Shouldn’t there be some kind of `time.sleep` in this loop?

mina86 (approved):

pytest/tests/mocknet/helpers/load_test_spoon_helper.py
nikurt: The sleep is supposed to be outside this function.

pytest/lib/account.py
nikurt: Good point, but I don't know how to avoid this code duplication.
The helper scripts run on each machine, and those scripts are not operating in terms of `BaseNode` or any other node.

pytest/lib/mocknet.py
```diff
@@ -1073,3 +1073,31 @@ def upgrade_node(node):
...
+                    f'Amount of {node_account.key.account_id} is {stake_amount}'
+                )
+                if stake_amount > (10**3) * NEAR_IN_YOCTONEAR:
+                    logger.info(
+                        f'Staking {stake_amount} for {node_account.key.account_id}'
+                    )
+                    node_account.send_stake_tx(stake_amount)
+                logger.info(
+                    f'Staked {stake_amount} for {node_account.key.account_id}')
+                return time.time()
```
mina86, line 1101: I think it makes more sense for the function to return a boolean

pytest/tests/mocknet/helpers/load_test_spoon_helper.py
mina86: I mean just a short sleep between attempts.

pmnoxx:
@nikurt Fails sanity checks removing `S-automerge`

PR #5522, opened 2021-11-29
fix: `PeerActorManager`: Remove hidden overflows

pmnoxx:
Adding a value to `active_peer.full_peer_info.chain_info.height`  could potentially cause an overflow error. It's better to use saturating add.

Refactor afterwards: https://github.com/near/nearcore/pull/5524


matklad (approved):
LGTM!

As a personal stylistic perference, I'd write this `filter_map` as a combination of separate `filter` and `map` invocations. `filter . map` composition can do *less* than a `filter_map`, so I find it to be more readable.

bowenwang1996 (approved):

Issue #4956, opened 2021-10-08
Add CI check for protocol version increase


bowenwang1996:
In the implementation of certain protocol features, we rely on the fact that no protocol version is skipped. As a result, if we increase protocol version by more than 1 in one release, it could lead to bugs. We should add a CI check for this.

mina86:
There’s still a corner case which this test won’t catch:
1. Cut a 1.42.0 release with protocol version 24.
2. Increase protocol version on master branch to version 25. This works because the test will compare against 1.42 binary.
3. Cut a 1.43.0-rc1 release with protocol version 25.
4. Increase protocol version on master branch to version 26. This works because the test will compare against 1.43-0-rc1 binary.
5. Cut a 1.43.0-rc2 release with protocol version 26.
6. Cut a 1.43.0 release with protocol version 26 which is two higher than previous stable release.

mzhangmzz:
Isn't this covered by the upgradable test? If we increase protocol version by more than 1 from the last stable release and some features will break because of that, the upgradeable test will not pass right?

mina86:
I’d imagine you’re not guaranteed to hit the scenario in which the test breaks.

PR #5523, opened 2021-11-29
pytest: make A/B tests against latest deployed mainnet binary

mina86:
Use latest_release and latest_deploy data stored on S3 rather than
relying on GitHub releases to figure out what is the latest binary run
on mainnet for performing A/B tests.

This fixes issues around release time when our CI suddenly breaks when
a new rc is made without it yet being released on either testnet or
mainnet.

For example, once 1.42.0-rc1 is release on GitHub, the code expects to
download a 1.42.0 binary from S3 regardless of whether such release
has been deployed or not.  With this change, the code will look at the
actual executable that is running in production on mainnet.

This also makes sure that all our tests are performed against what is
running on mainnet so that we can confirm that upgrading from mainnet
to current HEAD is possible.

Fixes: https://github.com/near/nearcore/issues/4956

### DIFF
--- a/pytest/lib/branches.py
+++ b/pytest/lib/branches.py
@@ -5,9 +5,13 @@ import sys
 import tempfile
 import typing
 
+import requests
 import semver
 from configured_logger import logger
-from github import Github
+
+_UNAME = os.uname()[0]
+_IS_DARWIN = _UNAME == 'Darwin'
+_BASEHREF = 'https://s3-us-west-1.amazonaws.com/build.nearprotocol.com'
 
 
 def current_branch():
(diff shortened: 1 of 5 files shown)


pytest/lib/branches.py
```diff
@@ -16,34 +20,22 @@ def current_branch():
...
+        res = requests.get(url)
+        res.raise_for_status()
+        return res.text
 
-def latest_rc_branch():
-    releases = list(
-        filter(
-            lambda release: (semver.VersionInfo.parse(release.title).prerelease
-                             or "").startswith("rc"), get_releases()))
+    release = download(basehref + '/latest_release')
```
nikurt, line 32: I slightly prefer the formatted string as you are using on line 25.

nikurt (approved):

PR #5525, opened 2021-11-29
docs: adding more documentation to DbCols

mm-near:


core/store/src/db.rs
```diff
@@ -49,22 +49,66 @@ impl Into<io::Error> for DBError {
 #[derive(PartialEq, Debug, Copy, Clone, EnumIter, BorshDeserialize, BorshSerialize, Hash, Eq)]
 pub enum DBCol {
```
pmnoxx, line 50: I think we should provide rust like interface to the database.
```rust
get<DBCol::ColBlock>(arg: CryptoHash) -> near_primitives::block_header::BlockHeader
```

pmnoxx: @matklad What do you think?

core/store/src/db.rs
```diff
@@ -49,22 +49,66 @@ impl Into<io::Error> for DBError {
...
     /// Column to indicate which version of database this is.
+    /// - *Rows*: single row [VERSION_KEY]
```
pmnoxx, line 52: Also, there is no documentation, which column is `is_rc()`.

mm-near: That's a good point - how do we do reference counting in our storage? (we can add it to our documentation in a separate PR)

core/store/src/db.rs
mm-near: that's an interesting idea - but it might be a larger scope than this PR intended.

Maybe let's start a zulip thread about it (how the storage interface should look like).
It might be interesting, as AFAIK Evgeny is thinking about changing the underlying storage to something else, so having an interface might help

core/store/src/db.rs
```diff
@@ -49,22 +49,66 @@ impl Into<io::Error> for DBError {
...
+    /// - *Content type*: [near_primitives::block_header::BlockHeader]
     ColBlockHeader = 3,
+    /// Column that stores mapping from block height to block hash.
+    /// - *Rows*: height (u64)
+    /// - *Content type*: block hash (CryptoHash)
     ColBlockHeight = 4,
+    /// TODO: this is something related to garbage collection - no idea yet.
     ColState = 5,
+    /// Mapping from BlockChunk to ChunkExtra
+    /// - *Rows*: BlockChunk (block hash + shard id)
```
matklad, line 74: Should `+` here and elsewhere be taken as concatenation? Ie, we don't numerically add shard id to the hash, we concatenate the two, right? I wonder if there's obvious, short way to write that.

core/store/src/db.rs
```diff
@@ -85,13 +129,22 @@ pub enum DBCol {
     ColReceiptIdToShardId = 27,
     _ColNextBlockWithNewChunk = 28,
     _ColLastBlockWithNewChunk = 29,
-    /// Map each saved peer on disk with its component id.
+    /// Network related: Map each saved peer on disk with its component id (a.k.a. nonce).
```
matklad, line 132: Is "component" the right word here? This is **not** connected component, right? Maybe just call this `nonce` instead? 

Things like "edges in this component" sound very confusing to me (unless I am already hopelessly confused, and nonces *are* somehow connected components as well).

core/store/src/db.rs
```diff
@@ -85,13 +129,22 @@ pub enum DBCol {
...
+    /// - *Rows*: peer_id
+    /// - *Column type*:  u64
     ColPeerComponent = 30,
-    /// Map component id with all edges in this component.
+    /// Network related: Map component id  (a.k.a. nonce) with all edges in this component.
+    /// These are all the edges that were purged and persisted to disk at the same time.
+    /// - *Rows*: nonce
```
matklad, line 138: if we `: nonce` here, then lets `: nonce` rather than `: u64` above.

core/store/src/db.rs
```diff
@@ -85,13 +129,22 @@ pub enum DBCol {
...
+    /// - *Column type*: `Vec<near_network::routing::Edge>`
     ColComponentEdges = 31,
-    /// Biggest nonce used.
+    /// Network related: Biggest component id (a.k.a nonce) used.
+    /// - *Rows*: single row (empty row name)
+    /// - *Column type*: u64
     ColLastComponentNonce = 32,
-    /// Transactions
+    /// Map of transactions (key is transaction hash)
```
matklad, line 145:
```suggestion
    /// Map of transactions
```

We don't explain that row=key anywhere else.

core/store/src/db.rs
```diff
@@ -49,22 +49,66 @@ impl Into<io::Error> for DBError {
 #[derive(PartialEq, Debug, Copy, Clone, EnumIter, BorshDeserialize, BorshSerialize, Hash, Eq)]
```
matklad, line 49: Let's maybe add comment for the enum as a whole (or maybe even `//!` comment at the start of the file) which explains terminology here. Ie, what are columns and rows? I haven't worked with key-value stores a lot, so it is a *bit* confusing to me.

matklad (approved):
:heavy_plus_sign: :100: 

Those are very high-impact docs, as knowing what's in the store more or less tells you everything about the overall architecture of the thing. 

I suggest linking this file from  https://github.com/near/nearcore/blob/master/docs/architecture.md#code-map -- it feels like of the entry points great for starting to read the code base.

core/store/src/db.rs
mm-near: done.

core/store/src/db.rs
mm-near: done

core/store/src/db.rs
mm-near: Yes, that is very unfortunate naming - I've updated the comment

core/store/src/db.rs
mm-near: I've added some explanation on top of DBCol - ptal

core/store/src/db.rs
pmnoxx: Yes, it's a topic for another discussion. Not blocking this PR.

core/store/src/db.rs
```diff
@@ -46,25 +46,72 @@ impl Into<io::Error> for DBError {
...
     ColBlock = 2,
+    /// Column that stores Block headers.
+    /// - *Rows*: block hash (CryptoHash)
+    /// - *Content type*: [near_primitives::block_header::BlockHeader]
     ColBlockHeader = 3,
+    /// Column that stores mapping from block height to block hash.
+    /// - *Rows*: height (u64)
+    /// - *Content type*: block hash (CryptoHash)
     ColBlockHeight = 4,
+    /// TODO: this is something related to garbage collection - no idea yet.
```
bowenwang1996, line 74: No. `ColState` stores state

core/store/src/db.rs
```diff
@@ -46,25 +46,72 @@ impl Into<io::Error> for DBError {
...
     ColEpochInfo = 11,
+    /// Mapping from BlockHash to BlockInfo
+    /// - *Rows*: BlockHash (CryptoHash)
+    /// - *Content type*: [near_primitives::epoch_manager::BlockInfo]
     ColBlockInfo = 12,
+    /// Mapping from ChunkHash to ShardChunk.
+    /// - *Rows*: ChunkHash (CryptoHash)
+    /// - *Content type*: [near_primitives::sharding::ShardChunk]
     ColChunks = 13,
+    /// Storage for  PartialEncodedChunkV1 (seems that only for version V1 and not for V2).
```
bowenwang1996, line 109: where is V2 stored then?

core/store/src/db.rs
mm-near: good point - fixed.

core/store/src/db.rs
mm-near: I misread the code - seems that both versions are stored here.

core/store/src/db.rs
```diff
@@ -46,25 +46,74 @@ impl Into<io::Error> for DBError {
...
+    /// Column that stores mapping from block height to block hash.
+    /// - *Rows*: height (u64)
+    /// - *Content type*: block hash (CryptoHash)
     ColBlockHeight = 4,
+    /// Column that stores the Trie state.
+    /// - *Rows*: trie_node_or_value_hash (CryptoHash)
+    /// - *Content type*: Serializd RawTrieNodeWithSize or value ()
     ColState = 5,
+    /// Mapping from BlockChunk to ChunkExtra
+    /// - *Rows*: BlockChunk (block hash, shard id)
```
mzhangmzz, line 79: It should be `(block_hash, shard_uid)`

mm-near: thx. done

core/store/src/db.rs
```diff
@@ -85,13 +134,40 @@ pub enum DBCol {
     ColReceiptIdToShardId = 27,
     _ColNextBlockWithNewChunk = 28,
     _ColLastBlockWithNewChunk = 29,
```
bowenwang1996, line 136: Could we document here that these two columns are deprecated?

bowenwang1996 (approved):

core/store/src/db.rs
mm-near: done

PR #5526, opened 2021-11-29
fix: Add debug logs to `upgradable.py` in case of failure.

pmnoxx:
We've been experiencing random test failures of `upgradable.py` see https://github.com/near/nearcore/pull/5524

'query' jsonrpc return an 'account' information, or an error json response, in case account is not found.
Let's start by printing the full response, this will allow us to see what the issue is.

### DIFF
--- a/pytest/lib/cluster.py
+++ b/pytest/lib/cluster.py
@@ -229,13 +229,14 @@ class BaseNode(object):
     def get_validators(self):
         return self.json_rpc('validators', [None])
 
-    def get_account(self, acc, finality='optimistic'):
+    def get_account(self, acc, finality='optimistic', do_assert=True):
         res = self.json_rpc('query', {
             "request_type": "view_account",
             "account_id": acc,
             "finality": finality
         })
-        assert 'error' not in res, res
+        if do_assert:
+            assert 'error' not in res, res
 
         return res
 
--- a/pytest/lib/cluster.py
+++ b/pytest/lib/cluster.py
@@ -230,11 +230,14 @@ class BaseNode(object):
         return self.json_rpc('validators', [None])
 
     def get_account(self, acc, finality='optimistic'):
-        return self.json_rpc('query', {
+        res = self.json_rpc('query', {
             "request_type": "view_account",
             "account_id": acc,
             "finality": finality
         })
+        assert 'error' not in res, res
+
+        return res
 
     def call_function(self,
                       acc,
--- a/pytest/tests/sanity/rosetta.py
+++ b/pytest/tests/sanity/rosetta.py
@@ -326,7 +326,7 @@ class RosettaTestCase(unittest.TestCase):
             require: If True, require that the account exists.
         """
         account_id = account.account_id
-        result = self.node.get_account(account_id)
+        result = self.node.get_account(account_id, do_assert=False)
         error = result.get('error')
         if error is None:
             amount = int(result['result']['amount'])


bowenwang1996:
should we close this PR if the fix does not work?

pmnoxx:
> should we close this PR if the fix does not work?
I think that the best course of action is to add debug logs.
In `jsonrpc`, we return an error in case account was not found, or some other issue happened. It's a good idea to print what the actual response is, this will allow us to debug it next time it happens/

pmnoxx:
@bowenwang1996 We should ship the debug logs.

pytest/tests/sanity/upgradable.py
```diff
@@ -164,8 +164,14 @@ def test_upgrade() -> None:
...
     assert 'Failure' not in res['result']['status'], res
 
-    hex_account_balance = int(
-        nodes[0].get_account(hex_account_id)['result']['amount'])
+    # We have seen test failures, where account was not found.
+
+    response = nodes[0].get_account(hex_account_id)
+
+    if 'result' not in response or 'amount' not in response['result']:
+        logger.error("got invalid response from node %s" % response)
```
mina86, line 172: Just `assert 'error' not in response, f'got error from node: {response["error"]}'`

pmnoxx:
@mina86 Here is the error
```
{'jsonrpc': '2.0', 'error': {'name': 'HANDLER_ERROR', 'cause': {'name': 'UNAVAILABLE_SHARD', 'info': {'requested_shard_id': 0}}, 'code': -32000, 'message': 'Server error', 'data': 'The node does not track the shard ID 0'}, 'id': 'dontcare'}

```

pytest/tests/sanity/upgradable.py
pmnoxx: Thanks, that's much better.

pytest/tests/sanity/upgradable.py
```diff
@@ -164,8 +164,13 @@ def test_upgrade() -> None:
     assert 'error' not in res, res
     assert 'Failure' not in res['result']['status'], res
 
-    hex_account_balance = int(
-        nodes[0].get_account(hex_account_id)['result']['amount'])
+    # We have seen test failures, where account was not found.
+
+    response = nodes[0].get_account(hex_account_id)
+
+    assert 'result' in response, f'got error from node: {response["error"]}'
```
mina86, line 171: By the way, looking at the rest of the code `assert 'error' not in res, res` is actually more consistent.

pytest/tests/sanity/upgradable.py
```diff
@@ -164,8 +164,13 @@ def test_upgrade() -> None:
...
+
+    hex_account_balance = int(response['result']['amount'])
```
mm-near, line 173: we have 16 places in our tests, where we use "['result']['amount']" -- maybe we should have a util function for that ? (this way it could check that account exists etc etc).

mm-near (approved):

pytest/tests/sanity/upgradable.py
pmnoxx: We have a few options:
1) Add a check to `get_account`

@mm-near @mina86 Another way is to add a check inside
`get_account`.

```
    def get_account(self, acc, finality='optimistic'):
        res = self.json_rpc('query', {
            "request_type": "view_account",
            "account_id": acc,
            "finality": finality
        })
        assert 'error' not in res, res

        return res
```

2) add argument  `assert_error=True` to `self.json_rpc`. 
Though, this may break some existing code if there is an error

3) add argument  `print_error=True` to `self.json_rpc`. 
Though, this add some extra prints to `json_rpc` shouldn't break anything, but will give us proper debug output.

My vote is on 2), but that may be a breaking change.
1) should be safe to do.

pmnoxx:
https://nayduck.near.org/#/run/2317

chefsale (approved):

miraclx (approved):

pmnoxx:
https://nayduck.near.org/#/run/2319

Issue #5452, opened 2021-11-24
build fails for v1.22.0


rencryptofish:
running `make release` getting the following build error 

```
error[E0658]: arbitrary expressions in key-value attributes are unstable
 --> runtime/near-vm-errors/src/lib.rs:1:10
  |
1 | #![doc = include_str!("../README.md")]
  |          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  |
  = note: see issue #78835 <https://github.com/rust-lang/rust/issues/78835> for more information
```

mina86:
Could you state which `rustc` version are you using? I.e. output of `rustc --version`?

mina86:
We generally require the latest rustc to compile the code. The actual minimal version the code expects is sest in [rust-toolchain.toml file](https://github.com/near/nearcore/blob/master/rust-toolchain.toml) and at the moment that’s 1.56. The error you’ve quoted indicates that you’re building the code with pre-1.54 compiler. Please update and then you’ll be able to build.

rencryptofish:
ah got it thanks, what about adding something like for node runners:
https://docs.rs/version_check/latest/version_check/

mina86:
There shouldn’t be a need. Cargo should update the toolchain according to the `rust-toolchain.toml`. The reason it didn’t work for you is that prior to [rustup 1.24 (released in April 2021)](https://blog.rust-lang.org/2021/04/27/Rustup-1.24.0.html), the file [was not recognised](https://github.com/rust-lang/rustup/issues/2652) due to `toml` extension. Going forward it’ll work on its own.

mina86:
Actually I lied a bit. `rust-toolchain.toml` will take effect if Rust wast installed through `rustup` only (so for example won’t take effect if it came from system’s package manager). Still, it’s not clear whether adding `build.rs` just to check the version is overall a positive change.

PR #5530, opened 2021-11-29
Add `rust-version` to Cargo configuration

mina86:
Add `package.rust-version` to all `Cargo.toml` files so that users don’t
get mysterious compiler failures when trying to build the project with
unsupported compilers.

Issue: https://github.com/near/nearcore/issues/5452


frol (approved):

miraclx (approved):

matklad:
Example error

```
λ make release
error: package `near-chunks-primitives v0.0.0 (/home/matklad/projects/nearcore/chain/chunks-primitives)` cannot be built because it requires rustc 1.59.0 or newer, while the currently active rustc version is 1.56.0
make: *** [Makefile:22: release] Error 101
```

Not a huge fan of "remember to upgrade this config in ten different places" workflows, as most likely we won't be able to keep this consistent. But it's worth a try!

matklad (approved):

matklad:
(one alternative is to pin this *only* for, eg `nearcore/Cargo.toml`, which should make, eg `make release` behave)

nikurt (approved):

nagisa:
I'm of a similar opinion as @matklad. Having `rust-version` specified for each crate would make most sense if we maintained and tested each crate separately. Since this is a monorepo-ish setup, specifying `rust-version` for just the most commonly depended on crate within the workspace is going to be easier to work with.

chefsale (approved):

miraclx:
Of the most commonly depended on, I think the best candidate would be `near-primitives`

matklad:
To clarify, given that we do publish things to crates IO (without much semver guarantees), I thing there *might* be some merit in just specifying this for everything.

pmnoxx (approved):

mina86:
Looks like opinions are split regarding whether this should be just in `near-primitives` or everywhere. I’m gonna mark it automerge in the current form (with `rust-version` added to all `Cargo.toml`). If anyone has strong opinions remove the label and discuss it.

mm-near (approved):

chefsale (approved):

pmnoxx:
As with every decision, there are benefits and drawbacks:

Benefits:
- external developers who use our packages, will know what we require new version, making it less confusing
- next time we upgrade Rust version, every code owner will get informed, we will be able to check what new features the next version offers, and decide whenever we should use any of them
- we already put `version` inside each crate, whenever we bump up version everyone already gets notified. So, as long as we do the version update, at the same time, as `rust_version` update, we will not be adding any more extra work, than we already are doing.

Drawback:
- Every code owners needs to accept the change. However, we already do that with every `version` bump. So, we are not really making the situation worse than it already is.

Did I miss anything?

PR #5532, opened 2021-11-30
feat: Fix `nayduck.py --run-locally` to work from any directory

pmnoxx:
`nayduck.py --run-locally` used to run assuming it gets executed inside `REPO/pytest`. After this PR `nayduck.py --run-locally` should work from any directory.


scripts/nayduck.py
```diff
@@ -205,17 +207,17 @@ def run_locally(tests):
                 '--',
                 '--exact',
                 fields[3])
-            cwd = None
+            cwd = os.path.join(REPO_DIR, 'pytest')
```
mina86, line 210: That should just be `REPO_DIR`

scripts/nayduck.py
```diff
@@ -205,17 +207,17 @@ def run_locally(tests):
...
         elif fields[0] in ('pytest', 'mocknet'):
             fields[0] = sys.executable
-            fields[1] = os.path.join('tests', fields[1])
+            fields[1] = os.path.join('./pytest/tests', fields[1])
             cmd = fields
-            cwd = 'pytest'
+            cwd = os.path.join(REPO_DIR, 'pytest')
```
mina86, line 215: I would leave `fields[1]` alone and than `cwd = REPO_DIR / 'pytest'`. As it stands pytests won’t work.

scripts/nayduck.py
```diff
@@ -205,17 +207,17 @@ def run_locally(tests):
...
         else:
             print(f'Unrecognised test category ‘{fields[0]}’', file=sys.stderr)
             continue
         print("RUNNING COMMAND cwd=%s cmd = %s", (cwd, cmd))
-        subprocess.check_call(cmd, cwd=cwd)
+        subprocess.check_call(cmd, cwd=str(cwd))
```
mina86, line 220: No need for `str()`. Pretty much everywhere where Python expects a path you can either pass `str` or `pathlib.Path` and the latter is preferred I’d say.

scripts/nayduck.py
pmnoxx: Thanks, fixed..

scripts/nayduck.py
pmnoxx: Thanks, fixed.

scripts/nayduck.py
pmnoxx: Nice, I see that removing `str`, still works with `subprocess`

mina86 (approved):

chefsale (approved):

PR #5533, opened 2021-11-30
feat: Implement `--timeout` in `nayduck.py`

pmnoxx:
Currently `nayduck.py` ignores `--timeout` parameter, this `PR` will add `--timeout` funcionality.


scripts/nayduck.py
```diff
@@ -182,11 +182,16 @@ def run_locally(tests):
         # See nayduck specs at https://github.com/near/nayduck/blob/master/lib/testspec.py
         fields = test.split()
 
+        timeout = ""
```
mina86, line 185: I guess using `timeout` tool is one way to do it. But if we do that we need `--verbose` argument to `timeout`. The way I would have done it is to initialise `timeout` with `None`, parse it as int and then use `subprocess.check_call(…, timeout=timeout)`.

scripts/nayduck.py
```diff
@@ -204,12 +209,16 @@ def run_locally(tests):
                 'expensive_tests',
                 '--',
                 '--exact',
-                fields[3])
+                fields[3]]
+            if timeout:
+                cmd = ["timeout", int(timeout)] + cmd
```
mina86, line 214: Since this is the same in both branches, this should go outside before the `subprocess.check_call` invocation. You also don’t want to `int(timeout)` since `timeout` can have a suffix (like `1h`).

scripts/nayduck.py
```diff
@@ -182,11 +182,16 @@ def run_locally(tests):
         # See nayduck specs at https://github.com/near/nayduck/blob/master/lib/testspec.py
         fields = test.split()
 
+        timeout = ""
         index = 1
+        ignored = []
         while len(fields) > index and fields[index].startswith('--'):
+            if fields[index].startswith("--timeout="):
+                timeout = fields[index].split("=")[1]
```
mina86, line 190: For consistency let’s use single quotes for stings. For the second line I would have done it with `timeout = fields[index][10:]` but I guess split also works.

scripts/nayduck.py
```diff
@@ -182,11 +182,16 @@ def run_locally(tests):
...
+            else:
+                ignored.append(fields[index])
             index += 1
 
-        ignored = fields[1:index]
         del fields[1:index]
         message = f'Running ‘{"".join(fields)}’'
```
mina86, line 196: By the way, since you’re editing this file, could you fix this join to be `" ".join(fileds)`? Same on line 198.

scripts/nayduck.py
```diff
@@ -182,11 +182,16 @@ def run_locally(tests):
         # See nayduck specs at https://github.com/near/nayduck/blob/master/lib/testspec.py
         fields = test.split()
 
+        timeout = ""
         index = 1
+        ignored = []
         while len(fields) > index and fields[index].startswith('--'):
+            if fields[index].startswith("--timeout="):
+                timeout = fields[index].split("=")[1]
+            else:
```
mina86, line 191: Might just as well do `elif fields[index] != '--skip-build':` so as to not bother people about that flag. ‘Skip build’ is the only behaviour this script handles after all.

scripts/nayduck.py
pmnoxx: ` subprocess.check_call` would be better, but then we would need some way of parsing time for example `2h` to seconds.

scripts/nayduck.py
pmnoxx: Ok, fixed.

pmnoxx:
@mina86 Is there anything else we should do. What's the best way of parsing `timeout` string to seconds?

scripts/nayduck.py
mina86: You could copy the code from nayduck.

```
def _parse_timeout(timeout: str) -> int:
    """Parses timeout interval and converts it into number of seconds.

    Args:
        timeout: An integer with an optional ‘h’, ‘m’ or ‘s’ suffix which
            multiply the integer by 3600, 60 and 1 respectively.
    Returns:
        Interval in seconds.
    """
    if timeout and (mul := {'h': 3600, 'm': 60, 's': 1}.get(timeout[-1])):
        timeout = timeout[:-1]
    else:
        mul = 1
    return int(timeout) * mul
```

pmnoxx: @mina86 Thanks, I added it to PR.

pmnoxx:
```

[2021-11-30T05:35:40Z] Nov 30 05:35:40.560  INFO neard: Version: trunk, Build: crates-0.10.0-88-gf395752c9, Latest Protocol: 49
--
  | [2021-11-30T05:35:40Z] Nov 30 05:35:40.563  INFO near: Generated node key, validator key, genesis file in /tmp/tmpgp4p1wcr
  | [2021-11-30T05:35:40Z] + python3 scripts/check_nightly.py
  | [2021-11-30T05:35:40Z] Traceback (most recent call last):
  | [2021-11-30T05:35:40Z]   File "scripts/check_nightly.py", line 8, in <module>
  | [2021-11-30T05:35:40Z]     import nayduck
  | [2021-11-30T05:35:40Z]   File "/var/lib/buildkite-agent/builds/buildkite-i-0c12a367396832d45-1/nearprotocol/nearcore/scripts/nayduck.py", line 191
  | [2021-11-30T05:35:40Z]     if timeout and (mul := {'h': 3600, 'm': 60, 's': 1}.get(timeout[-1])):
  | [2021-11-30T05:35:40Z]                         ^
  | [2021-11-30T05:35:40Z] SyntaxError: invalid syntax
  | [2021-11-30T05:35:40Z] 🚨 Error: The command exited with status 1

```

scripts/nayduck.py
```diff
@@ -188,7 +188,9 @@ def _parse_timeout(timeout: typing.Optional[str]) -> typing.Optional[int]:
     Returns:
         Interval in seconds.
     """
-    if timeout and (mul := {'h': 3600, 'm': 60, 's': 1}.get(timeout[-1])):
+    mul_ary = {'h': 3600, 'm': 60, 's': 1}
+    mul = mul_ary.get(timeout[-1])
```
mina86, line 192: `mul = timeout and mul_arr.get(timeout[-1])` won’t raise an exception if `timeout` is empty.

mina86:
Ah, yes, I’ve forgotten that you’re on Python 3.7. Walrus was added in 3.8.

scripts/nayduck.py
```diff
@@ -177,25 +177,50 @@ def github_auth(code_path: pathlib.Path):
     return code
 
 
+def _parse_timeout(timeout: typing.Optional[str]) -> typing.Optional[int]:
+    if not timeout:
+        return None
```
mina86, line 182: This needs to go below the docstring

mina86 (approved):

chefsale (approved):

PR #5534, opened 2021-11-30
pytest: better support running from arbitrary directory

mina86:
Fix a few more places which assume that the script is run from inside
of the ‘pytest’ directory.  Specifically, the ‘branches’ and ‘utils’
modules.


pmnoxx (approved):
LGTM, Thanks, this will make it even easier to use remaining scripts.

PR #5540, opened 2021-11-30
test: test for wasmer0-vs-wasmer2 differences

matklad:
We found this two curious cases when fuzzing wasmer0 versus wasmer2
during upgrade. Let's encode them as tests.


runtime/near-vm-runner/src/tests/runtime_errors.rs
```diff
@@ -749,3 +749,80 @@ fn test_contract_error_caching() {
...
+            ),
+            // wasmer0 incorrectly doesn't catch overflow during address calculation
+            VMKind::Wasmer0 => gas_and_error_match(actual, Some(61754076), None),
+        };
+    });
+}
+
+/// Uses `f32.copysign` to observe a sign of `NaN`.
+///
+/// Wasmer specification allows different behaviors here:
```
nagisa, line 789:
```suggestion
/// WASM specification allows different behaviors here:
```

nagisa (approved):
The tests themselves seem reasonable to me. I'm not sure why the order change in `with_vm_variants` is necessary. If it serves a purpose of some sort, a comment would be great.

olonho (approved):

PR #5542, opened 2021-11-30
re: continue refactoring `near-network`

pmnoxx:
There are a couple of improvements we can make in `near-network` to improve code quality.

### DIFF
--- a/chain/network/benches/graph.rs
+++ b/chain/network/benches/graph.rs
@@ -12,8 +12,8 @@ fn build_graph(depth: usize, size: usize) -> Graph {
 
     let mut graph = Graph::new(source.clone());
 
-    for i in 0..size {
-        graph.add_edge(&source, &nodes[i]);
+    for node in &nodes[..size] {
+        graph.add_edge(&source, node);
     }
 
     for layer in 0..depth - 1 {
--- a/chain/network/src/peer_manager/peer_manager_actor.rs
+++ b/chain/network/src/peer_manager/peer_manager_actor.rs
@@ -408,11 +408,11 @@ impl PeerManagerActor {
                     if !self.routing_table_view.is_local_edge_newer(other_peer, edge.nonce()) {
                         continue;
                     }
-                    // We belong to this edge.
+                    // Check whether we belong to this edge.
                     if self.connected_peers.contains_key(other_peer) {
                         // This is an active connection.
                         if edge.edge_type() == EdgeState::Removed {
-                            self.maybe_remove_connected_peer(ctx, edge.clone(), other_peer);
+                            self.maybe_remove_connected_peer(ctx, edge, other_peer);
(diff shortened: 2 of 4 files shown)


chain/network/benches/graph.rs
```diff
@@ -12,8 +12,8 @@ fn build_graph(depth: usize, size: usize) -> Graph {
 
     let mut graph = Graph::new(source.clone());
 
-    for i in 0..size {
-        graph.add_edge(&source, &nodes[i]);
+    for node in nodes.iter().take(size) {
```
matklad, line 15:
```suggestion
    for node in &nodes[..size] {
```

shorter, and also asserts that size is in bounds.

chain/network/src/peer_manager/peer_manager_actor.rs
```diff
@@ -392,22 +392,14 @@ impl PeerManagerActor {
...
-                            _ => {}
+                        if edge.edge_type() == EdgeType::Removed {
+                            self.maybe_remove_connected_peer(ctx, edge.clone(), other);
                         }
+                    } else if edge.edge_type() == EdgeType::Added {
+                        // We are not connected to this peer, but routing table contains
+                        // information that we do. We should wait and remove that peer
+                        // from routing table
+                        self.wait_peer_or_remove(ctx, edge.clone());
                     }
```
matklad, line 403: optional: another way to write such "decision tables" is to match on a tuple of conditions

```rust
let on_edge = self.active_peers.contains_key(other);
match (on_edge, edge.edge_type()) {
    (false, EdgeType::Added) => self.wait_peer_or_remove(ctx, edge.clone());
    (true, EdgeType::Removed) => self.maybe_remove_connected_peer(ctx, edge.clone(), other);
    (false, EdgeType::Removed) | (true, EdgeType::Added) => (),
}
```

The benefit is that it's more clear that you exhaustively check all combinations. The drawback is that, for many conditions, it might be hard to map from `(true, false, false)` pattern to what the conditions are.

matklad (approved):

chain/network/src/peer_manager/peer_manager_actor.rs
```diff
@@ -389,25 +389,18 @@ impl PeerManagerActor {
...
-                    } else {
-                        match edge.edge_type() {
-                            EdgeType::Added => {
-                                // We are not connected to this peer, but routing table contains
-                                // information that we do. We should wait and remove that peer
-                                // from routing table
-                                self.wait_peer_or_remove(ctx, edge.clone());
-                            }
-                            _ => {}
+                    // Check whenever we belong to this edge.
```
mm-near, line 392: nit: I think this should be 'check whether' rather than 'whenever'

mm-near (approved):

PR #5543, opened 2021-11-30
feat: Resolve `large_enum_variant` warning

pmnoxx:
In `NetworkViewClientResponses`, `PeerMessage` `enum`'s we have variants that can be quite large. This is bad, because size of enum is determined, by size of the largest variant. By putting `EpochSyncResponse` and `EpochSyncFinalizationResponse` inside `Box`, we minimize the size of the enum.

```
    /// ### Why is this bad?
    /// Enum size is bounded by the largest variant. Having a
    /// large variant can penalize the memory layout of that enum.
    ///
    /// ### Known problems
    /// This lint obviously cannot take the distribution of
    /// variants in your running program into account. It is possible that the
    /// smaller variants make up less than 1% of all instances, in which case
    /// the overhead is negligible and the boxing is counter-productive. Always
    /// measure the change this lint suggests.
```

### DIFF
--- a/chain/network/src/types.rs
+++ b/chain/network/src/types.rs
@@ -327,9 +327,9 @@ pub enum PeerMessage {
     HandshakeV2(HandshakeV2),
 
     EpochSyncRequest(EpochId),
-    EpochSyncResponse(EpochSyncResponse),
+    EpochSyncResponse(Box<EpochSyncResponse>),
     EpochSyncFinalizationRequest(EpochId),
-    EpochSyncFinalizationResponse(Box<EpochSyncFinalizationResponse>),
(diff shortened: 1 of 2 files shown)


chain/network/src/types.rs
```diff
@@ -328,9 +327,9 @@ pub enum PeerMessage {
     HandshakeV2(HandshakeV2),
 
     EpochSyncRequest(EpochId),
-    EpochSyncResponse(EpochSyncResponse),
+    EpochSyncResponse(Box<EpochSyncResponse>),
```
mm-near, line 330: @matklad - PTAL (from my perspective LGTM)

mm-near (approved):

bowenwang1996 (approved):

PR #5545, opened 2021-11-30
fix: Fix`lib.proxy.bridge` crashing in when header is not full read

pmnoxx:
`lib.proxy.bridge` function has been crashing while doing load testing. Some times, we could read less than 4 bytes of header, which could cause python proxy to crash. Now, we will read header, the same method that we read body.

That should fix some random test failures.


pytest/lib/proxy.py
```diff
@@ -212,6 +212,9 @@ async def bridge(reader, writer, handler_fn, global_stopped, local_stopped,
         while 0 == global_stopped.value and 0 >= local_stopped.value and 0 == error.value and 0 == bridge_stopped[
                 0]:
             header = await reader.read(4)
+            # It's possible that we will not read whole header at once, which can use a crash.
+            while len(header) < 4:
+                header += await reader.read(4 - len(header))
```
mina86, line 217: Perhaps define a helper function:

    async def _read_exact(reader, length):
        """Reads exactly `length` bytes from reader."""
        data = b''
        while True:
            buf = await reader.read(length - len(data))
            if not buf:
                raise RuntimeError('Endpoint closed')
            data += buf
            if len(data) >= length:
                return data

Though I’m not sure whether closed connection is handled correctly with this.

pytest/lib/proxy.py
```diff
@@ -212,6 +212,9 @@ async def bridge(reader, writer, handler_fn, global_stopped, local_stopped,
...
             if not header:
```
mina86, line 218: This, and assert on line 224, can never trigger.

mina86 (approved):

mina86:
By the way, `proxy.py::bridge` is not really Pythonic; `proxy.bridge` should suffice.

pytest/lib/proxy.py
pmnoxx: @mina86 After, I wrote my own load tester I saw a crash on line `assert len(header) == 4`

pmnoxx: Maybe that was caused by a closed connection, in the middle of reading.

pytest/lib/proxy.py
pmnoxx: I'm  not sure either. Anyway, thanks! I think it's a good idea to use the same reader for header and body.

pmnoxx:
> By the way, `proxy.py::bridge` is not really Pythonic; `proxy.bridge` should suffice.

You are right, I'm too used to `rust`

pytest/lib/proxy.py
```diff
@@ -203,6 +203,18 @@ def check_finish(server, global_stopped, local_stopped, error):
         local_stopped.value = 2
 
 
+async def _read_exact(reader, length):
+    """Reads exactly `length` bytes from reader."""
+    data = b''
+    while True:
+        buf = await reader.read(length - len(data))
+        if not buf:
+            raise RuntimeError('Endpoint closed')
```
mina86, line 212: So yeah, I think I don’t understand the code and perhaps this check should not be here? There’s bunch of NayDuck failures with this exception in logs.

PR #5549, opened 2021-11-30
re: Rename `EdgeType` to `EdgeState` in `near-network`

pmnoxx:
Add documentation to `EdgeType`/`EdgeState`. I'm proposing to rename `EdgeType` to `EdgeState` in order to make code more understandable.

It's easier to think about edges, by saying that each edge is in one of the two states. Starts in `Active` state, and it's state can change from `Removed` to `Active`, and vice-versa, each type increasing `nonce` by exactly 1.

Also, it's worth renaming `Added` to `Active`. Term "active state` is more self descriptive.

### DIFF
--- a/chain/network/src/routing/edge.rs
+++ b/chain/network/src/routing/edge.rs
@@ -132,7 +132,7 @@ impl Edge {
 
     /// Create the remove edge change from an added edge change.
     pub fn remove_edge(&self, my_peer_id: PeerId, sk: &SecretKey) -> Edge {
-        assert_eq!(self.edge_type(), EdgeType::Added);
+        assert_eq!(self.edge_type(), EdgeState::Active);
         let mut edge = self.0.as_ref().clone();
         edge.nonce += 1;
         let me = edge.key.0 == my_peer_id;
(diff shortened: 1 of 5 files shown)


chain/network/src/routing/edge.rs
```diff
@@ -283,18 +283,22 @@ impl SimpleEdge {
...
+/// State of a given edge.
+/// Every edge starts in `Active` state. It can be removed and go to `Removed` state, and then
+/// added back to go to `Active` state, etc.
 #[derive(BorshSerialize, BorshDeserialize, Clone, PartialEq, Eq, Debug, Hash)]
-pub enum EdgeType {
-    Added,
+pub enum EdgeState {
+    /// `Edge` is `Active` if there is an active connection between two peers on the network.
+    Active,
+    /// `Edge` is in `Removed` state if it was previously in `Active` state, but has been removed.
```
mm-near, line 302: AFAIK node is removed only if we get an 'explicit' Remove request that is signed by one of the peers right?

I'd also mention that:

if we didn't 'see' the connection for a while, it will still be 'active' - but we might remove it from the memory.

mm-near (approved):

chain/network/src/routing/edge.rs
pmnoxx: Thanks, I added extra explanation
```
   /// A signature of one of the peers is requires, otherwise the edge will stay active.
    /// Though, it may be removed  from memory if both peers become unreachable.
```

PR #5552, opened 2021-12-01
Re: Fix file not found error for `pytest/tests/sanity/restacked.py`

pmnoxx:
Fix file not found error for `pytest/tests/sanity/restacked.py`

http://nayduck.near.org/#/run/2271


mina86 (approved):

mina86:
Though it’s probably better to change `cluster.py` to include the trailing slash in `near_root`.

PR #5554, opened 2021-12-01
feat: Use `AcqRel` instead of `SeqCst` in `ThrottleController`

pmnoxx:
It's safe to use `AcqRel` instead of  `SeqCst` inside `ThrottleController`.

Also, we should get rid of `if self.semaphore.available_permits() == 0 {`, condition, because it's not correct due to how semaphore is implemented.
- we need to call `self.semaphore.add_permits(1);` to release semaphore to flush all writes to the current thread
- and then ensure that `ThrottleFrameRead` has a chance to read modified  `num_messages_in_progress` and `total_sizeof_messages_in_progress`

Detailed explanation, TL; DR:

Here is a race condition
- thread A: 
decreases `num_messages_in_progress` and ` total_sizeof_messages_in_progress`;
fetches `semaphore` with acquire
see's it's value is non-zero, and therefore does nothing.

- thread B:
gets `semaphore` and decreases it with `AcqRel`,
it fetches `num_messages_in_progress` and `total_sizeof_messages_in_progress` with `Acquire`, but gets values before thread A modifies them.

The issues is because, the thread didn't do `Release` on the atomic value inside the `semaphore`, which would cause all other writes from thread `A` to be flushed, and be ready to read by the time `A` gets data from `semaphore`.
By always changing permits inside `semaphore`, at the end of `A` we guarantee, that this race condition won't happen.


mm-near (approved):

pmnoxx:
```
```
 Can you take a look?

matklad (commented):
Looks reasonable! 

Does this *has* to be `AcRel`, or can it actually be `Relaxed`? It seems that, given that we synchronize via a semaphore, `Relaxed` should be fine?

pmnoxx:
> Looks reasonable!
> 
> Does this _has_ to be `AcRel`, or can it actually be `Relaxed`? It seems that, given that we synchronize via a semaphore, `Relaxed` should be fine?

I thought about it, we can safely change all variables to be relaxed. 


Unfortunately, the `semaphore` uses `Relaxed`  internally, which is not correct, it should be `AcqRel`. We should replace that semaphore, with a correct semaphore in the future.

matklad:
>Unfortunately, the semaphore uses Relaxed internally, which is not correct,

Where does it happen? It seems that they do use stronger-than-relaxed ordering? 

https://github.com/tokio-rs/tokio/blob/master/tokio/src/sync/batch_semaphore.rs

pmnoxx:
> > Unfortunately, the semaphore uses Relaxed internally, which is not correct,
> 
> Where does it happen? It seems that they do use stronger-than-relaxed ordering?
> 
> https://github.com/tokio-rs/tokio/blob/master/tokio/src/sync/batch_semaphore.rs

I think I misread the code, last time I was looking into it. Yes, everything is fine. They use Acquire/Release properly.
It's safe to switch this code to use `Ordering::Relaxed`.

PR #5555, opened 2021-12-01
feat: Remove `serde` from multiple crates in `/chain`

pmnoxx:
Remove `serde` from:

- `/chain/chain`
- `/chain/chunks`
- `/chain/client`
- `/chain/epoch_manager`


chain/chain/src/types.rs
```diff
@@ -797,7 +796,7 @@ pub trait RuntimeAdapter: Send + Sync {
 
 /// The last known / checked height and time when we have processed it.
 /// Required to keep track of skipped blocks and not fallback to produce blocks at lower height.
-#[derive(BorshSerialize, BorshDeserialize, Serialize, Debug, Clone, Default)]
+#[derive(BorshSerialize, BorshDeserialize, Debug, Clone, Default)]
```
mm-near, line 799: how do we know that this no longer needs Serialize?

(this is a public struct, so some other project could be using it, right?)

chain/client/Cargo.toml
```diff
@@ -13,7 +13,6 @@ futures = "0.3"
 chrono = { version = "0.4.4", features = ["serde"] }
```
mm-near, line 13: strange.. as you're just removing 'serde' from list of deps - maybe this line can be also removed?

mm-near (changes requested):

chain/client/Cargo.toml
pmnoxx: No, we still use chrono. 

Long explanation:
When we introduced `Clock`, we exported `Instant` from 'std', and `Utc` from `Datetime`, but the migration was only done half way. We never exported `chrono::Datetime` type, nor `chrono::Duration` etc. Right now, half of the import code from our `Clock` crate, half from `chrono`. This needs to be addresses separately.

chain/chain/src/types.rs
pmnoxx: This type is used only as an attribute of of `ChainStore` which itself is not serializable.

While, it's technically possible for someone create a new project, just to use this one type with `Serializable`. It doesn't make much sense for anyone to do so,In the future, we may add an optional `serde` support for those structs if someone requests it.  However, our code doesn't need it, so there is no need to have it by default.

```
pub struct ChainStore {
    /// Latest known.
    latest_known: Option<LatestKnown>,
```

mm-near (approved):

PR #5556, opened 2021-12-01
feat: Replace `DateTime`/`Instant`/`SystemTime` with `Time` in `near-network`

pmnoxx:
We are going to start using `Time` struct to replace `DateTime`/`Instant`/`SystemTime`  in `near-network`

Waiting for https://github.com/near/nearcore/pull/5177

### DIFF
--- a/chain/network-primitives/src/types.rs
+++ b/chain/network-primitives/src/types.rs
@@ -10,16 +10,14 @@
 /// - We also export publicly types from `crate::network_protocol`
 use actix::Message;
 use borsh::{BorshDeserialize, BorshSerialize};
-use chrono::DateTime;
 use near_crypto::SecretKey;
 use near_primitives::block::{Block, BlockHeader, GenesisId};
 use near_primitives::hash::CryptoHash;
 use near_primitives::network::{AnnounceAccount, PeerId};
 use near_primitives::syncing::{EpochSyncFinalizationResponse, EpochSyncResponse};
-use near_primitives::time::{Clock, Utc};
+use near_primitives::time::Time;
 use near_primitives::transaction::ExecutionOutcomeWithIdAndProof;
 use near_primitives::types::{AccountId, BlockHeight, EpochId, ShardId};
-use near_primitives::utils::{from_timestamp, to_timestamp};
 use near_primitives::views::{FinalExecutionOutcomeView, QueryResponse};
 use std::fmt::Debug;
 use std::hash::Hash;
@@ -141,7 +139,7 @@ pub enum KnownPeerStatus {
     Unknown,
     NotConnected,
     Connected,
-    Banned(ReasonForBan, u64),
+    Banned(ReasonForBan, Time),
 }
 
 impl KnownPeerStatus {
@@ -157,22 +155,24 @@ pub struct KnownPeerState {
     pub peer_info: PeerInfo,
     pub status: KnownPeerStatus,
     /// Unused
-    first_seen: u64,
-    pub last_seen: u64,
+    first_seen: Time,
+    pub last_seen: Time,
 }
 
 impl KnownPeerState {
     pub fn new(peer_info: PeerInfo) -> Self {
+        let now = Time::now();
         KnownPeerState {
             peer_info,
             status: KnownPeerStatus::Unknown,
-            first_seen: to_timestamp(Clock::utc()),
-            last_seen: to_timestamp(Clock::utc()),
+            first_seen: now,
+            last_seen: now,
         }
     }
 
-    pub fn last_seen(&self) -> DateTime<Utc> {
-        from_timestamp(self.last_seen)
+    pub fn banned_at(&mut self, ban_reason: ReasonForBan, now: Time) {
+        self.last_seen = now;
+        self.status = KnownPeerStatus::Banned(ban_reason, now);
     }
 }
 
@@ -447,5 +447,21 @@ mod tests {
                 92, 92, 92, 92, 92, 92, 92, 92, 92, 92, 92, 92, 92, 92, 92, 92, 92, 92, 92, 0, 0,
             ],
         );
+    }
+
+    #[test]
+    fn test_known_peer_state() {
+        let pi = PeerInfo::random();
+        let mut kps = KnownPeerState::new(pi);
+
+        let now = Time::now();
+        kps.last_seen = now;
+
+        assert_eq!(kps.last_seen, now);
+
+        let now = now + Duration::from_nanos(123);
+        kps.banned_at(ReasonForBan::Abusive, now);
+        assert_eq!(kps.status, KnownPeerStatus::Banned(ReasonForBan::Abusive, now));
+        assert_eq!(kps.last_seen, now);
     }
 }
--- a/chain/network/src/peer/rate_counter.rs
+++ b/chain/network/src/peer/rate_counter.rs
@@ -1,0 +1,87 @@
+// Copyright 2018 The Grin Developers
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use near_primitives::time::Time;
+use std::collections::VecDeque;
+use std::time::Duration;
+
+const MINUTE: Duration = Duration::from_secs(60);
+
+/// Stores list of entries for `RateCounter`
+struct Entry {
+    /// bytes since last reset
+    bytes: u64,
+    /// Time we created the entry.
+    recorded: Time,
+}
+
+/// A rate counter tracks number of transfers, the amount of data exchanged and the rate of transfer
+/// over the last minute.
+pub struct RateCounter {
+    entries: VecDeque<Entry>,
+    bytes_sum: u64,
+}
+
+impl RateCounter {
+    pub fn new() -> Self {
+        RateCounter { entries: VecDeque::new(), bytes_sum: 0 }
+    }
+
+    /// Increment number of bytes transferred, updating counts and rates.
+    pub fn increment(&mut self, bytes: u64) {
+        let now = Time::now();
+        self.entries.push_back(Entry { bytes, recorded: now });
+        self.bytes_sum += bytes;
+        self.truncate(now);
+    }
+
+    pub fn bytes_per_min(&self) -> u64 {
+        self.bytes_sum
+    }
+
+    pub fn count_per_min(&self) -> u64 {
+        self.entries.len() as u64
+    }
+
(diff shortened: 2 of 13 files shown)


chain/network-primitives/src/types.rs
```diff
@@ -671,26 +669,27 @@ impl KnownPeerStatus {
...
     pub status: KnownPeerStatus,
-    pub first_seen: u64,
-    pub last_seen: u64,
+    pub first_seen: UnixTime,
+    pub last_seen: UnixTime,
 }
 
 impl KnownPeerState {
     pub fn new(peer_info: PeerInfo) -> Self {
+        let now = Time::now().to_unix_timestamp();
```
matklad, line 678: :+1: for calling `now` once! 

Even better would be to pass `now` as an argument, but that's a separaet PR

chain/network/src/peer_manager/peer_store.rs
```diff
@@ -153,8 +153,9 @@ impl PeerStore {
         ban_reason: ReasonForBan,
     ) -> Result<(), Box<dyn std::error::Error>> {
         if let Some(peer_state) = self.peer_states.get_mut(peer_id) {
-            peer_state.last_seen = to_timestamp(Utc::now());
-            peer_state.status = KnownPeerStatus::Banned(ban_reason, to_timestamp(Utc::now()));
+            let now = Time::now().to_unix_timestamp();
+            peer_state.last_seen = now;
+            peer_state.status = KnownPeerStatus::Banned(ban_reason, now);
```
matklad, line 158:
```suggestion
            peer_state.banned_at(Time::now());
```

core/primitives/src/time.rs
```diff
@@ -1,155 +1,316 @@
...
-            if clock.is_mock {
-                clock.utc.push_back(mock_date);
-            } else {
-                panic!("Use MockClockGuard in your test");
-            }
-        });
+    impl BorshSerialize for Time {
+        fn serialize<W: std::io::Write>(&self, writer: &mut W) -> Result<(), std::io::Error> {
+            let nanos = self.to_unix_timestamp() as u64;
+            BorshSerialize::serialize(&nanos, writer).unwrap();
```
matklad, line 199: This' definitely shouldn't `.unwrap`, the writer may fail with IO error.

chain/network-primitives/src/types.rs
```diff
@@ -671,26 +669,27 @@ impl KnownPeerStatus {
 pub struct KnownPeerState {
     pub peer_info: PeerInfo,
     pub status: KnownPeerStatus,
-    pub first_seen: u64,
-    pub last_seen: u64,
+    pub first_seen: UnixTime,
+    pub last_seen: UnixTime,
```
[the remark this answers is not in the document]
matklad: Actually, we have `Time: BorshSerialize`, so we can just use `Time` for these fields, no?

core/primitives/src/time.rs
```diff
@@ -1,155 +1,316 @@
...
+            Self::UNIX_EPOCH + Duration::from_nanos(Clock::utc().timestamp_nanos() as u64)
+        }
+
+        pub fn duration_since(&self, rhs: &Self) -> Duration {
+            self.system_time.duration_since(rhs.system_time).unwrap_or(Duration::from_millis(0))
+        }
+
+        pub fn saturating_duration_since(&self, rhs: &Self) -> Duration {
+            self.system_time.duration_since(rhs.system_time).unwrap_or(Duration::from_millis(0))
+        }
```
matklad, line 240: The methods are identical, seems better to have just `duration_since` and document that all ops are saturating

chain/network-primitives/src/types.rs
pmnoxx: @matklad Yes, normally I would agree.

However, I checked all places which use `KnownPeerState::new`.
All those places call `new` exactly once, and the now wouldn't be useful for other purposes.

chain/network/src/peer/rate_counter.rs
```diff
@@ -22,7 +23,7 @@ struct Entry {
     /// bytes since last reset
     bytes: u64,
     /// Time we created the entry.
-    recorded: SystemTime,
+    recorded: Time,
```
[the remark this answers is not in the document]
pmnoxx: I agree. Btw, this code is going to be removed:
```
https://github.com/near/nearcore/pull/5225/files#diff-a771fcdab1669c3987af82fb22ca6f6960756ec7983128635cbf9e9422b30c49L21-L26
```

chain/network-primitives/src/types.rs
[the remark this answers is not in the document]
pmnoxx: @matklad Yes, that was my intention.

However, `KnownPeerState` derives `Borsh` and is saved to database.
In order to limit the scope of this PR, I decided to do that in the next PR.

I'm fine doing it in the same PR if you prefer.

We also use `u64` in `network-protocol`. I wanted to create a separate PR, where we change the time from `u64` to `Time`.

Otherwise, we will end up be discussing in this PR, whenever to do `network protocol` change or not.

chain/network/src/peer_manager/peer_store.rs
```diff
@@ -85,7 +85,7 @@ impl PeerStore {
             let peer_id: PeerId = key.try_into()?;
             let mut peer_state: KnownPeerState = value.try_into()?;
             // Mark loaded node last seen to now, to avoid deleting them as soon as they are loaded.
-            peer_state.last_seen = to_timestamp(Utc::now());
+            peer_state.last_seen = Time::now().to_unix_timestamp();
```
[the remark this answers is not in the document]
pmnoxx: @matklad I had to change it to `set_seen_at`, because `seen_at` already exists.

chain/network-primitives/src/types.rs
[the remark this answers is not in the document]
pmnoxx: @matklad Ok, I remove `UnixTime`.

chain/network-primitives/src/types.rs
matklad: >All those places call new exactly once,

Not really, here's one example where we call `new` in a loop, and already have time on our hand. 

```diff
diff --git a/chain/network/src/peer_manager/peer_store.rs b/chain/network/src/peer_manager/peer_store.rs
index 88e8a6dbf..209998715 100644
--- a/chain/network/src/peer_manager/peer_store.rs
+++ b/chain/network/src/peer_manager/peer_store.rs
@@ -58,6 +58,7 @@ impl PeerStore {
         let mut peer_states = HashMap::default();
         let mut addr_peers = HashMap::default();
 
+        let now = to_timestamp(Utc::now());
         for peer_info in boot_nodes.iter() {
             if !peer_states.contains_key(&peer_info.id) {
                 if let Some(peer_addr) = peer_info.addr {
@@ -70,7 +71,7 @@ impl PeerStore {
                             entry.insert(VerifiedPeer::signed(peer_info.id.clone()));
                             peer_states.insert(
                                 peer_info.id.clone(),
-                                KnownPeerState::new(peer_info.clone()),
+                                KnownPeerState::new(peer_info.clone(), now),
                             );
                         }
                     }
@@ -78,7 +79,6 @@ impl PeerStore {
             }
         }
 
-        let now = to_timestamp(Utc::now());
         for (key, value) in store.iter(ColPeers) {
             let key: Vec<u8> = key.into();
             let value: Vec<u8> = value.into();
```

Things like `add_indirect_peers` also seem to loop. 

Though, in any case this doesn't matter much.

pmnoxx: See https://github.com/near/nearcore/pull/5839/files

chain/network-primitives/src/types.rs
```diff
@@ -347,26 +345,36 @@ impl KnownPeerStatus {
 pub struct KnownPeerState {
     pub peer_info: PeerInfo,
     pub status: KnownPeerStatus,
-    pub first_seen: u64,
-    pub last_seen: u64,
+    first_seen: Time,
+    last_seen: Time,
```
matklad, line 349: Sorry for back and forth here, but I feel this should be 

```rust
    pub first_seen: Time,
    pub last_seen: Time,
```

and the getters/setters should be removed.

That is 

If we store raw `u64` internally, we should make fields private and provide well-typed accessor functions, to not expose raw representation. 

If we store `Time` internally, we should just make everything pub, because `fn foo(&self) / fn set_foo(&mut self)` pair is strictly worse than just making the field pub.

core/primitives/src/time.rs
```diff
@@ -287,12 +287,12 @@ mod tests {
     }
 }
 
-mod time {
+mod time_struct {
```
matklad, line 290: tbh, not sure this module pulls its weight at all. This is a realatively small (<1k lines) file, it feels like it doesnt' really need extra organization.

matklad (approved):

matklad (approved):

pmnoxx:
```
```
@mm-near Can you take a look?

pmnoxx:
```
```
@mm-near Can you take a look?

core/primitives/src/time.rs
```diff
@@ -179,9 +181,85 @@ impl Clock {
     }
 }
 
+#[derive(Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash, Debug)]
```
mm-near, line 184: please comment on when people should use Time vs other solutions.

core/primitives/src/time.rs
```diff
@@ -179,9 +181,85 @@ impl Clock {
...
+pub struct Time {
+    system_time: SystemTime,
+}
+
+impl BorshSerialize for Time {
```
mm-near, line 189: do we put Time somewhere in Borsh format already?

core/primitives/src/time.rs
```diff
@@ -285,124 +363,29 @@ mod tests {
...
-        fn add(self, other: Duration) -> Self {
-            Self { system_time: self.system_time + other }
-        }
-    }
+        let now_nc: Time = now_st.into();
 
-    impl Sub for Time {
-        type Output = Duration;
+        let t_nc = now_nc + Duration::from_nanos(123456);
+        let t_st = now_st + Duration::from_nanos(123456);
```
mm-near, line 374: can you also test subtraction?

mm-near: and that duration_since never returns 'negative' numbers (basically that it returns 0 when comparing newer and older)

mm-near (changes requested):

core/primitives/src/time.rs
pmnoxx: Not in the `master` branch. In this `PR`, we start using it with `KnownPeerState`, which gets written to, and loaded from the database. We don't use it for network protoctol yet.

core/primitives/src/time.rs
pmnoxx: Ok, I added a test.

mm-near:
did you push your changes? (as I don't see anything changed since last review)

pmnoxx:
> did you push your changes? (as I don't see anything changed since last review)

I had to rewrite this PR, due to huge merge conflicts.

core/primitives/src/time.rs
mm-near: right - so let's mention that Time's Borsch serialization is compatible with 'nanos' serialization - and let's add a test for that (serialize nanos, deserialize using borsch, compare)

mm-near (approved):

PR #5558, opened 2021-12-01
feat: Propagate `ActixMessageWrapper` between `PeerManagerActor` and `RoutingTableActor`

pmnoxx:
We want to track count and sizes of messages arriving from `PeerActor`, while they  travel to `PeerManagerActor`, and then to `RoutingTableActor`.

This PR extends message tracking per peer, while they travel through `actix` `InBox` system.

Continuation of https://github.com/near/nearcore/pull/5198

TODO:
- [x] documentation for methods

### DIFF
--- a/chain/network/src/routing/routing_table_actor.rs
+++ b/chain/network/src/routing/routing_table_actor.rs
@@ -773,7 +773,10 @@ impl Handler<ActixMessageWrapper<RoutingTableMessages>> for RoutingTableActor {
         let result = self.handle(msg, ctx);
 
         // TODO(#5155) Add support for DeepSizeOf to result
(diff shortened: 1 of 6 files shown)


pmnoxx:
@matklad @mm-near Can you take a look?

chain/network/src/routing/routing_table_actor.rs
```diff
@@ -477,6 +481,7 @@ pub enum RoutingTableMessages {
         prune: Prune,
         prune_edges_not_reachable_for: Duration,
     },
+    ValidateEdgeList(ValidateEdgeList),
```
mm-near, line 484: comment

mm-near (approved):

PR #5559, opened 2021-12-01
feat: Add `num_epoch_to_store` to `config.json` - with test

pmnoxx:
Same as https://github.com/near/nearcore/pull/5348, but with extra test.

### DIFF
--- a/integration-tests/tests/client/process_blocks.rs
+++ b/integration-tests/tests/client/process_blocks.rs
@@ -1597,10 +1597,11 @@ fn test_num_blocks_in_storage_config_setting() {
         env.process_block(0, block.clone(), Provenance::PRODUCED);
         env.process_block(1, block, Provenance::NONE);
     }
-    // Should crash, but doesn't?
+    // Check for GC'ed blocks.
     for i in 1..100 {
         assert!(env.clients[0].chain.get_block_by_height(i).is_err());
     }
+    // Check whenever we still keep blocks.
     for i in 101..epoch_length * (num_epochs_to_keep_store_data) + 2 {
         env.clients[0].chain.get_block_by_height(i).unwrap();
     }
--- a/integration-tests/tests/client/process_blocks.rs
+++ b/integration-tests/tests/client/process_blocks.rs
@@ -1559,6 +1559,30 @@ fn test_gc_after_state_sync() {
 }
 
 #[test]
+fn test_blocks_in_storage() {
+    let epoch_length = 1024;
+    let mut genesis = Genesis::test(vec!["test0".parse().unwrap(), "test1".parse().unwrap()], 1);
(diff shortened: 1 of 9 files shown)


nikurt (approved):

bowenwang1996:
@pmnoxx why not just modify the original PR to include the test?

integration-tests/tests/client/process_blocks.rs
```diff
@@ -1558,6 +1575,34 @@ fn test_gc_after_state_sync() {
...
+        ))
+        .build();
+    for i in 1..epoch_length * (num_epochs_to_keep_store_data - 1) + 2 {
+        let block = env.clients[0].produce_block(i).unwrap().unwrap();
+        env.process_block(0, block.clone(), Provenance::PRODUCED);
+        env.process_block(1, block, Provenance::NONE);
+    }
+    // Should crash, but doesn't?
+    for i in 4..epoch_length * (num_epochs_to_keep_store_data - 1) + 2 {
+        env.clients[0].chain.get_block_by_height(i).unwrap();
```
mzhangmzz, line 1602: Does this test pass? 

Should you check the first few blocks are gc'ed?

pmnoxx: @mzhangmzz This passes the test, but
Works:
``` (num_epochs_to_keep_store_data - 1) + 2```
Doesn't pass:
``` (num_epochs_to_keep_store_data) + 2```

You are right, I should also add a negative check in the test case.

pmnoxx: @mzhangmzz I updated the test, to check whenever blocks get GC'ed

jc-near:
Reopening this PR as [CP-25](https://nearinc.atlassian.net/jira/software/projects/CP/boards/17?selectedIssue=CP-25) requires it

mm-near:
@jc-near - let's adopt this PR so that it uses the same 'GCconfig' that is added by #6592

jc-near:
Meaning, once that PR is merged, we change this PR to use the same `GCconfig` struct and parametrize the variable there?

jc-near:
@mm-near  ^^

mina86:
> Meaning, once that PR is merged, we change this PR to use the same `GCconfig` struct and parametrize the variable there?

Yep. The idea now is to add gc_num_epochs_to_store to GCConfig struct. 

With the amount of conflicts this PR has it might be easier to just start anew.

mm-near:
@jc-near - yes, it should use the GCconfig struct. And its up to you @jc-near whether you want to adapt this PR or start a new one.

mm-near (changes requested):
.

mina86:
Obsolete by 535f4d793

PR #5562, opened 2021-12-01
fix: change request_partial_encoded_chunk to request chunk from a sin…

mzhangmzz:
…gle node instead of from all part owners for old blocks

This change is to avoid the network being flooded by partial chunks requests. Especially after we enable chunks to be requested for multiple blocks simultaneously

### DIFF
--- a/chain/chunks/src/lib.rs
+++ b/chain/chunks/src/lib.rs
@@ -449,6 +449,21 @@ impl ShardsManager {
             || self.runtime_adapter.will_care_about_shard(account_id, parent_hash, shard_id, is_me)
     }
 
+    /// Send requests for a chunk
+    /// Depending on whether the node tracks the shard, it either downloads the full chunk
+    /// or only parts that it concerns, specifically, the part that should be owned by the node if
+    /// it is a validator, plus the parts that are sent to the shards that the node is tracking
+    ///
+    /// `height`: block height of the chunk
+    /// `parent_hash`: hash of the prev block of the block that includes the requested chunk
+    /// `shard_id`: shard id of the requested chunk
+    /// `chunk_hash`: chunk hash of the requested chunk
+    /// `force_request_full`: if it is true, the node will request the full chunk regardless whether
+    ///    it is tracking the shard
+    /// `request_from_single_source`: if it is true, the node requests all parts from a single node
+    ///                               instead of sending one request to each part owner
+    /// `request_own_parts_from_others`: whether we should request parts owned by us
+    /// `request_from_archival`: whether to request from archival nodes
     fn request_partial_encoded_chunk(
         &mut self,
         height: BlockHeight,
@@ -456,6 +471,7 @@ impl ShardsManager {
         shard_id: ShardId,
         chunk_hash: &ChunkHash,
         force_request_full: bool,
+        request_from_single_source: bool,
         request_own_parts_from_others: bool,
         request_from_archival: bool,
     ) -> Result<(), near_chain::Error> {
@@ -506,7 +522,7 @@ impl ShardsManager {
             };
 
             if need_to_fetch_part {
-                let fetch_from = if request_from_archival {
+                let fetch_from = if request_from_archival || request_from_single_source {
                     shard_representative_target.clone()
                 } else {
                     let part_owner = self.runtime_adapter.get_part_owner(&parent_hash, part_ord)?;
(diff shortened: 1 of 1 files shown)


chain/chunks/src/lib.rs
```diff
@@ -449,13 +449,29 @@ impl ShardsManager {
             || self.runtime_adapter.will_care_about_shard(account_id, parent_hash, shard_id, is_me)
     }
 
+    /// Send requests for a chunk
+    /// Depending on whether the node tracks the shard, it either downloads the full chunk
+    /// or only parts that it concerns, specifically, the part that should be owned by the node if
```
bowenwang1996, line 454:
```suggestion
    /// or only parts that it cares about, specifically, the part that should be owned by the node if
```

chain/chunks/src/lib.rs
```diff
@@ -449,13 +449,29 @@ impl ShardsManager {
...
     fn request_partial_encoded_chunk(
         &mut self,
         height: BlockHeight,
         parent_hash: &CryptoHash,
         shard_id: ShardId,
         chunk_hash: &ChunkHash,
         force_request_full: bool,
+        request_from_single_source: bool,
         request_own_parts_from_others: bool,
         request_from_archival: bool,
```
bowenwang1996, line 476: I think it makes sense to create an enum here to reduce the number of arguments and potentially make the code cleaner

chain/chunks/src/lib.rs
```diff
@@ -690,6 +706,17 @@ impl ShardsManager {
...
+                    // This is to save the number of requests a node sends
+                    // For old blocks, we can be reasonably sure that some peer has downloaded
+                    // the full chunk and we can get the chunk from them
+                    // TODO: when we enable challenges, this may break data availability
+                    //    if not many nodes receive the chunk before a block gets old (no longer the head)
+                    //    We leave this issue here for now because it is not a problem now when
+                    //    each shard is tracked by the majority of nodes in the network
+                    //    and we may change the way how chunks are distributed to non-validators
+                    //    https://github.com/near/nearcore/issues/3890
+                    old_block,
```
bowenwang1996, line 719: Are we confident that this is indeed a better approach? The issue I see here is that if the node itself is a validator and it is slightly behind (>1 block), this will mean that the node will request parts from a single source, which may be less optimal than requesting from multiple validators. On the other hand, if the node is very behind, or that it is not a validator, then it probably doesn't matter that much. In short I don't think `old_block` captures what we really want here

mzhangmzz: I thought about making the condition `old_block && non_validator`, the issue is that if a validator is behind, then it will request missing chunks for multiple blocks. If all of these requests are sent to part owners, that seems like a lot of requests.

I am still trying to figure out why the OOM occurred for testnet canary. Maybe it is not related to sending too many chunk requests. Then I agree we should restrict the change to only `non-validators`.

chain/chunks/src/lib.rs
```diff
@@ -449,13 +449,29 @@ impl ShardsManager {
...
+    /// Send requests for a chunk
+    /// Depending on whether the node tracks the shard, it either downloads the full chunk
+    /// or only parts that it concerns, specifically, the part that should be owned by the node if
+    /// it is a validator, plus the parts that are sent to the shards that the node is tracking
+    ///
+    /// `height`: block height of the chunk
+    /// `parent_hash`: hash of the prev block of the block that includes the requested chunk
+    /// `shard_id`: shard id of the requested chunk
+    /// `chunk_hash`: chunk hash of the requested chunk
+    /// `force_request_full`: if it is true, the node will request the full chunk regardless whether
```
mm-near, line 461: s/if it is true/ if true

chain/chunks/src/lib.rs
mm-near: we should be consistent with naming: here you mention 'concerns'/'cares' - and below you say 'tracks'. I guess they all refer to the same thing, right?

chain/chunks/src/lib.rs
```diff
@@ -449,13 +449,29 @@ impl ShardsManager {
...
+    ///    it is tracking the shard
+    /// `request_from_single_source`: if it is true, the node requests all parts from a single node
+    ///                               instead of sending one request to each part owner
+    /// `request_own_parts_from_others`: whether we should request parts owned by us
```
mm-near, line 465: this is the part I don't understand - what does it mean 'owned'

chain/chunks/src/lib.rs
```diff
@@ -690,6 +706,17 @@ impl ShardsManager {
...
+                    // For old blocks, we can be reasonably sure that some peer has downloaded
+                    // the full chunk and we can get the chunk from them
+                    // TODO: when we enable challenges, this may break data availability
+                    //    if not many nodes receive the chunk before a block gets old (no longer the head)
+                    //    We leave this issue here for now because it is not a problem now when
+                    //    each shard is tracked by the majority of nodes in the network
+                    //    and we may change the way how chunks are distributed to non-validators
+                    //    https://github.com/near/nearcore/issues/3890
+                    old_block,
                     old_block,
```
mm-near, line 720: we have a lot of parameters here -- as Rust doesn't support named params, consider adding an inline comment to make reading easier:

false,     // force_request_full
old_block, // request_from_single_source 

etc..

chain/chunks/src/lib.rs
bowenwang1996: Each part has an assigned owner (a validator). Basically an encoded chunk is divided into n parts where n is the number of validators and each validator get a partial encoded chunk

mzhangmzz:
This change is hard to test with the current testing options and it's not easy to prove its theoretical correctness either. So I decided to abandon it

PR #5564, opened 2021-12-01
add comments for ShardLayout functions

mzhangmzz:


core/primitives/src/shard_layout.rs
```diff
@@ -61,10 +61,12 @@ impl ShardLayout {
         Self::v0(1, 0)
     }
 
+    /// Return a V0 Shardlayout
     pub fn v0(num_shards: NumShards, version: ShardVersion) -> Self {
         Self::V0(ShardLayoutV0 { num_shards, version })
     }
 
+    /// Return a V1 Shardlayout
```
mm-near, line 69: we might also want to mention (on top of this file) where is the current ShardLayout configured (the one that we use in mainnet and/or testnet)

mm-near (approved):

core/primitives/src/shard_layout.rs
```diff
@@ -61,10 +61,12 @@ impl ShardLayout {
         Self::v0(1, 0)
     }
```
matklad, line 62: unrelated, but that `default` feels weird. Seems it wants to be `impl Default for ShardLayout` or maybe renamed to `single_shad` (intuitively, it seems that theer isnt' good default shard layout)?

core/primitives/src/shard_layout.rs
```diff
@@ -207,6 +217,7 @@ fn is_top_level_account(top_account: &AccountId, account: &AccountId) -> bool {
...
 }
 
+/// ShardUId is an unique representation for shards from different shard layout
```
matklad, line 220: Might be helpful to contrast this with `ShardId`

core/primitives/src/shard_layout.rs
```diff
@@ -245,13 +262,16 @@ impl TryFrom<&[u8]> for ShardUId {
         Ok(Self { version, shard_id })
     }
 }
```
matklad, line 264: Unrelated to the PR, but, rather than `TryFrom<&[u8]>` I'd probably expect to see 

```rust
impl ShardUid {
  fn to_bytes(&self) -> [u8; 8] {}
  fn from_bytes(bytes: [u8; 8]) -> ShardId {}
}
```

No need to return a runtime error if the precondition can be expressed as a type.

matklad (approved):

core/primitives/src/shard_layout.rs
```diff
@@ -103,12 +106,16 @@ impl ShardLayout {
         )
     }
 
+    /// Given a parent shard id, return the shard uids for the shards in the current shard layout that
```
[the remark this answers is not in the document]
mzhangmzz: I added some doc at the beginning of the file

Longarithm (approved):

PR #5566, opened 2021-12-01
fix: Disable 'We expected `peer_id` to belong to a component' warning

pmnoxx:
Let's disable`We expected peer_id to belong to a component` warning for now

Resolves #5563

Doc in https://github.com/near/nearcore/pull/5618

Issue #5563, opened 2021-12-01
We expected `peer_id` to belong to a component warnings


mzhangmzz:
I started a new testnet rpc node recently and I saw the log is flooded with these warnings 

```
Dec 01 18:52:33.547  WARN near_network::routing::routing_table_actor: We expected `peer_id` to belong to a component 13059, but it doesn't belong anywhere
Dec 01 18:52:33.547  WARN near_network::routing::routing_table_actor: We expected `peer_id` to belong to a component 13059, but it doesn't belong anywhere
Dec 01 18:52:33.548  WARN near_network::routing::routing_table_actor: We expected `peer_id` to belong to a component 13059, but it doesn't belong anywhere
Dec 01 18:52:33.548  WARN near_network::routing::routing_table_actor: We expected `peer_id` to belong to a component 13059, but it doesn't belong anywhere
Dec 01 18:52:33.548  WARN near_network::routing::routing_table_actor: We expected `peer_id` to belong to a component 13059, but it doesn't belong anywhere
Dec 01 18:52:33.548  WARN near_network::routing::routing_table_actor: We expected `peer_id` to belong to a component 13059, but it doesn't belong anywhere
Dec 01 18:52:33.548  WARN near_network::routing::routing_table_actor: We expected `peer_id` to belong to a component 13059, but it doesn't belong anywhere
Dec 01 18:52:33.548  WARN near_network::routing::routing_table_actor: We expected `peer_id` to belong to a component 13059, but it doesn't belong anywhere
Dec 01 18:52:33.548  WARN near_network::routing::routing_table_actor: We expected `peer_id` to belong to a component 13059, but it doesn't belong anywhere
Dec 01 18:52:33.548  WARN near_network::routing::routing_table_actor: We expected `peer_id` to belong to a component 13059, but it doesn't belong anywhere
Dec 01 18:52:33.548  WARN near_network::routing::routing_table_actor: We expected `peer_id` to belong to a component 13059, but it doesn't belong anywhere
Dec 01 18:52:33.548  WARN near_network::routing::routing_table_actor: We expected `peer_id` to belong to a component 13059, but it doesn't belong anywhere
Dec 01 18:52:33.548  WARN near_network::routing::routing_table_actor: We expected `peer_id` to belong to a component 13059, but it doesn't belong anywhere
Dec 01 18:52:33.548  WARN near_network::routing::routing_table_actor: We expected `peer_id` to belong to a component 13059, but it doesn't belong anywhere```

[...]

PR #5567, opened 2021-12-01
fix: Fix database leak in `near-network::fetch_edges_for_peer_from_disk`

pmnoxx:
There is a small disk space leak in database.

Whenever we store a component of deleted edges on disk, we have a mapping peer_id -> compoment.

If we ever add edge, let's say (A, B), such that `B` is in that component, we will bring all edges back.
However, we will not delete mapping from `B -> compoment`.
This PR should fix that. 

Related to the issue: https://github.com/near/nearcore/pull/5566

See https://github.com/near/nearcore/issues/5563

### DIFF
--- a/chain/network/src/routing/routing_table_actor.rs
+++ b/chain/network/src/routing/routing_table_actor.rs
@@ -189,12 +189,17 @@ impl RoutingTableActor {
             let mut update = self.store.store_update();
 
             // Load all edges that were persisted in database in the cell - and add them to the current graph.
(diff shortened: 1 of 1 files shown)


bowenwang1996 (commented):
Please add a test

pmnoxx:
@bowenwang1996 I'll close this issue. I looked at the tests we have, the code is actually correct. Though it's rather obscure. I mean, there is no data leak to the database.

PR #5568, opened 2021-12-01
Fix handling eof in `_read_exact`

pmnoxx:
In python `if [0]:` returns false. We should check whenever the header wasn't ready, by checking whenever
the result of `await reader.read` has length of 0 instead of checking if all bytes are 0.

I think this way of handling the `EOF` should fix the issue.

### DIFF
--- a/pytest/lib/proxy.py
+++ b/pytest/lib/proxy.py
@@ -208,7 +208,10 @@ async def _read_exact(reader, length, *, allow_eof=False):
     data = await reader.read(length)
     if data or not allow_eof:
         while len(data) < length:
-            data += await reader.read(length - len(data))
+            new_data = await reader.read(length - len(data))
+            if not new_data:
+                break
+            data += new_data
     return data
 
 


mina86:
`if not data:` is equivalent to `is not len(data):`

    >>> bool(b'')
    False
    >>> bool(b'\0')
    True
    >>> bool(b'\0\0\0\0')
    True

pmnoxx:
> `if not data:` is equivalent to `if len(data) == 0:`
> 
> ```
> >>> bool(b'')
> False
> >>> bool(b'\0')
> True
> >>> bool(b'\0\0\0\0')
> True
> ```



> `if not data:` is equivalent to `if len(data) == 0:`
> 
> ```
> >>> bool(b'')
> False
> >>> bool(b'\0')
> True
> >>> bool(b'\0\0\0\0')
> True
> ```

That makes sense. 

My understanding is that, we get an empty array when EOF, happens, so the following code, could become an infinite loop.
```
while len(data) < length:
        data += await reader.read(length - len(data))
```

pmnoxx:
@mina86 I rewrote the PR, can you take a look again?

pmnoxx:
Ok, I'll close this PR.

PR #5571, opened 2021-12-01
re: Fix `clippy` warnings in `near-chain` 

pmnoxx:
There are a couple code style fixing we can do in the `near-chain`.
In addition we can remove some unnecessary usages of `clone()`.

Split:
- https://github.com/near/nearcore/pull/5655
- https://github.com/near/nearcore/pull/5656
- https://github.com/near/nearcore/pull/5658

### DIFF
--- a/chain/chain/src/chain.rs
+++ b/chain/chain/src/chain.rs
@@ -151,11 +151,15 @@ impl OrphanBlockPool {
     }
 
     fn add(&mut self, orphan: Orphan) {
-        let height_hashes =
-            self.height_idx.entry(orphan.block.header().height()).or_insert_with(|| vec![]);
+        let height_hashes = self
+            .height_idx
+            .entry(orphan.block.header().height())
+            .or_insert_with(std::vec::Vec::new);
         height_hashes.push(*orphan.block.hash());
-        let prev_hash_entries =
-            self.prev_hash_idx.entry(*orphan.block.header().prev_hash()).or_insert_with(|| vec![]);
+        let prev_hash_entries = self
+            .prev_hash_idx
+            .entry(*orphan.block.header().prev_hash())
+            .or_insert_with(std::vec::Vec::new);
         prev_hash_entries.push(*orphan.block.hash());
         self.orphans.insert(*orphan.block.hash(), orphan);
 
@@ -179,9 +183,9 @@ impl OrphanBlockPool {
                     break;
                 }
             }
-            self.height_idx.retain(|_, ref mut xs| xs.iter().any(|x| !removed_hashes.contains(&x)));
+            self.height_idx.retain(|_, ref mut xs| xs.iter().any(|x| !removed_hashes.contains(x)));
             self.prev_hash_idx
-                .retain(|_, ref mut xs| xs.iter().any(|x| !removed_hashes.contains(&x)));
+                .retain(|_, ref mut xs| xs.iter().any(|x| !removed_hashes.contains(x)));
 
             self.evicted += old_len - self.orphans.len();
         }
@@ -196,13 +200,13 @@ impl OrphanBlockPool {
         let ret = self.prev_hash_idx.remove(&prev_hash).map(|hs| {
             hs.iter()
                 .filter_map(|h| {
-                    removed_hashes.insert(h.clone());
+                    removed_hashes.insert(*h);
                     self.orphans.remove(h)
                 })
                 .collect()
         });
 
-        self.height_idx.retain(|_, ref mut xs| xs.iter().any(|x| !removed_hashes.contains(&x)));
+        self.height_idx.retain(|_, ref mut xs| xs.iter().any(|x| !removed_hashes.contains(x)));
 
         ret
     }
@@ -233,7 +237,7 @@ impl Chain {
         let (store, state_roots) = runtime_adapter.genesis_state();
         let store = ChainStore::new(store, chain_genesis.height);
         let genesis_chunks = genesis_chunks(
-            state_roots.clone(),
+            state_roots,
             runtime_adapter.num_shards(&EpochId::default())?,
             chain_genesis.gas_limit,
             chain_genesis.height,
@@ -258,7 +262,7 @@ impl Chain {
             runtime_adapter,
             orphans: OrphanBlockPool::new(),
             blocks_with_missing_chunks: MissingChunksPool::new(),
-            genesis: genesis.clone(),
+            genesis,
             transaction_validity_period: chain_genesis.transaction_validity_period,
             epoch_length: chain_genesis.epoch_length,
             block_economics_config: BlockEconomicsConfig::from(chain_genesis),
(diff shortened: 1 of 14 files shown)


chain/chain/src/store.rs
```diff
@@ -549,7 +547,10 @@ impl ChainStore {
         &self,
         id: &CryptoHash,
     ) -> Result<Vec<ExecutionOutcomeWithIdAndProof>, Error> {
-        Ok(self.store.get_ser(ColTransactionResult, id.as_ref())?.unwrap_or_else(|| vec![]))
+        Ok(self
+            .store
+            .get_ser(ColTransactionResult, id.as_ref())?
+            .unwrap_or_else(std::vec::Vec::new))
```
matklad, line 553:
```suggestion
            .unwrap_or_default())
```

chain/chain/src/store.rs
```diff
@@ -1013,14 +1014,14 @@ impl ChainStoreAccess for ChainStore {
     }
 
     fn get_blocks_to_catchup(&self, hash: &CryptoHash) -> Result<Vec<CryptoHash>, Error> {
-        Ok(self.store.get_ser(ColBlocksToCatchup, hash.as_ref())?.unwrap_or_else(|| vec![]))
+        Ok(self
+            .store
+            .get_ser(ColBlocksToCatchup, hash.as_ref())?
+            .unwrap_or_else(std::vec::Vec::new))
```
matklad, line 1020: or_default

chain/chain/src/store.rs
```diff
@@ -1013,14 +1014,14 @@ impl ChainStoreAccess for ChainStore {
...
     }
 
     fn is_block_challenged(&mut self, hash: &CryptoHash) -> Result<bool, Error> {
-        return Ok(self
-            .store
-            .get_ser(ColChallengedBlocks, hash.as_ref())?
-            .unwrap_or_else(|| false));
+        return Ok(self.store.get_ser(ColChallengedBlocks, hash.as_ref())?.unwrap_or(false));
```
matklad, line 1024:
```suggestion
        Ok(self.store.get_ser(ColChallengedBlocks, hash.as_ref())?.unwrap_or(false))
```

chain/chain/src/store.rs
```diff
@@ -2252,9 +2253,10 @@ impl<'a> ChainStoreUpdate<'a> {
 
     pub fn gc_outgoing_receipts(&mut self, block_hash: &CryptoHash, shard_id: ShardId) {
         let mut store_update = self.store().store_update();
-        match self.get_outgoing_receipts(block_hash, shard_id).map(|receipts| {
-            receipts.iter().map(|receipt| receipt.receipt_id.clone()).collect::<Vec<_>>()
-        }) {
+        match self
+            .get_outgoing_receipts(block_hash, shard_id)
+            .map(|receipts| receipts.iter().map(|receipt| receipt.receipt_id).collect::<Vec<_>>())
```
matklad, line 2258: I feel it's better to pull this operation into the `Ok` branch:

```
match  self.get_outgoing_receipts(block_hash, shard_id) {
  Ok(receipts) => {
    let receipt_ids = receipts.iter().map(|receipt| receipt.receipt_id).collect::<Vec<_>>();
  }
}
```

it might even make the collect unnecessary.

chain/chain/src/store.rs
```diff
@@ -2484,9 +2486,7 @@ impl<'a> ChainStoreUpdate<'a> {
                     Ok(m) => m.clone(),
                     Err(_) => HashMap::new(),
                 };
-            map.entry(block.header().epoch_id().clone())
-                .or_insert_with(|| HashSet::new())
-                .insert(*hash);
+            map.entry(block.header().epoch_id().clone()).or_insert_with(HashSet::new).insert(*hash);
```
matklad, line 2489:
```suggestion
            map.entry(block.header().epoch_id().clone()).or_default().insert(*hash);
```

chain/chain/src/chain.rs
```diff
@@ -151,11 +151,15 @@ impl OrphanBlockPool {
     }
 
     fn add(&mut self, orphan: Orphan) {
-        let height_hashes =
-            self.height_idx.entry(orphan.block.header().height()).or_insert_with(|| vec![]);
+        let height_hashes = self
+            .height_idx
+            .entry(orphan.block.header().height())
+            .or_insert_with(std::vec::Vec::new);
```
matklad, line 157:
```suggestion
            .or_default();
```

It definitlly doesn't make sense to spell `std::vec::Vec::new` in full.

chain/chain/src/chain.rs
```diff
@@ -729,15 +727,15 @@ impl Chain {
                 {
                     return Err(ErrorKind::InvalidChunk.into());
                 }
-            } else if chunk_header.height_created() == block.header().height() {
-                if !runtime_adapter.verify_chunk_header_signature(
+            } else if chunk_header.height_created() == block.header().height()
+                && !runtime_adapter.verify_chunk_header_signature(
```
matklad, line 731: TBH, I am not a huge fan of blindly merging `if cond_a { if cond_b }`. Often times, the second form is more readable.

matklad: OTOH, it can also be the case that the second form hides a bug, and in this sense it might be good to normalize all ifs as a rule of thumb.

chain/chain/src/chain.rs
```diff
@@ -2278,53 +2276,49 @@ impl Chain {
...
+        } else if level == 0 {
+            let maybe_hash = if index >= tree_size {
+                None
+            } else {
+                Some(*self.mut_store().get_block_hash_from_ordinal(index)?)
+            };
+            tree_nodes.insert((index, level), maybe_hash);
+            Ok(maybe_hash)
         } else {
-            if level == 0 {
```
matklad, line 2283: Yeah, especially here, I feel this could have been intentionally written with split iffs.

chain/chain/src/chain.rs
```diff
@@ -4020,22 +4005,18 @@ impl<'a> ChainUpdate<'a> {
...
-            {
-                return Err(ErrorKind::InvalidNextBPHash.into());
-            }
+        } else if header.next_bp_hash()
+            != &Chain::compute_bp_hash(
+                &*self.runtime_adapter,
+                header.next_epoch_id().clone(),
+                header.epoch_id().clone(),
+                header.prev_hash(),
+            )?
```
matklad, line 4014: It feels like the rhs here wants to be extracted in a name variable. It's a complex, multiline *and failable* expression, makes sense to pull it into a separate statment.

matklad (commented):
Reviewed parts of it, but then I realise that I can't carefully check that `if` transfomations are correct. To make this easire to review, I suggest to split this based on the kind of clippy lint this fixes. 

Something like this:

* A PR which changes all `.or_insert_with(|| vec![]);`
* A PR which revomes all unnesesary clones
* A PR which removes all `-> ()`
* A PR which merges all ifs (this one wold be the hardest to review and the most valuable to exntract)

pmnoxx:
> Reviewed parts of it, but then I realise that I can't carefully check that `if` transfomations are correct. To make this easire to review, I suggest to split this based on the kind of clippy lint this fixes.
> 
> Something like this:
> 
> * A PR which changes all `.or_insert_with(|| vec![]);`
> * A PR which revomes all unnesesary clones
> * A PR which removes all `-> ()`
> * A PR which merges all ifs (this one wold be the hardest to review and the most valuable to exntract)

Good idea. I'll split this PR.

pmnoxx:
@matklad I'll write a PR splitter, which splits PR into multiple ones. That should make reviews like this easier.

PR #5574, opened 2021-12-02
Remove `Copy` from `PublicKey`/`PrivateKey`

pmnoxx:
`Copy` derive is not needed for  `PublicKey`/`PrivateKey`.
We can remove it.


abacabadabacaba:
What's the point of this PR? Keys and hashes are trivially copyable, why not expose that?

matklad:
Yeah, it seems like a good idea -- these structs are pretty big: 

```
size_of::<PublicKey> = 192
size_of::<SecretKey> = 224
size_of::<usize> = 8
```

It's better to not pass them by value -- rustc is known for *not* eliminating all of unnecessary memcpys. At some point, it might even makes sense to box the data internally -- for sufficiently large data, it's cheaper to box it once and pass the pointer, rather than to pass size_of worth of bytes.

matklad (approved):

matklad:
Contrast this with eg https://github.com/near/nearcore/pull/5577, where size_of was borderline-reasonable.

matklad:
Going to automerge this to keep my review queue shorter, but this seems trivial to revert if folks thing this is a bad idea!

PR #5575, opened 2021-12-02
re+doc(account-id): refactor and fully document near-account-id

miraclx:
Exhaustively documents `near-account-id` including fuzz tests and refactors the API a bit.

What's different on the API?:

- `AccountId::is_implicit` is now a method on an already validated `AccountId`, no longer an associated method.

    ```rust
    // before
    AccountId::is_implicit("6bf3a4a59f8a23edc2fb96a94fc14be697353f49f3ec49b02a4764a5648fa4a5");

    // after
    "6bf3a4a59f8a23edc2fb96a94fc14be697353f49f3ec49b02a4764a5648fa4a5"
        .parse::<AccountId>()?
        .is_implicit();
    ```

cc @mm-near

### DIFF
--- a/core/account-id/README.md
+++ b/core/account-id/README.md
@@ -1,0 +1,50 @@
+# near-account-id
+
+This crate provides a type for representing a valid, unique account identifier on the [NEAR](https://near.org) network.
+
+[![crates.io](https://img.shields.io/crates/v/near-account-id?label=latest)](https://crates.io/crates/near-account-id)
+[![Documentation](https://docs.rs/near-account-id/badge.svg)](https://docs.rs/near-account-id)
+[![Version](https://img.shields.io/badge/rustc-1.56+-ab6000.svg)](https://blog.rust-lang.org/2021/10/21/Rust-1.56.0.html)
+[![Apache 2.0 licensed](https://img.shields.io/crates/l/near-account-id.svg)](https://github.com/near/nearcore/blob/master/licenses/LICENSE-APACHE)
+
+## Usage
+
+```rust
+use near_account_id::AccountId;
+
+let alice: AccountId = "alice.near".parse().unwrap();
+
+// Basic reports for why validation failed
+assert!(
+  matches!(
+    "z".parse::<AccountId>(),
+    Err(err) if err.kind().is_too_short()
+  )
+);
+
+assert!(
+  matches!(
+    "bob__carol".parse::<AccountId>(),
+    Err(err) if err.kind().is_invalid()
+  )
+);
+```
+
+## Account ID Rules
+
+- Minimum length is `2`
+- Maximum length is `64`
+- An Account ID consists of Account ID parts separated by `.`, example:
+  - `root` ✔
+  - `alice.near` ✔
+  - `bob.stage.testnet` ✔
+- Must not start or end with separators (`_`, `-` or `.`):
+  - `_alice.` ✗
+  - `.bob.near-` ✗
+- Each part of the Account ID consists of lowercase alphanumeric symbols separated either by `_` or `-`, example:
+  - `1_4m_n0t-al1c3.near` ✔
+- Separators are not permitted to immediately follow each other, example:
+  - `alice..near` ✗
+  - `not-_alice.near` ✗
+
+Learn more here: <https://docs.near.org/docs/concepts/account#account-id-rules>
--- a/core/account-id/README.md
+++ b/core/account-id/README.md
@@ -24,7 +24,24 @@ assert!(
 
 assert!(
   matches!(
+    // no caps
+    "MelissaCarver.near".parse::<AccountId>(),
+    Err(err) if err.kind().is_invalid()
+  )
+);
+
+assert!(
+  matches!(
+    // separators cannot immediately follow each other
     "bob__carol".parse::<AccountId>(),
+    Err(err) if err.kind().is_invalid()
+  )
+);
+
+assert!(
+  matches!(
+    // each part must be alphanumeric only (ƒ is not f)
+    "ƒelicia.near".parse::<AccountId>(),
     Err(err) if err.kind().is_invalid()
   )
 );
@@ -34,17 +51,19 @@ assert!(
 
 - Minimum length is `2`
 - Maximum length is `64`
-- An Account ID consists of Account ID parts separated by `.`, example:
-  - `root` ✔
-  - `alice.near` ✔
-  - `bob.stage.testnet` ✔
+- An **Account ID** consists of **Account ID parts** separated by `.`, example:
+  - `root` ✓
+  - `alice.near` ✓
+  - `app.stage.testnet` ✓
 - Must not start or end with separators (`_`, `-` or `.`):
   - `_alice.` ✗
   - `.bob.near-` ✗
-- Each part of the Account ID consists of lowercase alphanumeric symbols separated either by `_` or `-`, example:
-  - `1_4m_n0t-al1c3.near` ✔
+- Each part of the **Account ID** consists of lowercase alphanumeric symbols separated either by `_` or `-`, example:
+  - `ƒelicia.near` ✗ (`ƒ` is not `f`)
+  - `1_4m_n0t-al1c3.near` ✓
 - Separators are not permitted to immediately follow each other, example:
   - `alice..near` ✗
   - `not-_alice.near` ✗
+- An **Account ID** that is 64 characters long and consists of lowercase hex characters is a specific **implicit account ID**
 
 Learn more here: <https://docs.near.org/docs/concepts/account#account-id-rules>
--- a/core/account-id/src/errors.rs
+++ b/core/account-id/src/errors.rs
@@ -33,17 +33,17 @@ pub enum ParseErrorKind {
 }
 
 impl ParseErrorKind {
-    /// Returns `true` if the Account Id was too long.
+    /// Returns `true` if the Account ID was too long.
     pub fn is_too_long(&self) -> bool {
         matches!(self, ParseErrorKind::TooLong)
     }
 
-    /// Returns `true` if the Account Id was too short.
+    /// Returns `true` if the Account ID was too short.
     pub fn is_too_short(&self) -> bool {
         matches!(self, ParseErrorKind::TooShort)
     }
 
-    /// Returns `true` if the Account Id was marked invalid.
+    /// Returns `true` if the Account ID was marked invalid.
     ///
     /// This can happen for the following reasons:
     ///
--- a/core/account-id/src/lib.rs
+++ b/core/account-id/src/lib.rs
@@ -2,7 +2,7 @@
 //!
 //! ## Usage
 //!
-//! ```rust
+//! ```
 //! use near_account_id::AccountId;
 //!
 //! let alice: AccountId = "alice.near".parse().unwrap();
@@ -31,6 +31,7 @@
 //!   - `root` ✔
 //!   - `alice.near` ✔
 //!   - `bob.stage.testnet` ✔
+//!   - `ƒelicia.near` ✗ (`ƒ` is not `f`)
 //! - Must not start or end with separators (`_`, `-` or `.`):
 //!   - `_alice.` ✗
 //!   - `.bob.near-` ✗
@@ -40,7 +41,7 @@
 //!   - `alice..near` ✗
 //!   - `not-_alice.near` ✗
 //!
-//! Learn more here: https://docs.near.org/docs/concepts/account#account-id-rules
+//! Learn more here: <https://docs.near.org/docs/concepts/account#account-id-rules>
 
 use std::{fmt, str::FromStr};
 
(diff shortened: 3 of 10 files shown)


mm-near (approved):
Wow -- super nice documentation, thank you VERY much.

core/account-id/src/lib.rs
```diff
@@ -64,49 +281,34 @@ impl AccountId {
...
-        }
-        // Will not panic, since valid account id is utf-8 only and the length is checked above.
-        // e.g. when `near` creates `aa.near`, it splits into `aa.` and `near`
-        let (prefix, suffix) = self.0.split_at(self.len() - parent_account_id.len());
-
-        prefix.find('.') == Some(prefix.len() - 1) && suffix == parent_account_id.as_ref()
-    }
-
-    /// Returns true if the account ID length is 64 characters and it's a hex representation.
-    pub fn is_implicit(account_id: &str) -> bool {
```
frol, line 104: I feel the point of having the API this way is to be able to validate non-owned strings whether they are implicit accounts, so changing the API to require `&self` might be counter-productive, but I don't have a strong opinion here.

frol (approved):
Lovely docs! :heart:

core/account-id/README.md
```diff
@@ -0,0 +1,69 @@
+# near-account-id
+
+This crate provides a type for representing a valid, unique account identifier on the [NEAR](https://near.org) network.
+
+[![crates.io](https://img.shields.io/crates/v/near-account-id?label=latest)](https://crates.io/crates/near-account-id)
+[![Documentation](https://docs.rs/near-account-id/badge.svg)](https://docs.rs/near-account-id)
+[![Version](https://img.shields.io/badge/rustc-1.56+-ab6000.svg)](https://blog.rust-lang.org/2021/10/21/Rust-1.56.0.html)
```
matklad, line 7: updating this feels like a chore. I sugest *not* having this and instead rely on `rust-version` in Cargo.toml

core/account-id/src/lib.rs
```diff
@@ -1,6 +1,69 @@
...
+//!   - `.bob.near-` ✗
+//! - Each part of the **Account ID** consists of lowercase alphanumeric symbols separated either by `_` or `-`, example:
+//!   - `ƒelicia.near` ✗ (`ƒ` is not `f`)
+//!   - `1_4m_n0t-al1c3.near` ✔
+//! - Separators are not permitted to immediately follow each other, example:
+//!   - `alice..near` ✗
+//!   - `not-_alice.near` ✗
+//! - An **Account ID** that is 64 characters long and consists of lowercase hex characters is a specific **implicit account ID**
+//!
+//! Learn more here: <https://docs.near.org/docs/concepts/account#account-id-rules>
```
matklad, line 22: Same here, I'd probably just add a link to the reference here, rather than duplicate the reference.

core/account-id/src/lib.rs
```diff
@@ -1,6 +1,69 @@
...
+//!   )
+//! );
+//!
+//! assert!(
+//!   matches!(
+//!     // each part must be alphanumeric only (ƒ is not f)
+//!     "ƒelicia.near".parse::<AccountId>(),
+//!     Err(err) if err.kind().is_invalid()
+//!   )
+//! );
```
matklad, line 61: I think showing every error makes this documentation *less* useful, as it takes more time to skim through the unrelated bits. In general, the meta rule is to write as little as possible, while conveying all the information you want. Few people have time to spare for reading long-winded docs :)

core/account-id/src/lib.rs
```diff
@@ -9,23 +72,177 @@ mod serde;
 
 #[cfg(feature = "deepsize_feature")]
 use deepsize::DeepSizeOf;
-pub use error::{ParseAccountError, ParseErrorKind};
+pub use errors::{ParseAccountError, ParseErrorKind};
 
+/// Smallest valid length for a NEAR Account ID.
 pub const MIN_ACCOUNT_ID_LEN: usize = 2;
```
matklad, line 78: I wonder if this should be

```
impl AccountId {
  pub const MIN_LEN: usize = 2;
}
```

core/account-id/src/lib.rs
miraclx: Yeah, well the thing is: If you already have an `AccountId` to begin with, then you only need to check `id.is_implicit()` and if you don't then you need to parse that string, validating it first, before checking if it's implicit.

core/account-id/src/lib.rs
```diff
@@ -9,23 +72,177 @@ mod serde;
...
+/// Largest valid length for a NEAR Account ID.
 pub const MAX_ACCOUNT_ID_LEN: usize = 64;
 
-/// Account identifier. Provides access to user's state.
+/// NEAR Account Identifier.
+///
+/// This is a unique, validated, human-readable account identifier on the NEAR network.
```
matklad, line 84:
```suggestion
/// This is a unique, syntactically valid, human-readable account identifier on the NEAR network.
```

core/account-id/src/lib.rs
```diff
@@ -9,23 +72,177 @@ mod serde;
...
+    /// let near_tla: AccountId = "near".parse().unwrap();
+    /// assert!(near_tla.is_top_level_account_id());
+    ///
+    /// // "alice.near" is a sub account of "near" account
+    /// let alice: AccountId = "alice.near".parse().unwrap();
+    /// assert!(!alice.is_top_level_account_id());
+    /// ```
+    pub fn is_top_level_account_id(&self) -> bool {
+        self.len() >= MIN_ACCOUNT_ID_LEN
+            && self.len() <= MAX_ACCOUNT_ID_LEN
```
matklad, line 165: Why do we need these two conditions here? They should always be `true`, no?

core/account-id/src/lib.rs
```diff
@@ -9,23 +72,177 @@ mod serde;
...
+    /// use near_account_id::AccountId;
+    ///
+    /// let near_tla: AccountId = "near".parse().unwrap();
+    /// assert!(near_tla.is_top_level_account_id());
+    ///
+    /// // "alice.near" is a sub account of "near" account
+    /// let alice: AccountId = "alice.near".parse().unwrap();
+    /// assert!(!alice.is_top_level_account_id());
+    /// ```
+    pub fn is_top_level_account_id(&self) -> bool {
```
matklad, line 163:
```suggestion
    pub fn is_top_level(&self) -> bool {
```

core/account-id/src/lib.rs
```diff
@@ -9,23 +72,177 @@ mod serde;
...
+    /// use near_account_id::AccountId;
+    ///
+    /// let near_tla: AccountId = "near".parse().unwrap();
+    /// assert!(near_tla.is_top_level_account_id());
+    ///
+    /// let alice: AccountId = "alice.near".parse().unwrap();
+    /// assert!(alice.is_sub_account_of(&near_tla));
+    ///
+    /// let alice_app: AccountId = "app.alice.near".parse().unwrap();
+    /// assert!(alice_app.is_sub_account_of(&alice));
```
matklad, line 186: I feel `alice_app.is_sub_account_of(&near_tla)` would be a useful example. It's not super-obvisous if this a transitive relation or not.

core/account-id/src/lib.rs
```diff
@@ -64,49 +281,34 @@ impl AccountId {
...
-    /// Returns true if the account ID length is 64 characters and it's a hex representation.
-    pub fn is_implicit(account_id: &str) -> bool {
-        account_id.len() == 64
-            && account_id.as_bytes().iter().all(|b| matches!(b, b'a'..=b'f' | b'0'..=b'9'))
-    }
-
-    /// Returns true if the account ID is the system account.
     pub fn is_system(&self) -> bool {
         self.as_ref() == "system"
     }
```
matklad, line 314: Bellow it seems better to delete `test_account` function. I don't think that `test` has any special meaning for the protocol, so there shouldn't be a random fn to create that specific naem.

matklad: Bellow, why it is 

```
impl<T: ?Sized> AsRef<T> for AccountId
where
    Box<str>: AsRef<T>,
```

and not just 

```
impl AsRef<str> for AccountId
```

?

matklad: Bellow, in the Display impl, it's bettre to use `fmt::Display::fmt` rather than `write!` to get support for padding for free: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=464ccb8bc9a815d9c0193e742af51908

matklad: Bellow, we have `impl From<AccountId> for Box<str>` but not the opposite `TryFrom<Box<str>>`.

matklad: In addition to `AsRef`, I'd probably add

```rust
impl AccountId {
  pub fn as_str(&self) -> &str
}
```

core/account-id/src/lib.rs
```diff
@@ -9,23 +72,177 @@ mod serde;
...
+    /// ## Examples
+    ///
+    /// ```
+    /// use near_account_id::AccountId;
+    ///
+    /// let carol: AccountId = "carol.near".parse().unwrap();
+    /// assert_eq!(10, carol.len());
+    /// ```
+    #[allow(clippy::len_without_is_empty)]
     pub fn len(&self) -> usize {
```
matklad, line 143: I am curious why we have this specific method, but not all other methods available on `&str`?

core/account-id/src/errors.rs
```diff
@@ -1,14 +1,16 @@
 use std::fmt;
 
 /// An error occurred when parsing an invalid Account ID with [`AccountId::validate`](crate::AccountId::validate).
```
matklad, line 3: This reads a bit awkward to me: this is data, not action, and `validate` seems overly specific. I'd just steal from the great https://doc.rust-lang.org/stable/std/num/struct.ParseIntError.html

```
An error which can be returned when parsing a NEAR account ID.
```

core/account-id/src/errors.rs
```diff
@@ -1,14 +1,16 @@
...
-#[derive(Eq, Hash, Clone, Debug, PartialEq)]
+#[derive(Eq, Clone, Debug, PartialEq)]
```
matklad, line 4: :+1:

core/account-id/src/errors.rs
```diff
@@ -1,14 +1,16 @@
...
 pub struct ParseAccountError(pub(crate) ParseErrorKind, pub(crate) String);
 
 impl ParseAccountError {
+    /// Returns the corresponding [`ParseErrorKind`] for this error.
```
matklad, line 8: Again, I'd still the wording from the ParseIntError docs

core/account-id/src/errors.rs
```diff
@@ -1,14 +1,16 @@
...
     pub fn kind(&self) -> &ParseErrorKind {
         &self.0
     }
 
+    /// Returns the corresponding [`AccountId`](crate::AccountId) for this error.
     pub fn get_account_id(self) -> String {
```
matklad, line 14: This method should just not exist, the errors should not store the data which is available to the caller anyway.

matklad: Bellow: 

```rust
write!(f, "[{}]: {}", self.1, self.0)
```

this is somewhat unusal formatting. I think that's fine probably, as we can always change Display details. Though my gut feeling would be again to try to mimic ParseIntError

core/account-id/src/errors.rs
```diff
@@ -22,22 +24,32 @@ impl fmt::Display for ParseAccountError {
...
 impl ParseErrorKind {
+    /// Returns `true` if the Account ID was too long.
     pub fn is_too_long(&self) -> bool {
         matches!(self, ParseErrorKind::TooLong)
     }
 
+    /// Returns `true` if the Account ID was too short.
     pub fn is_too_short(&self) -> bool {
         matches!(self, ParseErrorKind::TooShort)
     }
```
matklad, line 44: I don't think we need these two methods, the caller can `matches!` or `==` just fine.

core/account-id/src/errors.rs
```diff
@@ -22,22 +24,32 @@ impl fmt::Display for ParseAccountError {
...
 
+    /// Returns `true` if the Account ID was marked invalid.
+    ///
+    /// This can happen for the following reasons:
+    ///
+    /// - An invalid character was detected (could be an uppercase character, symbol or space).
+    /// - Separators immediately follow each other.
+    /// - Separators begin or end the Account ID.
     pub fn is_invalid(&self) -> bool {
```
matklad, line 53: ditto.

core/account-id/src/errors.rs
```diff
@@ -22,22 +24,32 @@ impl fmt::Display for ParseAccountError {
 }
 
 /// A list of errors that occur when parsing an invalid Account ID.
-#[derive(Eq, Hash, Clone, Debug, PartialEq)]
+#[non_exhaustive]
+#[derive(Eq, Clone, Debug, PartialEq)]
 pub enum ParseErrorKind {
     TooLong,
     TooShort,
     Invalid,
```
matklad, line 32: probably makes sense to add docs to variants here

matklad: Feels like we might want to split `Invalid` into more fine graind componets: InvalidChar, DuplicateSeparator, TrailingSeparator. doing that aftre release would technically be a breaking change.

matklad (commented):
Ended up just reviewing the whole thing, sorry about that :sweat_smile: 

Extra notes: 

in `serde.rs`, we probably should use `de::Error::invalid_value` rather than `::custom`.

core/account-id/src/errors.rs
miraclx: Already did this on a local branch 😄 , I went with the following names though - `HasConsecutiveSeparators`, `HasUnterminatedSeparators` and `HasInvalidChars`.

core/account-id/README.md
```diff
@@ -0,0 +1,69 @@
...
+  - `_alice.` ✗
+  - `.bob.near-` ✗
+- Each part of the **Account ID** consists of lowercase alphanumeric symbols separated either by `_` or `-`, example:
+  - `ƒelicia.near` ✗ (`ƒ` is not `f`)
+  - `1_4m_n0t-al1c3.near` ✓
+- Separators are not permitted to immediately follow each other, example:
+  - `alice..near` ✗
+  - `not-_alice.near` ✗
+- An **Account ID** that is 64 characters long and consists of lowercase hex characters is a specific **implicit account ID**
+
```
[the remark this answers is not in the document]
miraclx: I agree, and was what I did at least initially, but some rules in there, didn't suffice to convey some cases (unless you read the regex on the page or jump into the parsing logic here), like:

- Account IDs not being allowed to start or terminate in separators
- Separators not being allowed to immediately follow each other

core/account-id/README.md
miraclx: My thinking was this wasn't going to change very often, since it's unlikely this crate changes a lot once stabilized. I'll drop it nonetheless.

core/account-id/src/lib.rs
miraclx: Right on!

core/account-id/src/errors.rs
matklad: sgtm! 

Another option for `ConsecutiveSeparators` and `TrailingSeparator` is to go dual. The rule isn't "you can't have two separators, and a trailing separator, and a leading separator". The rule is rather "separators separate non-empty parts". So the actual error here is `EmptyPart`. 


:thinking: having looked at the impl again, I do think we shouldn't try to distinguish separator-related errors, as `..` is erroneous for three reasons. That is, I am pretty sure there should be only one error variant here. So, either `EmptyPart` if we focus on parts, or `InvalidSeparator` if we focus on separators.

core/account-id/README.md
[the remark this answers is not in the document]
matklad: >but some rules in there, didn't suffice to convey some cases

Which feels like a good reason to send PR to *that* doc ;)

core/account-id/src/lib.rs
miraclx: Implemented `Deref<str>`.

core/account-id/src/errors.rs
miraclx: Yeah, we initially didn't have this, at least not until testing a bunch of IDs in a for-loop was failing, and I needed to inspect which one caused it. But that's a specific use case, and we can drop that since, ideally, the caller should be able to inspect the IDs before parsing.

core/account-id/src/errors.rs
miraclx: Yeah, initially I thought it could save the extra `ParseErrorKind` import if you could check `err.kind().is_too_long()`. But going by how rarely it's going to be used, it does make sense to drop it.

core/account-id/src/errors.rs
miraclx: The error description for `InvalidSeparator` I feel would be lacking to some degree. `EmptyPart` sounds good to me.

miraclx: So, `EmptyPart` is good for all error cases in `..` but not so much for `__` because according to the spec:

> **Account ID part** consists of lowercase alphanumeric symbols separated by either `_` or `-`. [ <sub> (doc) </sub> ](https://docs.near.org/docs/concepts/account#account-id-rules)

So, `_` and `-`, don't delimit a part. They're just a sort-of “spacer” within a part. Meaning, while `..` has 3 empty parts, `__` has one _non-empty, though invalid_ part.

Back to naming: maybe `ExtraneousSeparator`, `ExcessiveSeparator` or `RedundantSeparator`? 😪

I'm voting `RedundantSeparator`.

miraclx: Implemented in https://github.com/near/nearcore/pull/5660

PR #5577, opened 2021-12-02
feat: Remove `Copy` trait from `Cryptohash`

pmnoxx:
Rust is built around the concept of “zero cost abstraction”. I noticed that in our code base we do a lot of hidden copies, when it comes to `Cryptohash`.

`Cryptohash` currently uses `Copy` derive, which hides all copies of that structs. We could have avoided doing those copy operations, and also improved performance of our code.

My proposal is to remove `Copy` trait from `Cryptohash`.

We can continue doing that, until all crates no longer rely on `Copy` trait, and then simply remove it.

Part 1:
- No signature changes, no optimizations, this is just made so we can use `CryptoHash` without `Copy`

Part 2:
- Optimizations. That's optional, it will be up to individual code owner to decide to do them or not. This can be done later, or just after pt1.

### DIFF
--- a/chain/chain/src/chain.rs
+++ b/chain/chain/src/chain.rs
@@ -110,7 +110,7 @@ pub struct Orphan {
 
 impl BlockLike for Orphan {
     fn hash(&self) -> CryptoHash {
-        *self.block.hash()
+        self.block.hash().clone()
     }
 
     fn height(&self) -> u64 {
@@ -153,11 +153,13 @@ impl OrphanBlockPool {
     fn add(&mut self, orphan: Orphan) {
         let height_hashes =
             self.height_idx.entry(orphan.block.header().height()).or_insert_with(|| vec![]);
-        height_hashes.push(*orphan.block.hash());
-        let prev_hash_entries =
-            self.prev_hash_idx.entry(*orphan.block.header().prev_hash()).or_insert_with(|| vec![]);
-        prev_hash_entries.push(*orphan.block.hash());
-        self.orphans.insert(*orphan.block.hash(), orphan);
+        height_hashes.push(orphan.block.hash().clone());
+        let prev_hash_entries = self
+            .prev_hash_idx
+            .entry(orphan.block.header().prev_hash().clone())
+            .or_insert_with(|| vec![]);
+        prev_hash_entries.push(orphan.block.hash().clone());
+        self.orphans.insert(orphan.block.hash().clone(), orphan);
 
         if self.orphans.len() > MAX_ORPHAN_SIZE {
             let old_len = self.orphans.len();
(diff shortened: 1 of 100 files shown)


pmnoxx:
@matklad Do you think we should do it?

nikurt (commented):
Why do you want to do this change in stages? Isn't this PR self-contained and not requiring a follow-up?

matklad:
I am torn on this one. The structure is not small, but also not that large. It's like two wide pointers: 

```rust
#[test]
fn test_sizes() {
    assert_eq!(std::mem::size_of::<CryptoHash>(), 32);
    assert_eq!(std::mem::size_of::<&dyn Fn()>(), 16);
}
```

But Rust is known for producing some extra memcpys when copying large structs, and it feels like `fn foo(hash: &CryptoHash)` should be better than `fn foo(hash: CryptoHash)`. I don't have relevant experience to tell how meaningful the difference would be. I am curious if @nagisa knows better? 


My current thinking is to leave this as is: `CryptoHash: Copy` is OK. `CryptoHash: !Copy` might be better, but probably not by a lot. It also feels that this might not be the lowest hanging fruit? At the same time, actually changing this requires a lot of effort, as can be seen by the length of the reviewers list :)

nagisa:
> I don't have relevant experience to tell how meaningful the difference would be. I am curious if @nagisa knows better?

4 usizes will typically still fit in the registers allocated for calling convention use, which e.g. for the purposes of function that takes one or two arguments is still going to be close to optimal. I wouldn't worry particularly much about copies of 4-word structs unless profiles can show meaningful improvement by moving to references or similar alternative approach.

`Clone`is probably not a great solution overall, I feel. It increases the pressure on the inliner (longer compiles), and once the `derive(Copy)` is removed, `Clone` implementation will lose the memcpy optimization, hurting the performance further. Given the number of changes needed to add `clone`s everywhere I'm also wondering if code readability benefits to `Copy` might be a deciding factor here.

bowenwang1996:
So it appears that we should not do this change and I totally agree with @nagisa's point on readability!

pmnoxx:
I saw a few places, where we do the extra copy. Though, I see that we should do first more benchmark to see whenever it's worth to do it or not.

It's a trade of between performance, and reliability, I didn't expect it to be used in so many places.

pmnoxx:
@matklad @mm-near @bowenwang1996 
There are some performance benefits of using references, even for such small structs.
I bench-marked two functions, one with references, another with copys
```
pub fn combine_hash1(hash1: MerkleHash, hash2: MerkleHash) -> MerkleHash {
    CryptoHash::hash_borsh(&(hash1, hash2))
}
pub fn combine_hash2(hash1: &MerkleHash, hash2: &MerkleHash) -> MerkleHash {
    CryptoHash::hash_borsh(&(hash1, hash2))
}
```
The results:
```
test combine_hash1_bench  ... bench:         123 ns/iter (+/- 1)
test combine_hash2_bench  ... bench:         102 ns/iter (+/- 1)
```

Issue #3186, opened 2020-08-17
Add Rosetta tests to CI


nearmax:
We need to run Rosetta test suite as part of our CI to be able to know when something gets broken.

frol:
Plan
====

* [ ] Configure the test suite:
  a. Run `rosetta-cli check:construction` (configure `rosetta.cfg` with some scenarios)
  b. Run `rosetta-cli check:data` in parallel with `rosetta-cli check:construction`
* [ ] Add compilation of nearcore with the `rosetta_rpc` feature flag enabled and include the test suite execution to our CI setup

bowenwang1996:
@frol what is the status of this issue?

frol:
This issue is still relevant, but not urgent.

bowenwang1996:
@mina86 I think it is probably a good first task for you in terms of Rosetta

mina86:
I’m still going to be adding pytest tests but I think this can be closed.

frol:
@mina86 Well, actually, we needed to integrate the official Rosetta compliance checks: https://github.com/coinbase/rosetta-cli#checkdata (check:data and check:construction): https://github.com/near/nearcore/issues/3186#issuecomment-675124554

wacban:
please reopen if still relevant

PR #5578, opened 2021-12-02
pytest: kill node process through Popen class and then wait for it

mina86:
This fixes the following warning:

    ../usr/lib/python3.9/subprocess.py:1052: ResourceWarning: subprocess 91224 is still running
      _warn("subprocess %s is still running" % self.pid,

which is caused by LocalNode not waiting for the thread to terminate
after it’s killed.

Issue: https://github.com/near/nearcore/issues/3186
Issue: https://github.com/near/nearcore/issues/4618


mina86:
Note: This is on top of https://github.com/near/nearcore/pull/5576 so look at the second commit only.

pytest/lib/cluster.py
```diff
@@ -379,17 +381,15 @@ def start(self, *, boot_node: BootNode = None, skip_starting_proxy=False):
         env["RUST_LOG"] = "actix_web=warn,mio=warn,tokio_util=warn,actix_server=warn,actix_http=warn," + env.get(
             "RUST_LOG", "debug")
 
-        self.stdout_name = os.path.join(self.node_dir, 'stdout')
-        self.stderr_name = os.path.join(self.node_dir, 'stderr')
-        self.stdout = open(self.stdout_name, 'a')
-        self.stderr = open(self.stderr_name, 'a')
+        self._stdout = open(os.path.join(self.node_dir, 'stdout'), 'a')
```
mm-near, line 384: 'a' or 'ab'?

pytest/lib/cluster.py
```diff
@@ -398,24 +398,25 @@ def start(self, *, boot_node: BootNode = None, skip_starting_proxy=False):
         except:
             logger.error(
                 '=== failed to start node, rpc does not ready in 10 seconds')
-            self.stdout.close()
-            self.stderr.close()
             if os.environ.get('BUILDKITE'):
```
mm-near, line 401: please comment what is the special thing that we need to do for 'BUILDKITE'

pytest/lib/cluster.py
```diff
@@ -398,24 +398,25 @@ def start(self, *, boot_node: BootNode = None, skip_starting_proxy=False):
...
-                pass  # the process has already terminated
-            self.pid.value = 0
-
-            if self._proxy_local_stopped is not None:
-                self._proxy_local_stopped.value = 1
+        if self._proxy_local_stopped is not None:
+            self._proxy_local_stopped.value = 1
+        if self._process:
+            self._process.kill()
+            self._process.wait(5)
```
mm-near, line 418: this will add quite a delay to all the tests (and might introduce some flakiness). maybe we could keep checking that pid is free?

mm-near (changes requested):

pytest/lib/cluster.py
mina86: In the end it doesn’t matter. The file is passed to the subprocess.

pytest/lib/cluster.py
mina86: I don’t see why it would. The process is killed with SIGKILL so in practice this will never wait.

pytest/lib/cluster.py
mina86: With the other PR merged this PR no longer touches this piece of code so documenting this here feels out of place for this PR.

mina86:
@mm-near, PTAL

pmnoxx (approved):
LGTM

mm-near (approved):

PR #5581, opened 2021-12-02
chore: cleanup near-account-id API

miraclx:
Cleans up the `near-account-id` API in preparedness for stabilization.

Based on top of https://github.com/near/nearcore/pull/5575.

- Drops `AccountId::system_account()` and `AccountId::test_account()`.
- Introduces `AccountId::as_str()`
- Rewords `AccountId::is_top_level_account_id()` to `AccountId::is_top_level()`.
- Implements `TryFrom<Box<str>>` for `AccountId`.
- Implements `Deref<str>` for `AccountId` so `&*account_id`, `account_id.len()`, `account_id.as_bytes()` all work.
- Drops `ParseAccountError::get_account_id()`.
- Documents `AccountId::is_system()`.
- Rewords documentation.
- Licensed to `Apache-2.0 OR MIT`.
- Drops helper methods from `ParseErrorKind`.
- Moves crate-level length consts to be associated with the `AccountId` struct itself.
- Use custom error reporting for borsh + serde:
  - `borsh`: `invalid value: "Mantra", the account ID has an invalid format`
  - `serde`: `invalid value: "Mantra", the account ID has an invalid format at line 3 column 1`


khorolets (approved):

runtime/runtime/src/lib.rs
```diff
@@ -313,7 +313,7 @@ impl Runtime {
         result.gas_used += exec_fees;
         let account_id = &receipt.receiver_id;
         let is_the_only_action = actions.len() == 1;
-        let is_refund = AccountId::is_system(&receipt.predecessor_id);
+        let is_refund = receipt.predecessor_id.as_ref() == "system";
```
mm-near, line 316: what was the reason for removing 'is_system' (can you mention it in PR decription)

core/account-id/src/lib.rs
```diff
@@ -297,28 +297,38 @@ impl AccountId {
...
+}
 
-    pub fn is_system(&self) -> bool {
-        self.as_ref() == "system"
+#[doc(hidden)]
+#[allow(deprecated)]
+#[cfg(feature = "static_accounts")]
+impl AccountId {
+    pub fn near_account() -> Self {
+        Self::new_unvalidated("near")
```
mm-near, line 319: I wonder why do we use the 'invalidated' flow, rather than doing the 'normal' one ("near".parse()..)

as all these 4 names should pass the validation, right?

mm-near (changes requested):

core/account-id/src/lib.rs
miraclx: Yeah, they will. But since we know these names to be valid, we can skip the unnecessary checks.

runtime/runtime/src/lib.rs
miraclx: Done.

core/chain-configs/src/genesis_validate.rs
```diff
@@ -152,12 +152,12 @@ mod test {
     fn test_total_supply_not_match() {
         let mut genesis = Genesis::default();
         genesis.config.validators = vec![AccountInfo {
-            account_id: "test".parse().unwrap(),
+            account_id: AccountId::test_account(),
```
matklad, line 155: tbh, this change feels like a regression to me:

* more chars to type
* now the actual name of the account is obscured
* AccountId now concernts itself with the stuff which is only relevant to tests in unrelated crates.

runtime/runtime/src/lib.rs
matklad: Hm, `system` account is a protocol-level concept, so it actually seems fine to have one: https://nomicon.io/DataStructures/Account.html?highlight=system#system-account. 

And it is possible to observe system account id from the outside:

https://explorer.testnet.near.org/transactions/BBqGuajZrRBPpiQZMwJyJj1NNR3Q5P9KUD1w7HuCiz7C#HRKLfrnczeCsegJb5gVJzBhiMs7LXLe1Soq8RbqNvVLN

image

So I think `AccountId` should include public `is_system` method for completness, even if it is not very practically useful outside.

matklad (commented):
Heh, I feel like for every change here, I actually slightly but meaningfully prefer the old variant :-) See inline commets for details. 

No strong opinion here though!

runtime/runtime/src/lib.rs
miraclx: Understood, I'll restore it.

core/chain-configs/src/genesis_validate.rs
miraclx: Valid points, I'll go ahead to drop all the static_account methods all-together.

chefsale (approved):

integration-tests/tests/standard_cases/main.rs
```diff
@@ -1078,7 +1077,7 @@ pub fn test_access_key_reject_non_function_call(node: impl Node) {
         }),
     };
     let mut node_user = node.user();
-    let signer2 = InMemorySigner::from_random(AccountId::test_account(), KeyType::ED25519);
+    let signer2 = InMemorySigner::from_random("test".parse().unwrap(), KeyType::ED25519);
```
matklad, line 1080: unrelated to the PR, but it's not great that we have this many copies of the logic in the tests.

matklad (approved):

miraclx:
Applied patches from @matklad's review https://github.com/near/nearcore/pull/5575#pullrequestreview-821436213

core/account-id/src/lib.rs
```diff
@@ -154,17 +93,14 @@ impl AccountId {
...
-    /// assert!(!alice.is_top_level_account_id());
+    /// assert!(!alice.is_top_level());
     /// ```
-    pub fn is_top_level_account_id(&self) -> bool {
-        self.len() >= MIN_ACCOUNT_ID_LEN
-            && self.len() <= MAX_ACCOUNT_ID_LEN
-            && self.as_ref() != "system"
-            && !self.as_ref().contains('.')
+    pub fn is_top_level(&self) -> bool {
+        !self.is_system() && !self.contains('.')
```
matklad, line 103: This `self.contains` give me a pause with respect to `Deref` impl. 

Right now, the following code compiles: 

```rust
#[test]
fn hmm() {
    let near_tla: AccountId = "near".parse().unwrap();
    assert!(near_tla.is_top_level());

    let alice_app: AccountId = "app.alice.near".parse().unwrap();

    let hmm = alice_app.contains(&*near_tla);
}
```

Which I guess isn't *wrong*, but definitely looks a bit weird. Maybe stay conservative and don't do `Deref` immediatelly?

core/account-id/src/lib.rs
```diff
@@ -177,23 +113,21 @@ impl AccountId {
...
-            return false;
-        }
-        // Will not panic, since valid account id is utf-8 only and the length is checked above.
-        // e.g. when `near` creates `aa.near`, it splits into `aa.` and `near`
-        let (prefix, suffix) = self.0.split_at(self.len() - parent_account.len());
-
-        prefix.find('.') == Some(prefix.len() - 1) && suffix == parent_account.as_ref()
+    pub fn is_sub_account_of(&self, parent: &AccountId) -> bool {
+        self.strip_suffix(&parent[..])
+            .map_or(false, |s| !s.is_empty() && s.find('.') == Some(s.len() - 1))
```
matklad, line 130:
```suggestion
        self.strip_suffix(parent.as_str())
            .map_or(false, |s| s.ends_with('.'))
```

core/account-id/src/lib.rs
```diff
@@ -214,8 +148,26 @@ impl AccountId {
...
+    /// use near_account_id::AccountId;
+    ///
+    /// let alice: AccountId = "alice.near".parse().unwrap();
+    /// assert!(!alice.is_system());
+    ///
+    /// let system: AccountId = "system".parse().unwrap();
+    /// assert!(system.is_system());
+    /// ```
+    pub fn is_system(&self) -> bool {
+        &self[..] == "system"
```
matklad, line 170:
```suggestion
        self.as_str() == "system"
```

core/account-id/src/lib.rs
```diff
@@ -298,42 +238,35 @@ impl AccountId {
...
+}
 
-    pub fn is_system(&self) -> bool {
-        self.as_ref() == "system"
-    }
-
-    pub fn system_account() -> Self {
-        "system".parse().unwrap()
-    }
+impl std::ops::Deref for AccountId {
```
matklad, line 253: optional: i like to `impl ops::Trait for X` for all `ops` trait. That's usually the neatest way, as sometimes you impl more than one trait from there.

chefsale (approved):

core/account-id/src/lib.rs
miraclx: Well, this actually isn't what we want. `app.alice.near` is not a sub-account of `near`.

core/account-id/src/lib.rs
miraclx: I feel this is fine.. The coercion of `&AccountId` to `&str` is nice to have. We can pretty much treat an `AccountId` as a smart pointer to a memory location guaranteed to hold a valid NEAR Account ID so, the `&*near_tla` feels logical to me, although I think most people would probably just use `.as_str()` in that case.

core/account-id/src/lib.rs
miraclx: Yeah, only, in this case, with one use, I'm not convinced that it is as useful.

nikurt (approved):

matklad (approved):

core/account-id/README.md
```diff
@@ -1,69 +1,36 @@
...
-```
+Licensed under either of
+
+- Apache License, Version 2.0
+   ([LICENSE-APACHE](LICENSE-APACHE) or <http://www.apache.org/licenses/LICENSE-2.0>)
+- MIT license
+   ([LICENSE-MIT](LICENSE-MIT) or <http://opensource.org/licenses/MIT>)
 
-## Account ID Rules
+at your option.
```
mm-near, line 30: I'm surprised that AccountId rules disappeared from this documentation

mm-near: discussed offline: please add the link to the account rules: https://docs.near.org/docs/concepts/account#account-id-rules

mm-near (approved):

Issue #5173, opened 2021-11-08
Being able to use rosetta to determine transaction completion


asmolarek-bc:
This issue is related to the scope of data provided by Rosetta API 

**Describe the bug**
Rosetta does not have a way to determine which part of operation in a block is connected to which transaction because initial operation is connected to transaction id, and subsequent ones are identified by they specific receipt id
Because of that it is not possible to connect operations from single transaction together without using near RPC (which allows querying transaction with all of its receipts)

**To Reproduce**
Reconstruct transaction operations from data provided in rosetta blocks, try to determine transaction ~finality~ completeness only using rosetta API

**Expected behavior**
I am able to connect all operations reported in Rosetta blocks to original transaction, I am able to determine transaction ~finality~ completeness base only on data provided by Rosetta API

Version (please complete the following information):

nearcore tag 1.21.0
rustc 1.51.0
rustup 1.24.3 (ce5817a94 2021-05-31)
cargo 1.51.0
docker Docker version 20.10.8, build 3967b7d
local env with local network

node was compiled with --features rosetta_rpc and started with rosetta api enabled in conf file

bowenwang1996:
Do you mean you want one endpoint that returns this? It is still possible to reconstruct this information I believe

asmolarek-bc:
No I do not need single endpoint, I can reconstruct it from multiple calls, but what I found it is not possible via current implementation of rosetta I do not have a way to ask for transaction base on receipt id (those are the identifiers of subsequent operations), in general (from what I asked) it is not possible to get transaction by one of its receipts id

afaik rosetta api does not allow asking for specific transaction by transaction id
What is supported is asking for transaction by transaction id and block 
https://www.rosetta-api.org/docs/BlockApi.html#blocktransaction
But when I am receiving subsequent operations I do not have transaction id ad previous block, I only have receipt id and current block
And this endpoint returns transaction only in that block, which means it will not return operations related to receipts in subsequent blocks (am I right?)

for now workaround is to keep in memory transactions that are not yet final (there are pending outcomes) and ask near RPC to get transactions + all receipts, from this I have information about every transfer / refund and amount of tokens burnt (fee)

frol:
@asmolarek-bc Yep, there is indeed a gap in Rosetta implementation there (there was no notion of `related_transactions` before the very latest Rosetta 1.4.10 spec while we only implemented Rosetta 1.4.4).

[...]

PR #5582, opened 2021-12-02
pytest: add /block Rosetta RPC tests 

mina86:
Refactor Rosetta RPC test and most importantly add a test for /block
and /block/transaction endpoints.

Issue: https://github.com/near/nearcore/issues/3186


pytest/tests/sanity/rosetta.py
```diff
@@ -0,0 +1,305 @@
...
+import sys
+import time
+import typing
+import unittest
+
+import base58
+import ed25519
+import requests
+
+sys.path.append(str(pathlib.Path(__file__).resolve().parents[2] / 'lib'))
```
bowenwang1996, line 16: what is this magic

mina86: It figures out path to the `pytest/lib` directory and adds it to Python’s search path so modules from there can be imported.

`__file__` is path to the current file. `pathlib.Path(__file__).resolve()` makes in an absolute path. `parents[2]` returns it’s great-grandparent; that gives us path to `pytest` directory. `/ 'lib'` joins that with `lib` which is the directory we’re after.

bowenwang1996: I see....

frol (approved):

PR #5584, opened 2021-12-02
re: `near-network`: Limit visibility of `PeerStore` to `pub(crate)`

pmnoxx:
We can limit visibility of `PeerStore` methods in `pub(crate)`. They are not used anywhere outside the crate.


chain/network/src/peer_manager/peer_store.rs
```diff
@@ -17,7 +17,7 @@ use tracing::{debug, error};
 
 /// Level of trust we have about a new (PeerId, Addr) pair.
 #[derive(Eq, PartialEq, Debug, Clone)]
-pub enum TrustLevel {
+pub(crate) enum TrustLevel {
```
mm-near, line 20: btw- for such changes: how are you checking that no other near component (especially ones outside nearcore) is using them?

mm-near (approved):

chain/network/src/peer_manager/peer_store.rs
pmnoxx: @mm-near Rust prevents leaking private types inside public structs or public function interfaces. This type is only used inside private structs, so that's not an issue. 

Also. this type is not exported outside, because `peer_store.rs` by itself is not exported outside. Currently, there is no way to use it outside. 

Even if someone, someone got this type, there is no way they could actually use it. Besides, our api can change at any times, we don't make any gurantees.

pmnoxx: @mm-near At some point, we will have to define what belongs to our public api and what not.

By default, it's best to limit number of types/functions exported public-ally. Private functions, can change at any time, due to refactoring. By having a clearly defined api, it's easier to reason, what the crate does, even without having to look into the code.

PR #5586, opened 2021-12-02
re: `Cargo.toml`: limit the number of features used by tokio

pmnoxx:
We don't need all `tokio` features, only
```
tokio = { version = "1.1", features = ["net", "rt-multi-thread"] }
```
are needed.


chefsale (approved):

frol (approved):

matklad (approved):
LGTM!

Although, there's a caveat here: if you enable `full` everywhere, you'll compile tokio once. If you set specific subset of features for each crate, then you might compile tokio several times, depending on which crates you pass as `-p` to `cargo build -p foo -p bar`. On ballance, I think it's better to stick to minimal features right now.

bowenwang1996 (approved):

pmnoxx:
> Although, there's a caveat here: if you enable `full` everywhere, you'll compile tokio once. If you set specific subset of features 
for each crate, then you might compile tokio several times, depending on which crates you pass as `-p` to `cargo build -p foo -p bar`. On ballance, I think it's better to stick to minimal features right now.

@matklad That's a good point. It's more a choice between drawbacks.

We have a few options:
A) Each crate uses minimum number of features for `tokio`.
B) Use the same set, everywhere `["net", "rt-multi-thread", "sync", "macros"]`  - sum of all feature needed
C) Current, using ["full"] in one place, and sometimes, just part of features.

There are 4 use-caces
1)- compiling everything
2)- compiling a single crate
3)- compile multiple crates individually 
4)  number of imports when imported externally

Option A)
1) BEST the compile time is minimizes, we choose minimum subset of features
2) BEST - minimum features used
3)  BAD takes longer time, because tokio needs to be compiled multiple times.
4) BEST 

Option B)
1) BEST - same as previously
2)  BAD - compiling single packages takes slightly more time
3) BEST - avoids recompiling tokio
4) BAD - we may import more than needed


Option C)
1) BAD - Takes more time than in A, B
2) BAD - Takes more time than in A, B
3) BAD - Takes more time than B, but still less tan A
4) BAD - imports needed than A, B

This PR uses option A), it seems like a best choice, that minimizes number of packages used. Though, going for `B`  can be seen as reasonable choice.

mina86 (approved):

PR #5587, opened 2021-12-02
refactor: reduce sizes of the macros in imports.rs

matklad:
This cleans up and documents `import.rs`. The crux of the change is that now, instead of a single giant macro which does everything, we have three separate modules, one for each runtime we use, which contain logic related to that specific runtime. Moreover, most of the said logic now is written outside of the macro, so completion, formatting and the like work. Sadly, this requires a somewhat esoteric implementation approach with high-order macros :( I think, on ballance, it's well-worth it -- I personally had a really hard time understanding which parts of the old implementation related to which supported runtimes.


nagisa (approved):
Seems significantly more readable to me.

runtime/near-vm-runner/src/imports.rs
```diff
@@ -1,230 +1,73 @@
...
+//! ```ignore
+//! for host_fn in all_host_functions {
+//!    wasm_imports.define("env", host_fn.name, |args| host_fn(args))
+//! }
+//! ```
+//!
+//! The actual implementation is a bit more complicated, for two reasons. First,
+//! host functions have different signatures, so there isn't a trivial single
+//! type one can use to hold a host function. Second, we want to use direct
+//! calls in the jitted wasm, so we need to avoid dynamic dispatch and hand
```
olonho, line 21: jitted wasm -> "compiled WASM"?

runtime/near-vm-runner/src/imports.rs
```diff
@@ -1,230 +1,73 @@
...
+//! for host_fn in all_host_functions {
+//!    wasm_imports.define("env", host_fn.name, |args| host_fn(args))
+//! }
+//! ```
+//!
+//! The actual implementation is a bit more complicated, for two reasons. First,
+//! host functions have different signatures, so there isn't a trivial single
+//! type one can use to hold a host function. Second, we want to use direct
+//! calls in the compiled WASM, so we need to avoid dynamic dispatch and hand
+//! functions as ZSTs to the WASM runtimes. This basically means that we need to
```
olonho, line 22: ZST is not too common acronym, I'd use explicit `zero sized types`.

olonho (approved):
Overall looks good, but as always with macroses review may have missed something. Did you take a look on code withe expanded macroses? How did it change from previous version?

matklad:
> Did you take a look on code withe expanded macroses? How did it change from previous version?

Yeah, the expanded code is functionally the same. Syntactically, we no longer create a separate module for extension functions, and instead define them at the point wheer we use them: 

```rust
// before 

mod wasmer2_ext {
    #[cfg(...)]
    fn gas(...)
}

fn build_wamer2() {
    ...
    if #[cfg(...)] {
        namespace.insert("ecrecover",wasmer::Function::new_native_with_env(&store,env.clone(), wasmer2_ext::gas));
    }
    ...
}

// after
fn build_wamer2() {
    ...
    if #[cfg(...)] {
        fn gas(...)
        namespace.insert("ecrecover",wasmer::Function::new_native_with_env(&store,env.clone(), gas));
    }
    ...
}

```

Rust's `fn` items are not closures, so there's no need to put them into a separate module.

PR #5589, opened 2021-12-02
feat(dump-state): Support two modes of state dumping: InMemory and Streaming

nikurt:
InMemory mode works best for betanet and makes automation easy.
Streaming mode works best for testnet and mainnet.


test-utils/state-viewer/src/state_dump.rs
```diff
@@ -42,66 +42,99 @@ pub fn state_dump(
...
 
+fn get_records<'a, F, X>(
+    runtime: NightshadeRuntime,
+    state_roots: Vec<StateRoot>,
+    last_block_header: BlockHeader,
+    validators: &'a HashMap<AccountId, (X, Balance)>,
+    mut callback: F,
+) -> Balance
+where
+    F: FnMut(StateRecord),
```
matklad, line 113: It feels a bit weird that we need this to be generic. As a rule of thumb, I try to avoid generics in Rust, to avoid code bloat and longer compile times. Will try to take a closer look tomorrow whether this can be simplifield.

test-utils/state-viewer/src/commands.rs
```diff
@@ -50,8 +48,22 @@ pub(crate) fn state(home_dir: &Path, near_config: NearConfig, store: Arc<Store>)
     }
 }
 
+#[derive(ArgEnum, Debug)]
+pub enum DumpStateMode {
+    /// Produces a single `<near_home_dir>/output.json` which contains all of the state, but needs
```
pmnoxx, line 53: I think we should add an extra argument, and let the user specify the output file. Not everyone wants to output the dump result directly to their home directory.

test-utils/state-viewer/src/commands.rs
```diff
@@ -64,14 +76,34 @@ pub(crate) fn dump_state(
...
-        state_dump(runtime, state_roots.clone(), header, &near_config, &records_path);
-
-    println!("Saving state at {:?} @ {} into {}", state_roots, height, output_dir.display(),);
-    new_near_config.save_to_dir(&output_dir);
+    match dump_mode {
+        DumpStateMode::InMemory => {
+            let new_near_config =
+                state_dump(runtime, state_roots.clone(), header, &near_config, None);
+            let output_file = home_dir.join("output.json");
+            println!(
```
pmnoxx, line 85: It may be a good idea to use tracing here:
```
use tracing::info;

info!( message =  "Saving state at,
                ?state_roots,
                height,
                output_file.display(),
            );
```
This will print a nicely formatted output.

test-utils/state-viewer/Cargo.toml
```diff
@@ -9,6 +9,7 @@ edition = "2021"
 ansi_term = "0.12"
 borsh = "0.9"
 clap = "=3.0.0-beta.2"
+itertools = "0.10.0"
```
pmnoxx, line 12: I agree with @matklad it's good to try to minimize number of imports. Especially, if they are just used once, and can be replaced with a slightly different syntax.

test-utils/state-viewer/src/cli.rs
```diff
@@ -1,16 +1,13 @@
...
 use near_logger_utils::init_integration_logger;
 use near_primitives::types::{BlockHeight, ShardId};
 use near_primitives::version::{DB_VERSION, PROTOCOL_VERSION};
 use near_store::{create_store, Store};
 use nearcore::{get_default_home, get_store_path, load_config, NearConfig};
-
-use crate::commands::*;
+use once_cell::sync::Lazy;
+use std::path::{Path, PathBuf};
+use std::sync::Arc;
```
pmnoxx, line 10: Thanks, for fixing the input order!

test-utils/state-viewer/src/commands.rs
pmnoxx:
```
@@ -82,11 +84,11 @@ pub(crate) fn dump_state(
             let new_near_config =
                 state_dump(runtime, state_roots.clone(), header, &near_config, None);
             let output_file = home_dir.join("output.json");
-            println!(
-                "Saving state at {:?} @ {} into {}",
-                state_roots,
+            info!(
+                message = "Saving state",
+                ?state_roots,
                 height,
-                output_file.display(),
+                output_file = ?output_file.display()
             );
             new_near_config.genesis.to_file(&output_file);
         }
```

test-utils/state-viewer/src/cli.rs
```diff
@@ -104,13 +101,18 @@ impl StateViewerSubCommand {
 
 #[derive(Clap)]
 pub struct DumpStateCmd {
+    /// Optionally, can specify at which height to dump state.
     #[clap(long)]
     height: Option<BlockHeight>,
+    /// Use this mode if you want to dump state of networks like testnet and mainnet, this mode has
+    /// a much lower RAM requirements, but produces multiple files as the result.
+    #[clap(arg_enum, long, default_value = "in-memory")]
+    mode: DumpStateMode,
```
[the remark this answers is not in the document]
pmnoxx: I would two mutually exclusive options
```
--stream
--single-dump
```
and maybe optional
```
--file FILE
```

test-utils/state-viewer/src/state_dump.rs
```diff
@@ -1,25 +1,25 @@
...
 use near_store::TrieIterator;
 use nearcore::config::NearConfig;
 use nearcore::NightshadeRuntime;
+use serde::ser::{SerializeSeq, Serializer};
+use std::collections::HashMap;
+use std::fs;
+use std::fs::File;
+use std::path::Path;
 
 pub fn state_dump(
```
pmnoxx, line 17: It may we worth changing signature to
```
--- i/test-utils/state-viewer/src/state_dump.rs
+++ w/test-utils/state-viewer/src/state_dump.rs
@@ -16,7 +16,7 @@ use std::path::Path;
 
 pub fn state_dump(
     runtime: NightshadeRuntime,
-    state_roots: Vec<StateRoot>,
+    state_roots: &[StateRoot],
     last_block_header: BlockHeader,
     near_config: &NearConfig,
     records_path: Option<&Path>,
```
That way we can remove clone of `state_roots`

test-utils/state-viewer/src/cli.rs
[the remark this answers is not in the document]
nikurt: Thanks, implemented @pmnoxx's suggestion.

test-utils/state-viewer/src/state_dump.rs
```diff
@@ -42,66 +42,99 @@ pub fn state_dump(
...
-    seq.end().unwrap();
-
     let mut near_config = near_config.clone();
 
     let mut genesis_config = near_config.genesis.config.clone();
     genesis_config.genesis_height = genesis_height;
     genesis_config.validators = validators
+        .clone()
         .into_iter()
+        .sorted()
```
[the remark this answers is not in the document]
nikurt: I pulled in `itertools` to sort by `AccountId`.
There are two options:
1. Implement `Ord` for `AccountInfo` and sort as you suggest.
2. Convert a `HashMap` into a `Vec`, sort it, map to `AccountInfo`.

Without `itertools`, option 1 seems simpler.

test-utils/state-viewer/src/commands.rs
nikurt: Good idea, thanks.

test-utils/state-viewer/src/commands.rs
nikurt: The use of `tracing` was discussed in one of the previous PRs: https://github.com/near/nearcore/pull/4867#discussion_r717534200

test-utils/state-viewer/src/cli.rs
```diff
@@ -104,13 +101,35 @@ impl StateViewerSubCommand {
...
     height: Option<BlockHeight>,
+    /// Dumps state records and genesis config into separate files.
+    /// Has reasonable RAM requirements.
+    /// Use for chains with large state, such as mainnet and testnet.
+    #[clap(long)]
+    stream: bool,
+    /// Produces a single json file containing the state.
+    /// Needs the whole state to fit in RAM.
+    /// Use for chains with small state, such as betanet, guildnet, localnet.
+    #[clap(long)]
```
matklad, line 115: nit: I thing you can say something like `#[clap(long, conflicts_with(stream))]` or something like that to make them mutually exclusive.

test-utils/state-viewer/src/commands.rs
```diff
@@ -52,10 +49,22 @@ pub(crate) fn state(home_dir: &Path, near_config: NearConfig, store: Arc<Store>)
...
+    stream: bool,
+    mut single_file: bool,
+    file: Option<PathBuf>,
     home_dir: &Path,
     near_config: NearConfig,
     store: Arc<Store>,
 ) {
+    assert!(
+        !stream || !single_file,
+        "At most one of options --stream and --single_file can be set"
```
matklad, line 61: yeah, you def shouldn't be writing code for this, clap can do that: 
Might use `mark::hit!();` here: 

https://github.com/rust-analyzer/rust-analyzer/blob/d54e1157b66017e4aae38328cd213286e39ca130/crates/test_utils/src/mark.rshttps://docs.rs/clap/latest/clap/struct.Arg.html#method.conflicts_with

(not super obvious how to plug that into derive, but should be possilbe)

test-utils/state-viewer/src/state_dump.rs
```diff
@@ -42,16 +42,77 @@ pub fn state_dump(
...
+    genesis_config.genesis_height = genesis_height;
+    genesis_config.validators = validators
+        .iter()
+        .map(|(account_id, (public_key, amount))| AccountInfo {
+            account_id: account_id.clone(),
+            public_key: public_key.clone(),
+            amount: *amount,
+        })
+        .collect();
+    genesis_config.validators.sort();
```
matklad, line 57: Rather than implementing `Ord`, you can also `genesis_config.validators.sort_by_key(|it| (&it.account_id, &it.public_key, &it.amount));`. Though I think `Ord` is fine.

matklad (approved):

test-utils/state-viewer/src/state_dump.rs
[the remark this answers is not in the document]
pmnoxx: I would say, that it's not great. It hides an intermediate memory allocation. I would rather call sort directly. I like the usage of `.iter()` in general, there is a hidden assumption, that allocations don't happen in between. Unfortunately, something like `.sorted()` breaks them.

Option 1 is good. Adding `Ord` on `AccountInfo` is reasonable.
Also that way we avoid adding any intermediate memory allocations, by simply sorting at the end.

test-utils/state-viewer/src/commands.rs
pmnoxx: @nikurt Yes, I didn't mean in this pr. It makes sense to be consistent. 

I didn't mean to use tracing now, but rather we should consider changing  to using `tracing` in general rather than `println`. That is out of scope of this PR.

test-utils/state-viewer/src/state_dump.rs
pmnoxx: I agree, `sort_by_key` is even better.

pmnoxx (approved):
LGTM

test-utils/state-viewer/src/cli.rs
pmnoxx: @matklad Thanks, that's what I was looking for, for my own personal scripts.

test-utils/state-viewer/src/cli.rs
```diff
@@ -104,13 +101,35 @@ impl StateViewerSubCommand {
 
 #[derive(Clap)]
 pub struct DumpStateCmd {
+    /// Optionally, can specify at which height to dump state.
     #[clap(long)]
     height: Option<BlockHeight>,
+    /// Dumps state records and genesis config into separate files.
```
mm-near, line 107: so you can set both 'stream' and 'single file output' to true at the same time? What happens then?

test-utils/state-viewer/src/commands.rs
```diff
@@ -52,10 +49,22 @@ pub(crate) fn state(home_dir: &Path, near_config: NearConfig, store: Arc<Store>)
 
 pub(crate) fn dump_state(
     height: Option<BlockHeight>,
+    stream: bool,
+    mut single_file: bool,
```
mm-near, line 53: the fact that this argument is mutable is very surprising

test-utils/state-viewer/src/commands.rs
```diff
@@ -52,10 +49,22 @@ pub(crate) fn state(home_dir: &Path, near_config: NearConfig, store: Arc<Store>)
...
     home_dir: &Path,
     near_config: NearConfig,
     store: Arc<Store>,
 ) {
+    assert!(
+        !stream || !single_file,
+        "At most one of options --stream and --single_file can be set"
+    );
+    if !stream && !single_file {
+        println!("Assuming a single file output is requested");
```
mm-near, line 64: This logic (with stream and single file coming via two different bools, but then exactly one of them should be true) is a little bit confusing.

Maybe we should get rid of the single_file option and say:

if stream {
    // stuff
} else {
    // put stuff in a single file
}

test-utils/state-viewer/src/state_dump.rs
```diff
@@ -42,16 +42,77 @@ pub fn state_dump(
...
+            // `total_supply` is expected to change due to the natural processes of burning tokens and
+            // minting tokens every epoch.
+            genesis_config.total_supply = total_supply;
+            near_config.genesis = Genesis::new(genesis_config, records.into());
+        }
+    }
+    near_config
+}
 
+fn get_records(
```
mm-near, line 108: comment what this function does - (and consider renaming to something like 'iterate' - as it seems that you call the callback on each item rather than returning it)

test-utils/state-viewer/src/state_dump.rs
```diff
@@ -1,25 +1,25 @@
...
 use near_store::TrieIterator;
 use nearcore::config::NearConfig;
 use nearcore::NightshadeRuntime;
+use serde::ser::{SerializeSeq, Serializer};
+use std::collections::HashMap;
+use std::fs;
+use std::fs::File;
+use std::path::Path;
 
 pub fn state_dump(
```
mm-near, line 17: comment what this function does - especially the case where records_path is present vs missing

mm-near (changes requested):

test-utils/state-viewer/src/cli.rs
matklad: @nikurt @pmnoxx as a meta note, Clap is one of those crates where the answer to "How do I ..." is "Yes". It can do almost anything. Given that we have a fair amount of CLI here and there, it might make sense to dedicate some time to reading the docs: https://docs.rs/clap/latest/clap/struct.Arg.html

test-utils/state-viewer/src/cli.rs
nikurt: Updated. Those options are now mutually exclusive.

test-utils/state-viewer/src/cli.rs
```diff
@@ -104,13 +101,22 @@ impl StateViewerSubCommand {
...
+    /// Has reasonable RAM requirements.
+    /// Use for chains with large state, such as mainnet and testnet.
```
mm-near, line 109: I'd also mention:

if false - writes all the information into a single file, which is useful for smaller networks

test-utils/state-viewer/src/cli.rs
```diff
@@ -104,13 +101,22 @@ impl StateViewerSubCommand {
...
+    #[clap(long)]
+    stream: bool,
+    /// Location of the dumped state.
+    #[clap(long, parse(from_os_str))]
+    file: Option<PathBuf>,
```
mm-near, line 114: this is a file (if stream == false) or directory (if stream == true)

mm-near (approved):

PR #5590, opened 2021-12-02
documentation: documentation for different tools

mm-near:


bowenwang1996 (commented):
@mm-near for epoch manager, we have spec level documentation here https://github.com/near/NEPs. We may want to do cross link it here

mm-near:
thanks - done.

chain/epoch_manager/README.md
```diff
@@ -0,0 +1,40 @@
+# near-epoch-manager crate
+
+Epoch manager crate is responsible for code related to epochs and epoch switching.
+An epoch is a unit of time when validators of the network remain constant.
```
matklad, line 4:
```suggestion
An epoch is a unit of time when the set of validators of the network remains constant.
```

chain/epoch_manager/README.md
```diff
@@ -0,0 +1,40 @@
...
+
+You can read more about the epoch here: https://docs.near.org/docs/concepts/epoch
+
+You can read more about Epoch finalization and Epoch changes here: https://github.com/near/NEPs/blob/master/specs/BlockchainLayer/EpochManager/EpochManager.md 
+
+## EpochManager
+Main class that has two main functions:
+* it creates new epochs (EpochIds)
+* allows accessing information about past and current epochs.
```
matklad, line 13: I'd maybe start with more explicit list of what is the actual data? 

Ie, epoch is the set of validators, plus hashes of boundary blocks, etc.

chain/telemetry/README.md
```diff
@@ -0,0 +1,9 @@
+# Near Telemetry
+
+A small utility (TelemetryActor), that tries to send the telemetry (metrics) information as JSON over HTTP-post to selected list of servers.
+
+The data, that is sent over, is of type TelemetryInfo, and is signed with the server's key.
```
matklad, line 5:
```suggestion
The data that is sent over is of type TelemetryInfo, and is signed with the server's key.
```

chain/telemetry/README.md
```diff
@@ -0,0 +1,9 @@
+# Near Telemetry
+
+A small utility (TelemetryActor), that tries to send the telemetry (metrics) information as JSON over HTTP-post to selected list of servers.
```
matklad, line 3: Could you explain the reason why we need this? Specifically: 

* what are those select servers?
* what they do? 

Perhaps there's some dashbord we can link which displays the data?

core/chain-configs/README.md
```diff
@@ -0,0 +1,13 @@
...
+
+## Genesis config
+Genesis config is the one that 'defines' the chain. It was set at the beginning and generally is not mutable. 
+
+## Client config
+
+Client config is the part of the config that client can configure on their own - it controls things like: how many peers it should connect to before syncing, which shards to track etc.
+
+## Protocol config
+This is the class that is spanning over GenesisConfig and RuntimeConfig. People should not use it directly, but use the ProtocolConfigView class instead.
```
matklad, line 13:
```suggestion
This is the type that is spanning over GenesisConfig and RuntimeConfig. People should not use it directly, but use the ProtocolConfigView class instead.
```

I generally call things types or data types. A term of art used in Rust compilers is `ADT`, which is a short for "A Data Type" :)

runtime/runtime/src/genesis.rs
```diff
@@ -21,9 +21,13 @@ use near_store::{
 
 use crate::config::RuntimeConfig;
 use crate::Runtime;
-
+/// Computes the expected storage per account for a given set of StateRecord(s).
```
matklad, line 24:
```suggestion
/// Computes the expected storage per account for a given stream of StateRecord(s).
```

the fact that we do streaming is salient here -- otherwised this would be a function.

runtime/runtime/src/lib.rs
```diff
@@ -1453,7 +1453,8 @@ impl Runtime {
         state_update.commit(StateChangeCause::Migration);
     }
 
-    /// It's okay to use unsafe math here, because this method should only be called on the trusted
+    /// Computes the expected storage per account for a given set of StateRecord(s).
+    /// Note: It's okay to use unsafe math here, because this method should only be called on the trusted
     /// state records (e.g. at launch from genesis)
```
matklad, line 1458: The note should probably be moved to the `StorageComputer`

tools/storage-usage-delta-calculator/README.md
```diff
@@ -0,0 +1,6 @@
+# Storage delta calculator
+
+A small tool to compare the actual storage use with the one saved within the state.
+Useful as a sanity check if we do any updates (as storage has impact on the amount of tokens that we lock in people's accounts).
+
+WARNING: This was built as a one-time tool - to debug the issue during migration in May 2021 - and is no longer maintained.
```
matklad, line 6: Should we just drop it? Every binary we have in the repo adds substantial amount of linking time.

matklad (commented):
LGTM, but I object to `class` terminology, `type` would function better for Rust </terrible puns>

chain/epoch_manager/README.md
matklad: Actually, how do epoch interract with sharding? Am I correct that the epoch is the same for all shards, but each shard has its own set of validators?

EgorKulikov: theoretically yes, but currently all shards have same set of validators

chain/epoch_manager/README.md
```diff
@@ -0,0 +1,40 @@
...
+
+### New Epoch Creation
+When 'finalize_epoch' is called, the EpochManager will do all the necessary processing (like computing validator rewards, selecting validators for the next epoch etc) and create the new EpochId/EpochInfo.
```
EgorKulikov, line 16: We are actually selecting validators for epoch after the next at this point

EgorKulikov (approved):

core/chain-configs/README.md
bowenwang1996: @matklad wait I thought ADT stands for Algebraic Data Type?

bowenwang1996 (commented):
@mm-near please address comments

core/chain-configs/README.md
matklad: Yeah, the official correct definition for ADT is algebraic data types. 

In rust though, user-defined types are not, strictly speaking, ADTs: there are `unions`, product types come in two flavors (tuple and record structs), and recursive types are not possible without indirection. That's why calling some random Rust type ADT is pedantically wrong. The hack is to re-define "ADT" as an "A data type" :)

tools/storage-usage-delta-calculator/README.md
mm-near: This is a great question - let's discuss it on Zulip. 

From my side, I'm all for removing old tools (or at least disabling their compilation and/or moving to separate place)

chain/epoch_manager/README.md
mm-near: yes, epoch is the same for all shards. Number of shards can only change during epoch change.

chain/epoch_manager/README.md
mm-near: good catch - fixed.

chain/epoch_manager/README.md
mm-near: done. BTW - I don't think that epoch manager returns hashes of boundary blocks (at least I didn't see it)

chain/telemetry/README.md
mm-near: done.

chain/telemetry/README.md
mm-near: I don't know what these servers are (see the TODO at the end of the README).

@frol @telezhnaya - do you happen to know where does this telemetry data end up ?

core/chain-configs/README.md
mm-near: fixed.

runtime/runtime/src/genesis.rs
mm-near: done.

runtime/runtime/src/lib.rs
mm-near: done.

chain/telemetry/README.md
frol: [By default](https://github.com/near/nearcore/blob/1cbcc5eb89be9c78364d2ab984a192c77f9538aa/nearcore/src/config.rs#L849-L858), the telemetry [is configured](https://s3-us-west-1.amazonaws.com/build.nearprotocol.com/nearcore-deploy/mainnet/config.json) to send the telemetry to NEAR Explorer.

runtime/runtime/src/genesis.rs
```diff
@@ -32,7 +36,10 @@ impl<'a> StorageComputer<'a> {
         Self { result: HashMap::new(), config: &config.transaction_costs.storage_usage_config }
     }
 
+    /// Updates user's storage info based on the StateRecord.
     pub fn process_record(&mut self, record: &StateRecord) {
+        /// Note: It's okay to use unsafe math here, because this method should only be called on the trusted
+        /// state records (e.g. at launch from genesis)
```
matklad, line 42:
```suggestion
        // Note: It's okay to use unsafe math here, because this method should only be called on the trusted
        // state records (e.g. at launch from genesis)
```

`///` only works for definitions displayed by `cargo doc`. Three slashes is not a mere convention, it's a first-class syntax supported by the compiler for documentation comments on some, but not all, entities.

matklad (approved):
LGTM, though there are still some `class`es left.

bowenwang1996 (approved):

bowenwang1996:
@mm-near please fix CI

PR #5591, opened 2021-12-02
chain: rosetta: move network identifier checking to common function

mina86:
Introduce `check_network_identifier` function which checks whether
network identifier given by the user is what we expect.  This reduces
code duplication moving all those checks into a single place.

Furthermore, the function now checks blockchain and sub network
identifier fields as well.


chain/rosetta-rpc/src/lib.rs
```diff
@@ -45,7 +79,7 @@ async fn network_list(
         .map_err(|err| errors::ErrorKind::InternalError(err.to_string()))?;
     Ok(Json(models::NetworkListResponse {
         network_identifiers: vec![models::NetworkIdentifier {
-            blockchain: "nearprotocol".to_owned(),
+            blockchain: BLOCKCHAIN.to_owned(),
```
bowenwang1996, line 82: I think `to_string` is more idiomatic?

bowenwang1996 (approved):

mina86:
@frol, friendly ping

frol (approved):

frol:
@mina86 Somehow I missed this PR in my notifications. Thanks for pinging me!

PR #5596, opened 2021-12-03
feat: Optimize `RoutingTableView` iterators

pmnoxx:
I'm trying to improve `pub fn get_accounts_keys(&mut self)`.

Is there a way to avoid doing memory allocation here?

### DIFF
--- a/chain/network/src/peer_manager/peer_manager_actor.rs
+++ b/chain/network/src/peer_manager/peer_manager_actor.rs
@@ -676,7 +676,7 @@ impl PeerManagerActor {
             addr.do_send(SendMessage {
                 message: PeerMessage::SyncRoutingTable(RoutingTableUpdate::new(
                     known_edges,
-                    known_accounts,
+                    known_accounts.cloned().collect(),
                 )),
             });
 
@@ -1393,7 +1393,7 @@ impl PeerManagerActor {
                        reason = ?find_route_error,
                        ?msg,"Drop message",
                 );
-                trace!(target: "network", known_peers = ?self.routing_table_view.get_accounts_keys());
+                trace!(target: "network", known_peers = ?self.routing_table_view.get_accounts_keys().collect::<Vec<_>>(), "Known peers");
                 return false;
             }
         };
@@ -1471,7 +1471,6 @@ impl PeerManagerActor {
             known_producers: self
                 .routing_table_view
                 .get_announce_accounts()
-                .iter()
                 .map(|announce_account| KnownProducer {
                     account_id: announce_account.account_id.clone(),
                     peer_id: announce_account.peer_id.clone(),
--- a/chain/network/src/routing/routing_table_view.rs
+++ b/chain/network/src/routing/routing_table_view.rs
@@ -79,7 +79,7 @@ impl RoutingTableView {
         self.local_edges_info.get(other_peer).map_or(0, |x| x.nonce()) < nonce
     }
 
-    pub fn reachable_peers(&self) -> impl Iterator<Item = &PeerId> {
+    pub fn reachable_peers(&self) -> impl Iterator<Item = &PeerId> + ExactSizeIterator {
         self.peer_forwarding.keys()
     }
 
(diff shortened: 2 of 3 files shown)


pmnoxx:
@matklad The way I'm doing it doesn't seem to be working
```
                                                                                                                                                                                                                            
error[E0277]: `impl std::iter::Iterator` doesn't implement `std::fmt::Debug`                                                                                                                                                                                                                                                                                                                 
    --> chain/network/src/peer_manager/peer_manager_actor.rs:1377:17                                                                                                                                                                                                                                                                                                                         
     |                                                                                                                                                                                                                                                                                                                                                                                       
1377 |                 trace!(target: "network", message = "Known peers", known_peers = ?self.routing_table_view.get_accounts_keys());                                                                                                                                                                                                                                                       
     |                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ `impl std::iter::Iterator` cannot be formatted using `{:?}` because it doesn't implement `std::fmt::Debug`                                                                                                                                            
     |                                                                                                                                                                                                                                                                                                                                                                                       
     = help: the trait `std::fmt::Debug` is not implemented for `impl std::iter::Iterator`                                                                                                                                                                                                                                                                                                   
     = note: required because of the requirements on the impl of `std::fmt::Debug` for `&impl std::iter::Iterator`                                                                                                                                                                                                                                                                           
     = note: required because of the requirements on the impl of `tracing::Value` for `tracing::field::DebugValue<&impl std::iter::Iterator>`                                                                                                                                                                                                                                                
     = note: required for the cast to the object type `dyn tracing::Value`                                                                    
```

pmnoxx:
@matklad Is there a way to avoid using collect()?
```
       trace!(target: "network", message = "Known peers", known_peers = ?self.routing_table_view.get_accounts_keys().collect::<Vec<&AccountId>>());
          
```

pmnoxx:
@mm-near Can you take a look?

pmnoxx:
```
```

chain/network/src/routing/routing.rs
```diff
@@ -267,22 +267,23 @@ impl RoutingTableView {
...
-        self.account_peers.iter().map(|(k, _v)| (k.clone())).collect()
+    pub fn get_accounts_keys(&mut self) -> impl Iterator<Item = &AccountId> {
+        self.account_peers.iter().map(|(k, _v)| k)
     }
 
     /// Get announce accounts on cache.
-    pub fn get_announce_accounts(&mut self) -> Vec<AnnounceAccount> {
-        self.account_peers.iter().map(|(_k, v)| v).cloned().collect()
+    pub fn get_announce_accounts(&mut self) -> impl Iterator<Item = &AnnounceAccount> {
+        self.account_peers.iter().map(|(_k, v)| v)
```
matklad, line 286: You can return `Iterator + ExactSizeIterator`, and then `get_announce_accounts_size` becomes unnecessary

chain/network/src/peer_manager/peer_manager_actor.rs
```diff
@@ -1374,7 +1374,7 @@ impl PeerManagerActor {
                        find_route_error,
                        msg,
                 );
-                trace!(target: "network", "Known peers: {:?}", self.routing_table_view.get_accounts_keys());
+                trace!(target: "network", message = "Known peers", known_peers = ?self.routing_table_view.get_accounts_keys().collect::<Vec<&AccountId>>());
```
matklad, line 1377: I don't think there's an easy way to avoid this collect. My understanding is that tracing won't evaluate the expression unless the `trace` level is actually enabled, so this is probably not important. `.collect` here also should be cheaper than the actual string formatting.

matklad: Also, we ended up not using `message = `

pmnoxx:
```
```

mm-near (approved):

PR #999, opened 2019-06-11
Use dyn for trait object

bowenwang1996:
Rust is deprecating trait object without `dyn` keyword.

PR #5598, opened 2021-12-03
Fix an error, which confuses Rust plugin for Intellij

pmnoxx:
Rust plugin for intellij gets confused and prints an error
```
mismatches types [E308]
expecterd `&str`, found `&AccountId`
```


pmnoxx:
@matklad Can you take a look?

bowenwang1996:
I have seen a number of fixes related to some specific IDE. @matklad I wonder what your opinion is on this topic

matklad:
It's best to report such issues to https://github.com/intellij-rust/intellij-rust. I try to do this for PRs I see, but I've missed this one. @pmnoxx it might be less effort long-term to report bugs, as that would lead to fewer false errors in other places :) 

No strong opinion on whether we should place work-arounds for IDE bugs into our code. I'd say generally it doesn't matter either way.

pmnoxx:
> It's best to report such issues to https://github.com/intellij-rust/intellij-rust. I try to do this for PRs I see, but I've missed this one. @pmnoxx it might be less effort long-term to report bugs, as that would lead to fewer false errors in other places :)

@matklad I would prefer reporting the issue there.

 However, I have a few questions
- There are over 1200 open issues - should I try finding if there is a duplicate first?
- When reporting an issue, what's the best way to represent the issue? - Like, maybe giving them link to PR, such as this one?
- An example would be nice, on what is the preferred format on doing it. Just reporting an issue, and growing list of open issues, isn't going to help anyone, unless it's done in the right way.

Anyway, writing a fix, and submitting PR, only takes a few minutes, so it's fine.I only do it for files, that I'm working with.

matklad:
The general rule-of-thumb is that the more time you spend filing the issue, the faster it will be fixed. However it's OK to file issue without to much research, as long as it is reproducible. 

That is:

* "Something doesn't work for me sometimes, I don't know how to reproduce" is not useful to report, unless you are willing to do some back and forth with the maintainers to debug the thing
* "Hey, I see a false error in this file on this comment in this gigantic" repo is useful to report, but it'll take some time until a maintainer is able to checkout your repo and reproduce
* "Hey, I had this false error in my project at work, and I was able to minimize this to this single-file example. This also seems related, but not identical, to issue #999" would probably be fixed in a couple of weeks.

PR #5599, opened 2021-12-03
re: Simplify logic in `find_route_from_peer_id`

pmnoxx:
The code in `near-network::RoutingTableView::find_route_from_peer_id` is over complicated. It can be written in a simplified way.


pmnoxx:
@matklad Can you take a look?

mm-near (approved):

PR #5600, opened 2021-12-03
re: Remove utils.rs - refactor `fetch_ping_ping`

pmnoxx:
In Rust it's preferred to let the called choose how he wants the results to be represented. 
In case caller wants a different type, for example `Vec` instead of `HashMap`, extra copies can be avoided.

Let's remove `cache_to_hashmap` and `chain/network/src/routing/utils.rs`.

This improves code quality, which is always good.


pmnoxx:
@matklad Can you take a look?

mm-near (approved):

PR #5601, opened 2021-12-03
re: `near-network`: Rewrite `local_edge_info` in `RoutingTableView`

pmnoxx:
We can make a few code simplifications in `RoutingTableView`.
We currently only keep local edges in `local_edges_info`, once of peers is `my_peer_id`, so there is no need to store a pair anymore.

We can change signature of `local_edges_info`
```
pub local_edges_info: HashMap<(PeerId, PeerId), Edge>,
```
to
```
pub local_edges_info: HashMap<PeerId, Edge>,
```

### DIFF
--- a/chain/network/src/peer_manager/peer_manager_actor.rs
+++ b/chain/network/src/peer_manager/peer_manager_actor.rs
@@ -267,11 +267,11 @@ impl PeerManagerActor {
             .into_actor(self)
             .map(|response, act, ctx| match response {
                 Ok(RoutingTableMessagesResponse::RoutingTableUpdateResponse {
-                    local_edges_to_remove: edges_to_remove,
+                    local_edges_to_remove,
                     peer_forwarding,
                     peers_to_ban,
                 }) => {
-                    act.routing_table_view.remove_edges(&edges_to_remove);
+                    act.routing_table_view.remove_local_edges(&local_edges_to_remove);
                     act.routing_table_view.peer_forwarding = peer_forwarding;
                     for peer in peers_to_ban {
                         act.ban_peer(ctx, &peer, ReasonForBan::InvalidEdge);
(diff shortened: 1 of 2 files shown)


chain/network/src/routing/routing.rs
```diff
@@ -194,8 +193,9 @@ impl RoutingTableView {
 
     pub fn remove_edges(&mut self, edges: &Vec<Edge>) {
         for edge in edges.iter() {
-            assert!(edge.key().0 == self.my_peer_id || edge.key().1 == self.my_peer_id);
-            self.local_edges_info.remove(edge.key());
+            if let Some(other_peer) = edge.other(&self.my_peer_id) {
+                self.local_edges_info.remove(other_peer);
```
mm-near, line 197: now we're going to silently ignore the edges that don't start from the current peer (before we were asserting).

if this was intended - please add the comment to the function

mm-near: or maybe rename to something like 'remove_edges_if_local' ?

chain/network/src/peer_manager/peer_manager_actor.rs
```diff
@@ -1900,19 +1885,22 @@ impl PeerManagerActor {
...
-                            if edge.nonce() >= *nonce {
-                                self.local_peer_pending_update_nonce_request.remove(other);
+                if let Some(other_peer) = edge.other(&self.my_peer_id) {
+                    if edge.verify() {
+                        if self.routing_table_view.is_local_edge_newer(other_peer, edge.nonce()) {
+                            if let Some(nonce) =
+                                self.local_peer_pending_update_nonce_request.get(other_peer)
+                            {
+                                if edge.nonce() >= *nonce {
+                                    self.local_peer_pending_update_nonce_request.remove(other_peer);
```
mm-near, line 1895: add comment to explain this case

mm-near (approved):

chain/network/src/routing/routing.rs
pmnoxx: Ok, I renamed the method to `remove_local_edges` and I added panic!.

PR #5602, opened 2021-12-03
fix: Dont broadcast empty `SyncData`

pmnoxx:
There was one case, where we could broadcast an empty list of edges inside `add_verified_edges_to_routing_table`.
Let's fix that.

### DIFF
--- a/chain/network/src/peer_manager/peer_manager_actor.rs
+++ b/chain/network/src/peer_manager/peer_manager_actor.rs
@@ -301,12 +301,18 @@ impl PeerManagerActor {
             .map(move |response, act, ctx| match response {
                 Ok(RoutingTableMessagesResponse::AddVerifiedEdgesResponse(filtered_edges)) => {
                     // Broadcast new edges to all other peers.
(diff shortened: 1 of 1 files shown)


chain/network/src/peer_manager/peer_manager_actor.rs
```diff
@@ -301,12 +301,18 @@ impl PeerManagerActor {
             .map(move |response, act, ctx| match response {
                 Ok(RoutingTableMessagesResponse::AddVerifiedEdgesResponse(filtered_edges)) => {
                     // Broadcast new edges to all other peers.
-                    if broadcast_edges && act.adv_helper.can_broadcast_edges() {
-                        let new_data =
-                            SyncData { edges: filtered_edges, accounts: Default::default() };
+                    if broadcast_edges
```
mm-near, line 304: when we do fix like this - let's make sure that we add a test that covers it.

mm-near (changes requested):

pmnoxx:
```
```

pmnoxx:
```
```

pmnoxx:
```
```

PR #5603, opened 2021-12-03
re: `near-network`: Split `Graph` from `RoutingTableView`

pmnoxx:
I think code will become easier to read after splitting `Graph` from `RoutingTableView`.

Both structures are independent from each other. `Graph` is used by `RoutingTableActor`. while `RoutingTableView` is used by `PeerManagerActor`.


matklad (approved):
LGTM! Though, I'd say that the motivation here is not that the struct are big, but that they are independent: Graph doesn't use RoutingTableView, and RoutingTableView doesn't use Graph, so theer's little reason to keep them in the same module.

pmnoxx:
> LGTM! Though, I'd say that the motivation here is not that the struct are big, but that they are independent: Graph doesn't use RoutingTableView, and RoutingTableView doesn't use Graph, so theer's little reason to keep them in the same module.

Yes, both structures are independent from each other. `Graph` is used by `RoutingTableActor`. while `RoutingTableView` is used by `PeerManagerActor`.

mm-near (approved):

PR #5605, opened 2021-12-03
re: Simlify/Optimize `Decoder::decode`

pmnoxx:
Write `Decoder::decode` in a cleaner way.


chain/network/src/peer/codec.rs
```diff
@@ -78,9 +78,7 @@ impl Decoder for Codec {
             return Ok(None);
         }
 
-        let mut len_bytes: [u8; 4] = [0; 4];
-        len_bytes.copy_from_slice(&buf[0..4]);
-        let len = u32::from_le_bytes(len_bytes);
+        let len = u32::from_le_bytes([buf[0], buf[1], buf[2], buf[3]]);
```
matklad, line 81: Looks like we can use https://docs.rs/bytes/latest/bytes/buf/trait.Buf.html#method.get_u32_le ?

matklad (commented):
unrelated to the PR, but I wonder if we should use `type Item = Result<Bytes, ReasonForBan>;` rather than `Vec<u8>` to avoid copying?

chain/network/src/peer/codec.rs
pmnoxx: @matklad Thanks, that's even better.

matklad: Meta note on how to know such things: 

I've noticed that both versions do not exactly spark joy. At the same time, reading length prefix seems like a common operation one might do, and it feels like there *should* be a better way of doing this. I wanted to grep some tokio code to see if I can find the relevant pattern. But before that, I reasoned that it might be a good idea to check the docs for the type (it seems reasonable that the functionality can live on the `BytesMut` type itself). I actually was looking for something with `&mut BytesMut -> Option<[u8; N]>` signature, but just stumbled upon the very specific API.

chain/network/src/peer/codec.rs
```diff
@@ -78,9 +78,7 @@ impl Decoder for Codec {
             return Ok(None);
         }
 
-        let mut len_bytes: [u8; 4] = [0; 4];
-        len_bytes.copy_from_slice(&buf[0..4]);
-        let len = u32::from_le_bytes(len_bytes);
+        let len = buf.get_u32_le();
```
matklad, line 81: Wait, I think this is not actually buggy -- `get_u32_le` actually advances the buffer. That is, we should change the code bellow, `buf[4..4 + len as usize].to_vec()` to be just  `but.split_of(len).to_vec()`. Do we have any failing tests in this PR? If we don't we miss some!

chain/network/src/peer/codec.rs
pmnoxx: @matklad I tried looking for some apis, that convert `&[u8]` to `[u8;4]`, though I couldn't find anything. 
`fn get_u32_le(&mut self) -> u32` doesn't even have `[u8;4]` in it's signature either.

Though, I should have checked the `buf` structs. Anyway, it was a good idea to check.

chain/network/src/peer/codec.rs
pmnoxx: @matklad Yes, a lot of tests are failing.

Actually, consuming here is no a good idea. We first peek at the value, and then we try to determine whenever we have read enough bytes to read the whole message. We don't want to consume if that didn't happen.

We could make the Codec into a state machine, which remembers last length read, but I would rather avoid it.

I replaced the code above with
```rust
        let len = u32::from_le_bytes([buf[0], buf[1], buf[2], buf[3]]);
```
for now.

pmnoxx: @matklad Btw, this `to_vec()` is inefficient. I already have plans for it: https://github.com/near/nearcore/pull/5605

pmnoxx:
> unrelated to the PR, but I wonder if we should use `type Item = Result<Bytes, ReasonForBan>;` rather than `Vec<u8>` to avoid copying?

I think, we should do it like this to avoid copying: https://github.com/near/nearcore/pull/5605

PR #5610, opened 2021-12-03
feat: increase `OLDEST_PROTOCOL_VERSION`  from `34` to `LATEST - 1`

pmnoxx:
We haven't bumped minimum required version for more than a year, since 2020/10/01.

Let's upgrade minimum protocol version from `34` to `39`. This will allow us to drop deprecated code in `HandshakeV2`.

The current stable version used is '49`.

Follow up:
- [x] https://github.com/near/nearcore/pull/5611

### DIFF
--- a/core/primitives/src/version.rs
+++ b/core/primitives/src/version.rs
@@ -21,9 +21,6 @@ pub const DB_VERSION: DbVersion = 30;
 /// Protocol version type.
 pub use near_primitives_core::types::ProtocolVersion;
 
-/// Oldest supported version by this client.
-pub const OLDEST_BACKWARD_COMPATIBLE_PROTOCOL_VERSION: ProtocolVersion = 34;
-
 /// Minimum gas price proposed in NEP 92 and the associated protocol version
 pub const MIN_GAS_PRICE_NEP_92: Balance = 1_000_000_000;
 pub const MIN_PROTOCOL_VERSION_NEP_92: ProtocolVersion = 31;
@@ -142,15 +139,23 @@ pub enum ProtocolFeature {
     RoutingExchangeAlgorithm,
 }
 
-/// Current latest stable version of the protocol.
+/// Both, outgoing and incoming tcp connections to peers, will be rejected if `peer's`
+/// protocol version is lower than this.
+pub const PEER_MIN_ALLOWED_PROTOCOL_VERSION: ProtocolVersion = MAIN_NET_PROTOCOL_VERSION - 1;
+
+/// Current protocol version used on the main net.
 /// Some features (e. g. FixStorageUsage) require that there is at least one epoch with exactly
 /// the corresponding version
-#[cfg(not(feature = "nightly_protocol"))]
-pub const PROTOCOL_VERSION: ProtocolVersion = 50;
+const MAIN_NET_PROTOCOL_VERSION: ProtocolVersion = 50;
 
 /// Current latest nightly version of the protocol.
+const TEST_NET_PROTOCOL_VERSION: ProtocolVersion = 125;
+
+/// Version used by this binary.
+#[cfg(not(feature = "nightly_protocol"))]
+pub const PROTOCOL_VERSION: ProtocolVersion = MAIN_NET_PROTOCOL_VERSION;
 #[cfg(feature = "nightly_protocol")]
-pub const PROTOCOL_VERSION: ProtocolVersion = 125;
+pub const PROTOCOL_VERSION: ProtocolVersion = TEST_NET_PROTOCOL_VERSION;
 
 impl ProtocolFeature {
     pub const fn protocol_version(self) -> ProtocolVersion {
--- a/core/primitives/src/version.rs
+++ b/core/primitives/src/version.rs
@@ -148,14 +148,12 @@ pub const PEER_MIN_ALLOWED_PROTOCOL_VERSION: ProtocolVersion = MAIN_NET_PROTOCOL
 /// the corresponding version
 const MAIN_NET_PROTOCOL_VERSION: ProtocolVersion = 50;
 
-/// Current latest nightly version of the protocol.
-const TEST_NET_PROTOCOL_VERSION: ProtocolVersion = 125;
-
 /// Version used by this binary.
 #[cfg(not(feature = "nightly_protocol"))]
 pub const PROTOCOL_VERSION: ProtocolVersion = MAIN_NET_PROTOCOL_VERSION;
+/// Current latest nightly version of the protocol.
 #[cfg(feature = "nightly_protocol")]
-pub const PROTOCOL_VERSION: ProtocolVersion = TEST_NET_PROTOCOL_VERSION;
+pub const PROTOCOL_VERSION: ProtocolVersion = 125;
 
 impl ProtocolFeature {
     pub const fn protocol_version(self) -> ProtocolVersion {
--- a/chain/network/src/network_protocol.rs
+++ b/chain/network/src/network_protocol.rs
@@ -15,7 +15,7 @@ use near_primitives::network::{AnnounceAccount, PeerId};
 use near_primitives::syncing::{EpochSyncFinalizationResponse, EpochSyncResponse};
 use near_primitives::transaction::SignedTransaction;
 use near_primitives::types::{EpochId, ProtocolVersion};
-use near_primitives::version::{OLDEST_BACKWARD_COMPATIBLE_PROTOCOL_VERSION, PROTOCOL_VERSION};
+use near_primitives::version::{PEER_MIN_ALLOWED_PROTOCOL_VERSION, PROTOCOL_VERSION};
 use std::fmt::Formatter;
 use std::{fmt, io};
 
@@ -73,7 +73,7 @@ impl Handshake {
     ) -> Self {
         Handshake {
             protocol_version: version,
-            oldest_supported_version: OLDEST_BACKWARD_COMPATIBLE_PROTOCOL_VERSION,
+            oldest_supported_version: PEER_MIN_ALLOWED_PROTOCOL_VERSION,
             sender_peer_id: peer_id,
             target_peer_id,
             sender_listen_port: listen_port,
@@ -97,7 +97,7 @@ impl BorshDeserialize for Handshake {
 
         let version = u32::from_le_bytes(buf[..4].try_into().unwrap());
 
-        if OLDEST_BACKWARD_COMPATIBLE_PROTOCOL_VERSION <= version && version <= PROTOCOL_VERSION {
+        if PEER_MIN_ALLOWED_PROTOCOL_VERSION <= version && version <= PROTOCOL_VERSION {
             // If we support this version, then try to deserialize with custom deserializer
             HandshakeAutoDes::deserialize(buf).map(Into::into)
         } else {
@@ -158,7 +158,7 @@ impl HandshakeV2 {
     ) -> Self {
         Self {
             protocol_version: version,
-            oldest_supported_version: OLDEST_BACKWARD_COMPATIBLE_PROTOCOL_VERSION,
+            oldest_supported_version: PEER_MIN_ALLOWED_PROTOCOL_VERSION,
             sender_peer_id: peer_id,
             target_peer_id,
             sender_listen_port,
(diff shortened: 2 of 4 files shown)


matklad:
Wow, so we can actually just bump the oldest protocol version? I was under the impression that this isn't actually possible. It's great if we can do this. 

Two questions here for my own education: 

* What's our policy for bumping this version? What's the criteria for "can bump" and "should bump"? 
* What protocol version did the mainnet started from?

chefsale (approved):

pmnoxx:
> Wow, so we can actually just bump the oldest protocol version? I was under the impression that this isn't actually possible. It's great if we can do this.
> 
> Two questions here for my own education:
> 
> * What's our policy for bumping this version? What's the criteria for "can bump" and "should bump"?
> * What protocol version did the mainnet started from?

There are a few ways we should be doing version upgrades:
- If there are any security vulnerabilities fixed.  After, a while, we should bump up the version, and make sure old clients, with old vulnerable code are not present.
- New protocol features. Recently, we released a feature, which to split work on multiple shards. Most likely old client, don't work anyway, those that aren't aware of those features.

We haven't updated min protocol version for over a year. If I remember correctly, mainnet launched somewhere around that version. 

First phase of Mainnet launched in Apr. 22, 2020.

core/primitives/src/version.rs
```diff
@@ -22,7 +22,7 @@ pub const DB_VERSION: DbVersion = 30;
 pub use near_primitives_core::types::ProtocolVersion;
 
 /// Oldest supported version by this client.
-pub const OLDEST_BACKWARD_COMPATIBLE_PROTOCOL_VERSION: ProtocolVersion = 34;
+pub const OLDEST_BACKWARD_COMPATIBLE_PROTOCOL_VERSION: ProtocolVersion = 39;
```
mm-near, line 25: let's add a comment above on when/how this value  can be changed in the future.

(so that people know that it is changeable - but under some conditions)

mm-near: and just before merging - let's make sure that we notify people from nodeX and indexer (just in case)

mm-near (approved):

bowenwang1996 (commented):
I think we can remove more code with this change right? At least that code of 34..=38 could be removed.

pmnoxx:
> * re: `near-network`: Remove `HandshakeV2` #5611

Yes, I added all removals in re: near-network: Remove HandshakeV2 #5611
This way, we can focus on the main point of the review, which is the discussion about updating the protocol, rather than, whenever given code removal is bad or not.

core/primitives/src/version.rs
pmnoxx: Ok, I updated the documentation.

pmnoxx:
@matklad @mm-near This is my proposal.
```
/// Oldest supported version by this client on the network protocol layer.
/// That is, during handshake negotiation, connection will be rejected if below this version.
/// This should be updated updating `PROTOCOL_VERSION` below.
/// The change should be issued once all notes on `MAINNET` vote to use newer version,
/// that is, we commit to using new version.
pub const NETWORK_LAYER_PROTOCOL_VERSION: ProtocolVersion = 39;

/// Current latest stable version of the protocol on the chain layer.
/// the corresponding version
/// Note, we should update `OLDEST_BACKWARD_COMPATIBLE_PROTOCOL_VERSION` after updating
/// `PROTOCOL_VERSION`.
/// TODO: Rename to `CHAIN_PROTOCOL_VERSION`.
pub const PROTOCOL_VERSION: ProtocolVersion = 49;
```

bowenwang1996:
Instead of manually maintaining `NETWORK_LAYER_PROTOCOL_VERSION`, it probably makes sense to just set it to `PROTOCOL_VERSION - 1` since nodes running older version won't be able to follow the chain anyways (depends on the nature of the protocol change, but mostly likely it is the case).

pmnoxx:
> Instead of manually maintaining `NETWORK_LAYER_PROTOCOL_VERSION`, it probably makes sense to just set it to `PROTOCOL_VERSION - 1` since nodes running older version won't be able to follow the chain anyways (depends on the nature of the protocol change, but mostly likely it is the case).

That's a reasonable solution. It minimizes the amount of work done, not having to submit a PR.

However, there is some risk involved. After this PR, whenever we want to ship a new protocol, we need to note in the release notes, that both, the `NETWORK_LAYER_PROTOCOL_VERSION`, and `PROTOCOL_VERSION` got bumped at the same time.
Technically, we are making multiple changes at the same time, what if someone increases, protocol version not by 1, but by 2.
That is not likely to happen, but that adds extra constrains.

After the upgrade, it makes sense to update `NETWORK_LAYER_PROTOCOL_VERSION` to `PROTOCOL_VERSION - 1` .
And, after the voting is done, update the version to `PROTOCOL_VERSION`.

I think, it's better to be safe than sorry in those cases, and leave updating this value, to actual person, that verifies state of the network.

@mm-near @matklad What is your opinion on the subject?

matklad:
>what if someone increases, protocol version not by 1, but by 2. That is not likely to happen, but that adds extra constrains.

I believe this is not possible -- we guarantee that we increase protocol version by at most `1`, even if we stabilize multiple features. In other words, if we bump by 2, a lot of stuff will break, not only the network. So, defining as `NETWORK_LAYER_PROTOCOL_VERSION = PROTOCOL_VERSION - 1` makes sense. 

I wonder if it can be called `OLDEST_PEER_PROTOCOL_VERSION`? From the discussion here, it seems that the actual semantics is "the PROTOCOL_VERSION for `peers` we are ok connecting with".

mina86:
`pytest/tests/sanity/upgradable.py` will break if protocol version increases more than one compared to what is running in production.

pmnoxx:
> `pytest/tests/sanity/upgradable.py` will break if protocol version increases more than one compared to what is running in production.

Thanks! It's great to know what we have a check for that.

> I wonder if it can be called `OLDEST_PEER_PROTOCOL_VERSION`? From the discussion here, it seems that the actual semantics is "the PROTOCOL_VERSION for `peers` we are ok connecting with".

Yes, I'm also not sure about the naming. I tried `NETWORK_LAYER_PROTOCOL_VERSION`. However, after thinking about that name may suggest, we have two different protocols, each with a separate version.

> I believe this is not possible -- we guarantee that we increase protocol version by at most `1`, even if we stabilize multiple features. In other words, if we bump by 2, a lot of stuff will break, not only the network. So, defining as `NETWORK_LAYER_PROTOCOL_VERSION = PROTOCOL_VERSION - 1` makes sense.

Yes, it's not unreasonable idea. However, after we switch the protocol. And everything is done, no one using the old version anymore. We wouldn't we make `NETWORK_LAYER_PROTOCOL_VERSION = PROTOCOL_VERSION` ?

The are some benefits of setting them manually. For example, if I see that both values are equal, let's say to 49. Then I would know, just by reading the code, that the transition is done. I think it's important to know, whenever we are in the process of doing the transition or not. 

Anyway, I see 4 action items:
- Do first upgrade, make people aware, that we will be updating `OLDEST_BACKWARD_COMPATIBLE_PROTOCOL_VERSION`, and make sure other teams announce it properly. Make sure that there is no issues in the communication. 
- Do the upgrade to of `OLDEST_PEER_PROTOCOL_VERSION` to be equal to current `PROTOCOL_VERSION`, which is 49.
- Decide the name of `OLDEST_BACKWARD_COMPATIBLE_PROTOCOL_VERSION`
- Decide how to handle `OLDEST_BACKWARD_COMPATIBLE_PROTOCOL_VERSION` upgrades

As for how `OLDEST_BACKWARD_COMPATIBLE_PROTOCOL_VERSION` should be updated.

We have two proposals:
A) always set `OLDEST_BACKWARD_COMPATIBLE_PROTOCOL_VERSION` to be `PROTOCOL_VERSION - 1`.
B) Manually update  `OLDEST_BACKWARD_COMPATIBLE_PROTOCOL_VERSION` to be either current value of `PROTOCOL_VERSION - 1` or current value of `PROTOCOL_VERSION`. 
The advantage of this options is that we can say, that given `PROTOCOL_VERSION` will still be supported up to, let's say 2 weeks (or some other period), after transition is over. This gives people time to upgrade, and it makes it less confusing to the users.

The option `A` is better for developers. As that requires least amount of work. However, the drawback is that we will be doing to changes at the same time. 
The option `B` may be better for users of the network, as it makes the data times of when to do version upgrade, and restart the node, more transparent.

I would say, going for option `A` is perfectly reasonable choice, it requires least amount of maintenance, from our side.

bowenwang1996:
> The are some benefits of setting them manually. For example, if I see that both values are equal, let's say to 49. Then I would know, just by reading the code, that the transition is done. I think it's important to know, whenever we are in the process of doing the transition or not.

If the goal is to check whether the protocol version of some network, rpc is a much better way to do it. Given how our release process works, master could already move to another protocol version while on testnet or mainnet the upgrade is still ongoing and it does not make much sense to me to rely on what is in the code to determine the status of the upgrade. 

> Yes, it's not unreasonable idea. However, after we switch the protocol. And everything is done, no one using the old version anymore. We wouldn't we make NETWORK_LAYER_PROTOCOL_VERSION = PROTOCOL_VERSION ?

Unless we make another release, it doesn't help much. If we need to create a new release every time after the network is upgraded, it is too much overhead for too little gain.

I strongly suggest that we avoid manual upkeep if possible and I personally don't see much benefit of doing manual updates here.

pmnoxx:
> > Yes, it's not unreasonable idea. However, after we switch the protocol. And everything is done, no one using the old version anymore. We wouldn't we make NETWORK_LAYER_PROTOCOL_VERSION = PROTOCOL_VERSION ?
> 
> Unless we make another release, it doesn't help much. If we need to create a new release every time after the network is upgraded, it is too much overhead for too little gain.
> 
> I strongly suggest that we avoid manual upkeep if possible and I personally don't see much benefit of doing manual updates here.

There are a few issues I see:
Issue 1)
All messages, which we transport over the network are contained in enum `PeerMessage`.
On of those messages is `Routed(RoutedMessage),`.

Let's say, we upgraded the protocol, to version let's say `PROTOCOL_VERSION=50`.
The way, that I'm reading the current code is. If there is still a client on the network, let's call it `B`, that still uses version lower, let's say `49`.
And we add a new variant in the enum `RoutedMessage`, which requires version `50`. What is going to happen to messages that that are routed between from `A`, then through `B`, then to `C`?

`B` has to decode all messages, and what will happen those those messages?
My current understanding is that given message would be dropped, because `B` doesn't understand the message, leading to some packet loss, and therefore affecting network efficiency.

Therefore, upgrading all nodes to latest version has some advantages.

Issue 2)  Are you use you want to make `NETWORK_LAYER_PROTOCOL_VERSION = PROTOCOL_VERSION -  1`.
This will also effect the testnet. Now, we will require that all nodes on the `testnet`, which use `PROTOCOL_VERSION` 125, will be able to communicate only with versions `124` and above.

Are we use, that most people upgrade `testnet` nodes regularly? 

If those two issues I mentioned above, aren't a concern. We can switch
```
OLDEST_BACKWARD_COMPATIBLE_PROTOCOL_VERSION = PROTOCOL_VERSION - 1;
```

pmnoxx:
> I wonder if it can be called `OLDEST_PEER_PROTOCOL_VERSION`? From the discussion here, it seems that the actual semantics is "the PROTOCOL_VERSION for `peers` we are ok connecting with".

@matklad I like the name. How about `MIN_PEER_ALLOWED_PROTOCOL_VERSION'.

And the description would be 'Connections from a peer, that is below that version will be rejected'.

pmnoxx:
> I strongly suggest that we avoid manual upkeep if possible and I personally don't see much benefit of doing manual updates here.

That is valid reason. I'm all in for a solution that requires least maintenance overhead. However, I want to make sure we are on the same page, when it comes to small nuances, and there there is no miscommunication regarding them.

bowenwang1996:
> Now, we will require that all nodes on the testnet, which use PROTOCOL_VERSION 125, will be able to communicate only with versions 124 and above

That is not the case. Testnet follows stable protocols. You might be referring to betanet, and the short answer is we don't care there given how the network is run (frequent hard forks).

pmnoxx:
@bowenwang1996 Ok, now this blocks my from connecting to mainnet with build from the master branch.

pmnoxx:
https://github.com/near/nearcore/issues/5853

pmnoxx:
@bowenwang1996 Well, it turns out that changing minimum version to `PROTOCOL_VERSION -  1` wasn't such a good idea overall. Now I can't connect with, a build from `master` to MAINNET.

https://github.com/near/nearcore/issues/5853

PR #5611, opened 2021-12-03
re: `near-network`: Remove `HandshakeV2`

pmnoxx:
It has been almost a year, since we deprecated `HandshakeV2`.
We upgrade protocol version in https://github.com/near/nearcore/pull/5610

Now, it's time for cleanup.


chain/network/src/types.rs
```diff
@@ -334,7 +221,8 @@ pub enum PeerMessage {
     /// Gracefully disconnect from other peer.
     Disconnect,
     Challenge(Challenge),
-    HandshakeV2(HandshakeV2),
+    /// Deprecated
+    Unused,
```
mm-near, line 225: maybe let's call it Unused_HandshakeV2 (so that we remember what used to be here)

(or with _ prefix)

core/primitives/src/version.rs
```diff
@@ -22,7 +22,7 @@ pub const DB_VERSION: DbVersion = 30;
 pub use near_primitives_core::types::ProtocolVersion;
 
 /// Oldest supported version by this client.
-pub const OLDEST_BACKWARD_COMPATIBLE_PROTOCOL_VERSION: ProtocolVersion = 34;
+pub const OLDEST_BACKWARD_COMPATIBLE_PROTOCOL_VERSION: ProtocolVersion = 39;
```
mm-near, line 25: let's make sure that other PR gets merged before this one.

mm-near (approved):
Thanks for the cleanup !!

core/primitives/src/version.rs
pmnoxx: Ok, I made this into draft.

PR #5618, opened 2021-12-03
re: Document buggy behavior of `fetch_edges_for_peer_from_disk` 

pmnoxx:
Document buggy behavior of `fetch_edges_for_peer_from_disk`  in response to https://github.com/near/nearcore/pull/5566

Based on https://github.com/near/nearcore/pull/5566, meant to be shipped afterwards.

### DIFF
--- a/chain/network/src/routing/routing_table_actor.rs
+++ b/chain/network/src/routing/routing_table_actor.rs
@@ -176,6 +176,33 @@ impl RoutingTableActor {
     }
 
     /// If peer_id is not in memory check if it is on disk in bring it back on memory.
+    ///
+    /// Note: here an advanced example, which shows what's happening.
+    /// Let's say we have a full graph fully connected with nodes `A, B, C, D`.
+    /// Step 1 ) `A`, `B` get removed.
+    /// We store edges belonging to `A` and `B`: `<A,B>, <A,C>, <A, D>, <B, C>, <B, D>`
+    /// into component 1 let's call it `C_1`.
+    /// And mapping from `A` to `C_1`, and from `B` to `C_1`
+    ///
+    /// Note that `C`, `D` is still active.
+    ///
+    /// Step 2) 'C' gets removed.
+    /// We stored edges <C, D> into component 2 `C_2`.
+    /// And a mapping from `C` to `C_2`.
+    ///
+    /// Note that `D` is still active.
+    ///
+    /// Step 3) An active edge gets added from `D` to `A`.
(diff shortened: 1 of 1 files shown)


chain/network/src/routing/routing_table_actor.rs
```diff
@@ -177,6 +177,33 @@ impl RoutingTableActor {
     }
 
     /// If peer_id is not in memory check if it is on disk in bring it back on memory.
+    ///
+    /// Note: here an advanced example, which shows what's happening.
+    /// Let's say we have a full graph fully connected with nodes `A, B, C, D`.
+    /// Step 1 ) `A`, `B` get removed.
+    /// We store edges belonging to `A` and `B`: `<A,B>, <A,C>, <A, D>, <B, C>, <B, D>`
+    /// into component 1 let's call it `C_1`.
+    /// And mapping from `A` to `C_1`, and from `B` to `C_2`
```
mm-near, line 186: shouldn't be belong do 'C_1'?

chain/network/src/routing/routing_table_actor.rs
```diff
@@ -211,14 +238,10 @@ impl RoutingTableActor {
...
                                     peer_id.try_to_vec().unwrap().as_ref(),
                                 );
-                            } else {
-                                warn!("We expected `peer_id` to belong to component {}, but it belongs to {}",
-                                       component_nonce, cur_nonce);
                             }
-                        } else {
-                            warn!("We expected `peer_id` to belong to a component {}, but it doesn't belong anywhere",
-                                       component_nonce);
+                            // else case can happen, see explanation of this function.
```
mm-near, line 242: based on the explanation above, we don't have to print the warning if cur_nonce > component_nonce (as this is the case that you've described in explanation).

But what if cur_nonce > component_nonce ? 

let's discuss it during our meeting today.

pmnoxx: @mm-near That's completely normal. 



Case A) You add components C1, then C2.
Then you remove then in order C2, then C1.

Case B) You add components C1, then C2.
Then you remove then in order C1, then C2.

There is nothing special with `cur_nonce < component_nonce` vs  `cur_nonce > component_nonce`

pmnoxx: There are some advanced cases involving multiple cliques that I didn't cover. I think it's not worth documenting them.

pmnoxx:
```
```

pmnoxx:
```
```

mm-near (approved):

PR #5619, opened 2021-12-03
chain: rosetta: fix make Construction and Data APIs ids match

mina86:
Currently /construction/hash and /construction/submit APIs return
transaction identifiers where hash is simply the hash of the
transaction.  This does not match identifier returned by Data API
which prefixes the hashes with a ‘tx:’ (to distinguish them from
Rosetta transaction created as a result of other NEAR changes).

Introduce constructors for models::TransactionIdentifier and use them
throughout the code to make sure all places agree how the identifiers
are formatted.

### DIFF
--- a/chain/rosetta-rpc/src/models.rs
+++ b/chain/rosetta-rpc/src/models.rs
@@ -982,27 +982,27 @@ pub(crate) struct TransactionIdentifier {
 impl TransactionIdentifier {
     /// Returns an identifier for a NEAR transaction with given hash.
     pub(crate) fn transaction(tx_hash: &near_primitives::hash::CryptoHash) -> Self {
-        Self::construct("tx", tx_hash.to_base())
+        Self::from_prefix_and_hash("tx", tx_hash.to_base())
     }
 
     /// Returns an identifier for a NEAR receipt with given hash.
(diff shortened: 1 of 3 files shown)


chain/rosetta-rpc/src/models.rs
```diff
@@ -979,6 +979,34 @@ pub(crate) struct TransactionIdentifier {
...
+    pub(crate) fn receipt(receipt_hash: &near_primitives::hash::CryptoHash) -> Self {
+        Self::construct("receipt", receipt_hash.to_base())
+    }
+
+    /// Returns an identifier constructed as <prefix>:<hash>.
+    ///
+    /// Note: If constructing identifiers for transactions or receipts, use
+    /// [`transaction`] or [`receipt`] methods instead (this method will give
+    /// the wrong result).
+    pub(crate) fn new(
```
frol, line 998: This feels off after having `transaction` and `receipt` constructors, and given there is another `construct` method. I would have hard time deciding which method to use and why.

I would:
1. rename `construct` to `from_prefix_and_hash` and make it private
2. rename `new` to `block_event` or something along those lines

frol (approved):

PR #5620, opened 2021-12-03
re: Add `README.md` to `near-rate-limiter`

pmnoxx:
In this PR we add the first version of `README.md` for `near-rate-limiter`.


utils/near-rate-limiter/README.md
```diff
@@ -0,0 +1,52 @@
...
+However, neither `tokio`, nor `actix` solve the issue of load-balancing. 
+`near-rate-limiter` provides a `ThrottledFrameRead` as a replacement for `FramedRead` with a couple of extra features.
+
+## Implemented features:
+- Stopping/resuming reading from tcp socket at will.
+- `ThrottledFramedRead` can be controlled with a helper data structure `ThrottleController`.
+- For example, can be based on total memory usage, cpu, or other metrics, `ThrottleFramedRead` can stop or resume reading from socket.
+- An `ActixMessageWrapper`, wraps around `Actix` messages, can be used to track, the number, and size of all messages
+that originated from `TcpSocket`, and are still alive, and/or being transported inside `Actix` mailboxes, etc.
+The full design, needs its own separate section.
```
bowenwang1996, line 18: Might be helpful to create an issue for this so that we don't forget about it

bowenwang1996 (approved):
Looks good overall. Unrelated to the PR: should we put this crate inside `network` instead?

utils/near-rate-limiter/README.md
pmnoxx: Added TODO + https://github.com/near/nearcore/issues/5672

pmnoxx:
> Looks good overall. Unrelated to the PR: should we put this crate inside `network` instead?

That's a great question. However, if we were to do that, we all other crates would have to depend on `near-network`. Which is what we want to avoid. 

Btw, I'll be moving `ActixMessageWrapper`, which is now part of `near-network` to `near-rate-limiter`. That needs to be done to avoid the issue I mentioned above.

matklad:
We also do this

[/home/matklad/projects/nearcore/chain/indexer/src/lib.rs](https://github.com/near/nearcore/blob/960274091a2a1952b3dfca39b5d2f5572fb7f712/chain/indexer/src/lib.rs#L1)

to make sure that the docs are includes in `cargo doc` output.

PR #5623, opened 2021-12-03
doc: Add `near-network` reference level documentation `Routing table computation`

pmnoxx:

-  Routing table computation

### DIFF
--- a/docs/network.md
+++ b/docs/network.md
@@ -62,3 +62,60 @@ Responsibilities:
 `PeerManagerActor` open tcp server, which listens to incoming connection.
 It starts `RoutingTableActor`, which then starts `EdgeValidatorActor`.
 When connection incoming connection gets accepted, it starts a new `PeerActor` on its own thread.
+# 8 Routing table computation
+
+Routing table computation does a few things:
+- For each peer `B`, calculates set of peers `|C_b|`, such that each peer is on the shortest path to `B`.
+- Removing unreachable edges from memory and storing them to disk.
+- The distance is calculated as the minimum number of nodes on the path from given node `A`, to each other node 
+on the network. That is, `A` has a distance of `0` to itself. It's neighbors will have a distance of `1`.
+The neighbors of theirs neighbors will have a distance of `2`, etc.
+
+## 8.1 Step 1
+`PeerManagerActor` runs a `update_routing_table_trigger` every `UPDATE_ROUTING_TABLE_INTERVAL` seconds.
+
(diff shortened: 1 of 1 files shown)


docs/network.md
```diff
@@ -0,0 +1,55 @@
+# 8 Routing table computation
+
+Routing table computation does a few things:
+- for each peer `B`, calculates set of peers `|C_b|`, such that each peer is on the shortest path to `B`.
```
bowenwang1996, line 4:
```suggestion
- for each peer `B`, calculates set of peers `|C_b|`, such that each peer is on a shortest path to `B`.
```

bowenwang1996: Also worth documenting how "distance" is measured when talking about shortest paths.

pmnoxx: @bowenwang1996 Thanks, I added the following
```
- The distance is calculated as the minimum number of nodes on the path from given node `A`, to each other node 
on the network. That is, `A` has a distance of `0` to itself. It's neighbors will have a distance of `1`.
The neighbors of theirs neighbors will have a distance of `2`, etc.
```

docs/network.md
```diff
@@ -0,0 +1,58 @@
...
+## 8.3 Step 3
+`RoutingTableActor` sends `RoutingTableUpdateResponse` message back to `PeerManagerActor`.
+
+`PeerManagerActor` keep local copy of `edges_info`, called `local_edges_info` containing only edges adjacent to current node.
+
+- This message contains list of local edges, which `PeerManagerActor` should remove.
+- `peer_forwarding` which represent on how to route messages in the P2P network
+- `peers_to_ban` - list of peers to ban for sending us edges, which failed validation in `EdgeVerifierActor`.
+
+```rust,ignore
```
bowenwang1996, line 41: Same here

docs/network.md
```diff
@@ -0,0 +1,58 @@
...
+- The distance is calculated as the minimum number of nodes on the path from given node `A`, to each other node 
+on the network. That is, `A` has a distance of `0` to itself. It's neighbors will have a distance of `1`.
+The neighbors of theirs neighbors will have a distance of `2`, etc.
+
+## 8.1 Step 1
+`PeerManagerActor` runs a `update_routing_table_trigger` every `UPDATE_ROUTING_TABLE_INTERVAL` seconds.
+
+`RoutingTableMessages::RoutingTableUpdate` message is sent to `RoutingTableActor` to request routing table re-computation.
+
+```rust,ignore
```
bowenwang1996, line 15: why `ignore`?

bowenwang1996 (approved):

PR #5624, opened 2021-12-03
doc: Add `near-network` reference level documentation `Metrics reported to Prometheus's `

pmnoxx:

-  Metrics reported to Prometheus's 

See https://github.com/near/NEPs/pull/274 for full PR


bowenwang1996 (commented):
Do we need to have a dedicated section for prometheus counters? I think in-code documentation should suffice

pmnoxx:
> Do we need to have a dedicated section for prometheus counters? I think in-code documentation should suffice

Yes, we can close it. It's not that important.

PR #5625, opened 2021-12-03
doc: Add `near-network` reference level documentation `Message transportation layers`

pmnoxx:

-  Message transportation layers


docs/network.md
```diff
@@ -0,0 +1,55 @@
+# 10. Message transportation layers.
+
+This section describes different protocols of sending messages currently used in `Near`
+
+## 10.1 Messages between Actors.
+
+`Near` is build on `Actix`'s `actor` framework. (https://actix.rs/book/actix/sec-2-actor.html)
+Usually each actor runs on its own dedicated thread.
+Only messages implementing `actix::Message`, can be sent using between threads.
+Comment on lines +7 to +9
```
mm-near, line 10: ??

docs/network.md
```diff
@@ -0,0 +1,55 @@
+# 10. Message transportation layers.
+
+This section describes different protocols of sending messages currently used in `Near`
+
+## 10.1 Messages between Actors.
+
+`Near` is build on `Actix`'s `actor` framework. (https://actix.rs/book/actix/sec-2-actor.html)
+Usually each actor runs on its own dedicated thread.
```
mm-near, line 8: do we have cases, where we have multiple 'copies' of the actors (therefore running on a few separate threads?)

docs/network.md
```diff
@@ -0,0 +1,55 @@
...
+Only messages implementing `actix::Message`, can be sent using between threads.
```
mm-near, line 9: I'd also mention that those actors have their own queues - and that the processing of the requests is happening asynchronously.

mm-near: (but as these are all 'within' one program - we do not need to worry about backwards compatibility.

docs/network.md
```diff
@@ -0,0 +1,55 @@
...
+impl actix::Message for PeersRequest {
+    type Result = PeerRequestResult;
+}
+pub struct PeerRequestResult {
+    pub peers: Vec<PeerInfo>,
+}
+```
+
+## 10.2 Messages sent through TCP
+Near using `borsh` serialization to exchange messages between nodes (See https://borsh.io/).
```
mm-near, line 31: Near *is* using

docs/network.md
```diff
@@ -0,0 +1,55 @@
...
+Only messages implementing `BorshSerialize`, `BorshDeserialize` can be sent.
+
```
mm-near, line 33: (we also use borsh for storage)

docs/network.md
```diff
@@ -0,0 +1,55 @@
...
+Comment on lines +7 to +9
+
+ bowenwang1996 4 days ago Member
+
+We should not leak implementation details into the spec
+
+	Reply…
+Resolve conversation
+
+On example of such message is `PeersRequest` :
```
mm-near, line 19: I'd just mention, that people should look for 'impl actix::Message for XX' to see which messages can be passed such way

docs/network.md
```diff
@@ -0,0 +1,55 @@
...
+}
+```
+
+## 10.2 Messages sent through TCP
+Near using `borsh` serialization to exchange messages between nodes (See https://borsh.io/).
+Only messages implementing `BorshSerialize`, `BorshDeserialize` can be sent.
+
+Here is an example of on such message:
+```rust
+#[derive(BorshSerialize, BorshDeserialize)]
```
mm-near, line 36: this example is probably not needed (just saying that everything that has a 'derived BorshSerialize/Deserialize' can be sent over the network / written to storage)

docs/network.md
```diff
@@ -0,0 +1,55 @@
...
+pub struct SyncData {
+    pub(crate) edges: Vec<Edge>,
+    pub(crate) accounts: Vec<AnnounceAccount>,
+}
+```
+
+## 10.3 Messages sent/received through `chain/jsonrpc`
+Near runs a json rest server. (See `actix_web::HttpServer`).
```
mm-near, line 44: s/rest/REST

docs/network.md
```diff
@@ -0,0 +1,55 @@
...
+All messages sent and received must implement `serde::Serialize` and `serde::Deserialize`.
+
```
mm-near, line 46: and I'd mention that with such messages, we should keep backwards compatibility - until we are sure that customer updated/adjusted their code.

docs/network.md
mm-near: which means that we have to be VERY careful if we're changing any of those - as it is basically a protocol change.

mm-near (dismissed):

docs/network.md
pmnoxx: Yes. some of them have one thread per instance.

mm-near (approved):

PR #5626, opened 2021-12-03
doc: Add `near-network` reference level documentation `Message routing`

pmnoxx:

-  Message routing

### DIFF
--- a/docs/network.md
+++ b/docs/network.md
@@ -119,3 +119,25 @@ Those edges are then stored to disk.
 - updates local copy of`peer_forwarding`, used for routing messages.
 - removes `local_edges_to_remove` from `local_edges_info`.
 - bans peers, who sent us invalid edges.
+
+# 11. Code flow - routing a message
+This is the example of the message that is being sent between nodes (`RawRoutedMessage`) (https://github.com/near/nearcore/blob/fa8749dc60fe0de8e94c3046571731c622326e9f/chain/network-primitives/src/types.rs#L362)
+
+Each of these methods have a `target` - that is either the account_id or peer_id or hash (which seems to be used only for route back...). 
+If target is the account - it will be converted using `routing_table.account_owner` to the peer.
+
+Upon receiving the message, the `PeerManagerActor` will sign it (https://github.com/near/nearcore/blob/master/chain/network/src/peer_manager.rs#L1285)
+And convert into RoutedMessage (which also have things like TTL etc.).
+
+Then it will use the `routing_table`, to find the route to the target peer (add `route_back` if needed) and then send the message over the network as `PeerMessage::Routed`.
+Details about routing table computations are covered in section 8.
+
+When Peer receives this message (as `PeerMessage::Routed`), it will pass it to PeerManager (as `RoutedMessageFrom`), 
+which would then check if the message is for the current `PeerActor`. (if yes, it would pass it for the client) and if not - it would pass it along the network.
+
+All these messages are handled by `receive_client_message` in Peer. (`NetworkClientMessags`) - and transferred to `ClientActor` in (`chain/client/src/client_actor.rs`)
+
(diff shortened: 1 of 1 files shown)


docs/network.md
```diff
@@ -0,0 +1,19 @@
+# 11. Code flow - routing a message
+This is the example of the message that is being sent between nodes (RawRoutedMessage) (https://github.com/near/nearcore/blob/fa8749dc60fe0de8e94c3046571731c622326e9f/chain/network-primitives/src/types.rs#L362)
+
+Each of these methods have a ‘target’ - that is either the account_id or peer_id or hash (which seems to be used only for route back..). If target is the account - it will be converted using ‘routing_table.account_owner’ to the peer.
```
bowenwang1996, line 4: Yes hash is only used for routeback. Let's make that clear here.

bowenwang1996:
```suggestion
Each of these methods have a ‘target’ - that is either the account_id or peer_id or hash (which seems to be used only for route back..). If target is the account - it will be converted using `routing_table.account_owner` to the peer.
```

docs/network.md
```diff
@@ -0,0 +1,19 @@
...
+And convert into RoutedMessage (which also have things like TTL etc)
+
+Then it will use the routing_table, to find the route to the target peer (add route_back if needed) and then send the message over the network as PeerMessage::Routed.
+
+When Peer receives this message (as PeerMessage::Routed), it will pass it to PeerManager (as RoutedMessageFrom), which would then check if the message is for the current Peer (if yes, it would pass it for the client) and if not - it would pass it along the network.
+
+All these messages are handled by receive_client_message in Peer. (NetworkClientMessags) - and transferred to ClientActor in (chain/client/src/client_actor.rs)
+
+
+NetworkRequests to PeerManager actor are triggering the RawRoutedMessage
```
bowenwang1996, line 16:
```suggestion
NetworkRequests to PeerManager actor trigger RawRoutedMessages
```
Also I believe some network requests do not trigger routed messages. We may want to be more clear here.

docs/network.md
```diff
@@ -0,0 +1,19 @@
...
+
+Lib_rs (ShardsManager) has a network_adapter - coming from client’s network_adapter that comes from ClientActor that comes from start_client call that comes from start_with_config
```
bowenwang1996, line 18:
```suggestion
`lib.rs` (ShardsManager) has a network_adapter - coming from client’s network_adapter that comes from `ClientActor` that comes from start_client call that comes from start_with_config
```
Also it may make more sense to mention the network adapter in `ClientActor`. Other network adapters are passed from `ClientActor` down to inner structs.

docs/network.md
```diff
@@ -0,0 +1,20 @@
+# 11. Code flow - routing a message
+This is the example of the message that is being sent between nodes (`RawRoutedMessage`) (https://github.com/near/nearcore/blob/fa8749dc60fe0de8e94c3046571731c622326e9f/chain/network-primitives/src/types.rs#L362)
+
+Each of these methods have a `target` - that is either the account_id or peer_id or hash (which seems to be used only for route back...). 
+If target is the account - it will be converted using `routing_table.account_owner` to the peer.
+
+Upon receiving the message, the `PeerManagerActor` will sign it (https://github.com/near/nearcore/blob/master/chain/network/src/peer_manager.rs#L1285)
+And convert into RoutedMessage (which also have things like TTL etc)
+
+Then it will use the `routing_table`, to find the route to the target peer (add `route_back` if needed) and then send the message over the network as `PeerMessage::Routed`.
```
bowenwang1996, line 10: Can we add a couple sentences to explain how routing table determines the route to the target peer? If it is already mentioned in other places we can cross link it here.

bowenwang1996 (approved):

docs/network.md
pmnoxx: Ok, I added a sentence, and I mentioned that it's in section 8.

PR #5627, opened 2021-12-03
doc: Add `near-network` reference level documentation `Database`

pmnoxx:

-  Database

### DIFF
--- a/docs/network.md
+++ b/docs/network.md
@@ -67,7 +67,7 @@ When connection incoming connection gets accepted, it starts a new `PeerActor` o
 Routing table computation does a few things:
 - For each peer `B`, calculates set of peers `|C_b|`, such that each peer is on the shortest path to `B`.
 - Removing unreachable edges from memory and storing them to disk.
-- The distance is calculated as the minimum number of nodes on the path from given node `A`, to each other node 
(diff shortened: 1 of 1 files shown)


docs/network.md
```diff
@@ -0,0 +1,17 @@
+# 12  routing_table_view: RoutingTableView,
```
mm-near, line 1: is it needed?

docs/network.md
```diff
@@ -0,0 +1,17 @@
...
+
+# 12. Database
+### 12.1 Storage of deleted edges
+Everytime a group of peers becomes unreachable at the same time; We store edges belonging to
+them in components. We remove all of those edges from memory, and save them do database,
```
mm-near, line 6: s/do/to

docs/network.md
```diff
@@ -0,0 +1,17 @@
...
+If any of them were to be reachable again, we would re-add them.
+
```
mm-near, line 8: also mention why we're doing it at all (rather than just simply deleting them)

docs/network.md
```diff
@@ -0,0 +1,17 @@
...
+To store components, we have the following column in the DB.
```
mm-near, line 9: s/column/columns

mm-near (approved):

PR #5629, opened 2021-12-03
feat: Add bandwidth stats to `PeerManagerActor` pt2

pmnoxx:
Continuation of https://github.com/near/nearcore/pull/5455

We are in progress of determining, what are the safe ranges of limit thresholds to set for `ThrottleController`.

`num_messages_in_progress` is part of `ThrottleController`. It's associated with given `PeerActor`. I tracks the number of messages which are currently in progress of being processed.

I added stats, which help us track, what was the maximum value of `num_messages_in_progress` across all peers over period of 1m. This will help us setup the liimits.

### DIFF
--- a/utils/near-rate-limiter/src/framed_read.rs
+++ b/utils/near-rate-limiter/src/framed_read.rs
@@ -44,10 +44,12 @@ pin_project! {
 pub struct ThrottleController {
     /// Total count of all messages that are tracked by `RateLimiterHelper`
     num_messages_in_progress: Arc<AtomicUsize>,
+    /// Maximum value of `num_messages_in_progress` recorded since last `consume` call.
+    max_messages_in_progress: Arc<AtomicUsize>,
     /// Total size of all messages that are tracked by `RateLimiterHelper`
     total_sizeof_messages_in_progress: Arc<AtomicUsize>,
-    /// Stores amount of bandwidth used.
-    bandwidth_used: Arc<AtomicUsize>,
+    /// Keeps track of read bandwidth.
+    bandwidth_read: Arc<AtomicUsize>,
     /// Stores count of messages seen.
     msg_seen: Arc<AtomicUsize>,
     /// We have an unbounded queue of messages, that we use to wake `ThrottledRateLimiter`.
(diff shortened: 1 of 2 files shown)


utils/near-rate-limiter/src/framed_read.rs
```diff
@@ -46,8 +44,14 @@ pin_project! {
 pub struct ThrottleController {
     /// Total count of all messages that are tracked by `RateLimiterHelper`
     num_messages_in_progress: Arc<AtomicUsize>,
+    /// Maximum value of `num_messages_in_progress` recorded since last `consume` call.
+    max_record_num_messages_in_progress: Arc<AtomicUsize>,
     /// Total size of all messages that are tracked by `RateLimiterHelper`
     total_sizeof_messages_in_progress: Arc<AtomicUsize>,
+    /// Stores amount of bandwidth used.
+    bandwidth_used: Arc<AtomicUsize>,
```
mm-near, line 52: bandwidth -- is bytes sent and received?

utils/near-rate-limiter/src/framed_read.rs
```diff
@@ -368,4 +399,26 @@ mod tests {
...
+        throttle_controller.report_msg_seen();
+        assert_eq!(throttle_controller.consume_msg_seen(), 3);
+        assert_eq!(throttle_controller.consume_msg_seen(), 0);
+
+        assert_eq!(throttle_controller.consume_bandwidth_used(), 0);
+        throttle_controller.report_bandwidth_used(10);
+        throttle_controller.report_bandwidth_used(1000);
+        throttle_controller.report_bandwidth_used(5);
+        assert_eq!(throttle_controller.consume_bandwidth_used(), 1015);
+        assert_eq!(throttle_controller.consume_bandwidth_used(), 0);
```
mm-near, line 422: can you also test max messages in progress?

utils/near-rate-limiter/src/framed_read.rs
```diff
@@ -368,4 +399,26 @@ mod tests {
 
         assert_eq!(semaphore.available_permits(), 1);
     }
+
+    #[test]
```
mm-near, line 403: do we have any test for FrameReader? if not - can you add a very basic one, and see that throttler results are what we expect?

mm-near (approved):

utils/near-rate-limiter/src/framed_read.rs
pmnoxx: received.

utils/near-rate-limiter/src/framed_read.rs
pmnoxx: @mm-near I created a new PR for adding test: https://github.com/near/nearcore/pull/5758
It requires more work, I need to figure out why the test fails.

PR #5633, opened 2021-12-04
feat: Add extra logs to debug `routing::archival_node`

pmnoxx:
After adding extra logs, I see that the test fails, because of
```
   runner.push_action(check_direct_connection(0, 1));
```

The next step is to debug, the issue, why direct connection between those two nodes hasn't been created.


pmnoxx:
Ok, I figured out the issue. 

Our nodes have an upper connection limit. In practice they allow peers to connect for `100ms`. 
In that time, they have enough time to request list of peers for to connect to,

Due to the way, timings have changes, that no longer is the case. I'll work on a fix.

bowenwang1996 (approved):

PR #5634, opened 2021-12-04
re: `PeerManager`: Refactor`find_peers` 

pmnoxx:
`find_peers` is used to sample connected peers. It can be written in a more efficient way:
- avoiding cloning whole array, when less elements than max are requested
- splitting `filter_map` to `filter` and `map`.

### DIFF
--- a/chain/network/src/peer_manager/peer_store.rs
+++ b/chain/network/src/peer_manager/peer_store.rs
@@ -193,19 +193,20 @@ impl PeerStore {
 
     /// Return unconnected or peers with unknown status that we can try to connect to.
     /// Peers with unknown addresses are filtered out.
-    pub(crate) fn unconnected_peers(
+    pub(crate) fn unconnected_peer(
         &self,
         ignore_fn: impl Fn(&KnownPeerState) -> bool,
-        count: usize,
-    ) -> Vec<PeerInfo> {
+    ) -> Option<PeerInfo> {
         self.find_peers(
             |p| {
                 (p.status == KnownPeerStatus::NotConnected || p.status == KnownPeerStatus::Unknown)
                     && !ignore_fn(p)
                     && p.peer_info.addr.is_some()
             },
-            count,
+            1,
         )
+        .get(0)
+        .cloned()
     }
 
     /// Return healthy known peers up to given amount.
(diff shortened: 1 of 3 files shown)


pmnoxx:
@matklad Can you take a look?

chain/network/src/peer_manager/peer_store.rs
```diff
@@ -177,20 +177,17 @@ impl PeerStore {
         }
     }
 
-    fn find_peers<F>(&self, mut filter: F, count: u32) -> Vec<PeerInfo>
+    /// Find a random subset of peers based on filter.
```
mm-near, line 180: mention what happens if count is 0 or MAX

(btw - do we still need to keep this strange behavior for 0 ? )

chain/network/src/peer_manager/peer_store.rs
```diff
@@ -177,20 +177,17 @@ impl PeerStore {
...
+    fn find_peers<F>(&self, filter: F, count: usize) -> Vec<PeerInfo>
```
mm-near, line 181: do we have a test for this function ?

chain/network/src/peer_manager/peer_store.rs
```diff
@@ -177,20 +177,17 @@ impl PeerStore {
...
-        if count == 0 {
-            return peers;
+        let peers = self.peer_states.values().filter(filter).map(|p| &p.peer_info);
+        if count == 0 || count == usize::MAX {
+            return peers.cloned().collect();
         }
-        peers.shuffle(&mut thread_rng());
-        peers.iter().take(count as usize).cloned().collect::<Vec<_>>()
+        let peers: Vec<&PeerInfo> = peers.collect();
+        peers.choose_multiple(&mut thread_rng(), count).cloned().cloned().collect()
```
mm-near, line 190: I see that our behavior changed a little bit -- (for case when "count" > peers.len()), right ?

mm-near (approved):

chain/network/src/peer_manager/peer_store.rs
pmnoxx: @mm-near We don't. I simplified the logic, so that neither 0 nor usize::MAX is special in any way.

PR #5635, opened 2021-12-04
re: Fix `clippy` warnings in `near-store`

pmnoxx:
There are a lot of inefficiencies in `near-store`. It may be a good idea to start refactoring the code, and cutting down on many inefficiencies.

Let's start, by fixing some obvious issues.

Reduces number of `clippy` warnings from `830` to `808`.

### DIFF
--- a/core/store/src/migrations/v8_to_v9.rs
+++ b/core/store/src/migrations/v8_to_v9.rs
@@ -27,14 +27,18 @@ where
         store_update.update_refcount(column, key.as_ref(), &value, 1);
         batch_size += key.as_ref().len() + value.len() + 8;
         if batch_size > batch_size_limit {
-            store_update.commit().expect(&format!("Failed during recomputing column {:?}", column));
+            store_update
+                .commit()
+                .unwrap_or_else(|_| panic!("Failed during recomputing column {:?}", column));
             store_update = store.store_update();
             batch_size = 0;
         }
     }
 
     if batch_size > 0 {
-        store_update.commit().expect(&format!("Failed during recomputing column {:?}", column));
+        store_update
+            .commit()
+            .unwrap_or_else(|_| panic!("Failed during recomputing column {:?}", column));
     }
 }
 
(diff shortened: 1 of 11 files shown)


pmnoxx:
@matklad Can you take a look?

core/store/src/trie/split_state.rs
```diff
@@ -545,13 +545,12 @@ mod tests {
...
-                let memory_use_without_last_receipt =
-                    receipts[..receipts.len() - 1].iter().fold(0 as u64, |sum, receipt| {
-                        sum + receipt.try_to_vec().unwrap().len() as u64
-                    });
+                let total_memory_use = receipts
+                    .iter()
+                    .fold(0_u64, |sum, receipt| sum + receipt.try_to_vec().unwrap().len() as u64);
+                let memory_use_without_last_receipt = receipts[..receipts.len() - 1]
+                    .iter()
+                    .fold(0_u64, |sum, receipt| sum + receipt.try_to_vec().unwrap().len() as u64);
```
matklad, line 553: This would be better written as `sum` rather than `fold`. As a rule of thumb, never use `fold`.

matklad (approved):

core/store/src/migrations/v8_to_v9.rs
```diff
@@ -27,14 +27,18 @@ where
         store_update.update_refcount(column, key.as_ref(), &value, 1);
         batch_size += key.as_ref().len() + value.len() + 8;
         if batch_size > batch_size_limit {
-            store_update.commit().expect(&format!("Failed during recomputing column {:?}", column));
+            store_update
+                .commit()
+                .unwrap_or_else(|_| panic!("Failed during recomputing column {:?}", column));
```
mm-near, line 32: so unwrap_or_else with panic is better than expect ? (@matklad )

mm-near (approved):

core/store/src/migrations/v8_to_v9.rs
matklad: In this case, yes. The difference is that 

```rust
.expect(&format!("Failed during recomputing column {:?}", column));
```

creates a heap-allocated string even on the happy path. In contrast, 

```
.unwrap_or_else(|_| panic!("Failed during recomputing column {:?}", column));
```

only allocates on a sad path. 

THat is, `.expect("str literal")` is good, `.expect(&format!("string formatting at {}", "run time"))` is needlessly slow.

bowenwang1996 (commented):
Unrelated to this PR: there is no point fixing old migrations. We should just remove them as they won't be useful anymore.

pmnoxx:
@Longarithm Can you take a look?

PR #5636, opened 2021-12-04
change request_partial_encoded_chunk to filter peers that are up to date

mzhangmzz:
Currently, when a node requests chunk parts, it either request from a block producer or one of its peers. Right now, it doesn't check the peer height, so it's possible that this peer is behind and does not have the part, which means the node has to resend the request after 400ms. This change fix that.


chain/network-primitives/src/types.rs
```diff
@@ -338,23 +338,22 @@ pub enum PeerIdOrHash {
...
     pub only_archival: bool,
     pub account_id: Option<AccountId>,
     pub prefer_peer: bool,
-}
-
-impl AccountIdOrPeerTrackingShard {
-    pub fn from_account(shard_id: ShardId, account_id: AccountId) -> Self {
-        Self { shard_id, only_archival: false, account_id: Some(account_id), prefer_peer: false }
-    }
+    pub min_height: Option<BlockHeight>,
```
mzhangmzz, line 356: I'm not entirely sure I can add a field here without a protocol upgrade. I think this struct is only used between `ClientActor` and `PeerManagerActor`, so it should be fine?

pmnoxx: Yes, a protocol upgrade will be required. This is a part of

```rust
   /// Request chunk parts and/or receipts
   NetworkRequests:::PartialEncodedChunkRequest {
        target: AccountIdOrPeerTrackingShard,
        request: PartialEncodedChunkRequestMsg,
    },
```

pmnoxx: Actually, `AccountIdOrPeerTrackingShard` is not used as part of out protocol. `Bosh` serialization/deserializatin can be removed from that type. https://github.com/near/nearcore/pull/5638

chain/network/src/peer_manager/peer_manager_actor.rs
```diff
@@ -1699,6 +1699,10 @@ impl PeerManagerActor {
                         for (peer_id, active_peer) in self.active_peers.iter() {
                             if (active_peer.full_peer_info.chain_info.archival
                                 || !target.only_archival)
+                                && target.min_height.map_or_else(
```
pmnoxx, line 1702: nit:
```
diff --git i/chain/network/src/peer_manager/peer_manager_actor.rs w/chain/network/src/peer_manager/peer_manager_actor.rs
index a4ce019a2..4d56bbc5c 100644
--- i/chain/network/src/peer_manager/peer_manager_actor.rs
+++ w/chain/network/src/peer_manager/peer_manager_actor.rs
@@ -1699,10 +1699,9 @@ impl PeerManagerActor {
                         for (peer_id, active_peer) in self.active_peers.iter() {
                             if (active_peer.full_peer_info.chain_info.archival
                                 || !target.only_archival)
-                                && target.min_height.map_or_else(
-                                    || true,
-                                    |h| active_peer.full_peer_info.chain_info.height >= h,
-                                )
+                                && target.min_height.map_or(true, |h| {
+                                    active_peer.full_peer_info.chain_info.height >= h
+                                })
                                 && active_peer
                                     .full_peer_info
                                     .chain_info
```

chain/chunks/src/lib.rs
```diff
@@ -567,8 +597,7 @@ impl ShardsManager {
         &self,
         parent_hash: &CryptoHash,
         shard_id: ShardId,
-        request_from_archival: bool,
-    ) -> Result<AccountIdOrPeerTrackingShard, near_chain::Error> {
+    ) -> Result<Option<AccountId>, near_chain::Error> {
         let mut block_producers = vec![];
```
pmnoxx, line 601: Not related to this change, but creating `let mut block_producers = vec![];` for the purpose of getting one element seems a bit wasteful.

pmnoxx: After playing with my branch for a while
```
diff --git i/chain/chunks/src/lib.rs w/chain/chunks/src/lib.rs
index 7b46a51de..dd9f359f6 100644
--- i/chain/chunks/src/lib.rs
+++ w/chain/chunks/src/lib.rs
@@ -8,6 +8,7 @@ use cached::{Cached, SizedCache};
 use chrono::DateTime;
 use log::{debug, error, warn};
 use near_primitives::time::Utc;
+use rand::prelude::IteratorRandom;
 use rand::seq::SliceRandom;
 
 use near_chain::validate::validate_chunk_proofs;
@@ -598,26 +599,28 @@ impl ShardsManager {
         parent_hash: &CryptoHash,
         shard_id: ShardId,
     ) -> Result<Option<AccountId>, near_chain::Error> {
-        let mut block_producers = vec![];
         let epoch_id = self.runtime_adapter.get_epoch_id_from_prev_block(parent_hash).unwrap();
-        for (validator_stake, is_slashed) in
-            self.runtime_adapter.get_epoch_block_producers_ordered(&epoch_id, parent_hash)?
-        {
-            let account_id = validator_stake.take_account_id();
-            if !is_slashed
-                && self.cares_about_shard_this_or_next_epoch(
-                    Some(&account_id),
-                    &parent_hash,
-                    shard_id,
-                    false,
-                )
-                && self.me.as_ref() != Some(&account_id)
-            {
-                block_producers.push(account_id);
-            }
-        }
+        let block_producers =
+            self.runtime_adapter.get_epoch_block_producers_ordered(&epoch_id, parent_hash)?;
+        let block_producers =
+            block_producers.into_iter().filter_map(|(validator_stake, is_slashed)| {
+                let account_id = validator_stake.take_account_id();
+                if !is_slashed
+                    && self.cares_about_shard_this_or_next_epoch(
+                        Some(&account_id),
+                        &parent_hash,
+                        shard_id,
+                        false,
+                    )
+                    && self.me.as_ref() != Some(&account_id)
+                {
+                    Some(account_id)
+                } else {
+                    None
+                }
+            });
 
-        Ok(block_producers.choose(&mut rand::thread_rng()).cloned())
+        Ok(block_producers.choose(&mut rand::thread_rng()).clone())
     }
 
     fn get_tracking_shards(&self, parent_hash: &CryptoHash) -> HashSet<ShardId> {

```

pmnoxx: Btw, `get_epoch_block_producers_ordered` can be optimized as well, to not return vec and avoid allocation.

mzhangmzz: I'll make another PR for this change

mzhangmzz: https://github.com/near/nearcore/pull/5689

bowenwang1996 (commented):
What is a case where we would like `min_height` to be `None`?

pmnoxx (approved):

chain/network-primitives/src/types.rs
pmnoxx, line 356: It would be nice to have a description next to each field, what it does, that would make reading the doc much easier.

pmnoxx: The doc doesn't answer what the `min_height` field stores.

pmnoxx:
Ignore the `test`. I was testing whenever I can dismiss the review. I think we should document the `AccountIdOrPeerTrackingShard` a bit better. I still don't know what each field holds.

What is `min_height`. Minimum height of chain in what context? Maybe it's something like `min_height` at the beginning of epoch. Without, some context, reading rest of the code, I have no idea.

mzhangmzz:
> What is a case where we would like `min_height` to be `None`?

Honestly no case right now. I'll change that to be non-optional

chain/network-primitives/src/types.rs
mzhangmzz: Could you check the most updated version, I updated the doc

bowenwang1996 (approved):

mzhangmzz:
@pmnoxx @mm-near could you take a look (again) at this PR? I need one of your reviews

pmnoxx (approved):

PR #5641, opened 2021-12-04
re: Change formatting of messages in `PeerManager` to use `tracing`

pmnoxx:
We should improve readability of messages printed by `PeerManagerActor` in order to make debugging easier.
This will require more changes, but using `tracings` format of messages is a good start.

### DIFF
--- a/chain/network/src/peer_manager/peer_manager_actor.rs
+++ b/chain/network/src/peer_manager/peer_manager_actor.rs
@@ -212,8 +212,8 @@ impl PeerManagerActor {
         routing_table_addr: Addr<RoutingTableActor>,
     ) -> Result<Self, Box<dyn std::error::Error>> {
         let peer_store = PeerStore::new(store.clone(), &config.boot_nodes)?;
-        debug!(target: "network", "Found known peers: {} (boot nodes={})", peer_store.len(), config.boot_nodes.len());
-        debug!(target: "network", "Blacklist: {:?}", config.blacklist);
+        debug!(target: "network", message = "Found known peers", len = peer_store.len(), boot_noeds = config.boot_nodes.len());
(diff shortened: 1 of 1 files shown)


chain/network/src/peer_manager/peer_manager_actor.rs
```diff
@@ -218,8 +218,8 @@ impl PeerManagerActor {
         }
 
         let peer_store = PeerStore::new(store.clone(), &config.boot_nodes)?;
-        debug!(target: "network", "Found known peers: {} (boot nodes={})", peer_store.len(), config.boot_nodes.len());
-        debug!(target: "network", "Blacklist: {:?}", config.blacklist);
+        debug!(target: "network", message = "Found known peers", len = peer_store.len(), boot_noeds = config.boot_nodes.len());
```
mm-near, line 221: typo in boot_nodes

mm-near (approved):

chain/network/src/peer_manager/peer_manager_actor.rs
```diff
@@ -105,7 +105,7 @@ const MAX_MESSAGES_TOTAL_SIZE: usize = usize::MAX;
 macro_rules! unwrap_or_error(($obj: expr, $error: expr) => (match $obj {
     Ok(result) => result,
     Err(err) => {
-        error!(target: "network", "{}: {}", $error, err);
+        error!(target: "network", message = "", error = ?$error, err = ?err);
```
mm-near, line 108: where are we changing the implementation to use tracing ? (or does it happen automatically?)

pmnoxx: @mm-near We are already using `tracing` in `near-network`.

PR #5642, opened 2021-12-04
re: Move types only used by `PeerStore` into the same file

pmnoxx:
`KnownPeerState` and `KnownPeerStatus` are only used by `PeerStore`.
It makes sense to move them from `near-network-primitives` inside `peer_store.rs`.

### DIFF
--- a/chain/network-primitives/src/types.rs
+++ b/chain/network-primitives/src/types.rs
@@ -1,7 +1,6 @@
 use actix::dev::{MessageResponse, ResponseChannel};
 use actix::{Actor, Message};
 use borsh::{BorshDeserialize, BorshSerialize};
-use chrono::DateTime;
 #[cfg(feature = "deepsize_feature")]
 use deepsize::DeepSizeOf;
 use near_crypto::{KeyType, PublicKey, SecretKey, Signature};
@@ -17,10 +16,8 @@ use near_primitives::syncing::{
     EpochSyncFinalizationResponse, EpochSyncResponse, ShardStateSyncResponse,
     ShardStateSyncResponseV1,
 };
-use near_primitives::time::{Clock, Utc};
 use near_primitives::transaction::{ExecutionOutcomeWithIdAndProof, SignedTransaction};
 use near_primitives::types::{AccountId, BlockHeight, BlockReference, EpochId, ShardId};
-use near_primitives::utils::{from_timestamp, to_timestamp};
 use near_primitives::views::{FinalExecutionOutcomeView, QueryRequest, QueryResponse};
 use std::collections::{HashMap, HashSet};
 use std::fmt;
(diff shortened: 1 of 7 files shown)


chain/network-primitives/src/types.rs
```diff
@@ -648,60 +645,6 @@ impl FromStr for PatternAddr {
     }
 }
```
mm-near, line 647: I'd normally agree - but at the same time, these are BorchSerialized - which means that we persist them somewhere etc.

Wondering if maybe we should keep such kinds of types in network-primitives.. (but that requires a little bit more discussion)

pmnoxx: @matklad @mm-near Those types shouldn't be visible to the outside world. As, they are used only by `near-network`.

Proposals:
A) Move to `peer_store.rs`
B) That being said, having a separate `db_types.rs`, which is private inside `near-network` would be an alternative option. That way, we can more easily control any changes.
C) Keep as it is.

It seems to me that `A,B > C`. I'm fine with either `A` or `B`.

chain/network/src/peer_manager/db_types.rs
```diff
@@ -0,0 +1,55 @@
...
+    pub(crate) fn new(peer_info: PeerInfo) -> Self {
+        KnownPeerState {
+            peer_info,
+            status: KnownPeerStatus::Unknown,
+            first_seen: to_timestamp(Clock::utc()),
+            last_seen: to_timestamp(Clock::utc()),
+        }
+    }
+
+    pub(crate) fn last_seen(&self) -> DateTime<Utc> {
```
matklad, line 44: would be nice to get rid of chrono dependency here while we are at it, and if that's not too much bigger diff.

chain/network/src/peer_manager/db_types.rs
```diff
@@ -0,0 +1,55 @@
+use borsh::{BorshDeserialize, BorshSerialize};
+use chrono::{DateTime, Utc};
+use near_network_primitives::types::{PeerInfo, ReasonForBan};
+use near_primitives::time::Clock;
+use near_primitives::utils::{from_timestamp, to_timestamp};
+
+/// Status of the known peers.
+#[derive(BorshSerialize, BorshDeserialize, Eq, PartialEq, Debug, Clone)]
+pub enum KnownPeerStatus {
```
matklad, line 9: Can we `pub(crate)` it to make it obvious that this isn't used outside?

matklad (approved):

chain/network/src/peer_manager/db_types.rs
pmnoxx: @matklad See https://github.com/near/nearcore/pull/5556#pullrequestreview-831602436
It has been stuck in review queue for 2 weeks.

chain/network/src/peer_manager/db_types.rs
pmnoxx: @matklad It's needed.
I added the following.
```
/// `pub` is used by `test-utils` in `iter_peers_from_store`
```

Maybe, we should rewrite `iter_peers_from_store` not to export private information from this module.

pmnoxx:
```
```
@mm-near Can you take a look?

pmnoxx:
```
```
@mm-near Can you take a look?

PR #5644, opened 2021-12-04
re:  Simplify `PeerActorManager` by removing `IncomingCrutch`

pmnoxx:
We can remove `IncomingCrutch` and simplify `PeerManagerActor`. There is no reason, why we should be using `IncomingCrutch`. We can get rid of one atomic, at the same time.

Based on https://github.com/near/nearcore/pull/5645/files


chain/network/src/peer_manager/peer_manager_actor.rs
```diff
@@ -2268,7 +2240,11 @@ impl PeerManagerActor {
                 PeerManagerMessageResponse::OutboundTcpConnect(())
             }
             PeerManagerMessageRequest::InboundTcpConnect(msg) => {
-                self.handle_msg_inbound_tcp_connect(msg, ctx);
+                if self.peer_counter.load(Ordering::SeqCst)
+                    < self.config.max_num_peers as usize + LIMIT_PENDING_PEERS
+                {
+                    self.handle_msg_inbound_tcp_connect(msg, ctx);
+                }
```
mm-near, line 2247: so here we'd silently drop the connection if the limit is exceeded? might make debugging a lot harder. 

We should add some log here (and make sure that it is not printed too often)

chain/network/src/peer_manager/peer_manager_actor.rs
```diff
@@ -1503,38 +1482,32 @@ impl Actor for PeerManagerActor {
...
+                } else {
+                    return;
+                };
+
+                loop {
+                    let (conn, client_addr) =
+                        if let Ok((conn, client_addr)) = listener.accept().await {
+                            (conn, client_addr)
+                        } else {
+                            return;
```
mm-near, line 1501: so any error on the listener part will cause us to stop listening at all?

or is there some mechanism that calls 'started' again?

mm-near (changes requested):

chain/network/src/peer_manager/peer_manager_actor.rs
pmnoxx: Yes, that's how the current logic works. We can discuss changing it in another PR.

chain/network/src/peer_manager/peer_manager_actor.rs
pmnoxx: Currently, we continue on error. I changed the logic to be the same as before.

We will continue listening, and ignore all failures of `listener.accept()`. Same as before.

chain/network/src/peer_manager/peer_manager_actor.rs
```diff
@@ -1503,38 +1482,32 @@ impl Actor for PeerManagerActor {
...
-                    }));
-                    actix::fut::ready(())
-                },
-            ));
+            async move {
+                let listener = TcpListener::bind(server_addr).await.ok();
+                let listener = if let Some(listener) = listener {
+                    listener
+                } else {
+                    return;
```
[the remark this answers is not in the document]
pmnoxx: @mm-near I changed the logic to be the same as before. We will panic in case of failure.

pmnoxx:
@matklad Can you take a look? This part of code is much simpler than before, but it's good if you take a look at is as well.

chain/network/src/peer_manager/peer_manager_actor.rs
```diff
@@ -800,10 +795,10 @@ impl PeerManagerActor {
...
                     .take_while(|x| match x {
-                        Ok(_) => future::ready(true),
+                        Ok(_) => true,
                         Err(e) => {
                             warn!(target: "network", ?e, "Peer stream error");
-                            future::ready(false)
+                            false
                         }
                     })
                     .map(Result::unwrap),
```
matklad, line 804: One of those rare cases where there *should* be a better way to do this, but there isn't one currently. Opened https://github.com/tokio-rs/tokio/issues/4337.

chain/network/src/peer_manager/peer_manager_actor.rs
```diff
@@ -1508,62 +1503,38 @@ impl PeerManagerActor {
...
-                },
-            ));
+                loop {
+                    if let Ok((conn, client_addr)) = listener.accept().await {
+                        peer_manager_addr.do_send(PeerManagerMessageRequest::InboundTcpConnect(
+                            InboundTcpConnect::new(conn),
+                        ));
+                        debug!(target: "network", from = ?client_addr, "got new connection");
+                    }
+                }
```
matklad, line 1534: We don't seem to ever break out of this loop? Ie, when `accept()` returns an error, we will start busy-looping here. It's also not super clear if this handles cancellation for the happy case? How one does actually stop PeerManager?

matklad: Though it seems like pre-existing logic, so probably is out of scope for the PR.

matklad: Ok, upon further investigation, it is probably fine -- at the top-level, when we catch `ctlr_c`, we ask the System to stop, and presumably it should stop polling the future.

matklad (approved):

pmnoxx:
```
```

pmnoxx:
```
```

chain/network/src/peer_manager/peer_manager_actor.rs
matklad: btw, this was fixed upstream

pmnoxx: @matklad Thanks. We can make it even better in the future.

bowenwang1996 (commented):
Thanks!

bowenwang1996 (approved):

PR #5645, opened 2021-12-04
re: Change code to make Clions Rust Plugin not get confused

pmnoxx:
Change `PeerManager` to use `Context::<PeerManagerActor>::spawn(ctx,` instead of `ctx`.

Clion's Rust Plugin was getting confused and started printing errors.

Resolves https://github.com/near/nearcore/issues/5654


pmnoxx:
@matklad Can you take a look? Maybe you can figure out, what was causing an error in Clion's Rust Plugin here.

mm-near (approved):

PR #5651, opened 2021-12-05
`near-network`: Move network protocol files to `network_protocol.rs`

pmnoxx:
Our goal is to make `near-network` more readable.

We have a network protocol, which defines how nodes communicate between each other. However, the types which belong to the protocol and mixed up with types that do not.

By moving types related to network protocol from `near-network::types.rs`, to `network_protocol.rs`  gives us a few advantages:
-  Easier for anyone not familiar with `near-network`, which code belongs to network protocol.
-  Whenever, anyone makes changes to network protocol file, we need to review, whenever they don't break the network protocol, and whenever migration is needed.
- By splitting types into groups, that form logical connection to each other, we can make sure our network protocol doesn't change, as long as there are no changes to network protocol files. This not only applies to the `network protocol`, a similar split is also in progress for `near-jsonrpc`.

See https://github.com/near/nearcore/issues/5653


chain/network/src/network_protocol.rs
```diff
@@ -0,0 +1,424 @@
+/// Contains type that belong to the `network protocol.
+///
+/// We need to maintain backwards compatibility, all changes to this file needs to be reviews.
```
mm-near, line 3: This warning should be a LOT larger - and more scary...

WARNING ACHTUNG WARNING
WARNING ACHTUNG WARNING

Changing anything in this file might break our network protocol - please think about backwards compatibility and make sure to loop in someone from the core/network team in your PR.

chain/network/src/network_protocol.rs
```diff
@@ -0,0 +1,424 @@
+/// Contains type that belong to the `network protocol.
```
mm-near, line 1: s/type/types

And mention that any internal types should go to types.rs file

mm-near (approved):

PR #5652, opened 2021-12-05
chain: remove unnecessary ShardId redefinition

mina86:
On top of it, definition in shard_assignment module of near-epoch-manager
does not match definition in near-primitives-core (the former is usize
while the latter is u64) so this brings the two definition in sync.


pmnoxx (approved):
LGTM

mzhangmzz (approved):

PR #5655, opened 2021-12-05
`near-chain`: Fix `clippy::redundant-clone` issues.

pmnoxx:
Fix 'clippy::redundant-clone' clippy warnings in `near-chain`.

See https://github.com/near/nearcore/pull/5571

### DIFF
--- a/chain/chain/src/chain.rs
+++ b/chain/chain/src/chain.rs
@@ -348,7 +348,7 @@ impl Chain {
         let (store, state_roots) = runtime_adapter.genesis_state();
         let store = ChainStore::new(store, chain_genesis.height);
         let genesis_chunks = genesis_chunks(
-            state_roots.clone(),
+            state_roots,
             runtime_adapter.num_shards(&EpochId::default())?,
             chain_genesis.gas_limit,
             chain_genesis.height,
@@ -373,7 +373,7 @@ impl Chain {
             runtime_adapter,
             orphans: OrphanBlockPool::new(),
             blocks_with_missing_chunks: MissingChunksPool::new(),
-            genesis: genesis.clone(),
+            genesis: genesis,
             transaction_validity_period: chain_genesis.transaction_validity_period,
             epoch_length: chain_genesis.epoch_length,
             block_economics_config: BlockEconomicsConfig::from(chain_genesis),
@@ -1683,7 +1683,7 @@ impl Chain {
         );
         assert_eq!(&chunk_headers_root, sync_prev_block.header().chunk_headers_root());
 
-        let chunk = self.get_chunk_clone_from_header(&chunk_header.clone())?;
+        let chunk = self.get_chunk_clone_from_header(&chunk_header)?;
         let chunk_proof = chunk_proofs[shard_id as usize].clone();
         let block_header =
             self.get_header_on_chain_by_height(&sync_hash, chunk_header.height_included())?.clone();
--- a/chain/chain/src/store.rs
+++ b/chain/chain/src/store.rs
@@ -2978,7 +2978,7 @@ mod tests {
         let chain_genesis = ChainGenesis::test();
         let validators = vec![vec!["test1"]];
         let runtime_adapter = Arc::new(KeyValueRuntime::new_with_validators(
-            store.clone(),
+            store,
             validators
                 .into_iter()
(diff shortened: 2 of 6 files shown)


chain/chain/src/store.rs
```diff
@@ -110,9 +110,7 @@ pub trait ChainStoreAccess {
     ) -> Result<ShardChunk, Error> {
         let shard_chunk_result = self.get_chunk(&header.chunk_hash());
         match shard_chunk_result {
-            Err(_) => {
-                return Err(ErrorKind::ChunksMissing(vec![header.clone()]).into());
-            }
+            Err(_) => Err(ErrorKind::ChunksMissing(vec![header.clone()]).into()),
```
bowenwang1996, line 113: wait I don't think the transformation here is semantic preserving

pmnoxx: The result of `match shard_chunk_result {` is returned to `match shard_chunk_result {`
then to ` fn get_chunk_clone_from_header(`

pmnoxx: The result is preserved, by we can get @matklad opinion on this is you prefer.

bowenwang1996: > The result of match shard_chunk_result { is returned to match shard_chunk_result {
then to fn get_chunk_clone_from_header(

I don't think that is the case. See official rust doc here https://doc.rust-lang.org/rust-by-example/error/result/early_returns.html

pmnoxx: This is a simpler case than the one in the doc. 
@matklad Which code style should we use from the two above:
A) This is what we have in the code
```
fn multiply(first_number_str: &str, second_number_str: &str) -> Result<i32, ParseIntError> {
   match first_number_str.parse::<i32>() {
        Ok(first_number)  => first_number,
        Err(e) => return Err(e),
    };
}
```
B) This is how we can write it
```
fn multiply(first_number_str: &str, second_number_str: &str) -> Result<i32, ParseIntError> {
   match first_number_str.parse::<i32>() {
        Ok(first_number)  => first_number,
        Err(e) => e,
    };
}
```

@bowenwang1996 No that, there is `let match = `.
If you think that this change is not correct, we should open an issue with `clippy`. https://github.com/rust-lang/rust-clippy

pmnoxx: @bowenwang1996 Here is the warning from `clippy`:
```
warning: unneeded `return` statement
   --> chain/chain/src/store.rs:114:17
    |
114 |                 return Err(ErrorKind::ChunksMissing(vec![header.clone()]).into());
    |                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: remove `return`: `Err(ErrorKind::ChunksMissing(vec![header.clone()]).into())`
    |
    = note: `-W clippy::needless-return` implied by `-W clippy::all`
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#needless_return
```

pmnoxx: I removed this change from the PR, as it doesn't fix the clone() issue.

pmnoxx: This belongs to a different PR.

chain/chain/src/chain.rs
```diff
@@ -2278,53 +2278,49 @@ impl Chain {
...
-            Ok(hash.clone())
+            Ok(*hash)
+        } else if level == 0 {
+            let maybe_hash = if index >= tree_size {
+                None
+            } else {
+                Some(*self.mut_store().get_block_hash_from_ordinal(index)?)
+            };
+            tree_nodes.insert((index, level), maybe_hash);
+            Ok(maybe_hash)
```
matklad, line 2289: @pmnoxx could you carefully check that this doesn't do anything except removal of the clones?

yordanmadzhunkov (approved):

chain/chain/src/chain.rs
pmnoxx: @matklad It does now xD.

pmnoxx: @matklad Thanks, for noticing. I rewrote the change, to only fix `clippy::redundant-clone` issues.

PR #5659, opened 2021-12-05
feat: upgrade rust `1.57.0`

pmnoxx:
Rust `1.57.0` has been announced https://blog.rust-lang.org/2021/12/02/Rust-1.57.0.html

Perhaps. we should consider upgrading at some point.

Related PR's:
- #5663 
- #5674 
- #5675 
 
`1.57.0` introduces `panic! in const contexts`.

```
panic! in const contexts
With previous versions of Rust, the panic! macro was not usable in const fn and other compile-time contexts. Now, this has been stabilized. Together with the stabilization of panic!, several other standard library APIs are now usable in const, such as assert!.

This stabilization does not yet include the full formatting infrastructure, so the panic! macro must be called with either a static string (panic!("...")), or with a single &str interpolated value (panic!("{}", a)) which must be used with {} (no format specifiers or other traits).

It is expected that in the future this support will expand, but this minimal stabilization already enables straightforward compile-time assertions, for example to verify the size of a type:

const _: () = assert!(std::mem::size_of::<u64>() == 8);
const _: () = assert!(std::mem::size_of::<u8>() == 1);
```

I want to add asserts, in `near-network`, on sizes of enums, to make sure we don't increases their size too much. 
In `Rust` size of enum, is determined, by the worst variant; Adding a large enum variant can decrease the performance.

### DIFF
--- a/chain/network/src/types.rs
+++ b/chain/network/src/types.rs
@@ -105,6 +105,7 @@ impl Message for GetPeerId {
 #[cfg(feature = "test_features")]
 #[derive(MessageResponse, Debug, serde::Serialize)]
 pub struct GetPeerIdResult {
+    #[allow(unused)]
     pub(crate) peer_id: PeerId,
 }
 
--- a/chain/network/src/types.rs
+++ b/chain/network/src/types.rs
@@ -105,7 +105,6 @@ impl Message for GetPeerId {
 #[cfg(feature = "test_features")]
 #[derive(MessageResponse, Debug, serde::Serialize)]
 pub struct GetPeerIdResult {
-    #[allow(unused)]
     pub(crate) peer_id: PeerId,
 }
 
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,5 +1,10 @@
 [toolchain]
-# Please update rust-version in **/Cargo.toml files when changing channel:
-channel = "1.56.0"
+# This specified the version we use to build.
+channel = "1.57.0"
+# In order to provide provide backward compatibility, we define minimum version.
+# Current minimum version is "1.56.0", please change rust-version in `**/Cargo.toml`.
+# The minimum supported version is important for outside packages, importing packages
+# from `near/nearcore`. That means we are limited to features available in `1.56.0`.
+# However, we can still benifit from compiler optimizations, improved `cargofmt`, `clippy`, etc.
 components = [ "rustfmt" ]
 targets = [ "wasm32-unknown-unknown" ]
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,10 +1,7 @@
 [toolchain]
-# This specified the version we use to build.
+# This specifies the version of Rust we use to build.
+# Invidual crates in the project may support lower version, indicated by `rust-version` field of `Cargo.toml`.
+# The version specified below, should be at least as high as maximum `rust-version` out of all `**/Cargo.toml` files.
 channel = "1.57.0"
-# In order to provide provide backward compatibility, we define minimum version.
-# Current minimum version is "1.56.0", please change rust-version in `**/Cargo.toml`.
-# The minimum supported version is important for outside packages, importing packages
-# from `near/nearcore`. That means we are limited to features available in `1.56.0`.
-# However, we can still benifit from compiler optimizations, improved `cargofmt`, `clippy`, etc.
 components = [ "rustfmt" ]
 targets = [ "wasm32-unknown-unknown" ]
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,6 +1,6 @@
 [toolchain]
 # This specifies the version of Rust we use to build.
-# Invidual crates in the project may support lower version, indicated by `rust-version` field of `Cargo.toml`.
+# Individual crates in the workspace may support a lower version, as indicated by `rust-version` field in each crate's `Cargo.toml`.
 # The version specified below, should be at least as high as the maximum `rust-version` within the workspace.
 channel = "1.57.0"
 components = [ "rustfmt" ]
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,7 +1,7 @@
 [toolchain]
 # This specifies the version of Rust we use to build.
 # Invidual crates in the project may support lower version, indicated by `rust-version` field of `Cargo.toml`.
-# The version specified below, should be at least as high as maximum `rust-version` out of all `**/Cargo.toml` files.
+# The version specified below, should be at least as high as the maximum `rust-version` within the workspace.
 channel = "1.57.0"
 components = [ "rustfmt" ]
 targets = [ "wasm32-unknown-unknown" ]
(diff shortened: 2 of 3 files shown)


pmnoxx:
Success. Success.  http://nayduck.near.org/#/run/2278

miraclx (commented):
Is there a change that only compiles with this `rust-version` specification? If not, I'm unconvinced this is what we want. `rust-version` defines the MSRV for a crate and ideally, we should only bump this is we absolutely need to, to ensure maximum compatibility as this would unnecessarily force downstream dependents of our crates to have to upgrade too.

pmnoxx:
> Is there a change that only compiles with this `rust-version` specification? If not, I'm unconvinced this is what we want. `rust-version` defines the MSRV for a crate and ideally, we should only bump this if we absolutely need to, to ensure maximum compatibility as this would unnecessarily force downstream dependents of our crates to have to upgrade too.

There are some features from `1.57.0` we could benefit from. However, after thinking about it, I think that changing rust version should be separate, from changing minimum rust version. That way, we can support a range of versions, in this case from `1.56.0` to `1.57.0` in case someone wants to import library. At the same time, we will get benefits from improved `cargo fmt` / `cargo clippy` / `compiler optimizations` / etc.

We probably should have a separate discussion, whenever we want to bump minimum version, and then discuss the features. The purpose of this PR is just to use new version of the Rust for building.

miraclx:
@pmnoxx So first, I actually don't think all crates have to share the same `rust-version` spec. Any crate that needs to use new features and optimizations isn't restricted from doing that.

But I think, maybe the version in `rust-toolchain.toml` just has to reflect the maximum `rust-version` within the workspace.

We could maybe revisit this when we actually have a crate who's MSRV is `1.57.0`. Did you have a candidate in mind?

pmnoxx:
> @pmnoxx So first, I actually don't think all crates have to share the same `rust-version` spec. Any crate that needs to use new features and optimizations isn't restricted from doing that.
> 
> But I think, maybe the version in `rust-toolchain.toml` just has to reflect the maximum `rust-version` within the workspace.
> 
> We could maybe revisit this when we actually have a crate who's MSRV is `1.57.0`. Did you have a candidate in mind?

@pmnoxx Yes, `1.57.0` introduces `panic! in const contexts`.

```
panic! in const contexts
With previous versions of Rust, the panic! macro was not usable in const fn and other compile-time contexts. Now, this has been stabilized. Together with the stabilization of panic!, several other standard library APIs are now usable in const, such as assert!.

This stabilization does not yet include the full formatting infrastructure, so the panic! macro must be called with either a static string (panic!("...")), or with a single &str interpolated value (panic!("{}", a)) which must be used with {} (no format specifiers or other traits).

It is expected that in the future this support will expand, but this minimal stabilization already enables straightforward compile-time assertions, for example to verify the size of a type:

const _: () = assert!(std::mem::size_of::<u64>() == 8);
const _: () = assert!(std::mem::size_of::<u8>() == 1);```

I want to add asserts on sizes of enums, to make sure, we don't make some changes that increases their size;

File: chain/network/src/types.rs [hunk 1]
```diff
@@ -506,6 +506,7 @@ impl Message for GetPeerId {
 #[derive(MessageResponse, Debug)]
 #[cfg_attr(feature = "test_features", derive(serde::Serialize))]
 pub struct GetPeerIdResult {
+    #[allow(unused)]
```
frol, line 509: Seeing these I often ask myself "why?". Could you document all the new instances of `allow(unused)`?

miraclx (approved):
Yeah, that makes sense. ✔️

rust-toolchain.toml
```diff
@@ -1,5 +1,10 @@
 [toolchain]
-# Please update rust-version in **/Cargo.toml files when changing channel:
-channel = "1.56.0"
+# This specified the version we use to build.
+channel = "1.57.0"
+# In order to provide provide backward compatibility, we define minimum version.
+# Current minimum version is "1.56.0", please change rust-version in `**/Cargo.toml`.
+# The minimum supported version is important for outside packages, importing packages
+# from `near/nearcore`. That means we are limited to features available in `1.56.0`.
+# However, we can still benifit from compiler optimizations, improved `cargofmt`, `clippy`, etc.
```
frol, line 8: I feel that I would not be able to follow this comment unless I bring some prior context. Also, the comment seems to be misplaced.

miraclx, line 8: Can we reword this documentation to make it clear that individual crates can have a lower `rust-version` but never higher than this toolchain version to ensure it's actually compilable?
So whenever a crate's `rust-version` is bumped more than the current toolchain version, the toolchain version has to be updated too.

frol (commented):
@miraclx Thanks for bringing all those concerns to the table. I totally agree with those.

@pmnoxx Thanks for providing more context in the comments (ideally, we should have that context shared proactively :wink:)

frol (approved):

chain/network/src/types.rs
pmnoxx: @frol Rust, since `1.57.0` started detecting unused values with `pub(crate)` modifier.

frol: Do we know why we still need that field to be `pub(crate)` or even exist at all?

pmnoxx: @frol Yes, it's a test only feature. Take a look at https://github.com/near/nearcore/pull/5663

rust-toolchain.toml
pmnoxx: I'll rewrite this section to include this information.

pmnoxx: @frol Yes, I agree. I'll rewrite this section. I think, I should keep this PR changes to minimum, and the doc in a separate PR. To make bumping version non blocking, and independent from good documentation.

pmnoxx:
I'll update this PR, once we merge #5663,  #5675
to keep things simple.

rust-toolchain.toml
pmnoxx: @frol Can you take a another look. I rewrote the documentation based on the feedback. It should be easier to understand without prior context.

pmnoxx: @miraclx I rewrote the documentation. I think it's better now.

rust-toolchain.toml
```diff
@@ -1,5 +1,7 @@
 [toolchain]
-# Please update rust-version in **/Cargo.toml files when changing channel:
-channel = "1.56.0"
+# This specifies the version of Rust we use to build.
+# Invidual crates in the project may support lower version, indicated by `rust-version` field of `Cargo.toml`.
```
miraclx, line 3:
```suggestion
# Individual crates in the project may support a lower version, as indicated by `rust-version` field in each crate's `Cargo.toml`.
```

rust-toolchain.toml
```diff
@@ -1,5 +1,7 @@
...
+# The version specified below, should be at least as high as maximum `rust-version` out of all `**/Cargo.toml` files.
```
miraclx, line 4:
```suggestion
# The version specified below, should be at least as high as the maximum `rust-version` out of all `**/Cargo.toml` files in the workspace.
```

miraclx (approved):
Love it!

rust-toolchain.toml
miraclx: Or maybe this reads better?

```suggestion
# The version specified below, should be at least as high as the maximum `rust-version` within the workspace.
```

ailisp (approved):

matklad (approved):

Issue #5653, opened 2021-12-05
Separate `network protocol` types from rest of the code.


pmnoxx:
Our goal is to make network protocol easier to understand. 

This can be done in a few ways:
- Adding more documentation.
- Changing code structure, code refactoring.

I'm a big fan of making code readable, and self explanatory. Often, adding more documentation, which has it's uses, can be used to hide the actual problem. For example, if code is so hard to understand, that you actually to read the understand the documentation to understand the code, is not good. Often, when code is updated, the documentation may become out of data, leading to even more 
confusion and loss time.

I see that we don't have clearly defined structure of types in our code. Types, which are used in the `network protocol` are mixed with types that aren't. See `near-network:types`, `near-network-primitives:types`.

By moving types to `network_protocol` sub module, has a lot of benefits:
- There is one place to go, to see all types used in network protocol.
- By looking at `git log`, you can see, when PR modifies any part of network_protocol, and when protocol upgrade is needed
- Easy to verify: `network_protocol` submodules should only import types from other `network_protocol` submodules. That way, we can not only guarantee that the network protocol stays consistent, we could even write a tool that generates documentation automatically, based on the code structure.
- Easy to refactor: By separating `network protocol` types from internal types, any refactoring are easier to proof to be correct. For example, adding an internal field to a private structure, should be safe as long as type is not `network protocol`
- `BorshSerialize` / `BorshDeserialize` - those traits are not enough to determine whenever type belong to public protocol. We use Borsh to save private type to database for example. Btw, types saved to databases, should be defined in their own files, but that's a topic for another discussion.
- Easy to read - by splitting types into groups, it takes less time to learn the code. For example, if you want to first learn about `network protocol` types, you go to the right submodule. If you want to learn about actix messages used between crates, you should go to the right file. Etc.

Progress:
- [x] https://github.com/near/nearcore/pull/5651
- [x] https://github.com/near/nearcore/pull/5650
- [x] https://github.com/near/nearcore/issues/5653
- [ ] https://github.com/near/nearcore/pull/5944
- [ ] Move remaining `network_protocol.rs` away from `near-network`
- [ ] Create `near-network-protocol` package, move types away from `near-network-primitives`

PR #5663, opened 2021-12-05
re: Put `GetPeerIdResult` behind `test_features` flag

pmnoxx:
Rust `1.57.0` prints a warning about unused `peer_id`. It's unused, because it's only used with `GetPeerIdResult`.

The solution:
- put `GetPeerIdResult` behind `test_features` flag.
```
pub struct GetPeerIdResult {
    #[allow(unused)]
    pub(crate) peer_id: PeerId,
}
```


chain/network/src/types.rs
```diff
@@ -491,14 +491,17 @@ pub enum RegisterPeerResponse {
...
 pub struct GetPeerId {}
 
+#[cfg(feature = "test_features")]
 impl Message for GetPeerId {
     type Result = GetPeerIdResult;
 }
 
+#[cfg(feature = "test_features")]
 #[derive(MessageResponse, Debug)]
 #[cfg_attr(feature = "test_features", derive(serde::Serialize))]
```
frol, line 506: Given this structure will only be defined under `test_features`, we can flatten the `cfg_attr`:

```suggestion
#[cfg(feature = "test_features")]
#[derive(Debug, MessageResponse, serde::Serialize)]
```

P.S. Do we need `serde` here?

chain/network/src/types.rs
```diff
@@ -662,6 +666,7 @@ pub enum PeerManagerMessageResponse {
     PeerRequestResult(PeerRequestResult),
     PeersResponseResult(()),
     PeerResponse(PeerResponse),
+    #[cfg(feature = "test_features")]
     GetPeerIdResult(GetPeerIdResult),
```
frol, line 670: Does `PeerManagerMessageResponse` ever get borsh-serialized (I am not sure what `MessageResponse` derive does)? If so, we cannot skip the field since that would break the protocol due to serialization mismatch between nodes of different versions.

pmnoxx: @frol Great question! 

`PeerManagerMessageResponse` is just used internally by `PeerManagerActor`, so it's not going to get serialized.
In order to send a message between `Actors` internally.` Actix` requires a type implement `Message`, and it's reply `MessageResponse`. 

`GetPeerId` is the message we send during testing. While `GetPeerIdResult` is just a reply to that `Actix` mesasge.

Btw, I think we should separate our `network protocol` code, see: https://github.com/near/nearcore/issues/5653.
This will make it more clear.

chain/network/src/types.rs
pmnoxx: @frol Yes, we use `GetPeerIdResult` in `jsonrpc` for testing purposes.

frol (approved):

mm-near (approved):

PR #5665, opened 2021-12-05
feat: hint the position of invalidation for an Account ID

miraclx:
Merges into https://github.com/near/nearcore/pull/5660

Context: https://github.com/near/nearcore/pull/5660#discussion_r762553344

Wonder if it's useful to highlight which character and at what index was responsible for the error.. Something like;

`InvalidChar(2, 'ƒ')` from `tiƒƒany..near`
`RedundantSeparator(0, '_')` from `_root.near`

But, of course, when matching, you can discard that data with `RedundantSeparator(..)`.

- `fmt::Display` + `borsh`:
  - `invalid value: "alice.", the Account ID has a redundant separator '.' at index 5`
- `serde`:
  - `invalid value: "alice.", the Account ID has a redundant separator '.' at index 5 at line 4 column 9`

### DIFF
--- a/core/account-id/src/errors.rs
+++ b/core/account-id/src/errors.rs
@@ -1,20 +1,29 @@
 use std::fmt;
+use std::fmt::Write;
 
 /// An error which can be returned when parsing a NEAR Account ID.
 #[derive(Eq, Clone, Debug, PartialEq)]
-pub struct ParseAccountError(pub(crate) ParseErrorKind);
+pub struct ParseAccountError {
+    pub(crate) kind: ParseErrorKind,
+    pub(crate) char: Option<(usize, char)>,
+}
 
 impl ParseAccountError {
     /// Returns the specific cause why parsing the Account ID failed.
     pub fn kind(&self) -> &ParseErrorKind {
-        &self.0
+        &self.kind
     }
 }
 
 impl std::error::Error for ParseAccountError {}
 impl fmt::Display for ParseAccountError {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        fmt::Display::fmt(&self.0, f)
+        let mut buf = String::new();
+        buf.push_str(&self.kind.to_string());
+        if let Some((idx, char)) = self.char {
+            write!(buf, " {:?} at index {}", char, idx)?
+        }
+        buf.fmt(f)
     }
 }
 
@@ -38,13 +47,13 @@ pub enum ParseErrorKind {
     /// ends with or has separators immediately following each other.
     ///
     /// Cases: `jane.`, `angela__moss`, `tyrell..wellick`
-    RedundantSeparator { position: usize, separator: char },
+    RedundantSeparator,
     /// The Account ID contains an invalid character.
     ///
     /// This variant would be returned if the Account ID contains an upper-case character, non-separating symbol or space.
     ///
     /// Cases: `ƒelicia.near`, `user@app.com`, `Emily.near`.
-    InvalidChar { position: usize, character: char },
+    InvalidChar,
 }
 
 impl fmt::Display for ParseErrorKind {
@@ -52,16 +61,8 @@ impl fmt::Display for ParseErrorKind {
         match self {
             ParseErrorKind::TooLong => "the Account ID is too long".fmt(f),
             ParseErrorKind::TooShort => "the Account ID is too short".fmt(f),
-            ParseErrorKind::RedundantSeparator { position, separator } => format!(
-                "the Account ID has a redundant separator {:?} at index {}",
-                separator, position
-            )
-            .fmt(f),
-            ParseErrorKind::InvalidChar { position, character } => format!(
-                "the Account ID contains an invalid character {:?} at index {}",
-                character, position
-            )
-            .fmt(f),
+            ParseErrorKind::RedundantSeparator => "the Account ID has a redundant separator".fmt(f),
+            ParseErrorKind::InvalidChar => "the Account ID contains an invalid character".fmt(f),
         }
     }
 }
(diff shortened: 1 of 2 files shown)


core/account-id/src/errors.rs
```diff
@@ -38,24 +38,27 @@ pub enum ParseErrorKind {
...
     /// Cases: `jane.`, `angela__moss`, `tyrell..wellick`
-    RedundantSeparator,
+    RedundantSeparator(usize, char),
     /// The Account ID contains an invalid character.
     ///
     /// This variant would be returned if the Account ID contains an upper-case character, non-separating symbol or space.
     ///
     /// Cases: `ƒelicia.near`, `user@app.com`, `Emily.near`.
-    InvalidChar,
+    InvalidChar(usize, char),
```
frol, line 47: I would define these as inline structs rather tuples to be more explicit with the field names:

```
InvalidChar { invalid_char: char, position: usize }
```

P.S. let's also add a comment that position is the index of the Unicode character rather than an index in a byte array representing the account id string.

frol (approved):

matklad (commented):
I am -0 here:

* better error messages are useful!
* but I don't think that this is going to be *significantly* better
* and changing field-less enums to have data feels like a big commitment, API wise
* which is the reason why, I think, std doesn't expose such erros either https://doc.rust-lang.org/stable/std/num/enum.IntErrorKind.html

core/account-id/src/errors.rs
matklad: >P.S. let's also add a comment that position is the index of the Unicode character rather than an index in a byte array representing the account id string.

Wouldn't the two be the same? We only allow ascii, and we'll bail of the first error. 

I would say that, in the Rust API, we *should* use utf8 indices, and let the caller to convert from that to grapheme clusters or whatever else is appropriate for display. Index of unicode character is generally a wrong thing to use for any purpose.

frol: > I would say that, in the Rust API, we should use utf8 indices, and let the caller to convert from that to grapheme clusters or whatever else is appropriate for display. Index of unicode character is generally a wrong thing to use for any purpose.

Fair enough

miraclx: > Index of unicode character is generally a wrong thing to use for any purpose

To be clear, are you suggesting we drop `char ([u8; 4])` and read by bytes instead, much like before?

That would be convenient considering we only accept ASCII anyway, but that impairs the API a bit if we have this PR merged.

`InvalidChar { pos: usize, char: char }` is a lot more helpful than `InvalidChar { pos: usize, char: u8 }`

It's also not very clear what you meant by "utf8 indices".. There's `.char_indices()` for positional consistency, there's `.chars().enumerate()` for char size indifference, and there's raw bytes.

https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=d73bf43f591e42855c9b5783225f6cb1

miraclx: > I would define these as inline structs rather tuples to be more explicit with the field names:

Yeah, although I was hoping to model that after, for instance – `.enumerate()` with `(usize, item)`.

Plus, considering it's just two fields, it's not that hard to figure out. But I'll push a patch on another PR, so we can take a look and compare.

miraclx: \+ looking at semver, for instance, indexing is byte-wise but encodes the character at the invalid byte position as an actual `char` into the `ErrorKind`.

https://github.com/dtolnay/semver/blob/98163cf2fc/src/parse.rs#L177

matklad: `semver` is a very good example, thanks for bringing it up! It doesn't expose error kind at all: https://docs.rs/semver/latest/semver/struct.Error.html. This seems like a nice simplification. Do we have a concrete use-case for the consumer to inspect the kind?

matklad: >To be clear, are you suggesting we drop char ([u8; 4]) and read by bytes instead, much like before?

My original suggestion was to keep the impl as is, but document that `usize` is a byte index. That is, that, if you get a `pos` in the error when parsing `text` as AccountId, `&text[pos..]` is a valid `&str`. But now I lean towards not exposing the Kind at all.

miraclx: > Do we have a concrete use-case for the consumer to inspect the kind?

Not really, but I can imagine interfaces providing useful hints on Account ID construction.

I mean, for one, much like with `semver`: `serde` + `borsh` now helpfully report the position at which the ID was flagged to be invalid. Although it wouldn't hurt to retain that provision privately.

It's already beneficial that we have that information when `Display`-ed, it's unlikely users would need to be matching error kinds as well as position hints. So I'll drop it from the public API, but retain the hints when `Display`-ed.

matklad (approved):
>So I'll drop it from the public API, but retain the hints when Display-ed.

:+1: to sum up, I think we should do either of "public field-less enum OR private enum with bells and whistles"

miraclx:
> public field-less enum

Now, we have this:

```rust
let err = "alice.".parse::<AccountId>().unwrap_err();

println!("{}", err); // invalid value: "alice.", the Account ID has a redundant separator '.' at index 5
println!("{}", err.kind()); // invalid value: "alice.", the Account ID has a redundant separator
```

core/account-id/src/errors.rs
```diff
@@ -1,20 +1,29 @@
 use std::fmt;
+use std::fmt::Write;
 
 /// An error which can be returned when parsing a NEAR Account ID.
 #[derive(Eq, Clone, Debug, PartialEq)]
-pub struct ParseAccountError(pub(crate) ParseErrorKind);
+pub struct ParseAccountError {
+    pub(crate) kind: ParseErrorKind,
+    pub(crate) char: Option<(usize, char)>,
```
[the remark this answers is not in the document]
miraclx: Thanks. It's not on master yet though, can I get a review of https://github.com/near/nearcore/pull/5660?.

PR #5667, opened 2021-12-05
feat: `PeerManagerActor` Fix updating routing table adter adding/remove local edge

pmnoxx:
`PeerManagerActor` Fix updating routing table adter adding/remove local edge

### DIFF
--- a/chain/network/src/peer_manager/peer_manager_actor.rs
+++ b/chain/network/src/peer_manager/peer_manager_actor.rs
@@ -307,7 +307,7 @@ impl PeerManagerActor {
         self.routing_table_addr
             .send(RoutingTableMessages::RoutingTableUpdate { prune, prune_edges_not_reachable_for })
             .into_actor(self)
-            .map(|response, act, _ctx| match response {
+            .map(|response, act, ctx| match response {
                 Ok(RoutingTableMessagesResponse::RoutingTableUpdateResponse {
                     local_edges_to_remove,
                     peer_forwarding,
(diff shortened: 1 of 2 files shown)


pmnoxx:
http://nayduck.near.org/#/run/2279

pmnoxx:
```
```

pmnoxx:
```
```

chain/network/src/peer_manager/peer_manager_actor.rs
```diff
@@ -412,7 +423,7 @@ impl PeerManagerActor {
                 }
             }
             // Add new edge update to the routing table and broadcast it to peers.
-            self.add_verified_edges_to_routing_table(ctx, new_edges, true);
+            self.add_verified_edges_to_routing_table(ctx, new_edges, true, false);
```
mm-near, line 426: mention why we're not updating the routing table here.

chain/network/src/peer_manager/peer_manager_actor.rs
```diff
@@ -310,6 +313,14 @@ impl PeerManagerActor {
                             },
                         )
                     }
+                    // After we add or remove local peer, we should update routing table imediately.
+                    if update_routing_table_now {
+                        act.update_routing_table_and_prune_edges(
```
mm-near, line 318: so this issue was not caught by any tests? 

can you add a test that would cover this?

mm-near (changes requested):

chain/network/src/peer_manager/peer_manager_actor.rs
pmnoxx: It was, it was causing some random test failures.

pmnoxx: Usually, we update routing table after 1s. However, if local edge gets updated, we can do it immediately to avoid up to 1s delay.

chain/network/src/peer_manager/peer_manager_actor.rs
pmnoxx: We update routing table every 1s, so there is no need to do it immediately, after this trigger.

pmnoxx:
This PR is outdated, it need to be rewritten.

PR #5668, opened 2021-12-05
buf: Revert `commit 7d80bf000303a90` to fix tests

pmnoxx:
Many nightly tests are failing  after `commit 7d80bf000303a90599596d4c958d5dab4554e971`
master http://nayduck.near.org/#/run/2280

my branch http://nayduck.near.org/#/run/2279 -> http://nayduck.near.org/#/test/256695

The purpose of this PR is to check, whenever that commit causes the issue.


pmnoxx:
http://nayduck.near.org/#/run/2281

pmnoxx:
@mina86 What do you think we should do?

PR #5669, opened 2021-12-05
feat: Add `--regex` option to `nayduck.py`

pmnoxx:
Today, I wanted to debug `malicious_chain.py`, which is failing.

It's too  bad I can't run a command like this:
```
./scripts/nayduck.py --run-locally --regex malicious_chain.py
```

Ok, let's add `--filter <PATTERN>` to `nayduck.py`

### DIFF
--- a/scripts/nayduck.py
+++ b/scripts/nayduck.py
@@ -16,6 +16,7 @@ organisation on GitHub to authenticate (<https://github.com/orgs/near/people>).
 The source code for NayDuck itself is at <https://github.com/near/nayduck>.
 """
 
+import re
 import getpass
 import json
 import os
@@ -62,6 +63,9 @@ def _parse_args():
         '-n',
         action='store_true',
         help='Prints list of tests to execute, without doing anything')
+    parser.add_argument('--regex',
+                        '-e',
+                        help='Filter tests based on regex expression')
     args = parser.parse_args()
 
     return args
@@ -315,6 +319,9 @@ def main():
         tests = list(read_tests_from_stdin())
     else:
         tests = list(read_tests_from_file(pathlib.Path(args.test_file)))
+
+    if args.regex:
+        tests = filter(re.compile(args.regex).match, tests)
 
     if args.run_locally:
         run_locally(args, tests)


mina86:
I don’t think this is a good change. You can already do `./scripts/nayduck.py --dry-run | grep malicious_chain | ./scripts/nayduck.py -i`

pmnoxx:
> I don’t think this is a good change. You can already do `./scripts/nayduck.py --dry-run | grep malicious_chain | ./scripts/nayduck.py -i`. Not to mention that `grep malicious_chain nightly/pytest-aversarial.txt | ./scripts/nayduck.py -i` also works.

@mina86 That's a good point. When it comes to designing interfaces there is always a trade-off. Having more code, which does the same thing, vs making interface easier to use, and easier to document, and remember.

I see a few options:
Proposal A) keep the interface as it is, not changing `nayduck.py`
That means using one of the following
```
./scripts/nayduck.py --dry-run | grep malicious_chain | ./scripts/nayduck.py -i. 
```
```
grep malicious_chain nightly/pytest-aversarial.txt | ./scripts/nayduck.py -i also works.
```

Proposal B) 
Add interface like this
```
./scripts/nayduck.py --run-locally --filter malicious_chain
```

I you think proposal A) is better, then we can keep the interface as it is.

mina86:
A is better because B opens a can of worms of other possible features that one might reasonably expect. For example, why is filter a fixed string and not a regex? Should we add `-F` option? What if I want to run a sample of 30 tests? Should we add `shuf` feature? What if I want to run a malicious_chain tests ten times? Should we add `sed` feature? All of those features are already implemented and can already be used; there’s no reason for us to reimplement them in `nayduck.py`.

pmnoxx:
Ok, I'll close the PR.

pmnoxx:
@mina86 Btw, what do you think about adding `--build` option to `nayduck.py`. This options would reorder these when running locally, and build tests with appropriate flags.

mina86:
Honestly I’m not a fan of the `--run-locally` flag in the first place. To really make it work we would need to duplicate the build code from NayDuck making sure all the flags are set etc. I would prefer if we didn’t do that.

And is there actually a use case for running everything locally? The tests cannot be run in parallel (because they use the same paths and ports) so running everything locally would take ages.

pmnoxx:
@mina86 Whenever, I need to debug a test failure. I run the test locally, I can debug it much faster that way.
I also use my own scripts, that do a binary search to find when the change broke.

I've been thinking about writing my own script for working with nayduck and repo:
- able to run locally
- prints all jobs before running them.
- has filter option
- has option that runs given test multiple times, to try to debug a "flunky test"
- has a binary search option that given two commits, find where the change broke
- has a build option - can build locally the test, with the right flags
- has an option to take a link from `nayduck` to get list of tests that failed, and run those test locally, without having to run everything
- written in Rust
- don't submit job automatically by default, allow you to double check what to run before submitting
- has nice error output options, that shows what is going on, if you try to run nayduck, from not the right branch
- It will be left to people who use it to maintain it. If no one uses it, then we would delete it.
- That way we can keep `nayduck.py` as basic, and fail proof as possible

pmnoxx:
> Honestly I’m not a fan of the `--run-locally` flag in the first place. To really make it work we would need to duplicate the build code from NayDuck making sure all the flags are set etc. I would prefer if we didn’t do that.
> 
> And is there actually a use case for running everything locally? The tests cannot be run in parallel (because they use the same paths and ports) so running everything locally would take ages.

@mina86 This is mostly, for running a single test case, and trying to debug it. Or, scheduling only failed jobs to nayduck, to make iteration faster.

mina86:
My workflow for bisecting is to write a simple script `/tmp/a.sh`
which does something like:

    #!/bin/sh
    set -eux
    cargo build -p neard --features whatever,features,i,need
    cd pytest
    exec python3 test/whatever/test/i-am-checking.py

and then use `git bisect` to find the commit.  For flaky tests I can
run the test three times in the script.

I honestly think what you’re describing is over engineered.

To print jobs before running them just use `--dry-run` and if you’re
happy with what you’re seeing run the command again without that flag.
This also handles the case when you want to check what is going to be
run on NayDuck.

Filtering can be done with grep, sed etc. as described before.

Binary search can be done as described above and gives people infinite
amount of customisation of what exactly is being tested and how.

And the tool would need to handle all of those features combined in
arbitrary way.

Also, considering that the tool will mostly just run other
executables, I don’t really see a reason to write it in Rust.

I’m also not sure what you mean by ‘not the right branch’.

pmnoxx:
@mina86 Ok, no worries. In that case, let's just keep `nayduck.py` as it is. With only `--dry-run`, and `--run-locally`. I'm fine with removing `--run-locally`. I'll leave the decision up to you.

pmnoxx:
@mina86 Thanks, for suggesting regex. Now, the api is great!. I can do this:
```
./scripts/nayduck.py --regex 'pytest.*skip.*fea'
```
Finds:
```
pytest --timeout=240 sanity/skip_epoch.py --features nightly_protocol,nightly_protocol_features
```

scripts/nayduck.py
```diff
@@ -199,7 +203,7 @@ def _parse_timeout(timeout: typing.Optional[str]) -> typing.Optional[int]:
     return int(timeout) * mul
 
 
-def run_locally(tests):
+def run_locally(args, tests):
```
mina86, line 206: This argument is not used.

scripts/nayduck.py
```diff
@@ -297,8 +301,11 @@ def main():
     else:
         tests = list(read_tests_from_file(pathlib.Path(args.test_file)))
 
+    if args.regex:
+        tests = [test for test in tests if re.match(args.regex, test.strip())]
```
mina86, line 305: Alternatively, `tests = filter(re.compile(args.regex).match, tests)`.

mina86 (commented):
I remain of the opinion that this flag does not belong in this script.

pmnoxx:
> I remain of the opinion that this flag does not belong in this script.

@mina86 Yes, you could be right. Maybe it's not really needed.

I ended up writing my own script in `Rust`:
```
USAGE:
    nayduck nayduck [FLAGS] [OPTIONS] --log-level <log-level>

FLAGS:
    -b, --build                  If set, builds locally using `cargo build`
    -c, --continue-on-failure    Should continue on failure
    -h, --help                   Prints help information
        --locally                Set to run locally
        --remotely               Set to run remotely on nayduck
        --run                    Set to execute, otherwise only list of tasks / commands to execute
                                 wil be printed
    -v, --verbose                Enable for more output
    -V, --version                Prints version information

OPTIONS:
    -l, --log-level <log-level>                Minimum log level ERROR, INFO, DEBUG
    -n, --num-tests-limit <num-tests-limit>
            Takes fist `n` tests to run, that much `regex` [default: usize::MAX]

    -r, --regex <regex>                        Filter tests by regex expression
        --repeat <repeat>                      How man times to repeat [default: 1]
```

I ran it with a commend:
```
RUST_BACKTRACE=1 cargo run nayduck --regex "restaked.py" --num-tests-limit 1 --locally --build --run --repeat=10 --continue-on-failure
```

After a few minutes of waiting I got an output;
```
[2021-12-05 19:08:16] INFO: Cleaning up node 127.0.0.1:24577 on script exit
[2021-12-05 19:08:16] INFO: Executed store validity tests: 0
2021-12-06T03:08:16.306615Z  INFO nayduck::nayduck: status=Exited(1)
2021-12-06T03:08:16.306683Z  INFO nayduck::nayduck: status=Ok(Exited(1)) took=479.940876401s
2021-12-06T03:08:16.306706Z  INFO nayduck::nayduck: run=10 self.repeat=10 cmd=["tests/sanity/restaked.py"] successes=3 failures=7
2021-12-06T03:08:16.306761Z  INFO nayduck::nayduck: SUMMARY
2021-12-06T03:08:16.306775Z  INFO nayduck::nayduck: cmd=["tests/sanity/restaked.py"]
2021-12-06T03:08:16.306793Z  INFO nayduck::nayduck: success=3 failure=7
```

It looks like `"tests/sanity/restaked.py"` test is still flunky.  Anyway, maybe my script is useless, but I had fun writing it xD
Oh, well. I better get to fixing the test.

pmnoxx:
Oh, it looks like my fix works:
```

L .~/r/r/t/nayduck master (1/0/0) RUST_BACKTRACE=1 cargo run nayduck --regex "restaked.py" --num-tests-limit 1 --locally --build --run --repeat=10 --continue-on-failure

2021-12-06T03:36:15.633859Z  INFO nayduck::nayduck: status=Exited(0)
2021-12-06T03:36:15.633903Z  INFO nayduck::nayduck: status=Ok(Exited(0)) took=467.278068927s
2021-12-06T03:36:15.633915Z  INFO nayduck::nayduck: branch="piotr-fix-updating-routing-table-after-adding-edge" sha="bfe1027fc14acf6af8780b73ed97ea4dd7e32da9"
2021-12-06T03:36:15.633928Z  INFO nayduck::nayduck: run=10 self.repeat=10 cmd=["tests/sanity/restaked.py"] successes=10 failures=0
2021-12-06T03:36:15.633959Z  INFO nayduck::nayduck: SUMMARY
2021-12-06T03:36:15.633966Z  INFO nayduck::nayduck: cmd=["tests/sanity/restaked.py"]
2021-12-06T03:36:15.633983Z  INFO nayduck::nayduck: success=10 failure=0
2021-12-06T03:36:15.634000Z  INFO nayduck::nayduck: ABOUT TO RUN test="pytest --timeout=240 sanity/restaked.py"

```

pmnoxx:
I scheduled 100 runs on nayduck http://nayduck.near.org/#/run/2291 with command
```
RUST_BACKTRACE=1 cargo run nayduck --regex "restaked.py" --num-tests-limit 1 --remotely --build --run --repeat=100 --continue-on-failure

```

scripts/nayduck.py
pmnoxx: Thanks, that's a good idea.

mm-near:
What's the status of this PR ?

pmnoxx:
@mm-near The In terms of functionality, it's ready.
However, we need to decide whenever to merge it or not.

@mina86 Has a strong arguments for not merging it.

I think I should create a separate script instead, with all functionality that our team needs.

pmnoxx:
Ok, closing this issue to due to lack of interest

PR #5671, opened 2021-12-05
pytest: fix failure with tests that use LogTracker

mina86:
Commit ‘close stdout and stderr when local node is cleaned up’ removed
`stdout_name` and `stderr_name` fields from the LocalNode object but
as it turns out utils.LogTracker uses the latter.  In effect, that
change caused tests which use LogTracker to start failing at the
following:

      File "/datadrive/nearcore/pytest/lib/utils.py", line 77, in __init__
        self.fname = node.stderr_name
    AttributeError: 'LocalNode' object has no attribute 'stderr_name'

Fix the issue by bringing the two fields back to LocalNode class.


pmnoxx (approved):

mina86:
http://nayduck.near.org/#/run/2282

Issue #5672, opened 2021-12-06
Add documentation of full design of `near-rate-limiter`.


pmnoxx:
We need to add a section about full design, see `utils/near-rate-limiter/README.md`

See https://github.com/near/nearcore/pull/5620#discussion_r762627326

PR #5676, opened 2021-12-06
fe: Move `ActixMessageWrapper` from `near-network` to `near-rate-limiter`

pmnoxx:
We want to use `near-rate-limiter` across multiple crates. That implies, that we would have to use `ActixMessageWrapper` across mujltiple crates. However, `ActixMessageWrapper` currently is located in `near-network`. And we don't want other crates, to be directly importing from `near-network`.

I propose, we move `ActixMessageWrapper` to the same crate as `ThrottledFramedReader`, that is inside ` near-rate-limiter`.
Both `ActixMessageWrapper`, `ThrottledFramedReader` are meant to be used together, and it makes sense for them to be together.

Later, we may consider moving `near-rate-limiter`, to a new place. However, it's hard to say exactly where. 
Anyway, moving those two structs together, is definitely a good change. Let's do it now.

This is required for the next step. The plan will be added to https://github.com/near/nearcore/issues/5672.

### DIFF
--- a/chain/network/src/common/mod.rs
+++ b/chain/network/src/common/mod.rs
@@ -1,1 +1,0 @@
-pub(crate) mod message_wrapper;
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3003,6 +3003,7 @@ dependencies = [
 name = "near-rate-limiter"
 version = "0.0.0"
 dependencies = [
+ "actix",
  "bytes",
  "futures-core",
  "log",
--- a/chain/network/src/lib.rs
+++ b/chain/network/src/lib.rs
@@ -7,7 +7,6 @@ pub use crate::stats::metrics;
 // TODO(#5307)
 pub use near_network_primitives::types::PeerInfo;
 
-pub(crate) mod common;
 mod network_protocol;
 mod peer;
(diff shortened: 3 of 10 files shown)


matklad:
LGTM, I guess. I am not thrilled about adding actix dep to rate-limiter -- the limiter *can* be design in such a way as to leat the caller choose how to use it (which is proven by the fact that rate limiter was a separate crate). We do have several crates which want to use the message wrapper, and so we have to push it to the least common ancestor crate, and that is the rate-limiter crate. In the ideal world, we'd have some common utilities crate which depends on actix, but we don't have one now.


> We want to use near-rate-limiter across multiple crates. 

Hm, actually, which are those crates? I would imagine that we only need rate-limiter where we connect to the network, and that there is a single crate where we do that?

matklad (approved):

chain/network/src/common/mod.rs
```diff
@@ -1 +1 @@
-pub(crate) mod message_wrapper;
+
```
mm-near, line 1: do we still need the file if it is empty ?

mm-near (approved):

PR #5677, opened 2021-12-06
refactor: Keep IO and ops separately tracked inside GasCost

jakmeier:
This refactor of the fee estimator allows to print the x86 ops and I/O bytes as counted in QEMU separately.

Test plan
---------
Run the estimator and check that output contains new fields.
Example command: 
```
cargo run --release -p runtime-params-estimator --features required -- --metrics-to-measure WasmInstruction,StorageReadBase --warmup-iters 1 --docker --full
```
Expected output:
```
WasmInstruction                                            274_250 gas [                 2i 0r 0w]           (computed in 6.807445186s)
StorageReadBase                                     26_843_362_584 gas [            214665i 0r 1w] UNCERTAIN (computed in 127.336967362s)
```


runtime/runtime-params-estimator/src/testbed_runners/qemu.rs
```diff
@@ -0,0 +1,108 @@
...
+const CATCH_BASE: u32 = 0xcafebabe;
+const HYPERCALL_START_COUNTING: u32 = 0;
+const HYPERCALL_STOP_AND_GET_INSTRUCTIONS_EXECUTED: u32 = 1;
+const HYPERCALL_GET_BYTES_READ: u32 = 2;
+const HYPERCALL_GET_BYTES_WRITTEN: u32 = 3;
+
+#[derive(Clone, PartialEq, Eq)]
+pub(crate) struct QemuMeasurement {
+    // Integers are not enough because results are averaged over transactions in a block.
+    pub instructions: Ratio<u64>,
```
matklad, line 19: Let's use use just `u64` here -- it seems that this layer of abstraction should return just a raw number for single timed interval, while the logic for averaging that out over several measurements should live above it.

runtime/runtime-params-estimator/src/gas_cost.rs
```diff
@@ -18,7 +21,9 @@ pub(crate) struct GasCost {
     /// The smallest thing we are measuring is one wasm instruction, and it
     /// takes about a nanosecond, so we do need to account for fractional
     /// nanoseconds here!
-    value: Ratio<u64>,
+    time_ns: Ratio<u64>,
+    /// Value used for `GasMetric::ICount`
+    qemu_measurement: QemuMeasurement,
```
matklad, line 26: Here, I think it's best to flatten out all fields: 

```rust
struct GasCost {
    nanos: Ratio<u64>,
    instructions: Ratio<u64>,
    bytes_read: Ratio<u64>,
    bytes_written: Ratio<u64>,
}
```

runtime/runtime-params-estimator/src/gas_cost.rs
```diff
@@ -30,24 +35,40 @@ pub(crate) struct GasCost {
 }
 
 pub(crate) struct GasClock {
-    start: Consumed,
+    start: Option<Instant>,
```
matklad, line 38:
```suggestion
    start: Instant,
```

I think it's fine to just always track the time.

matklad, line 38:
```suggestion
    start: Instant,
```

I think it's fine to just always track the time.

runtime/runtime-params-estimator/src/gas_cost.rs
```diff
@@ -124,13 +178,29 @@ impl PartialOrd for GasCost {
...
+const WRITE_BYTE_COST: u64 = 47;
+
 impl GasCost {
     pub(crate) fn to_gas(&self) -> Gas {
-        ratio_to_gas(self.metric, self.value)
+        ratio_to_gas(self.metric, self.scalar_cost())
+    }
+
+    /// Reduces the detailed measurement to a single number.
+    pub(crate) fn scalar_cost(&self) -> Ratio<u64> {
```
matklad, line 195: I am somewhat surprised that we need this method, and for it to be public, I'd expect that `to_gas` would be enough.

jakmeier: Yeah this is a bit ugly right now. It's mostly to keep refactoring simple.

There are places where the code uses a single raw `u64` before applying the ratios, instead of `GasCost` or simply `Gas`. For example the `least_squares_method` in `runtime-params-estimator/src/estimator_context.rs` operates on such values and the conversion to gas happens afterwards.

I am pretty sure it could be rewritten to use `Gas` in all those places. It would also nicely lead to the removal of all calls to `ratio_to_gas` and `ratio_to_gas_signed`, as far as I can tell.

I will need to touch quite a bit more code but I think it is worth the effort.

matklad (approved):

runtime/runtime-params-estimator/src/vm_estimator.rs
```diff
@@ -105,22 +105,22 @@ pub fn cost_per_op(gas_metric: GasMetric, code: &ContractCode) -> Ratio<u64> {
     //    (measured * vm_config.regular_op_cost) /
     //       (outcome.burnt_gas * NUM_ITERATIONS),
     // as remaining can be computed with ratio_to_gas().
-    Ratio::new(measured * (vm_config.regular_op_cost as u64), NUM_ITERATIONS * outcome.burnt_gas)
+    measured * vm_config.regular_op_cost as u64 / NUM_ITERATIONS / outcome.burnt_gas
 }
 
 type CompileCost = (u64, Ratio<u64>);
```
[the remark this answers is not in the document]
jakmeier: > Can we change Ratio<u64> to Gas here?
I think yes, `Gas` makes sense here. And we should probably also use the raw instruction counter, without factoring in the IO bytes. 
I guess I haven't looked very closely while refactoring. I just made sure the function does what it did before.
But tbh, I still don't quite understand how and in what context we use this function.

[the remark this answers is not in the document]
jakmeier: > More generally, I think we might want to move ad-hoc functions from here to general infra in `lib.rs`, though I need to take a closer look to learn specifics.

Agreed, it seems cleaner. But I would argue that should go in a different PR.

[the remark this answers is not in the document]
jakmeier: Without moving any code to different files yet, I applied the changes to the function that think make sense. But it is a bit of guess work, since I don't know how we even use this.

matklad (approved):

PR #5681, opened 2021-12-06
chain: rosetta: introduce errors::Result type alias

mina86:
This saves a few kestrokes and I hear defining one’s own Result is
quite rustic.


frol (approved):
Fair enough

frol:
@mina86 run `cargo fmt` and commit to fix the failing CI

PR #5682, opened 2021-12-06
chain: rosetta: simplify TransactionIdentifier constructors

mina86:
Explicitly call `to_base` on the black hash when creating bloc event
transaction identifier to simplify and make the code consistent with
how NEAR transactions and receipts identifiers are constructed. This
does not change format of transaction identifier since CryptoHash is
Display’ed using base58.


chain/rosetta-rpc/src/models.rs
```diff
@@ -982,27 +982,26 @@ pub(crate) struct TransactionIdentifier {
 impl TransactionIdentifier {
     /// Returns an identifier for a NEAR transaction with given hash.
     pub(crate) fn transaction(tx_hash: &near_primitives::hash::CryptoHash) -> Self {
-        Self::from_prefix_and_hash("tx", tx_hash.to_base())
+        Self::from_prefix_and_hash("tx", tx_hash)
```
frol, line 985: “to_base” is explicit, and given there are just a few places, I would stick to it rather than implicitly depending on Display implementation of CryptoHash

frol:
@mina86 do yo have a strong opinion about it? Would you like to discuss?

chain/rosetta-rpc/src/models.rs
mina86: Should we use to_base in `block_event` as well then?

mina86:
I don’t. I was just confused that transactions and receipts were using `to_base` while `block_event` wasn’t. My first thought was that they were actually using different encodings (hence I even commented that `block_event` would give wrong result for transactions and receipts which is actually not true; you could do `block_event('tx', tx_hash)` and get correct result).

mina86:
@frol, PTAL

frol (approved):

PR #5687, opened 2021-12-06
documentation: 3 crates in chain (mostly to chunk handling)

mm-near:
Added documentation to 3 crates in chain - but most of it is about chunk handling.


chain/chunks/README.md
```diff
@@ -0,0 +1,30 @@
+# near-chunks
+
+This crate cotains functions to handle chunks. In NEAR - the block consists of multiple chunks - at most one per shard.
+
+When chunk is created, the creator encodes it using Reed Solomon encoding (ErasureCoding) - creating PartialChunks, that are later sent to all the validators (each validator gets a subset of them).
+
+Each validator is waiting to receive its share of PartialChunks (`has_all_parts`) - and once they are verified, it can accept the chunk (and if all chunks are accepted chain/client can sign the block).
```
mzhangmzz, line 7: I think it might be nice to differentiate between nodes and validators here because this algorithm/code is used for both non-validators and validators. For example, we could mention,

"A honest validator will only approve a block if it receives its assigned parts for all chunks in the block"
"For all nodes (validators and non-validators), they will only accept/process a block if the following requirements are satisfied: 
- for every shard it tracks, a node has to have the full chunk,
- for every shard it doesn't track, a node has have the receipts from this shard to all shards
"
Feel free to reword and choose the level of details you want to include in this README

chain/chunks/README.md
```diff
@@ -0,0 +1,30 @@
+# near-chunks
+
+This crate cotains functions to handle chunks. In NEAR - the block consists of multiple chunks - at most one per shard.
+
+When chunk is created, the creator encodes it using Reed Solomon encoding (ErasureCoding) - creating PartialChunks, that are later sent to all the validators (each validator gets a subset of them).
```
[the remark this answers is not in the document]
mzhangmzz: I think it would also be helpful to mention that `PartialEncodedChunks` is comprised of parts and receipts. A part denotes a part from the Reed Solomon encoding and they can be used to reconstruct the full chunk if a node receives enough parts. Cross-shard receipts are also sent as part of `PartialEncodedChunks`

chain/chunks/README.md
```diff
@@ -0,0 +1,30 @@
...
+If validator cares about the given shard (that is - it wants to have a whole content of the shard present in the local memory) - it will want to receive the necessary amount of PartialChunks to be able to reconstruct the whole chunk. As we use ReedSolomon, this means that they need `data_shard_count` PartialChunks (which is lower than `total_shard_count`). Afterwards, it will reconstruct the chunk and persist it in the local cache (ChunkCache).
+
+
+## ShardsManager
+Shard Manager is responsible for:
+
+* **fetching partial chunks** - it can ask other nodes for the partial chunks that it is missing. It keeps the track fo the request via RequestPool
+* **handling partial chunks requests** - when request for the partial chunk arrives, it handles reading it from store and returning to the requesting node
+
+* **storing partial chunks** - it stores partial chunks within local cache (`stored_partial_encoded_chunks`)
```
mzhangmzz, line 18: `stored_partial_encoded_chunks` is a temporary cache to store `PartialEncodedChunk`s before they can be added  to `ChunkCache`. With the existing implementation, we only merge a `PartialEncodedChunk` message in `ChunkCache` after it is verified, which requires the previous block of this chunk is processed. If the previous blocks are not processed, these `PartialEncodedChunk` messages are temporarily stored in `stored_partial_encoded_chunks`.

Also, not really affecting this PR, just fyi I plan to get rid of `stored_partial_encoded_chunks`

I would modify this section to be 
```**storing partial chunks** - it stores partial chunks within local cache before they can be used to reconstruct for the full chunk. `stored_partial_encoded_chunks` stores non-validated partial chunks while `ChunkCache` stores validated partial chunks```

File: chain/chunks/README.md [hunk 4]
```diff
@@ -0,0 +1,30 @@
...
+
+Each validator is waiting to receive its share of PartialChunks (`has_all_parts`) - and once they are verified, it can accept the chunk (and if all chunks are accepted chain/client can sign the block).
+
+If validator cares about the given shard (that is - it wants to have a whole content of the shard present in the local memory) - it will want to receive the necessary amount of PartialChunks to be able to reconstruct the whole chunk. As we use ReedSolomon, this means that they need `data_shard_count` PartialChunks (which is lower than `total_shard_count`). Afterwards, it will reconstruct the chunk and persist it in the local cache (ChunkCache).
+
+
+## ShardsManager
+Shard Manager is responsible for:
+
+* **fetching partial chunks** - it can ask other nodes for the partial chunks that it is missing. It keeps the track fo the request via RequestPool
```
mzhangmzz, line 15: It might be nice to add that `it keeps track of current pending requests and resend them periodically via RequestPool`

chain/chunks/README.md
mm-near: ack

mzhangmzz (approved):

chain/chunks/README.md
```diff
@@ -0,0 +1,37 @@
+# near-chunks
+
+This crate cotains functions to handle chunks. In NEAR - the block consists of multiple chunks - at most one per shard.
+
+When chunk is created, the creator encodes its contents using Reed Solomon encoding (ErasureCoding) and adds cross-shard receipts - creating PartialEncodedChunks, that are later sent to all the validators (each validator gets a subset of them). This is done for data availability reasons (so that we need only a part of the validators to reconstruct the whole chunk). You can read more about it in the Nightshade paper.
```
bowenwang1996, line 5:
```suggestion
When a chunk is created, the creator encodes its contents using Reed Solomon encoding (ErasureCoding) and adds cross-shard receipts - creating PartialEncodedChunks that are later sent to all the validators (each validator gets a subset of them). This is done for data availability reasons (so that we need only a part of the validators to reconstruct the whole chunk). You can read more about it in the Nightshade paper.
```
Could we link the nightshade paper here?

bowenwang1996 (commented):
It might make sense to document chunk forwarding as well, though we could do that in a separate PR.

PR #5690, opened 2021-12-06
feat: struct variants for account-id error-kinds

miraclx:
Merges into https://github.com/near/nearcore/pull/5665

Context: https://github.com/near/nearcore/pull/5665#discussion_r763005759


frol (approved):
I don't see any major downsides of having more descriptive error kinds, so I am approving this, but I am not a code owner anymore :stuck_out_tongue:

PR #5691, opened 2021-12-06
doc: restored-receipts-verifier

Longarithm:
Add README for the restored-receipts-verifier tool.
Also rename corresponding feature to make it more readable among other features.

## Testing

Existing tests.


tools/restored-receipts-verifier/README.md
```diff
@@ -0,0 +1,12 @@
...
+## Purpose
+
+Verifies that receipts restored in chain after `RestoreReceiptsAfterFixApplyChunks` protocol upgrade actually were created on chain at some heights.
+Because receipt hashes are unique, we only check for their presence.
+
+We needed such a restore, because these receipts got stuck in chain due to a bug fixed by a `FixApplyChunks` protocol upgrade. See [#4248](https://github.com/near/nearcore/pull/4248) for more details.
+
+## Requirements
+
+Mainnet archival node dump.
```
mm-near, line 12: so this was a one-time use tool?

mm-near (approved):

tools/restored-receipts-verifier/README.md
Longarithm: We need to keep it to allow anyone to verify that receipts which we "manually" inserted into chain were actually created.

Otherwise someone may complain that we allow arbitrary chain modifications.

PR #5693, opened 2021-12-07
re: PeerManagerActor Make exceptions readable when TcpSocket::bind unwrap panics

pmnoxx:
While debugging test failures I noticed that `let listener = listener.unwrap();` can sometimes panic.
Let's improve error handling in `PeerManagerActor`.

### DIFF
--- a/chain/network/src/peer_manager/peer_manager_actor.rs
+++ b/chain/network/src/peer_manager/peer_manager_actor.rs
@@ -1562,7 +1562,7 @@ impl Actor for PeerManagerActor {
 
             ctx.spawn(TcpListener::bind(server_addr).into_actor(self).then(
                 move |listener, act, ctx| {
-                    let listener = listener.unwrap();
+                    let listener = listener.unwrap_or_else(|_| panic!("Failed to PeerManagerActor at {}", server_addr));
                     let incoming = IncomingCrutch {
                         listener: tokio_stream::wrappers::TcpListenerStream::new(listener),
                     };


chain/network/src/peer_manager/peer_manager_actor.rs
```diff
@@ -1516,7 +1517,7 @@ impl Actor for PeerManagerActor {
 
             ctx.spawn(TcpListener::bind(server_addr).into_actor(self).then(
                 move |listener, act, ctx| {
-                    let listener = listener.unwrap();
+                    let listener = listener.with_context(||format!("Failed to PeerManagerActor at {}", server_addr)).unwrap();
```
pmnoxx, line 1520: @matklad Is there a way to format errors nicely, with similar syntax like with `tracing`

```
error!(message = "Failed to PeerManagerActor, server_addr)
```

frol:
We might want to consider more [fine-grained approach to error handling when it comes to lib crates](https://nick.groenen.me/posts/rust-error-handling/), so `thiserror` might be a better choice than `anyhow`, yet this PR already brings net improvement, so we might want to merge it now, and come back to fine-grained error handling later.

matklad:
If this is all we want to do in this PR, I suggest just 

```
listener.unwrap_or_else(|| panic!("Failed to PeerManagerActor at {}", server_addr));
```

there's no need for anyhow if what you want is just a nice panic message.

pmnoxx:
> If this is all we want to do in this PR, I suggest just
> 
> ```rust
> listener.unwrap_or_else(|| panic!("Failed to PeerManagerActor at {}", server_addr));
> ```
> 
> there's no need for anyhow if what you want is just a nice panic message.

Actually, I want to add error propagation in the next PR. It's useful for unit tests.

matklad:
> Actually, I want to add error propagation in the next PR. It's useful for unit tests.

I'd say that using `?` in tests is anti-pattern -- it's better if tests panic with an good error message, and you don't need anyhow for that, as it's possible to panic with format string. The reason why panic is better is that you immediately get a backtrace at the point of failure (anyhow supports backtraces only on nightly). You can also, at least in theory, configure the debugger to break on panics to speed up debugging (not sure if that actually works today -- I debug with prints :) ).

pmnoxx:
> > Actually, I want to add error propagation in the next PR. It's useful for unit tests.
> 
> I'd say that using `?` in tests is anti-pattern -- it's better if tests panic with an good error message, and you don't need anyhow for that, as it's possible to panic with format string. The reason why panic is better is that you immediately get a backtrace at the point of failure (anyhow supports backtraces only on nightly). You can also, at least in theory, configure the debugger to break on panics to speed up debugging (not sure if that actually works today -- I debug with prints :) ).
 
I didn't mean using `?` in tests. Yes, that's not a good pattern I agree.

In tat message, I meant something different, let me explain.
We have been experiencing a lot of random test failures, some of them were caused by crashes happening inside our production code.
Often, I see crashes in out production code, while running tests, it takes a bit of time to figure out what the crash actually means. My plan is to propagate 
```
let listener = listener.with_context(||format!("Failed to PeerManagerActor at {}", server_addr)).unwrap();
```

pmnoxx:
apply feedback

matklad:
I think we need to rever lockfile changes and `use anyhow::Context`, but otherwise LGTM and is a big improvemen!

chain/network/src/peer_manager/peer_manager_actor.rs
```diff
@@ -30,6 +30,7 @@ use actix::{
     Actor, ActorFuture, Addr, Arbiter, AsyncContext, Context, ContextFutureSpawner, Handler,
     Recipient, Running, StreamHandler, WrapFuture,
 };
+use anyhow::Context as _Context;
```
matklad, line 33: Note: you generally want `use anyhow::Context as _;`. That CLion bugs triggers in very rare circumstances, and I belive it was fixed after we reported it

matklad (approved):

bowenwang1996 (approved):

PR #5694, opened 2021-12-07
feat: Upgrade clap to `3.0.0-beta5`

pmnoxx:
We are using Clap to provide api for our rust binaries.

We are using `=3.0.0-beta.2` version, we should upgrade it to "=3.0.0-beta.5".


pmnoxx:
@matklad There is one small compilation issue, can you take a look:
```
                                                                                                                     
error[E0277]: the trait bound `StateViewerSubCommand: clap::Args` is not satisfied                                                                                                                                 
   --> neard/src/../src/cli.rs:107:17                                                                                                                                                                              
    |                                                                                                                                                                                                              
107 |     StateViewer(StateViewerSubCommand),                                                                                                                                                                      
    |                 ^^^^^^^^^^^^^^^^^^^^^ the trait `clap::Args` is not implemented for `StateViewerSubCommand`                                                                                                  
    |                                                                                                                                                                                                              
note: required by `augment_args`                                                                                                                                                                                   
   --> /home/pmnoxx/.cargo/registry/src/github.com-1ecc6299db9ec823/clap-3.0.0-beta.5/src/derive.rs:216:5                                                                                                          
    |                                                                                                                                                                                                              
216 |     fn augment_args(app: App<'_>) -> App<'_>;                                                                                                                                                                
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^                                                                                                                                                                
                                                              
````

**Comment by pmnoxx** [comment]
@matklad I can't get this to compile
```
use clap::Parser;

#[derive(Parser)]
pub(super) enum AAA {
    BBB(CCC)
}

#[derive(Parser)]
pub(super) enum CCC {
    DDD,
}
```

**Comment by matklad** [comment]
Something like this gets it compiling: 

```diff
diff --git a/neard/src/cli.rs b/neard/src/cli.rs
index 13d6e9b4e..50208a254 100644
--- a/neard/src/cli.rs
+++ b/neard/src/cli.rs
@@ -2,7 +2,7 @@ use super::{DEFAULT_HOME, NEARD_VERSION, NEARD_VERSION_STRING, PROTOCOL_VERSION}
 use clap::{AppSettings, Parser};
 use futures::future::FutureExt;
 use near_primitives::types::{Gas, NumSeats, NumShards};
-use near_state_viewer::StateViewerSubCommand;
+use near_state_viewer::cli::StateViewerCmd;
 use nearcore::get_store_path;
 use std::net::SocketAddr;
 use std::path::{Path, PathBuf};
@@ -59,7 +59,7 @@ impl NeardCmd {
                 fs::remove_dir_all(home_dir).expect("Removing data and config failed.");
             }
             NeardSubCommand::StateViewer(cmd) => {
-                cmd.run(&home_dir);
+                cmd.subcmd.run(&home_dir);
             }
         }
     }
@@ -104,7 +104,7 @@ pub(super) enum NeardSubCommand {
     UnsafeResetData,
     /// View DB state.
     #[clap(name = "view_state")]
-    StateViewer(StateViewerSubCommand),
+    StateViewer(StateViewerCmd),
 }
 
 #[derive(Parser)]
diff --git a/test-utils/state-viewer/src/cli.rs b/test-utils/state-viewer/src/cli.rs
index eaed87ca2..59f4e49d7 100644
--- a/test-utils/state-viewer/src/cli.rs
+++ b/test-utils/state-viewer/src/cli.rs
@@ -20,7 +20,7 @@ pub struct StateViewerCmd {
     #[clap(flatten)]
     opts: StateViewerOpts,
     #[clap(subcommand)]
-    subcmd: StateViewerSubCommand,
+    pub subcmd: StateViewerSubCommand,
 }
 
 impl StateViewerCmd {
@@ -47,6 +47,12 @@ impl StateViewerOpts {
     }
 }
 
+#[derive(Parser)]
+pub struct StateViewer {
+    #[clap(subcommand)]
+    subcmd: StateViewerSubCommand,
+}
+
 #[derive(Parser)]
 pub enum StateViewerSubCommand {
     #[clap(name = "peers")]
```

I don't think this is entirely correct patch (the home argument would get duplicated I think?), I haven't looked into why we need `StateViewerCmd` at all. 


TL;DR: don't nest enums directly, nest them via a `struct` with `subcmd`

matklad:
Note: we have two clap-related pieces of work:

* upgrade to beta.5
* remove all usages of clap 2

The seconds seems somewhat higher priority to me.

pmnoxx:
I see, `Parser` trait on `AAA`, requires `CCC` to implement `Args`.
By changing examples, I get:
```
use clap::Parser;

#[derive(Parser)]
pub enum AAA {
    BBB(CCC),
}

#[derive(clap::Args)]
pub enum CCC {
    DDD,
}
fn main() {}
```

This produces error:
```
error: `#[derive(Args)]` only supports non-tuple structs
 --> dump-analyzer/src/main.rs:8:10
  |
8 | #[derive(clap::Args)]
  |          ^^^^^^^^^^
  |
  = note: this error originates in the derive macro `clap::Args` (in Nightly builds, run with -Z macro-backtrace for more info)
```
This worked on `3.0.0-beta2`, but stopped since `3.0.0-beta3`. 
@matklad I need to find the relevant issue on https://github.com/clap-rs/clap/ to understand what's going on.

pmnoxx:
> Note: we have two clap-related pieces of work:
> 
> * upgrade to beta.5
> * remove all usages of clap 2
> 
> The seconds seems somewhat higher priority to me.

I would say that those issues are independent. Upgrading to `3.0.0` is a good idea, but clap still is in beta, and they broke the api recently, before we do any migrations, I think we should understand first, what is the current state of this library.

Is support that compile issue related to https://github.com/clap-rs/clap/discussions/2945 ?

pmnoxx:
Maybe, it's better to upgrade from `clap 2` to `clap 3.0.0-beta5` directly. Otherwise, we will have to do two migrations, which duplicate the amount of work. First to `beta2` then to `beta5`.

matklad:
There's an option to just keep using `beta2` -- I don't think it gives us a lot of direct value to proactively upgrade calp. I'd say we should only upgrade it when non-beta 3.0.0 is released, or if we find a bug in beta2. 

In contrast, getting rid of clap2 in Cargo.lock would help with compile times, which is a direct tangible benefit.

pmnoxx:
> There's an option to just keep using `beta2` -- I don't think it gives us a lot of direct value to proactively upgrade calp. I'd say we should only upgrade it when non-beta 3.0.0 is released, or if we find a bug in beta2.
> 
> In contrast, getting rid of clap2 in Cargo.lock would help with compile times, which is a direct tangible benefit.

That's a good option. By reading their code, I see they haven't implemented, in `3.0.0-beta5`, all feature which were supported `3.0.0-beta2`.  IMHO, doing the workarounds is not worth the efford.

However, having a discussion, and understanding the state of `clap` is good, regardless.

matklad:
The general state of clap has been "clap 3 is coming soon" for the last three years or so :-) 

My view of the state of the CLI parsing tools: 

* clap2 is a very mature library. It's not super convenient to use, and adds significant overhead in terms of binary size and compile times, but provides the best end-user experience. 
* `argh` and `gumdrop` are lighter on the binary size and more convenient to use (via derive proc macro), but they significantly compromize UX
* `structopt` is a hack which layers derived API on top of clap's runtime API. It's more convenient to use, but is not *super* intuitive, as, ultimately, is a hack which bends clap2 API to fit proc-macro
* clap3 is yet to be released version of clap, whose main feature is pulling `structopt` into the clap itself. As of now, it inherits some of struct opt awkwardness. Eg, we need to do [this](https://github.com/near/nearcore/blob/8cf5f52ee50d9efcbc9f5413fd0fbbb7862eb871/neard/src/main.rs#L34) because `clap` has the second lifitime parameter, which works ok for runtime-created parsers, but is awkward for derive, where it has to be `'static` (cc @epage -- I guess for clap3 it's best to stay as close to clap2 as possible, but for clap4 I'd love to see "native" prom macro API :) )
* there are pico_args, lexopt and xshell, which try to be as light as possible, at the expense of both developer and user experience. 

On ballance, it does make sense to use clap3-beta for us, but we shouldn't be trying to actively try every latest beta -- given the history of clap3 not being released for years, we shouldn't optimize for switching to a stable version in a nearest future.

pmnoxx:
Yes, it looks like staying with `3.0.0-beta2` long term, is the best choice.

epage:
> clap3 is yet to be released version of clap

FYI we are expecting to release rc0 this week.  The main thing holding back 3.0 after that is giving it time to collect any feedback (and us test-porting some clients to be involved in that)

> Eg, we need to do this because clap has the second lifitime parameter, which works ok for runtime-created parsers, but is awkward for derive, where it has to be 'static (cc @epage -- I guess for clap3 it's best to stay as close to clap2 as possible, but for clap4 I'd love to see "native" prom macro API :) )

We have added a `default_value_t` but it requires `impl Display` so we can render the default for the user.

One of my goals with clap is to make a toolbox out of it, split out the lexer, help generation, etc which will allow for experimenting with alternative arg parsers that have the full feature set of clap.  One of my goals with that is to give a native proc macro a try, where the builder isn't used.

pmnoxx:
> > clap3 is yet to be released version of clap
> 
> FYI we are expecting to release rc0 this week. The main thing holding back 3.0 after that is giving it time to collect any feedback (and us test-porting some clients to be involved in that)
> 
> > Eg, we need to do this because clap has the second lifitime parameter, which works ok for runtime-created parsers, but is awkward for derive, where it has to be 'static (cc @epage -- I guess for clap3 it's best to stay as close to clap2 as possible, but for clap4 I'd love to see "native" prom macro API :) )
> 
> We have added a `default_value_t` but it requires `impl Display` so we can render the default for the user.
> 
> One of my goals with clap is to make a toolbox out of it, split out the lexer, help generation, etc which will allow for experimenting with alternative arg parsers that have the full feature set of clap. One of my goals with that is to give a native proc macro a try, where the builder isn't used.

@epage Thanks, for doing great work on `clap`. It helps us a lot!

Issue #5649, opened 2021-12-04
Random test failure `'-p near-jsonrpc-tests --test rpc_query'`


pmnoxx:
We saw yet another random test failure: https://buildkite.com/nearprotocol/nearcore/builds/9182#abda1c38-7c83-4cbe-9a26-b67bca144c3f

```

Dec 04 13:13:02.761 DEBUG client: Some(AccountId("other")) transitions to no sync
--
  | Dec 04 13:13:02.761 DEBUG client: No peers: skip account announce
  | thread panicked while panicking. aborting.
  | error: test failed, to rerun pass '-p near-jsonrpc-tests --test rpc_query'
  |  
  | Caused by:
  | process didn't exit successfully: `/var/lib/buildkite-agent/builds/buildkite-i-03d14d7a4f367643e-1/nearprotocol/nearcore/target/debug/deps/rpc_query-c8bc7743f644c5dc` (signal: 4, SIGILL: illegal instruction)
  | 🚨 Error: The command exited with status 101
  | user command error: exit status 101


```

Issue #5628, opened 2021-12-03
bug:  `nearlib`: Random text failure `library/std/src/time.rs:292:48`


pmnoxx:
We saw yet, another, random test failure related to time going backwards.
Please, take a look at: https://buildkite.com/nearprotocol/nearcore/builds/9100#c87325c6-2262-4b2a-9dfb-6d5ab946b017

```
thread 'main' panicked at 'supplied instant is later than self', library/std/src/time.rs:292:48
--
  | stack backtrace:
  | 0:     0x5610b660d39c - std::backtrace_rs::backtrace::libunwind::trace::hd3a6cb8e32f9fc43
  | at /rustc/09c42c45858d5f3aedfa670698275303a3d19afa/library/std/src/../../backtrace/src/backtrace/libunwind.rs:90:5
  | 1:     0x5610b660d39c - std::backtrace_rs::backtrace::trace_unsynchronized::h6dcceebcd6a9f2ab
  | at /rustc/09c42c45858d5f3aedfa670698275303a3d19afa/library/std/src/../../backtrace/src/backtrace/mod.rs:66:5
  | 2:     0x5610b660d39c - std::sys_common::backtrace::_print_fmt::h4eacf4e9ad977499
  | at /rustc/09c42c45858d5f3aedfa670698275303a3d19afa/library/std/src/sys_common/backtrace.rs:67:5
  | 3:     0x5610b660d39c - <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt::h1e42c22572a7e7b4
  | at /rustc/09c42c45858d5f3aedfa670698275303a3d19afa/library/std/src/sys_common/backtrace.rs:46:22
  | 4:     0x5610b6636e4c - core::fmt::write::h072b578a8e90aa54
  | at /rustc/09c42c45858d5f3aedfa670698275303a3d19afa/library/core/src/fmt/mod.rs:1150:17
  | 5:     0x5610b6604dc5 - std::io::Write::write_fmt::hc32d3cabefff1d24
  | at /rustc/09c42c45858d5f3aedfa670698275303a3d19afa/library/std/src/io/mod.rs:1667:15
  | 6:     0x5610b660fb70 - std::sys_common::backtrace::_print::h058b85b6efc15780
  | at /rustc/09c42c45858d5f3aedfa670698275303a3d19afa/library/std/src/sys_common/backtrace.rs:49:5
[...]

Issue #5503, opened 2021-11-27
bug: test failure `./tests/sanity/restaked.py`


pmnoxx:
Test `pytest/tests/sanity/restacked.py` fails,
Build got broken somewhere between `b0a80d9e9a0ba0cbc008b05634e6a21de024c9d3` commit
and `b0a80d9e9a0ba0cbc008b05634e6a21de024c9d3~200`.

Need to write a script to automatically find the issue.

works:
```
> cd pytest
> git checkout b0a80d9e9a0ba0cbc008b05634e6a21de024c9d3~33; cargo build; python3 ./tests/sanity/restaked.py

ALL GOOD
```
doesn't
```
>cd pytest
>git checkout b0a80d9e9a0ba0cbc008b05634e6a21de024c9d3~32; cargo build; python3 ./tests/sanity/restaked.py

[2021-11-27 00:00:22] INFO: Starting restaked for /home/pmnoxx/.near/test0, rpc = 0.0.0.0:3041
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: "RpcError { error_struct: Some(HandlerError(Object({\"name\": String(\"UNKNOWN_EPOCH\")}))), code: -32000, message: \"Server error\", data: Some(String(\"Unknown Epoch\")) }"', tools/restaked/src/main.rs:86:48
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
Traceback (most recent call last):
  File "./tests/sanity/restaked.py", line 70, in <module>
    assert False, "Validator test0 is not in current validators after 5 epochs"
AssertionError: Validator test0 is not in current validators after 5 epochs
```

pmnoxx:
git log `b0a80d9e9a0ba0cbc008b05634e6a21de024c9d3~32`
```
commit 1a2d8a4c339ea9a302bdce6de9e6f68e2b098619 (HEAD)
Author: pmnoxx <piotr.mikulski@gmail.com>
Date:   Mon Oct 4 15:29:29 2021 -0700

    Move routing table exchange to another thread (#4886)
    
    In this PR we will creating copy of routing table to another thread.
```

pmnoxx:
```
[...]

PR #5696, opened 2021-12-07
`near-network` Try using `anyhow ` with `test_slot_map` test.

pmnoxx:
Usually, our tests don't handle panics well.

For example:
- https://github.com/near/nearcore/issues/5649
- https://github.com/near/nearcore/issues/5628
- https://github.com/near/nearcore/issues/5503
- ...

Let's see if it works with an example test `test_slot_map`.

### DIFF
--- a/chain/network/src/routing/ibf_peer_set.rs
+++ b/chain/network/src/routing/ibf_peer_set.rs
@@ -171,11 +171,12 @@ mod test {
     use crate::routing::ibf_peer_set::{IbfPeerSet, SlotMap};
     use crate::routing::ibf_set::IbfSet;
     use crate::test_utils::random_peer_id;
+    use anyhow::{Context, Result};
     use near_primitives::network::PeerId;
     use std::collections::HashMap;
 
     #[test]
-    fn test_slot_map() {
+    fn test_slot_map() -> Result<()> {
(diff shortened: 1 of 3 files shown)


chain/network/src/routing/ibf_peer_set.rs
```diff
@@ -185,12 +186,12 @@ mod test {
         let e2 = SimpleEdge::new(p1, p2, 3);
 
         let mut sm = SlotMap::default();
-        assert_eq!(0_u64, sm.insert(&e0).unwrap());
+        assert_eq!(0_u64, sm.insert(&e0).with_context(|| format!("failed with {:?}", &e0))?);
```
matklad, line 189: You don't need anyhow in tests, you can just supply the panic message: 

```rust
assert_eq!(Some(0_u64), sm.insert(&e0), "failed to insert {:?}", e0);
```

In general, in tests its preferable to use panics, as that's guaranteed to give you the right backtrace.

pmnoxx: @matklad Thanks, that works even better. I ready found a use for it: https://github.com/near/nearcore/pull/5711
I didn't realize, that by returning an error, the stack doesn't get printed, that's kind of sad.

PR #5698, opened 2021-12-07
docs: document structural logging style

matklad:
We stater using structured (key/value) logging that tracing provides, but there's more than one way to use it. Let's document our preferred style (the simplest one, and the one advertised by the upstream docs).

### DIFF
--- a/integration-tests/tests/network/runner.rs
+++ b/integration-tests/tests/network/runner.rs
@@ -172,7 +172,7 @@ impl StateMachine {
                           _ctx: &mut Context<WaitOrTimeoutActor>,
                           _runner| {
                         if can_write_log.swap(false, Ordering::Relaxed) == true {
-                            debug!(target: "network", message = "runner.rs: Action", num_prev_actions, action = ?action_clone);
+                            debug!(target: "network", num_prev_actions, action = ?action_clone, "runner.rs: Action");
                         }
                         let addr = info.read().unwrap().pm_addr[target].clone();
                         actix::spawn(
@@ -202,7 +202,7 @@ impl StateMachine {
                           _ctx: &mut Context<WaitOrTimeoutActor>,
                           _runner| {
                         if can_write_log.swap(false, Ordering::Relaxed) == true {
-                            debug!(target: "network", message = "runner.rs: Action", num_prev_actions, action = ?action_clone);
+                            debug!(target: "network", num_prev_actions, action = ?action_clone, "runner.rs: Action");
                         }
 
                         let addr = info.read().unwrap().pm_addr[u].clone();
@@ -230,7 +230,7 @@ impl StateMachine {
                       _ctx: &mut Context<WaitOrTimeoutActor>,
                       _runner| {
                     if can_write_log.swap(false, Ordering::Relaxed) == true {
-                        debug!(target: "network", message = "runner.rs: Action", num_prev_actions, action = ?action_clone);
+                        debug!(target: "network", num_prev_actions, action = ?action_clone, "runner.rs: Action");
                     }
 
                     let expected = expected
@@ -281,7 +281,7 @@ impl StateMachine {
                           _ctx: &mut Context<WaitOrTimeoutActor>,
                           _runner| {
                         if can_write_log.swap(false, Ordering::Relaxed) == true {
-                            debug!(target: "network", message = "runner.rs: Action", num_prev_actions, action = ?action_clone);
+                            debug!(target: "network", num_prev_actions, action = ?action_clone, "runner.rs: Action");
                         }
 
                         let expected_known: Vec<_> = known_validators
@@ -324,7 +324,7 @@ impl StateMachine {
                           _ctx: &mut Context<WaitOrTimeoutActor>,
                           _runner| {
                         if can_write_log.swap(false, Ordering::Relaxed) == true {
-                            debug!(target: "network", message = "runner.rs: Action", num_prev_actions, action = ?action_clone);
+                            debug!(target: "network", num_prev_actions, action = ?action_clone, "runner.rs: Action");
                         }
 
                         let target = info.read().unwrap().peers_info[target].id.clone();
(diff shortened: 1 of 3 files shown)


matklad:
@pmnoxx maybe I am missing something -- what was the reason we started  using  explicit `message = ` syntax?

integration-tests/tests/network/runner.rs
```diff
@@ -230,7 +230,7 @@ impl StateMachine {
                       _ctx: &mut Context<WaitOrTimeoutActor>,
                       _runner| {
                     if can_write_log.swap(false, Ordering::Relaxed) == true {
-                        debug!(target: "network", message = "runner.rs: Action", num_prev_actions, action = ?action_clone);
+                        debug!(target: "network", num_prev_actions, action = ?action_clone, "runner.rs: Action");
```
matklad, line 233: Note: we don't need to put the name of the file (`runner.rs`) into the logging call. Each logging event carries metadata, which contains full path to file:

https://docs.rs/tracing-core/0.1.21/tracing_core/event/struct.Event.html#method.metadata

If we need location info, we can adjust tracing subscriber to print file name and column.

matklad: Eg, see the Pretty format here: https://docs.rs/tracing-subscriber/latest/tracing_subscriber/fmt/index.html

```
Oct 24 12:57:29.386 fmt_pretty::yak_shave: shaving yaks
  at examples/examples/fmt/yak_shave.rs:38 on main
  in fmt_pretty::yak_shave::shaving_yaks with yaks: 3
```

pmnoxx: @matklad If we do that, we would be adding the file name to all logs, right? Usually, that's not needed, I just need it for this file. I'm grepping for `runner.rs` to find list of all actions executed during this test.

bowenwang1996:
> @pmnoxx maybe I am missing something -- what was the reason we started using explicit `message = ` syntax?

I think this was a suggestion from @nagisa

bowenwang1996 (approved):

pmnoxx:
> @pmnoxx maybe I am missing something -- what was the reason we started using explicit `message = ` syntax?

That was originally suggested by @nagisa. I thought that was a good change, and I started using it.

integration-tests/tests/network/runner.rs
pmnoxx: Just having `Oct 24 12:57:29.386 fmt_pretty::yak_shave: shaving yaks` would be enough.

At least having `` fmt_pretty::yak_shave```  is useful. Otherwise, I had to rely on the message name to find the right log.

nagisa:
The `message = ` style was in turn recommended to me by one of the authors of `tracing`. Not using `message =` requires a specific – and unintuitive – ordering of the arguments, sure, but there are a couple other reasons to use `message =` variant:

1. It naturally prevents people from mixing the key-values and formatting and the `message`s should pretty much never participate in formatting when utilizing structured events (e.g. in the example provided the threshold should _also_ be a field);
2. It isn't all that rare to have more than a single field with a string value, in which case it can become quite difficult to track what's where.

That said I don't care particularly strongly either way ^^.

matklad:
I am confused why this isn't recommendation in the docs then https://docs.rs/tracing/latest/tracing/. My heuristic here is "don't think, do what the docs suggest", but, if I think about this, I tend to agree that `message =` is better. 

Summon @davidbarsky I guess? When using string messages with tracing, is the best practice to rely on order (keys first, then message) or to specify `message = ` explicitly?

pmnoxx:
@matklad Personally, I think that "Adding message = " at the beginning is at good idea.
I prefer this coding style
```
error!(message = "Connection rejected", from = ?address, for_reason = ?reason);
```
rather than
```
error!( from = ?address, for_reason = ?reason, "Connection rejected");
```

Most likely, reason, why `message =` is needed, is for compatibility reasons. That way you can port from `log` to `tracing` without having to co code changes.
```
error!("Connection rejected {:?} because of {:?}", address, reason);
```

pmnoxx:
@matklad According to  https://docs.rs/tracing/latest/tracing/
```
Users of the log crate should note that tracing exposes a set of macros for creating Events (trace!, debug!, info!, warn!, and error!) which may be invoked with the same syntax as the similarly-named macros from the log crate. Often, the process of converting a project to use tracing can begin with a simple drop-in replacement.
```

In the examples, they show code style using `event!`
```
event!(Level::INFO, "something has happened!");
```

It seems time me like `event!`.

Personally, I like it to `event!(Level::INFO, "something has happened!");` is just as good ad `info!(message = "something has happened!");

As as site note, often we have to change imports, whenever someone add new `warn!` or `error!`, etc. `event!` minimizes number of import changes.

matklad:
>As as site note, often we have to change imports, whenever someone add new warn! or error!, etc. event! minimizes number of import changes.

My preferred way is to do `tracing::info!` -- that way you need zero imports:

https://github.com/near/nearcore/blob/21d9a31cc279a34892d49538c28382c3efc285f8/runtime/near-vm-runner/src/wasmer2_runner.rs#L371

davidbarsky:
> Summon @davidbarsky I guess? When using string messages with tracing, is the best practice to rely on order (keys first, then message) or to specify message = explicitly?

Sorry for the delay. I would (weakly) suggest the former (keys first, then message) for two reasons:
- I think explicitly adding `message = ` is a bit wordy, but this is entirely an aesthetic concern.
- If you ever want to use format arguments, it's easier with the trailing arguments.

> The message = style was in turn recommended to me by one of the authors of tracing. 

It's possible, but I don't think I ever really recall seeing strong guidance one way or the other.

matklad:
It seems that there are no super strong *direct* arguments one way or another. 

However, the argument of "do what the official docs do", though indirect, feels rather strong to me, as it makes sure that nearcore is consistent with the official way of doing things. So, let's not add `message = `. 

I've also added an explicit bit on preferring qualified `tracing::warn!` syntax.

pmnoxx:
> However, the argument of "do what the official docs do", though indirect, feels rather strong to me, as it makes sure that nearcore is consistent with the official way of doing things. So, let's not add `message = `.

I confirmed, that this change is cosmetic only. Generated output is the same. 
See https://github.com/near/near-memory-tracker/pull/59

PR #5699, opened 2021-12-07
Pass epoch length in tests setup

yordanmadzhunkov:
Fix `KeyValueRuntime` to take epoch length as parameter. It caused some bugs with epochs with different length then 5
Replace direct call of Utc::now with Clock::utc() so it can be mocked in tests

### DIFF
--- a/chain/chain/src/chain.rs
+++ b/chain/chain/src/chain.rs
@@ -5,7 +5,7 @@ use std::time::{Duration as TimeDuration, Instant};
 use borsh::BorshSerialize;
 use chrono::Duration;
 use itertools::Itertools;
-use near_primitives::time::{Clock, Utc};
+use near_primitives::time::Clock;
 use rand::rngs::StdRng;
 use rand::seq::SliceRandom;
 use rand::SeedableRng;
@@ -4221,7 +4221,7 @@ impl<'a> ChainUpdate<'a> {
         F: FnMut(ChallengeBody),
     {
         // Refuse blocks from the too distant future.
-        if header.timestamp() > Utc::now() + Duration::seconds(ACCEPTABLE_TIME_DIFFERENCE) {
+        if header.timestamp() > Clock::utc() + Duration::seconds(ACCEPTABLE_TIME_DIFFERENCE) {
             return Err(ErrorKind::InvalidBlockFutureTime(header.timestamp()).into());
         }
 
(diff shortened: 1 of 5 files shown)


chain/chain/src/chain.rs
```diff
@@ -4257,9 +4257,10 @@ impl<'a> ChainUpdate<'a> {
         let prev_header = self.get_previous_header(header)?.clone();
 
         // Check that epoch_id in the header does match epoch given previous header (only if previous header is present).
-        if &self.runtime_adapter.get_epoch_id_from_prev_block(header.prev_hash())?
-            != header.epoch_id()
-        {
+        let epoch_id_from_prev_block =
```
mm-near, line 4260: this doesn't seem to be related to the PR title. (or is it just a drive-by refactor? ;-)

chain/chain/src/tests/simple_chain.rs
```diff
@@ -36,6 +36,7 @@ fn build_chain() {
     let _mock_clock_guard = MockClockGuard::default();
     for i in 0..5 {
         Clock::add_utc(chrono::Utc.ymd(2020, 10, 1).and_hms_milli(0, 0, 3, 444 + i));
+        Clock::add_utc(chrono::Utc.ymd(2020, 10, 1).and_hms_milli(0, 0, 3, 444 + i));
```
mm-near, line 39: can you comment why are we adding it two times?

mm-near: is the second call here due to 'validate_header' call?

chain/chain/src/tests/simple_chain.rs
```diff
@@ -62,17 +63,17 @@ fn build_chain() {
...
     let count_instant = Clock::instant_call_count();
     let count_utc = Clock::utc_call_count();
-    assert_eq!(count_utc, 5);
+    assert_eq!(count_utc, 9);
     assert_eq!(count_instant, 0);
     #[cfg(feature = "nightly_protocol")]
     assert_eq!(
         chain.head().unwrap().last_block_hash,
-        CryptoHash::from_str("BNap11nsM7PEqYQertYU487g7gkCteQ8Fmee6QfyRdVQ").unwrap()
+        CryptoHash::from_str("JDNJcU8dbsY18Trzbts9gSphtQqw9C5fx1Abg3SzDdRZ").unwrap()
```
mm-near, line 71: can you explain why the header has changed?

mm-near (changes requested):

chain/chain/src/chain.rs
yordanmadzhunkov: yes, it's a drive by refactor. It was easier to debug in this version

chain/chain/src/tests/simple_chain.rs
yordanmadzhunkov: yes, the validate header consumes 4 more entries from the clock in this test scenario.

chain/chain/src/tests/simple_chain.rs
yordanmadzhunkov: I will revert them.

yordanmadzhunkov (commented):
try to merge

mm-near (approved):

PR #5700, opened 2021-12-07
Yordan/fuzzing works

yordanmadzhunkov:
Added sample fuzz_target that produces several blocks and can be used to collect coverage report. 
Added a big README.md file that show how to use fuzzing and generate coverage report. I will be happy, if at least one person can reproduce the steps in the readme.md file


bowenwang1996 (commented):
Please make sure that the title of your PR is descriptive

chain/chain/src/test_utils.rs
```diff
@@ -1400,13 +1404,84 @@ impl ChainGenesis {
...
+        let d = Duration::milliseconds(f);
+        if let Some(t) = utc.checked_add_signed(d) {
+            Clock::add_utc(utc.clone());
+            utc = t;
+        }
+    }
+}
+
+pub fn build_chain_from_random_data(data: &[u8]) {
+    if data.len() > 8 {
```
mm-near, line 1448: so if data.len is smaller than 8, the test does nothing (and 'passes')?

how about we do the assert instead? (this way if someone passes a small data package, at least they get the error)

chain/chain/src/test_utils.rs
```diff
@@ -1400,13 +1404,84 @@ impl ChainGenesis {
...
+        let f = exp.sample(&mut rng) as i64;
+        let d = Duration::milliseconds(f);
+        if let Some(t) = utc.checked_add_signed(d) {
+            Clock::add_utc(utc.clone());
+            utc = t;
+        }
+    }
+}
+
+pub fn build_chain_from_random_data(data: &[u8]) {
```
mm-near, line 1447: add comment on what this function does

mm-near: as it seems that length of the 'data' is important

chain/chain/src/test_utils.rs
```diff
@@ -1400,13 +1404,84 @@ impl ChainGenesis {
...
+    if data.len() > 8 {
+        let (a, b) = data.split_at(8);
+        let seed = u64::from_be_bytes(a.try_into().unwrap());
+        let num_blocks = b.len() as u64;
```
mm-near, line 1451: do we use contents of b somewhere?

chain/chain/src/test_utils.rs
```diff
@@ -1400,13 +1404,84 @@ impl ChainGenesis {
...
+        let _mock_clock_guard = MockClockGuard::default();
+        fill_mock_clock_utc(
+            chrono::Utc.ymd(2020, 10, 1).and_hms_milli(0, 0, 3, 444),
+            num_blocks * 2 + 1,
+            100.0,
```
mm-near, line 1456: this is the 'average interval' between blocks, right? I'd have expected it to be 1 second - or did you select 0.1 seconds on purpose?

chain/chain/src/test_utils.rs
```diff
@@ -1400,13 +1404,84 @@ impl ChainGenesis {
...
+            let block = Block::empty(&prev, &*signer);
+            if let Some(tip) = process_block_test(&mut chain, &None, block).unwrap() {
+                assert_eq!(tip.height, i + 1);
+            }
+        }
+        assert_eq!(chain.head().unwrap().height, num_blocks);
+        let count_instant = Clock::instant_call_count();
+        let count_utc = Clock::utc_call_count();
+        assert_eq!(count_utc, 2 * num_blocks + 1);
+        assert_eq!(count_instant, 0);
```
mm-near, line 1474: btw - as this test is deterministic now - can we actually check that hash of the header is equal to some constant?

chain/chain/fuzz/fuzz_targets/fuzz_build_chain.rs
```diff
@@ -0,0 +1,58 @@
...
+use std::str::FromStr;
+
+use near_chain::{Block, Error,Chain, Provenance};
+use near_chain::test_utils::{setup, build_chain_from_random_data};
+use near_primitives::time::{Clock, MockClockGuard};
+
+
+fuzz_target!(|data: &[u8]| {
+    build_chain_from_random_data(data);
+    /*
```
mm-near, line 14: should this code be commented out?

mm-near (changes requested):

chain/chain/src/test_utils.rs
yordanmadzhunkov: yes, I tried that. The last block hash can be passed as an argument. Anyway, I don't think it's a good idea to do it, for these specific cases.

mzhangmzz:
Closing this because no one is actively working on this

PR #5703, opened 2021-12-07
feat: `near` Print an error if the number of `peers` specified in config is higher than allowed.

pmnoxx:
`neard` binary reads a `client config` which contains `max_num_peers` peers parameter.
Internally `near-network` disallows having higher number of peers than `128`. 

It's a good idea to add this error to `CLI` directly in `verify()` method of `NetworkConfig`.


chain/network-primitives/src/types.rs
```diff
@@ -598,6 +600,13 @@ impl NetworkConfig {
             );
         }
 
+        if self.max_num_peers as usize >= MAX_NUM_PEERS {
+            error!(target: "network",
```
mm-near, line 604: after you change it to bail -> make sure to add a test.

mm-near (approved):

PR #5705, opened 2021-12-07
feat: Don't hide errors with parsing `network config`.

pmnoxx:
`client config` contains a sections for `near-network` called `NearNetworkConfig`.

In function `validate()` of `NearNetworkConfig` we verify all argument. However, if any argument is invalid, we will print an error an continue running an normal. 

The proper solution is to, throw an error, and make let the caller know, that the `network_config` is invalid. Btw, we don't check whenever the number of peers allowed is to high. See https://github.com/near/nearcore/pull/5703

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2995,6 +2995,7 @@ name = "near-network-primitives"
 version = "0.0.0"
 dependencies = [
  "actix",
+ "anyhow",
  "borsh 0.9.1",
  "chrono",
  "deepsize",
--- a/chain/indexer/src/lib.rs
+++ b/chain/indexer/src/lib.rs
@@ -1,8 +1,8 @@
 #![doc = include_str!("../README.md")]
 
+use anyhow::Context;
 use tokio::sync::mpsc;
 
-use anyhow::Result;
 pub use near_primitives;
 use near_primitives::types::Gas;
 pub use nearcore::{get_default_home, init_configs, NearConfig};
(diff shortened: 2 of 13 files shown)


chain/network-primitives/src/types.rs
```diff
@@ -584,15 +585,17 @@ impl NetworkConfig {
         }
     }
 
-    pub fn verify(&self) {
+    pub fn verify(&self) -> Result<()> {
```
mm-near, line 588: can you start a binary with mainnet/testnet config to see that these things don't crash ?

(or can you check the logs that none of these errors are printed?)

chain/network-primitives/src/types.rs
```diff
@@ -603,20 +606,20 @@ impl NetworkConfig {
         }
 
         if self.safe_set_size <= self.minimum_outbound_peers {
-            error!(target: "network",
+            bail!(
```
mm-near, line 609: can you add a few tests to  test these conditiions where we bail now ?

mm-near (dismissed):

chain/network-primitives/src/types.rs
pmnoxx: @mm-near Yes, I started testnet.

pmnoxx:
```
@mm-near: chain/network-primitives/ nearcore/ neard/
@khorolets: chain/indexer/ tools/indexer/
```
@khorolets @mm-near Can you take a look?

khorolets (approved):

chain/network-primitives/src/types.rs
mm-near: did you add the test?

mm-near (dismissed):

chain/network-primitives/src/types.rs
pmnoxx: I added a test last month. I rebased the PR.

PR #5710, opened 2021-12-08
feat: Reduce CI build time to 12m.

pmnoxx:
Reduce CI build time, by splitting `integration-tests` and the rest.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -1,18 +1,38 @@
 steps:
-  - label: "cargo test"
+  - label: "cargo test integration-tests*"
     command: |
       source ~/.cargo/env && set -eux
-      RUST_BACKTRACE=1 RUST_LOG=WARN cargo test --locked --workspace
+      RUST_BACKTRACE=1 RUST_LOG=WARN cargo test --locked -p 'integration-tests' -j 16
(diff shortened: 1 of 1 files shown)


chefsale (approved):

.buildkite/pipeline.yml
```diff
@@ -1,18 +1,38 @@
 steps:
-  - label: "cargo test"
+  - label: "cargo test integration-tests*"
     command: |
       source ~/.cargo/env && set -eux
-      RUST_BACKTRACE=1 RUST_LOG=WARN cargo test --locked --workspace
+      RUST_BACKTRACE=1 RUST_LOG=WARN cargo test --locked -p 'integration-tests' -j 16
```
matklad, line 5: Why do we need `-j 16` argument here? Cargo runs tests in parallel by default.

matklad (commented):
Nice win!

.buildkite/pipeline.yml
pmnoxx: @matklad Yes, I'm going to remove it; I think it doesn't make any difference. 

Anyway, I want to do this one next, as it's a clear win: https://github.com/near/nearcore/pull/5771

PR #5711, opened 2021-12-08
bug: Disable flunky `test_build_receipt_hashes_with_num_shard` check, but keep the test

pmnoxx:
Let's disable `test_build_receipt_hashes` test. This test is based on timing assumption that code `A` should be faster than `B`. That's not true while compiling tests with `-O2`. Let's disable tor remove this test.

This test always fails, when running with `-O2`.
```
L .~/r/nearcore.master piotr-add-tests-to-flunky-test NO_REMOTE cargo test --tests test_build_receipt_hashes
```

```
---- test_utils::test::test_build_receipt_hashes stdout ----
thread 'test_utils::test::test_build_receipt_hashes' panicked at 'naive duration vs production 7.855917ms 6.452102ms', chain/chain/src/test_utils.rs:1471:9
```

Closes https://github.com/near/nearcore/issues/5836


chain/chain/src/test_utils.rs
```diff
@@ -1468,10 +1468,16 @@ mod test {
...
+        assert!(
+            2 * naive_duration > 3 * prod_duration,
+            "naive duration vs production {:?} {:?}",
+            naive_duration,
+            prod_duration
+        );
     }
 
     #[test]
+    #[ignore]
```
mm-near, line 1480: add a comment with github issue that explains more info on why we ignored this test.

mm-near (approved):

chain/chain/src/test_utils.rs
pmnoxx: https://github.com/near/nearcore/issues/5836

bowenwang1996 (commented):
I suggest that we also figure out either how to fix the test or remove it altogether

pmnoxx:
> I suggest that we also figure out either how to fix the test or remove it altogether

First of all, what is the purpose of this test? `cargo test` runs code without optimizations by default. I would say, writing any test, which benchmarks performance is pointless, as you are not really testing the real code. And it should be removed.

mzhangmzz:
> > I suggest that we also figure out either how to fix the test or remove it altogether
> 
> First of all, what is the purpose of this test? `cargo test` runs code without optimizations by default. I would say, writing any test, which benchmarks performance is pointless, as you are not really bencharking the real code. I think it should be removed.

I think this test also tests that build_receipts_hashes and naive_build_receipts_hashes return the same result, which is not about performance, but correctness of the code. I'm ok with removing the timing line, but I think we should keep it the test there

pmnoxx:
> > > I suggest that we also figure out either how to fix the test or remove it altogether
> > 
> > 
> > First of all, what is the purpose of this test? `cargo test` runs code without optimizations by default. I would say, writing any test, which benchmarks performance is pointless, as you are not really bencharking the real code. I think it should be removed.
> 
> I think this test also tests that build_receipts_hashes and naive_build_receipts_hashes return the same result, which is not about performance, but correctness of the code. I'm ok with removing the timing line, but I think we should keep it the test there

@mzhangmzz I absolute agree. This test is good, it's a regression tests, which makes sure we don't break anything in the future. Let's remove just the line about the timing, and close the issue.

bowenwang1996:
Sounds good. Let's remove the timing check then

PR #5712, opened 2021-12-08
Fix small discrepancy in validator threshold

EgorKulikov:
When we calculated threshold we had not accounted that stake of validator should exceed not only total_stake * ratio, but actually (total_stake + stake) * ratio

### Test plan ###
Fixed test_validator_assignment_with_kickout and added new checks in it


nikurt (approved):

bowenwang1996:
@EgorKulikov should we close this PR?

EgorKulikov:
> @EgorKulikov should we close this PR?

No, I'll finish it after hotfix is merged

bowenwang1996 (commented):
@EgorKulikov isn't this a protocol change?

EgorKulikov:
> @EgorKulikov isn't this a protocol change?

It is, and that's why I am waiting on that hotfix

bowenwang1996:
> > @EgorKulikov isn't this a protocol change?
> 
> It is, and that's why I am waiting on that hotfix

What is the reason for making this change? It seems that the current behavior is also fine.

EgorKulikov:
> > > @EgorKulikov isn't this a protocol change?
> > 
> > 
> > It is, and that's why I am waiting on that hotfix
> 
> What is the reason for making this change? It seems that the current behavior is also fine.

Well, we display incorrect number as threshold for validator stake. Someone may base their stake action on it and miss out on an epoch of validating

bowenwang1996:
> Well, we display incorrect number as threshold for validator stake. Someone may base their stake action on it and miss out on an epoch of validating

In that case I suggest we fix the display instead

EgorKulikov:
> > Well, we display incorrect number as threshold for validator stake. Someone may base their stake action on it and miss out on an epoch of validating
> 
> In that case I suggest we fix the display instead

I think this is relatively small change as far as protocol features concerned, it helped to find one more place where we use incorrect epoch's version for chunk only producers (so parts of this PR would be incorporated in that fix if we will decide to forgo this PR), so I'd rather have correct number internally

chain/epoch_manager/src/validator_selection.rs
```diff
@@ -684,14 +711,69 @@ mod tests {
         // too low stakes are kicked out
         let kickout = epoch_info.validator_kickout();
         assert_eq!(kickout.len(), 2);
+        #[cfg(feature = "protocol_feature_fix_staking_threshold")]
+        let expected_threshold = 334;
+        #[cfg(not(feature = "protocol_feature_fix_staking_threshold"))]
+        let expected_threshold = 300;
```
matklad, line 717:
```suggestion
        let expected_threshold = if cfg!(feature = "protocol_feature_fix_staking_threshold") { 300 } else { 334 }
```

chain/epoch_manager/src/validator_selection.rs
```diff
@@ -684,14 +711,69 @@ mod tests {
...
+                &epoch_info,
+                proposals,
+                Default::default(),
+                Default::default(),
+                0,
+                PROTOCOL_VERSION,
+                PROTOCOL_VERSION,
+            )
+            .unwrap();
+            assert_eq!(num_validators + 1, epoch_info.validators_iter().len());
```
matklad, line 752: Not sure if it is worth it for this case, but, in general, I'd prefer to `cfg` *just* the asserts, rather than the whole test, such that the single test checks both the old and the new behavior.

matklad (approved):
>I think this is relatively small change as far as protocol features concerned

This is actually an interesting policy question: 

If we find that something is suboptimal in the protocol, but not outright broken, *and* fixing that is easy (ie, you don't need fork large swaths of code), should we make a protocol change? 

My gut feeling is that yes, we generally should do the change, as that creates a pressure to keep protocol-change mechanism reliable and efficient, and that'll indirectly improve the chance that an actually important protocol upgrade would go smoothly.

chain/epoch_manager/src/validator_selection.rs
EgorKulikov: I actually wrote it that way first, but I kinda prefer compile time for this

bowenwang1996 (approved):
@matklad convinced me that we should do this :)

PR #5713, opened 2021-12-08
Near-wasi and add it to runtime to run QuickJS and future C/C++ WASI builds on NEAR

ailisp:
See runtime/near-wasi/README.md for the full motivation

Test Plan
------------
- [ ] Existing CI
- [ ] Proper protocol upgrade and feature guard
- [ ] Unit test for non-stub WASI functions


matklad:
Can we implement this by pre-processing WASM instead? 

That is, we get 

```wat
(module 
  (import "wasi_snapshot_preview1" "fd_read"))
```

and we transfrom it to 

```wat
(module 
  (func $fd_read (impl of fd_read in terms of host functions here)))
```

Seems way less invasive that way

ailisp:
> Can we implement this by pre-processing WASM instead?
> ...
> Seems way less invasive that way

Sounds possible! It is more difficult to implement but avoid protocol change is great so I think it's a better approach overall. And in this case, it's going to be a separate near-wasi tool instead of a crate in nearcore. I'll close this PR

matklad:
I don't know, but I would guess that there's already tooling for "statically link" wasm modules. Ie, rather than writing the above transfomation by hand, it might be the case that we can just provide a `near-wasi.wasm` impl of wasi as a blob, and call something like `wabt link near-wasi.wasm user-contract.wasm` which does all the renumbering. 


Again, not sure if it exists, but looks like it should exist, see

https://github.com/WebAssembly/tool-conventions/blob/main/Linking.md

ailisp:
I'll take a look whether this works!

PR #5715, opened 2021-12-08
fix: correctly include WASM instruction cost into profile

matklad:
Our gas profiling was incomplete, as it didn't include the cost for wasm
instructions. One would thing that it is enough to subtract the sum of
all other costs from the total to get one, but that's not entirely
correct -- the total gas_burnt the user sees in the RPC includes various
fees computed outside of near_vm_runner. So let's explicitly account for
WASM execution costs in profile (luckily, adding new costs is a
backwrads-compatible change).

While we are at it, let's also add an integration tests that asserts
that this time we indeed accounted for all of the costs.

Test Plan
------------

See the added test

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -2,7 +2,7 @@ steps:
   - label: "cargo test"
     command: |
       source ~/.cargo/env && set -eux
-      RUST_BACKTRACE=1 RUSTFLAGS='-D warnings' cargo test --locked --workspace
+      RUST_BACKTRACE=1 RUST_LOG=WARN cargo test --locked --workspace
 
     timeout: 60
     agents:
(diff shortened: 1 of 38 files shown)


chain/client/src/test_utils.rs
```diff
@@ -1493,6 +1495,16 @@ impl TestEnv {
         }
     }
 
+    #[track_caller]
```
bowenwang1996, line 1498: What is this magic :)

bowenwang1996 (approved):

chain/client/src/test_utils.rs
matklad: A thing you stick on top of test helpers. 

If you do 

```
fn foo() {
    bar()
}

#[track_caller]
fn bar() {
    panic!("kaboom")
}
```

then the panic message would point at the `bar()` call inside `foo`, rather than at the `panic` inside of `bar()`.


In this case, this helper can panic if the transaction is not available, and its useful too see the `#[test]` where it happened, rather than the helper iteslf.

PR #5721, opened 2021-12-08
chain: rosetta: add `related_transactions` to transactions

mina86:
When returning a Rosetta transactions list receipts created as an
outcome of given NEAR transaction or receipt and return them in
`related_transactions` list.

This allows matching related NEAR transactions and receipts together
using only Rosetta API and figuring out where a transaction submitted
via Rosetta API finishes (it’s done once all forward related
transactions appear on chain).


mina86:
There’s this loop in `get_transaction_outcome` which I don’t really like. It’s what JSON RPC does so I’ve pretty much just copied it here.

cc: @bowenwang1996

bowenwang1996 (commented):
@mina86 I don't really see a better option here. What ideas do you have?

mina86:
I don’t have any other ideas. ;) It just feels like I was missing something and hoping there was a better way to get this working.

mina86:
@frol, friendly ping.

chain/rosetta-rpc/src/models.rs
```diff
@@ -948,12 +948,24 @@ pub(crate) struct Transaction {
...
+/// Transaction related to another [`Transaction`] such as matching transfer or
+/// a cross-shard transaction.
+#[derive(Debug, Clone, PartialEq, serde::Serialize, serde::Deserialize, Apiv2Schema)]
+pub(crate) struct RelatedTransaction {
+    // Rosetta API defines network_identifier as well but since all our related
+    // transactions are always on the same network we can leave out this field.
+    // pub network_identifier: NetworkIdentifier,
+    pub transaction_identifier: TransactionIdentifier,
+
+    pub direction: String,
```
frol, line 966: I would use a custom enum instead of a String (see `SubAccount`, `CurrencySymbol`, `SyncStage` as examples).

chain/rosetta-rpc/src/adapters/mod.rs
```diff
@@ -144,79 +146,313 @@ pub(crate) async fn convert_block_to_transactions(
...
+
+    /// Hash of the block for which list of transaction is created.  This is
+    /// used for generating Rosetta transaction identifiers for block events.
+    block_hash: &'a near_primitives::hash::CryptoHash,
+
+    /// Factory for generating [`crate::models::Transaction`] objects.  In
+    /// production code this is always Some and generates transactions with list
+    /// of related transactions.  In test code this may be None in which case
+    /// the transactions will be generated with empty related transactions list.
+    factory: Option<TransactionsFactory<'a>>,
```
frol, line 221: I would avoid adding Option to an otherwise strictly-typed field just for testing purposes, but I will let you make the final call here.

chain/rosetta-rpc/src/adapters/mod.rs
```diff
@@ -144,79 +146,313 @@ pub(crate) async fn convert_block_to_transactions(
...
+                            "Failed to fetch transaction: {}: {:?}",
+                            tx_hash, err
+                        ))
+                    })?;
+                if let Some(outcome) = status {
+                    break Ok(outcome);
+                }
+                tokio::time::sleep(self.polling_config.polling_interval).await;
+            }
+        })
```
frol, line 427: We don't need the whole TxStatus (which pulls and resolves all the receipts recursively and thus will block Rosetta block response until all the future receipts get executed).

Given that all you want to know is the list of produced receipts, you should be able to re-use (or consider moving it into ViewClient) what we have in Indexer Framework to deal with the [local receipts, and delayed local receipts](https://github.com/near/nearcore/blob/1037931fc45b3f51fb512bc9cb61ac663c8eb0c5/chain/indexer/src/streamer/mod.rs#L67-L161)

chain/rosetta-rpc/src/adapters/mod.rs
```diff
@@ -144,79 +146,313 @@ pub(crate) async fn convert_block_to_transactions(
...
+    /// If `tx_hash` is not None, fills out transaction’s `related_transactions`
+    /// with receipts for Near transaction with given hash.
+    async fn make_transaction(
+        factory: &mut Option<TransactionsFactory<'a>>,
+        hash: &String,
+        tx_hash: Option<near_primitives::hash::CryptoHash>,
+    ) -> crate::errors::Result<crate::models::Transaction> {
+        let id = crate::models::TransactionIdentifier { hash: hash.clone() };
+        let related_transactions = match (factory, tx_hash) {
+            (Some(factory), Some(tx_hash)) => factory.get_related_transactions(tx_hash).await?,
```
frol, line 318: Well, receipts should also have `related_transactions` since we want to be able to tie all the children receipts to the parent transaction

frol (changes requested):

frol:
@mina86 Sorry for the delay. The overall structure is fine, but I believe you will need to make substantial changes given that receipts currently miss the `relative_transaction` while they should have it.

chain/rosetta-rpc/src/adapters/mod.rs
mina86: Unfortunately I cannot think of a good way to deal with this. Because this uses async functions I cannot define a trait and than have a test mock that implements that trait. And `Addr<ViewClientActor>` isn’t currently mockable.

mina86:
@frol, PTAL. So this seems to be working but to be honest I don’t think I understand receipts so I don’t have confidence that it will always return all receipts. Specifically, indexer deals a lot with things like local and delayed receipts and after spending a lot of time looking at the code I’m embarrassed to admit that I still don’t get what that’s all about. (Still this PR should work as commit getting us in the right direction).

frol:
@mina86 I will take a closer look by lunch time tomorrow.

mina86:
I’ve talked with Bowen about this a bit and I think this code might actually be correct. Transactions will list receipts in related transactions but those receipt may later further list more receipts. I.e. one will need to recursively check transaction’s completion to see when a tranasction ends.

frol (approved):
The implementation looks right to me, so I am approving it, but there are a few nits regarding API design and naming

chain/rosetta-rpc/src/adapters/transactions.rs
```diff
@@ -1,69 +1,164 @@
...
+    /// Creates an empty mapping.  This is useful for tests.
+    #[cfg(test)]
+    fn empty() -> Self {
+        Self { map: Default::default() }
+    }
+
+    /// Returns list of related transactions for given NEAR transaction or
+    /// receipt.  Note that the hash is removed from the mapping so subsequent
+    /// calls of the method with the same hash will always return an empty
+    /// vector.
```
frol, line 41: > calls of the method with the same hash will always return an empty vector.

Why is this necessary? I always question myself what was the reason the code does extra work. It seems to me that `get_for_cause` does not rely on the fact that second call should lead to an empty vector.

frol (approved):
Overall, looks good to me now.

frol:
@mina86 Thanks for addressing the concerns I had! BTW, tests are failing

chain/rosetta-rpc/src/adapters/transactions.rs
```diff
@@ -1,69 +1,162 @@
...
+
+    /// Creates an empty mapping.  This is useful for tests.
+    #[cfg(test)]
+    pub(crate) fn empty() -> Self {
+        Self { map: Default::default() }
+    }
+
+    /// Returns list of related transactions for given NEAR transaction or
+    /// receipt.
+    fn pop_related(&self, exec_hash: CryptoHash) -> Vec<crate::models::RelatedTransaction> {
```
frol, line 40: Now it is not `pop`ing, so let's rename it

frol (approved):

mm-near (approved):

PR #3025, opened 2020-07-22
fix(client): fix trie_node_touched bug

mikhailOK:
Add read snapshots to storage and do refactoring where
Trie instances are no longer persisted.

Fixes #3020 

Test plan
---------
existing tests

PR #1683, opened 2019-11-11
Runtime fee per touched trie node

mikhailOK:

PR #5722, opened 2021-12-08
feat: trace node counter on receipt processing

Longarithm:
We need to track touched nodes quite frequently, this should simplify debugging.

Also disable `ext.reset_touched_nodes_counter()` on `VMLogic` creation. It should not affect the result because only counter differences matter, but it made counter harder to debug.

## Testing

Existing tests.

### DIFF
--- a/runtime/near-vm-logic/src/logic.rs
+++ b/runtime/near-vm-logic/src/logic.rs
@@ -108,7 +108,6 @@ impl<'a> VMLogic<'a> {
         memory: &'a mut dyn MemoryLike,
         current_protocol_version: ProtocolVersion,
     ) -> Self {
-        ext.reset_touched_nodes_counter();
         // Overflow should be checked before calling VMLogic.
         let current_account_balance = context.account_balance + context.attached_deposit;
         let current_storage_usage = context.storage_usage;
--- a/runtime/runtime/src/lib.rs
+++ b/runtime/runtime/src/lib.rs
@@ -1274,8 +1274,7 @@ impl Runtime {
                                    state_update: &mut TrieUpdate,
                                    total_gas_burnt: &mut Gas|
          -> Result<_, RuntimeError> {
-            let _span = tracing::debug_span!(target: "runtime", "Runtime::process_receipt", receipt_id = %receipt.receipt_id).entered();
-            tracing::debug!(target: "runtime", node_counter = state_update.trie.counter.get());
+            let _span = tracing::debug_span!(target: "runtime", "Runtime::process_receipt", receipt_id = %receipt.receipt_id, node_counter = state_update.trie.counter.get()).entered();
             let result = self.process_receipt(
                 state_update,
                 apply_state,
--- a/runtime/runtime/src/lib.rs
+++ b/runtime/runtime/src/lib.rs
@@ -1274,7 +1274,7 @@ impl Runtime {
                                    state_update: &mut TrieUpdate,
                                    total_gas_burnt: &mut Gas|
          -> Result<_, RuntimeError> {
-            let _span = tracing::debug_span!(target: "runtime", "Runtime::process_receipt", receipt_id = tracing::field::display(receipt.receipt_id)).entered();
+            let _span = tracing::debug_span!(target: "runtime", "Runtime::process_receipt", receipt_id = %receipt.receipt_id).entered();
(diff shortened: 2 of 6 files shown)


runtime/runtime/src/lib.rs
```diff
@@ -1276,17 +1274,19 @@ impl Runtime {
                                    state_update: &mut TrieUpdate,
                                    total_gas_burnt: &mut Gas|
          -> Result<_, RuntimeError> {
-            self.process_receipt(
+            let _span = tracing::debug_span!(target: "runtime", "Runtime::process_receipt", receipt_id = tracing::field::display(receipt.receipt_id)).entered();
```
matklad, line 1277:
```suggestion
            let _span = tracing::debug_span!(target: "runtime", "Runtime::process_receipt", receipt_id = %receipt.receipt_id).entered();
```

runtime/runtime/src/lib.rs
```diff
@@ -1276,17 +1274,19 @@ impl Runtime {
...
+            tracing::debug!(target: "runtime", node_counter = state_update.trie.counter.get());
```
matklad, line 1278: Seems like we can include this as a field in the span above?

matklad (approved):
LGTM, though I haven't double checked the logic that removing `reset_touched_nodes_counter` is sound. Should I?

Longarithm:
> LGTM, though I haven't double checked the logic that removing reset_touched_nodes_counter is sound. Should I?

@matklad yeah, I think here is the highest probability that I missed something important.

runtime/near-vm-logic/src/logic.rs
```diff
@@ -108,7 +108,6 @@ impl<'a> VMLogic<'a> {
         memory: &'a mut dyn MemoryLike,
         current_protocol_version: ProtocolVersion,
     ) -> Self {
-        ext.reset_touched_nodes_counter();
```
bowenwang1996, line 111: why?

bowenwang1996:
Please be careful here. We had a bug in the past where the incorrect management of the trie node touched counter caused nodes to diverge on state.

runtime/near-vm-logic/src/logic.rs
Longarithm: Because if I want to track nodes counter before and after receipt processing, resetting counter in the middle of execution leads to incorrect results.

Longarithm:
@bowenwang1996 that indeed makes sense. 

Anyway, I see that this reset was implemented here at first - https://github.com/near/nearcore/pull/1683 - and there is no good reason to have it, because everywhere we look at the diff.

matklad (approved):
Took a closer look here, the change looks correct -- we indeed care only about diff between the calls, and the calls can't overflow (we use u64, start from zero, and increment only by one). 

To phrase this in other way -- I don't understand why we added the reset functionality in the first place, it doesn't seem necessary. 

@Longarithm I believe you can remove `TouchedNodesCounter::reset` as well, it is not flagged as unused simply because it is public.

mzhangmzz (approved):
I'll approve to unblock, but I don't know enough about the vm-logic code to review it

bowenwang1996:
> @bowenwang1996 that indeed makes sense.
> 
> Anyway, I see that this reset was implemented here at first - #1683 - and there is no good reason to have it, because everywhere we look at the diff.

I believe this was introduced to circumvent inconsistencies resulted from view call executions that may be concurrent with transaction processing. Are we sure that this is not an issue?

Longarithm:
@bowenwang1996 I found that fix, it is not relevant to resetting counter: https://github.com/near/nearcore/pull/3025/files

Even if it was, resetting counter would not help because concurrent view calls and execution would spoil counter anyway.

Longarithm:
Looked a bit further - each view call, e.g. `call_function`, creates its own trie node counter via `new_trie_update_view` and doesn't reuse `TrieUpdate` created in `Runtime::apply`, so we are fine.

What feels wrong here is this is not obvious that `TrieUpdate` can't be incorrectly shared among different function calls, and why counter is atomic. We could consider somehow protecting this object or leaving some comment regarding it.

matklad:
Hm, right, this is such a footgun! This definitelly shouldn't use Atomics, and we should explain in the type system that we have unique access here.

PR #5724, opened 2021-12-09
feat: workspace crate compliance

miraclx:
Tracking issue: https://github.com/near/nearcore/issues/5763

This PR implements crate compliance within the workspace.

### TLDR;
This PR introduces a system for ensuring all the crates in the workspace are well-defined according to a series of defined rules.

### Backstory

Since https://github.com/near/nearcore/pull/5130 was merged, two PRs have introduced new _private_ crates missing the `publish = false` specification. Meaning, CI, armed with that update, wrongly attempted to publish those crates as though they were intentionally meant to be published. Luckily, in both cases, it failed for crate packaging related issues, so they weren't actually published. But would've if that didn't happen. (those crates have now been properly marked private in #5413 and #5670).

### Solution

CI check for workspace crate compliance. Runs on every PR, won't merge into master unless all the checks pass.

I've chosen to call this **Themis**, after the Greek Goddess of Justice, personification of law and order https://en.wikipedia.org/wiki/Themis.

Here's a sneak peek to see what the test looks like: ([CI](https://buildkite.com/nearprotocol/nearcore/builds/9587#e715f1ac-69fe-4e0d-a0f3-6571cfbbdafb)) ([Demo](https://asciinema.org/a/TiQ8tbA0Ze40Sm1pvG8yjlVIm))

To be clear, this doesn't check \*all\* `Cargo.toml` files in the repository. Just the ones that are members of the nearcore cargo workspace.


miraclx:
### Rules

- [X] Ensure all member crates in the workspace are unversioned (`v0.0.0`):

  All crates already share the same version as defined within the workspace manifest. For uniformity, all member crates should be unversioned. `cargo new` likes to version to `0.1.0` by default. This ensures new crates follow that system.

- [X] Ensure all member crates in the workspace have explicit `publish = <true/false>` spec:

  By default, cargo treats the omission of that option as `publish = true` which is not very helpful in our case. So this rule, enforced, makes sure every crate is explicitly publishable or not.

- [X] Ensure all member crates in the workspace specify a MSRV:

  All crates have to specify their Minimum Supported Rust Version.

- [X] Ensure no crate has a MSRV newer than the version in `rust-toolchain`:

  This makes sure that the workspace remains debuggable, and that the toolchain specified can sufficiently compile all the crates in the workspace.

- [X] Ensure all member crates have the same rust edition (for uniformity)

- [X] Ensures all member crates have the NEAR author:

  Ensures all crates are authored by `Near Inc <hello@nearprotocol.com>`, non-exclusively of course.

- [X] Ensures all **_non-private_** crates share the same license in their `Cargo.toml`:

  In our case, `MIT OR Apache-2.0`. Any deviations are flagged as a violation.

- [X] Ensures all **_non-private_** crates have a LICENSE file:

  In the case of `MIT OR Apache-2.0`, this checks for the existence of `LICENSE-APACHE` and `LICENSE-MIT`.

- [X] Ensures all **_non-private_** crates have a README file

- [X] Ensures all **_non-private_** crates have a description

- [X] Ensures all **_non-private_** crates have proper links to the nearcore repo:

  Ensures `repository` and `homepage` in each non-private crate points to https://github.com/near/nearcore

These are the rules I could think of that sort of make sense. If you have any ideas for good rules to add here, let me know.

miraclx:
CI fails because **Themis** is already doing it's job.. really well infact.

So I'll open up a series of PRs addressing the issues Themis highlighted:

- [ ] These packages should have the `publish` specification: [expected: publish = <true/false>]
  
  <details>
  <summary> Affected Crates </summary>

    • `near-chain-configs`
    • `near-crypto`
    • `near-account-id`
    • `near-primitives`
    • `near-primitives-core`
    • `near-rpc-error-macro`
    • `near-rpc-error-core`
    • `near-vm-errors`
    • `near-metrics`
    • `near-vm-logic`
    • `near-chain-primitives`
    • `near-chunks-primitives`
    • `near-network-primitives`
    • `near-client-primitives`
    • `near-jsonrpc-primitives`

  </details>


- [ ] These packages should specify a Minimum Supported Rust Version (MSRV):

  <details>
  <summary>Affected Crates</summary>

    • `themis`
    • `near-cache`

  </details>

- [ ] These packages have an unexpected rust edition: [expected: `2021`, used by 57 other packages in the workspace]

  <details>
  <summary>Affected Crates</summary>

    • `near-rate-limiter` (found: `2018`)
    • `near-jsonrpc-tests` (found: `2018`)

  </details>

- [ ] These packages need to be correctly authored: [expected: `Near Inc <hello@nearprotocol.com>`]

  <details>
  <summary>Affected Crates</summary>

    • `near-vm-runner-fuzz` (found: `[]`)
    • `testlib` (found: `[]`)
    • `near-rate-limiter` (found: `[]`)
    • `runtime-tester-fuzz` (found: `["Automatically generated"]`)
    • `storage-usage-delta-calculator` (found: `["Egor Kulikov <egor@near.org>"]`)
    • `restored-receipts-verifier` (found: `["Aleksandr Logunov <alex.logunov@near.org>"]`)
    • `near-cache` (found: `[]`)

  </details>

- [ ] These non-private packages have an unexpected license: [expected: `MIT OR Apache-2.0`, used by 18 other packages in the workspace]

  <details>
  <summary>Affected Crates</summary>

    • `near-account-id` (found: `Apache-2.0`)

  </details>

- [ ] These non-private packages should have a license file:

  <details>
  <summary>Affected Crates</summary>

    • `near-chain-configs`
    • `near-chain-primitives`
    • `near-chunks-primitives`
    • `near-network-primitives`
    • `near-client-primitives`
    • `near-jsonrpc-primitives`

  </details>

- [ ] These non-private packages should have a readme file:

  <details>
  <summary>Affected Crates</summary>

    • `near-chain-configs`
    • `near-crypto`
    • `near-primitives`
    • `near-primitives-core`
    • `near-rpc-error-macro`
    • `near-rpc-error-core`
    • `near-metrics`
    • `near-chain-primitives`
    • `near-chunks-primitives`
    • `near-network-primitives`
    • `near-client-primitives`
    • `near-jsonrpc-primitives`

  </details>

- [ ] These non-private packages need to have appropriate `repository` and `homepage` links: [expected: https://github.com/near/nearcore]

  <details>
  <summary>Affected Crates</summary>

    • `near-primitives-core`
    • `near-rpc-error-macro`
    • `near-rpc-error-core`
    • `near-vm-errors`
    • `near-vm-logic`

  </details>

matklad:
High level comment: this can be a separate binary, but it'd be simpler if it were just a test. Take a look at how rust-analyzer does similar thing:

https://github.com/rust-analyzer/rust-analyzer/blob/master/crates/rust-analyzer/tests/slow-tests/tidy.rs

(the "tidy" name comes from rust-lang/rust)

matklad:
Second high-level comment -- we definitely need to have *some* infra in place here. "check that the project as a whole obeys these idiosyncratic rules" is something that any sufficiently large project needs at some point, and it's better if there's a pattern that can be used to add such rules, rather than each check being an ad-hoc script. 

However, I think it's important to minimize the amount of accidental complexity. Spending 0.6k lines for this feels a bit to much. I see fancy colors, marco-based system which supports ignores, etc, and I feel that maybe we don't need most of those things. We are our own customers, so, instead of having a generic solution which is best across many dimensions, we could have a non-generic, minimal thing which does *exactly* what we need, and which can be configured by modifying source code. 

Some specific suggestions: 

* leverage `#[test]`  rather than our own "task runner"
* maybe drop ansi coloring and other more fancier reporting 
* maybe consider quick&dirty text processing rather than proper toml parsing. (again, [rust-anlyzer](cargo test --package rust-analyzer --test slow-tests --quiet -- tidy::check_licenses --exact --nocapture) is a bad example to lean from here).

miraclx:
> leverage `#[test]` rather than our own "task runner"

1. I needed all the compliance violations to be reported in one go. So you know all that is wrong all at once. Helpful in the case of CI so you aren't incrementally pushing fixes for fractions of the errors only to be hit by another wall of compliance violations.
2. First, we don't need concurrency. And If we're splitting all the rules into different test units, we need to share data between them. Specifically, the vector of workspace members and their data. To avoid running `cargo metadata` inside each test unit.
3. I think this would've been a great option if developers actually realistically tested the whole workspace before pushing. Personally, I find myself testing whatever specific project I'm working on but never really the whole workspace. Not really.
   So, the question arises of where to put these tests. But also, the time `cargo test --workspace` takes before it starts to run introduces unnecessary friction if the compliance checks are going to fail anyways. While `cargo run -p themis` only has to compile themis to run the checks.

> marco-based system which supports ignores

For one, the check to ensure all crates in the workspace have a `README.md` file is quite important. afaik, there's been some effort lately to achieve that and when we do, we best have a check to ensure no new crates are introduced without having documentation. But in the meantime, while that effort is WIP, we can skip over that check.

> maybe drop ansi coloring and other more fancier reporting

Helps readability, without it, in extreme cases it's just a blob of long-winded text, hard to parse through important information.

> can be configured by modifying source code

Well, this is sort of structured to be easy to modify so rules can be added in, removed, ignored, etc.

matklad:
To clarify, the main benefit of `cargo test`  is piggybacking on existing knowledge and tools: `cargo test` is something that everyone knows, while `cargo run -p themis` not so much. In a similar vein, if you go with `cargo test`, you don't have to update CI setup, you can use familiar workflow for re-running a single failing test, etc. There's a knowledge cost to custom tools.

I totally sympathise with the "our tests are rather useless for local development, so let's avoid tests" view :) But I think in this specific case, `cargo test -p themis` would work just as well as `cargo run -p themis`. I'd also would love us to fix that in a more fundamental way, by making `cargo test` actually useful (#4490).

I feel moderately strongly that this check should be a test (or a bunch of tests) for a somewhat soft "don't add N+1 tool" reason. 

> Helps readability

Yeah, no doubt about that! But it also adds to maintenance burden (a bit longer to compile, a bit more false positives when grepping the code, a bit more time to understand the logic behind the code). Given that we expect failures here to be relatively rare, my gut feeling is that maintenance cost outweigh the extra benefits. There's also a "less is more" angle -- one of the best way to improve readability is to just print less stuff. With the current fancy impl, I can not at a glance say if we can make the output more concise, simply because the logic for printing is very elaborate, and I don't understand how the output looks like without actually running the thing.

I guess, the last point is the main thing I am feeling a bit uncomfortable about -- the thing reads like a very well-written and thought-out system, but, as any complex system, it needs some careful reading for understanding. The complexity here doesn't seem proportional to the problem we are solving: intuitively, each check can be just a single `#[test]` function which contains logic, and does a `panic!("well formatted, concise error message")` at the end if something's wrong. 

Though, on this point I don't have super-strong feelings -- it's still just 0.6k lines of code, so nothing to worry about.

matklad:
>But in the meantime, while that effort is WIP, we can skip over that check.

There's a better way here: add an allow list of poorly-documented crates, and fail if:

* there's a crate which isn't in the allow list but doesnt' have a README
* there's a crate with README, but it's also in the allow list

That way, you create a ratchet that makes sure that documentedness doesn't regress, and that each new crate gets docs. That's what we did in ra until we docced all modules: 

https://github.com/rust-analyzer/rust-analyzer/blob/188b0f96f98feaa0771f941343887c46113c8ced/xtask/src/tidy.rs#L395-L413

frol (approved):
I don't have any hard feelings about how the thing is implemented, though tests might indeed be more maintainable by the team. Yet, I vote for that we don't spend too much time on it. 0.6k LOC of a standalone thing that is only run as a sanity check is fine with me. I would only worry about it if we want to get it into some standalone module with too much effort put into it right now

tools/themis/src/main.rs
```diff
@@ -0,0 +1,74 @@
...
+        Error::RuntimeError(err.into())
+    }
+}
+
+fn main() -> anyhow::Result<()> {
+    let workspace = utils::parse_workspace()?;
+
+    if workspace.members.is_empty() {
+        return Ok(());
+    }
```
matklad, line 49: Hm, why do we need this condition here, it seems like if we remove it than everything will just work? If anything, seems better to assert that the list of memebers is not empty here, to sanity check that we, in fact, do check something.

tools/themis/src/utils.rs
```diff
@@ -0,0 +1,135 @@
...
+                    reason.map_or("".to_string(), |reason| format!(", {}", reason)),
+                    c_expected = style::fg(style::Color::Color256(35))
+                        + &style::bg(style::Color::Gray { shade: 3 })
+                        + style::bold(),
+                    c_none = style::reset()
+                )),
+                c_heading = style::fg(style::Color::Color256(172)) + style::bold(),
+                c_none = style::reset()
+            );
+            let report = outliers.into_iter().fold(
```
matklad, line 82: This would be simpler with a `for` loop. It's pretty rare case that in Rust `.fold` makes things more readable.

tools/themis/src/utils.rs
```diff
@@ -0,0 +1,135 @@
...
+            );
+            eprintln!("{}", report);
+            Ok(true)
+        }
+        _ => Ok(false),
+    }
+}
+
+pub fn is_publishable(pkg: &Package) -> bool {
+    !matches!(pkg.raw["package"].get("publish"), Some(toml::Value::Boolean(false)))
```
matklad, line 113: No need to reach into `raw` to get this, this is exposed via cargo metadata: 

https://docs.rs/cargo_metadata/latest/cargo_metadata/struct.Package.html#structfield.publish

tools/themis/src/utils.rs
```diff
@@ -0,0 +1,135 @@
...
+    }};
+    (@ [$root:expr] $file:literal) => {{
+        $root.join($file).exists()
+    }};
+    (@ [$root:expr] ($file:expr)?) => {{
+        $file.as_ref().map_or(false, |file| $root.join(file).exists())
+    }};
+}
+
+pub(crate) use pub_missing;
```
matklad, line 135: As a general readability thing, it's rater hard to figure out what this macro is supposed to do.

tools/themis/src/main.rs
```diff
@@ -0,0 +1,74 @@
...
+            rules::has_unified_rust_edition,
+            rules::author_is("Near Inc <hello@nearprotocol.com>"),
+            rules::publishable_has_license,
+            rules::publishable_has_unified_license,
+            rules::publishable_has_license_file,
+            rules::publishable_has_description,
+            rules::publishable_has_readme,
+            rules::publishable_has_links("https://github.com/near/nearcore"),
+        }
+    };
```
matklad, line 67: It feels like this can be a for loop over an array of function pointers, rather than a macro.

tools/themis/src/rules.rs
```diff
@@ -0,0 +1,351 @@
...
+
+    if !outliers.is_empty() {
+        return Err(Error::OutcomeError {
+            msg: "These packages should have the `publish` specification".to_string(),
+            expected: Some(Expected { value: "publish = <true/false>".to_string(), reason: None }),
+            outliers,
+        });
+    }
+
+    return Ok(());
```
matklad, line 22:
```suggestion
    Ok(())
```

tools/themis/src/rules.rs
```diff
@@ -0,0 +1,351 @@
...
+}
+
+/// ensure all crates specify a MSRV
+pub fn has_rust_version(workspace: &Workspace) -> Result<(), Error> {
+    let outliers: Vec<_> = workspace
+        .members
+        .iter()
+        .filter(|pkg| pkg.raw["package"].get("rust-version").is_none())
```
matklad, line 30: No need to reach into `raw` here, this is exposed via cargo metadata.

tools/themis/src/rules.rs
```diff
@@ -0,0 +1,351 @@
...
+                pkg.parsed.version,
+                semver::Version {
+                    major: 0,
+                    minor: 0,
+                    patch: 0,
+                    ref pre,
+                    ref build,
+                } if pre == &semver::Prerelease::EMPTY
+                  && build == &semver::BuildMetadata::EMPTY
+            )
```
matklad, line 62:
```suggestion
            pkg != semver::Version::new(0, 0, 0)
```

tools/themis/src/rules.rs
```diff
@@ -0,0 +1,351 @@
...
+            outliers,
+        });
+    }
+
+    return Ok(());
+}
+
+/// ensure all crates have the appropriate author, non-exclusively of course.
+pub fn author_is<'a>(expected: &'a str) -> impl Fn(&'a Workspace) -> Result<(), Error> {
+    move |workspace| {
```
matklad, line 162: Not a huge fan of high-order functions. I'd just hard-code the author string.

tools/themis/src/rules.rs
```diff
@@ -0,0 +1,351 @@
...
+    let mut license_groups = HashMap::new();
+
+    for pkg in &workspace.members {
+        if let Some(license) = &pkg.parsed.license {
+            *license_groups.entry(license).or_insert(0) += 1;
+        }
+    }
+
+    let (most_common_license, n_compliant) =
+        license_groups.into_iter().reduce(|a, b| if a.1 > b.1 { a } else { b }).unwrap();
```
matklad, line 247: Why do we need to dynamicaly select most common license? Seems that the license we use is just a constant we know.

matklad (approved):
Seems fine to land as is, although I definitely get the feeling that the impl is a bit over engineered with macros and high-order functions ;)

tools/themis/src/main.rs
miraclx: Yeah, that makes sense.

tools/themis/src/utils.rs
miraclx: Well, I thought so too. Until I realized, according to the doc:

> Publishing is unrestricted if `None`, and forbidden if the `Vec` is empty.

This doesn't disambiguate between omitting the field and explicitly specifying `publish = true`. Whereas, in our case, we need to fail if the field is omitted.

tools/themis/src/utils.rs
miraclx: True.. could do with some documentation.. or just expanding the macro in all it's invocations..

Generally, it's supposed to check for the existence of files. By default with OR, but optionally, AND:

```rust
pub_missing!(pkg, "a" || "b" || ( "c" && "d" ) || "e")
```

Checks for the existence of either `a`, `b`, `e` or both `c` and `d`..

Could also do with better naming.

tools/themis/src/rules.rs
miraclx: Yeah, this is one of the refactorings I was going to get to.

tools/themis/src/rules.rs
miraclx: Also, yeah it seems like it should but apparently, while `rust-version` was introduced in v1.56, that field isn't going to be included into `cargo metadata` until v1.58.. which isn't even really out yet.

tools/themis/src/rules.rs
miraclx: Artifacts of, like you mentioned, a more generic design. I'll resolve it.

tools/themis/src/rules.rs
miraclx: Artifacts of, like you mentioned, a more generic design. I'll resolve it.

tools/themis/src/main.rs
```diff
@@ -0,0 +1,74 @@
...
+pub enum Error<'a> {
+    OutcomeError { msg: String, expected: Option<Expected>, outliers: Vec<PackageOutcome<'a>> },
+    RuntimeError(anyhow::Error),
+}
+
+impl<'a, E: Into<anyhow::Error>> From<E> for Error<'a> {
+    fn from(err: E) -> Self {
+        Error::RuntimeError(err.into())
+    }
+}
```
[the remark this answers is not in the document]
miraclx: I kinda wanted to disambiguate between a check failing from compliance violations and one failing from I/O or non-compliance related errors. I suppose the return signature could also be a `anyhow::Result<Option<Outcome>>`.

So:
- `Ok(None)` if there are no issues,
- `Ok(Some(Outcome { msg: .., expected: .., outliers: .. }))` if there are.
- `Err(anyhow::Error)` for actual runtime related errors.

tools/themis/src/utils.rs
miraclx: Simplified this to `exists!(pkg, <files...>)`

.buildkite/pipeline.yml
```diff
@@ -23,6 +23,7 @@ steps:
     command: |
       source ~/.cargo/env && set -eux
       rustc --version && cargo --version
+      RUSTFLAGS='-D warnings' cargo run -p themis
```
pmnoxx, line 26: RUST_LOG=WARN

miraclx: `RUSTFLAGS='-D warnings'` is for the compiler. `RUST_LOG=WARN` is for runtime related errors, like with `log`, `eyre`, `tracing`, etc..

miraclx: To be clear, for example: while `cargo run` typically warns when you have unused variables, even though it would build and run just fine, when you add `RUSTFLAGS="-D warnings"` then it would fail and cause a hard error.

tools/themis/src/rules.rs
```diff
@@ -0,0 +1,329 @@
...
+            expected: Some(Expected { value: EXPECTED_AUTHOR.to_string(), reason: None }),
+            outliers,
+        });
+    }
+
+    Ok(())
+}
+
+/// ensure all non-private crates have a license
+pub fn publishable_has_license(workspace: &Workspace) -> Result<(), Error> {
```
miraclx, line 171: Rethinking this, maybe this check should be for all crates, not just non-private ones.

tools/themis/src/rules.rs
```diff
@@ -0,0 +1,329 @@
...
+    }
+
+    Ok(())
+}
+
+/// ensure all non-private crates have a license file
+///
+/// Checks for either one LICENSE file, or two LICENSE files, one of which
+/// is the Apache License 2.0 and the other is the MIT license.
+pub fn publishable_has_license_file(workspace: &Workspace) -> Result<(), Error> {
```
miraclx, line 197: Same here..

tools/themis/src/rules.rs
```diff
@@ -0,0 +1,329 @@
...
+        });
+    }
+
+    Ok(())
+}
+
+const EXPECTED_LICENSE: &str = "MIT OR Apache-2.0";
+
+/// ensure all non-private crates use the the same expected license
+pub fn publishable_has_unified_license(workspace: &Workspace) -> Result<(), Error> {
```
miraclx, line 228: ditto..

tools/themis/src/rules.rs
```diff
@@ -0,0 +1,329 @@
...
+            expected: None,
+            outliers,
+        });
+    }
+
+    Ok(())
+}
+
+/// ensure all non-private crates have a README file
+pub fn publishable_has_readme(workspace: &Workspace) -> Result<(), Error> {
```
miraclx, line 274: This too, but maybe when we actually have all crates documented. For now, we can prioritize non-private crates.

.buildkite/pipeline.yml
pmnoxx: @miraclx Oh, sorry. You are right.

tools/themis/src/main.rs
```diff
@@ -0,0 +1,72 @@
...
+            rules::has_unified_rust_edition,
+            rules::author_is_near,
+            rules::publishable_has_license,
+            rules::publishable_has_unified_license,
+            rules::publishable_has_license_file,
+            rules::publishable_has_description,
+            rules::publishable_has_readme,
+            rules::publishable_has_near_link,
+        }
+    };
```
pmnoxx, line 65: My recommendation would be to avoid using macro when it's possible to write a single function instead.

I would rather write check("[workspace"], [
   
            rules::is_unversioned,
            rules::has_publish_spec,
            rules::has_rust_version,
            rules::has_debuggable_rust_version,
            rules::has_unified_rust_edition,
            rules::author_is_near,
            rules::publishable_has_license,
            rules::publishable_has_unified_license,
            rules::publishable_has_license_file,
            rules::publishable_has_description,
            rules::publishable_has_readme,
            rules::publishable_has_near_link,]);

It will be easier to maintain a function, rather than a macro.
I agree that macros have their place, but they take a lot of time to read.

tools/themis/src/main.rs
```diff
@@ -0,0 +1,72 @@
...
+
+    if !passed {
+        std::process::exit(1);
```
pmnoxx, line 68: It may be a good idea to print a human readable message in this case.

miraclx: A human-readable message has already been printed, check the CI output here - https://buildkite.com/nearprotocol/nearcore/builds/9824#af146b87-1a5d-40ab-8074-33752be92a41

pmnoxx: @miraclx Yes, that's good. It's hard to tell, by looking at the macro.

miraclx: You're right, resolved.

tools/themis/src/main.rs
```diff
@@ -0,0 +1,75 @@
...
+        rules::has_debuggable_rust_version,
+        rules::has_unified_rust_edition,
+        rules::author_is_near,
+        rules::publishable_has_license,
+        rules::publishable_has_unified_license,
+        rules::publishable_has_license_file,
+        rules::publishable_has_description,
+        rules::publishable_has_readme,
+        rules::publishable_has_near_link,
+    ];
```
pmnoxx, line 62: Can we add a rule, that checks, whenever all packages have the same version?

Maybe `has_unified_version` or `has_unified_crate_version` ?

tools/themis/src/utils.rs
```diff
@@ -0,0 +1,123 @@
...
+///
+/// ## Example
+///
+/// ```rust
+/// exists!(pkg, "a" || "b" || ( "c" && "d" ) || "e")
+/// ```
+///
+/// If the crate has either `a`, `b`, `e` or both `c` and `d`, this returns true.
+/// Only returns false if all the combinations don't exist.
+macro_rules! exists {
```
pmnoxx, line 104: I would recommend using a regular `Rust` function. This would be so much easier to review.

tools/themis/src/main.rs
miraclx: The first rule - `is_unversioned` is exactly this. We only care that all the crates in the workspace are `v0.0.0`.

tools/themis/src/rules.rs
```diff
@@ -0,0 +1,329 @@
+use std::collections::HashMap;
+
+use super::{style, utils, Error, Expected, PackageOutcome, Workspace};
+
+/// ensure all crates have the `publish = <true/false>` specification
```
pmnoxx, line 5: I with in general, we prefer to start comment for functions with an Upper Case letter.

tools/themis/src/rules.rs
```diff
@@ -0,0 +1,329 @@
...
+    if !outliers.is_empty() {
+        return Err(Error::OutcomeError {
+            msg: "These packages shouldn't be versioned".to_string(),
+            expected: Some(Expected { value: "0.0.0".to_string(), reason: None }),
+            outliers,
+        });
+    }
+
+    Ok(())
+}
```
pmnoxx, line 64: Oh, ignore my other comment. I thought it would be better, to rewrite this check, to make sure all versions are the same. 
This works too.

tools/themis/src/utils.rs
miraclx: meh..

```rust
exists!(pkg, "file1" || "file2" || ("file3" && "file4") || "file5")
```

at best, would end up as:

```rust
exists(pkg, "file1") || exists(pkg, "file2") || (exists(pkg, "file3") && exists(pkg, "file4")) || exists(pkg, "file5")
```

tools/themis/src/rules.rs
```diff
@@ -0,0 +1,329 @@
...
+    if !outliers.is_empty() {
+        return Err(Error::OutcomeError {
+            msg: "These packages should have the `publish` specification".to_string(),
+            expected: Some(Expected { value: "publish = <true/false>".to_string(), reason: None }),
+            outliers,
+        });
+    }
+
+    Ok(())
+}
```
pmnoxx, line 23: Theoretically, this can be written without if statements. I'm not saying we should, but just a theoretically experiment.

```
pub fn has_publish_spec(workspace: &Workspace) -> Result<(), Error> {
    once(
        workspace
            .members
            .iter()
            .filter(|pkg| pkg.raw["package"].get("publish").is_none())
            .map(|pkg| PackageOutcome { pkg, value: None })
            .collect::<Vec<_>>(),
    )
    .into_iter()
    .filter(|outliers| !outliers.is_empty())
    .map(|outliers| Error::OutcomeError {
        msg: "These packages should have the `publish` specification".to_string(),
        expected: Some(Expected { value: "publish = <true/false>".to_string(), reason: None }),
        outliers,
    })
    .next()
    .map(|e| Err(e))
    .unwrap_or(Ok(()))
}
```

pmnoxx: @matklad I wish there was a nicer API for doing things like this.

pmnoxx: Instead of ```
    .next()
    .unwrap_or(Ok(()))
```
something like
```
    .next_or(Ok(()))
```

Comment (pmnoxx) [reply tools/themis/src/rules.rs:23]: I wish I could write it like
```
    workspace
        .members
        .iter()
        .filter(|pkg| pkg.raw["package"].get("publish").is_none())
        .map(|pkg| PackageOutcome { pkg, value: None })
        .collect::<Vec<_>>()
    .once()
    .filter(|outliers| !outliers.is_empty())
    .map(|outliers| {
        Err(Error::OutcomeError {
            msg: "These packages should have the `publish` specification".to_string(),
            expected: Some(Expected { value: "publish = <true/false>".to_string(), reason: None }),
            outliers,
        })
    })
    .next_or(Ok(()))
```

or even better


```
    workspace
        .members
        .iter()
        .filter(|pkg| pkg.raw["package"].get("publish").is_none())
        .map(|pkg| PackageOutcome { pkg, value: None })
    .once()
    .filter(|outliers| !outliers.is_empty())
    .map(|outliers| {
        Err(Error::OutcomeError {
            msg: "These packages should have the `publish` specification".to_string(),
            expected: Some(Expected { value: "publish = <true/false>".to_string(), reason: None }),
            outliers: outliers.collect(),
        })
    })
    .next_or(Ok(()))
```

Comment (miraclx) [reply tools/themis/src/rules.rs:23]: > We don't have `is_empty` on `iter()`

While we actually do have [`is_empty`](https://doc.rust-lang.org/std/iter/trait.ExactSizeIterator.html#method.is_empty) on structs implementing `ExactSizeIterator`, in this case, we don't know how many times this would iterate. So this is not an `ExactSizeIterator`, which means the only way to check `is_empty` would be to consume the first item.

The logic as defined is evidently the clearer way to write this.

Comment (miraclx) [reply tools/themis/src/rules.rs:23]: > We don't have `is_empty` on `iter()`

While we actually do have [`is_empty`](https://doc.rust-lang.org/std/iter/trait.ExactSizeIterator.html#method.is_empty) on structs implementing `ExactSizeIterator`, in this case, we don't know how many times this would iterate. So this is not an `ExactSizeIterator`, which means the only way to check `is_empty` would be to consume the first item.

The logic as defined is evidently the clearer way to write this.

Comment (pmnoxx) [reply tools/themis/src/rules.rs:23]: Yes, I agree.

Another way, to write the logic would be to write it avoiding `return` statement.

```
    if !outliers.is_empty() {
        return Err(Error::OutcomeError {
            msg: "These packages shouldn't be versioned".to_string(),
            expected: Some(Expected { value: "0.0.0".to_string(), reason: None }),
            outliers,
        });
    }
    OK(())
```
would become
```
    if outliers.is_empty() {
        OK(())
    } else {
         Err(Error::OutcomeError {
            msg: "These packages shouldn't be versioned".to_string(),
            expected: Some(Expected { value: "0.0.0".to_string(), reason: None }),
            outliers,
        });
    }
```

**Review by pmnoxx — APPROVED** [review]
LGTM

File: tools/themis/src/utils.rs [hunk 24]
```diff
@@ -0,0 +1,123 @@
...
+    ($pkg:expr, $($files:tt)+) => {{
+        let pkg_root = $pkg.parsed.manifest_path.parent().unwrap();
```
pmnoxx, line 106: Maybe, it's worth changing macro, in such a way that it avoid hidden `unwrap().`

And instead returns `Result<bool, Error>`.

This is not  blocking, though it would be less surprising to the reader, that the function can panic.

pmnoxx:
@miraclx Btw, this is one of the cleanest written scripts in our repo!. Thanks for doing good work!

tools/themis/src/utils.rs
miraclx: Well, this actually isn't ever going to panic. Here's why:

For any crate that exists, the manifest path would be `<pkg>/Cargo.toml`. the parent, would be `<pkg>`

We know, for a fact that there would always be a parent if we even have a crate to begin with.

And, though unlikely, let's game this out and say, what if the crate exists on the root directory. In which case, for `/Cargo.toml`, the parent directory would then be `/`

This unwrap call will never panic and the guarantee is that there's a crate here at all to begin with.

https://docs.rs/camino/latest/camino/struct.Utf8Path.html#method.parent

pmnoxx: @miraclx I understand that it can't panic, and why. My point was that when reading the code, whenever I see `unwrap()` I need to spend a moment and thing about. Instead, adding a comment "can't panic", or writing code in such a way that avoids "unwrap()" would be better.

I'm not saying you have to change anything, just something to thinking about. This is a script, not a production code, so it's fine.

tools/themis/src/main.rs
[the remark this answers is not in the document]
matklad: You can disambiguate using anyhow's [downcasing support](https://docs.rs/anyhow/latest/anyhow/struct.Error.html#method.downcast): 

```rust
fn main() -> anyhow::Result<()> {
    let compliance_error = check().downcast::<ComplianceError>()?;
    report_compliance_error(compliance_error);
    std::process::exit(1)
}

fn check() -> anyhow::Result<()> { ... }
```

That is, both enum-style errors and `dyn` style errors support case analysis -- via `match` and `downcast` respectively. Using *both* mechanisms at the same time is redundant.

tools/themis/src/utils.rs
matklad: +1 to what @pmnoxx says. 

Looking at the second example, I can see what does it do. Looking at the first example, I have no idea what's going on, because the stuff after `!` is just custom made up syntax. So I have to either

a) guess intention of the macro writer judging by just how the surface syntax looks. 
b) carefully read the macro definition
c) ask the IDE to expand the macro and read the expansion.

tools/themis/src/rules.rs
matklad: @pmnoxx I think you are looking for find_map: https://doc.rust-lang.org/stable/std/iter/trait.Iterator.html#method.find_map

Though, in this case this won't work -- we indeed want to collect all of the results, instead of just the first one.

tools/themis/src/main.rs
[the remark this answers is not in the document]
miraclx: Yeah, that's nice. I guess the only irk here is `ComplianceError` now unconditionally has to impl `Display`.
Which makes sense considering there's a `msg` field, but besides not needing this `Display`-ability in our logic, the messages in that field begin with upper case which sorta breaks the convention for `Display`-ed Errors starting with lower caps. I'll implement it nonetheless.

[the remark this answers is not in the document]
miraclx: Keeping as-is, to avoid mitigating the `'static` requirement by repeatedly cloning `&Package`.

chefsale (approved):

PR #5732, opened 2021-12-09
nearcore: support dling compressed `genesis.json` and `config.json`

mina86:
Introduce `AutoXzDecoder` struct (which automatically detects
xz-compressed files and transparently decompresses them) and use it in
`download_file` to support compressed `genesis.json` and `config.json`
files.

Especially the former file may get quite large resulting in long
downloads and wasted bandwidth.  Being able to download compressed
JSON files will reduce the time it takes for `neard init` to finish.
For example, testnet’s `genesis.json` file is 5.5 GB and shrinks to 45
MB when compressed.

The default download URLs still point to uncompressed files (since
compressed ones are not yet available).


nearcore/src/config.rs
```diff
@@ -1108,20 +1112,210 @@ pub enum FileDownloadError {
...
+    /// The header did not match XZ stream header and thus the data is passed
+    /// through.
+    PlainText,
+
+    /// The header did match XZ stream header and thus the data is being
+    /// decompressed.
+    Compressed(xz2::stream::Stream, Box<[u8]>),
+}
+
+static XZ_HEADER_MAGIC: [u8; 6] = [0xFD, 0x37, 0x7A, 0x58, 0x5A, 0x00];
```
bowenwang1996, line 1139: Can we add a comment to explain what this magic is?

nearcore/src/config.rs
```diff
@@ -1108,20 +1112,223 @@ pub enum FileDownloadError {
...
+    let buffer = b"\xfd\x37\x7a\x58\x5a\x00\x00\x04\xe6\xd6\xb4\x46\
+                   \x02\x00\x21\x01\x1c\x00\x00\x00\x10\xcf\x58\xcc\
+                   \x01\x00\x19\x5a\x61\xc5\xbc\xc3\xb3\xc5\x82\xc4\
+                   \x87\x20\x67\xc4\x99\xc5\x9b\x6c\xc4\x85\x20\x6a\
+                   \x61\xc5\xba\xc5\x84\x00\x00\x00\x89\x4e\xdf\x72\
+                   \x66\xbe\xa9\x51\x00\x01\x32\x1a\x20\x18\x94\x30\
+                   \x1f\xb6\xf3\x7d\x01\x00\x00\x00\x00\x04\x59\x5a";
+    for chunk_size in [1, 2, 3, 4, 5, 6, 7, 8, 9, 10] {
+        let got = auto_xz_test_write_file(buffer, chunk_size).unwrap();
+        assert_eq!(got, "Zażółć gęślą jaźń".as_bytes());
```
mm-near, line 1301: ;-)

nearcore/src/config.rs
```diff
@@ -1108,20 +1112,223 @@ pub enum FileDownloadError {
     RemoveTemporaryFileError(std::io::Error, Box<FileDownloadError>),
 }
 
+/// Object which allows transparent XZ decoding when saving data to a file.
```
mm-near, line 1115: optional: I'd put this whole struct in a separate file.

nearcore/src/config.rs
```diff
@@ -1108,20 +1112,223 @@ pub enum FileDownloadError {
...
+            AutoXzState::PlainText => Ok(()),
+            AutoXzState::Compressed(ref mut stream, ref mut buffer) => {
+                Self::decompress(&mut self.file, stream, buffer, b"").await
+            }
+        }
+    }
+
+    /// If object is still in `Probing` state, read more data from the input to
+    /// determine whether it’s XZ stream or not.  Updates `state` accordingly.
+    /// If probing succeeded, returns number of bites from XZ header magic that
```
mm-near, line 1179: s/bites/bytes

nearcore/src/config.rs
```diff
@@ -1108,20 +1112,223 @@ pub enum FileDownloadError {
...
+        } else {
+            Some(0)
+        }
+    }
+
+    /// Writes data to the output file.  Panics if the object is still in
+    /// probing stage.
+    async fn write_all_impl(&mut self, chunk: &[u8]) -> Result<(), FileDownloadError> {
+        match self.state {
+            AutoXzState::Probing(_) => unreachable!(),
```
mm-near, line 1213: can you test 'empty' file and very short file that contains prefix of XZ_HEADER_MAGIC?
I think in such cases, when we call 'finish' - this condition will trigger.

mina86: `write_all_impl` is not called from `finish` so that’s not the case. Added another tests case nonetheless.

nearcore/src/config.rs
mina86: Yes, I’ve been thinking about that. It seems that we need an additional crate with just utility functions and classes which aren’t specific to any other NEAR thing.

mina86:
@mm-near, PTAL

mina86:
@mm-near, friendly ping.

mm-near (approved):

PR #5737, opened 2021-12-09
refactor: use `decode_value_with_rc` logic for `RawTrieNodeWithSize`

Longarithm:
We have similar logic for decoding/encoding DB key-value pairs and trie nodes, but it is implemented by different code, and I find it confusing.

I propose to unify this logic by reusing `decode_value_with_rc` function which shouldn't give overhead and makes it easier to understand.

The only problem could be that on encoding/decoding `rc` (refcount), we are switching from `u64` to `i64`. But decoding `i64` which was encoded as `u64` is correct because our refcount couldn't exceed `2**62` by absolute value.

There is a separate issue that we don't have strict logic of storing refcounts and handling negative values - at least it is not reflected in the code. But IMO it's better to have the same logic in all places.

## Testing 

Existing tests.

### DIFF
--- a/core/store/src/db/refcount.rs
+++ b/core/store/src/db/refcount.rs
@@ -78,6 +78,13 @@ pub(crate) fn encode_value_with_rc(data: &[u8], rc: i64) -> Vec<u8> {
     cursor.into_inner()
 }
 
+pub(crate) fn encode_value_with_rc_inplace(
+    data: &mut Vec<u8>,
+    rc: i64,
+) -> Result<(), std::io::Error> {
+    data.write_i64::<LittleEndian>(rc)
+}
+
 impl RocksDB {
     /// ColState has refcounted values.
     /// Merge adds refcounts, zero refcount becomes empty value.
--- a/core/store/src/trie/mod.rs
+++ b/core/store/src/trie/mod.rs
@@ -22,7 +22,7 @@ use crate::trie::trie_storage::{
     TouchedNodesCounter, TrieMemoryPartialStorage, TrieRecordingStorage, TrieStorage,
 };
 pub(crate) use crate::trie::trie_storage::{TrieCache, TrieCachingStorage};
-use crate::StorageError;
+use crate::{decode_value_with_rc, StorageError};
 
 mod insert_delete;
 pub mod iterator;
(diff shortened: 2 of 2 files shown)


core/store/src/db/refcount.rs
```diff
@@ -78,6 +78,13 @@ pub(crate) fn encode_value_with_rc(data: &[u8], rc: i64) -> Vec<u8> {
     cursor.into_inner()
 }
 
+pub(crate) fn encode_value_with_rc_inplace(
+    data: &mut Vec<u8>,
+    rc: i64,
+) -> Result<(), std::io::Error> {
```
matklad, line 84: There's no way this can fail with `io::Error`, this should return `()`.

Meta note: "can operation X fail in a way Y" is an important low-level coding concern which should be kept in mind. It's too easy to propagate errors which do not make sense, so it's important to be vigilant and notice when error management does't make sense. This is the case where static types make the life harder, because they describe the situation which can not actually arise in practice.

core/store/src/trie/mod.rs
```diff
@@ -384,7 +385,7 @@ impl RawTrieNode {
 impl RawTrieNodeWithSize {
     fn encode_into(&self, out: &mut Vec<u8>) -> Result<(), std::io::Error> {
         self.node.encode_into(out)?;
```
matklad, line 387: Actually, similarly here -- I don't see how encoding can be a failable operation. The only thing that can fail here is memory allocation, but Rust aborts on OOM.

core/store/src/db/refcount.rs
matklad: Can we get rid of `encode_value_with_rc` altogether? In general, it is idiomatic to let caller pass the buffers in, rather than allocate Vecs in callee.

mzhangmzz: I think you mixed `rc` and `memory_usage` here. The `RawTrieNodeWithSize` struct contains `memory_usage`, which records the size of the node, not the ref count.

Ref count is stored in the lower database layer, see `merge_refcounted_records`.

mzhangmzz (changes requested):

Longarithm:
> I think you mixed rc and memory_usage here. The RawTrieNodeWithSize struct contains memory_usage, which records the size of the node, not the ref count.
> 
> Ref count is stored in the lower database layer, see merge_refcounted_records.

@mzhangmzz oh that's right. It is even the size of the whole subtree as comment over `RawTrieNodeWithSize` says.
I think after refactoring this PR still makes sense, but the thing being changed is not a bug. Perhaps I will just close it soon

PR #5741, opened 2021-12-09
feat: record trie cache misses

Longarithm:
We've noticed that storage get time spikes correspond to cache misses, so let's record them in debug logs.

## Testing

Existing tests.

### DIFF
--- a/core/store/src/trie/trie_storage.rs
+++ b/core/store/src/trie/trie_storage.rs
@@ -171,6 +171,7 @@ impl TrieStorage for TrieCachingStorage {
         if let Some(val) = guard.cache_get(hash) {
             Ok(val.clone())
         } else {
+            tracing::debug!(target: "runtime", "trie cache miss", hash = %hash);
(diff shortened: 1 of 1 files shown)


core/store/src/trie/trie_storage.rs
```diff
@@ -171,6 +171,7 @@ impl TrieStorage for TrieCachingStorage {
         if let Some(val) = guard.cache_get(hash) {
             Ok(val.clone())
         } else {
+            tracing::debug!(target: "runtime", "trie cache miss for hash {}", hash);
```
bowenwang1996, line 174: This might lead to too much logging. A prometheus counter may be better

mzhangmzz: I agree here

bowenwang1996:
What is the plan with this PR?

Longarithm:
Close due to too much logging could be caused.

PR #5742, opened 2021-12-09
chain: make ChainUpdate::process_block take validated block only

mina86:


bowenwang1996:
@mina86 what happened to this PR?

mina86:
This was just a draft and turned out that changes to `ClientActor::process_block` broke the code. This is still salvageable though I think and I’m gonna have an updated version.

mina86:
Unfortunately this won’t work. The issue is as before, if we are processing an orphan block we cannot validate it because we don’t have the previous block. Currently, `ChainUpdate::process_block` will throw `ErrorKind::Orphan` if that is the case and the block ends up saved for later.

PR #5743, opened 2021-12-09
re: `near-network` Split private actix types of out `types.rs`

pmnoxx:
In order to more clearly define the public interface of `near-network`. 
We should move all actix messages, that are meat to be used only within the `near-network` to a separate `private_actix.rs` file.

### DIFF
--- a/chain/network/src/private_actix.rs
+++ b/chain/network/src/private_actix.rs
@@ -1,0 +1,174 @@
+/// This file is contains all types used for communication between `Actors` within this crate.
+/// They are not meant to be used outside.
+use crate::network_protocol::PeerMessage;
+use crate::network_protocol::{Edge, PartialEdgeInfo, SimpleEdge};
+use crate::peer::peer_actor::PeerActor;
+use crate::PeerInfo;
+use actix::dev::{MessageResponse, ResponseChannel};
+use actix::{Actor, Addr, Message};
+use conqueue::QueueSender;
+use near_network_primitives::types::{PeerChainInfoV2, PeerType};
+use near_primitives::network::PeerId;
+use near_primitives::version::ProtocolVersion;
+use near_rate_limiter::ThrottleController;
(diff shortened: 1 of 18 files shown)


chefsale (approved):

chain/jsonrpc/Cargo.toml
```diff
@@ -33,7 +33,7 @@ near-network-primitives = { path = "../network-primitives" }
 
 [features]
 dump_errors_schema = ["near-rpc-error-macro/dump_errors_schema"]
-test_features = ["near-client/test_features", "near-network/test_features", "near-jsonrpc-primitives/test_features", "near-jsonrpc-adversarial-primitives/ser_de"]
+test_features = ["near-client/test_features", "near-network/test_features", "near-jsonrpc-primitives/test_features", "near-jsonrpc-adversarial-primitives/ser_de", ]
```
mm-near, line 36: do we need this comma at the end ?

chain/network/src/private_actix.rs
```diff
@@ -0,0 +1,178 @@
+/// This file is contains all types used for communication between `Actors` within this crate.
+/// They are not meant to be used outside.
```
mm-near, line 2: And also they are not persisted anywhere, right ?

chain/network/src/routing/edge_validator_actor.rs
```diff
@@ -1,5 +1,6 @@
+use crate::private_actix::StopMsg;
+use crate::private_actix::ValidateEdgeList;
```
mm-near, line 2: any reason why not in a single line ?

mm-near (approved):

chain/network/src/private_actix.rs
pmnoxx: @mm-near No. The is no way to persist them, because they don't derive `borsh` nor `serde`.

PR #5747, opened 2021-12-09
Remove `stored_partial_encoded_chunks`

mzhangmzz:


bowenwang1996:
I am confused as to whether the PR is ready for review: on the one hand, the PR description says it is not ready for review. On the other hand, my review is requested 😅

mzhangmzz:
> I am confused as to whether the PR is ready for review: on the one hand, the PR description says it is not ready for review. On the other hand, my review is requested 😅

Sorry, I don't know how to cancel request. I converted the PR back to draft after I created it.

mzhangmzz:
@bowenwang1996 @mm-near @matklad @yordanmadzhunkov now it is ready for review. Sorry for the confusion before

chain/chunks/src/lib.rs
```diff
@@ -1202,79 +1226,163 @@ impl ShardsManager {
...
             }
         };
 
-        // 3. Checking chunk height
+        // 1.b Checking chunk height
         let chunk_requested = self.requested_partial_encoded_chunks.contains_key(&chunk_hash);
         if !chunk_requested {
             if !self.encoded_chunks.height_within_horizon(header.height_created()) {
-                return Err(Error::ChainError(ErrorKind::InvalidChunkHeight.into()));
+                return Ok(ProcessPartialEncodedChunkResult::Known);
```
bowenwang1996, line 1385: why? Does this change allow for an exploit given that we are not doing a protocol change here?

chain/client/src/client.rs
```diff
@@ -850,66 +850,61 @@ impl Client {
...
+    /// who are not marked as complete yet
+    pub fn check_incomplete_chunks(&mut self, prev_block_hash: &CryptoHash) -> Vec<AcceptedBlock> {
+        let mut accepted_blocks = vec![];
+        for chunk_header in self.shards_mgr.get_incomplete_chunks(prev_block_hash) {
+            debug!(target:"client", "try to process incomplete chunks {:?}, prev_block: {:?}", chunk_header.chunk_hash(), prev_block_hash);
+            // create a fake partial encoded chunk message with empty parts and receipts
+            // because we just want to reconstruct the chunk
+            let res = self.process_partial_encoded_chunk(MaybeValidated::from_validated(
+                PartialEncodedChunk::new(chunk_header, vec![], vec![]),
+            ));
```
bowenwang1996, line 863: This piece of code feels suboptimal. Is there a way to do it without this hack?

chain/chunks/src/lib.rs
```diff
@@ -1339,20 +1440,78 @@ impl ShardsManager {
...
+        let prev_block_hash = header.prev_block_hash();
+        let epoch_id = match self.runtime_adapter.get_epoch_id_from_prev_block(&prev_block_hash) {
+            Ok(epoch_id) => epoch_id,
+            Err(_) => {
+                return Ok(ProcessPartialEncodedChunkResult::NeedBlock);
+            }
+        };
+        let chunk_hash = header.chunk_hash();
+        // 4. validate the header again now that prev_block is processed
+        if !self.encoded_chunks.get(&chunk_hash).unwrap().header_fully_validated {
```
bowenwang1996, line 1482: could you add a comment on why it is safe to unwrap here?

chain/chunks/src/lib.rs
```diff
@@ -1741,17 +1877,18 @@ impl ShardsManager {
...
 
-        // Save the partial chunk for data availability
-        self.persist_partial_chunk_for_data_availability(&cache_entry, store_update);
-
         // Save this chunk into encoded_chunks.
-        self.encoded_chunks.insert(cache_entry.header.chunk_hash(), cache_entry);
+        self.encoded_chunks.merge_in_partial_encoded_chunk(&partial_chunk);
+
+        // Save the partial chunk for data availability
+        let cache_entry = self.encoded_chunks.get(&partial_chunk.header.chunk_hash()).unwrap();
```
bowenwang1996, line 1890: Same for this unwrap

chain/chunks/src/lib.rs
mzhangmzz: I don't think we ever handle the error returned by this function anyways. 

I can make a separate PR for this. When I made the PR, I didn't know the issue will occur on mainnet again so I included multiple changes related to processing partial encoded chunks

chain/chunks/src/lib.rs
```diff
@@ -1288,48 +1396,41 @@ impl ShardsManager {
...
+        // 1.e Checking receipts validity
         for proof in partial_encoded_chunk.receipts.iter() {
+            // TODO: only validate receipts we care about
             let shard_id = proof.1.to_shard_id;
-            if self.cares_about_shard_this_or_next_epoch(
-                self.me.as_ref(),
-                &prev_block_hash,
-                shard_id,
-                true,
-            ) {
```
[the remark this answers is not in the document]
mzhangmzz: because `prev_block_hash` is not ready and technically we don't know if we track this shard or not. Considering now every node tracks all shard, I think it's not a big issue. I'll create a TODO to track this

chain/chunks/src/lib.rs
```diff
@@ -1202,79 +1226,163 @@ impl ShardsManager {
...
+    ) -> Result<ProcessPartialEncodedChunkResult, Error> {
         let header = &partial_encoded_chunk.header;
         let chunk_hash = header.chunk_hash();
-
-        // 2. Leave if we received known chunk
+        // Verify the partial encoded chunk is valid and worth processing
+        // 1.a Leave if we received known chunk
+        // Note that we have to still process the chunk even if all parts and receipts included
+        // in `partial_encoded_chunk` are already in `encoded_chunks`. This is because
+        // `try_process_full_chunk` can only succ
```
mm-near, line 1374: comment not finished

chain/chunks/src/lib.rs
```diff
@@ -1288,48 +1396,41 @@ impl ShardsManager {
...
-                return Ok(ProcessPartialEncodedChunkResult::NeedBlock);
-            }
-        };
+        // 1.c checking header validity
+        match partial_encoded_chunk
+            .validate_with(|pec| self.validate_chunk_header(chain_head, &pec.header).map(|()| true))
+        {
+            Err(Error::ChainError(chain_error)) => match chain_error.kind() {
+                ErrorKind::DBNotFoundErr(_) => {
+                    return Ok(ProcessPartialEncodedChunkResult::NeedBlock)
```
mm-near, line 1405: comment

chain/chunks/src/lib.rs
mm-near: I'd add a comment to mention why we return "Known" (as I'd have expected InvalidChunkHeight as it was before)

chain/chunks/src/lib.rs
```diff
@@ -1202,79 +1226,163 @@ impl ShardsManager {
...
 
-        // 3. Checking chunk height
+        // 1.b Checking chunk height
         let chunk_requested = self.requested_partial_encoded_chunks.contains_key(&chunk_hash);
         if !chunk_requested {
             if !self.encoded_chunks.height_within_horizon(header.height_created()) {
-                return Err(Error::ChainError(ErrorKind::InvalidChunkHeight.into()));
+                return Ok(ProcessPartialEncodedChunkResult::Known);
             }
             // We shouldn't process unrequested chunk if we have seen one with same (height_created + shard_id)
```
mm-near, line 1387: I didn't  fully undestand this case -- as this partial_encoded_chunk might contain different parts of the chunk that we didn't see yet, right ?

chain/chunks/src/lib.rs
```diff
@@ -1339,20 +1440,78 @@ impl ShardsManager {
...
+                // Let's process those parts now.
+                let forwarded_chunk = PartialEncodedChunkV2 {
+                    header: header.clone(),
+                    parts: forwarded_parts.into_iter().map(|(_, part)| part).collect(),
+                    receipts: Vec::new(),
+                };
+                // Make sure we mark this chunk as being requested so that it is handled
+                // properly in the next call. Note no requests are actually sent at this point.
+                self.request_chunk_single_mark_only(header);
+                return self.process_partial_encoded_chunk(
```
mm-near, line 1460: I'd mention that here we call ourselves recursively, basically "simulating" as that forwarded part was just received from the network

mm-near (changes requested):

chain/chunks/src/lib.rs
mzhangmzz: This change is not relevant to this PR. I'll just remove it to avoid confusion. I just thought this error doesn't make sense and wanted to get rid of it. I'll do that in a different PR

chain/client/src/client.rs
mzhangmzz: You were right. CI was failing because of the hack introduced a bug. I fixed it.

chain/chunks/src/chunk_cache.rs
```diff
@@ -37,6 +40,13 @@ pub struct EncodedChunksCacheEntry {
     pub header: ShardChunkHeader,
     pub parts: HashMap<u64, PartialEncodedChunkPart>,
     pub receipts: HashMap<ShardId, ReceiptProof>,
+    /// whether this entry has all parts and receipts
+    pub complete: bool,
+    /// whether the header has been **fully** validated
+    /// every entry added to the cache already has their header "partially" validated
+    /// by validate_chunk_header. When the previous block is accepted, they must be
+    /// validated again to make sure they are fully validated
```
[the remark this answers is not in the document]
mzhangmzz: in validate_chunk_header

chain/chunks/src/chunk_cache.rs
```diff
@@ -90,27 +110,80 @@ impl EncodedChunksCache {
...
+                .insert(chunk_hash.clone());
+            self.incomplete_chunks
+                .entry(chunk_header.prev_block_hash())
+                .or_default()
+                .insert(chunk_hash.clone());
+        }
         self.encoded_chunks
             .entry(chunk_hash)
-            .or_insert_with(|| EncodedChunksCacheEntry::from_chunk_header(chunk_header.clone()))
+            .or_insert_with(|| EncodedChunksCacheEntry::from_chunk_header(chunk_header))
```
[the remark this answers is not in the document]
mzhangmzz: good point, changed back

chain/chunks/src/lib.rs
```diff
@@ -47,6 +48,82 @@ use near_primitives::epoch_manager::RngSeed;
...
+// Currently, X is set to be 1/3 of total validator seats (num_data_parts).
+//
+// **How chunks are propagated in the network
+// Instead sending the full chunk, chunk content is communicated between nodes through
+// PartialEncodedChunk, which includes the chunk header, some parts and receipts of the chunk.
+// Full chunk can be reconstructed if a node receipts enough chunk parts.
+// A node receives partial encoded chunks in three ways,
+// - by requesting for it and receive a PartialEncodedChunkResponse,
+// - by receiving a PartialEncodedChunk, which is sent from the original chunk producer to the part owners
+// - by receiving a PartialEncodedChunkForward, which is sent from part owners to validators.
```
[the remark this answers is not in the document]
mzhangmzz: That's a good question. I actually want to change that. Right now, part owners forward the part they own to other validators who need the part whenever they receive a partial encoded chunk message on the chunk, the partial encoded chunk message doesn't even need to include the part. That generates a lot of wasteful network messages

chain/chunks/src/lib.rs
bowenwang1996: Yeah that would make more sense

chain/chunks/src/lib.rs
```diff
@@ -47,6 +48,82 @@ use near_primitives::epoch_manager::RngSeed;
...
+//
+// ** Storing chunks
+// Before a chunk can be reconstructed fully, parts and receipts in the chunk are stored in
+// `encoded_chunks`. Full chunks will be persisted in the database storage after they are
+// reconstructed.
+//
+// ** Forwarding chunks
+// To save messages and time for chunks to propagate among validators, we implemented a feature
+// called ForwardChunkParts. When a validator receives a part it owns, it forwards the part to
+// other validators who track the shard through a PartialEncodedChunkForward message. This saves
```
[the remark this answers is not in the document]
mzhangmzz: Which validators track which shard is public information and is part of EpochInfo.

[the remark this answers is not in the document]
mzhangmzz: Or, maybe it is easier to understand this way, which validators are assigned to this shard. If a node decides to track this shard itself, the node wouldn't know.

For example, in the future when we have chunk only producers, if a chunk only producer produces chunk for shard 1 and just configures itself to track shard 2, then this ForwardChunkParts will only forward it parts for chunk 1, but not chunk 2.

chain/chunks/src/lib.rs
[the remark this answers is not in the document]
mzhangmzz: https://github.com/near/nearcore/issues/5885

chain/chunks/src/lib.rs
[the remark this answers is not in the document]
mzhangmzz: https://github.com/near/nearcore/issues/5886

chain/chunks/src/lib.rs
```diff
@@ -47,6 +48,82 @@ use near_primitives::epoch_manager::RngSeed;
...
+// Since blocks only contain chunk headers, full chunks must be communicated separately.
+// For data availability, information in a chunk is divided into parts by Reed Solomon encoding,
+// and each validator holds a subset of parts of each chunk (a validator is called the owner
+// of the parts that they hold). This way, a chunk can be retrieved from any X validators
+// where X is the threshold for the Reed Solomon encoding for retrieving the full information.
+// Currently, X is set to be 1/3 of total validator seats (num_data_parts).
+//
+// **How chunks are propagated in the network
+// Instead sending the full chunk, chunk content is communicated between nodes through
+// PartialEncodedChunk, which includes the chunk header, some parts and receipts of the chunk.
```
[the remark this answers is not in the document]
mzhangmzz: I'm not sure if that information is best put here. I think that is protocol level information, which should probably be put in nomicon or the create readme

chain/chunks/src/lib.rs
```diff
@@ -1442,34 +1639,14 @@ impl ShardsManager {
             self.seals_mgr.approve_chunk(height, &chunk_hash);
 
             self.encoded_chunks.remove_from_cache_if_outside_horizon(&chunk_hash);
+            self.encoded_chunks.mark_entry_complete(&chunk_hash);
             self.requested_partial_encoded_chunks.remove(&chunk_hash);
```
bowenwang1996, line 1643: It seems that we can extract these three lines out into a separate method?

bowenwang1996 (commented):
I have a meta question: given the invasive nature of this change, how could we make sure it does not break something unexpectedly?

mzhangmzz:
> I have a meta question: given the invasive nature of this change, how could we make sure it does not break something unexpectedly?

I will take a better look at existing CI tests for chunk management and add more tests if needed

mzhangmzz:
I thought a little more about testing code on chunks. I think there can be two kinds of tests:
1. These tests can assume data availability for chunks is given, that is, a chunk or the parts a node cares about in a chunk can be retrieved eventually. With this assumption, these tests test that chunks are processed correctly, blocks are processed correctly when chunks are ready. For validators, that means they can produce blocks, but the blocks may have missing chunks. For non-validators, that means they can process blocks, but may with delays.
2. These tests will test availability of chunk parts and full chunks, and will care about the delays in retrieving these parts. These tests will be important to guarantee that blocks can be produced without missing chunks. These tests will need to make some assumption on the network part, for example, the average delay between nodes. 

Implementation wise, 1) are much easier to think about and to implement because we don't need to worry about the network part there. I would also argue most of the existing tests in our codebase falls into that category. Some tests in `integration-tests/tests/client/chunks_management.rs` touched 2) to some extent by testing whether blocks can be produced if some chunks messages are dropped and varying the block production delays, but it is probably not enough. For example, it didn't catch the issue happened on mainnet during the launch that changed validator selection. 

Back to the test plan for this PR, it didn't change any part regarding chunk requesting, so I would argue type 1 tests will provide most of the coverage needed on this particular change. The existing tests in `chunks_maynagement.rs` cover the validator case and I have added a new test `test_processing_chunks_sanity` to cover the non-validator case. The only worry would be that it somehow increase the delay in requesting chunks, which I think it's quite unlikely logically speaking, but is not impossible. That could have two kinds of outcomes, blocks have missing chunks and rpc nodes go out of sync. The blocks have missing chunks issue I think can be covered by mocknet testing, and the rpc nodes go out of sync issue is exactly what this PR is trying to solve. We can deploy this change on some rpc nodes first and monitor the metrics.

I will continue to think more on type 2) tests.

bowenwang1996:
> 2\. These tests will test availability of chunk parts and full chunks, and will care about the delays in retrieving these parts. These tests will be important to guarantee that blocks can be produced without missing chunks. These tests will need to make some assumption on the network part, for example, the average delay between nodes.

It is always difficult to test for performance regression, especially if it is not hermetic. Collecting metrics in large-scale simulation tests and in the real world definitely helps, but I wonder whether there are better ways to do this. @matklad could you help us here :)

chain/chunks/src/chunk_cache.rs
```diff
@@ -90,24 +112,77 @@ impl EncodedChunksCache {
...
+    /// Mark an entry as complete, which means it has all parts and receipts needed
+    pub fn mark_entry_complete(&mut self, chunk_hash: &ChunkHash) {
+        match self.encoded_chunks.entry(chunk_hash.clone()) {
+            Occupied(entry) => {
+                let entry = entry.into_mut();
+                entry.complete = true;
+                let previous_block_hash = entry.header.prev_block_hash();
+                self.remove_incomplete_chunk(previous_block_hash, chunk_hash);
+            }
+            Vacant(_) => {}
```
bowenwang1996, line 124: why is it possible for this to happen? should we have a warning here?

chain/chunks/src/lib.rs
```diff
@@ -1357,20 +1465,108 @@ impl ShardsManager {
         );
         store_update.commit()?;
 
+        // Merge parts and receipts included in the partial encoded chunk into chunk cache
         self.encoded_chunks.merge_in_partial_encoded_chunk(partial_encoded_chunk);
 
-        // Forward my parts to others tracking this chunk's shard
+        // 3. Process the forwarded parts in chunk_forwards_cache
+        match self.chunk_forwards_cache.cache_remove(&chunk_hash) {
```
bowenwang1996, line 1472: `if let` works better here

chain/chunks/src/lib.rs
bowenwang1996: This comment is not addressed :)

chain/client/src/client.rs
```diff
@@ -850,66 +849,64 @@ impl Client {
...
     }
 
+    /// Try to process chunks in the chunk cache whose previous block hash is `prev_block_hash` and
+    /// who are not marked as complete yet
+    pub fn check_incomplete_chunks(&mut self, prev_block_hash: &CryptoHash) -> Vec<AcceptedBlock> {
+        let mut accepted_blocks = vec![];
+        for chunk_header in self.shards_mgr.get_incomplete_chunks(prev_block_hash) {
+            debug!(target:"client", "try to process incomplete chunks {:?}, prev_block: {:?}", chunk_header.chunk_hash(), prev_block_hash);
+            // create a fake partial encoded chunk message with empty parts and receipts
+            // because we just want to reconstruct the chunk
```
bowenwang1996, line 859: does this comment still make sense?

bowenwang1996 (approved):
Approving to unblock, but please address the comments

chain/client/src/client.rs
mzhangmzz: Good point

chain/chunks/src/lib.rs
```diff
@@ -47,6 +48,85 @@ use near_primitives::epoch_manager::RngSeed;
...
+// - by requesting it and receiving a PartialEncodedChunkResponse,
+// - by receiving a PartialEncodedChunk, which is sent from the original chunk producer to the part owners
+//   after the chunk is produced
+// - by receiving a PartialEncodedChunkForward, which is sent from part owners to validators who
+//   track the shard, when a validator first receives a part it owns.
+//   TODO: this is actually not the current behavior. https://github.com/near/nearcore/issues/5886
+// Note that last two messages are only be sent from validators to validators, so the only way a
+// non-validator receives a partial encoded chunk is by requesting for it.
+//
+// ** Requesting for chunks
```
mm-near, line 73: s/for//

chain/chunks/src/lib.rs
```diff
@@ -47,6 +48,85 @@ use near_primitives::epoch_manager::RngSeed;
...
+// yet. The requests are managed at the chunk level, instead of individual parts and receipts.
+// A new request can be added by calling function `request_chunk_single`. If it is not
+// in the pool yet, `request_partial_encoded_chunk` will be called, which checks which parts or
+// receipts are still needed for the chunk by checking `encoded_chunks` (see the section on
+// ** Storing chunks). This way, the node won't send requests for parts and receipts they already have.
+// It then figures out where to request them, either from the original
+// chunk producer, or a block producer or peer who tracks the shard, and sends out the network
+// requests. Check the logic there for details regarding how targets of requests are chosen.
+//
+// Once a request is added the pool, they can be asked to be resent through `resend_chunk_requests`,
```
mm-near, line 84: s/they can be asked to be resent/it can be resent

chain/chunks/src/lib.rs
```diff
@@ -1220,75 +1247,157 @@ impl ShardsManager {
...
+
+        match self.runtime_adapter.verify_chunk_header_signature(header, &epoch_id, &ancestor_hash)
+        {
             Ok(false) => {
                 byzantine_assert!(false);
-                return Err(Error::InvalidChunkSignature);
+                return if epoch_id_confirmed {
+                    Err(Error::InvalidChunkSignature)
+                } else {
+                    // we are not sure if we are using the correct epoch id for validation, so
```
mm-near, line 1329: shouldn't we move the bynzantine_assert into line 1327?

chain/chunks/src/lib.rs
```diff
@@ -1220,75 +1247,157 @@ impl ShardsManager {
...
+                Err(near_chain::Error::from(DBNotFoundErr(
+                    format!("block {:?}", header.prev_block_hash()).to_string(),
+                ))
+                .into())
+            };
         }
+        Ok(())
+    }
+
+    /// Process a partial encoded chunk message, which means
```
mm-near, line 1359: s/Process/Processes

chain/chunks/src/lib.rs
```diff
@@ -1220,75 +1247,157 @@ impl ShardsManager {
...
+    /// 1) Checks that the partial encoded chunk message is valid, including checking
+    ///    header, parts and receipts
+    /// 2) If the chunk message is valid, save the parts and receipts in cache
+    /// 3) Forwards newly received owned parts to other validators, if any
+    /// 4) Processes forwarded chunk parts that haven't been processed yet
+    /// 5) Checks if the chunk has all parts and receipts, if so and if the node cares about the shard,
+    ///    decode and persist the full chunk
```
mm-near, line 1366: s/decode/decodes (same for persists)

chain/chunks/src/lib.rs
```diff
@@ -1357,20 +1465,105 @@ impl ShardsManager {
...
 
-        let entry = self.encoded_chunks.get(&chunk_hash).unwrap();
+        self.check_chunk_have_all_parts_and_receipts(&partial_encoded_chunk.header, chain_store, rs)
+    }
+
+    /// Checks if the chunk has all parts and receipts, if so and if the node cares about the shard,
+    /// decode and persist the full chunk
+    /// `header`: header of the chunk. It must be known by `ShardsManager`, either
+    ///           by previous call to `process_partial_encoded_chunk` or `request_partial_encoded_chunk`
+    pub fn check_chunk_have_all_parts_and_receipts(
```
mm-near, line 1517: for future -- we should refactor/split this method - as the name suggests that 'we're only checking' -- while in practice we're also mutating storage.

chain/chunks/src/lib.rs
```diff
@@ -1357,20 +1465,105 @@ impl ShardsManager {
...
+        &mut self,
+        header: &ShardChunkHeader,
+        chain_store: &mut ChainStore,
+        rs: &mut ReedSolomonWrapper,
+    ) -> Result<ProcessPartialEncodedChunkResult, Error> {
+        // The logic from now on requires previous block is processed because
+        // calculating owner parts requires that
+        let prev_block_hash = header.prev_block_hash();
+        let epoch_id = match self.runtime_adapter.get_epoch_id_from_prev_block(&prev_block_hash) {
```
mm-near, line 1526: seems that we're already doing it before calling (in line 1495) -- maybe we should change this method to receive epoch_id in arguments?

mm-near (approved):

chain/chunks/src/lib.rs
mzhangmzz: It has to be checked in this function because it is not just for getting epoch_id, it is also to check the previous block has been processed by the chain. The functions `has_all_parts` requires that

chain/chunks/src/lib.rs
mzhangmzz: I renamed the function to try_process_chunk_parts_and_receipts

matklad (approved):
:facepalm: seems like I forgot to submit a review a did a while ago :(

matklad:
Don't have super concrete thoughts about testing guarantees here. 

One general thought (learned that from https://lamport.azurewebsites.net/video/videos.html or maybe https://lamport.azurewebsites.net/tla/paxos-algorithm.html) is that, to test availability and other network-related bits, you want to represent "the network" as a vector of messages which were ever send: `Vec<Message>`. In tests, rather then directly messing up with how `P1` sends a message to `P2`, you'd just have `P1` to enque its message onto this network vector and have some "network demon" decide in which order the messages are delivered 

If we want to test availability, we need to add explicit logical timestamps to the messages and constrain the "network demon" to always deliver messages within certain logical time delay. This should at least tell us if our algorithm is mathematically unsound.

Not sure what's the proper way to test availability on the intergration-test level, with real network, wall-clock timeouts and such. My hunch would be to read on how https://jepsen.io actually works under the hood (it uses real time and real network IIRC), and then use that.

PR #5749, opened 2021-12-10
fix: Fix race condition in tests related `open_port` 

pmnoxx:
While stress testing `archival_node` test I found out that there is a rare condition, which can cause test to fail.

In that test we all `open_port` function. It's possible that the test will crash, if any other service starts using given port, in between `open_port` is called, and the time `PeerManagerActor` calls `started()`.

Let's fix the race condition for `integration-tests/tests/network/src/runner.rs`

Something to consider:
- maybe fix other places that use `open_port`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2874,6 +2874,7 @@ name = "near-network"
 version = "0.0.0"
 dependencies = [
  "actix",
+ "anyhow",
  "bencher",
  "borsh 0.9.1",
  "bytes",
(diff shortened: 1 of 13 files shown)


pmnoxx:
```
thread panicked while panicking. aborting.
error: test failed, to rerun pass '-p near-jsonrpc-tests --test rpc_query'

Caused by:
  process didn't exit successfully: `/home/pmnoxx/repos/nearcore.master/target/debug/deps/rpc_query-1ac6b9974f6dae21` (signal: 4, SIGILL: illegal instruction)
```

pmnoxx:
Closing this PR. It become outdated, needs to be rewritten.

PR #5754, opened 2021-12-10
re: Disable yarn test

pmnoxx:
Yarn test keeps failing.

Let's disable it for now. This has been happening for a while.

Here is another failure.
https://buildkite.com/nearprotocol/nearcore/builds/9750#834c14d9-c161-45c7-b4e1-4acae3b92a13

My proposal is to move "yarn test" to nayduck.


chefsale (approved):

bowenwang1996:
@chefsale @pmnoxx this is our only safe guard to make sure that near-api-js does not break with nearcore changes. We should not disable this test. If the test is flaky, we should fix it. cc @mm-near

PR #5758, opened 2021-12-10
feat: Add test to `ThrottledFramedRead`

pmnoxx:
Lets add tests to `ThrottledFramedRead`.

For, some reason, the test times out. Any ideas?

### DIFF
--- a/utils/near-rate-limiter/src/framed_read.rs
+++ b/utils/near-rate-limiter/src/framed_read.rs
@@ -310,10 +310,21 @@ where
 
 #[cfg(test)]
 mod tests {
-    use crate::ThrottleController;
+    use crate::{ThrottleController, ThrottledFrameRead};
+    use bytes::{Buf, BytesMut};
+    use futures_core::Stream;
+    use std::error::Error;
+    use std::pin::Pin;
+    use std::ptr::null;
     use std::sync::atomic::Ordering::SeqCst;
     use std::sync::Arc;
+    use std::task::{Context, Poll, RawWaker, RawWakerVTable, Waker};
+    use std::time::Duration;
+    use tokio::io::AsyncWriteExt;
+    use tokio::net::{TcpListener, TcpStream};
     use tokio::sync::Semaphore;
+    use tokio::time::Instant;
+    use tokio_util::codec::Decoder;
     use tokio_util::sync::PollSemaphore;
 
     #[tokio::test]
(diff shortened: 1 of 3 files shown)


pmnoxx:
@matklad Can you take a look? Maybe you can figure out why I can't get this test to work.

utils/near-rate-limiter/src/framed_read.rs
```diff
@@ -413,4 +424,92 @@ mod tests {
...
+
+            match Pin::new(&mut read).poll_next(&mut context) {
+                Poll::Pending => {
+                    // this expected
+                }
+                Poll::Ready(x) => {
+                    assert!(false, "got {:?}", x);
+                    break;
+                }
+            }
```
matklad, line 498: What a conundrum! The bug here is that this loop doesn't contain await. So it never yields to the executor and we get a deadlock. Effectively, we have a cooperative multitasking situation, where task a is blocked on task b, but itself runns a `loop {}`, so task `b` never gets a chance to run.

matklad: A minimal, hacky fix is to insert `tokio::task::yield_now().await;`. A proper fix is to add a dev-dependency on `tokio-stream` and:

```rust
use tokio_stream::StreamExt;

let msg = read.next().await?;
```

instead of a manual poll-loop.

utils/near-rate-limiter/src/framed_read.rs
```diff
@@ -413,4 +424,92 @@ mod tests {
         assert_eq!(throttle_controller.consume_bandwidth_used(), 1015);
         assert_eq!(throttle_controller.consume_bandwidth_used(), 0);
     }
+
```
matklad, line 427: Unrelated to the PR, but rather than `ready!(PollSemaphore::poll_next(Pin::new(pinned.semaphore), cx));` we should do  `ready!(pinned.semaphore.poll_acquire(cx));` in the code above.

matklad (approved):

matklad:
Also, maybe I am missing something, but why we pass a semaphore in

```rust
    /// Creates a new `NearFramedRead` with the given `decoder`.
    pub fn new(
        inner: T,
        decoder: D,
        throttle_controller: ThrottleController,
        semaphore: PollSemaphore,
    ) -> ThrottledFrameRead<T, D> {
```

?

`throttle_controller` already has a semaphore, no?

pmnoxx:
> Also, maybe I am missing something, but why we pass a semaphore in
> 
> ```rust
>     /// Creates a new `NearFramedRead` with the given `decoder`.
>     pub fn new(
>         inner: T,
>         decoder: D,
>         throttle_controller: ThrottleController,
>         semaphore: PollSemaphore,
>     ) -> ThrottledFrameRead<T, D> {
> ```
> 
> ?
> 
> `throttle_controller` already has a semaphore, no?

`ThrottleController` is increasing the counter in `PollSemaphore`

`ThrottledFrameRead`, read from it, and decreases it.

utils/near-rate-limiter/src/framed_read.rs
pmnoxx: @matklad Thanks, it works!.

PR #5760, opened 2021-12-10
re: Fix `unnecessary 'unsafe' block` warnings

pmnoxx:
While reading through our CI logs i noticed this warnings:
```
warning: unnecessary `unsafe` block
--
  | --> src/lib.rs:7:5
  | \|
  | 7 \|     unsafe { core::arch::wasm32::unreachable() }
  | \|     ^^^^^^ unnecessary `unsafe` block
  | \|
  | = note: `#[warn(unused_unsafe)]` on by default
  |  
```

We can fix it, by removing `unsafe`.


matklad:
Context: this fn became safe relatively recently, that's why we didn't see this before

https://github.com/rust-lang/stdarch/pull/1196

matklad (approved):

PR #5761, opened 2021-12-10
fix(chain): fix block ordinal computation

bowenwang1996:
In #4793 we unfortunately did not fix the issue fully. The situation that we did not fix is as follows: if block h is a fork and block h+1 is on the canonical chain, they would have the same ordinal and if block h is received later than block h+1, it would overwrite block h+1 in `ColBlockOrdinal`, which is incorrect.

This PR fixes the issue completely by updating `ColBlockOrdinal` in a similar way to how `ColBlockHeight` is updated. Essentially we only update it when the header head change. When the header head changes, we overwrite the ordinals between the last header height and the new header head. This way we guarantee that only hashes of blocks on the canonical chain eventually stay in the column given how the fork choice rule works.

Test plan
--------------------------
`test_block_merkle_proof`


chain/chain/src/store.rs
```diff
@@ -1745,12 +1741,13 @@ impl<'a> ChainStoreUpdate<'a> {
...
         block_merkle_tree: PartialMerkleTree,
     ) {
-        self.chain_store_cache_update
-            .block_ordinal_to_hash
-            .insert(block_merkle_tree.size(), block_hash);
         self.chain_store_cache_update.block_merkle_tree.insert(block_hash, block_merkle_tree);
     }
 
+    pub fn save_block_ordinal(&mut self, block_hash: CryptoHash, block_ordinal: NumBlocks) {
+        self.chain_store_cache_update.block_ordinal_to_hash.insert(block_ordinal, block_hash);
```
mfornet, line 1748: I'm not very familiar with the code. Is this enough so that this change is ultimately reflected on the database?

mfornet (commented):
If possible, let's try to verify every block/ordinal is properly computed and update all affected values.

chain/chain/src/store.rs
bowenwang1996: Yes. It is saved when we call `ChainStoreUpdate::commit`

bowenwang1996:
@mfornet I added the DB migration. Also I changed it to a better implementation (essentially mimicking what we do to store the block height -> block hash mapping). This way we do not have to deal with some corner cases.

integration-tests/tests/client/process_blocks.rs
```diff
@@ -1923,11 +1924,10 @@ fn test_incorrect_validator_key_produce_block() {
     assert!(matches!(res, Ok(None)));
 }
 
-fn test_block_merkle_proof_with_len(n: NumBlocks) {
+fn test_block_merkle_proof_with_len(n: NumBlocks, rng: &mut StdRng) {
```
matklad, line 1927: :+1:

integration-tests/tests/client/process_blocks.rs
```diff
@@ -1949,8 +1954,23 @@ fn test_block_merkle_proof_with_len(n: NumBlocks) {
             cur_height += 1;
         }
     }
+
     let head = blocks.pop().unwrap();
     let root = head.header().block_merkle_root();
+    // verify that block ordinal to block hash map is correct
+    for h in 0..head.header().height() {
+        if let Ok(block) = env.clients[0].chain.get_block_by_height(h).map(Clone::clone) {
```
matklad, line 1962:
```suggestion
        if let Ok(block) = env.clients[0].chain.get_block_by_height(h).cloned() {
```

matklad (approved):
Can/should we add an assert somewhere checking that we never overwrite the ordinal?

mfornet:
> Can/should we add an assert somewhere checking that we never overwrite the ordinal?

Sometimes it is ok to rewrite ordinals, if they were part of a fork. We must make sure not to overwrite ordinals from finalised blocks. 

The other option is not to write ordinals until the block becomes final, and then making sure they are not overwritten.

matklad:
>The other option is not to write ordinals until the block becomes final

I was under the impression that this is exactly what this PR intends to do? To quote Bowen, 

> This PR fixes the issue completely by only updating ColBlockOrdinal for finalized blocks.

bowenwang1996:
@matklad sorry I forgot to update the description. Thanks for catching it!

integration-tests/tests/client/process_blocks.rs
```diff
@@ -1949,8 +1954,23 @@ fn test_block_merkle_proof_with_len(n: NumBlocks) {
             cur_height += 1;
         }
     }
+
     let head = blocks.pop().unwrap();
     let root = head.header().block_merkle_root();
+    // verify that block ordinal to block hash map is correct
```
mm-near, line 1960: what is "block hash map" ? I thought that we're comparing with ColBlockMerkleTree ?

nearcore/src/migrations.rs
```diff
@@ -478,6 +479,28 @@ pub fn migrate_24_to_25(path: &Path) {
     set_store_version(&store, 25);
 }
 
+/// Fix an issue with block ordinal (#5761)
+pub fn migrate_30_to_31(path: &Path, near_config: &NearConfig) {
+    let store = create_store(path);
+    if near_config.client_config.archive && &near_config.genesis.config.chain_id == "mainnet" {
```
mm-near, line 485: The fact that we're running it "only on mainnet" - means that this code will not be executed in testnet etc.

What I'd do is:
 - remove this condition (so that code runs on testnet / betanet etc)
 - in 491 -> say that for "mainnet" we start from block 47443088, while for other networks, we go from block 1

nearcore/src/migrations.rs
```diff
@@ -478,6 +479,28 @@ pub fn migrate_24_to_25(path: &Path) {
...
+        let genesis_height = near_config.genesis.config.genesis_height;
+        let mut chain_store = ChainStore::new(store.clone(), genesis_height);
+        let head = chain_store.head().unwrap();
+        let mut store_update = BatchedStoreUpdate::new(&store, 10_000_000);
+        // we manually checked mainnet archival data and the first block where the discrepancy happened is `47443088`.
+        for height in 47443088..=head.height {
+            if let Ok(block_hash) = chain_store.get_block_hash_by_height(height) {
+                let block_ordinal = chain_store.get_block_merkle_tree(&block_hash).unwrap().size();
+                store_update
```
mm-near, line 494: could we read the previous value and see if the new value differs ?

So that we could print the total amount of changed rows in the end ? 

(while this is not stricktly needed - it would help with debugging - and we should see the correct result when we run this code on the canary node)

nearcore/src/migrations.rs
```diff
@@ -478,6 +479,28 @@ pub fn migrate_24_to_25(path: &Path) {
     set_store_version(&store, 25);
 }
 
+/// Fix an issue with block ordinal (#5761)
+pub fn migrate_30_to_31(path: &Path, near_config: &NearConfig) {
```
mm-near, line 483: Maybe let's add the comment with "expected amount of time" such migration is going to take ?

nearcore/src/migrations.rs
```diff
@@ -478,6 +479,28 @@ pub fn migrate_24_to_25(path: &Path) {
...
+        let genesis_height = near_config.genesis.config.genesis_height;
+        let mut chain_store = ChainStore::new(store.clone(), genesis_height);
+        let head = chain_store.head().unwrap();
+        let mut store_update = BatchedStoreUpdate::new(&store, 10_000_000);
+        // we manually checked mainnet archival data and the first block where the discrepancy happened is `47443088`.
+        for height in 47443088..=head.height {
+            if let Ok(block_hash) = chain_store.get_block_hash_by_height(height) {
+                let block_ordinal = chain_store.get_block_merkle_tree(&block_hash).unwrap().size();
+                store_update
+                    .set_ser(DBCol::ColBlockOrdinal, &index_to_bytes(block_ordinal), &block_hash)
```
mm-near, line 495: do we have any test case that is executing this code ?

chain/chain/src/store.rs
```diff
@@ -3221,8 +3223,8 @@ mod tests {
             let mut store_update = chain.mut_store().store_update();
             store_update.save_block(block.clone());
             store_update.inc_block_refcount(block.header().prev_hash()).unwrap();
-            store_update.save_head(&Tip::from_header(block.header())).unwrap();
             store_update.save_block_header(block.header().clone()).unwrap();
```
mm-near, line 3226: so the ordering here is important ? was it a bug before ?

mm-near: should we add something (in separate PR) to enforce the proper ordering ?

mm-near (changes requested):

nearcore/src/migrations.rs
```diff
@@ -478,6 +479,28 @@ pub fn migrate_24_to_25(path: &Path) {
...
 
+/// Fix an issue with block ordinal (#5761)
+pub fn migrate_30_to_31(path: &Path, near_config: &NearConfig) {
+    let store = create_store(path);
+    if near_config.client_config.archive && &near_config.genesis.config.chain_id == "mainnet" {
+        let genesis_height = near_config.genesis.config.genesis_height;
+        let mut chain_store = ChainStore::new(store.clone(), genesis_height);
+        let head = chain_store.head().unwrap();
+        let mut store_update = BatchedStoreUpdate::new(&store, 10_000_000);
+        // we manually checked mainnet archival data and the first block where the discrepancy happened is `47443088`.
```
mm-near, line 490: did we verify every archival node that exists ? (as AFAIK this bug was not deterministic)

integration-tests/tests/client/process_blocks.rs
bowenwang1996: it is [block ordinal to block hash] map. Will rephrase it to make it easier to parse

chain/chain/src/store.rs
bowenwang1996: this is a piece of testing code

nearcore/src/migrations.rs
bowenwang1996: We verified it on multiple of our archival nodes and they all have the same issue.

nearcore/src/migrations.rs
bowenwang1996: > in 491 -> say that for "mainnet" we start from block 47443088, while for other networks, we go from block 1

I think this is debatable. Given that the check on mainnet (going from genesis block) took about 27 hours, I am not sure whether we want to do it on testnet as that will incur a significant amount of downtime for an archival node. At the same time, I manually tested this code on a mainnet archival node and it works fine (confirmed by Aurora).

@mfornet could you let us know whether it is important to run migration for testnet as well?

bowenwang1996:
Update: I ran the PR on a mainnet archival node. The migration was done successfully and the Aurora team confirmed that the fix works well.

nearcore/src/migrations.rs
mfornet: So far we haven't seen this issue on Testnet, but it is hard to know if the issue is present for some blocks in the past.
If it is the case it doesn't affect future txs, so from the POV of the bridge, running the migration in testnet is not necessary.

sept-en: @bowenwang1996 Yes, we haven't experienced that issue on Testnet so far, therefore, I don't think this is required to run a migration there.

N.B.: I will let you know if we encounter this issue there.

nearcore/src/migrations.rs
bowenwang1996: tested it manually on a mainnet archival node

integration-tests/tests/client/process_blocks.rs
bowenwang1996: This is a `Result`, so `cloned` doesn't work :(

nearcore/src/migrations.rs
bowenwang1996: This is difficult because it is a moving target. The time it took to migrate the node I used to test is not necessarily reflective of the time it will take to migrate a node when this change is released. Still, I added a note to document the amount of time it takes now.

mm-near (approved):

nearcore/src/migrations.rs
nemani: Is there a way to prevent migration if I am using https://github.com/aurora-is-near/partner-relayer-deploy/  and starting a new node from scratch ?

Issue #5767, opened 2021-12-11
`validatorrs` RPC errors out with UNKNOWN_EPOCH when on genesis block


mina86:
When no blocks have been generated (i.e. latest block is genesis block at height zero), sending `GetValidatorInfo { epoch_reference: EpochReference::Latest }` request to view client results in `UnknownEpoch`. This leads to spurious `restaked.py` failures. Should the view client or JSON RPC handle genesis block in some way?

PR #5768, opened 2021-12-11
tools: restaked: retry when getting UNKNOWN_EPOCH

mina86:
Request for validators information may fail with UNKNOWN_EPOCH even
when asking about the state at the latest epoch (which, presumably,
should be known to the node).

Make the `restaked` binary retry when that error is encountered (since
it eventually starts to work) rather than failing.

This attempts to mitigate some random restaked.py test failures.

Issue: https://github.com/near/nearcore/issues/5767

### DIFF
--- a/tools/restaked/src/main.rs
+++ b/tools/restaked/src/main.rs
@@ -3,7 +3,7 @@ use std::sync::Arc;
 use std::time::Duration;
 
 use clap::{App, Arg};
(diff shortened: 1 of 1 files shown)


mina86:
Actually, let’s close this for now. The issue happens when no blocks are generated (only genesis block). I’ll prepare a different fix.

PR #5769, opened 2021-12-11
chore: keep fuzz tests independent from workspace

miraclx:
I suppose this brings up a question: do we want fuzz tests as part of the workspace, or as independent units? So far, most of them are separate, except two. This PR removes those two from the workspace manifest. Relegating it to self-contained units.

What's the typical convention around fuzz tests? Should they all be independent or all be part of the workspace?


miraclx:
Also, it's worth noting that excluding these from the workspace means crate compliance checks https://github.com/near/nearcore/pull/5724 won't extend to these crates. Which isn't great.. For one, it seems the default `author` for new fuzz crates is `"Automatically generated"`. Ideally, this should always be the NEAR author.

mina86:
They fuzz tests were added to the workspace to avoid compilation failures. Without them in the workspace there’s nothing compiling the code in CI. This, IIRC, has caused two NayDuck failures in the past.

miraclx:
In this PR, I ensured both are compilable. Including the corresponding lockfiles. CI also seems to pass.

mina86:
I don’t have an opinion whether it should be part of workspace or not, but yes either all fuzz tests should be part of the workspace or none of them should and in the second case they probably should be compiled explicitly in CI.

miraclx:
Great then. I'll add checks for it. But wait for others to weigh in on this before we settle.

matklad:
What's the benefit of doing this? 

The benefit of the current approach (fuzz targets are part of the workpace) is that we have only single lockffile. This in turn has two benefits: 

* we fuzz the code we ship to the validators, rathre than some subtly different set of packages
* we have only one lockfile to worry about in terms of maintenance burden.

miraclx:
That's all well and good, and I agree. But then, as aforementioned, we should have some measure of uniformity. So, all other fuzz tests should in-turn be part of the workspace. Not just these two.

matklad:
Yeah, than we should move in the other direction and add other fuzz tests to the ws. While we are at it, another good  thing to do would be to actually run thouse other fuzz tests on CI: 

https://github.com/near/nearcore/blob/4f18d3b6a9c44f2cc95874d0bd33643cf2ab811e/nightly/fuzzing.txt#L1-L3

miraclx:
Closing in favor of https://github.com/near/nearcore/pull/5810

PR #5775, opened 2021-12-11
pytest: restaked: wait for at least one block to be generated

mina86:
This reduces flakiness of the `sanity/restaked.py` test though doesn’t
fix it completely.  It resolves failures caused by `UNKNOWN_EPOCH`
error such as:

    thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value:
    "RpcError { error_struct: Some(HandlerError(Object({\"name\": String(\"UNKNOWN_EPOCH\")}))),
                code: -32000,
                message: \"Server error\",
                data: Some(String(\"Unknown Epoch\")) }"', tools/restaked/src/main.rs:86:48

There’s still an issue where restake tool will send `StakeAction`
which is never executed.  This can be observed by there being only two
attempts at staking in:

    INFO: Starting restaked for /home/mpn/.near/test0, rpc = 0.0.0.0:3041
    INFO  restaked] Sending staking transaction test0 -> 50000000000000000000000000000000
    INFO  restaked] Sending staking transaction test0 -> 50000000000000000000000000000000
      File "/home/mpn/code/near/nearcore/pytest/tests/sanity/restaked.py", line 79, in <module>
        assert False, "Validator test0 is not in current validators after 5 epochs"
    AssertionError: Validator test0 is not in current validators after 5 epochs

Fixes: https://github.com/near/nearcore/issues/5767


pytest/tests/sanity/restaked.py
```diff
@@ -53,12 +56,16 @@ def start_restaked(node_dir, rpc_port, config):
 
 started = time.time()
 
+# Wait until at least one block is generated since otherwise we will get
+# UnknownEpoch errors inside of restaked binary when asking for validators info.
+while not (b := get_height(nodes[1])):
```
mm-near, line 61: I think we have a library util for that:
wait_for_blocks_or_timeout

pytest/tests/sanity/restaked.py
```diff
@@ -53,12 +56,16 @@ def start_restaked(node_dir, rpc_port, config):
...
+    time.sleep(0.1)
+
 restaked_pid = start_restaked(nodes[0].node_dir, nodes[1].rpc_port, {})
 nodes[0].kill()
 
 while time.time() - started < TIMEOUT:
-    status = nodes[1].get_status()
-    height = status['sync_info']['latest_block_height']
+    height = get_height(nodes[1])
```
mm-near, line 68: same here - we could use wait_for_blocks_or_timeout(EPOCH_LENGHT*5 + 5)

mm-near (changes requested):

pytest/tests/sanity/restaked.py
mina86: Yes and no. `wait_for_blocks_or_timeout` waits for given number of new blocks to appear. So if there were already 5 blocks produced and we call it to wait for one block it will wait for 6th block. However, here we just want for there to be at least on block and if there’s one ready that works for us.

Anyway, could we get this merged while I’m preparing a larger refactor for wait for blocks and some other code.

mina86:
@mm-near, PTAL

mina86:
@mm-near, friendly ping.

mina86:
@mm-near, friendly ping.

mina86:
@mm-near, friendly ping.

mm-near (approved):

Issue #4372, opened 2021-06-15
Remove or rename neard testnet command


austinabell:
As pointed out by [https://github.com/near/nearcore/pull/4356/files#r649480263](https://github.com/near/nearcore/pull/4356/files#r649480263), this command can be confusing that it is starting a local network rather than joining the testnet.

The options are:

*   Remove the subcommand
*   Rename it to be clear that it is initializing a new network (possibly `localnet`?)

So the direction should be chosen before it's picked up, but this
 should be a good first issue/bootcamp issue.

—

*   <span class="error">[x]</span> Add `localnet` command as an alias to `testnet`.
*   [ ] Some time in 2022Q2 remove `testnet` command.

PR #5778, opened 2021-12-12
chore: rename neard testnet subcommand to localnet

### LINKED ISSUE #4374: Fuzzing the entire runtime

webern:
The neard `testnet` subcommand had a confusing name because it starts
a local network instead of joining testnet.  Deprecate it and
introduce a new `localnet` to make it more clear what it is doing.

Issue: https://github.com/near/nearcore/issues/4372


webern:
Looks like CI depends on the subcommand. I'll track that down.

webern:
Interesting, some forward and backward compatibility testing would seem to make changing this command's API more difficult than it otherwise appeared. I'll mark this as a draft and learn a little bit more about how forward and backward compatibility testing is working.

mina86:
The problem here is that some of the tests download released binaries which run in production. Those will not have the change applied. You need a two-step approach: add ‘localnet’ subcommand while marking ‘testnet’ as deprecated; then, a quarter later, remove ‘testnet’ command.

We may also need to update things in `docs` repository.

neard/src/cli.rs
```diff
@@ -90,10 +94,14 @@ pub(super) enum NeardSubCommand {
...
     #[clap(name = "run")]
     Run(RunCmd),
-    /// Sets up testnet configuration with all necessary files (validator key, node key, genesis
-    /// and config)
+    /// Sets up local configuration with all necessary files (validator key, node key, genesis and
+    /// config)
+    #[clap(name = "localnet")]
+    Localnet(LocalnetCmd),
+    /// DEPRECATED: this command has been renamed to 'localnet' and will be removed in a future
+    /// release.
```
mina86, line 102:
```suggestion
    /// release.
    // TODO(#4372): Deprecated since 1.24.  Delete it in couple of
    // releases in 2022.
```

neard/src/cli.rs
```diff
@@ -10,8 +10,8 @@ use std::{env, fs, io};
 use tracing::debug;
```
mina86, line 10: Might just as well put `debug` into the group as well.

mina86 (approved):

webern:
> webern force-pushed the chore-rename-neard-subcommand branch from 11bd5f6 to 4cdef56

Add TODO comments for future removal of testnet command.

> webern force-pushed the chore-rename-neard-subcommand branch from 4cdef56 to 38d7810

Rebase

neard/src/cli.rs
webern: Done, also added this note to the relevant python test files.

webern:
> webern force-pushed the chore-rename-neard-subcommand branch from 38d7810 to ea76ff4

Group logging imports into a single use statement (except for `error` which is separate for a compile guard).

webern:
> We may also need to update things in `docs` repository.

I searched the docs repository for `neard` and did not see any invocations of the `testnet` subcommand.

I then verified by searching with this regex `neard .*testnet`. There are a few cases where `neard` is invoked with `--chain-id testnet` in the documentation, but I didn't find any cases where the `testnet` subcommand was used.

webern:
> webern force-pushed the chore-rename-neard-subcommand branch from ea76ff4 to 37324a7

Run python formatter as requested by CI automation.

neard/src/cli.rs
mina86: I don’t think it’s really needed in Python tests since removing `testnet` subcommand here will force change in Python code as well but either way is fine.

mina86: I don’t think it’s really needed in Python tests since removing `testnet` subcommand here will force change in Python code as well but either way is fine.

mina86:
@webrn, this now needs review from @bowenwang1996 or @mm-near an will merge automatically.

webern:
> webern force-pushed the chore-rename-neard-subcommand branch from 37324a7 to f86d233

Rebase

webern:
> webern force-pushed the chore-rename-neard-subcommand branch from f86d233 to b4397d5

Move a comment down to be directly above the subcommand it references.

bowenwang1996 (commented):
@mina86 could you submit a nayduck run to verify that this PR doesn't break nayduck?

mina86:
@bowenwang1996 looks good https://nayduck.near.org/#/run/2339

bowenwang1996 (approved):

PR #5783, opened 2021-12-12
re: Remove `From<Vec<u8>>` for `PeerId` and `KnownPeerState`

pmnoxx:
We can remove `impl TryFrom<Vec<u8>> for PeerId` and `impl TryFrom<Vec<u8>> for KnownPeerState`.
This will remove extra memory allocations and make code shorter.


matklad (approved):

chain/network/src/peer_manager/peer_store.rs
```diff
@@ -1,4 +1,4 @@
-use borsh::BorshSerialize;
+use borsh::{BorshDeserialize, BorshSerialize};
```
mm-near, line 1: do you need this ?

mm-near (approved):

chain/network/src/peer_manager/peer_store.rs
pmnoxx: @mm-near Yes, need it for `KnownPeerState::try_from_slice`.

By using `borsh` directly, we can avoid extra memory allocations; creating a temporary vector just to read from it.

PR #5785, opened 2021-12-12
feat: Optimize `gc_col` in `impl<'a> ChainStoreUpdate<'a> {`

pmnoxx:
We are doing memory allocation each time we access the database

Change:
`  fn gc_col(&mut self, col: DBCol, key: Vec<u8>) {`
to
`  fn gc_col(&mut self, col: DBCol, key: &[u8]) {`


pmnoxx:
````
@bowenwang1996 @mm-near @mzhangmzz @yordanmadzhunkov: /chain/
```
@bowenwang1996 @mm-near @mzhangmzz @yordanmadzhunkov Can you take a look?

PR #5786, opened 2021-12-12
re: Rewrite `PeerStore::new` in a functional way

pmnoxx:
`PeerStore::new` can be written in a functional style; it will be much easier to read it.

### DIFF
--- a/chain/network/src/peer_manager/peer_store.rs
+++ b/chain/network/src/peer_manager/peer_store.rs
@@ -3,7 +3,7 @@ use near_network_primitives::types::{
     KnownPeerState, KnownPeerStatus, NetworkConfig, PeerInfo, ReasonForBan,
 };
 use near_primitives::network::PeerId;
-use near_primitives::time::Utc;
+use near_primitives::time::{Clock, Utc};
 use near_primitives::utils::to_timestamp;
 use near_store::{ColPeers, Store};
 use rand::seq::IteratorRandom;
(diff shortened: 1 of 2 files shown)


chain/network/src/peer_manager/peer_store.rs
```diff
@@ -58,7 +58,7 @@ impl PeerStore {
         let mut peer_states = HashMap::default();
```
mm-near, line 58: comment on what these hashmaps are storing

chain/network/src/peer_manager/peer_store.rs
```diff
@@ -85,23 +85,25 @@ impl PeerStore {
...
 
-            if let Some(current_peer_state) = peer_states.get_mut(&peer_id) {
-                // This peer is a boot node and was already added so skip.
-                if peer_state.status.is_banned() {
-                    current_peer_state.status = peer_state.status;
+            match peer_states.entry(peer_id) {
+                Entry::Occupied(mut current_peer_state) => {
+                    // This peer is a boot node and was already added so skip.
+                    if peer_state.status.is_banned() {
+                        current_peer_state.get_mut().status = peer_state.status;
```
mm-near, line 97: comment about this case (I guess this means that peer was banned in the past, right?)

chain/network/src/peer_manager/peer_store.rs
```diff
@@ -85,23 +85,25 @@ impl PeerStore {
...
                 }
-                continue;
-            }
-
-            if let Some(peer_addr) = peer_state.peer_info.addr {
-                if let Entry::Vacant(entry) = addr_peers.entry(peer_addr) {
-                    entry.insert(VerifiedPeer::new(peer_state.peer_info.id.clone()));
-                    peer_states.insert(peer_id, peer_state);
+                Entry::Vacant(entry) => {
+                    if let Some(peer_addr) = peer_state.peer_info.addr {
```
mm-near, line 101: so if 2 peers are at the same address, we're going to add only one of them ? Shouldn't we somehow pick the "newer" one ? or it doesn't matter (as if we pick the old one - it will be marked as unreachable in 1h.. )

chain/network/src/peer_manager/peer_store.rs
```diff
@@ -85,23 +85,25 @@ impl PeerStore {
...
+                        if let Entry::Vacant(entry2) = addr_peers.entry(peer_addr) {
+                            entry2.insert(VerifiedPeer::new(peer_state.peer_info.id.clone()));
```
mm-near, line 103: We're adding it as "verified" peer --> Are we still doing any kind of signature check afterwards ?

mm-near (dismissed):

chain/network/src/peer_manager/peer_store.rs
pmnoxx: @mm-near This doesn't matter for bootnodes, as all of them are considered new.
There should be only one existing in the database.

chain/network/src/peer_manager/peer_store.rs
pmnoxx: We trust that all entries inserted to database, were verified before the insertion.

Yes, when connecting to given address, we will check whenever peer at given address has given `PeerId`.

chain/network/src/peer_manager/peer_store.rs
pmnoxx: There are a few cases:
- multiple peers have the same address in `boot_nodes` - we probably should panic!
- there is an entry for a peer in boot nodes, and it's different from entry in the database.
We currently, prefer the entry in boot_nodes. There is a bug, first_seen, seems to be wrongly updates.
- We assume that database entries are unique.

pmnoxx:
```
```
@mm-near Can you take a look?

pmnoxx:
```
```
@mm-near Can you take a look?

pmnoxx:
```
```
@mm-near Can you take a look?

chain/network/src/peer_manager/peer_store.rs
```diff
@@ -56,56 +56,82 @@ impl PeerStore {
...
-                                KnownPeerState::new(peer_info.clone()),
+                                KnownPeerState::new(peer_info.clone(), now),
                             );
                         }
                     }
                 }
             }
-        }
+            else {
+                // TODO: add panic!, `boot_nodes` list is wrolng
```
mm-near, line 87: typo in wrong.

Any reason why we don't add panic right now? 

(or at least an error message)

chain/network/src/peer_manager/peer_store.rs
```diff
@@ -56,56 +56,82 @@ impl PeerStore {
...
-                    peer_states.insert(peer_id, peer_state);
+                // Peer is not a boot node
+                Entry::Vacant(entry) => {
+                    if let Some(peer_addr) = peer_state.peer_info.addr {
+                        if let Entry::Vacant(entry2) = addr_2_peer.entry(peer_addr) {
+                            // Default case, add new entry.
+                            entry2.insert(VerifiedPeer::new(peer_state.peer_info.id.clone()));
+                            entry.insert(peer_state);
+                        }
+                        // else: There already exists a peer with a different addr, that's a boot node.
```
mm-near, line 128: s/different/same, right?

chain/network/src/peer_manager/peer_store.rs
```diff
@@ -56,56 +56,82 @@ impl PeerStore {
...
-                    current_peer_state.status = peer_state.status;
+            match peerid_2_state.entry(peer_id) {
+                // Peer is a boot node
+                Entry::Occupied(mut current_peer_state) => {
+                    if peer_state.status.is_banned() {
+                        // If it says in database, that peer should be banned, ban the peer.
+                        current_peer_state.get_mut().status = peer_state.status;
+                    }
+                    // Maybe a bug:
+                    // For boot node, it's first_seen = now
```
mm-near, line 116: what is first_seen used for? (from what I've seen in the code, seems that we're never reading this value)

mm-near (dismissed):

PR #5791, opened 2021-12-12
feat: Use `criterion` instead of `bencher`

pmnoxx:
We are currently using `bencher` for bench-marking. We should switch to `criterion`.
In this PR we will switch `near-network` to use `criterion`.

Reasoning:
- `bencher` hasn't been updated since Jan 2018.
- `criterion` provides a statistically correct data on samples
- `criterion` provides various profiling tools, etc:

https://nickb.dev/blog/guidelines-on-benchmarking-and-rust

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2839,7 +2839,6 @@ dependencies = [
  "borsh",
  "bytes",
  "bytesize",
- "cast",
  "conqueue",
  "criterion",
  "deepsize",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -715,6 +715,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "cast"
+version = "0.2.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "cc38c385bfd7e444464011bb24820f40dd1c76bcdfa1b78611cb7c2e5cafab75"
+dependencies = [
+ "rustc_version",
+]
+
+[[package]]
 name = "cc"
 version = "1.0.71"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1024,6 +1033,42 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "81156fece84ab6a9f2afdb109ce3ae577e42b1228441eded99bd77f627953b1a"
 dependencies = [
  "cfg-if 1.0.0",
+]
+
+[[package]]
+name = "criterion"
+version = "0.3.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "1604dafd25fba2fe2d5895a9da139f8dc9b319a5fe5354ca137cbbce4e178d10"
+dependencies = [
+ "atty",
+ "cast",
+ "clap 2.33.3",
+ "criterion-plot",
+ "csv",
+ "itertools",
+ "lazy_static",
+ "num-traits",
+ "oorandom",
+ "plotters",
+ "rayon",
+ "regex",
+ "serde",
+ "serde_cbor",
+ "serde_derive",
+ "serde_json",
+ "tinytemplate",
+ "walkdir",
+]
+
+[[package]]
+name = "criterion-plot"
+version = "0.4.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d00996de9f2f7559f7f4dc286073197f83e92256a59ed395f9aac01fe717da57"
+dependencies = [
+ "cast",
+ "itertools",
 ]
 
 [[package]]
@@ -1738,6 +1783,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "half"
+version = "1.8.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "eabb4a44450da02c90444cf74558da904edde8fb4e9035a9a6a4e15445af0bd7"
+
+[[package]]
 name = "hashbrown"
 version = "0.11.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2063,6 +2114,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "af25a77299a7f711a01975c35a6a424eb6862092cc2d6c72c4ed6cbc56dfc1fa"
 dependencies = [
  "libc",
+]
+
+[[package]]
+name = "js-sys"
+version = "0.3.55"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "7cc9ffccd38c451a86bf13657df244e9c3f37493cce8e5e21e940963777acc84"
+dependencies = [
+ "wasm-bindgen",
 ]
 
 [[package]]
@@ -2776,11 +2836,12 @@ name = "near-network"
 version = "0.0.0"
 dependencies = [
  "actix",
- "bencher",
  "borsh",
  "bytes",
  "bytesize",
+ "cast",
  "conqueue",
+ "criterion",
  "deepsize",
  "delay-detector",
  "futures",
@@ -3355,6 +3416,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "692fcb63b64b1758029e0a96ee63e049ce8c5948587f2f7208df04625e5f6b56"
 
 [[package]]
+name = "oorandom"
+version = "11.1.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "0ab1bc2a289d34bd04a330323ac98a1b4bc82c9d9fcb1e66b63caa84da26b575"
+
+[[package]]
 name = "opaque-debug"
 version = "0.3.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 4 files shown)


pmnoxx:
There is a version conflict:
```
381 │ ╭ rustc_version 0.2.3 registry+https://github.com/rust-lang/crates.io-index
--
  | 382 │ │ rustc_version 0.4.0 registry+https://github.com/rust-lang/crates.io-index
```

matklad:
If criterion makes it optimization work better for you, than sure, let's use it. 

I personally however am very mystified why people like criterion. To me, it feels like it creates way more problems than it solves. It seems like a bit of software which adds a lot of accidental complexity without actually solving any of the essential problems.

As a specific example, I've just followed https://bheisler.github.io/criterion.rs/book/getting_started.html with a fibbonacci example,  and here's the output I am getting in two subsequent runs: 

```console
λ cargo bench
    Finished bench [optimized] target(s) in 0.02s
     Running unittests (target/release/deps/ct-89e26ca72706676c)

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s

     Running unittests (target/release/deps/my_benchmark-326f953b20f97d4d)
WARNING: HTML report generation will become a non-default optional feature in Criterion.rs 0.4.0.
This feature is being moved to cargo-criterion (https://github.com/bheisler/cargo-criterion) and will be optional in a future version of Criterion.rs. To silence this warning, either switch to cargo-criterion or enable the 'html_reports' feature in your Cargo.toml.

Gnuplot not found, using plotters backend
fib 20                  time:   [14.005 us 14.010 us 14.015 us]                    
                        change: [-2.4323% -1.5819% -0.8679%] (p = 0.00 < 0.05)
                        Change within noise threshold.
Found 5 outliers among 100 measurements (5.00%)
  4 (4.00%) high mild
  1 (1.00%) high severe


14:28:27|~/tmp/ct|HEAD⚡?
λ cargo bench
    Finished bench [optimized] target(s) in 0.02s
     Running unittests (target/release/deps/ct-89e26ca72706676c)

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s

     Running unittests (target/release/deps/my_benchmark-326f953b20f97d4d)
WARNING: HTML report generation will become a non-default optional feature in Criterion.rs 0.4.0.
This feature is being moved to cargo-criterion (https://github.com/bheisler/cargo-criterion) and will be optional in a future version of Criterion.rs. To silence this warning, either switch to cargo-criterion or enable the 'html_reports' feature in your Cargo.toml.

Gnuplot not found, using plotters backend
fib 20                  time:   [16.115 us 16.300 us 16.440 us]                    
                        change: [+7.3299% +9.1826% +10.988%] (p = 0.00 < 0.05)
                        Performance has regressed.
```

Things to note:

* it detects performance regression in two identical subsequent runs for fibonacci on a desktop PC (way easier setup for benchmarking than a laptop or shard VM). This makes me completely distrust whatever perf analysis it is doing, and sort-of invalidates raison d'etre of the thing
* there's a useless to me message about HTML
* there's a useless to me message about gnuplot/plotters

The amount of *things* I need to type and use to get the benchmark is unreasonably high: 

```rust
use criterion::{black_box, criterion_group, criterion_main, Criterion};
use ct::fibonacci;

pub fn criterion_benchmark(c: &mut Criterion) {
    c.bench_function("fib 20", |b| b.iter(|| fibonacci(black_box(20))));
}

criterion_group!(benches, criterion_benchmark);
criterion_main!(benches);
```

<hr/>

What I suggest doing instead is writing our own "benchmarking framework", which can be as simple as using `std::time::Instant::elapsed`. I don't believe we are sufficiently optimized for fancy statistical analysis with p-values to matter. Everything else in criterion is just some syntactic sugar and fluff, which doesn't seem like it is necessary. 

I'd stick to something like this: 

```rust
use ct::fibonacci;

fn fibonacchi_bench() {
    let _ = fibonacci(20);
}

fn main() {
    let benches = [("fibonacchi_bench", fibonacchi_bench)];
    for (n, b) in benches {
        let t = std::time::Instant::now();
        b();
        let t = t.elapsed();
        eprintln!("{:<20} {:0.2?}", n, t)
    }
}
```

```
λ cargo bench --bench my_other_bench
    Finished bench [optimized] target(s) in 0.02s
     Running unittests (target/release/deps/my_other_bench-8075b8cfa337f1e1)
fibonacchi_bench     14.33µs
```

pmnoxx:
> If criterion makes optimization work better for you, than sure, let's use it.

Yes, it helps me a lot. 

> I personally however am very mystified why people like criterion. To me, it feels like it creates way more problems than it solves. It seems like a bit of software which adds a lot of accidental complexity without actually solving any of the essential problems.

It solves a few problems:
- by default `bencher` uses a fixed number of iterations, which makes it inaccurate, when measuring performance of small functions. On the other hand, criterion, works for `5s` by default, and generates enough data points to measure performance of both `fast` and `slow` functions.
- In addition. Cpu's have power states. By default, when not used, CPU will lower frequency by 50-70%. That's why it's important to do `warmup`. `criterion` does warmup for `3s` before doing test.
- Proper statistics - `criterion` detects whenever there are outliers, they often indicate that the machine is not configured properly, and the stats can be not reliable
- Configuration - take a look at `cargo-criterion`. There are a lot of useful options there, I can specify how long to run tests, to get enough statistical data. I typically run
 ```
 cargo criterion calculate_distance -- --measurement-time 60
```
- previously while writing `benchmark_write_then_read_successful_10m` in `near-store` benchmark. I had to write a test to write 10m keys each time, in order to get enough statistical data. It's painful to have to rewrite a function each time, and having to modify the code each time, you want to get statistical data. Unfortunately, increasing the number of keys written changes the test and makes it not comparable.
- Scientific notations - calculation of avg, standard deviation, `p` is useful. It lets me know quickly, by looking at one value that my machine is not configured properly.

> As a specific example, I've just followed https://bheisler.github.io/criterion.rs/book/getting_started.html with a fibbonacci example, and here's the output I am getting in two subsequent runs:

First of all ,they are in the middle of doing a migration. It's best to disable those warnings, and the output would be more readable. By adding 2 extra features, see my PR, you can disable warnings.


Here is my output when I ran `cargo criterion` a few times.
```console
L .~/r/n/c/network piotr-switch-to-criterion  for x in `seq 1 1 10`; do nice=-20 taskset -c 0 cargo criterion calculate_distance_3_3; done                        
calculate_distance_3_3  time:   [691.33 ns 691.62 ns 691.96 ns]                                    
                        change: [-0.3816% -0.2926% -0.1973%] (p = 0.00 < 0.05)
                        Change within noise threshold.

calculate_distance_3_3  time:   [694.80 ns 695.25 ns 695.74 ns]                                    
                        change: [+0.3990% +0.4742% +0.5464%] (p = 0.00 < 0.05)
                        Change within noise threshold.

calculate_distance_3_3  time:   [690.91 ns 691.12 ns 691.34 ns]                                    
                        change: [-0.6476% -0.5788% -0.5046%] (p = 0.00 < 0.05)
                        Change within noise threshold.

calculate_distance_3_3  time:   [695.33 ns 695.68 ns 696.05 ns]                                    
                        change: [+0.4244% +0.4994% +0.5717%] (p = 0.00 < 0.05)
                        Change within noise threshold.

calculate_distance_3_3  time:   [693.64 ns 694.04 ns 694.44 ns]                                    
                        change: [-0.2127% -0.1389% -0.0598%] (p = 0.00 < 0.05)
                        Change within noise threshold.
```
I think it did pretty good job.

> Things to note:
> 
> * it detects performance regression in two identical subsequent runs for fibonacci on a desktop PC (way easier setup for benchmarking than a laptop or shard VM). This makes me completely distrust whatever perf analysis it is doing, and sort-of invalidates raison d'etre of the thing

Criterion is statistically accurate, it detects that the conditions have changed. If you want the messages to be 100% reliable, you need to setup you desktop correctly. Otherwise, you need to take into the account the variance that comes from your cpu, running at different speeds at different times.  Anyway, I'm not saying you have to do it, but it's possible. Here are a couple of runs with a properly configured system:
```
L .~/r/n/c/network piotr-switch-to-criterion  nice=-20 taskset -c 0-3 cargo criterion calculate_distance_3_3 -- --measurement-time 60
    Finished bench [optimized + debuginfo] target(s) in 0.14s
calculate_distance_3_3  time:   [694.44 ns 694.69 ns 694.93 ns]                                   
                        change: [+0.0712% +0.1402% +0.2079%] (p = 0.00 < 0.05)
                        Change within noise threshold.

L .~/r/n/c/network piotr-switch-to-criterion  nice=-20 taskset -c 0-3 cargo criterion calculate_distance_3_3 -- --measurement-time 60
    Finished bench [optimized + debuginfo] target(s) in 0.14s
calculate_distance_3_3  time:   [692.21 ns 692.55 ns 692.93 ns]                                   
                        change: [-0.2432% -0.1653% -0.0946%] (p = 0.00 < 0.05)
                        Change within noise threshold.

```

> * there's a useless to me message about HTML
> * there's a useless to me message about gnuplot/plotters

I agree about the HTML, and `gnuplot/plotters`. They are in the process of removing it from `criterion`.
Hence the warning, it will not be generated by default in version `0.4.0`.




> The amount of _things_ I need to type and use to get the benchmark is unreasonably high:
> 
> ```rust
> use criterion::{black_box, criterion_group, criterion_main, Criterion};
> use ct::fibonacci;
> 
> pub fn criterion_benchmark(c: &mut Criterion) {
>     c.bench_function("fib 20", |b| b.iter(|| fibonacci(black_box(20))));
> }
> 
> criterion_group!(benches, criterion_benchmark);
> criterion_main!(benches);
> ```
> 
> What I suggest doing instead is writing our own "benchmarking framework", which can be as simple as using `std::time::Instant::elapsed`. I don't believe we are sufficiently optimized for fancy statistical analysis with p-values to matter. Everything else in criterion is just some syntactic sugar and fluff, which doesn't seem like it is necessary.
> 
> I'd stick to something like this:
> 
> ```rust
> use ct::fibonacci;
> 
> fn fibonacchi_bench() {
>     let _ = fibonacci(20);
> }
> 
> fn main() {
>     let benches = [("fibonacchi_bench", fibonacchi_bench)];
>     for (n, b) in benches {
>         let t = std::time::Instant::now();
>         b();
>         let t = t.elapsed();
>         eprintln!("{:<20} {:0.2?}", n, t)
>     }
> }
> ```
> 
> ```
> λ cargo bench --bench my_other_bench
>     Finished bench [optimized] target(s) in 0.02s
>      Running unittests (target/release/deps/my_other_bench-8075b8cfa337f1e1)
> fibonacchi_bench     14.33µs
> ```

I would argue, that the first example is shorter. Also, having to write a new main each time for each test adds up. Instead `criterion` provides command line arguments, for example:
- specify time to run
- specify number of iterations
- specify warmup time
- compares results between previous runs, I don't need to take notes myself - the time saving isn't much, but still,


Here are a few advanced options you can do, to make benchmark easier. Although, not always necessary, it saves a bit of time debugging.
- set up cpu frequency - you can do it easily on linux with `cpupower-gui` / or in bios / or some other tools on windows
- most laptops have bioses  / controls locked / though my be do able
- cpu doesn't throttle due to power
- doesn't throttle due to temperature
- on AMD - programs run on one CCX
- hype-threading can also affect benchmarking

Anyway, usually we want to have big performance gains, when we optimize performance by 2x, 10x, or some cases 4000x, like routing table computation, So, `bench` is kind of `ok` for that purpose. 

I agree, we can write our own tools, I wrote testing tool before, with properly calculated statistics. However, I'm not sure if it's worth spending a few days writing it, we would end up doing that `criterion` already has. 

As for, html, `gnuplot`, it's not useful for my by itself. I prefer to look at the raw data. However, it's a useful tool for doing presentations, and/or showing other people results of your work. It should be disabled by default, and they already committed to doing just that in `0.4.0` https://github.com/bheisler/criterion.rs/pull/496

matklad (approved):

pmnoxx:
```
```

pmnoxx:
```
```

pmnoxx:
```
```

mm-near (approved):

PR #5793, opened 2021-12-13
feat: `near-primitives` Improve `combine_hash` performance by `40%`.

pmnoxx:
`combine_hash` in `core/primitives/src/merkle.rs` can be written in a cleaner way, that also avoid memory allocations:

```
pub fn combine_hash(hash1: &MerkleHash, hash2: &MerkleHash) -> MerkleHash {
    CryptoHash::hash_borsh(&(hash1, hash2))
 }
```

> I did some benchmarking.
> 
> Before any changes
> 
> ```
> test combine_hash_bench  ... bench:         143 ns/iter (+/- 1)
> ```
> 
> After not using Vec::new() and making arguments use references
> 
> ```
> test combine_hash_bench  ... bench:         102 ns/iter (+/- 1)
> ```

Btw, `MerkleHash` is the same thing as `CryptoHash`. Using references instead of doing memory copies, has some performance advantages.

If we don't change the signature and run the code like this
```
pub fn combine_hash2(hash1: MerkleHash, hash2: MerkleHash) -> MerkleHash {
    CryptoHash::hash_borsh(&(hash1, hash2))
}
```
We get performance result of:
```
test combine_hash_bench  ... bench:         123 ns/iter (+/- 1)
```


core/primitives/src/merkle.rs
```diff
@@ -22,10 +22,8 @@ pub enum Direction {
     Right,
 }
 
-pub fn combine_hash(hash1: MerkleHash, hash2: MerkleHash) -> MerkleHash {
-    let mut combined: Vec<u8> = hash1.into();
-    combined.append(&mut hash2.into());
-    hash(&combined)
+pub fn combine_hash(hash1: &MerkleHash, hash2: &MerkleHash) -> MerkleHash {
```
mm-near, line 25: did you verify that it generates the same results as before? can you add a small test here, that would compute the combine_hash - and compare to a fixed result (and make sure that this result is the same before and after your PR)

mm-near (changes requested):

matklad (approved):
Yeah, would be great to confirm that some test fails if we actually mess up the impl.

core/primitives/src/merkle.rs
pmnoxx: @mm-near Yes, I did. I included a test in this PR.

pmnoxx:
> Yeah, would be great to confirm that some test fails if we actually mess up the impl.

@matklad I added a test for this method, it works fine.

pmnoxx:
Before
```
test combine_hash_bench  ... bench:         143 ns/iter (+/- 1)
```
After
```
test combine_hash_bench  ... bench:         102 ns/iter (+/- 1)
```

pmnoxx:
> I did some benchmarking.
> 
> Before any changes
> 
> ```
> test combine_hash_bench  ... bench:         143 ns/iter (+/- 1)
> ```
> 
> After not using Vec::new() and making arguments use references
> 
> ```
> test combine_hash_bench  ... bench:         102 ns/iter (+/- 1)
> ```

If we don't change the signature and run the code like this
```
pub fn combine_hash2(hash1: MerkleHash, hash2: MerkleHash) -> MerkleHash {
    CryptoHash::hash_borsh(&(hash1, hash2))
}
```
We get performance result of:
```
test combine_hash_bench  ... bench:         123 ns/iter (+/- 1)
```

pmnoxx:
@bowenwang1996 @mm-near It may be worth adding to readme, a not that 
`cargo test` needs `4-8gb of memory per thread`.

pmnoxx:
```
@bowenwang1996 @mm-near @mzhangmzz: /chain/
```
@bowenwang1996 @mm-near @mzhangmzz Can you take a look?

pmnoxx:
```
@bowenwang1996 @mm-near @mzhangmzz: chain/
```
@bowenwang1996 @mm-near @mzhangmzz Can you take a look?

pmnoxx:
```
@bowenwang1996 @mm-near @mzhangmzz: chain/
```
@bowenwang1996 @mm-near @mzhangmzz Can you take a look?

bowenwang1996 (approved):
Thanks!

mzhangmzz (approved):

mm-near (approved):

PR #5794, opened 2021-12-13
re: Rewrite `compute_result` in `near-network::Graph`

pmnoxx:
We can rewrite `compute_result` to be written in a more functional style.
This will improve performance by 50%.

### DIFF
--- a/chain/network/src/routing/routing.rs
+++ b/chain/network/src/routing/routing.rs
@@ -470,9 +470,16 @@ impl Graph {
             }
         }
 
+        // This takes 75% of the total time computation time of this function.
         self.compute_result(&mut routes, &distance)
     }
 
+    /// Converts representation of the result, from an array representation, to
+    /// a hashmap of PeerId -> Vec<PeerIds>
+    /// Arguments:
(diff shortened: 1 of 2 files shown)


chain/network/src/routing/routing.rs
```diff
@@ -516,13 +516,13 @@ impl Graph {
             {
```
mm-near, line 516: can you comment on what "compute_results" does ? and what are the arguments.

chain/network/src/routing/routing.rs
```diff
@@ -516,13 +516,13 @@ impl Graph {
...
                 continue;
             }
-            let mut peer_set: Vec<PeerId> = Vec::with_capacity(cur_route.count_ones() as usize);
-
-            for (id, &neighbor) in neighbors.iter().enumerate().take(MAX_NUM_PEERS) {
-                if (cur_route & (1u128 << id)) != 0 {
-                    peer_set.push(self.id2p[neighbor as usize].clone());
-                };
-            }
+            let peer_set = neighbors
```
mm-near, line 519: can you comment on what we're collecting here

mm-near (approved):

PR #5795, opened 2021-12-13
Fix crash in `near-network` benches

pmnoxx:
Due to recent changes. `chain/network/benches/routing_table_actor.rs` started crashing while `RoutingTableActor` due to the fact, that now it starts `EdgeValidatorActor` and `actix` System has not been running.

### DIFF
--- a/chain/network/benches/routing_table_actor.rs
+++ b/chain/network/benches/routing_table_actor.rs
@@ -1,5 +1,6 @@
 #[macro_use]
 extern crate bencher;
+extern crate actix;
 
 use bencher::{black_box, Bencher};
 use near_crypto::{KeyType, SecretKey, Signature};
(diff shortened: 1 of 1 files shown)


chain/network/benches/routing_table_actor.rs
```diff
@@ -1,18 +1,20 @@
...
-
 use near_network::test_utils::random_peer_id;
 use near_network::RoutingTableActor;
 use near_primitives::network::PeerId;
 use near_store::test_utils::create_test_store;
+use std::collections::HashMap;
+use std::sync::Arc;
 
 fn build_graph(depth: usize, size: usize) -> RoutingTableActor {
+    let _system = actix::System::new();
```
mm-near, line 16: add a comment why we need this here.

mm-near (approved):

Issue #5757, opened 2021-12-10
`near-api-js test` causes nondeterministic failures in CI.


pmnoxx:
Yarn test keeps failing. 

We have disabled it in https://github.com/near/nearcore/pull/5754

Here is an example of CI failure, caused by it.
https://buildkite.com/nearprotocol/nearcore/builds/9750#834c14d9-c161-45c7-b4e1-4acae3b92a13

We should either:
- Move it to `nayduck` and run it there. It may still fail, at least it won't block CI.
- Figure out why this test is flunky and fix it.

Related cleanup:
- #5796 
- #5797

bowenwang1996:
https://github.com/near/nearcore/pull/5754#issuecomment-991772399

bowenwang1996:
@pmnoxx @mm-near let's fix the flaky test and re-enable near-api-js tests. I agree that we should have a discussion about how in the future downstream libraries should test against nearcore changes, but for now it is important that we avoid breaking near-api-js which is the dominant client side library.

pmnoxx:
> @pmnoxx @mm-near let's fix the flaky test and re-enable near-api-js tests. I agree that we should have a discussion about how in the future downstream libraries should test against nearcore changes, but for now it is important that we avoid breaking near-api-js which is the dominant client side library.

@bowenwang1996  I'm glad we are in agreement, that the CI shouldn't have nondeterministic failures.

The first step to fixing those failures is to be try to reproduce them on localhost.

Running the following command
```shell
scripts/test_nearlib.sh
```
results in a failure
```
+ export RUST_BACKTRACE=full
+ RUST_BACKTRACE=full
+ export NEAR_PID=180029
+ NEAR_PID=180029
+ trap 'pkill -15 -P $NEAR_PID' 0
+ ./scripts/start_unittest.py --local
+ rm -rf near-api-js near-shell create-near-app
+ git clone https://github.com/near/near-api-js.git
Cloning into 'near-api-js'...
Starting unittest nodes with test.near account and seed key of alice.near
cargo 1.56.0 (4ed5d137b 2021-10-04)
   Compiling state-viewer v0.0.0 (/home/pmnoxx/repos/nearcore.master/test-utils/state-viewer)
remote: Enumerating objects: 24099, done.
remote: Counting objects: 100% (1845/1845), done.
remote: Compressing objects: 100% (797/797), done.
remote: Total 24099 (delta 1321), reused 1451 (delta 1040), pack-reused 22254
[...]

PR #5796, opened 2021-12-13
re: Make `test_nearlib.sh` work from any directory

pmnoxx:
`scripts/test_nearlib.sh` currently doesn't work on localhost.
Before, we fix `yarn test`, let's make this script work from any directory.

See #5757

### DIFF
--- a/scripts/test_nearlib.sh
+++ b/scripts/test_nearlib.sh
@@ -2,31 +2,47 @@
 set -ex
 
 ./scripts/requirements_check.sh
+repo_dir=$(realpath -e "$(dirname "${BASH_SOURCE[0]}")/..")
 
-export RUST_BACKTRACE=full
-./scripts/start_unittest.py --local &
+# Builds and starts `neard`, assumes it's running in ${repo_dir}'
+# May fail silently, leading to hard to debug CI failures
+# TODO: add option to build only and run checks, this will allow us to see those errors synchronously.
+
+# TODO: This test leaves `near` process running even after the test is over
+# pkill near || echo "Nothing to kill"
+(cd "${repo_dir}"; RUST_BACKTRACE=full "${repo_dir}"/scripts/start_unittest.py --local) &
 export NEAR_PID=$!
 trap 'pkill -15 -P $NEAR_PID' 0
 
-#./scripts/build_wasm.sh
+# "${repo_dir}"/scripts/build_wasm.sh
 
-rm -rf near-api-js near-shell create-near-app
-git clone https://github.com/near/near-api-js.git
-git clone https://github.com/near/near-shell.git
-git clone https://github.com/near/create-near-app.git
+rm -rf "${repo_dir}"/near-api-js "${repo_dir}"/near-shell "${repo_dir}"/create-near-app
+git clone https://github.com/near/near-api-js.git "${repo_dir}"/near-api-js
+git clone https://github.com/near/near-shell.git "${repo_dir}"/near-shell
+git clone https://github.com/near/create-near-app.git "${repo_dir}"/create-near-app
 
 # Make sure to use local nearcore for tests
 export NODE_ENV=local
-
-SRC_DIR=$(pwd)/$(dirname "${BASH_SOURCE[0]}")
-export HOME=$SRC_DIR/../testdir
+export HOME="{$repo_dir}"/testdir
 
 # Run near-api-js tests
-cd near-api-js
+pushd "${repo_dir}/near-api-js"
+
+# TODO(CREATE ISSUE) We should add a new file with all requirement checks.
+if [[ $(yarn --version) != "1."* ]]; then
+  echo "You version of yarn is too old $(yarn --version) < 1.0"
+  echo "Install with npm install --global yarn"
+  exit 1
+fi
+
 yarn
 yarn build
-../scripts/waitonserver.sh
-# Disabling yarn test for now
+yarn list
+popd
+
+"${repo_dir}"/scripts/waitonserver.sh
+
+# TODO(#5757) Disabling yarn test for now
 # yarn test
 # yarn doc
 


scripts/test_nearlib.sh
```diff
@@ -1,12 +1,16 @@
 #!/usr/bin/env bash
 set -ex
 
+SRC_DIR=$(pwd)/$(dirname "${BASH_SOURCE[0]}")
+export REPO_DIR=$(readlink -e "$SRC_DIR/..")
+pushd "${REPO_DIR}"
```
mina86, line 6: This is really confusing. First of all, no need to `export`, just `repo_dir=whatever`. Second of all, why `pushd`? Just `cd`. And lastly, the name resolution doesn’t seem correct. I think what you want is:

    repo_dir=$(realpath -e "$(dirname "${BASH_SOURCE[0]}")/..")

scripts/test_nearlib.sh
```diff
@@ -1,12 +1,16 @@
...
+
 export RUST_BACKTRACE=full
-./scripts/start_unittest.py --local &
+"${REPO_DIR}"/scripts/start_unittest.py --local &
```
mina86, line 9: That’s not necessary (here and in other places). With `cd` you’ve changed directory to repo dir so `./scripts/whatever` is fine.

scripts/test_nearlib.sh
```diff
@@ -16,14 +20,13 @@ git clone https://github.com/near/create-near-app.git
 # Make sure to use local nearcore for tests
 export NODE_ENV=local
 
-SRC_DIR=$(pwd)/$(dirname "${BASH_SOURCE[0]}")
-export HOME=$SRC_DIR/../testdir
+export HOME=$(readlink -e "$REPO_DIR/../testdir")
 
 # Run near-api-js tests
-cd near-api-js
+pushd "${REPO_DIR}"/near-api-js
```
mina86, line 26: Again, `cd` is fine and also this change isn’t necessary.

pmnoxx: Ok, I changed it back to `cd`.

This makes reading the code easier, as I don't need to follow, in which current directory we are at the moment.

scripts/test_nearlib.sh
pmnoxx: In different places of this script, we are either in `${repo_dir}` then in some other directory. Using absolute path, make it less error prone.

scripts/test_nearlib.sh
pmnoxx: `push` / `popd`, show which commands need to be executed in a specific directory.

This is a good example:
```
# Run near-api-js tests
pushd "${repo_dir}"/near-api-js
yarn
yarn build
"${repo_dir}"/scripts/waitonserver.sh
popd
```

mina86: I’d just `cd` each time. I’d say that’s more readable than having to maintain directory stack in your head when you read the script.

pmnoxx:
It turns out I had wrong version of `yarn` installed, I added a check. Now I have to fix the following issue
```
                                                                                             
                                                                                                                                                                                       
yarn install v1.22.17                                                                                                                                                                                              
[1/4] Resolving packages...                                                                                                                                                                                        
⠁                                                                                                                                                                                                                  
[2/4] Fetching packages...
[----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------] 0/844
error husky@7.0.4: The engine "node" is incompatible with this module. Expected version ">=12". Got "10.19.0"
error Found incompatible module.
info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
+ pkill -15 -P 91134

```

pmnoxx:
Added following to ignore issues with dependencies of `near-api-js`
```
yarn config set ignore-engines true
```

scripts/test_nearlib.sh
```diff
@@ -1,30 +1,47 @@
...
 
-export RUST_BACKTRACE=full
-./scripts/start_unittest.py --local &
+repo_dir=$(realpath -e "$(dirname "${BASH_SOURCE[0]}")/..")
+
+# Builds and starts `neard`, assumes it's running in ${repo_dir}'
+# May fail silently, leading to hard to debug CI failures
+# TODO: add option to build only and run checks, this will allow us to see those errors synchronously.
+
+pkill neard || echo "Nothing to kill"
```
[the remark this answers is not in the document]
pmnoxx: @mina86 This test is starting `neard` process.
We don't even kill it properly.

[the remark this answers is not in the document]
pmnoxx: @mina86 
I run this test, and after the test was over, "neard" still kept running.

I agree, this test should be rewritten, to do this in a better way.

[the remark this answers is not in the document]
pmnoxx: @mina86 Actually, it's `near`, not `neard`, which kills running after the test is over.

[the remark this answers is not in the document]
pmnoxx: Anyway, added TODO. Killing that process belongs to another PR.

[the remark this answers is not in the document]
mina86: It is starting it, but `pkill neard` will kill all processes called `neard` running on the machine including unrelated processes that user started.

The killing is supposed to happen on line 14 with `trap 'pkill -15 -P $NEAR_PID' 0`. One potential issue is that that will kill the Python script but won’t necessarily cause `neard` started by that process to die.

pmnoxx:
````
@chefsale @mhalambek @mina86 @nikurt: /scripts/
```
@chefsale @mhalambek @mina86 @nikurt Can you take a look?

nikurt (approved):

scripts/test_nearlib.sh
```diff
@@ -2,31 +2,47 @@
...
 ./scripts/requirements_check.sh
+repo_dir=$(realpath -e "$(dirname "${BASH_SOURCE[0]}")/..")
 
-export RUST_BACKTRACE=full
-./scripts/start_unittest.py --local &
+# Builds and starts `neard`, assumes it's running in ${repo_dir}'
+# May fail silently, leading to hard to debug CI failures
+# TODO: add option to build only and run checks, this will allow us to see those errors synchronously.
+
+# TODO: This test leaves `near` process running even after the test is over
```
mina86, line 11: `neard` process

scripts/test_nearlib.sh
```diff
@@ -2,31 +2,47 @@
...
+# pkill near || echo "Nothing to kill"
+(cd "${repo_dir}"; RUST_BACKTRACE=full "${repo_dir}"/scripts/start_unittest.py --local) &
```
mina86, line 13:
```
(cd "${repo_dir}" && RUST_BACKTRACE=full ./scripts/start_unittest.py --local) &
```
`cd` followed by the use of full path is rather redundant.

scripts/test_nearlib.sh
```diff
@@ -2,31 +2,47 @@
...
+repo_dir=$(realpath -e "$(dirname "${BASH_SOURCE[0]}")/..")
 
-export RUST_BACKTRACE=full
-./scripts/start_unittest.py --local &
+# Builds and starts `neard`, assumes it's running in ${repo_dir}'
+# May fail silently, leading to hard to debug CI failures
+# TODO: add option to build only and run checks, this will allow us to see those errors synchronously.
+
+# TODO: This test leaves `near` process running even after the test is over
+# pkill near || echo "Nothing to kill"
```
mina86, line 12: This probably should be rewritten in Python and then we can handle killing child processes with no issues.

scripts/test_nearlib.sh
```diff
@@ -2,31 +2,47 @@
...
 # Make sure to use local nearcore for tests
 export NODE_ENV=local
-
-SRC_DIR=$(pwd)/$(dirname "${BASH_SOURCE[0]}")
-export HOME=$SRC_DIR/../testdir
+export HOME="{$repo_dir}"/testdir
 
 # Run near-api-js tests
-cd near-api-js
+pushd "${repo_dir}/near-api-js"
```
mina86, line 29: I still think `pushd` is kinda pointless. `cd` is enough.

mina86 (approved):

PR #5797, opened 2021-12-13
re: Improve handling errors in `scripts/test_nearlib.sh`.

pmnoxx:
We are missing a check whenever `NEAR_PID` is defined in `scripts/test_nearlib.sh`.

It's better to add those checks ahead of the time, that will make debugging easier in case an issue happens.

### DIFF
--- a/scripts/waitonserver.sh
+++ b/scripts/waitonserver.sh
@@ -1,12 +1,7 @@
 #!/usr/bin/env bash
 set -e
 
-if [[ -z "${NEAR_PID}" ]]; then
-  echo "NEAR_PID is undefined"
-  exit 1
-fi
-
-echo "waiting on healthcheck NEAR_PID = ${NEAR_PID}" >&2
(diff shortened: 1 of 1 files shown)


scripts/waitonserver.sh
```diff
@@ -1,10 +1,16 @@
 #!/usr/bin/env bash
+set -e
 
-echo 'waiting on healthcheck' >&2
+if [[ -z "${NEAR_PID}" ]]; then
+  echo "NEAR_PID is undefined"
+  exit 1
+fi
+
+echo "waiting on healthcheck NEAR_PID = ${NEAR_PID}" >&2
```
mina86, line 9: This whole thing can be written as

    echo "waiting on health check; NEAR_PID = ${NEAR_PID:?undefined}" >&2

mina86 (approved):

scripts/waitonserver.sh
pmnoxx: @mina86 Thanks, I like it. That's a much better way to do it.

Issue #5800, opened 2021-12-13
Pull Request template is not showing up while creating pull request


Rishabhraghwendra18:
When you make pull request , you will not see the Pull Request template as discussed [here](https://github.com/near/nearcore/blob/master/CONTRIBUTING.md#submitting-the-pr) . It discuss about the format of the PR which I think should be visible while creating a PR to remind the contributor to follow the template.

bowenwang1996:
It is a known github limitation. I believe you cannot have more than one PR templates and still expect them to get triggered automatically when you create a PR.

Rishabhraghwendra18:
ok.. I don't know about that github has this kind of limitation . The current PR was not up-to-date and as per docs it had some other kind of PR template so I thought of making a separate file for this . If you thought that we could remove the old PR template then I would removed that .

No problem , I will close this issue

PR #5801, opened 2021-12-13
doc: Added `pull_request_template.md`

Rishabhraghwendra18:
issue: #5800

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE/pull_request_template.md
+++ b/.github/PULL_REQUEST_TEMPLATE/pull_request_template.md
@@ -1,0 +1,25 @@
+
+<!--
+
+Please prefix the title with following:
+
+Here are some examples of good titles:
+
+- refactor: Created `PartId` struct
+- fix: crash in near-network benches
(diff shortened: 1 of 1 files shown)


matklad:
I am not sure we are quite ready for this. PR and issue templates solve the problem of educating a stream of new contributors, not yet familiar with the code base. In the context of near, we do not yet have a firehose of contributors, so we can just talk to people and explain them how things are done internally. 

Dually, if do add a template, it risks becoming yet another bit of documentation which is outdated. It will also add friction to recurrent contributors, as they'd need to remove the text every time they create a PR.

Rishabhraghwendra18:
> I am not sure we are quite ready for this. PR and issue templates solve the problem of educating a stream of new contributors, not yet familiar with the code base. In the context of near, we do not yet have a firehose of contributors, so we can just talk to people and explain them how things are done internally.
> 
> Dually, if do add a template, it risks becoming yet another bit of documentation which is outdated. It will also add friction to recurrent contributors, as they'd need to remove the text every time they create a PR.

Ok.. no problem Aleksey . I will close the issue

PR #5802, opened 2021-12-13
feat: stabilize nonce range for implicit accounts

Longarithm:
Stabilize https://github.com/near/nearcore/pull/5482. Also, update changelog for 1.23.0 release.

## Reasoning

Previously, we changed formulas for tx nonce computation to avoid tx hash collisions: https://github.com/near/nearcore/issues/3779 
Later we figured out that it was not enabled for implicit accounts - on its creation, we simply created new access key with nonce 0.
So we need this to make formula consistent among all accounts, and to get rid of tx hash collisions completely. The fix itself is straightforward.

## Testing 

Existing tests.

PR #5803, opened 2021-12-13
refactor: drop dead code

matklad:
This field is unused. We still near `memory` to add it as a `memory`
import, but we don't need to store it in the env as well.

This didn't surface prior to #5587, as the struct was marked public and
compiled didn't realise it is actually unused.


matklad:
Not sure how this passed CI before -- I thought we used to deny warning on CI?

Longarithm:
This currently blocks https://github.com/near/nearcore/pull/5802.
Not sure why other PRs were not affected.

PR #5806, opened 2021-12-13
feat: single-threaded apply_range

Longarithm:
We recently observed that running multi-threaded apply range increases time of receipt processing up to 10x and make it less stable. So I suggest to add an option to run it in a single thread for making more precise estimations.

I considered using `RAYON_NUM_THREADS`, but it may be not a good idea because it disables rayon multithreading in all places, though we need to replicate only sequential blocks application.

**Note**: the big diff is caused only by indentation change, "Hide whitespace" makes it better. Real change is ~10 lines.

## Testing

Existing tests.


test-utils/state-viewer/src/apply_chain_range.rs
```diff
@@ -104,155 +104,190 @@ pub fn apply_chain_range(
 
     let all_blocks = end_height - start_height;
 
-    (start_height..=end_height).into_par_iter().for_each(|height| {
-            let mut chain_store = ChainStore::new(store.clone(), genesis.config.genesis_height);
-            let block_hash = match chain_store.get_block_hash_by_height(height) {
-                Ok(block_hash) => block_hash,
+    let process_height = |height| {
```
nikurt, line 107: Would you mind turning this closure into a function?

test-utils/state-viewer/src/apply_chain_range.rs
```diff
@@ -262,18 +297,45 @@ pub fn apply_chain_range(
...
+        inc_and_report_progress(
+            &processed_blocks_cnt,
+            &processed_blocks_timestamp,
+            all_blocks,
+            &skipped_blocks,
+        );
+    };
+    let range = (start_height..=end_height);
+
+    if cfg!(feature = "single_threaded") {
```
nikurt, line 334: It seems much simpler to pass a flag at runtime instead of recompiling with a different feature.

Longarithm: I was thinking that we may want to run more functions with such option, and passing it everywhere would become a routine.

But it's not the case now, so you are right.

nikurt (approved):
Leaving the compile-time-feature vs runtime-command-line-flag question up to you :)

PR #5808, opened 2021-12-13
scripts: fix `neard` binary name in `nodelib.py`

mina86:
The binary has been renamed from `near` to `neard` a long while ago
and the former is now deprecated.  Use new name.


mina86:
@nikurt, friendly ping

nikurt (approved):

Issue #5688, opened 2021-12-06
Evaluate the threshold used for kicking a validator node who didn't produce enough chunks


janewang:
A validator has reported that they have produced all the blocks, but was kicked because NotEnoughChunks

**Expected behavior**
Evaluate if the threshold used to kick a validator node for missing chunks is set to the appropriate number

**Screenshots**
<img width="1196" alt="Screen Shot 2021-12-06 at 3 17 55 PM" src="https://user-images.githubusercontent.com/734336/144925318-4b975780-8330-449c-82be-24ac9678d2b2.png">

**Version (please complete the following information): 1.23.0-rc1
- nearcore 1.23.0-rc.1
- testnet

bowenwang1996:
@EgorKulikov @nikurt please take a look

janewang:
Btw, if we are kicking of validators for missing chunks, we should probably alert. 

Some validators are complaining that they are confused when they are kicked:

> Thanks for the information, because I'm trying to figure it out and can not understand what the cause, the server is fine, blocks are produced (99-100% per epoch), but the node still kicked out.
 > same here. Just happened again. Without any of my alerts being triggered

nikurt:
I can see that validator was missing chunks both before and after the upgrade.
Some epochs have only a few chunks missing and don't result in kicking the validator.
90% Kickout threshold still makes sense to me.

```
$ ./neard view_state epoch_info --validator-account-id namdokmai.pool.f863973.m0 protocol_version 48 | grep 'Validator ' -A 3

epochs-48:Validator namdokmai.pool.f863973.m0 kickout: NotEnoughChunks {
epochs-48-    produced: 716,
epochs-48-    expected: 860,
epochs-48-}
--
epochs-48:Validator namdokmai.pool.f863973.m0 kickout: NotEnoughBlocks {
epochs-48-    produced: 110,
epochs-48-    expected: 216,
epochs-48-}
```

After upgrade:

```
$ ./neard view_state epoch_info --validator-account-id namdokmai.pool.f863973.m0 protocol_version 49 | grep 'Validator ' -A 3

epochs-49:Validator namdokmai.pool.f863973.m0 kickout: NotEnoughChunks {
epochs-49-    produced: 62,
epochs-49-    expected: 75,
epochs-49-}
--
epochs-49:Validator namdokmai.pool.f863973.m0 kickout: NotEnoughChunks {
epochs-49-    produced: 72,
epochs-49-    expected: 83,
epochs-49-}
--
[...]

PR #5809, opened 2021-12-13
Inspect epochs with state-viewer

nikurt:
Currently there is no way to directly inspect EpochInfos in the RocksDB.
This PR lets state-viewer make scans and queries to the EpochInfos column, and also lets us inspect blocks and chunks produced and expected by a given validator (#5688)

Will provide an example output of the tool soon.


test-utils/state-viewer/src/epoch_info.rs
```diff
@@ -0,0 +1,213 @@
...
+    epoch_manager: &mut EpochManager,
+    runtime_adapter: Arc<dyn RuntimeAdapter>,
+) {
+    println!("{:?}: {:#?}", epoch_id, epoch_info);
+    if epoch_info.epoch_height() >= *head_epoch_height {
+        println!("Epoch information for this epoch is not yet available, skipping.");
+        return;
+    }
+    if let Some(account_id) = validator_account_id.clone() {
+        if let Some(kickout) = epoch_info.validator_kickout().get(&account_id) {
```
mm-near, line 167: optional: I'd refactor 167- 212 into a separate method (that displays info from specific validator point of view)

test-utils/state-viewer/src/epoch_info.rs
```diff
@@ -0,0 +1,213 @@
...
+                    let block = chain_store.get_block(&block_hash).unwrap();
+                    if block.chunks()[shard_id as usize].height_included() != block_height {
+                        missing_chunks.push((block_height, shard_id));
+                    }
+                } else {
+                    missing_chunks.push((block_height, shard_id));
+                }
+            }
+            println!("Missing {} chunks: {:?}", missing_chunks.len(), missing_chunks);
+        }
```
mm-near, line 211: else --> print("validator didn't validate in this epoch")

mm-near (approved):

pmnoxx (approved):
LGTM

test-utils/state-viewer/src/cli.rs
```diff
@@ -243,3 +247,40 @@ impl DumpAccountStorageCmd {
         );
     }
 }
+#[derive(Clap)]
+pub struct EpochInfoCmd {
+    /// Fetch the given epoch.
```
pmnoxx, line 252: I looked at `get_epoch_ids`, it looks like those options are exclusive.

Even if you specify `epoch_id`, `epoch_height` option will be ignored.
It look to me like, all those options should belong to a group and be mutually exlusive.

test-utils/state-viewer/src/commands.rs
```diff
@@ -490,6 +491,45 @@ pub(crate) fn check_block_chunk_existence(store: Arc<Store>, near_config: NearCo
     }
     println!("Block check succeed");
```
pmnoxx, line 492: Although, this is not part of this PR. It may be  a good idea to switch to using tracing, rather than !println!

test-utils/state-viewer/src/epoch_info.rs
```diff
@@ -0,0 +1,213 @@
...
+use near_store::{DBCol, Store};
+use std::sync::Arc;
+
+pub(crate) fn print_epoch_info(
+    epoch_id: Option<EpochId>,
+    epoch_height: Option<EpochHeight>,
+    block_hash: Option<CryptoHash>,
+    block_height: Option<BlockHeight>,
+    protocol_version: Option<ProtocolVersion>,
+    validator_account_id: Option<AccountId>,
```
pmnoxx, line 19: maybe, should be enum, when I read `get_epoch_ids`, I realized, it only makes sense for one of those options to be specified.

PR #5810, opened 2021-12-13
chore: ensures all fuzz tests are part of the workspace

miraclx:
Context: https://github.com/near/nearcore/pull/5769


khorolets (approved):

Cargo.lock
```diff
@@ -4361,6 +4426,15 @@ dependencies = [
  "serde_json",
 ]
 
+[[package]]
+name = "rust-base58"
+version = "0.0.4"
```
matklad, line 4431: Where dose this come from? Seems like a weird dependency.

miraclx: `near-jsonrpc-fuzz`: https://github.com/near/nearcore/blob/376791e6783aa72cfd93665770a9fa865a04df69/chain/jsonrpc/fuzz/fuzz_targets/fuzz_target_1.rs#L126

matklad: We use `bs58` elsewhere, let's update the fuzz tests to use that as well?

mina86:
By the way, we might want to have a `scripts/check_fuzz.py` or something which like `check_pytest.py` and `check_nightly.py` would check if all fuzz tests are included in nightly.

pmnoxx (approved):
LGTM

matklad (approved):

PR #5812, opened 2021-12-13
test: add fuzz testing infra for account-id + jsonrpc

miraclx:
https://github.com/near/nearcore/pull/5810 ensures all fuzz testing units in the repo are part of the workspace. 

This ensures to run those fuzz tests with nayduck - https://github.com/near/nearcore/pull/5769#issuecomment-992437092

Not sure how the `pytest/tests` folder is meant to be organized, but I put these units under a new `nearcore` folder.

cc: @matklad


mina86:
LGTM.

> Not sure how the pytest/tests folder is meant to be organized, but I put these units under a new nearcore folder.

No one is. Maybe it would be better to put all fuzz tests into `pytest/tests/fuzz`?

pytest/tests/nearcore/fuzz_account_id.py
```diff
@@ -0,0 +1,11 @@
...
+import subprocess
+import sys
+
+sys.path.append('lib')
+
+import cargo_fuzz
+
+if __name__ == '__main__':
+    sys.exit(cargo_fuzz.run('core/account-id/fuzz', 'serde'))
+    sys.exit(cargo_fuzz.run('core/account-id/fuzz', 'borsh'))
```
mina86, line 11: Actually, those need to be separate fuzz scripts.

mina86 (changes requested):

pytest/tests/nearcore/fuzz_account_id.py
miraclx: Right..

nightly/fuzzing.txt
```diff
@@ -1,2 +1,5 @@
-pytest --skip-build --timeout=2h runtime/fuzz_runtime.py
-pytest --skip-build --timeout=2h runtime/fuzz_wasm_vm.py
+pytest --skip-build --timeout=2h fuzz/runtime.py
+pytest --skip-build --timeout=2h fuzz/wasm_vm.py
+pytest --skip-build --timeout=2h fuzz/jsonrpc.py
+pytest --skip-build --timeout=2h fuzz/account_id_serde.py
+pytest --skip-build --timeout=2h fuzz/account_id_borsh.py
```
mina86, line 5: It occurred to me that maybe it’d be better to have a single fuzz test runner and use arguments for each test rather than copying the same script over and over. As in [#5820](https://github.com/near/nearcore/pull/5820).

mina86 (commented):
LGTM, though the one comment is something to consider.

nightly/fuzzing.txt
miraclx: I certainly like yours better. I'll close this in favor of https://github.com/near/nearcore/pull/5820.

PR #5813, opened 2021-12-13
refactor: don't abuse serialization for type erasure

matklad:
vm-logic is "generic" over External implementation. We don't want
however to literally add a type parameter to it, so we use `dyn
External`. This creates a problem that, when a particular external
fails, we need to report the error, but we don't know error's concrete
types. So let's use `Box<dyn Any>` for that, and a bit of down-casting.

The previous impl did roughly equivalent thing, but via
serialize/deserialize route, which is more complicated than what we need
here.

As a result, we eliminate a couple of BorshSerializes and clones.

### DIFF
--- a/runtime/near-vm-errors/src/lib.rs
+++ b/runtime/near-vm-errors/src/lib.rs
@@ -488,30 +488,6 @@ impl<T: Any + Eq + Sized + Send + Sync> AnyEq for T {
     }
 }
 
-pub mod hex_format {
-    use hex::{decode, encode};
-
-    use serde::de;
-    use serde::{Deserialize, Deserializer, Serializer};
-
-    pub fn serialize<S, T>(data: T, serializer: S) -> Result<S::Ok, S::Error>
-    where
-        S: Serializer,
-        T: AsRef<[u8]>,
-    {
-        serializer.serialize_str(&encode(data))
-    }
-
-    pub fn deserialize<'de, D, T>(deserializer: D) -> Result<T, D::Error>
-    where
-        D: Deserializer<'de>,
-        T: From<Vec<u8>>,
-    {
-        let s = String::deserialize(deserializer)?;
-        decode(&s).map_err(|err| de::Error::custom(err.to_string())).map(Into::into)
-    }
-}
-
 #[cfg(test)]
 mod tests {
     use crate::{CompilationError, FunctionCallError, MethodResolveError, PrepareError, VMError};
--- a/runtime/near-vm-errors/src/lib.rs
+++ b/runtime/near-vm-errors/src/lib.rs
@@ -1,17 +1,17 @@
 #![doc = include_str!("../README.md")]
 
-use std::fmt::{self, Error, Formatter};
-
 use borsh::{BorshDeserialize, BorshSerialize};
 use near_account_id::AccountId;
 use near_rpc_error_macro::RpcError;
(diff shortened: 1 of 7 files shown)


pmnoxx (approved):
LGTM
Yes, it's good idea to return the error as it is, without doing the serialization if possible.

bowenwang1996 (approved):

runtime/near-vm-errors/src/lib.rs
```diff
@@ -432,27 +432,59 @@ impl std::fmt::Display for HostError {
...
-        decode(&s).map_err(|err| de::Error::custom(err.to_string())).map(Into::into)
+impl Eq for AnyError {}
+
+impl fmt::Debug for AnyError {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        fmt::Debug::fmt(self.any.as_any(), f)
+    }
+}
+
+trait AnyEq: Any + Send + Sync {
```
nagisa, line 470: Does this truly need to be using `Any`? Could `std::error::Error` be utilized? It also supports downcasting and is generally probably more accurately conveys the intent of what's happening here.

runtime/near-vm-errors/src/lib.rs
```diff
@@ -1,17 +1,17 @@
...
-
 use borsh::{BorshDeserialize, BorshSerialize};
 use near_account_id::AccountId;
 use near_rpc_error_macro::RpcError;
 use serde::{Deserialize, Serialize};
+use std::any::Any;
+use std::fmt::{self, Error, Formatter};
 
-#[derive(Debug, Clone, PartialEq, Eq)]
+#[derive(Debug, PartialEq, Eq)]
```
nagisa, line 10: The `PartialEq` and `Eq` impls here don't appear to be used at all. Is there any reason why we need to keep these implementations in place?

Comparing errors for equality (as opposed to matching their structure) in general seems like a pretty weird thing to need.

nagisa: If these are dropped, then the whole `AnyEq` thing could be replaced with just `Box<dyn std::error::Error + Send + Sync>` fields. Which is pretty common way to do this kind of type erasure in Rust.

runtime/near-vm-errors/src/lib.rs
matklad: Good question! 

On the one hand, yes, this could use `std::error::Error`, and that does seem like a better fit (it's an error!). On the other hand, the intention here is subtly different -- we are not going to use this as a dynamic error, the sole purpose here is to downcast this to a concrete repr down the line. In this sense, `Any` is a better, because it does strictly what we need, and nothing more. 

Perhaps more important thing here is practicality -- if I use `Error` here, than I'd have to write `Display` for `ExternalError`, but that would effectively be dead code. 

So, I'd say let's keep this as Any. But thanks for the sugestions -- using Error didn't occure to me!

runtime/near-vm-errors/src/lib.rs
matklad: These are used in our tests :sob: And yeah, this is also problematic, as it, eg, prevents us from using `io::Error`. I do want to fix that (by rewriting our test), but not in this PR.

nagisa (approved):

PR #5815, opened 2021-12-13
refactor: test contract build script $PWD agnostic

nagisa:
The scripts runs a couple of dangerous commands (such as `rm -rf`) with
relative directory arguments. This can result in an unintended data
loss. Make sure the script has $PWD set to an expected location before
doing anything else.


matklad (approved):
Makes sense!

We might also want to just move this logic to Rust via `build.rs` or something. This wasn't too trivial last time I've looked though.

PR #5820, opened 2021-12-13
nightly: add account-id and jsonrpc fuzz targets

mina86:


nightly/fuzzing.txt
```diff
@@ -1,2 +1,5 @@
-pytest --skip-build --timeout=2h runtime/fuzz_runtime.py
-pytest --skip-build --timeout=2h runtime/fuzz_wasm_vm.py
+pytest --skip-build --timeout=2h fuzz.py test-utils/runtime-tester/fuzz runtime-fuzzer
```
mm-near, line 1: 2h timeout ? is that intended ? How long are these tests running ?

mm-near (approved):

nightly/fuzzing.txt
mina86: Yes. The test run as long as we give them. It’s fuzzing so it will continue trying random inputs until failure is found or we kill it.

PR #5823, opened 2021-12-13
feat: dump contracts command in state viewer

Longarithm:
Special tool for dumping all **distinct** contracts to file.

Should be useful to analyze all active contracts.

## Testing

Existing tests.


test-utils/state-viewer/src/commands.rs
```diff
@@ -50,6 +52,37 @@ pub(crate) fn state(home_dir: &Path, near_config: NearConfig, store: Arc<Store>)
...
+    store: Arc<Store>,
+) {
+    let (runtime, state_roots, header) = load_trie(store, home_dir, &near_config);
+    println!("Storage roots are {:?}, block height is {}", state_roots, header.height());
+    let mut distinct_codes: HashMap<Vec<u8>, AccountId> = HashMap::default();
+    for (shard_id, state_root) in state_roots.iter().enumerate() {
+        let trie = runtime.get_trie_for_shard(shard_id as u64, header.prev_hash()).unwrap();
+        let trie = TrieIterator::new(&trie, state_root).unwrap();
+        // Optimization: contract nodes form a contiguous segment in the set of nodes with values, so if we touched a
+        // contract node and then touched a non-contract node, then we can stop iterating.
```
ailisp, line 68: :+1:

test-utils/state-viewer/src/commands.rs
```diff
@@ -50,6 +52,37 @@ pub(crate) fn state(home_dir: &Path, near_config: NearConfig, store: Arc<Store>)
...
+        let mut touched_contract_node = false;
+        for item in trie {
+            let (key, value) = item.unwrap();
+            if is_contract_code_key(&key) {
+                touched_contract_node = true;
+                let account_id = parse_account_id_from_contract_code_key(&key).unwrap();
+                distinct_codes.insert(value.clone(), account_id.clone());
```
ailisp, line 75: do we want to collect all account id of a given code?

test-utils/state-viewer/src/commands.rs
```diff
@@ -50,6 +52,37 @@ pub(crate) fn state(home_dir: &Path, near_config: NearConfig, store: Arc<Store>)
...
+            } else if touched_contract_node {
+                break;
+            }
+        }
+    }
+    let mut codes_to_dump: HashMap<_, _> =
+        distinct_codes.iter().map(|(code, account_id)| (account_id, to_base64(code))).collect();
+    std::fs::write(output, serde_json::to_string_pretty(&codes_to_dump).unwrap());
```
ailisp, line 83: Is that more useful to write contract to separate binary files instead of single json?

ailisp (approved):
Looks great

pmnoxx (approved):
LGTM
Nice! It would be even better if we made `dump_contracts` return `anyhow::Result` and have proper error handling.

nikurt (approved):

test-utils/state-viewer/README.md
```diff
@@ -88,6 +88,18 @@ Flags:
...
+Dump all **distinct** contracts from the current state of the DB to JSON file.
+
+File format is a dictionary which maps account ID to base64-encoded contract code.
+
+As of 13/12/2021, takes 1m to run, size of resulting file is ~100 MB.
+
+Flags:
+
+* `--output` specifies a file to which contracts will be saved.
+
```
matklad, line 102: Unrelated to the PR, but it seems that this info duplicates what should be available as `--help`. I'd probably default to putting most of the into into the doc-comments (which clap uses to generate help messages).

test-utils/state-viewer/src/commands.rs
```diff
@@ -50,6 +52,37 @@ pub(crate) fn state(home_dir: &Path, near_config: NearConfig, store: Arc<Store>)
...
+        // Optimization: contract nodes form a contiguous segment in the set of nodes with values, so if we touched a
+        // contract node and then touched a non-contract node, then we can stop iterating.
+        let mut touched_contract_node = false;
+        for item in trie {
+            let (key, value) = item.unwrap();
+            if is_contract_code_key(&key) {
+                touched_contract_node = true;
+                let account_id = parse_account_id_from_contract_code_key(&key).unwrap();
+                distinct_codes.insert(value.clone(), account_id.clone());
+            } else if touched_contract_node {
```
matklad, line 76: Seems like `!` is missing somewhere, right now the code is equivalent to just breaking at the end of `if is_contract_code_key(&key)` block.

test-utils/state-viewer/src/commands.rs
matklad: Yeah, it's better to output a directory with files named like `alice.near.wasm`.

test-utils/state-viewer/src/commands.rs
```diff
@@ -50,6 +52,37 @@ pub(crate) fn state(home_dir: &Path, near_config: NearConfig, store: Arc<Store>)
     }
 }
 
+pub(crate) fn dump_contracts(
+    output: &Path,
+    home_dir: &Path,
+    near_config: NearConfig,
+    store: Arc<Store>,
+) {
```
matklad, line 60: In general, we try to improve our error reporting, so perhaps it's better to `-> anyhow::Result<()>` here and output useful error messages when the stuff goes wrong?

test-utils/state-viewer/src/commands.rs
matklad, line 83: You probably want to `std::fs::create_dir_all(output.parent())` for convenience.

matklad: better = more convenient for consumer *and* allows for streaming implementation

test-utils/state-viewer/src/commands.rs
matklad: Agree with the overall sentiment! My gut feeling is that `anyhow` is the right default with high probability. 

* It resists over-engineering quite well, as it is a deliberately "sloppy" type everything else converts to
* it actually makes writing the code easier, as `?` is shorter than `.unwrap()`
* At the same time, when you need to add `.context` to the error, the thing is already there .

That is, I refactored enough of stuff to `anyhow`, that now I tend to just always start with it. 

But this is definitelly unrelated to the PR.

pmnoxx: @nikurt @matklad This is a good example, on how we can make `neard` cli command errors more readable.

pmnoxx: At some point in time I realized, that having to read the source code, to figure out what the issue with arguments is. Is a huge waste of time. Usually, I simply want to know, which arguments are wrong, without having to waste time each time reading the code.

test-utils/state-viewer/src/commands.rs
Longarithm: I don't see value in it, however it should be easy to modify it to aggregate total contract usage, or contract id with top usage.

test-utils/state-viewer/src/commands.rs
Longarithm: This is intended. Contracts form a contiguous segment in state, so if we observed non-contract code key, we can break. If there would be `!`, we would break at the first iteration.

mm-near: so we're expecting a bunch of non-contract nodes at the beginning of the iterator? (if yes - then mention it in the code comment)

test-utils/state-viewer/src/commands.rs
```diff
@@ -48,6 +49,48 @@ pub(crate) fn state(home_dir: &Path, near_config: NearConfig, store: Arc<Store>)
...
+        // contract node and then touched a non-contract node, then we can stop iterating.
+        let mut touched_contract_node = false;
+        for (i, item) in trie.enumerate() {
+            let (key, value) = item.unwrap();
+            if i % 100 == 0 {
+                if i >= 25_000 {
+                    panic!("reached limit");
+                }
+                // tracing::info!(target: "neard", "{} {:?}", i, StateRecord::from_raw_key_value(key.clone(), value.clone()));
+                tracing::info!(target: "neard", "{}", i);
```
mm-near, line 78: maybe write  more than just 'integer' ?

Longarithm:
@mm-near Sorry, I unexpectedly requested a review and at the same time made lots of debug changes here. I will do it in separate branch next time
Until the slow performance issue is not resolved, I convert this to draft

akhi3030:
I suspect this PR is no longer being actively worked on.  Feel free to reopen when you are ready to start working on it again.

nikurt:
Sad to see this one closed :cry:

PR #5831, opened 2021-12-15
pytest: refactors tests to use `utils.wait_for_blocks`

mina86:


pytest/lib/utils.py
```diff
@@ -306,20 +307,35 @@ def compute_merkle_root_from_path(path, leaf_hash):
...
-        time.sleep(check_sec)
+        sync_info = node.get_status(**kw)['sync_info']
+        height = sync_info['latest_block_height']
+        if height != previous:
+            yield height, sync_info['latest_block_hash']
+            previous = height
+        time.sleep(poll_interval)
+
+
+def wait_for_blocks(node: LocalNode,
```
mm-near, line 327: let's add comments to all the util functions (so wait_for_blocks and. poll_blocks)

pytest/lib/utils.py
```diff
@@ -306,20 +307,35 @@ def compute_merkle_root_from_path(path, leaf_hash):
...
+def poll_blocks(node: LocalNode,
+                *,
+                timeout: float = 120,
+                poll_interval: float = 0.25,
+                **kw) -> typing.Iterable[typing.Tuple[int, str]]:
     started = time.time()
-    while max_height < start_height + num_blocks:
+    previous = -1
+    while True:
         assert time.time() - started < timeout
```
mm-near, line 318: let's add more info withi this assert (for example: how many blocks were yielded) - this might help with test debugging..

pytest/lib/utils.py
```diff
@@ -306,20 +307,35 @@ def compute_merkle_root_from_path(path, leaf_hash):
...
-        status = node.get_status()
-        max_height = status['sync_info']['latest_block_height']
-        if callback is not None:
-            if callback():
-                break
-        time.sleep(check_sec)
+        sync_info = node.get_status(**kw)['sync_info']
```
mm-near, line 319: do we handle errors properly ? can get_status return an error ?

mm-near (approved):
very nice !

pytest/lib/utils.py
mina86: I don’t think that it can. It’ll raise HTTP exception if there’s an error.

pytest/lib/utils.py
mina86: I’ll do that as separate PR

PR #5832, opened 2021-12-15
re: Rewrite `monitor_peers_trigger` in a cleaner way

pmnoxx:
The logic in computing `monitor_peers_trigger` is too complicated.
This PR rewrites logic, to an equivalent logic, that's easier to comprehend.

We can have a discussion, whenever this logic makes sense, but that should be another PR.

TODO:
- [x] Move 10ms to constants
- [x] Move 60s to constants

### DIFF
--- a/chain/network/src/peer_manager/peer_manager_actor.rs
+++ b/chain/network/src/peer_manager/peer_manager_actor.rs
@@ -47,8 +47,7 @@ use near_rate_limiter::{
 use near_store::Store;
 use rand::seq::{IteratorRandom, SliceRandom};
 use rand::thread_rng;
-use std::cmp;
-use std::cmp::max;
+use std::cmp::{max, min};
 use std::collections::{HashMap, HashSet};
 use std::net::SocketAddr;
 use std::pin::Pin;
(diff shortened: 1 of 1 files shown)


chain/network/src/peer_manager/peer_manager_actor.rs
```diff
@@ -1168,7 +1160,13 @@ impl PeerManagerActor {
     ///  - bootstrap outbound connections from known peers,
     ///  - unban peers that have been banned for awhile,
     ///  - remove expired peers,
```
mm-near, line 1162: add a comment explaining the difference between 'interval', 'default_interval' and 'max_interval

chain/network/src/peer_manager/peer_manager_actor.rs
```diff
@@ -1168,7 +1160,13 @@ impl PeerManagerActor {
...
-    fn monitor_peers_trigger(&mut self, ctx: &mut Context<Self>, max_interval: Duration) {
+    fn monitor_peers_trigger(
+        &mut self,
+        ctx: &mut Context<Self>,
+        mut interval: Duration,
+        default_interval: Duration,
+        max_interval: Duration,
```
mm-near, line 1168: optional:
we have 3 Duration types next to each other.. 
Maybe we should pass 'default_interval' and 'max_interval' as a tuple?

mm-near (approved):

chain/network/src/peer_manager/peer_manager_actor.rs
pmnoxx: Ok, I added an explanation.

chain/network/src/peer_manager/peer_manager_actor.rs
pmnoxx: @mm-near I make it into a tuple. It looks better now.

Issue #1313, opened 2019-09-14
Use Box for large enums


evgenykuzyakov:
Once borsh supports Box serialization, we should switch to it: [https://github.com/nearprotocol/borsh/issues/19](https://github.com/nearprotocol/borsh/issues/19)

Enums preallocate memory for the largest item. For example the same enum (e.g. `PeerMessage`) contains `Handshake` and `Block` or `ChunkOnePart`. `Handshake` enum value will have the same fixed memory size as the largest item

Rishabhraghwendra18:
@pmnoxx is this issue resolved ?

pmnoxx:
@Rishabhraghwendra18 More work is required. I have a script, which prints me all places, where this issue happens.

Rishabhraghwendra18:
Ok.. no problem @pmnoxx . I will find some other issue to work on

pmnoxx:
@Rishabhraghwendra18 I started working on fixing it for `near-network`. https://github.com/near/nearcore/pull/5834
The issue is a bit complicated, as some other code changes / refactoring are required. I don't think it's a good issue to start with.

PR #5834, opened 2021-12-15
feat: Move `RoutedMessage` inside `Box` to reduce `PeerMessage` enum size

pmnoxx:
We should reduce size of `PeerMessage` enum to 64 bytes or less.
We can start by doing this for `PeerMessage::Routed` variant.

See #1313 

TODO
- [ ] variant `Challenge`: 1143 bytes                                                                                                                                                                                         
- [ ] variant `Transaction`: 287 bytes 
- [ ] variant `Handshake`: 215 bytes
- [ ] variant `HandshakeV2`: 207 bytes 
- [ ] variant `HandshakeFailure`: 127 bytes
- [ ] variant `RequestUpdateNonce`: 87 bytes

### DIFF
--- a/chain/network/src/network_protocol.rs
+++ b/chain/network/src/network_protocol.rs
@@ -153,6 +153,10 @@ pub enum HandshakeFailureReason {
     GenesisMismatch(GenesisId),
     InvalidTarget,
 }
+const _: () = assert!(
+    std::mem::size_of::<HandshakeFailureReason>() <= 64,
+    "HandshakeFailureReason > 64 bytes"
+);
 
 impl fmt::Display for HandshakeFailureReason {
     fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
@@ -198,7 +202,7 @@ pub enum PeerMessage {
     Block(Block),
 
     Transaction(SignedTransaction),
-    Routed(RoutedMessage),
+    Routed(Box<RoutedMessage>),
 
     /// Gracefully disconnect from other peer.
     Disconnect,
@@ -212,6 +216,7 @@ pub enum PeerMessage {
     #[cfg(feature = "protocol_feature_routing_exchange_algorithm")]
     RoutingTableSyncV2(RoutingSyncV2),
 }
+const _: () = assert!(std::mem::size_of::<PeerMessage>() <= 1144, "PeerMessage > 1144 bytes");
 
 #[cfg(feature = "protocol_feature_routing_exchange_algorithm")]
 #[cfg_attr(feature = "deepsize_feature", derive(deepsize::DeepSizeOf))]
@@ -219,6 +224,8 @@ pub enum PeerMessage {
 pub enum RoutingSyncV2 {
     Version2(RoutingVersion2),
 }
+#[cfg(feature = "protocol_feature_routing_exchange_algorithm")]
+const _: () = assert!(std::mem::size_of::<RoutingSyncV2>() <= 80, "RoutingSyncV2 > 80 bytes");
 
 #[cfg(feature = "protocol_feature_routing_exchange_algorithm")]
 #[cfg_attr(feature = "deepsize_feature", derive(deepsize::DeepSizeOf))]
--- a/chain/network/src/peer/peer_actor.rs
+++ b/chain/network/src/peer/peer_actor.rs
@@ -680,15 +680,10 @@ impl StreamHandler<Result<Vec<u8>, ReasonForBan>> for PeerActor {
             self.routed_message_cache.put(key, now);
         }
         if let PeerMessage::Routed(routed) = &peer_msg {
-            match routed.as_ref() {
-                RoutedMessage { body: RoutedMessageBody::ForwardTx(_), .. } => {
-                    self.txns_since_last_block.fetch_add(1, Ordering::AcqRel);
-                }
-                _ => {
-                    self.txns_since_last_block.store(0, Ordering::Release);
-                }
+            if let RoutedMessage { body: RoutedMessageBody::ForwardTx(_), .. } = routed.as_ref() {
+                self.txns_since_last_block.fetch_add(1, Ordering::AcqRel);
             }
-        } else {
+        } else if let PeerMessage::Block(_) = &peer_msg {
             self.txns_since_last_block.store(0, Ordering::Release);
         }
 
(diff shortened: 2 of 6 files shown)


chain/network/src/network_protocol.rs
```diff
@@ -424,3 +424,27 @@ impl PeerMessage {
...
+    }
+
+    #[test]
+    fn test_assert_size() {
+        assert_size!(HandshakeFailureReason, 64);
+        assert_size!(PeerMessage, 1144);
+        #[cfg(feature = "protocol_feature_routing_exchange_algorithm")]
+        assert_size!(RoutingSyncV2, 80);
+    }
+}
```
matklad, line 450: It might make sense to use `static_assert` crate here.

matklad (approved):

chain/network/src/network_protocol.rs
pmnoxx: @matklad The best way would be to use `1.57.0` features 
```const _: () = assert!(std::mem::size_of::<u64>() == 8);```

Unfortunately, `const_format` is not fully stable yet. https://docs.rs/const_format/latest/const_format/
I can write a static assert like this:
```
assert!(std::mem::size_of::<u64>() == 8, "u64");
```
However, `assert!` doesn't support formatting in `static` case on rust stable.
I think it's best to add `TODO`, once that feature becomes stable.

Comment (matklad) [reply chain/network/src/network_protocol.rs:450]: Yeah, I mean that we can use this crate https://crates.io/crates/static_assertions which we already use, and which contains all the necessary polyfils. 

It was possible to "hand-write" this even before 1.57 (`[()][!cond]`), and I generally do perfer hand-writing such simple utilities, but, as we *already* depend on the crate, we might as use it. Though, no strong opinion here, the current approach is also fine!

Comment (pmnoxx) [reply chain/network/src/network_protocol.rs:450]: @matklad I looked into https://docs.rs/static_assertions/latest/static_assertions/
However, it's functionally the same as `1.57.0` static asserts. Neither of those examples contains proper prints on assert with format.

Comment (pmnoxx) [reply chain/network/src/network_protocol.rs]: Ok, I changed the code to use static asserts from `1.57.0`, they look good.

Comment (pmnoxx) [reply chain/network/src/network_protocol.rs]: https://doc.rust-lang.org/nightly/std/macro.const_format_args.html#

File: chain/network/src/network_protocol.rs [hunk 2]
```diff
@@ -424,3 +424,8 @@ impl PeerMessage {
         }
     }
 }
+
+const _: () =
+    assert!(std::mem::size_of::<HandshakeFailureReason>() <= 64, "HandshakeFailureReason");
+const _: () = assert!(std::mem::size_of::<PeerMessage>() <= 1144, "PeerMessage");
+const _: () = assert!(std::mem::size_of::<RoutingSyncV2>() <= 80, "RoutingSyncV2");
```
matklad, line 431: I'd probably put each of those after the corresponding struct. That way, when you look at the type's definiion, you realize that this is the place where we care about size.

pmnoxx:
````
@mm-near @pmnoxx: /chain/network-primitives/ /chain/network/
```
@mm-near @pmnoxx Can you take a look?

pmnoxx:
```
```
@mm-near Can you take a look?

pmnoxx:
```
```
@mm-near Can you take a look?

chain/network/Cargo.toml
```diff
@@ -4,7 +4,7 @@ version = "0.0.0"
 authors = ["Near Inc <hello@nearprotocol.com>"]
 publish = false
 # Please update rust-toolchain.toml as well when changing version here:
-rust-version = "1.56.0"
+rust-version = "1.57.0"
```
mm-near, line 7: ? should this be a part of this PR? 

@matklad - can we update the rust version only for a single cargo? (as others still have 1.56, right?)

chain/network/src/network_protocol.rs
```diff
@@ -267,6 +267,8 @@ pub enum HandshakeFailureReason {
     GenesisMismatch(GenesisId),
     InvalidTarget,
 }
+const _: () =
+    assert!(std::mem::size_of::<HandshakeFailureReason>() <= 64, "HandshakeFailureReason");
```
mm-near, line 271: add better comment for this assert (and others)

mm-near (changes requested):

chain/network/Cargo.toml
matklad: We can, but it doesn't make sense. 

It also doesn't make sense that we have `1.57` in `rust-toolchain.toml` and `1.56` in all of the cargo-tomls. I guess we should have updated cargo.tomls as a part of https://github.com/near/nearcore/pull/5659. 

So:

* this change should be reverted from this PR
* we should send a separate PR to upgrade rust-version everywhere
  * it probably makes sense to fold this with https://github.com/near/nearcore/issues/6120
* we should patch https://github.com/near/nearcore/blob/6d86ccd4fee74c1567dc2d83d384b1952826e796/tools/themis/src/rules.rs#L28-L48 to also check that `rust-version` is the right one.

matklad: filed https://github.com/near/nearcore/issues/6124

matklad: To say this more explicitly: 

It's ok to actually use 1.57 features in this PR -- 1.57 is the version of Rust we use. It's a per-existing bug that Cargo.toml specifies older-than-actually-required rust version.

chain/network/src/peer/peer_actor.rs
```diff
@@ -688,12 +688,16 @@ impl StreamHandler<Result<Vec<u8>, ReasonForBan>> for PeerActor {
...
-        {
-            self.txns_since_last_block.fetch_add(1, Ordering::AcqRel);
-        } else if let PeerMessage::Block(_) = &peer_msg {
+        if let PeerMessage::Routed(routed) = &peer_msg {
+            match routed.as_ref() {
+                RoutedMessage { body: RoutedMessageBody::ForwardTx(_), .. } => {
+                    self.txns_since_last_block.fetch_add(1, Ordering::AcqRel);
+                }
+                _ => {
+                    self.txns_since_last_block.store(0, Ordering::Release);
```
mm-near, line 697: you changed the behavior here..


before the PR - we were 'storing' 0 only if we got a 'block' transaction. Now we're storing '0' on each non ForwardTX one.

mm-near (changes requested):

chain/network/src/peer/peer_actor.rs
pmnoxx: @mm-near You are right, thanks. That was a good catch.

mm-near (approved):

PR #5835, opened 2021-12-15
feat: Remove deprecated `near` build target

pmnoxx:
We renamed `near` to `neard` on  Fri Apr 3 12:18:20 2020 -0700 See 253d99407eae6572a3f94124fd6cd2c661de122b
During this period, we kept both `near` and `neard`.

We should get rid of old `near` binary, it's not needed anymore.

See issue at https://github.com/near/nearcore/issues/5840


pmnoxx:
https://nayduck.near.org/#/run/2315

mina86:
NayDuck is not necessarily the only place. There may be some outdated documentation in `docs` repository or some other scripts in other repositories or other people’s computers.

pmnoxx:
> NayDuck is not necessarily the only place. There may be some outdated documentation in `docs` repository or some other scripts in other repositories or other people’s computers.

The proper thing to do is add a huge warning, which gets printed every time someone starts the binary with `near`.

miraclx (approved):

chefsale (approved):

mm-near (approved):

PR #5837, opened 2021-12-15
refactor: make sure there's only one code path an InconsistentStateError can take through runtime

matklad:
This started as pure refactor to remove duplicate logic, but it uncovered a surprising behavior, at least in theory. It seems today runtime can upgrade `TreNodeMissing` error into `InconsistentStateError`, is this expected? 

Builds on top of #5813 and can be viewed per commit. The actuall change is the last commit, the stuff before are pure refactors which make the change more obvious. 

Previously, there were two different code paths an
StorageError could take:

Path 1:
* RuntimeExt emits StorageError
* RuntimeExt calls wrap_storage_error to wrap it into VMLogicError::ExternalError
* near_vm_runner translates that into VMError::ExternalError
* actions.rs matches on AnyError and emits a StorageError

Path 2
* RuntimeExt emits StorageError
* RuntimeExt wraps that into VMLogicError::InconsistentStateError. Note
  that this state is especially weird: `InconsistentStateError` is a
  *subset* of `StorageError`. But here RuntimeExt treats *any* storage
  error as InconsistentStateError.
* near_vm_runner translates that into VMError::InconsistentStateError
* actions.rs translates that into `StorageError::InconsistentStateError`

Note in particular how on the path 2 we can start with
`StorageError::TrieNodeMissing` and end with
`StorageError::StorageInconsistentState("TrieNodeMissing")`, which feels
like a bug maybe?

Turns out, that the second path is actually dead code, so we can safely remove it, what we actually do in this PR.


runtime/runtime/src/ext.rs
```diff
@@ -178,11 +191,7 @@ impl<'a> External for RuntimeExt<'a> {
         let data_keys = self
             .trie_update
             .iter(&trie_key_parsers::get_raw_prefix_for_contract_data(self.account_id, prefix))
-            .map_err(|err| {
-                VMLogicError::InconsistentStateError(InconsistentStateError::StorageError(
-                    err.to_string(),
-                ))
-            })?
```
matklad, line 185: @EgorKulikov this is the place where we unconditionally upgrade *any* StorageError to InconsistentStateError. Is this a bug?

EgorKulikov: It does seem weird, I'll investigate

EgorKulikov: Though enclosing method does not seem to be used anywhere in our code and External in not exposed from near-vm-logic, so maybe it is just some old dead code?

EgorKulikov: Wait, no, it is exposed

EgorKulikov: But it is still not used anywhere

matklad: Thanks for looking into this! Indeed it seems that this is both buggy and dead code. I've added a comment to the function that it is dead. I'd probably say that we should just remove it completely (to make sure that, if we do need in the future, we add proper tests).

runtime/near-vm-errors/src/lib.rs
```diff
@@ -239,11 +239,10 @@ pub enum VMLogicError {
 
 impl std::error::Error for VMLogicError {}
 
-/// An error that is caused by an operation on an inconsistent state.
-/// E.g. a deserialization error or an integer overflow.
+/// An error that is caused by an operation on an inconsistent state. such as
```
EgorKulikov, line 242:
```suggestion
/// An error that is caused by an operation on an inconsistent state, such as
```

runtime/runtime/src/actions.rs
```diff
@@ -210,8 +206,9 @@ pub(crate) fn action_function_call(
                 ExternalError::ValidatorError(err) => Err(RuntimeError::ValidatorError(err)),
             };
         }
-        Some(VMError::InconsistentStateError(err)) => {
-            return Err(StorageError::StorageInconsistentState(err.to_string()).into());
+        Some(VMError::InconsistentStateError(err @ InconsistentStateError::IntegerOverflow)) => {
+            let message = err.to_string();
```
EgorKulikov, line 210: Why double to_string here?

EgorKulikov (approved):

runtime/runtime/src/actions.rs
matklad: Ah, good catch, copy-paste error

bowenwang1996 (approved):
Perspective of a protocol archaeologist: I believe we at one point want to have the ability for smart contract to remove storage by prefix, but later on realized that there is no a good way to charge for such potentially very expensive operation and decided not to do it.

PR #5838, opened 2021-12-15
NixOS support for upgradeable.py test

nagisa:
The downloaded binaries have the elf interpreter set to a path that does not exist on NixOS, which results in attempts to execute those binaries return a ENOENT.

To facilitate local testing on NixOS machines, this change makes the test script patch the interpreter for the downloaded binaries.


pytest/lib/branches.py
```diff
@@ -95,6 +95,33 @@ def _compile_current(branch: str) -> Executables:
     return Executables(_OUT_DIR, neard, state_viewer)
 
 
+def patch_binary(binary: pathlib.Path) -> None:
+    # Are we running on NixOS and require patching…?
```
mina86, line 99: Perhaps add some docstring?

mina86 (approved):
I don’t know much about NixOS but Python code looks good.

Issue #5840, opened 2021-12-15
Remove `near` build target.


pmnoxx:
We added `neard` build target in May 2020. We should consider removing `near` build target.

Steps
- https://github.com/near/nearcore/pull/5841
- https://github.com/near/nearcore/pull/5835

PR #5841, opened 2021-12-15
feat: Add warning to `near` binary

pmnoxx:
We should phase out usage of `near` binary. We created a new `neard` build target back in May 2020.
We can start by adding a warning and `panic!`.

Once we ship this PR, we will have to wait for 1-2 protocol updates. After that, it will be guaranteed, that anyone connecting to MAINNET will have fixed their scripts.

The main advantage of adding `panic!` is that it allow people to quickly find out that they use wrong binary in their scripts. It's better than using outdated binary by accident.

After, we make sure noone uses `near`, we will be able to safely drop it.

See https://github.com/near/nearcore/issues/5840

The alternative, would be to remove `near` right way.

### DIFF
--- a/neard/src/main.rs
+++ b/neard/src/main.rs
@@ -59,6 +59,7 @@ fn main() {
         println!("WARNING WARNING WARNING");
         println!("Usage of `./near` binary is deprecated since May 2020!!!");
         println!("Use `./neard` instead");
+        panic!("Use ./neard build target");
     }
 
     // We use it to automatically search the for root certificates to perform HTTPS calls
--- a/neard/src/main.rs
+++ b/neard/src/main.rs
@@ -42,7 +42,25 @@ static ALLOC: MyAllocator<tikv_jemallocator::Jemalloc> =
 #[global_allocator]
 static ALLOC: tikv_jemallocator::Jemalloc = tikv_jemallocator::Jemalloc;
 
+fn prog() -> Option<String> {
+    env::args()
+        .next()
+        .as_ref()
+        .map(std::path::Path::new)
+        .and_then(std::path::Path::file_name)
+        .and_then(std::ffi::OsStr::to_str)
+        .map(String::from)
+}
+
 fn main() {
+    if prog().map(|name| name == "near").unwrap_or_default() {
+        println!("WARNING WARNING WARNING");
+        println!("WARNING WARNING WARNING");
+        println!("WARNING WARNING WARNING");
+        println!("Usage of `./near` binary is deprecated since May 2020!!!");
+        println!("Use `./neard` instead");
+    }
+
     // We use it to automatically search the for root certificates to perform HTTPS calls
     // (sending telemetry and downloading genesis)
     openssl_probe::init_ssl_cert_env_vars();
--- a/neard/src/main.rs
+++ b/neard/src/main.rs
@@ -1,6 +1,6 @@
 mod cli;
 
-use std::env;
+use std::{env, thread};
 
 use self::cli::NeardCmd;
 use clap::crate_version;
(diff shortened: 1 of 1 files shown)


matklad:
TBH, my gut feeling is that we can remove the old binary without any notice. Human operators who actively use the thing probably noticed that `neard` exists, and automation scripts which rely on `near` won't read the warnings.

mina86:
There isn’t a value in panicking. If we are willing to just break, it’s best to break at compilation time by removing the `near` target. The two options are i) remove `near` binary or ii) change this to be an obnoxious, multi-line `error!` with a 10 second sleep but which then lets users continue to use the binary. The first option is obviously simpler, the second would provide a grace period and maybe help some people notice.

The first option isn’t unreasonable. Commenting on the other PR I just meant that we need to look through our other repositories to see if all places are updated. Especially if there’s some documentation or build scripts we offer to the public which still references `near` than we probably should keep the `near` binary for a bit longer.

But with the second option we shouldn’t panic and we should check `CARGO_BIN_NAME` since what we care is whether someone used `near` build target rather than the name of the file. (Compiling `neard` and renaming it `near` is perfectly fine; compiling `near` and renaming it `neard` isn’t).

miraclx:
> But with the second option we shouldn’t panic and we should check CARGO_BIN_NAME since what we care is whether someone used near build target rather than the name of the file. (Compiling neard and renaming it near is perfectly fine; compiling near and renaming it neard isn’t).

+1 for this idea. Seems reasonable enough.

neard/src/main.rs
```diff
@@ -42,7 +42,33 @@ static ALLOC: MyAllocator<tikv_jemallocator::Jemalloc> =
...
+        .next()
+        .as_ref()
+        .map(std::path::Path::new)
+        .and_then(std::path::Path::file_name)
+        .and_then(std::ffi::OsStr::to_str)
+        .map(String::from)
+}
+
 fn main() {
+    if prog().map(|name| name == "near").unwrap_or_default() {
```
mina86, line 56: This should be `env!("CARGO_BIN_NAME") == "near"`.

neard/src/main.rs
```diff
@@ -42,7 +42,33 @@ static ALLOC: MyAllocator<tikv_jemallocator::Jemalloc> =
...
+        println!("WARNING WARNING WARNING");
+        println!("WARNING WARNING WARNING");
+        println!("WARNING WARNING WARNING");
+        println!("WARNING WARNING WARNING");
+        println!("WARNING WARNING WARNING");
+        println!("WARNING WARNING WARNING");
+        println!("WARNING WARNING WARNING");
+        println!("Usage of `./near` binary is deprecated since May 2020!!!");
+        println!("Use `./neard` instead");
+        println!("Use ./neard build target");
```
mina86, line 68: `tracing::error!(…)` or `tracing::warn!(…)` would be better than
`println!`.  Also, the message should better describe what is going
on.  Perhaps something like:

```
use tracing::warn;
```
```
        warn!(target: "neard",
              "This executable has been compiled as a ‘near’ bin target.");
        warn!(target: "neard",
              "That target has been deprecated since May 2020 and will \
               eventually be removed.");
        warn!(target: "neard",
              "Please update your build process to use the new ‘neard’ target,\
               e.g. by using ‘make release’ or ‘make neard’ commands.");
        warn!(target: "neard", "See {} for more information.",
              "https://docs.near.org/docs/develop/node/validator/compile-and-run-a-node");
        warn!(target: "neard",
              "I’m now going to sleep for 30 seconds to be obnoxious about \
               it.  ;)");
```

pmnoxx:
> There isn’t much value in panicking. If we are willing to just break, it’s best to break at compilation time by removing the `near` target. The two options are i) remove `near` binary or ii) change this to be an obnoxious, multi-line `error!` with a 10 second sleep but which then lets users continue to use the binary. The first option is obviously simpler, the second would provide a grace period and maybe help some people notice.
> 
> The first option isn’t unreasonable. Commenting on the other PR I just meant that we need to look through our other repositories to see if all places are updated. Especially if there’s some documentation or build scripts we offer to the public which still references `near` than we probably should keep the `near` binary for a bit longer.
> 
> But with the second option we shouldn’t panic and we should check `CARGO_BIN_NAME` since what we care is whether someone used `near` build target rather than the name of the file. (Compiling `neard` and renaming it `near` is perfectly fine; compiling `near` and renaming it `neard` isn’t).

I think we should just remove `near`. I has been 1.5 years. Adding warnings, is something that we should have done 1.5 year ago. Though, In general for those kind of migrations, I think is the way to go. And, I have seen  many companies do. It's definitely a practice, we should adopt for future migrations.

Although, there is some value in doing this anyway. It's a good time to think how to improve communication in general, and what kind of "tools" we have at our disposal, to make usage of nodes easier.

Issue #5708, opened 2021-12-08
Expose chunk stats in validator rpc endpoint


bowenwang1996:
Currently [the validator rpc endpoint](https://docs.near.org/docs/api/rpc/network#validation-status) only exposes block stats and not chunk stats. Validators would like to understand, in addition to the number of blocks produced, the number of chunks produced and expected to get a better understanding of the status of their nodes.

PR #5843, opened 2021-12-15
feat(rpc): add chunk production stats to the /validators method

marcelo-gonzalez:
It's a small change to add this info, as it's already present in
the EpochSummary and EpochInfoAggregator structs. This should not
be a breaking change, since fields are only being added.

Test plan
---------
Since this isn't adding much new logic, just exposing the result of
existing logic, it should be sufficient to just manually check that
the output looks correct/in line with the current chain

Issue: https://github.com/near/nearcore/issues/5708


core/primitives/src/views.rs
```diff
@@ -1416,6 +1416,8 @@ pub struct CurrentEpochValidatorInfo {
     pub shards: Vec<ShardId>,
     pub num_produced_blocks: NumBlocks,
     pub num_expected_blocks: NumBlocks,
+    pub num_produced_chunks: NumBlocks,
+    pub num_expected_chunks: NumBlocks,
```
matklad, line 1420: This is a `Borsh` structure, so adding fields is generally breaking (it changes offsets of everything). However, it seems that this is BorsSerialize accidently. Let's apply the following diff to make sure borsh isn't a concern: 

```diff
diff --git a/core/primitives/src/views.rs b/core/primitives/src/views.rs
index b45b9c7bc..73d531f32 100644
--- a/core/primitives/src/views.rs
+++ b/core/primitives/src/views.rs
@@ -1378,7 +1378,7 @@ impl TryFrom<ReceiptView> for Receipt {
 
 /// Information about this epoch validators and next epoch validators
 #[cfg_attr(feature = "deepsize_feature", derive(deepsize::DeepSizeOf))]
-#[derive(BorshSerialize, BorshDeserialize, Serialize, Deserialize, Debug, PartialEq, Eq, Clone)]
+#[derive(Serialize, Deserialize, Debug, PartialEq, Eq, Clone)]
 pub struct EpochValidatorInfo {
     /// Validators for the current epoch
     pub current_validators: Vec<CurrentEpochValidatorInfo>,
@@ -1406,7 +1406,7 @@ pub struct ValidatorKickoutView {
 }
 
 #[cfg_attr(feature = "deepsize_feature", derive(deepsize::DeepSizeOf))]
-#[derive(BorshSerialize, BorshDeserialize, Serialize, Deserialize, Debug, PartialEq, Eq, Clone)]
+#[derive(Serialize, Deserialize, Debug, PartialEq, Eq, Clone)]
 pub struct CurrentEpochValidatorInfo {
     pub account_id: AccountId,
     pub public_key: PublicKey,
```

Though, even for JSON, what if an RPC client which uses new version contracts a near node which is still at the old version? I think the client would  fail to deserialize this. So maybe let's also `#[serde(default)]`?

marcelo-gonzalez: Oof good catch. Updated.

marcelo-gonzalez: Also on this note, I wonder if https://github.com/near/nearcore/pull/4221 broke anything? I would guess the same issues would happen there, but maybe nobody noticed :/

matklad: I think not, because, as far as I know, the borsh impl actually unused.

chain/epoch_manager/src/lib.rs
```diff
@@ -1028,11 +1032,19 @@ impl EpochManager {
...
                             .unwrap_or_else(|| &ValidatorStats { produced: 0, expected: 0 })
                             .clone();
+
+                        let mut chunk_stats = ValidatorStats { produced: 0, expected: 0 };
+                        for (_shard, tracker) in aggregator.shard_tracker.iter() {
+                            if let Some(stats) = tracker.get(&(validator_id as u64)) {
+                                chunk_stats.produced += stats.produced;
+                                chunk_stats.expected += stats.expected;
+                            }
+                        }
```
matklad, line 1047: This feels quadratic (n validators X n shards), though probably is ok.

matklad (approved):

bowenwang1996 (approved):

PR #5844, opened 2021-12-15
feat: Improve `neard` output while running `./target/debug/neard init`

pmnoxx:
I tried running `neard init`. I think that the output can be improved a little bit in case of failure.

I did it for one case, as an example:

```
L .~/r/nearcore.master piotr-improve-neard (1/0/0) cargo build -p neard && ./target/debug/neard init
Dec 15 08:39:52.430  INFO neard: Version: trunk, Build: aab139777-modified, Latest Protocol: 50
thread 'main' panicked at 'NeardCmd::parse_and_run failed: Init command failed InitCmd { download_genesis: false, download_config: false, fast: false, account_id: None, chain_id: None, download_genesis_url: None, download_config_url: None, genesis: None, boot_nodes: None, num_shards: 1, test_seed: None, max_gas_burnt_view: None }

Caused by:
    0: nearcore::init_configs
    1: Can't create new config, it already exists.
                    Use `neard unsafe_reset_all` to delete it.
                    folder = "/home/pmnoxx/.near"
                    chain_id = "test-chain-1ORDx"
                    path = "/home/pmnoxx/.near/genesis.json"', neard/src/main.rs:51:31
```

I'm debugging, why node doesn't start on my machine. And, it's hard to figure out why, without reading the code deeply. I'm adding information to code paths, which failed for me.

### DIFF
--- a/nearcore/src/config.rs
+++ b/nearcore/src/config.rs
@@ -3,8 +3,9 @@ use std::fs::File;
 use std::io::{Read, Write};
 use std::path::Path;
 use std::sync::Arc;
-use std::time::Duration;
+use std::time::{Duration, Instant};
 
+use anyhow::bail;
 use hyper::body::HttpBody;
 use near_primitives::time::Clock;
 use num_rational::Rational;
@@ -816,13 +817,23 @@ pub fn init_configs(
     download_config_url: Option<&str>,
     boot_nodes: Option<&str>,
     max_gas_burnt_view: Option<Gas>,
-) {
+) -> anyhow::Result<()> {
     fs::create_dir_all(dir).expect("Failed to create directory");
     // Check if config already exists in home dir.
     if dir.join(CONFIG_FILENAME).exists() {
         let config = Config::from_file(&dir.join(CONFIG_FILENAME));
-        let genesis_config = GenesisConfig::from_file(&dir.join(config.genesis_file));
-        panic!("Found existing config in {} with chain-id = {}. Use unsafe_reset_all to clear the folder.", dir.display(), genesis_config.chain_id);
+        let path = &dir.join(config.genesis_file);
+        let genesis_config = GenesisConfig::from_file(path);
+        bail!(
+            "Can't create new config, it already exists.
+             Use `neard unsafe_reset_all` to delete it.
+             folder = {:?}
+             chain_id = {:?}
+             path = {:?}",
+            dir.display(),
+            genesis_config.chain_id,
+            path
+        );
     }
 
     let mut config = Config::default();
@@ -978,6 +989,7 @@ pub fn init_configs(
             info!(target: "near", "Generated node key, validator key, genesis file in {}", dir.display());
         }
     }
+    Ok(())
 }
 
 pub fn create_testnet_configs_from_seeds(
@@ -1201,7 +1213,10 @@ impl From<NodeKeyFile> for KeyFile {
 
 pub fn load_config_without_genesis_records(dir: &Path) -> NearConfig {
     let config = Config::from_file(&dir.join(CONFIG_FILENAME));
+    info!(message = "Loading genesis started");
+    let started = Instant::now();
     let genesis_config = GenesisConfig::from_file(&dir.join(&config.genesis_file));
+    info!(message = "Loading genesis done", took = ?started.elapsed());
     let genesis_records_file = if let Some(genesis_records_file) = &config.genesis_records_file {
         dir.join(genesis_records_file)
     } else {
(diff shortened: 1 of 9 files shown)


nearcore/src/config.rs
```diff
@@ -816,13 +817,23 @@ pub fn init_configs(
     download_config_url: Option<&str>,
     boot_nodes: Option<&str>,
     max_gas_burnt_view: Option<Gas>,
-) {
+) -> Result<()> {
```
matklad, line 820:
```suggestion
) -> anyhow::Result<()> {
```

we have many different results, so being explicit helps

bowenwang1996 (approved):
Thanks!

pmnoxx:
> Thanks!

I ended up writing my own `near-up`; `nay-duck`; `unit-tester` scripts in `Rust`; As well as other equivalents of scripts we have. I found them to be too hard to use.
As a rule of thumb. I ask myself "Why do I need to, read a code, to figured out how to pass arguments, such that the CLI doesn't panic?".

Here is an example of the `near-up` I write for myself.

I can run a command like this:
```console
near-up run mainnet --log-level DEBUG --init --build --logs
``` 

Here are arguments I implemented.
- RUN <mainnet|testnet|bettanet> - command to run node, takes a choice of the network
- (OPTIONAL) `-- build` -- it build it with proper flags for `mainnet `
- (OPTIONAL) `--log-level` -- specified log level to use for l`oggin`
- (OPTIONAL) `--init` -- tells that if not initialized, it initialized and gets genesis file and configs
- (OPTIONAL) `--logs` -- tails logs afterwards
- (OPTIONAL) `--docker` --runs in a docker
- (OPTIONAL) `--binary-path` -- runs from binary path
- ... some other config related arguments, because I don't want to edit config manually each time

It's simple, can be installed with
```
cargo install --path ~/repos/near-up
```

It has a lot of other functionality, I use for my self as well. 

Anyway, `CLI` for `neard` is bad compared to many other internal `CLI` I have written in the past.

pmnoxx:
```
file="/home/pmnoxx/repos/nearcore/chain/indexer/src/lib.rs" owners=["@khorolets"]
file="/home/pmnoxx/repos/nearcore/runtime/runtime-params-estimator/src/main.rs" owners=["@olonho", "@matklad"]
file="/home/pmnoxx/repos/nearcore/test-utils/state-viewer/src/cli.rs" owners=["@nikurt"]
```
@matklad @khorolets @nikurt It looks like your approvals are needed.

pmnoxx:
```
@matklad @olonho: /runtime/runtime-params-estimator/
@khorolets: /chain/indexer/
@nikurt: /test-utils/state-viewer
```
@khorolets @matklad @nikurt @olonho Can you take a look?

neard/src/main.rs
```diff
@@ -48,5 +48,5 @@ fn main() {
     openssl_probe::init_ssl_cert_env_vars();
     near_performance_metrics::process::schedule_printing_performance_stats(Duration::from_secs(60));
 
-    NeardCmd::parse_and_run()
+    NeardCmd::parse_and_run().expect("NeardCmd::parse_and_run failed")
```
matklad, line 51: This would be better: 

```rust
fn main() -> anyhow::Result<()> {
    // We use it to automatically search the for root certificates to perform HTTPS calls
    // (sending telemetry and downloading genesis)
    openssl_probe::init_ssl_cert_env_vars();
    near_performance_metrics::process::schedule_printing_performance_stats(Duration::from_secs(60));

    NeardCmd::parse_and_run()
}
```

neard/src/cli.rs
```diff
@@ -173,11 +179,12 @@ impl InitCmd {
             self.download_config_url.as_deref(),
             self.boot_nodes.as_deref(),
             self.max_gas_burnt_view,
-        );
+        )
+        .with_context(|| "nearcore::init_configs")
```
matklad, line 183: Context is what the end user will see, so this maybe shouldn't mention Rust-specific concepts like modules. Maybe `.context("failed to initialize configuration")`.

matklad (approved):
Feels like an improvement already!

nikurt (approved):

miraclx (approved):

pmnoxx:
```
@khorolets: chain/indexer/
```
@khorolets Can you take a look?

khorolets (approved):

miraclx (commented):
@pmnoxx, what's the status of this PR?

#6343, which is based on this, merges with master and resolves all the merge conflicts.

mm-near (approved):

PR #5850, opened 2021-12-15
re: Fix Clion error in `near-client`

pmnoxx:
There is one place in `near-client`, where Rust plugin for Clion prints an error about Display not being implemented.
Let's fix it.


pmnoxx:
```
@mm-near: /chain/
```
@mm-near Can you take a look?

mm-near (approved):

PR #5852, opened 2021-12-15
pytest: introduce `BaseNode.get_latest_block` method

mina86:


pytest/tests/adversarial/fork_sync.py
```diff
@@ -64,26 +64,20 @@
 logger.info("all nodes restarted")
 
 while cur_height < TIMEOUT:
```
mm-near, line 66: comment on what this while loop is testing

pytest/lib/mocknet.py
```diff
@@ -360,8 +359,7 @@ def transfer_between_nodes(nodes):
     logger.info(f'Alice initial balance: {alice_initial_balance}')
     logger.info(f'Bob initial balance: {bob_initial_balance}')
 
-    last_block_hash = node.get_status()['sync_info']['latest_block_hash']
-    last_block_hash_decoded = base58.b58decode(last_block_hash.encode('utf8'))
+    last_block_hash = node.get_latest_block().hash_bytes
```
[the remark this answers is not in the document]
mina86: `.hash_bytes` (as opposed to `.hash`) does the decode.

pytest/tests/adversarial/fork_sync.py
mina86: I mean… How should I know. I didn’t write the test. ;)

PR #5858, opened 2021-12-16
pytest: remove `utils.Unbuffered`

mina86:
The tests use logging which writes to standard error so making
standard output unbuffered doesn’t do anything.  Furthermore, there is
no reason to make test’s output unbuffered.

### DIFF
--- a/pytest/tests/stress/stress.py
+++ b/pytest/tests/stress/stress.py
@@ -34,11 +34,9 @@ sys.path.append(str(pathlib.Path(__file__).resolve().parents[2] / 'lib'))
 
 from cluster import init_cluster, spin_up_node, load_config
 from configured_logger import logger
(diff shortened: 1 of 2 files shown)


pytest/tests/stress/stress.py
```diff
@@ -79,9 +77,8 @@ def stress_process(func):
     def wrapper(stopped, error, *args):
         try:
             func(stopped, error, *args)
-        except Exception as e:
-            traceback.print_exc()
```
mm-near, line 83: this is no longer needed ?

mm-near (approved):

pytest/tests/stress/stress.py
mina86: No. `logger.info` will print exception when `exc_info` is passed.

PR #5866, opened 2021-12-16
feat: Change `SyncLruCache` api to match `LruCache`

pmnoxx:
While implementing wrappers, it's generally a good idea, to make the api similar to the underlying wrapped structure.

- Changed how `put` behaves,
- Added better documentation
- Some other minor code quality improvements
- add `len` method
- add `clear` method


utils/near-cache/src/lib.rs
```diff
@@ -10,37 +10,46 @@ pub struct SyncLruCache<K, V> {
...
-            return result;
-        }
-        let val = f(&key);
-        let val_clone = val.clone();
-        self.inner.lock().unwrap().put(key, val_clone);
-        val
+    /// Return the value of they key in the cache otherwise computes the value and inserts it into
+    /// the cache. If the key is already in the cache, they gets gets moved to the head of
+    /// the LRU list.
+    pub fn get_or_put(&self, key: K, f: impl FnOnce(&K) -> V) -> V {
```
matklad, line 24: I'd keep the `insert` terminology. I agree that "don't change wrapped API" is a good rule of thumb, as generally that API uses specific idioms for a reason. 

In this specific case though, I think the reason for `put` terminology is mostly historical. The API originated from a very old version of Rust, when the terminology wasn't ironed out at all: https://doc.rust-lang.org/0.12.0/std/collections/lru_cache/struct.LruCache.html. Note `0.12.0` in the URL. 

Today, Rust uses insert everywhere, 

* https://doc.rust-lang.org/stable/std/?search=insert
* https://doc.rust-lang.org/stable/std/?search=put

utils/near-cache/src/lib.rs
```diff
@@ -10,37 +10,46 @@ pub struct SyncLruCache<K, V> {
...
+        self.get(&key).unwrap_or_else(|| {
+            let val = f(&key);
+            self.inner.lock().unwrap().put(key, val.clone());
```
matklad, line 27: Note that this re-introduces a minor perf bug where we clone the value in the scope of the lock. That's a rule-of-thumb: everywhere you do `.lock()`, take a second look at the code and see if you can minimize the extent at which the lock is held.

utils/near-cache/src/lib.rs
matklad: Actaully, I take this back. I didn't realise that we are going to use non-sync version as well, as in https://github.com/near/nearcore/pull/5871. If that's the case, than, yeah, it definitelly is better to align names.

matklad (approved):

PR #5868, opened 2021-12-17
Update code owners for `/utils`

pmnoxx:
I added `/utils` a while back. We have code there that's used in production, and it's better to make sure I review all changes to it.


CODEOWNERS
```diff
@@ -9,6 +9,7 @@
 /chain/jsonrpc-primitives/ @khorolets @miraclx @chefsale
 /chain/indexer/ @khorolets
 /chain/rosetta-rpc/ @frol @mina86
+/utils @pmnoxx
```
mm-near, line 12: please add me there too (as we should have at least 2 people for each directory here)

mm-near (approved):

PR #5869, opened 2021-12-17
chain: indexer: avoid allocation of temporary vector

mina86:
In the two places where `fetch_chunks` function is called, it is used
to fetch chunks for a block.  By moving code which extracts list of
chunk hashes from a block view into the `fetch_chunks` function we
reduce code duplication and eliminate an allocation of a temporary
vector object.

To better describe what the function does, rename it to
`fetch_block_chunks`.


khorolets (approved):
Thank you!

PR #5874, opened 2021-12-17
nightly: fix expensive integration tests after test move

mina86:
Commit d3efd9942: ‘Move testbs for `integration-tests` from `/tests`
to `/src/test`’ moved tests around affecting their path but did not
correctly update nightly list with the correct test paths.  Fix that.


pmnoxx (approved):
LGTM

PR #5875, opened 2021-12-17
re: Fix clippy warnings in `/core/store/src/trie`

pmnoxx:
There are a couple of `clippy` warnings, which we can fix in `/core/store/src/trie`


matklad (approved):

pmnoxx:
```
@longarithm @mzhangmzz: /core/store/src/trie
```
@longarithm @mzhangmzz Can you take a look?

Longarithm (approved):

PR #5873, opened 2021-12-17
re: Fix `clippy::redundant_clone `

pmnoxx:
Refactor `near-client`, fixes clippy issues in files touched by https://github.com/near/nearcore/pull/5872

PR #5876, opened 2021-12-17
(chore) (primitives) Some small readability fixes suggested by lint

EgorKulikov:


EgorKulikov:
Converting to draft as some fixes are done in #5873 , will bring this back after that is merged

core/primitives/src/utils.rs
```diff
@@ -470,7 +470,7 @@ impl fmt::Debug for dyn CompiledContractCache {
 /// objects using tracing.
 ///
 /// tracing::debug!(target: "diagnostic", value=%ser(&object));
-pub fn ser<'a, T>(object: &'a T) -> Serializable<'a, T>
+pub fn ser<T>(object: &T) -> Serializable<T>
```
matklad, line 473: It's a bit weird that clippy suggests that. I *think* the idiomatic way to write these nowdays is 

```rust
pub fn ser<T>(object: &T) -> Serializable<'_, T>
```

That is, you do want to indicate to the reader that there's a lifitime in the struct.

matklad (approved):

PR #5851, opened 2021-12-15
bug: Remove `unwrap_or_error!` from `near-network`

pmnoxx:
It looks like we were not using `unwrap_or_error` correctly.

If an error within `monitor_peers_trigger` where to happen, the node would stop connecting to new peers. That issues was obfuscated by usage of   `unwrap_or_error` macro.

PR #5877, opened 2021-12-17
fix: Fix `PeerManager::monitor_peers_trigger` stopping on error

### LINKED ISSUE #5878: PeerManager::monitor_peers_trigger may stop in case of an error
After doing cleanup in #5851
We found out that the unwrap_or_error! macro, was used inside a trigger.
This would cause "return" to be called in case of an error, and that it turn would result in monitor_peers_trigger trigger stopping.

Let's remove those returns.

pmnoxx:
After doing cleanup in https://github.com/near/nearcore/pull/5851
We found out that the `unwrap_or_error! ` macro, was used inside a trigger.
This would cause "return" to be called in case of an error, and that it turn would result in `monitor_peers_trigger` trigger stopping.

Let's remove those returns.

Closes #5878


chain/network/src/peer_manager/peer_manager_actor.rs
```diff
@@ -1170,9 +1170,6 @@ impl PeerManagerActor {
         for peer_id in to_unban {
             if let Err(err) = self.peer_store.peer_unban(&peer_id) {
                 error!(target: "network", message = "Failed to unban a peer", ?err);
-                // TODO: Do we really want to return?
```
mm-near, line 1173: can we have tests to cover these error cases ?

mm-near (changes requested):

pmnoxx:
```
```

chain/network/src/peer_manager/peer_manager_actor.rs
mm-near: This PR is fixing a bug - so we should have a test case, that 'fails' before this fix and 'works' afterwards.

(even if this is 'just' a unittest that simulates database read/write failure)

pmnoxx: @mm-near Currently, we don't have a good way to test triggers, nor simulate database failures.

Besides what do we event want to test? 
a) That this function continues to run, and there is no return?
The code has been now rewritten, we can see that there is no "return", that can be easily seen.
b) Test that trigger works. 
Triggers needs to be rewritten to be unit testable. I have a plan to do it, but that's a outside of scope of this  pr.

mm-near (approved):

PR #5881, opened 2021-12-17
Add a text describing runtimes and their requirements

nagisa:
It is nice to agree on and have a document describing our requirements and priorities and criteria that we evaluate the runtimes based on and try to maintain.

This PR adds a document that attempts to primarily document our requirements and criteria. While I also added some text describing two backends we have considered, I wouldn't consider those sections to be the primary focus of this PR. Instead I think it would make sense to update them in a follow up alongside the discussion on the frontend options we've considered.


nagisa:
I haven't yet proof-read this and there's definitely a lot that can be added to this still (e.g. comparison between the frontents (wasmer vs wasmtime), but I appreciate all nitpicks ^^

runtime/near-vm-runner/RUNTIMES.md
```diff
@@ -0,0 +1,176 @@
...
+The NEAR validator implementation currently targets `x86_64-linux`, so stellar support for this
+target is a hard requirement.
+
+However, developers may not necessarily have a `x86_64` machine to develop with, or they might be
+using a device running a different operating system. With the recent surge in popularity of devices
+based on the Arm CPU architecture chips and generally high popularity of the macOS and Windows
+operating systems it would be a huge boon to the development experience if the runtime used also
+supported these other targets.
+
+## Runtime performance
```
matklad, line 89: Let's maybe split this into two primary metrics we care about: 

* *latency* to run a simple function in a big contract (rationale: many contracts are relatively simple in terms of business logic and do little compute)
* *throughput* to run a relatively long wasm computation as fast as possible (rationale: at the same time, there are some compute-heavy contracts, especially those that run interpreters in wasm)

matklad: In general, I feel like our runtime model has three parts to it: 

* compilation (read wasm, compile to machine code, write machine code to disk)
* linking/loading (load machine code from disk, load it into the executable memory, unload it after execution)
* actual execution

runtime/near-vm-runner/RUNTIMES.md
```diff
@@ -0,0 +1,176 @@
...
+stack operations, pick more efficient instructions and sometimes avoid some operations such as NaN
+canonicalization altogether.
+
+Presence of this state comes at a cost of testability of implementation. The behaviour of the
+codegen can depend on prior Wasm instructions. Because of this there's a combinatorial explosion of
+the test cases that need to be considered and verified before confidence in the results can be
+established. Relatedly, at times it may also be difficult to validate changes made to the
+implementation of the `wasmer-singlepass` codegen. The global state is definitely a source of
+potential spooky action at a distance problems where changing code generation of a specific
+instruction affects correctness and behaviour of another one.
```
matklad, line 147: Curious, is this theoretical, or was 2.1.0 miscompilation an example of that?

runtime/near-vm-runner/RUNTIMES.md
```diff
@@ -0,0 +1,176 @@
...
+validator. Cranelift gets high scores for its Correctness, Reliability and Security properties.
+It supports targeting the Arm, s390x and x86_64 architectures as well as a variety of
+operating systems, netting it a high score on the platform support metric as well.
+
+Upon a brief investigation in December of 2021 generating machine code with Cranelift should
+broadly be `O(n)` over input size with the exception of register allocation, which due to their
+soon-to-be replaced regalloc algorithm is currently `O(n²)`.
+
+One detail where Cranelift may fall short is in the ability to produce super-optimized machine code
+sequences for hot operations such as gas counting.
```
matklad, line 176: I feel something's missing here... Especially, my main worry with cranelift is that it'd be hard to estimate costs for generated code. Basically, with `singlepass` we somewhat confidente that the compiler isn't being overly smart. With cratelift, it feels like it could happen that we are benchmarking happy case.

runtime/near-vm-runner/RUNTIMES.md
```diff
@@ -0,0 +1,176 @@
...
+validator. Some of them are already listed in the
+[FAQ](FAQ.md#practical-aspects-of-the-execution-engines) document. Listed roughly in the order of
+importance:
+
+* correctness – does the runtime do what it claims to do;
+* security – how well does the runtime deal with untrusted input;
+* reliability – how much confidence is there in the implementation of runtime;
+* platform support – does the runtime support targets that we want to target;
+* runtime performance – how quickly can the runtime execute the prepared Wasm code;
+* preparation performance – how quickly can the runtime start running Wasm code;
```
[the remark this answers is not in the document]
nagisa: I consider `x86_64-linux` to be a part of such platform support requirement. It is not inconceivable that otherwise a very nice wasm compiler might only support targeting, say, [JVM bytecode](https://github.com/cretz/asmble), at which point it is most likely not applicable for our use cases, regardless of performance it can achieve.

runtime/near-vm-runner/RUNTIMES.md
```diff
@@ -0,0 +1,176 @@
...
+usability of the whole system as a whole.
+
+One of more costly operations that occurs in a typical contract quite often is accounting for gas
+costs, so it is quite important that the runtime gives sufficient flexibility to the NEAR
+validator to implement such operations efficiently. And, of course, general code quality also
+matters.
+
+On a scale of trade-offs, runtime performance is probably one of the less important metrics. Slower
+execution of a contract doesn't make NEAR protocol unsound as an idea, whereas something like a
+non-deterministic execution outcome would.
```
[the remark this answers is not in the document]
nagisa: I don't think we can afford correctness to be a “want”. Wasm is a target for optimizing compilers, and they can definitely make optimizations with an underlying presumption that the generated code will be ran according to the spec. With a non-conformant runtime it can take very innocuous deviation in behaviour before e.g. a wrong branch is taken, unauthorized user initiates a token transfer, etc.

runtime/near-vm-runner/RUNTIMES.md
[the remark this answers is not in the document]
matklad: Yeah, we def need x86 support, support for everything else is niceto have.

runtime/near-vm-runner/RUNTIMES.md
```diff
@@ -0,0 +1,176 @@
...
+The NEAR validator implementation currently targets `x86_64-linux`, so stellar support for this
+target is a hard requirement.
+
+However, developers may not necessarily have a `x86_64` machine to develop with, or they might be
+using a device running a different operating system. With the recent surge in popularity of devices
+based on the Arm CPU architecture chips and generally high popularity of the macOS and Windows
+operating systems it would be a huge boon to the development experience if the runtime used also
+supported these other targets.
+
+## Runtime performance
```
jakmeier, line 89: I think there is another important point to be mentioned about performance: Predictability

Essentially, our blockchain is a real-time system, as a transaction that consumes e.g. 1TGas should finish within 1ms. Optimizing performance for a RT system is a completely different engineering problem than optimization otherwise, since in RT, it is always the worst-case that counts.

This might sound like something that's only relevant to the preparation cost. But I would argue it is just as important for the generated code. For example, an optimization which tries to be smart with arrangement of more/less likely branches could be problematic for us, for several reasons that pop to my mind:
1) It doesn't improve worst-case performance but creates extra compiler overhead.
2) Gaining some extra performance on average doesn't help us a bit in terms of overall system performance, since we have to be conservative in how many transactions go in a block anyway.
3) It makes it harder for us to estimate a realistic worst-case execution time. As the average time per branch goes down, we are even tempted to lower the gas fee parameters for it. Which can lead to subtle security vulnerabilities which are potentially hard to spot.

runtime/near-vm-runner/RUNTIMES.md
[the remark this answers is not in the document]
matklad: :thinking: that's a very good point actually, and something I didn't consider before. To give a specific example: wasmer0 has this fun property that some memory accesses wrap around, rather than trap (toroidal virtual memory). I thought that that's a weird quirk, but not too bad, as it doesn't violate security. But this reasoning is wrong. While indeed modulo arithmetic for addresses doesn't allow the contract to escape sandbox, it may make the smart-contracts themselves exploitable.

runtime/near-vm-runner/RUNTIMES.md
nagisa: 2.1.0 calling convention miscompile is not strictly an example of the issue per se. It was a problem with how calling convention was handled and the issue was introduced by changes to the area of code that was dealing with calling conventions.

On the other hand, it is in some sense such an example. In order to reproduce this issue a number of other instructions preceding the miscompiled call were necessary. These instructions seem to have been responsible for setting up the global state just the right way to trigger the invalid case in the code for call codegen.

runtime/near-vm-runner/RUNTIMES.md
nagisa: I attempted rewording the sections around performance to introduce more detailed view of the requirements and the parts that go into it.

runtime/near-vm-runner/RUNTIMES.md
```diff
@@ -0,0 +1,214 @@
+This document describes the considerations that go into picking a Wasm runtime as well as the
+conclusions we made after evaluating a number of VM backends and frontends.
+
+# Criteria
+
+There are a number criteria which have been used to evaluate a Wasm runtime for use in a NEAR
+validator. Some of the criteria are already listed in the [FAQ] document and here we take a more
```
bowenwang1996, line 7: I don't think we should specify `validator` here. For this purpose, it should be the same for all nodes that validate some shard

runtime/near-vm-runner/RUNTIMES.md
```diff
@@ -0,0 +1,214 @@
...
+thorough look. Listed roughly in the order of importance:
+
+* security – how well does the runtime deal with untrusted input;
+* correctness – does the runtime do what it claims to do;
+* reliability – how much confidence is there in the implementation of runtime;
+* platform support – does the runtime support targets that we want to target;
+* performance – how quickly can the runtime execute the requested operations;
```
bowenwang1996, line 14: This is interesting. What is reason for putting platform support above performance?

runtime/near-vm-runner/RUNTIMES.md
```diff
@@ -0,0 +1,214 @@
...
+the memory is reused by the contract without zeroing it, or where the memory accesses aren't
+validated correctly and allow access outside of the memory regions allocated for the contract.
+
+## Correctness
+
+A VM implementation must implement the Wasm specification precisely in order for it to be
+considered correct. Typically the Wasm code will be generated as an output by compiling a program
+implemented in a higher level programming language. Often the compiler will optimize the produced
+Wasm code with the understanding that Wasm instructions behave as specified in the specification.
+Were a VM implementation deviate from specification in some way, there's a non-negligible risk that
```
bowenwang1996, line 44:
```suggestion
Were a VM implementation to deviate from specification in some way, there's a non-negligible risk that
```

runtime/near-vm-runner/RUNTIMES.md
```diff
@@ -0,0 +1,214 @@
...
+the contract would invoke behaviours not intended by the contract authors (for example allow
```
bowenwang1996, line 45:
```suggestion
a contract would invoke behaviours not intended by the contract authors (for example allow
```

runtime/near-vm-runner/RUNTIMES.md
```diff
@@ -0,0 +1,214 @@
...
+correctly. It is well known that the x87 co-processor does not necessarily round the results
+correctly, so a special attention to how a runtime implements the floating point operations on the
+32-bit x86 targets is warranted.
+
+## Reliability
+
+The required and desirable properties of a runtime implementation must remain true while the
+changes are made to it. Compilers and, by extension, runtimes are complex systems and a poor
+architecture may result in even simple changes introducing correctness or security bugs, which
+aren't necessarily easily discovered.
```
bowenwang1996, line 74:
```suggestion
aren't necessarily easily discoverable.
```

runtime/near-vm-runner/RUNTIMES.md
```diff
@@ -0,0 +1,214 @@
...
+
+A reliable code base in this sense should be architected in such a way that the algorithms they
+implement and the way they execute Wasm code remains correct by construction and is easy to verify.
+
+Size of a test suite, architecture of the code base and the number of developers/users are all
```
bowenwang1996, line 79:
```suggestion
Size of a test suite, architecture of the code base, and the number of developers/users are all
```

runtime/near-vm-runner/RUNTIMES.md
```diff
@@ -0,0 +1,214 @@
...
+### Compilation
+
+The NEAR protocol charges gas fees to deploy a contract. One part of a deployment involves
+preparing the contract Wasm code for execution. This may involve instrumentation, compilation,
+serialization of the machine code and other similar operations.
+
+An estimate of the deployment cost will typically involve a function which uses the size of the
+input Wasm code as its primary input. Such a function can only exist if we have a good knowledge of
+the runtime’s time complexity properties are known. For our purposes a linear or `O(n log n)`
+relationship between the input size and execution time is the highest we can accept.
```
bowenwang1996, line 132: Actually, what is the reason why a quadratic algorithm is absolutely unacceptable? @matklad

runtime/near-vm-runner/RUNTIMES.md
```diff
@@ -0,0 +1,214 @@
...
+### Latency
+
+Whenever a function within a contract is invoked, the serialized form of the contract will be
+instantiated: loaded into memory and prepared for execution (though processes such as e.g.
+linking). Only then it is possible to start executing the contract code.
+
+When executing a tiny function part of a larger contract these operations will dominate and
+contribute greatly to the observed latency of the contract execution. These overheads contribute to
+the fees paid by anybody using the protocol, making any unnecessary overhead a potential roadblock
+in NEAR protocol's adoption.
```
bowenwang1996, line 143: N00b question: is the overhead here fundamentally unavoidable, i.e, is there some way to only deserialize part of the compiled wasm module to reduce the cost when only a tiny function is invoked?

runtime/near-vm-runner/RUNTIMES.md
```diff
@@ -0,0 +1,214 @@
...
+
+### Throughput
+
+The faster we can execute the contract code, the more elaborate are the ideas that developers can
+express via contracts running on the NEAR protocol. The cheaper it is to execute contracts, the
+more approachable and usable the protocol is, so runtime performance is a direct consequence of the
+usability of the whole system as a whole.
+
+One of more costly operations that occurs in a typical contract quite often is accounting for gas
```
bowenwang1996, line 152:
```suggestion
One of the more costly operations that occurs in a typical contract quite often is accounting for gas
```

runtime/near-vm-runner/RUNTIMES.md
```diff
@@ -0,0 +1,214 @@
...
+stack operations, pick more efficient instructions and sometimes avoid some operations such as NaN
+canonicalization altogether.
+
+Presence of this state comes at a cost of testability of implementation. The behaviour of the
+codegen can depend on prior Wasm instructions. Because of this there's a combinatorial explosion of
+the test cases that need to be considered and verified before confidence in the results can be
+established. Relatedly, at times it may also be difficult to validate changes made to the
+implementation of the `wasmer-singlepass` codegen. The global state is definitely a source of
+potential spooky action at a distance problems where changing code generation of a specific
+instruction affects correctness and behaviour of another one.
```
bowenwang1996, line 185: I have trouble parsing this sentence. Is there a typo somewhere?

bowenwang1996 (commented):
This is great! It provides a very good overview of the problems we face and our priorities. One question: it seems that we did not cover wasmtime in the document. Is that intentional?

runtime/near-vm-runner/RUNTIMES.md
nagisa: I reasoned about this in https://github.com/near/nearcore/pull/5881#discussion_r771518710. In short, I think our requirements to support `x86_64-linux` bring this above performance. But given that the question comes up a second time I'm wondering if either the description is lacking or this really should be less important compared to performance.

runtime/near-vm-runner/RUNTIMES.md
nagisa: In theory it is possible to only load the functions necessary for execution by constructing a call graph. In practice a non-negligible number of contracts will utilize the `call_indirect` instruction which makes this sort of analysis much more complicated.

An alternative could be to load the machine code into memory lazily, as it is accessed. This is exactly what operating systems like Linux do when they execute a program.

runtime/near-vm-runner/RUNTIMES.md
nagisa: Gave a shot at rewording this.

runtime/near-vm-runner/RUNTIMES.md
nagisa: Reworded some occurences of validator to refer to the NEAR protocol instead, where makes sense.

runtime/near-vm-runner/RUNTIMES.md
matklad: The max size of the contract is 4mb, so we can have up to roughly `10^6` "things" (functions, instruction in a single function, totall call depth, etc). `10^6 ^ 2` would be prohibitively expensive to compute. 

That being said, we *could* accept wore time complexities if we reflect them in the cost model. For example, if we now that our register allocation is, eg, quadratic in the number of local variables, we can charge quadratic cost for that. Non-linear cost models feel like the can of worms though.

nagisa: I think it could be feasible to utilize quadratic algorithms for select stages of our runtime if we did meter and charge for those stages separately from everything else. That way the linear operations with larger constant factors would be isolated from the quadratic stuff with potentially small factors and either approach would be reasonably feasible.

That said, since we're trying to estimate fees from the input wasm code, establishing the relationship between input wasm and the work an algorithm such as regalloc would do would probably involve a fair amount of tessomancy.

runtime/near-vm-runner/RUNTIMES.md
```diff
@@ -0,0 +1,218 @@
+This document describes the considerations that go into picking a Wasm runtime as well as the
+conclusions we made after evaluating a number of VM backends and frontends.
+
+# Criteria
+
+A number criteria have been used to evaluate a Wasm runtime for use in the implementation of the
+NEAR protocol. Some of the criteria are already listed in the [FAQ] document and this document
+gives a more thorough look. Listed roughly in the order of importance:
+
+* security – how well does the runtime deal with untrusted input;
```
olonho, line 10: "how well" is somewhat vague, we'd better define what are exact criteria here: i.e. have ability to limit resource consumption, trap on incorrect input, etc.

runtime/near-vm-runner/RUNTIMES.md
```diff
@@ -0,0 +1,218 @@
...
+* correctness – does the runtime do what it claims to do;
+* reliability – how much confidence is there in the implementation of runtime;
+* platform support – does the runtime support targets that we want to target;
```
olonho, line 13: Guess this sums up to "Linux/x86 for development and deployment and macOS arm/x64 for development".

runtime/near-vm-runner/RUNTIMES.md
olonho: Guess we need to define "requested operations".

nagisa: I want to avoid defining specifics in this list of criteria as the requirements section that comes after is better suited for elaboration in this aspect.

Re-reading the section on the performance requirements below I think it does list the kinds of operations we tend to ask a runtime to execute, but I do agree that the wording does not necessarily make the connection obvious.

runtime/near-vm-runner/RUNTIMES.md
nagisa: Correct, this is also described in the Plaform Support section within Requirements.

runtime/near-vm-runner/RUNTIMES.md
nagisa: Right the requirement section lists some of the… requirements with regards to security but I definitely missed e.g. resource consumption aspect. Will add that in.

matklad (approved):

Issue #4791, opened 2021-09-06
Feature request of Unsafe fast start of nearcore


khorolets:
After the recent incident with Indexer for Explorer, we've started the <span class="error">["Improvement uptime" discussion]</span>([https://github.com/near/near-indexer-for-explorer/discussions/161](https://github.com/near/near-indexer-for-explorer/discussions/161))

One of the important stuff for an indexer is to be able to start very quickly even if it is not safe to do so, but very required.

We want to make sure we can quickly restart indexer-for-explorer even on testnet (where the genesis file is huge), maybe at a cost of adding `--UNSAFE-fast-start` flag which would disable **genesis validation in nearcore somehow** (I believe that is the main contributor to the start time)

/cc @frol

frol:
@janewang @bowenwang1996 I would like to ask Node Experience team to consider implementing this feature.

Currently, Indexer Framework uses `nearcore::config::load_config_without_genesis_records`, but even then it takes tens of minutes to boot on testnet

https://github.com/near/nearcore/blob/b84b3fcdca7e3c221fa4adee35efeac6c2170637/chain/indexer/src/lib.rs#L104-L105

I feel we might want the same feature for `neard`. What do you think?

bowenwang1996:
> I feel we might want the same feature for neard. What do you think?

@frol how urgent is this request? @posvyatokum could you take a look?

frol:
This feature is not urgent at the moment, but it would help a lot in a face of incidents that may happen any day

nikurt:
Did some experiments with commenting out genesis validation on an archival testnet node. Disabling validation reduces the time between running `neard` and it getting to opening the DB by 1 minute.

## Without validation: 59 sec and 47 sec

```
Nov 29 19:23:43.235  INFO neard: Version: 1.22.0, Build: 5a6fb2bd2-modified, Latest Protocol: 48
Nov 29 19:24:42.306  WARN near_chain_configs::genesis_validate: Skipping validation of genesis
Nov 29 19:24:42.306  INFO near: Opening store database at "/home/ubuntu/.near/data"
```

```
Nov 29 19:31:51.720  INFO neard: Version: 1.22.0, Build: 5a6fb2bd2-modified, Latest Protocol: 48
Nov 29 19:32:38.803  WARN near_chain_configs::genesis_validate: Skipping validation of genesis
Nov 29 19:32:38.803  INFO near: Opening store database at "/home/ubuntu/.near/data"
[...]

PR #5888, opened 2021-12-17
Avoid reading the genesis json file twice upon startup

nikurt:
`nearcore` reads the genesis json file once to get GenesisConfig, and the second time to read the records and get Genesis.
Also, confusingly, the function `load_config_without_genesis_records` does read the genesis records in all cases.

This PR drops `load_config_without_genesis_records` in favor of `load_config` and ensures the genesis json file is read exactly once.

Additionally, this PR adds a command line option to skip genesis validation.

Next steps: Figure out why opening RocksDB takes minutes.

Test results on an archival testnet node:
`neard-1.23 run`: 115 seconds from start until starting to open RocksDB
`neard-1.23-with-fix run`: 90 seconds from start until starting to open RocksDB
`neard-1.23-with-fix --unsafe-fast-startup run`: 67 seconds from start until starting to open RocksDB

Issue #4791


core/chain-configs/src/genesis_config.rs
```diff
@@ -468,37 +468,75 @@ impl GenesisJsonHasher {
...
         let genesis = Self { config, records: GenesisRecords(vec![]), records_file };
-        validate_genesis(&genesis);
+        match genesis_validation {
+            GenesisValidationMode::Full => {
+                validate_genesis(&genesis);
+            }
+            GenesisValidationMode::UnsafeFast => {
+                warn!("Skipped genesis validation");
+            }
+        }
```
matklad, line 507: Might make sense to dedupe the code. Usually it's best when there's one "primary constructor", everything else delegates to.

core/chain-configs/src/genesis_config.rs
```diff
@@ -468,37 +468,75 @@ impl GenesisJsonHasher {
...
             serde_json::from_reader(reader).expect("Failed to deserialize the genesis records.");
-        validate_genesis(&genesis);
+        match genesis_validation {
+            GenesisValidationMode::Full => {
+                validate_genesis(&genesis);
+            }
+            GenesisValidationMode::UnsafeFast => {
+                warn!("Skipped genesis validation");
+            }
+        }
```
matklad, line 523: Oh, with thee copies, I feel strongly that we should dedupe this

core/chain-configs/src/genesis_config.rs
```diff
@@ -468,37 +468,75 @@ impl GenesisJsonHasher {
...
+    Full,
+    UnsafeFast,
+}
+
 impl Genesis {
-    pub fn new(config: GenesisConfig, records: GenesisRecords) -> Self {
+    pub fn new(
+        config: GenesisConfig,
+        records: GenesisRecords,
+        genesis_validation: GenesisValidationMode,
```
matklad, line 480: All calls to `new` pass in a literal `GenesisValidationMode` so perhaps this signature doesn't need to change?

runtime/runtime-params-estimator/src/main.rs
```diff
@@ -117,7 +118,7 @@ fn main() -> anyhow::Result<()> {
             )
             .expect("failed to init config");
 
-            let near_config = load_config(&state_dump_path);
+            let near_config = load_config(&state_dump_path, GenesisValidationMode::Full);
```
bowenwang1996, line 121: Do we really need to validate genesis in parameter estimator? @matklad

bowenwang1996 (commented):
@nikurt could you document somewhere how to start neard without genesis validation?

core/chain-configs/src/genesis_config.rs
```diff
@@ -470,39 +470,77 @@ impl GenesisJsonHasher {
...
+        genesis.validate(genesis_validation)
     }
 
+    fn validate(self, genesis_validation: GenesisValidationMode) -> Self {
+        match genesis_validation {
+            GenesisValidationMode::Full => {
+                validate_genesis(&self);
+            }
+            GenesisValidationMode::UnsafeFast => {
+                warn!("Skipped genesis validation");
```
frol, line 539: let's add `target` to the warn!

frol (approved):

nikurt:
> @nikurt could you document somewhere how to start neard without genesis validation?

@bowenwang1996. Done, added flag documentation.

bowenwang1996 (approved):

frol (approved):

runtime/runtime-params-estimator/src/main.rs
matklad: I don't think it matters either way -- I'd expect the performance to be dominated by other stuff, and an extra sanity check won't harm.

matklad (approved):

khorolets (approved):

PR #5889, opened 2021-12-18
Upgrade clap to `3.0.0-beta5`

pmnoxx:
Previously, we couldn't upgrade `clap` to `3.0.0-beta5` due to technical issues. 
I figured out a walk-around, that keeps the api exactly the same.

Clap is expecting to release `rc0` within a week: https://github.com/near/nearcore/pull/5694#issuecomment-987950282
There is nothing holding us back from upgrading.

See previous discussion https://github.com/near/nearcore/pull/5694


pmnoxx:
```                                                                                                                                       
2021-12-18T02:09:29.709732Z  INFO refactor::cargo: cmd="cargo-deny --all-features check bans"                                                                                                                                                                      
error[B004]: found 2 duplicate entries for crate 'textwrap'                                                                                                                                                                                   
    ┌─ /home/pmnoxx/repos/nearcore.master/Cargo.lock:436:1                                                                                                                                                                                    
    │                                                                                                                                                                                                                                         
436 │ ╭ textwrap 0.11.0 registry+https://github.com/rust-lang/crates.io-index                                                                                                                                                                 
437 │ │ textwrap 0.14.2 registry+https://github.com/rust-lang/crates.io-index                                                                                                                                                                 
    │ ╰─────────────────────────────────────────────────────────────────────^ lock ent
```

neard/src/cli.rs
```diff
@@ -15,7 +15,7 @@ use tracing::metadata::LevelFilter;
 use tracing_subscriber::EnvFilter;
 
 /// NEAR Protocol Node
-#[derive(Clap)]
+#[derive(clap_derive::Parser)]
```
nikurt, line 18: I haven't seen the docs for the most recent version of `clap`, but as a reader I'm somewhat confused.
Why do we need `use clap::Parser` if here we derie `clap_derive::Parser`?

Checked the docs, the first example derives `clap::Parser`:
https://docs.rs/crate/clap/3.0.0-rc.7

nikurt (approved):

neard/src/cli.rs
pmnoxx: @nikurt Good question.
Note that `use clap::Parser` and `clap_derive::Parser`, share the same name, but one is trait, another is `proc-macro`.

`use clap::Parser` is a trait, which adds `parse` method to ` CliArgs::parse();`

`clap_derive::Parser` is needed for derive.
In Rust all derives / `proc_macro` crates, need to be separate.

neard/Cargo.toml
```diff
@@ -13,7 +13,8 @@ path = "src/main.rs"
 name = "neard"
 
 [dependencies]
-clap = "=3.0.0-beta.2"
+clap = { version = "3.0.0-rc.7", features = ["cargo"] }
+clap_derive = "3.0.0-rc.7"
```
matklad, line 17:
```suggestion
clap = { version = "3.0.0-rc.7", features = ["cargo", "derive"] }
```

I believe that's the intended idiomatic way of using clap. For something neare the bottom of dep graph, like `serde`, depdending on runtime crate and derive crate separately *might* help with compile times, but that's not the case for clap.

pmnoxx: @matklad Thanks, I updated the code.

pmnoxx:
```
@matklad: /runtime/runtime-params-estimator/ /runtime/near-vm-runner-standalone/
@mm-near: /neard/
@khorolets: /tools/indexer/
```
@matklad @mm-near @khorolets Can you take a look?

neard/src/cli.rs
```diff
@@ -82,7 +82,7 @@ impl NeardOpts {
     }
 }
 
-#[derive(Clap)]
+#[derive(clap::Parser)]
```
matklad, line 85: :heart: qualified imports!

matklad (approved):

pmnoxx:
```
@bowenwang1996 @mm-near: neard/
@khorolets: tools/indexer/
```
@bowenwang1996 @khorolets @mm-near Can you take a look?

pmnoxx:
```
@bowenwang1996 @mm-near: neard/
@khorolets: tools/indexer/
```
@bowenwang1996 @khorolets @mm-near Can you take a look?

nikurt (approved):

mm-near (approved):

runtime/near-vm-runner-standalone/Cargo.toml
```diff
@@ -22,8 +22,7 @@ to make sure it has expected behavior once deployed to the blockchain.
 [dependencies]
 serde = { version = "1", features = ["derive"] }
 serde_json = "1"
-clap = "=3.0.0-beta.2"
-clap_derive = "=3.0.0-beta.2"
+clap = { version = "3.0.0-rc.7", features = ["derive"] }
```
nagisa, line 25: This should be at least `^3.0.0`, I think?

runtime/near-vm-runner-standalone/src/main.rs
```diff
@@ -3,7 +3,7 @@
 mod script;
 
 use crate::script::Script;
-use clap::Clap;
+use clap::Parser;
```
nagisa, line 6: Is this use necessary at all, since the derive is using `clap::Parser` in the first place

pmnoxx: @nagisa That's a good question. 
We are importing `clap::Parser` which is used as trait. It's needed for `let cli_args = CliArgs::parse();`.

runtime/near-vm-runner-standalone/Cargo.toml
pmnoxx: The version, which is in `Cargo.lock` matters. We use `3.0.12`.

I will change this to be `version = "3.0.0"` to be less confusing.

PR #5894, opened 2021-12-18
feat: add `.github/workflows/nearcore.yml `

pmnoxx:
We can make our CI much faster by switching to github pipelines.

For example, running `clippy` check took only 25s,
same for `cargo-deny`.

Advantages:
- caching - part of build process is cached, and only need to be redone on changes to `Cargo.lock`.
- reliability - github does good job maintaining it
- works with forks - each fork can run pipeline, without even having to submit PR
- full build only needed when updating `Cargo.lock`.
- We can add self hosted runners: https://docs.github.com/en/actions/hosting-your-own-runners/adding-self-hosted-runners

By default runners only have 2 cores, that's why build may take more time,
but on the other hand, we don't have to build dependencies anymore every time.

We could use custom runners, with more resources.


matklad:
cc @chefsale I guess? Personally, I am all for switching to stock GitHub actions -- the main benefit I see there is that it makes the project significantly less weird, and that's a rather important property for the open ecosystem we are building. There's also personal productivity angle -- buildkite is complemelty opaque to me, while GHA is something I am used to. 

I haven't actually compared buildkite with GHA in details though. 

Even if this a good idea, this probably should be a bigger than a PR discussion.

.github/workflows/rust.yml
```diff
@@ -0,0 +1,80 @@
...
+        cargo clippy --all-features --no-deps -p near-cache -p near-performance-metrics -p near-performance-metrics-macros -p near-rate-limiter -p near-stable-hasher -- -D clippy::all
+
+  cargo-deny:
+
+    runs-on: ubuntu-20.04
+
+    steps:
+    - uses: actions/checkout@v2
+    - uses: actions/cache@v2
+      id: deny-cache
```
miraclx, line 40: cargo deny has no build artifacts, there's no need to cache. If anything, we only need to keep track of changes to the lockfiles.

.github/workflows/rust.yml
```diff
@@ -0,0 +1,80 @@
...
+      run: |
+        python3 -m pip install -q --user yapf
+        yapf -pdr pytest scripts
+
+  cargo-fmt:
+    runs-on: ubuntu-20.04
+
+    steps:
+    - uses: actions/checkout@v2
+    - uses: actions/cache@v2
```
miraclx, line 70: `cargo fmt` doesn't require caching build artifacts either.

.github/workflows/rust.yml
```diff
@@ -0,0 +1,80 @@
...
+        path: |
+          ~/.cargo/bin/
+          ~/.cargo/registry/index/
+          ~/.cargo/registry/cache/
+          ~/.cargo/git/db/
+          target/
+        key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}-clippy
+    - name: Run clippy
+      run: |
+        cargo clippy --all-features --no-deps -p near-cache -p near-performance-metrics -p near-performance-metrics-macros -p near-rate-limiter -p near-stable-hasher -- -D clippy::all
```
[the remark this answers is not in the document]
pmnoxx: @miraclx Well, we should first remaining 900 Clippy errors first. I wrote fixes for `near-network`, and I'm still waiting for reviews.

.github/workflows/rust.yml
pmnoxx: @miraclx The caching is for the installation of `cargo-deny`. It's not preinstalled, and it takes 6m to install.

mm-near:
@pmnoxx - thanks for the suggestion. As this is the larger change - let's discuss this during out team meeting - and later also with SREs/other teams. 

I've personally never used GitHub Pipelines before - so I'll need some time to read up on them.

.github/workflows/rust.yml
```diff
@@ -0,0 +1,80 @@
+name: Rust
```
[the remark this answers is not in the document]
miraclx:
```suggestion
name: nearcore
```

PR #5897, opened 2021-12-18
Prefer `&str` to `&String`

mina86:
Taking `&String` argument limits how function can be used without
providing any additional features.  Specifically, when caller has
a string slice they have to make unnecessary conversion to String
before passing it to the function.

Similarly, returning `&String` doesn’t offer benefits over returning
`&str` since there’s nothing that can be done with the former that
cannot be done with the latter as well.

As such, prefer using `&str` instead of `&String`.

### DIFF
--- a/integration-tests/src/tests/nearcore/node_cluster.rs
+++ b/integration-tests/src/tests/nearcore/node_cluster.rs
@@ -42,7 +42,7 @@ pub fn start_nodes(
             if i == 0 { first_node } else { open_port() },
             genesis.clone(),
         );
-        rpc_addrs.push(near_config.rpc_addr().unwrap().clone());
+        rpc_addrs.push(near_config.rpc_addr().unwrap().to_owned());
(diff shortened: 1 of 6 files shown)


integration-tests/src/tests/nearcore/node_cluster.rs
```diff
@@ -42,7 +42,7 @@ pub fn start_nodes(
             if i == 0 { first_node } else { open_port() },
             genesis.clone(),
         );
-        rpc_addrs.push(near_config.rpc_addr().unwrap().clone());
+        rpc_addrs.push(near_config.rpc_addr().unwrap().to_owned());
```
matklad, line 45: I personally used to use semantically pedantically correct `to_owned()` in this situations, but since switched to `.to_string()`, as that is a tiny bit easier in terms of cognitive load. Both are fine though!

matklad (approved):
:+1:

nikurt (approved):

mm-near (approved):

PR #5901, opened 2021-12-19
Synchronize block and chunk production

EgorKulikov:
Hotfix to address #5688
While all chunk producers are block producers we should use same validator to produce chunks that would later produce block out of them, to avoid network delays

### DIFF
--- a/chain/chain/src/tests/simple_chain.rs
+++ b/chain/chain/src/tests/simple_chain.rs
@@ -25,7 +25,7 @@ fn empty_chain() {
     #[cfg(feature = "nightly_protocol")]
     assert_eq!(hash, CryptoHash::from_str("DHFJfDH25yEPmEpKrUN9upXY3gFcCmsGyHMTk2sugyXd").unwrap());
     #[cfg(not(feature = "nightly_protocol"))]
-    assert_eq!(hash, CryptoHash::from_str("BkrZeGCDnYJGAdc3G1fb1P8FMbnNDhBn6w5DjhQeNAdp").unwrap());
+    assert_eq!(hash, CryptoHash::from_str("7mewJ6tbV1hZBg47qaCjtRfPQG4fBHykg2mzBjEbXwv9").unwrap());
     assert_eq!(count_utc, 1);
     assert_eq!(count_instant, 0);
 }
@@ -49,7 +49,7 @@ fn build_chain() {
     #[cfg(not(feature = "nightly_protocol"))]
     assert_eq!(
         prev_hash,
-        CryptoHash::from_str("6C474NRQK1cLkGYrdaCpeVPKbBbsTxGeB7ZZWkdbD3uL").unwrap()
+        CryptoHash::from_str("HdzZUsY2p3hmursbjgcG2PdYaw3UgchA9rXnFoEJLGH1").unwrap()
     );
 
     for i in 0..4 {
@@ -72,7 +72,7 @@ fn build_chain() {
     #[cfg(not(feature = "nightly_protocol"))]
     assert_eq!(
         chain.head().unwrap().last_block_hash,
-        CryptoHash::from_str("5pXLqgQe98JSdhtQ66VQFjQzBRPdZaEiMaTpdGuziF4H").unwrap()
+        CryptoHash::from_str("3vG7bLc3AjfYKdESnu5qNX24gjD6SKB1PuiRBDpcBjWC").unwrap()
     );
 }
 
--- a/chain/epoch_manager/src/validator_selection.rs
+++ b/chain/epoch_manager/src/validator_selection.rs
@@ -546,13 +546,11 @@ mod tests {
             let bp = epoch_info.sample_block_producer(h);
             for shard_id in 0..num_shards {
                 let cp = epoch_info.sample_chunk_producer(h, shard_id);
-                if checked_feature!("stable", SynchronizeBlockChunkProduction, PROTOCOL_VERSION)
-                    && !checked_feature!(
-                        "protocol_feature_chunk_only_producers",
-                        ChunkOnlyProducers,
-                        PROTOCOL_VERSION
-                    )
-                {
+                if !checked_feature!(
+                    "protocol_feature_chunk_only_producers",
+                    ChunkOnlyProducers,
+                    PROTOCOL_VERSION
+                ) {
                     assert_eq!(bp, cp);
                 }
             }
--- a/chain/epoch_manager/src/validator_selection.rs
+++ b/chain/epoch_manager/src/validator_selection.rs
@@ -511,8 +511,9 @@ mod tests {
 
     #[test]
     fn test_block_producer_sampling() {
+        let num_shards = 2;
         let epoch_config = create_epoch_config(
-            1,
+            num_shards,
             2,
             0,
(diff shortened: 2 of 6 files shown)


chain/epoch_manager/src/validator_selection.rs
```diff
@@ -543,6 +544,18 @@ mod tests {
         let mut counts: [i32; 2] = [0, 0];
         for h in 0..100_000 {
             let bp = epoch_info.sample_block_producer(h);
+            for shard_id in 0..num_shards {
+                let cp = epoch_info.sample_chunk_producer(h, shard_id);
+                if checked_feature!("stable", SynchronizeBlockChunkProduction, PROTOCOL_VERSION)
+                    && !checked_feature!(
+                        "protocol_feature_chunk_only_producers",
+                        ChunkOnlyProducers,
+                        PROTOCOL_VERSION
```
bowenwang1996, line 553: why does this check appear in the test? I thought unless otherwise specified, we always test the stable protocol

integration-tests/tests/client/challenges.rs
```diff
@@ -613,6 +613,10 @@ fn test_fishermen_challenge() {
 /// If there are two blocks produced at the same height but by different block producers, no
 /// challenge should be generated
 #[test]
+// Something weird happens here. For len_in_blocks = 13 this test passes for versions up to 49,
+// but fails for version 50 (because of chunk validator sampling changes). But if we set it to 20,
+// it fails for version 49 as well
```
bowenwang1996, line 618: It is probably worth investigating. While challenges are not enabled by any means, the failure here seems related to your change and not to challenges per se.

bowenwang1996 (commented):
I would like to see more testing to this change:
* Could we test with at least 4 shards?
* Could we test that when blocks are missed, validators do not get kicked out for missing chunks? I think we could even add some debug assert that `chunk_produced/chunk_expected` should not be smaller than `block_produced/block_expected`

chain/chain/src/tests/simple_chain.rs
```diff
@@ -25,7 +25,7 @@ fn empty_chain() {
     #[cfg(feature = "nightly_protocol")]
     assert_eq!(hash, CryptoHash::from_str("DHFJfDH25yEPmEpKrUN9upXY3gFcCmsGyHMTk2sugyXd").unwrap());
     #[cfg(not(feature = "nightly_protocol"))]
-    assert_eq!(hash, CryptoHash::from_str("BkrZeGCDnYJGAdc3G1fb1P8FMbnNDhBn6w5DjhQeNAdp").unwrap());
+    assert_eq!(hash, CryptoHash::from_str("7mewJ6tbV1hZBg47qaCjtRfPQG4fBHykg2mzBjEbXwv9").unwrap());
```
mm-near, line 28: (unrelated to this line of code) can you improve the PR description ? (currently it just says "hotfix for XX")

core/primitives/src/version.rs
```diff
@@ -128,6 +128,7 @@ pub enum ProtocolFeature {
     /// https://github.com/near/NEPs/pull/167 for general description, note that we would not
     /// introduce chunk-only validators with this feature
     AliasValidatorSelectionAlgorithm,
+    SynchronizeBlockChunkProduction,
```
mm-near, line 131: add comment on what this feature is about

nearcore/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "nearcore"
-version = "0.0.0"
+version = "1.23.0"
```
mm-near, line 3: we never updated this version in the past .. is it needed now? or should we leave it as 0.0. ?

neard/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "neard"
-version = "0.0.0"
+version = "1.23.0"
```
mm-near, line 3: same as in the other toml file

core/primitives/src/epoch_manager.rs
```diff
@@ -832,7 +832,22 @@ pub mod epoch_info {
...
+                    let seed = if checked_feature!(
+                        "stable",
+                        SynchronizeBlockChunkProduction,
+                        protocol_verstion
+                    ) && !checked_feature!(
+                        "protocol_feature_chunk_only_producers",
+                        ChunkOnlyProducers,
+                        protocol_verstion
+                    ) {
+                        // 32 bytes from epoch_seed, 8 bytes from height
```
mm-near, line 845: you should mention that here, we are using the same seed as block producer -- which should lead to matching results.

mm-near (changes requested):

integration-tests/tests/client/challenges.rs
EgorKulikov: It fails with len_in_blocks = 20 even without my change

nearcore/Cargo.toml
EgorKulikov: This is part of release commit for 1.23, it shows with my changes just because I started with that tag

integration-tests/tests/client/challenges.rs
yordanmadzhunkov: I also had problems with that test during working on other issues. It should be investigated

core/primitives/src/epoch_manager.rs
```diff
@@ -832,7 +832,23 @@ pub mod epoch_info {
...
+                        // This is same seed that used for determining block producer
```
matklad, line 846: Should this be a `fn block_produce_seed`  then?

nearcore/Cargo.toml
EgorKulikov: @bowenwang1996 This was meant for testing, tbh. To merge this to master would require dealing with version issues, and I, of course, would rebase it (or create separate PR)

mm-near (approved):

core/primitives/src/epoch_manager.rs
```diff
@@ -832,7 +827,19 @@ pub mod epoch_info {
                     shard_cps[(height as u64 % (shard_cps.len() as u64)) as usize]
                 }
                 Self::V3(v3) => {
-                    let seed = {
+                    let protocol_verstion = self.protocol_version();
+                    let seed = if checked_feature!(
+                        "stable",
+                        SynchronizeBlockChunkProduction,
+                        protocol_verstion
```
bowenwang1996, line 834: looks like a typo here :)

EgorKulikov: where?

bowenwang1996: fixed it myself :)

bowenwang1996:
Merged to 1.23.1

core/primitives/src/epoch_manager.rs
EgorKulikov: Ah, I see now, I though you meant there is typo in feature name

integration-tests/tests/client/challenges.rs
bowenwang1996: @EgorKulikov I suggest creating an issue for this

EgorKulikov: @bowenwang1996 #5914

PR #5902, opened 2021-12-19
fix: Fix memory leak in `near-network` affecting `indexer` and probably all nodes

pmnoxx:
There is a huge memory leak, which is affecting indexer and `near-network`.

After rewriting `memory-analyzer` from `C++` to `Rust`. I noticed that, there is a memory leak in `Codec`. The codec is not being cleared after it's length drops to 0.

I added a logic for clearing both `read` and `write` buffers. This should help indexer, and regular nodes as well

```shell
> sudo rust-memory-analyzer analyze --pid `pidof indexer-example` --print-raw-symbols 2>&1

 INFO memory_analyzer::analyze: func="_ZN5bytes9bytes_mut8BytesMut13reserve_inner17h599b80e87834931bE" count=52 size_mb=993
 INFO memory_analyzer::analyze: took=1.116597038s total_size_mb=1774
 INFO memory_analyzer::analyze: took=1.116604928s resident_but_not_used_mb=144 allocated_with_proxy_mb=1055 mapped_files_mb=575
```

After having indexer run for 10m, it was using almost `2gb` before, now it's down to `900mb`. That's over 1gb of memory saving.


chain/network/src/peer/codec.rs
```diff
@@ -90,6 +91,10 @@ impl Decoder for Codec {
         } else {
             let res = Some(Ok(buf[4..4 + len as usize].to_vec()));
             buf.advance(4 + len as usize);
+            // Fix memory leak see (#TODO)
+            if buf.is_empty() && buf.is_empty().not() {
```
matklad, line 95:
```suggestion
            if buf.is_empty() && !buf.is_empty() {
```

One *can* `.not()`, but this isn't remotely idiomatic. 

Also -- this is `false`, no?

pmnoxx: @matklad I updated the logic. This isn't really ready for review, hence the "draft".

pmnoxx:
```
@mm-near: chain/network/ utils
```
@mm-near Can you take a look?

chain/network/src/peer/codec.rs
```diff
@@ -72,24 +72,28 @@ impl Decoder for Codec {
...
-        if buf.len() < 4 + len as usize {
+        if buffer.len() < 4 + len as usize {
             // not enough bytes, keep waiting
             Ok(None)
         } else {
-            let res = Some(Ok(buf[4..4 + len as usize].to_vec()));
-            buf.advance(4 + len as usize);
+            let res = Some(Ok(buffer[4..4 + len as usize].to_vec()));
+            buffer.advance(4 + len as usize);
+            // Fix memory leak see (#TODO)
```
bowenwang1996, line 93: Could we add the issue number?

chain/network/src/peer/codec.rs
```diff
@@ -72,24 +72,28 @@ impl Decoder for Codec {
...
+            if buffer.is_empty() && buffer.capacity() >= MIB as usize {
+                std::mem::take(buffer);
```
bowenwang1996, line 95: is it more idiomatic to use https://docs.rs/bytes/latest/bytes/struct.BytesMut.html#method.truncate ?

pmnoxx: Unfortunately, `truncate` doesn't free used memory, see https://docs.rs/bytes/latest/src/bytes/bytes_mut.rs.html#399-405

pmnoxx: and https://docs.rs/bytes/latest/src/bytes/bytes_mut.rs.html#484

bowenwang1996 (approved):

chain/network/src/peer/codec.rs
```diff
@@ -72,24 +72,28 @@ impl Decoder for Codec {
     type Item = Result<Vec<u8>, ReasonForBan>;
     type Error = Error;
 
-    fn decode(&mut self, buf: &mut BytesMut) -> Result<Option<Self::Item>, Self::Error> {
```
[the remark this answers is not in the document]
pmnoxx: Ok, we can keep it as `buf`.

chain/network/src/peer/codec.rs
pmnoxx: Thanks, I added it.

mm-near (approved):

PR #5904, opened 2021-12-20
refactor(themis): dyn error handling case analysis

miraclx:
Context: https://github.com/near/nearcore/pull/5724#discussion_r766659405

cc. @matklad

#### Interface Change

Thanks for the cleanup hints @matklad

##### Before

```console
(i) These packages have an incompatible `rust-version`: [expected: <=1.57.0, as defined in the `rust-toolchain`]
 • near-cache v0.0.0 (utils/near-cache/Cargo.toml) [found: 1.58.0]
```

##### After

```console
(i) These packages have an incompatible `rust-version`: [expected: <=1.57.0, as defined in the `rust-toolchain`]
 ↳ utils/near-cache/Cargo.toml (found: 1.58.0)
```

### DIFF
--- a/tools/themis/src/types.rs
+++ b/tools/themis/src/types.rs
@@ -1,0 +1,93 @@
+use std::fmt;
+use std::sync::Arc;
+
+use cargo_metadata::camino::Utf8PathBuf;
+
+use super::style;
+
+#[derive(Debug)]
+pub struct Package {
+    pub parsed: cargo_metadata::Package,
+    pub raw: toml::Value,
+}
+
+#[derive(Debug)]
+pub struct Workspace {
+    pub root: Utf8PathBuf,
+    pub members: Vec<Arc<Package>>,
+}
+
+#[derive(Debug)]
+pub struct PackageOutcome {
+    pub pkg: Arc<Package>,
+    pub value: Option<String>,
+}
+
+#[derive(Debug)]
+pub struct Expected {
+    pub value: String,
+    pub reason: Option<String>,
+}
+
+#[derive(Debug)]
+pub struct ComplianceError {
+    pub msg: String,
+    pub expected: Option<Expected>,
+    pub outliers: Vec<PackageOutcome>,
+}
+
+impl fmt::Display for ComplianceError {
+    fn fmt(&self, _: &mut fmt::Formatter<'_>) -> fmt::Result {
+        unimplemented!()
+    }
+}
+
+impl std::error::Error for ComplianceError {}
+
+impl ComplianceError {
+    pub fn report(&self, workspace: &Workspace) {
+        let mut report = format!(
+            "{c_heading}(i) {}:{c_none}{}",
+            self.msg,
+            match &self.expected {
+                None => "".to_string(),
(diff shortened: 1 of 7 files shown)


matklad (approved):
Lgtm, though I'd personally try to write this smaller still. 

It seems that we only need package in the outcome to do the following reporting: 

```rust
                "\n \u{2022} {c_name}{}{c_none} v{} {c_path}({}){c_none}{}",
                pkg.name,
                pkg.version,
                pkg.manifest_path.strip_prefix(&workspace.root).unwrap(),
```

Here, we use name, version and path fields. It seems that name and version are redundant -- they can easily be inferred from the path, and the path is what's interesting, because you are going to edit that file to fix the error. 

So, I'd rather just store `PathBuf` rather than `Package` in `PackageOutcome`. This'll remove the neeed for `unimplmented` Display and `report` function, as the relative path can then be computed at the point where we emit the error.

miraclx:
Agreed, and that's a neat insight. Especially since we have unversioned crates, the only place where the `version` info is needed, we have "expected/found" indicating the version. I'll clean it up.

tools/themis/src/types.rs
```diff
@@ -65,25 +64,21 @@ impl ComplianceError {
...
-            let Package { parsed: pkg, .. } = &**pkg;
+        for Outlier { path, found } in &self.outliers {
             report.push_str(&format!(
-                "\n \u{2022} {c_name}{}{c_none} v{} {c_path}({}){c_none}{}",
-                pkg.name,
-                pkg.version,
-                pkg.manifest_path.strip_prefix(&workspace.root).unwrap(),
-                match value {
+                "\n {c_path}\u{21b3} {}{c_none}{}",
+                path.strip_prefix(&workspace.root).unwrap(),
```
matklad, line 70: I'd push this logic to the place where we *create* an `Outlier`, that way we could get rid of the custom report.

tools/themis/src/types.rs
```diff
@@ -33,7 +32,7 @@ pub struct Expected {
 pub struct ComplianceError {
     pub msg: String,
     pub expected: Option<Expected>,
-    pub outliers: Vec<PackageOutcome>,
+    pub outliers: Vec<Outlier>,
```
matklad, line 35: Love the naming!

matklad (approved):

tools/themis/src/types.rs
miraclx: I'm unconvinced this would be any better. Aside from unnecessarily duplicating that strip call across all compliance checks, it doesn't alleviate the need for the custom report. It also walks back on the static ownership requirement for `ComplianceError` when used with `bail!()`

matklad: I mean that's the only reason why we need to pass in `workspace: &Workspace`. If we remove that, than `Display`'s signature would work, and we can do 

```
        if let Err(err) = rule(&workspace) {
            failed |= true;
            eprlintln!("{}", err.downcast::<types::ComplianceError>()?);
        }
```

matklad: No strong opinion though!

PR #5906, opened 2021-12-20
re: Replace `Arc<Store>` with `Store`

### LINKED ISSUE #4837: refactor: replace `Arc<Store>` with `Store` throughout the codebase?
We use `Arc<Store>` a bunch. But the Store is defined as 

https://github.com/near/nearcore/blob/3080cf7d305c8e1d25d09d378c7f0ab1f841ec27/core/store/src/lib.rs#L48-L52

Ie, it is already an `Arc`. I don't think there can be a conceivable reason for a double indirection.

*However*, I see that `#[derive(Clone)]` was a late addition: 

https://github.com/near/nearcore/pull/3310/files#diff-bc42e726bca3d291bcb277ae2857c8dd36ae5831d1190902b12082d74bc6e302R47

So maybe it's that derive that is wrong? Looking at the code, it seems that Clone is fine:

* there are no store methods which require exclusive access 
* `create_store` already creates an `Arc`, so we never even *have* exclusive access to the store. 

So I suggest replacing all `Arc<Store>` with just `Store`. While we are at it, we have a bunch of `get_store(&self) -> Arc<Store>` methods. I think a more idiomatic signature for those would be `store(&self) -> &Store`.

pmnoxx:
I was always curious, why we need `Pin` inside `Store`. It looks like we don't, and we can change
```
#[derive(Clone)]
pub struct Store {
    storage: Pin<Arc<dyn Database>>,
}
```
to
```
#[derive(Clone)]
pub struct Store {
    storage: Arc<dyn Database>,
}
```

Closes #4837

### DIFF
--- a/chain/chain/src/store.rs
+++ b/chain/chain/src/store.rs
@@ -1,7 +1,6 @@
 use std::collections::hash_map::Entry;
 use std::collections::{HashMap, HashSet};
 use std::io;
-use std::sync::Arc;
 
 use borsh::{BorshDeserialize, BorshSerialize};
 use lru::LruCache;
@@ -291,7 +290,7 @@ pub trait ChainStoreAccess {
 
 /// All chain-related database operations.
 pub struct ChainStore {
-    store: Arc<Store>,
+    store: Store,
     /// Genesis block height.
     genesis_height: BlockHeight,
     /// Latest known.
@@ -357,7 +356,7 @@ pub fn option_to_not_found<T>(res: io::Result<Option<T>>, field_name: &str) -> R
 }
 
 impl ChainStore {
-    pub fn new(store: Arc<Store>, genesis_height: BlockHeight) -> ChainStore {
+    pub fn new(store: Store, genesis_height: BlockHeight) -> ChainStore {
         ChainStore {
             store,
             genesis_height,
@@ -390,8 +389,8 @@ impl ChainStore {
         }
     }
 
-    pub fn owned_store(&self) -> Arc<Store> {
-        self.store.clone()
+    pub fn owned_store(&self) -> &Store {
+        &self.store
     }
 
     pub fn store_update(&mut self) -> ChainStoreUpdate<'_> {
@@ -571,7 +570,7 @@ impl ChainStore {
     ) -> Result<&HashMap<EpochId, HashSet<CryptoHash>>, Error> {
         option_to_not_found(
             read_with_cache(
-                &*self.store,
+                &self.store,
                 ColBlockPerHeight,
                 &mut self.block_hash_per_height,
                 &index_to_bytes(height),
@@ -766,7 +765,7 @@ impl ChainStore {
 
 impl ChainStoreAccess for ChainStore {
     fn store(&self) -> &Store {
-        &*self.store
+        &self.store
     }
     /// The chain head.
     fn head(&self) -> Result<Tip, Error> {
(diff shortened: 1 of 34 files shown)


core/store/src/lib.rs
```diff
@@ -44,11 +43,11 @@ mod trie;
 
 #[derive(Clone)]
 pub struct Store {
-    storage: Pin<Arc<dyn Database>>,
```
matklad, line 47: If we remove `Pin` here, we should remove `PhantomPinned` from `RocksDB` struct as well. 

I *think* these are fine to remove, but what gives me a pause is that I don't understand the reasonong behind their original addition: https://github.com/near/nearcore/pull/3025/files#r772310668.

pmnoxx: @bowenwang1996 I see, you reviewed [#3025](https://github.com/near/nearcore/pull/3025), could you tel us more about the reason why  `PhantomPinned` was added?

chain/chain/src/store.rs
```diff
@@ -390,7 +389,7 @@ impl ChainStore {
         }
     }
 
-    pub fn owned_store(&self) -> Arc<Store> {
+    pub fn owned_store(&self) -> Store {
```
matklad, line 392: This feels redundant now, as one can `.store().clone()`

core/store/src/lib.rs
```diff
@@ -44,11 +44,11 @@ mod trie;
 
 #[derive(Clone)]
```
matklad, line 45: @mm-near this feels like an important type to document. I am especially wondering about our threading strategy here -- this type is `Clone`, so many threads may access the database at the same time, and they can also modify it concurrently via `Store::store_update`. And then there are things like iterators which, at least judging from the types, can happen concurrently with modification.

matklad (approved):

nikurt (approved):

chain/chain/src/store.rs
pmnoxx: Fixed.

pmnoxx:
```
@mm-near: /chain/network/ /chain/ /nearcore/
@longarithm: /core/store/src/trie
```
@mm-near @longarithm Can you take a look?

pmnoxx:
```
@bowenwang1996 @mm-near: /chain/ /chain/network/ /nearcore/
@longarithm @mzhangmzz: /core/store/src/trie
```
@bowenwang1996 @longarithm @mm-near @mzhangmzz Can you take a look?

Longarithm (approved):

pmnoxx:
```
@longarithm @mzhangmzz: core/store/src/trie
```
@longarithm @mzhangmzz Can you take a look?

core/store/src/lib.rs
mzhangmzz: I can comment on how store is used in the chain and runtime level and @pmnoxx can comment on how it is accessed in `PeerManagerActor` and other network related threads. My guess is that the network threads access completely different database columns than chain and runtime would access, so they don't conflict with each other.

The general story is that although `Store` can be modified concurrently, in practice we don't do that. In chain and runtime, `Store` is used by three structs, `Chain`, `NightshadeRuntime` and `EpochManager`. `Chain` is accessed by one thread (`ClientActor`) while `NightshadeRuntime` and `EpochManager` are accessed by multiple threads (`ViewClientActor`, `ClientActor` and temporary threads when we apply blocks). The trick is, only `Chain` actually modifies store through `store_update`, `EpochManager` and `NightshadeRuntime` only use `Store` to read information, they never write (except for maybe during initialization).

mzhangmzz (approved):

PR #5907, opened 2021-12-20
feat: Don't cache `CacheError` inside `cache.rs`

pmnoxx:
Change caching logic to not cache catastrophic errors. in `Result<Result<wasmer::Module, CompilationError>, CacheError>`, the inner error should be cached, but the outer error should be propagated


matklad:
Ah, sorry, I forgot linking https://github.com/near/nearcore/pull/5856 from the issue. I think we should close this in favor of #5856, as that does things a bit better.

PR #5911, opened 2021-12-20
fix: don't log host functions by default

matklad:
We received feedback that near_vm_runner produces too much output couple
of times. At the same time, there was once instance where having a log
line for something as trivial `read_register` was useful.

So, let's downgrade the logging here to trace for now, and maybe revisit
in the future when we have a clearer understanding of which amount of
logs is acceptable.

### DIFF
--- a/runtime/near-vm-runner/src/imports.rs
+++ b/runtime/near-vm-runner/src/imports.rs
@@ -256,7 +256,7 @@ pub(crate) mod wasmer {
                     let _span = if IS_GAS {
                         None
                     } else {
-                        Some(tracing::debug_span!(target: "host-function", stringify!($func)).entered())
+                        Some(tracing::trace_span!(target: "host-function", stringify!($func)).entered())
                     };
                     let logic: &mut VMLogic<'_> = unsafe { &mut *(ctx.data as *mut VMLogic<'_>) };
                     logic.$func( $( $arg_name, )* )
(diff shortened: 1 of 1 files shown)


nagisa (commented):
The change made here would suggest to me that we are logging span entries and exits. Ability to do so to the best of my knowledge is really intended to be a debugging tool for people implementing subscribers or similar parts of the tracing ecosystem, and probably should not be left enabled in a production system as is.

The primary reason is indeed just how spammy it is (and formatting all those messages can't possibly be very cheap). Mitigating the amount of messages the way is done here, however, has a downside that the alternative subscribers, which natively understand spans, will no longer receive them, either. In my view the spans should generally remain at `info!` level so that tools like opentelemetry collector can actually take advantage of the data that's available to them.

If we absolutely must output some information about span durations to the standard output alongside other log messages, I would probably approach this from the angle of implementing a subscriber which consumes spans of the form:

```rust
info_span!(..., span_logger.warn_deadline = 10ms)
```

and outputs a single message _if_ the span did indeed exceed the specified duration.

Alternatively I would investigate our ability to utilize ecosystem tooling which natively understands tracing data such as `opentelemetry+jaeger` so that we don't need to try and coerce logs into a trace system.

nagisa (approved):
That said, I believe this discussion should be held more widely than just within the runtime team and so I'm fine with this landing as a temporary measure, but we should definitely keep in mind reverting this when we have some real tracing infrastructure in place.

PR #5913, opened 2021-12-20
Synchronize block and chunk production

EgorKulikov:
This is port of hotfix to address #5688 to master
While all chunk producers are block producers we should use same validator to produce chunks that would later produce block out of them, to avoid network delays

Originally implemented in #5901


core/primitives/src/version.rs
```diff
@@ -144,7 +147,7 @@ pub const PEER_MIN_ALLOWED_PROTOCOL_VERSION: ProtocolVersion = MAIN_NET_PROTOCOL
 /// Current protocol version used on the main net.
 /// Some features (e. g. FixStorageUsage) require that there is at least one epoch with exactly
 /// the corresponding version
-const MAIN_NET_PROTOCOL_VERSION: ProtocolVersion = 50;
+const MAIN_NET_PROTOCOL_VERSION: ProtocolVersion = 51;
```
matklad, line 150: :thinking: I am a bit surprised that we do version bump here. Was this indeed the case that we upgraded mainnet version? That is, that the next stable release should generally get protocol version 52?

matklad (approved):

core/primitives/src/version.rs
EgorKulikov: It should get version 51 and we already have stabilized feature, that's why I bumped it (for that feature)

bowenwang1996 (approved):
I think you might not be able to merge this until we publish 1.23.1 due to the check on protocol version.

EgorKulikov:
> I think you might not be able to merge this until we publish 1.23.1 due to the check on protocol version.

Yeah, that's why I had not solicited reviews yet

core/primitives/src/version.rs
mzhangmzz: I'm still confused why we need to bump the version. The protocol version of testnet is already 50 which included the hotfix. So it seems we are bumping protocol version for nothing?

PR #5915, opened 2021-12-20
chain: fix ‘remove unnecessary ShardId redefinition’ commit

mina86:
Commit 15c648c4c2: ‘remove unnecessary ShardId redefinition’
introduced `shard_pos` variable in `assign_with_possible_repeats`
function but failed to properly update it when necessary.  This lead
to incorrect vector index being used when accessing result vector
elements.  Fix that.

### DIFF
--- a/chain/epoch_manager/src/shard_assignment.rs
+++ b/chain/epoch_manager/src/shard_assignment.rs
@@ -116,13 +116,12 @@ fn assign_with_possible_repeats<T: HasStake + Eq, I: Iterator<Item = (usize, T)>
             .expect("cp_iter should contain enough elements to minimally fill each shard");
         let (least_validator_count, shard_stake, shard_id) =
             shard_index.pop().expect("shard_index should never be empty");
-        let shard_pos = usize::try_from(shard_id).unwrap();
 
         if assignment_index < num_chunk_producers {
             // no need to worry about duplicates yet; still on first pass through validators
             shard_index.push((least_validator_count + 1, shard_stake + cp.get_stake(), shard_id));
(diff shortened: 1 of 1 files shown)


chain/epoch_manager/src/shard_assignment.rs
```diff
@@ -136,15 +135,15 @@ fn assign_with_possible_repeats<T: HasStake + Eq, I: Iterator<Item = (usize, T)>
...
+                if result[usize::try_from(shard_id).unwrap()].contains(&cp) {
                     buffer.push((least_validator_count, shard_stake, shard_id))
                 } else {
                     shard_index.push((
                         least_validator_count + 1,
                         shard_stake + cp.get_stake(),
                         shard_id,
                     ));
-                    result[shard_pos].push(cp);
+                    result[usize::try_from(shard_id).unwrap()].push(cp);
```
mm-near, line 146: let's add a unittest for this function, that would have caught this issue.

mm-near (changes requested):

chain/epoch_manager/src/shard_assignment.rs
mina86: Could I get an approval for this PR so we get the test failure fixed. We can worry about unit tests later.

mina86: There actually is unrelated bug in `assign_with_possible_repeats` and it’ll take me more time to get another fix and unit tests so let’s get this merged. There already is NayDuck test that catches this failure and the test is currently failing.

mm-near (approved):

PR #5916, opened 2021-12-20
Count the number of chunks included in processed blocks

nikurt:


chain/client/src/client_actor.rs
```diff
@@ -980,11 +980,18 @@ impl ClientActor {
                 accepted_block.provenance,
             );
             let block = self.client.chain.get_block(&accepted_block.hash).unwrap();
+            let mut chunks_in_block = 0;
+            for chunk in block.chunks().iter() {
```
EgorKulikov, line 984: We have bool slice of chunks included in block.header.chunk_mask()

chain/client/src/client_actor.rs
```diff
@@ -980,11 +980,18 @@ impl ClientActor {
...
+                if chunk.height_included() == block.header().height() {
+                    chunks_in_block += 1;
+                }
+            }
+            // let chunks_in_block = block.chunks().len();
```
bowenwang1996, line 989: should we remove this line?

chain/client/src/metrics.rs
```diff
@@ -39,6 +39,16 @@ pub static SENT_BYTES_PER_SECOND: Lazy<IntGauge> = Lazy::new(|| {
 pub static BLOCKS_PER_MINUTE: Lazy<IntGauge> = Lazy::new(|| {
     try_create_int_gauge("near_blocks_per_minute", "Blocks produced per minute").unwrap()
 });
+pub static CHUNKS_PER_MINUTE: Lazy<IntGauge> = Lazy::new(|| {
```
bowenwang1996, line 42: why do we want to have this metric (per minute specifically)?

chain/client/src/info.rs
```diff
@@ -103,6 +107,11 @@ impl InfoHelper {
         let avg_bls = (self.num_blocks_processed as f64)
             / (self.started.elapsed().as_millis() as f64)
             * 1000.0;
+        let chunks_per_block = if self.num_blocks_processed > 0 {
+            (self.num_chunks_in_blocks_processed as f64) / (self.num_blocks_processed as f64)
+        } else {
+            0.
```
mzhangmzz, line 113: How often does the metric get recorded? Is it possible that between the two data points of the metrics, no block is processed, and this metric is set to 0, which will make the metric artificially low.

Is there a way to output something like nan/null/none in metrics so the data point will be ignored?

nikurt: Once every 10 seconds.
This is possible, but seems like 0 blocks per minute is a bigger problem than 0 chunks per block.
The value needs to be an `i64`, and generally dealing with nan/null/none values in dashboards is problematic.

bowenwang1996 (approved):

chain/client/src/info.rs
mzhangmzz: If it's 10 seconds, then it could be zero if no block is processed in the 10 seconds window.

I agree that it's problematic if no block is processed in 10 seconds. However, if that does happen, the value for this metric will also drop to zero, which is not ideal because it will cause confusion for people who are reading the metric. Basically people will need to make a mental note that, if the metric is zero, it could be one of the two cases, either block processing is slow, or the blocks have no chunks. If we have more metrics like this, it will be very hard to keep track of. 

I would suggest to cache the previous non-zero value, and return that value. We save the last value of `(self.num_chunks_in_blocks_processed as f64) / (self.num_blocks_processed as f64)` when `self.num_blocks_processed` is not zero, and return this value if it is.

PR #5929, opened 2021-12-21
Chunks and blocks statistics: produced/expected by validators

nikurt:
Tested by running a few epochs on localnet, and adding and removing some of the validators.
The kicked validators remain in the metrics, but their metric values are set to 0 until the end of the `neard` process.


matklad (approved):

nikurt:
Thanks @matklad !

mzhangmzz (approved):

Issue #5932, opened 2021-12-21
`assign_shards` is broken when trying to assign last validator to shard it was already assigned to


mina86:
Trying to assign validators with balances `[5, 4, 3]` to two shards with minimum number of validators per shard being two results in a panic.

PR #5933, opened 2021-12-21
chain: demonstrate failure in `assign_shards` function

mina86:
Introduce a test case which demonstrates a failure in the code.
Upcoming changes will fix the issue.


mina86:
This is going to go on top of https://github.com/near/nearcore/pull/5915

mm-near (approved):

Issue #4422, opened 2021-06-28
Node cannot download any blocks after restart


danielimkk:
**Describe the bug**
After a restart of the node, no more blocks are being downloaded. So it is stuck at block 41175921

**To Reproduce**
Start a node, using a snapshot. And at some moment of it running, restart it.

**Expected behavior**
I'd expect a node to continue synchronization after a restart

**Screenshots**
![image](https://user-images.githubusercontent.com/45372041/123588005-efd53f00-d808-11eb-857a-de2c504448e1.png)

**Version (please complete the following information):**
- nearcore commit/branch 1.19.1
- rust version (if local) rustc 1.49.0-nightly
- mainnet

**Additional context**
I have built the neard binary from sources. This issue has happened the second time, last time I had to remove all current database and start the node using a snapshot provided in documentation. Do you have any ideas on how to prevent this issue happening and what could have caused it? Thank you!

A bit more information about setup:
Process manager: supervisor
Binary: neard, built from sources
Version of neard: 1.19.1

bowenwang1996:
Did you ever change your config.json?

danielimkk:
> Did you ever change your config.json?

Nope, it was not changed before or after the restart and update
Here is the config file contents:

```
{
    "genesis_file": "genesis.json",
    "genesis_records_file": null,
    "validator_key_file": "validator_key.json",
    "node_key_file": "node_key.json",
    "rpc": {
        "addr": "0.0.0.0:3030",
        "cors_allowed_origins": [
            "*"
        ],
        "polling_config": {
            "polling_interval": {
                "secs": 0,
                "nanos": 500000000
            },
            "polling_timeout": {
                "secs": 10,
                "nanos": 0
            }
        },
        "limits_config": {
            "json_payload_max_size": 10485760
        }
    },
    "telemetry": {
        "endpoints": [
            "https://explorer.mainnet.near.org/api/nodes"
        ]
    },
    "network": {
        "addr": "0.0.0.0:24567",
        "external_address": "",
[...]

PR #5934, opened 2021-12-21
logs: add a bit more info in Block sync debug logs

marcelo-gonzalez:
This makes two changes to try and provide a bit more debugging
info. One is that we show which peer was chosen to requset a block
from in the block sync debug message. The other is that we also warn
if there were no suitable peers, since otherwise the logs could
possibly be misleading if for example we're connected to 3 peers, none
of which are archival, we're trying to sync from a chain that's > 5
epochs behind and we see a debug message saying:

sync: Block sync: 12345/123456 requesting block ABCDE from 3 peers

when really we're not about to send any request at all.

Issue: https://github.com/near/nearcore/issues/4422


chain/client/src/sync.rs
```diff
@@ -557,12 +555,17 @@ impl BlockSync {
...
+            debug!(target: "sync", "Block sync: {}/{} requesting block {} from {} (out of {} peers)",
+		   head.height, header_head.height, next_hash, peer.peer_info.id, highest_height_peers.len());
             self.network_adapter.do_send(PeerManagerMessageRequest::NetworkRequests(
                 NetworkRequests::BlockRequest {
                     hash: request.hash,
                     peer_id: peer.peer_info.id.clone(),
                 },
             ));
+        } else {
+            warn!(target: "sync", "Block sync: {}/{} No available {}peers to request block {} from",
```
mm-near, line 567: (in a separate PR) - we should consider having a metric with this (exported to grafana) - to see often this happens.

mm-near (approved):

chain/client/src/sync.rs
marcelo-gonzalez: Yeah would be good to have stats on that, agreed

Issue #5900, opened 2021-12-18
Validator RPC not responding in a timely manner after update to 1.23.0


JiuKelo:
Hey guys,

After updating to 1.23.0 our RPC requests stoped working. During debug we've discovered that it answers but with *HUGE* delay
At epoch change yesterday it stated working instantly again, today (after epoch change) it began lagging again.

An example here:

```
time curl -s -X POST -H "Content-Type: application/json" -d '{"jsonrpc":"2.0","id":"1","method":"validators","params":[null]}' 127.0.0.1:3030

.....

real 7m19.798s
user 0m0.017s
sys 0m0.000s
```

Using nearcore v1.23.0 with rustc 1.57.0 (f1edd0429 2021-11-29) on mainnet

bowenwang1996:
Looks like this is caused by https://github.com/near/nearcore/pull/5125 where we changed to use a separate runtime for view client. The issue is that it initializes a separate epoch manager and https://github.com/near/nearcore/blob/acaa632eb6a4ac80dba344e715c2fda30aa3cf43/chain/epoch_manager/src/lib.rs#L965 relies on https://github.com/near/nearcore/blob/acaa632eb6a4ac80dba344e715c2fda30aa3cf43/chain/epoch_manager/src/lib.rs#L1400, which has the (undocumented) assumption that the aggregator is shared between client and view client so that view client could easily extract the current epoch information. With a separate epoch manager, however, this is no longer the case and it means view client potentially has to traverse up to 43200 blocks to compute this information, which in turn significantly degrades the rpc performance towards the end of an epoch.

mzhangmzz:
Yes, sorry for the bug, I will just revert the original commit

bowenwang1996:
Fixed in #5935

kucharskim:
When this fix is scheduled to be released in mainnet?

bowenwang1996:
> When this fix is scheduled to be released in mainnet?

On Feb 7

PR #5935, opened 2021-12-21
(fix) Revert "make view client use a different runtime object to avoid cont…

mzhangmzz:
…ention (#5125)"

This reverts commit e7433e6844c46026f57a7aaaa2511f8c271c5ba2.

To fix issue https://github.com/near/nearcore/issues/5900


bowenwang1996 (approved):
Do we still plan to have a separate `NightshadeRuntime` for view client?

mzhangmzz:
> Do we still plan to have a separate `NightshadeRuntime` for view client?

I incline to no, until we refactor `EpochManager`, because of problems like this might arise (I can fix all the existing issues like this but people might change the code in the future), and there is no way in the code to prevent that from happening.

PR #5937, opened 2021-12-21
chain: simplify `assign_with_possible_repeats` function

mina86:
The `assignment_index < num_chunk_producers` special case present in
the function can be handled by the general case.  It’s not really
necessary and doesn’t really speed things up since on first pass all
vectors in `result` will be empty thus `contains` calls will all
immediately return `false`.  Removing that special case simplifies the
function making `num_chunk_producers` argument and enumerating chunk
producers  unnecessary.

While at it, remove the `num_shards` argument since it can be inferred
from `shard_index.len()` and thus doesn’t need to be passed directly.


bowenwang1996 (commented):
@EgorKulikov @mzhangmzz please review

mina86:
@mzhangmzz, FYI, this still doesn’t fix the issue (just a refactoring) but commit after this will finally be a fix.

pmnoxx (approved):
LGTM

chain/epoch_manager/src/shard_assignment.rs
```diff
@@ -30,13 +30,13 @@ pub fn assign_shards<T: HasStake + Eq + Clone>(
...
 
     let mut result: Vec<Vec<T>> = (0..num_shards).map(|_| Vec::new()).collect();
+    let mut cp_iter = chunk_producers.into_iter();
 
     // Place validators into shards while there are still some without the
     // minimum required number.
     if required_validator_count == num_chunk_producers {
         // In this case there are at least enough chunk producers that no one will need
         // to be assigned to multiple shards, so we don't need to worry about that case.
-        let mut cp_iter = chunk_producers.into_iter();
```
mzhangmzz, line 39: The moving the this line seems to be unnecessary?

mzhangmzz (approved):

bowenwang1996 (approved):

EgorKulikov (approved):
Thanks, this really makes this code simpler

PR #5938, opened 2021-12-22
re: Remove `unwrap_option_or_return` macro

pmnoxx:
There is a macro, which we are using exactly ones. It's better to inline it.


pmnoxx:
```
@bowenwang1996 @mm-near: /chain/network/ /core/
```

matklad (approved):

mm-near (approved):

PR #5939, opened 2021-12-22
re: Remove pub use of external `PeerInfo`

pmnoxx:
`PeerInfo` is stored right now in `near-network-primitives`.
It should not be exported by `near-network`.


pmnoxx:
```
@bowenwang1996 @mm-near: /chain/
```

mm-near (approved):

PR #5941, opened 2021-12-22
chain: refactor `shard_assignments` module testing code

mina86:
Introduce `tests::assign_shards` wrapper which takes care of calling
the actual `assign_shards` function and performs basic validation on
the result.  Furthermore, make it return aggregated data which is
easier to assert on.

In that new function check whether all chunk producers have been
assigned to at least one shard and whether there isn’t imbalance in
how many shards chunk producers are assigned to.

With all that, adjust test cases to use the new function (which
usually makes them shorter) and make `test_distribution_common` a bit
more verbose when reporting failures.


mina86:
@mm-near, @mzhangmzz, friendly ping

mzhangmzz (approved):
LGTM!

mzhangmzz:
One question though, this PR doesn't fix https://github.com/near/nearcore/issues/5932, right?

mina86:
> One question though, this PR doesn't fix #5932, right?

Correct. I have subsequent changes that will fix it.

PR #5942, opened 2021-12-22
re:  `PeerManagerActor` use `checked_feature` correctly

pmnoxx:
Currently `PeerManagerActor` uses `checked_feature!` in a weird way.
It turns out that `checked_feature` can take two code path as argument, one if the condition is met, and other one other wise.
By using the other syntax, the code will be much easier to read.


pmnoxx:
```
```

matklad (approved):

mm-near (approved):

PR #5944, opened 2021-12-22
re: Move `Edge` related types to `near-network-primitives`

pmnoxx:
Move `chain/network/src/routing/network_protocol.rs → `/chain/network-primitives/src/network_protocol/edge.rs`

Currently `chain/network/src/routing/network_protocol.rs` contains types related to network protocol - `Edges`.
This file doesn't depend on `near-network` at all, it makes sense to move it outside of `near-network`.

As a side effect:
 - `jsonrpc-adversarial-primitives` will no longer depend on `near-network` only on `-near-network-primitives`.

The eventual goal is to split `network protocol` to a separate create.
The intermediate goal is to move those types outside of `near-network`. That will decrease the number of dependencies to `near-network`, and improve compilation time as a side effect.

In the next PR:
- Move out `chain/network/src/network_protocol.rs` outside of `near-network`.


pmnoxx:
```
@mm-near: /chain/ /chain/network-primitives/
```
@mm-near Can you take a look?

pmnoxx:
```
```
@mm-near Can you take a look?

mm-near (approved):

PR #5945, opened 2021-12-22
re: Derive `MessageResponse` instead of implementing it

pmnoxx:
We can refactor code related to `MessageResponse`. We don't need to implement the trait, we can simply derive it.

### DIFF
--- a/chain/network-primitives/src/types.rs
+++ b/chain/network-primitives/src/types.rs
@@ -8,8 +8,7 @@
 ///
 /// NOTE:
 /// - We also export publicly types from `crate::network_protocol`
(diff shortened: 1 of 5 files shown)


pmnoxx:
```
@mm-near: /chain/network-primitives/ /chain/network/ /utils
```
@mm-near Can you take a look?

pmnoxx:
```
@mm-near: chain/network/ chain/network-primitives/ utils
```
@mm-near Can you take a look?

matklad (approved):

mm-near (approved):

PR #5946, opened 2021-12-22
re: Simplify `AnnounceAccount`

pmnoxx:
`AnnounceAccount` code can be inlined, and it will be much easier to read what's going on.

### DIFF
--- a/chain/client/src/client_actor.rs
+++ b/chain/client/src/client_actor.rs
@@ -27,7 +27,6 @@ use near_chain::{
 use near_chain_configs::ClientConfig;
 #[cfg(feature = "test_features")]
 use near_chain_configs::GenesisConfig;
-use near_crypto::Signature;
 use near_network::types::{
     NetworkClientMessages, NetworkClientResponses, NetworkInfo, NetworkRequests,
     PeerManagerAdapter, PeerManagerMessageRequest,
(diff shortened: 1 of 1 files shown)


pmnoxx:
```
@bowenwang1996 @mm-near: /chain/ /core/
```

chain/client/src/client_actor.rs
```diff
@@ -766,14 +754,17 @@ impl ClientActor {
...
             self.last_validator_announce_time = Some(now);
-            let signature = self.sign_announce_account(&next_epoch_id).unwrap();
 
             self.network_adapter.do_send(PeerManagerMessageRequest::NetworkRequests(
                 NetworkRequests::AnnounceAccount(AnnounceAccount {
+                    signature: validator_signer.sign_account_announce(
+                        validator_signer.validator_id(),
+                        &self.node_id,
+                        &next_epoch_id,
+                    ),
```
matklad, line 764: I'd write this as

```
let signature = validator_signer.sign_account_announce(
                        validator_signer.validator_id(),
                        &self.node_id,
                        &next_epoch_id,
                    );
```

matklad (approved):
Also :+1: for the general pattern of "handle only relevant errors". That's a relatively common problem with Rust-style strongly-typed error handling -- types force you to handle some error condition, while in reality this condition isn't really possible.

mm-near (approved):

PR #5948, opened 2021-12-22
Fix clippy warnings in runtime

pmnoxx:
We can fix following clippy warnings in `near-runtime`

- clippy::redundant_closure 
- clippy::unnecessary_lazy_evaluations 
- clippy::useless_conversion 
- clippy::map_clone 
- clippy::len_zero 
- clippy::redundant_field_names

### DIFF
--- a/integration-tests/src/tests/network/runner.rs
+++ b/integration-tests/src/tests/network/runner.rs
@@ -670,8 +670,7 @@ impl Runner {
             .collect();
         let ports: Vec<_> = (0..self.num_nodes).map(|_| open_port()).collect();
 
-        let validators: Vec<_> =
-            accounts_id.iter().map(|x| x.clone()).take(self.num_validators).collect();
+        let validators: Vec<_> = accounts_id.iter().cloned().take(self.num_validators).collect();
 
         let mut peers_info =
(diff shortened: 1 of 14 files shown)


pmnoxx:
```
@matklad: /core/ /runtime/near-vm-runner/ /runtime/runtime-params-estimator/ /test-utils/runtime-tester
```
@matklad Can you take a look?

integration-tests/src/tests/network/runner.rs
```diff
@@ -670,8 +670,7 @@ impl Runner {
             .collect();
         let ports: Vec<_> = (0..self.num_nodes).map(|_| open_port()).collect();
 
-        let validators: Vec<_> =
-            accounts_id.iter().map(|x| x.clone()).take(self.num_validators).collect();
+        let validators: Vec<_> = accounts_id.iter().cloned().take(self.num_validators).collect();
```
matklad, line 673: Microoptimizatin: 

```suggestion
        let validators: Vec<_> = accounts_id.iter().take(self.num_validators).cloned().collect();
```

matklad (approved):

nagisa:
If we intend to keep clippy warnings low/nonexistent, there should definitely be a CI job that runs clippy with these lints denied.

PR #5954, opened 2021-12-22
re: `PeerActor` Simplify logic for receiving messages

pmnoxx:
It turns out that one of the arguments to `match` isn't even used. We can write a logic in a simpler way.


pmnoxx:
```
```

pmnoxx:
```
```

mm-near (approved):

PR #5955, opened 2021-12-23
re: Implement `ProjectedHashMap`

pmnoxx:
We talked before about replacing `HashMap` with `HashSet` in https://github.com/near/nearcore/pull/5475

I think it would be better to have a generic implementation, that doesn't depend on `PeerId` or `Edge`. The code will be much cleaner that way.

### DIFF
--- a/chain/network/src/routing/projected_hash_map.rs
+++ b/chain/network/src/routing/projected_hash_map.rs
@@ -3,12 +3,19 @@ use std::collections::HashSet;
 use std::hash::{Hash, Hasher};
 use std::marker::PhantomData;
 
+pub trait ProjectedMapKey<Key>
+where
+    Key: Hash,
+{
+    fn projected_map_key(&self) -> &Key;
+}
+
 /// Wraps around `Edge` struct. The main feature of this struct, is that it's hashed by
 /// `(Edge::key.0, Edge::key.1)` pair instead of `(Edge::key.0, Edge::key.1, Edge::nonce)`
 /// triple.
 struct HashMapHelper<T, V>
 where
-    V: Borrow<T>,
+    V: ProjectedMapKey<T>,
     T: Hash + PartialEq + Eq,
 {
     inner: V,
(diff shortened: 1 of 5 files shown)


pmnoxx:
```
```

chain/network/src/routing/projected_hash_map.rs
```diff
@@ -0,0 +1,147 @@
...
+#[derive(std::cmp::Eq)]
+pub struct HashMapHelper<T, V>
+where
+    V: Borrow<T>,
+    T: Hash + PartialEq + Eq,
+{
+    inner: V,
+    pd: PhantomData<T>,
+}
+
```
matklad, line 18: I'd keep this private with 

```diff
diff --git a/chain/network/src/routing/projected_hash_map.rs b/chain/network/src/routing/projected_hash_map.rs
index cb4ee6a8c..ba128ac43 100644
--- a/chain/network/src/routing/projected_hash_map.rs
+++ b/chain/network/src/routing/projected_hash_map.rs
@@ -6,8 +6,7 @@ use std::marker::PhantomData;
 /// Wraps around `Edge` struct. The main feature of this struct, is that it's hashed by
 /// `(Edge::key.0, Edge::key.1)` pair instead of `(Edge::key.0, Edge::key.1, Edge::nonce)`
 /// triple.
-#[derive(std::cmp::Eq)]
-pub struct HashMapHelper<T, V>
+struct HashMapHelper<T, V>
 where
     V: Borrow<T>,
     T: Hash + PartialEq + Eq,
@@ -36,6 +35,13 @@ where
     }
 }
 
+impl<T, V> Eq for HashMapHelper<T, V>
+where
+    V: Borrow<T>,
+    T: Hash + PartialEq + Eq,
+{
+}
+
 impl<T, V> Hash for HashMapHelper<T, V>
 where
     V: Borrow<T>,
@@ -69,7 +75,6 @@ impl<T, V> ProjectedHashMap<T, V>
 where
     V: Borrow<T>,
     T: Hash + PartialEq + Eq,
-    HashMapHelper<T, V>: Eq,
 {
     /// Note: We need to remove the value first.
     /// The insert inside HashSet, will not remove existing element if it has the same key.
```

As a rule of thumb, making something called `Helper` a public type signals that something's iffy going on.

chain/network/src/routing/routing_table_actor.rs
```diff
@@ -326,8 +329,10 @@ impl RoutingTableActor {
         let edges_to_remove = self
             .edges_info
             .iter()
-            .filter_map(|(key, edge)| {
-                if peers_to_remove.contains(&key.0) || peers_to_remove.contains(&key.1) {
+            .filter_map(|edge| {
```
matklad, line 332: unrelated tho the PR, but I'd write this as `.filter(||).cloned()` rather than as a `.filter_map`

chain/network/src/routing/routing_table_actor.rs
```diff
@@ -44,18 +45,20 @@ pub enum Prune {
...
     /// Current view of the network represented by undirected graph.
     /// Nodes are Peers and edges are active connections.
-    pub raw_graph: Graph,
+    pub(crate) raw_graph: Graph,
     /// Active PeerId that are part of the shortest path to each PeerId.
-    pub peer_forwarding: Arc<HashMap<PeerId, Vec<PeerId>>>,
+    pub(crate) peer_forwarding: Arc<HashMap<PeerId, Vec<PeerId>>>,
     /// Last time a peer was reachable through active edges.
-    pub peer_last_time_reachable: HashMap<PeerId, Instant>,
+    pub(crate) peer_last_time_reachable: HashMap<PeerId, Instant>,
```
matklad, line 61: :+1: love this `pub(crate)`'s, they make this immediately clear that the thing isn't accessed outside of this compilation unit.

matklad (approved):

chain/network/src/routing/routing_table_actor.rs
pmnoxx: @matklad This whole function should be rewritten. However, I'm waiting with refactor after this PR gets shipped.

mm-near (approved):

PR #5957, opened 2021-12-23
re: Use `Ordering::Relaxed` in `near-network`

pmnoxx:
Orderings other than `Relaxed` makes sense to use only if the atomic value is meant to be used a memory barrier.
For the purpose of `near-network` only `Ordering::Relaxed` makes sense.

There is no  performance difference on `x86_64`, but this is not done for performance, but code readability.


pmnoxx:
```
```

pmnoxx:
```
```

matklad (approved):
LGTM!

I don't think strongly between `SeqCst` or `Relaxed` (both "use SeqCst unless benchmarked" and "use the ordering you need to reason about correctness of code" schools of thought have merit).

I definitely think that replacing `AcqRel` with either of those is a good idea.

mm-near (approved):

PR #5960, opened 2021-12-23
refactor: use the lower level wasmer APIs

nagisa:
This is intended to be a non-functional change.

Here we replace the use of the `wasmer` crate and the wrappers it
provides with the lower level APIs exposed by various wasmer subcrates
such as `wasmer-engine`, `wasmer-vm`, etc.

These changes are primarily intended to reduce the burden of maintaining
and publishing the `wasmer-near` crate. Since our requirements for a
runtime are scoped to very narrow portion of what the `wasmer` crate
provides, the complexity of the code in `nearcore` didn't
change as much as I feared it would, while also making certain portions
of the codebase significantly more straightforward.

The `imports`-related code is perhaps the highest gain in complexity of
the code, as it has become necessary to take care of ensuring a matching
ABI. On the other hand, the changes there also reduced non-local
assumptions about the lifetime of the `VMLogic` by moving it this concern
entirely to the funcition which handles instantiations.

There are definitely more opportunities for simplifying the code around
this part of the codebase, but these will need to happen as a follow up
as this one is already getting quite large, and some of the changes
would touch other runners, not just Wasmer2.

### DIFF
--- a/runtime/near-vm-runner/src/imports.rs
+++ b/runtime/near-vm-runner/src/imports.rs
@@ -54,7 +54,7 @@ macro_rules! imports {
         $func:ident < [ $( $arg_name:ident : $arg_type:ident ),* ] -> [ $( $returns:ident ),* ] >,)*
     ) => {
         macro_rules! for_each_available_import {
-            ($protocol_version:ident, $M:ident) => {$(
+            ($protocol_version:expr, $M:ident) => {$(
                 $(#[cfg(feature = $feature_name)])*
                 if true
                     $(&& near_primitives::checked_feature!($feature_name, $feature, $protocol_version))*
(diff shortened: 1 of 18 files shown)


runtime/near-vm-runner/src/wasmer2_runner.rs
```diff
@@ -40,25 +40,36 @@ const WASMER_FEATURES: wasmer::Features = wasmer::Features {
...
+                &MemoryType::new(Pages(initial_memory_pages), Some(max_pages), false),
+                &MemoryStyle::Static { bound: max_pages, offset_guard_size: 0x1000 },
             )
-            .expect("TODO creating memory cannot fail"),
-        ))
+            .expect("creating memory must not fail"),
+        )))
+    }
+
+    fn view(&self) -> MemoryView<u8> {
```
matklad, line 58: Seems like we can just `-> &[u8]` here?

nagisa: `MemoryView` derefs to `&[Cell<u8>]` really, allowing mutation as necessary for e.g. `write_memory`.

matklad: Yeah, and it seems that this Deref is the only bit of MemoryView we use? So it seems that we can use `[Cell<u8>]` *directly* instead and avoid `MemoryView` abstraction altogether?

nagisa: Changed this area of code significantly (and added tests!)

runtime/near-vm-runner/src/imports.rs
```diff
@@ -274,55 +274,150 @@ pub(crate) mod wasmer {
...
+    macro_rules! return_ty {
+        () => {
+            type Ret = ();
+            fn make_ret() -> () {}
+        };
+        ($($return: ident),*) => {
+            #[repr(C)]
+            struct Ret($(<$return as Wasmer2Type>::Wasmer),*);
+            fn make_ret($($return: $return),*) -> Ret { Ret($($return.to_wasmer()),*) }
         }
```
matklad, line 324: No strong opinion here, but I'd:

* Pass the `Ret` ident in, pretending that macros are hygenic
* use `$returns` rather than `$return` to be consistent with `add_import!` and to avoid messing mental syntax highlighting up: `$return` looks like some kind of hygenic `return`

runtime/near-vm-runner/src/imports.rs
```diff
@@ -274,55 +274,150 @@ pub(crate) mod wasmer {
...
+                        let env = unsafe { &mut *env };
+                        const IS_GAS: bool = str_eq(stringify!($func), "gas");
+                        let _span = if IS_GAS {
+                            None
+                        } else {
+                            Some(tracing::trace_span!(
+                                target: "host-function",
+                                stringify!($func)
+                            ).entered())
+                        };
```
matklad, line 359: Let's move the tracing bit into the `catch_unwind` boundary? Just to limit the amount of code outside to the absolute minimum.

runtime/near-vm-runner/src/wasmer2_runner.rs
```diff
@@ -40,76 +40,103 @@ const WASMER_FEATURES: wasmer::Features = wasmer::Features {
...
-        ))
+            .expect("creating memory must not fail"),
+        )))
     }
 
-    pub fn clone(&self) -> Memory {
-        self.0.clone()
+    // Returns the pointer to memory at the specified offset and the size of the buffer starting at
+    // the returned pointer.
+    fn data_offset(&self, offset: u64) -> Option<(*mut u8, usize)> {
```
matklad, line 63: Naively, I'd expect this to return `Option<*mut [u8]>`. Am I correct that we don't do that here because `*mut [u8]` is impossible to construct without going via `&mut [u8]`, which we rather avoid?

runtime/near-vm-runner/src/wasmer2_runner.rs
```diff
@@ -40,76 +40,103 @@ const WASMER_FEATURES: wasmer::Features = wasmer::Features {
...
-            wasmer::InstantiationError::Start(e) => e.into_vm_error(),
-            wasmer::InstantiationError::HostEnvInitialization(_) => {
-                VMError::FunctionCallError(FunctionCallError::CompilationError(
-                    CompilationError::PrepareError(PrepareError::Instantiate),
-                ))
-            }
+            InstantiationError::Start(e) => e.into_vm_error(),
         }
     }
 }
```
matklad, line 137: Mostly unrelated to the PR, but consider inlining this impl into a single place where it is called.

runtime/near-vm-runner/src/wasmer2_runner.rs
```diff
@@ -40,76 +40,103 @@ const WASMER_FEATURES: wasmer::Features = wasmer::Features {
...
+            let memory = self.get_memory_buffer(offset, buffer.len());
+            // SAFETY: we verified indices into are valid and the pointer will always be valid as
+            // well. Our runtime is currently only executing Wasm code on a single thread, so data
+            // races aren't a concern here.
+            std::ptr::copy_nonoverlapping(buffer.as_ptr(), memory, buffer.len());
+        }
     }
 }
 
-impl IntoVMError for wasmer::InstantiationError {
```
matklad, line 93: No strong feelings, but I'd probably stick with `wasmer_engine::InstantiationError` here -- when translating errors, its useful to know at a glance which crate the error comes from.

runtime/near-vm-runner/src/wasmer2_runner.rs
```diff
@@ -301,62 +271,245 @@ pub(crate) fn wasmer2_vm_hash() -> u64 {
...
-    // We only support singlepass compiler at the moment.
-    assert_eq!(WASMER2_CONFIG.compiler, WasmerCompiler::Singlepass);
-    let compiler = Singlepass::new();
-    // We only support universal engine at the moment.
-    assert_eq!(WASMER2_CONFIG.engine, WasmerEngine::Universal);
-    let engine = wasmer::Universal::new(compiler).features(WASMER_FEATURES).engine();
-    Store::new(&engine)
+#[derive(Clone)]
+pub(crate) struct VMArtifact {
+    artifact: Arc<dyn Artifact>,
```
matklad, line 276: ditto here, I'd probably spell this as `wasmer_engine::Artifact`, just so it's obvious which one of the two artifacts is ours.

runtime/near-vm-runner/src/wasmer2_runner.rs
```diff
@@ -301,62 +271,245 @@ pub(crate) fn wasmer2_vm_hash() -> u64 {
...
+    _engine: UniversalEngine,
```
matklad, line 277: Add a comment explaining why we need this unused field.

runtime/near-vm-runner/src/wasmer2_runner.rs
```diff
@@ -301,62 +271,245 @@ pub(crate) fn wasmer2_vm_hash() -> u64 {
...
+                let _span = tracing::debug_span!(target: "vm", "run_method/instantiate").entered();
+                // SAFETY/FIXME: this transmute shouldn't be necessary, but we are limited by
+                // wasmer's API right now.
+                //
+                // An important caveat is that the `'static` lifetime here refers to the lifetime
+                // of `VMLogic` reference to which is retained by the `InstanceHandle` we create.
+                // However this `InstanceHandle` only lives during the execution of this body, so
+                // we can be sure that `VMLogic` remains live and valid at any time.
+                let import =
+                    std::mem::transmute::<_, &mut Wasmer2Imports<'static, 'static>>(&mut import);
```
matklad, line 362:
```suggestion
                    std::mem::transmute::<&mut Wasmer2Imports<_, _>, &mut Wasmer2Imports<'static, 'static>>(&mut import);
```

just to be extra-paranoid that we convert reference of the right type.

runtime/near-vm-runner/src/wasmer2_runner.rs
```diff
@@ -301,62 +271,245 @@ pub(crate) fn wasmer2_vm_hash() -> u64 {
...
+                        // by the virtue of it being contained within `import` which lives for the
+                        // entirety of this function.
+                        InstanceConfig::new_with_counter(gas),
+                    )
+                    .map_err(|err| translate_instantiation_error(err, import.vmlogic))?;
+                // SAFETY: being called immediately after instantiation.
+                artifact
+                    .artifact
+                    .finish_instantiation(self, &handle)
+                    .map_err(|err| translate_instantiation_error(err, import.vmlogic))?;
```
matklad, line 382: This begets a question -- do we actually have a test which checks what happens if gas is exhausted when `_start` is called? I don't recall one.

runtime/near-vm-runner/src/wasmer2_runner.rs
```diff
@@ -301,62 +271,245 @@ pub(crate) fn wasmer2_vm_hash() -> u64 {
...
-pub(crate) struct Wasmer2VM;
+unsafe impl wasmer_vm::TrapHandler for Wasmer2VM {
+    fn as_any(&self) -> &dyn std::any::Any {
+        self
+    }
+
+    fn custom_trap_handler(&self, _: &dyn Fn(&wasmer_vm::TrapHandlerFn) -> bool) -> bool {
+        false
+    }
+}
```
matklad, line 460: Feels like something we potentially don't need :)

matklad (commented):
LGTM!

https://github.com/near/nearcore/pull/5960/files#r778106409 is I think the only place which we need to fix

runtime/near-vm-runner/src/wasmer2_runner.rs
nagisa: Right, this trait is required due to trait bounds on `wasmer_call_trampoline`. I could split out `Wasmer2VM` into two types, but this seemed less onerous to me.

runtime/near-vm-runner/src/wasmer2_runner.rs
nagisa: I had initially made this a `&mut [u8]` but then realized this becomes a `&T -> &mut T` function which is a big red flag something is going wrong, so I went with `*mut u8` instead and left some load-bearing safety comments. Maybe in the future there's space to improve on this, but it seems… involved.

runtime/near-vm-runner/src/wasmer2_runner.rs
nagisa: We do have one here: https://github.com/near/nearcore/blob/2a033f42eeefea2fbc867e26ec4281132b6ae447/runtime/near-vm-runner/src/tests/runtime_errors.rs#L32-L44

nagisa:
Are there any remaining fundamental blockers (besides the merge conflict?)

matklad (approved):

PR #5964, opened 2021-12-23
refactor: Generalize parameter estimator 

jakmeier:
- Change function interfaces to use Gas as a common unit
- Integrate CPU benchmark used to estimate x86 op cost
- Integrate gas metering estimation
- Preparation for more general estimation cases to be added
- Remove unused code

### DIFF
--- a/runtime/runtime-params-estimator/src/config.rs
+++ b/runtime/runtime-params-estimator/src/config.rs
@@ -1,16 +1,5 @@
-use near_primitives::types::AccountId;
 use near_vm_runner::internal::VMKind;
 use std::path::PathBuf;
-
-/// Get account id from its index.
-pub fn get_account_id(account_index: usize) -> AccountId {
-    AccountId::try_from(format!("near_{}_{}", account_index, account_index)).unwrap()
-}
-
-/// Total number of transactions that we need to prepare.
-pub fn total_transactions(config: &Config) -> usize {
-    config.block_sizes.iter().sum::<usize>() * config.iter_per_block
-}
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 pub enum GasMetric {
@@ -37,6 +26,6 @@ pub struct Config {
     pub metric: GasMetric,
     /// VMKind used
     pub vm_kind: VMKind,
-    /// When non-none, only the specified metrics will be measured.
-    pub metrics_to_measure: Option<Vec<String>>,
+    /// When non-none, only the specified costs will be measured.
+    pub costs_to_measure: Option<Vec<String>>,
 }
--- a/runtime/runtime-params-estimator/src/cost.rs
+++ b/runtime/runtime-params-estimator/src/cost.rs
@@ -87,6 +87,8 @@ pub enum Cost {
     AltBn128G1SumByte,
 
     // Costs used only in estimator
+    ContractCompileBase, // TODO: Needs estimation function
+    ContractCompileBytes, // TODO: Needs estimation function
     GasMeteringBase,
     GasMeteringOp,
     IoReadByte,
--- a/runtime/runtime-params-estimator/src/cost.rs
+++ b/runtime/runtime-params-estimator/src/cost.rs
@@ -26,10 +26,6 @@ pub enum Cost {
     ActionDeleteKey,
     ActionDeleteAccount,
 
-    // TODO: Remove these two, compilation is a part of `ActionDeploy`.
-    ContractCompileBase,
-    ContractCompileBytes,
-
     HostFunctionCall,
     WasmInstruction,
     ReadMemoryBase,
@@ -89,6 +85,15 @@ pub enum Cost {
     AltBn128PairingCheckByte,
     AltBn128G1SumBase,
     AltBn128G1SumByte,
+
+    // Costs used only in estimator
+    GasMeteringBase,
+    GasMeteringOp,
+    IoReadByte,
+    IoWriteByte,
+    CpuBenchmarkSha256,
+    OneCPUInstruction,
+    OneNanosecond,
 
     __Count,
 }
--- a/runtime/runtime-params-estimator/src/cost.rs
+++ b/runtime/runtime-params-estimator/src/cost.rs
@@ -87,7 +87,7 @@ pub enum Cost {
     AltBn128G1SumByte,
 
     // Costs used only in estimator
-    ContractCompileBase, // TODO: Needs estimation function
+    ContractCompileBase,  // TODO: Needs estimation function
     ContractCompileBytes, // TODO: Needs estimation function
     GasMeteringBase,
     GasMeteringOp,
--- a/runtime/runtime-params-estimator/src/estimator_params.rs
+++ b/runtime/runtime-params-estimator/src/estimator_params.rs
@@ -1,0 +1,55 @@
+//! Some parameters are use within the estimator to transform measurements to gas costs.
+//! These parameters have been estimated manually and are now hard-coded for a more deterministic estimation of runtime parameters.
+//! This module contains the hard-coded constants as well as the code to manually re-estimate them.
+
+use num_rational::Ratio;
+
+use crate::{config::GasMetric, gas_cost::GasCost};
+
+// All constant below are measured in Gas, respectively, in fractions thereof.
+
+/// How much gas there is in a nanosecond worth of computation.
+pub(crate) const ONE_NANOSECOND: Ratio<u64> = Ratio::new_raw(1_000_000, 1);
+// We use factor of 8 to approximately match the price of SHA256 operation between
+// time-based and icount-based metric as measured on 3.2Ghz Core i5.
+pub(crate) const ONE_CPU_INSTRUCTION: Ratio<u64> = Ratio::new_raw(1_000_000, 8);
+
+// See runtime/runtime-params-estimator/emu-cost/README.md for the motivation of constant values.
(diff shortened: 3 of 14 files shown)


jakmeier:
This PR is mostly to give us a base to discuss the direction that I am heading with my refactoring.
This is what I was able to extract as a stable change from a larger set of changes that I have been playing around with.

I guess the main point to discuss here is: How elaborate should the parameter estimator binary be? Should we extend it so that it can run `ICount` and `Time` measurements all in one go? (This is probably necessary to have a meaningful estimation for `Cost::OneCPUInstruction`) Or should we keep it simple and do these comparisons only on the generated cost tables?

If we want to compare different measurements, it might make sense to have multiple more specific calls into Docker+Qemu, instead of one that forwards everything as we have it now. This is part of the motivation for `QemuCommandBuilder`. (The other is that I wanted more debug options.) I could go further and also generalize the docker command.

runtime/runtime-params-estimator/src/gas_metering.rs
```diff
@@ -237,9 +238,16 @@ pub fn compute_gas_metering_cost(
         );
         assert!(result.1.is_none());
     }
-    let total_raw_no_gas = start.elapsed().scalar_cost().to_i128().unwrap();
+    let total_raw_no_gas = start.elapsed().to_gas();
 
-    // println!("with gas: {}; no gas {}", total_raw_with_gas, total_raw_no_gas);
+    // TODO: This seems to fail almost always but has some non-determinism to it.
+    assert!(
```
jakmeier, line 244: This is what I am looking at right now with higher priority than the general refactoring, as it seems related to RocksDB compaction. Here. it looks like the IO costs are sometimes attributed to one and sometimes to the other execution. It's always 64kiB for one and 0B for the other. The order in which the calls with/without gas metering happen does not affect it that much. If at least one of the test iterations has the IO cost attributed to the case without gas-metering, then this check fails.

This test also fails in the CI.

runtime/runtime-params-estimator/src/cost.rs
```diff
@@ -26,10 +26,6 @@ pub enum Cost {
     ActionDeleteKey,
     ActionDeleteAccount,
 
-    // TODO: Remove these two, compilation is a part of `ActionDeploy`.
-    ContractCompileBase,
-    ContractCompileBytes,
```
matklad, line 31: I'd actually maybe kept these, as a "helper cost" to answer the question of "how much of deploy cost is compilation, and how much is everything else" . So, move it to the block below, and actually add the relevan functiosn to `lib.rs`

runtime/runtime-params-estimator/src/estimator_params.rs
```diff
@@ -0,0 +1,55 @@
...
+//! This module contains the hard-coded constants as well as the code to manually re-estimate them.
+
+use num_rational::Ratio;
+
+use crate::{config::GasMetric, gas_cost::GasCost};
+
+// All constant below are measured in Gas, respectively, in fractions thereof.
+
+/// How much gas there is in a nanosecond worth of computation.
+pub(crate) const ONE_NANOSECOND: Ratio<u64> = Ratio::new_raw(1_000_000, 1);
```
matklad, line 12:
```suggestion
pub(crate) const ONE_NANOSECOND: Ratio<Gas> = Ratio::new_raw(1_000_000, 1);
```

runtime/runtime-params-estimator/src/estimator_params.rs
```diff
@@ -0,0 +1,55 @@
...
+// We use factor of 8 to approximately match the price of SHA256 operation between
+// time-based and icount-based metric as measured on 3.2Ghz Core i5.
+pub(crate) const ONE_CPU_INSTRUCTION: Ratio<u64> = Ratio::new_raw(1_000_000, 8);
```
matklad, line 15:
```suggestion
pub(crate) const ONE_CPU_INSTRUCTION: Ratio<Gas> = Ratio::new_raw(1_000_000, 8);
```

runtime/runtime-params-estimator/src/gas_cost.rs
```diff
@@ -65,8 +67,8 @@ impl GasCost {
     pub(crate) fn from_raw(raw: Ratio<u64>, metric: GasMetric) -> GasCost {
         let mut result = GasCost::zero(metric);
         match metric {
-            GasMetric::ICount => result.instructions = raw,
-            GasMetric::Time => result.time_ns = raw,
+            GasMetric::ICount => result.instructions = raw / ONE_CPU_INSTRUCTION,
+            GasMetric::Time => result.time_ns = raw / ONE_NANOSECOND,
```
matklad, line 71: Maybe name these `GAS_IN_NS` and `GAS_IN_INSTR`?

runtime/runtime-params-estimator/src/gas_metering.rs
```diff
@@ -25,15 +23,16 @@ fn test_gas_metering_cost(metric: GasMetric) {
         if true {
             // Here we test gas metering costs for forward branch cases.
             let nested_contract = make_deeply_nested_blocks_contact(depth);
-            let cost =
-                compute_gas_metering_cost(metric, VMKind::Wasmer0, REPEATS, &nested_contract);
+            let cost = compute_gas_metering_cost(metric, vm_kind, REPEATS, &nested_contract);
+            #[cfg(test)]
             println!("nested {} {}", depth, cost / (REPEATS as u64));
```
matklad, line 28: I'd just remove prints

runtime/runtime-params-estimator/src/gas_cost.rs
```diff
@@ -65,8 +67,8 @@ impl GasCost {
     pub(crate) fn from_raw(raw: Ratio<u64>, metric: GasMetric) -> GasCost {
```
matklad, line 67:
```suggestion
    pub(crate) fn from_raw(raw: Ratio<Gas>, metric: GasMetric) -> GasCost {
```

runtime/runtime-params-estimator/src/config.rs
```diff
@@ -37,6 +26,6 @@ pub struct Config {
     pub metric: GasMetric,
     /// VMKind used
     pub vm_kind: VMKind,
-    /// When non-none, only the specified metrics will be measured.
-    pub metrics_to_measure: Option<Vec<String>>,
+    /// When non-none, only the specified costs will be measured.
+    pub costs_to_measure: Option<Vec<String>>,
```
matklad, line 30: Not sure if i mentioned this alraedy, but it makes sense to rename the cli flag to jsut `--costs` as well.

runtime/runtime-params-estimator/src/gas_metering.rs
```diff
@@ -42,29 +41,27 @@ fn test_gas_metering_cost(metric: GasMetric) {
...
+    #[cfg(test)]
+    println!("backward branches: {} gas base {} gas per op", cost2_base, cost2_op,);
+
+    let cost_base = std::cmp::max(cost1_base, cost2_base).round().to_integer() as u64;
+    let cost_op = std::cmp::max(cost1_op, cost2_op).round().to_integer() as u64;
+    (cost_base, cost_op)
 }
 
 #[test]
 fn test_gas_metering_cost_time() {
```
matklad, line 61: we moved this to `lib.rs`, so the tests can be safely removed.

runtime/runtime-params-estimator/src/gas_metering.rs
matklad: Uhu, let's than maybe comment this out and land just the refactor?

matklad (approved):
LGTM as is, but let's maybe fix the following in this PR:

* drop prints altogether, rather than cfg(test) them
* drop `#[test]` for stuff we moved over to `lib.rs`
* rename constants like `ONE_NANOSECOND` to be more clear

Adding actual separate costs for compilation time and gas can be folow-ups!

runtime/runtime-params-estimator/src/qemu.rs
```diff
@@ -39,3 +41,72 @@ fn hypercall(index: u32) -> u64 {
...
+        self.started = started;
+        self
+    }
+    /// Print the counters on every close() syscall
+    pub fn print_on_every_close(mut self, on_every_close: bool) -> Self {
+        self.on_every_close = on_every_close;
+        self
+    }
+    /// Instantiate different counters for each thread
+    pub fn count_per_thread(mut self, count_per_thread: bool) -> Self {
```
matklad, line 65: I like the following naming convention for such setters (seen in some of BurntSushi's code)

```suggestion
    pub fn count_per_thread(mut self, yes: bool) -> Self {
```

runtime/runtime-params-estimator/src/gas_metering.rs
jakmeier: Only commenting the assertion out won't help, the subtraction below will panic with an underflow. But I can add `#[ignore]` to the affected tests

jakmeier: Ah never mind, those tests can be removed anyway

jakmeier:
> LGTM as is, but let's maybe fix the following in this PR:
> 
>     * drop prints altogether, rather than cfg(test) them
> 
>     * drop `#[test]` for stuff we moved over to `lib.rs`
> 
>     * rename constants like `ONE_NANOSECOND` to be more clear
> 
> 
> Adding actual separate costs for compilation time and gas can be folow-ups!

Cool, thanks for the quick review! I pushed those changes and also renamed `from_raw` to `from_gas` to make the (new) semantics of that function clear.

Issue #5485, opened 2021-11-26
Missing, malformed or inaccessible config file(s) cause code panics


nagisa:
_**Describe the bug**_ 

Running `neard` commands with an absent configuration file will cause `neard` invocations to panic, rather than to print out an error message and terminate regularly.

For example:

 1\. ./target/release/neard view_state peers
 Nov 26 17:56:58.036 INFO neard: Version: trunk, Build: crates-0.10.0-64-gdcab6f416-modified, Latest Protocol: 49
 thread 'main' panicked at 'Could not open config file: `~/.near/config.json`', nearcore/src/config.rs:469:33
 stack backtrace:
 0: rust_begin_unwind
 at /rustc/09c42c45858d5f3aedfa670698275303a3d19afa/library/std/src/panicking.rs:517:5
 1: core::panicking::panic_fmt
 at /rustc/09c42c45858d5f3aedfa670698275303a3d19afa/library/core/src/panicking.rs:101:14
 2: nearcore::config::Config::from_[file::](file::)([file::](file::))`closure`
 3: nearcore::config::Config::from_file
 4: nearcore::config::load_config_without_genesis_records
 5: nearcore::config::load_config
 6: state_viewer::cli::StateViewerSubCommand::run
 7: neard::cli::NeardCmd::parse_and_run
 note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
 ```

 Touching the config file causes a different, more opaque panic:

[...]

PR #5967, opened 2021-12-23
refactor: clean up some panics in favor of returning a Result

marcelo-gonzalez:
Here we change things so that functions that might panic on bad user input return a result instead, so we can exit normally from the `init` command. This doesn't get rid of every case, but tries to chip away at it a little.


core/crypto/src/key_file.rs
```diff
@@ -16,18 +16,16 @@ pub struct KeyFile {
...
-        let mut file = File::create(path).expect("Failed to create / write a key file.");
+    pub fn write_to_file(&self, path: &Path) -> std::io::Result<()> {
+        let mut file = File::create(path)?;
         #[cfg(platform = "unix")]
         {
             use std::os::unix::fs::PermissionsExt;
             let perm = std::fs::Permissions::from_mode(u32::from(libc::S_IWUSR | libc::S_IRUSR));
-            file.set_permissions(perm).expect("Failed to set permissions for a key file.");
-        }
-        let str = serde_json::to_string_pretty(self).expect("Error serializing the key file.");
```
pmnoxx, line 27: It may be a good idea to print the path, where this operation failed. That, would make it easier to debug if something happens. Otherwise, you kind of have to modify the code, to figure out what failed.

marcelo-gonzalez: Ah yea so what I tried to do here was make the callers print out the path in config.rs, like:

```rust
network_signer
                .write_to_file(&path)
                .with_context(|| format!("Error writing key file to {}", path.display()))?
```
But now that I look at it again maybe it's easier to do that in this function here, yeah. Since we'd have to do that in every caller anyway. Maybe the way to go is to make this function return an `anyhow::Result<()>` and call `with_context` in here?

bowenwang1996 (approved):

pmnoxx (approved):
LGTM

nikurt (approved):

Issue #5969, opened 2021-12-23
Prefer `neard-sandbox` to `near-sandbox`


mina86:
Currently `make sandbox{,-release}` creates a binary called `near-sandbox`.  That naming is inconsistent with how we call the plain executable.  It would be better if we used `neard-sandbox` instead (notice the `d`).

Just changing the name is of course a trivial affair but we are also uploading sandbox release binary under `near-sandbox` name and documentation refer to it.  Those would need to be updated as well.

matklad:
To play devil's advocate, the current naming isn't necessary bad `neard` is "near daemon", and `near-sandbox` is, well, "near sandbox", not "near daemon sandbox". Today the two are rather similar, but I can see them diverging in the future, and becoming two separaet binaries using a common shared library. For example, it doesn't **really** make sense for the sandbox to carry the whole rocksdb, and pure in-memory store can make sense for it. 

There was also the idea that sandbox name is provisional, and that we might want to rename the thing once we figure out the overall strategy. So something like `neard-workspace` might make sense? 

@ChaoticTempest as the primary customer is probably the right person to make a call on what the name *should* be.

mina86:
> well, "near sandbox", not "near daemon sandbox".

But is it a ‘near sandbox daemon’? ;) I think of it with analogy to daemon built with nightly features which I would call `neard-nightly` rather than `near-nightly`.

Also note that `make docker-nearcore-sandbox` uses a `neard` executable name.

PR #5970, opened 2021-12-23
Makefile: add `neard-sandbox` executable name

mina86:
For consistency with plain release naming, we probably should prefer
`neard-sandbox` to `near-sandbox` executable name.  This commit
changes `sandbox` and `sandbox-release` make targets to provide this
new name.

This does not remove the old name since God knows how many things
depend on it (we are uploading sandbox binary as `near-sandbox` and
documentation refers to that name) but having the new name will allow
us to phase out the legacy one with time.

Issue: https://github.com/near/nearcore/issues/5969


mina86:
@mm-near, friendly ping

pytest/tests/sandbox/patch_state.py
```diff
@@ -22,18 +22,18 @@
...
+        for exe in ('neard-sandbox', 'near-sandbox', 'neard'):
             if (root / exe).exists():
                 logger.info(
                     f'Using {(root / exe).relative_to(repo_dir)} binary')
                 CONFIG['near_root'] = str(root)
                 CONFIG['binary_name'] = exe
                 return
-    assert False, ('Unable to figure out location of near-sandbox binary; '
+    assert False, ('Unable to figure out location of neard-sandbox binary; '
                    'Did you forget to run `make sandbax`?')
```
mm-near, line 37: typo in sandbox?

mm-near (approved):

pytest/tests/sandbox/patch_state.py
mina86: Indeed, fixed.

PR #5981, opened 2021-12-30
Use make targets when building docker images

mina86:
Rather than running `cargo build` to compile `neard` executable when
building Docker images, use make target which produce a better
optimised version of the binary.


Dockerfile
```diff
@@ -26,16 +26,11 @@ VOLUME [ /near ]
...
-ARG features=default
-RUN cargo build -p neard --release --features $features && \
-    mkdir /tmp/build && \
-    cd /tmp/target/release && \
-    mv ./neard /tmp/build
-
-COPY scripts/run_docker.sh /tmp/build/run.sh
+ARG make_target=
+RUN make CARGO_TARGET_DIR=/tmp/target \
+         RUSTC_FLAGS='-C target-cpu=x86-64' \
```
nagisa, line 32: This environment variable doesn't seem standard. I think this may be doing nothing and can be removed.

mina86: By the way, do you know what `PORTABLE` might be?

nagisa: :shrug: but it isn't an environment variable rust toolchain inspects either.

mina86:
@nagisa, friendly ping.

Makefile
```diff
@@ -8,55 +8,65 @@ export NEAR_RELEASE_BUILD = no
...
-nightly-release:
-	cargo build -p neard --release --features nightly_protocol,nightly_protocol_features,performance_stats,memory_stats
+
+nightly-release: neard-nightly-release
 	cargo build -p state-viewer --release --features nearcore/nightly_protocol,nearcore/nightly_protocol_features,nearcore/performance_stats,nearcore/memory_stats
 	cargo build -p store-validator --release --features nearcore/nightly_protocol,nearcore/nightly_protocol_features,nearcore/performance_stats,nearcore/memory_stats
 	cargo build -p runtime-params-estimator --release --features nightly_protocol,nightly_protocol_features,nearcore/nightly_protocol,nearcore/nightly_protocol_features,nearcore/performance_stats,nearcore/memory_stats
 	cargo build -p genesis-populate --release --features nearcore/nightly_protocol,nearcore/nightly_protocol_features,nearcore/performance_stats,nearcore/memory_stats
 
+neard-nightly-release:
```
nagisa, line 66: Very nice.

nagisa (approved):

PR #5984, opened 2022-01-04
show log in VMOutcome debug, used in vm-runner-standalone output

ailisp:
I'm using this in quickly test a contract with near-vm-runner-standalone. Example output after this change:
```
StandaloneOutput {
    outcome: Some(
        VMOutcome: balance 10000000000000000000000000 storage_usage 100 return data None burnt gas 1823742379209 used gas 1823742379209 logs ["Hello Near"],
    ),
    err: None,
    receipts: [],
    state: State(
        {},
    ),
}
```


Rishabhraghwendra18 (approved):

runtime/near-vm-logic/src/logic.rs
```diff
@@ -2554,8 +2554,8 @@ impl std::fmt::Debug for VMOutcome {
         };
         write!(
             f,
-            "VMOutcome: balance {} storage_usage {} return data {} burnt gas {} used gas {}",
-            self.balance, self.storage_usage, return_data_str, self.burnt_gas, self.used_gas
+            "VMOutcome: balance {} storage_usage {} return data {} burnt gas {} used gas {} logs {:?}",
```
nagisa, line 2557: How likely is it that `logs` will be shown with many lines in the output? If it is “quite likely”, then I think it may be quite important that we figure out how to pass through the `#` formatting flag here. This way the callers of this `Debug` implementation can get a choice of printing 1 log line per line of output (i.e. pretty printed).

You can check for the flag by calling [`Formatter::alternate`](https://doc.rust-lang.org/stable/std/fmt/struct.Formatter.html#method.alternate)

matklad (commented):
This is potentially problematic: today, the `Debug` impl is deliberately written to avoid arbitrary-long outputs -- for return data, we print just the number of bytes, rather than the whole output. 

Logs can be rather sizable (16kilobyte per outcome), I think we should add logic to make sure that we don't print too much. 

We can do either of the following:

**Approach 1**: use `alternate` to erase log and data for `{:?}` but to print them in full in `{:#?}`

```rust
            let return_data_str = match &self.return_data {
                ReturnData::None => "None".to_string(),
                ReturnData::ReceiptIndex(_) => "Receipt".to_string(),
                ReturnData::Value(v) => format!("Value [{} bytes]", v.len()),
            };
            f.debug_struct("VMOutcome")
                .field("ballance", &self.balance)
                .field("storage_usage", &self.storage_usage)
                .field(
                    "return_data",
                    if f.alternate() {
                        &self.storage_usage
                    } else {
                        &format_args!("{}", return_data_str)
                    },
                )
                .field("logs", if f.alternate() { &self.logs.len() } else { &self.logs.len() })
                .finish()
```

**Approach 2** Just unconditionally sanitize logs to some reasonable amount (like at most 3 messages at most 100 characters long). 

I'd probably favor Approach 1, as we do use debug output, eg here: 

https://github.com/near/nearcore/blob/960c4112111bc4a2f214af65c37bfa55a1552062/runtime/runtime/src/state_viewer/mod.rs#L258

making those lines significantly bigger due to logs doesn't seem great. 

While we are at it, I'd maybe changed the format to follow standard rust layout:

```rust 
impl std::fmt::Debug for VMOutcome {
    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
        let return_data_str = match &self.return_data {
            ReturnData::None => "None".to_string(),
            ReturnData::ReceiptIndex(_) => "Receipt".to_string(),
            ReturnData::Value(v) => format!("Value [{} bytes]", v.len()),
        };
        f.debug_struct("VMOutcome")
            .field("ballance", &self.balance)
            .field("storage_usage", &self.storage_usage)
            .field("return_data", &format_args!("{}", return_data_str))
            .field("logs", &self.logs)
            .finish()
    }
}
```

Actually, **Alternative 3** -- let's just change code in the standalone runner to print logs  explicitely? This'll also allow us to use `Display` for strings rather than debug.

ailisp:
THanks! I didn't know the story behind `:#?` and logs. I think what i want is simply knows the log from vm-standalone but not the neard tracing logs. Alternative 3 is the est, and also print the log in the best way (E.g. `\n` renders to newline)

runtime/near-vm-runner-standalone/src/main.rs
```diff
@@ -176,6 +176,13 @@ fn main() {
         }
     );
 
+    if let Some(outcome) = outcome.clone() {
+        println!("\nLogs:");
+        for log in outcome.logs {
+            println!("{}\n", log);
+        }
+    }
```
matklad, line 184:
```suggestion
    if let Some(outcome) = &outcome {
        println!("\nLogs:");
        for log in &outcome.logs {
            println!("{}", log);
        }
    }
```

matklad (approved):

pmnoxx (approved):
LGTM

runtime/near-vm-runner-standalone/src/main.rs
ailisp: The `&` sounds good. but i don't suggest to remove `\n` because people always log without explicit `\n` (I don't see an exception, because logs are stored as elements in vec and rendered as different lines on explorer):
```
        env::log(
            format!(
                "@{} deposited {}. New unstaked balance is {}",
                account_id, amount, account.unstaked
            )
            .as_bytes(),
        );
```
So I think newline after each log here will make it readable and matches explorer displays.

PR #5986, opened 2022-01-04
feat: tool for measuring store stats

Longarithm:
Tool for measuring RocksDB stats for each column: 
- number of entries
- column size
- total keys size
- total values size

We use `sst_dump` for this, because native tools doesn't fully satisfy our needs, especially ones ported to Rust.

# Test plan

Existing tests.


tools/rocksdb-stats/src/main.rs
```diff
@@ -0,0 +1,85 @@
...
+    for (i, line) in lines.iter().enumerate() {
+        if line.contains("Process") {
+            sst_file_breaks.push(i);
+        }
+    }
+    sst_file_breaks.push(lines.len());
+
+    let mut column_data: HashMap<String, Data> = HashMap::new();
+
+    for i in 1..sst_file_breaks.len() {
```
nagisa, line 72: If `slice::array_windows` was stable, you could do:

```rust
for [start, end] in sst_file_breaks.array_windows::<2>() { 
    let sst_file_dump = &lines[start..end];
    // <snip>
}
```

…alas :(

tools/rocksdb-stats/src/main.rs
```diff
@@ -0,0 +1,85 @@
...
+        let split_line: Vec<&str> = line.split(':').collect();
+        if split_line.len() < 2 {
+            continue;
+        }
+        let line = split_line[0].trim();
+        let value = split_line[1].trim();
+        eprintln!("{} {}", line, value);
+
+        match line {
+            "column family name" => data.col = value.to_string(),
```
nagisa, line 38: Would be nice to have an assert of some kind to ensure that this line does not appear in the output multiple times.

(Or a comment if the `sst` format guarantees that this line can only appear 0 or 1 times in the output)

nagisa (commented):
Overall looks good to me.

tools/rocksdb-stats/src/main.rs
```diff
@@ -0,0 +1,107 @@
...
+                .get(SST_FILE_DUMP_LINES[2])
+                .unwrap()
+                .parse::<u64>()
+                .unwrap(),
+            raw_key_size: values.get(SST_FILE_DUMP_LINES[3]).unwrap().parse::<u64>().unwrap(),
+            raw_value_size: values.get(SST_FILE_DUMP_LINES[4]).unwrap().parse::<u64>().unwrap(),
+        }
+    }
+
+    pub fn merge(&mut self, other: &Data) {
```
pmnoxx, line 63: nit: other: &Self

tools/rocksdb-stats/src/main.rs
```diff
@@ -0,0 +1,107 @@
...
+        let data = Data::from_sst_file_dump(&lines[sst_file_breaks[i - 1]..sst_file_breaks[i]]);
+        if let Some(x) = column_data.get_mut(&data.col) {
+            x.merge(&data);
+        } else {
+            column_data.insert(data.col.clone(), data);
+        }
+    }
+
+    let mut column_data_list: Vec<&Data> = column_data.values().collect();
+    column_data_list.sort_by_key(|data| -(data.estimated_table_size as i64));
```
pmnoxx, line 104: nit: Can be written without casting to i64.
```rust
    column_data_list.sort_by_key(|data| u64::MAX - data.estimated_table_size);
```

tools/rocksdb-stats/src/main.rs
```diff
@@ -0,0 +1,107 @@
...
+    }
+}
+
+fn main() {
+    let home_dir = get_default_home();
+    let store_dir = get_store_path(&home_dir);
+    let mut cmd = Command::new("sst_dump");
+    cmd.arg(format!("--file={}", store_dir.to_str().unwrap()))
+        .arg("--show_properties")
+        .arg("--command=none"); // For some reason, adding this argument makes execution 20x faster
```
pmnoxx, line 77: nit: It may be a good idea to use `clap`. We have a lot of examples in our repo.

pmnoxx (approved):
LGTM

tools/rocksdb-stats/src/main.rs
Longarithm: Oh, this is cool

matklad:
Consider adding this as a subcommand to `neard`. This would have two primary benefits: 

* reducing compile times, as we'd have one copy of rocksdb to link less
* ability to just invoke the tool on validators/archival nodes without the need to compile yet another binary.

tools/rocksdb-stats/src/main.rs
```diff
@@ -0,0 +1,107 @@
...
+        // Mapping from SST file dump key to value.
+        let mut values: HashMap<&str, &str> = Default::default();
+
+        for line in lines {
+            let split_line: Vec<&str> = line.split(':').collect();
+            if split_line.len() < 2 {
+                continue;
+            }
+            let line = split_line[0].trim();
+            let value = split_line[1].trim();
```
matklad, line 32: https://doc.rust-lang.org/stable/std/primitive.str.html#method.split_once ?

tools/rocksdb-stats/src/main.rs
```diff
@@ -0,0 +1,107 @@
...
+    estimated_table_size: u64,
+    // Total size of all keys in bytes.
+    raw_key_size: u64,
+    // Total size of all values in bytes.
+    raw_value_size: u64,
+}
+
+// SST file dump keys we use to collect statistics.
+const SST_FILE_DUMP_LINES: [&str; 5] =
+    ["column family name", "# entries", "(estimated) table size", "raw key size", "raw value size"];
```
matklad, line 19: Alternative: 

```suggestion
const SST_FILE_DUMP_LINES: &[&str] =
    &["column family name", "# entries", "(estimated) table size", "raw key size", "raw value size"];
```

tools/rocksdb-stats/src/main.rs
```diff
@@ -0,0 +1,107 @@
...
+
+    eprintln!("Parsing output ...");
+    let out = std::str::from_utf8(&output.stdout).unwrap();
+    let mut sst_file_breaks: Vec<usize> = vec![];
+    let lines: Vec<&str> = out.lines().collect();
+    for (i, line) in lines.iter().enumerate() {
+        if line.contains("Process") {
+            sst_file_breaks.push(i);
+        }
+    }
```
pmnoxx, line 89: Or use:
```rust
    for (i, line) in lines.iter().enumerate().filter(|(_, line)|line.contains("Process")) {
        sst_file_breaks.push(i);
    }
```

tools/rocksdb-stats/src/main.rs
```diff
@@ -0,0 +1,107 @@
...
+        self.raw_key_size += other.raw_key_size;
+        self.raw_value_size += other.raw_value_size;
+    }
+}
+
+fn main() {
+    let home_dir = get_default_home();
+    let store_dir = get_store_path(&home_dir);
+    let mut cmd = Command::new("sst_dump");
+    cmd.arg(format!("--file={}", store_dir.to_str().unwrap()))
```
matklad, line 75: Would the following work? 

```suggestion
    cmd.arg("--file").arg(&store_dir)
```

? Path are not strings, better not to call `to_str` and the like.

tools/rocksdb-stats/src/main.rs
```diff
@@ -0,0 +1,107 @@
...
+        .arg("--show_properties")
+        .arg("--command=none"); // For some reason, adding this argument makes execution 20x faster
+    eprintln!("Running {:?} ...", cmd);
+    let output = cmd.output().expect("sst_dump command failed to start");
```
matklad, line 79: I suggest using `anyhow` rather than `.expect` / `.unwrap` from the get go.

matklad: misses 

```rust
if !output.status().success() {
   bail!("...")
}
```

check

tools/rocksdb-stats/src/main.rs
```diff
@@ -0,0 +1,107 @@
...
+        let data = Data::from_sst_file_dump(&lines[sst_file_breaks[i - 1]..sst_file_breaks[i]]);
+        if let Some(x) = column_data.get_mut(&data.col) {
+            x.merge(&data);
+        } else {
+            column_data.insert(data.col.clone(), data);
+        }
+    }
+
+    let mut column_data_list: Vec<&Data> = column_data.values().collect();
+    column_data_list.sort_by_key(|data| u64::MAX - data.estimated_table_size);
```
matklad, line 104:
```suggestion
    column_data_list.sort_by_key(|data| std::cmp::Reverse(data.estimated_table_size));
```

tools/rocksdb-stats/src/main.rs
```diff
@@ -0,0 +1,107 @@
...
+    eprintln!("Printing stats ...");
+    println!("{}", serde_json::to_string(&column_data_list).unwrap());
```
matklad, line 106:
```suggestion
    println!("{}", serde_json::to_string_pretty(&column_data_list).unwrap());
```

tools/rocksdb-stats/src/main.rs
```diff
@@ -0,0 +1,107 @@
...
+    let mut column_data: HashMap<String, Data> = HashMap::new();
+
+    for i in 1..sst_file_breaks.len() {
+        let data = Data::from_sst_file_dump(&lines[sst_file_breaks[i - 1]..sst_file_breaks[i]]);
+        if let Some(x) = column_data.get_mut(&data.col) {
+            x.merge(&data);
+        } else {
+            column_data.insert(data.col.clone(), data);
+        }
+    }
```
matklad, line 101: Isn't this just https://doc.rust-lang.org/stable/std/primitive.slice.html#method.split ?

tools/rocksdb-stats/src/main.rs
```diff
@@ -0,0 +1,107 @@
...
+                    if values.contains_key(line) {
+                        panic!(
+                            "Line {} was presented twice and contains values {} and {}",
+                            line,
+                            values.get(line).unwrap(),
+                            value
+                        );
+                    } else {
+                        values.insert(line, value);
+                    }
```
matklad, line 45:
```suggestion
                    let prev = values.insert(line, value);
                    assert!(prev.is_none(),                             "Line {} was presented twice and contains values {} and {}",
                            line,
                            values.get(line).unwrap(),
                            value
)
```

tools/rocksdb-stats/src/main.rs
pmnoxx, line 106: It may be a good idea to use `tracing` instead of `println`.

Longarithm: I'm not sure. I want output of script to be a raw json which could be easily passed to some other script.

But as for `eprintln`s, that's right.

pmnoxx: @Longarithm Yes, I was talking about  `eprintln!("Printing stats ...")`.

PR #5987, opened 2022-01-04
refactor: Runtime structure now contains VMConfig

nagisa:
Conceptually this matches what is happening somewhat better – VM is
constructed as configured in some specific way and then can be used to
run multiple contracts or somesuch, as long as the configuration
matches.

This kind of construction can help, in the future, to reuse more state
between contract invocations. Since `VMConfig` doesn't change within the
same run of `neard` typically, we could reuse the same memory map to
hold the contract code and cache the allocation of the working memory as
well (so long as we clear it between contract invocations), thus saving
on page faults and similar cost centres.

### DIFF
--- a/runtime/near-vm-runner-standalone/src/script.rs
+++ b/runtime/near-vm-runner-standalone/src/script.rs
@@ -127,7 +127,6 @@ impl Script {
                         &step.method,
                         &mut external,
                         step.vm_context.clone(),
-                        &self.vm_config,
                         runtime_fees_config,
                         &step.promise_results,
                         self.protocol_version,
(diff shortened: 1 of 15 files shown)


matklad:
Makes sense to me!

One extra thing we can do is to move part of `get_contract_cache_key` to a `fingerprint` method on a Runtmie, as Runtime now encapsulates "sameness" of the runtime envrionment. Not neccessary for this PR though. 

>Since VMConfig doesn't change within the same run of neard typically,

I want to unpack this a bit: VMConfig doesn't change within a single block, but it totally can change during a protocol upgrade, and protocol upgrade doesn't require rebooting the node. Want to be a bit nitpicky here, as we can't in general cater to "typical" use-cases, for the standrard problem with caches.

matklad (approved):

Issue #5255, opened 2021-11-14
Add store configuration to `config.json`


pmnoxx:
We would like to be able to adjust cache configuration per machine basis. Right now, column cache configuration is hard coded inside `db.rs`.

Done:
- In this change, we will add default column cache size, and ability to choose column cache size per each column separately.

In another PR: (if requested)
- We may add number of open files limit
- Block size, etc.

As discussed in https://github.com/near/nearcore/issues/5203#issuecomment-968284705

PR https://github.com/near/nearcore/pull/5249

PR #5998, opened 2022-01-05
Add database cache settings to config.json

pmnoxx:
Hey, this is another epic change. We will add cache configuration for our store. to `config.json`
However, some big changes are needed.

This will allow us to specify cache, for db columns, etc. 

Resolves https://github.com/near/nearcore/issues/5255

TODO
- [x] Test on a node

### DIFF
--- a/core/store/src/db.rs
+++ b/core/store/src/db.rs
@@ -49,6 +49,8 @@ impl Into<io::Error> for DBError {
 /// This enum holds the information about the columns that we use within the RocksDB storage.
 /// You can think about our storage as 2-dimensional table (with key and column as indexes/coordinates).
 // TODO(mm-near): add info about the RC in the columns.
+pub const NUM_DB_COLS: usize = 50;
+
 #[derive(PartialEq, Debug, Copy, Clone, EnumIter, BorshDeserialize, BorshSerialize, Hash, Eq)]
 pub enum DBCol {
     /// Column to indicate which version of database this is.
@@ -201,9 +203,6 @@ pub enum DBCol {
     ColStateChangesForSplitStates = 49,
 }
 
-// Do not move this line from enum DBCol
-pub const NUM_COLS: usize = 50;
-
 impl std::fmt::Display for DBCol {
     fn fmt(&self, formatter: &mut std::fmt::Formatter<'_>) -> Result<(), std::fmt::Error> {
         let desc = match self {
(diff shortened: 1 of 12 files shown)


core/store/src/db.rs
```diff
@@ -201,9 +203,6 @@ pub enum DBCol {
     ColStateChangesForSplitStates = 49,
 }
 
-// Do not move this line from enum DBCol
-pub const NUM_COLS: usize = 50;
```
matklad, line 205: Seems better if this line is last? So that it's next to `ColStateChangesForSplitStates = 49,`. In the current state, note that the "This enum holds" documentation comment is misplaced.

core/store/src/db.rs
```diff
@@ -751,11 +758,15 @@ fn choose_cache_size(col: DBCol) -> usize {
...
+fn rocksdb_column_options(col: DBCol, extra_db_memory: usize) -> Options {
     let mut opts = Options::default();
     opts.set_level_compaction_dynamic_level_bytes(true);
-    let cache_size = choose_cache_size(col);
-    opts.set_block_based_table_factory(&rocksdb_block_based_options(cache_size));
+
+    // In config there is `extra_db_memory_bytes` setting. It specifies extra memory tht can be used
+    // to improve database performance. For now, let's simply add the same cache size for each column.
+    // We can add a better way of using extra memory later.
+    let cache_size = choose_cache_size(col) + extra_db_memory / NUM_DB_COLS;
```
matklad, line 768: no strong opinion here, but it feel that ` extra_db_memory / NUM_DB_COLS`  logic belongs to the caller.

pmnoxx:
```
@bowenwang1996 @mm-near: chain/ core/ nearcore/
@nikurt: genesis-tools/
```
@bowenwang1996 @mm-near @nikurt Can you take a look?

nikurt (approved):

nearcore/src/config.rs
```diff
@@ -433,6 +433,8 @@ pub struct Config {
     /// If set, overrides value in genesis configuration.
     #[serde(skip_serializing_if = "Option::is_none")]
     pub max_gas_burnt_view: Option<Gas>,
+    /// Extra memory database can use for purpose of caches.
+    pub extra_db_memory_bytes: usize,
```
mm-near, line 437: I'd do it a little bit different - and create the show "sub struct" that would keep the db settings.

(as we will probably play with other settings too - like what is the size of files, compression etc etc).

mm-near (changes requested):

PR #5999, opened 2022-01-05
re: `near-network` Remove `chain/network/src/routing/routing.rs`

pmnoxx:
`chain/network/src/routing/routing.rs` is no longer used, and should be removed.


pmnoxx:
```
```

mm-near (approved):

PR #6002, opened 2022-01-06
chain: document that `assign_shards` expects producers to be sorted

mina86:
The `assign_shards` function assumes that chunk producers are given in
the descending order.  This has never been stated explicitly though so
document that and also add a check to verify this is the case.


chain/epoch_manager/src/shard_assignment.rs
```diff
@@ -11,21 +11,26 @@ use std::cmp;
...
     let num_chunk_producers = chunk_producers.len();
     if num_chunk_producers < min_validators_per_shard {
         // Each shard must have `min_validators_per_shard` distinct validators,
         // however there are not that many, so we cannot proceed.
         return Err(NotEnoughValidators);
     }
+
+    // We’re using greedy algorithm so make sure chunk producers are sorted from
+    // the one with biggest stake.
+    chunk_producers.sort_by(|a, b| b.get_stake().cmp(&a.get_stake()));
```
mzhangmzz, line 27: I think the chunk producers here is already sorted, because of how `select_validators` works. It should rank validator stake from high to low and select by order.

I think it's probably better to add in the specs of `assign_shards` that it assumes `chunk_producers` are sorted by stake and adds a check in the function that checks the chunk producer stake is in decreasing order.

mzhangmzz (changes requested):

chain/epoch_manager/src/shard_assignment.rs
mina86: Checking is actually more code than simply sorting it turns out (`is_sorted_by` is a nightly feature) and `sort_by` is O(n) if sequence is already sorted.  With that I think that we should either sort or document and I’d prefer the former since that will guard against any potential in the future where someone forgets that the function expects producers to be ordered. Thoughts?

mzhangmzz: In my personal opinion, I would still prefer the check it's the same cost as sort and it is not that complicated, and it does not add unnecessary code that may cause confusion for people who are reading the code. Plus, I don't think the `assign_shards` function will be used by any where else in which it is not taking input from `select_validators` (the check actually serves as a guard for implementation errors in `select_validators`), expect in tests.

mina86: PTAL.

chain/epoch_manager/src/shard_assignment.rs
```diff
@@ -3,34 +3,46 @@ use near_primitives::types::{Balance, NumShards, ShardId};
...
     min_validators_per_shard: usize,
 ) -> Result<Vec<Vec<T>>, NotEnoughValidators> {
-    let num_chunk_producers = chunk_producers.len();
-    if num_chunk_producers < min_validators_per_shard {
-        // Each shard must have `min_validators_per_shard` distinct validators,
-        // however there are not that many, so we cannot proceed.
-        return Err(NotEnoughValidators);
+    for (idx, pair) in chunk_producers.windows(2).enumerate() {
+        if pair[0].get_stake() > pair[1].get_stake() {
+            log::warn!(target: "epoch_manager",
```
mzhangmzz, line 28: I think it's safe to panic here. If it's not sorted, that's a big issue and the node should stop.

mzhangmzz (commented):
Thanks!

chain/epoch_manager/src/shard_assignment.rs
mina86: PTAL

mzhangmzz (approved):

PR #6003, opened 2022-01-06
re: `PeerManagerActor` Simplify `send_message`

pmnoxx:
`send_message` function can be simplified in `PeerManagerActor` by changing it's signature from `self` to `Self`.

This is important step for adding tests to `PeerManagerActor` .

### DIFF
--- a/chain/network/src/peer_manager/peer_manager_actor.rs
+++ b/chain/network/src/peer_manager/peer_manager_actor.rs
@@ -22,6 +22,8 @@ use actix::{
     Recipient, Running, StreamHandler, WrapFuture,
 };
 use futures::task::Poll;
+#[cfg(feature = "protocol_feature_routing_exchange_algorithm")]
+use futures::FutureExt;
 use futures::{future, Stream, StreamExt};
 use near_network_primitives::types::{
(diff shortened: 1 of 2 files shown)


chain/network/src/peer_manager/peer_manager_actor.rs
```diff
@@ -1394,33 +1382,19 @@ impl PeerManagerActor {
...
-                })
-                .map(|_, _, _| ())
-                .spawn(ctx);
+            actix::spawn(connected_peer.addr.send(SendMessage { message }));
             true
         } else {
             debug!(target: "network",
                    to = ?peer_id,
-                   num_connected_peers = self.connected_peers.len(),
+                   num_connected_peers = connected_peers.len(),
```
mm-near, line 1397: hmm.. why does this debug log says 'sending message' - if message is not really sent?

mm-near (approved):

chain/network/src/peer_manager/peer_manager_actor.rs
pmnoxx: I changed it to "failed sending message"

PR #6004, opened 2022-01-06
Add metrics related to the epoch, sync state and node version

nikurt:
Adds the following metrics to the node:

* Epoch height
* "Header head" metric to complement the existing "block head" metric in case of sync issues
* Protocol version supported by the node
* DB version expected by the node
* Sync status of the node
* And the expected block height of the next protocol upgrade


chain/chain/src/chain.rs
```diff
@@ -4397,6 +4397,7 @@ impl<'a> ChainUpdate<'a> {
             let tip = Tip::from_header(header);
             self.chain_store_update.save_header_head_if_not_challenged(&tip)?;
             debug!(target: "chain", "Header head updated to {} at {}", tip.last_block_hash, tip.height);
+            metrics::HEADER_HEIGHT_HEAD.set(tip.height as i64);
```
mzhangmzz, line 4400: I think `Header_HEAD_HEIGHT` would make more sense for naming

chain/client-primitives/src/types.rs
```diff
@@ -153,6 +153,18 @@ impl SyncStatus {
             _ => true,
         }
     }
+
+    pub fn cast_to_int(&self) -> u8 {
+        match self {
+            SyncStatus::AwaitingPeers => 0,
+            SyncStatus::NoSync => 1,
```
mzhangmzz, line 160: It might make more sense to make `SyncStatus::NoSync` be 0 since we expect most nodes are at NoSync

chain/client/src/metrics.rs
```diff
@@ -75,3 +75,23 @@ pub static BLOCKS_AHEAD_OF_HEAD: Lazy<IntGauge> = Lazy::new(|| {
...
+pub static SYNC_STATUS: Lazy<IntGauge> =
+    Lazy::new(|| try_create_int_gauge("near_sync_status", "Node sync status").unwrap());
+pub static EPOCH_HEIGHT: Lazy<IntGauge> = Lazy::new(|| {
+    try_create_int_gauge("near_epoch_height", "Height of the epoch at the head of the blockchain")
+        .unwrap()
+});
+pub static PROTOCOL_UPGRADE_BLOCK_HEIGHT: Lazy<IntGauge> = Lazy::new(|| {
+    try_create_int_gauge(
+        "near_protocol_upgrade_block_height",
+        "Estimated block height of the protocol upgrade",
```
mzhangmzz, line 87: Just curious, what's the purpose for this metric?

nikurt: It lets validators set alerts about an imminent protocol upgrade. Multiple validators expressed interest in this alert.

chain/client-primitives/src/types.rs
```diff
@@ -153,6 +153,19 @@ impl SyncStatus {
             _ => true,
         }
     }
+
+    pub fn cast_to_int(&self) -> u8 {
```
bowenwang1996, line 157: `repr` is potentially a better name

bowenwang1996 (approved):

mzhangmzz (approved):

chefsale (approved):

PR #6005, opened 2022-01-06
refactor: move rocksdb stats to state viewer

Longarithm:
Apply suggestions from https://github.com/near/nearcore/pull/5986.

Also, move script to state viewer to be able to call it from both `state-viewer` and `neard`.

## Test plan

Existing tests.


nikurt (approved):

test-utils/state-viewer/src/cli.rs
```diff
@@ -279,3 +285,16 @@ impl EpochInfoCmd {
...
+#[derive(Clap)]
+pub struct RocksDBStatsCmd {
+    /// Location of the dumped Rocks DB stats.
+    #[clap(long, parse(from_os_str))]
+    file: Option<PathBuf>,
+}
+
+impl RocksDBStatsCmd {
+    pub fn run(self, home_dir: &Path) {
+        assert_matches!(get_rocksdb_stats(home_dir, self.file), Ok(_));
```
matklad, line 298: I don't think its worth pulling in `assert_matches!` as a runtime dependency just for this.

test-utils/state-viewer/src/rocksdb_stats.rs
```diff
@@ -68,31 +72,31 @@ impl Data {
...
         .arg("--show_properties")
         .arg("--command=none"); // For some reason, adding this argument makes execution 20x faster
     eprintln!("Running {:?} ...", cmd);
-    let output = cmd.output().expect("sst_dump command failed to start");
+    let output = cmd.output()?;
+    if !output.status.success() {
+        anyhow::bail!(
+            "failed to run sst_dump, {}, stderr: {}",
+            output.status,
+            std::str::from_utf8(&output.stderr)?
```
matklad, line 87: For diagnostic purposes, you generally want 

https://doc.rust-lang.org/stable/std/string/struct.String.html#method.from_utf8_lossy

test-utils/state-viewer/src/rocksdb_stats.rs
```diff
@@ -101,7 +105,13 @@ fn main() {
...
-    eprintln!("Printing stats ...");
-    println!("{}", serde_json::to_string(&column_data_list).unwrap());
+    column_data_list.sort_by_key(|data| std::cmp::Reverse(data.estimated_table_size));
+    let result = serde_json::to_string_pretty(&column_data_list).unwrap();
+
+    eprintln!("Dumping stats ...");
+    match file {
+        None => println!("{}", result),
+        Some(file) => std::fs::write(file, result)?,
+    }
```
matklad, line 115: Rather than passing in an `Option<PathBuf>`, a more general solution is to pass `&mut dyn Write`, that way the code doesn't have to care where does the output go.

test-utils/state-viewer/src/rocksdb_stats.rs
```diff
@@ -101,7 +105,13 @@ fn main() {
...
     let mut column_data_list: Vec<&Data> = column_data.values().collect();
-    column_data_list.sort_by_key(|data| u64::MAX - data.estimated_table_size);
-    eprintln!("Printing stats ...");
-    println!("{}", serde_json::to_string(&column_data_list).unwrap());
+    column_data_list.sort_by_key(|data| std::cmp::Reverse(data.estimated_table_size));
+    let result = serde_json::to_string_pretty(&column_data_list).unwrap();
+
+    eprintln!("Dumping stats ...");
+    match file {
+        None => println!("{}", result),
```
matklad, line 113: :+1: for using `stderr` for human output and `stdout` for JSON, that's the right way to do it!

test-utils/state-viewer/src/cli.rs
```diff
@@ -83,6 +85,9 @@ pub enum StateViewerSubCommand {
     /// Print `EpochInfo` of an epoch given by `--epoch_id` or by `--epoch_height`.
     #[clap(name = "epoch_info")]
     EpochInfo(EpochInfoCmd),
+    /// Dump stats for the RocksDB storage.
+    #[clap(name = "rocksdb_stats")]
```
matklad, line 89: unrelated to PR, but I am surprised that we need all these `name = ` attributes -- feels like something clap should just infer from the enum names...

matklad (approved):
LGTM! I'd probably remove assert_matches, everything else is relatively minor

Issue #5995, opened 2022-01-04
Robust IO cost measurements in estimator


jakmeier:
We have to make sure our IO measurements for the parameter estimator are:
* isolated from what runs before (e.g. setup) and afterwards
* immune to artifacts caused by the storage architecture

To get there, I think I need to first understand our storage layers better and we should probably reflect on what the goal of the measurement is.

### Todos
1. We have to understand the factors that cause IO.
    - Can we control the background threads for IO?
    - How does the shape of the Trie affect storage accesses?
    - What storage access pattern produce what IO patterns?
    - What is the cheapest way for an attacker to produce lots of IO?
    - (and many more questions)
2. Define what we *want* to measure
    - What are the assumptions behind IO measurements? E.g. which caches are enabled? What is the prior state?
3. Ensure the measurements **in all estimations** follow the defined assumptions and are appropriately isolated

### Background

When running the parameter estimator inside QEMU with the ICount metric, it reports IO bytes read and written. This is done by intercepting syscall which should give the most accurate picture for how much actual IO is triggered.

However, there is a number of challenges with this approach. Simply looking at the counters before and after the execution of some code *does not* give an accurate picture. There are at least 3 effects we should take into account.

 1. IO is not produced evenly throughout execution. Often, IO comes in large blocks with zero IO in between. This is due to caches and write buffers in different layers.
 2. The prior state of the database changes how expensive it is to access an entry in that DB. Both the shape of the Trie as well as the shape of the RocksDB LSM-Tree are relevant here.
 3. RocksDB handles IO in background threads, including flushes/write-backs as well as compaction.

### Prior Status Quo

Our IO measurements already disable (some) caches in an attempt to address point 1. Point 2 is somewhat addressed by using the same state every time, which at least makes the estimations consistent. Point 3 is not well understood right now, especially in the context of the estimator (at least by me).

# Work items

- [x] #3194
- [x] #3097
- [x] #3193
- [x] #5996 
- [x] #4566
- [x] #7053
- [x] #7440
- [x] #7060
- [x] #7681 
- [x] #7683
- [x] #7519 

# Proposal / ideas
- https://github.com/near/nearcore/issues/6445

bowenwang1996:
The work that @Longarithm did on visualizing the trie could be helpful here

jakmeier:
Some thoughts with regards to background threads:
[...]

PR #6006, opened 2022-01-06
feat: Reintroduce single_thread_rocksdb

jakmeier:
This has been temporarily removed from Cargo.toml to wait for
a rocksdb release (see #3835)

Test plan
---------
Running existing store tests but with --features single_thread_rocksdb

### DIFF
--- a/core/store/Cargo.toml
+++ b/core/store/Cargo.toml
@@ -46,6 +46,7 @@ harness = false
 [features]
 default = []
 no_cache = []
+single_thread_rocksdb = []
 test_features = []
 protocol_feature_chunk_only_producers = []
 nightly_protocol = []
--- a/core/store/Cargo.toml
+++ b/core/store/Cargo.toml
@@ -46,7 +46,7 @@ harness = false
 [features]
 default = []
 no_cache = []
-single_thread_rocksdb = []
+single_thread_rocksdb = [] # Deactivate RocksDB IO background threads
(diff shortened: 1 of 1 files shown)


ailisp:
Looks good to me, As we discussed, to run param estimator with "single_thread_rocksdb" would require some work to c++ rocksdb to avoid blocking. That, if make changes to rocksdb, will need to bump rocksdb-sys in rust-rocksdb, and wait rust-rocksdb's new release. Before all that happen, we might need to use git dependency of rust-rocksdb again.

ailisp (approved):

jakmeier:
Yes, this problem still exists. (I replicated the hang you described to me, i.e. after writing lots of data.)

So, running the full estimator without background IO threads is not possible without said C++ changes. But after some more thoughts since our last exchange, I am no longer chasing that strategy. My current opinion is we should embrace the fact that IO happens in background threads in all cases where IO is relevant. (More on that over at #5995)

But "single_thread_rocksdb" *might* be useful to run estimations that really shouldn't have any IO but I am still experimenting with that.

The main reason I created this PR is because we currently have some code behind this feature flag which is completely dead because it no longer listed. (I probably should have written that when I submitted the PR)

matklad (approved):
SGTM!

*Might* make sense to add a short comment to Cargo.toml explaining what does this do.

PR #6015, opened 2022-01-07
pytest: fix `state_sync_routed.py` test

mina86:
Commit c247ea5b: ‘Change formatting of messages in `PeerManager` to
use `tracing`’ changed formatting of log lines produced by the peer
manager.  Most notably, it changed ‘Consolidated connection’ line such
that `state_sync_routed.py` test could no longer see it.

Update the pattern the test is using to match the new line format and
thus fix the test.


bowenwang1996 (approved):
Would be nice if we have a better way to catch the error earlier

mm-near (approved):

PR #6018, opened 2022-01-07
feat: Implement download file progress bar

### LINKED ISSUE #4617: `neard init --download-genesis` should have some progress reporting
Currently, one just need to know that it takes a while to download the genesis file and otherwise it appears to be stuck on that command "forever". We may either just warn that we expect to download a bunch of data or ideally provide a progress reporting.

cc @janewang @chefsale

itegulov:
Fixes #4617 

This PR introduces a download progress bar using [indicatif](https://github.com/console-rs/indicatif). I used the default download style they showcase in one of their [examples](https://github.com/console-rs/indicatif/blob/main/examples/download.rs) because I think it fits pretty well to the existing color scheme here. But let me know if you prefer something to be restyled. The gif below shows what it looks like (with middle cut out for the sake of brevity):

progress-bar


miraclx (approved):
Love this, nice work!

matklad (approved):

nearcore/src/config.rs
```diff
@@ -1160,10 +1161,19 @@ async fn download_file_impl(
     let https_connector = hyper_tls::HttpsConnector::new();
     let client = hyper::Client::builder().build::<_, hyper::Body>(https_connector);
     let mut resp = client.get(uri).await?;
+    let bar_opt = resp.size_hint().upper().map(|file_size| {
+        let bar = ProgressBar::new(file_size);
+        bar.set_style(ProgressStyle::default_bar()
+            .template("{spinner:.green} [{elapsed_precise}] [{wide_bar:.cyan/blue}] {bytes}/{total_bytes} ({eta})")
+            .progress_chars("#>-"));
+        bar
+    });
```
matklad, line 1170: GitHub ate my comment?

```suggestion
    }).unwrap_or_else(ProgressBar::hidden);
```

to avoid dealing with options down the line.

nearcore/src/config.rs
```diff
@@ -1160,10 +1161,19 @@ async fn download_file_impl(
     let https_connector = hyper_tls::HttpsConnector::new();
     let client = hyper::Client::builder().build::<_, hyper::Body>(https_connector);
     let mut resp = client.get(uri).await?;
+    let bar_opt = resp.size_hint().upper().map(|file_size| {
+        let bar = ProgressBar::new(file_size);
+        bar.set_style(ProgressStyle::default_bar()
+            .template("{spinner:.green} [{elapsed_precise}] [{wide_bar:.cyan/blue}] {bytes}/{total_bytes} ({eta})")
```
mina86, line 1167: Maybe also include `per_sec`?

nearcore/src/config.rs
mina86: Even better, use a style which would shows how much has been downloaded and `per_sec`.

mina86 (approved):

nearcore/src/config.rs
itegulov: Nice idea!

nearcore/src/config.rs
itegulov: Done

nearcore/src/config.rs
```diff
@@ -1161,12 +1162,33 @@ async fn download_file_impl(
...
+        }
+    };
     while let Some(next_chunk_result) = resp.data().await {
         let next_chunk = next_chunk_result.map_err(FileDownloadError::HttpError)?;
         file.write_all(next_chunk.as_ref())
             .await
             .map_err(|e| FileDownloadError::WriteError(path.to_path_buf(), e))?;
+        bar.inc(next_chunk.len() as u64);
     }
+    bar.finish();
```
mina86, line 1191: I’m assuming it’s ok not ta call `bar.finish()` in case of an early return?

nearcore/src/config.rs
```diff
@@ -1161,12 +1162,33 @@ async fn download_file_impl(
     let https_connector = hyper_tls::HttpsConnector::new();
     let client = hyper::Client::builder().build::<_, hyper::Body>(https_connector);
     let mut resp = client.get(uri).await.map_err(FileDownloadError::HttpError)?;
+    let bar = match resp.size_hint().upper() {
```
mina86, line 1165: I’d use `if let` but `match` also works I suppose.

mina86 (approved):
LGTM

nearmax:
Nit. @miraclx @austinabell I would've made progress bar optional, e.g. through verbosity flag, to make output easier to pipe with bash.

mina86:
> General progress bar behaviors:
>
> * if a non terminal is detected the progress bar will be completely hidden. This makes piping programs to logfiles make sense out of the box.

So I think we’re good.

nearcore/src/config.rs
itegulov: It will just leave an unfinished progress bar like this one which I think is fine:

latest-screenshot

nearcore/src/config.rs
itegulov: Yeah, I guess that is the more canonical way of doing that

itegulov:
@bowenwang1996 @mm-near I need a codeowner review on this PR. Could you please take a look?

mm-near (approved):

Issue #5423, opened 2021-11-22
Good places to refactor


pmnoxx:
The purpose of this issue is to track ant remaining refactoring that need to be done, created issue for. This is to make sure we don't forget really good comments. And, at the same time we limit scope of changes, so we don't do too code changes in one piece of code.

- [x] Rewrite `RateCounter` #5225 
- [ ] Refactor `PeerStatsResult` https://github.com/near/nearcore/pull/5225/files#r754529253
```rust
pub struct Throughput {
  pub n_messages: usize,
  pub n_bytes: usize
}

struct PeerStatsResult {
  send_per_sec: Throughput,
  recv_per_sec: Throughput,
  is_abusive: bool,
  chain_info: ChainInfo,
}
```
- [x] #5402 
- [x] #5371
- [x] #5416 
- [ ] avoid conditional imports https://github.com/near/nearcore/pull/5373#issuecomment-976584358
```rust
#[cfg(feature = "deepsize")]
use deepsize::DeepSizeOf;
pub use error::{ParseAccountError, ParseErrorKind};

#[cfg_attr(feature = "deepsize", derive(DeepSizeOf))]
#[derive(Eq, Ord, Hash, Clone, Debug, PartialEq, PartialOrd)]
pub struct AccountId(Box<str>);
```

==>>

```rust
pub use error::{ParseAccountError, ParseErrorKind};

#[cfg_attr(feature = "deepsize", derive(deepsize::DeepSizeOf))]
#[derive(Eq, Ord, Hash, Clone, Debug, PartialEq, PartialOrd)]
pub struct AccountId(Box<str>);
```
- [ ] Move out `peer_id_type_field_len`, `is_forward_tx` out of `near-network::codec.rs`
- [ ]  Use debug!(message="...", ?A, ?B, ?C); syntax in rest of the code base simialar to    https://github.com/near/nearcore/pull/5392
- [ ] Hide feature inside delay detector   https://github.com/near/nearcore/pull/5436#pullrequestreview-814734494
- [ ] Don't compute current time multiple times in `near-network` https://github.com/near/nearcore/pull/5556#discussion_r769589102

bowenwang1996:
It is worth mentioning what kind of refactoring you refer to here. "Remaining refactorings" is not very descriptive.

pmnoxx:
> It is worth mentioning what kind of refactoring you refer to here. "Remaining refactorings" is not very descriptive.

Yes, we can write a custom CI pipeline, which does the check.

[...]

PR #6020, opened 2022-01-07
feat: `RateCounter` rewrite

pmnoxx:
As @mm-near noticed, `truncate` method is not called before getting bandwidth stats. This can lead to displayed bandwidth data to be incorrect, especially if no data is send/received.

We will also refactor this code: https://github.com/near/nearcore/pull/5218

Edit:
- Rewrote the code completely from scratch, this will help us improve code quality and get rid of license.

See #5423


pmnoxx:
```
```
@mm-near Can you take a look?

mm-near (approved):

PR #6023, opened 2022-01-07
feat: Print cargo test output only on error.

pmnoxx:
Change logger to print output of `cargo test` correctly.

https://github.com/near/nearcore/issues/4490#issuecomment-1007554120


matklad (approved):
<3

PR #6024, opened 2022-01-07
fix(error messages): fix error messages when downloading the config

marcelo-gonzalez:
This was missed, but https://github.com/near/nearcore/pull/5967 did
not interact very well with `enum FileDownloadError`, so this is an
attempt to fix that, and present better error messages.

The reason for getting rid of the #[from] in the HttpError variant is
that it doesn't really play nicely with the anyhow {:#} formatter and
gives us ugliness like:

ERROR neard::cli: Failed to initialize configs: Failed to download the config file from http://localhost:8002/config.json: error trying to connect: tcp connect error: Connection refused (os error 111): tcp connect error: Connection refused (os error 111): Connection refused (os error 111)

Test Plan:
Of course check that downloading the file still works, but also inject
errors and make sure the error messages look ok. For reviewer
convenience, error msgs corresponding to the variants below (except
for RemoveTemporaryFileError. Didn't get to that one... but I think
it's probably good)

HttpError:

ERROR neard::cli: Failed to initialize configs: Failed to download the config file from http://localhost:8002/config.json: error trying to connect: tcp connect error: Connection refused (os error 111)

OpenError:

ERROR neard::cli: Failed to initialize configs: Failed to download the config file from http://localhost:8002/config.json: Failed to open temporary file: No such file or directory (os error 2) at path "/asdf/.tmp2g9Pqe"

WriteError:

ERROR neard::cli: Failed to initialize configs: Failed to download the config file from http://localhost:8000/config.json: Failed to write to temporary file at /tmp/asdfa/.tmpNdy4iL: Bad file descriptor (os error 9)

RenameEror:

ERROR neard::cli: Failed to initialize configs: Failed to download the config file from http://localhost:8000/config.json: Failed to rename temporary file /tmp/asdfa/.tmpRvBBEn to /tmp/asdfa/config.json : Permission denied (os error 13)

UriError:

ERROR neard::cli: Failed to initialize configs: Failed to download the config file from http://localhost:8002:3:a:b/config.json: Invalid URI: invalid authority

### DIFF
--- a/nearcore/src/config.rs
+++ b/nearcore/src/config.rs
@@ -1,7 +1,7 @@
 use std::fs;
 use std::fs::File;
 use std::io::{Read, Write};
-use std::path::Path;
+use std::path::{Path, PathBuf};
 use std::sync::Arc;
 use std::time::Duration;
 
@@ -1139,16 +1139,16 @@ pub fn get_config_url(chain_id: &str) -> String {
 pub enum FileDownloadError {
     #[error("{0}")]
     HttpError(hyper::Error),
-    #[error("Failed to open temporary file: {0}")]
-    OpenError(std::io::Error),
-    #[error("Failed to write to temporary file at {0}: {1}")]
-    WriteError(String, std::io::Error),
-    #[error("Failed to rename temporary file {0} to {1} : {2}")]
-    RenameError(String, String, std::io::Error),
+    #[error("Failed to open temporary file")]
+    OpenError(#[source] std::io::Error),
+    #[error("Failed to write to temporary file at {0:?}")]
+    WriteError(PathBuf, #[source] std::io::Error),
+    #[error("Failed to rename temporary file {0:?} to {1:?}")]
+    RenameError(PathBuf, PathBuf, #[source] std::io::Error),
     #[error("Invalid URI")]
     UriError(#[from] hyper::http::uri::InvalidUri),
-    #[error("Failed to remove temporary file: {0}. Download previously failed: {1}")]
-    RemoveTemporaryFileError(std::io::Error, Box<FileDownloadError>),
+    #[error("Failed to remove temporary file: {0}. Download previously failed")]
+    RemoveTemporaryFileError(std::io::Error, #[source] Box<FileDownloadError>),
 }
 
 /// Downloads resource at given `uri` and saves it to `file`.  On failure,
@@ -1165,7 +1165,7 @@ async fn download_file_impl(
         let next_chunk = next_chunk_result.map_err(FileDownloadError::HttpError)?;
         file.write_all(next_chunk.as_ref())
             .await
-            .map_err(|e| FileDownloadError::WriteError(path.to_string_lossy().into_owned(), e))?;
+            .map_err(|e| FileDownloadError::WriteError(path.to_path_buf(), e))?;
     }
     Ok(())
 }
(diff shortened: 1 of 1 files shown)


nearcore/src/config.rs
```diff
@@ -1137,65 +1137,73 @@ pub fn get_config_url(chain_id: &str) -> String {
 
 #[derive(thiserror::Error, Debug)]
 pub enum FileDownloadError {
-    #[error("Failed to download the file: {0}")]
-    HttpError(#[from] hyper::Error),
-    #[error("Failed to open file: {0}")]
+    #[error("{0}")]
```
nagisa, line 1140: This should be `#[error(transparent)]` I think.

---

Pre-existing, but in general `thiserror` messages should _not_ format in the cause error's description into its own. Instead `#[source]` mechanism should be utilized to specify the causation chain. This is how the ecosystem expects the errors to be, making it least likely to get duplicated output like you had described in the PR description. I have, in the past, also written [a block post on error handling](https://kazlauskas.me/entries/errors). While this is not directly applicable to this discussion, it kind of demonstrates what a conventional thiserror use looks like.

nearcore/src/config.rs
```diff
@@ -1137,65 +1137,73 @@ pub fn get_config_url(chain_id: &str) -> String {
...
     let client = hyper::Client::builder().build::<_, hyper::Body>(https_connector);
-    let mut resp = client.get(uri).await?;
+    let mut resp = client.get(uri).await.map_err(FileDownloadError::HttpError)?;
     while let Some(next_chunk_result) = resp.data().await {
-        let next_chunk = next_chunk_result?;
-        file.write_all(next_chunk.as_ref()).await.map_err(FileDownloadError::WriteError)?;
+        let next_chunk = next_chunk_result.map_err(FileDownloadError::HttpError)?;
+        file.write_all(next_chunk.as_ref())
+            .await
+            .map_err(|e| FileDownloadError::WriteError(path.to_string_lossy().into_owned(), e))?;
```
nagisa, line 1168: I would argue that the error should store a `PathBuf` and then the error message should “just” use the `{:?}` formatter to print out the path. There are a couple of benefits to this, most important of which is that non-utf8 paths are still understandable in the output.

nearcore/src/config.rs
marcelo-gonzalez: Ah yeah, I actually tried this originally but it gives the same repition of info. Looks like something wrong with hyper itself?

```rust
println!("orig {}", &e);
let mut source = e.source();

while let Some(err) = source {
  println!("source {}", err);
  source = err.source();
}
```

this prints out:

```
orig error trying to connect: tcp connect error: Connection refused (os error 111)
source tcp connect error: Connection refused (os error 111)
source Connection refused (os error 111)
```

So just using "{0}" is the best workaround for that I can think of here. But as for the others, using #[source] instead of formatting it explicitly seems way better yeah, I like the blog post! updated the other ones to use #[source] now, and checked that the output is still good. (Last one is a little tricky since there are really two sources, but we can just pick the first one I guess)

bowenwang1996 (approved):

nearcore/src/config.rs
nagisa: I see. I think it might be a good idea to report this to hyper.

nagisa: https://github.com/hyperium/hyper/issues/2732

nagisa (approved):

PR #6025, opened 2022-01-07
test-uitls: fix test output not being captured by `cargo test`

mina86:
Rust’s `libtest` has a… limitation where it doesn’t capture data
written directly to `std::io::stdout` or `std::io::stderr` (see
https://github.com/rust-lang/rust/issues/90785).  Since we are test
logger to use the latter, none of the log messages are captured by the
test harness.

The solution is to use `TestWriter` as the writer instead which
cooperates `libtest` better.


pmnoxx:
Duplicate of https://github.com/near/nearcore/pull/6023

PR #6026, opened 2022-01-07
re: Remove dependency on `near-network` from genesis tools

pmnoxx:
There is need to `genesis-populate` depend on `near-network`. That only decreases build time.


pmnoxx:
```
@nikurt: genesis-tools/
```
@nikurt Can you take a look?

nikurt (approved):

PR #6031, opened 2022-01-09
integration-tests: fix some expensive tests marking

mina86:
Firstly, the proper way to denote a test as expensive is to use
`#[cfg]` directive before its definition rather than adding a return
inside of the function.  The latter will cause the test to show up in
list of tests and always pass giving people false sense that running
the test without `expensive_tests` feature actually does something.

After fixing `sync_state_nodes_multishard` also add it to nightly run.

Secondly, the `check_nightly.py` script is rather crude and the order
in which `#[cfg]` and `#[test]` attributes are applied do matter.
With `#[test]` being first, the script failed to recognise
`test_catchup` as an expensive test.

Thirdly, the `check_nightly.py` script also doesn’t understand
`#[cfg]` matching multiple features.  Because of that it does not
detect `test_highload` as an expensive test either.  Since we’re not
using `regression_tests` feature for anything simply remove it.

### DIFF
--- a/integration-tests/Cargo.toml
+++ b/integration-tests/Cargo.toml
@@ -58,7 +58,6 @@ portpicker = "0.1.1"
 
 [features]
 performance_stats = ["nearcore/performance_stats", "near-network/performance_stats"]
-regression_tests = []
 expensive_tests = []
 test_features = ["nearcore/test_features"]
 protocol_feature_alt_bn128 = [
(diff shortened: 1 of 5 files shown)


mina86:
https://nayduck.near.org/#/run/2349

matklad:
Question: why do we need to look for `#[cfg]` in python at all? Can we just run all tests on nayduck? Presumably, non-expensive tests shouldn't increase runtime that much.

matklad (approved):
That being said, this is a clear improvement by itself

mina86:
> Question: why do we need to look for #[cfg] in python at all? Can we just run all tests on nayduck? Presumably, non-expensive tests shouldn't increase runtime that much.

We run each test separately on NayDuck and thus we need to enumerate which tests should be run.

Issue #6008, opened 2022-01-06
Add --log-level as a command line option OR config.json file option


mintiayields:
I have seen this discussed as a RUST_LOG environment variable usage, however it will have a lot more sense as a config file option or a command line option. I sometimes get a hundred warnings a second written to my neard logs and there is not much I can do to prevent this. 
I suggest that this is put in config.json and has multiple options, for instance: "log_level = info, error"

mina86:
I’m honestly not convinced this needs to be a flag or config option. The feature is available and adding a flag would just introduce unnecessary code and complicate interaction between RUST_LOG and said flag. Adding config option would make it even more complicated since config file is read at later stage during startup.

mintiayields:
Exporting RUST_LOG variable will not produce any effect in my case. When printenv clearly shows that RUST_LOG is set to error, the log file continues to be filled with numerous warnings and info messages. I wouldn't mind the latter as they show valuable info, but the warnings are mostly junk and inflate the log file like crazy.

mina86:
No, it’s good. Currently waiting for review.

PR #6036, opened 2022-01-10
cli: better respect RUST_LOG

mina86:
At the moment, when configuring what data is logged we first
initialise EnvFilter with a predefined defaults and only later if
RUST_LOG environment variable is set add that on top of the.  This
means that setting RUST_LOG to `warn` doesn’t actually filter messages
to include warnings and up only.

Change the code such that if RUST_LOG environment variable is set
(even if empty), the value is taken as is without including the
defaults.  If the variable is not present, the defaults are used as
before.

Fixes: https://github.com/near/nearcore/issues/6008


neard/src/cli.rs
```diff
@@ -387,10 +387,14 @@ impl LocalnetCmd {
 }
 
 fn init_logging(verbose: Option<&str>) {
-    let mut env_filter = EnvFilter::new(
-        "tokio_reactor=info,near=info,stats=info,telemetry=info,delay_detector=info,\
-         near-performance-metrics=info,near-rust-allocator-proxy=info",
-    );
+    const DEFALUT_RUST_LOG: &'static str =
```
mm-near, line 390: typo in DEFAULT

mm-near (approved):

Issue #6010, opened 2022-01-06
init and unsafe_reset_all subcommands enhancement


mintiayields:
When neard is run with "init" parameter it will overwrite any existing files without warning, including node_key.json and validator_key.json. Yes, validators should backup their keys, data, etc. But adding a simple rename function to the code once it finds the existing file would be a safety feature. The rename pattern may be "original_filename_<cur-timestamp>.original_extension.bak
For unsafe_reset_all it will make sense not to truncate the whole dir, but to delete only files that neard knows and uses. Wiping the whole directory came as a surprise to me when I first faced this.

mina86:
Dunno, `unsafe_reset_all` is explicitly market as unsafe command so I’m not convinced we need extra safety features. The documentation might be updated though to explicitly state that the entire directory is deleted.

mintiayields:
unsafe_reset_all - I agree. Yet, this was a surprise to get the whole directory wiped. A warning or a confirmation request would be nice though - like it is accepted in linux systems.

mina86:
Again, I dunno. `rm -r directory` will delete the directory without any warning or confirmation.

Were you surprised because the documentation didn’t mention that the whole directory is deleted? I.e. would #6038 solve the confusion?

mintiayields:
Adding this behavior description in documentation is good. What do you think about adding an explicit notice at run and a request to confirm with "yes/no"? If the latter is not possible,  #6038 will be the only option.

mina86:
It’s not that it’s not possible, but it’s more code I’m not sure is necessary plus it’s not clear what to do if command is snot run on the terminal. Should it not delete the directory then? That might break existing scripts.

mintiayields:
I am just not aware of the cases when this kind of command can be run not from the terminal. If you do not think my suggestion is feasible, feel free to close it :) I learned my lesson. The question is what other node runners' experiences will be.

mina86:
> I am just not aware of the cases when this kind of command can be run not from the terminal.

Tests, startup scripts.

PR #6037, opened 2022-01-10
neard: don’t overwrite validator and node keys when initialising chain

mina86:
Currently, the ‘init’ subcommand will happily overwrite node and
validator key files if they are present in the home directory.  Change
the code such that if those files exist, rather than overwriting them,
the subcommand will only verify that they are correct.

Fixes: https://github.com/near/nearcore/issues/6010


nearcore/src/config.rs
```diff
@@ -790,13 +790,58 @@ fn add_account_with_key(
...
-    signer
-        .write_to_file(path)
-        .with_context(|| format!("Error writing key file to {}", path.display()))
+enum KeyToGenerate<'a> {
+    Node,
+    Validator(Option<AccountId>),
+    TestValidator(AccountId, Option<&'a str>),
+}
+
+fn generate_key(
```
mm-near, line 799: maybe we should call it 'load_or_generate_key'? (or  generate_if_missing ?)

nearcore/src/config.rs
```diff
@@ -948,26 +990,19 @@ pub fn init_configs(
             })?;
 
             let account_id = account_id.unwrap_or_else(|| "test.near".parse().unwrap());
+            let signer = generate_key(
```
mm-near, line 993: I'd mention here, that we're overwriting the key that exists

nearcore/src/config.rs
```diff
@@ -790,13 +790,58 @@ fn add_account_with_key(
...
-    info!(target: "near", "Use key {} for {} to stake.", signer.public_key(), account_id);
-    signer
-        .write_to_file(path)
-        .with_context(|| format!("Error writing key file to {}", path.display()))
+enum KeyToGenerate<'a> {
+    Node,
+    Validator(Option<AccountId>),
+    TestValidator(AccountId, Option<&'a str>),
+}
+
```
mm-near, line 798: please add tests for this function

mm-near (changes requested):

nearcore/src/config.rs
```diff
@@ -790,13 +790,58 @@ fn add_account_with_key(
...
+fn generate_key(
+    key: KeyToGenerate<'_>,
+    home_dir: &Path,
+    filename: &str,
+    overwrite: bool,
+) -> anyhow::Result<Option<InMemorySigner>> {
+    let path = home_dir.join(filename);
```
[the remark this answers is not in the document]
mina86: Either works, but since joining happens in all places that call this function I figured it’s easier if it’s pushed to the function so it doesn’t have to be repeated over and over.

mina86:
@mm-near, PTAL

matklad (approved):

mina86:
@mm-near, friendly ping.

nearcore/src/config.rs
```diff
@@ -792,13 +792,116 @@ fn add_account_with_key(
...
-        .write_to_file(path)
-        .with_context(|| format!("Error writing key file to {}", path.display()))
+/// Generates or loads a signer key from given file.
+///
+/// If the file already exists, loads the file (panicking if the file is
+/// invalid), checks that account id in the file matches `account_id` if it’s
+/// given and returns the key.  `test_seed` is ignored in this case.
+///
+/// If the file does not exist and `account_id` is not `None`, generates a new
+/// key, saves it in the file and returns it.  If `test_seed` is note `None`,
```
mm-near, line 802: s/note/not

mm-near (approved):

Issue #4370, opened 2021-06-15
chain id parsed from nodelib scripts doesn't handle valid argument syntax


austinabell:
**Describe the bug**

Currently, the `scripts/nodelib.py` script
https://github.com/near/nearcore/blob/5d12069ca305cc781c77d63e3c932e1c11f06ac4/scripts/nodelib.py#L52

parses the chain id assuming this syntax:
```bash
neard init --chain-id=network
```

But the following is also valid and ignored:
```bash
neard init --chain-id network
```

Added a TODO in #4356 

**To Reproduce**
Steps to reproduce the behavior.

**Expected behavior**

Both valid argument syntax should be supported

**Screenshots**

**Version (please complete the following information):**
- nearcore commit/branch
- rust version (if local)
- docker (if using docker)
- mainnet/testnet/betanet/local

**Additional context**

PR #6039, opened 2022-01-10
nodelib: better support ‘--chain-id’ flag

mina86:
Change get_chain_id_from_flags function to support ‘--chain-id
<chain>’ flag syntax (without the equals sign).

Fixes: https://github.com/near/nearcore/issues/4370#event-5866878005


scripts/nodelib.py
```diff
@@ -48,15 +48,16 @@ def nodocker_init(home_dir, is_release, init_flags):
...
-    # TODO this doesn't handle the `--chain-id my-id` syntax which should also be valid
-    chain_id_flags = [flag for flag in flags if flag.startswith('--chain-id=')]
-    if len(chain_id_flags) == 1:
-        return chain_id_flags[0][len('--chain-id='):]
-    return ''
+    """Retrieve requested chain id from the flags."""
+    chain_id = None
+    flags_iter = iter(flags)
+    for flag in flags_iter:
+        if flag.startswith('--chain-id='):
```
nikurt, line 56: I'd like to avoid the magic number 11. For example, it's not obvious whether 11 includes the equal sign or not.

chefsale (approved):

PR #6041, opened 2022-01-11
Revert the changes to use low level wasmer2 implementation

nagisa:
It broke betanet. The breakage is related to the memory sizes, wherein we trigger the first assert in [this section of code](https://github.com/near/wasmer/blob/68da83cfb903ae14093b84cc041048cd0f21e48b/lib/engine/src/resolver.rs#L242-L275)

Looks like there is a contract that is attempting to import 65536 pages worth of memory whereas our limits config specifies a 2048 maximum. Presumably the preparation code should standardize these sizes, so not sure how the larger numbers end up reaching wasmer.

Reverting to give us time to investigate.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -439,7 +439,7 @@ dependencies = [
  "cfg-if 1.0.0",
  "libc",
  "miniz_oxide",
- "object",
+ "object 0.23.0",
  "rustc-demangle",
 ]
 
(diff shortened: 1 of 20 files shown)


jakmeier (approved):
Yeah definitely makes sense to revert for now

matklad (approved):

PR #6043, opened 2022-01-11
re: restore the low level wasmer2 refactor

nagisa:
This reverts commit 353b4b4aa3c5955460e89d8ab7c654d61d29201d.

### DIFF
--- a/runtime/near-vm-runner/src/tests/cache.rs
+++ b/runtime/near-vm-runner/src/tests/cache.rs
@@ -71,14 +71,12 @@ fn make_cached_contract_call_vm(
     let promise_results = vec![];
     context.prepaid_gas = prepaid_gas;
     let code = ContractCode::new(code.to_vec(), None);
-    let runtime = vm_kind.runtime().expect("runtime has not been compiled");
-
+    let runtime = vm_kind.runtime(config).expect("runtime has not been compiled");
     runtime.run(
         &code,
         method_name,
         &mut fake_external,
         context.clone(),
-        &config,
         &fees,
         &promise_results,
         LATEST_PROTOCOL_VERSION,
--- a/runtime/near-vm-runner/src/tests/cache.rs
+++ b/runtime/near-vm-runner/src/tests/cache.rs
@@ -9,6 +9,7 @@ use near_primitives::types::CompiledContractCache;
 use near_vm_errors::VMError;
 use near_vm_logic::mocks::mock_external::MockedExternal;
 use near_vm_logic::{VMConfig, VMOutcome};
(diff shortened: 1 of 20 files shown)


nagisa:
Note, this is a draft, pending investigation into what exactly causes us to attempt an import of a 65536 page memory region.

nagisa:
I believe I figured out the problem that we had. It isme forgetting to bump the `WASMER2_CONFIG` constant (again) so that potential changes to the artifacts are taken into account for caching purposes.

The change in this particular instance was in the `Tunables` implementation which now returns correctly bounded `MemoryStyle`. Previously even if we allocated memories with up-to 2048 pages, the `MemoryStyle` would still say something along the lines of “this is up to 0xFFFF pages” or some such.

Now that these are consistent, across the board, loading an artifact generated with an old version of implementation would give us a module with a `MemoryStyle` that's… vzery different.

runtime/near-vm-runner/src/tests/cache.rs
```diff
@@ -71,21 +75,41 @@ fn make_cached_contract_call_vm(
...
         &promise_results,
         LATEST_PROTOCOL_VERSION,
         Some(cache),
     )
 }
 
+#[test]
+fn test_wasmer2_artifact_output_stability() {
+    // If this test has failed, you want to adjust the necessary constants so that `cache::vm_hash`
+    // changes (and the only then the hashes here).
```
matklad, line 94: Something's up with grammar here? "the only then the hashes"? love the alliteration though

runtime/near-vm-runner/src/tests/cache.rs
```diff
@@ -71,21 +75,41 @@ fn make_cached_contract_call_vm(
...
+    let prepared_code = prepare::prepare_contract(code, &config).unwrap();
+    let mut features = CpuFeature::set();
+    features.insert(CpuFeature::AVX);
+    let triple = "x86_64-unknown-linux-gnu".parse().unwrap();
+    let target = Target::new(triple, features);
+    let vm = Wasmer2VM::new_for_target(config, target);
+    let artifact = vm.compile_uncached(&prepared_code).unwrap();
+    let serialized = artifact.artifact().serialize().unwrap();
+    serialized.hash(&mut hasher);
+    assert_eq!(hasher.finish(), 7017891230624373240, "WASMER2_CONFIG needs version change");
```
matklad, line 110: :+1: that's a very good test, thanks for realizing we can do this. It did happen in the past that we've *almost* forgot to bust the caches.

matklad (approved):

Issue #5701, opened 2021-12-07
Check that we can verify block hash from rpc result


bowenwang1996:
For context, https://stackoverflow.com/questions/70155175/calculate-and-verify-block-hash-from-light-client-block-in-near-protocol. 

We should check that we indeed can verify block hash from the result returned by the `block` rpc endpoint and add a test for it.

Rishabhraghwendra18:
Hi @bowenwang1996 . I am new to near . I will like to work on this issue . May you specify me what I have to do in order to resolve this issue ?
Thanks

bowenwang1996:
@Rishabhraghwendra18 please read the stack overflow link first and let me know if something is still not clear.

Rishabhraghwendra18:
@bowenwang1996 I read the issue on stack overflow and the link in the issue too .  I don't get what is `epoch` ?
I am new to blockchain technology and Near is the first platform I am learning (I don't know ethereum) . But I am able to get some gist of the problem that user wants to verify a block hash suing LightClient but it returned him a wrong block hash . So he is asking for a way to verify & calculate the blockhash in Near  .

Rishabhraghwendra18:
Hi @bowenwang1996 . I think this issue is currently hard for my level of understanding . I am really sad that I can't contribute to the issue , sorry . I will hide my comments to keep the issue clean .

PR #6044, opened 2022-01-12
pytest: add test for computing block hash from JSON RPC response

mina86:
Fixes: https://github.com/near/nearcore/issues/5701


mina86:
https://nayduck.near.org/#/run/2354

pytest/tests/sanity/rpc_hash.py
```diff
@@ -0,0 +1,199 @@
...
+        try:
+            nodes = cluster.start_cluster(2, 0, 1, None, genesis_overrides, {})
+            node = nodes[0]
+            block_id = utils.wait_for_blocks(node, target=1)
+            header = node.get_block(block_id.hash)['result']['header']
+        finally:
+            for node in nodes:
+                node.cleanup()
+
+        #logger.info('header = %s', json.dumps(header, indent=True))
```
mm-near, line 156: remove commented out code

pytest/tests/sanity/rpc_hash.py
```diff
@@ -0,0 +1,199 @@
...
+import messages
+import messages.crypto
+import messages.block
+
+
+def serialize(msg: typing.Any) -> bytes:
+    return serializer.BinarySerializer(messages.schema).serialize(msg)
+
+
+def compute_block_hash(header: typing.Dict[str, typing.Any],
```
mm-near, line 29: if we have this function already - - how about we extend this test a little?

we could create couple 'static' json files - load them and check that the hashes computed by this are correct.
(and those json files could come from 'real' blocks on the blockchain)

(yes, this means that we'd have a 'test' for a 'test' method ;-) - but long term, we might want to move this function into some python utils.

 (and there are a bunch of TODOs below - which we'll be able to cover with these 'static' tests)

pytest/tests/sanity/rpc_hash.py
```diff
@@ -0,0 +1,199 @@
...
+        #logger.info('header = %s', json.dumps(header, indent=True))
+        self.assertEqual((block_id.height, block_id.hash),
+                         (header['height'], header['hash']),
+                         (block_id, header))
+
+        got = compute_block_hash(header, msg_version)
+        self.assertEqual(header['hash'], got)
+
+    def test_block_hash_v1(self):
+        """Starts a cluster, fetches the first block and computes its hash.
```
mm-near, line 165: could we test a non-first block too?

mm-near (changes requested):

pytest/tests/sanity/rpc_hash.py
mm-near: do you plan to also use/test the code from: https://github.com/near/nearcore/blob/master/pytest/lib/lightclient.py#L33 ?

pytest/tests/sanity/rpc_hash.py
mina86: It actually doesn’t matter much.  In practice the tests takes ~4th block.

pytest/tests/sanity/rpc_hash.py
mina86: Regarding light client, I am planning to change it so that it uses `messages.schema` rather than rolling the schema on its own, but otherwise I don’t have any plans to use the code here.

As for TODOs and taking header from mainnet, blocks on mainnet don’t have those fields set either.  Challenges aren’t implemented at

mina86:
@mm-near, PTAL

mm-near (approved):

PR #6046, opened 2022-01-12
feat: compute rocksdb metrics

Longarithm:
Add feature to compute rocksdb metrics to grafana per each column:
* column size
* num entries
* total size of keys
* total size of values

The plan is to enable 1-day periodical computation on some selected rpc and archive nodes and collect dynamics within a week.
I don't like the idea of enabling it everywhere, because `sst_dump` is kind of custom script which is even not available on Ubuntu 18, and computation can be heavy (~40m for an archival node).

## Test plan

* Launched neard node, observed metrics in the list of all metrics.


chain/client/src/client_actor.rs
```diff
@@ -1508,6 +1512,38 @@ impl ClientActor {
...
     }
+
+    /// Periodically compute RocksDB stats.
+    fn compute_rocksdb_stats(&self, ctx: &mut Context<Self>) {
+        if let Some(period) = self.client.config.compute_rocksdb_stats_period {
+            near_performance_metrics::actix::run_later(ctx, period, move |act, ctx| {
+                info!(target: "neard", "Computing rocksdb stats...");
+                let rocksdb_stats = get_rocksdb_stats(
+                    self.client.chain.store().store().get_rocksdb().unwrap().path(),
+                );
```
matklad, line 1523: Hm, this looks like a rather scary thing to do...

My understanding is what happens here is that once in a while we are going to run external process to analyse database, while the database is still in use. 

The biggest immediate thing that'll go wrong is that this processing would completely block client's actor loop -- no other stuff will be happening while the sst dump is in progress. 

More generally, it doesn't seem like `sst_dump` is a thing which is intended to be run in production like this, so I think this can add a fair amount of unknown unknowns. 

My gut feel is that these stats should be computed and send by a different process altogether -- rather than running the thing from within a live node, setup some cron job to run a script periodically (the script *can* be `neard send-db-telemety `~/.neard`. It's OK to keep the code for this in the same binary).

I see however how  this can be operationally more convenient if we just do this from within the process itself... As a litmus test, the way we intend to use this code, would it be ok if it outright panicked? If we are OK with that, than, yeah, probably fine to keep this. Although, even in that case, rather than spawning this from-within a clinet, I'd spawn a whole dedicated thread for this from `main`, like we do here: 

https://github.com/near/nearcore/blob/93efd262fff035d9825898bd9e906fb9efcf19af/neard/src/main.rs#L48

core/store/Cargo.toml
```diff
@@ -9,6 +9,7 @@ rust-version = "1.56.0"
 edition = "2021"
 
 [dependencies]
+anyhow = "1"
```
matklad, line 12: `anyhow` is meant for "sloppy" error management which you do at the edges of the system (CLI). `store` feels like the place where we care about pedantically correct error management, so it feels weird that it has to depend on anyhow.

mm-near:
I've checked rocksdb, and it seems that their SST files are immutable once created. This means that this stats job could be running as a completely separate binary (even without a 'write' access to those files). 

Moreover, it could only read the files that were created since last run, therefore drastically lowering the cost of running.

chain/client/src/client_actor.rs
mm-near: +1 to this comment. We should try running those stats from the outside.

mm-near (changes requested):

chain/client/src/client_actor.rs
```diff
@@ -1508,6 +1512,38 @@ impl ClientActor {
             },
         );
     }
+
+    /// Periodically compute RocksDB stats.
+    fn compute_rocksdb_stats(&self, ctx: &mut Context<Self>) {
+        if let Some(period) = self.client.config.compute_rocksdb_stats_period {
```
pmnoxx, line 1518: It may be a good idea to rename this to add `_trigger_` suffix. This will make easier to distinguish trigger from not trigger functions..

akhi3030:
I suspect this PR is no longer being actively worked on.  Feel free to reopen when you are ready to start working on it again.

PR #6048, opened 2022-01-12
state-viewer: Get a receipt from the DB

nikurt:


nagisa (approved):
I'm not super qualified to review this, but what I see here looks good to me.

PR #6049, opened 2022-01-12
state-viewer: Get chunks and partial chunks from the DB

nikurt:


nagisa (approved):

test-utils/state-viewer/src/cli.rs
```diff
@@ -324,3 +330,29 @@ impl ReceiptsCmd {
...
+
+#[derive(Clap)]
+pub struct ChunksCmd {
+    #[clap(long)]
+    chunk_hash: String,
+}
+
+impl ChunksCmd {
+    pub fn run(self, near_config: NearConfig, store: Arc<Store>) {
+        let chunk_hash = ChunkHash::from(CryptoHash::from_str(&self.chunk_hash).unwrap());
```
nagisa, line 342: If this was a end-user facing CLI tool, I would probably try to somehow preesent a proper error message here since `chunk_hash` comes as an input from the user, but this is okay for an internal tool as well.

PR #6050, opened 2022-01-12
Fix comments and some implementation style in chunks 

mzhangmzz:
This PR does not include any change to the algorithm or behavior, it is to address comments that I didn't address in https://github.com/near/nearcore/pull/5747


mzhangmzz:
@matklad I addressed your comments in https://github.com/near/nearcore/pull/5747 here, since 5747 was closed

chain/chunks/src/chunk_cache.rs
```diff
@@ -143,23 +141,21 @@ impl EncodedChunksCache {
...
     // Use `mark_entry_complete` instead for outside calls
     fn remove_incomplete_chunk(&mut self, prev_block_hash: CryptoHash, chunk_hash: &ChunkHash) {
-        if let Some(mut chunks) = self.incomplete_chunks.remove(&prev_block_hash) {
-            chunks.remove(chunk_hash);
-            if !chunks.is_empty() {
-                self.incomplete_chunks.insert(prev_block_hash, chunks);
+        if let Occupied(mut entry) = self.incomplete_chunks.entry(prev_block_hash) {
+            entry.get_mut().remove(chunk_hash);
+            if !entry.get().is_empty() {
+                entry.remove();
```
matklad, line 158: I think this is buggy: `!` is not needed. Does it fail any tests?

chain/chunks/src/lib.rs
```diff
@@ -1,54 +1,4 @@
...
-    AccountIdOrPeerTrackingShard, PartialEncodedChunkForwardMsg, PartialEncodedChunkRequestMsg,
-    PartialEncodedChunkResponseMsg,
-};
-use near_primitives::epoch_manager::RngSeed;
-use near_primitives::shard_layout::{account_id_to_shard_id, ShardLayout};
-use rand::Rng;
-
-// This file implements ShardManager, which handles chunks requesting and processing.
+//! This module implements ShardManager, which handles chunks requesting and processing.
 // Since blocks only contain chunk headers, full chunks must be communicated separately.
```
matklad, line 2: You need to add `!` to every line to make this a doc comment.

matklad (changes requested):

chain/chunks/src/chunk_cache.rs
mzhangmzz: Oh good catch, apparently no existing test is failed. I'll add a test

chain/chunks/src/chunk_cache.rs
```diff
@@ -294,28 +292,56 @@ mod tests {
...
+        let header1 = create_chunk_header(1, 1);
+        cache.try_insert(&header0);
+        cache.merge_in_partial_encoded_chunk(&PartialEncodedChunkV2 {
+            header: header1.clone(),
+            parts: vec![],
+            receipts: vec![],
+        });
+        assert_eq!(
+            cache.get_incomplete_chunks(&CryptoHash::default()).unwrap(),
+            &vec![header0.chunk_hash(), header1.chunk_hash()].into_iter().collect::<HashSet<_>>()
```
matklad, line 329:
```suggestion
            &HashSet::from([header0.chunk_hash(), header1.chunk_hash()])
```

chain/chunks/src/chunk_cache.rs
```diff
@@ -143,39 +141,41 @@ impl EncodedChunksCache {
...
-    pub fn try_insert(&mut self, chunk_hash: ChunkHash, header: &ShardChunkHeader) {
-        self.get_or_insert_from_header(chunk_hash, header);
+    pub fn try_insert(&mut self, header: &ShardChunkHeader) {
+        self.get_or_insert_from_header(header);
     }
 
     // Create an empty entry from the header and insert it if there is no entry for the chunk already
     // Return a mutable reference to the entry
-    pub fn get_or_insert_from_header(
+    fn get_or_insert_from_header(
```
matklad, line 174: :+1:

matklad (approved):

chain/chunks/src/chunk_cache.rs
```diff
@@ -143,39 +141,41 @@ impl EncodedChunksCache {
...
         &mut self,
-        chunk_hash: ChunkHash,
         chunk_header: &ShardChunkHeader,
     ) -> &mut EncodedChunksCacheEntry {
+        let chunk_hash = chunk_header.chunk_hash();
         if !self.encoded_chunks.contains_key(&chunk_hash) {
```
pmnoxx, line 179: This should be inside the if.
```rust
        self.encoded_chunks
            .entry(chunk_hash)
            .or_insert_with(|| EncodedChunksCacheEntry::from_chunk_header(chunk_header.clone()))
```

Then it can be written more simply as.
```rust
        self.encoded_chunks
            .insert(chunk_hash, EncodedChunksCacheEntry::from_chunk_header(chunk_header.clone()))
```

mzhangmzz: chunk_hash is used in line 179

`.entry.or_insert_with` saves a call to construct the EncodedChunksCacheEntry object if the entry already exists.

mm-near (approved):

PR #6051, opened 2022-01-12
fix: use pwasm-utils 0.18 for more correct stack accounting

matklad:
This forward-ports the change from 1.23.1

We upgrade our version of pwasm-utils -- the old one severely
undercounted stack usage in some cases. As this is a protocol change, we
keep the old implementation around

Test Plan
=========

* added test for stack overflow behavior for the bad case
* added test for protocol upgrade related to stack overflow
* fuzzed new pwasm utils implementation locally, confirmed that it only
  differs for calls which stack overflow

### DIFF
--- a/core/primitives-core/src/config.rs
+++ b/core/primitives-core/src/config.rs
@@ -31,6 +31,8 @@ pub struct VMLimitConfig {
     /// See <https://wiki.parity.io/WebAssembly-StackHeight> to find out
     /// how the stack frame cost is calculated.
     pub max_stack_height: u32,
+    /// Whether a legacy version of stack limiting should be used, see
+    /// [`StackLimiterVersion`].
     #[serde(default = "StackLimiterVersion::v0")]
     pub stack_limiter_version: StackLimiterVersion,
 
@@ -84,9 +86,18 @@ pub struct VMLimitConfig {
     pub max_functions_number_per_contract: Option<u64>,
 }
 
+/// Our original code for limiting WASM stack was buggy. We fixed that, but we
+/// still have to use old (`V0`) limiter for old protocol versions.
+///
+/// This struct here exists to enforce that the value in the config is either
+/// `0` or `1`. We could have used a `bool` instead, but there's a chance that
+/// our current impl isn't perfect either and would need further tweaks in the
+/// future.
 #[derive(Debug, Clone, Copy, Hash, PartialEq, Eq)]
 pub enum StackLimiterVersion {
+    /// Old, buggy version, don't use it unless specifically to support old protocol version.
     V0,
+    /// What we use in today's protocol.
     V1,
 }
 
--- a/core/primitives-core/src/config.rs
+++ b/core/primitives-core/src/config.rs
@@ -31,6 +31,8 @@ pub struct VMLimitConfig {
     /// See <https://wiki.parity.io/WebAssembly-StackHeight> to find out
     /// how the stack frame cost is calculated.
     pub max_stack_height: u32,
+    #[serde(default = "StackLimiterVersion::v0")]
+    pub stack_limiter_version: StackLimiterVersion,
 
     /// The initial number of memory pages.
     /// NOTE: It's not a limiter itself, but it's a value we use for initial_memory_pages.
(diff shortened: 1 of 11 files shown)


core/primitives/src/version.rs
```diff
@@ -128,6 +128,9 @@ pub enum ProtocolFeature {
     /// Make block producers produce chunks for the same block they would later produce to avoid
     /// network delays
     SynchronizeBlockChunkProduction,
+    /// Change the algorithm to count WASM stack usage to avoid undercounting in
+    /// some cases.
+    CorrectStackLimit,
```
nagisa, line 133: if we ever need to correct the algorithm again, it'll be hard to come up with a name for it ^^

nagisa (approved):

core/primitives/src/version.rs
matklad: This name we can retroactively change in the future, as it is not serialized anywhere, so tried not to over-thing this )

core/primitives/src/version.rs
```diff
@@ -181,7 +184,8 @@ impl ProtocolFeature {
             | ProtocolFeature::LimitContractFunctionsNumber
             | ProtocolFeature::BlockHeaderV3
             | ProtocolFeature::AliasValidatorSelectionAlgorithm => 49,
-            ProtocolFeature::SynchronizeBlockChunkProduction => 50,
+            ProtocolFeature::SynchronizeBlockChunkProduction
+            | ProtocolFeature::CorrectStackLimit => 50,
```
mm-near, line 188: shouldn't we use a new protocol version for this? (or was it already deployed somehow?)

runtime/near-vm-runner/src/prepare.rs
```diff
@@ -172,26 +197,175 @@ impl<'a> ContractModule<'a> {
...
+                        imported_mem_type = Some(memory_type);
+                        continue;
+                    }
+                    _ => continue,
+                };
+
+                let Type::Function(ref _func_ty) =
+                    types.get(*type_idx as usize).ok_or(PrepareError::Instantiate)?;
+
+                // TODO: Function type check with Env
```
mm-near, line 327: optional: I'm not a fan of 'orphaned' TODOs ;-) we should either have a github/jira issue that is tracking it (if this is a lot more complex fix), or author name + additional info on why this is complex enough that we're not fixing it in this PR.

runtime/near-vm-runner/src/prepare.rs
```diff
@@ -172,26 +197,175 @@ impl<'a> ContractModule<'a> {
...
+                /*
+
+                let ext_func = env
```
mm-near, line 330: do you need this code?

mm-near (approved):

core/primitives/src/version.rs
matklad: Yeah, that was deployed as a part of 1.23.1: https://github.com/near/nearcore-ghsa-62f6-rgx9-xcw2/pull/1

core/primitives-core/src/config.rs
```diff
@@ -82,6 +84,53 @@ pub struct VMLimitConfig {
     pub max_functions_number_per_contract: Option<u64>,
 }
 
+#[derive(Debug, Clone, Copy, Hash, PartialEq, Eq)]
+pub enum StackLimiterVersion {
+    V0,
+    V1,
```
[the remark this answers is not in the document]
matklad: Yeah, good call, added some comments.

runtime/near-vm-runner/src/prepare.rs
```diff
@@ -23,18 +21,48 @@ pub(crate) const WASM_FEATURES: wasmparser::WasmFeatures = wasmparser::WasmFeatu
...
+    wasmparser::Validator::new()
+        .wasm_features(WASM_FEATURES)
+        .validate_all(original_code)
+        .map_err(|_| PrepareError::Deserialization)?;
+
+    match config.limit_config.stack_limiter_version {
+        // Support for old protocol versions, where we incorrectly didn't
+        // account for many locals of the same type.
+        //
+        // See `test_stack_instrumentation_protocol_upgrade` test.
```
matklad, line 45: FWIW, there's a cool trick to make such links to tests checkable (ie, the test fails if it doesn't actually touch the relevant code): https://docs.rs/cov-mark/latest/cov_mark/.

I am still on the fence where maintaining such coverage marks helps overall.

runtime/near-vm-runner/src/prepare.rs
matklad: Yeah, I agree that it's better to just remove wishful-thinking comments from this module. In general, every time I open this file I want to refactor it. 

I'd rather not do that in the PR though, as the goal here is to change a bit of the behavior we want to change, and to hold everything else constant.

PR #6053, opened 2022-01-12
refactor: don't panic in nearcore::load_config()

marcelo-gonzalez:
This cleans up a few more panics in favor of presenting a more user-friendly error when things go wrong.


chain/indexer/src/lib.rs
```diff
@@ -96,8 +96,15 @@ impl Indexer {
...
-            nearcore::config::load_config(&indexer_config.home_dir, GenesisValidationMode::Full);
+        let near_config = match nearcore::config::load_config(
+            &indexer_config.home_dir,
+            GenesisValidationMode::Full,
+        ) {
+            Ok(c) => c,
+            Err(e) => {
+                panic!("Error loading config: {:#}", e);
+            }
+        };
```
matklad, line 107:
```suggestion
        let near_config = nearcore::config::load_config(
            &indexer_config.home_dir,
            GenesisValidationMode::Full,
        ).unwrap_or_else(|e| panic!("Error loading config: {:#}", e));
```

core/crypto/src/key_file.rs
```diff
@@ -28,10 +28,10 @@ impl KeyFile {
         file.write_all(str.as_bytes())
     }
 
-    pub fn from_file(path: &Path) -> Self {
-        let mut file = File::open(path).expect("Could not open key file.");
+    pub fn from_file(path: &Path) -> std::io::Result<Self> {
+        let mut file = File::open(path)?;
         let mut content = String::new();
-        file.read_to_string(&mut content).expect("Could not read from key file.");
-        serde_json::from_str(&content).expect("Failed to deserialize KeyFile")
```
matklad, line 35: while we are at it, can simplify to 

```
let content = fs::read_to_string(path)?;
```

core/crypto/src/signer.rs
```diff
@@ -60,8 +60,8 @@ impl InMemorySigner {
         Self { account_id, public_key: secret_key.public_key(), secret_key }
     }
 
-    pub fn from_file(path: &Path) -> Self {
-        KeyFile::from_file(path).into()
+    pub fn from_file(path: &Path) -> std::io::Result<Self> {
+        Ok(KeyFile::from_file(path)?.into())
```
matklad, line 64: Alternative, not necessary better, way to write this: 

```rust
KeyFile::from_file(path).map(InMemorySigner::from)
```

core/crypto/src/key_file.rs
```diff
@@ -28,10 +28,10 @@ impl KeyFile {
         file.write_all(str.as_bytes())
     }
 
-    pub fn from_file(path: &Path) -> Self {
-        let mut file = File::open(path).expect("Could not open key file.");
+    pub fn from_file(path: &Path) -> std::io::Result<Self> {
```
matklad, line 31:
```suggestion
    pub fn from_file(path: &Path) -> io::Result<Self> {
```

That's the idiomatic way to spell io result

test-utils/state-viewer/src/cli.rs
```diff
@@ -109,7 +109,13 @@ pub enum StateViewerSubCommand {
...
 impl StateViewerSubCommand {
     pub fn run(self, home_dir: &Path, genesis_validation: GenesisValidationMode) {
-        let near_config = load_config(home_dir, genesis_validation);
+        let near_config = match load_config(home_dir, genesis_validation) {
+            Ok(c) => c,
+            Err(e) => {
+                tracing::error!("Error loading config: {:#}", e);
+                return;
+            }
+        };
```
matklad, line 118: I think it might be better to `.unwrap()` rather than `return` here and elsewhere, such that the exit code is non-zero.

nearcore/src/config.rs
```diff
@@ -1257,21 +1257,29 @@ impl From<NodeKeyFile> for KeyFile {
...
     let genesis_file = dir.join(&config.genesis_file);
-    let validator_signer = if dir.join(&config.validator_key_file).exists() {
+    let validator_file = dir.join(&config.validator_key_file);
+    let validator_signer = if validator_file.exists() {
         let signer =
-            Arc::new(InMemoryValidatorSigner::from_file(&dir.join(&config.validator_key_file)))
-                as Arc<dyn ValidatorSigner>;
+            Arc::new(InMemoryValidatorSigner::from_file(&validator_file).with_context(|| {
+                format!("Failed initializing validator signer from {}", validator_file.display())
+            })?) as Arc<dyn ValidatorSigner>;
```
matklad, line 1271: This expression is a bit too busy now, with both error handling and `as` casing. I'd re-shuffled it a bit

```reust
let signer = InnMemoryValidatorSigner::from_file(&validator_file).with_context(|| {
    format!("Failed initializing validator signer from {}", validator_file.display())
})?;
Some(Arc::new(signer) as Arc<dyn ValidatorSigner>)
```

matklad (approved):

test-utils/state-viewer/src/cli.rs
marcelo-gonzalez: done. @matklad what do you think about returning a result from main(), maybe in another PR. just so we can exit normally here and not have to panic to get a nonzero exit code

core/crypto/src/signer.rs
marcelo-gonzalez: yah I like that better too

test-utils/state-viewer/src/cli.rs
matklad: Yeah, we should have 

```
fn main() -> anyhow::Result<()>
```

pmnoxx: > ```
> fn main() -> anyhow::Result<()>
> ```

Or even better:
```rust
fn main() -> Result<(), anyhow::Error>
```

nearcore/src/config.rs
```diff
@@ -1233,11 +1233,11 @@ struct NodeKeyFile {
...
-    fn from_file(path: &Path) -> Self {
-        let mut file = File::open(path).expect("Could not open key file.");
+    fn from_file(path: &Path) -> std::io::Result<Self> {
+        let mut file = File::open(path)?;
         let mut content = String::new();
-        file.read_to_string(&mut content).expect("Could not read from key file.");
-        serde_json::from_str(&content).expect("Failed to deserialize KeyFile")
+        file.read_to_string(&mut content)?;
+        Ok(serde_json::from_str(&content)?)
     }
```
pmnoxx, line 1241: I would add `with_context(||format!("path={:?}", path)`, or something similar to indicate, which file couldn't be opened.

pmnoxx (approved):
LGTM

nearcore/src/config.rs
```diff
@@ -1257,21 +1257,28 @@ impl From<NodeKeyFile> for KeyFile {
...
 }
 
-pub fn load_config(dir: &Path, genesis_validation: GenesisValidationMode) -> NearConfig {
-    let config = Config::from_file(&dir.join(CONFIG_FILENAME)).unwrap();
+pub fn load_config(
+    dir: &Path,
+    genesis_validation: GenesisValidationMode,
+) -> anyhow::Result<NearConfig> {
```
pmnoxx, line 1263: I would rather use, `Result<NearConfig, anyhow::Error>`; This should be much easier to understand, the result type is from std, while the only thing that's different is the error type.

marcelo-gonzalez: done

nearcore/src/config.rs
marcelo-gonzalez: Yeah, I made the caller in load_config() attach that context with the pathname. Felt more natural to me to do it that way since the caller knows what path it tried to open. No super strong opinion about that though. Do you think its better to add the context here?

test-utils/state-viewer/src/cli.rs
matklad: Hm, I'd say that `anyhow::Result<()>` would be better. That is

* both `Result<(), anyhow::Error>` and `anyhow::Result<()>` mean exactly the same, they are different spellings of the same type.
* but `anyhow::Result<()>` is a preferred spelling:
  * it's a bit more idiomatic (compare, ig, `io::Result` or `fmt::Result`)
  * it's just reads better from "qualified names" point of view -- `anyhow` is the prefix, rather than the middle. 

No strong feelings here though!

matklad: To expand on this: 

* yes, `Result<(), anyhow::Error>` is a bit easier to understand, as it is more direct, you don't have to know about `foo::Result<T>` idiom
* but this is outweighed by qualified result form being idiomatic -- code generally should be written in idiomatic form for consistency, and because idioms, while taking some time to learn, do generally make the code better. 

For qualified form being idiomatic:

* example from stdlib https://doc.rust-lang.org/stable/std/io/index.html (io::Result)
* documentation for the idiom https://blog.burntsushi.net/rust-error-handling/#the-result-type-alias-idiom
* the alias that exists in anyhow

pmnoxx: @matklad I see, personally I think that `Result<(), anyhow::Error>` is better.

However, on the other hand, Rust std already uses `io::Result`, `fmt::Result`. I agree that it's better to be consistent. Most often consistent approach is preferred, even though it may not be an optimal approach.
In that case, let's use `anyhow::Result<()> `

nikurt (approved):

nearcore/src/config.rs
pmnoxx: Either way is fine. It depends on how many callers are there, if this function is only called 1-3 times, then doing it on the caller side is fine.

tools/restaked/src/main.rs
```diff
@@ -70,9 +70,23 @@ fn main() {
...
+            return;
+        }
+    };
     // Support configuring if there is another key.
-    let signer = InMemorySigner::from_file(&home_dir.join(&config.validator_key_file));
+    let signer = match InMemorySigner::from_file(&key_path) {
+        Ok(s) => s,
+        Err(e) => {
+            error!("Failed to initialize signer from key file at {:?}: {:#}", key_path, e);
+            return;
```
mm-near, line 87: shouldn't we return a non-zero error code to os?

mm-near (changes requested):

genesis-tools/genesis-populate/src/main.rs
```diff
@@ -27,7 +27,13 @@ fn main() {
         .value_of("additional-accounts-num")
         .map(|x| x.parse::<u64>().expect("Failed to parse number of additional accounts."))
         .unwrap();
-    let near_config = load_config(home_dir, GenesisValidationMode::Full);
+    let near_config = match load_config(home_dir, GenesisValidationMode::Full) {
+        Ok(c) => c,
+        Err(e) => {
+            eprintln!("Error loading config: {:#}", e);
+            return;
```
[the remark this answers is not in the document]
marcelo-gonzalez: oof yea good catch. @matklad brought that up in the comment above, I just forgot about these somehow. Updated now to make these panic with the plan to have `main()` return an `anyhow::Result` like discussed there

tools/restaked/src/main.rs
```diff
@@ -70,9 +70,19 @@ fn main() {
...
 
     let config = Config::from_file(&home_dir.join(CONFIG_FILENAME)).expect("can't load config");
-    let key_file = KeyFile::from_file(&home_dir.join(&config.validator_key_file));
+
+    let key_path = home_dir.join(&config.validator_key_file);
+    let key_file = match KeyFile::from_file(&key_path) {
+        Ok(f) => f,
+        Err(e) => {
+            error!("Failed to open key file at {:?}: {:#}", &key_path, e);
+            return;
```
matklad, line 79: `.unwrap()` to exit with non-zero here as welll?

genesis-tools/genesis-populate/src/main.rs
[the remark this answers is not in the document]
marcelo-gonzalez: @mm-near friendly ping

khorolets (approved):

bowenwang1996 (approved):

marcelo-gonzalez:
cc @mina86 last merge commit touches your `generate_or_load_key()`

mina86 (approved):

marcelo-gonzalez:
@mina86 oops one more ping... Didn't realize that this would break the `test_generate_or_load_key_panic` test since it doesnt panic anymore. Moved that logic into the existing test, PTAL.

mina86 (approved):

PR #6054, opened 2022-01-13
neard: deprecate unsafe_reset_{all,data} commands

mina86:
Those commands are literally just ‘rm -r’ but with custom names which
only serves to confuse users about what they might be doing.  In
addition, it’s unnecessary code in neard which can easily be deleted
without loss of features.

Also, don’t advertise removing all the data if expected chain id
doesn’t match one in the file on disk.  Deleting the entire home
directory is a destructive task so we shouldn’t willy-nilly recommend
it when neard fails to start.

Issue: https://github.com/near/nearcore/issues/6010


neard/src/cli.rs
```diff
@@ -72,6 +72,15 @@ impl NeardCmd {
     }
 }
 
+fn unsafe_reset(command: &str, path: &std::path::Path, what: &str, default: &str) {
+    let dir =
+        path.to_str().map(|path| shell_escape::unix::escape(path.into())).unwrap_or(default.into());
+    warn!(target: "neard", "The ‘{}’ command is deprecated", command);
```
nagisa, line 78: Worth mentioning when it'll be removed. If somebody runs this command for the first time in Summer of 2022, they may not realize that this command will be come in a couple of months.

matklad (approved):

mm-near (approved):

PR #6056, opened 2022-01-13
Move tests taking over 60 s to nightly

mina86:

### DIFF
--- a/integration-tests/src/tests/client/chunks_management.rs
+++ b/integration-tests/src/tests/client/chunks_management.rs
@@ -271,6 +271,7 @@ fn chunks_produced_and_distributed_one_val_per_shard() {
 /// give up on getting the part from test1 and will get it from test2 (who will have it because
 /// `validator_groups=2`)
 #[test]
+#[cfg_attr(not(feature = "expensive_tests"), ignore)]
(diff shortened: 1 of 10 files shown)


mina86:
Blocked by https://github.com/near/nearcore/pull/6032 for now

matklad:
One improvement is to add `## Testing` to cross-cutting concerns section of architecture md: 

https://github.com/near/nearcore/blob/master/docs/architecture.md#cross-cutting-concerns

There, we say something like:

* we expect tests to take less than 60s
* if they take more, use this `cfg_attr` spell
* see `nightly/readme` for how the stuff is actually run

That is, the goal is not to be super thorough, but to just have a place where we can easily add our tribal knowledge about testing. 

If we *do* want to be thorough, it probably makes sense to create https://github.com/near/nearcore/tree/master/docs/testing.md, and move contents from https://github.com/near/nearcore/wiki/Writing-tests-for-nearcore (and maybe nightly/readme?) there.

mina86:
> One improvement is to add ## Testing to cross-cutting concerns section of architecture md:

https://github.com/near/nearcore/pull/6060

matklad (approved):

Issue #3284, opened 2020-09-03
Bound size of `TransactionPool` in memory


birchmd:
Presently transactions are simply stored in a `BTreeMap` in memory, however, if we receive too many transactions too quickly, then this could grow over time, especially if the transactions are large.

nikurt:
Not sure what to do if a pool reaches its maximum size but transactions keep coming in.
Is it possible that rejecting incoming transactions will also the one transaction that needs to be processed next, i.e. a transaction with the lowest nonce?
@mzhangmzz do you have suggestions?

nikurt:
To get an idea of a reasonable tx pool size bound, will add a metric get back to it.

bowenwang1996:
> Not sure what to do if a pool reaches its maximum size but transactions keep coming in.

We might need to stop accepting transactions at that point. cc @mm-near

akashin:
I'll pick this issue in the context of congestion work (https://github.com/near/nearcore/milestone/26, https://github.com/near/nearcore/issues/8878).

As a first step, I plan to investigate two naive approaches:
- [ ] Introducing hard capacity limit for transaction pool (e.g. 1000 transactions)
- [ ] Introducing TTL for transactions in the pool (e.g. 1 minute)

I highly suspect that both approaches will break some tests/assumptions that the clients make. My first goal will be to understand:
- [ ] What will actually break
- [ ] How easy would it be to fix it
- [ ] How should the clients behave in the face of these limits: https://github.com/near/nearcore/issues/8879

akashin:
We do indeed have tests that rely on putting many transactions into the pool. Example failures are:
- https://buildkite.com/nearprotocol/nearcore/builds/27048 (with size limit 20)
- https://buildkite.com/nearprotocol/nearcore/builds/27050 (with size limit 2)

Concrete test https://github.com/near/nearcore/blob/055a9e27f2652e0f13a398d74660db20e1e9acd1/integration-tests/src/tests/client/benchmarks.rs#L25 puts 20 transactions into the pool before starting to process them.

I'm adding explicit checks for the returned transaction status https://github.com/near/nearcore/pull/8976 which would make it much easier to catch tests broken by introducing transaction pool size limits.

Retrying the failure in all those tests would likely be too burdensome, so we'll have to use some high-enough limit that will cover the majority of tests and only introduce in the retries in tests that actually exercise the congestion scenario.

akashin:
[...]

PR #6057, opened 2022-01-13
Metric for the transaction pool size

nikurt:
Issue #3284


pmnoxx (approved):
LGTM

chain/pool/src/lib.rs
```diff
@@ -86,8 +90,10 @@ impl TransactionPool {
             if remove_entry {
                 self.transactions.remove(&key);
             }
-            for hash in hashes {
-                self.unique_transactions.remove(&hash);
+            for hash in &hashes {
+                if self.unique_transactions.remove(&hash) {
+                    metrics::TRANSACTION_POOL_TOTAL.dec();
```
mm-near, line 95: do we have any way of checking this value in the unittests below?

mm-near (approved):

chain/pool/src/lib.rs
nikurt: Unittesting is complicated because metrics are global variables. As unittests run in parallel, those values get randomly overwritten.
That can be fixed by adding a layer of abstraction - a local object proxying metric operations - but that adds quiet a lot of complexity.

PR #6062, opened 2022-01-13
fix(chain|: fix a bug where the order of receipts from different blocks are reversed

mzhangmzz:
We found a bug that receipt created at the later block may be executed first, and this is because `get_incoming_receipts_for_shard` includes incoming receipts from blocks in a backward order. This PR fixes that.

Test Plan:
- Add a unit test for get_incoming_receipts_for_shard


chain/chain/src/store.rs
```diff
@@ -1193,14 +1193,18 @@ impl<'a> ChainStoreUpdate<'a> {
...
 
+    /// Return all incoming receipts that need to be processed by shard `shard_id` at block with
+    /// block hash `block_hash`. `last_chunk_height_included` is the height of the last block that
+    /// contains a non-empty chunk for the shard before `block_hash`. The results will include
+    /// all incoming receipts sent to shard `shard_id` from the block at height
+    /// `last_chunk_height_included` + 1 to the block at `block_hash`.
     pub fn get_incoming_receipts_for_shard(
         &mut self,
         shard_id: ShardId,
         mut block_hash: CryptoHash,
```
mm-near, line 1204: i'm surprised that block_hash is mutable

chain/chain/src/store.rs
```diff
@@ -1213,14 +1217,18 @@ impl<'a> ChainStoreUpdate<'a> {
             }
 
             let prev_hash = *header.prev_hash();
+            blocks.push(block_hash);
+
+            block_hash = prev_hash;
+        }
 
+        let mut ret = vec![];
+        for block_hash in blocks.into_iter().rev() {
```
mm-near, line 1226: add a comment explaining why we're going in 'reverse'

mm-near (approved):

mzhangmzz:
Not ready to merge yet, I think this fix needs a protocol upgrade

chain/chain/src/store.rs
bowenwang1996: yeah this piece of code is not optimal

akhi3030:
I think this is not being actively worked on.

PR #6063, opened 2022-01-13
refactor: inline highest_height_peer()

marcelo-gonzalez:
It seems the original structure was needed for the changes in
3530e7d058, but then undone in b90822f37b.


chain/client/src/sync.rs
```diff
@@ -53,14 +53,7 @@ pub const NS_PER_SECOND: u128 = 1_000_000_000;
 
 /// Get random peer from the hightest height peers.
 pub fn highest_height_peer(highest_height_peers: &Vec<FullPeerInfo>) -> Option<FullPeerInfo> {
```
pmnoxx, line 55: Consider also changing
```
highest_height_peers: &Vec<FullPeerInfo>
```
to
```
highest_height_peers: &[FullPeerInfo]
```

pmnoxx (approved):
LGTM

chain/client/src/sync.rs
marcelo-gonzalez: yea, that sounds good

chain/client/src/sync.rs
```diff
@@ -52,15 +52,8 @@ pub const MAX_PENDING_PART: u64 = MAX_STATE_PART_REQUEST * 10000;
...
-pub fn highest_height_peer(highest_height_peers: &Vec<FullPeerInfo>) -> Option<FullPeerInfo> {
-    if highest_height_peers.len() == 0 {
-        return None;
-    }
-
-    match highest_height_peers.iter().choose(&mut thread_rng()) {
-        None => highest_height_peers.choose(&mut thread_rng()).cloned(),
-        Some(peer) => Some(peer.clone()),
-    }
+pub fn highest_height_peer(highest_height_peers: &[FullPeerInfo]) -> Option<FullPeerInfo> {
```
pmnoxx, line 55: It may be a good idea to simply inline this method. I think that the inline code is easier to read. By in-lining the code, the reader will not have to check what the method does, as it should be obvious,

I would rather see:
```
         if let Some(peer) = highest_height_peers.choose(&mut thread_rng()).cloned() {
```
than
```
         if let Some(peer) = highest_height_peer(highest_height_peers) {
```

pmnoxx: I looked at the code. It looks like the clone may be be even needed in some of the cases.

marcelo-gonzalez: good point, didn't even think of that...

chain/client/src/sync.rs
```diff
@@ -201,7 +189,7 @@ impl HeaderSync {
             *sync_status =
                 SyncStatus::HeaderSync { current_height: header_head.height, highest_height };
             self.syncing_peer = None;
-            if let Some(peer) = highest_height_peer(highest_height_peers) {
+            if let Some(peer) = highest_height_peers.choose(&mut thread_rng()).cloned() {
```
mm-near, line 192: nit: do you need cloned here?

mm-near (approved):

chain/client/src/sync.rs
marcelo-gonzalez: yeah I think here it is needed actually, because of the `self.syncing_peer =` part, so the struct has to be cloned somewhere eventually

PR #6065, opened 2022-01-14
refactor: remove dead code

matklad:
This was only used by the simulator, which is deprecated. The current
implementation diverged from the main `fn outcome`, so it's better just
to remove this.


pmnoxx (approved):
LGTM

Btw, is there a good programatical way of finding unused `pub fn` functions?

mm-near (approved):

matklad:
@pmnoxx nothing that would've work out of the box, as far as I know. Couple of approaches here: 

* trial deletion: replace `pub` with `pub(crate)`, run `cargo check` if that compiles (with warnings), than everything that reaches the `fn` is unused (ie, sometimes *just* deleting the thing doesn't work, because it is used by some other `pub` thing which is itself unused)
* using IntelliJ Rust / rust-analyzer to find things with zero usages.

PR #6066, opened 2022-01-14
Instrument gas usage rather than opcodes

nagisa:
Here we adjust the instrumentation to insert `gas(...)` where the argument
is the actual gas amount rather than the number of WASM instructions.

This results in larger number being passed to the `gas` function, and
therefore fewer instructions being possible per-block – with current
`regular_op_cost` that's about 5.5k instructions per block maximum.

As a benefit, however, this allows to remove a multiplication with
significant data dependencies every time gas cost is charged. This makes
sense, as the `regular_op_cost` won't change between protocol versions,
and therefore computing block's gas cost can be done once – at compile time.

Every time `regular_op_cost` changes, the contract has will already
change due to it including a hash of the `VMConfig`, so the contracts
previously compiled will naturally get “evicted” from the cache.

The limitation pertaining to instruction count could be dealt with by changing
instrumentation to emit calls to the newly added `gas64` function, or perhaps
by changing the `gas` function to accept the `u64` argument directly.

(Similar changes would be necessary for the intrinsic codegen implementation)

---

I'm sure I'm missing some details in this implementation, but I'm otherwise confident
we could make all the changes necessary to make this work. Potentially even without
a protocol break.

### DIFF
--- a/runtime/near-vm-logic/src/gas_counter.rs
+++ b/runtime/near-vm-logic/src/gas_counter.rs
@@ -69,7 +69,6 @@ impl GasCounter {
     pub fn new(
         ext_costs_config: ExtCostsConfig,
         max_gas_burnt: Gas,
-        opcode_cost: u32,
         prepaid_gas: Gas,
         is_view: bool,
     ) -> Self {
@@ -81,7 +80,7 @@ impl GasCounter {
             fast_counter: FastGasCounter {
                 burnt_gas: 0,
                 gas_limit: min(max_gas_burnt, prepaid_gas),
-                opcode_cost: Gas::from(opcode_cost),
+                opcode_cost: 1,
             },
             max_gas_burnt: max_gas_burnt,
             promises_gas: 0,
@@ -153,11 +152,6 @@ impl GasCounter {
         } else {
             HostError::GasExceeded
         }
-    }
-
-    pub fn pay_wasm_gas(&mut self, opcodes: u32) -> Result<()> {
-        let value = Gas::from(opcodes) * self.fast_counter.opcode_cost;
-        self.burn_gas(value)
     }
 
     /// Very special function to get the gas counter pointer for generated machine code.
@@ -290,7 +284,7 @@ mod tests {
     use near_primitives_core::types::Gas;
 
     fn make_test_counter(max_burnt: Gas, prepaid: Gas, is_view: bool) -> super::GasCounter {
-        super::GasCounter::new(ExtCostsConfig::test(), max_burnt, 1, prepaid, is_view)
+        super::GasCounter::new(ExtCostsConfig::test(), max_burnt, prepaid, is_view)
     }
 
     #[test]
--- a/runtime/near-vm-logic/src/logic.rs
+++ b/runtime/near-vm-logic/src/logic.rs
@@ -119,7 +119,6 @@ impl<'a> VMLogic<'a> {
         let gas_counter = GasCounter::new(
             config.ext_costs.clone(),
             max_gas_burnt,
-            config.regular_op_cost,
             context.prepaid_gas,
             context.is_view(),
         );
@@ -1045,15 +1044,27 @@ impl<'a> VMLogic<'a> {
         Ok(false as u64)
     }
 
-    /// Called by gas metering injected into Wasm. Counts both towards `burnt_gas` and `used_gas`.
+    /// Potentially called by gas metering injected into Wasm. Counts both towards `burnt_gas` and
+    /// `used_gas`.
+    ///
+    /// Today we implement this operation directly within the codegen so this function is called
+    /// only in non-standard situations – use of the different runtime, or if testing with
+    /// intrinsification disabled.
     ///
     /// # Errors
     ///
     /// * If passed gas amount somehow overflows internal gas counters returns `IntegerOverflow`;
     /// * If we exceed usage limit imposed on burnt gas returns `GasLimitExceeded`;
     /// * If we exceed the `prepaid_gas` then returns `GasExceeded`.
-    pub fn gas(&mut self, opcodes: u32) -> Result<()> {
-        self.gas_counter.pay_wasm_gas(opcodes)
+    pub fn gas(&mut self, gas: u32) -> Result<()> {
+        let gas = Gas::from(gas);
+        self.gas_counter.burn_gas(gas)
+    }
+
+    /// See [`Self::gas`]. This takes u64 gas count.
+    pub fn gas64(&mut self, gas: u64) -> Result<()> {
+        let gas = Gas::from(gas);
+        self.gas_counter.burn_gas(gas)
     }
 
     // ################
--- a/runtime/near-vm-logic/src/tests/gas_counter.rs
+++ b/runtime/near-vm-logic/src/tests/gas_counter.rs
@@ -2,7 +2,7 @@ use crate::tests::fixtures::get_context;
 use crate::tests::helpers::*;
 use crate::tests::vm_logic_builder::VMLogicBuilder;
 use crate::types::Gas;
-use crate::{VMConfig, VMLogic};
+use crate::VMLogic;
 
 #[test]
 fn test_dont_burn_gas_when_exceeding_attached_gas_limit() {
@@ -24,15 +24,13 @@ fn test_dont_burn_gas_when_exceeding_attached_gas_limit() {
 #[test]
 fn test_limit_wasm_gas_after_attaching_gas() {
     let gas_limit = 10u64.pow(14);
-    let op_limit = op_limit(gas_limit);
-
     let mut logic_builder = VMLogicBuilder::default().max_gas_burnt(gas_limit * 2);
     let mut logic = logic_builder.build_with_prepaid_gas(gas_limit);
 
     let index = promise_create(&mut logic, b"rick.test", 0, 0).expect("should create a promise");
     promise_batch_action_function_call(&mut logic, index, 0, gas_limit / 2)
         .expect("should add action to receipt");
-    logic.gas((op_limit / 2) as u32).expect_err("should fail with gas limit");
+    logic.gas64(Gas::from(gas_limit) / 2).expect_err("should fail with gas limit");
     let outcome = logic.outcome();
 
     assert_eq!(outcome.used_gas, gas_limit);
@@ -43,12 +41,9 @@ fn test_limit_wasm_gas_after_attaching_gas() {
 #[test]
 fn test_cant_burn_more_than_max_gas_burnt_gas() {
     let gas_limit = 10u64.pow(14);
-    let op_limit = op_limit(gas_limit);
-
     let mut logic_builder = VMLogicBuilder::default().max_gas_burnt(gas_limit);
     let mut logic = logic_builder.build_with_prepaid_gas(gas_limit * 2);
-
-    logic.gas(op_limit * 3).expect_err("should fail with gas limit");
+    logic.gas64(Gas::from(gas_limit) * 3).expect_err("should fail with gas limit");
     let outcome = logic.outcome();
 
     assert_eq!(outcome.burnt_gas, gas_limit);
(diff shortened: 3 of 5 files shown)


matklad:
Something along these lines sounds reasonable. In my mind, there are three alternatives we can have:

* limit gas, instrument with the number of instructions, do multiplication in the generated code (current impl)
* limit gas, instrument with gas, do multiplication during compilation (this RFC)
* limit number of instructions, instrument number of instructions, do multiplication on host fn entry/exit (something we've vaguely discussed)

It does feel like this proposal should be the fastest way to do this, as we move all multiplications to compile time. 

Do we have rough idea of how much faster will this be?

matklad:
>Every time regular_op_cost changes, the contract has will already
change due to it including a hash of the VMConfig, so the contracts
previously compiled will naturally get “evicted” from the cache.

That's a great observation! Basically, every time we change any cost, we want to recompile all contracts. This might be fine, but also seems like something we can potentially improve in the future. Though, it seems quite likely that, when we want to change `regular_op_cost` specifically, we actually want to recompile the world, as this cost reduction most likely indicates overal vm upgrade or some such.

nagisa:
> Do we have rough idea of how much faster will this be?

So, here's the assembly for what is roughly current assembly that we generate for `gas` in wasmer, after intrinsification:

```asm
mov    0x10(%rax,%riz,1),%rcx
mov    $0x64,%rdx
mulq   %rdx,%rcx
mov    0x0(%rax,%riz,1),%rdx
addq    %rcx,%rdx
jo     .overflow
cmp    %rdx,0x8(%rax,%riz,1)
mov    %rdx,0x0(%rax,%riz,1)
jbe    .overflow
```

changing nothing other than removing the multiplication, which is possible after this PR, would produce something more like this:

```asm
mov    $56320000,%rcx
mov    0x0(%rax,%riz,1),%rdx
addq    %rcx,%rdx
jo     .overflow
cmp    %rdx,0x8(%rax,%riz,1)
mov    %rdx,0x0(%rax,%riz,1)
jbe    .overflow
```

If this code is run for 100 loop iterations (imagine something along the lines of `for _ in 0..100 {}` that's fully unrolled), here are the numbers:

| | current znver1 | new znver1 | current haswell | new skylake |
| - | - | - | - | - |
| cycles | 318 | 215 | 265 | 213 |
| ipc | 2.83 | 3.26 | 3.40 | 3.29 |
| block reciprocal throughput | 2.8 | 2.0 | 2.5 | 2.0 |

So it is a modest 20%-or-so improvement to the throughput of `gas`.

Eventually, after the multiplication is removed, the machine code can be further tweaked to reduce the hot instructions down to something along the lines of

```asm
subq $56320000, 0x0(%rax,%riz,1)
js .overflow
```

This would require some computation after a VM return to figure out the actual amount of gas consumed, but during normal execution would improve the throughput of `gas` by another 50%-or-so.

NB: the measurements are synthesized by `llvm-mca` and the perceived performance different may be quite different in practice depending on the surrounding code.

runtime/near-vm-logic/src/gas_counter.rs
```diff
@@ -155,9 +154,8 @@ impl GasCounter {
         }
     }
 
-    pub fn pay_wasm_gas(&mut self, opcodes: u32) -> Result<()> {
-        let value = Gas::from(opcodes) * self.fast_counter.opcode_cost;
-        self.burn_gas(value)
+    pub fn pay_wasm_gas(&mut self, gas: Gas) -> Result<()> {
+        self.burn_gas(gas)
     }
```
pmnoxx, line 159: Consider in-lining this method. It loos like it can be called directly, with the same arguments.

pmnoxx (approved):
LGTM

nagisa:
The test failure in upgradeable looks like this:

```
Jan 18 22:00:49.839 DEBUG Process block{#23}: chain: Failed to validate chunk extra: Error { inner:    0: failure::backtrace::Backtrace::new
   1: near_chain::chain::ChainUpdate::apply_chunks_preprocessing
   2: near_chain::chain::ChainUpdate::process_block
   3: near_chain::chain::Chain::process_block_single
   4: near_client::client::Client::process_block
   5: near_client::client_actor::ClientActor::process_block
   6: <near_client::client_actor::ClientActor as actix::handler::Handler<near_network::types::NetworkClientMessages>>::handle
   7: <actix::address::envelope::SyncEnvelopeProxy<M> as actix::address::envelope::EnvelopeProxy<A>>::handle
   8: <actix::contextimpl::ContextFut<A,C> as core::future::future::Future>::poll
   9: tokio::runtime::task::raw::poll
  10: tokio::task::local::LocalSet::tick
  11: <core::future::from_generator::GenFuture<T> as core::future::future::Future>::poll
  12: neard::cli::NeardCmd::parse_and_run
  13: std::sys_common::backtrace::__rust_begin_short_backtrace
  14: main
  15: __libc_start_main
  16: _start


Invalid State Root Hash }
Jan 18 22:00:49.887 DEBUG Process block{#23}: chain: Verifying challenges []
Jan 18 22:00:49.887 DEBUG Process block{#23}:process_state_update: runtime: enter
Jan 18 22:00:49.887 DEBUG Process block{#23}:process_state_update: runtime: block height: 22, is next_block_epoch_start false
Jan 18 22:00:49.887 DEBUG Process block{#23}:process_state_update: runtime: epoch height: 1, epoch id: EpochId(`11111111111111111111111111111111`), current_protocol_version: 50, is_first_block_of_version: false
Jan 18 22:00:49.887 DEBUG Process block{#23}:process_state_update:Runtime::apply: runtime: enter
Jan 18 22:00:49.887 DEBUG Process block{#23}:process_state_update:Runtime::apply: runtime: close time.busy=96.5µs time.idle=2.26µs
Jan 18 22:00:49.887 DEBUG Process block{#23}:process_state_update: runtime: close time.busy=165µs time.idle=2.43µs
Jan 18 22:00:49.887 DEBUG Process block{#23}: chain: close time.busy=50.1ms time.idle=2.66µs
Jan 18 22:00:49.887  WARN client: receive bad block: Invalid Chunk State 
 Cause: Unknown    
Jan 18 22:00:50.442 DEBUG client: Receive approval Approval { inner: Endorsement(`CqZsKfem42cAuZ17wz1U8WrcXSApDVD6bknA1hnRye6j`), target_height: 24, signature: ed25519:4kDRL9uRVS2jUfczU9KHmVmDrQkqWvjdcjJeVj8geRJk35W7JhdDHMUy1UWNb71g833b1fWAiGjWVPeQJKtxMTNu, account_id: AccountId("test1") } from peer ed25519:FXXrTXiKWpXj1R6r5fBvMLpstd8gPyrBq3qMByqKVzKF    
Jan 18 22:00:51.357  INFO stats: #      22 2WqEAiciweWKx9A1nC9PQ8ZjynQ3pjN6eGCp2bpwJ6jT V/2  1/1/40 peers ⬇ 3.4kiB/s ⬆ 6.1kiB/s 1.30 bps 1.18 Tgas/s CPU: 1%, Mem: 50.3 MiB    
Jan 18 22:00:52.160 DEBUG client: Receive approval Approval { inner: Skip(23), target_height: 25, signature: ed25519:4MzZQeT7cTmEK6Tn7SZAfMP432k3cvHP7TWXrK7AMf2M1nzthqeLkyFLJVCNAR6cjjSgmDfQWr2gYWKdiw6XSZjS, account_id: AccountId("test1") } from peer ed25519:FXXrTXiKWpXj1R6r5fBvMLpstd8gPyrBq3qMByqKVzKF    
Jan 18 22:00:54.486 DEBUG client: Receive approval Approval { inner: Skip(23), target_height: 26, signature: ed25519:38heWjmoova1t4i1HEhQScDiSY8FhPkaM6SF1GKC7DLUSBFVoYz5ojSmMzCa5GRuunH22iA8stJxmktRi2ZDznQP, account_id: AccountId("test1") } from peer ed25519:FXXrTXiKWpXj1R6r5fBvMLpstd8gPyrBq3qMByqKVzKF    
Jan 18 22:00:57.124 DEBUG client: Receive approval Approval { inner: Skip(23), target_height: 27, signature: ed25519:32c9Pue37T9WpH94yJyd2qrLZ2cTdCVyCDF2vqmfT9kcbHx17R22ntffiy6DD3KRHqpvSAUQGhdQ4CMNgeYAESSc, account_id: AccountId("test1") } from peer ed25519:FXXrTXiKWpXj1R6r5fBvMLpstd8gPyrBq3qMByqKVzKF    
Jan 18 22:00:57.366  WARN jsonrpc: Timeout: tx_polling method. tx_info Transaction(SignedTransaction { transaction: Transaction { signer_id: AccountId("test0"), public_key: ed25519:He7QeRuwizNEhBioYG3u4DZ8jWXyETiyNzFD3MkTjDMf, nonce: 12000001, receiver_id: AccountId("test_account.test0"), block_hash: `EmVSHwnJfQqoMTwUc37Wks3k5H1DX3Roi7v3aWvQJ9Sz`, actions: [FunctionCall(FunctionCallAction { method_name: call_promise, args: (295)`[{"create": {"account_id": "test_account.test0", "method_name": "call_promise", "arguments": [], "amount": "0", "gas": 300000000…`, gas: 90000000000000, deposit: 0 })] }, signature: ed25519:fPhwHXJoiX7qDi8NDt32kocaANSAm2CM4WUgHYhres8oyUpcbiHgC56iPo1GDBhBWnHfFKq4jfd6CkVc8bRUJsL, hash: `Extn58VAAojQ2zx2XbvbXyikMJiiJ87BVB6amiXCbrT3`, size: 448 })
```

Not entirely sure how to interpret this, but I guess there's a hash/protocol change happening somewhere?

bowenwang1996:
> Not entirely sure how to interpret this, but I guess there's a hash/protocol change happening somewhere?

Yes. This usually means there is a different state transition after this PR, i.e, it is not compatible with the current protocol

nagisa:
Right, my changes… changed the amount of gas charged for memory.grow instruction, hence the incompatibility.

nagisa:
NB: I marked this as ready, but I would like to hear thoughts on how to best approach validating whether the reduction of maximum instructions per block to 5.5k is valid.

matklad:
On 5.5k, question, the best way to investigate this would be to use https://github.com/near/nearcore/pull/5823 to fetch just all of the contracts we have, and then see what's the maximal block size actually is. It's also useful to understand what's our current limit is. https://github.com/bytecodealliance/wasm-tools/blob/ac9028a4c52dbf1b0e78b24be33b7e5ce9618ca6/crates/wasmparser/src/limits.rs#L28 presumably is something else? 

The 5.5 limit seems a bit arbitrary, as it depends on specific gas cost. I'd feel more comfortable with us enforcing a limit which is not tied to other runtime parameters.

Alternatively, the reason why I think we hit 5.5 limit is that we use u32, and we use u32 precisely because we are counting instructions, rather than gas. If we want to make this change, we might keep all the calculation in u64 and sidestep the problem, perhaps?

matklad:
What's the corresponding wasmer change?

nagisa:
> What's the corresponding wasmer change?

`wasmer` needs no change to land this at the moment – we get the correct behaviour by setting `opcode_cost` to `1` in `FastGasCounter`, which will make wasmer to multiply by 1 for now.

Or are you saying you'd like to see what the improvements on wasmer side look like before considering landing this change?

---

> Alternatively, the reason why I think we hit 5.5 limit is that we use u32, and we use u32 precisely because we are counting instructions, rather than gas. If we want to make this change, we might keep all the calculation in u64 and sidestep the problem, perhaps?

Yes, I think ideally we would want to use `gas64` (as added in this PR). Implementing this would require changes to pwasm_utils since, AFAIK, it only supports emitting regular `gas(u32)` call today.

nagisa:
Here're the potential real-world improvements that could be had (with the changes on wasmer side as seen [here](https://github.com/near/wasmer/compare/nagisa/gas-opt)).

`regular_op_cost` computed by

```
cargo run --release --package runtime-params-estimator --bin runtime-params-estimator -- --accounts-num 1000 --additional-accounts-num 1000 --iters 100000 --warmup-iters 100 --metric time --costs WasmInstruction
```

goes down to a very consistent 202kgas down from 205kgas~210kgas on my machine. Not as significant as I might have hoped originally, but an improvement nevertheless.

```
cargo run --release --package runtime-params-estimator --bin runtime-params-estimator -- --accounts-num 1000 --additional-accounts-num 1000 --iters 100000 --warmup-iters 100 --metric icount --costs WasmInstruction --docker --full
```

computes 273264gas before and 244592gas after.

So <3% improvement when comparing wall clock time and ~10% improvement when counting icount.

---

```
cargo test -p aurora-engine-tests --features mainnet-test test_uniswap_input_multihop --release
```

Sees a wall-clock time reduction of… unknown.

The contracts this test attempts to run actually run into a insn/block limit with the `regular_op_cost = 822756` it sets! Either way its a good indicator that `gas64` work needs to happen one way or another before this can land.

matklad:
>~10% improvement when counting icount.

Wow, that's more than I expected! I wonder if this is good or bad though -- our estimation is still based on a grand total of one benchmark, and reducing the cost by 10% while the wall-clock time reduces by 3% reduces our reliability gap (but it still might be correct!)

My gut feeling is that it probably best to table this for know -- the improvement is meaningful, but not breathtaking, and the amount of work to do to productionalize this (gas64, potential protocol change, etc), seems not worth it at the time being, as there seems to be bigger fish to fry.

matklad:
Let's close this for now -- we probably won't forget about this when the time comes to revisit this

PR #6067, opened 2022-01-14
fix debug make target

marcelo-gonzalez:
the neard-debug target doesn't exist, so "make debug" currently fails

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -33,9 +33,10 @@ neard-release: NEAR_RELEASE_BUILD=release
 neard-release:
 	cargo build -p neard --release
 
-
-debug:
+neard-debug:
 	cargo build -p neard
+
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -34,7 +34,7 @@ neard-release:
 	cargo build -p neard --release
 
 
-debug: neard-debug
+debug:
 	cargo build -p neard
```
mina86, line 38: Sorry about that. Better still (so this target follows other targets),
add:

```
neard-debug:
	cargo build -p neard
```

and remove building `neard` from `debug` target.

mina86 (approved):

Makefile
marcelo-gonzalez: yeah good point, done.

Issue #5516, opened 2021-11-29
feat: Separate `jsonrpc` from rest of codebase, to prevent `rpc`s to breaking due to refactoring.


pmnoxx:
We should refactor the code to separate `jsonrpc` from rest of the code base.

Steps 1: remove usage of serde from crates other than `jsonrpc`. `serde` is used mainly for `Serialize`/`Deserialize` of messages. `serde` is used in following creates, simply because `jsonrpc` needs to serialize to json, to read/write messages to network. By removing `serde` from those crates, we will be one step closer to cleaning `jsonrpc`
- [x] #5515
- [x] #5508
- [x] removing `serde` from `near-primitives` (only uses `serde` with `test_fetures`)
- [x] #5546

Step 2:
- reorganize `jsonrpc` and cleanup api, make sure it contains all types used.

See https://github.com/near/nearcore/pull/5428#pullrequestreview-813563252

matklad:
I'd make this even more fine-grained: conceptually, `jsonrpc` consists of two parts: 

* definition of data types and routes that comprise our public, stable API
* implementation of actix-based client and server

It's important to separate data type definitions into a separate crate. That way, if someone wants use the API using a different networking stack (eg, based on blocking IO), they get to re-use definitions of the JSON API iteslf, without brining in tokio and actix ecosystems. 

I think the end state here is that we publish `near_rpc_types` crate which:

* doesn't depend on random nearcore crates
* doesn't depend on any "runtime" crates like actix, hyper, tokio, etc.
* doesn't have a lot of dependencies in general, mostly just serde 
* defines just the data types
* has a stable API which we are committed to not break
* versioned properly (eg, `1.0.142`, like serde)

My suggested path there: 

* pick `near-jsonrpc-primitives` crate as the one that would become `near_rpc_types`
* move serde usage from other crates into `near-jsonrpc-primitives`
* remove extra deps from `near-jsonrpc-primitives`
* rename the crate to have a short, catchy public name (`nrpc_types`? The call site would be `nrpc_types::Block`, etc)
* publish `0.1` to crates.io
* audit API *very* *very* carefully and publish `1.0` with the commitment to *not* make `2.0` for at least couple of years.

matklad:
Also, from me personally: this is a hugely important bit of work which has multiplicative implications for both how effective are folks working at nearcore itself, as well as how robust is the ecosystem which builds on top of us. Kudos for spearheading the effort @pmnoxx !

matklad:
[...]

PR #6068, opened 2022-01-14
Adding comments to all the database columns

mm-near:
Added comments on what is the row & column type to all the database columns.


pmnoxx (approved):
LGTM

core/store/src/db.rs
```diff
@@ -110,27 +110,60 @@ pub enum DBCol {
...
     ColAccountAnnouncements = 24,
-    /// Next block hashes in the sequence of the canonical chain blocks
+    /// Next block hashes in the sequence of the canonical chain blocks.
+    /// - *Rows*: BlockHash (CryptoHash)
+    /// - *Content type*: next block: BlockHash (CryptoHash)
     ColNextBlockHashes = 25,
-    /// `LightClientBlock`s corresponding to the last final block of each completed epoch
+    /// `LightClientBlock`s corresponding to the last final block of each completed epoch.
+    /// - *Rows*: EpochId (CryptoHash)
+    /// - *Content type*: LightClientBlockView
```
matklad, line 162: This ... feels fishy. My understanding that `*View` types are for RPC, not for storing directly in DB...

matklad (approved):
It almost feels like these comments want to be types: 

```rust
impl DbCol for BlockExtraCol {
  const ID: u32 = 20;
  type Key = BlockHash;
  type Value = BlockExtra;
}
```

Definitely shouldn't be doing that right now, but potentially reifying columns as types could help us down the line.

core/store/src/db.rs
mm-near: I found this code:
https://github.com/near/nearcore/blob/master/chain/chain/src/store.rs#L2603

And this type is BorshSerializable:
https://github.com/near/nearcore/blob/master/core/primitives/src/views.rs#L1433

matklad: Yeah, it's not a statement about this particular change, it's a statement about pre-existing state of code. 

My understanding is that `*View`-types are intended to be `API` types we use in JSON RPC. So, eg, we use [`Action`](https://github.com/near/nearcore/blob/518685af6a67838e27bf727fd2a677830133f7ad/core/primitives/src/transaction.rs#L48-L66) as internal representation of what we store in the database, and we use [`ActionView`](https://github.com/near/nearcore/blob/518685af6a67838e27bf727fd2a677830133f7ad/core/primitives/src/views.rs#L755-L794) for the RPC (which is reachable from the [`chunk`](https://github.com/near/nearcore/blob/518685af6a67838e27bf727fd2a677830133f7ad/chain/jsonrpc/src/lib.rs#L902-L912) method). 

Some of the `*View` types are not only JSON-encodable, but also borsh-encodable. My understanding that this bottoms out in [`RoutedMessageBody::FinalExecutionOutcomeView`](https://github.com/near/nearcore/blob/518685af6a67838e27bf727fd2a677830133f7ad/chain/network-primitives/src/network_protocol.rs#L181). I am not sure if that should've use `*View` struct or not: it feels weird that [here](https://github.com/near/nearcore/blob/518685af6a67838e27bf727fd2a677830133f7ad/chain/network-primitives/src/types.rs#L325-L330) we have both `*View` and not `*View` struct. So, my current understanding is "some view types are borsh because we use then in p2p messages, which I am not sure why we do". 

What I am noticing here is that there's also at least one `*View` type which we store in the database. And that's surprising, because that means that we store *both* , eg, `ValidatorStakeView` (through `ColEpochLightClientBlocks`) and `ValidatorStake` (through `ColEpochValidatorInfo`) in the database.

Overall, it seems that the codebase as a whole is confused about what `*View` really means, and that some things are miss-classified as `*View`.

matklad: cc https://github.com/near/nearcore/issues/5516

PR #6073, opened 2022-01-15
re: Make `rosetta-rpc` depend on `near-network-primitives`

pmnoxx:
Currently `rosetta-rpc` depends on `near-network`, because it needs access to `NetworkClientMessages` and `NetworkClientResponses`.

We can move those two types to `near-network-primitives` from `near-network`. This will allow us to remove dependency on `near-network`.

There are a lot of simplifications that can be done like this. However, let's do then one step at a time.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3126,7 +3126,7 @@ dependencies = [
  "near-client",
  "near-client-primitives",
  "near-crypto",
- "near-network",
+ "near-network-primitives",
  "near-primitives",
  "paperclip",
  "serde",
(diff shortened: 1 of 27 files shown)


pmnoxx:
```
@chefsale @khorolets: chain/jsonrpc/
@frol @mina86: chain/rosetta-rpc/
```
@chefsale @frol @khorolets @mina86 @mm-near Can you take a look?

mina86 (approved):
I’m honestly not sure if this is worth the effort. It’s not like we have a separate Rosetta binary which can be built without near-network.

PR #6074, opened 2022-01-15
re: Refactor `near-network::private_actix` 

pmnoxx:
`Actix` uses messages for communication between actors. All messages need to implement `Message` or `MessageResponse` traits to be used in actix.

We can make code cleaner, by deriving those traits, instead of implementing them directly. We can start by fixing `near-network::private_actix` 

See https://github.com/near/nearcore/issues/6075

### DIFF
--- a/chain/network/src/private_actix.rs
+++ b/chain/network/src/private_actix.rs
@@ -18,7 +18,8 @@ use std::sync::{Arc, Mutex};
 /// Actor message which asks `PeerManagerActor` to register peer.
 /// Returns `RegisterPeerResult` with `Accepted` if connection should be kept
 /// or a reject response otherwise.
-#[derive(Clone, Debug)]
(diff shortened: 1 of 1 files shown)


chain/network/src/private_actix.rs
```diff
@@ -73,30 +70,15 @@ pub struct Unregister {
...
+#[derive(Message, Clone, Debug)]
+#[rtype(result = "PeerRequestResult")]
 pub struct PeersRequest {}
 
-impl Message for PeersRequest {
-    type Result = PeerRequestResult;
-}
-
-#[derive(Debug)]
+#[derive(MessageResponse, Debug)]
```
mm-near, line 77: does this 'derive MessageResponse' do the same thing as the 'handle' before?

mm-near (changes requested):

mm-near (approved):

PR #6078, opened 2022-01-15
re: `near-network` elide casts to `usize`

pmnoxx:
There are a few places in `near-network`, where we have to cast types to `usize`. We can simplify the code by keeping certain types as `usize`.

### DIFF
--- a/chain/jsonrpc-primitives/src/types/network_info.rs
+++ b/chain/jsonrpc-primitives/src/types/network_info.rs
@@ -22,7 +22,7 @@ pub struct RpcKnownProducer {
 pub struct RpcNetworkInfoResponse {
     pub active_peers: Vec<RpcPeerInfo>,
     pub num_active_peers: usize,
-    pub peer_max_count: u32,
+    pub peer_max_count: usize,
     pub sent_bytes_per_sec: u64,
     pub received_bytes_per_sec: u64,
     /// Accounts of known block and chunk producers from routing table.
--- a/nearcore/src/config.rs
+++ b/nearcore/src/config.rs
@@ -154,19 +154,19 @@ pub const MAX_INFLATION_RATE: Rational = Rational::new_raw(1, 20);
 pub const PROTOCOL_UPGRADE_STAKE_THRESHOLD: Rational = Rational::new_raw(4, 5);
 
 /// Maximum number of active peers. Hard limit.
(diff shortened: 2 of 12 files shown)


pmnoxx:
```
@mm-near: chain/network/ chain/network-primitives/ nearcore/
@chefsale @khorolets: chain/client-primitives/ chain/jsonrpc-primitives/
```
@chefsale @khorolets @mm-near Can you take a look?

chain/jsonrpc-primitives/src/types/network_info.rs
```diff
@@ -22,7 +22,7 @@ pub struct RpcKnownProducer {
 pub struct RpcNetworkInfoResponse {
     pub active_peers: Vec<RpcPeerInfo>,
     pub num_active_peers: usize,
-    pub peer_max_count: u32,
+    pub peer_max_count: usize,
```
mm-near, line 25: is this change backwards compatible? (I remember that json and u64 had issues when serializing - as you had to use strings in JSON).

nearcore/src/config.rs
```diff
@@ -203,27 +203,27 @@ pub struct Network {
     pub boot_nodes: String,
     /// Maximum number of active peers. Hard limit.
     #[serde(default = "default_max_num_peers")]
-    pub max_num_peers: u32,
+    pub max_num_peers: usize,
```
mm-near, line 206: same here - is it backwards compatible?

mm-near (changes requested):

chain/jsonrpc-primitives/src/types/network_info.rs
pmnoxx: Yes, the change is backward compatible.

The only difference is that before, we wouldn't decode some specially prepared JSON messages. Instead, the process will fail on "validate()" method.

The main reason why you don't want to use `usize`, `u64` , is when you store values greater than `i64::MAX`, because those can't be represented by java. This matters when you store, for example 8 bytes hashes. However, that's not the case for us. So there is nothing to worry about.

nearcore/src/config.rs
pmnoxx: Same as above.

mm-near (approved):

miraclx (approved):

PR #6079, opened 2022-01-15
re: `near-network` Change `Ordering` to `Relaxed`

pmnoxx:
`near-network` is not using any advanced synchronizations techniques. Therefore, we should use `Ordering::Relaxed` by default. This will improve code readability.

### DIFF
--- a/chain/network/src/peer/peer_actor.rs
+++ b/chain/network/src/peer/peer_actor.rs
@@ -556,7 +556,7 @@ impl PeerActor {
     /// If so, drop the transaction.
     fn should_we_drop_msg_without_decoding(&self, msg: &[u8]) -> bool {
         if utils::is_forward_transaction(msg).unwrap_or(false) {
-            let r = self.txns_since_last_block.load(Ordering::Acquire);
+            let r = self.txns_since_last_block.load(Ordering::Relaxed);
             if r > MAX_TRANSACTIONS_PER_BLOCK_MESSAGE {
                 return true;
             }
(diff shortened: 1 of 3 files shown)


pmnoxx:
> If we don't want to think about orderings, a `Mutex` should be used, rather than defaulting the ordering to something specific. If there's no contention, `Mutex` is going to be reasonably cheap.
> 
> Otherwise, I think defaulting to the _weakest_ ordering is a mistake. Any sort of default ordering in context of atomics is a terrible practice, but defaulting to the weakest one makes the code most likely to be incorrect. This is especially relevant in context of us looking at supporting Aarch64 natively in the future.

I think we shouldn't be using Mutex. It's my an order of magnitude slower then using atomics.

I think it's not worth to think about it. I would rather use `Ordering::SeqCst` everywhere. First of all, there is no performance different for it on x86 platform, and i haven't seen any compelling benchmark of how big performance impact it has on `Aarch64`.

@matklad @nagisa Anyway, we should agree on one default `Ordering` and use it in our code. We rarely use atomics, so any performance optimizations, don't matter that much. Besides, we should switch to `parking_lot::Mutex`, instead of `std:sync::Mutex`, which we use a lot, and has much higher performance impact, than usage of atomics.

matklad:
Yeah, to clarify, I don't believe we have any meaningful atomics (as in unsafe lock-free data structures), we are mostly talking about counters here. I think there are two reasonable stances here: 

* use `SeqCst` as that's the strongest ordering and has the highest chance of making the code correct
* avoid probabilistic reasoning about correctness, and use the weakest ordering which makes the code correct. In practice, that means using either Acq/Rel or Relaxed, as `SeqCst` is a really tricky thing, which is needed only for some rather fancy data structures. (curious: what are real world examples of **safe** code where one needs something stronger than `Relaxed`?)

I personally am a proponent of the second approach, but the first one is more agreeable, so it might make more sense for a large-ish team like ours. 

We definitelly should refactor `txns_since_last_block` to not use `Acq/Rel`, as that contradicts both positions :)

mm-near (approved):

PR #6080, opened 2022-01-16
re: chore: Fix unused `msg` paramater warnings in `near-network`

pmnoxx:
By making improvements to `near-performance-metrics-macros`, we can change the logic of `perf` derive not to use `msg` argument.
This will allow us to use `perf` for any function, even ones that don't contain `msg` argument.
Also, we will be able to fix warnings about unused `msg` agument.


pmnoxx:
```
@mm-near: chain/network/ utils
```
@mm-near Can you take a look?

mm-near (approved):

PR #6083, opened 2022-01-16
feat: Optimize `find_route_from_peer_id` in `near-network`

pmnoxx:
`find_route_from_peer_id` can be written without doing extra `clone`, memory allocations `Vec` nor `unwraps`.

### DIFF
--- a/chain/network/src/routing/routing_table_view.rs
+++ b/chain/network/src/routing/routing_table_view.rs
@@ -103,11 +103,10 @@ impl RoutingTableView {
                     // max nonce - threshold.
                     self.route_nonce.put(
                         next_hop.clone(),
-                        // TODO, should be max(min_v + 1, max_v.saturating_sub(ROUND_ROBIN_MAX_NONCE_DIFFERENCE_ALLOWED))
                         std::cmp::max(
-                            min_v,
+                            min_v + 1,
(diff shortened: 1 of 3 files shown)


chain/network/src/routing/routing_table_view.rs
```diff
@@ -88,31 +89,33 @@ impl RoutingTableView {
...
+                None => Err(FindRouteError::Disconnected),
+                // Neighbor with minimum and maximum nonce respectively.
+                Some(((min_v, next_hop), (max_v, _))) => {
+                    // Strategy similar to Round Robin. Select node with least nonce and send it. Increase its
+                    // nonce by one. Additionally if the difference between the highest nonce and the lowest
+                    // nonce is greater than some threshold increase the lowest nonce to be at least
+                    // max nonce - threshold.
+                    self.route_nonce.put(
+                        next_hop.clone(),
+                        // TODO, should be max(min_v + 1, max_v.saturating_sub(ROUND_ROBIN_MAX_NONCE_DIFFERENCE_ALLOWED))
```
mm-near, line 109: does it matter? or should we just update the comment in line 106 and leave it as is?

mm-near (approved):

chain/network/src/routing/routing_table_view.rs
pmnoxx: @mm-near It doesn't matter. I removed the comment.

PR #6092, opened 2022-01-17
feat: RocksDB gas estimation

jakmeier:
Gas estimation functions for raw RocksDB inertion and read operations.
The new options with a rdb- prefix can be used to change DB settings for these estimations.

Test plan
---------
To run the two estimations:
```
cargo run --release -p runtime-params-estimator --features required -- --metric time --costs RocksDbInsertValueByte,RocksDbReadValueByte
```


runtime/runtime-params-estimator/src/rocksdb.rs
```diff
@@ -0,0 +1,311 @@
...
+    /// (`RocksDb*` estimations only)
+    #[clap(long, name = "rdb-block-cache", long)]
+    pub block_cache: bool,
+    /// Print RocksDB debug output where available
+    #[clap(skip)]
+    pub debug_rocksdb: bool,
+    /// Pseudo-random input data dump
+    /// (`RocksDb*` estimations only)
+    #[clap(long, name = "rdb-pr-data-path", long)]
+    pub pr_data_path: Option<PathBuf>,
```
matklad, line 52: nit: this dosen't *have* to be PR, so maybe a better name is possible for the flag. Not that it matters. 

pr is also not the most immediatelly obvious prefix.

runtime/runtime-params-estimator/src/lib.rs
```diff
@@ -156,6 +159,8 @@ static ALL_COSTS: &[(Cost, fn(&mut EstimatorContext) -> GasCost)] = &[
     (Cost::TouchingTrieNode, touching_trie_node),
     (Cost::GasMeteringBase, gas_metering_base),
     (Cost::GasMeteringOp, gas_metering_op),
+    (Cost::RocksDbInsertValueByte, rocks_db_insert_value_byte),
+    (Cost::RocksDbReadValueByte, rocks_db_read_value_byte),
```
matklad, line 163:
```suggestion
    (Cost::RocksDbInsertValueByte, rocks::db_insert_value_byte),
    (Cost::RocksDbReadValueByte, rocks::db_read_value_byte),
```

For "tablish, rectangular" code, qualified imports usually work better.

runtime/runtime-params-estimator/src/rocksdb.rs
```diff
@@ -0,0 +1,311 @@
...
+    let db = new_test_db(&tmp_dir, &data, &db_config);
+
+    if db_config.debug_rocksdb {
+        println!("# {:?}", db_config);
+        println!("# After setup / before measurement:");
+        print_levels_info(&db);
+    }
+
+    let mut prng: XorShiftRng = rand::SeedableRng::seed_from_u64(SETUP_PRANDOM_SEED);
+    let mut keys: Vec<usize> = (0..db_config.setup_insertions).map(|_| prng.gen()).collect();
```
matklad, line 151: `iter::repat_with(|| prng.gen()).take(db_config.setup_insertions)` is another way to write this.

runtime/runtime-params-estimator/src/rocksdb.rs
```diff
@@ -0,0 +1,311 @@
...
+    } else {
+        // give it another shuffle to make lookup order different from insertion order
+        keys.shuffle(&mut prng);
+    }
+
+    let gas_counter = GasCost::measure(config.metric);
+
+    for i in 0..db_config.op_count {
+        let key = keys[i as usize % keys.len()];
+        db.get(&key.to_string()).unwrap();
```
matklad, line 163: We *might* want to push `.to_string()` outside of the benched loop, though it sholudn't matter at this scale I hope.

runtime/runtime-params-estimator/src/rocksdb.rs
```diff
@@ -0,0 +1,311 @@
...
+            .write(true)
+            .append(true)
+            .create(true)
+            .open("names-to-stats.txt")
+            .unwrap();
+        let stats_num = std::io::BufReader::new(&stats_file).lines().count();
+        let data_dump_path = format!("data_dump_{:<04}.bin", stats_num);
+
+        let mut dump = std::fs::File::create(&data_dump_path).unwrap();
+        dump.write_all(&data).unwrap();
```
matklad, line 189: std::fs::write ?

runtime/runtime-params-estimator/src/rocksdb.rs
```diff
@@ -0,0 +1,311 @@
...
+        db.flush().expect("Flush failed");
+    }
+    if force_compaction {
+        db.compact_range::<&[u8], &[u8]>(None, None);
+    }
+}
+
+fn input_data(db_config: &RocksDBTestConfig, data_size: usize) -> Vec<u8> {
+    let mut data = vec![0u8; data_size];
+    if let Some(path) = db_config.pr_data_path.as_ref() {
```
matklad, line 255:
```suggestion
    if let Some(path) = &db_config.pr_data_path {
```

runtime/runtime-params-estimator/src/rocksdb.rs
```diff
@@ -0,0 +1,311 @@
...
+        let mut input = std::fs::File::open(path).unwrap();
+        input.read_exact(&mut data).unwrap();
```
matklad, line 257: Do we want to assert if file contains *more* data?

matklad (approved):

runtime/runtime-params-estimator/src/lib.rs
jakmeier: These are actually two local functions that delegate to the functions in RocksDB.
I feel the only reason to keep the indirection is that the inner function can use the `Config` but not `EstimatorContext`, which makes it very clear that it is separate from the usual testbed and DB.
But no strong opinion on my side. I can also remove the indirection.

runtime/runtime-params-estimator/src/rocksdb.rs
jakmeier: I had the same thought, allocating memory inside benchmarks is usually not good.
OTOH, looping through a huge buffer of keys is also not ideal as it affect caches.
So, I decided to go with the simpler approach.

runtime/runtime-params-estimator/src/rocksdb.rs
jakmeier: Ah yes, that's a bit nicer here :)

runtime/runtime-params-estimator/src/lib.rs
matklad: Ah, that makes total sense to me know, yeah, this is indeed better.

runtime/runtime-params-estimator/src/rocksdb.rs
jakmeier: Right! I completely forgot that exists.
Also, while at it, I extracted some duplicated code into `fn backup_input_data(data: &[u8])`

runtime/runtime-params-estimator/src/rocksdb.rs
```diff
@@ -0,0 +1,303 @@
...
+        db.compact_range::<&[u8], &[u8]>(None, None);
+    }
+}
+
+fn input_data(db_config: &RocksDBTestConfig, data_size: usize) -> Vec<u8> {
+    let mut data = vec![0u8; data_size];
+    if let Some(path) = &db_config.input_data_path {
+        let mut input = std::fs::File::open(path).unwrap();
+        input.read_exact(&mut data).unwrap();
+        assert_eq!(input.seek(SeekFrom::End(0)).unwrap(), 0, "Provided input file has wrong size");
```
matklad, line 232: at this point, `std::fs::read` might be eashier.

matklad (approved):

runtime/runtime-params-estimator/src/rocksdb.rs
jakmeier: Now that I think about it, yes, we should do that.
I've added a check for that, using `input.seek(SeekFrom::End(0))` to read the file size. That's the best way I know on stable

runtime/runtime-params-estimator/src/rocksdb.rs
jakmeier: Ah okay, I see what you mean. Yes that should be easier

runtime/runtime-params-estimator/src/rocksdb.rs
jakmeier: Changes this to `rdb-input-data-path` which personally I like much better but it is still not perfect (names are hard)

PR #6094, opened 2022-01-17
make near-vm-runner build /w --no-default-features

nagisa:
I wanted to utilize the features here to disable and not deal with
wasmer0 in some experiments, but was surprised to find out the code
simply does not compile if the feature is disabled.

### DIFF
--- a/runtime/near-vm-runner/src/cache.rs
+++ b/runtime/near-vm-runner/src/cache.rs
@@ -1,9 +1,6 @@
 use crate::errors::ContractPrecompilatonResult;
 use crate::prepare;
 use crate::vm_kind::VMKind;
-use crate::wasmer2_runner::wasmer2_vm_hash;
-use crate::wasmer_runner::wasmer0_vm_hash;
-use crate::wasmtime_runner::wasmtime_vm_hash;
 use borsh::{BorshDeserialize, BorshSerialize};
 use near_primitives::contract::ContractCode;
 use near_primitives::hash::CryptoHash;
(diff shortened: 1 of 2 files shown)


runtime/near-vm-runner/src/cache.rs
```diff
@@ -111,12 +117,12 @@ impl fmt::Debug for MockCompiledContractCache {
 #[cfg(not(feature = "no_cache"))]
 const CACHE_SIZE: usize = 128;
 
-#[cfg(not(feature = "no_cache"))]
+#[cfg(all(feature = "wasmer0_vm", not(feature = "no_cache")))]
 static WASMER_CACHE: once_cell::sync::Lazy<
```
matklad, line 121: Can we move these inside the corresponding modules?

runtime/near-vm-runner/src/cache.rs
```diff
@@ -111,12 +117,12 @@ impl fmt::Debug for MockCompiledContractCache {
 #[cfg(not(feature = "no_cache"))]
 const CACHE_SIZE: usize = 128;
```
matklad, line 118: Thias *might* give unused if both are disabled?

matklad (approved):
Heh, I think this is now significantly less painful to achieve than it used to be, thanks to your Runtime refactor!

One obvious question is "should we enforce this on CI"? I'd say at this point, as this requires an extra CI job, it probably isn't worth it. My bet is that, aftre wasmtime upgrade and wasmer0 removal, we should work on removing default features, which would make extra CI unnecessary.

runtime/near-vm-runner/src/cache.rs
nagisa: There are a fair amount of unused lints when none of the implementations are enabled, including for this item. I'm not entirely sure it is worth the effort to go and clean up everything here, as I only really needed this for some hacking with aurora contracts.

I don't anticipate disabling all the runtimes will be all that useful thing to need.

PR #6096, opened 2022-01-17
refactor: Split VMLogic and MemoryLike

Ekleog:
This gets us ready for an (upcoming) PR that will bump wasmtime, as such
a PR will need a different &dyn MemoryLike depending on whether we're in
the host before calling into wasm, or in a host function called by wasm
code.

For now, for convenience VMLogicPlusMemory has been added. Future work
could consider removing it, though wasmer code will probably keep
needing it for now.

-----

I'll undraft the PR once the wasmtime bump PR will be ready and confirmed to work well with this refactor

### DIFF
--- a/runtime/near-vm-runner/src/imports.rs
+++ b/runtime/near-vm-runner/src/imports.rs
@@ -222,7 +222,7 @@ imports! {
 #[cfg(feature = "wasmer0_vm")]
 pub(crate) mod wasmer {
     use super::str_eq;
-    use near_vm_logic::{ProtocolVersion, VMLogic, VMLogicError};
+    use near_vm_logic::{ProtocolVersion, VMLogicError, VMLogicPlusMemory};
     use std::ffi::c_void;
 
     #[derive(Clone, Copy)]
@@ -232,7 +232,7 @@ pub(crate) mod wasmer {
 
     pub(crate) fn build(
         memory: wasmer_runtime::memory::Memory,
-        logic: &mut VMLogic<'_>,
+        logic: &mut VMLogicPlusMemory<'_>,
         protocol_version: ProtocolVersion,
     ) -> wasmer_runtime::ImportObject {
         let raw_ptr = logic as *mut _ as *mut c_void;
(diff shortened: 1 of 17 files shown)


runtime/near-vm-runner/src/imports.rs
```diff
@@ -455,13 +455,14 @@ pub(crate) mod wasmtime {
         };
     }
 
-    pub(crate) fn link(
+    pub(crate) unsafe fn link(
         linker: &mut wasmtime::Linker,
         memory: wasmtime::Memory,
-        raw_logic: *mut c_void,
+        raw_logic_plus_memory: *mut c_void,
```
nagisa, line 461: Preexisting, but I would argue that the context should be passed in as a proper type here, and that the thread local should own the entire `VMLogicPlusMemory` rather than doing this raw pointer contraption.

If possible.

matklad:
Yeah, would love to see this in the context of wasmtime upgrade. 

On the one hand, yeah, passing memory into logic as an argument makes sense, because memory *can* be different for every separate host function call. 

On the other hand, one can think of `VMLogic` as one-shot object used to a specific host function which needs to be reset in some form for each call.

nagisa (commented):
One thing that comes to mind immediately is that we may very much like to not pass memory to every host function as is done now. Definitely feels like there may be some further improvement lurking around this area of the code.

runtime/near-vm-runner/src/imports.rs
Ekleog: I did it this way just to get the refactor out of the way, with https://github.com/near/nearcore/pull/6127 the whole unsafe thing gets removed away, as well as wasmtime's usage of VMLogicPlusMemory :)

Ekleog:
The wasmtime upgrade is now available at https://github.com/near/nearcore/pull/6127 :)

Also yes, there are definitely still improvements that could be made, for now in this refactor I just went the fewest-changes way to get to a data format that let us upgrade wasmtime, as I don't feel confident enough yet in my overall understanding of the code to know which specific refactor would be the best one to pick

Ekleog:
Closing for now, I'll reopen a version focused on erasing unsafe from wasmer2 and wasmtime backends

PR #6097, opened 2022-01-17
tests: fix flaky integration test

marcelo-gonzalez:
the download_file() test might fail if download_file() runs before
the HTTP server gets a chance to start. Fix it by waiting until the server
is already listening to call download_file(). This requires moving the
tokio runtime block_on() call to a wrapper function so we don't have problems
starting a runtime inside a runtime in the test code

### DIFF
--- a/integration-tests/src/tests/test_download_file.rs
+++ b/integration-tests/src/tests/test_download_file.rs
@@ -5,8 +5,8 @@ use std::fs::File;
 use std::io::Read;
 use std::net::SocketAddr;
 
-#[test]
-fn test_file_download() {
+#[tokio::test]
+async fn test_file_download() {
     let port = portpicker::pick_unused_port().expect("No ports free");
 
     async fn handle_request(_req: Request<Body>) -> Result<Response<Body>, Infallible> {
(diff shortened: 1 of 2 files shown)


marcelo-gonzalez:
flaky run is here: https://buildkite.com/nearprotocol/nearcore/builds/12012#_

I guess I don't really have any more evidence that this is what caused it, but I'm pretty sure it is, because if you add a one-second sleep right before starting the server in the current test, you get the same conn refused error

integration-tests/src/tests/test_download_file.rs
```diff
@@ -5,33 +5,39 @@ use std::fs::File;
...
+    rx.await.unwrap();
+
     nearcore::config::download_file(
         &format!("http://localhost:{}", port),
         tmp_downloaded_file_path,
     )
+    .await
     .unwrap();
 
     let mut downloaded_file = File::open(tmp_downloaded_file_path).unwrap();
```
matklad, line 43: could use `std::fs::read` here

integration-tests/src/tests/test_download_file.rs
```diff
@@ -5,33 +5,39 @@ use std::fs::File;
...
-        tokio::runtime::Runtime::new().unwrap().block_on(async move {
-            let addr = SocketAddr::from(([127, 0, 0, 1], port));
-            let make_svc =
-                make_service_fn(|_conn| async { Ok::<_, Infallible>(service_fn(handle_request)) });
-            let server = Server::bind(&addr).serve(make_svc);
-            if let Err(e) = server.await {
-                eprintln!("server error: {}", e);
-            }
-        });
+    let (tx, rx) = tokio::sync::oneshot::channel();
```
matklad, line 17: Doesn't matter in this case, but in general the rule-of-thumb primitive to use for synchronization in tests is barrier: https://docs.rs/tokio/latest/tokio/sync/struct.Barrier.html. It more directly expresses "wait util the coordinating threads have finished setup work", and generalizes for more than one thread.

matklad (approved):
:heart: always great to fix flaky tests!

Also: does this need to be an `integraton-test`? It seems that this doesn't depend on any of the shared testing machinery, and can be a unit-test in the `nearcore` crate.

integration-tests/src/tests/test_download_file.rs
marcelo-gonzalez: ah cool, that does seem more clear. might as well update it

marcelo-gonzalez:
> Also: does this need to be an `integraton-test`? It seems that this doesn't depend on any of the shared testing machinery, and can be a unit-test in the `nearcore` crate.

hmm good point. Could maybe be good to do that in another PR

mm-near (approved):

PR #6099, opened 2022-01-18
chain: fix and refactor assign_shards function

mina86:
Refactor assign_shards by eliminating the special case handling
situation where chunk producers need to be assigned to multiple shards
to satisfy the minimum number of validators requirement.  This fixes
the issue of assign_shards function panicking in certain situations.

Fixes: https://github.com/near/nearcore/issues/5932


mina86:
@mzhangmzz, @mm-near friendly ping

chain/epoch_manager/src/shard_assignment.rs
```diff
@@ -141,10 +104,17 @@ fn assign_with_possible_repeats<T: HasStake + Eq, I: Iterator<Item = T>>(
                 // assign a chunk producer right now.
                 let (least_validator_count, shard_stake, shard_id) =
                     shard_index.pop().expect("shard_index should never be empty");
-                if result[usize::try_from(shard_id).unwrap()].contains(&cp) {
```
mzhangmzz, line 144: I'm still confused about the logic here. It's hard for me to reason through that this algorithm for all cases.

Here is a simpler proposal. In this same loop here, we keep looking for the shard that doesn't have cp yet. When we find that shard, we assign cp to that shard and move the last validator in that shard to the current shard. This way, we always find a shard for every cp in the iterator and increase the validator count for the shard that has the least validators.

mina86: > When we find that shard, we assign cp to that shard and move the last validator in that shard to the current shard.

I don’t see how we can be sure that the last validator of that shard can be added to current shard.

The way I think about the algorithm is that we loop over all chunk producers (possibly repeating them) and for each we find the first shard which has fewer than min_validators_per_shard validators and cp is not assigned to yet. If there is no such shard, we skip the chunk producer. Once all shards get the minimum number of validators, we end the algorithm.

mzhangmzz: If we skip the chunk producer, how do you ensure all shards will get the minimum number of validators at the end once cp_iter is exhausted. Right now, it holds because num_chunk_producers >= min_validators_per_shard and we don't skip chunk producers. 

How about, when we find that shard, we assign cp to that shard and find a validator in that shard that has is not in the current shard. If we can't find such validator, then it means the other shard has the same number of validators as the current shard, then we can just assign cp to that shard.

mina86: We will only skip chunk producers if num_chunk_producers > min_validators_per_shard in which case it’s alright because we have chunk producers to spare. If we have exactly num_chunk_producers == min_validators_per_shard than first cp will get assigned to first shard, then to second, then to third etc. The second cp will get assigned to second shard, than to first, than to third etc. And so on.

mzhangmzz: I don't see how when num_chunk_producers == min_validators_per_shard, the algorithm will assign the first cp to the first shard and then to second. It will assign first cp to first, second cp to second, until it loops the cps again and again. But in the following loops, where the cp is assigned to is hard to say, since it depends on whether it was already assigned to the shard. For example, let's say there is 5 validators with the same stake and 5 shards, and min_validators_per_shard is 2. The length of cp_iter will be 10. Then, after the first 5 iterations, the shard assignment will become 
[0], [1], [2], [3], [4]. Then 0 will be assigned to shard [1] because it's already in shard 0, 1 will be assigned to shard 0, 2 assigned to shard 3, 3 assigned to shard 2 (this may depends on how the heap order identical values, so this counter-example may not exactly work, but it showed my point that it is not trivial to prove the correctness of this algorithm). When the algorithm is trying to assign 4, the shards look like, [0,1], [0,1], [2,3], [2,3], [4], so it has to drop 4. But now we exhausted cp_iter and couldn't fill every shard with the minimum validators requirement.

mina86: Right, yes, I was vastly simplifying. 

> The length of cp_iter will be 10.

`cp_iter` is infinite. Once the algorithm drops 4 it’ll move back to 0 and assign it to the last shard. `cp_iter` being limited in length was pretty much the bug in the algorithm. Now, it’s true that this will affect balance of how many shards each validator will be assigned to, but all shards will get their assignments.

The question is how much we care about optimising this case. In general this is an NP problem and we’re already doing heuristics to perform the assignments.

chain/epoch_manager/src/shard_assignment.rs
```diff
@@ -214,17 +190,9 @@ mod tests {
     /// the code will assign validator 81 to it.  In the last step, shard 0 will
     /// have only one validator so the code will try to assign validator 100 to
     /// it.  However, that validator is already assigned to that shard.
```
mm-near, line 192: please update this comment with the results of your changes.

chain/epoch_manager/src/shard_assignment.rs
mm-near: agreed, that if the cp_iter is infinite, we'll find the assignment - thanks for a great example @mzhangmzz and for the explanation @mina86 .

Maybe let's add a test case with the this example.

mm-near (approved):

chain/epoch_manager/src/shard_assignment.rs
mina86: The test case that this commit is fixing checks such a scenario.

mzhangmzz (approved):

mzhangmzz:
LGTM. Thanks for the fix! Sorry for the delay of approval.

PR #6105, opened 2022-01-18
feat: benchmark for iterating trie

Longarithm:
Add benchmark for iterating trie items because work on https://github.com/near/nearcore/pull/5823 showed that it requires unexpectedly long time.

Though moving from one item to other requires some ascending and descending steps, I would still expect it to be faster than 50 items per second. In particular, this is our current dumping state speed.

I find this form useful for writing other benchmarks and running this one from time to time.

## Test plan

Existing tests.


nearcore/benches/store.rs
```diff
@@ -0,0 +1,78 @@
...
+/// TODO: make separate bench runs independent. As of 18/01/2022, first run gives speed of 50 items per second, and all
+/// next runs give ~ 30k items per second.
+fn read_trie_items(bench: &mut Bencher, num_trie_items: usize, shard_id: usize) {
+    init_integration_logger();
+    let home_dir = get_default_home();
+    let near_config = load_config(&home_dir, GenesisValidationMode::UnsafeFast);
+
+    bench.iter(move || {
+        tracing::info!(target: "neard", "{:?}", home_dir);
+        let store = create_store(&get_store_path(&home_dir));
```
mm-near, line 23: can you add the option to use the readonly store? (I wonder if it is faster then -as it doesn't do the compactions, right?)

can be done in separate pr

mm-near (approved):

nearcore/benches/store.rs
Longarithm: My experiment showed that performance is the same.

PR #6106, opened 2022-01-18
Betanet loadtest using mocknet scripts

nikurt:
Replace javascript betanet loadtest with an adapted mocknet script.
This script runs on one of the betanet RPC nodes, generating 1.5 tps.

Next step is to let the mocknet scripts submit transactions asynchronously. This will remove the bottleneck and let us generate an arbitrary amount of traffic.


pytest/tests/mocknet/load_test_betanet.py
```diff
@@ -0,0 +1,257 @@
...
+    tx_res = node_account.send_call_contract_raw_tx(node_account.key.account_id,
+                                                    'ft_transfer',
+                                                    s.encode('utf-8'), 1)
+    logger.info(
+        f'{node_account.key.account_id} ft_transfer to {account.key.account_id} {tx_res}'
+    )
+
+
+def get_test_accounts_from_args(argv):
+    assert (len(argv) == 7)
```
mina86, line 189: Superfluous parenthesis.

mina86 (approved):

PR #6110, opened 2022-01-18
Fix NixOS patching implementation in pytest

nagisa:
Turns out my initial testing was spotty in that the binary failed to
patch once, was built from source and then we never hit the patching
code path again.


pytest/lib/branches.py
```diff
@@ -18,7 +18,7 @@
 
 def current_branch():
     return os.environ.get('BUILDKITE_BRANCH') or subprocess.check_output([
-        "git", "rev-parse", "--symbolic-full-name", "--abbrev-ref", "HEAD"
+        "git", "rev-parse", "HEAD"
```
nagisa, line 21: NB: this is a drive-by fix. Previous command would just return a `HEAD` for me in some cases, which would make this script fail to restore the checkout back to the original state.

pytest/lib/branches.py
```diff
@@ -67,11 +66,14 @@ def _compile_binary(branch: str) -> Executables:
...
     subprocess.check_output(['git', 'checkout', str(branch)])
-    subprocess.check_output(['git', 'pull', 'origin', str(branch)])
-    result = _compile_current(branch)
-    subprocess.check_output(['git', 'checkout', prev_branch])
-    if stash_output != b"No local changes to save\n":
-        subprocess.check_output(['git', 'stash', 'pop'])
+    try:
+        subprocess.check_output(['git', 'pull', 'origin', str(branch)])
+        result = _compile_current(branch)
+    except:
```
mina86, line 72: Should this be `finally`? Also, this should be a separate commit?

pytest/lib/branches.py
mina86: Well, yeah, but now without `--symbolic-full-name` this will always return sha of the commit so if someone is at a branch foo, after they run the test they’ll end up at detached head. You are getting `HEAD` when you are on detached head. This needs a different solution. I’ll prepare a commit.

matklad (approved):

Ekleog (approved):
Ideally there'd be one derivation for the build-time deps like patchelf and one for the runtime deps like gcc.cc.lib, so as to avoid that the final binary depends on patchelf, but it'd be some more code and this is test code only so as long as it works it's all good :) (and it's still better than before anyway)

mina86 (approved):

PR #6111, opened 2022-01-18
pytest: print more information about nodes

nagisa:
Without knowing how the script works it is pretty hard to investigate
issues when there is no output and it is not obivious how to run the
command!

Hopefully these changes help a little. They helped me, at least.

### DIFF
--- a/pytest/lib/cluster.py
+++ b/pytest/lib/cluster.py
@@ -412,6 +412,18 @@ class LocalNode(BaseNode):
         if self._start_proxy is not None:
             self._proxy_local_stopped = self._start_proxy()
 
+    def output_logs(self):
+        stdout = pathlib.Path(self.node_dir) / 'stdout'
+        stderr = pathlib.Path(self.node_dir) / 'stderr'
+        if os.environ.get('BUILDKITE'):
+            logger.info('=== stdout: ')
+            logger.info(stdout.read_text('utf-8', 'replace'))
+            logger.info('=== stderr: ')
+            logger.info(stderr.read_text('utf-8', 'replace'))
+        else:
+            logger.info(f'=== stdout: available at {stdout}')
+            logger.info(f'=== stderr: available at {stderr}')
+
     def start(self, *, boot_node: BootNode = None, skip_starting_proxy=False):
         if self._proxy_local_stopped is not None:
             while self._proxy_local_stopped.value != 2:
@@ -445,11 +457,6 @@ class LocalNode(BaseNode):
         except:
             logger.error(
                 '=== failed to start node, rpc does not ready in 10 seconds')
-            if os.environ.get('BUILDKITE'):
-                logger.info('=== stdout: ')
-                logger.info(stdout.read_text('utf-8', 'replace'))
-                logger.info('=== stderr: ')
-                logger.info(stderr.read_text('utf-8', 'replace'))
 
     def kill(self):
         if self._proxy_local_stopped is not None:
@@ -486,6 +493,8 @@ class LocalNode(BaseNode):
         if os.path.exists(target_path) and os.path.isdir(target_path):
             shutil.rmtree(target_path)
         os.rename(self.node_dir, target_path)
+        self.node_dir = target_path
+        self.output_logs()
         self.cleaned = True
 
     def stop_network(self):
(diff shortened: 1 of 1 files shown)


pytest/lib/cluster.py
```diff
@@ -446,10 +457,7 @@ def start(self, *, boot_node: BootNode = None, skip_starting_proxy=False):
             logger.error(
                 '=== failed to start node, rpc does not ready in 10 seconds')
             if os.environ.get('BUILDKITE'):
```
mina86, line 459: You could remove the environment variable check from here.

pytest/lib/cluster.py
```diff
@@ -636,10 +645,6 @@ def spin_up_node(config,
     is_local = config['local']
 
     args = make_boot_nodes_arg(boot_node)
-    logger.info("Starting node %s %s" %
-                (ordinal,
-                 ('with ' + '='.join(args) if args else 'as BOOT NODE')))
-
```
mina86, line 642: Removing this means that we no longer see the ordinal in the logs and also that we don’t get this line when doing remote run. The second point isn’t maybe that crucial since at this point I doubt tests can actually run with remote nodes, but ordinal is something useful in the log.

pytest/lib/cluster.py
```diff
@@ -412,6 +412,16 @@ def start_proxy_if_needed(self):
...
 
+    def output_logs(self):
+        if os.environ.get('BUILDKITE'):
+            logger.info('=== stdout: ')
+            logger.info(self.stdout_name.read_text('utf-8', 'replace'))
+            logger.info('=== stderr: ')
+            logger.info(self.stderr_name.read_text('utf-8', 'replace'))
+        else:
+            logger.info(f'=== stdout: available at {self.stdout_name}')
+            logger.info(f'=== stderr: available at {self.stderr_name}')
```
mina86, line 423: Those two lines aren’t true. When nodes run they use `test#` as their home directory but once they are finished those directories are renamed to `test#_finished`. Those two lines will print the old path.

pytest/lib/cluster.py
nagisa: ordinal is printed immediately after the arguments:

```
[2022-01-18 20:14:22] INFO: Starting node: /var/lib/buildkite-agent/builds/buildkite-i-00824bbff719e81c2-1/nearprotocol/nearcore/target/debug/neard-1.23.1-c0551c84ba9f54509abe05cb1291dbeb33cbd85d --home /tmp/near/backward/test0 run
[2022-01-18 20:14:25] INFO: node 0 started
[2022-01-18 20:14:25] INFO: Starting node: /var/lib/buildkite-agent/builds/buildkite-i-00824bbff719e81c2-1/nearprotocol/nearcore/target/debug/neard-nagisa-pytest-more-info --home /tmp/near/backward/test1 run --boot-nodes He7QeRuwizNEhBioYG3u4DZ8jWXyETiyNzFD3MkTjDMf@127.0.0.1:24577
[2022-01-18 20:14:30] INFO: node 1 started
```

Do you think we need to put the ordinal in the message with arguments still?

pytest/lib/cluster.py
mina86: Using `self.node_dir` doesn’t really change the output. `self.node_dir` points at the `test#` directory.

nagisa: I made sure to set `self.node_dir` to the value of the moved directory and and also moved calls to `output_logs` to after the files are moved to their final destination.

nagisa: (In retrospective I should've made more commits to better demonstrate changes being made…)

mina86: Sorry, missed that change.

mina86 (approved):

pytest/lib/cluster.py
nagisa: My goal was to make it feasible to mostly copy and paste the command locally. Trying to post-process the invocation to e.g. relativize the paths doesn't feel super straightforward, to be honest.

That said my change has another problem I just noticed – it drops the print for GCloudNode. I don't really want to fiddle with this further so I'll just revert.

mina86: > My goal was to make it feasible to mostly copy and paste the command locally.
I’ve never needed that feature but it makes sense.

> That said my change has another problem I just noticed – it drops the print for GCloudNode.

Yeah, that’s what I meant about remote runs. We could choose to ignore this though since we don’t use that feature and it may very well be that it doesn’t actually work anyway.

PR #6112, opened 2022-01-18
pytest: improve getting current branch name

mina86:
`git rev-parse --symbolic-full-name HEAD` returns HEAD when on
detached head.  Since HEAD does not relate to specific commit, it
cannot be used to track previous checked-out commit.  When on detached
head, we want to get the commit sha instead.  Unfortunately, there’s
no single git command which gets us that but combination of `git
symbolic-ref` and `git rev-parse` can make that work.

### DIFF
--- a/pytest/lib/branches.py
+++ b/pytest/lib/branches.py
@@ -16,10 +16,19 @@ _REPO_DIR = pathlib.Path(__file__).resolve().parents[2]
 _OUT_DIR = _REPO_DIR / 'target/debug'
 
 
-def current_branch():
-    return os.environ.get('BUILDKITE_BRANCH') or subprocess.check_output([
-        "git", "rev-parse", "--symbolic-full-name", "--abbrev-ref", "HEAD"
-    ]).strip().decode()
+def current_branch() -> str:
+    """Returns checked out branch name or sha if we’re on detached head."""
+    branch = os.environ.get('BUILDKITE_BRANCH')
+    if branch:
+        return branch
+    try:
+        return subprocess.check_output(
+            ('git', 'symbolic-ref', '--short', '-q', '@')).strip().decode()
+    except subprocess.CalledProcessError as ex:
+        if ex.returncode != 1:
+            raise
+        # We’re on detached HEAD
+    return subprocess.check_output(('git', 'rev-parse', '@')).strip().decode()
 
 
 def __get_latest_deploy(chain_id: str) -> typing.Tuple[str, str]:


pytest/lib/branches.py
```diff
@@ -16,10 +16,19 @@
 _OUT_DIR = _REPO_DIR / 'target/debug'
 
 
-def current_branch():
-    return os.environ.get('BUILDKITE_BRANCH') or subprocess.check_output([
-        "git", "rev-parse", "--symbolic-full-name", "--abbrev-ref", "HEAD"
-    ]).strip().decode()
+def current_branch() -> str:
+    """Returns checked out branch name or sha if we’re on detached head."""
+    branch = os.environ.get('BUILDKITE_BRANCH')
```
mm-near, line 21: can you explain (in comment) - why we treat BUILDKITE in a special way?

mm-near (approved):

pytest/lib/branches.py
mina86: I mean, I wasn’t the one writing the original code and don’t really have any knowledge other than anyone who would read this code would have. There is a risk that whatever I comment will not be correct.

nagisa:
This doesn't work for me locally:

```
fatal: No such ref: @
  File "pytest/tests/sanity/backward_compatible.py", line 129, in <module>
    main()
  File "pytest/tests/sanity/backward_compatible.py", line 25, in main
    executables = branches.prepare_ab_test()
  File "/sb/i/clients/near/nearcore/pytest/lib/branches.py", line 229, in prepare_ab_test
    current = _compile_current(current_branch())
  File "/sb/i/clients/near/nearcore/pytest/lib/branches.py", line 25, in current_branch
    return subprocess.check_output(
  File "/nix/store/hq6mrm0pc6xn6j8y6lm4qcgg9rwmqd8q-python3-3.8.9/lib/python3.8/subprocess.py", line 415, in check_output
    return run(*popenargs, stdout=PIPE, timeout=timeout, check=True,
  File "/nix/store/hq6mrm0pc6xn6j8y6lm4qcgg9rwmqd8q-python3-3.8.9/lib/python3.8/subprocess.py", line 516, in run
    raise CalledProcessError(retcode, process.args,
subprocess.CalledProcessError: Command '('git', 'symbolic-ref', '--short', '-q', '@')' returned non-zero exit status 128.
```

mina86:
Sorry about that. Turns out `symbolic-ref` does not understand `@`. I’m confused why this change passed CI though. https://github.com/near/nearcore/pull/6152

PR #6114, opened 2022-01-18
re: Rename `active_peer` to `connected_peer`

pmnoxx:
There is still some code left in `PeerManagerActor` where we use `active_peer` terminology, but `connected_peer` terminology should be used. Let's fix those remaining places.


pmnoxx:
```
```

mm-near (approved):

PR #6121, opened 2022-01-19
pytest: ensure branch is always reset

nagisa:
Even if compilation fails...

### DIFF
--- a/pytest/lib/branches.py
+++ b/pytest/lib/branches.py
@@ -75,12 +75,14 @@ def _compile_binary(branch: str) -> Executables:
     # TODO: download pre-compiled binary from github for beta/stable?
     prev_branch = current_branch()
     stash_output = subprocess.check_output(['git', 'stash'])
-    subprocess.check_output(['git', 'checkout', str(branch)])
     try:
-        subprocess.check_output(['git', 'pull', 'origin', str(branch)])
-        result = _compile_current(branch)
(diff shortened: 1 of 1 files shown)


mina86 (approved):

pytest/lib/branches.py
```diff
@@ -76,11 +76,13 @@ def _compile_binary(branch: str) -> Executables:
...
-    if stash_output != b"No local changes to save\n":
-        subprocess.check_output(['git', 'stash', 'pop'])
+    try:
+        subprocess.check_output(['git', 'pull', 'origin', str(branch)])
+        result = _compile_current(branch)
+    finally:
+        subprocess.check_output(['git', 'checkout', prev_branch])
+        if stash_output != b"No local changes to save\n":
+            subprocess.check_output(['git', 'stash', 'pop'])
     return result
```
mina86, line 86: Actually, what if `git checkout` fails? This needs nested try blocks
or a flag:

```python
    check_back = False
    try:
        subprocess.check_output(['git', 'checkout', str(branch)])
        check_back = True
        subprocess.check_output(['git', 'pull', 'origin', str(branch)])
        return _compile_current(branch)
    finally:
        if check_back:
            subprocess.check_output(['git', 'checkout', '--force, prev_branch])
        if stash_output != b"No local changes to save\n":
            subprocess.check_output(['git', 'stash', 'pop'])
```

PR #6128, opened 2022-01-19
docs: link Zulip&Discourse

matklad:


matklad:
There's probably more links we can add here (there's Discord as well). In general, it probably makes sense to refresh our readme, as, right now, it's not super obvious who is the target audience:

* Non-necessary technical participants in NEAR ecosystem
* nearcore contributors
* dapps authors

Decided not to do that and just add zulip link in at least some form.

bowenwang1996 (approved):
Thanks!

PR #6129, opened 2022-01-19
re: Use std::mem::take instead of std::mem::swap

pmnoxx:
Cleanup `peer_manager_actor.rs` to use `std::mem::take` instead of `std::mem::swap`.


bowenwang1996 (commented):
hmm could do we `ibf_mesg.edges.clear()` instead?

pmnoxx:
> hmm could do we `ibf_msg.edges.clear()` instead?

@bowenwang1996 Hey, that's a good question.

Yes, we can use pair of `(ibf_msg.edges.clone()` and then `ibf_msg.edges.clear())`, to get list of edges, and then clear it.
However, that involves doing a full copy of this data structure, which is not needed.

In this case, we are interested in passing list of all edges into `validate_edges_and_add_to_routing_table`, which consumes that list. Rust offers two ways, to avoid a copy, either `std::mem::swap` or `std::mem::take`.
`std::mem::take` is more idiomatic way of doing that.

bowenwang1996 (approved):

mm-near (approved):

PR #6132, opened 2022-01-19
feat: Change actix not to block sender.

pmnoxx:
In this PR, we will start using a patched version of `Actix`, which doesn't block the Actors, when their queues are full, and prints a stack trace every time the queue size is `0 mod 1024`, and `>= 1024`.

This is still WIP, but it's worth trying to see if that fixes the memory issue.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5,11 +5,11 @@ version = 3
 [[package]]
 name = "actix"
 version = "0.11.0-beta.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "8eb78f9871feb3519e06b947c2becbf2cc7f67ce786e510e6bd3f9a27da3dbf1"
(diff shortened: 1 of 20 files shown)


frol:
@pmnoxx git dependencies break our crates.io publishing. Can we patch the dependency in the root Cargo.toml?

pmnoxx:
@frol Probably, the best solution would be to publish this crate under a different name. Maybe `near-actix`.

I need to make a few other changes, to add useful stats to actix, that's seems like a right way to do it.

frol:
> I need to make a few other changes, to add useful stats to actix, that's seems like a right way to do it.

That does not seem like a right way to maintain actix. Can we PR the changes to actix?

pmnoxx:
> > I need to make a few other changes, to add useful stats to actix, that's seems like a right way to do it.
> 
> That does not seem like a right way to maintain actix. Can we PR the changes to actix?

There are a lot of other issues we currently have with Actix.
- one is lack of stats
- other is lack of prioritization, which causes nodes to OOM

I don't see an easy way to incorporate those changes into Actix directly. Other than either creating a fork, or doing a rewrite.
I'll close this PR for now.

PR #6133, opened 2022-01-19
doc: Add comments to storage estimations

jakmeier:


jakmeier:
@matklad This is a suggestion for how I could add documentation on all cost estimations, starting with the storage costs. 
As I go through all other costs, I will add more comments. If you had something different in mind when we discussed this earlier this week, then please let know.

I'm still a bit unsure what kind of documentation adds the most value here. But I figured explicitly writing down the intention behind each estimation right at the enum variant  definition is a good start. We can always move things around later.

matklad:
Yeah, this is roughly what I have in mind. I think it would be useful to have three sections for each costs:

* when it is charged
* what operation is actually done
* how we estimate it

```
    /// Estimates `ExtCost::storage_write_base` which is charged once per call to `write_storage`.
    ///
    /// `write_storage` records a single "key/value" pair, initially in the prospective changes in-memory hash map, 
    /// and than, in the on-disk trie.
    ///
    /// We estimate this base cost by measuring how long does it take contract to make N small writes
    StorageWriteBase,
```

It also maybe makes sense to split this per-operation rathre than per-cost. Ie, `StorageWriteBase` and `StorageWriteByte` should probably be described by a single comment.

jakmeier:
> Yeah, this is roughly what I have in mind. I think it would be useful to have three sections for each costs:
> 
>     * when it is charged
> 
>     * what operation is actually done
> 
>     * how we estimate it
> 

Interestingly, I started with exactly these three points. I ended up not including it in my first suggestion, since the second point felt like repeating what is already documented elsewhere. (On the `VMLogic` public interface defined in [logic.rs](https://github.com/near/nearcore/blob/master/runtime/near-vm-logic/src/logic.rs)) And the third point seemed like translating code into comments, where anyone who is interested in should really be technically-skilled enough to just read the code.

But I gave it another try and now I think there is good value in writing it all down here afterall. If nothing else, it at least ensures we actually look through all the code again and validate that it makes sense as it is.

For the estimations that are not directly related to a fee, such as RocksDB benchmarks,I think this schema doesn't work so well. I would suggest to just make the intention clear in a comment and describe how it relates to other estimations. Pretty much what I have done already. (But I am open to other suggestions)

> It also maybe makes sense to split this per-operation rathre than per-cost. Ie, `StorageWriteBase` and `StorageWriteByte` should probably be described by a single comment.

Some grouping definitely makes sense. But I would also like to keep comments associated with the enum variants, at least until we reached completeness. (For progress tracking.) Thus I suggest we describe the operation once right before the actual costs. And then for each cost we have 1) when it is charged and 2) how we estimate it. This is reflected in the latest commit.

runtime/runtime-params-estimator/src/cost.rs
```diff
@@ -51,25 +51,79 @@ pub enum Cost {
     EcrecoverBase,
     LogBase,
     LogByte,
+
+    // `storage_write` records a single key-value pair, initially in the prospective changes in-memory hash map, and then once a full block has been processed, in the on-disk trie.
+    // If there was already a value stored, it overwritten and the old value is returned to the caller.
+    /// Estimates `ExtCost::storage_write_base` which is charged once per call to `storage_write`.
+    ///
+    /// Estimation: Contract call that writes N small values and divide the cost by N.
```
matklad, line 59: Let's wrapp comments at 80-100 characters. It's a bit hard to read otherwise

matklad (approved):

PR #6138, opened 2022-01-20
Mocknet: Read genesis records from a separate file

nikurt:


pytest/tests/mocknet/helpers/load_testing_add_and_delete_helper.py
```diff
@@ -292,7 +297,8 @@ def main(argv):
     assert delay >= 1
     for i, account in enumerate(test_accounts):
         logger.info(f'Deploying contract for account {i}')
-        account.send_deploy_contract_tx(mocknet.WASM_FILENAME)
+        mocknet_helpers.retry_and_ignore_errors(
```
mina86, line 300: Do we really want to ignore errors if this fails?

pytest/lib/mocknet.py
```diff
@@ -440,84 +440,68 @@ def compress_and_upload(nodes, src_filename, dst_filename):
...
-            pmap(
-                lambda node: start_genesis_updater(
-                    node, 'genesis_updater.py', genesis_filename_in,
-                    '/home/ubuntu/.near/genesis.json', chain_id,
-                    validator_node_names, rpc_node_names, done_filename,
-                    epoch_length, node_pks, increasing_stakes, num_seats),
-                validator_nodes + rpc_nodes)
-            pmap(lambda node: wait_genesis_updater_done(node, done_filename),
-                 validator_nodes + rpc_nodes)
+        assert update_genesis_on_machine
```
mina86, line 460: What’s the point of the argument if it’s always required True?

pytest/lib/mocknet.py
```diff
@@ -440,84 +440,68 @@ def compress_and_upload(nodes, src_filename, dst_filename):
...
                               node_pks=None,
                               increasing_stakes=0.0,
-                              num_seats=100):
+                              num_seats=100,
+                              sharding=True,
+                              all_node_pks=None,
+                              node_ips=None):
     logger.info(
         f'create_and_upload_genesis: validator_nodes: {validator_nodes}')
     assert chain_id
```
mina86, line 455: What’s the point of default None if that default is never correct?

pytest/lib/mocknet.py
```diff
@@ -440,84 +440,68 @@ def compress_and_upload(nodes, src_filename, dst_filename):
...
+        assert update_genesis_on_machine
+        logger.info(
+            'Assuming that genesis_updater.py is available on the instances.')
+        validator_node_names = [node.instance_name for node in validator_nodes]
+        rpc_node_names = [node.instance_name for node in rpc_nodes]
+        assert '-spoon' in chain_id, f'Expecting chain_id like "testnet-spoon" or "mainnet-spoon", got {chain_id}'
+        chain_id_in = chain_id.split('-spoon')[0]
+        genesis_filename_in = f'/home/ubuntu/.near/${chain_id_in}/genesis.json'
+        records_filename_in = f'/home/ubuntu/.near/${chain_id_in}/records.json'
+        config_filename_in = f'/home/ubuntu/.near/${chain_id_in}/config.json'
```
mina86, line 469: Do you mean to have `$` in the resulting string? As in, is this meant to be expanded with environment variables?

pytest/lib/mocknet.py
```diff
@@ -440,84 +440,68 @@ def compress_and_upload(nodes, src_filename, dst_filename):
...
+        done_filename = f'/home/ubuntu/genesis_update_done_{int(time.time())}.txt'
```
mina86, line 470: Perhaps:

    stamp = time.strftime('%Y%m%d-%H%M%S', time.gmtime())
    done_filename = f'/home/ubuntu/genesis_update_done_{stamp}.txt'

pytest/lib/mocknet.py
```diff
@@ -847,39 +844,52 @@ def reset_data(node, retries=0):
...
+def start_genesis_updater_script(
+        script, genesis_filename_in, genesis_filename_out, records_filename_in,
+        records_filename_out, config_filename_in, config_filename_out, chain_id,
+        validator_nodes, rpc_nodes, done_filename, epoch_length, node_pks,
+        increasing_stakes, num_seats, sharding, all_node_pks, node_ips):
     return '''
         cd {dir}
         rm -f {done_filename}
-        nohup ./venv/bin/python {script} {genesis_filename_in} {genesis_filename_out} {chain_id} {validator_nodes} {rpc_nodes} {done_filename} {epoch_length} {node_pks} {increasing_stakes} {num_seats} 1> genesis_updater.out 2> genesis_updater.err < /dev/null &
+        nohup ./venv/bin/python {script} {genesis_filename_in} {genesis_filename_out} {records_filename_in} {records_filename_out} {config_filename_in} {config_filename_out} {chain_id} {validator_nodes} {rpc_nodes} {done_filename} {epoch_length} {node_pks} {increasing_stakes} {num_seats} {sharding} {all_node_pks} {node_ips} 1> genesis_updater.out 2> genesis_updater.err < /dev/null &
```
mina86, line 855: There’s so many parameters at this point that I would wonder if it’d
be better to do a join:

    cmd = ' '.join(shlex.quote(str(arg)) for arg in (
        'nohup',
        './venv/bin/python',
        script,
        genesis_filename_in,
        genesis_filename_out,
        records_filename_in,
        records_filename_out,d
        config_filename_in,
        config_filename_out,
        chain_id,
        ','.join(validator_nodes),
        ','.join(rpc_nodes),
        done_filename,
        epoch_length,
        ','.join(node_pks),
        increasing_stakes,
        num_seats,
        sharding,
        ','.join(all_node_pks),
        ','.join(node_ips)
    ))
    return f'''
        cd {shlex.quote(dir)}
        rm -f -- {shlex.quote(done_filename)}
        {cmd} >genesis_updater.out 2>genesis_updater.err </dev/null &
    '''

pytest/lib/mocknet.py
```diff
@@ -440,10 +440,9 @@ def compress_and_upload(nodes, src_filename, dst_filename):
 # We assume that the nodes already have the .near directory with the files
 # node_key.json, validator_key.json and config.json.
 def create_and_upload_genesis(validator_nodes,
```
mina86, line 442: FYI, if you want to require those to be passed as keyword arguments you can add `*` as argument in the definition, e.g.:

    def create_and_upload_genesis(*,
                                  validator_nodes,
                                  chain_id,
                                  …):

pytest/lib/mocknet.py
```diff
@@ -849,29 +846,22 @@ def start_genesis_updater_script(
...
+    cmd = ' '.join([
+        shlex.quote(str(arg)) for arg in [
+            'nohup', './venv/bin/python', script, genesis_filename_in,
+            genesis_filename_out, records_filename_in, records_filename_out,
+            config_filename_in, config_filename_out, chain_id, ','.join(
+                validator_nodes), ','.join(rpc_nodes), done_filename,
+            str(epoch_length), ','.join(node_pks),
+            str(increasing_stakes),
+            str(num_seats),
+            str(sharding), ','.join(all_node_pks), ','.join(node_ips)
```
mina86, line 858: `str` is in `shlex.quote` invocation so it’s not needed here.

pytest/tests/mocknet/helpers/load_testing_add_and_delete_helper.py
mina86: Right, but if we fail deploying the contract doesn’t that break other tests?

mina86 (approved):

nikurt:
> Right, but if we fail deploying the contract doesn’t that break other tests?

Nothing catastrohpic will happen if a couple of contracts fail to deploy.

PR #6139, opened 2022-01-20
Alt-bn128 audit fixes and tests

rahulghangas:
- [ ] Full testing

### DIFF
--- a/runtime/near-vm-logic/src/alt_bn128.rs
+++ b/runtime/near-vm-logic/src/alt_bn128.rs
@@ -27,25 +27,23 @@ pub fn alt_bn128_g1_multiexp_sublinear_complexity_estimate(
     const MULTIEXP_ITEM_SIZE: u64 = std::mem::size_of::<(G1, Fr)>() as u64;
 
     // A+C*n/(log2(n)+4) - B*n - discount
-
-    let n = (n_bytes - 1).checked_add(MULTIEXP_ITEM_SIZE).ok_or(HostError::IntegerOverflow)?
+    let n = (n_bytes)
+        .checked_add(MULTIEXP_ITEM_SIZE).ok_or(HostError::IntegerOverflow)?
+        .checked_add(MULTIEXP_ITEM_SIZE).ok_or(HostError::IntegerOverflow)?
         / MULTIEXP_ITEM_SIZE;
-    let res = A
-        .checked_add(if n == 0 {
-            0
-        } else {
-            // set linear complexity growth for n > 32768
-            let growth_factor = std::cmp::min(ilog2(n), 15);
-
-            // (n * (growth_factor + 3) +  (1 << (1 + growth_factor)))  * C / ((growth_factor + 4) * (growth_factor + 5))
-            (n.checked_mul(growth_factor + 3).ok_or(HostError::IntegerOverflow)?)
+        
+    let res = A;
+    if n != 0 {
+        let growth_factor = std::cmp::min(ilog2(n), 15);
+        res.checked_add(
+            n.checked_mul(growth_factor + 3).ok_or(HostError::IntegerOverflow)?
                 .checked_add(1 << (1 + growth_factor))
                 .ok_or(HostError::IntegerOverflow)?
                 .checked_mul(C)
                 .ok_or(HostError::IntegerOverflow)?
                 / ((growth_factor + 4) * (growth_factor + 5))
-        })
-        .ok_or(HostError::IntegerOverflow)?;
+        ).ok_or(HostError::IntegerOverflow)?;
+    }
 
     Ok(res.saturating_sub(
         B.checked_mul(n)
--- a/runtime/near-vm-logic/src/alt_bn128.rs
+++ b/runtime/near-vm-logic/src/alt_bn128.rs
@@ -12,7 +12,6 @@ const NOT_IN_FIELD: &str = "integer is not less than modulus";
 #[inline]
 pub fn ilog2(n: u64) -> u64 {
     assert!(n > 0);
-    let mut r = 0;
     63 - n.leading_zeros() as u64
 }
 
--- a/runtime/near-vm-logic/src/alt_bn128.rs
+++ b/runtime/near-vm-logic/src/alt_bn128.rs
@@ -29,8 +29,8 @@ pub fn alt_bn128_g1_multiexp_sublinear_complexity_estimate(n_bytes: u64, discoun
         0
     } else {
         // set linear complexity growth for n > 32768
-        let l = std::cmp::min(ilog2(n), 15);
-        (n * (l + 3) + (1 << (1 + l))) * C / ((l + 4) * (l + 5))
+        let growth_factor = std::cmp::min(ilog2(n), 15);
+        (n * (growth_factor + 3) + (1 << (1 + growth_factor))) * C / ((growth_factor + 4) * (growth_factor + 5))
     };
 
     res.saturating_sub(B*n + discount)
--- a/runtime/near-vm-logic/src/alt_bn128.rs
+++ b/runtime/near-vm-logic/src/alt_bn128.rs
@@ -272,7 +272,7 @@ pub fn alt_bn128_g1_sum(data: &[u8]) -> crate::logic::Result<Vec<u8>> {
         .collect::<Vec<_>>();
 
     let mut acc = G1::zero();
-    acc = items.iter().fold(acc, |acc, &(sign, e)| if sign {acc - e} else {acc} );
+    acc = items.iter().fold(acc, |acc, &(sign, e)| if sign {acc - e} else {acc + e} );
 
     let result = WrapG1(acc)
         .try_to_vec()
--- a/runtime/near-vm-logic/src/alt_bn128.rs
+++ b/runtime/near-vm-logic/src/alt_bn128.rs
@@ -99,15 +99,15 @@ impl BorshDeserialize for WrapFq {
 
 impl BorshSerialize for WrapFq2 {
     fn serialize<W: Write>(&self, writer: &mut W) -> Result<(), Error> {
-        WrapFq(self.0.imaginary()).serialize(writer)?;
-        WrapFq(self.0.real()).serialize(writer)
+        WrapFq(self.0.real()).serialize(writer)?;
+        WrapFq(self.0.imaginary()).serialize(writer)
     }
 }
 
 impl BorshDeserialize for WrapFq2 {
     fn deserialize(buf: &mut &[u8]) -> Result<Self, Error> {
-        let im = WrapFq::deserialize(buf)?.0;
         let re = WrapFq::deserialize(buf)?.0;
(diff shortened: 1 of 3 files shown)


EgorKulikov (approved):
LGTM, very clean and easy to understand. Can't vouch for change in comment about algorithm though

abacabadabacaba:
One issue about this code that I noted before and which remains is that the new host functions introduced here accept their input and produce their output in Borsh-serialized form. None of the existing host functions do that. Because of this, the SDK will likely need to do extra memory allocations and copies. It would be much more natural to instead pass the data as a slice of items (with each item being `(G1, Fr)`, `(bool, G1)` or `(G1, G2)`). The difference with Borsh is that in the Borsh-serialized form, the data is prefixed with the number of items, which is unnecessary as it can be derived from the length.

@matklad WDYT about this?

runtime/near-vm-logic/src/alt_bn128.rs
```diff
@@ -47,20 +28,31 @@ pub fn alt_bn128_g1_multiexp_sublinear_complexity_estimate(n_bytes: u64, discoun
...
-        (n * (l + 3) + (1 << (1 + l))) * C / ((l + 4) * (l + 5))
-    };
-
-    if res < B * n + discount {
-        return 0;
-    }
-    res -= B * n + discount;
-    res
+    let n = (n_bytes - 1).checked_add(MULTIEXP_ITEM_SIZE).ok_or(HostError::IntegerOverflow)?
+        / MULTIEXP_ITEM_SIZE;
```
abacabadabacaba, line 32: This will underflow if `n_bytes == 0`. This also assumes that (normally) `n_bytes` is a multiple of the item size, but it actually includes extra 4 bytes for the number of items, so `n` will be one greater than the number of items.

runtime/near-vm-logic/src/alt_bn128.rs
```diff
@@ -47,20 +28,31 @@ pub fn alt_bn128_g1_multiexp_sublinear_complexity_estimate(n_bytes: u64, discoun
...
+    let res = A
+        .checked_add(if n == 0 {
+            0
+        } else {
```
abacabadabacaba, line 36: Why not mutate `res` conditionally instead?

runtime/near-vm-logic/src/alt_bn128.rs
```diff
@@ -47,20 +28,31 @@ pub fn alt_bn128_g1_multiexp_sublinear_complexity_estimate(n_bytes: u64, discoun
 
     // A+C*n/(log2(n)+4) - B*n - discount
```
abacabadabacaba, line 29: This formula doesn't match what is computed below.

runtime/near-vm-logic/src/logic.rs
```diff
@@ -781,11 +781,16 @@ impl<'a> VMLogic<'a> {
...
+                (alt_bn128_g1_multiexp_byte as u64)
+                    .checked_mul(len)
+                    .ok_or(HostError::IntegerOverflow)?,
+            )
+            .ok_or(HostError::IntegerOverflow)?)
             / alt_bn128_g1_multiexp_sublinear as u64;
         let sublinear_complexity =
-            crate::alt_bn128::alt_bn128_g1_multiexp_sublinear_complexity_estimate(len, discount);
+            crate::alt_bn128::alt_bn128_g1_multiexp_sublinear_complexity_estimate(len, discount)?;
         self.gas_counter.pay_per(alt_bn128_g1_multiexp_sublinear, sublinear_complexity)?;
```
abacabadabacaba, line 794: The cost formula seems overly complicated. It uses `alt_bn128_g1_multiexp_base` and `alt_bn128_g1_multiexp_byte` (which were already charged before), divides them by `alt_bn128_g1_multiexp_sublinear` and later multiplies it back. In general, all costs should be computed like this: first compute some value that doesn't depend on any costs (e.g. the number of bytes), and then multiply it by a fixed cost value (e.g. cost per byte).

abacabadabacaba (changes requested):

abacabadabacaba:
Another issue with the API is that the pairing check function is very specific: it can only check that the result is equal to one, but not return it. This may limit the number of possible use cases. We may want to have a more general API (which may also include addition and multiplication for G2 and Gt and hash to curve).

runtime/near-vm-logic/src/alt_bn128.rs
```diff
@@ -47,20 +28,31 @@ pub fn alt_bn128_g1_multiexp_sublinear_complexity_estimate(n_bytes: u64, discoun
```
abacabadabacaba, line 28: These constants seem to be estimated. They should be together with the other estimated costs, not here.

abacabadabacaba (changes requested):

matklad:
@abacabadabacaba I agree with the concern about borsh. I'd also add that using `borsh` adds extra error conditions. In particular, `AltBn128SerializationError` I think is de-facto dead code (it should be impossible to trigger this error). For our ABI, I feel it's relatively important to not have such issues. Defining ABI in terms of `#[repr(C)]` structs feels more proper.

rahulghangas:
@abacabadabacaba @matklad apologies, github hasn't been sending me notifications for your reviews. Will try to address them today

runtime/near-vm-logic/src/alt_bn128.rs
```diff
@@ -253,6 +246,11 @@ pub fn alt_bn128_g1_multiexp(data: &[u8]) -> crate::logic::Result<Vec<u8>> {
         .into_iter()
         .map(|e| (e.0 .0, e.1 .0))
         .collect::<Vec<_>>();
+    // Upper bounded by 2^9 buckets
+    // Reference: https://github.com/zeropoolnetwork/bn/blob/b25e138cd5a7d98ad16c091c20b4fa273ca1f993/src/groups/mod.rs#L71
+    if items.len() > 5000 {
+        return Err(VMLogicError::HostError(HostError::AltBn128MaxNumberOfItemsExceeded));
+    }
```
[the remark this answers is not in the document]
rahulghangas: the bn library used can cause an OOM error because it allocates buckets exponentailly, which is why this limit was added to circumvent that. The other solution would be to fork the library and add the constraint there

rahulghangas:
>Another issue with the API is that the pairing check function is very specific: it can only check that the result is equal to one, but not return it. This may limit the number of possible use cases. We may want to have a more general API (which may also include addition and multiplication for G2 and Gt and hash to curve)
Hmm, will need to think about this. 

Btw, @abacabadabacaba @matklad apologies for the massive delay, this keeps getting pushed in my backlog due to other work, only got around to it recently

runtime/near-vm-logic/src/alt_bn128.rs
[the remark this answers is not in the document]
rahulghangas: https://github.com/zeropoolnetwork/bn/blob/b25e138cd5a7d98ad16c091c20b4fa273ca1f993/src/groups/mod.rs#L71

nearmax:
The concerns raised by @abacabadabacaba and @matklad are not blocking.

We had an extremely long, more than a year long, discussion about the best signature for these methods. There is has to be a limit for how long this conversation can go, especially given the diminishing impact of arguing about serializer or generalization of math operation. Consider that the decision on the method signatures has been made long time ago and we will re-open it only if we later encounter performance or usage issues with the current API confirmed by the real projects. The goal of this PR is to stabilize the code and address the security findings, not to revamp the API.

Unless there are safety or validity concerns, @rahulghangas , please address @abacabadabacaba 's [comment about cost computation](https://github.com/near/nearcore/pull/6139#discussion_r813040108), and let's proceed with merging this PR.

matklad:
I would say that presence of "AltBn128SerializationError, AltBn128DeserializationError" in the ABI is a non-blocking safety concern. If these errors can happen, we should have a test for them. If they can’t happen, that means we have dead code, and should probably comment it as such to avoid confusion.

Though, to re-iterate, this doesn’t feel too significant to me.

runtime/near-vm-logic/src/alt_bn128.rs
```diff
@@ -27,25 +27,23 @@ pub fn alt_bn128_g1_multiexp_sublinear_complexity_estimate(
     const MULTIEXP_ITEM_SIZE: u64 = std::mem::size_of::<(G1, Fr)>() as u64;
 
     // A+C*n/(log2(n)+4) - B*n - discount
-
-    let n = (n_bytes - 1).checked_add(MULTIEXP_ITEM_SIZE).ok_or(HostError::IntegerOverflow)?
```
matklad: Note: I don't fully understand the change from `(n_bytes + size - 1) / size` to `(n_bytes + 2*size) / size`. The former compute ceil divion, the latter feels like bug.

runtime/near-vm-logic/src/alt_bn128.rs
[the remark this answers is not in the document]
matklad: I don't understand this reasoning. It seems that we allocate (`2^ln(items.len())`) buckets, where `ln` is a natural (base e) logarithm. That's going to be smaller than `items.len()`.

matklad:
Ok, I've spend today getting up to speed, and my impression is:

* This pr feels like an overall improvement
* At the same time, I am fairly certain that this doesn't yet meet the standard of quality we want to hold for this part of the codebase. In particular:
  * cost computations are all over the place: we charge per byte, not per field element, the "sublinear" stuff overall feels needlessly complicated to me (we don't even have machinery for estimating less-then-linear components of costs), there are numerous rounding bugs related to borsh's extra bytes with length
  * I think the `< 5000` condition is wrong. 
  * On the closer reading, using borsh here makes the code **way** more complicated than it needs to be. 
  * Test coverage is incomplete. In particular, we don't seem to have tests for invalid field elements

I suggest merging the PR as is, and iterating in follow ups. I'll also be in charge of  pushing this over the finish line and stabilizing.


@abacabadabacaba WDYT in particular about merging this? As this is still feature-gated, I don't see harm here, and I see benefit in not keeping PR in limbo.

matklad:
https://github.com/near/nearcore/pull/6505 is proposed fix for the problems outlined above.

runtime/near-vm-logic/src/alt_bn128.rs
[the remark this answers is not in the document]
matklad: Yeah, manual experiments didn't show exponential memory usage here.

bowenwang1996:
> I suggest merging the PR as is, and iterating in follow ups. I'll also be in charge of pushing this over the finish line and stabilizing.

@matklad let's do it

matklad:
@rahulghangas could you rebase the PR so that we can merge it?

rahulghangas:
@matklad done

matklad:
Prefect, thanks @rahulghangas!

I also note that the PR description has a `tests` checkbox -- do you by any chance have some test verctors ready somewhere?

Issue #6120, opened 2022-01-19
Upgrade rust to 1.58


matklad:
1.58 is out, we should upgrade to it, preferably early in the release cycle.

matklad:
Probably makes sense to wait for 1.58.1: https://blog.rust-lang.org/2022/01/20/cve-2022-21658.html

Issue #6124, opened 2022-01-19
Check that `rust-version` in Cargo.tomls is consistent with what we declare in `rust-toolchain.toml`


matklad:
The check should go here:

[https://github.com/near/nearcore/blob/6d86ccd4fee74c1567dc2d83d384b1952826e796/tools/themis/src/rules.rs#L28-L48](https://github.com/near/nearcore/blob/6d86ccd4fee74c1567dc2d83d384b1952826e796/tools/themis/src/rules.rs#L28-L48)

See [https://github.com/near/nearcore/pull/5834/files#r787711401](https://github.com/near/nearcore/pull/5834/files#r787711401) for context.

matklad:
(as an alternative, we can also just remove `rust-version` from Cargo.toml. I am still not sure if, on ballance, the field is worth it. #5834 shifted my opinion more in the "not worth it" direction, but that could have played differently if we did have the CI check)

matklad:
Hm, I think what we actually should do here is

* remove `rust-version` from Cargo.toml
* add it on-publish, the same way we add `version`

@miraclx WDYT?

miraclx:
I'm not sure if this is an optimal solution. `rust-version` specifies MSRV for a crate. If we decide to keep a consistent version for all crates in the workspace, there would undoubtedly be crates that run just fine on lower versions, but this constraint would prohibit their use.

As currently defined, the `rust-version` spec and compliance checks we have allows each crate to have its own minimum rust version spec, which can be anything less than or equal to the version in `rust-toolchain`.

I'd advise we only bump the MSRV on a crate by crate basis, based on utility. Whatever runs just fine on a prior rust version doesn't need to be updated.

However, the `rust-toolchain` version can and should be whatever the latest rust version is.

matklad:
There's no way for us to understand *when* we we need to upgrade `rust-version`. We only run CI with what's specified in `rust-toolchain.toml`, and only that version is guaranteed to work. 

We *can* add MSRV CI check for every crate, but I think this is clearly unreasonable -- nearcore crates do not generally promise any kinds of stability or compatibility.

But year, *select* crates which do want to make extra promises, like `near-account-id` can manually specify lower `rust-version` in Cargo.toml, provided that there's a CI check for that as well.

pmnoxx:
> 



> LGTM as-is.
> 
> See my comment in [#6124](https://github.com/near/nearcore/issues/6124#issuecomment-1018708432)
[...]

PR #6142, opened 2022-01-20
feat: Upgrade rust to 1.58.1

pmnoxx:
This is a proposal to upgrade to rust 1.58.0

See https://blog.rust-lang.org/2022/01/13/Rust-1.58.0.html

The most notable improvement is to string formatting.
It's not possible to write 
```rust
println!("Hello, {}!", get_person());                // implicit position
println!("Hello, {0}!", get_person());               // explicit index
println!("Hello, {person}!", person = get_person()); // named

let person = get_person();
// ...
println!("Hello, {person}!"); // captures the local `person`

let ident = ...;
panic!("{ident}")
```
This code would capture ident from the current score, and put it inside the format.

See https://github.com/near/nearcore/issues/6120

No new warnings were generated.

TODO:
- [x] wait for 1.58.1 - https://blog.rust-lang.org/2022/01/20/cve-2022-21658.html


pmnoxx:
> We should also upgrade all Cargo.toml to say `rust-version = "1.58"`

Or, even better, let's remove `rust-version = "1.56"` from or `Cargo.toml` files.

pmnoxx:
@matklad Otherwise, we should have to involve everyone in the company, every time, we need to make a change to `rust-version`.

matklad:
Yeah, that's not really sustainable. I think I've found a better option -- set this field on publish, the same way we do with `version`: https://github.com/near/nearcore/issues/6124#issuecomment-1018698670. 

So, I suggest the following: 

* accept that `rust-version` today are wrong and don't spend cycles fixing them
* merge this PR without worrying about that.

chain/client/src/tests/catching_up.rs
```diff
@@ -694,7 +694,7 @@ enum ChunkGrievingPhases {
 
 // TODO(#3180): seals are disabled in single shard setting
 #[test]
-#[cfg_attr(not(feature = "expensive_tests"), ignore)]
+// #[cfg_attr(not(feature = "expensive_tests"), ignore)]
```
matklad, line 697: Why is this commented out? It's a but puzzling to see newly commented out code in a review.

pmnoxx: @matklad  With `1.58.0.`, this generates a warning about unused `ignore`.

miraclx (approved):
LGTM as-is.

<details>
<summary>See my comment in <a href="https://github.com/near/nearcore/issues/6124#issuecomment-1018708432">#6124</a> </summary>

  > I'm not sure if this is an optimal solution. `rust-version` specifies MSRV for a crate. If we decide to keep a consistent version for all crates in the workspace, there would undoubtedly be crates that run just fine on lower versions, but this constraint would prohibit their use.
  >
  > As currently defined, the `rust-version` spec and compliance checks we have allows each crate to have its own minimum rust version spec, which can be anything less than or equal to the version in `rust-toolchain`.
  > 
  > I'd advise we only bump the MSRV on a crate by crate basis, based on utility. Whatever runs just fine on a prior rust version doesn't need to be updated.
  > 
  > However, the `rust-toolchain` version can and should be whatever the latest rust version is.

</details>

chain/client/src/tests/catching_up.rs
matklad: Ah I see, we have a blanket `#[ignore]` at the next line. Let's just remove the commented-out line -- I see how it communicates "when you unignore the test, make sure to mark it as expensive", but I don't think that adds enough real value -- people would just be puzzled over this.

pmnoxx:
> 



> LGTM as-is.
> 
> See my comment in [#6124](https://github.com/near/nearcore/issues/6124#issuecomment-1018708432)
> > I'm not sure if this is an optimal solution. `rust-version` specifies MSRV for a crate. If we decide to keep a consistent version for all crates in the workspace, there would undoubtedly be crates that run just fine on lower versions, but this constraint would prohibit their use.
> > As currently defined, the `rust-version` spec and compliance checks we have allows each crate to have its own minimum rust version spec, which can be anything less than or equal to the version in `rust-toolchain`.
> > I'd advise we only bump the MSRV on a crate by crate basis, based on utility. Whatever runs just fine on a prior rust version doesn't need to be updated.
> > However, the `rust-toolchain` version can and should be whatever the latest rust version is.

We currently, don't verify whenever a crate will work with `rust-version`, which is specified. 

I see a few choices:
- option 1) keep things as they are - we will surely have some crates that claim to work, for example with `1.56.0`, but they don't, for example, due to someone starting to use a features from newer version of Rust.
- option 2) add CI check, which checks whenever each crate, can be compiled with version it specifies
- option 3 ) keep the requirement, that each crate should work with the version specified in `rust-toolchain`, then we won't need per crate configuration

Questions:
- do we want to ensure that each crate works with the minimum version

miraclx:
My vote is for option 2. A simple `cargo +<msrv> check -p <crate>` for each crate. [`cargo-msrv version`](https://foresterre.github.io/cargo-msrv/commands/verify.html) is well suited to help with this.

Even if we don't adopt this for all the crates, we should at least for the _non-private_ ones. Because otherwise, automated publishing would attempt to compile those crates before publishing them. Best we catch that discrepancy on the PR level than have breaking CI on master.

matklad (approved):

chain/client/src/tests/catching_up.rs
```diff
@@ -694,7 +694,6 @@ enum ChunkGrievingPhases {
 
 // TODO(#3180): seals are disabled in single shard setting
 #[test]
```
mm-near, line 696: is it related to this PR?

mm-near (approved):

chain/client/src/tests/catching_up.rs
matklad: Yes: new version of the compiler contains a bug fix which makes it correctly emit a warning about duplicate `ignore`

PR #6127, opened 2022-01-19
debt: Upgrade wasmtime to the latest version

Ekleog:
This builts on top of #6096, so I'll undraft after it'll have landed and I'll have rebased it.

PR #6144, opened 2022-01-20
debt: upgrade wasmtime to latest version

Ekleog:
Hacky version of #6127, as per https://near.zulipchat.com/#narrow/stream/295306-nearinc.2Fcontract-runtime/topic/wasmtime.20upgrade

Hopefully we can rekindle #6096 after this lands to eliminate the need for unsafe code :)

### DIFF
--- a/runtime/near-vm-runner/src/imports.rs
+++ b/runtime/near-vm-runner/src/imports.rs
@@ -438,25 +438,8 @@ pub(crate) mod wasmtime {
         static EMBEDDER_ERROR: RefCell<Option<VMLogicError>> = RefCell::new(None);
     }
 
-    // Wasm has only i32/i64 types, so Wasmtime 0.17 only accepts
-    // external functions taking i32/i64 type.
-    // Remove, once using version with https://github.com/bytecodealliance/wasmtime/issues/1829
-    // fixed. It doesn't affect correctness, as bit patterns are the same.
-    #[cfg(feature = "wasmtime_vm")]
-    macro_rules! rust2wasm {
-        (u64) => {
-            i64
-        };
-        (u32) => {
-            i32
-        };
-        ( () ) => {
-            ()
-        };
-    }
-
     pub(crate) fn link(
-        linker: &mut wasmtime::Linker,
+        linker: &mut wasmtime::Linker<()>,
         memory: wasmtime::Memory,
         raw_logic: *mut c_void,
         protocol_version: ProtocolVersion,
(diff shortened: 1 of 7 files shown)


Ekleog:
Also, FWIW I'm not sure which is worse, between handling a `*mut c_void` and extracting it as `Foo<'_>` and handling a `Foo<'static>` whose lifetime is actually a lie, so if you have a specific preference I can definitely make the switch!

runtime/near-vm-runner/src/wasmtime_runner.rs
```diff
@@ -8,65 +8,63 @@ use near_primitives::runtime::fees::RuntimeFeesConfig;
...
     }
 
     fn write_memory(&mut self, offset: u64, buffer: &[u8]) {
-        // data_unchecked_mut() is unsafe.
-        unsafe {
+        CALLER.with(|caller| {
             let offset = offset as usize;
             for i in 0..buffer.len() {
-                self.0.data_unchecked_mut()[i + offset] = buffer[i];
+                self.0.data_mut(caller.borrow_mut().as_mut().unwrap())[i + offset] = buffer[i];
```
matklad, line 65: Can pull borrowing out of the loop

runtime/near-vm-runner/src/wasmtime_runner.rs
```diff
@@ -8,65 +8,63 @@ use near_primitives::runtime::fees::RuntimeFeesConfig;
...
 
     fn read_memory(&self, offset: u64, buffer: &mut [u8]) {
-        let offset = offset as usize;
-        // data_unchecked() is unsafe.
-        unsafe {
+        CALLER.with(|caller| {
+            let offset = offset as usize;
             for i in 0..buffer.len() {
-                buffer[i] = self.0.data_unchecked()[i + offset];
+                buffer[i] = self.0.data(caller.borrow_mut().as_mut().unwrap())[i + offset];
```
matklad, line 52: can pull borrowing out of the loop

runtime/near-vm-runner/src/imports.rs
```diff
@@ -469,21 +452,28 @@ pub(crate) mod wasmtime {
...
+                    // the below is bad. don't do this at home. it probably works thanks to the exact way the system is setup.
+                    // Thanksfully, this doesn't run in production, and hopefully should be possible to remove before we even
+                    // consider doing so.
                     let data = CALLER_CONTEXT.with(|caller_context| {
                         unsafe {
                             *caller_context.get()
                         }
                     });
+                    unsafe {
+                        // transmute the lifetime of caller so it's possible to put it in a thread-local
```
matklad, line 471: I presonally have a soft preference for grammar and punctuation in coments: 

```
// Transmute the lifetime of caller so it's possible to put it in a thread-local.
```

Not because it makes the code that much more readable, but because it just encourages you to write better comments (writing a proper paragraph is serious business, while a mere comment is just that).

matklad (approved):
Nice! This indeed will make it much easier to untangle API refactors from upgrade per se

PR #6146, opened 2022-01-20
re:  rewrite `prune_unreachable_edges_and_save_to_db` to functional style pt2

pmnoxx:
`prune_unreachable_edges_and_save_to_db` can be rewritten to a cleaner functional style. We can get rid of 2 variables as a side effect.

Made on top of https://github.com/near/nearcore/pull/6145

### DIFF
--- a/chain/network/src/routing/routing_table_actor.rs
+++ b/chain/network/src/routing/routing_table_actor.rs
@@ -307,15 +307,16 @@ impl RoutingTableActor {
         let now = Instant::now();
         // Save nodes on disk and remove from memory only if elapsed time from oldest peer
         // is greater than `SAVE_PEERS_MAX_TIME`
-        if !(force_pruning
+        let do_pruning = force_pruning
             || (self.peer_last_time_reachable.values())
-                .any(|&last_time| now.saturating_duration_since(last_time) >= SAVE_PEERS_MAX_TIME))
-        {
+                .any(|&last_time| now.saturating_duration_since(last_time) >= SAVE_PEERS_MAX_TIME);
+        if !do_pruning {
             return Vec::new();
         }
 
         // We compute routing graph every one second; we mark every node that was reachable during that time.
         // All nodes not reachable for at last 1 hour(SAVE_PEERS_AFTER_TIME) will be moved to disk.
+        // TODO - use drain_filter once it becomes stable - #59618
         let peers_to_remove = (self.peer_last_time_reachable.iter())
             .filter(|(_, &last_time)| {
                 now.saturating_duration_since(last_time) >= prune_edges_not_reachable_for
@@ -325,12 +326,7 @@ impl RoutingTableActor {
 
         debug!(target: "network", "try_save_edges: We are going to remove {} peers", peers_to_remove.len());
 
-        let current_component_nonce = self.next_available_component_nonce;
-        self.next_available_component_nonce += 1;
-
         let mut update = self.store.store_update();
-        // Stores next available nonce.
-        let _ = update.set_ser(ColLastComponentNonce, &[], &self.next_available_component_nonce);
 
         // Sets mapping from `peer_id` to `component nonce` in DB. This is later used to find
         // component that the edge belonged to.
@@ -338,26 +334,29 @@ impl RoutingTableActor {
             let _ = update.set_ser(
                 ColPeerComponent,
                 peer_id.try_to_vec().unwrap().as_ref(),
-                &current_component_nonce,
+                &self.next_available_component_nonce,
             );
 
             self.peer_last_time_reachable.remove(peer_id);
         }
 
-        let component_nonce = index_to_bytes(current_component_nonce);
-        let edges_to_remove = self
-            .edges_info
-            .iter()
-            .filter_map(|(key, edge)| {
-                if peers_to_remove.contains(&key.0) || peers_to_remove.contains(&key.1) {
-                    Some(edge.clone())
-                } else {
-                    None
-                }
+        // TODO - use drain_filter once it becomes stable - #59618
+        let edges_to_remove = (self.edges_info.values())
+            .filter(|edge| {
+                peers_to_remove.contains(&edge.key().0) || peers_to_remove.contains(&edge.key().1)
             })
+            .cloned()
             .collect();
 
-        let _ = update.set_ser(ColComponentEdges, component_nonce.as_ref(), &edges_to_remove);
+        let _ = update.set_ser(
+            ColComponentEdges,
+            &index_to_bytes(self.next_available_component_nonce),
+            &edges_to_remove,
+        );
+
+        self.next_available_component_nonce += 1;
+        // Stores next available nonce.
+        let _ = update.set_ser(ColLastComponentNonce, &[], &self.next_available_component_nonce);
 
         if let Err(e) = update.commit() {
             warn!(target: "network", "Error storing network component to store. {:?}", e);


chain/network/src/routing/routing_table_actor.rs
```diff
@@ -306,63 +306,56 @@ impl RoutingTableActor {
...
         let now = Instant::now();
-        let mut oldest_time = now;
+        // Save nodes on disk and remove from memory only if elapsed time from oldest peer
+        // is greater than `SAVE_PEERS_MAX_TIME`
+        if !(force_pruning
+            || (self.peer_last_time_reachable.values())
+                .any(|&last_time| now.saturating_duration_since(last_time) >= SAVE_PEERS_MAX_TIME))
+        {
+            return Vec::new();
+        }
```
matklad, line 316: I wonder if this could be made more readable? I see two problems here: 

* it's just a complex multiline condition
* with this implementation we iterate `peer_last_time_reachable` twice -- once to see if any peer is older than `SAVE_PEERS_MAX_TIME`, and once to collect all peers older than prune_edges_not_reachable_for. I don't worry about this from perf perspective, but I do worry about logical duplication. 

The second is probably OK -- this is essentially a fast-path check, so it is defensible that we don't want to allocate a hash set here. 

Yeah, really struggle to suggest a strictly better alternative here. Maaybe just de-morgan this? 

```rust
        // Fast path: if we saw everyone somewhat recently, don't bother with pruning edges. 
        if !force_pruning
            && self.peer_last_time_reachable.values()
                .all(|&last_time| now.saturating_duration_since(last_time) < SAVE_PEERS_MAX_TIME))
        {
            return Vec::new();
        }
```

chain/network/src/routing/routing_table_actor.rs
```diff
@@ -306,63 +306,56 @@ impl RoutingTableActor {
...
-                    Some(edge.clone())
-                } else {
-                    None
-                }
-            })
+        // TODO - use drain_filter once it becomes stable - #59618
+        let edges_to_remove = (self.edges_info.iter())
+            .filter(|(key, _)| peers_to_remove.contains(&key.0) || peers_to_remove.contains(&key.1))
+            .map(|(_, edge)| edge.clone())
             .collect();
```
matklad, line 348:
```suggestion
        let edges_to_remove = self.edges_info.values()
            .filter(|edge| peers_to_remove.contains(&edge.key().0) || peers_to_remove.contains(&edge.key().1))
            .cloned()
            .collect();
```

chain/network/src/routing/routing_table_actor.rs
```diff
@@ -306,63 +306,56 @@ impl RoutingTableActor {
...
+            index_to_bytes(self.next_available_component_nonce).as_ref(),
+            &edges_to_remove,
+        );
+
+        self.next_available_component_nonce += 1;
+        // Stores next available nonce.
+        let _ = update.set_ser(ColLastComponentNonce, &[], &self.next_available_component_nonce);
 
         if let Err(e) = update.commit() {
             warn!(target: "network", "Error storing network component to store. {:?}", e);
```
matklad, line 361: unrelated to PR, but feels a bit odd that we we update in-memory hash maps even if ommid didn't succeed.

chain/network/src/routing/routing_table_actor.rs
pmnoxx: @matklad I moved the condition to a separate variable, to be more readable.
```rust
    let do_pruning = force_pruning
            || (self.peer_last_time_reachable.values())
                .any(|&last_time| now.saturating_duration_since(last_time) >= SAVE_PEERS_MAX_TIME);
        if !do_pruning {
            return Vec::new();
        }
```

chain/network/src/routing/routing_table_actor.rs
pmnoxx: @matklad That's a good point. However, there isn't much we can do when a database commit fails, we don't handle database failures anywhere in our code.

chain/network/src/routing/routing_table_actor.rs
```diff
@@ -306,63 +306,56 @@ impl RoutingTableActor {
...
-            .iter()
-            .filter_map(|(peer_id, last_time)| {
-                oldest_time = std::cmp::min(oldest_time, *last_time);
-                if now.saturating_duration_since(*last_time) >= prune_edges_not_reachable_for {
-                    Some(peer_id.clone())
-                } else {
-                    None
-                }
+        // TODO - use drain_filter once it becomes stable - #59618
+        let peers_to_remove = (self.peer_last_time_reachable.iter())
```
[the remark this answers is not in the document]
pmnoxx: @matklad I think it's better to write it with brackets to prevent `rustfmt` from splitting it in muiltiple lines.

I see a lot of code like this:
```rust
   &genesis
            .config
            .validators
            .iter()
            .map(|account_info| {
                (
                    account_info.account_id.clone(),
                    account_info.public_key.clone(),
                    account_info.amount,
                )
            })
            .collect::<Vec<_>>(),
```

This would be more readable to write it simply as
```rust
   &(genesis.config.validators.iter())
            .map(|account_info| {
                (
                    account_info.account_id.clone(),
                    account_info.public_key.clone(),
                    account_info.amount,
                )
            })
            .collect::<Vec<_>>(),
```

As a good practice, I think putty brackets around `(... .iter()) ` or  `(... .values())` makes code easier to read.  I really don't see any reason as to why you would want to split into a new line every time a dot appears.

A good rule of thumb, would be to split lines, such that, there is at most one function call per line. 
The following code, does make sense following that rule:
```rust
           (self.peer_last_time_reachable.values())
            .any(|&last_time| now.saturating_duration_since(last_time) >= SAVE_PEERS_MAX_TIME);
```

I probably should write this as a suggestion to `rustfmt`.

There is a reason, why we don't always split every time after every dot, some code would look ridiculous, for example:
```rust
   &genesis
            .config
            .validators
            .iter()
            .map(|account_info| {
                (
                    account_info
                        .account_id
                        .clone(),
                    account_info
                         .public_key
                          .clone(),
                    account_info
                          .amount,
                )
            })
            .collect::<Vec<_>>(),
```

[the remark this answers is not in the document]
matklad: Ah, I see, that's a good point. Won't clippy complain about the extra parens though?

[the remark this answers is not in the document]
pmnoxx: @matklad No, `clippy` doesn't complain about it.

chain/network/src/routing/routing_table_actor.rs
```diff
@@ -306,63 +306,58 @@ impl RoutingTableActor {
         prune_edges_not_reachable_for: Duration,
     ) -> Vec<Edge> {
         let now = Instant::now();
-        let mut oldest_time = now;
+        // Save nodes on disk and remove from memory only if elapsed time from oldest peer
+        // is greater than `SAVE_PEERS_MAX_TIME`
+        let do_pruning = force_pruning
+            || (self.peer_last_time_reachable.values())
+                .any(|&last_time| now.saturating_duration_since(last_time) >= SAVE_PEERS_MAX_TIME);
```
pmnoxx, line 313: @matklad Btw, this code is actually faster. Most of the time this condition is false, so in most common case, we will avoid doing extra memory allocations.

matklad (approved):

mm-near (approved):

PR #6147, opened 2022-01-20
re: chore: Rewrite `index_to_bytes` to not do hidden memory allocations

pmnoxx:
`index_to_bytes` does extra memory allocation by default, this can be avoided by using `u64::to_le_bytes`.


matklad (approved):
Good catch!

bowenwang1996 (approved):

PR #6148, opened 2022-01-20
fix: reduce number of chunk requesting messages

mzhangmzz:
This is a quick fix to reduce the number of chunk requests that we send. It changes the number of orphan blocks for which we requests for chunks from 5 to 2. 2 means we only request for missing chunks for at most 2 blocks if there are no forks. We should get most of the benefit from requesting chunks early with the number 2, since chunk requesting time is ~200-400ms, less than the block processing time these days.

Test plan:
- Start a node on testnet with the change and verify that it can catch up
- I restarted the node three times and each time it caught up every fast, CPU and memory usage was normal. I also verified in the log that there were indeed cases when the node requested chunks for two blocks at a time.
- https://grafana.near.org/d/zzGDUdcnk/chunks-delay-monitoring?orgId=1&var-chain_id=testnet&var-role=min-testing&var-node_id=All&from=now-3h&to=now


bowenwang1996 (commented):
What issue does this PR intend to fix? Are we confident that it actually fixes the issue? It is not clear to me that this change is actually beneficial to us.

nearcore/src/config.rs
```diff
@@ -82,7 +82,7 @@ const BLOCK_HEADER_FETCH_HORIZON: BlockHeightDelta = 50;
 const CATCHUP_STEP_PERIOD: u64 = 100;
 
 /// Time between checking to re-request chunks.
-const CHUNK_REQUEST_RETRY_PERIOD: u64 = 400;
+const CHUNK_REQUEST_RETRY_PERIOD: u64 = 800;
```
mm-near, line 85: optional: should we also change any settings in
https://github.com/near/nearcore/blob/master/chain/chunks/src/lib.rs#L134 ?

(in theory - we can leave them as is)

mm-near (approved):

nearcore/src/config.rs
bowenwang1996: @mzhangmzz are we confident that this won't degrade the performance, i.e, the time it takes to reconstruct a chunk? Maybe we should test this PR and see how well it performs

Issue #5340, opened 2021-11-17
Should nearcore exit with error code after upgrade panic


kucharskim:
Binary is still running but is useless and fills the logs with flood of mailbox errors.
```
Nov 17 16:46:24 near03 neard[1445878]: Nov 17 16:46:24.672  INFO stats: #52890638 HzgsmYrzpyayMRfmwh2bdYm4k
VkTa79y1c27VfubRVwR V/60 31/26/40 peers ⬇ 487.5kiB/s ⬆ 742.6kiB/s 0.90 bps 83.61 Tgas/s CPU: 31%, Mem: 7.2 GiB
Nov 17 16:46:34 near03 neard[1445878]: Nov 17 16:46:34.674  INFO stats: #52890646 EnR1GrLyJaiC7M9n5XGFqfHTc
1Y1oo2jh2jJVcRvTzhx V/60 31/24/40 peers ⬇ 511.0kiB/s ⬆ 826.0kiB/s 0.80 bps 60.54 Tgas/s CPU: 29%, Mem: 7.2 GiB
Nov 17 16:46:44 near03 neard[1445878]: Nov 17 16:46:44.677  INFO stats: #52890655 BjeJz7bSxtowHp1mJfWSjrrhJ
r5Q9WZoMjc6i3NVU4Qn V/60 31/21/40 peers ⬇ 498.8kiB/s ⬆ 749.5kiB/s 0.90 bps 115.41 Tgas/s CPU: 33%, Mem: 7.2 GiB
Nov 17 16:46:54 near03 neard[1445878]: Nov 17 16:46:54.679  INFO stats: #52890663 ESkVGwasLLUXn97oAY9uDthGi
beAprrC7F9KgMtEub5r V/60 31/20/40 peers ⬇ 505.8kiB/s ⬆ 800.6kiB/s 0.70 bps 47.68 Tgas/s CPU: 21%, Mem: 7.2 GiB
Nov 17 16:46:55 near03 neard[1445878]: thread 'main' panicked at 'The client protocol version is older than
 the protocol version of the network. Please update nearcore', chain/client/src/client.rs:1011:21
Nov 17 16:46:55 near03 neard[1445878]: note: run with `RUST_BACKTRACE=1` environment variable to display a 
backtrace
Nov 17 16:46:55 near03 neard[1445878]: Nov 17 16:46:55.365 ERROR network: Received error sending message to
[...]

PR #6149, opened 2022-01-21
feat: Change `EpochManager` to use `Mutex` instead of `RwLock`

pmnoxx:
In general `Mutex` is faster on Unix, than `RwLock`.  We should change `epoch_manager` to use `Mutex` instead of `RwLock`.

### DIFF
--- a/nearcore/src/runtime/mod.rs
+++ b/nearcore/src/runtime/mod.rs
@@ -2,7 +2,7 @@ use std::cmp::Ordering;
 use std::collections::{HashMap, HashSet};
 use std::fs;
 use std::path::Path;
-use std::sync::{Arc, RwLock};
+use std::sync::{Arc, Mutex};
 
 use borsh::ser::BorshSerialize;
 use borsh::BorshDeserialize;
@@ -77,10 +77,10 @@ const STATE_DUMP_FILE: &str = "state_dump";
 const GENESIS_ROOTS_FILE: &str = "genesis_roots";
 
 /// Wrapper type for epoch manager to get avoid implementing trait for foreign types.
-pub struct SafeEpochManager(pub Arc<RwLock<EpochManager>>);
+pub struct SafeEpochManager(pub Arc<std::sync::Mutex<EpochManager>>);
 
-impl AsRef<RwLock<EpochManager>> for SafeEpochManager {
-    fn as_ref(&self) -> &RwLock<EpochManager> {
+impl AsRef<Mutex<EpochManager>> for SafeEpochManager {
+    fn as_ref(&self) -> &Mutex<EpochManager> {
         self.0.as_ref()
     }
 }
@@ -92,7 +92,7 @@ impl EpochInfoProvider for SafeEpochManager {
         last_block_hash: &CryptoHash,
         account_id: &AccountId,
     ) -> Result<Option<Balance>, EpochError> {
-        let mut epoch_manager = self.0.write().expect(POISONED_LOCK_ERR);
+        let mut epoch_manager = self.0.lock().expect(POISONED_LOCK_ERR);
         let slashed = epoch_manager.get_slashed_validators(last_block_hash)?;
         if slashed.contains_key(account_id) {
             return Ok(None);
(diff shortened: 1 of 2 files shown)


matklad:
Agree with this change, although I'd phrase the rationale differently: we are not taking advantage of `RwLock` capabilities (we only ever `.write`), so we should use `Mutex` to make the code simpler. 

If we really care about performance here, than we should switch to `parking_lot` (https://docs.rs/parking_lot/latest/parking_lot/).

matklad (approved):

pmnoxx:
@matklad I did benchmarks again:

```rust
   {
        let x = RwLock::new(0u64);
        c.bench_function("std::sync::RwLock", |b| {
            b.iter(|| {
                *x.write() += 1;
            });
        });
    }
    {
        let x = parking_lot::RwLock::new(0u64);
        c.bench_function("parking_log::RwLock", |b| {
            b.iter(|| {
                *x.write() += 1;
            });
        });
    }
    {
        let x = parking_lot::Mutex::new(0u64);
        c.bench_function("parking_log::Mutex", |b| {
            b.iter(|| {
                *x.lock() += 1;
            });
        });
    }
    {
        let x = std::sync::Mutex::new(0u64);
        c.bench_function("std::sync::Mutex", |b| {
            b.iter(|| {
                *x.lock().unwrap() += 1;
            });
        });
    }
```
And the results are:
```shell
std::sync::RwLock       time:   [3.6103 ns 3.6208 ns 3.6317 ns]       
parking_log::RwLock     time:   [3.5678 ns 3.5797 ns 3.5921 ns]  
parking_log::Mutex      time:   [3.8528 ns 3.8582 ns 3.8640 ns]     
std::sync::Mutex        time:   [4.7532 ns 4.7732 ns 4.7951 ns]    
```

I think we should leave `std::sync::RwLock` as it is.
There is no reason to use `parking_log::RwLock` over `std::sync::RwLock`.

I think it's better to simply use `std::sync::RwLock`, instead of `std::sync::Mutex ` rather than adding extra dependency for `parking_log::Mutex`

pmnoxx:
> Agree with this change, although I'd phrase the rationale differently: we are not taking advantage of `RwLock` capabilities (we only ever `.write`), so we should use `Mutex` to make the code simpler.
> 
> If we really care about performance here, than we should switch to `parking_lot` (https://docs.rs/parking_lot/latest/parking_lot/).

We should either consider switching `std::sync::Mutex` to `parking_log::Mutex` or simply use `std::sync::RwLock` for performance reasons.

matklad:
@pmnoxx I think in your benchmark, `std::sync::RwLock ` is actually `std::sync::Mutex` 

* it doesn't make sense that std::Mutex is slower than std::RwLock, as RwLock is strictly more powerful/has strictly more work to do
* you don't have `.unwrap()` there, which means that it's probably praking's lot lock, which doesn't have poisoning built-in.

pmnoxx:
@matklad You are right. Probably, we should switch to parking lot in that case, or at the very least, ship this PR.-

bowenwang1996:
I believe we intentionally chose to use the std locks because of lock poisoning. @matklad do you think we should switch to parking lot?

matklad:
No, I think we should stick with std. 

The main benefit of std is that it is not an external dependency, so less things to worry about. The second benefit is support for lock poisoning, but that I don't think matters for us, at least today. Lock poisoning matters when you write the code to recover from panics (catch_unwind somewhere). If you do want to outright crash the application on panic (as I think we currently do), than it's comparatively less important. Though, poisoning can still can help somewhat in this case if some panics do not crash the app, and I think this is the case for some of our code (In general, I am not at all sure what happens to neard if something somewhere panics. cc https://github.com/actix/actix/pull/111 as something relevant).

The main benefit of parking_lot is better performance. The second benefit is conceptual simplicity: `std::mutex` is pthread mutex, which carries POSIX and libc legacy, while parking lot directly does futex syscall, bypassing extra layers of abstraction. I don't think "performance" is a good reason to switch though: while parking lot is a clear winner in micro benchmark, it's not at all clear if will have any impact on the overall system's performance. Ie, if we spend 0.001% of time locking mutexes, making that 2x faster won't help. And, if high-level benchmark does show that mutexes are a problem, I'd argue that a better solution would be to change architecture to require less mutexes in the first place, and not to use faster mutexes. 

That is to say, **if** we care about mutex performance, we should be using parking_lot, but I will be very surprised if we care about mutex performance. Overall, the codebase doesn't seem nearly as well optimized as it should be to make a particular mutex implementation a significant performance problem.

bowenwang1996:
> If you do want to outright crash the application on panic (as I think we currently do), than it's comparatively less important

I am not sure that this is what we wanted. I believe to address https://github.com/near/nearcore/issues/5340, we want to catch panics and shut down neard gracefully to avoid database corruption. cc @nikurt @marcelo-gonzalez

matklad:
I would say that needing to catch panic to avoid DB corruptions feels very suspect. Usually, the way this is done is by making sure all necessary cleanup is implemented in `Drop`, and then letting the panic to properly terminate the process. Indeed, it seems that the problem with `#5340` is exactly that the panic *doesn't* crash the process, so the user have to abruptly kill the process with a signal, which doesn't run drops.

mm-near (approved):

PR #6151, opened 2022-01-21
Measure time of successful block processing

nikurt:
We already measure time of block processing, but only as little as 0.1% of block processing attempts are successful.

Ran a betanet nodeto test the change:
```
# HELP near_block_processed_total Total number of blocks processed successfully
# TYPE near_block_processed_total counter
near_block_processed_total 338
# HELP near_block_processing_attempts_total Total number of block processing attempts.
# TYPE near_block_processing_attempts_total counter
near_block_processing_attempts_total 23170
# HELP near_block_processing_time Time taken to process blocks successfully
# TYPE near_block_processing_time histogram
near_block_processing_time_bucket{le="0.005"} 272
near_block_processing_time_bucket{le="0.01"} 320
near_block_processing_time_bucket{le="0.025"} 336
near_block_processing_time_bucket{le="0.05"} 338
near_block_processing_time_bucket{le="0.1"} 338
near_block_processing_time_bucket{le="0.25"} 338
near_block_processing_time_bucket{le="0.5"} 338
near_block_processing_time_bucket{le="1"} 338
near_block_processing_time_bucket{le="2.5"} 338
near_block_processing_time_bucket{le="5"} 338
near_block_processing_time_bucket{le="10"} 338
near_block_processing_time_bucket{le="+Inf"} 338
near_block_processing_time_sum 1.2205153949999998
near_block_processing_time_count 338
```


mzhangmzz:
Maybe it's better to add the metric inside `process_block_single`, instead of at places it is called?

chain/chain/src/chain.rs
```diff
@@ -905,6 +905,9 @@ impl Chain {
     {
         let block_hash = *block.hash();
         let timer = metrics::BLOCK_PROCESSING_TIME.start_timer();
+        let instant = Instant::now();
+        // We measure time once, because the histogram measures time of all block processing
+        // attempts, and filtering needs to be done manually.
```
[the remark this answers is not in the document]
matklad: Oh, one more version which think would definitelly be strinctly better than the current one, as it fires up only one timer: 

```diff
diff --git a/chain/chain/src/chain.rs b/chain/chain/src/chain.rs
index fdfe9335b..31afcf71f 100644
--- a/chain/chain/src/chain.rs
+++ b/chain/chain/src/chain.rs
@@ -905,7 +905,6 @@ impl Chain {
     {
         let block_hash = *block.hash();
         let timer = metrics::BLOCK_PROCESSING_TIME.start_timer();
-        let instant = Instant::now();
         // We measure time once, because the histogram measures time of all block processing
         // attempts, and filtering needs to be done manually.
         let res = self.process_block_single(
@@ -917,10 +916,10 @@ impl Chain {
             block_orphaned_with_missing_chunks,
             on_challenge,
         );
-        timer.observe_duration();
+        let t = timer.stop_and_record();
         if res.is_ok() {
             metrics::BLOCK_PROCESSED_SUCCESSFULLY_TOTAL.inc();
-            metrics::BLOCK_SUCCESSFUL_PROCESSING_TIME.observe(instant.elapsed().as_secs_f64());
+            metrics::BLOCK_SUCCESSFUL_PROCESSING_TIME.observe(t);
 
             if let Some(new_res) = self.check_orphans(
                 me,
```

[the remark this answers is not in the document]
nikurt: Awesome, thanks

chain/chain/src/chain.rs
```diff
@@ -905,6 +905,9 @@ impl Chain {
     {
         let block_hash = *block.hash();
         let timer = metrics::BLOCK_PROCESSING_TIME.start_timer();
+        let success_timer = metrics::BLOCK_SUCCESSFUL_PROCESSING_TIME.start_timer();
+        // We measure time once, because the histogram measures time of all block processing
+        // attempts, and filtering needs to be done manually.
```
matklad, line 910: I feel like this comment is better be placed on `pub static BLOCK_SUCCESSFUL_PROCESSING_TIME:`

chain/chain/src/metrics.rs
```diff
@@ -15,6 +15,13 @@ pub static BLOCK_PROCESSED_SUCCESSFULLY_TOTAL: Lazy<IntCounter> = Lazy::new(|| {
...
+pub static BLOCK_SUCCESSFUL_PROCESSING_TIME: Lazy<Histogram> = Lazy::new(|| {
+    try_create_histogram(
+        "near_block_successful_processing_time",
+        "Time taken to successfully process blocks",
+    )
+    .unwrap()
+});
 pub static BLOCK_PROCESSING_TIME: Lazy<Histogram> = Lazy::new(|| {
     try_create_histogram("near_block_processing_time", "Time taken to process blocks").unwrap()
 });
```
matklad, line 27:
```
BLOCK_PROCESSED_TOTAL
BLOCK_PROCESSED_SUCCESSFULLY_TOTAL
BLOCK_SUCCESSFUL_PROCESSING_TIME
BLOCK_PROCESSING_TIME
```

successs/no-succes order of statcis is inconsistent

word-order within is also inconsistent, but not sure how to fix that

bowenwang1996 (commented):
Why do we care about unsuccessful ones? Could we just remove the other metric?

chain/chain/src/metrics.rs
```diff
@@ -18,6 +18,13 @@ pub static BLOCK_PROCESSED_SUCCESSFULLY_TOTAL: Lazy<IntCounter> = Lazy::new(|| {
...
+pub static BLOCK_PROCESSING_SUCCESSFULLY_TIME: Lazy<Histogram> = Lazy::new(|| {
+    try_create_histogram(
+        "near_block_successful_processing_time",
```
matklad, line 23: Var name `BLOCK_PROCESSING_SUCCESSFULLY_TIME` should probably be consistent with counter name `near_block_successful_processing_time`

nikurt:
@bowenwang1996 good point.
Re-purposed `BLOCK_PROCESSING_TIME` for the successful attempts.

chain/chain/src/chain.rs
```diff
@@ -1575,7 +1577,7 @@ impl Chain {
                 debug!(target: "chain", "Check orphans: found {} orphans", orphans.len());
                 for orphan in orphans.into_iter() {
                     let block_hash = orphan.hash();
-                    let timer = metrics::BLOCK_PROCESSING_TIME.start_timer();
+                    let success_timer = metrics::BLOCK_PROCESSING_TIME.start_timer();
```
mm-near, line 1580: maybe we should simply put the timer inside process_block_single?

(or are there too many returns in that function?)

chain/chain/src/metrics.rs
```diff
@@ -4,19 +4,23 @@ use near_metrics::{
 };
 use once_cell::sync::Lazy;
 
-pub static BLOCK_PROCESSED_TOTAL: Lazy<IntCounter> = Lazy::new(|| {
-    try_create_int_counter("near_block_processed_total", "Total number of blocks processed")
-        .unwrap()
+pub static BLOCK_PROCESSING_ATTEMPTS_TOTAL: Lazy<IntCounter> = Lazy::new(|| {
+    try_create_int_counter(
+        "near_block_processing_attempts_total",
+        "Total number of block processing attempts.",
```
mm-near, line 10: maybe explain how 'attempts' differ from 'successful processing' below?

mm-near (changes requested):
I'm not sure if we're putting the monitoring in the right place - it feels that it should be inside the process_block_single function rather than in places that are calling it

chain/chain/src/metrics.rs
```diff
@@ -4,19 +4,23 @@ use near_metrics::{
...
+    )
+    .unwrap()
 });
-pub static BLOCK_PROCESSED_SUCCESSFULLY_TOTAL: Lazy<IntCounter> = Lazy::new(|| {
+pub static BLOCK_PROCESSED_TOTAL: Lazy<IntCounter> = Lazy::new(|| {
     try_create_int_counter(
-        "near_block_processed_successfully_total",
+        "near_block_processed_total",
```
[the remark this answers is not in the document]
nikurt: I really tried finding the usage of this metric in grafana, and found none. Can't claim that my search was exhaustive, it's hard to click on every graph of every dashboard :)

Preserving the history of a metric is usually a good idea. In this case losing history should be acceptable, because I haven't found a single use case for the original metric.

chain/chain/src/chain.rs
nikurt: A timer outside is simpler, because the function has 5 returns: 4 question marks and the end of the function.

matklad: It feels odd that we have some metrics measured on the inside, and some on the outside. I'd say yeah, it probably does make sense to split the function as 

```rust
    fn process_block_single<F, F1, F2, F3>(
        &mut self,
        me: &Option<AccountId>,
        block: MaybeValidated<Block>,
        provenance: Provenance,
        mut block_accepted: F,
        mut block_misses_chunks: F1,
        mut orphan_misses_chunks: F2,
        on_challenge: F3,
    ) -> Result<Option<Tip>, Error>
    where
        F: FnMut(AcceptedBlock),
        F1: Copy + FnMut(BlockMissingChunks),
        F2: Copy + FnMut(OrphanMissingChunks),
        F3: FnMut(ChallengeBody),
    {
        metrics::BLOCK_PROCESSED_TOTAL.inc();
        metrics::NUM_ORPHANS.set(self.orphans.len() as i64);
        let res = self.process_block_single_impl(
            me,
            block,
            provenance,
            block_accepted,
            block_misses_chunks,
            orphan_misses_chunks,
            on_challenge,
        );
        if res.is_ok() {
        } else {
        }
        res
    }

    fn process_block_single_impl<F, F1, F2, F3>(
        &mut self,
        me: &Option<AccountId>,
        block: MaybeValidated<Block>,
        provenance: Provenance,
        mut block_accepted: F,
        mut block_misses_chunks: F1,
        mut orphan_misses_chunks: F2,
        on_challenge: F3,
    ) -> Result<Option<Tip>, Error>
    where
        F: FnMut(AcceptedBlock),
        F1: Copy + FnMut(BlockMissingChunks),
        F2: Copy + FnMut(OrphanMissingChunks),
        F3: FnMut(ChallengeBody),
    {

        let prev_head = self.store.head()?;
```

It looks very ugly because the signature is horrendous, but that seems like an orthogonal problem, tried to tackle that a bit in https://github.com/near/nearcore/pull/6206.

nikurt: Done

matklad (approved):

mzhangmzz (approved):

chain/chain/src/chain.rs
```diff
@@ -1221,6 +1217,45 @@ impl Chain {
...
+        if res.is_ok() {
+            metrics::BLOCK_PROCESSED_TOTAL.inc();
+            success_timer.stop_and_record();
+        } else {
+            success_timer.stop_and_discard();
+        }
+        res
+    }
+
+    fn process_block_single_impl<F, F1, F2, F3>(
```
mm-near, line 1258: please add comments on what this function does (and how does it differ from 'process_block_single')

mm-near (approved):
Looks great - thanks.

Issue #6100, opened 2022-01-18
Error during requesting partial encoded chunk


frol:
**Describe the bug**

After the latest testnet upgrade to 1.24.0-rc.1 there were a number of reports in Discord `#validators` channel seeing the following errors:

```
Jan 17 19:22:31.651 ERROR chunks: Error during requesting partial encoded chunk: DB Not Found Error: 9L9ekNWCDhzRXBNx3TPu5zBTzQgAAxZogA66jqwh59jh
 Cause: Unknown
Jan 17 19:22:31.652 ERROR chunks: Error during re-requesting partial encoded chunk. Cannot determine whether to request from an archival node, defaulting to not: DB Not Found Error: 9L9ekNWCDhzRXBNx3TPu5zBTzQgAAxZogA66jqwh59jh
 Cause: Unknown
```

**To Reproduce**

Probably, just run 1.24.0-rc.1 release.

**Expected behavior**

No errors should happen

**Screenshots**

<img width="826" alt="image" src="https://user-images.githubusercontent.com/304265/149876212-c6a1769d-962c-4e44-9960-9c03a0053660.png">

**Version (please complete the following information):**
- nearcore commit/branch: 1.24.0-rc.1
- testnet

mzhangmzz:
I found where the bug is and I will submit the fix soon. There is no serious issue, as in the bug does not affect the normal running of validator nodes. Please ignore the error messages for now until the fix

PR #6155, opened 2022-01-21
fix a bug in chunk requesting

mzhangmzz:
This resolves https://github.com/near/nearcore/issues/6100

The issue was, if a validator is very behind, it receives a chunk forward message before it processes the previous block of the chunk and before it sends requests for the chunk, the existing chunk forward handling logic adds a request to the request pool, but the request requires the previous block is processed. When the `resend_chunk_requests` is called in the future, it can't find the previous block and throws errors.

Test plan: 
- modified existing unit test to reproduce the error message locally and verified that it passes after the fix


chain/chunks/src/lib.rs
```diff
@@ -1477,9 +1483,6 @@ impl ShardsManager {
                 parts: forwarded_parts.into_iter().map(|(_, part)| part).collect(),
                 receipts: Vec::new(),
             };
-            // Make sure we mark this chunk as being requested so that it is handled
-            // properly in the next call. Note no requests are actually sent at this point.
-            self.request_chunk_single_mark_only(header);
```
mzhangmzz, line 1482: This line was for by-passing the check about `chunk_requested` at line 1409 for the following call of `self.process_partial_encoded_chunk`.

chain/chunks/src/lib.rs
```diff
@@ -1403,15 +1410,14 @@ impl ShardsManager {
...
-            // We shouldn't process unrequested chunk if we have seen one with same (height_created + shard_id)
+            // We shouldn't process unrequested chunk if we have seen one with same (height_created + shard_id) but different chunk_hash
             if let Ok(hash) = chain_store
                 .get_any_chunk_hash_by_height_shard(header.height_created(), header.shard_id())
             {
                 if *hash != chunk_hash {
                     warn!(target: "client", "Rejecting unrequested chunk {:?}, height {}, shard_id {}, because of having {:?}", chunk_hash, header.height_created(), header.shard_id(), hash);
                     return Err(Error::DuplicateChunkHeight.into());
                 }
-                return Ok(ProcessPartialEncodedChunkResult::Known);
```
mzhangmzz, line 1414: I'm not sure why we need to ignore incoming chunks that were not requested. We check the chunk header anyways, so this line seems unnecessary to me and caused issue for chunk forwarding

bowenwang1996 (approved):
Thanks!

PR #6158, opened 2022-01-21
feat(sandbox): fast forwarding

ChaoticTempest:
With this PR, we can now fast forward sandbox by a specific height. This is a delta supplied to the request as to not need to retrieve the current block height. This updates the doomslug tip to the newest advanced block height as that prevented producing blocks past a certain block w.r.t. time. Could have gotten rid of doomslug completely, but that just made block production speeds super rampant and no longer really mimicked what a live network would produce.

To see it in action, add a line to print out what block height we are at in the node:

```diff
diff --git a/chain/client/src/client_actor.rs b/chain/client/src/client_actor.rs
index eb1858420..88f78d0b9 100644
--- a/chain/client/src/client_actor.rs
+++ b/chain/client/src/client_actor.rs
@@ -962,6 +962,7 @@ impl ClientActor {
     /// Produce block if we are block producer for given `next_height` height.
     /// Can return error, should be called with `produce_block` to handle errors and reschedule.
     fn produce_block(&mut self, next_height: BlockHeight) -> Result<(), Error> {
+        println!("Producing: {}", next_height);
         match self.client.produce_block(next_height) {
             Ok(Some(block)) => {
                 let peer_id = self.node_id.clone();

```

Compile and run it:
```
make sandbox

./target/debug/neard-sandbox --home sandbox_home init
./target/debug/neard-sandbox --home sandbox_home run
```

Then fast-forward 10000 blocks via curl:
```
curl http://localhost:3030 -H 'content-type: application/json' -d '{"jsonrpc": "2.0", "id":1, "method":"sandbox_fast_forward", "params":{ "delta_height": 10000 }}'
```

@miraculous, after this PR, we'll need to add in fast-forward feature into `near-jsonrpc-client` for sandbox. Let me know if I can help out with that


chain/network-primitives/src/types.rs
```diff
@@ -29,6 +29,9 @@ use std::time::Duration;
 use strum::AsStaticStr;
 use tokio::net::TcpStream;
 
+#[cfg(feature = "sandbox")]
+use near_primitives::types::BlockHeightDelta;
+
```
pmnoxx, line 34: This import is only used once. Let's inline this import, and use the full import path inside `NetworkSandboxMessage`.

pmnoxx (changes requested):

chain/client/src/client_actor.rs
```diff
@@ -20,6 +20,8 @@ use near_chain::chain::{
     BlockCatchUpResponse, StateSplitRequest, StateSplitResponse,
 };
 use near_chain::test_utils::format_hash;
+#[cfg(feature = "sandbox")]
```
pmnoxx, line 23: I think it may be a good idea to avoid using conditional imports.
I know, that we still use conditional imports here, but in general we are trying to phase it out.

pmnoxx (approved):
LGTM

Adding approve to unblock, but I still think that we should use qualified imports, rather than adding more conditional imports.

chain/client/src/client_actor.rs
ChaoticTempest: Sounds good, I moved all the other conditional imports inplaced into their respective areas

ailisp (approved):
Nice! Very neat implementation

chefsale (approved):

miraclx (approved):

bowenwang1996 (commented):
Would be nice to have an integration test that verifies a node with sandbox enabled can run and you can interact with it through rpc. This can be run in CI to make sure that we don't regress sandbox

mikedotexe:
I just want to say that I'm very excited. And I appreciate you all. Can't wait til we have this, it will be the final step toward having a singular, powerful testing framework for our protocol. Massive massive deal, and thanks to everyone involved

khorolets (approved):

bowenwang1996:
> @bowenwang1996 I thought `integration-tests/src/tests/client/sandbox.rs::test_fast_forward` would be equivalent to that besides processing the request through rpc. Is there some other way I'm missing out on? Maybe using `NodeCluster` then?

No this does not test the node as a whole. We should run a node like how it is done in the python tests and interact it through rpc instead of testing through internals.

ChaoticTempest:
@bowenwang1996 got it, let me know if the python test I added works with what you were thinking of

chain/chain/src/chain.rs
```diff
@@ -3947,11 +3948,15 @@ impl<'a> ChainUpdate<'a> {
...
         let is_next = block.header().prev_hash() == &head.last_block_hash;
 
-        // A heuristic to prevent block height to jump too fast towards BlockHeight::max and cause
-        // overflow-related problems
-        let block_height = block.header().height();
-        if block_height > head.height + self.epoch_length * 20 {
-            return Err(ErrorKind::InvalidBlockHeight(block_height).into());
+        // Sandbox allows fast-forwarding, so only enable when not within sandbox
+        #[cfg(not(feature = "sandbox"))]
+        {
```
mina86, line 3953: This could be written as `if !cfg!(feature = "sandbox") {` which would mena `epoch_length` does not need to be marked as dead code.

integration-tests/src/tests/client/sandbox.rs
```diff
@@ -109,3 +118,40 @@ fn test_patch_account() {
     let test1_after = env.query_account("test1".parse().unwrap());
     assert_eq!(test1_after.amount, 10);
 }
+
+#[test]
+#[cfg(feature = "sandbox")]
+fn test_fast_forward() {
```
mina86, line 124: FYI, this test will be pretty much never executed. BuildKite pipeline does not include sandbox when running tests.

pytest/tests/sandbox/fast_forward.py
```diff
@@ -0,0 +1,50 @@
+#!/usr/bin/env python3
+# Fast forward sandbox
```
mina86, line 2: Please describe *exactly* what the test is doing, why it is doing that and what are expected results.

Also, this test needs to be added to `nightly/sandbox.txt`.

pytest/tests/sandbox/fast_forward.py
```diff
@@ -0,0 +1,50 @@
...
+        for exe in ('neard-sandbox', 'near-sandbox', 'neard'):
+            if (root / exe).exists():
+                logger.info(
+                    f'Using {(root / exe).relative_to(repo_dir)} binary')
+                config['near_root'] = str(root)
+                config['binary_name'] = exe
+                return config
+
+    assert False, ('Unable to figure out location of neard-sandbox binary; '
+                   'Did you forget to run `make sandbox`?')
```
mina86, line 33: We need to move this to a module so we don’t duplicate the same code in multiple tests.

mina86 (changes requested):

integration-tests/src/tests/client/sandbox.rs
ChaoticTempest: I see, would there be anything extra I'd need to do for this? If not, the python tests covers the same case either way.

integration-tests/src/tests/client/sandbox.rs
```diff
@@ -109,3 +118,40 @@ fn test_patch_account() {
     let test1_after = env.query_account("test1".parse().unwrap());
     assert_eq!(test1_after.amount, 10);
 }
+
+#[test]
+#[cfg(feature = "sandbox")]
```
mina86, line 123: By the way, `#[cfg(feature = "sandbox")]` isn’t necessary. Neither here nor anywhere else in the file. The file is conditionally included in `mod.rs` and won’t be compiled if `sandbox` feature is not enabled.

integration-tests/src/tests/client/sandbox.rs
mina86: We need to add `cargo test -p integration-tests --features sandbox -- tests::client::sandbox::` somewhere to `.buildkite/pipeline.yml`. Probably to ‘cargo test integration-tests*’ step. Maybe also to nightly as well; not sure if we care.

pytest/tests/sandbox/patch_state.py
```diff
@@ -2,42 +2,16 @@
...
 import base64
 import pathlib
 
 sys.path.append(str(pathlib.Path(__file__).resolve().parents[2] / 'lib'))
 
 from cluster import start_cluster
-from configured_logger import logger
 from transaction import sign_deploy_contract_tx, sign_function_call_tx
-from utils import load_test_contract
+from utils import load_test_contract, figure_out_sandbox_binary
```
mina86, line 12: Alternatively change this to `import utils` and use `utils.load_test_contract` and `utils.figure_out_sandbox_binary` in the code.

mina86 (approved):

bowenwang1996 (approved):

integration-tests/src/tests/client/sandbox.rs
ChaoticTempest: yeah, I'll just leave them out for now then. From the pytests already covering these cases, it shouldn't really matter. We can revisit in the future

Issue #6137, opened 2022-01-20
node cannot catch up after the 1.24.0-rc.1 release


bowenwang1996:
An archival node that runs 1.24.0-rc.1 release on mainnet cannot catch up with the network. It processes blocks at a speed of less than 0.5 bps with relatively high cpu usage (around 200%). We should investigate what happened and whether the new logic for block catchup introduced in 1.24.0-rc.1 is problematic.

mzhangmzz:
We figured out the issue, it's because when a node is sufficiently behind, it starts to ignore blocks sent by other nodes, but only relies on requesting block one at a time. This is inefficient. I'm working on a fix that makes the node request for more than block in blocks catching up mode.

PR #6160, opened 2022-01-21
change block sync to request 5 blocks at a time

mzhangmzz:
Fixes https://github.com/near/nearcore/issues/6137

This PR also fixes a bug in the existing block requesting logic, that it only checks if a block is in store, which means, block sync will request blocks that are stored in the orphan or the MissingChunks pool

I also reorganized the Unfit error kind. 

Test Plan:
- modified existing unit test
- cherry-picked the PR on top of 1.24, while setting the number of blocks for which we request chunks concurrently to 2. Ran the binary on testnet and mainnet. Checked that the catching up  speed. There should be no difference in bps between when the node is more than 500 blocks behind and when it is less than. There was a difference before because when the node was 500 blocks behind, it discards all incoming blocks and only relies on block sync to request blocks. With this PR and concurrent chunk requesting, bps for catching up on testnet was improved from ~4 to ~9 and from ~1.5 to ~2.5 on mainnet (the stats on mainnet is only an approximate since mainnet's blocks are more full and bps can depend on the load)


chain/client/src/sync.rs
```diff
@@ -41,6 +41,9 @@ pub const MAX_BLOCK_HEADERS: u64 = 512;
 /// Maximum number of block header hashes to send as part of a locator.
 pub const MAX_BLOCK_HEADER_HASHES: usize = 20;
 
+/// Maximum number of block requested at once in BlockSync
+const MAX_BLOCK_REQUESTS: usize = 3;
```
bowenwang1996, line 45: Why 3?

chain/chain-primitives/src/error.rs
```diff
@@ -70,9 +70,9 @@ pub struct Error {
 
 #[derive(Clone, Eq, PartialEq, Debug, Fail)]
 pub enum ErrorKind {
-    /// The block doesn't fit anywhere in our chain.
-    #[fail(display = "Block is unfit: {}", _0)]
```
mzhangmzz, line 74: I think this unfit message is very confusing, it sounds like the block is invalid, so I renamed it.

chain/client/src/sync.rs
mzhangmzz: Requesting many blocks in advance won't help that much because the bottleneck is at requesting chunks. Since now we know that we can't afford to request chunks for 5 blocks, setting this number to >= 5 doesn't make much sense. Out of the numbers that are >=2 but < 5, I chose 3 since it's the middle

chain/chain-primitives/src/error.rs
```diff
@@ -388,3 +388,25 @@ impl From<StorageError> for Error {
...
 }
+
+#[derive(Clone, Eq, PartialEq, Debug, Fail)]
+pub enum BlockKnownError {
+    KnownInHeader,
+    KnownInHead,
+    KnownInOrphan,
+    KnownInMissingChunks,
+    KnownInStore,
+}
```
bowenwang1996, line 399: 👍

bowenwang1996 (approved):

chain/client/src/sync.rs
```diff
@@ -515,47 +519,64 @@ impl BlockSync {
...
+                ));
+            } else {
+                warn!(target: "sync", "Block sync: {}/{} No available {}peers to request block {} from",
+                      head.height, header_head.height, if request_from_archival { "archival " } else { "" }, request.hash);
+            }
         }
 
-        self.last_request = Some(request);
+        // we know that requests is non-empty
+        self.last_request = Some(requests.pop().unwrap());
```
mm-near, line 579: What should we do if any of the request hit that 'warn' part ? 
(as in such case, the request was never sent, right?)

should we set the last_request to the last one that was actually successfully sent over the network?

chain/client/src/sync.rs
```diff
@@ -469,12 +472,13 @@ impl BlockSync {
             return Ok(true);
         }
 
-        let reference_hash = match &self.last_request {
-            Some(request) if chain.is_chunk_orphan(&request.hash) => request.hash,
-            _ => chain.head()?.last_block_hash,
-        };
-
+        // reference_hash is the last block on the canonical chain that is in store (processed)
```
mm-near, line 475: can you explain a little bit more about the 'last_request' path in this match? when does it happen?

mm-near (changes requested):

chain/chain/src/chain.rs
```diff
@@ -325,6 +325,69 @@ pub struct OrphanMissingChunks {
...
+        }
+        // Check if this block is in the set of known orphans.
+        if self.orphans().contains(block_hash) {
+            return Ok(Err(BlockKnownError::KnownInOrphan));
+        }
+        if self.blocks_with_missing_chunks().contains(block_hash) {
+            return Ok(Err(BlockKnownError::KnownInMissingChunks));
+        }
+        self.check_known_store(block_hash)
+    }
```
matklad, line 388: Non-trivial default method impls are a bit of an anti-pattern. As written, an impl of `ChainAccess` would be able to override this logic, although it really shouldn't. Slightly more purist way to write this would be to use a free-standing function: 

```rust
pub fn check_known(chan: &dyn ChainAccess, block_hash: &CryptoHash) -> Result<Result<(), BlockKnownError>, Error> {
   ...
}
```

But this doesn't really matter much.

chain/chain-primitives/src/error.rs
```diff
@@ -388,3 +388,25 @@ impl From<StorageError> for Error {
...
+        let output = match self {
+            BlockKnownError::KnownInHeader => "already known in header",
+            BlockKnownError::KnownInHead => "already known in head",
+            BlockKnownError::KnownInOrphan => "already known in orphan",
+            BlockKnownError::KnownInMissingChunks => "already known in missing chunks",
+            BlockKnownError::KnownInStore => "already known in store",
+        };
+        Display::fmt(&output, f)
+    }
+}
```
matklad, line 412:
```suggestion
impl fmt::Display for BlockKnownError {
    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
        let output = match self {
            BlockKnownError::KnownInHeader => "already known in header",
            BlockKnownError::KnownInHead => "already known in head",
            BlockKnownError::KnownInOrphan => "already known in orphan",
            BlockKnownError::KnownInMissingChunks => "already known in missing chunks",
            BlockKnownError::KnownInStore => "already known in store",
        };
        fmt::Display::fmt(&output, f)
    }
}
```

to minimize the amount of requried imports to jsut `fmt`

chain/client/src/sync.rs
```diff
@@ -515,47 +519,64 @@ impl BlockSync {
...
+                        break;
+                    }
+                    _ => return Err(e),
+                },
+            }
+            if let Ok(()) = chain.check_known(&next_hash)? {
+                let next_height = chain.get_block_header(&next_hash)?.height();
+                let request =
+                    BlockSyncRequest { height: next_height, hash: next_hash, when: Clock::utc() };
+                debug!(target:"sync", "add block {:?}", next_hash);
```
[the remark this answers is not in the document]
mzhangmzz: This debug print was added accidentally. Removed. I added height information in the other debug print below.

chain/client/src/sync.rs
```diff
@@ -515,47 +519,64 @@ impl BlockSync {
...
+                       head.height, header_head.height, request.hash, peer.peer_info.id, highest_height_peers.len());
+                self.network_adapter.do_send(PeerManagerMessageRequest::NetworkRequests(
+                    NetworkRequests::BlockRequest {
+                        hash: request.hash,
+                        peer_id: peer.peer_info.id.clone(),
+                    },
+                ));
+            } else {
+                warn!(target: "sync", "Block sync: {}/{} No available {}peers to request block {} from",
+                      head.height, header_head.height, if request_from_archival { "archival " } else { "" }, request.hash);
```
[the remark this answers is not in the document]
mzhangmzz: Yes I think so. If one of the peers catch up and have the block, the next request can succeed, otherwise the node will just keep waiting

chain/client/src/sync.rs
```diff
@@ -469,12 +472,13 @@ impl BlockSync {
...
 
-        let reference_hash = match &self.last_request {
-            Some(request) if chain.is_chunk_orphan(&request.hash) => request.hash,
-            _ => chain.head()?.last_block_hash,
-        };
-
+        // reference_hash is the last block on the canonical chain that is in store (processed)
         let reference_hash = {
+            let reference_hash = match &self.last_request {
+                Some(request) if chain.is_chunk_orphan(&request.hash) => request.hash,
```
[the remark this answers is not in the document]
mzhangmzz: Yeah I think that's what "chunk orphan" meant. Will rename it in a separate PR

chain/chain-primitives/src/error.rs
```diff
@@ -70,9 +70,9 @@ pub struct Error {
 
 #[derive(Clone, Eq, PartialEq, Debug, Fail)]
 pub enum ErrorKind {
-    /// The block doesn't fit anywhere in our chain.
-    #[fail(display = "Block is unfit: {}", _0)]
-    Unfit(String),
+    /// The block is already known
+    #[fail(display = "Block is known: {:?}", _0)]
+    BlockKnown(BlockKnownError),
```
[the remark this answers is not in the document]
mzhangmzz: That's a good point. For context, `process_block` returns Error in many cases that is not problematic in the sense that the block is invalid or something wrong occurred during the process. It simply means that the block was not processed successfully and the reason why it wasn't processed successfully is encoded in the error types. For example, the block is already known, or the block is an orphan (we don't know the previous block yet), or the block is invalid, all these cases correspond to some chain error type. And the caller of `process_block` will process these errors accordingly. It is definitely true that such info can be returned as value instead of as error, it is probably better that way too.

chain/client/src/sync.rs
mzhangmzz: honestly I think this case could be removed. When I digged to the original PR, it said "download either the block that builds directly on top of head, or one block ahead if the next block is already in the blocks with missing chunks.", which made sense. The last_request path was referring to the case "if the next block is already in the blocks with missing chunks". But then there is another PR later to fix a bug " if the current head is not on the canonical header chain, the block sync would continuously try to fetch a block that build on top of the current head, which is not what needs to be fetched". The fix added lines https://github.com/near/nearcore/blob/master/chain/client/src/sync.rs#L499-L516 to deal with the fork case, but now these lines made the "next block is already in the blocks with missing chunks" case obsolete because `reference_hash` now is the effective head in all cases.

I'll remove the case and add some comments in the code.

chain/client/src/sync.rs
mzhangmzz: Good point, changed it

PR #6161, opened 2022-01-21
re: Fix qualified imports inside `client_actor.rs`

pmnoxx:
We should remove usage of conditional imports from `chain/client/src/client_actor.rs` and replace them with `qualified` imports.


pmnoxx:
```
@bowenwang1996 @mm-near @mzhangmzz: chain/
```
@bowenwang1996 @mm-near @mzhangmzz Can you take a look?

matklad:
LGTM, though I'd probably ordered https://github.com/near/nearcore/pull/5436#pullrequestreview-814734494 before this PR, to not to have touch each delay detector invocation twice

pmnoxx:
> LGTM, though I'd probably ordered [#5436 (review)](https://github.com/near/nearcore/pull/5436#pullrequestreview-814734494) before this PR, to not to have touch each delay detector invocation twice

@matklad I we want to do it, here is the PR: https://github.com/near/nearcore/pull/6176

mm-near (approved):

PR #6164, opened 2022-01-24
test artifact stability with more contracts

nagisa:
The basic simple contract does not really cover all that much

### DIFF
--- a/runtime/near-test-contracts/Cargo.toml
+++ b/runtime/near-test-contracts/Cargo.toml
@@ -9,6 +9,9 @@ edition = "2021"
 license = "Apache-2.0"
 
 [dependencies]
-once_cell = "1"
+once_cell = "1.5.0"
 wat = "1.0.40"
-wasm-encoder = "0.10"
+wasm-encoder = "0.11.0"
+wasm-smith = "0.10.0"
+rand = { version = "0.8.0", features = ["small_rng"] }
+arbitrary = "1.1.0"
--- a/runtime/near-vm-runner/src/tests/cache.rs
+++ b/runtime/near-vm-runner/src/tests/cache.rs
@@ -103,19 +103,68 @@ fn test_wasmer2_artifact_output_stability() {
     //
     // Note that this test is a best-effort fish net. Some changes that should modify the hash will
     // fall through the cracks here, but hopefully it should catch most of the fish just fine.
-    let mut hasher = StableHasher::new();
-    let code = near_test_contracts::trivial_contract();
-    let config = VMConfig::test();
-    let prepared_code = prepare::prepare_contract(code, &config).unwrap();
-    let mut features = CpuFeature::set();
-    features.insert(CpuFeature::AVX);
-    let triple = "x86_64-unknown-linux-gnu".parse().unwrap();
-    let target = Target::new(triple, features);
-    let vm = Wasmer2VM::new_for_target(config, target);
-    let executable = vm.compile_uncached(&prepared_code).unwrap();
-    let serialized = executable.serialize().unwrap();
-    serialized.hash(&mut hasher);
-    assert_eq!(hasher.finish(), 2128441495928970645, "WASMER2_CONFIG needs version change");
+    let seeds = [2, 3, 5, 7, 11, 13, 17];
+    let prepared_hashes = [
+        4956511664052417334,
+        13266258595277126792,
+        11794508683191255560,
+        4769993623179083711,
+        10062359189500799005,
+        12997491198579188147,
+        4736731759770276858,
+    ];
+    let mut got_prepared_hashes = Vec::with_capacity(seeds.len());
+    let compiled_hashes = [
+        2482866579380424992,
+        12522401137476924800,
+        14535116066862320239,
+        14541379784010931226,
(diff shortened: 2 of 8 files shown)


runtime/near-vm-runner/src/tests/cache.rs
```diff
@@ -95,19 +95,30 @@ fn test_wasmer2_artifact_output_stability() {
...
-    let serialized = artifact.artifact().serialize().unwrap();
-    serialized.hash(&mut hasher);
-    assert_eq!(hasher.finish(), 18352148442716835594, "WASMER2_CONFIG needs version change");
+    let expected_hashes =
+        [18352148442716835594, 9541580911438661949, 1877160384252030064, 1939770487874739745];
+    let contracts = [
+        near_test_contracts::trivial_contract(),
+        near_test_contracts::rs_contract(),
+        near_test_contracts::ts_contract(),
+        near_test_contracts::fuzzing_contract(),
```
matklad, line 104: Hm, I think these three are expected to change somewhat frequently, and fixing the hashes can be a touch annoying. Even something like a rustc upgrade may change hashes, as we compile the contracts on the fly. 

I wonder if there's some smarter way to do this? For example, can we use wasm_smith's arbitrary to generate a deterministic, large contract?

nagisa: A good idea, I'll look into it.

nagisa:
So, initially there was a problem with `wasm-smith` generating modules that used features from proposals such as multi_value. These have been fixed, but now `parity-wasm` parser is finding itself unable to parse some valid WASM modules. I had encountered this sort of issue already in the past.

This is the sort of WASM it chokes on

```
0x0000 | 00 61 73 6d | version 1
       | 01 00 00 00
0x0008 | 04 08       | table section
0x000a | 01          | 1 count
0x000b | 70 01 c7 02 | [table 0] TableType { element_type: FuncRef, initial: 327, maximum: Some(274215) }
       | a7 de 10   
0x0012 | 05 01       | memory section
0x0014 | 00          | 0 count
0x0015 | 06 01       | global section
0x0017 | 00          | 0 count
0x0018 | 09 08       | element section
0x001a | 01          | 1 count
0x001b | 02 00       | element FuncRef table[0]
0x001d | 41 09       | I32Const { value: 9 }
0x001f | 0b          | End
0x0020 | 00 00       | 0 items

```

nagisa:
This I believe is now actually ready! It took a while, but I believe working on this helped us to make significant improvements all over the place.

runtime/near-test-contracts/Cargo.toml
```diff
@@ -11,3 +11,6 @@ license = "Apache-2.0"
 [dependencies]
 once_cell = "1"
 wat = "1.0.40"
+wasm-smith = "0.10.0"
+rand = { version = "0.8", features = ["small_rng"] }
+arbitrary = "1.1"
```
matklad, line 16: Very minor, but I now think it's good to specify patch version for dep as well https://users.rust-lang.org/t/psa-please-specify-precise-dependency-versions-in-cargo-toml/71277

runtime/near-vm-runner/src/tests/cache.rs
```diff
@@ -102,19 +102,58 @@ fn test_wasmer2_artifact_output_stability() {
...
+    // the implications of the change.
+    //
+    // May need a protocol version change, and definitely wants a `WASMER2_CONFIG version update
+    // too, as below. Maybe something else too.
+    assert_eq!(got_input_hashes, input_hashes, "contract preparation has changed");
+    // In this case you will need to adjust the WASMER2_CONFIG version so that the cached contracts
+    // are evicted from the contract cache.
+    assert_eq!(got_hashes, expected_hashes, "VM output has changed");
+    // Once it has been confirmed that these steps have been done, the expected hashes in this test
+    // can be adjusted.
```
matklad, line 156: Might through `insta` at this, though probably it isn't worth it here.

runtime/near-vm-runner/src/tests/cache.rs
```diff
@@ -102,19 +102,58 @@ fn test_wasmer2_artifact_output_stability() {
...
+    ];
+    let mut got_input_hashes = Vec::with_capacity(seeds.len());
+    let mut got_hashes = Vec::with_capacity(seeds.len());
+    for seed in seeds {
+        let mut hasher = StableHasher::new();
+        let contract = near_test_contracts::arbitrary_contract(seed);
+        let mut ch = StableHasher::new();
+        contract.hash(&mut ch);
+        let config = VMConfig::test();
+        let prepared_code = prepare::prepare_contract(&contract, &config).unwrap();
```
matklad, line 132: I wonder if we want to add

```
let hash = StableHasher::stable_hash((&contract, &prepared_code))
```

API for convenience, while we are at it? It's a bit hard to keep track of hasher and ch.

matklad (approved):

runtime/near-vm-runner/src/tests/cache.rs
nagisa: I made it better by reordering the code and adding newlines.

PR #6165, opened 2022-01-24
Revert to wasmer-near 2.1.0 for the 1.24.0 cycle

nagisa:


runtime/near-vm-runner/src/wasmer2_runner.rs
```diff
@@ -395,6 +392,15 @@ impl crate::runner::VM for Wasmer2VM {
             %method_name
         )
         .entered();
+        // NaN behavior is deterministic as of now: https://github.com/wasmerio/wasmer/issues/1269
+        // So doesn't require x86. However, when it is on x86, AVX is required:
+        // https://github.com/wasmerio/wasmer/issues/1567
+        #[cfg(not(feature = "no_cpu_compatibility_checks"))]
+        if (cfg!(target_arch = "x86") || !cfg!(target_arch = "x86_64"))
```
matklad, line 399: Wait, doesn't `cfg!(target_arch = "x86")` imply `!cfg!(target_arch = "x86_64")`? That is, can this be simplified to just `!cfg!(target_arch = "x86_64")`?

nagisa: This is code I pulled out from history from back when we had manual cpu checks (which was necessary because wasmer-near 2.1.0 does not have these internally yet.) Given that this code is gone in master anyway, I'm not entirely sure its worth the effort scrutinizing it.

matklad (approved):
LGTM!

Though I'd probably add a test to sanity check that:

* the problem exists before revert
* the revert fixes it

cc @posvyatokum @bowenwang1996 we want to get this into current testnet/future mainnet.

bowenwang1996 (approved):

PR #6168, opened 2022-01-24
chain: rosetta: guarantee order of operations in genesis block

mina86:
Because we were using `buffer_unordered` as well as a hash map the
order in which we consider accounts on genesis block is
nondeterministic.  This means that operations in the ‘genesis
transaction’ are in nondeterministic order as well.  This perhaps
wouldn’t be an issue except that we assign operation indexes to each
operation based on their position in the list.

This means, that querying genesis block multiple times may result in
transactions whose operation indexing changes.

To make sure that doesn’t happen and operation indexes remain
consistent, switch from using a hash map to a B-tree map.

Note that this fixes the issue for genesis block only.  Other blocks
still suffer from the ordering being nondeterministic.


frol (approved):
@mina86 Good catch!

frol:
P.S. perhaps, we need to add you and me to that rosetta.py file CODEOWNERS

mina86:
> P.S. perhaps, we need to add you and me to that rosetta.py file CODEOWNERS

Fun fact: I am.  [GitHub is just a broken mess](https://github.community/t/do-not-require-owner-approval-if-the-pull-request-is-from-an-owner/369)

PR #6176, opened 2022-01-24
Make `delay-detector` required, and make it's functionality flag dependent.

pmnoxx:
We can get rid of conditional imports from our code by making `delay-detected` a required dependency. However, it's functionality will be put under `delay_detector` flag.

### DIFF
--- a/chain/chain/Cargo.toml
+++ b/chain/chain/Cargo.toml
@@ -23,7 +23,7 @@ strum = "0.20"
 thiserror = "1.0"
 tracing = "0.1.13"
 
-delay-detector = { path = "../../tools/delay_detector", optional = true}
+delay-detector = { path = "../../tools/delay_detector"}
 near-chain-configs = { path = "../../core/chain-configs" }
 near-chain-primitives = { path = "../chain-primitives" }
 near-crypto = { path = "../../core/crypto" }
@@ -40,7 +40,7 @@ near-logger-utils = {path = "../../test-utils/logger"}
 byzantine_asserts = []
 expensive_tests = []
 test_features = []
-delay_detector = ["delay-detector"]
+delay_detector = ["delay-detector/delay_detector"]
 no_cache = ["near-store/no_cache"]
 protocol_feature_chunk_only_producers = [
   "near-chain-configs/protocol_feature_chunk_only_producers",
--- a/chain/network/src/peer_manager/peer_manager_actor.rs
+++ b/chain/network/src/peer_manager/peer_manager_actor.rs
@@ -1496,9 +1496,9 @@ impl PeerManagerActor {
         ctx: &mut Context<Self>,
         throttle_controller: Option<ThrottleController>,
     ) -> NetworkResponses {
-        #[cfg(feature = "delay_detector")]
-        let _d =
-            delay_detector::DelayDetector::new(format!("network request {}", msg.as_ref()).into());
+        let _d = delay_detector::DelayDetector::new(|| {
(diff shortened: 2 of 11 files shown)


chain/network/src/peer_manager/peer_manager_actor.rs
```diff
@@ -1953,7 +1952,6 @@ impl PeerManagerActor {
     #[perf]
     fn handle_msg_inbound_tcp_connect(&mut self, msg: InboundTcpConnect, ctx: &mut Context<Self>) {
         {
-            #[cfg(feature = "delay_detector")]
             let _d = delay_detector::DelayDetector::new("inbound tcp connect".into());
         }
```
matklad, line 1956: Hm, this feels wrong? Why we delay-detect an empty block?

tools/delay_detector/src/lib.rs
```diff
@@ -1,18 +1,24 @@
+#[cfg(features = "delay_detector")]
```
matklad, line 1:
```suggestion
#[cfg(feature = "delay_detector")]
```

every cfg in this file has this typo.

tools/delay_detector/src/lib.rs
```diff
@@ -75,3 +83,13 @@ impl<'a> Drop for DelayDetector<'a> {
         }
     }
 }
```
matklad, line 85: To avoid a lot of cfgs here, I'd do something like this: 

```
#[cfg(feature = "delay_detector")]
mod enabled;
#[cfg(feature = "delay_detector")]
pub use enabled::DelayDetector;

#[cfg(not(feature = "delay_detector"))]
mod disabled;
pub use disabled::DelayDetector;
```

chain/chain/Cargo.toml
```diff
@@ -40,7 +40,7 @@ near-logger-utils = {path = "../../test-utils/logger"}
 byzantine_asserts = []
 expensive_tests = []
 test_features = []
-delay_detector = ["delay-detector"]
+delay_detector = ["delay-detector/delay_detector"]
```
matklad, line 43: as a follow up, we might think about removing all these forwarding Cargo.features. Rather than having 

```toml
[features]
delay_detector = ["delay-detector/delay_detector"]
```

in every intermediate Cargo.toml, we now can have just one at the top level in `neard/Cargo.toml`, as that would enable delay detection globally.

pmnoxx: @matklad Yes, that's a good idea. I'll have to think about it. Let's move the discussion to a separate PR.

tools/delay_detector/src/lib.rs
pmnoxx: Fixed.

tools/delay_detector/src/lib.rs
```diff
@@ -1,76 +1,96 @@
...
-        if let Some((s, started)) = self.last_snapshot.take() {
-            self.snapshots.push((
-                (s, msg.to_string()),
-                Snapshot {
-                    real_time: now - started.real_time,
-                    cpu_time: started.cpu_time.elapsed(),
-                },
-            ));
+    impl<'a> DelayDetector<'a> {
+        pub fn new(msg: impl FnOnce() -> Cow<'a, str>) -> Self {
```
matklad, line 44: Good call to make this a lambda!

matklad (approved):

mzhangmzz (approved):

bowenwang1996 (approved):

PR #6177, opened 2022-01-25
bug: Change the way listening on port happens for `PeerManagerActor`

pmnoxx:
According to the official documentation in tokio https://docs.rs/tokio/latest/tokio/net/struct.TcpListener.html
You are supposed to stop the `loop` on the first error inside `listener.accept()` loop.

Let's change our code, to follow the official guidelines.

### DIFF
--- a/chain/network/src/peer_manager/peer_manager_actor.rs
+++ b/chain/network/src/peer_manager/peer_manager_actor.rs
@@ -209,6 +209,9 @@ impl Actor for PeerManagerActor {
                                 ),
                             );
                             debug!(target: "network", from = ?client_addr, "got new connection");
+                        } else {
+                            info!(at = ?server_addr, "Stopping listening");
+                            break;
                         }
                     },
                     Err(e) => {
--- a/chain/network/src/peer_manager/peer_manager_actor.rs
+++ b/chain/network/src/peer_manager/peer_manager_actor.rs
@@ -210,7 +210,7 @@ impl Actor for PeerManagerActor {
                             );
                             debug!(target: "network", from = ?client_addr, "got new connection");
                         } else {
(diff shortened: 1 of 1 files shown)


chain/network/src/peer_manager/peer_manager_actor.rs
```diff
@@ -209,6 +209,9 @@ impl Actor for PeerManagerActor {
                                 ),
                             );
                             debug!(target: "network", from = ?client_addr, "got new connection");
+                        } else {
+                            debug!(at = ?server_addr, "stopping listening");
```
mm-near, line 213: could we print the error inside the log?

mm-near (approved):

matklad:
Hm, is this actually a correct thing to do? Citing the doc (*emphasis mine*): 

> Note that accepting a connection can lead to various errors and not all of them are necessarily fatal ‒ for example having too many open file descriptors or **the other side closing the connection while it waits in an accept queue**. These would terminate the stream if not handled in any way.

It seems that this opens us up to an attack when a malicious peer ques up accepts, and then closes one connection before we are able to get to it, which would cause us to exit the loop and to become unable to accept connections from other peers?

matklad:
Looked a bit more into this, and I must say I am thoroughly confused as to what's the actual proper way to handle this. Asked the question on users.rust-lang.org:  https://users.rust-lang.org/t/how-to-handle-errors-when-accepting-connections/70838

pmnoxx:
@matklad You are right. I think we should solve the issue of potentially having a busy loop.
However, adding `break` is not the correct way to do it.

I think it's best to close this PR, and go back to this problem later, once we figure out the correct way to write this code.

PR #6179, opened 2022-01-25
feat: Make jsonrpc-primitives types derive `Debug`

itegulov:
This PR just adds `Debug` derivations to all jsonrpc-primitives types as we need them to be able to print RPC debug logs in [workspaces-rs](https://github.com/near/workspaces-rs).


miraclx (commented):
Nice work! If it isn't much of an ask, can we maybe extend this implementation to all the JSON RPC types?

EDIT: (clarifying: request/response/error types)

miraclx (approved):

Issue #5979, opened 2021-12-29
add rust version to `/status` endpoint


kucharskim:
Recently we had issues on different blockchain (Solana) where process was crash looping. With long streak of troubleshooting it turned out, Rust version was the culprit and not Solana codebase itself. Well both rust+solana, but we were only one of very few validators hitting problematic code path, because of Rust, used to compile Solana binary. Please add `rust` version to the `neard` binary, how it was compiled. For example here:
```
$ curl -s 10.130.0.52:3030/status | jq -Mcr .version
{"version":"1.23.0","build":"1.23.0"}
```

kucharskim:
Also, what I found useful, a commit hash of git repo checkout and it the checkout clean or not (local uncommitted changes).

Rishabhraghwendra18:
@bowenwang1996 I think I can help on this issue . I run the `cargo run -- --home ~/.near run` in `neard` directory . It started the node showing block height and all others stuff . But I don't get a way to interact with blocks (or something else) on which I can make a request to `/status` endpoint to reproduce the issue to know how it's currently . May you tell me how can I see the `/status` endpoint and in which file I have to make the changes ?

bowenwang1996:
> @bowenwang1996 I think I can help on this issue . I run the `cargo run -- --home ~/.near run` in `neard` directory . It started the node showing block height and all others stuff . But I don't get a way to interact with blocks (or something else) on which I can make a request to `/status` endpoint to reproduce the issue to know how it's currently . May you tell me how can I see the `/status` endpoint and in which file I have to make the changes ?

It can be accessed at `http://localhost:3030/status`

mina86:
> Also, what I found useful, a commit hash of git repo checkout and if the checkout is clean or not (local uncommitted changes).

That’s `build` in the status, no?

kucharskim:
not sure does it say repo was clean or not, I didn't test it, but yeah, that hash below is from a repo, without letter `g`:

```
{
  "version": "1.24.0-rc.1",
  "build": "crates-0.11.0-67-gfdd3cffea"
}
```

```
nearcore@near04:~/nearcore$ git log -n1 fdd3cffea | wc -c
733
```

kucharskim:
I think having a hash should always be there. I cannot experiment now, but `trunk` doesn't really say anything, right? Trunk today, tomorrow, last month, next year? It cannot be translated into specific checkout.

mina86:
[...]

PR #6180, opened 2022-01-25
client: add rustc_version to /status page

mina86:
Issue: https://github.com/near/nearcore/issues/5979

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2605,6 +2605,8 @@ dependencies = [
  "once_cell",
  "rand 0.7.3",
  "reed-solomon-erasure",
+ "rustc_version_runtime",
+ "semver 0.9.0",
  "serde_json",
  "strum",
  "sysinfo",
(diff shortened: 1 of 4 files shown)


matklad:
i'd rather write our own `build.rs` than use `rustc_version_runtime`:

* that crate doesn't seem particularly popular
* it uses semver 0.9, and semver 1.0 is a big improvement over that (dtolnay took over the maintenance and rewrote it to dtolnay's standard of quality)
* it just does a lot of extra stuff -- I don't think we want to parse rustc version, it seems best to just include the literal output of `rustc --version` (maaaybe `rustc --version --verbose`)
* it doesn't do enough of stuff -- there's no reason why this needs to use `once_cell` on top. This is a bit of static data, there shouldn't be any need for runtime initialization here.

matklad:
Unrelated thing: we have the following bit of code: 

https://github.com/nearprotocol/nearcore/blob/34eef742285287ba384b192c4240af64dc767c4e/neard/src/main.rs#L16-L33

it seems to do conceptually the same thing. So it feels like both that (getting version of `neard` itself) and this (getting version of the compiler/standard library used to build near) should be implemented in a same, single place, and be exposed both via `RPC` and via `neard --version`. 

But overall big :+1: obviously!

mina86:
> i'd rather write our own build.rs than use rustc_version_runtime:

I thought about that but consider that we already transitively depend on `rustc_version` (which `rustc_version_runtime` is just a run-time wrapper around).

matklad:
The fact that it depends on `rustc_version 0.2` rather than current one `rustc_version 0.4` is sort-of red flag for me. Given how simple the actual thing we want to achieve, it seems that doing that ourselves would be less headache down the line.

PR #6181, opened 2022-01-25
fix: support relative TARGET_DIR in near-test-contracts

matklad:
near-test-contracts builds some wasm contracts for use in testing. It
does so by recursively invoking `cargo` from `build.rs`. Before this
commit, we tried to re-use parent's `CARGO_TARGET_DIR` to figure out
where we should put the data. That was rather hacky, as cargo doesn't
expose that information to the build scripts in a reliable way:

rust-lang/cargo#9661 (comment)

Naturally, our hacks broken when when the `CARGO_TARGET_DIR` was set to
a relative path, because build.rs doesn't know where workspace root
lives.

The fix is to use `OUT_DIR` rather than `CARGO_TARGET_DIR`, which I
think is the supported way to this in the first place. Eg, the `cc`
crate uses `OUT_DIR` to store intermediate `.o` files, which I think
matches our use case pretty closely.


Longarithm:
It seems to be conflicting with https://github.com/near/nearcore/pull/6171

matklad:
Yeah, I think this PR is still an improvement over that:

* overall, the approach is simpler and should be more robust: `OUT_DIR` is something supported by Cargo, while fetching TARGET_DIR is hacky
* this way, we'll actually honor the `TARGET_DIR` set from the outside (after #6171, the code would actually be build into a local directory in src)

Longarithm (approved):

matklad:
@nikurt could you merge this when the time is convenient? I thing this PR is a good idea, but, if #6171 works ok, and we are in the process of releasing something (which I don't know), I don't want to add extra posibility of something's going wrong just now.

chefsale (approved):

nikurt (approved):

PR #6182, opened 2022-01-25
Update to wasmer-near 2.2.0

nagisa:
This pulls in a fair amount of development that has occurred in the fork
over the past couple weeks.

Most notably it has a fix for a crash, functionality to simplify
profiling of the code generated by JIT, a stack limiting safeguard, and
a removal of baked in CPU capability checks at instantiation time, 
thus allowing cost estimation work again.

cc @jakmeier @olonho

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5517,9 +5517,9 @@ dependencies = [
 
 [[package]]
 name = "wasmer-compiler-near"
-version = "2.1.2"
+version = "2.2.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "2992fc265c0207ea423d6565a27adf1a01210eae4bb4db7aa18778a2b84f588d"
+checksum = "c11074b5b8f4170b5ebf0744e811728befb01a70757395c43b528b6441e9c924"
 dependencies = [
  "enumset",
  "loupe",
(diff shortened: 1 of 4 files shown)


jakmeier:
Ah yes, then I can stop using the development branch for my estimations! :)

matklad (approved):

matklad:
LGTM! We are early in the release cycle, so it's the perfect opportunity to land this!

@Ekleog after the PR lands, could you look into fuzzing this a bit more thoroughly than our unusual CI does? Might be a good moment to look into https://github.com/near/nearcore/issues/6184 as well.

matklad:
To spell this more explicitly, for the current release we want to merge https://github.com/near/nearcore/pull/6165, but we won't be forward-porting that to master, as this PR supersedes that fix.

Ekleog:
I'll do! Just, to get an idea of the timeline, when does the release cycle end? (to know whether I should keep working on the correct infra for nayduck machines, or just spin up a gcp vm and have it fuzz because we're in a hurry − can also do both if the cycle is short-ish)

nagisa:
By my estimation there're ~4 weeks before this specific PR gets into testnet and 10 weeks before mainnet release, if I understand our release process correctly.

Ekleog:
I can confirm a few weeks of fuzzing didn't find anything here, and hopefully we'll have continuous fuzzing running on the whole nayduck infrastructure soon enough to make sure no regression happens even without manual fuzzing :)

PR #6190, opened 2022-01-25
Add more unit test for ShardsManager

mzhangmzz:


chain/chunks/src/lib.rs
```diff
@@ -1428,7 +1428,9 @@ impl ShardsManager {
                 // validate_chunk_header returns DBNotFoundError if the previous block is not ready
                 // in this case, we return NeedBlock instead of error
                 ErrorKind::DBNotFoundErr(_) => {
-                    return Ok(ProcessPartialEncodedChunkResult::NeedBlock)
+                    debug!(target:"client", "Dropping partial encoded chunk {:?} because we don't have enough information to validate it",
```
mm-near, line 1431: is there an easy way to print the block height too?

chain/chunks/src/lib.rs
```diff
@@ -1472,6 +1474,7 @@ impl ShardsManager {
 
         // Merge parts and receipts included in the partial encoded chunk into chunk cache
         self.encoded_chunks.merge_in_partial_encoded_chunk(partial_encoded_chunk);
+        debug!(target:"client", "merged {:?}", self.encoded_chunks.get(&header.chunk_hash()).unwrap().parts);
```
mm-near, line 1477: same here - having block height helps debugging a lot

chain/chunks/src/lib.rs
```diff
@@ -2290,11 +2294,77 @@ mod test {
         assert!(seals_manager.past_seals.get(&fixture.mock_height).is_none());
     }*/
 
+    #[test]
+    fn test_resend_chunk_requests() {
+        // Test that resending chunk requests won't request for parts the node already receives
```
mm-near, line 2299: s/receives/received?

chain/chunks/src/lib.rs
```diff
@@ -2290,11 +2294,77 @@ mod test {
...
+                    _ => {}
+                }
+            }
+            parts
+        };
+        let requested_parts = collect_request_parts(&mut fixture);
+        assert_eq!(requested_parts, (1..fixture.mock_chunk_parts.len() as u64).collect());
+
+        // process chunk part 1
+        let partial_encoded_chunk = fixture.make_partial_encoded_chunk(&[1]);
```
mm-near, line 2344: BTW - what happens when we get some part with id > fixture.mock_chunk_parts.len() ?

chain/chunks/src/lib.rs
```diff
@@ -2290,11 +2294,77 @@ mod test {
...
+                MaybeValidated::from(&partial_encoded_chunk),
+                Some(&fixture.mock_chain_head),
+                &mut fixture.chain_store,
+                &mut fixture.rs,
+            )
+            .unwrap();
+        assert_matches!(result, ProcessPartialEncodedChunkResult::NeedBlock);
+
+        // resend request and check chunk part 0 and 1 are not requested again
+        std::thread::sleep(Duration::from_millis(2 * CHUNK_REQUEST_RETRY_MS));
```
mm-near, line 2356: in the future - we should consider passing a 'timer' to the resend_chunk_requests instead

chain/chunks/src/lib.rs
```diff
@@ -2290,11 +2294,77 @@ mod test {
...
+        shards_manager.resend_chunk_requests(&fixture.mock_chain_head);
+
+        let requested_parts = collect_request_parts(&mut fixture);
+        assert_eq!(requested_parts, (2..fixture.mock_chunk_parts.len() as u64).collect());
+    }
```
mm-near, line 2361: maybe run resend_chunk_request again - and check that it doesn't request chunks anymore (CHUNK_REQUEST_RETRY_MS didn't pass).

// Note that this might be flaky (hopefully not) - so we should use that time injection to make it more stable

mm-near (approved):

chain/chunks/src/lib.rs
mzhangmzz: This line was added accidentally. I will remove it

chain/chunks/src/lib.rs
mzhangmzz: Good question, I added another test for this case, it seems that the code handled it correctly

PR #6193, opened 2022-01-25
chain: rosetta: guarantee order of operations in transactions

mina86:
Because we are using a hash set when identifying accounts which were
changed in a block, the order in which change events are collected is
nondeterministic.  This means that operations in transactions which
cause the changes in nondeterministic order as well.

This is an issue because we assign operation indexes based position in
the list.  This means, that querying the same block multiple times may
result in transactions whose operation indexing changes.

To make sure that doesn’t happen and operation indexes remain
consistent store accounts in a vector in the same order the view
client returns them.

### DIFF
--- a/chain/rosetta-rpc/src/adapters/mod.rs
+++ b/chain/rosetta-rpc/src/adapters/mod.rs
@@ -119,6 +119,10 @@ pub(crate) async fn convert_block_to_transactions(
         .await?
         .unwrap();
 
+    // TODO(mina86): Do we actually need ‘seen’?  I’m kinda confused at this
+    // point how changes are stored in the database and whether view_client can
+    // return duplicate AccountTouched entries.
+    let mut seen = std::collections::HashSet::new();
     let touched_account_ids = state_changes
         .into_iter()
         .filter_map(|x| {
@@ -128,7 +132,12 @@ pub(crate) async fn convert_block_to_transactions(
                 None
             }
         })
-        .collect::<std::collections::HashSet<_>>();
+        .filter(move |account_id| {
+            // TODO(mina86): Convert this to seen.get_or_insert_with(account_id,
+            // Clone::clone) once hash_set_entry stabilises.
+            seen.insert(account_id.clone())
+        })
+        .collect::<Vec<_>>();
 
     let prev_block_id = near_primitives::types::BlockReference::from(
         near_primitives::types::BlockId::Hash(block.header.prev_hash),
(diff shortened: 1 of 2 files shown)


chain/rosetta-rpc/src/adapters/transactions.rs
```diff
@@ -154,35 +154,52 @@ pub(crate) fn convert_block_changes_to_transactions(
...
+                    "queried AccountChanges, but received {:?}.",
+                    unexpected_value
+                )))
+            }
+        };
+        changes_by_account.entry(account_id).or_insert_with(Vec::new).push((change.cause, account));
+    }
+
+    for (account_id, changes) in changes_by_account {
+        let mut previous_account_state = accounts_previous_state.remove(&account_id);
```
mm-near, line 180: there is a behavior change:

in previous code - accounts_previous_state had 'correct' values after the end of the run (as we were doing 'inserts').

Now - this accounts_previous_state will be missing values for every entry that we touched.

if this is working as intended - we should mention it in the function description.

mm-near (approved):

chain/rosetta-rpc/src/adapters/transactions.rs
mina86: `accounts_previous_state` is moved by value into the function and never returned. It doesn’t matter that it’s value changes and it’s not a behaviour that is observable by callers of the function.

mina86:
@bowenwang1996, we synced up on this outside of GitHub and decided this isn’t proper solution. Rather than sorting the operations the code should leverage deterministic order returned by the view client.

mina86:
@frol, PTAL. It indeed gets massively simpler if we assume view client returns defined order.

chain/rosetta-rpc/src/adapters/mod.rs
```diff
@@ -119,6 +119,10 @@ pub(crate) async fn convert_block_to_transactions(
         .await?
         .unwrap();
 
+    // TODO(mina86): Do we actually need ‘seen’?  I’m kinda confused at this
+    // point how changes are stored in the database and whether view_client can
+    // return duplicate AccountTouched entries.
```
frol, line 124: In a single block a single account can be touched in several receipts / transactions and also receive rewards.

Well, to be honest, I feel it will be much easier to introduce a new view-client method to just return full state changes in a given block, so there is no need to fetch those through "give me all the affected accounts" -> "give me full state changes for all the affected accounts, and here is the list", and instead there is just "give me all state changes in the block".

frol (approved):
Looks good to me! Though, consider implementing the view client method I described in the comment

chain/rosetta-rpc/src/adapters/mod.rs
mina86: Yes, that’s exactly what I was thinking as well. I think it should be somewhat trivial to add `StateChangesRequestView::AllAccountsChanges` which would do exactly what we need.

For now I’m going to merge this so that the issue is resolved but I do plan to make the optimisation.

PR #6194, opened 2022-01-25
re: chore: Rewrite `Codec::decode` in a way to avoid unwraps

pmnoxx:
`Codec::decode` function can be rewritten in a functional way, which avoids calls to  `unwrap`.


matklad (approved):

chain/network/src/peer/codec.rs
```diff
@@ -78,19 +78,20 @@ impl Decoder for Codec {
...
-        let len = u32::from_le_bytes(buf[..4].try_into().unwrap()) as usize;
+        let len = u32::from_le_bytes(len_buf) as usize;
         if len > NETWORK_MESSAGE_MAX_SIZE_BYTES {
             // If this point is reached, abusive peer is banned.
             return Ok(Some(Err(ReasonForBan::Abusive)));
         }
 
-        if 4 + len <= buf.len() {
-            let res = Some(Ok(buf[4..4 + len].to_vec()));
+        if let Some(data_buf) = buf.get(4..4 + len) {
```
mm-near, line 93: from what I see below, tests don't cover the test, when buffer is too low (and decoder returns None).

Can you add the test for such case?

mm-near (changes requested):

chain/network/src/peer/codec.rs
mm-near: also - maybe we should test what happens if buf size < 4

pmnoxx: Thanks, that a good point. I added a test.

mm-near (approved):

chain/network/src/peer/codec.rs
```diff
@@ -78,19 +78,20 @@ impl Decoder for Codec {
     type Error = Error;
 
     fn decode(&mut self, buf: &mut BytesMut) -> Result<Option<Self::Item>, Self::Error> {
-        if buf.len() < 4 {
+        let len_buf = match buf.get(..4).and_then(|s| <[u8; 4]>::try_from(s).ok()) {
```
nagisa, line 81: Alternative way to write this could be:

```rust
let len = match buf {
    [a, b, c, d, ..] => u32::from_le_bytes([a, b, c, d]) as usize,
    _ => return Ok(None),
};
```

PR #6195, opened 2022-01-26
fix(indexer): Indexer Framework was unable to index the genesis block.

khorolets:
* `convert_transactions_sir_into_local_receipts` is fetching a previous block that doesn't exist in the case of the genesis block. Also, it was fetching a previous block even if zero transactions were passed.
* example indexer had a duplicate record in subscription for the logs

This PR resolves https://github.com/near/near-indexer-for-explorer/issues/146 and is related to https://github.com/near/near-explorer/issues/849

### DIFF
--- a/chain/indexer/src/streamer/mod.rs
+++ b/chain/indexer/src/streamer/mod.rs
@@ -77,6 +77,7 @@ async fn build_streamer_message(
         as near_primitives::types::NumShards;
 
     let mut shards_outcomes = fetch_outcomes(&client, block.header.hash).await?;
+
     let mut indexer_shards = (0..num_shards)
         .map(|shard_id| IndexerShard { shard_id, chunk: None, receipt_execution_outcomes: vec![] })
         .collect::<Vec<_>>();
--- a/chain/indexer/src/streamer/mod.rs
+++ b/chain/indexer/src/streamer/mod.rs
@@ -77,7 +77,6 @@ async fn build_streamer_message(
         as near_primitives::types::NumShards;
 
     let mut shards_outcomes = fetch_outcomes(&client, block.header.hash).await?;
-
     let mut indexer_shards = (0..num_shards)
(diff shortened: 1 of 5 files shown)


chain/indexer/src/streamer/mod.rs
```diff
@@ -77,6 +77,7 @@ async fn build_streamer_message(
         as near_primitives::types::NumShards;
 
     let mut shards_outcomes = fetch_outcomes(&client, block.header.hash).await?;
+
```
khorolets, line 80:
```suggestion
```

frol (approved):

telezhnaya (approved):

mina86:
> is fetching a previous block that doesn't exist in the case of the genesis block

This is still true though, no? If `txs` is non-empty, the function will still try to fetch the previous block. So what really fixes the issue is that the function is never called on genesis block and non-empty list of transactions?

matklad:
Feels like a test would be worthwhile here.

khorolets:
@mina86 and this is expected behavior. We want to convert sir transactions to local receipts and we need prev_block gas price anyway in order to achieve it. The logic of the conversion is taken from nearcore.

khorolets:
> Feels like a test would be worthwhile here.

what kind of test?

matklad:
I mean in general -- this looks like a change to the observable behavior, so we should have a test which pins down the behavior we want. As to how actually write the test -- I don't know, not familiar with this part of the codebase. It might be the case that the answer is "yeah, it'd be great to have a test here, but we don't have a necessary test infrastructure in place to do this in a nice way, sadly".

khorolets:
> "yeah, it'd be great to have a test here, but we don't have a necessary test infrastructure in place to do this in a nice way, sadly".
@matklad yeah, from my point of view it's the case

PR #6196, opened 2022-01-26
chore: update crates to use single hashbrown version

matklad:
This upgrades:

* borsh
* bn
* lru

which together ensures that we have only one version of hashbrown -- a
sizable crate we don't want to have duplicates of.


mm-near (approved):

pmnoxx (commented):
LGTM

PR #6205, opened 2022-01-27
gitignore: allow target to be a symlink

miraclx:
Allows the `target` build directory to be ignored when symlinked from another partition. With the `target/` spec, git doesn't ignore symlinks because it expects a directory, of which a symlink is not.

Prior Art: https://github.com/rust-lang/cargo/pull/4944

As seen in;

- https://github.com/rust-lang/rust/blob/6abb6385b2cb7249f67b9b3ce7522527767dd907/.gitignore#L47
- https://github.com/rust-lang/cargo/blob/1c034752de0df744fcd7788fcbca158830b8bf85/.gitignore#L1


.gitignore
```diff
@@ -1,6 +1,6 @@
 # Generated by Cargo
 # will have compiled files and executables
-target/
+target
 target_expensive/
```
miraclx, line 4: Leaving `target_expensive` out since, as-per https://github.com/near/nearcore/pull/5971, it's only used on NayDuck. cc: @mina86

mina86: Do we want to do the same with sandbox? It’s used by `make release` and whatever reasons people have to use symlink with `target` they could apply to `sandbox` as well?

mina86 (approved):

matklad:
We might want to do `/target` to match what Cargo does, but I am not off the top of my head sure if we aren't using nested `target`s somewhere...

mina86:
We are using targets in runtime/near-test-contracts/*/target

PR #6208, opened 2022-01-27
feat: Rework deployment fee estimation

jakmeier:
Changes deployment estimation from a sum of deployment and compilation
to just executing the action in a transaction and using least-squares
on these measurements.

Fixes contract code caching during the estimation. Now every
compilation should have a cache-miss for estimations.

Also includes more detailed measurements around deployment and
compilation costs.

Test plan
---------
Manually run the estimator and verify numbers make sense.
This will soon become automated with tracked gas fee changes.


jakmeier:
So here is a suggestion how to change deployment and compilation cost estimations.

After this lands, I would probably do some quick refactoring:
1. Move all helper functions at the bottom of `lib.rs` into a separate utils module
2. Change how the new `ApplyBlock` cost is subtracted: I think this can be hidden inside the helper functions used to measure functions and transactions, so that not every caller-site has to think about block overhead.

matklad:
What's the actual estimation before/after this change, and how that compares to today's cost?

runtime/runtime-params-estimator/src/cost.rs
```diff
@@ -7,12 +7,52 @@ use std::str::FromStr;
...
     ActionCreateAccount,
+    // Deploying a new contract on the blockchain triggers a compilation process
+    // and the compiled code is stored in the DB. The costs charged for such a
+    // contract deployment action is `ActionDeployContractBase` + `N` *
+    // `ActionDeployContractPerByte`, where `N` is the number of bytes in the
+    // WASM code.
+    /// Estimates `action_creation_config.deploy_contract_cost`, which is
+    /// charged once per contract deployment
+    ///
+    /// Estimation: Measure deployment cost of a "smallest" contract.
```
matklad, line 48: Let's rephrase this to highlight that:

* raw wasm is stored in a trie
* compiled code stored in a db

runtime/runtime-params-estimator/src/lib.rs
```diff
@@ -943,6 +1017,33 @@ fn touching_trie_node(ctx: &mut EstimatorContext) -> GasCost {
...
+    let mut testbed = ctx.testbed();
+
+    let blocks = {
+        let n_blocks = testbed.config.warmup_iters_per_block + testbed.config.iter_per_block;
+        let mut blocks = Vec::with_capacity(n_blocks);
+        for _ in 0..n_blocks {
+            let block = Vec::new();
+            blocks.push(block)
+        }
+        blocks
```
matklad, line 1034:
```suggestion
        let mut blocks = vec![vec![]; n_blocks]
```

runtime/runtime-params-estimator/src/lib.rs
```diff
@@ -1169,3 +1270,49 @@ pub fn read_resource(path: &str) -> Vec<u8> {
     std::fs::read(&path)
         .unwrap_or_else(|err| panic!("failed to load test resource: {}, {}", path.display(), err))
 }
+
+fn least_squares_method(xs: &Vec<u64>, ys: &Vec<u64>) -> (Ratio<i128>, Ratio<i128>, Vec<i128>) {
```
matklad, line 1274: Let's move it to a dedicated module, and add some unit-tests.

matklad: ah, you probably want to do that as a follow-up, :+1:

jakmeier:
> What's the actual estimation before/after this change, and how that compares to today's cost?

Sorry, there were some delays in gathering the data. Now I have it (almost) complete and I have quite bit of comments to add on the numbers. :)  

Below I have one table for each metric, comparing *current parameter* vs *estimation before* vs *estimation after* of all action costs.

The big change is obviously the deployment cost. The icount based measurement had the deployment cost at an insane 7.6Tgas while we currently only charge 0.184Tgas  (kind of the reason why I jumped on this). The new version is surprisingly close to the parameter with 0.189Tgas.

Note that the new base estimation has no least-squares method component in it anymore, as we just use the time it takes to compile a smallest contract. But isolated compilation cost is still estimated in `ContractCompileBase` as before, just not used in the formula. And these values are still big, which at first glance makes no sense at all!

However, the reason for the massive change is that the least-squares method acts unstable here since we measure several points with large contracts ( 120kB < x < 350kB ) and then we extrapolate to x = 0, which is far outside the sampled range.

Maybe we should add some smaller contracts to make it more mathematically stable.
Or we can consider completely different strategies, see comments and implementation of `ContractCompileBaseV2` for some alternative line-fitting ideas that give us more sane results. 

### QEMU-based instruction count
| Cost | Parameter value | Before | After | P/B | P/A | B/A |
| --- | --- | --- | --- | --- | --- | --- |
| ActionReceiptCreation                |   108059500000.0 |   305848525000.0 |   285213933750.0 | 0.35 | 0.38 | 1.07 |
| ActionSirReceiptCreation             |                  |   223602098750.0 |   218900551250.0 | 0.00 | 0.00 | 1.02 |
| ActionDeployContractBase             |   184_765_750_000.0 |  7_688_923_375_000.0 |   189_013_373_750.0 | 0.02 | 0.98 | 40.68 |
| ActionDeployContractPerByte          |        6_812_999.0 |      401_988_745.0 |        *still running*          | 0.02 | 0.00 | 0.00 |
| ApplyBlock                           |                  |                  |     9_295_500_000.0 | 0.00 | 0.00 | 0.00 |
| ContractCompileBase                  |                  |                  |  3_957_136_093_750.0 | 0.00 | 0.00 | 0.00 |
| ContractCompileBytes                 |                  |                  |      396_718_750.0 | 0.00 | 0.00 | 0.00 |
| ContractCompileBaseV2                |                  |                  |    39_629_221_512.0 | 0.00 | 0.00 | 0.00 |
| ContractCompileBytesV2               |                  |                  |      350_351_597.0 | 0.00 | 0.00 | 0.00 |


### Wall-clock time measurements

| Cost | Parameter value | Before | After | P/B | P/A | B/A |
| --- | --- | --- | --- | --- | --- | --- |
| ActionReceiptCreation                |   108059500000.0 |   310059240000.0 |   277000600000.0 | 0.35 | 0.39 | 1.12 |
| ActionSirReceiptCreation             |                  |   228268910000.0 |   230981020000.0 | 0.00 | 0.00 | 0.99 |
| ActionDeployContractBase             |   184_765_750_000.0 |  1_331_957_418_177.0 |   486_187_530_000.0 | 0.14 | 0.38 | 2.74 |
| ActionDeployContractPerByte          |        6_812_999.0 |      288_310_494.0 |     1_323_000_000.0 | 0.02 | 0.01 | 0.22 |
| ApplyBlock                           |                  |                  |    20_987_000_000.0 | 0.00 | 0.00 | 0.00 |
| ContractCompileBase                  |                  |                  | 18_771_781_250_000.0 | 0.00 | 0.00 | 0.00 |
| ContractCompileBytes                 |                  |                  |      206_250_000.0 | 0.00 | 0.00 | 0.00 |
| ContractCompileBaseV2                |                  |                  |   737_554_420_193.0 | 0.00 | 0.00 | 0.00 |
| ContractCompileBytesV2               |                  |                  |      235_486_572.0 | 0.00 | 0.00 | 0.00 |


`ActionDeployContractPerByte` is still missing for the icount metric. I will add it when I have it, it has been running for 7 hours now and I am not sure why it hasn't finished... (time measurement of the same cost only took 5min)

### Other action costs

Aside from that, some actions like account creation are also affected with up to 20%. But I wouldn't read too much into that, yet, as this is going to change again. The function `transaction_cost` does not subtract `ApplyBlock` (yet) and this is included in the formula for most actions.

But if you want to have a look, the numbers are hidden below.

<details>

### QEMU-based instruction count
| Cost | Parameter value | Before | After | P/B | P/A | B/A |
| --- | --- | --- | --- | --- | --- | --- |
| ActionFunctionCallBase               |  2319861500000.0 |   103604723750.0 |                  | 22.39 | 0.00 | 0.00 |
| ActionReceiptCreation                |   108059500000.0 |   305848525000.0 |   285213933750.0 | 0.35 | 0.38 | 1.07 |
| ActionSirReceiptCreation             |                  |   223602098750.0 |   218900551250.0 | 0.00 | 0.00 | 1.02 |
| ActionTransfer                       |   115123062500.0 |    62378528750.0 |    78831221250.0 | 1.85 | 1.46 | 0.79 |
| ActionCreateAccount                  |    99607375000.0 |    73811367500.0 |    90674175000.0 | 1.35 | 1.10 | 0.81 |
| ActionDeleteAccount                  |   147489000000.0 |   291264278750.0 |   265108575000.0 | 0.51 | 0.56 | 1.10 |
| ActionAddFullAccessKey               |   101765125000.0 |    61698683750.0 |    74558255000.0 | 1.65 | 1.36 | 0.83 |
| ActionAddFunctionAccessKeyBase       |   102217625000.0 |    62456670000.0 |    73772278750.0 | 1.64 | 1.39 | 0.85 |
| ActionAddFunctionAccessKeyPerByte    |        1925331.0 |       47106380.0 |       46518312.0 | 0.04 | 0.04 | 1.01 |
| ActionDeleteKey                      |    94946625000.0 |    34988302500.0 |    39946443750.0 | 2.71 | 2.38 | 0.88 |
| ActionStake                          |   102217625000.0 |   127501297500.0 |   132520533750.0 | 0.80 | 0.77 | 0.96 |


### Wall-clock time measurements

| Cost | Parameter value | Before | After | P/B | P/A | B/A |
| --- | --- | --- | --- | --- | --- | --- |
| ActionReceiptCreation                |   108059500000.0 |   310059240000.0 |   277000600000.0 | 0.35 | 0.39 | 1.12 |
| ActionSirReceiptCreation             |                  |   228268910000.0 |   230981020000.0 | 0.00 | 0.00 | 0.99 |
| ActionTransfer                       |   115123062500.0 |    61359110000.0 |    82641700000.0 | 1.88 | 1.39 | 0.74 |
| ActionCreateAccount                  |    99607375000.0 |    60258670000.0 |    86613760000.0 | 1.65 | 1.15 | 0.70 |
| ActionDeleteAccount                  |   147489000000.0 |   282373230000.0 |   302465940000.0 | 0.52 | 0.49 | 0.93 |
| ActionAddFullAccessKey               |   101765125000.0 |    57260020000.0 |    60004680000.0 | 1.78 | 1.70 | 0.95 |
| ActionAddFunctionAccessKeyBase       |   102217625000.0 |    59525860000.0 |    61631630000.0 | 1.72 | 1.66 | 0.97 |
| ActionAddFunctionAccessKeyPerByte    |        1925331.0 |       48943904.0 |       49666660.0 | 0.04 | 0.04 | 0.99 |
| ActionDeleteKey                      |    94946625000.0 |    28469770000.0 |    31545670000.0 | 3.33 | 3.01 | 0.90 |
| ActionStake                          |   102217625000.0 |   104301170000.0 |   105689550000.0 | 0.98 | 0.97 | 0.99 |
| ActionFunctionCallBase               |  2319861500000.0 |   464761610000.0 |   442363810000.0 | 4.99 | 5.24 | 1.05 |
| ActionFunctionCallPerByte            |        2235934.0 |         374991.0 |         371772.0 | 5.96 | 6.01 | 1.01 |


</details>

matklad:
>Note that the new base estimation has no least-squares method component in it anymore, as we just use the time it takes to compile a smallest contract.

:+1: great find! Yeah, I believe what you are doing here (deploying the smallest contract) is correct: we don't need to extrapolate for `x=0`, when we can just *run* with `x=0`.

runtime/runtime-params-estimator/src/lib.rs
```diff
@@ -441,77 +468,124 @@ fn action_stake(ctx: &mut EstimatorContext) -> GasCost {
...
+fn action_deploy_contract_per_byte(ctx: &mut EstimatorContext) -> GasCost {
+    let mut xs = vec![];
+    let mut ys = vec![];
+
+    for (contract, pivot_fn) in REAL_CONTRACTS_SAMPLE {
+        let code = read_resource(contract);
+        xs.push(code.len() as u64);
+        let cost = deploy_contract_cost(ctx, code, Some(pivot_fn.as_bytes()));
+        ys.push(cost);
+    }
```
matklad, line 493: There's a bit of a problem here. Previous estimate has this bit of logic in it: 

```
 // We multiply `b` by 5/4 to accommodate for the fact that test contracts are typically 80% code,
    // so in the worst case it could grow to 100% and our costs still give better upper estimation.
    // Safety muliplication with 5/4.
    let safety_numer = 5u64;
    let safety_denom = 4u64;
```

I think we should preserve it. 

Basically, any contract contains some data, and some code. It takes almost no time to compile data, but compiling code is slow. The sample of *real* contracts we use contains about 20% of data, and we should account for that.

runtime/runtime-params-estimator/src/lib.rs
```diff
@@ -441,77 +468,124 @@ fn action_stake(ctx: &mut EstimatorContext) -> GasCost {
...
-    let (compilation_base_cost, _) = compilation_cost_base_per_byte(ctx);
-    let base_cost = deploy_contract_base(ctx);
-    base_cost + compilation_base_cost
-}
-fn action_deploy_contract_per_byte(ctx: &mut EstimatorContext) -> GasCost {
-    let total_cost = {
-        let code = read_resource(if cfg!(feature = "nightly_protocol_features") {
-            "test-contract/res/nightly_large_contract.wasm"
-        } else {
-            "test-contract/res/stable_large_contract.wasm"
```
matklad, line 459: I wonder if `stable_large_contract` becomes unused after this?

matklad:
One thing I am unsure about is that we used to use stable_large_contract to estimate "not compilation"  component of the cost before, and now we use sample contracts, which we primarily tailored for "compilation" component. I think, to answer whether this is OK or not, it' be useful to understand which fraction of the overal deploy cost compilation accounts for.

matklad (approved):
Approving: this feels like an improvement enough already, especially the bit that fixes the base cost. So, I think we can merge. I do want to follow up on two points: 

* add back 5/4 multiplier for compilaton cost, on the account of 20% of sample contracts being data
* checking if it's OK to use sample contracts, rather than the artificial one, for the "non compilation" costs (access to DB basically)

jakmeier:
> * add back 5/4 multiplier for compilaton cost, on the account of 20% of sample contracts being data
> 
> * checking if it's OK to use sample contracts, rather than the artificial one, for the "non compilation" costs (access to DB basically)

These are two good points and I completely agree that we should include this in the new estimation.
I think the tricky bit here is that now the deployment action is measured holistically. Applying the multiplier to the full deployment cost isn't quite the same as applying it to just the compilation part. But I will come up with something that  think makes sense in a follow-up PR.

matklad:
>Applying the multiplier to the full deployment cost isn't quite the same as applying it to just the compilation part.

One hypothesis I have here is that deployment is *mostly* compilation, so it might actually be pretty similar in effect. Hence the question of which part of deploy is compilation, and which one is deploy per se.

PR #6211, opened 2022-01-28
feat(sandbox): fast forward timestamp and epoch height

ChaoticTempest:
Realized was missing some extra stuff like updating timestamps/epoch-height after fast forwarding special thanks to @TrevorJTClarke. This is a corrollary PR to https://github.com/near/nearcore/pull/6158 to address and add timestamps and epoch-height updates.

### How it works:
The timestamps are updated via the delta block height converted to a fast-forward timestamp delta; which then gets added to a `Clock::utc()` to get the fast-forward timestamp. Then the next block with use the previous timestamp to go from there. Epoch height update uses current block height / epoch_length to get the height instead of the usual increment by one. This is handled this way since not much is passed into it besides the `BlockInfo` and `EpochManager` is separate from everything else such as chain/client/client-actor.

Also, unfortunately, there’s no easy way that I can see of for testing the epoch height update since that is apart of `near-epoch-manager` crate and is created separately from everything else; so can't really test it with block production. Let me know if there is something I'm missing out on


chain/client/src/client.rs
```diff
@@ -962,6 +988,18 @@ impl Client {
...
+    /// Gets the advanced timestamp delta in nanoseconds for sandbox once it has been fast-forwarded
+    #[cfg(feature = "sandbox")]
+    pub fn sandbox_delta_time(
+        &self,
+        delta_height: near_primitives::types::BlockHeightDelta,
+    ) -> u64 {
+        let avg_block_prod_time = (self.config.min_block_production_delay.as_nanos()
+            + self.config.max_block_production_delay.as_nanos())
+            / 2;
+        (delta_height as u128 * avg_block_prod_time) as u64
```
ailisp, line 1000: Theoratically it's better to check overflow and panic. but I guess it's fine as u64 can hold sufficiently big number of nano seconds.

integration-tests/src/tests/client/sandbox.rs
```diff
@@ -122,36 +122,68 @@ fn test_patch_account() {
...
+
+                        // Check if final block timestamp is within fast forwarded time min and max:
+                        assert!(
+                            (min_forwarded_time..max_forwarded_time).contains(&timestamp),
+                            "Forwarded timestamp {} is not within expected range ({},{})",
+                            timestamp,
+                            min_forwarded_time,
+                            max_forwarded_time
+                        );
+
```
ailisp, line 179: also check epoch height is bumped here?

ailisp (approved):
lgtm! I didn't expect timestamp could also be update in a straightforward way. Historically we had a PR to "time travel": https://github.com/near/nearcore/pull/3661 which was complicated. I think your did both fast forward block height and time travel!

integration-tests/src/tests/client/sandbox.rs
ChaoticTempest: Epoch height isn't available unfortunately within integration-tests since that is provided by the `EpochManager` crate. Not sure how I could tie that in since epoch manager isn't being tested with that the rest of nearcore here. And it seems like epoch manager crate was meant to be separate anyways since I see docs saying it could potentially be turned into a contract

chain/client/src/client.rs
ChaoticTempest: Converted this to a `try_into().expect()` to panic if necessary

nikurt (approved):

integration-tests/src/tests/client/sandbox.rs
ChaoticTempest: Actually, was able to test epoch height in `fast_forward.py` instead since those tests have everything end-to-end

pytest/tests/sandbox/fast_forward.py
```diff
@@ -5,23 +5,62 @@
 import sys, time
 import pathlib
 
+from datetime import datetime as dt, timedelta
```
mina86, line 8: Please don’t rename to `dt`. I consider importing symbols from modules bad in and of itself, renaming things makes it even worse.

pytest/tests/sandbox/fast_forward.py
```diff
@@ -5,23 +5,62 @@
...
+
+min_forwarded_secs = timedelta(0, BLOCKS_TO_FASTFORWARD * MIN_BLOCK_PROD_TIME)
+max_forwarded_secs = timedelta(0, BLOCKS_TO_FASTFORWARD * MAX_BLOCK_PROD_TIME)
+min_forwarded_time = earliest + min_forwarded_secs
+max_forwarded_time = earliest + max_forwarded_secs
+
+assert min_forwarded_time < latest < max_forwarded_time
+
+# wait a little longer for epoch height to be updated:
+time.sleep(7)
```
mina86, line 62: Why 7? If we’re waiting for a new epoch, shouldn’t we use `utils.wait_for_blocks` or something? Besides, why do we even need to wait for any new blocks?

pytest/tests/sandbox/fast_forward.py
```diff
@@ -5,23 +5,62 @@
...
+nodes = start_cluster(1, 0, 1, CONFIG, [["epoch_length", EPOCH_LENGTH]], {})
 
 # request to fast forward
 nodes[0].json_rpc('sandbox_fast_forward', {
-    "delta_height": BLOCKS_TO_FASTFORARD,
+    "delta_height": BLOCKS_TO_FASTFORWARD,
 })
 
 # wait a little for it to fast forward
 time.sleep(3)
```
mina86, line 42: Actually, this made me realise this should be `utils.wait_for_blocks(nodes[0], target=BLOCKS_TO_FAST_FORWARD, timeout=3)`.

chain/client/src/client.rs
```diff
@@ -516,6 +523,24 @@ impl Client {
...
+                Block::BlockV2(ref mut block) => &mut block.header,
+            };
+            let mut inner_lite = match header {
+                BlockHeader::BlockHeaderV1(ref mut header) => &mut header.inner_lite,
+                BlockHeader::BlockHeaderV2(ref mut header) => &mut header.inner_lite,
+                BlockHeader::BlockHeaderV3(ref mut header) => &mut header.inner_lite,
+            };
+
+            // Add in the accrued delta timestamp:
+            inner_lite.timestamp = to_timestamp(Clock::utc()) + self.sandbox_delta_time();
```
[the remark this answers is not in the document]
ChaoticTempest: Originally was this, but the next timestamp for the next block produced is only incremented by 1 (happens in `Block::produce`), which isn't the intended behavior, so we need to set it every single time instead

pytest/tests/sandbox/fast_forward.py
ChaoticTempest: so `wait_for_blocks` doesn't work as we'd like is the thing. Fast forwarding means jumping a specific block height relative to the current block height. We won't know what the current block height is at the time of fast forwarding, and this function checks for a very specific block to be produced

pytest/tests/sandbox/fast_forward.py
ChaoticTempest: hmm, I can switch it over to the more verbose form, otherwise `datetime import datetime` will be in conflict with `import datetime`

pytest/tests/sandbox/fast_forward.py
ChaoticTempest: good point, this was because it only updated the next_next_epoch height. Thanks for pointing this out. Made the relevant changes to update the next_epoch height as well

pytest/tests/sandbox/fast_forward.py
mina86: Please do that.

pytest/tests/sandbox/fast_forward.py
mina86: No, the function checks for a block at least equal to the target to be
produced:

        if latest.height >= target:
            return latest

so you can call it with a target equal the number of blocks being
skipped, and it’ll return as soon as the node goes past that.

chain/chain/src/chain.rs
```diff
@@ -4193,8 +4193,11 @@ impl<'a> ChainUpdate<'a> {
         provenance: &Provenance,
         on_challenge: &mut dyn FnMut(ChallengeBody),
     ) -> Result<(), Error> {
-        // Refuse blocks from the too distant future.
-        if header.timestamp() > Clock::utc() + Duration::seconds(ACCEPTABLE_TIME_DIFFERENCE) {
+        // Refuse blocks from the too distant future. Only exception to this is while we're within
+        // sandbox, we're allowed to jump to the future via fast forwarding.
+        #[cfg(not(feature = "sandbox"))]
+        if header.timestamp() > Clock::utc() + chrono::Duration::seconds(ACCEPTABLE_TIME_DIFFERENCE)
```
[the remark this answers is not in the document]
mina86: `!cfg!(feature = "sandox")` rather than `cfg!(feature = "sandbox")`.

mina86 (approved):

chain/chain/src/chain.rs
```diff
@@ -4193,8 +4193,11 @@ impl<'a> ChainUpdate<'a> {
         provenance: &Provenance,
         on_challenge: &mut dyn FnMut(ChallengeBody),
     ) -> Result<(), Error> {
-        // Refuse blocks from the too distant future.
-        if header.timestamp() > Clock::utc() + Duration::seconds(ACCEPTABLE_TIME_DIFFERENCE) {
+        // Refuse blocks from the too distant future. Only exception to this is while we're within
+        // sandbox, we're allowed to jump to the future via fast forwarding.
+        if cfg!(feature = "sandbox")
```
bowenwang1996, line 4198: shouldn't this be `!cfg!(feature = "sandbox")`?

chain/chain/src/chain.rs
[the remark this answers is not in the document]
ChaoticTempest: somehow missed that, thanks for pointing it out!

pytest/tests/sandbox/fast_forward.py
ChaoticTempest: hmm weird, it threw an `AssertionError` at me before with not being able to get to `BLOCKS_TO_FAST_FORWARD` even though it went past that. Might've been out of date at that time

chain/chain/src/chain.rs
ChaoticTempest: Yup, thanks for pointing it out!

pytest/tests/sandbox/fast_forward.py
```diff
@@ -40,7 +40,8 @@
 })
 
 # wait a little for it to fast forward
-time.sleep(3)
+# NOTE: epoch takes a couple of blocks before it gets updated, so +10 for this
+utils.wait_for_blocks(nodes[0], target=BLOCKS_TO_FASTFORWARD + 10)
 
 # Assert at the end that the node is past the amounts of blocks we specified
 assert nodes[0].get_latest_block().height > BLOCKS_TO_FASTFORWARD
```
mina86, line 47: This is redundant.

khorolets (approved):

chain/client/src/client.rs
```diff
@@ -960,6 +985,18 @@ impl Client {
         Ok(())
     }
 
+    /// Gets the advanced timestamp delta in nanoseconds for sandbox once it has been fast-forwarded
+    #[cfg(feature = "sandbox")]
+    pub fn sandbox_delta_time(&self) -> u64 {
+        let avg_block_prod_time = (self.config.min_block_production_delay.as_nanos()
+            + self.config.max_block_production_delay.as_nanos())
+            / 2;
```
bowenwang1996, line 993: No this is not representative of average block time. What do you need it for?

ChaoticTempest: I'm using it for estimating how much of a jump in the timestamp needs to be added to the current time, for the new timestamp after fast-forwarding/time-travelling. It doesn't have to be entirely accurate. Just need to be within a threshold. Is `min_block_production_delay` and `max_block_production_delay` not the right config to use then? It seemed so far to me to have controlled how fast block production while I was testing it when I was playing with it

bowenwang1996: > I'm using it for estimating how much of a jump in the timestamp needs to be added to the current time, for the new timestamp after fast-forwarding/time-travelling

What do you use this estimation for?

chain/epoch_manager/src/lib.rs
```diff
@@ -410,6 +411,12 @@ impl EpochManager {
             )
         };
         let next_next_epoch_config = self.config.for_protocol_version(next_version);
+        #[cfg(feature = "sandbox")]
+        {
+            *next_epoch_info.epoch_height_mut() =
+                block_info.height() / next_next_epoch_config.epoch_length;
+        }
```
bowenwang1996, line 418: Could you describe why this change is necessary? In general the relationship described here does not always hold.

chain/epoch_manager/src/lib.rs
```diff
@@ -445,6 +452,12 @@ impl EpochManager {
         // This epoch info is computed for the epoch after next (T+2),
         // where epoch_id of it is the hash of last block in this epoch (T).
         self.save_epoch_info(store_update, &next_next_epoch_id, next_next_epoch_info)?;
+
+        // Store next_epoch info again since sandbox changes it:
+        if cfg!(feature = "sandbox") {
+            self.save_epoch_info(store_update, &next_epoch_id, next_epoch_info)?;
+        }
```
bowenwang1996, line 459: This logic looks quite dubious to me. Why do you need to change epoch info retroactively? Also we should have unit test for this.

bowenwang1996 (changes requested):
You can get access to epoch height from view client through `GetValidatorInfo`. I think the PR could benefit from describing the use case more clearly and having more tests

chain/client/src/client.rs
ChaoticTempest: after we fast-forward the sandbox by a specific delta height, we need the new blocks produced to reflect the timestamps at that new block height as well. So this estimation is done specifically for that.

chain/epoch_manager/src/lib.rs
ChaoticTempest: this updates the epoch height to after fast-forwarding the block height. This was the only way I could feasibly update it since `epoch_manager` is separate from everything else. Also, what cases does this not hold for? This is just for sandbox so it might not be necessary to account for them

chain/epoch_manager/src/lib.rs
ChaoticTempest: So, there's a clone that happens when we grab `next_epoch_info`. This is why we need to update it again. I'll try to add a unit test for this

chain/client/src/client.rs
bowenwang1996: > we need the new blocks produced to reflect the timestamps at that new block height as well

I am confused by this statement. Which of the following do you mean here:
1. Sandbox assumes a certain relationship between block height and block timestamp and it is crucial for the functionality of Sandbox that such relationship holds.
2. Block height is used as a proxy for block timestamp so that you only need to update one value to make fast forwarding work

chain/epoch_manager/src/lib.rs
bowenwang1996: > Also, what cases does this not hold for?

While it is guaranteed that the block height of the end of an epoch is at least `epoch_length` + the height of the block in the beginning of the epoch, there is no guarantee on the upper bound. In fact, it is possible for a block height in one epoch to be arbitrarily large. Could you let me know the reason for modifying epoch height? What guarantees do you need to have here?

chain/client/src/client.rs
ChaoticTempest: More so 1, but 2 also holds as a side-effect of how I implemented it. The functionality I'm trying to simulate here is as if the blocks were produced as normal. So let's say we fast-forward 10,000 blocks. That translates to producing 10,000 blocks where the latest block timestamp should reflect an almost equivalent amount of time has passed after 10,000 blocks were produced. So roughly the new timestamp should be `time::now() + 10,000 * avg_block_prod_time`.

Let me know if that doesn't make sense since this feature is more so based on what the sdk-sim had with producing blocks to simulate fast-forwarding/time-traveling

chain/epoch_manager/src/lib.rs
ChaoticTempest: > there is no guarantee on the upper bound. In fact, it is possible for a block height in one epoch to be arbitrarily large

This should be fine since when we receive a fast-forward request, it happens in the middle of an epoch and the newly forward block height would trigger updating the epoch-height at an arbitrary block height.

The guarantees needed for epoch height updates would be what you mentioned at the end of an epoch:
- block-height lower bound: `epoch_length + height of block in beginning of the epoch`
- block-height upper bound: unbounded

> Could you let me know the reason for modifying epoch height?

This is needed because what I mentioned earlier in the other post about producing blocks and simulating those. So after producing X amount of blocks (fast-forwarding), the user of a contract would intuitively assume that the epoch height also got updated to reflect X blocks being produced.

Also, correct me on any of these things, but that's my thinking at least

chain/client/src/client.rs
bowenwang1996: I see. What would break in sandbox if the time isn't fast-forwarded? In other words, is this a nice-to-have feature or does it serve some crucial purpose?

chain/epoch_manager/src/lib.rs
bowenwang1996: > So after producing X amount of blocks (fast-forwarding), the user of a contract would intuitively assume that the epoch height also got updated to reflect X blocks being produced.

Actually this should be done automatically as long as block height is more than `epoch_length` + height of block in the beginning of the epoch. I don't think we need custom logic here.

chain/client/src/client.rs
ChaoticTempest: For instance, the cron.cat needs it to test out scheduling in their contracts. So I would vote for crucial purpose. @TrevorTJClarke can talk more about it from his side and how he uses it

chain/epoch_manager/src/lib.rs
ChaoticTempest: hmm interesting. But wouldn't that mean the epoch-height is only incremented by 1? There was only logic I saw that incremented the epoch-height by 1 when the next epoch was in view: https://github.com/near/nearcore/blob/31493321477fac77b466305c019117f61c930b63/chain/epoch_manager/src/validator_selection.rs#L194

bowenwang1996: Yes. Do you require it to be incremented by more than 1? If so, what is the use case?

ChaoticTempest: not so much that it needs to be incremented more than 1 right there, but the custom logic is what I'm saying allows for us to skip to the designated epoch height after fast forwarding.

> Actually this should be done automatically as long as block height is more than epoch_length + height of block in the beginning of the epoch. I don't think we need custom logic here.

But now I'm not sure what you're referring to then with the above statement. Is there some other logic that would allow us to jump to the new epoch height after fast forwarding? I didn't see anything and epoch height remained the same as if fast-forwarding the block height never happened

bowenwang1996: I think you just need to produce three more blocks to end the epoch, which could be implemented as part of the sandbox

ChaoticTempest: It shouldn't need to produce more blocks to end the epoch though, but regardless I tested it without the custom logic and it's not producing what's expected after fast forwarding. I'm getting epoch height of 3, when it should be 503, after waiting for roughly 50-60 blocks to be produced. You can see for yourself with the following diff and running `python3 pytests/tests/sandbox/fast_forward.py`:
```diff
diff --git a/chain/epoch_manager/src/lib.rs b/chain/epoch_manager/src/lib.rs
index b54af9e47..b1406c930 100644
--- a/chain/epoch_manager/src/lib.rs
+++ b/chain/epoch_manager/src/lib.rs
@@ -411,18 +411,6 @@ impl EpochManager {
             )
         };
         let next_next_epoch_config = self.config.for_protocol_version(next_version);
-        #[cfg(feature = "sandbox")]
-        {
-            // explaination on calculation:
-            // (block_height / length) gives us epoch_height, but we index from 1, so +1 at the end.
-            // Note, this is a retroactive update for sandbox since we can receive fast-forward
-            // requests in the middle of an epoch. block_height is different based on when it gets
-            // called. When fast-forwarded, we need to retroactively adjust the epoch height, but
-            // with a normal `finalize_epoch`, it's the start of the epoch's next block height, so
-            // (height - 1) is needed after adding +1 at the end.
-            *next_epoch_info.epoch_height_mut() =
-                ((block_info.height() - 1) / next_next_epoch_config.epoch_length) + 1;
-        }
 
         let next_next_epoch_info = match proposals_to_epoch_info(
             next_next_epoch_config,
@@ -460,12 +448,6 @@ impl EpochManager {
         // where epoch_id of it is the hash of last block in this epoch (T).
         self.save_epoch_info(store_update, &next_next_epoch_id, next_next_epoch_info)?;
 
-        // Store next_epoch info again since sandbox changes it. Need to retroactively update
-        // since a fast-forward request can happen in the middle of an epoch.
-        if cfg!(feature = "sandbox") {
-            self.save_epoch_info(store_update, &next_epoch_id, next_epoch_info)?;
-        }
-
         Ok(())
     }
 
diff --git a/pytest/tests/sandbox/fast_forward.py b/pytest/tests/sandbox/fast_forward.py
index 569981ab3..87287fd40 100644
--- a/pytest/tests/sandbox/fast_forward.py
+++ b/pytest/tests/sandbox/fast_forward.py
@@ -61,4 +61,9 @@ assert min_forwarded_time < latest < max_forwarded_time
 
 # Check to see that the epoch height has been updated correctly:
 epoch_height = nodes[0].get_validators()['result']['epoch_height']
-assert epoch_height > BLOCKS_TO_FASTFORWARD / EPOCH_LENGTH
+# assert epoch_height > BLOCKS_TO_FASTFORWARD / EPOCH_LENGTH
+print(f"-> epoch_height {epoch_height}")
+
+utils.wait_for_blocks(nodes[0], target=BLOCKS_TO_FASTFORWARD + 50)
+epoch_height = nodes[0].get_validators()['result']['epoch_height']
+print(f"-> epoch_height {epoch_height}")
```
And after including the custom logic again, the python tests show the correct epoch height of 503, so it's definitely needed here

bowenwang1996 (approved):
I'll approve to unblock, but please work with @mm-near to figure out whether there is a way to avoid adding conditional statements in epoch manager

chain/epoch_manager/src/lib.rs
```diff
@@ -445,6 +459,13 @@ impl EpochManager {
         // This epoch info is computed for the epoch after next (T+2),
         // where epoch_id of it is the hash of last block in this epoch (T).
         self.save_epoch_info(store_update, &next_next_epoch_id, next_next_epoch_info)?;
+
+        // Store next_epoch info again since sandbox changes it. Need to retroactively update
+        // since a fast-forward request can happen in the middle of an epoch.
+        if cfg!(feature = "sandbox") {
+            self.save_epoch_info(store_update, &next_epoch_id, next_epoch_info)?;
```
mm-near, line 466: maybe move it next to 424

chain/epoch_manager/src/lib.rs
```diff
@@ -410,6 +411,19 @@ impl EpochManager {
...
         let next_next_epoch_config = self.config.for_protocol_version(next_version);
+        #[cfg(feature = "sandbox")]
+        {
+            // explaination on calculation:
+            // (block_height / length) gives us epoch_height, but we index from 1, so +1 at the end.
+            // Note, this is a retroactive update for sandbox since we can receive fast-forward
+            // requests in the middle of an epoch. block_height is different based on when it gets
+            // called. When fast-forwarded, we need to retroactively adjust the epoch height, but
+            // with a normal `finalize_epoch`, it's the start of the epoch's next block height, so
+            // (height - 1) is needed after adding +1 at the end.
```
mm-near, line 422: do you have any doc describing how the chain looks like after the 'fast forward'?

Because I see that you change some things, but don't change the others 
(for example: here you'd update the epoch_height -- but this might also mean that we'll have 'holes' in the epoch history).

The approach, I'd have preferred is that we do move timestamp forward (fine) - but for blocks, what we do - is we generate 2-3 blocks per epoch (to get finality and generate the next list of validators) - and jump to another epoch.

This would mean that such chain is still a 'correct' chain.

mm-near (changes requested):

chain/chain/src/chain.rs
```diff
@@ -4334,8 +4334,11 @@ impl<'a> ChainUpdate<'a> {
         provenance: &Provenance,
         on_challenge: &mut dyn FnMut(ChallengeBody),
     ) -> Result<(), Error> {
-        // Refuse blocks from the too distant future.
-        if header.timestamp() > Clock::utc() + Duration::seconds(ACCEPTABLE_TIME_DIFFERENCE) {
+        // Refuse blocks from the too distant future. Only exception to this is while we're within
+        // sandbox, we're allowed to jump to the future via fast forwarding.
+        if !cfg!(feature = "sandbox")
+            && header.timestamp() > Clock::utc() + Duration::seconds(ACCEPTABLE_TIME_DIFFERENCE)
```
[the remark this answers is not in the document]
ChaoticTempest: Think that would be overkill for this. Mocking out Clock is pretty complicated according to this PR https://github.com/near/nearcore/pull/3661/files which attempted it already before. It's easier just to have block production timestamps change for the contract testing side. There was only one place that checked for block timestamps and that's already being ignored under sandbox. Maybe something to revisit in the future if we want to implement time-traveling overall, but this should be sufficient for just smart contracts to use.

chain/epoch_manager/src/lib.rs
ChaoticTempest: ownership rules are funky in this function which is why I put it down there ¯\_(ツ)_/¯

chain/epoch_manager/src/lib.rs
ChaoticTempest: how the chain looks like after the fast forward (after like 10k blocks) is the following:
```
block_timestamp: T + 10000 * avg_block_prod_time
block_height: H + 10000
epoch_height: E + 10000 / epoch_length
```
where the variables before fast forwarding are the following:
```
block_timestamp: T
block_height: H
epoch_height: E
```

I'm not sure I follow with the holes. Do you mean not being able to refer to a previous epoch? The history would just end up looking like the following with epoch length of 5:
```
E1 ........... E2 ..... E11 .............
B1 B2 B3 B4 B5 B6 B7 B8 B50 B51 B52 ...
                        ^^^
                        fast forward request w/ 42 blocks
```
where
```
E = epoch height
B = block height
```

We'd still be able to refer to the previous epoch, since this is all being referenced via the epoch-id, which hasn't been changed. So I'm not sure what a "correct" chain entails wrt epochs/blocks. What would one look like, or any docs I can look at?

ChaoticTempest:
@bowenwang1996 @mm-near updated this PR with the new epoch boundary logic, where we jump to each epoch boundary, produce some blocks, then update the epoch height, so no custom sandbox related epoch_manager logic is needed. Let me know if I'm missing anything

mm-near (changes requested):

chain/client/src/client.rs
```diff
@@ -516,17 +522,87 @@ impl Client {
...
         metrics::BLOCK_PRODUCED_TOTAL.inc();
 
         Ok(Some(block))
     }
 
+    /// Set the fast-forwarded block timestamp to the supplied `block`. This is
+    /// current time + total amount of time forwarded from fast-forwarding. This function will
+    /// update the block hash since the timestamp was originally used to compute it.
+    #[cfg(feature = "sandbox")]
+    fn sandbox_alter_block_for_fast_forwarded(
```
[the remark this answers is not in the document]
ChaoticTempest: that works as well and would make things simpler, but didn't want to add another variable to that already long list of arguments. But if think that's best, I'll just add it there

chain/client/src/client_actor.rs
```diff
@@ -384,6 +384,13 @@ impl ClientActor {
                         self.fastforward_delta = Some(delta_height);
```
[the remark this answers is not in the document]
ChaoticTempest: oh that's true, yeah let's return an error here if a fast-forward request is already happening. We can add onto this later if the need arises for consecutive fast forwards

chain/chain/src/chain.rs
```diff
@@ -106,6 +106,16 @@ pub const NUM_EPOCHS_TO_KEEP_STORE_DATA: u64 = 5;
 /// Maximum number of height to go through at each step when cleaning forks during garbage collection.
 const GC_FORK_CLEAN_STEP: u64 = 1000;
 
+/// Grab the acceptable time difference. This may differ between sandbox and regular networks
+const fn acceptable_time_difference() -> i64 {
```
mm-near, line 110: move it to line 94 - and do a simpler thing:

#[cfg(feature = "sandbox")]
ACCEPTABLE_TIME_DIFFERENCE = 60 * 60 * 24 * 365 * 10000;
#[!cfg(not(feature = "sandbox"))]
ACCEPTABLE_TIME_DIFFERENCE = 12*10;

chain/client/src/client_actor.rs
```diff
@@ -880,6 +893,79 @@ impl ClientActor {
         }
     }
 
+    #[cfg(feature = "sandbox")]
+    fn sandbox_process_fast_forward(
```
mm-near, line 897: comment what this method does

chain/client/src/client.rs
```diff
@@ -445,6 +451,9 @@ impl Client {
             prev_next_bp_hash
         };
 
+        #[cfg(feature = "sandbox")]
+        let delta_time = self.sandbox_delta_time();
```
mm-near, line 455: move it closer to 511

core/primitives/src/block.rs
```diff
@@ -243,8 +244,16 @@ impl Block {
         );
 
         let new_total_supply = prev.total_supply() + minted_amount.unwrap_or(0) - balance_burnt;
-        let now = to_timestamp(Clock::utc());
-        let time = if now <= prev.raw_timestamp() { prev.raw_timestamp() + 1 } else { now };
+        let time = if let Some(timestamp) = timestamp_override {
+            timestamp
+        } else {
+            let now = to_timestamp(Clock::utc());
+            if now <= prev.raw_timestamp() {
```
mm-near, line 251: maybe we should still apply this condition, even if someone has provided the 'override timestamp'?

I'd simply replace the old line 246 with 
let now = timestamp_override.unwrap_or_else(to_timestamp(Clock::utc))

mm-near (approved):

chain/chain/src/chain.rs
ChaoticTempest: a bit opinionated but had the function instead since we can have it all in one place. But it's not a big deal

chain/client/src/client.rs
ChaoticTempest: Had it there because of borrow rules here, but I can move the declaration of `let (at, timestamp) = ...` close to this declaration instead

core/primitives/src/block.rs
ChaoticTempest: not a bad idea to still apply the condition, since fast forward should never be going backwards anyways

chain/client/src/client.rs
```diff
@@ -514,13 +536,11 @@ impl Client {
...
             block_merkle_root,
+            timestamp_override,
         );
 
         // Update latest known even before returning block out, to prevent race conditions.
-        self.chain.mut_store().save_latest_known(LatestKnown {
-            height: next_height,
-            seen: to_timestamp(Clock::utc()),
-        })?;
+        self.chain.mut_store().save_latest_known(LatestKnown { height: next_height, seen: at })?;
```
matklad, line 543: :thinking: I don't like that `at` here at `LatestKnown` and the `timestamp` of the block are different, but this is pre-existing behavior. So, no need to change anything in this PR, but I woneder if the following change makes sense as a follow-up:

```diff
diff --git a/chain/client/src/client.rs b/chain/client/src/client.rs
index c6091c459..f5982fa6b 100644
--- a/chain/client/src/client.rs
+++ b/chain/client/src/client.rs
@@ -451,21 +451,10 @@ impl Client {
             prev_next_bp_hash
         };
 
-        #[cfg(feature = "sandbox")]
-        let delta_time = self.sandbox_delta_time();
-
-        let at = to_timestamp(Clock::utc());
+        #[cfg(not(feature = "sandbox"))]
         let timestamp_override = Option::<u64>::None;
-
-        // Sandbox can override the timestamp normally produced from a Block. This new timestamp
-        // is from when a fast_forward request that has been processed.
         #[cfg(feature = "sandbox")]
-        let (at, timestamp_override) = if delta_time > 0 {
-            let at = at + delta_time;
-            (at, Some(at))
-        } else {
-            (at, timestamp_override)
-        };
+        let timestamp_override = to_timetamp(Clock::now()) + self.sandbox_delta_time();
 
         // Get block extra from previous block.
         let mut block_merkle_tree =
@@ -538,9 +527,11 @@ impl Client {
             block_merkle_root,
             timestamp_override,
         );
+        let latest_known =
+            LatestKnown { height: block.header().height(), seen: block.header().timestamp() };
 
         // Update latest known even before returning block out, to prevent race conditions.
-        self.chain.mut_store().save_latest_known(LatestKnown { height: next_height, seen: at })?;
+        self.chain.mut_store().save_latest_known(latest_known)?;
 
         metrics::BLOCK_PRODUCED_TOTAL.inc();
 
```

chain/client/src/test_utils.rs
```diff
@@ -66,6 +66,11 @@ use near_primitives::utils::MaybeValidated;
 
 pub type PeerManagerMock = Mocker<PeerManagerActor>;
 
+/// min block production time in milliseconds
+pub const MIN_BLOCK_PROD_TIME: u64 = 100;
```
matklad, line 70: Consider making these durations, then the comment becomes unnnecessary.

matklad (approved):

chain/network-primitives/src/types.rs
```diff
@@ -278,12 +278,15 @@ pub enum NetworkSandboxMessage {
...
     SandboxPatchStateStatus,
     SandboxFastForward(near_primitives::types::BlockHeightDelta),
+    SandboxFastForwardStatus,
 }
 
 #[cfg(feature = "sandbox")]
 #[derive(Eq, PartialEq, Debug)]
 pub enum SandboxResponse {
     SandboxPatchStateFinished(bool),
+    SandboxFastForwardFinished(bool),
```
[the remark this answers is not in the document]
ChaoticTempest: I went with what was already there for patching state. It's only being used internally right now, so not too important to change IMO

chain/client/src/client.rs
ChaoticTempest: I ended up changing it to what you suggested, since I had to change it already for swapping the return type of `sandbox_delta_time` to `Duration`. And pretty much same logic at the end of the day too, so it should work the same

chain/client/src/client_actor.rs
```diff
@@ -890,6 +903,86 @@ impl ClientActor {
...
+    /// Process the sandbox fast forward request. If the change in block height is past an epoch,
+    /// we fast forward to just right before the epoch, produce some blocks to get past and into
+    /// a new epoch, then we continue on with the residual amount to fast forward.
+    #[cfg(feature = "sandbox")]
+    fn sandbox_process_fast_forward(
+        &mut self,
+        block_height: BlockHeight,
+    ) -> Result<Option<near_chain::types::LatestKnown>, Error> {
+        let mut delta_height = match self.fastforward_delta.take() {
+            Some(delta_height) if delta_height > 0 => delta_height,
```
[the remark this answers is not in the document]
ChaoticTempest: good point, changed it to be a non-option type

chain/client/src/client.rs
```diff
@@ -67,6 +67,10 @@ pub struct Client {
     #[cfg(feature = "test_features")]
     pub adv_produce_blocks_only_valid: bool,
 
+    /// Fast Forward accrued delta height used to calculate fast forwarded timestamps for each block.
+    #[cfg(feature = "sandbox")]
+    pub(crate) accrued_fastforward_delta: near_primitives::types::BlockHeightDelta,
```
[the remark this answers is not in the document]
ChaoticTempest: yeah, this was definitely a hack to circumvent that. Maybe we can revisit this in the future, but for now, this was pretty simple to do just for contract testing side of things

PR #6212, opened 2022-01-28
fix: replace pop with get in `TrieCachingStorage`

Longarithm:
While working with https://github.com/near/nearcore/pull/5871, because of a typo (?) we replaced some `cache_get` with `pop`, so we remove each entry from `TrieCachingStorage` on read: https://github.com/near/nearcore/pull/5871/files#diff-e073548a40d97af14f75cf143fab41a1cffe61d159e0b9a6297daeab0b2a5d45L171

Here we restore the original behaviour.

## Test plan

Existing tests.

### DIFF
--- a/core/store/src/trie/trie_storage.rs
+++ b/core/store/src/trie/trie_storage.rs
@@ -167,7 +167,7 @@ impl TrieCachingStorage {
 impl TrieStorage for TrieCachingStorage {
     fn retrieve_raw_bytes(&self, hash: &CryptoHash) -> Result<Vec<u8>, StorageError> {
         let mut guard = self.cache.0.lock().expect(POISONED_LOCK_ERR);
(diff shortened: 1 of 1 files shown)


matklad:
Oh wow, great catch. 

>Existing tests.

I ... don't think this does the bug justice. At the same time, I don't think writing a unit-test specifically for this would be an improvement. This should have been caught by some kind of hight-level integrated benchmark.

matklad (approved):

matklad:
cc https://near.zulipchat.com/#narrow/stream/295558-nearinc.2Fprotocol-core/topic/Stresstesting.2Fbenchmarking.20client.20.26.20runtime.20.28idea.29

mm-near (approved):

mzhangmzz (approved):

pmnoxx (approved):
LGTM

Issue #4833, opened 2021-09-17
Upgrade the docker image we use for the estimation


matklad:
At the moment, we use rustup to install the appropriate Rust version in the estimator. That made sense while we were using Rust-nightly, but today, we should just use the official Rust image. IN theory, that should be as easy as applying the following diff:

```diff
diff --git a/runtime/runtime-params-estimator/emu-cost/Dockerfile b/runtime/runtime-params-estimator/emu-cost/Dockerfile
index ec9a4fee0..04baa5038 100644
--- a/runtime/runtime-params-estimator/emu-cost/Dockerfile
+++ b/runtime/runtime-params-estimator/emu-cost/Dockerfile
@@ -1,15 +1,9 @@
 # our local base image
 # FROM ubuntu
-FROM rust
+FROM rust:1.55
 
 LABEL description="Container for builds"
 
-# RUN rustup update nightly
-RUN rustup toolchain uninstall nightly
-# Must match nearcore/rust-toolchain.
-RUN rustup default 1.55.0
-RUN rustup target add wasm32-unknown-unknown
-
 # install build dependencies for QEMU
 RUN apt-get update && apt-get install -y  g++ rsync zip openssh-server \
      make apt-utils git sudo pkg-config libglib2.0-dev curl clang gdb llvm-dev cmake
```

That however, fails for me at runtime, with qemu not finding libnettle.so.6. Seems that we also need to re-build the qemu plugin.

bowenwang1996:
@matklad is this issue resolved?

matklad:
Not really, need to revive https://github.com/near/nearcore/pull/5071

PR #6213, opened 2022-01-28
fix: Upgrade docker image for params-estimator

jakmeier:
The latest build requires rust 1.58.1.
To make it run again, I also had to rebuild the qemu binary and plugin.


jakmeier:
Incidentally fixes #4833. It seems the problem there was that the same that I ran into: The `qemu-x86_64` binary needs to be update, as the old one required shared libraries that are not present in the newer image.

On a side note, am I seeing this right that after updating the Dockerfile, every user of the params-estimator needs to run `./build.sh` in emu-cost again manually? If yes, is this something worth improving, maybe with version tags of the local `rust-emu` image?

matklad:
>On a side note, am I seeing this right that after updating the Dockerfile, every user of the params-estimator needs to run ./build.sh in emu-cost again manually? 

I think something like this is true, yeah. I am actually pretty ignorant of docker, so not super sure about details. I just do `docker volume rm (docker volume ls -q)` and `docker rm (docker ps -a -q)` which I think cleans everything. THe image is then rebuild automatically when you `cargo run`. 

That is, we already have the logic to build the image, we just don't have a way to signal that it needs to be rebuild. @jakmeier if you know an easy way to achieve that, let's do that before this PR, so we can check that the logic actually works!

jakmeier:
@matklad I squeezed the change to automate rebuilding the docker image into this PR, I think it logically fits in here. But I'm happy to create a separate PR if you think that would be better.

The idea is that instead of using the default image tag that is simply called `latest`, we use an explicit image tag when building the image.
Before:
```
docker build --tag rust-emu runtime/runtime-params-estimator/emu-cost
```
Now:
```
docker build --tag rust-emu:rust-1.58.1 runtime/runtime-params-estimator/emu-cost
```

Now if I `cargo run`, it will crate a second image under the name `rust-emu` as shown below.
```
$ docker images rust-emu
REPOSITORY   TAG           IMAGE ID       CREATED       SIZE
rust-emu     latest        ef36120bd78b   7 hours ago   2.02GB
rust-emu     rust-1.58.1   ef36120bd78b   7 hours ago   2.02GB
```
In my case, both are the same since I already manually rebuilt the old one. For anyone else, these should be different.

If you want to regain the space, you will need to delete the old image manually. We _could_ delete it automatically, but I think it is better to keep it. Then it works nicely for bisecting and the like, since older code versions will automatically use the right image without rebuilding every time.

matklad (approved):

Issue #5750, opened 2021-12-10
Repeated epoch heights in testnet


marcelo-gonzalez:
Querying the testnet archival RPC, it's possible to observe a repeated epoch height:

<pre>$ http post [https://archival-rpc.testnet.near.org](https://archival-rpc.testnet.near.org) jsonrpc=2.0 id=dontcare method=validators params:='

 {"epoch_id": "5zb5CbwJuVNeV9pju6bQND68ku5Z5gDuArspS27tKiwT"} 

' | jq ".result.epoch_height, .result.epoch_start_height"
 3
 42463323</pre>

<pre>$ http post [https://archival-rpc.testnet.near.org](https://archival-rpc.testnet.near.org) jsonrpc=2.0 id=dontcare method=validators params:='

 {"epoch_id": "765QmBomc5Np4pcXqySt1hbVpsqV4yX9nEhH24KLEZi1"} 

' | jq ".result.epoch_height, .result.epoch_start_height"
 3
 42506523</pre>

Both epochs give height = 3 although they have different IDs and starting block heights. This doesn't seem to be the expected behavior, so worth looking into why this happened/whether it could happen again.

marcelo-gonzalez:
Ah so the same thing exists in mainnet, also at height 3

```bash
$ http post https://archival-rpc.mainnet.near.org jsonrpc=2.0 id=dontcare method=validators params:='{"epoch_id": "12RxgqCjctDewhRpvX37b4vN3pQVQnjzqAQRvEKUcigC"}' | jq ".result.epoch_height, .result.epoch_start_height"
3
9906614
```
```bash
$ http post https://archival-rpc.mainnet.near.org jsonrpc=2.0 id=dontcare method=validators params:='{"epoch_id": "BoifkxfpSLrqpEHBUdb9ogsTiqNZyTjugp9HkAH1xUeJ"}' | jq ".result.epoch_height, .result.epoch_start_height"
3
9949814
```

bowenwang1996:
Does it only repeat for epoch height 3? Might be related to the fact that we have two epochs with height 1

marcelo-gonzalez:
@bowenwang1996 Yup it's only height 3

PR #6219, opened 2022-01-31
pytest: fix a failure in skip_epoch.py

marcelo-gonzalez:
In skip_epoch.py, there are two periods where a block producer
responsible for some shard is down. The first is when a validator
never went online to begin with, and the second is when we kill one
after the chain has been running for a while. In each case, we wait
for a couple epochs to pass and then check that the balances resulting
from a bunch of transactions are applied correctly, the idea being
that we want to check that some online validator is made responsible
for the offline shard and is able to sync its state.

In the second case, we record the latest block height before killing
the validator and then wait for EPOCH_LENGTH * 3 - 1 blocks, so that
we land at least in epoch T+2. But this may not be a long enough wait
period. If the validator had already produced some blocks in the epoch
in which it's killed, it might not actually drop below the kickout threshold
for that epoch, in which case we won't expect to see that shard's chunks' "height_included"
move up to HEAD until epoch T+3.

So fix it by waiting for that long, and actually make it explicit by
introducing a utils.poll_epochs() function that does basically the same
thing as poll_blocks(), but for epochs instead


marcelo-gonzalez:
btw, I didn't change the first part of the test to use the poll_epochs() function because I tried it locally and I think it runs into https://github.com/near/nearcore/issues/5750, which complicates it.

Also failing tests are like this one and many others recently: https://nayduck.near.org/#/test/273491

mina86:
You need to run `./scripts/formatting --fix`.

mina86 (approved):

bowenwang1996 (commented):
@marcelo-gonzalez could we add more document on what the test *actually* does?

marcelo-gonzalez:
> @marcelo-gonzalez could we add more document on what the test _actually_ does?

@bowenwang1996 Now that I've looked at it some, I think it would be worth just refactoring the test so that the code itself is a little more clear. There are comments sprinkled throughout the code that explain the idea behind the test, but it's not so easy to follow. Feels more worthwhile to me to make the test code itself clearer than to just stick a big comment at the top of the current test

PR #6229, opened 2022-02-01
feat: Tracing span for deploy action

jakmeier:


runtime/runtime/src/actions.rs
```diff
@@ -436,6 +436,7 @@ pub(crate) fn action_deploy_contract(
     apply_state: &ApplyState,
     current_protocol_version: ProtocolVersion,
 ) -> Result<(), StorageError> {
+    let _span = tracing::debug_span!(target: "vm", "action_deploy_contract").entered();
```
matklad, line 439:
```suggestion
    let _span = tracing::debug_span!(target: "runtime", "action_deploy_contract").entered();
```

this is still runtime

runtime/runtime-params-estimator/src/main.rs
```diff
@@ -71,6 +71,8 @@ struct CliArgs {
     /// Print extra debug information
     #[clap(long, multiple(true), possible_values=&["io", "rocksdb", "least-squares"])]
     debug: Vec<String>,
+    #[clap(long)]
+    tracing_span_tree: bool,
```
matklad, line 75: :facepalm: it literally didn't occur to me that we can make this a command-line argument.

matklad (approved):
LGTM with vm changed to runtime in one case!

runtime/runtime-params-estimator/src/main.rs
matklad: Could you add a doc comment here though? So that it gets to the `--help` message

bowenwang1996 (approved):

PR #6232, opened 2022-02-02
fuzz: introduce fuzzer configuration

Ekleog:
For use by https://github.com/near/nayduck/pull/28


matklad:
This definitely needs some docs somewhere :)

Ekleog:
Right! I'll definitely write up the docs if the format and nayduck#28 confirm this is the way we want to go :)

Basically:
- there's a branch array, that lists the branches to fuzz along with the relative weight of the branch (more means more fuzzing)
- there's a target array, that lists the targets to fuzz along with the relative weight of the fuzz target (more means more fuzzing), plus fuzzing flags to be able to customize the fuzz target configurations

matklad:
and what we fuzz is essentially the cross-product of the two?

Ekleog:
Yup that's it!

matklad:
This looks reasonable to me, but I'll let @mina86 to approve design here. 

One bit of a doubt I have is that so far we used simple `.txt` files to collect tasks for Nayduck, and it's not clear that introducing toml just for fuzzing would be worth it. 

Oh, and one more thing: I'd rather we not add this as a literal top-level file. Top-level entries of the project are a somewhat scarce and valuable resource, and are better kept for really important things like `docs/` or `CONTRIBUTING.md`. If this is related to nayduck, moving it to `nightly/` makes some sense perhaps.

mina86:
> One bit of a doubt I have is that so far we used simple .txt files to collect tasks for Nayduck, and it's not clear that introducing toml just for fuzzing would be worth it.

I don’t have strong opinions here.  It certainly help that we don’t need a custom parser. The text files work and have various advantages (such as  being able to do `./scripts/nayduck -n | grep whatevern | ./scripts/nayduck -i` but it does parsing a bit more convoluted.

> If this is related to nayduck, moving it to nightly/ makes some sense perhaps.

Yes, that’s my thought as well. I don’t really like pollutting root of the repository with random files.

Overall looks good. Let’s wait with getting this merged once we agree on the NayDuck change.

Ekleog:
I've just moved `fuzz.toml` to `nightly/` :)

mina86 (approved):

matklad (approved):

PR #6236, opened 2022-02-03
refactor: Params-estimator new modules and tests

jakmeier:
Let us try to have only estimation functions inside the lib.rs file,
therefore all the helper functions are now in utils.rs.

The least-squares method goes into a separate module with tests.

The least-squares method for gas cost now gracefully propgagates
negative values as an error instead of panicking. The caller sites still
panick to keep this a pure refactor without functional changes.
It also has tests now.


runtime/runtime-params-estimator/src/gas_cost.rs
```diff
@@ -80,70 +80,115 @@ impl GasCost {
     }
 
     /// Performs least squares using a separate variable for each component of the gas cost.
-    pub(crate) fn least_squares_method_gas_cost(xs: &Vec<u64>, ys: &Vec<Self>) -> (Self, Self) {
+    ///
+    /// Least-squares linear regression is able to produce negative parameters
+    /// even when all input values are negative. However, negative gas costs
```
matklad, line 85:
```suggestion
    /// even when all input values are positive. However, negative gas costs
```

runtime/runtime-params-estimator/src/gas_cost.rs
```diff
@@ -80,70 +80,115 @@ impl GasCost {
...
+    /// make no sense for us. What we really want to solve is non-negative least
+    /// squares (NNLS) but that is algorithmically much more complex. Instead,
+    /// when we do get negative values, we return (A,B), where A has negative
+    /// values set to zero and B contains the values so that solution = A-B.
+    pub(crate) fn least_squares_method_gas_cost(
+        xs: &[u64],
+        ys: &[Self],
+    ) -> Result<(Self, Self), ((Self, Self), (Self, Self))> {
```
matklad, line 93: Returning a `Result` here feels a bit too complex -- we shouldn't be using negative results for anything. Let's maybe just clamp to zero and set the `GasCost` as `uncertain` (we might might want to change uncertain from a `bool` to `enum` with reasons).

runtime/runtime-params-estimator/src/gas_cost.rs
```diff
@@ -271,3 +316,148 @@ impl GasCost {
...
+        Err(((a_pos, b_pos), (a_neg, b_neg))) => {
+            assert_eq!(a_pos, GasCost::new_icount_based(1000, 1001, 1002));
+            assert_eq!(a_neg, GasCost::new_icount_based(0, 0, 0));
+
+            assert_eq!(b_pos, GasCost::new_icount_based(0, 0, 0));
+            assert_eq!(b_neg, GasCost::new_icount_based(1, 2, 3));
+        }
+        Ok(_) => panic!("Least squares got no negative values where it should have."),
+    }
+}
```
matklad, line 463: I see a minor smell here -- all these tests are coded directly to the API of the type, so, if the API of the type changes, you need to change every test. For example, if we move from returning `Result`s to using `uncertain` flag, you'd have to change every match. I think it's usually best to make the tests more data driven, by extracting a single driver function which interacts with the API, like this

```rust
#[cfg(test)]
mod tests {
  #[track_caller]
  fn check(xs: &[u64], ys: &[GasCost], expected: Result<(GasCost, GasCost), ()>) {}
  #[test]
  fn least_squares_method_gas_cost_icount_ok() {
    check(&[10, 20, 30], [
        GasCost::new_icount_based(990, 981, 972),
        GasCost::new_icount_based(980, 961, 942),
        GasCost::new_icount_based(970, 941, 912),
    ], Ok((GasCost::new_icount_based(50, 60, 70), GasCost::new_icount_based(50, 60, 70))))
  }
}
```

matklad: Oh, and yeah, I think it's better to default to having a `tests module -- that way you can have additional imports and heplers. Theres' `tmod` snippet in both IntelliJ and RA to generate that.

runtime/runtime-params-estimator/src/gas_cost.rs
jakmeier: Yeah, I am not happy with the complexity of the Result, either. But, I have consider quite carefully how I want to use this function interface and I think clamping + some flag is not sufficient. Let me share my thoughts why we need more flexibility:

* Let's say the base cost is 100k instructions + 500 bytes read + 100 bytes write. The extra cost per byte might be 50 instructions and 0 bytes each. But least-squares will most likely not return 0 bytes but instead a small number with positive or negative sign. In that case, it seems perfectly okay to ignore the negative values.
* What a "small-enough" number is depends on how large the number is compared to total value. But I don't want to fix this inside the least-squares function but rather the caller should decide the margin.
* Having a `tolerance: f64` parameter would sort of solve considerations so far.
* In addition,  sometimes we don't need `base` and `factor`. For example, in deployment cost per byte, we already have the base estimated differently. If least-squares produces a negative base but a decent factor, this is probably okay. I would only want to do some sanity checks that the base is still close to zero.
* In conclusion, the caller should be able to receive also the negative results and decide what to do with them.
* Then, mabye it seems like the easiest solution is to use `Ratio<i64>` instead inside `GasCost`. But I *really* dislike that, as it breaks the guarantee that an estimating function returning `GasCost` can only return positive values.
* Lastly, we could make `GasCost<T>` generic and use `Ratio<T>` inside. Least-squares uses `GasCost<i128>` as result but everywhere else we use `GasCost<u64>`. (Or we use an entirely new type `LeastSquaresResult` that look eerily similar to `GasCost<i128>`.) 

Personally, I dislike my last suggestion slightly more than what I have now but I am open to hear your thoughts!

runtime/runtime-params-estimator/src/gas_cost.rs
jakmeier: Thanks for the suggestions, definitely makes sense to do as you write. I will work on the change

runtime/runtime-params-estimator/src/gas_cost.rs
matklad: :thinking: yeah, indeed, it seems we don't have super good choices here. Convinced that the current signature is OK. Though, prehaps we can generalize from "the caller should be able to receive also the negative results and decide what to do with them" and "Having a tolerance: f64" to having some kind of `tolerance: Tolerance`, which allows the caller to pick one of the fixed strategies?

runtime/runtime-params-estimator/src/lib.rs
```diff
@@ -492,7 +499,8 @@ fn action_deploy_contract_per_byte(ctx: &mut EstimatorContext) -> GasCost {
         ys.push(cost);
     }
 
-    let (_base, per_byte_cost) = GasCost::least_squares_method_gas_cost(&xs, &ys);
+    let (_base, per_byte_cost) =
+        GasCost::least_squares_method_gas_cost(&xs, &ys).expect("Returned negative parameters");
```
matklad, line 503: Here, I think it might be better to flag the cost as uncertain, rather than panic, so that, if you run a bunch of estimates, you still can conclude the rest of them.

jakmeier: Heh, yeah I'm already running such code to actually get the estimate :)

I thought I will do all the functional changes in a separate PR, including the 5/4 factor and a few other things. But I can include this change here if you want.

matklad: no opinion on this!

runtime/runtime-params-estimator/src/gas_cost.rs
jakmeier: Okay, i gave it another shot. I found it rather tricky to introduce this generalization layer without adding too much complexity, but still allowing the caller-site flexibility I want.

My current solution is a  bit of trade-off:
1. The public method `least_squares_method_gas_cost` now doesn't have a complex return type
2. The new type `LeastSquaresTolerance` is public and must be understood by the caller
3. More code inside the `gas_cost.rs` module (but at least this is mostly "private code")
4. Caller can only chose a fixed tolerance strategy and this strategy is always on the entire `GasCost`. (Cannot differentiate between things like "IO reads can be negative but instructions must be positive". )
5. The caller doesn't know if an uncertain result was due to least-squares negative values, or because the inputs were already uncertain.
6. On the positive, I think I managed to write the code in a way that we could add support for 4 with new tolerance variations and 5 with converting `uncertain` to an enum. Both of these can be done, if we really need it at some point. But for now, I left it out to not blow up the code for things that maybe are not necessary at all.

Personally, I still don't really like my solution here... But I think it at least a small improvement on what I had before.

runtime/runtime-params-estimator/src/gas_cost.rs
jakmeier: This should be addressed with the newest commit, if I understood your suggestions right

runtime/runtime-params-estimator/src/lib.rs
jakmeier: Changed this to use the new tolerance factor, which should mean we avoid all runtime panics

runtime/runtime-params-estimator/src/gas_cost.rs
jakmeier: @matklad what do you think of the new function interface for least squares?

matklad (approved):

PR #6237, opened 2022-02-03
aarch64: fix compilation issue

### LINKED ISSUE #6222: Different PeerMessage size on aarch64
It seems that `PeerMessage` has a different size on aarch64 platforms. When I try to build neard on a M1 MacBook, using the native target `aarch64-apple-darwin`, the build fails.

```
jakmeier@Jakobs-Air nearcore % make                               
[...]
error[E0080]: evaluation of constant value failed
   --> chain/network/src/network_protocol.rs:219:15
    |
219 | const _: () = assert!(std::mem::size_of::<PeerMessage>() <= 1144, "PeerMessage > 1144 bytes");
    |               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the evaluated program panicked at 'PeerMessage > 1144 bytes', chain/network/src/network_protocol.rs:219:15
    |
    = note: this error originates in the macro `$crate::panic::panic_2021` (in Nightly builds, run with -Z macro-backtrace for more info)

For more information about this error, try `rustc --explain E0080`.
```

This assertion has been added with #5834. I am not sure what the implications are here for having the larger struct size on aarch64. 

@pmnoxx can you maybe comment on this?

My thoughts are, maybe that is okay, since aarch64 is only local-development and not for production? In that case, maybe it's enough to ignore the assertion on aarch64.

Ekleog:
The bound is too tight for aarch64 for now. Given that we don't run
aarch64 in production yet and would have to audit all the code before we
do anyway, let's just disable this check for aarch64.

Fixes #6222

### DIFF
--- a/chain/network/src/network_protocol.rs
+++ b/chain/network/src/network_protocol.rs
@@ -216,6 +216,7 @@ pub enum PeerMessage {
     #[cfg(feature = "protocol_feature_routing_exchange_algorithm")]
     RoutingTableSyncV2(RoutingSyncV2),
 }
+#[cfg(target_arch = "x86_64")] // Non-x86_64 doesn't match this requirement yet but it's not bad as it's not production-ready
 const _: () = assert!(std::mem::size_of::<PeerMessage>() <= 1144, "PeerMessage > 1144 bytes");
 
 #[cfg(feature = "protocol_feature_routing_exchange_algorithm")]


jakmeier (approved):
Makes sense to me!

bowenwang1996 (approved):

Issue #3487, opened 2020-10-14
Garbage collect headers


SkidanovAlex:
Today we store all the headers from genesis. Headers are rather large, and it won't take long before storing them starts consuming unacceptable amount of storage.

This issue tracks adding support for garbage collecting old headers.

PR #6238, opened 2022-02-03
client: add GetBlockHash request and use it in Rosetta RPC

mina86:
The `/network/status` Rosetta RPC returns, among other things, height
and hash of the genesis block.  Currently, it does it by fetching the
genesis block.  Sadly, in non-archival nodes, genesis block is usually
unavailable which leads to failures in serving the RPC.  To address
this issue, introduce a `GetBlockHash` request which returns only the
hash of a block.  Since height→hash mapping is not garbage collected,
this request will always work even for garbage collected blocks.

### DIFF
--- a/chain/chain/src/chain.rs
+++ b/chain/chain/src/chain.rs
@@ -2727,6 +2727,12 @@ impl Chain {
         self.store.get_block(&hash)
     }
 
+    /// Gets block hash from the current chain by height.
+    #[inline]
+    pub fn get_block_hash_by_height(&mut self, height: BlockHeight) -> Result<CryptoHash, Error> {
+        self.store.get_block_hash_by_height(height)
+    }
+
     /// Gets a block header by hash.
     #[inline]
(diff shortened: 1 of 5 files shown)


mina86:
Then again, maybe it would be even better to add `GetBlockHash` instead? That’s all Rosetta needs and it wouldn’t introduce dependency on storing all headers in the database.

frol (approved):

frol:
I believe this approach if fair enough, so we can live with it until nearcore needs to do anything about the fact of storing all the block headers.

mina86:
> I believe this approach if fair enough, so we can live with it until nearcore needs to do anything about the fact of storing all the block headers.

I actually already had a commit with the change just as you commented. ;) I’m gonna push it to this PR. If others have preferences for one or the other variant (`GetBlockHeader` or `GetBlockHash`) I can easily swap between the two.

bowenwang1996 (approved):
This would work fine as a stop-gap solution, but when #3487 is implemented, we need a different solution.

khorolets (approved):

PR #6239, opened 2022-02-03
chore: bump version to 0.12.0

miraclx:
Tagged: https://github.com/near/nearcore/releases/tag/crates-0.12.0

Notable changes:

- https://github.com/near/nearcore/pull/6179
- https://github.com/near/nearcore/pull/6158

Check here for a full changelog: https://github.com/near/nearcore/compare/crates-0.11.0...crates-0.12.0

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -54,7 +54,7 @@ members = [
 ]
 
 [workspace.metadata.workspaces]
-version = "0.11.0"
+version = "0.12.0"
 
 [patch.crates-io]
 


austinabell (approved):

austinabell:
Note: I don't have context around the unrelated changes since 0.11. I tried to skim through, but if anyone wants to point out any key changes that might be breaking, that might be helpful.

miraclx:
Yeah, there's no lib-specific changelog yet, suppose there should be..

But I took some time to look through the diff https://github.com/near/nearcore/compare/crates-0.11.0...d8d7448b146c6a793475efac3c0dde0fe35f755a, sorted out the crates we're actually publishing and highlighted the notable changes above. We should be good.

Then again, the versions are relatively inexpensive, if we have an issue, we can just cut another release.

PR #6241, opened 2022-02-03
feat: cache chunk nodes

Longarithm:
This is currently a draft for discussing implementation.

I also include here delayed update with trie cache size increase to 100_000 and max value size reduction to 1_000 ([link to reasoning](https://near.zulipchat.com/#narrow/stream/308695-nearinc.2Fprivate/topic/gas.20usage.20mainnet/near/264523342))

TODO:
* protocol feature
* tests
* performance stats
* cost stats


Longarithm:
I changed implementation a bit.

* Removed `bool` with more meaningful `RetrievalCost` which should be self-descriptive.
* Make `chargeable_retrieve_raw_bytes` a trait method with default implementation which should make it less hacky. 
Alternatively we could require returning `(Vec<u8>, RetrievalCost)` from `retrieve_raw_bytes` but we don't need it in the majority of cases.

@abacabadabacaba

core/store/src/trie/trie_storage.rs
```diff
@@ -13,25 +13,120 @@ use std::cell::RefCell;
...
 
     pub fn clear(&self) {
-        self.0.lock().expect(POISONED_LOCK_ERR).clear()
+        let mut guard = self.0.lock().expect(POISONED_LOCK_ERR);
+        guard.cache_state = CacheState::CachingShard;
+        guard.shard_cache.clear();
+        guard.chunk_cache.clear();
+    }
+
+    pub fn flip_caching_chunk_state(&self) {
```
matklad, line 116: Hm, so, from perspective of idiomatic Rust which cares about ownership and aliasing, this is as bad a signature as it gets. 

The type of this function tells you "'it's safe to call this concurrently from any thread in any moment, because it won't affect concurrent users". The reality is very much the opposite -- the order in which this method and others are called is very much important, there's a specific ordered protocol where you must first flip, then do some operations, than flip back, and it very much relies on exclusive access. 

The fundamental issue here is that our "read" operation isn't really a read, in a sense that it is not idempotent, and has observable side effects (I think this became the case when we added touching `TouchedNodesCounter` which uses the same faux-atomics, while in fact it requires unique access). In this case we have an atypical cache, which is not transparent to the caller, and where we, on the contrary, care very much about internal operation. From Rust's perspective, we really want to have `&mut self` even for `.get` here. 

An extra smell here is that, I think, we only this extra cache to be in effect for the duration of `apply` method: 

https://github.com/near/nearcore/blob/6f9aea9925812ce17b285ccc0876e17c22e762c2/runtime/runtime/src/lib.rs#L1169-L1180

But we are forced to make it global, and force every client of the trie to use implementation with this extra accounting. This also makes stuff more confusing (hence my question of what the heck is chunk and why we care about it here). At the level of abstraction where we implement this, we have chunks and blocks, and care about them. At the level of abstraction where we use it (`Runtime::apply`), we don't have those, we just have a bunch of transactions/recipts.

<hr/>

I don't have a super specific and confident suggestion for the incremental step to make this better. This is a typical hard case where we want to ship "cool feature", but we have "suboptimal architecture" in place, and really need a "big refactor". But doing refactoring and feature-work at the same time is pushing opposing incentives. 

That being said, I wonder if the following would work. Rather than adding the cache to the trie itself, we can *inject* the cache as a parameter to the `get` method. That way, we can create the cache at the beginning of the `apply` method, and have `&mut` access to it. This *cache* would than be able to do the required accounting. In code,

```rust
pub struct Trie {
    pub(crate) storage: Box<dyn TrieStorage>,
    // NB: no node counter here
}

/// A trait to "decorate" access to storage
pub trait RetrieveRawBytes {
    fn retrieve_raw_bytes(
        &mut self,
        key: &CryptoHash,
        // undecorated access
        thunk: impl FnOnce() -> Result<Vec<u8>, StorageError>,
    ) -> Result<Vec<u8>, StorageError> {
    }
}

/// Trivial decorator, which just calls the thunk directly
impl RetrieveRawBytes for () {
    fn retrieve_raw_bytes(
        &mut self,
        key: &CryptoHash,
        thunk: impl FnOnce(&CryptoHash) -> Result<Vec<u8>, StorageError>,
    ) -> Result<Vec<u8>, StorageError> {
        thunk()
    }
}

impl Trie {
    // lookup delegates to lookup_with with trivial retrieve_bytes
    fn lookup(
        &self,
        root: &CryptoHash,
        key: NibbleSlice<'_>,
    ) -> Result<Option<(u32, CryptoHash)>, StorageError> {
        self.lookup_with(root, key, &mut ())
    }

    // The core of the pattern: the caller can customize the behavior here by
    // injecting a custom RetrieveRawBytes. Note also aliasing interplay:
    // `self` is `&`, only `r` is `&mut`.
    fn lookup_with(
        &self,
        root: &CryptoHash,
        mut key: NibbleSlice<'_>,
        r: &mut impl RetrieveRawBytes,
    ) -> Result<Option<(u32, CryptoHash)>, StorageError> {
        let mut hash = *root;

        loop {
            if hash == Trie::empty_root() {
                return Ok(None);
            }
            // Here we call the decorator, passing in the undecorated thunk to it.
            let bytes = r.retrieve_raw_bytes(&hash, || self.storage.retrieve_raw_bytes(&hash))?;
            ...
        }
    }

    // `lookup` is an internal method, we need to do the similar `foo/foo_with` bifurcation of the whole public API.
    pub fn get(&self, root: &CryptoHash, key: &[u8]) -> Result<Option<Vec<u8>>, StorageError> {
        self.get_with(root, key, &mut ())
    }

    pub fn get_with(
        &self,
        root: &CryptoHash,
        key: &[u8],
        r: &mut impl RetrieveRawBytes,
    ) -> Result<Option<Vec<u8>>, StorageError> { ... }
}

pub struct TrieUpdate {
    pub trie: Rc<Trie>,
    // That's our `RetrieveRawBytes`
    trie_node_cache: TrieNodeCache,
    root: CryptoHash,
    committed: RawStateChanges,
    prospective: TrieUpdates,
}

/// Mediator for trie node access, which counts the number of touched nodes and maintains an extra cache. 
/// Note that we no longer need atomics here, as we gained `&mut`
struct TrieNodeCache {
    n_touched_nodes: u64,
    nodes: HashMap<CryptoHash, Vec<u8>>,
}

impl RetrieveRawBytes for TrieNodeCache {
    fn retrieve_raw_bytes(
        &mut self,
        key: &CryptoHash,
        thunk: impl FnOnce() -> Result<Vec<u8>, StorageError>,
    ) -> Result<Vec<u8>, StorageError> {
        // The actual logic of caching access and counting the misses.
        let res = match self.nodes.entry(*key) {
            Entry::Occupied(entry) => entry.get().clone(),
            Entry::Vacant(entry) => {
                let value = thunk()?;
                self.n_touched_nodes += 1;
                entry.insert(value.clone());
                value
            }
        };
        Ok(res)
    }
}

impl TrieUpdate {
   /// Note that this is `&mut self` now, which makes sense. When you do accesses via `TrieUpdate`, 
   /// you want to count cache misses, and to do that correctly you generally want exclusive access
   pub fn get(&mut self, key: &TrieKey) -> Result<Option<Vec<u8>>, StorageError> {
        let key = key.to_vec();
        if let Some(key_value) = self.prospective.get(&key) {
            return Ok(key_value.value.as_ref().map(<Vec<u8>>::clone));
        } else if let Some(changes_with_trie_key) = self.committed.get(&key) {
            if let Some(RawStateChange { data, .. }) = changes_with_trie_key.changes.last() {
                return Ok(data.as_ref().map(<Vec<u8>>::clone));
            }
        }

        // Here we used extended Trie API by passing `self.trie_node_cache` as the decorator for the underlying node access. 
        self.trie.get_with(&self.root, &key, &mut self.trie_node_cache)
    }
}
```

matklad:
One question here is would this mandatory cache increase worst-case memory usage? Ie, before this change we generally assume that we can always just load trie nodes from disk, and don't have to keep them in memory, while now we promise that all touched nodes are in memory?

In this PR, we increase the LRU limit to 100_000. One estimate for the max number of nodes we can touch is 1000TGas / touching_trie_node cost which is `1000*10**12 / 16101955926 = 62104`. That is, we have a guarantee that all touched nodes will be present in the LRU. So, implementation wise, maybe we don't need a second cache and the logic to move the values between the two, but rather just the set of nodes we touched? 

```rust
// Current Code
struct TrieCache {
    cache_state: CacheState,
    shard_cache: LruCache<CryptoHash, Vec<u8>>,
    chunk_cache: HashMap<CryptoHash, Vec<u8>>,
}

/// Suggestion
struct TrieCache {
    cache_state: CacheState,
    shard_cache: LruCache<CryptoHash, Vec<u8>>,
    touched_nodes: HashSet<CryptoHash>,
}

#[test] 
fn lru_is_big_enough() {
    let chnuk_limit = 1000*10.pow(12);
    let touching_trie_node_cost = RuntimeConfigStore::for_protocol_version(PROTOCOL_VERSION).touching_trie_node;
    let max_touched_nodes = block_limit / touching_trie_node_cost;
    assert!(TRIE_MAX_CACHE_SIZE > max_touched_nodes, "we assume that all nodes touched in a single chunk are present in in-memory cache")
}
```

Longarithm:
@matklad 

> One question here is would this mandatory cache increase worst-case memory usage? Ie, before this change we generally assume that we can always just load trie nodes from disk, and don't have to keep them in memory, while now we promise that all touched nodes are in memory?

Yeah we have this promise. The estimate for the max number of nodes is even 31052 (because we care only about receipts gas limit which is 500 Tgas), so cache size for all shards won't exceed `131K * 4 * 1K = 524M`.
In reality it will be even lower, because as you mention below we expect most of this nodes occur in LRU anyway.

> So, implementation wise, maybe we don't need a second cache and the logic to move the values between the two, but rather just the set of nodes we touched?

Great question! The problem here is that we also have lots of uncharged trie accesses everywhere, e.g. we call `get_access_key` on add/delete key actions. So unfortunately we don't have a good estimation of the number of touched nodes, and we don't have a guarantee that chunk nodes won't be incorrectly eliminated from LRU. [(Zulip discussion)](https://near.zulipchat.com/#narrow/stream/296545-nearinc.2Fprotocol-core.2Fprivate/topic/uncharged.20trie.20accesses/near/269325953)

In the long term, we plan to charge for these touches as well.

core/store/src/trie/trie_storage.rs
```diff
@@ -13,25 +13,120 @@ use std::cell::RefCell;
...
+    ShardCache(Vec<u8>),
+    ChunkCache(Vec<u8>),
+}
+
+enum CacheState {
+    CachingShard,
+    CachingChunk,
+}
+
+pub enum RetrievalCost {
```
bowenwang1996, line 35: I think this name is too general. Mabye `TrieNodeRetrievalCost` is better?

core/store/src/trie/trie_storage.rs
```diff
@@ -13,25 +13,120 @@ use std::cell::RefCell;
...
+    Free,
+    Full,
+}
 
 impl TrieCache {
+    fn get_cache_position(&mut self, hash: &CryptoHash) -> CachePosition {
+        match self.chunk_cache.get(hash) {
+            Some(value) => CachePosition::ChunkCache(value.clone()),
```
bowenwang1996, line 43: do we want to always clone here?

core/store/src/trie/trie_storage.rs
```diff
@@ -13,25 +13,120 @@ use std::cell::RefCell;
...
+            None => match self.shard_cache.get(hash) {
+                Some(value) => CachePosition::ShardCache(value.clone()),
+                None => CachePosition::None,
+            },
+        }
+    }
+
+    pub fn get_with_cost(&mut self, hash: &CryptoHash) -> (Option<Vec<u8>>, RetrievalCost) {
```
bowenwang1996, line 51: I suggest that we create a new type for the return value here. It is a bit hard to read otherwise

core/store/src/trie/trie_storage.rs
```diff
@@ -162,31 +267,42 @@ impl TrieCachingStorage {
...
                 // not StorageError::TrieNodeMissing because it's only for TrieMemoryPartialStorage
                 Err(StorageError::StorageInconsistentState("Trie node missing".to_string()))
             }
         }
     }
 
+    pub fn prepare_trie_cache(&self) {
+        let mut guard = self.cache.0.lock().expect(POISONED_LOCK_ERR);
+        guard.cache_state = CacheState::CachingShard;
+        guard.drain_chunk_cache();
```
bowenwang1996, line 297: if `drain_chunk_cache` is only used here, we might want to inline the implementation to avoid confusion

Longarithm:
@matklad I like the idea of making `TrieCache` a part of `TrieUpdate`. In the current PR version I reflected it. It allows us to get rid of "flips" and `TouchedNodesCounter` in trie storage.

Unfortunately there are many assumptions in the code that `TrieUpdate` is manipulated using immutable reference, and some of them are nontrivial to fix (see `storage_get` and `check_balance`).

So now I see a choice between two hard decisions:
* fix immutable `TrieUpdate` references in many places, carefully ensuring that nothing breaks;
* leave `TrieCache` in `TrieStorage` but also have these dangerous flips.

PolyProgrammist:
Good job:)

matklad:
I think there's a third option: put the thing in `TrieUpdate`, but use non-thread-safe interior mutability for now, with the hope to refactor to `&mut` later. 

```rust
pub trait RetrieveRawBytes {
    fn retrieve_raw_bytes(
        &self,
        key: &CryptoHash,
        // undecorated access
        thunk: impl FnOnce() -> Result<Vec<u8>, StorageError>,
    ) -> Result<Vec<u8>, StorageError> {
    }
}

struct TrieNodeCache {
    n_touched_nodes: Cell<u64>,
    nodes: RefCell<HashMap<CryptoHash, Vec<u8>>>,
}
```

:thinking: Actually, I think the interior mutability here might not be horribly bad, as long as it is confined to a single thread (so, `RefCell` rather than `Mutex`).

core/store/src/trie/trie_storage.rs
Longarithm: It can be improved, but so far the only usage of this function (`retrieve_raw_bytes`) requires cloning.

On the other hand, `retrieve_raw_bytes` itself can be improved by returning a reference, but I would treat this as a separate issue https://github.com/near/nearcore/issues/6270

core/store/src/trie/trie_tests.rs
```diff
@@ -126,3 +127,41 @@ fn test_reads_with_incomplete_storage() {
         }
     }
 }
+
+// Helper for tests ensuring the correct behaviour of trie counter.
+// For example, on testing set of keys `[b"aaa", b"abb", b"baa"]`, we expect 6 touched nodes to get value for the first
+// key: Branch -> Extension -> Branch -> Extension -> Leaf plus retrieving the value by its hash.
+fn get_touched_nodes_numbers(with_chunk_cache: bool) -> Vec<u64> {
```
bowenwang1996, line 134: shouldn't it also take keys as an input :)

bowenwang1996 (commented):
I think it would be helpful if you could describe how chunk cache and shard cache interact with each other

Longarithm:
Superseded by https://github.com/near/nearcore/pull/6292.

PR #6245, opened 2022-02-04
refactor: remove unused RPC error

matklad:
ContractCallError was never exposed through RPC (it is intentionally not
serialize), so we shouldn't make it an RpcError.


matklad:
Actually, let me thing more about it: it might be the case that the problem is that we *don't* use this type as an error...

PR #6246, opened 2022-02-04
fix(network-primitives): pin actix-derive to fix isolated compilation issues

miraclx:
Context: https://buildkite.com/nearprotocol/nearcore-release/builds/1802#48b4d213-9613-462b-bb5d-350a3ad87d3a/106-1194

```console
$ cargo package -p near-network-primitives
   Compiling near-network-primitives v0.12.0 (/var/lib/buildkite-agent/builds/buildkite-i-08e77618f4765ff4c-1/nearprotocol/nearcore-release/target/package/near-network-primitives-0.12.0)
error[E0412]: cannot find type `OneshotSender` in module `actix::dev`
   --> src/types.rs:322:17
    |
322 | #[derive(Debug, actix::MessageResponse)]
    |                 ^^^^^^^^^^^^^^^^^^^^^^ not found in `actix::dev`
    |
    = note: this error originates in the derive macro `actix::MessageResponse` (in Nightly builds, run with -Z macro-backtrace for more info)

error[E0049]: method `handle` has 0 type parameters but its trait declaration has 1 type parameter
   --> src/types.rs:322:39
    |
322 | #[derive(Debug, actix::MessageResponse)]
    |                                       ^ found 0 type parameters, expected 1
    |
    = note: this error originates in the derive macro `actix::MessageResponse` (in Nightly builds, run with -Z macro-backtrace for more info)
```

We didn't have this issue before https://github.com/near/nearcore/pull/5945 which uses `actix_derive`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2884,6 +2884,7 @@ name = "near-network-primitives"
 version = "0.0.0"
 dependencies = [
  "actix",
+ "actix_derive",
  "anyhow",
  "borsh",
  "chrono",
--- a/chain/network-primitives/Cargo.toml
+++ b/chain/network-primitives/Cargo.toml
@@ -13,6 +13,7 @@ publish = true
 [dependencies]
 anyhow = "1.0.51"
 actix = "=0.11.0-beta.2"
+actix_derive = "=0.6.0-beta.1" # Pinned dependency in addition to actix dependecy (remove this line once the pinning is not needed)
 borsh = "0.9"
 chrono = { version = "0.4.4", features = ["serde"] }
 deepsize = { version = "0.2.0", optional = true }


matklad:
Hm, why doesn't it pick the lockfile? As we have a lockfile, we shouldn't need to pin in Cargo.toml.

miraclx:
`cargo package` doesn't use the lock file. It copies the source and tries to build it in isolation before packaging it.

matklad:
I suggest passing `--no-verify` to `cargo package` then. We only guarantee that versions from the `.lock` work. We *could* spend extra mental and CI cycles making sure that the code builds even without Cargo.lock, but that doesn't seem on ballance beneficial for stuff which isn't public API.

miraclx:
To note, this isn't an issue we aren't already familiar with. And we already have this patch in nearcore. (exhibit: [A](https://github.com/near/nearcore/blob/9110028c663b5f9c1a353e064a535f0ebedd82e0/nearcore/Cargo.toml#L14))

Should be fixed with the recently released versions of actix. But we still need to cut a new release of nearcore crates, and [we have so far](https://user-images.githubusercontent.com/16881812/152619180-3609860f-5988-4586-a3e3-de41460dcc0d.png), at least until this error was reported while attempting to publish `near-network-primitives`.

This is just a temporary fix and would be cleaned up when upgrading actix.

Also, adding `--no-verify` might be counter-intuitive, because then, we'll be publishing a crate that doesn't actually compile. And that's a bad user experience downstream.

mm-near (approved):

matklad:
>To note, this isn't an issue we aren't already familiar with.

Yeah, that makes sense then! 

>Also, adding --no-verify might be counter-intuitive, because then, we'll be publishing a crate that doesn't actually compile. And would be forcing downstream users of said crates to pin their actix_derive versions, when instead, we can save them the stress by doing this in here.

I would like to push back a bit against this line of thinking -- I feel this kinds of local improvements can lead to trouble down the line, where upstream and downstream become de-facto entangled, which would make changes harder to make. To me, it seems that the downstream having *our* actix in *their* lockfile is a problem in itself. If someone cares that we use actix internally, that means that it'd be hard to upgrade actix or replace it with something else (and, given the volatile nature of rust web-space, such change doesn't seem unlikely over the long time horizon to me). 

There's a useful feedback here -- if some dependency causes pain downstream, it's a good signal that we want to better encapsulate this dependency and make the downstream not care about it at all, rather than just making it less painful.

miraclx:
@matklad, you're right. Having `actix` in our dep tree at all for primitive types is one of those odd little things we had to live with because the JSON RPC server and client shared the same types and the server implemented `actix::MailboxError` for the RPC error types. By the time we're done decoupling the types, we should have this resovled. So depending on `near-jsonrpc-primitives` doesn't pull much more than is needed.

Issue #6183, opened 2022-01-25
Remove state-viewer as a standalone tool


matklad:
After https://github.com/near/nearcore/pull/5333, state-viewer is available as a subcommand of `neard`. That means we can get rid of separate state-viewer binary:

* https://github.com/near/nearcore/blob/96a4d1a2f76ec10280bf79ba3a5fa6035081e158/test-utils/state-viewer/src/main.rs
* https://github.com/near/nearcore/blob/96a4d1a2f76ec10280bf79ba3a5fa6035081e158/Makefile#L24

nikurt:
and also move the code of state-viewer out of the test-utils directory

PR #6251, opened 2022-02-07
Remove references to the state-viewer tool

nikurt:
https://github.com/near/nearcore/issues/6183

Doesn't remove the state-viewer binary yet.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,7 +21,6 @@ docker-nearcore-nightly:
 
 
 release: neard-release
-	cargo build -p state-viewer --release
 	cargo build -p store-validator --release
 	cargo build -p runtime-params-estimator --release
 	cargo build -p genesis-populate --release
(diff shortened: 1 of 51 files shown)


Makefile
```diff
@@ -21,7 +21,6 @@ docker-nearcore-nightly:
 
 
 release: neard-release
-	cargo build -p state-viewer --release
```
mina86, line 24: Don’t forget to change near-ops as well.

pytest/lib/branches.py
```diff
@@ -57,7 +57,6 @@ def download(url: str) -> str:
 class Executables(typing.NamedTuple):
     root: pathlib.Path
```
mina86, line 58: At this point we can also remove `root` and replace it with `root = property(lambda self: self.neard.parent)`. Not really sure if it’s worth the change though.

mina86 (approved):

PR #6252, opened 2022-02-07
pytest: fix sync_chunks_from_archival.py test

mina86:


pytest/tests/sanity/sync_chunks_from_archival.py
```diff
@@ -239,35 +239,45 @@ def get_validators(node):
...
+    # nodes so we’re again checking nodes 2 and 3.
     for h in range(new_height - 5, new_height - 1):
-        for shard in [0, 1]:
-            for producer in [2, 3]:
-                assert (h, shard, producer) in requests, h
-                assert (h, shard, producer) in responses, h
+        for shard in (0, 1):
+            found = False
+            for node in (2, 3):
+                if (height, shard, node) in requests:
```
mm-near, line 277: should it be height or h ?

mm-near (dismissed):

pytest/tests/sanity/sync_chunks_from_archival.py
mina86: Good catch. Fixed.

mm-near (approved):

PR #6253, opened 2022-02-07
chain: refactor assign_with_possible_repeats in shard_assignment.rs

mina86:
Restructure the code a little such that `shard_index` is popped/peeked
in only one location and the ‘if→loop→if+else-if+else-if+…’ structure
is simplified to ‘loop→match’ which is hopefully slightly easier to
comprehend.

While at it, use `peek` when accessing the `shard_index` heap.  This
way there’s only one place in the code where elements are popped and
added to buffer.  Again, this hopefully simplifies data flow inside of
the function.


core/primitives/src/utils/min_heap.rs
```diff
@@ -20,9 +20,19 @@ impl<T: Ord> MinHeap<T> {
...
+    pub fn peek_mut(&mut self) -> Option<PeekMut<'_, Reverse<T>>> {
+        self.inner.peek_mut()
+    }
+
     pub fn len(&self) -> usize {
         self.inner.len()
     }
+
+    // This sadly leaks an abstraction.
+    pub fn extend<I: std::iter::IntoIterator<Item = Reverse<T>>>(&mut self, iter: I) {
```
matklad, line 33:
```suggestion
    pub fn extend<I: IntoIterator<Item = Reverse<T>>>(&mut self, iter: I) {
```

IntoIterator is in prelude I think

core/primitives/src/utils/min_heap.rs
```diff
@@ -20,9 +20,19 @@ impl<T: Ord> MinHeap<T> {
         self.inner.peek().map(|Reverse(t)| t)
     }
 
+    // This sadly leaks an abstraction.
+    pub fn peek_mut(&mut self) -> Option<PeekMut<'_, Reverse<T>>> {
```
matklad, line 24: We can define our own `MinHeapPeakMut` to plug the leak. Not sure if it's worth it. Though, if we do wrap the type, we might as well wrap fully?

matklad (approved):
Approving to unblock, didn't check that the logic is equivalent. How resilient the code is to the mutation testing here? Do test fail if we forget some condition?

core/primitives/src/utils/min_heap.rs
```diff
@@ -134,4 +164,27 @@ mod tests {
...
+
+        let top = heap.peek_mut().unwrap();
+        assert_eq!(17, *top);
+        PeekMut::pop(top);
+        assert_eq!(Some(&37), heap.peek());
+
+        let mut top = heap.peek_mut().unwrap();
+        assert_eq!(37, *top);
+        *top = 111;
+        core::mem::drop(top);
```
mm-near, line 185: can you comment about this

chain/epoch_manager/src/shard_assignment.rs
```diff
@@ -87,55 +87,44 @@ fn assign_with_possible_repeats<T: HasStake + Eq, I: Iterator<Item = T>>(
...
-                let (least_validator_count, shard_stake, shard_id) =
-                    shard_index.pop().expect("shard_index should never be empty");
-                if least_validator_count >= min_validators_per_shard {
+        // Decide which shard to assign this chunk producer to.  We mustn’t
+        // assign producers to a single shard multiple times.
+        loop {
+            match shard_index.peek_mut() {
+                None => {
+                    // No shards left which don’t already contain this chunk
+                    // producer.  Skip it and move to another validator.
```
mm-near, line 96: nit: you sometimes say "chunk producer" and sometimes "validator" -- if you're talking about the same entity, let's use the same naming

chain/epoch_manager/src/shard_assignment.rs
```diff
@@ -87,55 +87,44 @@ fn assign_with_possible_repeats<T: HasStake + Eq, I: Iterator<Item = T>>(
...
+                    break;
+                }
+                Some(top) if top.0 >= min_validators_per_shard => {
                     // All remaining shards have min_validators_per_shard chunk
-                    // producers assigned to them.  Don’t assign current cp to
-                    // any shard and move to next cp.
-                    buffer.push((least_validator_count, shard_stake, shard_id));
+                    // producers assigned to them.  Don’t assign current
```
mm-near, line 101: producers assigned to them, as shard_index is sorted by number of validators assigned.

chain/epoch_manager/src/shard_assignment.rs
```diff
@@ -87,55 +87,44 @@ fn assign_with_possible_repeats<T: HasStake + Eq, I: Iterator<Item = T>>(
...
+                    // validator to any shard and move to next cp.
```
mm-near, line 102: actually in this case - we'll probably also fail the line 86 condition - and completely exit.

mm-near (approved):

chain/epoch_manager/src/shard_assignment.rs
mina86: No. We may have a shard which still needs another chunk producer assigned to it but the current producer is already assigned to that shard. In this case condition on line 86 will be true, we then get to the case on line 105 and skip the shard for now and get back to looking at next shard which might turn out to have all the validators we need.

Issue #6226, opened 2022-02-01
Use custom profiles to set LTO and such


matklad:
Rust 1.57 stabilized custom profiles feature of cargo: 

https://github.com/rust-lang/cargo/pull/9943/

That means that we can switch from using env variables for LTO

https://github.com/near/nearcore/blob/d1d45597abe41228e60f725e56c216e87a4c9b84/Makefile#L1-L2

to using a proper `--profile production`

Steps here: 

* Add custom profile to root Cargo.toml
* Search all the places that currently set CARGO_PROFILE_RELEASE_LTO (makefile, estimator, maybe something else?)
* Update them all to pass `--profile production` instead.

matklad:
cc @jakmeier for estimator changes.

mina86:
Turns out this isn’t as trivial as I thought at first. Issue is that with `--profile production` the binary ends up at `target/production/neard` which means that places which refer to `target/release` would need to be updated as well.

I wonder that maybe it would be better to enable LTO in release profile and introduce dev-release profile equivalent to current release? Just thinking out loud.

matklad:
Good point. I'd probably still lean towards for introduction of `prod` profile: today, to build production artifacts, you are going to use a custom, unusual workflow anyway (`make`), so adding more stuff there doesn't affect conceptual complexity (it's still a custom command!). In contrast, for workflows like "running a test in release mode to access performance" the standard `cargo test --release` works, and making that into non-standard `cargo test --profile dev-release`  does add to conceptual complexity. 

Ideally, `dev-release` would be a standard Cargo profile but, alas, it isn't

abacabadabacaba:
I think that release profile should be what we are actually releasing. Anything less is not a release, it's "debug with optimizations" or something like that.

mina86:
> I'd probably still lean towards for introduction of prod profile: today, to build production artifacts, you are going to use a custom, unusual workflow anyway (make), so adding more stuff there doesn't affect conceptual complexity (it's still a custom command!).

The issue is that people expect `make release` to put all the files in `target/release` which will no longer be true with `prod` profile. I’d rather avoid a situation where someone was testing some code with `cargo build --release` and later made a release by running `make release` and copying files from `target/release`. We can copy the files over in Makefile but then we just made the build commands more complicated for no discernable reason. As you know, there is `--out-dir` but it’s unstable so we probably don’t want to use it.

> I think that release profile should be what we are actually releasing.

[...]

PR #6254, opened 2022-02-07
cargo: enable ‘-D warnings’ by default

mina86:
Making ‘-D warnings’ the default may be a minor annoyence to developers
who now will need to fix warnings even while they work on the feature
(as opposed to only having to do that once they are ready to send the
code for review) but it does simplify the code a small amount.

Issue: https://github.com/near/nearcore/issues/6226

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -2,7 +2,7 @@ steps:
   - label: "cargo test integration-tests*"
     command: |
       source ~/.cargo/env && set -eux
-      RUST_BACKTRACE=1 RUSTFLAGS='-D warnings' cargo test --locked -p 'integration-tests'
+      RUST_BACKTRACE=1 cargo test --locked -p 'integration-tests'
 
     timeout: 60
     agents:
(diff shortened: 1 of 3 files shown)


bowenwang1996 (commented):
Is there a reason for removing the check from CI? I suppose we should still check for warnings in CI to make sure they are fixed

mina86:
With the flag being set in `.cargo/config.toml` setting RUSTFLAGS is now redundant.

matklad:
I don't think we should do this. To me, it seems that while the annoyance *can* be minor, it can also be very significant. 

For example, I often do quick&dirty exploratory work (building the code with two versions of some library, etc) where I have a ton of warnings. 

Even if it were true that the annoyance is minor, i'd argue that fixing a minor, repeated annoyance for everyone is well worth to spend a couple of extra CI lines on. 

There's also a point that we are setting `RUST_BACKTRACE=1` anyway, so it's not like we make anything *meaningfully* simpler, we still have the repeated "set up common env" step. It seems to be that the actual problem here is that we can't set env-variable once, like we can do with github actions: 

https://github.com/rust-analyzer/rust-analyzer/blob/9b1978a3ed405c2a5ec34703914ec1878b599e14/.github/workflows/ci.yaml#L10-L16

mina86:
> It seems to be that the actual problem here is that we can't set env-variable once, like we can do with github actions:

Actually it can: https://github.com/near/nearcore/pull/6269

Issue #6250, opened 2022-02-06
Archival nodes do not need to store trie changes for old blocks


bowenwang1996:
`ColTrieChanges` accounts for about 2TB of archival node's storage on mainnet. However, we can reconstruct them easily if we have prev state, transactions, and receipts for a give chunk. To save storage space, we can compute them on the fly if needed.

abacabadabacaba:
In fact, I think that trie changes can be reconstructed more efficiently by diffing the tries. No need to actually reexecute anything. So, I think that we should eventually get rid of ColTrieChanges completely, for all nodes.

frol:
@abacabadabacaba diffs won't give you the reason, e.g. why the balance has changed.

marcelo-gonzalez:
@mzhangmzz @bowenwang1996 @mina86  Am i correct here in saying that getting rid of this is actually not that complicated for archival nodes?

So looks to me like the only place `ColTrieChanges` is accessed after storing to it is when clearing old data in `clear_block_data()`. This in turn is only called from `clear_data()` (and `clear_forks_data()`, which is called by `clear_data()`) and `reset_data_pre_state_sync()`:

https://github.com/near/nearcore/blob/a83db115e4ef01152bbe2f89cf11b0f3da32006d/chain/client/src/client.rs#L1050-L1055

https://github.com/near/nearcore/blob/a83db115e4ef01152bbe2f89cf11b0f3da32006d/chain/client/src/client_actor.rs#L1504-L1506

so in both cases only if archival=false. So as a quick solution before maybe just getting rid of it completely like @abacabadabacaba suggested, should be possible to just get rid of the whole column now and avoid storing to it in the future for archival nodes

mina86:
I think you’re actually right. The only two reads I could find are in `clear_block_data` which is not called in archival nodes.

And the changes is just a trie key→value list so if we ever need that information diffing two tries seem like obvious approach.

bowenwang1996:
@marcelo-gonzalez yeah please do that (removing the whole column)

mina86:
FYI, in https://github.com/near/nearcore/pull/6477 I’m adding an option to recompress-storage to clean out that column.

PR #6255, opened 2022-02-07
refactor(indexer): Move `state_changes` from `StreamerMessage` root to `IndexerShard` instead

khorolets:
* Extract `indexer::streamer::types` to separate crate `indexer-primitives` for publishing
* Add `GetStateChangesWithCauseInBlockForTrackedShards` to `client-primitives` and corresponding `view_client` handler to return `StateChangesWithCause` for each shard separately
* Breaking changes in `StreamerMessage` struct, there is no `state_changes` in the root anymore


chain/client/src/view_client.rs
```diff
@@ -758,6 +758,58 @@ impl Handler<GetStateChangesWithCauseInBlock> for ViewClientActor {
...
+                StateChangeValue::ContractCodeUpdate { account_id, .. } => account_id.clone(),
+                StateChangeValue::ContractCodeDeletion { account_id } => account_id.clone(),
+            };
+            let shard_id = if let Ok(shard_id) =
+                self.runtime_adapter.account_id_to_shard_id(&account_id, &msg.epoch_id)
+            {
+                shard_id
+            } else {
+                return Err(GetStateChangesError::IOError {
+                    error_message: format!("Failed to get ShardID from AccountID {}", account_id),
```
frol, line 794: According to the `account_id_to_shard_id` implementation, the error could happen only when we cannot get shard layout (either due to some internal error or when the epoch never happened), so while it is more or less fair to abuse IOError, I would at least make the error message more helpful to both future developers and those who might potentially observe the error message. Thus, I suggest we format Err into the error message.

chain/client/src/view_client.rs
```diff
@@ -758,6 +758,58 @@ impl Handler<GetStateChangesWithCauseInBlock> for ViewClientActor {
...
+
+            let mut state_changes = if let Some(changes_with_cause) =
+                state_changes_with_cause_split_by_shard_id.remove(&shard_id)
+            {
+                changes_with_cause
+            } else {
+                Vec::<StateChangeWithCauseView>::new()
+            };
+            state_changes.push(state_change_with_cause.into());
+            state_changes_with_cause_split_by_shard_id.insert(shard_id, state_changes);
```
frol, line 806: You have spent too much time with near-sdk-rs 😄 

You don't need to re-add a record inside a hashmap when you only want to mutate the internal object (a vector). Use [`entry(&shard_id).or_default()`](https://doc.rust-lang.org/stable/std/collections/struct.HashMap.html#method.entry) to get a mutable `state_changes_in_shard` vector, which you can push into.

chain/indexer-primitives/Cargo.toml
```diff
@@ -0,0 +1,16 @@
...
+rust-version = "1.56.0"
+edition = "2021"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+serde = { version = "1", features = [ "derive" ] }
+serde_json = "1.0.55"
+
+near-primitives = { path = "../../core/primitives" }
```
frol, line 19: @miraclx Could you review the newly added crate config, please? (We plan to get it published to crates.io)

chain/indexer/CHANGELOG.md
```diff
@@ -1,56 +1,68 @@
...
+## Breaking changes
+
+The field `state_changes` is moved from the root of `StreamerMessage`
+to the `IndexerShard.state_changes` and now contains only changes related
+to the specific shard.
+
 ## 0.10.1
 
-* (mainnet only) Add additional handler to inject restored receipts to the block #47317863. See [PR 4248](https://github.com/near/nearcore/pull/4248) for reference
+- (mainnet only) Add additional handler to inject restored receipts to the block #47317863. See [PR 4248](https://github.com/near/nearcore/pull/4248) for reference
```
frol, line 17: I suggest we rollback the changes regarding the use of `*` vs `-` and other formatting changes unless there is some contribution guide that defined this rule.

chain/indexer-primitives/Cargo.toml
```diff
@@ -0,0 +1,16 @@
+[package]
+name = "indexer-primitives"
```
frol, line 2: Following the pattern, the crate should be still named with `near-` prefix.

chain/indexer/Cargo.toml
```diff
@@ -18,6 +18,7 @@ serde = { version = "1", features = [ "derive" ] }
 serde_json = "1.0.55"
 tokio = { version = "1.1", features = ["time", "sync"] }
 
+indexer-primitives = { path = "../indexer-primitives" }
```
frol, line 21:
```suggestion
near-indexer-primitives = { path = "../indexer-primitives" }
```

and move it down the list maintaining the alphabetical order if rustfmt does not do it automatically.

chain/indexer/src/streamer/mod.rs
```diff
@@ -17,18 +17,17 @@ use self::fetchers::{
     fetch_block_by_hash, fetch_block_by_height, fetch_block_chunks, fetch_latest_block,
     fetch_outcomes, fetch_state_changes, fetch_status,
 };
-pub use self::types::{
+use self::utils::convert_transactions_sir_into_local_receipts;
+use crate::streamer::fetchers::fetch_protocol_config;
+use crate::INDEXER;
+pub use indexer_primitives::{
```
frol, line 23: Please, fix the order of the imports 🙏

chain/indexer/src/streamer/mod.rs
```diff
@@ -77,14 +76,31 @@ async fn build_streamer_message(
...
+        near_primitives::types::EpochId(block.header.epoch_id.clone()),
+    )
+    .await?;
     let mut indexer_shards = (0..num_shards)
-        .map(|shard_id| IndexerShard { shard_id, chunk: None, receipt_execution_outcomes: vec![] })
+        .map(|shard_id| IndexerShard {
+            shard_id,
+            chunk: None,
+            receipt_execution_outcomes: vec![],
+            state_changes: vec![],
```
frol, line 90: Extract the state_changes for the appropriate shard_id here immediately, since given that execution outcomes could happen when chunk was not produced, state changes will also be there outside of the chunk, so iteration over chunks won't be enough, and here is the right place to have the shard state changes populated.

chain/indexer/src/streamer/mod.rs
```diff
@@ -77,14 +76,31 @@ async fn build_streamer_message(
...
 
     for chunk in chunks {
         let views::ChunkView { transactions, author, header, receipts: chunk_non_local_receipts } =
             chunk;
 
+        let shard_id = header.shard_id.clone() as usize;
+
+        indexer_shards[shard_id].state_changes = state_changes
+            .remove(&header.shard_id)
+            .expect("StateChanges for given shard should be present");
```
frol, line 102: See above

chain/indexer/src/streamer/utils.rs
```diff
@@ -5,7 +5,7 @@ use node_runtime::config::tx_cost;
 
 use super::errors::FailedToFetchData;
 use super::fetchers::fetch_block_by_hash;
-use super::IndexerTransactionWithOutcome;
+use indexer_primitives::IndexerTransactionWithOutcome;
```
frol, line 8: order 🙏

frol (changes requested):

chain/client/src/view_client.rs
```diff
@@ -758,6 +758,58 @@ impl Handler<GetStateChangesWithCauseInBlock> for ViewClientActor {
...
+            let account_id = match &state_change_with_cause.value {
+                StateChangeValue::AccountUpdate { account_id, .. } => account_id.clone(),
+                StateChangeValue::AccountDeletion { account_id } => account_id.clone(),
+                StateChangeValue::AccessKeyUpdate { account_id, .. } => account_id.clone(),
+                StateChangeValue::AccessKeyDeletion { account_id, .. } => account_id.clone(),
+                StateChangeValue::DataUpdate { account_id, .. } => account_id.clone(),
+                StateChangeValue::DataDeletion { account_id, .. } => account_id.clone(),
+                StateChangeValue::ContractCodeUpdate { account_id, .. } => account_id.clone(),
+                StateChangeValue::ContractCodeDeletion { account_id } => account_id.clone(),
+            };
```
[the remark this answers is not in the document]
frol: BTW, it might be beneficial to extract getting affected account id into a dedicated method on `StateChangeValue` type.

chain/indexer-primitives/Cargo.toml
miraclx: Looks good to me. 👍🏽

miraclx (approved):

chain/indexer/CHANGELOG.md
khorolets: It was an imposter - VSCode :( I didn't notice it. Thanks!

chain/indexer/CHANGELOG.md
```diff
@@ -1,5 +1,17 @@
 # Changelog
 
+## 1.25.x (UNRELEASED)
+
+* `state_changes` is moved to `IndexerShard` struct and represent only
+  the state changes happened on the specific shard
+* All the NEAR Indexer Framework types was extracted to a separate crate `indexer-primitives`
```
frol, line 7:
```suggestion
* All the NEAR Indexer Framework types were extracted to a separate crate `near-indexer-primitives`
```

chain/indexer/CHANGELOG.md
```diff
@@ -1,5 +1,17 @@
 # Changelog
 
+## 1.25.x (UNRELEASED)
+
+* `state_changes` is moved to `IndexerShard` struct and represent only
+  the state changes happened on the specific shard
```
frol, line 6:
```suggestion
* `state_changes` field is moved from the top-level `StreamerMessage` to `IndexerShard` struct to align better with the sharded nature of NEAR protocol. In the future, when nearcore will be able to track only a subset of shards, this API will work naturally, so we take pro-active measures to solidify the APIs
```

chain/indexer/src/streamer/fetchers.rs
```diff
@@ -6,11 +6,11 @@ use actix::Addr;
 use futures::stream::StreamExt;
 use tracing::warn;
 
+use near_indexer_primitives::IndexerExecutionOutcomeWithOptionalReceipt;
 pub use near_primitives::hash::CryptoHash;
 pub use near_primitives::{types, views};
```
frol, line 11: Off-topic; we don't need to solve it now, yet I want to bring this to attention. Do we actually need them to be `pub`? It caught my eye that this API surface is a bit odd.

chain/client/src/view_client.rs
```diff
@@ -758,6 +758,40 @@ impl Handler<GetStateChangesWithCauseInBlock> for ViewClientActor {
...
+        _: &mut Self::Context,
+    ) -> Self::Result {
+        let state_changes_with_cause_in_block =
+            self.chain.store().get_state_changes_with_cause_in_block(&msg.block_hash)?;
+
+        let mut state_changes_with_cause_split_by_shard_id: HashMap<ShardId, StateChangesView> =
+            HashMap::new();
+        for state_change_with_cause in state_changes_with_cause_in_block {
+            let account_id = &state_change_with_cause.value.get_affected_account_id();
+            let shard_id = match self.runtime_adapter.account_id_to_shard_id(&account_id, &msg.epoch_id) {
```
frol, line 779: I would expect this to work just fine:

```suggestion
            let account_id = state_change_with_cause.value.get_affected_account_id();
            let shard_id = match self.runtime_adapter.account_id_to_shard_id(account_id, &msg.epoch_id) {
```

frol (approved):
Looks good to me.

P.S. Run fmt to fix CI

chain/indexer-primitives/Cargo.toml
```diff
@@ -0,0 +1,19 @@
+[package]
+name = "near-indexer-primitives"
```
matklad, line 2: Could you add a readme explaining why this crates needs to exist, what should belong to this crate, and what shouldn't belong to this crate? "this crates holds types" doesn't answer those questions for me -- we might as well keep them in indexer, and we also might add some logic here, why not? 

It *seems* to me that what actually happens is that "this crate defines over-the-wire schema for JSON messages used by indexer HTTP API". Ie, it seems that this is not about Rust, but rather about a network protocol/API we expose. That is, if true, a very important bit of information, and it currently is missing from the description. 


:thinking: wait, I think the above is not true. I've just tried removing `Serialize, Deserialize` from all intexer messages, and that seems to work -- nothing in *this* repository seems to rely on serialization. I guess, I am just confused as to why this particular bunch of types needs special treatment and is worth a separate crate :)

matklad:
unrelated, but `s/IndexerChunkView/IndexerChunk` might make naming a bit more consistent.

chain/indexer-primitives/Cargo.toml
khorolets: @matklad I'll add a little README. 

We are making [NEAR Lake](https://github.com/near/near-lake) - in a few words it's an indexer that is going to save all the data from the network to AWS S3 in JSON files. We are starting some NEAR Lake Framework which is going to read from AWS S3 and stream all the data. Both of the mentioned projects will use indexer primitives, in nearest future, there are going to be even more projects. 

The first goal is to make it possible that projects like NEAR Lake Framework and the ones that use it as a dependency won't require an entire `nearcore` in dependencies in order to actually use only those primitives.

And we are going to extend the number of types in indexer primitives and adjust the serde_json parameters to improve the way the types are serialized.

I hope that makes sense. Let me know if you have questions regarding this extraction to a separate crate :)

matklad: Ah, this makes sense! And yeah, I want to stress that the salient point of this thing then is *not* the Rust types, but rather the JSON schema, and all of the related backwards compatibility concerns.

And yeah, this totally needs to be a separate crate. I wonder though if `-types` or `-views` would be a better public name than `-primitives`. In today's usage `primitives` is essentially a kitchen sink, and it defines much more than just data schema. 

I also wonder if, long term, this should live in some other repository. It feels odd that this defines a data schema, but the actual data isn't used by the repo (so, it's easy for nearcore developers to break this stuff without realizing this). Contrast this with `views`, which also do define the schema, but this schema is than used by the code in this repository (the node's RPC API). It seems nicer if `indexer-framework` defines just the Rust API (without Serialize), and let's the consumers define their own data schema. 

:thinking: yeah, I guess long-term I can see two worlds:

* In one world, "indexer schema" is something defined completely outside of nearcore as a part of the lake framework. In this world, there's a separate `indexer-types` crate, but it lives in the repo which implements the actual indexer. We can also imagine two different lakes by two different providers with different schema
* In another world, we say that "indexer schema"  is something universal, and is as much a part of nearcore as JSON RPC API. In this world, I would expect this to be a *module* in the crate that defines JSON RPC API (so, `near_primitives::views::indexer` today). In this world, while there might be alternative implementations of the lake, we'd expect them to use the same schema.

But yeah, this is long-term, starting with moving this to a separate crate might make sense (but also do consider just creating a views submodule)

chain/indexer/src/streamer/fetchers.rs
khorolets: I don't remember the reason so I've dropped `pub` from the places I believe it's not necessary

chain/indexer-primitives/Cargo.toml
frol: @matklad TL;DR: There are several things at play, and you properly identified the long-term plans we actually have.

The types defined in this crate are Rust types for Indexer Framework, which is now tightly coupled with nearcore. You don't see any use of these APIs in nearcore repo because it is designed to be a library (you can find the usage in indexer-for-* projects out there), and we want to stabilize this API in the long term in terms of Rust types and while we already have our own top-level structure, there are still quite some types used in depth (e.g. Transaction and ExecutionOutcome types). Given there is an initiative to decouple JSON RPC types #5516, we decided to postpone solving all the problems.

Second, indeed, we want to be able to serialize the types for near-lake and deserialize it on the reader side without pulling the whole nearcore (we could have get away with pulling the whole nearcore, but due to #3803 it makes things a bit harder on our side, since we often receive questions about M1 support in our tooling and we use M1 ourselves)

Third, there is a plan to further solidify the APIs and extract near-indexer into a separate repo, but to make that happen, nearcore should be published as a crate, and its APIs also stable.

matklad (approved):

core/primitives/src/types.rs
```diff
@@ -233,6 +233,21 @@ pub enum StateChangeValue {
...
+            StateChangeValue::AccountUpdate { account_id, .. } => account_id,
+            StateChangeValue::AccountDeletion { account_id } => account_id,
+            StateChangeValue::AccessKeyUpdate { account_id, .. } => account_id,
+            StateChangeValue::AccessKeyDeletion { account_id, .. } => account_id,
+            StateChangeValue::DataUpdate { account_id, .. } => account_id,
+            StateChangeValue::DataDeletion { account_id, .. } => account_id,
+            StateChangeValue::ContractCodeUpdate { account_id, .. } => account_id,
+            StateChangeValue::ContractCodeDeletion { account_id } => account_id,
+        }
+    }
```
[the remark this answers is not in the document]
frol: > If the thing is called account_id in all of the variants, might as well re-use the name.

Well, I would argue that `account_id` name in those struct variants is not specific enough, and following bad naming does not necessarily a good thing. Having a more descriptive name is actually net positive.

chain/indexer-primitives/src/lib.rs
```diff
@@ -8,7 +8,6 @@ pub use near_primitives::{types, views};
 pub struct StreamerMessage {
     pub block: views::BlockView,
     pub shards: Vec<IndexerShard>,
```
mm-near, line 10: do you have to worry about backwards compatibility ? is this message used in communication between two different binaries ?

chain/indexer/src/streamer/mod.rs
```diff
@@ -77,14 +76,29 @@ async fn build_streamer_message(
...
+    .await?;
     let mut indexer_shards = (0..num_shards)
-        .map(|shard_id| IndexerShard { shard_id, chunk: None, receipt_execution_outcomes: vec![] })
+        .map(|shard_id| IndexerShard {
+            shard_id,
+            chunk: None,
+            receipt_execution_outcomes: vec![],
+            state_changes: state_changes
+                .remove(&shard_id)
+                .expect("StateChanges for given shard should be present"),
```
mm-near, line 92: But some shards (chunks) might be missing in a block, right ?

chain/indexer-primitives/src/lib.rs
```diff
@@ -42,4 +41,5 @@ pub struct IndexerShard {
     pub shard_id: types::ShardId,
     pub chunk: Option<IndexerChunkView>,
     pub receipt_execution_outcomes: Vec<IndexerExecutionOutcomeWithReceipt>,
+    pub state_changes: views::StateChangesView,
```
mm-near, line 44: do we have any tests for this ?

mm-near (changes requested):

chain/indexer-primitives/src/lib.rs
khorolets: We had a conversation re: indexer and tests with @matklad recently

Here's the start https://github.com/near/nearcore/pull/6195#issuecomment-1023053576

chain/indexer/src/streamer/mod.rs
khorolets: Oh, really, my bad, thanks!

chain/indexer-primitives/src/lib.rs
khorolets: A little bit above in this PR we have talked about it, let me know if that doesn't answer your question

https://github.com/near/nearcore/pull/6255#discussion_r801527633

bowenwang1996 (commented):
This is fine as is, but I'd like to note the following:
* We probably should change the underlying storage to store state changes by shard.
* @frol @khorolets we plan to stop storing state changes in the database in the future https://github.com/near/nearcore/issues/6250. It shouldn't affect how this PR works, but just want to give you a heads up.

khorolets:
> This is fine as is, but I'd like to note the following:
> 
> * We probably should change the underlying storage to store state changes by shard.
> * @frol @khorolets we plan to stop storing state changes in the database in the future #6250. It shouldn't affect how this PR works, but just want to give you a heads up.

Thanks for the heads up @bowenwang1996! In #6250 you're saying it should be possible to retrieve StateChanges on the fly, so it means we would be able to return StateChanges from the NEAR Indexer side. 

cc @frol

chain/indexer/src/lib.rs
```diff
@@ -113,8 +113,8 @@ impl Indexer {
     }
 
     /// Boots up `near_indexer::streamer`, so it monitors the new blocks with chunks, transactions, receipts, and execution outcomes inside. The returned stream handler should be drained and handled on the user side.
-    pub fn streamer(&self) -> mpsc::Receiver<streamer::StreamerMessage> {
-        let (sender, receiver) = mpsc::channel(16);
+    pub fn streamer(&self) -> mpsc::Receiver<StreamerMessage> {
+        let (sender, receiver) = mpsc::channel(100);
```
mm-near, line 117: why was it changed to 100? can you add a comment ?

chain/indexer-primitives/src/lib.rs
mm-near: so - no tests ? :-(

mm-near (approved):

chain/indexer/src/lib.rs
khorolets: In the PR description 

> Increase the streaming channel size from 16 to 100 (reindexing speed)

rtsai123 (approved):

Issue #2454, opened 2020-04-13
Improve UX of the neard status line (peers count, etc)


frol:
**Motivation**

Looking at the current status line reported by neard every 10 seconds:

```
Apr 13 19:46:10.190  INFO stats: # 2135246 Downloading headers 86% -/4  7/7/40 peers ⬇ 21.9kiB/s ⬆ 3.1kiB/s 0.00 bps 0 gas/s CPU: 18%, Mem: 36.3 MiB
```

As a user, I have a hard time understanding some of the values:

* `-/4` (I know that instead of the "-" there might be "F" [fisherman] or "V" [validator], and while those are also not intuitive, "-" is completely misleading when you look at it in the context of `-/4` -- I read it as "nothing (or unknown) out of four")
* `7/7/40 peers` (7/7 are number of peers and number of peers with the latest chain)

What it is even worse is that those values are impossible to search in the docs (at least I do not have a clue where to even start). UPD: https://docs.nearprotocol.com/docs/validator/staking

**Version (please complete the following information):**
- nearcore commit/branch: 0.4.13 build 650b6986-modified
- docker: 19.03.8-ce
- betanet

nikurt:
Can confirm that status lines are confusing, experienced that while experimenting with running a node without any prior knowledge.

bowenwang1996:
> Why does a node periodically print a transaction hash? How is the transaction hash useful?

It is the hash of some block, not of some transaction.

> What is 7/7/40 peers?

The first number means number of connected peers. The second number means the number of peers that have highest height close to the node. The third number means the maximum number of peers.

bowenwang1996:
Assigning to @janewang to figure out how to improve the UX

janewang:
Suggested UX
`-/4  7/7/40 peers =>  - Fisherman | 4 Validator  7 Connected | 7 Highest Height  | 40 Max`

ilblackdragon:
I would drop completely "highest height" part and "fisherman" given those were designed for different chain algorithm. Also max peers is not that useful as it doesn't really change. Probably better to report that at the launch of the nearcore with some other configuration information?

janewang:
Thanks, `-/4  7/7/40 peers` => `4 validators   7 connected peers` The 40 max peers is defined in the config.json.

janewang:
`-/4  7/7/40 peers` This should be changed to  `Validator  | 4 Validators  | 7 Connected Peers`
[...]

PR #6260, opened 2022-02-07
chain: make status line more descriptive

mina86:
Remove some overly verbose information from status line and at the
same time make the status more descriptive.

Before: `-/4 7/7/40 peers`
 After: `4 validators 7 peers`

Information that this commit removes:
* the ‘-’ indicating whether node was validator or not,
* second ‘7’ indicating the number of peers that have highest height
  close to the node and
* ‘40’ indicating maximum number of peers.

Fixes: https://github.com/near/nearcore/issues/2454

### DIFF
--- a/chain/client/src/info.rs
+++ b/chain/client/src/info.rs
@@ -98,14 +98,12 @@ impl InfoHelper {
         let sync_status_log = Some(display_sync_status(sync_status, head, genesis_height));
 
         let validator_info_log = validator_info.as_ref().map(|info| {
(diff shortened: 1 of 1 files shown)


chain/client/src/info.rs
```diff
@@ -95,17 +95,18 @@ impl InfoHelper {
             Some(text) => ansi_term::Style::default().paint(text),
         };
 
+        let s = |num| if num == 1 { "" } else { "s" };
```
bowenwang1996, line 98: Nice :)

bowenwang1996 (approved):

PR #6268, opened 2022-02-08
feat: add metrics about chunk fullness

Ekleog:
The metrics added look like this on a localnet node without transactions:
```
near_chunk_tgas_used_hist_bucket{shard="0",le="50"} 12
near_chunk_tgas_used_hist_bucket{shard="0",le="100"} 12
near_chunk_tgas_used_hist_bucket{shard="0",le="300"} 12
near_chunk_tgas_used_hist_bucket{shard="0",le="500"} 12
near_chunk_tgas_used_hist_bucket{shard="0",le="700"} 12
near_chunk_tgas_used_hist_bucket{shard="0",le="800"} 12
near_chunk_tgas_used_hist_bucket{shard="0",le="900"} 12
near_chunk_tgas_used_hist_bucket{shard="0",le="950"} 12
near_chunk_tgas_used_hist_bucket{shard="0",le="1000"} 12
near_chunk_tgas_used_hist_bucket{shard="0",le="1050"} 12
near_chunk_tgas_used_hist_bucket{shard="0",le="1100"} 12
near_chunk_tgas_used_hist_bucket{shard="0",le="1150"} 12
near_chunk_tgas_used_hist_bucket{shard="0",le="1200"} 12
near_chunk_tgas_used_hist_bucket{shard="0",le="1250"} 12
near_chunk_tgas_used_hist_bucket{shard="0",le="1300"} 12
near_chunk_tgas_used_hist_bucket{shard="0",le="+Inf"} 12
near_chunk_tgas_used_hist_sum{shard="0"} 0
near_chunk_tgas_used_hist_count{shard="0"} 12
```

Grafana is able to display that as a heatmap: https://grafana.com/docs/grafana/latest/visualizations/heatmap/

### DIFF
--- a/chain/client/src/client_actor.rs
+++ b/chain/client/src/client_actor.rs
@@ -983,6 +983,9 @@ impl ClientActor {
 
             let last_final_hash = *block.header().last_final_block();
 
+            for chunk in block.chunks().iter() {
+                self.info_helper.chunk_processed(chunk.shard_id(), chunk.gas_used());
+            }
             self.info_helper.block_processed(gas_used, chunks_in_block as u64);
             self.check_send_announce_account(last_final_hash);
         }
--- a/chain/client/src/client_actor.rs
+++ b/chain/client/src/client_actor.rs
@@ -983,9 +983,16 @@ impl ClientActor {
 
             let last_final_hash = *block.header().last_final_block();
 
-            for chunk in block.chunks().iter() {
+            let chunks = block.chunks();
+            let included_chunks = chunks
+                .iter()
+                .zip(block.header().chunk_mask().iter())
+                .filter(|(_, mask)| **mask)
+                .map(|(chunk, _)| chunk);
+            for chunk in included_chunks {
                 self.info_helper.chunk_processed(chunk.shard_id(), chunk.gas_used());
             }
+
             self.info_helper.block_processed(gas_used, chunks_in_block as u64);
             self.check_send_announce_account(last_final_hash);
         }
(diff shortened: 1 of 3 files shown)


chain/client/src/client_actor.rs
```diff
@@ -983,6 +983,9 @@ impl ClientActor {
 
             let last_final_hash = *block.header().last_final_block();
 
+            for chunk in block.chunks().iter() {
```
nikurt, line 986: Please check if the chunk is included.

chain/client/src/info.rs
```diff
@@ -71,6 +71,13 @@ impl InfoHelper {
         }
     }
 
+    pub fn chunk_processed(&mut self, shard: ShardId, gas_used: Gas) {
+        let teragas = 1_000_000_000_000f64;
```
nikurt, line 75: I suggest to use `chunk.gas_limit()` instead of hardcoding the constant

Ekleog: To the best of my knowledge, the less we pre-compute stuff before putting it in prometheus the better, because it's the way the ecosystem is designed.

So IMO it would be better to have the metrics expressed in Tgas (to be vaguely human-readable without any loss of information), and then have a `chunk_gas_limit` metric (gauge) that we could then use to divide in grafana.

That said to avoid hardcoding one more constant, I just hoisted the previously-existing teragas constant to the top of the file, hopefully that looks ok? :)

chain/client/src/client_actor.rs
Ekleog: Hopefully done, if I understood correctly how chunk masks work

nikurt (approved):

bowenwang1996 (approved):

PR #6269, opened 2022-02-08
buildkite: set common environment variables in `env` section

mina86:


.buildkite/pipeline.yml
```diff
@@ -1,8 +1,12 @@
+env:
+  RUST_BACKTRACE: short
+  RUSTFLAGS: -D warnings -W unreachable-pub -W rust-2021-compatibility
```
matklad, line 3: `-W unreachable-pub` will probably break every crate of ours. I do think it's a generally good thing to have, but I am not sure if it's good enough to actually enable for our use-case. But this should be a great way to test if this actually works.

matklad: If we want to poach from rust-analyzer, than `CARGO_INCREMENTAL=0` *might* be interesting avenue to explore to speed up CI: https://matklad.github.io/2021/09/04/fast-rust-builds.html#ci-caching. Needs substantial investigation though

matklad (approved):
:heart: 

This have been bothering me for ages, would be great if it works (I didn't find this in buildkite's docs)

matklad:
Hey, it does work, :tada: 

```
Compiling proc-macro-crate v0.1.5
--
  | error: unreachable `pub` field
  | --> tools/rpctypegen/macro/src/lib.rs:16:5
  | \|
  | 16 \|     pub schema: BTreeMap<String, ErrorType>,
  | \|     ---^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

```

https://buildkite.com/nearprotocol/nearcore/builds/13184#6ce34762-ad12-491a-b85e-4d5b151d406f/94-443

mhalambek (approved):

PR #6273, opened 2022-02-08
chore: reformat Cargo.toml lists with many elements in one line

marcelo-gonzalez:
this will make them easier to edit and the git diffs will be nicer.
Done with a python script that basically says:

  if len(list) > 1:
     reformat(list)

Issue: https://github.com/near/nearcore/issues/5328

### DIFF
--- a/chain/chain/Cargo.toml
+++ b/chain/chain/Cargo.toml
@@ -45,6 +45,10 @@ no_cache = ["near-store/no_cache"]
 protocol_feature_chunk_only_producers = ["near-chain-configs/protocol_feature_chunk_only_producers", "near-primitives/protocol_feature_chunk_only_producers"]
 
 protocol_feature_routing_exchange_algorithm = []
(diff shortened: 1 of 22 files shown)


mina86 (approved):
I think we want the lines reformatted even if there are only two elements.  The point is to have cleaner diffs when elements are added or removed.  Having list on a single line when there’s only one element is a reasonable compromise for the files not ending up too verbose.

khorolets (approved):

marcelo-gonzalez:
> I think we want the lines reformatted even if there are only two elements. The point is to have cleaner diffs when elements are added or removed. Having list on a single line when there’s only one element is a reasonable compromise for the files not ending up too verbose.

Fair enough. I initially chose > 2 because it seemed fine to leave stuff like:

```
default = ["borsh", "serde"]
```

but I'll update it shortly, maybe manually omitting some like that one

bowenwang1996 (approved):
Thanks!

pmnoxx (approved):
lgtm

matklad (approved):

nikurt (approved):

PR #6276, opened 2022-02-09
Export ED25519SecretKey

noch3:
The type `ED25519SecretKey` appears in multiple public APIs for `SecretKey` but it is not exported and therefore can only be used opaquely. Notably, there is no way to construct one apart from doing tricks with `FromStr`.

### DIFF
--- a/core/crypto/src/lib.rs
+++ b/core/crypto/src/lib.rs
@@ -1,8 +1,8 @@
 pub use errors::{ParseKeyError, ParseKeyTypeError, ParseSignatureError};
 pub use key_file::KeyFile;
 pub use signature::{
-    ED25519PublicKey, KeyType, PublicKey, Secp256K1PublicKey, Secp256K1Signature, SecretKey,
-    Signature,
+    ED25519PublicKey, ED25519SecretKey, KeyType, PublicKey, Secp256K1PublicKey, Secp256K1Signature,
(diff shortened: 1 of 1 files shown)


noch3:
I can imagine an additional improvement would be to make `ED25519SecretKey` a wrapper around `bip25519_dalek::Keypair` (instead of the binary encoding of that type). Then constructing one is much less error-prone. If you like this idea, now is the time to do it since that change would still be backwards compatible at this point.

matklad:
This makes sense to me locally, but, globally, I don't think we intende `near-crypto` to have any kind of public API. @noch3 could you give some context for *why* you need to access `near-crypto` in the first place? Perhaps something *else* should wrap its APIs into an actually usable form.

matklad:
We publish nearcore crates to crates.io, but we generally don't make any promises that API are fit for use outside of nearcore. This is essentially an automated publishing process a-la rustc ap crates (https://crates.io/crates/rustc-ap_rustc_ast). You *can* use those published crates, but the API would break pretty frequently. 

l am not sure what's the officially supported way to get this functionality though, let me ping core-tools folks.

noch3:
Oh I see, so backwards compatibility isn’t a strong concern. Fine with me

matklad (approved):

PR #6278, opened 2022-02-10
feat: Differentiate code and data bytes

jakmeier:
The new deployment cost estimation now also treats data bytes and code
bytes differently  in the fromulas.
The new cost `DeployBytes` is there to assert the assumption that
compilation cost is the dominant factor.

### DIFF
--- a/runtime/runtime-params-estimator/src/cost.rs
+++ b/runtime/runtime-params-estimator/src/cost.rs
@@ -243,10 +243,11 @@ pub enum Cost {
     ContractCompileBytesV2,
     /// The cost of contract deployment per byte, without the compilation cost.
     ///
-    /// Estimation: Measure the cost of two contract that have the same code but
-    /// include different data sections. One data section is significantly
-    /// larger. The cost difference is pure overhead of moving around bytes that
-    /// are not code. Divide this cost by the difference of bytes.
+    /// Estimation: Measure the deployment costs of two data-only contracts,
+    /// where one data sections is empty and the other is max size as allowed by
+    /// contract size limits. The cost difference is pure overhead of moving
+    /// around bytes that are not code. Divide this cost by the difference of
+    /// bytes.
     DeployBytes,
     GasMeteringBase,
     GasMeteringOp,
--- a/runtime/runtime-params-estimator/src/cost.rs
+++ b/runtime/runtime-params-estimator/src/cost.rs
@@ -241,6 +241,13 @@ pub enum Cost {
     /// produces the steepest line.
     ContractCompileBaseV2,
     ContractCompileBytesV2,
+    /// The cost of contract deployment per byte, without the compilation cost.
+    ///
+    /// Estimation: Measure the cost of two contract that have the same code but
+    /// include different data sections. One data section is significantly
+    /// larger. The cost difference is pure overhead of moving around bytes that
+    /// are not code. Divide this cost by the difference of bytes.
+    DeployBytes,
     GasMeteringBase,
     GasMeteringOp,
     /// Cost of inserting a new value directly into a RocksDB instance.
--- a/runtime/runtime-params-estimator/src/lib.rs
+++ b/runtime/runtime-params-estimator/src/lib.rs
@@ -627,7 +627,6 @@ fn pure_deploy_bytes(ctx: &mut EstimatorContext) -> GasCost {
     let cost_4mb = deploy_contract_cost(ctx, large_code, Some(b"main"));
 
     if cost_4mb < cost_empty {
-        eprintln!("High variance in pure_deploy_bytes: Deployment cost of small contract is higher than large contract.");
         let mut cost = GasCost::zero(ctx.config.metric);
         cost.set_uncertain(true);
         cost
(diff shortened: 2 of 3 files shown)


runtime/runtime-params-estimator/src/lib.rs
```diff
@@ -613,6 +617,34 @@ fn contract_compile_base_per_byte_v2(ctx: &mut EstimatorContext) -> (GasCost, Ga
...
+        "test-contract/res/nightly_small_contract.wasm"
+    } else {
+        "test-contract/res/stable_small_contract.wasm"
+    });
+    // 1MiB
+    let large_code = read_resource(if cfg!(feature = "nightly_protocol_features") {
+        "test-contract/res/nightly_large_contract.wasm"
+    } else {
+        "test-contract/res/stable_large_contract.wasm"
+    });
```
matklad, line 632: I think for this cost it doesn't really matter which one we use, so we can always pick `stable_large_contract.wasm`?

Alternatively, perhaps it'd be even better to synthesize empty wasm "by hand" here.

jakmeier: Ahh yeah this is indeed unnecessary to differentiate nightly/stable.

But what is the benefit of using a synthesized wasm?

matklad: >But what is the benefit of using a synthesized wasm?

It'd be simpler -- `stable_large_contract.wsm` is a complex beast with all kinds of stuff inside. While it seems that what we want here is `(module (data "bla-bla-bla-bla-bla..."))`

runtime/runtime-params-estimator/src/cost.rs
```diff
@@ -241,6 +241,13 @@ pub enum Cost {
     /// produces the steepest line.
     ContractCompileBaseV2,
     ContractCompileBytesV2,
+    /// The cost of contract deployment per byte, without the compilation cost.
+    ///
+    /// Estimation: Measure the cost of two contract that have the same code but
+    /// include different data sections. One data section is significantly
+    /// larger. The cost difference is pure overhead of moving around bytes that
+    /// are not code. Divide this cost by the difference of bytes.
+    DeployBytes,
```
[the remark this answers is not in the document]
jakmeier: Ok, yeah I can create such a contract and it should be more accurate.

Second observations makes sense to me. The way we measure it, what we call "compilation cost" includes this additional IO produced by code and the "pure deployment" cost does not. I guess that makes the estimation of `DeployBytes` a bit inaccurate. I can think of a few ways to improve this but none of them are particularly simple. And since we only use this to get a feeling for which fraction of the deployment cost is compilation, I'm not sure if it's worth it to spend the time on that. WDYT?

jakmeier:
Using empty data-only contracts for the deployment costs per byte actually make quite a difference, basically cutting the result in half for both metrics.

```
DeployBytes   30_803_625 gas [      139.13i 0.18r 2.18w]           (computed in 1094.733588051s)
DeployBytes    3_005_059 gas [                      3ns]           (computed in 10.023759771s)
```

runtime/runtime-params-estimator/src/utils.rs
```diff
@@ -223,3 +225,10 @@ pub(crate) fn generate_fn_name(index: usize, len: usize) -> Vec<u8> {
     }
     name
 }
+
+/// Create a WASM module that is empty except for a main method and a single data entry with n characters
+pub(crate) fn generate_data_only_contract(data_size: usize) -> Vec<u8> {
+    let payload = rand::thread_rng().sample_iter(&Alphanumeric).take(data_size).collect::<String>();
```
matklad, line 231: I'd rather not make this random, again, from simplicity's point of view. I'd use

```rust
"All work and no play makes Jack a dull boy".chars().cycle().take(data_size)
```

Or are we caring about this being non-comperssible? If we do, let's use a random string with a fixed seed.

runtime/runtime-params-estimator/src/utils.rs
```diff
@@ -223,3 +225,10 @@ pub(crate) fn generate_fn_name(index: usize, len: usize) -> Vec<u8> {
...
+    let wat_code = format!("(module (data \"{}\") (func (export \"main\")))", payload);
```
matklad, line 232:
```suggestion
    let wat_code = format!("(module (data \"{payload}\") (func (export \"main\")))");
```

the future is here :)

runtime/runtime-params-estimator/src/lib.rs
```diff
@@ -613,6 +618,23 @@ fn contract_compile_base_per_byte_v2(ctx: &mut EstimatorContext) -> (GasCost, Ga
...
+    let large_code = generate_data_only_contract(bytesize::mb(4u64) as usize);
+    let small_code_len = small_code.len();
+    let large_code_len = large_code.len();
+    let cost_empty = deploy_contract_cost(ctx, small_code, Some(b"main"));
+    let cost_4mb = deploy_contract_cost(ctx, large_code, Some(b"main"));
+
+    if cost_4mb < cost_empty {
+        eprintln!("High variance in pure_deploy_bytes: Deployment cost of small contract is higher than large contract.");
+        let mut cost = GasCost::zero(ctx.config.metric);
+        cost.set_uncertain(true);
```
matklad, line 632: Do we need a `eprintln` here? Human attention is a very scarce resource, so any ad hoc printouts have a very high cost to them. 

Maybe we need `cost.set_uncertain("high variance in ... ");` instead, and display this when printing the cost? 

If ^ is a good idea, it's probably do the refacto in a follow-up though.

jakmeier: I think we can just remove the output. A literal zero cost + uncertain flag in the output seems good enough to avoid confusing the output with an actual measurement. Latest commit has it removed.

We can consider making the uncertain flag more sophisticated. But it forces us to come up with some rule how we treat it in addition/subtraction or how we reduce many flags into one. That's the main reason I like the simple boolean, it keeps the `GasCost` closer to a simple primitive.

matklad (approved):

PR #6279, opened 2022-02-11
Warn user if the state is probably corrupted (one of previous migrations was interrupted)

EdvardD:
Config changes:
- Added a flag 'fail_if_migration_is_in_progress' (default to false) to fail node if previously migration was interrupted.

State changes:
- Added a key 'migration_is_in_progress' to the column BlockMisc. Normally the flag doesn't exist or equals false. Equals true iff the last migration was interrupted in the middle. If the flag is true the error message will be shown on the node start. Depending on the flag 'fail_if_migration_is_in_progress' node start will fail.
- Added a key 'store_is_probably_broken'. Normally the flag doesn't exist. Equals true iff previously the migration was done having another interrupted migration. The flag is permanent and is used to show an error message so the user is aware that the state can be corrupted.


nearcore/src/lib.rs
```diff
@@ -74,8 +74,78 @@ pub fn get_default_home() -> PathBuf {
     PathBuf::default()
 }
 
+mod r#impl {
```
mm-near, line 77: can you explain why like this (and not putting the struct directly  ? 

(or optionally moving this struct to a separate file)

nearcore/src/lib.rs
```diff
@@ -74,8 +74,78 @@ pub fn get_default_home() -> PathBuf {
...
+    const MIGRATION_IS_IN_PROGRESS_STORE_KEY: &'static [u8] = b"migration_is_in_progress";
+    const STORE_IS_PROBABLY_CORRUPTED_STORE_KEY: &'static [u8] = b"store_is_probably_corrupted";
+    use near_store::Store;
+    use near_store::db::DBCol::ColBlockMisc;
+    use tracing::error;
+
+    pub struct MigrationIsInProgress<'a> {
```
mm-near, line 84: please comment what this struct does (that it is RAII for migrations etc)

nearcore/src/lib.rs
```diff
@@ -74,8 +74,78 @@ pub fn get_default_home() -> PathBuf {
...
+    pub fn check_if_migration_is_in_progress(store: &Store, fail_if_migration_is_in_progress: bool) {
+        let store_is_probably_corrupted =
+            match store.get_ser::<bool>(ColBlockMisc, &STORE_IS_PROBABLY_CORRUPTED_STORE_KEY) {
+                Ok(Some(x)) => x,
+                Ok(None) => false,
+                Err(e) => panic!("Can't read DB, {:?}", e),
+            };
+        if store_is_probably_corrupted {
+            error!(concat!("The state was earlier migrated while another migration was interrupted",
+                           " so the current state can be corrupted and lead to undefined failures."));
```
mm-near, line 116: I'd also mention:

'consider recovering the state from the backup'

nearcore/src/lib.rs
```diff
@@ -74,8 +74,78 @@ pub fn get_default_home() -> PathBuf {
...
+                Err(e) => panic!("Can't read DB, {:?}", e),
+            };
+
+        if migration_is_in_progress {
+            error!(concat!("The state migration was earlier interrupted so the current state can ",
+                   "be corrupted, it's recommended to recover the node from a backup otherwise ",
+                   "this can lead to undefined failures."));
+            if fail_if_migration_is_in_progress {
+                panic!(concat!("Failed to start because the state is probably corrupted and ",
+                       "the flag fail_if_migration_is_in_progress is enabled"));
```
mm-near, line 132: I'd rephrase it - to say something like:

if you want to proceed anyway, rerun with --fail_if_migration_is_in_progress=false

nearcore/src/lib.rs
```diff
@@ -74,8 +74,78 @@ pub fn get_default_home() -> PathBuf {
...
+            let mut store_update = store.store_update();
+            store_update.set_ser(ColBlockMisc, &MIGRATION_IS_IN_PROGRESS_STORE_KEY, &true).unwrap();
+            store_update.commit().unwrap();
+            MigrationIsInProgress{store}
+        }
+    }
+
+    impl<'a> Drop for MigrationIsInProgress<'a> {
+        fn drop(&mut self) {
+            if !std::thread::panicking() {
```
mm-near, line 99: let's also add a small unittest for this struct

mm-near (changes requested):

nearcore/src/lib.rs
EdvardD: I was thinking of moving it out to a separate file but there were not many files so I thought it's intended to have only several files.

This is needed so the struct can't be created directly but only using ::new method.

nikurt:
I've just made https://github.com/near/nearcore/pull/6282 which doesn't write anything to DB, but makes a persistent checkpoint to disk. This way we can not only detect if DB is probably corrupted, but can also recover without re-downloading a snapshot.

EdvardD:
> Why don't use transactions to make migrations atomic?

It seems a single migration can be too big (the whole storage update) and putting it into a single transaction can lead to issues. That's my guess seeing the implementation.

Anyways @nikurt 's change uses checkpoints to guarantee migrations being transactional so I guess this change is not needed anymore.

mm-near:
let's abandon this PR for now - in favor of @nikurt approach with checkpoints.

PR #6282, opened 2022-02-11
Safe DB migrations using RocksDB Checkpoints

nikurt:
RocksDB checkpoints hardlink files and therefore are free as long as the files are on the same filesystem. Because we usually a separate disk into ~/.near/data, this PR by default writes checkpoints inside that directory.
The PR detects if a checkpoint already exists and if it does refuses a DB migration.
Checkpoints are deleted after successful migrations.

Test plan:
* [x] Enable checkpoints even if no DB migrations are needed and run localnet
* [x] Use 1.21 version on localnet. Check that restarting with the current branch creates a checkpoint before applying DB migrations.
* [x] Use 1.21 version on localnet. When restarting with the current branch, kill the process before the migration is complete.
	* [x] Check that the checkpoint still exists
	* [x] Check that neard panics if tries to restart
	* [x] Restore from the checkpoint, delete the checkpoint and check that the migration starts over.
* [x] Repeat the test above with `use_checkpoints_for_db_migration = false` and see that the DB gets corrupted and can't be recovered.


bowenwang1996 (commented):
What is the test plan?

nearcore/src/lib.rs
```diff
@@ -74,17 +71,87 @@ pub fn get_default_home() -> PathBuf {
...
+    } else {
+        path
+    }
+}
+
+const DB_CHECKPOINT_PREFIX: &str = "db_checkpoint_";
+
+fn find_db_checkpoint(path: &Path, near_config: &NearConfig) -> Result<Option<PathBuf>, io::Error> {
+    let db_checkpoints_path = db_checkpoint_root_dir(path, near_config);
+    for entry in std::fs::read_dir(db_checkpoints_path)?.filter(|entry| {
```
EdvardD, line 86: Probably it makes sense to check here that exactly one snapshot exists and panic otherwise.

nearcore/src/lib.rs
```diff
@@ -74,17 +71,87 @@ pub fn get_default_home() -> PathBuf {
...
     if db_version == near_primitives::version::DB_VERSION {
         return;
     }
 
+    if near_config.config.use_checkpoints_for_db_migration {
+        let existing_checkpoint_path = find_db_checkpoint(path, near_config);
+        match existing_checkpoint_path {
+            Ok(existing_checkpoint_path) => match existing_checkpoint_path {
+                Some(existing_checkpoint_path) => {
+                    panic!("Detected an existing DB migration checkpoint: '{}'. If the previous DB migration attempt failed, please delete the checkpoint and restart. Otherwise, please restore from that checkpoint and restart.",existing_checkpoint_path.display());
```
EdvardD, line 130: This can be inconvenient for users. What about using that snapshot?

EdvardD, line 130: > please restore from that checkpoint

How should this be done? As far as I see we always panic if checkpoint exists.

nikurt: Technically, yes, a node can copy the files from a snapshot by itself. However, it needs to be done extremely carefully. I don't know a good way to separate files created by RocksDB vs user-created files. Thus, I can't assume that everything in ~/.near/data can be safely deleted.
Safe options of moving a checkpoint to '/tmp/' are not feasible, because '/tmp' can be a different filesystem.
As DB migrations are rare events, it should be ok to expect a user to recover manually.

An alternative to this PR is downloading a large public backup, which also requires a lot of manual work and a lot of time.

nikurt:
> What is the test plan?

@bowenwang1996 Manual testing. Updated description. Please let me know if you think we need an automated test for it.

core/store/src/db.rs
```diff
@@ -872,6 +873,10 @@ impl RocksDB {
             Ok(())
         }
     }
+
+    pub fn checkpoint(&self) -> Result<Checkpoint, DBError> {
```
mm-near, line 877: please comment what this method does (does it create a new checkpoint ?)

nearcore/src/config.rs
```diff
@@ -435,6 +439,12 @@ pub struct Config {
     /// If set, overrides value in genesis configuration.
     #[serde(skip_serializing_if = "Option::is_none")]
     pub max_gas_burnt_view: Option<Gas>,
+    /// Checkpoints let the user recover from interrupted DB migrations.
+    #[serde(default = "default_use_checkpoints_for_db_migration")]
+    pub use_checkpoints_for_db_migration: bool,
+    /// Absolute path to the root directory for DB checkpoints.
```
mm-near, line 445: it must be set if checkpoints for db_migrations. are enabled ?
what happens if it is not set ?

nearcore/src/lib.rs
```diff
@@ -74,17 +71,87 @@ pub fn get_default_home() -> PathBuf {
...
+            }
+        }
+        return false;
+    }) {
+        return Ok(Some(PathBuf::from(entry?.path())));
+    }
+    Ok(None)
+}
+
+fn create_db_checkpoint(path: &Path, near_config: &NearConfig) -> Result<PathBuf, anyhow::Error> {
```
mm-near, line 99: please add comments

(and mention that path here, is the path to the current rocksdb and checkpoint will be created in the directory that might come from the config)

nearcore/src/lib.rs
```diff
@@ -74,17 +71,87 @@ pub fn get_default_home() -> PathBuf {
...
+    if let Some(db_checkpoints_path) = &near_config.config.db_checkpoints_path {
+        &db_checkpoints_path
+    } else {
+        path
+    }
+}
+
+const DB_CHECKPOINT_PREFIX: &str = "db_checkpoint_";
+
+fn find_db_checkpoint(path: &Path, near_config: &NearConfig) -> Result<Option<PathBuf>, io::Error> {
```
mm-near, line 84: I'd add a comment that this finds the FIRST checkpoint in the directory

nearcore/src/lib.rs
```diff
@@ -74,17 +71,87 @@ pub fn get_default_home() -> PathBuf {
...
+        let existing_checkpoint_path = find_db_checkpoint(path, near_config);
+        match existing_checkpoint_path {
+            Ok(existing_checkpoint_path) => match existing_checkpoint_path {
+                Some(existing_checkpoint_path) => {
+                    panic!("Detected an existing DB migration checkpoint: '{}'. Probably a DB migration got interrupted and your DB is corrupted. Please replace the contents of '{}' with data from that checkpoint, delete the checkpoint and try again.",existing_checkpoint_path.display(), path.display());
+                }
+                None => {}
+            },
+            Err(err) => {
+                warn!(target: "near", "Failed to check if a DB checkpoint already exists: {:#?}", err);
```
mm-near, line 135: shouldn't we panic in such case ?

nearcore/src/lib.rs
```diff
@@ -74,17 +71,87 @@ pub fn get_default_home() -> PathBuf {
...
     if db_version == near_primitives::version::DB_VERSION {
         return;
     }
 
+    if near_config.config.use_checkpoints_for_db_migration {
+        let existing_checkpoint_path = find_db_checkpoint(path, near_config);
+        match existing_checkpoint_path {
+            Ok(existing_checkpoint_path) => match existing_checkpoint_path {
+                Some(existing_checkpoint_path) => {
+                    panic!("Detected an existing DB migration checkpoint: '{}'. Probably a DB migration got interrupted and your DB is corrupted. Please replace the contents of '{}' with data from that checkpoint, delete the checkpoint and try again.",existing_checkpoint_path.display(), path.display());
```
mm-near, line 130: I wonder if here, instead of a long instruction - we shouldn't point people to some README on our webpage (near.org/dev/recover-from-failed-migration) - where we can have more detailed steps.

mm-near (changes requested):

nearcore/src/lib.rs
EdvardD: I meant that we can check that no more than one snapshot exists otherwise it means the user did migration having another pending migration in a snapshot

nearcore/src/lib.rs
nikurt: Yes, I will follow up after I write that README.

nearcore/src/lib.rs
nikurt: Removed this codepath.

nearcore/src/lib.rs
```diff
@@ -74,17 +71,87 @@ pub fn get_default_home() -> PathBuf {
...
+
+    let checkpoint_path = if near_config.config.use_checkpoints_for_db_migration {
+        match create_db_checkpoint(path, near_config) {
+            Ok(checkpoint_path) => {
+                info!(target: "near", "Created a DB checkpoint before a DB migration: '{}'. Please recover from this checkpoint if the migration gets interrupted.", checkpoint_path.display());
+                Some(checkpoint_path)
+            }
+            Err(err) => {
+                warn!(target: "near", "Failed to create a DB checkpoint before a DB migration: {}", err);
+                None
```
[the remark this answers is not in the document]
nikurt: Mentioned that this issue can be resolved by disabling the checkpoints feature, now it should be acceptable to panic.

nearcore/src/config.rs
```diff
@@ -435,6 +439,13 @@ pub struct Config {
...
+    /// If not set, defaults to the database location, i.e. '$home/data/'.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub db_checkpoints_path: Option<PathBuf>,
```
matklad, line 448: It seems that other files here (genesis_file, validator_key_file) are treated as relative to the nearhome directory, we might want to do the same treatment here.

nearcore/src/lib.rs
```diff
@@ -74,17 +71,89 @@ pub fn get_default_home() -> PathBuf {
...
+        path
+    }
+}
+
+/// Returns the path of the DB checkpoint.
+fn db_checkpoint_path<'a>(path: &'a Path, near_config: &'a NearConfig) -> PathBuf {
+    let checkpoint_path = db_checkpoint_root_dir(path, near_config);
+    let checkpoint_path = checkpoint_path.join(DB_CHECKPOINT_NAME);
+    checkpoint_path
+}
```
matklad, line 89: This feels a bit more verbose than it needs to be, something like this should work:

```suggestion
/// Returns the path of the DB checkpoint.
fn db_checkpoint_path<'a>(path: &'a Path, near_config: &'a NearConfig) -> PathBuf {
    near_config.config.db_checkpoints_path.unwrap_or(path).join(DB_CHECKPOINT_NAME)
}
```

nearcore/src/lib.rs
```diff
@@ -74,17 +71,89 @@ pub fn get_default_home() -> PathBuf {
...
+                info!(target: "near", "Created a DB checkpoint before a DB migration: '{}'. Please recover from this checkpoint if the migration gets interrupted.", checkpoint_path.display());
+                Some(checkpoint_path)
+            }
+            Err(err) => {
+                panic!("Failed to create a DB checkpoint before a DB migration: {}. If this persists, you can disable checkpoints in `config.json`:\n```\"use_checkpoints_for_db_migration\": false\n```", err);
+            }
+        }
+    } else {
+        None
+    };
```
matklad, line 155: Nit, but to subsequent blocks which start with `if near_config.config.use_checkpoints_for_db_migration` feel like something which can be compressed. 

Why do we need separate "exits/create" API? Can we have just a single `create_db_checkpoint` which fails if there's an existing checkpoints? This'll avoid TOCTOU-prone API.

nearcore/src/lib.rs
```diff
@@ -274,6 +343,21 @@ pub fn apply_store_migrations(path: &Path, near_config: &NearConfig) {
...
+            match std::fs::remove_dir_all(&checkpoint_path) {
+                Ok(_) => {
+                    info!(target: "near", "Deleted DB checkpoint at '{}'", checkpoint_path.display());
+                }
+                Err(err) => {
+                    error!(target: "near", "Failed to delete a DB checkpoint at '{}'. Error: {:#?}. Please delete it manually before the next start of the node.", checkpoint_path.display(), err);
+                }
+            }
+        }
+    }
```
matklad, line 360:
```suggestion
        if let Some(checkpoint_path) = checkpoint_path {
            info!(target: "near", "Deleting DB checkpoint at '{}'", checkpoint_path.display());
            match std::fs::remove_dir_all(&checkpoint_path) {
                Ok(_) => {
                    info!(target: "near", "Deleted DB checkpoint at '{}'", checkpoint_path.display());
                }
                Err(err) => {
                    error!(target: "near", "Failed to delete a DB checkpoint at '{}'. Error: {:#?}. Please delete it manually before the next start of the node.", checkpoint_path.display(), err);
                }
            }
        }
```

The `Option` is a bearer of "do we have a checkpoint" info, we don't need any extra checks.

nearcore/src/lib.rs
```diff
@@ -274,6 +343,21 @@ pub fn apply_store_migrations(path: &Path, near_config: &NearConfig) {
         let db_version = get_store_version(path);
         debug_assert_eq!(db_version, near_primitives::version::DB_VERSION);
     }
+
+    // DB migration was successful, remove the checkpoint to avoid it taking up precious disk space.
+    if near_config.config.use_checkpoints_for_db_migration {
+        if let Some(checkpoint_path) = checkpoint_path {
+            info!(target: "near", "Deleting DB checkpoint at '{}'", checkpoint_path.display());
+            match std::fs::remove_dir_all(&checkpoint_path) {
```
matklad, line 351: Note that `std::fs::remove_dir_all` is broken on widows for large dirs, and https://docs.rs/remove_dir_all/latest/remove_dir_all/ should be preferred instead. We probably shouldn't care about this at this point though.

nearcore/src/lib.rs
```diff
@@ -74,17 +71,89 @@ pub fn get_default_home() -> PathBuf {
...
         return;
     }
 
+    // Before attemping a DB migration check if the current DB is corrupted. We assume that
+    // checkpoints get created before each migration attempt and get removed after successful
+    // attempts.
+    if near_config.config.use_checkpoints_for_db_migration {
+        match check_db_checkpoint_exists(path, near_config) {
+            Some(existing_checkpoint_path) => {
+                panic!("Detected an existing DB migration checkpoint: '{}'. Probably a DB migration got interrupted and your DB is corrupted. Please replace the contents of '{}' with data from that checkpoint, delete the checkpoint and try again.",existing_checkpoint_path.display(), path.display());
```
matklad, line 135: This is quite hard to read on a finite display, you can wrap this using `\`: 

```
        match check_db_checkpoint_exists(path, near_config) {
            Some(existing_checkpoint_path) => {
                panic!(
                    "Detected an existing DB migration checkpoint: '{}'. \
                     Probably a DB migration got interrupted and your DB is corrupted. \
                     Please replace the contents of '{}' with data from that checkpoint, \
                     delete the checkpoint and try again.",
                    existing_checkpoint_path.display(),
                    path.display()
                );
            }
            None => {}
        }
```

We might also hard wrap this (with `\n`), as terminals are often bad at wrapping, but I don't know what's the best practice here.

nearcore/src/config.rs
nikurt: Treating it as a relative path is complicated as the path needs to exist.
If a user provides "./sub/dir", then ~/.near/sub/dir needs to exist, which can be confusing.
Instead I added a check that the given path is an absolute path.

mm-near (approved):

nearcore/src/lib.rs
```diff
@@ -74,17 +71,81 @@ pub fn get_default_home() -> PathBuf {
...
+            "Detected an existing database migration snapshot: '{}'.\n\
+             Probably a database migration got interrupted and your database is corrupted.\n\
+             Please replace the contents of '{}' with data from that checkpoint, delete the checkpoint and try again.",checkpoint_path.display(), path.display());
+    }
+
+    let db = RocksDB::new(path)?;
+    let checkpoint = db.checkpoint()?;
+    info!(target:"near","Creating a database migration snapshot in '{}'",checkpoint_path.display());
+    checkpoint.create_checkpoint(&checkpoint_path)?;
+    info!(target:"near","Created a database migration snapshot in '{}'",checkpoint_path.display());
```
mm-near, line 107: nit: no space after commas? I thought that code-style should have fixed that?

bowenwang1996: > I thought that code-style should have fixed that?

I think it doesn't work well with macros.

nearcore/src/lib.rs
```diff
@@ -74,17 +71,81 @@ pub fn get_default_home() -> PathBuf {
...
+
+/// Creates a consistent DB checkpoint and returns its path.
+/// By default it creates checkpoints in the DB directory, but can be overridden by the config.
+fn create_db_checkpoint(path: &Path, near_config: &NearConfig) -> Result<PathBuf, anyhow::Error> {
+    let checkpoint_path = db_checkpoint_path(path, near_config);
+    if checkpoint_path.exists() {
+        panic!(
+            "Detected an existing database migration snapshot: '{}'.\n\
+             Probably a database migration got interrupted and your database is corrupted.\n\
+             Please replace the contents of '{}' with data from that checkpoint, delete the checkpoint and try again.",checkpoint_path.display(), path.display());
```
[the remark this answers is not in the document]
nikurt: Changed to return an `Err`. The only concern is that this error message is verbose, and the callsite handling error messages is verbose too.

nearcore/src/lib.rs
nikurt: Thanks, haven't noticed this before.
Removing `target` lets the tool format this line. But an `error!` macro further down couldn't be formatted. Maybe its format string is too long.

bowenwang1996 (commented):
@nikurt could you add documentation on how to use this feature? cc @janewang

PR #6284, opened 2022-02-11
feat(params-estimator): Initialize state in home

jakmeier:


jakmeier:
Creating temporary state from scratch every time is not very practical when running with millions of accounts, taking hours to insert them. Using `--home some-dir` with manually initialized state today doesn't work nicely because the estimator has assumptions on what contracts are deployed.

This new flag allows to prepare a state directory exactly as the temporary state for estimation is produced. But at a persistent location that can be reused for repeated runs.

runtime/runtime-params-estimator/src/main.rs
```diff
@@ -102,49 +107,48 @@ fn main() -> anyhow::Result<()> {
...
+            Arc::new(near_config.genesis),
+            store,
+        )
+        .add_additional_accounts(cli_args.additional_accounts_num)
+        .add_additional_accounts_contract(contract_code)
+        .print_progress()
+        .build()
+        .unwrap()
+        .dump_state()
+        .unwrap();
```
matklad, line 150: I wonder if we can `?` now

runtime/runtime-params-estimator/src/main.rs
```diff
@@ -28,6 +28,9 @@ struct CliArgs {
     /// to generate appropriate data.
     #[clap(long)]
     home: Option<PathBuf>,
+    /// If a home directory has been specified, set this flag if it needs to be initialized.
+    #[clap(long)]
+    init_home: bool,
```
matklad, line 33: Can we just infer/DWIM this? Ie, if the directory is empty, it is populated

matklad (approved):

runtime/runtime-params-estimator/src/main.rs
jakmeier: Not out of the box:
```
error[E0277]: the size for values of type `dyn std::error::Error` cannot be known at compilation time
   --> runtime/runtime-params-estimator/src/main.rs:149:22
    |
149 |         .dump_state()?;
    |                      ^ doesn't have a size known at compile-time
    |
    = help: the trait `Sized` is not implemented for `dyn std::error::Error`
    = note: required because of the requirements on the impl of `std::error::Error` for `Box<dyn std::error::Error>`
    = note: required because of the requirements on the impl of `From<Box<dyn std::error::Error>>` for `anyhow::Error`
    = note: required because of the requirements on the impl of `FromResidual<Result<Infallible, Box<dyn std::error::Error>>>` for `Result<(), anyhow::Error>`

error[E0277]: `dyn std::error::Error` cannot be sent between threads safely
   --> runtime/runtime-params-estimator/src/main.rs:149:22
    |
149 |         .dump_state()?;
    |                      ^ `dyn std::error::Error` cannot be sent between threads safely
    |
    = help: the trait `Send` is not implemented for `dyn std::error::Error`
    = note: required because of the requirements on the impl of `Send` for `Unique<dyn std::error::Error>`
    = note: required because it appears within the type `Box<dyn std::error::Error>`
    = note: required because of the requirements on the impl of `From<Box<dyn std::error::Error>>` for `anyhow::Error`
    = note: required because of the requirements on the impl of `FromResidual<Result<Infallible, Box<dyn std::error::Error>>>` for `Result<(), anyhow::Error>`
```

runtime/runtime-params-estimator/src/main.rs
jakmeier: Yes that seems like a better approach here!

runtime/runtime-params-estimator/src/main.rs
matklad: Ah yeah, that's the general problem that, due to coherence, there may only be *one* dynamically typed error. ideally we should migrate that other crate to anyhow.

matklad (approved):

Issue #6150, opened 2022-01-21
NEP 264 tracking issue


matklad:
See https://github.com/near/NEPs/pull/264.

Implementation PR: https://github.com/near/nearcore/pull/6285

The NEP is pre-approved, but not approved. We want to have experimental nightly protocol feature implementation, but we don't yet commit to stabilization. 

Plan for implementation: 

Implementation wise, gas reshuffling should happen here: 

https://github.com/near/nearcore/blob/1cbcc5eb89be9c78364d2ab984a192c77f9538aa/runtime/near-vm-logic/src/logic.rs#L2487

I'd imagine something like this 

```
    pub fn outcome(self) -> VMOutcome {
        let unused_gas = if self.context.is_view() { 0 } else { self.context.prepaid_gas - self.gas_counter.used_gas() };
        let distributed_gas = self.ext.distribute_unused_gas(unused_gas);
        self.gas_counter.prepay_gas(distributed_gas);

        let burnt_gas = self.gas_counter.burnt_gas();
        let used_gas = self.gas_counter.used_gas();
```

That is, `dyn External` would the thing maintaining the required state, like this: 

```diff
diff --git a/runtime/near-vm-logic/src/logic.rs b/runtime/near-vm-logic/src/logic.rs
index ef64d75b2..c72ae1fb2 100644
--- a/runtime/near-vm-logic/src/logic.rs
+++ b/runtime/near-vm-logic/src/logic.rs
@@ -2482,6 +2482,10 @@ impl<'a> VMLogic<'a> {
 
     /// Computes the outcome of execution.
     pub fn outcome(self) -> VMOutcome {
+        let unused_gas = if self.context.is_view() { 0 } else { self.context.prepaid_gas - self.gas_counter.used_gas() };
+        let distributed_gas = self.ext.distribute_unused_gas(unused_gas);
+        self.gas_counter.prepay_gas(distributed_gas);
+
         let burnt_gas = self.gas_counter.burnt_gas();
         let used_gas = self.gas_counter.used_gas();
 
diff --git a/runtime/runtime/src/ext.rs b/runtime/runtime/src/ext.rs
index 17fefa28d..b49b9dc28 100644
--- a/runtime/runtime/src/ext.rs
+++ b/runtime/runtime/src/ext.rs
@@ -25,6 +25,10 @@ pub struct RuntimeExt<'a> {
     trie_update: &'a mut TrieUpdate,
     account_id: &'a AccountId,
[...]

PR #6285, opened 2022-02-12
feat: implement NEP264, function call gas weight

austinabell:
More of a rough draft because I don't have context around the runtime as to the best migration plan of these APIs. I've followed the recommendation of #6150 to avoid refactoring to move action receipts to `VMLogic`, but this means the `External` trait needs to be updated, and it's unclear if anyone has preferences.

TODO:
- e2e test to cover `RuntimeExt` as current tests just cover `MockedExternal` (same functionality though)
  - Also these tests don't check that gas was distributed to the correct function calls
- Determine the best path for trait migration (would we do the refactor mentioned above before any releases?)
- Where does the protocol version number come from, and what should this one be numbered?
- Benchmark to gauge gas cost and verify existing impl doesn't have higher cost
- Would be nice to fuzz test this to make sure there aren't any weird edge cases


runtime/runtime/src/ext.rs
```diff
@@ -35,6 +35,18 @@ pub struct RuntimeExt<'a> {
...
     current_protocol_version: ProtocolVersion,
+
+    #[cfg(feature = "protocol_feature_function_call_ratio")]
+    distribute_leftover_gas_to: Vec<GasRatioMetadata>,
+    #[cfg(feature = "protocol_feature_function_call_ratio")]
+    gas_ratio_sum: u64,
+}
+
+#[cfg(feature = "protocol_feature_function_call_ratio")]
+struct GasRatioMetadata {
```
nearmax, line 46: We try to avoid `u*` types in the runtime as much as possible. If you find usage of `u*` type anywhere please file a Github issue.

runtime/near-vm-logic/src/dependencies.rs
```diff
@@ -486,4 +531,11 @@ pub trait External {
 
     /// Returns total stake of validators in the current epoch.
     fn validator_total_stake(&self) -> Result<Balance>;
+
+    /// Distribute the gas among the scheduled function calls that specify a gas ratio.
+    /// Returns the amount of distributed gas.
+    ///
+    /// * `gas` - amount of unused gas to distribute
```
nearmax, line 538: Please describe the returned value.

runtime/runtime/src/ext.rs
```diff
@@ -389,4 +445,31 @@ impl<'a> External for RuntimeExt<'a> {
...
+                        .action_receipts
+                        .get_mut(receipt_index)
+                        .and_then(|(_, receipt)| receipt.actions.get_mut(action_index))
+                    {
+                        *gas += assign_gas;
+                    } else {
+                        panic!("Invalid index for assigning unused gas ratio");
+                    }
+
+                    assign_gas
```
nearmax, line 468: It is preferable if we always distribute all of the unused gas to avoid emitting gas refund receipt. It is okay if the last receipt gas a little bit extra gas. Right now we have leftover gas because of the division. We then would need to write tests to make sure refund receipt is never emitted. Once we do it the return value is always going to be either equal to `gas` or zero, which means it will become semantically a bool.

runtime/runtime/src/ext.rs
matklad: I must say I don't entirely understand what should be used instead. As this interfaces with `VmLogic` which used raw `u64/u32`, we actually have a whole bunch of `u64`s in various signatures in this module. 

But yeah, I'd probably spell this as:

```rust
struct RuntimeExt {
  ...
  gas_weights: Vec<(FunctionCallActionIndex, GasWeight)>
}
struct GasWeight(u64);

struct FunctionCallActionIndex {
  receipt_index: usize,
  action_inde
}

impl RuntimeExt {
    fn get_function_call_action_mut(&mut self, index: FunctionCallActionIndex) -> &mut FunctionCallAction {
    }
}
```

* Separate the two-dimensional index into its own type
* Add a typed wrapper to the weight
* Remove `gas_ratio_sum` field -- it seems we can compute it in the `distribute` function. To check if it's zero, you can check if `gas_weights` is not empty

runtime/runtime/src/ext.rs
```diff
@@ -254,6 +276,40 @@ impl<'a> External for RuntimeExt<'a> {
...
+        );
+
+        if gas_ratio > 0 {
+            self.distribute_leftover_gas_to.push(GasRatioMetadata {
+                receipt_index: receipt_index as usize,
+                action_index,
+                gas_ratio,
+            });
+            self.gas_ratio_sum =
+                self.gas_ratio_sum.checked_add(gas_ratio).ok_or(HostError::IntegerOverflow)?;
```
matklad, line 307: This error condition doesn't feel great to me. This is a code path which could be triggered by the user, it's not just dead code, so it adds to the number of code paths which we need to exercise. "You can't use `u64::MAX - 1` as a ratio" is also an error condition the user needs to worry about. Let's implement the logic in a way which can deal with arbitrary ratios. Using `u128` to sum the ratios should work I think.

runtime/near-vm-logic/src/tests/gas_counter.rs
```diff
@@ -102,6 +102,53 @@ fn test_hit_prepaid_gas_limit() {
...
+
+    // Split two
+    function_call_ratio_check([(0, 3), (0, 2)]);
+
+    // Split two with static gas
+    function_call_ratio_check([(1_000_000, 3), (3_000_000, 2)]);
+
+    // Many different gas ratios
+    function_call_ratio_check([(1_000_000, 3), (3_000_000, 2), (0, 1), (1_000_000_000, 0), (0, 4)]);
+}
```
matklad, line 150: We need to add case which hits total_gas_ratio overflow in the current implementation.

matklad (commented):
Feels broadly right to me! 

One thing which I feel suuuper uneasy though is the fact that the tests here check our *mock* implementation, rather than the real one. Here's  what we can do here:

* Have at least one integration test here. I don't know what's the right place for it though. They are some more integration tests in `near-vm-runner/tests/rs_contract`, but they still use mocked external
* Move promise handling logic from `runtime` to `vm-logic`. Today vm-logic depends on primitives, so that should be fine. It'll also allow us to reduce duplication in MockedExternal. 
* Alternatively, keep actual promises data in `runtime`, but move bookeeping to `vm-logic`

Though, I'd personally be fine langing this PR without proper tests for new functionality, as long as the feature is experimental. That is, we absolutelly have to add tests which exercise the real thing, but we can do that in a follow-up PR

runtime/near-vm-logic/src/logic.rs
```diff
@@ -1506,6 +1506,62 @@ impl<'a> VMLogic<'a> {
...
+
+    /// Appends `FunctionCall` action to the batch of actions for the given promise pointed by
+    /// `promise_idx`. This function allows not specifying a specific gas value and allowing the
+    /// runtime to assign remaining gas based on a ratio.
+    ///
+    /// # Gas
+    ///
+    /// Gas can be specified using a static amount, a ratio of remaining prepaid gas, or a mixture
+    /// of both. To omit a static gas amount, [`u64::MAX`] can be passed for the `gas` parameter.
+    /// To omit assigning remaining gas, [`u64::MAX`] can be passed as the `gas_ratio` parameter.
```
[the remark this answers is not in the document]
austinabell: ah, yes. I had updated this in the NEP but forgot to update here. Good catch

runtime/near-vm-logic/src/dependencies.rs
```diff
@@ -276,6 +276,51 @@ pub trait External {
...
+    ///     100u64,
+    ///     2,
+    /// ).unwrap();
+    /// ```
+    ///
+    /// # Panics
+    ///
+    /// Panics if the `receipt_index` does not refer to a known receipt.
+    #[cfg(feature = "protocol_feature_function_call_ratio")]
+    // TODO: unclear if this should be a supertrait to avoid semver breaking changes
```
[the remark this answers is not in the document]
austinabell: so are you suggesting modifying `append_action_function_call` or keeping as is?

runtime/runtime/src/ext.rs
austinabell: Do you want the u64 removed from the trait interface or just internally in the implementation? I notice for the other types in the signature the types are aliases in the trait definition but then just use the primitive type in the implementation.

As for the proposed API, it wouldn't work with the current implementation because the gas weights need to be borrowed mutably so I can't take a mutable reference on `self` when it's used. Seems more complex to read with this change to me, but happy to change to whatever is preferred.

runtime/near-vm-logic/src/dependencies.rs
austinabell: The returned value will be removed based on the following comment, as all gas will be distributed.

runtime/runtime/src/ext.rs
austinabell: The only issue I see to this is the overhead of getting the modulus of the division before the operations or having to re-index the initial `gas % n` items if we want to micro optimize. Will do this change and then will re-visit later to see if implications on the cost of the operation

runtime/runtime/src/ext.rs
austinabell: Seems like a bit of a foot-gun because the gas maximum is u64::MAX, so having the ratio larger doesn't do anything and goes into the fallback case, which could silently cause unexpected behaviour. Should never realistically happen so maybe not worth handling this case.

runtime/runtime/src/ext.rs
austinabell: So I've changed to just wrapping internally for now, but can expose this type through the trait interface if that's what we want to move the API toward

runtime/runtime/src/ext.rs
austinabell: Made changes with Max's suggestion for now for simplicity, but will update once we resolve these edge cases :)

austinabell:
> * Move promise handling logic from `runtime` to `vm-logic`. Today vm-logic depends on primitives, so that should be fine. It'll also allow us to reduce duplication in MockedExternal.
> * Alternatively, keep actual promises data in `runtime`, but move bookeeping to `vm-logic`

You indicated in the issue that you thought this refactor should be done in a separate PR, has your opinion on this changed or are you referring to doing this after this PR?

austinabell:
Okay I'm going to open this PR for review so I can start getting other feedback while I wait for resolution on a few things or input on those things.

The outstanding things are:
- Where does the protocol version number come from? I picked the next number I could see, but not sure if it's overlapping
- Move promise handling logic from `runtime` to `vm-logic` based on a suggestion from @matklad. I was originally avoiding this based on #6150 recommending doing this in a separate PR, unclear if the plan is changed and now is intended to do in this one.
- e2e test to cover `RuntimeExt`. Scoping this out it seems like this is more involved since there isn't the test scaffolding set up to allow testing these, and maybe not that necessary if logic is moved from the previous point
- Should the `GasWeight` type be exposed to the trait API? https://github.com/near/nearcore/pull/6285#discussion_r805474236 and related comments suggested this, but it is unlike other types in related files that are aliases of the primitive types. I've made the newtype part of the API here [`78d4a95` (#6285)](https://github.com/near/nearcore/pull/6285/commits/78d4a95212d4caec39e020fadfddf4bb5096369b) but I don't think it's very consistent if we don't update others.

runtime/runtime/src/ext.rs
```diff
@@ -389,4 +439,47 @@ impl<'a> External for RuntimeExt<'a> {
...
+                .gas_weights
+                .iter()
+                .map(|(action_index, GasWeight(weight))| {
+                    let assigned_gas = gas_per_weight * weight;
+
+                    distribute_gas(action_index, assigned_gas);
+
+                    assigned_gas
+                })
+                .sum();
```
bowenwang1996, line 473: I suggest rewriting this logic using a for loop to make it easier to understand

runtime/near-vm-logic/src/logic.rs
```diff
@@ -2481,7 +2554,19 @@ impl<'a> VMLogic<'a> {
...
+        #[cfg(feature = "protocol_feature_function_call_weight")]
+        if !self.context.is_view() {
+            // Distribute unused gas to scheduled function calls
+            let unused_gas = self.context.prepaid_gas - self.gas_counter.used_gas();
+
+            // Distribute the unused gas and prepay for the gas.
+            if self.ext.distribute_unused_gas(unused_gas) {
+                self.gas_counter.prepay_gas(unused_gas).unwrap();
+            }
+        }
```
bowenwang1996, line 2568: This function now has a nontrivial side effect. We might want to rename it and add a comment about what it does

bowenwang1996 (approved):

runtime/near-vm-logic/src/logic.rs
austinabell: took a stab at this, open to any suggestions of naming or any other information to include in the rustdocs since it's opinionated

runtime/runtime/src/ext.rs
```diff
@@ -140,13 +154,18 @@ impl<'a> RuntimeExt<'a> {
...
             .get_mut(receipt_index as usize)
             .expect("receipt index should be present")
             .1
-            .actions
-            .push(action);
+            .actions;
+
+        actions.push(action);
+
+        // Return index that action was inserted at
```
nagisa, line 167: This should be a function doc-comment.

nagisa (approved):
Runtime bits broadly LGTM, some nits inline.

runtime/near-vm-logic/src/dependencies.rs
```diff
@@ -486,4 +531,17 @@ pub trait External {
...
+    /// Distribute the gas among the scheduled function calls that specify a gas weight.
+    /// Returns the amount of distributed gas.
+    ///
+    /// # Arguments
+    ///
+    /// * `gas` - amount of unused gas to distribute
+    ///
+    /// # Returns
+    ///
+    /// Function returns true if gas was distributed, false if no ratios existed to distribute to.
```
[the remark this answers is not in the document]
austinabell: Ah, it would be an outdated comment above, I can update that.

I'm a little skeptical to add an additional type to the public trait API, are you sure this is in line with all usages? Seems as if most of the API is primitive types

runtime/near-vm-logic/src/logic.rs
```diff
@@ -1541,7 +1594,27 @@ impl<'a> VMLogic<'a> {
...
+            arguments,
+            amount,
+            gas,
+            gas_weight,
+        )?;
+
+        #[cfg(not(feature = "protocol_feature_function_call_weight"))]
+        {
+            let _ = gas_weight;
+            self.ext.append_action_function_call(
```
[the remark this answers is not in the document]
austinabell: I just did this to avoid code duplication from `promise_batch_action_function_call`. Are you suggesting to panic if the weight is something other than 0 without the protocol feature?

runtime/near-vm-logic/src/logic.rs
```diff
@@ -2480,8 +2553,24 @@ impl<'a> VMLogic<'a> {
...
     }
 
-    /// Computes the outcome of execution.
-    pub fn outcome(self) -> VMOutcome {
+    /// Computes the outcome of the execution.
+    ///
+    /// If `FunctionCallWeight` protocol feature (127) is enabled, unused gas will be
+    /// distributed to functions that specify a gas weight. If there are no functions with
+    /// a gas weight, the outcome will contain unused gas as usual.
+    #[allow(unused_mut)]
```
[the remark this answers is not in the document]
austinabell: This function is used without the feature, are you suggesting duplicating the function logic and calling the respective functions based on if the feature is set or not? Seems a bit jankier this way

runtime/runtime/src/ext.rs
austinabell: Added comment to the function, but kept this. Would you prefer if I removed it?

runtime/near-vm-logic/src/logic.rs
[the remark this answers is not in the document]
nagisa: Ah my bad. I meant 

```rust
#[cfg_attr(not(feature = "protocol_feature_function_call_weight"), allow(unused_mut))]
```

`cfg_attr` here applies the `allow(unused_mut)` based on absence of the `protocol_feature_function_call_weight` feature. Since the `mut` is only unused when `cfg!(feature = "protocol_feature_function_call_weight")`. This has the benefit of making it less easy to forget to remove the lint disable later when we remove the feature altogether and also allows to check that the rest of the code in the function does not accidentally violate this lint in future edits.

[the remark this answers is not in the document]
austinabell: oh oops, yes, I misread when skimming as well. Of course

runtime/near-vm-logic/src/logic.rs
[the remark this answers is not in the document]
nagisa: For deduplication I would probably add an inherent method which takes a callback/closure to call at the end.

So e.g.

```rust
    fn promise_batch_action_function_call_generic<F: FnOnce(...) -> Result<()>>(
        &mut self, 
        promise_idx: u64,
        method_name_len: u64,
        method_name_ptr: u64,
        arguments_len: u64,
        arguments_ptr: u64,
        amount_ptr: u64,
        gas: Gas,
        callback: F,
    ) -> Result<()> {
        /* the deduplicated logic... */
        callback()
    }

    pub fn promise_batch_action_function_call(
        &mut self,
        promise_idx: u64,
        method_name_len: u64,
        method_name_ptr: u64,
        arguments_len: u64,
        arguments_ptr: u64,
        amount_ptr: u64,
        gas: Gas,
    ) -> Result<()> {
        self.promise_batch_action_function_call_generic(
            promise_idx,
            method_name_len,
            method_name_ptr,
            arguments_len,
            arguments_ptr,
            amount_ptr,
            gas,
            |...| {
                self.ext.append_action_function_call(...)
            }
        )
    }

    #[cfg(feature = "protocol_feature_function_call_weight")]
    pub fn promise_batch_action_function_call_weight(...) {
        self.promise_batch_action_function_call_generic(..., |...| {
            self.ext.append_action_function_call_weight(...)
        })
    }
```

But to answer your question, the fact that a call to `promise_batch_action_function_call_weight` can be called with or without this protocol feature enabled seems pretty sketchy to me, as well as the fact that it transparently can choose to behave exactly like `promise_batch_action_function_call`.

runtime/near-vm-logic/src/dependencies.rs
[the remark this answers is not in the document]
nagisa: My concern is largely rooted in the fact that this function mutates some state in a semi-fallible looking manner and makes its outcome easy to ignore, or handle incorrectly. There seems to be little in the way to prevent somebody from forgetting to deal with the gas that has been distributed by the call to this function. Not to mention a `false` return is not really distinguishable between “this is called twice, neard is buggy” and “there is no calls with weights”.

Are there use-cases where distributing the gas but not prepaying it is useful? Maybe prepaying inside the implementation of this method would make sense?

> Seems as if most of the API is primitive types

What do you mean? I believe use of primitive types in signatures is only required for VM host function exports, but this isn't such a function.

runtime/runtime/src/ext.rs
nagisa: I think its fine either way, thanks!

runtime/near-vm-logic/src/dependencies.rs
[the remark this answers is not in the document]
austinabell: yeah, this is fair. I think I'll go with a `must_use` enum in this case, since having no ratios isn't actually an "error".

> What do you mean? I believe use of primitive types in signatures is only required for VM host function exports, but this isn't such a function.

I was just noting that it's not ergonomic to have a bunch of custom types as part of the trait API, requiring implementations to import the type and requiring importing this type when calling the function.

runtime/near-vm-logic/src/mocks/mock_external.rs
```diff
@@ -209,6 +248,46 @@ impl External for MockedExternal {
     fn validator_total_stake(&self) -> Result<Balance> {
         Ok(self.validators.values().sum())
     }
+
+    #[cfg(feature = "protocol_feature_function_call_weight")]
+    fn distribute_unused_gas(&mut self, gas: Gas) -> GasDistribution {
```
mm-near, line 253: comment (especially the part that you're actually modifying the assigned_gas for the receipts

runtime/near-vm-logic/src/mocks/mock_external.rs
```diff
@@ -209,6 +248,46 @@ impl External for MockedExternal {
...
+            let mut distribute_gas = |metadata: &FunctionCallActionIndex, assigned_gas: u64| {
+                let FunctionCallActionIndex { receipt_index, action_index } = metadata;
+                if let Some(Action::FunctionCall(FunctionCallAction { ref mut gas, .. })) = self
+                    .receipts
+                    .get_mut(*receipt_index)
+                    .and_then(|receipt| receipt.actions.get_mut(*action_index))
+                {
+                    *gas += assigned_gas;
+                } else {
+                    panic!("Invalid index for assigning unused gas weight");
```
mm-near, line 268: maybe also print what was the index (and also some other identifier - transaction/receipt id?)

( I assume that this should never ever happen, right?)

runtime/near-vm-logic/src/mocks/mock_external.rs
```diff
@@ -209,6 +248,46 @@ impl External for MockedExternal {
     fn validator_total_stake(&self) -> Result<Balance> {
         Ok(self.validators.values().sum())
     }
+
+    #[cfg(feature = "protocol_feature_function_call_weight")]
+    fn distribute_unused_gas(&mut self, gas: Gas) -> GasDistribution {
+        let gas_weight_sum: u128 =
+            self.gas_weights.iter().map(|(_, GasWeight(weight))| *weight as u128).sum();
+        if gas_weight_sum != 0 {
+            let gas_per_weight = (gas as u128 / gas_weight_sum) as u64;
```
mm-near, line 257: if someone sets very large 'gas_weight' - then gas_per_weight can be 0, right? 
(we should mention it in the comment)

runtime/near-vm-logic/src/tests/gas_counter.rs
```diff
@@ -96,12 +96,76 @@ fn test_hit_prepaid_gas_limit() {
...
+            static_gas,
+            GasWeight(gas_weight),
+        )
+        .expect("batch action function call should succeed");
+
+        weight_sum += gas_weight as u128;
+    }
+    let outcome = logic.compute_outcome_and_distribute_gas();
+
+    // Verify that all gas is used when only one ratio is specified
```
mm-near, line 130: confusing comment

runtime/near-vm-logic/src/tests/gas_counter.rs
```diff
@@ -96,12 +96,76 @@ fn test_hit_prepaid_gas_limit() {
...
+    println!("{} {} {}", outcome.used_gas, gas_limit, weight_sum);
```
mm-near, line 131: improve the print - or remove (as currently it will print just 3 numbers)

runtime/near-vm-logic/src/tests/gas_counter.rs
```diff
@@ -96,12 +96,76 @@ fn test_hit_prepaid_gas_limit() {
...
+    assert!(outcome.used_gas as u128 + weight_sum - 1 >= gas_limit as u128);
+}
+
+#[cfg(feature = "protocol_feature_function_call_weight")]
+#[test]
+fn function_call_weight_single_smoke_test() {
+    // Single function call
```
mm-near, line 138: are we testing how much each call got?

runtime/near-vm-logic/src/tests/gas_counter.rs
```diff
@@ -96,12 +96,76 @@ fn test_hit_prepaid_gas_limit() {
...
+        .expect("batch action function call should succeed");
+
+        weight_sum += gas_weight as u128;
+    }
+    let outcome = logic.compute_outcome_and_distribute_gas();
+
+    // Verify that all gas is used when only one ratio is specified
+    println!("{} {} {}", outcome.used_gas, gas_limit, weight_sum);
+    assert!(outcome.used_gas as u128 + weight_sum - 1 >= gas_limit as u128);
+}
```
mm-near, line 133: should we also check that used_gas < gas_limit? 

And maybe return how much each promise got? (so that we can verify it below)

runtime/near-vm-logic/src/tests/gas_counter.rs
```diff
@@ -96,12 +96,76 @@ fn test_hit_prepaid_gas_limit() {
...
+    function_call_weight_check([(0, 1)]);
+
+    // Single function with static gas
+    function_call_weight_check([(888, 1)]);
+
+    // Large weight
+    function_call_weight_check([(0, 88888)]);
+
+    // Weight larger than gas limit
+    function_call_weight_check([(0, 11u64.pow(14))]);
```
mm-near, line 148: right -- and in this case, how much gas do we expect that it gets? 

(as according to the code, we'll NOT give it any additional gas)

runtime/near-vm-logic/src/tests/gas_counter.rs
```diff
@@ -96,12 +96,76 @@ fn test_hit_prepaid_gas_limit() {
...
+
+    // Split two
+    function_call_weight_check([(0, 3), (0, 2)]);
```
mm-near, line 151: I'd also have a test with one '0' and one-nonzero

And a test with two '0' -- to verify that we don't give them any additional gas.

runtime/runtime/src/ext.rs
```diff
@@ -389,4 +440,44 @@ impl<'a> External for RuntimeExt<'a> {
...
+                {
+                    *gas += assigned_gas;
+                } else {
+                    panic!("Invalid index for assigning unused gas weight");
+                }
+            };
+
+            let mut distributed = 0;
+            for (action_index, GasWeight(weight)) in &self.gas_weights {
+                let assigned_gas = gas_per_weight * weight;
```
mm-near, line 466: let's mention that this should not overflow ;-)

runtime/runtime/src/ext.rs
```diff
@@ -389,4 +440,44 @@ impl<'a> External for RuntimeExt<'a> {
             .validator_total_stake(self.epoch_id, self.prev_block_hash)
             .map_err(|e| ExternalError::ValidatorError(e).into())
     }
+
+    #[cfg(feature = "protocol_feature_function_call_weight")]
+    fn distribute_unused_gas(&mut self, gas: u64) -> GasDistribution {
```
mm-near, line 445: I see that we have the same code here and inside the mock -- could we somehow combine it together ?

mm-near (changes requested):

runtime/near-vm-logic/src/mocks/mock_external.rs
austinabell: This is documented on the trait. Are you suggesting all implementations to give details about how specifically it distributes gas?

runtime/near-vm-logic/src/mocks/mock_external.rs
austinabell: yeah, this is currently unreachable, but just have a static error message to help debug if the functionality ever changes to re-order actions or receipts.

Using formatting machinery to include indices comes at a cost, can you confirm this is ideal before I make the change?

runtime/near-vm-logic/src/dependencies.rs
```diff
@@ -276,6 +278,49 @@ pub trait External {
         prepaid_gas: Gas,
     ) -> Result<()>;
 
+    /// Attach the [`FunctionCallAction`] action to an existing receipt.
+    ///
```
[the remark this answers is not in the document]
austinabell: Assumed the docs for this would be lighter since it's documented in `promise_batch_action_function_call_weight`, but I see the benefit in giving some info here also.

runtime/near-vm-logic/src/mocks/mock_external.rs
austinabell: Added docs in vm logic for this, let me know if it's preferable to duplicate docs to the external trait and/or impls

runtime/near-vm-logic/src/tests/gas_counter.rs
austinabell: ah, yes, this is also an outdated assertion since now all gas is used. Updated

runtime/near-vm-logic/src/tests/gas_counter.rs
austinabell: Seemed weird to assert details of the MockedExternal as there are no methods to view scheduled actions, but I will do this in a minute. These tests were just intended to be smoke tests where specific details weren't asserted

runtime/runtime/src/ext.rs
austinabell: So the reason I didn't do here is that there is a plan, as mentioned in #6150, to move the logic for receipt handling into VM logic which would remove this duplication. I assume this change would follow this PR. 

The types for action_receipts are currently slightly different and the implementations are in different crates, so creating a temporary abstraction that is exposed publicly doesn't make much sense to me

austinabell:
@mm-near I've tried again to find a clean way to assert the amount of gas allocated with the current testing scaffolding based on your comments, but I'll try to point out why it's more involved and should come in a separate change:
- `External` in VM logic is `&'a mut dyn External` so it would require downcasting just to get a `MockedExternal` to assert specific values on a mock implementation (doesn't seem that valuable)
  - Would also be asserting specific values assuming gas usage on a mocked implementation. This also doesn't seem valuable and is very brittle for this test breaking for other reasons unless the test is involved and recalculates everything
- There is no testing scaffolding to test non-mock implementations of `External`. This would be much easier and more valuable if testing after migration of receipt logic from external to VM logic
  - Quite an involved change for this, and didn't want to block this implementation

If you think this is imperative to test specific allocated values in this PR I can do the suggestion above with downcasting and re-calculating gas allocations, but I wanted to be clear about context before doing this as it feels janky to me to do in this PR.

runtime/near-vm-logic/src/dependencies.rs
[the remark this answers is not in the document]
mm-near: it is fine to have lighter docs - but then let's make sure that we have a 'pointer' to the place where people can read more details.

(I actually don't know which file is read by users - this one or logic.rs?)

runtime/near-vm-logic/src/mocks/mock_external.rs
mm-near: yes, if possible. It helps a lot during code review and later code reading.

runtime/near-vm-logic/src/mocks/mock_external.rs
mm-near: Thanks for adding those. I'd also add a one-line comment here, that we're using a floor, and we know that it might be 0 if the sum is too large.

runtime/near-vm-logic/src/mocks/mock_external.rs
mm-near: but wait -  this cost would be paid only if the panic actually triggers, right?

runtime/runtime/src/ext.rs
mm-near: uh, I didn't notice that the action_receipts are different. We should clean it up, but doing it in the next PR is ok.

runtime/near-vm-logic/src/tests/gas_counter.rs
mm-near: ping.

runtime/near-vm-logic/src/tests/gas_counter.rs
mm-near: fine - but then let's have a test that actually test this new funcionality

runtime/near-vm-logic/src/tests/gas_counter.rs
mm-near: ok.

mm-near (approved):

mm-near:
I'm approving to unblock you - but I'd really like to see a PR with a test that covers this functionality - as from what I see, if I change the logic to (for example) assign all the remaining gas to the last receipt (and ignore the weights) - all the tests would still pass..

PR #6286, opened 2022-02-12
chain: add epoch info to StatusSyncInfo

### LINKED ISSUE #2914: Extend RPC /status to calc epoch progress
It would be nice to have `epoch_id` and `start_epoch_height` in response to the status method instead of making 3 different method calls just to get the needed data to calculate epoch progress.

mooori:
**Problem:** RPC endpoint `status` should return more epoch information.

**Proposed solution:** Add `epoch_id` and `epoch_start_height` to `StatusSyncInfo`

Fixes #2914


core/primitives/src/views.rs
```diff
@@ -302,6 +302,8 @@ pub struct QueryResponse {
 #[cfg_attr(feature = "deepsize_feature", derive(deepsize::DeepSizeOf))]
 #[derive(Serialize, Deserialize, Debug)]
 pub struct StatusSyncInfo {
+    pub epoch_id: EpochId,
+    pub epoch_start_height: BlockHeight,
```
matklad, line 306: Do we need to add `serde::default()` or `Option` here to maintain compatibility?

mooori: When deserializing a `StatusSyncInfo` without these fields, I think the default values of `EpochId` and `BlockHeight` could be misleading. So I went with `Option`.

mm-near (changes requested):

khorolets (approved):

mm-near (approved):

PR #6292, opened 2022-02-15
feat: cache chunk nodes

Longarithm:
Extend existing trie cache to support caching nodes visiting on a chunk.

Suggested model:
* By default, we are in the `CachingShard` mode.
* On funcall execution, we temporarily enable `CachingChunk` mode.

Because all nodes touched in such mode will be charged with 16 Ggas, the chunk cache size will be limited by 500 Tgas / 16 Ggas = 31250 elements. For each such node, we charge for touching it only once - on putting it into the cache.

The current implementation does not have effect on costs, because the way of enabling `CachingChunk` state is under discussion. With enabled costs, it will save ~33% of the cost for Aurora and ~87.5% of the touching trie node cost in particular.

The difference of receipt processing times is 2%, but this can be mitigated later, see https://near.zulipchat.com/#narrow/stream/295558-nearinc.2Fcore/topic/chunk.20nodes.20cache.20prototype for more details.

## Testing

* new trie cache tests in `trie_storage`

### DIFF
--- a/core/store/src/trie/trie_storage.rs
+++ b/core/store/src/trie/trie_storage.rs
@@ -8,24 +8,31 @@ use crate::db::refcount::decode_value_with_rc;
 use crate::trie::POISONED_LOCK_ERR;
 use crate::{ColState, StorageError, Store};
 use lru::LruCache;
-use near_primitives::block::CacheState;
 use near_primitives::shard_layout::ShardUId;
+use near_primitives::types::TrieCacheState;
 use std::cell::RefCell;
 use std::io::ErrorKind;
 
 #[derive(Clone)]
 pub struct SyncTrieCache(Arc<Mutex<TrieCache>>);
 
+/// Cache for Trie nodes.
+/// Maps hash of the encoded node to the encoded node bytes.
+/// Note that "node" here means both the trie node and the value stored in trie.
 pub(crate) struct TrieCache {
-    pub(crate) cache_state: CacheState,
+    pub(crate) cache_state: TrieCacheState,
     shard_cache: LruCache<CryptoHash, Vec<u8>>,
     chunk_cache: HashMap<CryptoHash, Vec<u8>>,
 }
 
+/// Position of the value in cache.
 #[derive(Debug)]
 pub(crate) enum CachePosition {
+    /// Value is not presented.
     None,
+    /// Value is presented in the shard cache.
     ShardCache(Vec<u8>),
+    /// Value is presented in the chunk cache.
     ChunkCache(Vec<u8>),
 }
 
@@ -37,14 +44,16 @@ pub enum TrieNodeRetrievalCost {
 
 #[derive(Debug, Eq, PartialEq)]
 pub(crate) struct RawBytesWithCost {
+    /// Bytes of the retrieved node. None if no value was found in cache.
     pub(crate) value: Option<Vec<u8>>,
+    /// Cost of node retrieval.
     pub(crate) cost: TrieNodeRetrievalCost,
 }
 
 impl TrieCache {
     pub fn new(shard_cache_size: usize) -> Self {
         TrieCache {
-            cache_state: CacheState::CachingShard,
+            cache_state: TrieCacheState::CachingShard,
             shard_cache: LruCache::new(shard_cache_size),
             chunk_cache: Default::default(),
         }
@@ -66,7 +75,7 @@ impl TrieCache {
                 RawBytesWithCost { value: None, cost: TrieNodeRetrievalCost::Full }
             }
             CachePosition::ShardCache(value) => {
-                if let CacheState::CachingChunk = self.cache_state {
+                if let TrieCacheState::CachingChunk = self.cache_state {
                     let value = self
                         .shard_cache
                         .pop(key)
@@ -78,8 +87,8 @@ impl TrieCache {
             CachePosition::ChunkCache(value) => RawBytesWithCost {
                 value: Some(value),
                 cost: match self.cache_state {
-                    CacheState::CachingShard => TrieNodeRetrievalCost::Full,
-                    CacheState::CachingChunk => TrieNodeRetrievalCost::Free,
+                    TrieCacheState::CachingShard => TrieNodeRetrievalCost::Full,
+                    TrieCacheState::CachingChunk => TrieNodeRetrievalCost::Free,
                 },
             },
         }
@@ -91,7 +100,7 @@ impl TrieCache {
         }
         let value = value.to_vec();
 
-        if let CacheState::CachingChunk = &self.cache_state {
+        if let TrieCacheState::CachingChunk = &self.cache_state {
             self.shard_cache.pop(&key);
             self.chunk_cache.insert(key, value);
         } else {
@@ -118,12 +127,12 @@ impl SyncTrieCache {
 
     pub fn clear(&self) {
         let mut guard = self.0.lock().expect(POISONED_LOCK_ERR);
-        guard.cache_state = CacheState::CachingShard;
+        guard.cache_state = TrieCacheState::CachingShard;
         guard.shard_cache.clear();
         guard.chunk_cache.clear();
     }
 
-    pub fn set_state(&self, state: CacheState) {
+    pub fn set_state(&self, state: TrieCacheState) {
         let mut guard = self.0.lock().expect(POISONED_LOCK_ERR);
         guard.cache_state = state;
     }
@@ -392,13 +401,13 @@ mod tests {
         let value = vec![1u8];
         let key = hash(&value);
 
-        assert_matches!(trie_cache.cache_state, CacheState::CachingShard);
+        assert_matches!(trie_cache.cache_state, TrieCacheState::CachingShard);
         assert_matches!(trie_cache.get_cache_position(&key), CachePosition::None);
 
         trie_cache.put(key, &value);
         assert_matches!(trie_cache.get_cache_position(&key), CachePosition::ShardCache(_));
 
-        trie_cache.cache_state = CacheState::CachingChunk;
+        trie_cache.cache_state = TrieCacheState::CachingChunk;
         assert_matches!(trie_cache.get_cache_position(&key), CachePosition::ShardCache(_));
         assert_eq!(
             trie_cache.get_with_cost(&key),
@@ -424,7 +433,7 @@ mod tests {
         assert_eq!(trie_cache.shard_cache.get(&key), Some(&value));
         assert!(!trie_cache.chunk_cache.contains_key(&key));
 
-        trie_cache.cache_state = CacheState::CachingChunk;
+        trie_cache.cache_state = TrieCacheState::CachingChunk;
         let value = vec![2u8];
         trie_cache.put(key, &value);
         assert!(!trie_cache.shard_cache.contains(&key));
--- a/core/store/src/trie/trie_storage.rs
+++ b/core/store/src/trie/trie_storage.rs
@@ -8,20 +8,124 @@ use crate::db::refcount::decode_value_with_rc;
 use crate::trie::POISONED_LOCK_ERR;
 use crate::{ColState, StorageError, Store};
 use lru::LruCache;
+use near_primitives::block::CacheState;
 use near_primitives::shard_layout::ShardUId;
 use std::cell::RefCell;
 use std::io::ErrorKind;
 
 #[derive(Clone)]
-pub struct TrieCache(Arc<Mutex<LruCache<CryptoHash, Vec<u8>>>>);
+pub struct SyncTrieCache(Arc<Mutex<TrieCache>>);
+
+pub(crate) struct TrieCache {
+    pub(crate) cache_state: CacheState,
+    shard_cache: LruCache<CryptoHash, Vec<u8>>,
+    chunk_cache: HashMap<CryptoHash, Vec<u8>>,
+}
+
+#[derive(Debug)]
+pub(crate) enum CachePosition {
+    None,
+    ShardCache(Vec<u8>),
+    ChunkCache(Vec<u8>),
+}
+
+#[derive(Debug, Eq, PartialEq)]
+pub enum TrieNodeRetrievalCost {
+    Free,
+    Full,
+}
+
+#[derive(Debug, Eq, PartialEq)]
+pub(crate) struct RawBytesWithCost {
+    pub(crate) value: Option<Vec<u8>>,
+    pub(crate) cost: TrieNodeRetrievalCost,
+}
 
 impl TrieCache {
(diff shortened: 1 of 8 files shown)


core/store/src/trie/trie_storage.rs
```diff
@@ -9,34 +9,144 @@ use crate::trie::POISONED_LOCK_ERR;
...
+pub(crate) struct TrieCache {
+    pub(crate) cache_state: TrieCacheState,
+    shard_cache: LruCache<CryptoHash, Vec<u8>>,
+    chunk_cache: HashMap<CryptoHash, Vec<u8>>,
+}
+
+/// Position of the value in cache.
+#[derive(Debug)]
+pub(crate) enum CachePosition {
+    /// Value is not presented.
```
mm-near, line 32: s/presented/present

core/store/src/trie/trie_storage.rs
```diff
@@ -9,34 +9,144 @@ use crate::trie::POISONED_LOCK_ERR;
...
+    None,
+    /// Value is presented in the shard cache.
```
mm-near, line 34: s/presented/present

core/store/src/trie/trie_storage.rs
```diff
@@ -9,34 +9,144 @@ use crate::trie::POISONED_LOCK_ERR;
...
+    ShardCache(Vec<u8>),
+    /// Value is presented in the chunk cache.
```
mm-near, line 36: same as above

core/store/src/trie/trie_storage.rs
```diff
@@ -9,34 +9,144 @@ use crate::trie::POISONED_LOCK_ERR;
...
+    ChunkCache(Vec<u8>),
+}
+
+#[derive(Debug, Eq, PartialEq)]
+pub enum TrieNodeRetrievalCost {
```
mm-near, line 41: comment

core/store/src/trie/trie_storage.rs
```diff
@@ -9,34 +9,144 @@ use crate::trie::POISONED_LOCK_ERR;
...
-pub struct TrieCache(Arc<Mutex<LruCache<CryptoHash, Vec<u8>>>>);
+pub struct SyncTrieCache(Arc<Mutex<TrieCache>>);
+
+/// Cache for Trie nodes.
+/// Maps hash of the encoded node to the encoded node bytes.
+/// Note that "node" here means both the trie node and the value stored in trie.
+/// TODO (#5920): enable chunk nodes caching in Runtime::apply.
+pub(crate) struct TrieCache {
+    pub(crate) cache_state: TrieCacheState,
+    shard_cache: LruCache<CryptoHash, Vec<u8>>,
```
mm-near, line 25: later we should check how much faster would it be to use Arc<Vec<u8>> here (and 1 line below)

core/store/src/trie/trie_storage.rs
```diff
@@ -9,34 +9,144 @@ use crate::trie::POISONED_LOCK_ERR;
...
+            CachePosition::None => {
+                RawBytesWithCost { value: None, cost: TrieNodeRetrievalCost::Full }
+            }
+            CachePosition::ShardCache(value) => {
+                if let TrieCacheState::CachingChunk = self.cache_state {
+                    let value = self
+                        .shard_cache
+                        .pop(key)
+                        .expect("If position is ShardCache then value must be presented");
+                    self.chunk_cache.insert(key.clone(), value);
```
mm-near, line 84: comment what we're doing here

core/store/src/trie/trie_storage.rs
```diff
@@ -9,34 +9,144 @@ use crate::trie::POISONED_LOCK_ERR;
...
+        if value.len() >= TRIE_LIMIT_CACHED_VALUE_SIZE {
+            return;
+        }
+        let value = value.to_vec();
+
+        if let TrieCacheState::CachingChunk = &self.cache_state {
+            self.shard_cache.pop(&key);
+            self.chunk_cache.insert(key, value);
+        } else {
+            if self.chunk_cache.contains_key(&key) {
```
mm-near, line 108: comments..

core/store/src/trie/trie_storage.rs
```diff
@@ -9,34 +9,144 @@ use crate::trie::POISONED_LOCK_ERR;
...
+impl SyncTrieCache {
     pub fn new() -> Self {
-        Self(Arc::new(Mutex::new(LruCache::new(TRIE_MAX_CACHE_SIZE))))
+        Self(Arc::new(Mutex::new(TrieCache::new(TRIE_MAX_SHARD_CACHE_SIZE))))
     }
 
     pub fn clear(&self) {
-        self.0.lock().expect(POISONED_LOCK_ERR).clear()
+        let mut guard = self.0.lock().expect(POISONED_LOCK_ERR);
+        guard.cache_state = TrieCacheState::CachingShard;
```
mm-near, line 131: consider adding "clear" method to TrieCache and calling it from here

core/store/src/trie/trie_storage.rs
```diff
@@ -208,3 +333,111 @@ impl TouchedNodesCounter {
...
+        assert_eq!(
+            trie_cache.get_with_cost(&key),
+            RawBytesWithCost { value: Some(value.clone()), cost: TrieNodeRetrievalCost::Full }
+        );
+        assert_matches!(trie_cache.get_cache_position(&key), CachePosition::ChunkCache(_));
+        assert_eq!(
+            trie_cache.get_with_cost(&key),
+            RawBytesWithCost { value: Some(value), cost: TrieNodeRetrievalCost::Free }
+        );
+
```
mm-near, line 422: and now switch to CachingShard - and check that you're being charged on "get"

core/store/src/trie/trie_storage.rs
```diff
@@ -9,34 +9,144 @@ use crate::trie::POISONED_LOCK_ERR;
...
-        self.0.lock().expect(POISONED_LOCK_ERR).clear()
+        let mut guard = self.0.lock().expect(POISONED_LOCK_ERR);
+        guard.cache_state = TrieCacheState::CachingShard;
+        guard.shard_cache.clear();
+        guard.chunk_cache.clear();
+    }
+
+    pub fn set_state(&self, state: TrieCacheState) {
+        let mut guard = self.0.lock().expect(POISONED_LOCK_ERR);
+        guard.cache_state = state;
```
mm-near, line 138: and where do we have the code that "moves" stuff from chunk_cache to shard_cache when state changes ?

core/store/src/trie/trie_storage.rs
```diff
@@ -208,3 +333,111 @@ impl TouchedNodesCounter {
...
+        trie_cache.put(key, &value);
+        assert_eq!(trie_cache.shard_cache.get(&key), Some(&value));
+        assert!(!trie_cache.chunk_cache.contains_key(&key));
+
+        trie_cache.cache_state = TrieCacheState::CachingChunk;
+        let value = vec![2u8];
+        trie_cache.put(key, &value);
+        assert!(!trie_cache.shard_cache.contains(&key));
+        assert_eq!(trie_cache.chunk_cache.get(&key), Some(&value));
+    }
```
mm-near, line 442: we might have a nasty bug here...

trie.push(key, small_value)
assert trie.get(key) == small_value
trie.push(key, large_value) 
assert trie.get(key) == ??? // I think that current code will return small_value if "large_value" is > TRIE_MAX_SHARD_CACHE_SIZE

mm-near (changes requested):

matklad:
Hm, still not very happy with the shape of code here. Let me try to summarize the issues: 

* Our goal here is to implement a non-transparent caching -- a cache which effectively allows reading "hit rate" which itself affects the protocol via costs
* This is a tricky, stateful business: we have parts of code where we do want to meter cache access, and parts of code where we don't. We definitely do not want non-metered parts to affect the metered ones. 
* The fundamental owned of all the data is the `NightshadeRuntime` struct. In particular, `tries: ShardTries,` is the owner of trie caches.
* The caching logic we want to add is tricky: we can't just *layer* it on top of existing cache, we want to modify the caching logic.
  * In particular, while we *can* just layer a second cache on top, it'll lead to caching the data twice, and that's prohibitive amount of memory. We do want to "move" the cached entries between transparent and opaque caches. 
* `NightshadeRuntime` is shared across the threads. Most of the code intreacts with `NightshadeRuntime` via `RuntimeAdapter` trait, which uses `&self` (shared access everywhere). Essentially, we have one clinet and a bunch of view clients, and they all concurrently access the same `NightshadeRuntime` object, which requires `NightshadeRuntime` to be thread safe.
* It can be argued that maybe each client should have its own independent copy of `NightshadeRuntime` so that they can use single-threaded caches, and avoid synchronization, but this is a huuge change to architecture of the whole thing. 

Which together brings us to the main tech-debt problem with the implementation: the "metering state" toggle, `cache_state: TrieCacheState,` is stored in a `Sync` data structure and it is possible to *concurrently* toggle this state, which is obviously completely invalid. This not entirely theoretical concern -- in the past, we had a bug where concurrent view calls affected metering results for main client. 

<hr/>

A new thing I've realised recently is that, while `ShardTries` data structure needs to be thread-safe, as a part of `NightshadeRuntime`, the `Trie` and in particular `TrieStorage` don't need to be thread safe. In particular, the following diff works:

```diff
diff --git a/core/store/src/trie/trie_storage.rs b/core/store/src/trie/trie_storage.rs
index f302757f8..625b38265 100644
--- a/core/store/src/trie/trie_storage.rs
+++ b/core/store/src/trie/trie_storage.rs
@@ -196,15 +196,15 @@ impl TrieStorage for TrieCachingStorage {
 /// Trie increments it on every call to TrieStorage::retrieve_raw_bytes()
 #[derive(Default)]
 pub struct TouchedNodesCounter {
-    counter: AtomicU64,
+    counter: std::cell::Cell<u64>,
 }
 
 impl TouchedNodesCounter {
     pub fn increment(&self) {
-        self.counter.fetch_add(1, Ordering::SeqCst);
+        self.counter.set(self.counter.get() + 1);
     }
 
     pub fn get(&self) -> u64 {
-        self.counter.load(Ordering::SeqCst)
+        self.counter.get()
     }
 }
```

and makes the current code much less scary -- it makes it clear that it's impossible to touch nodes concurrently and actually actively prevents that via Rust's type system. (It might make sense to send that diff as a PR, as it improves today's code).

I think we can take advantage of this, and move the state-toggle from shared, thread-safe `TrieCache` to `CachingTrieStorage`, which doesn't need to be thread safe:

```diff
diff --git a/core/store/src/trie/mod.rs b/core/store/src/trie/mod.rs
index 556a85389..6a5477058 100644
--- a/core/store/src/trie/mod.rs
+++ b/core/store/src/trie/mod.rs
@@ -407,7 +407,6 @@ impl RawTrieNodeWithSize {
 
 pub struct Trie {
     pub(crate) storage: Box<dyn TrieStorage>,
-    pub counter: TouchedNodesCounter,
 }
 
 /// Stores reference count change for some key-value pair in DB.
diff --git a/core/store/src/trie/trie_storage.rs b/core/store/src/trie/trie_storage.rs
index f302757f8..737bbc6fb 100644
--- a/core/store/src/trie/trie_storage.rs
+++ b/core/store/src/trie/trie_storage.rs
@@ -135,6 +135,10 @@ pub struct TrieCachingStorage {
     pub(crate) store: Store,
     pub(crate) cache: TrieCache,
     pub(crate) shard_uid: ShardUId,
+
+    cache_state: std::cell::Cell<CacheState>,
+    chunk_cache: std::cell::RefCell<HashMap<CryptoHash, Vec<u8>>>,
+    touched_trie_nodes: std::cell::Cell<u64>,
 }
 
 impl TrieCachingStorage {
```

Differences from the current impl: 

* the `state` flag is stored in a `!Sync` data structure, so we don't have to worry about concurrent access
* touched_trie_nodes and chunk_cache are stored in the same data structure, which makes sense, as these two bits of state track the same info -- how much we should chare for trie ops

Similarities to the current impl:

* we still make use of interior mutability, just more bening, thread-unsafe version. In general, interior mutability is bad, and, if we make caching non-transparent, it is possible to argue that we should refactor call-sites to use `&mut Trie` or something. At the same time, "counting the number of times a read operation was issued" sounds like a defensible use of interior mutability. 
* accounting logic is moved from trie to trie_storage (in the PR, it is split between the two, in my suggestion, it is fully in the storage). I think this actually is a bug in both implementaitons -- if I understand the code correctly, we are going to use `TrieMemoryPartialStorage` when processing challenges, and that *should* contain the same cache metering logic. The current code (on mastre branch) doesn't have this problem, as accounting is done by the trie itself, not by the storage.

matklad:
:thinking: I've just realised that by moving entries between the two caches we distort the LRU picture, potentially making the cache less efficient.

Longarithm:
> I've just realised that by moving entries between the two caches we distort the LRU picture, potentially making the cache less efficient.

Yeah, this is a concern for us, because we want tx/receipt processing to be as fast as possible. The current implementation leaves the room for it.

> I think this actually is a bug in both implementaitons -- if I understand the code correctly, we are going to use TrieMemoryPartialStorage when processing challenges

The problem with `TrieMemoryPartialStorage` is valid. I think because we don't have challenges we can leave it as TODO for now.

The number of problems and discussions occurred with this change makes me think that it is not possible solve all them at once. I suggest to solve them sequentially, starting from this PR. What do you think? @matklad 

Another idea would be to store mutable reference to `TrieCachingStorage` in the **first** `TrieUpdate` which took `Rc<Trie>` and toggle cache state using this reference.

matklad:
> I suggest to solve them sequentially, starting from this PR. What do you think?

I am all for solving the problems sequentially, but, in this particular case, `pub fn set_state(&self, state: TrieCacheState) {`  (putting the state toggle into a shared data structure) doesn't feel like an incremental step towards the solution, if we are going to move that data out of that structure anyway. 

I'd see the incremental steps here as:

* Figure out who owns what in the end state
* Setup from https://github.com/near/nearcore/pull/6292#issuecomment-1040415712 makes most sense to me. This PRs setup really doesn't feel good to me. That's "architecture" concerns rather than "correctness" concerns, so, if people do feel that the ownership setup we have here is workable, we can ship it! 
* *If* the setup from that comment feels better, than incremental steps towards that:
  * changing TouchedTrieNodeCounter to use `Cell`, rather than `Atomic`
  * moving the counter from `Trie` to `TrieStorage` and adjusting the call-sites (I feel this might fail some tests for `RecordTrieStorage`, if we have them)
   * adding the "start counting/stop counting API" (this'll probably let us to assert that counter is zero when we start counting, to make sure we don't accidentally have "leftover" counts)
   * implementing the caching proper.

Longarithm:
Closing in favour of https://github.com/near/nearcore/pull/6414.

PR #6294, opened 2022-02-15
State-viewer uses read-only RocksDB

### LINKED ISSUE #5194: Ensure `state-viewer` is strictly read-only
Steps to reproduce:
* Run 1.21 `neard` binary
* Stop the `neard` binary
* Try to open the DB with a 1.22 `state-viewer` binary
* Restart 1.21 `neard` binary, get an error about an unknown column family `col49` in the DB.

This shouldn't happen. When investigating incidents people are prone to using the wrong versions of the binaries, which can make the incident more complicated.

nikurt:
Fix #5194

Tested by running 1.21, 1.22, 1.23 and 1.24 versions in localnet and attempting to dump their state.

Without this change, attempting to dump state can add a column (col49) which makes it impossible to dump state with state-viewer of the corresponding version.
With this change, state-viewer panics earlier and state-viewer of the corresponding version remains able to dump state:

```
Tue 14:17:50 ~/code/nearcore-rocksdb18-2 % RUST_BACKTRACE=all ./target/release/neard --home ~/.near/localnet/node0/ view_state dump_state
thread 'main' panicked at 'Failed to start Epoch Manager: IOErr(Unexpected length of input)', nearcore/src/runtime/mod.rs:198:18
stack backtrace:
   0: rust_begin_unwind
             at /rustc/db9d1b20bba1968c1ec1fc49616d4742c1725b4b/library/std/src/panicking.rs:498:5
   1: core::panicking::panic_fmt
             at /rustc/db9d1b20bba1968c1ec1fc49616d4742c1725b4b/library/core/src/panicking.rs:107:14
   2: core::result::unwrap_failed
             at /rustc/db9d1b20bba1968c1ec1fc49616d4742c1725b4b/library/core/src/result.rs:1613:5
   3: nearcore::runtime::NightshadeRuntime::new
   4: nearcore::runtime::NightshadeRuntime::with_config
   5: state_viewer::commands::load_trie_stop_at_height
   6: state_viewer::commands::dump_state
   7: state_viewer::cli::StateViewerSubCommand::run
   8: neard::cli::NeardCmd::parse_and_run
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
```

### DIFF
--- a/core/store/src/db.rs
+++ b/core/store/src/db.rs
@@ -488,6 +488,10 @@ impl Default for RocksDBOptions {
     }
 }
 
+fn col_name(col: DBCol) -> String {
+    format!("col{}", col as usize)
+}
+
 impl RocksDBOptions {
     /// Once the disk space is below the `free_disk_space_warn_threshold`, RocksDB will emit an warning message every [`interval`](RocksDBOptions::check_free_space_interval) write.
     pub fn free_disk_space_warn_threshold(mut self, warn_treshold: bytesize::ByteSize) -> Self {
@@ -524,11 +528,14 @@ impl RocksDBOptions {
 
     /// Opens a read only database.
     pub fn read_only<P: AsRef<std::path::Path>>(self, path: P) -> Result<RocksDB, DBError> {
-        let options = self.rocksdb_options.unwrap_or_default();
-        let cf_names: Vec<_> = self.cf_names.unwrap_or_else(|| vec!["col0".to_string()]);
-        let db = DB::open_cf_for_read_only(&options, path, cf_names.iter(), false)?;
-        let cfs =
-            cf_names.iter().map(|n| db.cf_handle(n).unwrap() as *const ColumnFamily).collect();
+        use strum::IntoEnumIterator;
+        let options = self.rocksdb_options.unwrap_or_else(rocksdb_options);
+        let cf_with_opts = DBCol::iter().map(|col| (col_name(col), rocksdb_column_options(col)));
+        let db = DB::open_cf_with_opts_for_read_only(&options, path, cf_with_opts, false)?;
+        let cfs = DBCol::iter()
+            .map(|col| db.cf_handle(&col_name(col)).unwrap() as *const ColumnFamily)
+            .collect();
+
         Ok(RocksDB {
             db,
             cfs,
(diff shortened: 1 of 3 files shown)


core/store/src/db.rs
```diff
@@ -524,11 +528,14 @@ impl RocksDBOptions {
...
     /// Opens a read only database.
     pub fn read_only<P: AsRef<std::path::Path>>(self, path: P) -> Result<RocksDB, DBError> {
-        let options = self.rocksdb_options.unwrap_or_default();
-        let cf_names: Vec<_> = self.cf_names.unwrap_or_else(|| vec!["col0".to_string()]);
-        let db = DB::open_cf_for_read_only(&options, path, cf_names.iter(), false)?;
-        let cfs =
-            cf_names.iter().map(|n| db.cf_handle(n).unwrap() as *const ColumnFamily).collect();
+        use strum::IntoEnumIterator;
+        let options = self.rocksdb_options.unwrap_or_else(rocksdb_options);
+        let cf_with_opts = DBCol::iter().map(|col| (col_name(col), rocksdb_column_options(col)));
```
matklad, line 533: Feels a bit strange that we do `elf.cf_names.unwrap_or_else(|| DBCol::iter().map(|col| col_name(col)).collect());` for rw but just `DBCol::iter().map(|col| (col_name(col), rocksdb_column_options(col)));` for ro

core/store/src/db.rs
```diff
@@ -524,11 +528,14 @@ impl RocksDBOptions {
...
+        let db = DB::open_cf_with_opts_for_read_only(&options, path, cf_with_opts, false)?;
```
mm-near, line 534: do we depend on 0.18 for this ?

nikurt: Yes, unfortunately.

core/store/src/db.rs
```diff
@@ -525,11 +529,14 @@ impl RocksDBOptions {
 
     /// Opens a read only database.
     pub fn read_only<P: AsRef<std::path::Path>>(self, path: P) -> Result<RocksDB, DBError> {
```
mm-near, line 531: should it be public function?

mm-near (approved):

core/store/src/db.rs
matklad: Sorry, I was unclear: I meant that it feels strange that we ignore `self.cf_names` and just go for all the columns.

PR #6298, opened 2022-02-16
pytest: fuzz: fix test not killing fuzz processes correctly

mina86:
`subprocess.call(..., timeout=...)` will send KILL signal to the
process that has been spawned once the timeout passes but that may
leave its descendants still running.  To cover that case, handle
timeout ourselves by waiting and then killing the entire process tree.

### DIFF
--- a/pytest/tests/fuzz.py
+++ b/pytest/tests/fuzz.py
@@ -12,6 +12,25 @@ import psutil
 
 
 def run(directory: str, fuzz_target: str, timeout=typing.Optional[int]) -> int:
+    """Executes `cargo fuzz run` on given target in given directory.
+
+    Args:
+        directory: Directory within the repository to execute the `cargo fuzz`
+            command in.
+        fuzz_target: The `cargo fuzz run` target.
+        timeout: Optional timeout in seconds denoting maximum time the fuzzer
+            can run.  If given, the fuzzer will be killed after given time
+            passes and run is considered a success, i.e. function returns 0.
+
+    Returns:
+        The exit status of the fuzzer except that if timeout is specified and
+        fuzzer runs over allotted interval it is killed and the function returns
+        zero denoting success.  Similarly, if script is interrupted with Ctrl+C,
+        the fuzzer is killed and run is considered a success.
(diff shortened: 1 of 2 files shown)


Ekleog (approved):
Makes sense, I hadn't thought of `psutil`. An alternative would be to start the subprocess with `start_new_session=True` and then call `killpg(getpgid(proc.pid))` similar to what was done in nayduck#28.

mina86:
`psutil` is what NayDuck was using when I took it over and I never changed it even though I think it is a more fragile method.  A test can escape process tree accidentally by doing a double fork while escaping a process group needs more deliberate action.  Nonetheless, so far we didn’t have issue with killing process tree so for now it’s probably good enough.

Ekleog (approved):

pytest/tests/fuzz.py
```diff
@@ -2,27 +2,64 @@
...
+import signal
 import subprocess
 import sys
 import typing
 
+import psutil
 
-def run(directory: str, fuzz_target: str) -> int:
+
+def run(directory: str, fuzz_target: str, timeout=typing.Optional[int]) -> int:
```
mm-near, line 14: what is the unit for timeout ?

pytest/tests/fuzz.py
```diff
@@ -2,27 +2,64 @@
...
+        except psutil.NoSuchProcess:
+            pass
 
 
 def _get_timeout() -> typing.Optional[int]:
+    """Returns test timeout configured on NayDuck
+
+    When run on NayDuck, returns timeout that NayDuck gives us to execute.  If
+    we run over that time, the test will be reported as timed out.  When not
+    running on NayDuck, returns None denoting no timeout.
```
mm-near, line 61: what is the unit of the response?

pytest/tests/fuzz.py
mm-near, line 14: also mention was it the response

mm-near (changes requested):

mm-near (approved):

Issue #6270, opened 2022-02-08
Optimize unnecessary clones in `retrieve_raw_bytes`


Longarithm:
It seems that `retrieve_raw_bytes` can return `&Vec<u8>` or `&<span class="error">[u8]</span>` and avoid cloning, because in many places we even don't use the result or use just a reference: [https://github.com/near/nearcore/blob/f158f333daca1f34d41a2dc44b7d9ad04766e110/core/store/src/trie/mod.rs#L662](https://github.com/near/nearcore/blob/f158f333daca1f34d41a2dc44b7d9ad04766e110/core/store/src/trie/mod.rs#L662)

We know that the bytes stay in cache for the usage lifetime, so we could optimize unnecesary clones and see if this optimizes performance.

PR #6303, opened 2022-02-16
feat: replace `Vec<u8>` with `Arc<[u8]>` in trie cache

Longarithm:
https://github.com/near/nearcore/issues/6270

Currently, `retrieve_raw_bytes` returns `Vec<u8>`, but in fact we don't need it, and sometimes it leads to unnecessary clones. Here we replace it with `Arc<[u8]>`. We can't use `&[u8]` at least for now, because trie cache may be used by different clients (i.e. view clients).

## Benchmarking

I tried two approaches to benchmark this:

### Running state-viewer `apply_range` on Aurora shard

I couldn't see significant performance improvement here.

### Running near-store `trie_lookup` benchmark. 

The difference is still negligible, though the version in PR always gave smaller time.
However, in current bench implementation
* keys are random `Vec<u8>`s, so the number of intermediate visited nodes should be small relatively to the number of values;
* perhaps on the long distance, running bench with the same set of kv pairs gives the same results.

So I modified benchmark https://github.com/near/nearcore/compare/master...trie-bench and looked at the first time estimations:

```
master

test trie_lookup ... took 13.026505ms
took 12.420749ms
took 11.799348ms
took 11.420096ms
took 11.310058ms
took 12.060703ms
took 11.923342ms
took 11.418761ms
took 11.250021ms
took 12.76402ms
took 11.329514ms
took 11.303166ms

cache-arc-test

test trie_lookup ... took 11.316463ms
took 10.719466ms
took 10.952066ms
took 10.971129ms
took 10.921156ms
took 10.979292ms
took 11.390993ms
took 12.171543ms
took 10.805069ms
took 10.880024ms
took 11.095865ms
took 10.831956ms
```

I think this is enough to prove the usefulness of the change. Also this should help to extend TrieCache with new chunk cache.
## Testing

Existing tests.

### DIFF
--- a/core/store/src/trie/mod.rs
+++ b/core/store/src/trie/mod.rs
@@ -3,6 +3,7 @@ use std::cmp::Ordering;
 use std::collections::HashMap;
 use std::fmt;
 use std::io::{Cursor, Read, Write};
+use std::sync::Arc;
 
 use borsh::{BorshDeserialize, BorshSerialize};
 use byteorder::{LittleEndian, ReadBytesExt, WriteBytesExt};
@@ -611,7 +612,10 @@ impl Trie {
         }
     }
 
-    pub(crate) fn retrieve_raw_bytes(&self, hash: &CryptoHash) -> Result<Vec<u8>, StorageError> {
+    pub(crate) fn retrieve_raw_bytes(
+        &self,
+        hash: &CryptoHash,
+    ) -> Result<Arc<Vec<u8>>, StorageError> {
         self.counter.increment();
         self.storage.retrieve_raw_bytes(hash)
     }
@@ -624,7 +628,7 @@ impl Trie {
         match RawTrieNodeWithSize::decode(&data) {
             Ok(value) => {
                 let memory_usage = TrieNodeWithSize::from_raw(value).memory_usage;
-                Ok(StateRootNode { data, memory_usage })
+                Ok(StateRootNode { data: data.to_vec(), memory_usage })
             }
             Err(_) => Err(StorageError::StorageInconsistentState(format!(
                 "Failed to decode node {}",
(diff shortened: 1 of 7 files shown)


core/store/src/trie/trie_storage.rs
```diff
@@ -176,6 +176,7 @@ impl TrieStorage for TrieCachingStorage {
                 .get(ColState, key.as_ref())
                 .map_err(|_| StorageError::StorageInternalError)?;
             if let Some(val) = val {
+                let val: Arc<[u8]> = val.into();
```
matklad, line 179: In the grand scheme on things, this copy is avoidable. Here's what happens here:

The data lives in memory, in some internal rocks datastructure. We then copy the data from rocks into a Rust Vec (there's an alternative API to get an equivalent of `[u8]`, but we don't use that). And then we copy data from Vec into Arc. 

The optimization here would be to elide intermediete copy, and to copy data from rocks directly into an Arc.

I don't think we need to do this in the PR, but it's something to keep in mind when thinking about storage.

core/store/src/trie/mod.rs
```diff
@@ -699,7 +700,9 @@ impl Trie {
 
     pub fn get(&self, root: &CryptoHash, key: &[u8]) -> Result<Option<Vec<u8>>, StorageError> {
```
matklad, line 701: Unrelated to the PR, but we should evaluate whether returning an `Arc` rather than `Vec` makes sense here. Could you create the issue for that if/when this PR is merged?

matklad (approved):
It's useful to take a broader look at what happens here:

Before this PR, we always interacted with `Vec<u8>`. That is, anyone who wanted to read a database record, got a freshly allocated `Vec<u8>`. This fundamentally doesn't play well with the idea of a concurrent cache -- if we want several threads to interact with the same DB record, we need some sort of a reference count somewhere to figure out when we can free the data. 

And that's exactly what this PR does -- now a db record is not a `Vec<u8>`, but has a reference count attached. So you can, eg, put the data in the concurrent cache.


We might want to have *more* of this, and even change the low-level DB API to directly return ref-counted objects, and change callers to take advantage of that. If we go that way, we probably should evaluate using `Bytes` over `Arc<[u8]>`, as that is more optimized for the use case.

core/store/src/trie/mod.rs
mm-near: I guess that the same question applies to the "iterator.rs" part - as we also keep returning a Vec<> in the items there.

mm-near (approved):

mzhangmzz (approved):

PR #6305, opened 2022-02-16
client: add rustc_version to --version page and /status response

mina86:
Fixes: https://github.com/near/nearcore/issues/5979

### DIFF
--- a/core/primitives/src/version.rs
+++ b/core/primitives/src/version.rs
@@ -8,6 +8,8 @@ use crate::types::Balance;
 pub struct Version {
     pub version: String,
     pub build: String,
+    #[serde(default)]
+    pub rustc_version: String,
 }
 
 /// Database version.
--- a/neard/build.rs
+++ b/neard/build.rs
@@ -9,6 +9,9 @@
 use anyhow::{anyhow, Result};
 
 /// Returns value of given environment variable or error if missing.
+///
+/// This also outputs necessary ‘cargo:rerun-if-env-changed’ tag to make sure
+/// build script is rerun if the environment variable changes.
 fn env(key: &str) -> Result<std::ffi::OsString> {
     println!("cargo:rerun-if-env-changed={}", key);
     std::env::var_os(key).ok_or_else(|| anyhow!("missing ‘{}’ environment variable", key))
@@ -76,6 +79,9 @@ fn try_main() -> Result<()> {
     println!("cargo:rustc-env=NEARD_VERSION={}", version);
 
     println!("cargo:rustc-env=NEARD_BUILD={}", get_git_version()?);
+
+    println!("cargo:rerun-if-env-changed=RUSTC");
+    println!("cargo:rustc-env=NEARD_RUSTC_VERSION={}", rustc_version::version()?);
 
     Ok(())
 }
--- a/neard/build.rs
+++ b/neard/build.rs
@@ -80,7 +80,6 @@ fn try_main() -> Result<()> {
 
     println!("cargo:rustc-env=NEARD_BUILD={}", get_git_version()?);
 
-    println!("cargo:rerun-if-env-changed=RUSTC");
     println!("cargo:rustc-env=NEARD_RUSTC_VERSION={}", rustc_version::version()?);
 
     Ok(())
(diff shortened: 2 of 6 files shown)


neard/build.rs
```diff
@@ -8,27 +8,58 @@
...
 fn env(key: &str) -> Result<std::ffi::OsString> {
     println!("cargo:rerun-if-env-changed={}", key);
     std::env::var_os(key).ok_or_else(|| anyhow!("missing ‘{}’ environment variable", key))
 }
 
+/// Returns value of given environment variable or given default value.
+///
+/// This also outputs necessary ‘cargo:rerun-if-env-changed’ tag to make sure
+/// build script is rerun if the environment variable changes.
+fn env_or(key: &str, default: impl std::convert::Into<std::ffi::OsString>) -> std::ffi::OsString {
```
matklad, line 38:
```suggestion
fn env_or(key: &str, default: impl Into<std::ffi::OsString>) -> std::ffi::OsString {
```

`Into` in in prelude

neard/build.rs
```diff
@@ -8,27 +8,58 @@
...
+    println!("cargo:rerun-if-env-changed={}", key);
+    std::env::var_os(key).unwrap_or_else(|| default.into())
```
matklad, line 40: Very weak opinion, but using `var_os` (over `var`) to make code pedantically utf8-correct probably isn't worth it. Even `cargo` sometimes just panics on invalid utf8, and I don't think anyone complained in practice about that: 

```console
[matklad@Ishmael:~/projects/nearcore]$  cargo build --manifest-path $(echo -e '\xFF')
thread 'main' panicked at 'unexpected invalid UTF-8 code point', /cargo/registry/src/github.com-1ecc6299db9ec823/clap-2.33.3/src/args/arg_matches.rs:118:40
```

core/primitives/src/version.rs
```diff
@@ -8,6 +8,7 @@ use crate::types::Balance;
 pub struct Version {
     pub version: String,
     pub build: String,
+    pub rustc_version: String,
```
matklad, line 11: I always wonder if it is correct to add fields to `derive(Deserialize)` structs. If this is used by, eg, indexer, then the indexer would fail to deserialize responses from older nodes.

neard/build.rs
```diff
@@ -45,16 +76,28 @@ fn get_git_version() -> Result<String> {
...
             Ok("unknown".to_string())
-        }
-    }
+        },
+    )
+}
+
+/// Returns rustc version.
+fn get_rustc_version() -> Result<String> {
+    command(&env_or("RUSTC", "rustc"), &["-vV"])
```
matklad, line 90: Why do we want to run command by hand, if we already depend on `rustc_version` and can use [rustc_version](https://docs.rs/rustc_version/latest/rustc_version/index.html)::[version_meta](https://docs.rs/rustc_version/latest/rustc_version/fn.version_meta.html#)?

neard/build.rs
```diff
@@ -45,16 +76,28 @@ fn get_git_version() -> Result<String> {
...
+        .and_then(|output| Ok(rustc_version::version_meta_for(&output)?))
+        .map(|meta| {
+            let channel = match meta.channel {
+                rustc_version::Channel::Dev => " dev",
+                rustc_version::Channel::Nightly => " nightly",
+                rustc_version::Channel::Beta => " beta",
+                rustc_version::Channel::Stable => "",
+            };
+            format!("{}{}", meta.semver, channel)
+        })
```
matklad, line 100:
```suggestion
        .map(|meta| meta.semver.to_string())
```

Semver already includes channel info: `1.59.0-beta.1`

matklad, line 100: Minor stylistic preferences, but I'd use `?` over `.and_then` and `.map` here -- closures add to cognitive load

neard/build.rs
mina86: I think the three lines that we need to handle OsStr are kinda ok cost for the complication. ;) The code has mechanism for bubbling errors so we might just as well uses them IMO.

neard/build.rs
```diff
@@ -77,5 +80,8 @@ fn try_main() -> Result<()> {
 
     println!("cargo:rustc-env=NEARD_BUILD={}", get_git_version()?);
 
+    println!("cargo:rerun-if-env-changed=RUSTC");
```
matklad, line 83: This I think isn't necessary -- Cargo already has the logic to re-run the build if anything about the compiler changes. Though, this doesn't hurt either.

matklad (approved):

mm-near (approved):

mina86:
@matklad could you take a look at the deny.toml change? We have some transitive dependencies which have old rustc_version in their build-dependiencies.

PR #6307, opened 2022-02-17
End-to-end test for canaries

nikurt:


pytest/lib/account.py
```diff
@@ -62,9 +63,11 @@ def prep_tx(self):
         self.tx_timestamps.append(time.time())
         self.nonce += 1
 
-    def send_transfer_tx(self, dest_account_id, base_block_hash=None):
+    def send_transfer_tx(self,
+                         dest_account_id,
+                         transfer_amount=100,
+                         base_block_hash=None):
```
mina86, line 69: Perhaps

    def send_transfer_tx(self,
                         dest_account_id,
                         *,
                         transfer_amount=100,
                         base_block_hash=None):

to make sure existing callers don’t pass base block hash as transfer
amount?  This will force using keyword arguments.

pytest/lib/mocknet_helpers.py
```diff
@@ -9,34 +9,40 @@
...
+        'account_id': account_id,
+        'finality': finality
+    },
+                           addr=addr,
+                           port=port)
     logger.info(f'get_nonce_for_pk {account_id}')
     assert access_keys['result']['keys'], account_id
     for k in access_keys['result']['keys']:
         if k['public_key'] == pk:
             return k['access_key']['nonce']
```
mina86, line 40: Not related to this PR, but shouldn’t this be something like:

    nonce = next((key['access_key']['nonce']
                  for key in access_keys['result']['keys']
                  if key['public_key'] == pk), None)
    if nonce is None:
        raise KeyError(f'Nonce for {account_id} {pk} not found')
    return nonce

pytest/tests/endtoend/test.py
```diff
@@ -0,0 +1,114 @@
...
+    --accounts <account_node1,account_node2>
+    --interval-sec 60
+    --port 3030
+    --public-key <public_key of test accounts>
+    --private-key <private key of test account>
+    --rpc-server-addr rpc.testnet.near.org
+    --rpc-server-port 80
+    --metrics-port 3040
+"""
+#!/usr/bin/env python3
```
mina86, line 19: This needs to be the first line of the file.

pytest/tests/endtoend/test.py
```diff
@@ -0,0 +1,114 @@
...
+import account as account_mod
+import key as key_mod
+import mocknet
+import mocknet_helpers
+from concurrent.futures import ThreadPoolExecutor
+
+from configured_logger import logger
+import prometheus_client
+
+balance_histogram = prometheus_client.Gauge(
```
mina86, line 37: That’s not really a histogram. It’s just a gauge.

mina86 (approved):

pytest/lib/account.py
nikurt: > existing callers

Checked that none of the existing callers get affected by this change.

pytest/lib/mocknet_helpers.py
mina86: I meant that the previous loop is now unnecessary with this, i.e.:

```python
def get_nonce_for_pk(account_id,
                     pk,
                     finality='optimistic',
                     addr=LOCAL_ADDR,
                     port=RPC_PORT):
    request = {
        'request_type': 'view_access_key_list',
        'account_id': account_id,
        'finality': finality
    }
    access_keys = json_rpc('query', request, addr=addr, port=port)
    logger.info(f'get_nonce_for_pk {account_id}')
    nonce = next((key['access_key']['nonce']
                  for key in access_keys['result']['keys']
                  if key['public_key'] == pk), None)
    if nonce is None:
        raise KeyError(f'Nonce for {account_id} {pk} not found')
    return nonce
```

Or perhaps I don’t understand the point of the `assert`?  Current code
checks if account has keys and then looks for one matching the
argument.  If the account has no keys than it raises AssertionError
but if it has keys but no key matches given one it returns None.

It felt to me like the intention was to raise an exception when the
nounce cannot be found?  Which is what my proposed code does.

nikurt: Thanks

PR #6309, opened 2022-02-17
NEP205: Delay voting for new version until a certain date and time

nikurt:
See https://github.com/near/NEPs/issues/205
This PR enables clients to extend the protocol upgrade window, but indirectly. This PR lets clients upgrade early but not announce the fact of the upgrade until a certain date in the future.

Normal upgrades:
* Clients will delay voting until a certain date.

Emergency upgrades:
* Voting for the new version starts immediately and a protocol upgrade becomes effective in 1-2 epochs.


core/primitives/src/upgrade_schedule.rs
```diff
@@ -0,0 +1,143 @@
...
+
+pub(crate) fn get_protocol_version_internal(
+    next_epoch_protocol_version: ProtocolVersion,
+    client_protocol_version: ProtocolVersion,
+    schedule: Option<ProtocolUpgradeVotingSchedule>,
+) -> ProtocolVersion {
+    if next_epoch_protocol_version >= client_protocol_version {
+        client_protocol_version
+    } else if let Some(voting_start) = schedule {
+        if chrono::Utc::now() < voting_start.get() {
```
matklad, line 28: `chorono::Utc::now` potentially can go backwards, in effect a node can change its mind whether it supports the new protocol version or not. A somewhat more robust approach in theory would be to persist the decision to upgrade to the database. Or, perhaps even better, to base the decision on the the timestamp of the last block.

core/primitives/src/upgrade_schedule.rs
```diff
@@ -0,0 +1,143 @@
+use chrono::{DateTime, NaiveDateTime, ParseResult, Utc};
+use near_primitives_core::types::ProtocolVersion;
+
+pub(crate) struct ProtocolUpgradeVotingSchedule(pub &'static str);
+impl ProtocolUpgradeVotingSchedule {
```
bowenwang1996, line 5: formatting question: shouldn't there be a new line between these two lines?

core/primitives/src/version.rs
```diff
@@ -164,6 +164,29 @@ pub const PROTOCOL_VERSION: ProtocolVersion = MAIN_NET_PROTOCOL_VERSION;
 #[cfg(feature = "nightly_protocol")]
 pub const PROTOCOL_VERSION: ProtocolVersion = 126;
 
+#[allow(dead_code)]
+const MAIN_NET_PROTOCOL_UPGRADE_VOTING_START: Option<ProtocolUpgradeVotingSchedule> =
+    Some(ProtocolUpgradeVotingSchedule("2022-02-17 12:44:00"));
```
bowenwang1996, line 169: what is this hardcoded number?

core/primitives/src/version.rs
```diff
@@ -164,6 +164,29 @@ pub const PROTOCOL_VERSION: ProtocolVersion = MAIN_NET_PROTOCOL_VERSION;
 #[cfg(feature = "nightly_protocol")]
 pub const PROTOCOL_VERSION: ProtocolVersion = 126;
 
+#[allow(dead_code)]
+const MAIN_NET_PROTOCOL_UPGRADE_VOTING_START: Option<ProtocolUpgradeVotingSchedule> =
```
bowenwang1996, line 168: this will also be used for testnet right? If so, shouldn't we just name it `STABLE_PROTOCOL_UPGRADE_VOTING_START`?

bowenwang1996 (commented):
Do we plan to manually set the upgrade time in code every time we do a new release for both testnet and mainnet? If so, how could we make sure that we don't forget to do it?

core/primitives/src/version.rs
nikurt: Arbitrary point in time used for testing. Removed.

core/primitives/src/version.rs
bowenwang1996: Let's change everything to `STABLE_PROTOCOL_` then

bowenwang1996:
@nikurt please answer my question above https://github.com/near/nearcore/pull/6309#pullrequestreview-889165958

nikurt:
@bowenwang1996 sorry for missing the question.
Yes, we'll have to manually update it every time.
Also we can't assume that this date is always set, because emergency releases should leave the schedule empty.
I'm struggling to find a good way to automate the check whether the field was intentionally updated or intentionally not updated (multiple emergency releases in a row).

Every release requires a manual update of version in two `Cargo.toml` files. Maybe if we somehow move the schedule definition to `Cargo.toml` it will make the human error less likely? However it doesn't exclude the human error completely.

@abacabadabacaba do you have suggestions?

mzhangmzz:
> @bowenwang1996 sorry for missing the question. Yes, we'll have to manually update it every time. Also we can't assume that this date is always set, because emergency releases should leave the schedule empty. I'm struggling to find a good way to automate the check whether the field was intentionally updated or intentionally not updated (multiple emergency releases in a row).
> 
> Every release requires a manual update of version in two `Cargo.toml` files. Maybe if we somehow move the schedule definition to `Cargo.toml` it will make the human error less likely? However it doesn't exclude the human error completely.
> 
> @abacabadabacaba do you have suggestions?

Instead of only saving the current update time, how about store a dictionary from protocol versions to update times, so each time we upgrade, we have to add a new entry to the dictionary. For emergency update, the entry value will be None, but there still must be an entry.

core/primitives/src/upgrade_schedule.rs
```diff
@@ -0,0 +1,150 @@
+use chrono::{DateTime, NaiveDateTime, ParseResult, Utc};
+use near_primitives_core::types::ProtocolVersion;
+
+/// Defines the point in time after which validators are expected to vote on the new protocol version.
+pub(crate) struct ProtocolUpgradeVotingSchedule(pub &'static str);
```
mm-near, line 5: I'm surprised that this is a str.

how about we keep ti as DateTime inside and implement the FromStr trait? 
(then you don't need things like 'is_valid')

core/primitives/src/upgrade_schedule.rs
```diff
@@ -0,0 +1,150 @@
...
+    pub fn get(self) -> DateTime<Utc> {
+        DateTime::<Utc>::from_utc(self.parse().unwrap(), Utc)
+    }
+
+    fn parse(self) -> ParseResult<NaiveDateTime> {
+        NaiveDateTime::parse_from_str(&self.0, "%Y-%m-%d %H:%M:%S")
+    }
+}
+
+pub(crate) fn get_protocol_version_internal(
```
mm-near, line 22: add comments on what these arguments are.

core/primitives/src/version.rs
```diff
@@ -152,20 +152,46 @@ pub enum ProtocolFeature {
...
-pub const PROTOCOL_VERSION: ProtocolVersion = MAIN_NET_PROTOCOL_VERSION;
+pub const PROTOCOL_VERSION: ProtocolVersion = STABLE_PROTOCOL_VERSION;
 /// Current latest nightly version of the protocol.
 #[cfg(feature = "nightly_protocol")]
 pub const PROTOCOL_VERSION: ProtocolVersion = 126;
 
+/// The moment in time after which the voting for the latest client version should start on the
+/// mainnet version of the protocol.
+#[allow(dead_code)]
+const STABLE_PROTOCOL_UPGRADE_VOTING_START: Option<ProtocolUpgradeVotingSchedule> = None;
```
mm-near, line 172: so before each release, we should set some value here, right? 

(if yes - we should add more comments).

(and also I like the idea of this being more 'explicit' - by using the map from protocol version to the expected upgrade_voting_start)

mm-near (changes requested):

core/primitives/src/upgrade_schedule.rs
nikurt: FromStr doesn't work as a const expression.
FromStr actually gives us a simpler interface but needs non-const initialization, PTAL.

core/primitives/src/version.rs
nikurt: > so before each release, we should set some value here, right?
Yes, if we want the protocol upgrade to be delayed.

Implemented. I'm slightly concerned that this dictionary needs to be populated in the release branch, and thus will only contain at most one value.

mm-near (approved):

bowenwang1996:
@nikurt could you document this change in go/nearcore-release ?

nikurt:
@bowenwang1996 documented

PR #6313, opened 2022-02-17
fuzz: introduce fuzzing configuration

Ekleog:


mina86:
I’ve already updated this in the branch. In the future this will need to happen as part of cutting a release I guess.

PR #6315, opened 2022-02-17
refactor: move nodes counter to the caching storage

Longarithm:
Work on https://github.com/near/nearcore/pull/6241 showed the following issues with trie counter:
* it's implemented as atomic, but using it in different threads doesn't make sense;
* node accounting logic makes more sense nearly to caching, because we use it only on `TrieCachingStorage` and the chunk nodes cache feature will impact it as well.

So here we move counter to the `TrieCachingStorage` and remove some unnecessary counter usages.

## Testing

Add currently missing tests:
* for counting touched nodes and trie cache size;
* same but for repeated values.


core/store/src/trie/mod.rs
```diff
@@ -577,7 +573,6 @@ impl Trie {
         if *hash == Trie::empty_root() {
             Ok(memory.store(TrieNodeWithSize::empty()))
         } else {
-            self.counter.increment();
```
matklad, line 580: Do we have any protocol changes here? In particular, I see there's `self.storage.retrieve_raw_bytes(hash)?;` in `delete_value`, and that now would increment the counter.

core/store/src/trie/mod.rs
```diff
@@ -612,7 +607,6 @@ impl Trie {
     }
 
     pub(crate) fn retrieve_raw_bytes(&self, hash: &CryptoHash) -> Result<Vec<u8>, StorageError> {
-        self.counter.increment();
         self.storage.retrieve_raw_bytes(hash)
     }
```
matklad, line 611: Can just remove this function now

core/store/src/trie/trie_storage.rs
```diff
@@ -59,6 +58,10 @@ pub trait TrieStorage {
     fn as_partial_storage(&self) -> Option<&TrieMemoryPartialStorage> {
         None
     }
+
+    fn get_touched_nodes_count(&self) -> u64 {
+        unimplemented!();
+    }
```
matklad, line 64:
```suggestion
    fn get_touched_nodes_count(&self) -> u64;
```

Longarithm: @matklad why? This method doesn't make sense for impls except `TrieCachingStorage` and `TrieRecordingStorage`, so if we incorrectly call it for some other impl, I prefer to see general unimplemented error instead of implementing it for each `TrieStorage` impl separately.

core/store/src/trie/mod.rs
Longarithm: Great question!

From what I see, all logic leading to incrementing counter here happens in `Trie::update` which itself is called after all tx/receipt execution. I feel that some counter increments makes us think that we charge for them somewhere, although we don't do it in reality. So it's better to remove them. But I also will be happy if someone double-check this.

core/store/src/trie/trie_storage.rs
Longarithm: However, I filed an issue that this counter should work with recording storage as well https://github.com/near/nearcore/issues/6316

core/store/src/trie/mod.rs
bowenwang1996: Is there a way to test this?

bowenwang1996 (commented):
@EgorKulikov @mzhangmzz please review the PR

core/store/src/trie/mod.rs
mzhangmzz: @Longarithm I still don't understand why the change in `delete_value` won't cause any change in gas counting. Could you elaborate more on that?

One way to test this manually, is maybe run state viewer apply chain on some mainnet blocks to verify that gas counting doesn't change.

mzhangmzz: I think `delete_value` will change here, because previously it only calls `self.storage.retrieve_raw_bytes(hash)?` without incrementing the counter. After your change, since you moved the counter incrementing logic inside `storage.retrieve_raw_bytes`, `delete_value` will increase the counter. 

I'm not super familiar with the gas counting logic, so I'm not sure why it is called in `Trie::update` won't affect gas counting.

Longarithm: Yeah, you are right.

> I'm not sure why it is called in Trie::update won't affect gas counting.

I've looked over all `Trie::update` calls. It is called for four purposes:
* tests
* splitting states (`add_values_to_split_states_impl`)
* applying genesis (`finalize_genesis`)
* in `TrieUpdate::finalize`

The latter function is called for similar purposes - I think it doesn't make much sense to list all them. Also it is called in the end of `Runtime::apply` to apply changes to trie, when all tx/receipts are already executed and burnt/used gas is computed.

Moreover, trie counter is used for gas counting only during function call execution, which is triggered only by `process_receipt`, this is another reason why `Trie::update` change doesn't affect it.

Sorry I didn't explain it before properly. On my side, it was a fast look over CLion usages, so I thought it didn't worth attention.

mzhangmzz (approved):
Will approve to unblock. I suggest that as a final test, run state-viewer apply-chain on a mainnet node to make sure the gas counting logic doesn't change. 

(I'm convinced that it didn't. But this part of logic is quite complicated and I just want to be sure)

core/store/src/trie/trie_storage.rs
matklad: No strong opinion here, it's just that `unimplemented!` as a default impl feels like a smell.

`unimplemented!()` in a trait impl is a smell -- if someone could be unimplemented, it's generally better to split a trait in two. I don't think we can easy make this change here (and I'd probably argue that an impl which returns 0 would be more correct), but we should make this `unimplemented!()` stand out more by explicitly writing this for every type where this method doesn't make sense.

matklad (approved):

core/store/src/trie/trie_storage.rs
mm-near: +1 to moving the 'unimplemented' part into each of the implementation classes.

core/store/src/trie/mod.rs
```diff
@@ -612,16 +606,11 @@ impl Trie {
         }
     }
 
-    pub(crate) fn retrieve_raw_bytes(&self, hash: &CryptoHash) -> Result<Arc<[u8]>, StorageError> {
```
mm-near, line 615: @matklad - more Rust question:

in such case - should we 'keep' this method (and have it simply call self.storage.retrieve_raw_bytes) or should we allow other classes to call XX.storage.retrieve_raw_bytes (as Alex did in this PR)?

For me - one advantage of keeping this method - is that we keep better 'layers' - and avoid having other classes 'peek' into Trie) - but I wonder if it has some performance implications

Longarithm:
@mzhangmzz I run state-viewer on blocks `[55290778, 55290878)` on Aurora shard - sequence from the end of my DB snapshot - showed full equality of both burnt_gas and touching_trie_node_cost for all receipts. Same for shard 0.

I attach the logs for Aurora shard:
[logs.zip](https://github.com/near/nearcore/files/8218244/logs.zip)

core/store/src/trie/mod.rs
matklad: There won't be performance implications here. In general, it's a pretty safe assumption in Rust that linguistic abstractions don't introduce overhead -- static nature of the language and unboxed memory layout guarantee that compiler "sees through" any abstractions. 

Regarding layering, my general observation is that layering is good only when it is enforced: it's better to focus on specific places where layering is important, rather than to try to enforce it recursively throughout the stack. There should be places where there's explicitly *no* layers.

For Rust, the best enforcement is aligning layers with crate boundaries: that's why I am relatively picky about `pub` vs `pub(crate)`, but relatively indifferent between `pub(crate)` and `/*private*/`.

As an aside, I think we can make storage private rather `pub(crate)` if we put trie tests into submodule of the `trie` module. We only need `pub(crate)` for tests, but "`pub(crate)` for tests" is a bit of oxymoron -- `pub(crate)` guarantees that this is for *unit* tests, and unit tests can always be made a submodule (in Rust, submodule see the full contents of partent modules)

PR #6318, opened 2022-02-18
Log to WARN information about invalid chunk state and invalid challenges

nikurt:
This could help us investigate Issue #6317

### DIFF
--- a/chain/chain/src/chain.rs
+++ b/chain/chain/src/chain.rs
@@ -1015,7 +1015,7 @@ impl Chain {
         ) {
             Ok(_) => {}
             Err(err) => {
-                warn!(target: "chain", "Invalid challenge: {}", err);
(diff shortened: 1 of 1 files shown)


chain/chain/src/chain.rs
```diff
@@ -3626,7 +3626,7 @@ impl<'a> ChainUpdate<'a> {
                         chunk_header,
                     )
                     .map_err(|e| {
-                        debug!(target: "chain", "Failed to validate chunk extra: {:?}", e);
+                        warn!(target: "chain", "Failed to validate chunk extra: {:?}", e);
```
mm-near, line 3629: let's also mention the block hash + chunk hash (or shard id)

mm-near (approved):

PR #6324, opened 2022-02-21
feat(sandbox): add debug log host function to sandbox

ailisp:
We need a debug print host function that prints arbitrary length during contract development. `env.log_utf8` would log to execution outcome and has 16K? size limit. The most appropriate place for this feature would be in sandbox node only. Then on SDK side, debug log can be conditionally compiled to call debug host function on sandbox, and noop on production.

Test Plan
---------
`make sandbox`, export debug log for near-sdk-js. Write a smart contract in JS that calls debug log:
```
export function test_debug_log() {
    debug.log('this will be in sandbox log')
}
```
As a result I can see this line in sandbox log:
```
Feb 21 10:31:04.238 DEBUG sandbox: this will be in sandbox log    
```

And existing CI tests.


matklad:
Who is the target audience for this feature? Is this to make the life of sandbox developers easier, or is this intended as a help for contract developers? In general, it'll be useful to understand what kind of a problem we are solving here -- at the first glance, using `log!` directly feels like a very hacky approach.

ailisp:
@matklad It's intended as a help for contract developers. 

> at the first glance, using log! directly feels like a very hacky approach.
This is true given it is some method in sandbox access host io. However, it's limited to only sandbox node. As we know sandbox has more risky operations like patch state. If your concern is the use of `log`, it's a choice we made from a long discussion. The user story is contract developer will always use near-workspaces, to test their contracts. A sandbox node is launched and controlled by near-workspaces. Near-workspaces has pipe and can filter sandbox's stdout. Then our log with topic can help solve "capture debug log from sandbox" and show it in workspace. There can be more complex solution like send debug log over socket, but we decide that's not necessary as sandbox is always a local node. cc @MaximusHaximus

matklad:
>If your concern is the use of log, it's a choice we made from a long discussion.

Yeah, this is the point I am surprised about. I see three problems with log:

* Fundamentally, it's a kludge and uses the wrong tool for the job. This will probably create unspecific problems down the line -- for example, if we change the format of logging messages, the parsing code in the workspaces would break. 
* Our log is unstructured, so the parsing would have to be based on heuristics. If the user prints a multiline message using `{:#?}`, the workspaces would have to do extra gymnastics to read this out.
* Due to asynchronous nature of NEAR, a single workspaces API call can lead to the execution of serval contract. On the workspaces side, it would be hard to correlate log messages with specific contract calls and decide which function call printed which message. 

Naively, I'd expect debug-logs to be added as a field to the `ExecutionMetadata`, and fetched by the RPC API. We *already* have a communication channel to communicate normal and "debug" (profile data) execution results, it seems we can use the same for debug log? 

I don't necessary want to re-iterate design discussion here -- if dev-platform is sure that literal logging is the way to go here, it's fine by me. In particular, as this is a private API to use by sandbox, we can always change it later.

runtime/near-vm-logic/Cargo.toml
```diff
@@ -31,6 +31,7 @@ near-primitives-core = { path = "../../core/primitives-core" }
 near-vm-errors = { path = "../near-vm-errors" }
 
 bn = { package = "zeropool-bn", version = "0.5.11", features = [], optional = true }
+log = { version = "0.4", optional = true}
```
matklad, line 34: Please use `tracing` instead of `log`.

runtime/near-vm-logic/src/logic.rs
```diff
@@ -5,6 +5,8 @@ use crate::types::{PromiseIndex, PromiseResult, ReceiptIndex, ReturnData};
 use crate::utils::split_method_names;
 use crate::ValuePtr;
 use byteorder::ByteOrder;
+#[cfg(feature = "sandbox")]
+use log::debug;
```
matklad, line 9:
```suggestion
```

runtime/near-vm-logic/src/logic.rs
```diff
@@ -2516,6 +2518,13 @@ impl<'a> VMLogic<'a> {
         let new_used_gas = self.gas_counter.used_gas();
         self.gas_counter.process_gas_limit(new_burn_gas, new_used_gas)
     }
+
+    #[cfg(feature = "sandbox")]
+    pub fn debug_log(&mut self, len: u64, ptr: u64) -> Result<()> {
+        let message = self.get_utf8_string(len, ptr)?;
+        debug!(target: "sandbox", "{}", message);
```
matklad, line 2525:
```suggestion
        tracing::debug!(target: "sandbox", "{}", message);
```

runtime/near-vm-logic/src/logic.rs
```diff
@@ -2516,6 +2518,13 @@ impl<'a> VMLogic<'a> {
         let new_used_gas = self.gas_counter.used_gas();
         self.gas_counter.process_gas_limit(new_burn_gas, new_used_gas)
     }
+
+    #[cfg(feature = "sandbox")]
+    pub fn debug_log(&mut self, len: u64, ptr: u64) -> Result<()> {
```
matklad, line 2523: This is a public API, so it must be documented. Additionally, it's an API which is subtly different from all other APIs here (as it sandbox-only), so it needs specific docs for that as well.

matklad: Let's move this function to after `storage_has_key`, to keep it with other host functions.

ailisp:
Thanks @matklad . For ExecutionMetadata option - I have some reading in the code. Am I right that profile data is always enabled? ExecutionMetadata is currently V2(ProfileData) for all contract call actions?

To add a new debug log section, it could be `V3(ProfileData, Option<DebugLog>)`. However whether to leave this sandbox only feature to introduce a V3 is a problem. Should we guard V3 or `Option<DebugLog>` with sandbox feature? I think add the feature guard breaks the principle on when not use feature (like you did to remove profile-data feature). 

I also considered the problems with log.

About multiline - this can be address by some hack. The most straightforward and reliable one would be log the length of message before the message.

For correlate log to the contract - Our current "near-workspaces-js" setup, every new test is use a new spinned workspace. This is obviously not efficient, and me forget about correlate log from the test problem. Workspace user have control over when to spin a new workspace, when and which contract is tested on which workspace. A new official recommended setup we're in consensus is: each contract use a new sandbox instance, tests of a single contract is run sequentially on the same sandbox and tests of different contracts are paralleled because of multiprocess of sandbox. Under this setup it's easy to figure out the log is coming from which contract and which test. But it'd be better if log/call correlation can be established under any setup. I'll let @MaximusHaximus to make the decision.

matklad:
@ailisp yeah, I think adding `V3` or `V2Sandbox` to `ExecutionMetadata` (which we store in the db) and adding a field to `ExecutionMetadataView` (whiche describes the JSON API) is the way to go. The question about features would be interesting. I'd say we probably shouldn't use one for the db-side of things, just to make sure sandbox and non-sandbox node can access the same data. 

>Under this setup it's easy to figure out the log is coming from which contract and which test

Hm, I think there's still a case of cross-contract calls? Workspace calls contract `A`, but it then calls `B` and `C`, and those two are executed concurrently in some future block. 

@MaximusHaximus I see there are two tracks here: 

* Showing source-mapped backtrace when contract throws an exception
* Support for the *live* streaming of stderr

For the first use-case, I think the intended API is for the SDK to use `panic_utf8` host function: 

https://github.com/near/nearcore/blob/ecc175634276910d816871b9abda0f3e4f897e0b/runtime/near-vm-logic/src/logic.rs#L2013-L2028

It allows the contract to fail with a specific error message, which then should be available via API. This is a stable, not-only-sandbox API, so it can be used to, eg, debug cases where an already deployed contracts fail for some reason. If you haven't already, please consider using this existing API to solve this narrow problem.

For the broader question of "streaming output", I agree that, if we do want to have streaming output, that the current PR is probably the best step. Using log is hacky, but it's an incremental first step. I am not quite sure we want streaming output, as it seems a bit at odds with our execution model. Besides the logs, you'd probably want to know if the tx failed or not, and you'll want to say things (on the tooling level): "this contract call failed with this error. Here are debug logs". With the suggested impl, the picture would be like this: workspaces submits a tx, this single tx leads to execution of A, B, and C contracts in the future, and then workspaces receives three results for the txes, plus an unrelated single stream of debug logs. 

The PR basically looks good to me though, I don't think this'll add to the burden on the nearcore side, I am worrying more about workspaces side.

runtime/near-vm-logic/src/logic.rs
```diff
@@ -2389,6 +2389,22 @@ impl<'a> VMLogic<'a> {
...
+    ///
+    /// # Errors
+    ///
+    /// * If string is not UTF-8 returns `BadUtf8`
+    ///
+    /// # Cost
+    ///
+    /// 0
+    #[cfg(feature = "sandbox")]
+    pub fn debug_log(&mut self, len: u64, ptr: u64) -> Result<()> {
```
matklad, line 2402:
```suggestion
    pub fn sandbox_debug_log(&mut self, len: u64, ptr: u64) -> Result<()> {
```

let's add `sandbox_` prefix to all of the sandbox calls, so the "sandbox-only" aspect of this is also visible at the call-site in the contract.

runtime/near-vm-logic/src/logic.rs
```diff
@@ -2389,6 +2389,22 @@ impl<'a> VMLogic<'a> {
...
 
+    /// Debug print given utf-8 string to node log. It's only available in Sandbox node
+    ///
+    /// # Errors
+    ///
+    /// * If string is not UTF-8 returns `BadUtf8`
+    ///
+    /// # Cost
+    ///
+    /// 0
```
matklad, line 2400: I don't think this is correct, `get_utf8_string` costs some gas.

matklad (commented):
Left couple of more comments which *should* be fixed, but otherwise the impl looks good to me.

However, this also doesn't have any automated tests. While we can shrug it off and say "nah, that's sandbox-only API which we don't worry about breaking", I think we **must** test at least the fact that our macro correctly gates sandbox-only API. 

Please write a test in `compile_errors.rs` that the contract that tries to call `sandbox_debug_log` works if and only if sandbox feature is enabled.

runtime/near-vm-logic/src/logic.rs
ailisp: Fixed. Use a sandbox version of get utf8 string. The idea is we still want this cost 0, so adding debug log to a function won't cause run out of gas for an existing funcall.

ailisp:
@matklad ptal

runtime/near-vm-runner/src/tests/compile_errors.rs
```diff
@@ -235,3 +235,51 @@ fn test_limit_locals() {
...
+    .unwrap()
+}
+
+#[cfg(not(feature = "sandbox"))]
+#[test]
+fn test_sandbox_only_function() {
+    with_vm_variants(|vm_kind| {
+        let wasm = call_sandbox_debug_log();
+        let res = make_simple_contract_call_vm(&wasm, "main", vm_kind);
+        println!("{:?}", res);
```
nagisa, line 257: Remove this?

neard/src/cli.rs
```diff
@@ -431,6 +431,10 @@ fn init_logging(verbose: Option<&str>) {
         }
     }
 
+    if cfg!(feature = "sandbox") {
+        env_filter = env_filter.add_directive("sandbox=debug".parse().unwrap());
```
nagisa, line 435: Could you add a comment describing why this is being done? Something along the lines of

> The `env.sandbox_debug_log` VMLogic function implements output via this `sandbox` logging target.

runtime/near-vm-logic/src/logic.rs
```diff
@@ -2391,6 +2389,22 @@ impl<'a> VMLogic<'a> {
...
+    #[cfg(feature = "sandbox")]
+    pub fn debug_log(&mut self, len: u64, ptr: u64) -> Result<()> {
+        let message = self.get_utf8_string(len, ptr)?;
+        tracing::debug!(target: "sandbox", "{}", message);
```
nagisa, line 2417: I believe this is better written as `tracing::debug!(target: "sandbox", message);`. The primary text of the log message is conventionally represented as the `message` key-value and this is guaranteed to avoid invoking the formatting machinery.

nagisa (approved):
Aleksey is out, I took a look. r=me *after* these comments are fixed.

runtime/near-vm-logic/src/logic.rs
ailisp: Seems doesn't compile:
```
nearcore (sandbox-debug-log) make sandbox
cargo build -p neard --features sandbox
   Compiling near-vm-logic v0.0.0 (/home/bo/workspace/nearcore/runtime/near-vm-logic)
error[E0277]: the trait bound `std::string::String: tracing::Value` is not satisfied
    --> runtime/near-vm-logic/src/logic.rs:2417:9
     |
2417 |         tracing::debug!(target: "sandbox", message);
     |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `tracing::Value` is not implemented for `std::string::String`
     |
     = note: required for the cast to the object type `dyn tracing::Value`
     = note: this error originates in the macro `$crate::valueset` (in Nightly builds, run with -Z macro-backtrace for more info)

For more information about this error, try `rustc --explain E0277`.
error: could not compile `near-vm-logic` due to previous error
make: *** [Makefile:82: neard-sandbox] Error 101
```

ailisp:
ping @EgorKulikov @bowenwang1996 @mm-near @mzhangmzz

runtime/near-vm-logic/src/logic.rs
nagisa: Ah, I missed this is a `String`. You'll need an `&str`, [like this](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=3f556a015b53c73b8d87cb8b218fbdf9).

mm-near (approved):

runtime/near-vm-logic/src/logic.rs
ailisp: Got it, thanks! I learned more syntax about it on tracing's docs page. I also need a `message =`.

PR #6326, opened 2022-02-21
Graceful shutdown on panic

### LINKED ISSUE #5340: Should nearcore exit with error code after upgrade panic

nikurt:
Panic of `ClientActor` is handled to gracefully shutdown RocksDB and exit. Supported only on Linux.

Fix #5340


neard/src/cli.rs
```diff
@@ -373,6 +402,11 @@ impl RunCmd {
             actix::System::current().stop();
         });
         sys.run().unwrap();
+        // Wait a few extra seconds to let RocksDB cleanly shut down.
+        // It's unknown how much time the process actually needs.
+        const ROCKS_DB_SLEEP_SECONDS: u64 = 3;
+        info!(target: "neard", "Waiting for {} seconds to let RocksDB shut down cleanly...",ROCKS_DB_SLEEP_SECONDS);
+        std::thread::sleep(Duration::from_secs(ROCKS_DB_SLEEP_SECONDS));
```
matklad, line 409: Yeah, this is quite suboptimal, for two reasons: 

* it penalises every shutdown by three seconds, which feels pretty bad, especially for sandbox  (which might be started and shut-down in tests at all)
* it doesn't guarantee that the DB gets properly closed. If something somewhere runs a "once in 10 seconds" loop or some such, and holds a reference to db, the db won't get closed

The proper way to handle such situations is to:

* rely on RAII -- making sure that all resources are cleaned up when `Drop` is finished
* rely on structured concurrency -- making sure that all outstanding tasks and threads are joined when we exit scopes

I am not sure how to do this properly in actix -- I've spend quite a bit of time reading into actix code at this point, and I am still hopelessly confused as to what the intended coding patterns are there -- it seems that, eg, `SyncArbiter` doesn't have a way to wait for actor termination at all. 

So, I think this might be OK to solve the immediate problems, but this is definitely very messed up and feels like something we should intentionally fix some day.

abacabadabacaba:
Sleeping to prevent a race condition is a big kludge. To avoid RocksDB errors, it is needed to make sure that RocksDB object is dropped before the main thread exits.

neard/src/cli.rs
nikurt: The issue is that the graceful shutdown is performed by another thread. The current hack also breaks a lot of tests, because tests generously use `System.stop()`.

Let me see if it's possible to directly synchronize the `select!` and `ClientActor` drop.

Printing `std::thread::current().id()` I get the following:

```
Feb 22 10:42:04.958  INFO neard: Got DIED, stopping...
Feb 22 10:42:04.959 ERROR near_store::db: RocksDB cancelling all background work: My thread id is ThreadId(4)
Feb 22 10:42:04.959 ERROR neard::cli: Waiting for another thread to close RocksDB. My thread id is ThreadId(1)
Feb 22 10:42:04.959  INFO neard: Waiting for 3 seconds to let RocksDB shut down cleanly...
Feb 22 10:42:04.959 ERROR near_store::db: RocksDB cancelled all background work and dropped: My thread id is ThreadId(4)
```

In some runs RocksDB was dropped by the same thread, but usually by a different one.


The backtrace at the point of dropping RocksDB object is
```
   0: <near_store::db::RocksDB as core::ops::drop::Drop>::drop
   1: core::ptr::drop_in_place<near_store::db::RocksDB>
   2: alloc::sync::Arc<T>::drop_slow
   3: core::ptr::drop_in_place<core::cell::UnsafeCell<near_epoch_manager::EpochManager>>
   4: alloc::sync::Arc<T>::drop_slow
   5: core::ptr::drop_in_place<nearcore::runtime::NightshadeRuntime>
   6: alloc::sync::Arc<T>::drop_slow
   7: core::ptr::drop_in_place<near_chunks::SealsManager>
   8: core::ptr::drop_in_place<near_client::client::Client>
   9: core::ptr::drop_in_place<near_client::client_actor::ClientActor>
  10: core::ptr::drop_in_place<tokio::runtime::task::core::Stage<actix::contextimpl::ContextFut<near_client::client_actor::ClientActor,actix::context::Context<near_client::client_actor::ClientActor>>>>
  11: tokio::runtime::task::harness::Harness<T,S>::shutdown
  12: tokio::macros::scoped_tls::ScopedKey<T>::set
  13: <tokio::task::local::LocalSet as core::ops::drop::Drop>::drop
  14: core::ptr::drop_in_place<actix_rt::runtime::Runtime>
  15: actix_rt::system::SystemRunner::run
  16: neard::cli::NeardCmd::parse_and_run
  17: std::sys_common::backtrace::__rust_begin_short_backtrace
  18: std::rt::lang_start::{{closure}}
  19: core::ops::function::impls::<impl core::ops::function::FnOnce<A> for &F>::call_once
             at /rustc/db9d1b20bba1968c1ec1fc49616d4742c1725b4b/library/core/src/ops/function.rs:259:13
      std::panicking::try::do_call
             at /rustc/db9d1b20bba1968c1ec1fc49616d4742c1725b4b/library/std/src/panicking.rs:406:40
      std::panicking::try
             at /rustc/db9d1b20bba1968c1ec1fc49616d4742c1725b4b/library/std/src/panicking.rs:370:19
      std::panic::catch_unwind
             at /rustc/db9d1b20bba1968c1ec1fc49616d4742c1725b4b/library/std/src/panic.rs:133:14
      std::rt::lang_start_internal::{{closure}}
             at /rustc/db9d1b20bba1968c1ec1fc49616d4742c1725b4b/library/std/src/rt.rs:128:48
      std::panicking::try::do_call
             at /rustc/db9d1b20bba1968c1ec1fc49616d4742c1725b4b/library/std/src/panicking.rs:406:40
      std::panicking::try
             at /rustc/db9d1b20bba1968c1ec1fc49616d4742c1725b4b/library/std/src/panicking.rs:370:19
      std::panic::catch_unwind
             at /rustc/db9d1b20bba1968c1ec1fc49616d4742c1725b4b/library/std/src/panic.rs:133:14
      std::rt::lang_start_internal
             at /rustc/db9d1b20bba1968c1ec1fc49616d4742c1725b4b/library/std/src/rt.rs:128:20
  20: main
  21: __libc_start_call_main
  22: __libc_start_main@GLIBC_2.2.5
  23: _start
```

neard/src/cli.rs
```diff
@@ -283,6 +286,26 @@ pub(super) struct RunCmd {
...
+
+impl Actor for SystemStoppedActor {
+    type Context = Context<Self>;
+
+    // Does nothing but needs to be implemented for the Actor trait trait.
+    fn started(&mut self, _: &mut Self::Context) {}
+}
+
+use once_cell::sync::Lazy;
+pub static SYSTEM_STOPPED_CHANNEL: Lazy<Channel<bool, [bool; 1]>> = Lazy::new(|| Channel::new());
```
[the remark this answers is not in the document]
nikurt: I gave it a try and couldn't make it work.
Stuck at this panic:

```
thread 'main' panicked at 'Cannot start a runtime from within a runtime. This happens because a function (like `block_on`) attempted to block the current thread while the thread is being used to drive asynchronous tasks.',
```

chain/client/src/client_actor.rs
```diff
@@ -1534,6 +1534,8 @@ impl ClientActor {
 impl Drop for ClientActor {
     fn drop(&mut self) {
         self.state_parts_client_arbiter.stop();
+        info!(target:"chain", "As ClientActor is dropped, initiate a graceful shutdown.");
+        System::current().stop();
```
mm-near, line 1538: btw - will this still work after #6333 ?

core/store/src/db.rs
```diff
@@ -911,6 +928,29 @@ impl Drop for RocksDB {
...
+        let mut num_instances = lock.lock().unwrap();
+        *num_instances += 1;
+        cvar.notify_all();
+        Self {}
+    }
+}
+
+impl Drop for InstanceCounter {
+    fn drop(&mut self) {
+        debug!(target:"db","Dropped an instance of RocksDB");
```
mm-near, line 946: same here - mention how many are there

core/store/src/db.rs
```diff
@@ -911,6 +928,29 @@ impl Drop for RocksDB {
     }
 }
 
+struct InstanceCounter {}
+
+impl InstanceCounter {
```
mm-near, line 933: maybe add a comment why do we need to track rocksdb instances at all.

neard/src/cli.rs
```diff
@@ -283,6 +287,34 @@ pub(super) struct RunCmd {
...
 
+/// SystemStoppedActor is used to monitor whether the actix System was stopped.
+struct SystemStoppedActor {
+    tx: Option<oneshot::Sender<bool>>,
+}
+
+impl Actor for SystemStoppedActor {
+    type Context = Context<Self>;
+
+    // Does nothing but needs to be implemented for the Actor trait trait.
```
mm-near, line 298: trait twice.

neard/src/cli.rs
```diff
@@ -373,6 +412,8 @@ impl RunCmd {
             actix::System::current().stop();
         });
         sys.run().unwrap();
+        info!(target: "neard", "Waiting for RocksDB to gracefully shutdown");
+        RocksDB::block_until_all_instances_are_dropped();
```
mm-near, line 416: which threads are still 'alive' when we're at this point?

mm-near (changes requested):

marcelo-gonzalez:
A question I would have is what is the purpose of the `SystemStoppedActor`? My gut feeling is that defining and starting an actor whose purpose is only to wait for the system to be stopped is overly complicated, since that's exactly what `System::run()` does anyway, right? Before this change, we `select!` on receiving either one of `SIGINT` or `SIGTERM`, and then we `stop()` and `await` the servers, then call `System::stop()`. So if the goal is to have the dropping of the `ClientActor` trigger all these things, could we not do it more simply by just passing the `ClientActor` the `oneshot::Sender` you created in this PR?

```rust
pub struct ClientActor {
    _stopped: Option<oneshot::Sender<()>>,
    /// Adversarial controls
    pub adv: Arc<std::sync::RwLock<crate::AdversarialControls>>,

    client: Client,
    ...
    ...
```

Then you wouldn't need to do anything extra, as the rx side will get the message when `self._stopped` is dropped, and the only change you would have to make to `neard/src/cli.rs` would be adding the `rx` side to the `select!` statement like:

```
diff --git a/neard/src/cli.rs b/neard/src/cli.rs
index 684cf098f..c0eb057bc 100644
--- a/neard/src/cli.rs
+++ b/neard/src/cli.rs
@@ -10,6 +10,7 @@ use std::{env, fs, io};
 use tracing::metadata::LevelFilter;
 use tracing::{debug, error, info, warn};
 use tracing_subscriber::EnvFilter;
+use tokio::sync::oneshot;
 
 /// NEAR Protocol Node
 #[derive(Clap)]
@@ -347,10 +348,11 @@ impl RunCmd {
             }
         }
 
+        let (tx, rx) = oneshot::channel::<()>();
         let sys = actix::System::new();
         sys.block_on(async move {
             let nearcore::NearNode { rpc_servers, .. } =
-                nearcore::start_with_config(home_dir, near_config).expect("start_with_config");
+                nearcore::start_with_config(home_dir, near_config, tx).expect("start_with_config");
 
             let sig = if cfg!(unix) {
                 use tokio::signal::unix::{signal, SignalKind};
@@ -358,7 +360,8 @@ impl RunCmd {
                 let mut sigterm = signal(SignalKind::terminate()).unwrap();
                 futures::select! {
                     _ = sigint .recv().fuse() => "SIGINT",
-                    _ = sigterm.recv().fuse() => "SIGTERM"
+                    _ = sigterm.recv().fuse() => "SIGTERM",
+                    _ = rx.fuse() => "STOP"
                 }
             } else {
                 tokio::signal::ctrl_c().await.unwrap();
```

marcelo-gonzalez:
Also, instead of using the `InstanceCounter` to wait for all rocksb instances to be dropped, did you consider trying to just make sure all threads get the signal to stop, joining them all in `main`? I realize that might be hairy though given the different types of actix actors we start (`SyncArbiter`, etc), but if that's possible then I think that would achieve the goal but without having to treat rocksdb in any special way.

neard/src/cli.rs
nikurt: In the experiments it was always one of `ViewClientActor` threads. RocksDB gets dropped when all references are dropped. Therefore the thread of the last dropped actor is the thread dropping RocksDB.

chain/client/src/client_actor.rs
nikurt: Yes, and it should fix the `pthread destroy mutex: Device or resource busy` issue.

nikurt:
@marcelo-gonzalez good idea to avoid introducing `SystemStoppedActor`.

bowenwang1996 (commented):
@nikurt how did you test this change?

chain/client/src/client_actor.rs
mm-near: my point is, that it feels 'strange' that the ClientActor destructor is the thing that initiates 'global' shutdown.

mm-near (changes requested):

nikurt:
@bowenwang1996 Tested manually on localnet. Mocknet wouldn't be much different from localnet.
The tests on localnet was injecting panics or trying to join a network with a much higher protocol version.

mm-near (approved):

chain/client/src/client_actor.rs
```diff
@@ -62,6 +63,9 @@ const BLOCK_HORIZON: u64 = 500;
 const HEAD_STALL_MULTIPLIER: u32 = 4;
 
 pub struct ClientActor {
+    /// Synchronization measure to allow graceful shutdown.
+    /// Informs the system when a ClientActor gets dropped.
+    _sender: Option<oneshot::Sender<()>>,
```
matklad, line 68: Let's put this field last: rust drops struct fields in order, and we want this field to be dropped last to make sure that everything is fully drop by the time we get the signal.

chain/client/src/client_actor.rs
```diff
@@ -124,6 +128,7 @@ impl ClientActor {
         enable_doomslug: bool,
         rng_seed: RngSeed,
         ctx: &Context<ClientActor>,
+        sender: Option<oneshot::Sender<()>>,
```
matklad, line 131: is there a better name here? `shutdown_signal` perhaps?

core/store/src/db.rs
```diff
@@ -830,7 +832,22 @@ fn rocksdb_column_options(col: DBCol) -> Options {
...
+    Lazy::new(|| (Mutex::new(0), Condvar::new()));
+
 impl RocksDB {
+    /// Blocks until all RocksDB instances (usually 0 or 1) gracefully shutdown.
+    pub fn block_until_all_instances_are_dropped() {
+        debug!(target: "db", "Blocking until all RocksDB instances perform graceful shutdown");
+        let (lock, cvar) = &*ROCKSDB_INSTANCES_COUNTER;
+        let mut num_instances = lock.lock().unwrap();
+        while *num_instances != 0 {
+            num_instances = cvar.wait(num_instances).unwrap();
```
matklad, line 846: Might want to add a log line here -- it feels like the place where we can get stuck if, for some reason, some of the threads *don't* shut down.

core/store/src/db.rs
```diff
@@ -830,7 +832,22 @@ fn rocksdb_column_options(col: DBCol) -> Options {
     opts
 }
 
+// Number of RocksDB instances in the process.
+pub static ROCKSDB_INSTANCES_COUNTER: Lazy<(Mutex<usize>, Condvar)> =
```
matklad, line 836: Can this be at least `pub(crate)` ? Exposing global state to the world isn't great.

matklad (approved):
SGTM!

Even when we fix this properly by making sure view threads die properly, I think we should have something like this regardless, just as a debug aid, to make sure the problem doesn't resurface.

nearcore/src/lib.rs
```diff
@@ -71,84 +71,17 @@ pub fn get_default_home() -> PathBuf {
...
     if db_version > near_primitives::version::DB_VERSION {
         error!(target: "near", "DB version {} is created by a newer version of neard, please update neard or delete data", db_version);
         std::process::exit(1);
     }
-
     if db_version == near_primitives::version::DB_VERSION {
         return;
     }
 
-    // Before starting a DB migration, create a consistent snapshot of the database. If a migration
```
marcelo-gonzalez, line 126: hmm why do we want to delete this? Even with the graceful shutdown here, receiving SIGINT in the middle of a migration would still leave the DB in a bad state, right?

marcelo-gonzalez (approved):

PR #6327, opened 2022-02-21
scripts: accept NEAR_HOME from environment and other minor improvements

mina86:
run_docker.sh unconditionally sets value of NEAR_HOME environment
variable to /srv/near which means that running ‘docker run -e
NEAR_HOME=/home/near ... nearcore’ has no effect.  To be fair, user
probably won’t need to set near home directory inside of the container
since really what matters is path outside of the container which is
set via ‘docker run -v/path/outside/of/container:/srv/near ...’ but at
the same time there’s no reason not to respect the path.

Speaking of NEAR_HOME, since neard reads that environment variable
there is no need to specify --home argument if its exported.  Since
run_docker.sh does export it the argument can be skipped simplifying
the script.

Lastly, reformat the script a little so it reads better (most notably
wrap long lines).


nikurt (approved):

nikurt:
> reformat the script a little so it reads better (most notably wrap long lines)

Do we have a tool for that?

mina86:
> Do we have a tool for that?

Not for shell scripts.

PR #6328, opened 2022-02-21
chain: replace O(N) with O(1) hash lookup in assign_with_possible_reats

mina86:
To avoid adding chunk producer to a shard multiple times, the code
would go through the list of already assigned producers and look for
the one being considered.  This means a O(N) lookup each time so
overall a O(N²) algorithm.

Replace the linear scan with a hash set offering O(1) lookup which
reduces complexity of the algorithm.

### DIFF
--- a/chain/epoch_manager/src/shard_assignment.rs
+++ b/chain/epoch_manager/src/shard_assignment.rs
@@ -45,7 +45,7 @@ pub fn assign_shards<T: HasStake + Eq + Clone>(
     // First, distribute chunk producers until all shards have at least the
     // minimum requested number.  If there are not enough validators to satisfy
     // that requirement, assign some of the validators to multiple shards.
-    let mut chunk_producers = chunk_producers.into_iter().cycle();
+    let mut chunk_producers = chunk_producers.into_iter().enumerate().cycle();
     assign_with_possible_repeats(
         &mut shard_index,
         &mut result,
@@ -64,7 +64,7 @@ pub fn assign_shards<T: HasStake + Eq + Clone>(
             .map(|(count, stake, shard_id)| (stake, count, shard_id))
             .collect();
 
-        for cp in chunk_producers.take(remaining_producers) {
+        for (_, cp) in chunk_producers.take(remaining_producers) {
             let (least_stake, least_validator_count, shard_id) =
                 shard_index.pop().expect("shard_index should never be empty");
             shard_index.push((least_stake + cp.get_stake(), least_validator_count + 1, shard_id));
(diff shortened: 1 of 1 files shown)


chain/epoch_manager/src/shard_assignment.rs
```diff
@@ -75,16 +75,19 @@ pub fn assign_shards<T: HasStake + Eq + Clone>(
...
 
-fn assign_with_possible_repeats<T: HasStake + Eq, I: Iterator<Item = T>>(
+fn assign_with_possible_repeats<T: HasStake + Eq, I: Iterator<Item = (usize, T)>>(
     shard_index: &mut MinHeap<(usize, Balance, ShardId)>,
     result: &mut Vec<Vec<T>>,
     cp_iter: &mut I,
     min_validators_per_shard: usize,
 ) {
     let mut buffer = Vec::with_capacity(shard_index.len());
+    let mut seen = std::collections::HashSet::<(ShardId, usize)>::with_capacity(
```
mm-near, line 85: comment

chain/epoch_manager/src/shard_assignment.rs
```diff
@@ -103,19 +106,7 @@ fn assign_with_possible_repeats<T: HasStake + Eq, I: Iterator<Item = T>>(
...
-                    // shard once we figure out what to do with current `cp`).
-                    //
-                    // TODO(mina86): The contains check in the condition makes
-                    // this an O(N^2) algorithm.  At the moment there aren’t too
-                    // many chunk producers so it should be fine but if that’s
-                    // becomes an issue we should switch to a hash set.
-                    buffer.push(PeekMut::pop(top));
-                }
-                Some(mut top) => {
+                Some(mut top) if seen.insert((top.2, cp_index)) => {
```
mm-near, line 109: I'm missing something.

What's preventing us from adding the same cp twice to the same shard?

as cp_iter is 'cycle' - it means that it can have the same cp multiple times with multiple 'cp_indexes', right?

mm-near (changes requested):

chain/epoch_manager/src/shard_assignment.rs
mina86: The cycle is done after enumeration so the same cp will have the same cp_index.

mm-near: ah, ok - that's what I've missed. Let's comment about it then (that cp_index is basically like a 'hash'/unique_identifier of the cp) and we're good to go.

mm-near (approved):

PR #6333, opened 2022-02-22
start client actor in a separate thread

mm-near:
It seems that starting the actor in the 'current' thread (as we do right now) - is causing a bunch of issues (impacting the messages that are going to other actors)

### DIFF
--- a/chain/client-primitives/src/types.rs
+++ b/chain/client-primitives/src/types.rs
@@ -346,6 +346,13 @@ pub enum QueryError {
     },
     #[error("The node reached its limits. Try again later. More details: {error_message}")]
     InternalError { error_message: String },
+    #[error(
+        "The data for block #{block_height} is garbage collected on this node, use an archival node to fetch historical data"
+    )]
+    GarbageCollectedBlock {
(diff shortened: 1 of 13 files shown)


matklad:
Yeah, this makes total sense! One thing to note that, as we now spawn `ClientActor` to a different thread, we exacerbate shutdown problems. Ideally, we'd return an `Arbiter`, rather than `ArbiterHandle` from this function, and `.join()` it in main to make sure that all cleanup work finished. 

And yeah, I feel that the fundamental problem is that whatever actix is doing internally is poorly documented, opaque, and doesn't actually help us much. Given the size and complexity of near, it makes sense for us to own our event loop, rather than to hide it in third-party actor framework without significant usage.

bowenwang1996 (approved):
Epic! I saw this failure in CI `pthread destroy mutex: Device or resource busy` which is interesting.

nikurt:
> `pthread destroy mutex: Device or resource busy`

I hope that https://github.com/near/nearcore/pull/6326 will fix this.

PR #6335, opened 2022-02-22
localnet with many shards

mm-near:
Support num-shards correctly (before this PR, if you run the init command with shard_num > 1 the neard crashes).

After this PR, you can run commands like: 
./target/debug/neard --home ~/.near_tmp init --chain-id localnet --num-shards=5


nearcore/src/config.rs
```diff
@@ -1058,6 +1060,20 @@ pub fn init_configs(
                 CryptoHash::default(),
             );
             add_protocol_account(&mut records);
+            let shards = if num_shards > 1 {
+                ShardLayout::v1(
+                    (0..num_shards - 1)
+                        .map(|f| {
+                            AccountId::from_str(format!("shard{}.test.near", f).as_str()).unwrap()
```
mzhangmzz, line 1067: nit: maybe we should generate these account dividers randomly and sort them. In the current way, the middle shards will only have one account.

bowenwang1996: good point

bowenwang1996 (approved):

nearcore/src/config.rs
mm-near: these are not 'dividers' but 'fixed_shards' - and they have one account on purpose (as this makes later load-testing a lot easier).

So in the current setup, if we ask for 3 shards, we get:
shard 0: only account named 'shard0.test.near'
shard 1: only account named 'shard1.test.near'
shard 2: all the other accounts.

PR #6337, opened 2022-02-22
Redirect to the 0-copy wasmer branch

nagisa:
This makes the changes necessary on the nearcore side to start
experimenting with – and testing – the branch of runtime implementing
0-copy deserialization.

@Ekleog I wonder how difficult would it be to throw this at the fuzzing infrastructure?

### DIFF
--- a/runtime/near-vm-runner/src/cache.rs
+++ b/runtime/near-vm-runner/src/cache.rs
@@ -338,7 +338,7 @@ pub mod wasmer2_cache {
                 vm.engine
                     .load_universal_executable(&executable)
                     .map(|v| Arc::new(v) as _)
-                    .map_err(|err| CompilationError::WasmerCompileError { msg: err.to_string() })
+                    .map_err(|err| panic!("could not load the executable: {}", err.to_string()))
             })),
             Some(cache) => {
                 let serialized = cache.get(&key.0).map_err(|_io_err| CacheError::ReadError)?;
--- a/runtime/near-vm-runner/src/cache.rs
+++ b/runtime/near-vm-runner/src/cache.rs
@@ -253,17 +253,19 @@ pub mod wasmer0_cache {
 pub mod wasmer2_cache {
     use crate::wasmer2_runner::{VMArtifact, Wasmer2VM};
     use near_primitives::contract::ContractCode;
+    use wasmer_engine::Executable;
 
     use super::*;
 
     pub(crate) fn compile_module_wasmer2(
+        vm: &Wasmer2VM,
         code: &[u8],
         config: &VMConfig,
-    ) -> Result<VMArtifact, CompilationError> {
+    ) -> Result<wasmer_engine_universal::UniversalExecutable, CompilationError> {
         let _span = tracing::debug_span!(target: "vm", "compile_module_wasmer2").entered();
         let prepared_code =
             prepare::prepare_contract(code, config).map_err(CompilationError::PrepareError)?;
-        Wasmer2VM::new(config.clone()).compile_uncached(&prepared_code)
+        vm.compile_uncached(&prepared_code)
     }
 
     pub(crate) fn compile_and_serialize_wasmer2(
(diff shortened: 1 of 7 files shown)


Ekleog:
It's just a matter of pushing this branch to nearcore and then adding a `branch` block to https://github.com/near/nearcore/blob/master/nightly/fuzz.toml ; then on the next refresh (every ~24 hours) the fuzzing infrastructure should pick the branch up.

That said, right now we have a single machine running fuzzing because we're still testing out the deployment to validate nothing goes bad, so depending on the timeline of fuzzing this (hopefully we'll have the full fuzzing infra running production in 1-2 weeks at most) it may also be worth it just spinning up a VM to validate at least the basics faster. (the fuzzing infrastructure doesn't allow picking manually which targets to run and would not necessarily run the targets that are of interest to us, as it runs only 4 target/branch couples per day)

runtime/near-vm-runner/src/wasmer2_runner.rs
```diff
@@ -378,29 +332,45 @@ impl Wasmer2VM {
...
+                    .map_err(|err| {
+                        use wasmer_engine::InstantiationError::{self, *};
+                        match err.downcast_ref::<InstantiationError>() {
+                            Some(Start(err)) => translate_runtime_error(err.clone(), import.vmlogic),
+                            Some(Link(e)) => VMError::FunctionCallError(FunctionCallError::LinkError { msg: e.to_string() }),
+                            Some(CreateInstance(e)) => VMError::FunctionCallError(FunctionCallError::LinkError { msg: e.to_string() }),
+                            Some(CpuFeature(e)) => {
+                                panic!("host does not support the CPU features required to run contracts: {}", e)
+                            }
+                            None => VMError::FunctionCallError(FunctionCallError::LinkError { msg: err.to_string() })
```
matklad, line 344: This looks sketchy -- what kind of other errors can we get here? Can we change upstream API to avoid downcasts here? If we cant, than I think a commment about when this branch happens might be useful.

runtime/near-vm-runner/src/wasmer2_runner.rs
```diff
@@ -447,8 +417,19 @@ impl Wasmer2VM {
...
             current_protocol_version,
         );
-        let import = imports::wasmer2::build(vmmemory, &mut logic, current_protocol_version);
-        if let Err(e) = get_entrypoint_index(&*artifact.artifact, method_name) {
+
+        let import = imports::wasmer2::build(
+            vmmemory,
+            &mut logic,
+            current_protocol_version,
+            // FIXME: make sure tricky case below is impossible.
```
matklad, line 425: Yeah, fixing that would be great!

runtime/near-vm-runner/src/cache.rs
```diff
@@ -326,8 +332,14 @@ pub mod wasmer2_cache {
         config: &VMConfig,
         cache: Option<&dyn CompiledContractCache>,
     ) -> Result<Result<VMArtifact, CompilationError>, CacheError> {
+        let vm = Wasmer2VM::new(config.clone());
         match cache {
-            None => Ok(compile_module_wasmer2(code.code(), config)),
+            None => Ok(compile_module_wasmer2(&vm, code.code(), config).and_then(|executable| {
+                vm.engine
+                    .load_universal_executable(&executable)
```
matklad, line 339: Hm, what kind of error can we get during loading? I'd assume that this step can only fail catastrophically (mmap fails, or something), and that we should just panic here. But I assume the reality doesn't work like that? :)

matklad (commented):
LGTM, but I have a couple of questions about error handling!

runtime/near-vm-runner/src/cache.rs
nagisa: Failure modes here are indeed representing mostly internal failure modes – function body being longer than 32-bits, failure to decode a data initializer, allocate data and such.

runtime/near-vm-runner/src/wasmer2_runner.rs
nagisa: This branch should be unreachable. The only reason we're dealing with a dynamic error here is due to the way crate graph looks like (`wasmer_vm` contains the `Artifact` trait and `wasmer_engine` contains `InstantiationError` – I should just add an associated type to the trait one of those days...)

runtime/near-vm-runner/src/cache.rs
matklad: Hm, "failure to allocate data" in particular seems like something non-deterministic (can happen on one machine, but not the other), and that's a big no-no for us. This feels like something we should fix the sooner the better. Although, if this doesn't make the situation worse (ie, we already were treating allocation errors as compilation/loading errors), I think it's fine to land this.

runtime/near-vm-runner/src/wasmer2_runner.rs
matklad: Than maybe put `unreachable` here or at least `tracing::error`?

runtime/near-vm-runner/src/cache.rs
nagisa: Yeah, I've been thinking about this a lot lately. Mostly in context of also making things faster with some arenas ^^

runtime/near-vm-runner/src/wasmer2_runner.rs
matklad: Actually, probably log_assert would be best?

https://github.com/near/nearcore/blob/bcd5744f307ae4bcbf504b358d22b43da58a58a2/core/o11y/src/lib.rs#L186

nagisa: I don't see how that'd work here, since its non-diverging, unless the idea is to also return an error anyway?

matklad: Year, `log_assert > tracing::error`, and they both are incomparable with `unreachable`

matklad (approved):
LGTM!

But let's prioritize tightening up the errors! I wonder if this might be a good refactor for @Ekleog to dive into codebase?

PR #6340, opened 2022-02-23
mock network first iteration

mzhangmzz:
First iteration of mock network (different from Nikolay's mocknet, better name suggestion welcomed)
A mock network simulates a test environment for ClientActor by replacing PeerManagerActor, instead of sending network messages out to other peers, it simulates a network and reads the needed block and chunk content from storage.

In this first iteration, the mock network only simulates one peer and all messages get instant response from the mock network. I will add more complicated peer logic in future iteration.

### DIFF
--- a/chain/chain/src/chain.rs
+++ b/chain/chain/src/chain.rs
@@ -3018,6 +3018,10 @@ impl Chain {
             .ok_or_else(|| ErrorKind::DBNotFoundErr(format!("EXECUTION OUTCOME: {}", id)).into())
     }
 
+    /// Retrieve the up to `max_headers_returned` headers on the main chain
+    /// `hashes`: a list of block "locators". This function will find the first block in `hashes`
+    ///           that is on the main chain and returns the blocks after this block. If none of the
+    ///           blocks in `hashes` are on the main chain, the function returns an empty vector.
     pub fn retrieve_headers(
         &mut self,
         hashes: Vec<CryptoHash>,
--- a/chain/chain/src/chain.rs
+++ b/chain/chain/src/chain.rs
@@ -3019,7 +3019,8 @@ impl Chain {
     }
 
     /// Retrieve the up to `max_headers_returned` headers on the main chain
-    /// `hashes`: a list of block "locators". This function will find the first block in `hashes`
+    /// `hashes`: a list of block "locators". `hashes` should be ordered from older blocks to
+    ///           more recent blocks. This function will find the first block in `hashes`
     ///           that is on the main chain and returns the blocks after this block. If none of the
     ///           blocks in `hashes` are on the main chain, the function returns an empty vector.
     pub fn retrieve_headers(
--- a/chain/chain/src/chain.rs
+++ b/chain/chain/src/chain.rs
@@ -3018,6 +3018,30 @@ impl Chain {
             .ok_or_else(|| ErrorKind::DBNotFoundErr(format!("EXECUTION OUTCOME: {}", id)).into())
     }
 
+    pub fn retrieve_headers(
+        &mut self,
+        hashes: Vec<CryptoHash>,
+        max_headers_returned: u64,
+    ) -> Result<Vec<BlockHeader>, Error> {
+        let header = match self.find_common_header(&hashes) {
+            Some(header) => header,
+            None => return Ok(vec![]),
+        };
+
+        let mut headers = vec![];
+        let max_height = self.header_head()?.height;
+        // TODO: this may be inefficient if there are a lot of skipped blocks.
+        for h in header.height() + 1..=max_height {
+            if let Ok(header) = self.get_header_by_height(h) {
+                headers.push(header.clone());
+                if headers.len() >= max_headers_returned as usize {
+                    break;
+                }
+            }
+        }
+        Ok(headers)
+    }
+
     /// Returns a vector of chunk headers, each of which corresponds to the previous chunk of
     /// a chunk in the block after `prev_block`
     /// This function is important when the block after `prev_block` has different number of chunks
--- a/chain/chain/src/chain.rs
+++ b/chain/chain/src/chain.rs
@@ -3026,6 +3026,7 @@ impl Chain {
         &mut self,
         hashes: Vec<CryptoHash>,
         max_headers_returned: u64,
+        max_height: Option<BlockHeight>,
     ) -> Result<Vec<BlockHeader>, Error> {
         let header = match self.find_common_header(&hashes) {
             Some(header) => header,
@@ -3033,7 +3034,8 @@ impl Chain {
         };
 
         let mut headers = vec![];
-        let max_height = self.header_head()?.height;
+        let header_head_height = self.header_head()?.height;
+        let max_height = max_height.unwrap_or(header_head_height);
         // TODO: this may be inefficient if there are a lot of skipped blocks.
         for h in header.height() + 1..=max_height {
             if let Ok(header) = self.get_header_by_height(h) {
--- a/integration-tests/src/bin/start_mock_network.rs
+++ b/integration-tests/src/bin/start_mock_network.rs
@@ -6,13 +6,15 @@ use futures::{future, FutureExt};
 use std::path::Path;
 
 use integration_tests::mock_network::setup::{setup_mock_network, SyncMode};
-use integration_tests::mock_network::GetChainHistoryFinalBlockHeight;
+use integration_tests::mock_network::GetChainTargetBlockHeight;
 use near_actix_test_utils::run_actix;
 use near_chain_configs::GenesisValidationMode;
 use near_client::GetBlock;
 use near_crypto::{InMemorySigner, KeyType};
 use near_logger_utils::init_integration_logger;
 use near_network::test_utils::WaitOrTimeoutActor;
+use near_primitives::types::BlockHeight;
+use std::time::Duration;
 
 /// Program to start a testing environment for one client and a mock network environment
 /// The mock network simulates the entire network by reading a pre-generated chain history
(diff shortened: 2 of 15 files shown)


chain/chain/src/chain.rs
```diff
@@ -3018,6 +3018,30 @@ impl Chain {
             .ok_or_else(|| ErrorKind::DBNotFoundErr(format!("EXECUTION OUTCOME: {}", id)).into())
     }
 
+    pub fn retrieve_headers(
```
mm-near, line 3021: please add comments on what this function does (as  I thought that it would return the block headers for these hashes, but from what I see it returns headers for the chain that starts from the lowest hash -- quite confusing)

integration-tests/src/mock_network/mod.rs
```diff
@@ -0,0 +1,287 @@
...
+
+/// MockPeerManagerActor mocks PeerManagerActor and responds to messages from ClientActor.
+/// Instead of sending these messages out to other peers, it simulates a network and reads
+/// the needed block and chunk content from storage.
+/// MockPeerManagerActor has the following responsibilities
+/// - Responds to BlockRequest, BlockHeadersRequest and PartialEncodedChunkRequest
+/// - Sends NetworkInfo to ClientActor periodically
+/// - Simulates block production and sends the most "recent" block to ClientActor
+struct MockPeerManagerActor {
+    /// Client address for the node that we are testing
```
mm-near, line 28: I'd also mention that this will be used to send the 'reponses' for the requests that we get.

integration-tests/src/mock_network/mod.rs
```diff
@@ -0,0 +1,287 @@
...
+            client_addr,
+            chain_history_access: ChainHistoryAccess { chain },
+            network_info,
+            block_production_delay,
+        }
+    }
+
+    /// This function gets called periodically
+    /// When it is called, it increments peer heights by 1 and sends the block at that height
+    /// to ClientActor. In a way, it simulates peers that get upgraded and broadcast blocks
```
mm-near, line 81: what do you mean by 'upgraded'?

integration-tests/src/mock_network/setup.rs
```diff
@@ -0,0 +1,177 @@
...
+    block_production_delay: Duration,
+) -> MockPeerManagerActor {
+    let chain =
+        Chain::new_for_view_client(runtime, chain_genesis, DoomslugThresholdMode::NoApprovals)
+            .unwrap();
+    MockPeerManagerActor::new(client_addr, genesis_config, chain, 0, block_production_delay)
+}
+
+#[allow(dead_code)]
+fn setup_mock_network(
```
mm-near, line 47: can you explain the arguments? how does 'client' and 'network' homedirs differ?

integration-tests/src/mock_network/setup.rs
```diff
@@ -0,0 +1,177 @@
...
+                client_actor1.recipient(),
+                &genesis_config,
+                &chain_genesis,
+                block_production_delay,
+            )
+        });
+    network_adapter.set_recipient(mock_network_actor.clone().recipient());
+    (mock_network_actor, client_actor, view_client)
+}
+
```
mm-near, line 102: Great - now we're only missing a binary, where you 'point' it at existing (for example mainnet) .near/ home directory and an 'empty' one -- and then it should start 'syncing' ...

integration-tests/src/mock_network/setup.rs
```diff
@@ -0,0 +1,177 @@
...
+        near_config1.client_config.min_num_peers = 1;
+        run_actix(async move {
+            let (_mock_network, _client, view_client) =
+                setup_mock_network(dir1.path().clone(), dir.path().clone(), &near_config1);
+            WaitOrTimeoutActor::new(
+                Box::new(move |_ctx| {
+                    actix::spawn(view_client.send(GetBlock::latest()).then(|res| {
+                        if let Ok(Ok(block)) = res {
+                            if block.header.height >= 20 {
+                                System::current().stop()
```
mm-near, line 165: could we maybe also check that block hash from here matches the one from line 142?

mm-near (approved):

integration-tests/src/mock_network/mod.rs
mzhangmzz: yeah sorry for being misleading. I just meant nodes broadcasting new blocks

integration-tests/src/mock_network/setup.rs
mzhangmzz: added

integration-tests/src/mock_network/mod.rs
```diff
@@ -0,0 +1,304 @@
...
+        _ctx: &mut Self::Context,
+    ) -> Self::Result {
+        self.chain_history_access.chain.head().unwrap().height
+    }
+}
+
+/// This class provides access a pre-generated chain history
+struct ChainHistoryAccess {
+    chain: Chain,
+}
```
bowenwang1996, line 176: feels like you only need `ChainStore` here

mzhangmzz: Actually `Chain` is needed because some functions here such as `retrieve_headers` do rely on logic that is implemented in `Chain` not `ChainStore`.

chain/chain/src/chain.rs
```diff
@@ -3018,6 +3018,36 @@ impl Chain {
             .ok_or_else(|| ErrorKind::DBNotFoundErr(format!("EXECUTION OUTCOME: {}", id)).into())
     }
 
+    /// Retrieve the up to `max_headers_returned` headers on the main chain
+    /// `hashes`: a list of block "locators". This function will find the first block in `hashes`
```
mm-near, line 3022: s/first/oldest ?

integration-tests/src/mock_network/README.md
```diff
@@ -0,0 +1,15 @@
+# mock-network
+This crate hosts libraries to start a test env for a single node by replacing the network module with a mock network environment. 
+The mock network environment simulates the interaction that the client will usually have with other nodes by 
+responding to the client's network messages and broadcasts new blocks. The mock network reads a pre-generated chain
```
mm-near, line 4: s/broadcasts/broadcasting

integration-tests/src/bin/start_mock_network.rs
```diff
@@ -0,0 +1,89 @@
...
+        let (mock_network, _client, view_client) = setup_mock_network(
+            Path::new(&client_home_dir),
+            home_dir,
+            &near_config,
+            sync_mode,
+            network_delay,
+            args.target_height,
+        );
+        let target_height =
+            async { mock_network.send(GetChainTargetBlockHeight).await }.await.unwrap();
```
mm-near, line 72: why don't we use args.target_height?

integration-tests/src/bin/start_mock_network.rs
```diff
@@ -0,0 +1,89 @@
...
+        WaitOrTimeoutActor::new(
```
mm-near, line 73: comment what this timeout actor does

integration-tests/src/mock_network/mod.rs
```diff
@@ -33,6 +35,10 @@ struct MockPeerManagerActor {
     network_info: NetworkInfo,
     /// Block production speed for the network that we are simulating
     block_production_delay: Duration,
+    /// Simulated network delay
+    network_delay: Duration,
+    /// The simulated peers will stop producing new blocks at this height
+    target_height: BlockHeight,
```
mm-near, line 41: which block do they 'start' producing from?

integration-tests/src/mock_network/README.md
```diff
@@ -0,0 +1,15 @@
...
+
+The crate has two files and another binary sits in ../bin/start_mock_network
+
+## mod.rs
+Implements `ChainHistoryAccess` and `MockPeerManagerActor`, which is the main 
+components of the mock network.
+## setup.rs
+Provides functions for setting up a mock network from configs and home dirs.
+## start_mock_network.rs
+A binary that starts a mock network from the home dir of an existing node
```
mm-near, line 15: maybe also add the example command line here?

Also - the part that confused me is 'sync' vs 'no sync' (let me know if I understand it correctly:
if no sync:
- everyone starts from 'genesis_height'
- the 'mock' network will keep 'producting' block until 'target_height'.

If sync:
- the 'mock' network starts from 'target_height' and does NOT produce any more blocks, right?
- and the 'node' is just trying to sync 


(then the option I'm missing -- but can be added in the next PR is the mixture, where:

- the 'mock' network starts from height X, and keeps producing 1 block per second until height Y
- the 'node' tries to 'sync' and then catchup and keep getting the new blocks that are produced.
(this last mode is the most similar to what is happening in production today).

mm-near (approved):

chain/chain/src/chain.rs
mzhangmzz: It's the same, because it assumes that `hashes` is ordered. I'll add that.

integration-tests/src/bin/start_mock_network.rs
mzhangmzz: We pass `args.target_height` to `setup_mock_network` so the `GetChainTargetBlockHeight` function will return information with target height incorporated. 

The reason is that `args.target_height` could exceed the the pre-generated chain history, and we want to use the minimum among the two values for termination condition.

integration-tests/src/mock_network/README.md
mzhangmzz: Yeah that's right. I do plan to add these options later, not now because we can't support it yet, because the client can only start from genesis height for now.

Later, in the full version, the binary take  `client_start_height`, `target_height`. 
If no sync, everyone starts from 'client_start_height' and the 'mock' network will keep 'producting' block until 'target_height'.
If sync, the binary can take another argument `network_start_height`. Client starts from `client_start_height` and the network starts from `network_start_height` and will keep producing blocks until `target_height`.

mzhangmzz: Example command line will be printed by the binary, because of how clap works.

mzhangmzz:
@mm-near @bowenwang1996 I am going to force merge this PR since I don't want to bother the other reviewers. And it's not a critical change that would affect production code

Issue #4340, opened 2021-05-29
Misleading error message when querying on garbage collected data through JSON RPC


frol:
***Describe the bug***

Querying historical data through JSON RPC from a non-archival node I get the following error message:

> The node does not track the shard ID 0

It is not accurate

***To Reproduce***

```
 curl [https://rpc.mainnet.near.org](https://rpc.mainnet.near.org) -H 'Content-Type: application/json' --data '{"jsonrpc": "2.0", "method": "query", "params": 

 {"request_type": "call_function", "account_id": "tipbot.app.near", "block_id": "ArRFkwFsRHMAGxndEMP4G3b1PBSjLVVrk91y9iFYRZRQ", "method_name": "get_deposit", "args_base64": "eyJhY2NvdW50X2lkIjogImZyb2wubmVhciJ9"} 

, "id": "dontcare"}'
 ```

```
 curl [https://rpc.mainnet.near.org](https://rpc.mainnet.near.org) -H 'Content-Type: application/json' --data '{"jsonrpc": "2.0", "method": "query", "params": 

 {"request_type": "view_account", "account_id": "frol.near", "block_id": 30000001} 

, "id": "dontcare"}'
 ```

***Expected behavior***

Ideally, we should be able to tell something like "the data for block X is garbage collected on this node, use archival node to fetch the historical data".

***Version (please complete the following information):***

*   nearcore commit/branch: 1.19.x releases
*   rust version (if local): n/a
*   docker (if using docker): n/a
*   mainnet/testnet/betanet/local: all networks

***Additional context***

This was part of the #3944 refactoring where we introduced fine-grained errors.

bowenwang1996:
@frol @khorolets any update on this?

miraclx:
@frol, can't we reuse the `UnknownBlock` variant, though? Seems to be documented as a catch-all, including the case where we have garbage collected blocks.

https://github.com/near/nearcore/blob/f951d6ed8e491ef842da6bd9448d3242472c126b/chain/client-primitives/src/types.rs#L349-L350

https://github.com/near/nearcore/pull/6342 introduces the `GarbageCollectedBlock` error variant, which is very specific but conflicts with the semantics for `UnknownBlock`.

mina86:
Reporting that the block has been garbage collected relies on DBCol::BlockHeader not being garbage collected but this isn’t something we necessarily want to be true (see #3488).  We probably don’t want to add another dependency on that column.

frol:
[...]

PR #6342, opened 2022-02-23
chain: add RpcQueryError::GarbageCollectedBlock

mooori:
Querying garbage collected historical data from a non-archival node yields `RpcQueryError::UnavailableShard`, see #4340.

#### Changes introduced by this PR
Before returning `RpcQueryError::UnavailableShard`, `ViewClientActor` checks in `handle_query` if the block might have been garbage collected. If yes, it returns `QueryError::GarbageCollectedBlock` (new error variant).

#### Testing
- Start two validator nodes: one non-archival, one archival.
- Let them produce blocks until garbage collection kicks in.
- Query data that might have been garbage collected and verify:
  - The non-archival node returns `QueryError::GarbageCollectedBlock`.
  - The archival node returns the data.

### DIFF
--- a/chain/client/src/view_client.rs
+++ b/chain/client/src/view_client.rs
@@ -195,15 +195,6 @@ impl ViewClientActor {
         }
     }
 
-    fn block_maybe_garbage_collected(&self, block: &BlockHeader) -> Result<bool, Error> {
-        if self.config.archive {
-            return Ok(false);
-        }
-        let tip = self.chain.head()?;
-        let gc_stop_height = self.runtime_adapter.get_gc_stop_height(&tip.last_block_hash);
-        Ok(block.height() < gc_stop_height)
-    }
-
     fn handle_query(&mut self, msg: Query) -> Result<QueryResponse, QueryError> {
         let header = match msg.block_reference {
             BlockReference::BlockId(BlockId::Height(block_height)) => {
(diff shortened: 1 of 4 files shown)


chain/client/src/view_client.rs
```diff
@@ -256,10 +265,18 @@ impl ViewClientActor {
             .shard_id_to_uid(shard_id, header.epoch_id())
             .map_err(|err| QueryError::InternalError { error_message: err.to_string() })?;
 
+        let block_maybe_gc = self.block_maybe_garbage_collected(&header);
```
mm-near, line 268: let's move this check to 272 - so that it triggers only if we see that block is not in DB

mm-near (changes requested):

khorolets (approved):

miraclx (approved):

miraclx (commented):
Once this merges, please update the **"What could go wrong?"** section in the docs:

https://github.com/near/docs/blob/master/docs/api/rpc/contracts.md#what-could-go-wrong-what-could-go-wrong

chain/client/src/view_client.rs
mooori: Addressed in [ced9ad2](https://github.com/near/nearcore/pull/6342/commits/ced9ad28a40bcf25e704ffad304b8a7a2bd588f6). It required some minor refactoring due to `get_chunk_extras` in the outer scope taking `&mut self`.

mm-near (approved):

PR #6345, opened 2022-02-23
refactor: Align gas metering estimation

jakmeier:
Gas metering estimation now also respects CLI parameters
(`--iters` and --`warmups-iters`)

Also avoid runtime panic in case of unstable measurements


matklad (approved):
As a general state of the things, I think gas metering cost was most useful when we were switching to the current impl of the gas counting logic. Nowdays, it's less useful, so I wondn't be against removing the code altogether.

jakmeier:
> As a general state of the things, I think gas metering cost was most useful when we were switching to the current impl of the gas counting logic. Nowdays, it's less useful, so I wondn't be against removing the code altogether.

I guess you have a point there. But I thought this could also be an interesting data point to look at once we have continuous estimation, as this would immediately tell us if we mess up gas metering optimizations somehow.

PR #6347, opened 2022-02-24
feat(params-estimator): Clear IO caches

jakmeier:
Setting up state for estimating benchmarks leads to warm chaches.
But the goal is to measure the times used when caches are cold.
Flushing write buffers and clearing the OS cache helps here.

### DIFF
--- a/runtime/runtime-params-estimator/src/utils.rs
+++ b/runtime/runtime-params-estimator/src/utils.rs
@@ -24,7 +24,8 @@ pub fn read_resource(path: &str) -> Vec<u8> {
 /// Attempts to clear OS page cache on Linux based system. Will fail on
 /// other systems. Requires write access to /proc/sys/vm/drop_caches
 pub fn clear_linux_page_cache() -> io::Result<()> {
-    Command::new("sync").output().expect("sync failed");
+    let out = Command::new("sync").output().expect("sync failed");
+    assert!(out.status.success());
     fs::write("/proc/sys/vm/drop_caches", b"1")
 }
 
--- a/runtime/runtime-params-estimator/src/utils.rs
+++ b/runtime/runtime-params-estimator/src/utils.rs
@@ -4,7 +4,6 @@ use crate::gas_cost::GasCost;
 use crate::transaction_builder::TransactionBuilder;
 
 use std::collections::HashMap;
-use std::process::Command;
 use std::{fs, io};
 
 use near_primitives::transaction::SignedTransaction;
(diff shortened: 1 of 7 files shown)


jakmeier:
Maybe I should also note: The part that drops OS caches requires `sudo` and only works on Linux based systems. While this is not ideal, all others solutions that I tried are even more hacky and not more portable either. Let me know if you have ideas to improve it.

For now, to at least hide this for users that don't need 100% accuracy, this new feature is hidden behind the flag `--drop-os-cache`.

nikurt (approved):

runtime/runtime-params-estimator/src/utils.rs
```diff
@@ -19,6 +21,13 @@ pub fn read_resource(path: &str) -> Vec<u8> {
         .unwrap_or_else(|err| panic!("failed to load test resource: {}, {}", path.display(), err))
 }
 
+/// Attempts to clear OS page cache on Linux based system. Will fail on
+/// other systems. Requires write access to /proc/sys/vm/drop_caches
+pub fn clear_linux_page_cache() -> io::Result<()> {
+    Command::new("sync").output().expect("sync failed");
```
matklad, line 27: Let's assert exit code as well, just to not think about it.

matklad (approved):

runtime/runtime-params-estimator/src/utils.rs
```diff
@@ -19,6 +21,14 @@ pub fn read_resource(path: &str) -> Vec<u8> {
         .unwrap_or_else(|err| panic!("failed to load test resource: {}, {}", path.display(), err))
 }
 
+/// Attempts to clear OS page cache on Linux based system. Will fail on
+/// other systems. Requires write access to /proc/sys/vm/drop_caches
+pub fn clear_linux_page_cache() -> io::Result<()> {
+    let out = Command::new("sync").output().expect("sync failed");
```
nagisa, line 27: To the best of my knowledge this is not necessary. That said, if you'd like to keep it still, I'd change this to call `libc::sync` instead, which is among other things is infallible.

jakmeier: Hm, I was not able to find anything confirming tor deny the necessity of a sync here. From [man proc(5)](https://man7.org/linux/man-pages/man5/proc.5.html)

>  /proc/sys/vm/drop_caches (since Linux 2.6.16)
              Writing to this file causes the kernel to drop clean
              caches, dentries, and inodes from memory, causing that
              memory to become free.  This can be useful for memory
              management testing and performing reproducible filesystem
              benchmarks.  Because writing to this file causes the
              benefits of caching to be lost, it can degrade overall
              system performance.

It sounds to me that recently written pages, that might still be dirty, would not necessarily be cleaned. So to be sure, I would like to keep the sync.

Using libc makes a lot of sense, though! I also decided to put it behind some `cfg!`, seems cleaner.

nagisa (approved):

PR #526, opened 2019-02-05
Fix network deadlock and improve tests

nearmax:

PR #6351, opened 2022-02-25
Upgrade rust-rocksdb from 0.16 to 0.18

EdvardD:
Upgrade rust-rocksdb from 0.16 to 0.18;
Make sse4.1, sse4.2 cpu instructions a requirement for running a node.

Brings a log of compilation warnings while compiling on aarch64 because of this issue: https://github.com/rust-lang/cargo/issues/10439.

Commit blocking 0.16->0.18 switch: 'Make SSE inclusion conditional for target features (#526)' (`81a9edea83473012378e808606cdd92c1212c076`)
Performance after the change: https://pastebin.com/z0uu0dyg
Performance in master: https://pastebin.com/ECjLFptg


Makefile
```diff
@@ -1,7 +1,7 @@
 export CARGO_PROFILE_RELEASE_CODEGEN_UNITS = 1
 export CARGO_PROFILE_RELEASE_LTO = fat
 export DOCKER_BUILDKIT = 1
-export RUSTFLAGS = -D warnings
+export RUSTFLAGS = "-D warnings -C target-cpu=native"
```
nagisa, line 4: We cannot do this. The binary built by the `Makefile` should be able to run on any machine that satisfies the minimum requirements, which are going to be significantly lower than those available on a typical build machine.

EdvardD: It seems `rust-rocksdb` 0.16 have unconditionally enabled sse2/sse4.1/sse4.2 now: https://github.com/rust-rocksdb/rust-rocksdb/blob/v0.16.0/librocksdb-sys/build.rs#L106.
Let me look for a workaround.

EdvardD: What about enabling only sse2/sse4.1/sse4.2 since they're in fact already a requirement (implicit) for nodes?

nikurt (approved):

Makefile
nagisa: I originally suggested doing something like that in https://github.com/near/nearcore/issues/5487. I personally think it makes sense to enable at least some of these features unconditionally, but others might disagree.

Makefile
```diff
@@ -1,7 +1,7 @@
 export CARGO_PROFILE_RELEASE_CODEGEN_UNITS = 1
 export CARGO_PROFILE_RELEASE_LTO = fat
 export DOCKER_BUILDKIT = 1
-export RUSTFLAGS = "-D warnings -C target-cpu=native"
+export RUSTFLAGS = "-D warnings -C target-feature=+sse2 -C target-feature=+sse4.1 -C target-feature=+sse4.2"
```
nagisa, line 4: you can avoid some of the duplication by

```
-Ctarget-feature=+sse2,+sse4.1,+sse4
```

Additionally `sse2` is part of the baseline `x86_64` target already, so there is no need to specify it.

mina86: Is there a way we could specify this in Cargo.toml or some place like that? While our release binaries use Makefile and our instructions for validators are to build with Makefile it would be nice to have faster code during testing and development as well.

mina86: This line isn’t necessary since we have .cargo/config.toml, is it?

Makefile
EdvardD: I see. Yeah that's cool. Sorry I didn't understand how it's related first time you shared the PR. How do we sync these 2 PRs? Should I wait for you or I can go ahead?

.cargo/config.toml
```diff
@@ -0,0 +1,2 @@
+[build]
+rustflags = ["-Ctarget-feature=+sse4.1,+sse4.2"]
```
nagisa, line 2: These are target-specific flags and should only be applied to x86_64 builds in order to not break aarch64 builds.

```
[target.cfg(target_arch="x86_64")]
rustflags = ["-Ctarget-feature=+sse4.1,+sse4.2"]
```

nagisa (commented):
Target feature changes make sense to me. I'm not sure however if we need to audit the rocksdb changes to see if there's any potential protocol breaking changes that might occur as a result of this upgrade.

Makefile
mina86: > I see. Yeah that's cool. Sorry I didn't understand how it's related first time you shared the PR. How do we sync these 2 PRs? Should I wait for you or I can go ahead?

To me it seems like we should have three separate PRs:
1. Enable SSE et al in build flags.
2. Add CPU features checking to release builds.
3. Upgrade RocksDB.

Currently we have two PRs. One which covers 1 and 2 and another which covers 1 and 3.

nagisa: > I see. Yeah that's cool. Sorry I didn't understand how it's related first time you shared the PR. How do we sync these 2 PRs? Should I wait for you or I can go ahead?

Feel free to land your changes first; I don't see myself coming back to my own PR anytime soon.

---

> Add CPU features checking to release builds.

This is already a thing, my PR just refactors it, but lets discuss that particular detail over there rather than here.

> To me it seems like we should have three separate PRs:

On one hand I'm inclined to agree. More granluar changesets would avoid blocking part of the PR that seems good based on concerns with a different part of the PR. Smaller changesets are easier to review as well. On another hand, there are also a reason to keep 1 and 3 together as well – there isn't all that much of a reason to make the 1st change in isolation if we end up never making the 3rd one.

.cargo/config.toml
EdvardD: Thanks, fixed. Shouldn't smth similar be done in the Makefile?

EdvardD:
@nagisa rocksdb 0.18 was already merged before (notice that I've cherry-picks @nikurt 's commit here and added the needed flags) and reverted because of rocksdb read performance degradation. Now we found the reason behind it so I suppose we should be okay to go.

Makefile
EdvardD: I'm okay with splitting this to 2 commits. Please let me know if it's a blocker.

On the other hand I agree that those should go together: hypothetically if later on we decide to revert flags change it can lead to performance degradation if we don't revert 0.16 -> 0.18 as well.

nagisa (approved):

Makefile
```diff
@@ -1,15 +1,17 @@
...
 export CARGO_PROFILE_RELEASE_LTO = fat
 export DOCKER_BUILDKIT = 1
-export RUSTFLAGS = -D warnings
 export NEAR_RELEASE_BUILD = no
 export CARGO_TARGET_DIR = target
 
+export RUSTFLAGS = -D warnings
+ifeq ($(shell uname -m),x86_64)
+  export RUSTFLAGS := ${RUSTFLAGS} -Ctarget-feature=+sse4.1,+sse4.2
+endif
```
mina86, line 10: So I’m always unsure about such changes to Makefile.  Specifically, this is GNU Make syntax and it’s my understanding that we support builds on Macs and thus must work with intersection supported by BSD Make.

jakmeier (approved):
Adding my approval for params estimator changes.

.cargo/config.toml
```diff
@@ -0,0 +1,2 @@
+[target.'cfg(target_arch="x86_64")']
+rustflags = ["-Ctarget-feature=+sse4.1,+sse4.2"]
```
mm-near, line 2: I'd also update the name of the PR (to mention that we're enabling SSE flags)

mm-near (approved):

Makefile
EdvardD: In fact MacOS uses GNU make so no problem with that. .cargo/config.toml doesn't apply for some reason. I'm trying to figure out why that's the case.

EdvardD: Okay I found the issue: 'RUSTFLAGS = -D warnings' was overriding rustflags coming from config so I moved it to the config as well.

EdvardD (commented):
It turned out to be impossible configuring flags per target (x86_64) so they're passed to rust-rocksdb/librocksdb-sys/build.rs so we do that with `[build]` unconditionally adding flags for all cpu types. We expect this to not break M1 compilation though.

`-Dwarnings` flag was moved from Makefile to config.toml so it doesn't override flags specified in the config.

.cargo/config.toml
```diff
@@ -0,0 +1,2 @@
+[build]
+rustflags = ["-Dwarnings", "-Ctarget-feature=+sse4.1,+sse4.2"]
```
nagisa, line 2: I think having `-Dwarnings` here will mean that during development there cannot be any warnings at any time, otherwise the compilation will fail. In my experience this is not a deal breaker, but having to keep codebase warning-free at all times definitely makes development way less straightforward.

EdvardD: Either we do this or having GNU make syntax (which works for Mac BTW).

nagisa: We could plausibly not set `-Dwarnings` in either place. The most appropriate place for `-Dwarnings` would be in some CI job, probably.

EdvardD: And yeah if we have it this way it won't compile on M1 processors because sse4.1 warnings will turn into errors. I would propose to go with having it in Makefile.

nagisa: `-Dwarnings` in the Makefile would still mean being unable to use the Makefile when building for aarch64.

EdvardD: It won't if we enable sse4.1 only for x86_64:
```ifeq ($(shell uname -m),x86_64)
  export RUSTFLAGS := ${RUSTFLAGS} -Ctarget-feature=+sse4.1,+sse4.2
endif```

Comment (EdvardD) [reply .cargo/config.toml:2]:
```
[build.rustflags](https://doc.rust-lang.org/cargo/reference/config.html#buildrustflags)
Type: string or array of strings
Default: none
Environment: CARGO_BUILD_RUSTFLAGS or RUSTFLAGS
Extra command-line flags to pass to rustc. The value may be a array of strings or a space-separated string.

There are three mutually exclusive sources of extra flags. They are checked in order, with the first one being used:

RUSTFLAGS environment variable.
All matching target.<triple>.rustflags and target.<cfg>.rustflags config entries joined together.
build.rustflags config value.
```

nagisa (approved):

bowenwang1996 (approved):

PR #6354, opened 2022-02-28
Adding debug page to nearcore with latest blocks

mm-near:
It is disabled by default - you have to set 'enable_debug_rpc' in the config to enable.


chain/chain/src/chain.rs
```diff
@@ -392,6 +397,61 @@ pub fn check_known(
...
+
+impl CryptoHashTimer {
+    pub fn new(key: CryptoHash) -> Self {
+        CryptoHashTimer { key, start: Clock::instant() }
+    }
+    pub fn get_timer_value(key: CryptoHash) -> Option<u64> {
+        match CRYPTO_HASH_TIMER_RESULTS.lock().unwrap().get(&key) {
+            Some(&val) => Some(val),
+            None => None,
+        }
```
mina86, line 427: That’s just `CRYPTO_HASH_TIMER_RESULTS.lock().unwrap().get(&key).cloned()`.

chain/jsonrpc/src/lib.rs
```diff
@@ -70,6 +74,10 @@ pub struct RpcConfig {
     pub polling_config: RpcPollingConfig,
     #[serde(default)]
     pub limits_config: RpcLimitsConfig,
+    // If true, enable some debug RPC endpoints (like one to get the latest block).
+    // We disable it by default, as some of those endpoints might be quite CPU heavy.
+    #[serde(default = "default_enable_debug_rpc")]
+    pub enable_debug_rpc: bool,
```
mina86, line 80: I’m tempted to say we should just always enable it without configuration option. This doesn’t leak any sensitive information (I don’t think) and if someone has access to JSON RPC end point they can easily issue more expensive requests if they want to DoS the node.

core/primitives/src/views.rs
```diff
@@ -320,6 +320,31 @@ pub struct ValidatorInfo {
     pub is_slashed: bool,
 }
 
+#[cfg_attr(feature = "deepsize_feature", derive(deepsize::DeepSizeOf))]
+#[derive(Serialize, Deserialize, Debug)]
+pub struct DebugChunkStatus {
+    pub shard_id: u64,
+    pub chunk_hash: ChunkHash,
+    pub gas_used: u64,
+    pub processing_time_ms: u64,
```
mina86, line 329: Alternatively make it an Option. Or maybe skip serialisation if zero? Same in DebugBlockStatus.

mina86 (approved):

chain/jsonrpc/res/debugz.html
```diff
@@ -0,0 +1,7 @@
+<html>
+
+<body>
+    <h1><a href="/debugz/last_blocks">Last blocks</a></h1>
```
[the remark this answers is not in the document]
mm-near: this is the 'index' page - when we add more pages (like yours) - then we can add the URLs here.

chain/jsonrpc/src/lib.rs
```diff
@@ -1413,6 +1484,9 @@ pub fn start_http(
             )
             .service(web::resource("/network_info").route(web::get().to(network_info_handler)))
             .service(web::resource("/metrics").route(web::get().to(prometheus_handler)))
+            .service(web::resource("/debugz/api/last_blocks").route(web::get().to(debugz_handler)))
```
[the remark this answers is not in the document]
mm-near: This is done above the functions: 
#[get("/debugz")]

integration-tests/src/tests/nearcore/stake_nodes.rs
```diff
@@ -231,81 +234,97 @@ fn test_validator_kickout() {
...
+                                                        }
+                                                        _ => panic!("wrong return result"),
+                                                    }
+                                                }),
+                                        );
+                                    }
 
-                                if finalized_mark1.iter().all(|mark| mark.load(Ordering::SeqCst)) {
-                                    System::current().stop();
+                                    if finalized_mark1
```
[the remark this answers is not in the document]
mm-near: yeah.. I've changed one line in 240 - and then the editor refactored everything..

I'll check what kind of style settings we should be using

chain/jsonrpc/src/lib.rs
```diff
@@ -70,6 +74,10 @@ pub struct RpcConfig {
     pub polling_config: RpcPollingConfig,
     #[serde(default)]
     pub limits_config: RpcLimitsConfig,
+    // If true, enable some debug RPC endpoints (like one to get the latest block).
+    // We disable it by default, as some of those endpoints might be quite CPU heavy.
+    #[serde(default = "default_enable_debug_rpc")]
```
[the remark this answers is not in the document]
mm-near: default value is set by the function above (line 64).

The values are normally read from config.json file in the HOME dir (for example ~/.near/config.json)

chain/jsonrpc/src/lib.rs
mm-near: I'd rather play it safe - and start with disabled. We can always enable it in the future with 1-line PR

chain/jsonrpc/res/last_blocks.html
```diff
@@ -0,0 +1,126 @@
...
+
+        tr.active {
+            background-color: #eff8bf;
+        }
+    </style>
+    <script src="https://ajax.googleapis.com/ajax/libs/jquery/3.5.1/jquery.min.js"></script>
+    <script>
+        $(document).ready(function () {
+            var root = $("#blocks_table_tbody");
+            root[0].innerHTML = "Loading...";
```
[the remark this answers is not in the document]
mm-near: good point - fixed.

chain/jsonrpc/res/last_blocks.html
```diff
@@ -0,0 +1,126 @@
...
+                            if (chunk.shard_id > max_shard) {
+                                max_shard = chunk.shard_id;
+                            }
+                        });
+                    });
+
+                    prepareTableHeader(max_shard);
+                    data.detailed_debug_status.last_blocks.forEach(element => {
+                        console.log(element.block_hash);
+                        var row = document.createElement("tr");
```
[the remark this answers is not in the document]
mm-near: good idea - done.

chain/chain/src/chain.rs
mm-near: done - thanks.

core/primitives/src/views.rs
mm-near: good idea - done

chain/jsonrpc/src/lib.rs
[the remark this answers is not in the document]
jc-near: Cool.. So what does this line do then?

chain/chain/src/chain.rs
```diff
@@ -1288,6 +1345,8 @@ impl Chain {
     ) -> Result<Option<Tip>, Error> {
         metrics::BLOCK_PROCESSING_ATTEMPTS_TOTAL.inc();
         metrics::NUM_ORPHANS.set(self.orphans.len() as i64);
+        let block_hash = *block.hash();
+        let _timer = CryptoHashTimer::new(block_hash);
```
mzhangmzz, line 1349: This timer may not generate the time that you want. See how 'success_timer' is done here. `process_block` may return early because the block is orphan, or missing chunks, or the block is already known.

chain/chain/src/chain.rs
```diff
@@ -391,6 +396,58 @@ pub fn check_known(
     check_known_store(chain, block_hash)
 }
 
+pub static CRYPTO_HASH_TIMER_RESULTS: Lazy<Mutex<LruCache<CryptoHash, u64>>> =
+    Lazy::new(|| Mutex::new(LruCache::new(10000)));
```
mzhangmzz, line 400: Probably deserves some comments here on what crypto hashes and times you are putting here, since this is a global variable.

core/primitives/src/views.rs
```diff
@@ -322,6 +322,33 @@ pub struct ValidatorInfo {
...
+}
+
+#[cfg_attr(feature = "deepsize_feature", derive(deepsize::DeepSizeOf))]
+#[derive(Serialize, Deserialize, Debug)]
+pub struct DebugBlockStatus {
+    pub block_hash: CryptoHash,
+    pub block_height: u64,
+    pub chunks: Vec<DebugChunkStatus>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub processing_time_ms: Option<u64>,
```
mzhangmzz, line 342: comment the fields `processing_time_ms` and `timestamp_delta`, what they mean, how they are calculated

chain/client/src/client_actor.rs
```diff
@@ -662,6 +664,50 @@ impl Handler<Status> for ClientActor {
...
+                        processing_time_ms: CryptoHashTimer::get_timer_value(chunk.chunk_hash().0),
+                    })
+                    .collect();
+
+                blocks_debug.push(DebugBlockStatus {
+                    block_hash: last_block_hash,
+                    block_height: block.header().height(),
+                    chunks,
+                    processing_time_ms: CryptoHashTimer::get_timer_value(last_block_hash),
+                    timestamp_delta: if last_block_timestamp > 0 {
```
mzhangmzz, line 697: If I understand correctly, `timestamp_delta` is the time difference between the timestamp of this block and the next block. I think this could be confusing. Intuitively, when I look at the stats, I would assume that this number is the timestamp between the previous block and this block. 

I understand implementation wise, the current way is easier. But this may cause confusion when looking at the debug page number

chain/jsonrpc/src/lib.rs
```diff
@@ -864,7 +874,11 @@ impl JsonRpcHandler {
         near_jsonrpc_primitives::types::status::RpcHealthResponse,
         near_jsonrpc_primitives::types::status::RpcStatusError,
     > {
-        Ok(self.client_addr.send(Status { is_health_check: true }).await??.into())
+        Ok(self
+            .client_addr
+            .send(Status { is_health_check: true, get_detailed_info: false })
```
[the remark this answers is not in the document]
mm-near: yes - this is a good idea. I'll do it in a separate PR.

chain/jsonrpc/src/lib.rs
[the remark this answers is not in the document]
mm-near: There are two ways how do 'assign' the URL:
if we want to use the JSONRPC class - we pass the web resource here.

If we just 'return' as static HTML - I'm using that #get option.

chain/chain/src/chain.rs
mm-near: yes, but the 'final' run should be the successful one, right? 
 (as without it, the block would not be added to the chain)

chain/chain/src/chain.rs
mm-near: good idea - done

core/primitives/src/views.rs
mm-near: Done.

chain/client/src/client_actor.rs
mm-near: yes, this is easier implementation wise, but I agree that it might be confusing in the current HTML.

very soon, I plan to update that HTML (adding colors etc) - and it should be more clear for the user.

chain/chain/src/chain.rs
mzhangmzz: Not necessarily. It's possible that the node receives a block that's already processed and `process_block` will return early because the block is known (see `check_known`)

mzhangmzz (approved):
Will approve to unblock, but please see my comments

chain/chain/src/chain.rs
```diff
@@ -391,6 +396,60 @@ pub fn check_known(
     check_known_store(chain, block_hash)
 }
 
+// Cache with the mappint from CryptoHash (blocks, chunks) to the number milliseconds that it took to process them.
+// Used only for debugging purposes.
+pub static CRYPTO_HASH_TIMER_RESULTS: Lazy<Mutex<LruCache<CryptoHash, u64>>> =
+    Lazy::new(|| Mutex::new(LruCache::new(10000)));
```
matklad, line 402: Can we make this a field of some struct, rather than a global variable? Global state is bad: it's easy to add, but hard to remove. For diagnostics use-case this is usually acceptable, but best avoided anyway.

chain/chain/src/chain.rs
```diff
@@ -391,6 +396,60 @@ pub fn check_known(
     check_known_store(chain, block_hash)
 }
 
+// Cache with the mappint from CryptoHash (blocks, chunks) to the number milliseconds that it took to process them.
+// Used only for debugging purposes.
+pub static CRYPTO_HASH_TIMER_RESULTS: Lazy<Mutex<LruCache<CryptoHash, u64>>> =
```
matklad, line 401:
```suggestion
pub static CRYPTO_HASH_TIMER_RESULTS: Lazy<Mutex<LruCache<CryptoHash, Duration>>> =
```

Seems we can stick to strong types here?

chain/chain/src/chain.rs
```diff
@@ -391,6 +396,60 @@ pub fn check_known(
...
+        CRYPTO_HASH_TIMER_RESULTS.lock().unwrap().get(&key).cloned()
+    }
+}
+
+impl Drop for CryptoHashTimer {
+    fn drop(&mut self) {
+        CRYPTO_HASH_TIMER_RESULTS
+            .lock()
+            .unwrap()
+            .put(self.key, (Clock::instant() - self.start).as_millis() as u64);
```
matklad, line 434: Let's extract `(Clock::instant() - self.start).as_millis() as u64` to a variable, so that it's outside of the lock.

chain/chain/src/chain.rs
```diff
@@ -391,6 +396,60 @@ pub fn check_known(
...
+/// {
+///     let _timer = CryptoHashTime(block.hash);
+///     // Do stuff with the block
+///     // timer gets released at the end of this block
+/// }
+/// ```
+pub struct CryptoHashTimer {
+    key: CryptoHash,
+    start: Instant,
+}
```
matklad, line 418: Consider moving this machinery to a separate file/module -- it feels relatively self-contained, and the present file is already quite large.

chain/chain/src/chain.rs
```diff
@@ -391,6 +396,60 @@ pub fn check_known(
...
+        CRYPTO_HASH_TIMER_RESULTS
+            .lock()
+            .unwrap()
+            .put(self.key, (Clock::instant() - self.start).as_millis() as u64);
+    }
+}
+
+#[test]
+fn test_crypto_hash_timer() {
+    let crypto_hash = CryptoHash::from_str("s3N6V7CNAN2Eg6vfivMVHR4hbMZeh72fTmYbrC6dXBT").unwrap();
```
matklad, line 440:
```suggestion
    let crypto_hash: CryptoHash = "s3N6V7CNAN2Eg6vfivMVHR4hbMZeh72fTmYbrC6dXBT").parse().unwrap();
```

Rule of thumb: only import `FromStr` if you want to write an `impl`. The interface for *using* this trait is `str::parse` method.

matklad: For the future: rather than gating imports with `#[cfg(test)] use ...` use the following pattern instead:

```rust
#[cfg(test)]
mod tests {
  // Bring everything from the parent module into the scope
  use super::*;

  // Any additional imports now don't need to be gated
  use foo::bar::baz;

  #[test]
  fn test_crypto_hash_timer() {
  }
}
```

If you really have just a single test function, the following would also work:

```rust
#[test]
fn test_crypto_hash_timer() {
  use foo::bar::baz;
}
```

chain/chain/src/chain.rs
matklad, line 434: We probably should document *overriding* behavior. Also, do we want to override here? Perhaps it's better to accumulate (`+`) instead?

chain/chain/src/chain.rs
```diff
@@ -391,6 +396,60 @@ pub fn check_known(
...
+    // Timer should be missing.
+    assert_eq!(CryptoHashTimer::get_timer_value(crypto_hash), None);
+    let mock_clock_guard = near_primitives::time::MockClockGuard::default();
+    mock_clock_guard.add_instant(Instant::now());
+    mock_clock_guard.add_instant(Instant::now() + std::time::Duration::from_secs(1));
+    {
+        let _timer = CryptoHashTimer::new(crypto_hash);
+    }
+    // Timer should show exactly 1 second (1000 ms).
+    assert_eq!(CryptoHashTimer::get_timer_value(crypto_hash), Some(1000));
```
matklad, line 450: Nice that we can actually write such tests now! Hopefully one day we'll replace global `Clock::instant` with explicitly injected

```rust
fn foo(clock: &Clock) {
  let now = clock.instant();
}
```

chain/client-primitives/src/types.rs
```diff
@@ -358,6 +358,8 @@ pub enum QueryError {
 
 pub struct Status {
     pub is_health_check: bool,
+    // If true - return more detailed information about the current status (recent blocks etc).
+    pub get_detailed_info: bool,
```
matklad, line 362: maybe just

```suggestion
    pub detailed: bool,
```

?

chain/jsonrpc/res/debugz.html
```diff
@@ -0,0 +1,7 @@
...
+</body>
+
+</html>
```
matklad, line 7: trailing newline missing

chain/jsonrpc/res/last_blocks.html
```diff
@@ -0,0 +1,120 @@
...
+            vertical-align: center;
+            padding: 8px;
+            background-color: lightgrey;
+        }
+
+        tr.active {
+            background-color: #eff8bf;
+        }
+    </style>
+    <script src="https://ajax.googleapis.com/ajax/libs/jquery/3.5.1/jquery.min.js"></script>
```
matklad, line 33: No strong opinion, but I'd stick to modern built-in APIs like `fetch` and friends.

chain/jsonrpc/res/last_blocks.html
```diff
@@ -0,0 +1,120 @@
...
+                },
+                contentType: "application/json; charset=utf-8",
+            })
+
+        });
+
+        function prepareTableHeader(max_shard) {
+            var headers = '<th>Block id</th> <th>Block hash</th><th>Processing Time(ms)</th><th>Next block time(s)</th>';
+            for (let i = 0; i < max_shard + 1; i += 1) {
+                headers += '<th colspan=3>Shard ' + i + ' (hash/gas/time)</th>';
```
matklad, line 99:
```suggestion
                headers += `<th colspan=3>Shard ${i} (hash/gas/time)</th>`;
```

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals

We live in the future -- every language supports string interpolation!

chain/jsonrpc/src/lib.rs
matklad: Second mina86's opinion. I do think that we can enable this later, but this would be more risky: it's easier to audit new code for production readiness than to audit existing debug-only code.

chain/client/src/client_actor.rs
```diff
@@ -671,6 +673,50 @@ impl Handler<Status> for ClientActor {
...
+                    Err(_) => break,
+                };
+                let chunks = block
+                    .chunks()
+                    .iter()
+                    .map(|chunk| DebugChunkStatus {
+                        shard_id: chunk.shard_id(),
+                        chunk_hash: chunk.chunk_hash(),
+                        gas_used: chunk.gas_used(),
+                        processing_time_ms: CryptoHashTimer::get_timer_value(chunk.chunk_hash().0),
```
matklad, line 697: This feels potentially abusable -- here we lock that global mutex 50 times, which might contend with the actual processing work.

chain/jsonrpc/src/lib.rs
```diff
@@ -873,7 +887,30 @@ impl JsonRpcHandler {
...
     > {
-        Ok(self.client_addr.send(Status { is_health_check: false }).await??.into())
+        Ok(self
+            .client_addr
+            .send(Status { is_health_check: false, get_detailed_info: false })
+            .await??
+            .into())
+    }
+
+    pub async fn debugz(
```
matklad, line 897: Why `debugz` rather just `debug`?

core/primitives/src/views.rs
```diff
@@ -322,6 +322,35 @@ pub struct ValidatorInfo {
...
 }
 
+#[cfg_attr(feature = "deepsize_feature", derive(deepsize::DeepSizeOf))]
+#[derive(Serialize, Deserialize, Debug)]
+pub struct DebugChunkStatus {
+    pub shard_id: u64,
+    pub chunk_hash: ChunkHash,
+    pub gas_used: u64,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub processing_time_ms: Option<u64>,
```
matklad, line 332: Regarding `Duration` vs `u64`, I think this (JSON RPC layer) is the only place were we need to use `u64`. Everything else should use Duration.

In theory, we can even use `Duration` here, as it is `Serialize`, but, IIRC, it's JSON is rather ugly.

chain/jsonrpc/src/lib.rs
```diff
@@ -1360,6 +1409,21 @@ fn get_cors(cors_allowed_origins: &[String]) -> Cors {
         .max_age(3600)
 }
 
+lazy_static_include::lazy_static_include_str! {
```
matklad, line 1412: :+1: this is indeed the intended use-case for lazy_static_include!

chain/jsonrpc/src/lib.rs
```diff
@@ -1310,6 +1347,18 @@ fn status_handler(
...
+    handler: web::Data<JsonRpcHandler>,
+) -> impl Future<Output = Result<HttpResponse, HttpError>> {
+    let response = async move {
+        match handler.debugz().await {
+            Ok(value) => Ok(HttpResponse::Ok().json(&value)),
+            Err(_) => Ok(HttpResponse::MethodNotAllowed().finish()),
+        }
+    };
+    response.boxed()
+}
```
matklad, line 1360: I wonder if this can be just

```suggestion
async fn debugz_handler(
    handler: web::Data<JsonRpcHandler>,
) -> Result<HttpResponse, HttpError> {
        match handler.debugz().await {
            Ok(value) => Ok(HttpResponse::Ok().json(&value)),
            Err(_) => Ok(HttpResponse::MethodNotAllowed().finish()),
        }
}
```

?

Can't check immediatelly -- waiting for nearcore to compile on my trusty old laptop :) Already miss my 12 core machine :(

matklad, line 1360: I wonder if this can be just

```suggestion
async fn debugz_handler(
    handler: web::Data<JsonRpcHandler>,
) -> Result<HttpResponse, HttpError> {
        match handler.debugz().await {
            Ok(value) => Ok(HttpResponse::Ok().json(&value)),
            Err(_) => Ok(HttpResponse::MethodNotAllowed().finish()),
        }
}
```

?

Can't check immediatelly -- waiting for nearcore to compile on my trusty old laptop :) Already miss my 12 core machine :(

chain/chain/src/chain.rs
```diff
@@ -391,6 +396,60 @@ pub fn check_known(
...
+///
+/// Example use:
+/// ```rust, ignore
+/// {
+///     let _timer = CryptoHashTime(block.hash);
+///     // Do stuff with the block
+///     // timer gets released at the end of this block
+/// }
+/// ```
+pub struct CryptoHashTimer {
```
matklad, line 415:
```suggestion
/// ```
#[must_use]
pub struct CryptoHashTimer {
```

to give a warning if the user forgets `let _time =`

chain/jsonrpc/src/lib.rs
matklad: Compiled!

So, yeah, this seems to work

```diff
diff --git a/chain/jsonrpc/src/lib.rs b/chain/jsonrpc/src/lib.rs
index dcbe658f1..92c00b0c3 100644
--- a/chain/jsonrpc/src/lib.rs
+++ b/chain/jsonrpc/src/lib.rs
@@ -1347,16 +1347,13 @@ fn status_handler(
     response.boxed()
 }
 
-fn debugz_handler(
+async fn debugz_handler(
     handler: web::Data<JsonRpcHandler>,
-) -> impl Future<Output = Result<HttpResponse, HttpError>> {
-    let response = async move {
-        match handler.debugz().await {
-            Ok(value) => Ok(HttpResponse::Ok().json(&value)),
-            Err(_) => Ok(HttpResponse::MethodNotAllowed().finish()),
-        }
-    };
-    response.boxed()
+) -> Result<HttpResponse, HttpError> {
+    match handler.debugz().await {
+        Ok(value) => Ok(HttpResponse::Ok().json(&value)),
+        Err(_) => Ok(HttpResponse::MethodNotAllowed().finish()),
+    }
 }
 
 fn health_handler(
```

While working on, it I've noticed another thing: we treat any error as debug being disabled by configuration. That is, if gettind debug info fails at runtime, the error will be silently swallowed. We'll see a 400, not a 500. To fix *that*, I'd do

```diff
diff --git a/chain/jsonrpc/src/lib.rs b/chain/jsonrpc/src/lib.rs
index dcbe658f1..7d77942d4 100644
--- a/chain/jsonrpc/src/lib.rs
+++ b/chain/jsonrpc/src/lib.rs
@@ -897,20 +897,17 @@ impl JsonRpcHandler {
     pub async fn debugz(
         &self,
     ) -> Result<
-        near_jsonrpc_primitives::types::status::RpcStatusResponse,
+        Option<near_jsonrpc_primitives::types::status::RpcStatusResponse>,
         near_jsonrpc_primitives::types::status::RpcStatusError,
     > {
         if self.enable_debug_rpc {
-            Ok(self
-                .client_addr
-                .send(Status { is_health_check: false, get_detailed_info: true })
-                .await??
-                .into())
-        } else {
-            Err(near_jsonrpc_primitives::types::status::RpcStatusError::InternalError {
-                error_message: "Debug not enabled in configuration.".to_string(),
-            })
+            return None;
         }
+        let status_response = self
+            .client_addr
+            .send(Status { is_health_check: false, get_detailed_info: true })
+            .await??;
+        Ok(Some(status_response.into()))
     }
 
     /// Expose Genesis Config (with internal Runtime Config) without state records to keep the
@@ -1347,16 +1344,12 @@ fn status_handler(
     response.boxed()
 }
 
-fn debugz_handler(
-    handler: web::Data<JsonRpcHandler>,
-) -> impl Future<Output = Result<HttpResponse, HttpError>> {
-    let response = async move {
-        match handler.debugz().await {
-            Ok(value) => Ok(HttpResponse::Ok().json(&value)),
-            Err(_) => Ok(HttpResponse::MethodNotAllowed().finish()),
-        }
-    };
-    response.boxed()
+async fn debugz_handler(handler: web::Data<JsonRpcHandler>) -> Result<HttpResponse, HttpError> {
+    match handler.debugz().await {
+        Ok(Some(value)) => Ok(HttpResponse::Ok().json(&value)),
+        Ok(None) => Ok(HttpResponse::MethodNotAllowed().finish()),
+        Err(_) => Ok(HttpResponse::ServiceUnavailable().finish()),
+    }
 }
 
 fn health_handler(
```

chain/chain/src/chain.rs
mm-near: ack

chain/chain/src/chain.rs
mm-near: done

chain/chain/src/chain.rs
mm-near: done.

chain/chain/src/chain.rs
mm-near: done

chain/client-primitives/src/types.rs
mm-near: done

chain/client/src/client_actor.rs
mm-near: yes - that's why I wanted this to be disabled by default (and enabled only on some selected canary nodes that we own)

chain/jsonrpc/res/debugz.html
mm-near: done - but my editor keeps removing it..

chain/jsonrpc/res/last_blocks.html
mm-near: my javascript skills are still stuck very much in the past ;-)

chain/jsonrpc/res/last_blocks.html
mm-near: I'll read about this -- but that would be a separate PR

chain/jsonrpc/src/lib.rs
mm-near: done

chain/jsonrpc/src/lib.rs
mm-near: ack

chain/jsonrpc/src/lib.rs
mm-near: done

core/primitives/src/views.rs
mm-near: ok

chain/chain/src/chain.rs
mm-near: good idea - done.

chain/chain/src/chain.rs
mm-near: good idea - will do.

chain/chain/src/chain.rs
mm-near: Yes, I'd also have liked to avoid global variable - but I don't see a good way around it right now (unless we want to push this object through couple layers).

As you've mentioned - this is for diagnostics, so it should be ok.

chain/chain/src/crypto_hash_timer.rs
```diff
@@ -0,0 +1,72 @@
...
+use std::{
+    sync::Mutex,
+    time::{Duration, Instant},
+};
+
+use once_cell::sync::Lazy;
+
+// Cache with the mapping from CryptoHash (blocks, chunks) to the number milliseconds that it took to process them.
+// Used only for debugging purposes.
+pub static CRYPTO_HASH_TIMER_RESULTS: Lazy<Mutex<LruCache<CryptoHash, Duration>>> =
```
matklad, line 12:
```suggestion
static CRYPTO_HASH_TIMER_RESULTS: Lazy<Mutex<LruCache<CryptoHash, Duration>>> =
```

I thing should be enough. If not, that's suspect -- publicly visible global state is especially not great

chain/chain/src/crypto_hash_timer.rs
```diff
@@ -0,0 +1,72 @@
...
+    pub fn get_timer_value(key: CryptoHash) -> Option<Duration> {
+        CRYPTO_HASH_TIMER_RESULTS.lock().unwrap().get(&key).cloned()
+    }
+}
+
+impl Drop for CryptoHashTimer {
+    fn drop(&mut self) {
+        let time_passed = Clock::instant() - self.start;
+        let mut guard_ = CRYPTO_HASH_TIMER_RESULTS.lock().unwrap();
+        let previous = guard_.get(&self.key).unwrap_or(&(Duration::ZERO.clone())).clone();
```
matklad, line 45:
```suggestion
        let previous = guard_.get(&self.key).copied().unwrap_or_default();
```

chain/jsonrpc/res/last_blocks.html
```diff
@@ -0,0 +1,120 @@
...
+                },
+                contentType: "application/json; charset=utf-8",
+            })
+
+        });
+
+        function prepareTableHeader(max_shard) {
+            var headers = '<th>Block id</th> <th>Block hash</th><th>Processing Time(ms)</th><th>Next block time(s)</th>';
+            for (let i = 0; i < max_shard + 1; i += 1) {
+                headers += '<th colspan=3>Shard ${i} (hash/gas/time)</th>';
```
matklad, line 99: Hm, does this work? You want backtics here, not single quotes I think (``` ' vs ` ```)

matklad (approved):
LGTM, but I think theer's a wrong quote (=bug)

rtsai123 (approved):

Issue #6242, opened 2022-02-03
Archivel nodes should GC ColPartialChunks


mina86:
ColPartialChunks is the largest column in archival nodes while at the same time all information stored in it are available in ColChunks column. It would be big win if we could GC the column in archival nodes.

PR #6356, opened 2022-02-28
chain: refactor `process_partial_encoded_chunk_request` method

mina86:
Extract two more methods from `process_partial_encoded_chunk_request`
which correspond to `if` bodies that used to be in it.  This makes each
function shorter and thus easier to read especially as in the future
more branches will be added to the method.

Furthermore, move sending of the message to the method changing
`maybe_send_partial_encoded_chunk_response` into method which prepares
the response only.

This is a pure refactoring with no changes to the behaviour.

### DIFF
--- a/chain/chunks/src/lib.rs
+++ b/chain/chunks/src/lib.rs
@@ -998,70 +998,98 @@ impl ShardsManager {
     ) {
         debug!(target: "chunks", "Received partial encoded chunk request for {:?}, part_ordinals: {:?}, receipts: {:?}, I'm {:?}", request.chunk_hash.0, request.part_ords, request.tracking_shards, self.me);
 
-        // Check if we have the chunk in our cache
-        if let Some(entry) = self.encoded_chunks.get(&request.chunk_hash) {
-            // Create iterators which _might_ contain the requested parts.
-            let parts_iter = request.part_ords.iter().map(|ord| entry.parts.get(ord).cloned());
-            let receipts_iter = request
-                .tracking_shards
-                .iter()
-                .map(|shard_id| entry.receipts.get(shard_id).cloned());
-
-            // Pass iterators to function which will evaluate them. Since iterators are lazy
-            // we will clone as few elements as possible before realizing not all are present.
-            // In the case all are present, the response is sent.
-            return self.maybe_send_partial_encoded_chunk_response(
-                request.chunk_hash,
-                route_back,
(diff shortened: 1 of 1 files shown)


chain/chunks/src/lib.rs
```diff
@@ -998,78 +998,106 @@ impl ShardsManager {
...
-                .map(|shard_id| present_receipts.get(shard_id).map(|x| *x).cloned());
-
-            // Pass iterators to function, same as cache case.
-            return self.maybe_send_partial_encoded_chunk_response(
-                request.chunk_hash,
-                route_back,
-                parts_iter,
-                receipts_iter,
-            );
+            Self::prepare_partial_encoded_chunk_response_from_partial(request, partial_chunk)
```
mm-near, line 1004: nit: I'd call it "_from_storage" rather than "from_partial"

chain/chunks/src/lib.rs
```diff
@@ -998,78 +998,106 @@ impl ShardsManager {
...
+        } else {
+            None
         };
+
+        if let Some(response) = response {
+            self.peer_manager_adapter.do_send(PeerManagerMessageRequest::NetworkRequests(
+                NetworkRequests::PartialEncodedChunkResponse { route_back, response },
+            ))
+        }
```
mm-near, line 1013: in separate PR:

maybe we should have some stats - on how many times we don't have the response (as normally this should not happen).

chain/chunks/src/lib.rs
```diff
@@ -998,78 +998,106 @@ impl ShardsManager {
...
-    ) where
+    ) -> Option<PartialEncodedChunkResponseMsg>
+    where
         A: Iterator<Item = Option<PartialEncodedChunkPart>>,
         B: Iterator<Item = Option<ReceiptProof>>,
     {
         let maybe_known_parts: Option<Vec<_>> = parts_iter.collect();
         let parts = match maybe_known_parts {
             None => {
                 debug!(target:"chunks", "Not responding, some parts are missing");
```
mm-near, line 1099: maybe we should also add 'chunk_hash' to those debug logs

mm-near (approved):

chain/chunks/src/lib.rs
mina86: Problem with that name is that I’m working on `_from_chunk` method which also reads data from storage.

PR #6358, opened 2022-03-01
fix(params-estimator): Fix imports on non-linux OS

jakmeier:


jakmeier:
I broke Mac builds on master yesterday... This time, I tested compiling on a MB air.

Ekleog (approved):
The diff definitely makes sense to me, though I haven't tested compiling it as you said you already did it.

Unfortunately I guess this kind of issue couldn't even be detected on an aarch64 linux VM as the important factor is linux vs darwin. So until a cloud platform gets access to darwin vms I guess there's nothing much we can do to improve the process here…

olonho (approved):

PR #6360, opened 2022-03-01
doc(params-estimator): Log related costs docs

jakmeier:


jakmeier:
My changes so far describe the status-quo. But there are some things I don't understand why they are as they are. For that, I created #6361.

Once we understand this a bit more, we could merge this PR to move forward the goal of documenting all estimations. And keep the linked issue to track necessary logical changes.

matklad (approved):
LGTM, though I think we've since divided some constants by 10

PR #6362, opened 2022-03-01
feat(params-estimator): Cleanup fn call estimation

jakmeier:
Various changes around function call estimation.

1. Estimation for function call cost related to loading contract code from storage.
2. Change fn call estimations to use CLI arguments for iteration counts and reduce inner iterations to something that runs faster.
3. Add some documentation to estimations and rename a few things to make the code clearer.

### DIFF
--- a/runtime/runtime-params-estimator/src/cost.rs
+++ b/runtime/runtime-params-estimator/src/cost.rs
@@ -79,10 +79,27 @@ pub enum Cost {
     /// a transaction. Subtract base costs and apply least-squares on the
     /// results to find the per-byte costs.
     ActionDeployContractPerByte,
+    /// Estimates `action_creation_config.function_call_cost`, which is the base
+    /// cost for adding a `FunctionCallAction` to a receipt. It aims to account
+    /// for all costs of calling a function that are already known on the caller
+    /// side.
+    /// 
+    /// Estimation: Measure the cost to execute a transaction with an empty
+    /// function with no arguments. Subtract the receipt creating cost from
+    /// that, as that is already charged separately.
+    /// 
+    /// TODO[jakmeier][#6353]: Today, the estimation falsely includes the
+    /// loading cost of the estimator test contract.
     ActionFunctionCallBase,
+    /// Estimates `action_creation_config.function_call_cost_per_byte`, which is
+    /// the incremental cost for each byte of the method name and method
+    /// arguments cost for adding a `FunctionCallAction` to a receipt.
+    ///
+    /// Estimation: Measure the cost for a transaction with an empty function
+    /// call with a large argument value. Subtract the cost of an empty function
+    /// call with no argument. Divide the difference by the length of the
+    /// argument.
     ActionFunctionCallPerByte,
-    ActionFunctionCallBaseV2,
-    ActionFunctionCallPerByteV2,
     /// Estimates `action_creation_config.transfer_cost` which is charged for
     /// every `Action::Transfer`, the same value for sending and executing.
     ///
@@ -556,6 +573,26 @@ pub enum Cost {
     /// around bytes that are not code. Divide this cost by the difference of
     /// bytes.
     DeployBytes,
+    /// Estimates `contract_compile_base` which is charged once per contract
+    /// that is loaded from the database to execute a method on it.
+    /// (will be renamed to `contract_loading_base`, see
+    /// https://github.com/near/nearcore/issues/5962)
+    ///
+    /// Estimation: Measure the cost to execute an empty contract method
+    /// directly on a runtime instance, using different sizes of contracts.
+    /// Use least-squares to calculate base and per-byte cost.
+    /// The contract size is scaled by adding more methods to it. This has been
+    /// identified as a particular expensive in terms of per-byte loading time.
+    /// This makes it a better scaling strategy than, for example, adding large
+    /// constants in the data section.
+    ContractLoadingBase,
+    /// Estimates `contract_compile_byte` which is charged for each byte in a
+    /// contract when it is loaded from the database to execute a method on it.
+    /// (will be renamed to `contract_loading_base`, see
+    /// https://github.com/near/nearcore/issues/5962)
+    ///
+    /// Estimation: See `ContractLoadingBase`.
+    ContractLoadingPerByte,
     GasMeteringBase,
     GasMeteringOp,
     /// Cost of inserting a new value directly into a RocksDB instance.
--- a/runtime/runtime-params-estimator/src/cost.rs
+++ b/runtime/runtime-params-estimator/src/cost.rs
@@ -584,7 +584,7 @@ pub enum Cost {
     /// Estimates the executable loading part of `wasm_contract_loading_bytes`
     /// which is charged for each byte in a contract when it is loaded as an
     /// executable.
-    /// 
+    ///
     /// This cost also has to cover the reading of the code from the database.
     /// So technically, it covers code loading from database and also executable
     /// loading. But it is still charged right before executable loading because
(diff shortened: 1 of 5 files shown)


jakmeier:
Estimations will only work after merging #6958.

jakmeier:
Storage estimations are still not perfect, so take these numbers with a grain of salt. But here are the time-based estimation numbers, without safety multiplier:

|  | gcloud time | qemu icount |
| --- | --- | --- |
| ContractLoadingBase                             |    32_484.778_944 Mgas | 6_972.125_000 Mgas |
| ContractLoadingPerByte                          |        40.436_971 Mgas  | 13.625_000 Mgas |
| FunctionCallPerStorageByte                     |          1.011_164 Mgas | 0.278_971 Mgas |


We don't have enough information yet to draw conclusion about how to change the parameter values. This should be tackled in the next quarter.

The main take-away here is that the storage related cost in today's estimation is much smaller than the executable loading cost. Therefore, the current cost model should be good enough, for now.

(If, on the contrary, the storage costs would have been dominant, than I would have suggested to add a parameter for just the storage cost, which can be charged even earlier. But this would require major refactoring to allow gas charging that is neither a receipt cost nor a cost charged inside a VM implementation. But since we can just hide the cost in the executable loading cost, I think we can avoid doing that.)

runtime/runtime-params-estimator/src/cost.rs
```diff
@@ -79,10 +79,24 @@ pub enum Cost {
     /// a transaction. Subtract base costs and apply least-squares on the
     /// results to find the per-byte costs.
     ActionDeployContractPerByte,
+    /// Estimates `action_creation_config.function_call_cost`, which is the base
+    /// cost for adding a `FunctionCallAction` to a receipt. It aims to account
+    /// for all costs of calling a function that are already known on the caller
+    /// side.
```
matklad, line 85: Let's rephrase this: I *think* this is not so much about knowing, as about where the actual work is being done. Even if we know the size of the contract when making a call, we'd still like to charge for that on exec (which might be different shard, etc)

runtime/runtime-params-estimator/src/cost.rs
```diff
@@ -79,10 +79,24 @@ pub enum Cost {
...
+    ///
+    /// Estimation: Measure the cost to execute a transaction with an empty
+    /// function with no arguments. Subtract the receipt creating cost from
+    /// that, as that is already charged separately.
```
matklad, line 89: So this *would* include a cost of actually executing the function, right? I wonder if we can abuse our unncessary early bail-out for empty method names for estimation?

runtime/runtime-params-estimator/src/function_call.rs
```diff
@@ -1,35 +1,38 @@
...
+/// Estimates linear cost curve for a function call execution cost per byte of
+/// total contract code. The contract size is increased by adding more methods
+/// to it. This cost is pure VM cost, without the loading from storage.
+pub(crate) fn function_call_cost_per_code_byte(config: &Config) -> (GasCost, GasCost) {
     let mut xs = vec![];
     let mut ys = vec![];
-    const REPEATS: u64 = 50;
+    let repeats = config.iter_per_block as u64;
+    let warmup_repeats = config.warmup_iters_per_block as u64;
     for method_count in vec![5, 20, 30, 50, 100, 200, 1000] {
```
matklad, line 21:
```suggestion
    for method_count in [5, 20, 30, 50, 100, 200, 1000] {
```

still cant' believe this now just works :D

runtime/runtime-params-estimator/src/lib.rs
```diff
@@ -668,46 +668,65 @@ fn action_function_call_base(ctx: &mut EstimatorContext) -> GasCost {
     total_cost.saturating_sub(&base_cost, &NonNegativeTolerance::PER_MILLE)
 }
 fn action_function_call_per_byte(ctx: &mut EstimatorContext) -> GasCost {
```
matklad, line 670: So here we vary only the size of the argument, not the size of a method name. Could you sanity-check that long method names are not problematic (not sure if theer's some small size limit there).

runtime/runtime-params-estimator/src/lib.rs
```diff
@@ -668,46 +668,65 @@ fn action_function_call_base(ctx: &mut EstimatorContext) -> GasCost {
...
-    let total_cost = {
-        let mut make_transaction = |tb: &mut TransactionBuilder| -> SignedTransaction {
-            let sender = tb.random_unused_account();
-            tb.transaction_from_function_call(sender, "noop", vec![0; 1024 * 1024])
-        };
-        transaction_cost(ctx, &mut make_transaction)
-    };
-
-    let base_cost = noop_function_call_cost(ctx);
+    let xs = [1, 1_000_000, 4_000_000];
```
matklad, line 671: Hm, naively, just 3 points seems a bit too little for least squares. Do we want more? Or do we want just two points and simple division? 

Actually, that's intereresting -- a lot of per-byte costs are computed as `(10k - base) / 10k`, but that's just a degenerate case of least squares!

matklad (approved):
Left a bunch of comments, but good to go as is. 

One significant (but pre-existing) issue I think is that our function_call_base effectively includes contract_loading_base. Worth looking into!

runtime/runtime-params-estimator/src/cost.rs
jakmeier: `FunctionCallAction` does have a send and an execution part though. I would argue all at this point known cost should be part of the receipt `execution`. Using VM costs has the downside that it relies on user-attached gas which is less robust against parameter value changes.

runtime/runtime-params-estimator/src/cost.rs
jakmeier: Yes, it includes the execution of the empty function. I guess now that we are properly estimating the loading cost separately, this is an option. Let me see if it makes a difference.

runtime/runtime-params-estimator/src/lib.rs
jakmeier: Agreed that it is a bit odd, so I added a comment to explain the motivation.

The problem is that adding more X values below 1M increases overall variance because often f(x=500k) will be smaller than f(x=100k) in time-based estimation. So I picked three values that are at least sorted.

Of course, making the estimation more stable would be ideal. But since it is so insignificantly small, I couldn't really justify spending the time on this specific estimate.

runtime/runtime-params-estimator/src/lib.rs
```diff
@@ -668,46 +668,65 @@ fn action_function_call_base(ctx: &mut EstimatorContext) -> GasCost {
...
+    let ys: Vec<GasCost> = xs
+        .iter()
+        .map(|&arg_len| inner_action_function_call_per_byte(ctx, arg_len as usize))
+        .collect();
 
-    let bytes_per_transaction = 1024 * 1024;
+    let negative_base_tolerance = 0u64;
+    let rel_factor_tolerance = 0.001;
```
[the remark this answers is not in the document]
jakmeier: Done.

runtime/runtime-params-estimator/src/function_call.rs
```diff
@@ -1,35 +1,38 @@
...
 use near_vm_runner::internal::VMKind;
-use num_rational::Ratio;
 use std::fmt::Write;
 use std::sync::Arc;
 
-pub(crate) fn test_function_call(metric: GasMetric, vm_kind: VMKind) -> (Ratio<i128>, Ratio<i128>) {
+/// Estimates linear cost curve for a function call execution cost per byte of
+/// total contract code. The contract size is increased by adding more methods
+/// to it. This cost is pure VM cost, without the loading from storage.
+pub(crate) fn function_call_cost_per_code_byte(config: &Config) -> (GasCost, GasCost) {
```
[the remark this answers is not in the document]
jakmeier: done

runtime/runtime-params-estimator/src/lib.rs
jakmeier: There is this parameter: `max_length_method_name: 256`

Quick sanity checks suggest that there is no measurable difference between different method lengths in the range 1-255. (Mostly because the variance is just too high.)

runtime/runtime-params-estimator/src/cost.rs
jakmeier: It does seem to make a sizeable difference from my quick hacky tests I did. But I think implementing that is asks for a separate PR, it's more than just a trivial change in the existing function.

PR #3036, opened 2020-07-24
(fix): Making stress.py with node_restart mode pass, and fixing #2916

SkidanovAlex:
After this change stress.py node_restart passes relatively consistently, and is reintroduced to nightly.

Nearcore fixes:

- We had a bug in the syncing logic (with a low chance of being
triggered in the wild): if a block is produced, and between 1/3 and 2/3
of block producers received it, and the rest have not, the system
stalls, because no 2/3 of block producers have the same head, but also
nobody is two blocks behind the highest peer to start syncing. Fixing it
by forcing sync if we've been 1 block behind for too long.
stress.py was reproducing this issue in every run

- (#2916) we had an issue that if a node produced a chunk, and then
crashed, on recovery it was not able to serve it because it didn't have
all the parts and receipts stored in the storage from which we recover
cache entries in the shards manager. Fixing it by always storing all the
parts and receipts (redundantly) for chunks in the shards we care about.

Test fixes

[v] Fixing a scenario in which a failure to send a transaction to all
validators resulted in recording an incorrect tx hash alongside the tx.
Later when checking balances using the incorrect hash resulted in
getting incorrect success value, and thus applying incorrect corrections
to the expected balances;

[v] Changing the order of magnitude of staking transactions, so that the
validator set actually changes.

Other issues discovered while fixing stress.py:
- https://github.com/nearprotocol/nearcore/issues/2906

PR #6366, opened 2022-03-01
refactor: simplify if statements in persist_partial_chunk_for_data_av…

marcelo-gonzalez:
…ailability()

there are multiple branches with the same condition as others that can be removed


marcelo-gonzalez:
btw, I think what happened is that the review comments in the PR that added these were slightly misinterpreted: https://github.com/near/nearcore/pull/3036#discussion_r460326792

bowenwang1996 (approved):
Thanks!

PR #6367, opened 2022-03-01
chore: impl `Error` on near-primitives errors that missed them

ChaoticTempest:
Saw that we were missing a couple impls for `std::error::Error` for some of the errors in near-primitives. Would be useful if we could convert them into `&dyn Error` or `anyhow` errors to be used in libraries such as `workspaces`

### DIFF
--- a/core/primitives/src/errors.rs
+++ b/core/primitives/src/errors.rs
@@ -21,6 +21,8 @@ pub enum TxExecutionError {
     InvalidTxError(InvalidTxError),
 }
 
+impl std::error::Error for TxExecutionError {}
+
 impl Display for TxExecutionError {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> Result<(), std::fmt::Error> {
         match self {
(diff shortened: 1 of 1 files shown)


core/primitives/src/errors.rs
```diff
@@ -21,6 +21,8 @@ pub enum TxExecutionError {
     InvalidTxError(InvalidTxError),
 }
 
+impl std::error::Error for TxExecutionError {}
```
miraclx, line 24: To be clear, `TxExecutionError` isn't directly an `Rpc*Error` but is instead embedded deep inside `FinalExecutionOutcomeView` that's the `Rpc*Response` of both the `broadcast_tx_commit` and `tx` RPC methods.

miraclx (approved):
Nice catch. LgTm!

ChaoticTempest:
> > Would be useful if we could convert them into &dyn Error or anyhow errors to be used in libraries such as workspaces
> 
> I'd recommend against converting near-primitives errors directly to `anyhow`: this makes `near-primitives` a public dependency, as the underlying error can be extracted via `downcasting`. This not suuper important though.

@matklad gotcha, this conversation can probably be moved over to workspaces, but it's just that some error handling needs specific error data on the user side if they want to handle it. The thing was that these error types coming from `near-primitives` are pretty sprawling and large in number, so maintaining a 1 to 1 might prove to be difficult if we were to just provide our own set of error types in workspaces. I can see mismatch errors happening when we use a newer version of sandbox that provides newer types. I'm already seeing similar issues with `near-jsonrpc-client` where it fails to deserialize some newer error types. What would you recommend in this case?

core/primitives/src/errors.rs
matklad: Hm, taking a second look here, this impl in particular feels suspect to me. `TxExecutionError` should **not** be an `std::error::Error`, it should **not** work with the `?` syntax. 

It's not a runtime error a-la io error. It is a value which signifies that the tx was succesfully processed, and that, as a result of this processing, the runtime concluded that the tx is invalid.

PR #6371, opened 2022-03-02
fuzz: avoid runtime-tester false positives

Ekleog:
System time is a high-variability metric. CPU time consumed by the
process should be much less volatile, and so hopefully will lead to both
fewer false positives, but also to a lesser need for safety margins,
thus allowing the lowering of 3 seconds to 2 seconds when fuzzing.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4338,6 +4338,7 @@ name = "runtime-tester"
 version = "0.0.0"
 dependencies = [
  "byteorder",
+ "cpu-time",
  "libfuzzer-sys",
  "near-chain",
  "near-chain-configs",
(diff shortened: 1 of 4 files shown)


Ekleog:
To add some context here, the fuzzer for runtime-tester frequently generates false positives (around once a day). This is probably due to fighting for resources with other processes (my guess would be it actually uses more than one thread, though I haven't checked that yet).
In addition to this, we need to SIGSTOP/SIGCONT the fuzzer process when nayduck has a job coming in. If we keep using Instant, this'll break every time the fuzzer gets sigstop'd.

In order to work around both issues, using the CPU time instead of wall clock time should be enough.

The only drawback is that it does not account for time spent in syscalls, but I'm thinking lowering the limit at the same time as we make this switch should be enough to work around this limitation.

PR #6376, opened 2022-03-02
pytest: check reading chunks from storage in block_sync_archival.py

mina86:
The ShardManager keeps cache of encoded chunks going back 1024 heights
which means that nodes which request partial chunks for a recent block
will be satisfied by data in the cache.  This means that since only
a hundred blocks are generated in block_sync_archival.py test, the
code path where data is read from storage is never executed.

Extend the test such that it generates 1500 blocks to make sure that
both code paths are executed.


pytest/tests/sanity/block_sync_archival.py
```diff
@@ -5,8 +5,15 @@
 Starts the validator and waits until several epochs worth of blocks are
 generated.  Then, starts the observer node and makes sure that it properly
 synchronises the full history.
+
+When called with --long-run the test will generate enough blocks so that entries
+in EncodedChunksCache start being evicted.  That it, it’ll generated more than
```
mm-near, line 10: s/generated/generate

pytest/tests/sanity/block_sync_archival.py
```diff
@@ -71,22 +93,44 @@ def get_all_blocks(node: cluster.BaseNode, *,
...
-    fred_blocks = get_all_blocks(fred, head=latest)
-    if boot_blocks != fred_blocks:
-        for a, b in zip(boot_blocks, fred_blocks):
-            if a != b:
-                logger.error(f'{a} != {b}')
-        assert False
+def main(argv: typing.Sequence[str]) -> None:
+    target_height = SHORT_TARGET_HEIGHT
+    min_delay = None
+    for arg in argv[1:]:
```
mm-near, line 99: should we use 'argparse' instead?

mm-near (approved):

PR #6377, opened 2022-03-02
chain: ShardManager: add ability to serve partial chunks from ShardChunk

mina86:
Add ability to respond to PartialEncodedChunkRequest from ShardChunk
objects in addition to PartialEncodedChunk.  In practice this is currently
dead code since there is no scenario in which the former is in the storage
while the latter isn’t but the plan is to start garbage collecting
ColPartialChunks column at which point we’ll have to serve requests from
data in ColChunks column.


chain/chunks/src/lib.rs
```diff
@@ -1075,6 +1089,136 @@ impl ShardsManager {
...
 
+    fn prepare_partial_encoded_chunk_response_from_chunk(
+        &mut self,
+        request: PartialEncodedChunkRequestMsg,
+        chain_store: &mut ChainStore,
+        rs: &mut ReedSolomonWrapper,
+        chunk: ShardChunk,
+    ) -> Option<PartialEncodedChunkResponseMsg> {
+        // TODO(mina86): Is it correct to use rs.data_shard_count() and
+        // rs.total_shard_count() here?  Can those values change over time?
```
mm-near, line 1100: Currently this is the number from 'num_blocks_producer_seats' in genesis.

We might want to change it in the future -- but that would be a larger change anyway (as we'd have to start storing this number somewhere in the header/epoch info)

chain/chunks/src/lib.rs
```diff
@@ -1075,6 +1089,136 @@ impl ShardsManager {
         )
     }
 
+    fn prepare_partial_encoded_chunk_response_from_chunk(
+        &mut self,
+        request: PartialEncodedChunkRequestMsg,
+        chain_store: &mut ChainStore,
+        rs: &mut ReedSolomonWrapper,
+        chunk: ShardChunk,
+    ) -> Option<PartialEncodedChunkResponseMsg> {
```
mm-near, line 1098: we should have a test for this - and check that the 'final' hashes are matching

chain/chunks/src/lib.rs
```diff
@@ -995,13 +995,27 @@ impl ShardsManager {
...
+        rs: &mut ReedSolomonWrapper,
     ) {
         debug!(target: "chunks", "Received partial encoded chunk request for {:?}, part_ordinals: {:?}, shards: {:?}, I'm {:?}", request.chunk_hash.0, request.part_ords, request.tracking_shards, self.me);
 
         let response = if let Some(entry) = self.encoded_chunks.get(&request.chunk_hash) {
             Self::prepare_partial_encoded_chunk_response_from_cache(request, entry)
         } else if let Ok(partial_chunk) = chain_store.get_partial_chunk(&request.chunk_hash) {
             Self::prepare_partial_encoded_chunk_response_from_partial(request, partial_chunk)
+        } else if let Ok(chunk) = chain_store.get_chunk(&request.chunk_hash).map(|ch| ch.clone()) {
+            // Note: we need to clone the chunk because otherwise we would be
```
mm-near, line 1007: we should have metrics for all 3 cases here (can be done in separate PR)

mm-near (changes requested):

chain/chunks/src/lib.rs
```diff
@@ -1075,6 +1089,136 @@ impl ShardsManager {
         )
     }
 
+    fn prepare_partial_encoded_chunk_response_from_chunk(
```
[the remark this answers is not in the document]
mina86: It’s not immediately obvious to be me what should be moved to separate file though to be honest. Code that handles partial encoded chunk requests is less than 200 lines so moving just it wouldn’t really help reduce the size of lib.rs. This would need some more care and thought to figure out which of the methods in the struct belong together in a separate file.

chain/chunks/src/lib.rs
mina86: Do you have an idea for a better comment? Or should I just drop it altogether?

mzhangmzz: I'd suggest using `self.runtime_adapter.num_total_parts()`, since that's how we access total number of parts throughout the code and we will reimplement this function if we decide to change `num_blocks_producer_seats` in the future.

chain/chunks/src/lib.rs
```diff
@@ -1075,6 +1092,143 @@ impl ShardsManager {
...
+            .runtime_adapter
+            .get_shard_layout_from_prev_block(prev_block_hash)
+            .map_err(|err| {
+                debug!(target: "chunks", "Not sending {:?}, failed to get shard layout: {}", request.chunk_hash.0, err);
+            })
+            .ok()?;
+
+        // Get outgoing receipts for the chunk and construct vector of their
+        // proofs.
+        let outgoing_receipts = chain_store
```
mzhangmzz, line 1137: Here you can just use `chunk.receipts()` instead of reading from store, since we already have the chunk content

chain/chunks/src/lib.rs
```diff
@@ -1075,6 +1092,143 @@ impl ShardsManager {
...
+        let (outgoing_receipts_root, outgoing_receipts_proofs) =
+            merklize(&outgoing_receipts_hashes);
+        if header.outgoing_receipts_root() != outgoing_receipts_root {
+            error!(target: "chunks",
+                   "Not sending {:?}, expected outgoing receipts root doesn’t match calculated: {} != {}",
+                   request.chunk_hash.0, header.outgoing_receipts_root(), outgoing_receipts_root);
+            return None;
+        }
+
+        let mut receipts_by_shard =
```
mzhangmzz, line 1158: This logic of constructing receipts_proofs has already appeared in both `distribute_encoded_chunk` and `create_and_persist_partial_chunk`, probably better to create a new method in `Chain` to encapsulate this logic to avoid code duplication

mzhangmzz (commented):
I agree with Marcin that we want to have some measurement of how long it takes to process a partial encode chunk part request by reconstructing the chunk parts, since it might not be cheap. Another question we need to answer after we have the measurement is whether we want to cache the results for some time, since the node requesting for these parts may resend the request again.

chain/chunks/src/lib.rs
mina86: In the end `rs` is created from `self.runtime_adapter.num_total_parts()`:

```rust
impl Client {
    pub fn new(
        /* ... */
    ) -> Result<Self, Error> {
        /* ... */
        let data_parts = runtime_adapter.num_data_parts();
        let parity_parts = runtime_adapter.num_total_parts() - data_parts;
        /* ... */
        Ok(Self {
            /* ... */
            rs: ReedSolomonWrapper::new(data_parts, parity_parts),
            /* ... */
        })
    }
}
```

so sounds like using `rs` is correct here.  I’d rather not call
`self.runtime.num_total_parts()` and `num_data_parts()` here because
that would require creating a new `ReedSolomonWrapper` which, from
what I understand, is non-trivial operation.

My question and worry is whether number of parts can change over
time.  E.g. if someone requests a partial block from one year ago can
its encoding be different than encoding of a chunk created at head.

chain/chunks/src/lib.rs
```diff
@@ -1075,6 +1092,139 @@ impl ShardsManager {
         )
     }
 
+    /// Constructs a vector of [`ReceiptsProof`]s for specified chunk.
+    fn make_outgoing_receipts_proofs<T>(
+        &self,
+        chunk_header: &ShardChunkHeader,
+        outgoing_receipts: &Vec<Receipt>,
+        make_elem: impl Fn(ReceiptProof) -> T,
+    ) -> Result<Vec<T>, near_chunks_primitives::Error> {
```
matklad, line 1101: My personal style is to write such things as

```suggestion
        sink: &mut dyn FnMut(ReceiptProof),
    ) -> Result<(), near_chunks_primitives::Error> {
```

That is, without generics. The primary motivation is compile time -- with generics, this closure will be "copy pasted" several times, and it feels like it can add up other time.

I can not definitely say that using vs not-using such pattern would lead to some definitive X difference in compile time.

However, in rust-analyzer we did get observable and measurable compile-time improments out of this style of changes.

For example, https://github.com/rust-analyzer/rust-analyzer/commit/76b60efbfb8cdc94cf14fef1be50c60f250bc293 shows about 10% percent improvement in terms of amount of LLVM IR generated, when we replaced one generic call-back with a dynamic call-back (in the context of the codebase which uses generics rather sparingly)

chain/chunks/src/lib.rs
```diff
@@ -1075,6 +1092,139 @@ impl ShardsManager {
...
+            .get_shard_layout_from_prev_block(chunk_header.prev_block_hash_ref())?;
+
+        let hashes = Chain::build_receipts_hashes(&outgoing_receipts, &shard_layout);
+        let (root, proofs) = merklize(&hashes);
+        if chunk_header.outgoing_receipts_root() != root {
+            return Err(near_chunks_primitives::Error::Other(format!(
+                "unexpected outgoing receipts root; want: {}, but got: {}",
+                chunk_header.outgoing_receipts_root(),
+                root
+            )));
```
matklad, line 1114: Yeah, I'd expect something like this here:

```suggestion
            return Err(near_chunks_primitives::Error::UnexpectedOutgoingRemeiptsRoot {
                want: chunk_header.outgoing_receipts_root(),
                got: root
            });
```

chain/chunks/Cargo.toml
```diff
@@ -33,3 +34,4 @@ assert_matches = "1.5.0"
 [features]
 byzantine_asserts = ["near-chain/byzantine_asserts"]
 expensive_tests = []
+test_features = []
```
mm-near, line 37: comment. Also maybe we should call it 'reconstruct_partial_chunk_on_the_fly' or something like that?

chain/chunks/src/lib.rs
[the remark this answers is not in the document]
matklad: Yeah, I don't think *this* makes sense to move to a separate file, as it seems to be pretty entangled with the rest.

The easiest wins in terms of keeping files small and logic clean are usually around various small helpers. Things like `RequestPool` or `SealsManager` seem like better candidates for extraction.

chain/chunks/src/lib.rs
```diff
@@ -1089,78 +1092,81 @@ impl ShardsManager {
         )
     }
 
+    /// Constructs a vector of [`ReceiptsProof`]s for specified chunk.
+    fn make_outgoing_receipts_proofs<T>(
+        &self,
+        chunk_header: &ShardChunkHeader,
+        outgoing_receipts: &Vec<Receipt>,
+        make_elem: impl Fn(ReceiptProof) -> T,
```
mm-near, line 1100: I'd expect that we'd return vector of ReceiptProof, but here I see that we have this 'make_elem' instead. Can you explain in the comment why? is it due to performance reasons?

chain/chunks/src/lib.rs
```diff
@@ -1075,6 +1092,139 @@ impl ShardsManager {
...
+        &mut self,
+        request: PartialEncodedChunkRequestMsg,
+        rs: &mut ReedSolomonWrapper,
+        chunk: ShardChunk,
+    ) -> Option<PartialEncodedChunkResponseMsg> {
+        let total_parts = rs.total_shard_count();
+        for ord in request.part_ords.iter() {
+            let ord: usize = (*ord).try_into().unwrap();
+            if ord >= total_parts {
+                warn!(target:"chunks", "Not sending {}, requested part_ord {} but we only expect {} total",
```
matklad, line 1153: This probably worth an `error`, right? Ie, this case signals a bug. (low level on confidence on this one)

mm-near (approved):

core/primitives/src/sharding.rs
```diff
@@ -303,6 +303,17 @@ impl ShardChunkHeader {
         }
     }
 
+    // TODO(mina86): Change return type of prev_block_hash to &CryptoHash and
```
matklad, line 306: :+1:

integration-tests/Cargo.toml
```diff
@@ -66,7 +66,7 @@ performance_stats = [
   "near-network/performance_stats",
 ]
 expensive_tests = []
-test_features = ["nearcore/test_features"]
+test_features = ["nearcore/test_features", "near-chunks/test_features"]
```
matklad, line 69: I'd rather

```toml
diff --git a/chain/client/Cargo.toml b/chain/client/Cargo.toml
index 68bb99de9..3a854466d 100644
--- a/chain/client/Cargo.toml
+++ b/chain/client/Cargo.toml
@@ -54,6 +54,7 @@ expensive_tests = []
 test_features = [
   "near-network/test_features",
   "near-chain/test_features",
+  "near-chunks/test_featurs",
 ]
 delay_detector = [
   "near-chain/delay_detector",
diff --git a/integration-tests/Cargo.toml b/integration-tests/Cargo.toml
index fce626a17..c69cb0ba8 100644
--- a/integration-tests/Cargo.toml
+++ b/integration-tests/Cargo.toml
@@ -66,7 +66,7 @@ performance_stats = [
   "near-network/performance_stats",
 ]
 expensive_tests = []
-test_features = ["nearcore/test_features", "near-chunks/test_features"]
+test_features = ["nearcore/test_features"]
 protocol_feature_alt_bn128 = [
     "near-primitives/protocol_feature_alt_bn128",
     "node-runtime/protocol_feature_alt_bn128",
```

we should try to make it impossible to compile only *some* crates with `test_features`.

chain/chunks-primitives/src/error.rs
```diff
@@ -14,6 +14,7 @@ pub enum Error {
     KnownPart,
     ChainError(near_chain_primitives::Error),
     IOError(std::io::Error),
+    Other(String),
```
matklad, line 17: Adding `Other(String)` feels like a particularly slippery slope.

chain/chunks/src/lib.rs
mina86: I honestly dunno. It’s a situation which does not prevent the node from operating (and for all we know the problem is on the other side) so that’s why I opted to using warning.

chain/chunks/Cargo.toml
mina86: I’m planning to make GCing a configurable feature.

chain/chunks/src/lib.rs
```diff
@@ -1075,6 +1092,142 @@ impl ShardsManager {
...
+    ) -> Result<Vec<T>, near_chunks_primitives::Error> {
+        let shard_id = chunk_header.shard_id();
+        let shard_layout = self
+            .runtime_adapter
+            .get_shard_layout_from_prev_block(chunk_header.prev_block_hash_ref())?;
+
+        let hashes = Chain::build_receipts_hashes(&outgoing_receipts, &shard_layout);
+        let (root, proofs) = merklize(&hashes);
+        if chunk_header.outgoing_receipts_root() != root {
+            return Err(near_chunks_primitives::Error::UnexpectedOutgoingRemeiptsRoot {
```
mzhangmzz, line 1114: I think we should assert here like the previously instead of returning an error. This means we have accepted an invalid chunk in the storage and there is something seriously wrong.

chain/chunks/src/lib.rs
mzhangmzz: Agree that warning is fine here because the request is sent from another node whose behavior we cannot control

chain/chunks/src/lib.rs
mzhangmzz: Right now it doesn't since we never change the number of validator seats in our chain, but that is a very valid concern because that might and I think will change in the future. And that is why I suggested using `runtime_adapter.num_total_parts` here because if in the future we do change that, we will definitely need to change `num_total_parts`, so it's easy to track what needs to be changed in the code.

How about we use `num_total_parts` here but add an assert that it's the same value as `rs.total_shard_count`? For now, it's the same anyways. And this is more rust to future changes.

chain/chunks/src/lib.rs
matklad: It's definitely not a clear win, it's a trade off between "write more code" and "wait more for the code to compile", and its indeed unclear how much of the second one we gain/loose here, so no strong opinion. 

But, having taken a second look here, it seems that this can just return an iterator? That seems like a clear win -- no geneircs, no closures, more flexibility to the caller:

```diff
diff --git a/chain/chunks/src/lib.rs b/chain/chunks/src/lib.rs
index cb55bac6a..247684ad9 100644
--- a/chain/chunks/src/lib.rs
+++ b/chain/chunks/src/lib.rs
@@ -1097,12 +1097,11 @@ impl ShardsManager {
     /// The `make_elem` argument converts `ReceiptsProof` values into objects
     /// which are returned from this function.  In practice this is used to
     /// either return `ReceiptProof` directly or wrap the object into `Arc`.
-    fn make_outgoing_receipts_proofs<T>(
+    fn make_outgoing_receipts_proofs(
         &self,
         chunk_header: &ShardChunkHeader,
         outgoing_receipts: &[Receipt],
-        make_elem: impl Fn(ReceiptProof) -> T,
-    ) -> Result<Vec<T>, near_chunks_primitives::Error> {
+    ) -> Result<impl Iterator<Item = ReceiptProof>, near_chunks_primitives::Error> {
         let shard_id = chunk_header.shard_id();
         let shard_layout = self
             .runtime_adapter
@@ -1119,20 +1118,13 @@ impl ShardsManager {
 
         let mut receipts_by_shard =
             Chain::group_receipts_by_shard(outgoing_receipts.to_vec(), &shard_layout);
-        let proofs = proofs
-            .into_iter()
-            .enumerate()
-            .map(|(proof_shard_id, proof)| {
-                let proof_shard_id = proof_shard_id as u64;
-                let receipts = receipts_by_shard.remove(&proof_shard_id).unwrap_or_else(Vec::new);
-                let shard_proof = ShardProof {
-                    from_shard_id: shard_id,
-                    to_shard_id: proof_shard_id,
-                    proof: proof,
-                };
-                make_elem(ReceiptProof(receipts, shard_proof))
-            })
-            .collect();
+        let proofs = proofs.into_iter().enumerate().map(move |(proof_shard_id, proof)| {
+            let proof_shard_id = proof_shard_id as u64;
+            let receipts = receipts_by_shard.remove(&proof_shard_id).unwrap_or_else(Vec::new);
+            let shard_proof =
+                ShardProof { from_shard_id: shard_id, to_shard_id: proof_shard_id, proof: proof };
+            ReceiptProof(receipts, shard_proof)
+        });
         Ok(proofs)
     }
 
@@ -1165,11 +1157,12 @@ impl ShardsManager {
         // proofs.
         let outgoing_receipts = chunk.receipts();
         let receipt_proofs = self
-            .make_outgoing_receipts_proofs(&header, &outgoing_receipts, |p| p)
+            .make_outgoing_receipts_proofs(&header, &outgoing_receipts)
             .map_err(|err| {
                 warn!(target: "chunks", "Not sending {}, {}", request.chunk_hash.0, err);
             })
-            .ok()?;
+            .ok()?
+            .collect::<Vec<_>>();
 
         // Construct EncodedShardChunk.  If we earlier determined that we will
         // need parity parts, instruct the constructor to calculate them as
@@ -2064,7 +2057,8 @@ impl ShardsManager {
         store_update: &mut ChainStoreUpdate<'_>,
     ) -> Result<(), Error> {
         let header = encoded_chunk.cloned_header();
-        let receipts = self.make_outgoing_receipts_proofs(&header, &outgoing_receipts, |p| p)?;
+        let receipts =
+            self.make_outgoing_receipts_proofs(&header, &outgoing_receipts)?.collect::<Vec<_>>();
         let partial_chunk = PartialEncodedChunkV2 {
             header,
             parts: encoded_chunk
@@ -2115,8 +2109,10 @@ impl ShardsManager {
             entry.push(part_ord);
         }
 
-        let receipt_proofs =
-            self.make_outgoing_receipts_proofs(&chunk_header, &outgoing_receipts, Arc::new)?;
+        let receipt_proofs = self
+            .make_outgoing_receipts_proofs(&chunk_header, &outgoing_receipts)?
+            .map(Arc::new)
+            .collect::<Vec<_>>();
         for (to_whom, part_ords) in block_producer_mapping {
             let part_receipt_proofs = receipt_proofs
                 .iter()
```

mina86: Well, yeah, I considered that and that works for me but notice that…

> That seems like a clear win -- no geneircs,

…there are generics here.

chain/chunks/src/lib.rs
mina86: Works for me. Makes the code simpler.

chain/chunks/src/lib.rs
matklad: There's no generics here, `impl Trait` syntax does very different things before and after `->`

```rust
// This is generic: for every distinct type of `x`, we'll get a separate copy of `generic`: `generic<i32>`, `generic<Vec<()>>`, etc
fn generic(x: &impl Clone) {
  let _x = x.clone();
}


// This is not generic -- there's only one copy of `monomorphic`, it return a single type (`u32`), it's just that we are too lazy to spell it out
fn monomorphic() -> impl Clone {
  92u32
}
```

mina86: Ah, yeah, you’re right.

mzhangmzz (approved):

PR #6380, opened 2022-03-03
a loadtest tool which fetches a fragment of the chain

pompon0:
Implemented a loadtest tool which fetches a fragment of the chain at a given QPS.

Additionally implemented a bunch of structured concurrency primitives.

### DIFF
--- a/tools/chainsync-loadtest/Cargo.toml
+++ b/tools/chainsync-loadtest/Cargo.toml
@@ -1,0 +1,38 @@
+[package]
+name = "chainsync-loadtest"
+version = "0.0.0"
+authors = ["Near Inc <hello@nearprotocol.com>"]
+publish = false
+# Please update rust-toolchain.toml as well when changing version here:
+rust-version = "1.56.0"
+edition = "2021"
+default-run = "chainsync-loadtest"
+
+[[bin]]
+path = "src/main.rs"
+name = "chainsync-loadtest"
+
+[dependencies]
+clap = "=3.0.0-beta.2"
+actix = "=0.11.0-beta.2"
+tracing = "0.1.13"
+tracing-subscriber = "0.2.4"
+openssl-probe = "0.1.4"
+futures = "0.3"
+anyhow = "1.0.55"
+rand = "0.8"
+dirs = "3"
+tokio = "1.16.1"
+log = "0.4"
+
+near-chain-configs = { path = "../../core/chain-configs" }
+near-crypto = { path = "../../core/crypto" }
+near-primitives = { path = "../../core/primitives" }
+near-store = { path = "../../core/store" }
+nearcore = { path = "../../nearcore" }
+near-network = { path = "../../chain/network" }
+near-network-primitives = { path = "../../chain/network-primitives" }
+
+[features]
+[package.metadata.workspaces]
+independent = true
--- a/tools/chainsync-loadtest/Cargo.toml
+++ b/tools/chainsync-loadtest/Cargo.toml
@@ -15,6 +15,7 @@ name = "chainsync-loadtest"
 [dependencies]
 clap = "=3.0.0-beta.2"
 actix = "=0.11.0-beta.2"
+parking_lot = "0.11.2"
 tracing = "0.1.13"
 tracing-subscriber = "0.2.4"
 openssl-probe = "0.1.4"
--- a/tools/chainsync-loadtest/src/concurrency/ctx.rs
+++ b/tools/chainsync-loadtest/src/concurrency/ctx.rs
@@ -1,0 +1,183 @@
+use crate::concurrency::Once;
+use std::fmt;
+use std::ops;
+use std::sync::Arc;
+use tokio::time;
+
+// CtxErr represents an error that Context can return.
+#[derive(Debug, Clone, PartialEq)]
+pub enum CtxErr {
+    Timeout,   // context (or one of the ancestors) has timed out.
+    Cancelled, // context (or one of the ancestors) has been manually cancelled.
+}
+
+impl fmt::Display for CtxErr {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        return write!(f, "{:?}", self);
+    }
+}
+
+impl std::error::Error for CtxErr {
+    fn source(&self) -> Option<&(dyn std::error::Error + 'static)> {
+        return None;
+    }
+}
+
+struct Ctx_ {
+    label: Option<String>,
+    done: Arc<Once<CtxErr>>,
+    deadline: Option<time::Instant>,
+    parent: Option<Arc<Ctx_>>,
+}
+
+impl Drop for Ctx_ {
+    fn drop(&mut self) {
+        self.done.set(CtxErr::Cancelled);
+    }
+}
+
+// Ctx is an implementation of https://pkg.go.dev/context.
+// Ctxs are a mechanism of broadcasting cancellation to concurrent routines (aka futures).
+// The routines are expected to react to context cancellation on their own (they are not preempted)
+// and perform a graceful shutdown. In general the graceful shutdown is expected to
+// be non-blocking, but it still can be asynchronous (if it is known that an async call will
+// be non-blocking as well). Ctx is similar to a rust lifetime but for runtime:
+// Ctx is expected to be passed down the call stack and to the spawned subroutines.
+// At any level of the call stack the Ctx can be narrowed down via Ctx::with_cancel (which
+// allows the routine to cancel context of subroutines it spawns) or via
+// Ctx::with_timeout/Ctx::with_deadline, which cancels the scope automatically after a given time.
+// If is NOT possible to extend the context given by the parent - a subroutine is expected
+// to adhere to the lifetime it represents and finish as soon as it gets cancelled (or earlier).
+#[derive(Clone)]
+pub struct Ctx(Arc<Ctx_>);
+
+impl std::fmt::Debug for Ctx {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        self.0.parent.as_ref().map(|p| Ctx(p.clone()).fmt(f));
+        self.0.label.as_ref().map(|l| f.write_fmt(format_args!("::{}", &l)));
+        Ok(())
+    }
+}
+
+impl Ctx {
+    // background() returns a top level context equivalent to 'static lifetime.
+    // It should be called directly from main.rs.
+    pub fn background() -> Ctx {
+        return Ctx(Arc::new(Ctx_ {
+            label: None,
+            parent: None,
+            deadline: None,
+            done: Arc::new(Once::new()),
+        }));
+    }
+
+    fn cancel(&self) {
+        self.0.done.set(CtxErr::Cancelled);
+    }
+
+    // err() returns the CtxErr if the Context was already cancelled,
+    // None otherwise.
+    pub fn err(&self) -> Option<CtxErr> {
+        self.0.done.get()
+    }
+
+    // done() waits until the context gets cancelled and returns the CtxErr.
+    pub async fn done(&self) -> CtxErr {
+        self.0.done.wait().await
+    }
+
+    // wrap() executes the future f to completion, or until the context gets cancelled.
+    // This function is expected to be called only by the low level code, which
+    // wraps context-unaware code (context-aware code is expected to take ctx as an argument
+    // instead).
+    pub async fn wrap<F, T>(&self, f: F) -> Result<T, CtxErr>
+    where
+        F: std::future::Future<Output = T>,
+    {
+        tokio::select! {
+            v = f => Ok(v),
+            e = self.0.done.wait() => Err(e),
+        }
+    }
+
+    // wait_until() wraps tokio::time::sleep_until. Returns Ok(()) if deadline
+    // has passed, or CtxErr if the context got cancelled.
+    pub async fn wait_until(&self, deadline: time::Instant) -> Result<(), CtxErr> {
+        self.wrap(time::sleep_until(deadline)).await
+    }
+
+    // wait() wraps tokio::time::sleep (see also wait_until()).
+    pub async fn wait(&self, duration: time::Duration) -> Result<(), CtxErr> {
+        self.wrap(time::sleep(duration)).await
+    }
+
+    fn new_child(&self, label: Option<String>, deadline: Option<time::Instant>) -> Ctx {
+        let deadline = match (deadline, self.0.deadline) {
+            (Some(cd), Some(pd)) => Some(std::cmp::min(cd, pd)),
+            (None, pd) => pd,
+            (cd, None) => cd,
+        };
+        let done = Arc::new(Once::<CtxErr>::new());
+        tokio::spawn({
+            let parent_done = self.0.done.clone();
+            let child_done = done.clone();
+            async move {
+                let deadline_wait = async {
+                    if let Some(d) = deadline {
+                        time::sleep_until(d).await
+                    } else {
+                        std::future::pending().await
+                    }
+                };
+                tokio::select! {
+                    _ = deadline_wait => { child_done.set(CtxErr::Timeout); }
+                    e = parent_done.wait() => { child_done.set(e); }
+                    _ = child_done.wait() => {}
+                }
+            }
+        });
+        return Ctx(Arc::new(Ctx_ { label, parent: Some(self.0.clone()), done, deadline }));
+    }
+
+    // with_cancel creates a child context which can be manually cancelled.
+    // If the parent context gets cancelled, the child will also be cancelled,
+    // but not vice versa (you cannot affect the parent context).
+    pub fn with_cancel(&self) -> CtxWithCancel {
+        return CtxWithCancel(self.new_child(None, None));
+    }
+
+    // with_deadline() creates a child context which will be automatically cancelled,
+    // once deadline passes (or parent context gets cancelled).
+    pub fn with_deadline(&self, deadline: tokio::time::Instant) -> Ctx {
+        return self.new_child(None, Some(deadline));
+    }
+
+    // with_label() is a debugging utility which allows to create child context
+    // with a custom label. The path of labels of the ancestors is returned
+    // in the Debug implementation for Ctx.
+    pub fn with_label(&self, label: &str) -> Ctx {
+        return self.new_child(Some(label.to_string()), None);
+    }
+
+    // with_timeout() is the same as with_deadline() but you provide a duration,
+    // rather than an instant.
+    pub fn with_timeout(&self, timeout: time::Duration) -> Ctx {
+        return self.with_deadline(time::Instant::now() + timeout);
+    }
+}
+
+#[derive(Clone)]
+pub struct CtxWithCancel(Ctx);
+
+impl ops::Deref for CtxWithCancel {
+    type Target = Ctx;
+    fn deref(&self) -> &Ctx {
+        return &self.0;
+    }
+}
+
+impl CtxWithCancel {
+    pub fn cancel(&self) {
+        self.0.cancel();
+    }
+}
--- a/tools/chainsync-loadtest/src/concurrency/ctx.rs
+++ b/tools/chainsync-loadtest/src/concurrency/ctx.rs
@@ -8,7 +8,7 @@ use tokio::time;
 #[derive(Debug, Clone, PartialEq)]
 pub enum CtxErr {
     /// Context (or one of the ancestors) has timed out.
-    Timeout,   
+    Timeout,
     /// Context (or one of the ancestors) has been manually cancelled.
     Cancelled,
 }
--- a/tools/chainsync-loadtest/src/concurrency/ctx.rs
+++ b/tools/chainsync-loadtest/src/concurrency/ctx.rs
@@ -15,7 +15,12 @@ pub enum Error {
 
 impl fmt::Display for Error {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        return write!(f, "{:?}", self);
+        f.write_str(match self {
+            Self::Timeout => "ctx::Error::Timeout: context or one of the ancestors has timed out",
+            Self::Cancelled => {
+                "ctx::Error::Cancelled: context or one of the ancestors has been cancelled"
+            }
+        })
     }
 }
 
--- a/tools/chainsync-loadtest/src/concurrency/ctx.rs
+++ b/tools/chainsync-loadtest/src/concurrency/ctx.rs
@@ -4,22 +4,22 @@ use std::ops;
 use std::sync::Arc;
 use tokio::time;
 
-// CtxErr represents an error that Context can return.
+// Error represents an error that Context can return.
 #[derive(Debug, Clone, PartialEq)]
-pub enum CtxErr {
+pub enum Error {
     /// Context (or one of the ancestors) has timed out.
     Timeout,
     /// Context (or one of the ancestors) has been manually cancelled.
     Cancelled,
 }
 
-impl fmt::Display for CtxErr {
+impl fmt::Display for Error {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         return write!(f, "{:?}", self);
     }
 }
 
-impl std::error::Error for CtxErr {
+impl std::error::Error for Error {
     fn source(&self) -> Option<&(dyn std::error::Error + 'static)> {
         return None;
     }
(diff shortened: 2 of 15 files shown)


tools/chainsync-loadtest/Cargo.toml
```diff
@@ -0,0 +1,38 @@
+[package]
+name = "chainsync-loadtest"
```
mm-near, line 2: please also add README.md file with info about this tool and examples on how to run it.

tools/chainsync-loadtest/src/fetch_chain.rs
```diff
@@ -0,0 +1,68 @@
...
+use log::info;
+
+use crate::concurrency::{Ctx, Scope};
+use crate::network;
+use anyhow::Context;
+use tokio::time;
+
+use near_primitives::hash::CryptoHash;
+
+pub async fn run(
```
mm-near, line 12: comment

tools/chainsync-loadtest/src/fetch_chain.rs
```diff
@@ -0,0 +1,68 @@
...
+    ctx: Ctx,
+    network: Arc<network::Network>,
+    start_block_hash: CryptoHash,
+) -> anyhow::Result<()> {
+    info!("SYNC start");
+    let peers = network.info(&ctx).await?;
+    let target_height = peers.highest_height_peers[0].chain_info.height;
+    info!("SYNC target_height = {}", target_height);
+    Scope::run(&ctx, |ctx, s| async move {
+        s.spawn_weak({
```
mm-near, line 22: what does spawn_weak do? do we have to clean it up/stop somehow afterwards?

tools/chainsync-loadtest/src/fetch_chain.rs
```diff
@@ -0,0 +1,68 @@
...
+                    info!("stats = {:?}", network.stats);
+                    ctx.wait(time::Duration::from_secs(2)).await?;
+                }
+            }
+        });
+
+        let mut last_hash = start_block_hash;
+        let mut last_height = 0;
+        while last_height < target_height {
+            let mut headers = network.fetch_block_headers(&ctx, &last_hash).await?;
```
mm-near, line 36: comment that this is fetching 512 headers after last_hash

tools/chainsync-loadtest/src/network.rs
```diff
@@ -0,0 +1,330 @@
...
+            }
+            NetworkClientMessages::PartialEncodedChunkResponse(resp) => {
+                self.chunks.get(&resp.chunk_hash.clone()).map(|p| p.set(resp));
+            }
+            _ => {}
+        }
+    }
+}
+
+pub struct ClientActor {
```
mm-near, line 273: maybe we should call it 'FakeClientActor'? (or Mock)

tools/chainsync-loadtest/src/network.rs
```diff
@@ -0,0 +1,330 @@
...
+        };
+        info!("view_request: {}", name);
+        return NetworkViewClientResponses::NoResponse;
+    }
+}
+
+impl Handler<NetworkClientMessages> for ClientActor {
+    type Result = NetworkClientResponses;
+    fn handle(&mut self, msg: NetworkClientMessages, _ctx: &mut Context<Self>) -> Self::Result {
+        // info!("network.notify() on thread {:?}",std::thread::current().name());
```
mm-near, line 326: remove

tools/chainsync-loadtest/src/concurrency/ctx.rs
```diff
@@ -0,0 +1,183 @@
+use crate::concurrency::Once;
+use std::fmt;
+use std::ops;
+use std::sync::Arc;
+use tokio::time;
+
+// CtxErr represents an error that Context can return.
```
mm-near, line 7: FYI: I didn't spend much time reviewing these libraries - as currently they will be used only by this tool.

But if in future, we want to move them to the 'core' - then we should do another thorough review + tests.

tools/chainsync-loadtest/src/network.rs
```diff
@@ -0,0 +1,330 @@
...
+            parts_per_chunk: config.genesis.config.num_block_producer_seats,
+            rate_limiter: RateLimiter::new(
+                time::Duration::from_secs(1) / qps_limit,
+                qps_limit as u64,
+            ),
+            request_timeout: time::Duration::from_secs(2),
+        })
+    }
+
+    fn keep_sending(
```
mm-near, line 115: comment

tools/chainsync-loadtest/src/network.rs
```diff
@@ -0,0 +1,330 @@
...
+                        .network_adapter
+                        .do_send(PeerManagerMessageRequest::NetworkRequests(new_req(peer.clone())));
+                    self_.stats.msgs_sent.fetch_add(1, Ordering::Relaxed);
+                    ctx.wait(self_.request_timeout).await?;
+                }
+            }
+        }
+    }
+
+    pub async fn info(self: &Arc<Self>, ctx: &Ctx) -> anyhow::Result<Arc<NetworkInfo>> {
```
mm-near, line 139: comment

tools/chainsync-loadtest/src/network.rs
```diff
@@ -0,0 +1,330 @@
...
+    header_start: AtomicU64,
+    header_done: AtomicU64,
+    block_start: AtomicU64,
+    block_done: AtomicU64,
+    chunk_start: AtomicU64,
+    chunk_done: AtomicU64,
+}
+
+struct NetworkData {
+    info_futures: Vec<oneshot::Sender<Arc<NetworkInfo>>>,
```
mm-near, line 54: comment

tools/chainsync-loadtest/src/network.rs
```diff
@@ -0,0 +1,330 @@
...
+
+    fn keep_sending(
+        self: &Arc<Self>,
+        ctx: &Ctx,
+        new_req: impl Fn(FullPeerInfo) -> NetworkRequests + Send,
+    ) -> impl Future<Output = anyhow::Result<()>> + Send {
+        let self_ = self.clone();
+        let ctx = ctx.with_label("keep_sending");
+        async move {
+            loop {
```
mm-near, line 123: when do we exit this loop?

mm-near (changes requested):

tools/chainsync-loadtest/src/concurrency/ctx.rs
pompon0: yes, this is just a PoC. I'd like to get some traction on the ideas encoded here and some opinions of our Rust experts before we go any further with that. Then we should definitely invest into adding tests and finding a decent location for this code.

tools/chainsync-loadtest/src/fetch_chain.rs
pompon0: done

tools/chainsync-loadtest/Cargo.toml
pompon0: done

tools/chainsync-loadtest/src/fetch_chain.rs
pompon0: PTAL at the extensive comment in the scope.rs.
Scope awaits for all routines to complete before the scope is completed.
If any routine returns an error, the whole scope is cancelled, and cancellation is propagated to all other routines of the scope.
Still there are 2 kinds of routines in the scope: "main" and "weak". As soon as all "main" routines complete, the scope gets cancelled (even if there was no error), and the cancellation is propagated to the remaining ("weak") routines. The idea is that "weak" routines are not expected to ever complete on their own.

tools/chainsync-loadtest/src/fetch_chain.rs
pompon0: Done. Added a comment on the fech_block_headers() function.

tools/chainsync-loadtest/src/network.rs
pompon0: done

tools/chainsync-loadtest/src/network.rs
pompon0: done

tools/chainsync-loadtest/src/network.rs
pompon0: Only if the context expires.
The happy path looks as follows:
1. we spawn keep_sending()
2. we await a response
3. once we get the response, we cancel keep_sending(), by cancelling the context.

tools/chainsync-loadtest/src/network.rs
pompon0: done

tools/chainsync-loadtest/src/network.rs
pompon0: done

tools/chainsync-loadtest/src/network.rs
pompon0: done

mm-near (approved):

tools/chainsync-loadtest/src/concurrency/ctx.rs
bowenwang1996: I'd ask some experts to review the libraries :) @matklad @nagisa could you take a look if you have time

tools/chainsync-loadtest/src/concurrency/rate_limiter.rs
```diff
@@ -0,0 +1,54 @@
...
+    interval: time::Duration,
+    burst: u64,
+    tokens: u64,
+    ticks_processed: u64,
+    start: time::Instant,
+}
+
+impl RateLimiter_ {
+    fn ticks(&self, t: time::Instant) -> u64 {
+        return ((t - self.start).as_secs_f64() / self.interval.as_secs_f64()) as u64;
```
nagisa, line 15: This can sometimes (in very rare instances) produce a `(0.0/0.0) as u64 = 0` for `self.interval = 0ns`, even though the intent is probably for it to be a `+INF as u64` at all times.

tools/chainsync-loadtest/src/concurrency/rate_limiter.rs
```diff
@@ -0,0 +1,54 @@
...
+            start: time::Instant::now(),
+            ticks_processed: 0,
+        })));
+    }
+
+    // See semantics of https://pkg.go.dev/golang.org/x/time/rate
+    pub async fn allow(&self, ctx: &Ctx) -> anyhow::Result<()> {
+        let mut rl = ctx.wrap(self.0.lock()).await?;
+        let ticks_now = rl.ticks(time::Instant::now());
+        rl.tokens = std::cmp::min(rl.burst, rl.tokens + (ticks_now - rl.ticks_processed));
```
nagisa, line 44: Wrapping arithmetic here when `ticks_now` approaches `u64::MAX`, I believe. Will `debug_assert!`. Use `wrapping_add/sub`.

tools/chainsync-loadtest/src/concurrency/rate_limiter.rs
pompon0: The intent is to panic if interval is 0.

tools/chainsync-loadtest/src/concurrency/once.rs
```diff
@@ -0,0 +1,58 @@
...
+}
+
+impl<T: Clone + Send + Sync> Once<T> {
+    pub fn new() -> Once<T> {
+        return Once { value: RwLock::new(None), notify: tokio::sync::Notify::new() };
+    }
+
+    // set() sets the value of Once to x and returns true.
+    // If the value was already set, it just returns false.
+    pub fn set(self: &Self, x: T) -> bool {
```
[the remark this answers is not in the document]
pompon0: why?

tools/chainsync-loadtest/src/concurrency/rate_limiter.rs
pompon0: I want it to panic in such case. time::Instant also is capable of representing a limited time range and it is not like the time will stop afterwards. I'm not expecting for the code to work in such a case.

tools/chainsync-loadtest/src/concurrency/once.rs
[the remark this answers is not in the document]
nagisa: To give a brief answer, because the return value here indicates a success/failure of the requested operation, and `Result` does a much better job at documenting this than `bool`, both to the user and the compiler.

tools/chainsync-loadtest/src/concurrency/once.rs
```diff
@@ -0,0 +1,58 @@
...
+    is_send::<Once<T>>();
+    is_sync::<Once<T>>();
+}
+
+// Once is a synchronization primitive, which stores a single value.
+// This value can be set at most once, and multiple consumers are
+// allowed to wait for that value.
+pub struct Once<T: Clone + Send + Sync> {
+    value: RwLock<Option<T>>,
+    notify: tokio::sync::Notify,
```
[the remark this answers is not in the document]
pompon0: > Mixing in a synchronous RwLock into what otherwise seems like a type for use in async contexts seems like a potential footgun.

Synchronization of asynchronous tasks requires the use of synchronous primitives at some layer of abstraction. It is as footgun'ish as using mutexes in general.

I missed the fact that std::sync::RwLock is not guaranteed to be fair. How about I just replace it with parking_lot::RwLock ? It should be robust enough, given that by design there should be just 1 write, so there is no point in optimizing for multiple writers (here we just care about correctness).

tools/chainsync-loadtest/src/concurrency/weak_map.rs
```diff
@@ -0,0 +1,73 @@
...
+use std::collections::hash_map::Entry;
+use std::collections::HashMap;
+use std::hash::Hash;
+use std::ops::Deref;
+use std::sync::{Arc, Mutex, Weak};
+
+// WeakMap is a collection of weak pointers.
+// Once the last reference to an element of the map is dropped,
+// the weak pointer is removed from the map.
+pub struct WeakMap<K: Hash + Eq + Clone, V> {
```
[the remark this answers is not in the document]
pompon0: Can you elaborate? I don't see your comments on type-level bounds anywhere else.

tools/chainsync-loadtest/src/concurrency/weak_map.rs
```diff
@@ -0,0 +1,73 @@
...
+    }
+}
+
+impl<K: Hash + Eq + Clone, V> Drop for Ref<K, V> {
+    fn drop(&mut self) {
+        // Be careful to not to drop an Arc here,
+        // because that might trigger this function
+        // recursively and cause a deadlock.
+        let mut m = self.map.inner.lock().unwrap();
+        let e = match m.entry(self.key.clone()) {
```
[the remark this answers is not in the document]
pompon0: Isn't that a microoptimization, given that I need the key to be cloned in multiple places anyway?

tools/chainsync-loadtest/src/concurrency/once.rs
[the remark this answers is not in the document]
pompon0: Is it part of some style guide? I'd expect "cargo fmt" to do that for me if it was.

tools/chainsync-loadtest/src/concurrency/ctx.rs
```diff
@@ -0,0 +1,183 @@
...
+#[derive(Debug, Clone, PartialEq)]
+pub enum CtxErr {
+    Timeout,   // context (or one of the ancestors) has timed out.
+    Cancelled, // context (or one of the ancestors) has been manually cancelled.
+}
+
+impl fmt::Display for CtxErr {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        return write!(f, "{:?}", self);
```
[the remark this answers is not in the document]
pompon0: are you sure? This will end up in a stack trace, so it should rather be developer-friendly. End user is expected to just copy-paste the stack trace.

tools/chainsync-loadtest/src/concurrency/once.rs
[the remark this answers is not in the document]
pompon0: Yes, however tokio-provided async primitives (which internally are using sync primitives anyway) didn't satisfy my requirements, therefore I've augmented them to construct another async primitive (also by using some sync primitives internally).

tools/chainsync-loadtest/src/concurrency/ctx.rs
```diff
@@ -0,0 +1,183 @@
+use crate::concurrency::Once;
+use std::fmt;
+use std::ops;
+use std::sync::Arc;
+use tokio::time;
+
+// CtxErr represents an error that Context can return.
+#[derive(Debug, Clone, PartialEq)]
+pub enum CtxErr {
```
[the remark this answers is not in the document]
pompon0: done

tools/chainsync-loadtest/src/concurrency/once.rs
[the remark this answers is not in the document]
pompon0: ack, done.

tools/chainsync-loadtest/src/concurrency/weak_map.rs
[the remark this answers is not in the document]
nagisa: I was referring to https://github.com/near/nearcore/pull/6380#discussion_r820644264

tools/chainsync-loadtest/src/concurrency/once.rs
```diff
@@ -0,0 +1,58 @@
...
+#[allow(dead_code)]
+fn test<T: Clone + Send + Sync>() {
+    is_send::<Once<T>>();
+    is_sync::<Once<T>>();
+}
+
+// Once is a synchronization primitive, which stores a single value.
+// This value can be set at most once, and multiple consumers are
+// allowed to wait for that value.
+pub struct Once<T: Clone + Send + Sync> {
```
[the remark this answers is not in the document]
pompon0: done

tools/chainsync-loadtest/src/concurrency/ctx.rs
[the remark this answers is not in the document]
nagisa: Even if the user here is the developer, I would argue that a more extensive description is going to be more actionable than just the variant name most of the time: it directly explains in the output what is happening, without anybody having to dig into the code/documentation to obtain more context.

There is another way to think about it – there isn't all that much of a purpose in documenting an error variant if its meaning is obvious from just the variant name itself, is it? I'd say similar line of reasoning holds for using `Debug` in implementing `Display`.

tools/chainsync-loadtest/src/concurrency/once.rs
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+    // set() sets the value of Once to x and returns true.
+    // If the value was already set, it just returns false.
+    pub fn set(self: &Self, x: T) -> bool {
+        let mut v = self.value.write().unwrap();
+        if v.is_some() {
+            return false;
+        }
+        *v = Some(x.clone());
```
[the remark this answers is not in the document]
pompon0: done

tools/chainsync-loadtest/src/concurrency/weak_map.rs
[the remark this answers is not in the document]
pompon0: Here it doesn't work, because these bounds are required by the Drop trait.

tools/chainsync-loadtest/src/concurrency/once.rs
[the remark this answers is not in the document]
nagisa: I'm not sure if there's a style guide for this anywhere. `self`, `&self` and `&mut self` predate the ability to specify the self type at all, so for at least references everybody has been writing it the old way (both because its terser and because everybody else is doing it the old way.)

`rustfmt` fixing this up sounds like a good idea, but explicit self types are so obscure that this likely just never came up. Mind filing an issue against `rustfmt`, or should I?

tools/chainsync-loadtest/src/concurrency/ctx.rs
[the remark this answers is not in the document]
pompon0: Personally, I never trust the error messages and all I expect from them is to point me precisely to the code location where they originated from (I need to do that anyway to do the debugging) and a rust identifier is easier to search for than a human-readable string, which may: be non-unique, contain whitespaces which are harder to search for, be broken into multiple lines, or in the worst case be constructed dynamically.

How about a trade-off: I'll put both rust identifier and a human friendly message to the error message?

tools/chainsync-loadtest/src/concurrency/once.rs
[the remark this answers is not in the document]
pompon0: If you consider it worth it, please file a request. Like, IMHO converging towards self:&Self, rather than towards &self would make the syntax more uniform (since for wrapping types you have to state type explicitly anyway), so personally I don't know if it is a good idea in the first place. I was just curious whether this decision has already been "set in stone" in some sense.

nagisa (approved):
Just FWIW my review on the libraries here isn't blocking landing of this PR in any way.

It is nice if we can make this code a little bit more polished, and my review was focusing primarily on that aspect of the libraries.

If it came to moving these to a more permanent location like “core”, I would really debate whether we should maintain synchronization primitives of our own, let alone in tree. My immediate reaction to any PR trying to do so would be to find ways to implement a similar functionality using the tools already available in the ecosystem, and only when it is impossible – to use our own. In that sort of situation I would definitely spend much more time in reviewing the correctness of the primitives.

tools/chainsync-loadtest/src/concurrency/scope.rs
```diff
@@ -0,0 +1,184 @@
...
+    pub fn spawn<F>(self: &Arc<Self>, f: impl Send + FnOnce(Ctx, Arc<Self>) -> F)
+    where
+        F: Future<Output = anyhow::Result<()>> + Send + 'static,
+    {
+        let fut = f((*self.ctx).clone(), self.clone());
+        let s = self.clone();
+        s.main_futures.inc();
+        tokio::spawn(async move {
+            s.complete(fut.await);
+            s.main_futures.dec();
```
matklad, line 96: This is suspect -- if `fut` panics, we don't do `dec` (and, in general, we swallow the panic).

tools/chainsync-loadtest/src/concurrency/scope.rs
```diff
@@ -0,0 +1,184 @@
...
+    // - Even though the spawned futures are guaranteed to have a bounded lifetime, I failed so
+    // far to express that reliably in rust lifetimes, therefore for now all futures are required
+    // to be 'static.
+    // - As a consequence the Scope object lifetime is also 'static, so there is no compile-time
+    // protection from having scope outlive the run() call.
+    // - There is no compile-time protection from having "weak" futures spawn stuff. Any
+    // suggestions how to achieve that are welcome.
+    pub async fn run<F, T>(ctx: &Ctx, f: impl FnOnce(Ctx, Arc<Scope>) -> F) -> anyhow::Result<T>
+    where
+        F: Future<Output = anyhow::Result<T>> + Send + 'static,
```
matklad, line 150: Ok for this limeted scope, but we shouldn't enshrine `anyhow` at this layer of abstraction.

tools/chainsync-loadtest/src/concurrency/scope.rs
```diff
@@ -0,0 +1,184 @@
...
+    // Ctx) but won't get preempted: they will be able to shutdown gracefully.
+    //
+    // run() is a lambda-abstraction of the Scope:
+    // 1. it creates a new scope, spawns f(ctx,scope) and adds it to the scope.
+    // 2. This first future of the scope is allowed to spawn additional futures in the scope.
+    // 3. run() is blocking until all members of the scope complete.
+    //
+    // Since from the point of view of the caller of Scope::run, it is just a blocking call, which
+    // might spawn coroutines internally, but all of them will complete before Scope::run returns,
+    // therefore Scope::run allows for expressing STRUCTURED CONCURRENCY.
```
matklad, line 126: **Important Note** this is going to be only best-effort structured concurrency, due to how cancellation-via-drop works, it's not really possible to express water-tight structured concurrency in `async` Rust, as far as I am aware. Here's a test that shows a task outliving it's scope:

```rust
#[tokio::test]
async fn break_structured_concurrency() {
    {
        let ctx = Ctx::background();
        let scope = Scope::run(&ctx, |_ctx, scope| async move {
            scope.spawn(|_ctx, scope| async move {
                loop {
                    tokio::time::sleep(Duration::from_millis(500)).await;
                    eprintln!("still running");
                }
            });
            Ok(())
        });
        let sleep = tokio::time::sleep(Duration::from_millis(100));
        tokio::pin!(sleep);
        tokio::select! {
            biased;
            _ = scope => {}
            _ = sleep => {}
        }

    }

    loop {
        tokio::task::yield_now().await
    }
}
```

matklad (commented):
Belated review!

PR #6381, opened 2022-03-03
Chunk delay metric: report delay per shard

nikurt:
Adds a new metric by tracking the event of requesting a chunk.

Changes how the timestamps are stored. Instead of triggering metric update at every opportunity, wait for the block to be processed and then update the metrics. Makes things a lot simpler.

Updates the existing chunks delay metric to report a histogram:
* Time between learning about a block and receiving the last of its chunks

Metric "blocks ahead of head" is removed as redundant, as it is mostly the same as the number of orphan blocks.

### DIFF
--- a/chain/client/src/chunks_delay_tracker.rs
+++ b/chain/client/src/chunks_delay_tracker.rs
@@ -1,4 +1,3 @@
-use std::cmp::max;
 use std::collections::{BTreeMap, HashMap};
 use std::time::{Duration, Instant};
 
@@ -19,17 +18,15 @@ pub(crate) struct ChunksDelayTracker {
     heights: BTreeMap<BlockHeight, HeightInfo>,
 }
 
-const CHUNKS_DELAY_TRACKER_HORIZON: u64 = 10;
-
 #[derive(Debug, Default)]
 struct HeightInfo {
     block_received: Option<Instant>,
     chunks_received: HashMap<ShardId, Instant>,
-    chunks_requested: Option<Instant>,
+    chunks_requested: HashMap<ShardId, Instant>,
 }
 
 impl HeightInfo {
-    pub fn get_delay(&self) -> Duration {
+    pub fn get_block_chunk_delay(&self) -> Duration {
         let block_received = self.block_received.unwrap();
         let latest_chunk_received = self.chunks_received.values().max().unwrap();
         latest_chunk_received.saturating_duration_since(block_received)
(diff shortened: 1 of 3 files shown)


chain/client/src/chunks_delay_tracker.rs
```diff
@@ -19,174 +18,76 @@ pub(crate) struct ChunksDelayTracker {
     heights: BTreeMap<BlockHeight, HeightInfo>,
 }
 
-const CHUNKS_DELAY_TRACKER_HORIZON: u64 = 10;
-
 #[derive(Debug, Default)]
 struct HeightInfo {
     block_received: Option<Instant>,
     chunks_received: HashMap<ShardId, Instant>,
+    chunks_requested: HashMap<ShardId, Instant>,
```
mm-near, line 25: please comment what these fields are

chain/client/src/chunks_delay_tracker.rs
```diff
@@ -19,174 +18,76 @@ pub(crate) struct ChunksDelayTracker {
     heights: BTreeMap<BlockHeight, HeightInfo>,
```
mm-near, line 18: you might want to update the comment above

chain/client/src/chunks_delay_tracker.rs
```diff
@@ -19,174 +18,76 @@ pub(crate) struct ChunksDelayTracker {
...
+    chunks_requested: HashMap<ShardId, Instant>,
 }
 
 impl ChunksDelayTracker {
-    // Blocks below this heights are not tracked.
-    fn lowest_height(head_height: BlockHeight) -> BlockHeight {
-        head_height.saturating_sub(CHUNKS_DELAY_TRACKER_HORIZON)
-    }
-
     fn remove_old_entries(&mut self, head_height: BlockHeight) {
```
mm-near, line 29: comment - what does head_hight mean here?

chain/client/src/chunks_delay_tracker.rs
mm-near: also - maybe rather than 'heights' I'd call it 'blocks_in_progress'?

And then 'remove_old_entries' would be something like 'remove_completed_blocks'.

mm-near (changes requested):

chain/client/src/chunks_delay_tracker.rs
```diff
@@ -1,192 +1,89 @@
...
-        assert_eq!(tracker.get_blocks_ahead(head_height), 1);
-
-        head_height = 5;
-        tracker.add_block_timestamp(4, head_height, now);
-        assert_eq!(tracker.get_blocks_ahead(head_height), 0);
-
-        head_height = 999;
-        tracker.add_block_timestamp(4, head_height, now);
-        assert_eq!(tracker.get_blocks_ahead(head_height), 0);
+    pub fn processed_block(&mut self, processed_block_height: BlockHeight) {
```
mzhangmzz, line 71: nit: I would rename this function to be `finish_block_processing` to indicate that this function should be called when a block is fully processed

chain/client/src/chunks_delay_tracker.rs
```diff
@@ -1,192 +1,89 @@
...
+        let heights_to_drop: Vec<BlockHeight> =
```
mzhangmzz, line 72: I think it's better to just drop the block at the specific height, because blocks before the height may or may not be finished if there are forks. And again, I think it's better to index blocks by hash instead of by height.

chain/client/src/chunks_delay_tracker.rs
```diff
@@ -1,192 +1,89 @@
...
-/// 2) the difference between the latest known block and the current head.
-/// If a chunk or a block is received multiple times, only the first time is recorded.
+/// Provides monitoring information about the delays of receiving and requesting blocks and their corresponding chunks.
+/// Keeps timestamps for all future blocks and chunks that were requested or received.
+/// Updates the metrics when the head gets updated. Assumes that if the head moves forward, all needed chunks were received.
+/// If a chunk or a block is received or requested multiple times, only the first time is recorded.
 #[derive(Debug, Default)]
 pub(crate) struct ChunksDelayTracker {
-    heights: BTreeMap<BlockHeight, HeightInfo>,
+    blocks_in_progress: BTreeMap<BlockHeight, BlockInProgress>,
```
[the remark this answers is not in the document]
mm-near: yes, that's a good idea.

mm-near (changes requested):
+1 to Min's suggestion about keying them as hash rather than height

mzhangmzz (commented):
Good point on there is no block hash in chunk headers. Indexing by block hash will need to more complicated then.

Unfortunately, we can't key by `prev_block_hash` because more than one blocks may depend on the same previous block. We have to key by the current block hash. Because we can't get block hash from chunk headers, we will need to store a map from chunk hash to block hash in `ChunksDelayTracker`. New items can be added to the map in `record_receive_block_timestamp` because there we have information of the full block.

nikurt:
Updated the PR to track both blocks and chunks by hashes. The caveat is that sometimes chunks are received before the corresponding block is received. To handle this the PR records timestamps of all received/requested chunks. To avoid unlimited memory growth, the chunks' timestamps are stored in an LRU cache.

chain/client/src/chunks_delay_tracker.rs
```diff
@@ -1,192 +1,129 @@
...
-                }
+    pub fn add_chunk_timestamp(&mut self, chunk_hash: &ChunkHash, timestamp: Instant) {
+        match self.chunks_in_progress.get_mut(chunk_hash) {
+            Some(entry) => {
+                entry.chunk_received.get_or_insert(timestamp);
+            }
+            None => {
+                self.chunks_in_progress.put(
+                    chunk_hash.clone(),
+                    ChunkInProgress { chunk_received: Some(timestamp), chunk_requested: None },
```
mzhangmzz, line 55: It's ok to not track chunks that are not requested since we care about the delay between when a chunk is requested vs when it is received. If it is received before it is requested, we can just ignore this data. This also simplified the implementation so you can get rid of the None case here.

chain/client/src/chunks_delay_tracker.rs
```diff
@@ -1,192 +1,129 @@
...
-        if let Some((&latest, _)) = self.heights.iter().next_back() {
-            latest.saturating_sub(head_height)
-        } else {
-            0
+    pub fn requested_chunk(&mut self, chunk_hash: &ChunkHash, timestamp: Instant) {
+        match self.chunks_in_progress.get_mut(chunk_hash) {
+            Some(entry) => {
+                entry.chunk_requested.get_or_insert(timestamp);
+            }
+            None => {
```
mzhangmzz, line 66: Similarly, you can get rid of the "Some" case here

chain/client/src/chunks_delay_tracker.rs
```diff
@@ -1,192 +1,129 @@
...
-            self.heights.entry(height).or_default().block_received.get_or_insert(timestamp);
+        if let Some(block_received) = self.blocks_received.get(&block_hash) {
+            for (chunk_hash, shard_id) in chunks {
+                let chunk_in_progress = self.chunks_in_progress.peek(&chunk_hash.clone());
+                if let Some(chunk_in_progress) = chunk_in_progress {
+                    if let Some(chunk_received) = chunk_in_progress.chunk_received {
+                        metrics::BLOCK_CHUNKS_DELAY
+                            .with_label_values(&[&format!("{}", shard_id)])
+                            .observe(
+                                chunk_received
```
mzhangmzz, line 102: It's probably better to track the time between "block_received" and "chunk_requested" instead of between "block_received" and "chunk_received", since we have another metric for between "chunk_requested" and "chunk_received"

chain/client/src/chunks_delay_tracker.rs
```diff
@@ -1,192 +1,129 @@
...
 
 #[derive(Debug, Default)]
-struct HeightInfo {
-    block_received: Option<Instant>,
-    chunks_received: HashMap<ShardId, Instant>,
+struct ChunkInProgress {
+    /// Timestamps of receiving each of the shards in a block.
+    chunk_received: Option<Instant>,
+    /// Timestamps of requesting missing chunks.
+    /// Chunks are usually requested when a client falls out-of-sync with the head of the chain.
```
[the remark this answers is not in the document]
nikurt: Done

chain/client/src/chunks_delay_tracker.rs
```diff
@@ -1,192 +1,129 @@
...
-        let mut max_delay = Duration::ZERO;
-        for (_, v) in self.heights.range(..=head_height) {
-            if let Some(block_received) = v.block_received {
-                if let Some(latest_chunk) = v.chunks_received.values().max() {
-                    if latest_chunk > &block_received {
-                        let delay = latest_chunk.duration_since(block_received);
-                        max_delay = max(max_delay, delay);
-                    }
-                }
+    pub fn add_chunk_timestamp(&mut self, chunk_hash: &ChunkHash, timestamp: Instant) {
```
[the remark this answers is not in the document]
nikurt: Done

chain/client/src/chunks_delay_tracker.rs
```diff
@@ -1,192 +1,129 @@
...
+    /// Timestamps of requesting missing chunks.
+    /// Chunks are usually requested when a client falls out-of-sync with the head of the chain.
+    chunk_requested: Option<Instant>,
 }
 
 impl ChunksDelayTracker {
-    // Blocks below this heights are not tracked.
-    fn lowest_height(head_height: BlockHeight) -> BlockHeight {
-        head_height.saturating_sub(CHUNKS_DELAY_TRACKER_HORIZON)
+    pub fn add_block_timestamp(&mut self, block_hash: &CryptoHash, timestamp: Instant) {
```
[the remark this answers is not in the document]
nikurt: Done

chain/client/src/chunks_delay_tracker.rs
```diff
@@ -1,192 +1,129 @@
...
+        self.update_block_chunks_metric(processed_block_hash, &chunks);
+        self.update_chunks_metric(&chunks);
+        self.blocks_received.remove(&processed_block_hash);
     }
 
-    fn update_metrics(&mut self, head_height: BlockHeight) {
-        self.update_chunks_metric(head_height);
-        self.update_blocks_ahead_metric(head_height);
+    pub fn get_chunks(block: &Block) -> Vec<(ChunkHash, ShardId)> {
+        block.chunks().iter().map(|chunk| (chunk.chunk_hash(), chunk.shard_id())).collect()
```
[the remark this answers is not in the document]
nikurt: That will work well enough short term but after next steps of sharding are launched, this assumption will no longer be true.

chain/client/src/chunks_delay_tracker.rs
```diff
@@ -1,192 +1,129 @@
...
-    fn update_blocks_ahead_metric(&mut self, head_height: BlockHeight) {
-        metrics::BLOCKS_AHEAD_OF_HEAD.set(self.get_blocks_ahead(head_height) as i64);
+    pub fn finish_block_processing(
+        &mut self,
+        processed_block_hash: &CryptoHash,
+        chunks: &[(ChunkHash, ShardId)],
+    ) {
+        self.update_block_chunks_metric(processed_block_hash, &chunks);
+        self.update_chunks_metric(&chunks);
+        self.blocks_received.remove(&processed_block_hash);
```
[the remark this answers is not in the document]
nikurt: Done

chain/client/src/chunks_delay_tracker.rs
nikurt: Agreed

chain/client/src/chunks_delay_tracker.rs
nikurt: Done

chain/client/src/chunks_delay_tracker.rs
nikurt: Done

chain/client/src/chunks_delay_tracker.rs
```diff
@@ -1,192 +1,129 @@
...
     }
 
-    pub fn add_block_timestamp(
+    fn update_block_chunks_metric(
         &mut self,
-        height: BlockHeight,
-        head_height: BlockHeight,
-        timestamp: Instant,
+        block_hash: &CryptoHash,
+        chunks: &[(ChunkHash, ShardId)],
```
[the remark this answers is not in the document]
nikurt: Same as above.

chain/client/src/chunks_delay_tracker.rs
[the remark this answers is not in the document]
mzhangmzz: That will still be true actually, the sharding change won't affect the block format. Blocks will still contain all chunk headers from all shards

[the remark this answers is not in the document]
nikurt: Thanks for the explanation. Changed as you suggested. Let's revisit this when we get to 100k chunks :)

mzhangmzz (approved):
Thanks for addressing all the previous comments! Last few comments, otherwise the PR looks great!

chain/client/src/chunks_delay_tracker.rs
```diff
@@ -1,192 +1,103 @@
...
-struct HeightInfo {
-    block_received: Option<Instant>,
-    chunks_received: HashMap<ShardId, Instant>,
+#[derive(Debug)]
+/// Records timestamps of requesting and receiving a chunk. Assumes that each chunk is requested
+/// before it is received.
+struct ChunkInProgress {
+    /// Timestamps of requesting missing chunks.
+    chunk_requested: Instant,
+    /// Timestamps of receiving each of the shards in a block.
```
mm-near, line 28: nit: you say 'timestamps' - but there is a single timestamp, right?

mm-near: and is it 'each of the shards"? as I see only a single 'option<instant>'?

chain/client/src/chunks_delay_tracker.rs
```diff
@@ -1,192 +1,103 @@
...
     }
 
-    fn update_blocks_ahead_metric(&mut self, head_height: BlockHeight) {
-        metrics::BLOCKS_AHEAD_OF_HEAD.set(self.get_blocks_ahead(head_height) as i64);
-    }
-
-    fn update_metrics(&mut self, head_height: BlockHeight) {
-        self.update_chunks_metric(head_height);
-        self.update_blocks_ahead_metric(head_height);
+    pub fn get_chunks(block: &Block) -> Vec<ChunkHash> {
```
mm-near, line 64: nit: I find it strange to see this method here (as it has nothing to do with chunks_delay_tracking) - maybe we could inline it in that place where it is used?

chain/client/src/client_actor.rs
```diff
@@ -1067,7 +1068,10 @@ impl ClientActor {
             }
 
             self.info_helper.block_processed(gas_used, chunks_in_block as u64);
+            let block_hash = block.header().hash().clone();
+            let chunks = ChunksDelayTracker::get_chunks(block);
             self.check_send_announce_account(last_final_hash);
+            self.client.record_accepted_block(&block_hash, &chunks);
```
mm-near, line 1074: on one side - I'd rather have this code directly in 'client' (rather than client actor)..

chain/client/src/chunks_delay_tracker.rs
```diff
@@ -1,192 +1,103 @@
...
-/// 2) the difference between the latest known block and the current head.
-/// If a chunk or a block is received multiple times, only the first time is recorded.
+/// Provides monitoring information about the delays of receiving and requesting blocks and their corresponding chunks.
+/// Keeps timestamps for all future blocks and chunks that were requested or received.
+/// Updates the metrics when the head gets updated. Assumes that if the head moves forward, all needed chunks were received.
+/// If a chunk or a block is received or requested multiple times, only the first time is recorded.
 #[derive(Debug, Default)]
 pub(crate) struct ChunksDelayTracker {
-    heights: BTreeMap<BlockHeight, HeightInfo>,
+    /// Timestamp of trying to process a block for the first time.
```
mm-near, line 15: and mention that the hashmap should contain only the blocks that are being processed right now.

(as the name ' blocks_received' suggests that we keep all the blocks there..
maybe blocks_in_progress would be a better name?)

mm-near (approved):

chain/client/src/chunks_delay_tracker.rs
nikurt: Done

chain/client/src/chunks_delay_tracker.rs
nikurt: Done

chain/client/src/client_actor.rs
nikurt: Moved this into `on_block_accepted()`

chain/client/src/chunks_delay_tracker.rs
nikurt: Clarified

bowenwang1996 (commented):
@nikurt how does this relate to #6431 ?

nikurt:
@bowenwang1996 IIUC, #6431 measures time spent by the node itself responding to a request.
This PR (#6381) measures time the peers take to send back the response, which includes all kinds of network latency.

PR #6383, opened 2022-03-03
doc(params-estimator): Crypto and promise cost doc

jakmeier:


runtime/runtime-params-estimator/src/cost.rs
```diff
@@ -82,14 +82,35 @@ pub enum Cost {
...
+    // Cryptographic host functions:
+    // The runtime provides host functions for sha256, keccak256, keccak512,
+    // ripemd160 hashes. Plus, there is a ECDSA signature verification host
+    // function.
+    // All these host functions are estimated by executing a transaction with a
+    // single function call in them, that just invokes the given host function.
+    // To measure the cost of additional bytes or blocks, a function call with a
+    // large argument is measured and the total cost divided by total bytes or
+    // blocks
+    /// Estimates `sha256_base`.
```
nagisa, line 95: The comment above mentions that the total cost is divided by total bytes or blocks, but it isn't clear exactly which functions are measured by a block and which ones by a byte. Might be worthwhile to add this information to each of the types?

runtime/runtime-params-estimator/src/cost.rs
```diff
@@ -198,10 +219,37 @@ pub enum Cost {
...
+    /// available from the local EpochManager.
+    ///
+    /// Estimation: Currently not estimated
     ValidatorStakeBase,
+    /// Estimates `validator_total_stake_base` which is charged for each call to
+    /// `validator_total_stake`, covering the cost for looking up the total
+    /// staked tokens for the current epoch. This information is
+    /// available from the local EpochManager.
+    ///
+    /// Estimation: Currently not estimated
```
nagisa, line 252: These seem like they ideally want to be a separate commit, but since merging will squash the commits anyway, I guess its fine to leave as is.

nagisa (approved):

runtime/runtime-params-estimator/src/cost.rs
jakmeier: The clue is in the name, `Sha256Byte` + `Keccak256Byte` + `Keccak512Byte` are measured in bytes and only `Ripemd160Block` is measured in blocks.
But I guess block is rather ambiguous in this context, at least I should probably mention that it is a compression block (as opposed to blockchain-block).

On top of that, do you think it is still worth it to add the information to each variant? I kind of would like to have a conclusive doc comment on each variant. But here it just felt a bit too repetitive to me.

nagisa: I myself try to err on the side of too much documentation than too little & I haven't yet had an opportunity to see any documentation that I thought was excessive ^^

But I definitely won't object if you decide to land the documentation here as-is.

jakmeier: You are right, more documentation definitely doesn't hurt here.

PR #6386, opened 2022-03-04
feat(params-estimator): Block overhead consistency

jakmeier:
So far, the fix block overhead has been ignored in many estimations.
This is only around 8Ggas and is usually divided by the number of
transaction used in a block. But for small costs, this is non-negible.

This change subtracts the block overhead when the per-block results
are aggreagated. This changes the behaviour of the often used helper
functions `transaction_cost`  and `fn_cost_*` variations.

As a result, this slightly changes many estimation results.
Most affected costs go down. But since some estimations are
computed as several measurments subtracted from each others,
the opposite is also possible.

There is also a correction included in how `action_receipt_creation` is
estimated. Previously, the block overhead was subtracted after dividing
by the block size, which made the effect of block overhead 100 times
larger than it should have.

### DIFF
--- a/runtime/runtime-params-estimator/src/lib.rs
+++ b/runtime/runtime-params-estimator/src/lib.rs
@@ -242,21 +242,12 @@ fn action_receipt_creation(ctx: &mut EstimatorContext) -> GasCost {
         return cached;
     }
 
-    let testbed = ctx.testbed();
-
     let mut make_transaction = |tb: &mut TransactionBuilder| -> SignedTransaction {
         let (sender, receiver) = tb.random_account_pair();
 
         tb.transaction_from_actions(sender, receiver, vec![])
     };
-    let total_cost = transaction_cost(testbed, &mut make_transaction);
-
-    // Two blocks per action receipt will be processed in the measurement, so
-    // the overhead is subtracted here
-    // TODO: Move block overhead subtraction into `Testbed::measure_blocks` for
-    // all measurements to benefit from this
-    let block_overhead_cost = apply_block_cost(ctx);
-    let cost = total_cost - block_overhead_cost * 2;
+    let cost = transaction_cost(ctx, &mut make_transaction);
 
     ctx.cached.action_receipt_creation = Some(cost.clone());
     cost
@@ -267,18 +258,13 @@ fn action_sir_receipt_creation(ctx: &mut EstimatorContext) -> GasCost {
         return cached;
     }
 
-    let testbed = ctx.testbed();
-
     let mut make_transaction = |tb: &mut TransactionBuilder| -> SignedTransaction {
         let sender = tb.random_account();
         let receiver = sender.clone();
 
         tb.transaction_from_actions(sender, receiver, vec![])
     };
-    let total_cost = transaction_cost(testbed, &mut make_transaction);
-
-    let block_overhead_cost = apply_block_cost(ctx);
-    let cost = total_cost - block_overhead_cost;
+    let cost = transaction_cost(ctx, &mut make_transaction);
 
     ctx.cached.action_sir_receipt_creation = Some(cost.clone());
     cost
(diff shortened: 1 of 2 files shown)


jakmeier:
Full list of changes based on a Icount measurement with `--warmup-iters 1 --iters 5`:

| Cost | Before | New Estimate | New/Before |
| -------| ------ | ---------- | ------------------|
| ActionAddFullAccessKey | 88_908_622_750 | 77_330_148_750 | 0.87 |
| ActionAddFunctionAccessKeyBase | 75_810_334_250 | 74_303_981_000 | 0.98 |
| ActionAddFunctionAccessKeyPerByte | 48_045_566 | 48_487_814 | 1.01 |
| ActionCreateAccount | 93_535_666_250 | 76_808_645_250 | 0.82 |
| ActionDeleteAccount | 291_724_139_750 | 283_614_677_250 | 0.97 |
| ActionDeleteKey | 45_596_674_750 | 36_264_508_500 | 0.80 |
| ActionDeployContractBase | 264_979_756_250 | 255_124_923_500 | 0.96 |
| ActionReceiptCreation | 298_175_805_000 | 321_065_233_250 | 1.08 |
| ActionSirReceiptCreation | 227_837_083_750 | 236_518_536_500 | 1.04 |
| ActionStake | 137_369_329_500 | 123_690_069_500 | 0.90 |
| ActionTransfer | 83_147_641_500 | 61_275_486_500 | 0.74 |
| ApplyBlock | 9_914_425_000 | 7_956_600_000 | 0.80 |
| ContractCompileBase | 3_944_114_843_750 | 3_995_317_343_750 | 1.01 |
| ContractCompileBaseV2 | 45_980_055_666 | 39_124_212_713 | 0.85 |
| DataReceiptCreationBase | 67_810_481_987 | 66_398_377_062 | 0.98 |
| DataReceiptCreationPerByte | 1_124_427 | 1_109_420 | 0.99 |
| GasMeteringBase | 953_875_000 | 967_000_000 | 1.01 |
| GasMeteringOp | 8_875_000 | 9_000_000 | 1.01 |
| RocksDbInsertValueByte | 28_186_446 | 28_870_869 | 1.02 |
| RocksDbReadValueByte | 66_542_592 | 64_740_145 | 0.97 |
| StorageHasKeyBase | 22_168_994_790 | 21_557_381_421 | 0.97 |
| StorageHasKeyByte | 9_855_166 | 9_970_981 | 1.01 |
| StorageReadBase | 22_274_055_715 | 25_213_444_109 | 1.13 |
| StorageReadKeyByte | 10_574_761 | 10_392_711 | 0.98 |
| StorageReadValueByte | 4_690_638 | 4_869_519 | 1.04 |
| StorageRemoveBase | 30_432_547_065 | 30_773_337_509 | 1.01 |
| StorageWriteBase | 15_854_885_665 | 15_262_250_759 | 0.96 |
| StorageWriteEvictedByte | 16_501_572 | 16_389_711 | 0.99 |
| StorageWriteValueByte | 17_285_672 | 16_772_331 | 0.97 |
| TouchingTrieNode | 6_364_015_918 | 7_203_841_174 | 1.13 |
| Utf8DecodingBase | 368_633_037 | 366_739_933 | 0.99 |

bowenwang1996 (commented):
What is our plan to update the costs?

jakmeier:
This change in particular does not affect our strategy to update costs. Today's parameter are not perfectly in line with estimations anyway and the changes from this correction is relatively small. (To be clear: The comparison above is against old estimations, not against current parameters.)


The more general strategy is to finish fixing all known issues with estimations, then compare the full list of estimations with current parameters. We should have this data by the end of the month. Then we can decide which costs need to be updated.

matklad:
I agree with the problem, but I am not 100% convinced that this is the right solution ) Most of our estimates are based on estimating the difference between measured cost and the base cost. I *think* this is a sound and robust approach, if both measured and based costs are total (ie, are not themselves computed as a subtraction). In this setup, all overheads (like block production) cancel each other. This seems to be more robust than artificially subtracting the overheads twice from both sides of the difference. 

However, our problem seems to be that the base costs are not always total -- they themselves often have *their* bases subtracted. 

So, what about alternative approach: 

* we separate total measured costs and costs that go in runtime config (might use `total_cost` for naming in code)
* we make sure that all differences are always computed between two total costs (actually, we might get fancy here and add a type for a TotalGasCost)
* we make sure that we only cache total costs

jakmeier:
Thinking about this for a while, I quite like the idea of separating total costs from what goes into the runtime configuration. A new type might also make sense. I could give it a try to refactor it this way.
About naming, maybe I would call it `GasCostMeasurement` or just `Measurement` instead of `TotalGasCost`?

However, I just remembered why I wanted to subtract the cost inside of `measure_blocks()`. The second problem  that each loop inside of `process_blocks_until_no_receipts()` adds another block overhead. Right now, we lose the information when we return measurements from `measure_blocks()`. Not sure how to cover this case cleany if we decide against subtracting it always.

PR #6387, opened 2022-03-04
feat: added feature flag to core crypto to support wasm build

russellwmy:
### Background:
There is not a good package to support `secp256k1` in WASM build.  Therefore, we should add a feature flag to disable `secp256k1` to allow developer to build without `secp256k1`.

### Changes:
- added feature flag `secp256k1`
- added `secp256k1` as default feature

### How to test:
- Run `cargo build`, it should build successfully
- Run `cargo build --no-default-features --target wasm32-unknown-unknown`, it should build successfully

### DIFF
--- a/core/crypto/Cargo.toml
+++ b/core/crypto/Cargo.toml
@@ -22,7 +22,6 @@ ed25519-dalek = "1"
 primitive-types = "0.10"
 once_cell = "1.5.2"
 libc = "0.2"
-parity-secp256k1 = "0.7"
 rand = "0.7"
 rand_core = "0.5"
 serde = { version = "1", features = [ "derive" ] }
@@ -32,11 +31,19 @@ thiserror = "1"
 near-account-id = { path = "../account-id" }
 deepsize = { version = "0.2.0", optional = true }
 
+[target.'cfg(target_arch = "wasm32")'.dependencies]
+getrandom = { version = "0.2", features=["js"]}
+
+[target.'cfg(not(target_arch = "wasm32"))'.dependencies]
+parity-secp256k1 = "0.7"
+
 [dev-dependencies]
 hex-literal = "0.2"
 sha2 = ">=0.8,<0.10"
 
 [features]
+default = ["secp256k1"]
+secp256k1 = []
 deepsize_feature = [
   "deepsize",
   "near-account-id/deepsize_feature",
--- a/core/crypto/Cargo.toml
+++ b/core/crypto/Cargo.toml
@@ -43,7 +43,6 @@ sha2 = ">=0.8,<0.10"
 
 [features]
 secp256k1 = []
-wasm_feature = []
 deepsize_feature = [
   "deepsize",
   "near-account-id/deepsize_feature",
--- a/core/crypto/Cargo.toml
+++ b/core/crypto/Cargo.toml
@@ -42,8 +42,8 @@ hex-literal = "0.2"
 sha2 = ">=0.8,<0.10"
 
 [features]
-default = ["secp256k1"]
 secp256k1 = []
+wasm_feature = []
 deepsize_feature = [
   "deepsize",
   "near-account-id/deepsize_feature",
--- a/core/crypto/src/signature.rs
+++ b/core/crypto/src/signature.rs
@@ -7,12 +7,14 @@ use std::str::FromStr;
 
 use borsh::{BorshDeserialize, BorshSerialize};
 use ed25519_dalek::ed25519::signature::{Signature as _, Signer, Verifier};
+#[cfg(feature = "secp256k1")]
 use once_cell::sync::Lazy;
 use primitive_types::U256;
 use rand_core::OsRng;
+#[cfg(feature = "secp256k1")]
 use secp256k1::Message;
 use serde::{Deserialize, Serialize};
-
+#[cfg(feature = "secp256k1")]
 pub static SECP256K1: Lazy<secp256k1::Secp256k1> = Lazy::new(secp256k1::Secp256k1::new);
 
 #[derive(Debug, Copy, Clone, Serialize, Deserialize)]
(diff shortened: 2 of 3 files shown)


matklad:
If we want to do this in the future, one question I would have is *why* do we want this crate to be compatible with wasm

russellwmy:
### My use case:
I am working with yew.rs to build Dapp. However, there is only `near-api-js` for a web app to interact with the chain. So I have to build a Rust version of api sdk to enable yew.rs to interact with the chain.

### For future use case:
It enables developers can build web Dapp in Rust.

## My motivation:
I build product in one programming language. So my team can manage the product from frontend to backend. It is important to me to enable scaling team smoothly. Since everyone work in Rust.

I am not sure if it is the way near will do. I will keep the logic in my repository first.

russellwmy:
@matklad  if you are interested  to the SDK, you can check https://github.com/russellwmy/near-api-rs. But it is still WIP.

austinabell:
> ### My use case:
> I am working with yew.rs to build Dapp. However, there is only `near-api-js` for a web app to interact with the chain. So I have to build a Rust version of api sdk to enable yew.rs to interact with the chain.
> 
> ### For future use case:
> It enables developers can build web Dapp in Rust.
> 
> ### My motivation:
> I build product in one programming language. So my team can manage the product from frontend to backend. It is important to me to enable scaling team smoothly. Since everyone work in Rust.
> 
> I am not sure if it is the way near will do. I will keep the logic in my repository first.

Can you be more specific about what you are looking for? Is there some functionality that is missing from the secp/ed libraries in Rust, or is it just that you'd like a library that handles the different key types together with a shared API?



> if you are interested to the SDK, you can check [russellwmy/near-api-rs](https://github.com/russellwmy/near-api-rs). But it is still WIP.

Unlikely it would come into the SDK, since it sounds like this would not be used within a contract. Some library could potentially be built out to expose a simple API that handles NEAR supported crypto operations, which could be used in the SDK or any other Rust tooling lib.

russellwmy:
@austinabell 
> Can you be more specific about what you are looking for? Is there some functionality that is missing from the secp/ed libraries in Rust, or is it just that you'd like a library that handles the different key types together with a shared API?
I am looking for a Rust SDK can run in browser. So my application can interact with Near platform in browser. Which my application frontend is built with Rust framework yew.rs. I only find  Rust SDK for interact with smart contract.
> Unlikely it would come into the SDK, since it sounds like this would not be used within a contract. Some library could potentially be built out to expose a simple API that handles NEAR supported crypto operations, which could be used in the SDK or any other Rust tooling lib.
That would be good to have Rust SDK to do that. I did research for awhile. I cannot find one.
> Edit: it seems like you just wanted ed key support, but would it be beneficial to have secp also included with a wasm compatible package? If only using ED25519, I'm missing the motivation for using the nearcore internal crate
The internal crate using https://crates.io/crates/secp256k1. I try to build it but it seems it cannot compile into wasm. Maybe you have idea to make it works. But I noticed `near-api-js` also does not implement that `key_type`. So I add feature flag to disable until there is a method to build or find replacement.

austinabell:
> I am looking for a Rust SDK can run in browser. So my application can interact with Near platform in browser. Which my application frontend is built with Rust framework yew.rs. I only find Rust SDK for interact with smart contract.

You can use https://github.com/near/near-jsonrpc-client-rs (cc @miraclx) or even https://github.com/near/workspaces-rs/ (although this is primarily just used as a testing framework for now)

The Rust SDK is for building smart contracts, not for interacting with them

> 
> > Unlikely it would come into the SDK, since it sounds like this would not be used within a contract. Some library could potentially be built out to expose a simple API that handles NEAR supported crypto operations, which could be used in the SDK or any other Rust tooling lib.
> 
> That would be good to have Rust SDK to do that. I did research for awhile. I cannot find one.
> 
> > Edit: it seems like you just wanted ed key support, but would it be beneficial to have secp also included with a wasm compatible package? If only using ED25519, I'm missing the motivation for using the nearcore internal crate
> 
> The internal crate using [crates.io/crates/secp256k1](https://crates.io/crates/secp256k1). I try to build it but it seems it cannot compile into wasm. Maybe you have idea to make it works. But I noticed `near-api-js` also does not implement that `key_type`. So I add feature flag to disable until there is a method to build or find replacement.

But you could just use a ed25519 library that compiles to wasm? What comes with the nearcore crate that doesn't exist there? A simpler API?

russellwmy:
>You can use https://github.com/near/near-jsonrpc-client-rs (cc @miraclx) or even https://github.com/near/workspaces-rs/ (although this is primarily just used as a testing framework for now)
The Rust SDK is for building smart contracts, not for interacting with them
Screenshot 2022-03-10 at 7 52 57 AM

> But you could just use a ed25519 library that compiles to wasm? What comes with the nearcore crate that doesn't exist there? A simpler API?

In that case, I can do that, just create another things to the same things that nearcore provides. But I am thinking I am reinventing the wheel. Btw, I also think `core/primitives` and `jsonrpc-primitives` should be  the package with less dependencies and allow other developer to develop thing base on that, cause they are core types.

matklad:
Drive by, but I've learned today that https://docs.rs/ed25519-compact/latest/ed25519_compact/ exists, and, from the cursory look, it might be better suited for wasm than `ed25519-dalek` which we use for nearcore.

russellwmy:
@matklad Yes, we can replace the library. But we should replace `parity-secp256k1 ` with `libsecp256k1` for `specp256k1`. it is a blocker for compiling the `crypto-core` in WASM. `parity-secp256k1` had not update for 2 years.
 And we need to review the version of other libraries. I found number of libraries are out-dated.

russellwmy:
> Drive by, but I've learned today that https://docs.rs/ed25519-compact/latest/ed25519_compact/ exists, and, from the cursory look, it might be better suited for wasm than `ed25519-dalek` which we use for nearcore.

I just check their code. It is cool, small and less dependencies. We can check what is  the security level first if we want to employ to be nearcore.

Issue #6186, opened 2022-01-25
Suspicious unwrap in Account's deserialize


matklad:
The following code looks somewhat suspicious to me: 

https://github.com/nearprotocol/nearcore/blob/47620d3ec7cdd0d9c30d981d537234ffeb8c0b47/core/primitives-core/src/account.rs#L114-L130

`Borsh` generally doesn't have a contract that the input is valid, and I think we should return a borsh error in this case. 

That code is reachable from 

https://github.com/nearprotocol/nearcore/blob/47620d3ec7cdd0d9c30d981d537234ffeb8c0b47/core/store/src/lib.rs#L334-L340

which converts borsh serialization errors to inconsistent state errors, which sounds like the right logic here.

matklad:
@EgorKulikov could you take a look when you have time?

EgorKulikov:
Yep, I agree

Ekleog:
Overall one thing that bothers me with this assertion is that borsh appears to have as an expectation that BorshDeserialize will advance the buffer by the number of bytes consumed, which sounds to me like it could always pass in a slice bigger than the actual data to deserialize. So I'm not sure even returning an error would actually be implementing the BorshDeserialize as specified.

But then I'm having doubts, because in practice it seems like the code we currently have, that panics on error, does work, so… maybe I'm misunderstanding the BorshDeserialize contract?

frol:
Should this issue be closed as #6388 was merged?

matklad:
No, the code still violates the contract of BorshDeserialize trait, by panicing on slices of invalid lenght

frol:
I would just turn the panic into a warning log. WDYT?

Ekleog:
I'm not sure it'd be much better than the statu quo. The reason for removing the panic is (IMO) mostly to simplify code, as if it were actually reachable there's a good chance that it'd already have triggered. Plus primitives-core doesn't depend on any logging crate yet, so I'm having doubts warn-logging would be of much help here.

That said, it'd be different if we actually thought there's a chance that the panic actually triggers, in which case we would have to think about whether this is:
- expected behavior (which I think, as it's in the documented BorshDeserialize API), in which case warn-logging would be surprising for the end-user as there's actually nothing wrong to warn about
- unexpected behavior, in which case we may have just deserialized an invalid account and crashing might be a better choice than continuing with broken data

To sum up, the decision tree as I see it:
- Do we expect it to ever trigger?
[...]

PR #6388, opened 2022-03-04
fix: fix account deserialization on arm64 while keeping the behavior of panicking if deserializing too big a slice

### LINKED ISSUE #6293: Account deserialization fails on ARM64
**Describe the bug**

The versioning check for Account type deserialization is invalid since it compares the size of the buffer holding serialized data (72 bytes) with a memory layout size of the struct (platform-dependent due to alignment and it seems to be 72 bytes on amd64, while it is 80 bytes on arm64 [Apple M1]) instead of the size of the serialized struct.

https://github.com/near/nearcore/blob/5568df666a0f5025c724027814069050acb194c6/core/primitives-core/src/account.rs#L113-L130

cc @EgorKulikov @mm-near 

**To Reproduce**

1. Compile neard for Apple M1 (#6200 is already on master)
2. Init neard: `./target/debug/neard --home /tmp/near-home init`
3. Run neard: `./target/debug/neard --home /tmp/near-home run`

**Expected behavior**

It should deserialize the value just fine, and it does when I just remove the condition and let it deserialize the LegacyAccount struct.

image

- nearcore commit/branch: master 5568df666a0f5025c724027814069050acb194c6
- rust version (if local): 1.58.1
- docker (if using docker): no
- mainnet/testnet/betanet/local

Ekleog:
Fixes #6293 

Note I'm not sure why we were panicking upon invalid buffer length before: it looks to me like something that was left over from an experiment of adding a second version of account serialization and no longer makes sense, but I'm keeping it for now until @EgorKulikov comments about its use.

I tried building and running the two neard command lines listed in #6293 on both arm64 and amd64 and they succeeded properly.

cc @frol @EgorKulikov @bowenwang1996 @mm-near

### DIFF
--- a/core/primitives-core/src/account.rs
+++ b/core/primitives-core/src/account.rs
@@ -112,20 +112,19 @@ struct LegacyAccount {
 
 impl BorshDeserialize for Account {
     fn deserialize(buf: &mut &[u8]) -> Result<Self, io::Error> {
-        if buf.len() == std::mem::size_of::<LegacyAccount>() {
-            // This should only ever happen if we have pre-transition account serialized in state
-            // See test_account_size
-            let deserialized_account = LegacyAccount::deserialize(buf)?;
-            Ok(Account {
-                amount: deserialized_account.amount,
-                locked: deserialized_account.locked,
-                code_hash: deserialized_account.code_hash,
-                storage_usage: deserialized_account.storage_usage,
-                version: AccountVersion::V1,
-            })
-        } else {
-            unreachable!();
+        // This should only ever happen if we have pre-transition account serialized in state
+        // See test_account_size
+        let deserialized_account = LegacyAccount::deserialize(buf)?;
+        if buf.len() != 0 {
+            panic!("Tried deserializing a buffer that is not exactly the size of an account");
         }
+        Ok(Account {
+            amount: deserialized_account.amount,
+            locked: deserialized_account.locked,
+            code_hash: deserialized_account.code_hash,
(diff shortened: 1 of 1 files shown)


matklad:
cc https://github.com/near/nearcore/issues/6186 for the panic issue.

Ekleog:
Oh I hadn't seen that issue, thank you for the link!

core/primitives-core/src/account.rs
```diff
@@ -112,20 +112,19 @@ struct LegacyAccount {
 
 impl BorshDeserialize for Account {
     fn deserialize(buf: &mut &[u8]) -> Result<Self, io::Error> {
-        if buf.len() == std::mem::size_of::<LegacyAccount>() {
```
matklad, line 115: I'd just hard-coded the constant here, deserializing LegacyAccount every time feels like it can affect performance negatively. We can than have a test which compares the constant with the size of serialized account.

matklad (commented):
Out of curiosity, does grepping for size_of turns out more patterns like this?

core/primitives-core/src/account.rs
matklad: Ah yes, you are completely right!

matklad (approved):

Ekleog:
I could not find other patterns similar to this, the closest one I could find was https://github.com/near/nearcore/blob/master/core/primitives-core/src/hash.rs#L67 ; which doesn't suffer from the same issue as it's checking for the length only after a successful deserialization :)

PR #6389, opened 2022-03-04
feat: state_viewer: add an apply_chunk command

marcelo-gonzalez:
This command applies txs and receipts in a chunk even if we don't have
a block including this chunk on disk yet. Requires the chunk's
prev_block to be available. You can specify a list of tx or
receipt hashes to only apply those particular txs or receipts:
    
$ ./target/debug/neard view_state apply_chunk --chunk-hash <hash> \
  [--txs <tx1> <tx2> ...] [--receipts <receipt1> <receipt2 ...]


tools/state-viewer/src/commands.rs
```diff
@@ -649,3 +660,118 @@ pub fn format_hash(h: CryptoHash) -> String {
 pub fn chunk_mask_to_str(mask: &[bool]) -> String {
     mask.iter().map(|f| if *f { '.' } else { 'X' }).collect()
 }
+
+fn filter_hashes<T: Clone, F: Fn(&T) -> CryptoHash>(
+    items: &Vec<T>,
```
bowenwang1996, line 665: why not `&[T]`

tools/state-viewer/src/commands.rs
```diff
@@ -649,3 +660,118 @@ pub fn format_hash(h: CryptoHash) -> String {
...
+    ));
+    let chunk = chain_store.get_chunk(&chunk_hash)?;
+    let chunk_header = chunk.cloned_header();
+
+    let prev_block_hash = chunk_header.prev_block_hash();
+    let shard_id = chunk.shard_id();
+    let prev_state_root = chunk.prev_state_root();
+
+    let transactions = filter_txs(chunk, tx_hashes)?;
+    let receipts = filter_receipts(chunk, receipt_hashes)?;
```
bowenwang1996, line 740: Actually this is not what we want because we would like to reproduce the exact same output as if it is applied in production. What we need here is to give transaction/receipt hash as input and then find the right chunk to apply and filter the outcome for that specific transaction or receipt.

Also receipts in the chunk body are outgoing receipts, not incoming receipts. We need incoming receipts for a shard when applying a chunk. You can take a look at https://github.com/near/nearcore/blob/d305aac67b07daf5c168f6b5644fcf58edb1faa9/chain/chain/src/chain.rs#L3521 to see how it is implemented

bowenwang1996 (changes requested):

tools/state-viewer/src/commands.rs
marcelo-gonzalez: oh good catch... don't know how I missed that. Updated it to try and mimic `save_incoming_receipts_from_block()` by looking in `ColChunkHashesByHeight`

Ok, so I misunderstood what the `--tx` filter here should do. Updated to print out the apply outcomes for any tx or receipt hashes given on the command line. did you mean something like that?

marcelo-gonzalez: btw, I checked that resulting `state_root` is the same as with the `apply` cmd, but the `outcome_root` won't be the same because of the local receipts that hve different hashes.

marcelo-gonzalez: ah also forgot to mention, this last commit still asks for a chunk hash on the command line. I'm trying to think of a good command line interface that would leave it out. With the current `--txs` and `--receipts` flags, you would need to iterate over different chunks (I guess the ones included in `ColChunkHashesByHeight[--target-height]`) and look for one containing your transaction. Then it's a little hairy when you ask what happens if there are two txs given in `--txs` that appear in two different chunks at that height. Do we apply both chunks? or fail? Seems simpler to just ask for a chunk hash to me, since you probably can find the chunk hash you're looking for if you're in a position where you're using this command, but could be wrong about that

tools/state-viewer/src/commands.rs
```diff
@@ -649,3 +659,168 @@ pub fn format_hash(h: CryptoHash) -> String {
...
+                    return Err(*hash);
+                }
+            }
+            Ok(())
+        }
+        None => Ok(()),
+    }
+}
+
+pub(crate) fn apply_chunk(
```
nikurt, line 748: The new command requires quite a lot of logic, please consider moving it to a separate file.
And if you move it to a separate file, would it be possible to add a simple unit test?

marcelo-gonzalez: moved and added test

nikurt (approved):

bowenwang1996 (approved):
I think it is fine as is, but I'd like to explorer the possibility of passing in transaction or receipt hash directly. I think this is doable:
* If the transaction or receipt is contained in some block we can look up which block it is in from `ColTransactionResult`
* Otherwise it is not in any block yet. We can search for all the unapplied chunks and apply all of the chunks that contain the transaction/receipt. There may be multiple but we can just print out all of the them

Issue #1808, opened 2019-12-02
Remove dead code


nearmax:
Some of our integration tests contain `#[allow(dead_code)]`. We should examine whether this code is actually not used and delete it if we are not planning to use it.

nearmax:
Moving to Ice Box since it does not affect readability, performance, safety or DevX.

PR #6390, opened 2022-03-06
chore(tests): remove dead code

kakoc:
#1808

### DIFF
--- a/chain/epoch_manager/src/lib.rs
+++ b/chain/epoch_manager/src/lib.rs
@@ -38,6 +38,7 @@ mod reward_calculator;
 #[cfg(feature = "protocol_feature_chunk_only_producers")]
 mod shard_assignment;
 pub mod test_utils;
+#[cfg(test)]
 mod tests;
 mod types;
 mod validator_selection;
--- a/chain/epoch_manager/src/tests/random_epochs.rs
+++ b/chain/epoch_manager/src/tests/random_epochs.rs
@@ -21,35 +21,6 @@ fn test_random_epochs() {
     run_with_seed_range(2, 10, 1000);
 }
 
-#[allow(dead_code)]
-fn minimal_repro(epoch_length: u64, num_heights: u64, max_seed: u64) {
-    let mut epoch_length = epoch_length;
-    let mut num_heights = num_heights;
-    if !std::panic::catch_unwind(move || run_with_seed_range(epoch_length, num_heights, max_seed))
-        .is_err()
-    {
-        println!("PASS");
-        return;
-    }
-    while epoch_length > 1
-        && std::panic::catch_unwind(move || {
-            run_with_seed_range(epoch_length - 1, num_heights, max_seed)
-        })
-        .is_err()
-    {
-        epoch_length -= 1;
-    }
-    while num_heights > 1
-        && std::panic::catch_unwind(move || {
-            run_with_seed_range(epoch_length, num_heights * 4 / 5, max_seed)
-        })
-        .is_err()
-    {
-        num_heights = num_heights * 4 / 5;
-    }
-    panic!("Found! run_with_seed_range({}, {}, {})", epoch_length, num_heights, max_seed);
-}
-
 fn run_with_seed_range(epoch_length: u64, num_heights: u64, max_seed: u64) {
     for test_seed in 0..max_seed {
         if let Err(e) =
--- a/runtime/runtime-params-estimator/src/testbed.rs
+++ b/runtime/runtime-params-estimator/src/testbed.rs
@@ -15,8 +15,7 @@ use std::sync::Arc;
 
 pub struct RuntimeTestbed {
     /// Directory where we temporarily keep the storage.
-    #[allow(dead_code)]
-    pub(crate) workdir: tempfile::TempDir,
+    _workdir: tempfile::TempDir,
     tries: ShardTries,
     root: MerkleHash,
     runtime: Runtime,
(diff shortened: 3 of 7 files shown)


chain/epoch_manager/src/tests/random_epochs.rs
```diff
@@ -1,418 +0,0 @@
...
-use near_primitives::epoch_manager::block_info::BlockInfo;
-use near_primitives::epoch_manager::epoch_info::EpochInfo;
-use near_primitives::epoch_manager::SlashState;
-use near_primitives::hash::CryptoHash;
-use near_primitives::types::validator_stake::ValidatorStake;
-use near_primitives::types::{AccountId, Balance, EpochId, ValidatorKickoutReason};
-
-const DEBUG_PRINT: bool = false;
-
-#[test]
```
matklad, line 19: This seems to be a real test we are deleting, no?

runtime/runtime-params-estimator/src/testbed.rs
```diff
@@ -15,8 +15,6 @@ use std::sync::Arc;
 
 pub struct RuntimeTestbed {
     /// Directory where we temporarily keep the storage.
-    #[allow(dead_code)]
-    pub(crate) workdir: tempfile::TempDir,
```
matklad, line 19: This needs to be here to make sure it's not dropped prematurely (dropping `TempDir` value deletes the underlying temporary directory). I suggest the following change insteed:

```
_workdir: tempfile::TempDir,
```

* remove `pub`
* use `_` rather than allow to silence the warning. The `_` prefix is idiomatic way to signal "this is here only for drops".

chain/epoch_manager/src/tests/random_epochs.rs
kakoc: Yep, are right. Was a little bit confused since `cargo check` showed me that there was an unused code. Placed all that staff under a tests module.

runtime/runtime-params-estimator/src/testbed.rs
kakoc: Changed.
Thanks for the review.

chain/epoch_manager/src/lib.rs
```diff
@@ -38,7 +38,6 @@ mod reward_calculator;
 #[cfg(feature = "protocol_feature_chunk_only_producers")]
 mod shard_assignment;
 pub mod test_utils;
-mod tests;
```
matklad, line 41: Curious, dose `#[cfg(test)]` work here? If it does, that would be better than introducing the nested test module.

kakoc: Am I right that your proposal is trying to move that tests module from https://github.com/near/nearcore/pull/6390/files#diff-d53b6e2ee18a5c4ef85c47a6f020198cfb91d4e32d96ec2b799e412effc3c606R1 and place it near thttps://github.com/near/nearcore/blob/622e4439097a36856d1bf9839f0dac9b071282b8/chain/epoch_manager/src/lib.rs#L1498 ?

matklad: Oh, this is a bit confusing, let me elaborate!

So, in the current state of the PR, you have this `#[cfg(test)]` module, but it feels that the module *itself* is already inside test-only code, so `#[cfg(test)]` should be pulled up, and guard the whole thing.

It seems like the `mod tests;` declaration in the `src/lib.rs` is the place where we need to add a cfg.

Taking a second look here, it seems that, in the current state, the code is dead because we miss `mod tests;` altogether.

This is what I think should work:

```diff
diff --git a/chain/epoch_manager/src/lib.rs b/chain/epoch_manager/src/lib.rs
index eeb299d00..5e70221ce 100644
--- a/chain/epoch_manager/src/lib.rs
+++ b/chain/epoch_manager/src/lib.rs
@@ -41,6 +41,9 @@ pub mod test_utils;
 mod types;
 mod validator_selection;
 
+#[cfg(test)]
+mod tests;
+
 const EPOCH_CACHE_SIZE: usize = if cfg!(feature = "no_cache") { 1 } else { 50 };
 const BLOCK_CACHE_SIZE: usize = if cfg!(feature = "no_cache") { 5 } else { 1000 }; // TODO(#5080): fix this
 const AGGREGATOR_SAVE_PERIOD: u64 = 1000;
diff --git a/chain/epoch_manager/src/tests/random_epochs.rs b/chain/epoch_manager/src/tests/random_epochs.rs
index 9b9a4f5ff..7b5dfe303 100644
--- a/chain/epoch_manager/src/tests/random_epochs.rs
+++ b/chain/epoch_manager/src/tests/random_epochs.rs
@@ -1,5 +1,5 @@
-#[cfg(test)]
-mod test {
+// #[cfg(test)]
+// mod test {
     use rand::prelude::StdRng;
     use rand::{Rng, SeedableRng};
     use std::collections::{BTreeMap, BTreeSet, HashMap};
@@ -406,4 +406,4 @@ mod test {
             .map(|stake| stake.account_and_stake())
             .collect::<HashMap<_, _>>()
     }
-}
+// }
```

kakoc: Yes, I really missed an idea about marking the whole module as tests module on the top level.
Thanks.

matklad (approved):
LGTM now, thanks for a good cleanup @kakoc !

mm-near (approved):

mm-near:
Thanks for contributing @kakoc

PR #6391, opened 2022-03-07
fuzz: fight fuzzer false positives by attempting to reduce the allowed memory

Ekleog:
Context: grafana for at least two of the last spurious crashes since switching to measuring CPU time shows a memory peak of ~4G just around the time of the crash report.

I'm thinking if one iteration of one of these fuzzers takes more than 1G there's probably something wrong, so might as well lower the limit.

I'm also seeing to integrate the cAdvisor exporter in order to get cgroup-specific metrics, which should help identify the root cause of the memory peak in case it was not actually one fuzzer causing the issues.

### DIFF
--- a/nightly/fuzz.toml
+++ b/nightly/fuzz.toml
@@ -2,34 +2,34 @@
 name = "master"
 weight = 10
 
-[[branch]]
-name = "1.25.0"
-weight = 10
-
-[[branch]]
-name = "1.24.0"
-weight = 10
+#[[branch]]
+#name = "1.25.0"
+#weight = 10
+#
+#[[branch]]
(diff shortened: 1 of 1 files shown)


Ekleog:
I'm also taking advantage of this PR to comment out 1.24.0 and 1.25.0 for now, because they're subject to the false positive issue and this'll help me double down on figuring out where this false positive is coming from and how to eliminate it on a branch that has all the latest fixes, as it's the biggest remaining thing that's preventing me from declaring the fuzzer ready for wider adoption.

nagisa (approved):

mina86 (approved):

PR #6396, opened 2022-03-07
add support for start_height in start_mock_network

mzhangmzz:
Add an option for `start_mock_network` to start the client at a given height instead of from genesis. This is useful if we want to replay mainnet history using `start_mock_network`.


chain/chain/src/store.rs
```diff
@@ -2465,6 +2466,103 @@ impl<'a> ChainStoreUpdate<'a> {
         Ok(())
     }
 
+    /// Only used in mock network
+    /// Copy the necessary chain state related to `block_hash` from `source_store` to the current
+    /// store.
+    pub fn copy_chain_state_as_of_block(
```
bowenwang1996, line 2472: why do we need to update all the caches?

chain/epoch_manager/src/lib.rs
```diff
@@ -142,6 +142,62 @@ impl EpochManager {
         Ok(epoch_manager)
     }
 
+    /// Only used in mock network
```
bowenwang1996, line 145: if that is the case it is better to guard this function with some feature flag

chain/chain/src/store.rs
```diff
@@ -2465,6 +2466,103 @@ impl<'a> ChainStoreUpdate<'a> {
         Ok(())
     }
 
+    /// Only used in mock network
```
EgorKulikov, line 2469: maybe hide it behind feature then?

chain/chain/src/store.rs
mzhangmzz: It's not for updating the cache, eventually these changes to will be commit to the database

chain/epoch_manager/src/lib.rs
mzhangmzz: Done

chain/chain/src/store.rs
mzhangmzz: Done

chain/chain/src/chain.rs
```diff
@@ -3699,6 +3700,10 @@ impl<'a> ChainUpdate<'a> {
                         ))));
                     }
 
+                    // if we are running mock_network, ignore this check because
+                    // this check may require old block headers, which may not exist in storage
+                    // of the client in the mock network
```
bowenwang1996, line 3705: I am confused. Why do we special case this protocol feature?

integration-tests/src/bin/start_mock_network.rs
```diff
@@ -19,15 +19,30 @@ use std::time::Duration;
...
+/// In NoNewBlocks mode, the mock network will not simulate block production.
+/// The simulated peers will start from the target height and no new blocks are produced.
 ///
-/// In NoSync mode, the client and mock network start at the same height and new blocks will
-/// be produced. They both start from genesis height and the mock network will produce
-/// blocks until target height.
+/// In ProduceNewBlocks mode, new blocks will be produced, i.e., the client will receive new
+/// blocks from the mock network. User can provide a number in this mode, which will be the starting
+/// height the new produced blocks. If no number is provided, the mock network will start at
+/// the same height as the client, which is specified by client_start_height.
```
bowenwang1996, line 30: What are in those blocks? Are they empty or filled with transactions? If they are not empty how are the blocks produced?

integration-tests/src/mock_network/setup.rs
```diff
@@ -97,6 +128,124 @@ pub fn setup_mock_network(
...
+            .copy_epoch_info_as_of_block(
+                &hash,
+                header.epoch_id(),
+                header.next_epoch_id(),
+                &mut mock_epoch_manager,
+            )
+            .unwrap();
+        info!(target:"mock_network", "Done preparing epoch info");
+
+        // copy state for all shards
```
bowenwang1996, line 200: Wouldn't this potentially take a very long time?

integration-tests/src/bin/start_mock_network.rs
mzhangmzz: These are from an existing storage. The mock network takes an existing data dir from a node. It reads the chain history from that source data dir and simulate the generation of new blocks by feeding these blocks to the client.

integration-tests/src/mock_network/setup.rs
mzhangmzz: Yes, but that's unavoidable. I generally run this once to set up the directory and then just use the same directory to avoid copying state again

chain/chain/src/store.rs
```diff
@@ -2465,6 +2467,104 @@ impl<'a> ChainStoreUpdate<'a> {
...
+    /// Copy the necessary chain state related to `block_hash` from `source_store` to the current
+    /// store.
+    #[cfg(feature = "mock_network")]
```
mzhangmzz, line 2473: This is a hack to make the binary compile without the feature. I can't add this feature as a default feature to integration-tests, because that breaks some tests in integration-tests.

I plan to move mock_network outside of integration-tests to be its own package so I can set mock_network as default feature for the package, but I want to do that in another PR to make review easier

integration-tests/src/mock_network/README.md
```diff
@@ -12,4 +12,54 @@ components of the mock network.
...
+```
+To replay mainnet history, pass a home dir of a mainnet node as chain-history-home-dir to the commmand.
+The command also supports specifying a start_height for the client. If `client-start-height` is specified,
+the binary will set up the data dir before starting the client, by copying the state at the specified height
+and other chain info necessary for processing the blocks afterwards (such as block headers and blocks).
+
+Note that the start height must be the last block of an epoch.
+
+The initial setup for client dir takes a long time, so we suggest specifying a client dir
+so you can reuse it again without having to copy the state again. If you run the above command,
```
mm-near, line 57: maybe we could add concrete instructions on how to do it?
(as I guess people should 'snapshot' the directory then, right?)

Also - maybe mention what does 'long time' mean - for example for mainnet ? 10 minutes? 8 hours?

integration-tests/src/mock_network/setup.rs
```diff
@@ -97,6 +141,131 @@ pub fn setup_mock_network(
...
+            network_tail_height,
+            network_head_height
+        );
+        let hash = network_chain_store.get_block_hash_by_height(start_height).unwrap();
+        if !mock_network_runtime.is_next_block_epoch_start(&hash).unwrap() {
+            panic!("start height must be the last block of an epoch");
+        }
+
+        // copy chain info
+        #[cfg(feature = "mock_network")]
```
mm-near, line 181: shouldn't the whole setup_mock_network be under 'mock_network' feature?

chain/chain/src/store.rs
```diff
@@ -2465,6 +2467,104 @@ impl<'a> ChainStoreUpdate<'a> {
...
+    pub fn copy_chain_state_as_of_block(
```
mm-near, line 2474: Could we change it to constructor instead?  (so that it returns ChainStoreUpdate?)

(as using it on already partially created ChainStoreUpdate object might not be a good idea)

chain/epoch_manager/src/lib.rs
```diff
@@ -142,6 +142,63 @@ impl EpochManager {
...
+    /// Copy the necessary epoch info related to `block_hash` from `source_epoch_manager` to
+    /// the current epoch manager.
+    /// Note that this function doesn't copy info stored in EpochInfoAggregator, so `block_hash` must be
+    /// the last block in an epoch in order for the epoch manager to work properly after this function
```
mm-near, line 149: Do we check that it is the last block of the epoch? what kind of error would happen if it is not?

chain/epoch_manager/src/lib.rs
```diff
@@ -142,6 +142,63 @@ impl EpochManager {
...
+            epoch_id,
+            source_epoch_manager.get_epoch_info(epoch_id)?.clone(),
+        )?;
+        // save next epoch info too
+        self.save_epoch_info(
+            &mut store_update,
+            next_epoch_id,
+            source_epoch_manager.get_epoch_info(next_epoch_id)?.clone(),
+        )?;
+        // save next next epoch info if the block is the last block
```
mm-near, line 171: wait - but according to the comment - this should always be true, right? (it should always be  last block)

chain/epoch_manager/src/lib.rs
```diff
@@ -142,6 +142,63 @@ impl EpochManager {
...
+            let next_next_epoch_id =
+                source_epoch_manager.get_next_epoch_id_from_prev_block(block_hash)?;
+            self.save_epoch_info(
+                &mut store_update,
+                &next_next_epoch_id,
+                source_epoch_manager.get_epoch_info(&next_next_epoch_id)?.clone(),
+            )?;
+        }
+
+        let block_info = source_epoch_manager.get_block_info(block_hash)?.clone();
```
mm-near, line 182: as we're getting block info from the source_epoch_manager, could we also get the epoch_id and next_epoch_id from there?
(we'd avoid 2 function parameters)

integration-tests/src/mock_network/setup.rs
```diff
@@ -97,6 +141,131 @@ pub fn setup_mock_network(
...
+                    info!(
+                        target: "mock_network",
+                        "Done {}/{} parts",
+                        finished_parts_count_num + 1,
+                        num_parts
+                    );
+                }));
+            }
+            for handler in handlers {
+                handler.join().expect("Problem in copying state parts");
```
mm-near, line 264: consider also printing part_id  so that we know which part caused trouble.

integration-tests/src/mock_network/setup.rs
```diff
@@ -97,6 +141,131 @@ pub fn setup_mock_network(
...
+        let next_block = network_chain_store.get_block(&next_hash).unwrap();
+        for (shard_id, chunk_header) in next_block.chunks().iter().enumerate() {
+            info!(target:"mock_network", "Preparing state for shard {}", shard_id);
+            let shard_id = shard_id as u64;
+            let state_root = chunk_header.prev_state_root();
+            let state_root_node =
+                mock_network_runtime.get_state_root_node(shard_id, &hash, &state_root).unwrap();
+            let num_parts = get_num_state_parts(state_root_node.memory_usage);
+            let mut handlers = vec![];
+            let finished_parts_count = Arc::new(RwLock::new(0));
```
mm-near, line 231: use AtomicUsize instead of RwLock -- then you can use fetch_add to increment

integration-tests/src/mock_network/setup.rs
```diff
@@ -97,6 +141,131 @@ pub fn setup_mock_network(
...
+                            num_parts,
+                            &state_part,
+                            &mock_network_runtime1.get_epoch_id_from_prev_block(&hash1).unwrap(),
+                        )
+                        .unwrap();
+                    let finished_parts_count_num = *finished_parts_count1.read().expect("");
+                    *finished_parts_count1.write().expect("") = finished_parts_count_num + 1;
+                    info!(
+                        target: "mock_network",
+                        "Done {}/{} parts",
```
mm-near, line 257: I'd also mention the shard_id

mm-near (approved):

integration-tests/src/mock_network/setup.rs
mzhangmzz: yes, but I was having some issues that I don't remember now.

I plan to move the mock_network crate outside of integration-tests. That will make the feature setup cleaner.

chain/epoch_manager/src/lib.rs
mzhangmzz: Yes, I might change the code in the future to get rid of the requirement that it has to be the last block of the epoch

chain/epoch_manager/src/lib.rs
mzhangmzz: We don't check it in this function but we check it in setup_mock_network

PR #6397, opened 2022-03-07
feat(params-estimator): Increase deployment cost

jakmeier:
Set the deployment cost per byte to the maximum that still allows
deploying a 4MiB contract. This is to cover compilation costs.

Test plan
---------
A new integration test checks that a 4MiB can still be deployed.


jakmeier:
This is a protocol change. Since this is the first one I submit, I hope I got the technicalities for this right. Please let me know otherwise.

For now I updated the cost for version 52 but unless it is included in the next release, I will have to update this PR and create a new cost json file.

core/primitives/res/runtime_configs/52.json
```diff
@@ -32,7 +32,7 @@
       "deploy_contract_cost_per_byte": {
         "send_sir": 6812999,
         "send_not_sir": 6812999,
-        "execution": 6812999
+        "execution": 64572944
```
mm-near, line 35: I think that you might have to create version 53.. as 52 (AFAIK) was already pushed to testnet - but you can check with @mina86 who is doing that release.

integration-tests/src/tests/runtime/deployment.rs
```diff
@@ -0,0 +1,69 @@
...
+    assert_eq!(transaction_result.receipts_outcome.len(), 2);
+
+    // Deploy contract
+    let wasm_binary = near_test_contracts::sized_contract(contract_size as usize);
+    let transaction_result =
+        node_user.deploy_contract(test_contract_id, wasm_binary.to_vec()).unwrap();
+    assert_eq!(transaction_result.status, FinalExecutionStatus::SuccessValue(to_base64(&[])));
+    assert_eq!(transaction_result.receipts_outcome.len(), 1);
+    assert!(
+        transaction_result.receipts_outcome[0].outcome.gas_burnt
```
mm-near, line 66: how much gas does it burn now? how much is the headroom?

mm-near (changes requested):

Longarithm (approved):

runtime/near-test-contracts/src/lib.rs
```diff
@@ -12,6 +12,26 @@ pub fn trivial_contract() -> &'static [u8] {
         .as_slice()
 }
 
+/// Contract with exact size in bytes.
+pub fn sized_contract(size: usize) -> &'static [u8] {
+    static CONTRACT: OnceCell<Vec<u8>> = OnceCell::new();
+    CONTRACT
```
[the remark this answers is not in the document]
jakmeier: good catch, removed the caching

jakmeier:
From my point of view, this is ready for final review.

I contacted everyone that I could identify to be affected by the change. And besides the test that I wrote already, I cannot think of anything that could be tested here.

If possible, I would like to include this in NEAR TestNet Release - 1.26.0-rc.1, scheduled for Wed, March 23.

bowenwang1996 (approved):

mm-near (approved):

Issue #3120, opened 2020-08-08
Worst case Wasm CPU soak test


maxzaver:
# Motivation

We are currently using this code [https://github.com/nearprotocol/nearcore/blob/master/runtime/runtime-params-estimator/test-contract/src/lib.rs#L681](https://github.com/nearprotocol/nearcore/blob/master/runtime/runtime-params-estimator/test-contract/src/lib.rs#L681) to compute the ratio `<num CPU instructions>/<num Wasm operations>`. However this code might not be pessimistic enough, it might not be emitting the worst case Wasm operations. This might lead us to underestimating the cost of Wasm operations which might open the potential abuse vector.

# Goal

We need to write Wasm code that emits as many as possible CPU instructions per Wasm operation. It cannot use thread-specific stuff so atomics cannot be used.

willemneal:
This seems like a good use case for a fuzzer.

bowenwang1996:
@olonho what is the status of this issue?

bowenwang1996:
@matklad @olonho

PR #6405, opened 2022-03-09
feat(params-estimator): Wasm Op estimation V2

jakmeier:
Introduce an alternative way to estimate Wasm instruction cost, that
executes a single type of instruction in a tight loop.
    
The current wasm instruction estimation only looks at memory load/store.
While this might be generally  the slowest operation, for Icount based
measurements this is not really relevant. We need to estimate the
operation that produces the most x86 ops.
Also, memory-prefetchers can often avoid memory latency, which might
introduce errors into the current time based measurements.

### DIFF
--- a/runtime/runtime-params-estimator/src/estimator_context.rs
+++ b/runtime/runtime-params-estimator/src/estimator_context.rs
@@ -1,12 +1,21 @@
 use std::collections::HashMap;
+use std::sync::Arc;
 
+use near_primitives::contract::ContractCode;
+use near_primitives::runtime::config_store::RuntimeConfigStore;
 use near_primitives::transaction::SignedTransaction;
-use near_vm_logic::ExtCosts;
+use near_primitives::types::CompiledContractCache;
+use near_store::{create_store, StoreCompiledContractCache};
+use near_vm_logic::mocks::mock_external::MockedExternal;
+use near_vm_logic::{ExtCosts, ProtocolVersion};
+use nearcore::get_store_path;
+use tempfile::TempDir;
 
 use crate::config::{Config, GasMetric};
 use crate::gas_cost::GasCost;
 use crate::testbed::RuntimeTestbed;
 use crate::utils::get_account_id;
+use crate::vm_estimator::create_context;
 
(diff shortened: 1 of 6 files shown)


runtime/runtime-params-estimator/src/estimator_context.rs
```diff
@@ -31,6 +40,20 @@ pub(crate) struct CachedCosts {
     pub(crate) apply_block: Option<GasCost>,
 }
 
+/// A raw runtime to estimate costs that don't need anything external. Also
+/// allows to compare VMs using the --vm-kind option.
+pub(crate) struct EstimatorRuntime {
```
matklad, line 45: Let's call it `WASM` runtime (or, well, contract runtime), that's more descriptive than Raw or estimator

runtime/runtime-params-estimator/src/lib.rs
```diff
@@ -767,6 +770,36 @@ fn wasm_instruction(ctx: &mut EstimatorContext) -> GasCost {
...
+        eprintln!("i64 add {add:?}");
+        eprintln!("i64 mul {mul:?}");
+
+        eprintln!("f64 div {f_div:?}");
+        eprintln!("f64 add {f_add:?}");
+        eprintln!("f64 mul {f_mul:?}");
+        eprintln!("f64 sqrt {f_sqrt:?}");
+    }
+
+    div.max(rem).max(add).max(mul).max(f_div).max(f_add).max(f_mul).max(f_sqrt)
```
matklad, line 800: I'd put the bunch into an array and `.into_iter().max().unwrap()` to save typing.  Actually, these are all identical, why don't we *start* with something like

```rust
let binary_ops = [("i64", "div_s"), ("i64", "rem_s")]
```

?

And then just loop, rather than copy-paste.

bowenwang1996:
Unrelated but this is a github bug?
Screen Shot 2022-03-23 at 4 07 06 PM

matklad:
realted issue: https://github.com/near/nearcore/issues/3120

akhi3030:
I think this is not being actively worked on.

PR #6406, opened 2022-03-09
fix shard tracking issue for localnet

mzhangmzz:
Right now, localnet use a validator seat config that will split validators to all shards, as a result, every validator will track only a subset of shards in the localnet, even when ChunkOnly producer is not enabled. This is not desired since we want to run localnet similar to the setup of testnet and mainnet right now, where all validators track all shards. This PR changes that by changing how the localnet genesis config is generated.


chain/epoch_manager/src/lib.rs
```diff
@@ -1179,8 +1181,17 @@ impl EpochManager {
         shard_id: ShardId,
     ) -> Result<bool, EpochError> {
         let epoch_info = self.get_epoch_info(&epoch_id)?;
-        let chunk_producers = epoch_info.chunk_producers_settlement();
-        for validator_id in chunk_producers[shard_id as usize].iter() {
+        let validators = if !checked_feature!(
```
matklad, line 1184: let's mabybe flip branches to avoid `!`? There's :bulb: for flipping an if.

bowenwang1996:
@mzhangmzz is this PR ready for review?

mzhangmzz:
> @mzhangmzz is this PR ready for review?

Yeah it is, sorry missed your comment and just forgot about this PR in general

bowenwang1996:
looks like CI is failing :(

core/primitives/src/utils.rs
```diff
@@ -409,17 +408,7 @@ pub fn to_timestamp(time: DateTime<chrono::Utc>) -> u64 {
...
-            let remainder = num_seats % num_shards;
-            let num = if i < remainder as u64 {
-                num_seats / num_shards + 1
-            } else {
-                num_seats / num_shards
-            };
-            max(num, 1)
-        })
-        .collect()
+    vec![num_seats; num_shards as usize]
```
mm-near, line 411: so we'll assign the same number of seats for each shard? 

This changes the meaning of this function - and we should at least update the comment.

mm-near (changes requested):

core/primitives/src/utils.rs
mzhangmzz: Originally this function splits the validators among shards, so each validator is only assigned to one shard. I don't think this is what we want right now, especially for localnet, since in the current stage all validators track all shards. I'll update the comments.

mzhangmzz:
I reverted the change in "cares_about_shard_in_epoch" because I realized it's actually not right. I misread the code in sample_chunk_producer.

core/primitives/src/utils.rs
```diff
@@ -407,19 +406,11 @@ pub fn to_timestamp(time: DateTime<chrono::Utc>) -> u64 {
...
-                num_seats / num_shards + 1
-            } else {
-                num_seats / num_shards
-            };
-            max(num, 1)
-        })
-        .collect()
+pub fn get_num_seats_per_shard(
+    num_shards: NumShards,
+    num_seats_per_shard: NumSeats,
```
mm-near, line 411: this is a little bit tricky -- as you're changing the meaning of the argument (and this method is called in many places).

matklad: Should be maybe just inline it everywhere? It exists mostly for tests, and seem to create more confusion than any future-proofing.

mzhangmzz: Yeah the tests are a little tricky. I didn't go through all of the tests, but most of them I saw are only using one shard. I'll remove the function and just try inlining it everywhere so I can check all the tests affected.

mzhangmzz: Actually I was being dumb. The current change should be good.

mm-near (approved):

mina86:
Looks like this broke https://nayduck.near.org/#/test/321011

PR #6407, opened 2022-03-09
add a debug page with sync and network info

jc-near:
This PR is to add a debugging page that reflects the sync status in a browser window.

JIRA Issue: [CP-10](https://nearinc.atlassian.net/browse/CP-10)

Test plan
 - Build with `make neard`
 - Launch using command `nearup stop && nearup run localnet --binary-path ./target/release`
 - Open your browser and go to http://[host_address[:3030/debug/
 - Expect to see the link "Sync Info"
 - Click the link "Sync Info"
 
 Screenshot

(on localnet)
Screen Shot 2022-03-09 at 1 19 22 PM

(on testnet)
Screen Shot 2022-03-15 at 4 50 54 PM

First-round Reviewers
 - @mm-near 
 - @mzhangmzz

### DIFF
--- a/chain/client/src/client_actor.rs
+++ b/chain/client/src/client_actor.rs
@@ -758,7 +758,11 @@ impl Handler<Status> for ClientActor {
                 last_block_height = block.header().height();
             }
 
-            Some(DetailedDebugStatus { last_blocks: blocks_debug })
+            Some(DetailedDebugStatus {
+                last_blocks: blocks_debug,
+                network_info: self.network_info.clone().into(),
+                sync_status: self.client.sync_status.as_variant_name().to_string(),
+            })
         } else {
             None
         };
--- a/chain/jsonrpc/res/sync_info.html
+++ b/chain/jsonrpc/res/sync_info.html
@@ -31,12 +31,12 @@
     </style>
     <script src="https://ajax.googleapis.com/ajax/libs/jquery/3.5.1/jquery.min.js"></script>
     <script>
-        $(document).ready(function () {
+        $(document).ready(() => {
             $('span').text("Loading...");
             $.ajax({
                 type: "GET",
                 url: "/debug/api/sync_info",
-                success: function (data) {
+                success: data => {
                     let sync_status = data.detailed_debug_status.sync_status;
                     let network_info = data.detailed_debug_status.network_info;
                     $('.js-sync-status').text(sync_status);
--- a/chain/jsonrpc/res/sync_info.html
+++ b/chain/jsonrpc/res/sync_info.html
@@ -1,0 +1,93 @@
+<html>
+<head>
+    <style>
+        table {
+            width: 100%;
+            border-collapse: collapse;
+        }
+
+        table,
+        th,
+        td {
+            border: 1px solid black;
+        }
+
+        td {
+            text-align: left;
+            vertical-align: top;
+            padding: 8px;
+        }
+
+        th {
+            text-align: center;
+            vertical-align: center;
+            padding: 8px;
+            background-color: lightgrey;
+        }
+
+        tr.active {
+            background-color: #eff8bf;
+        }
+    </style>
+    <script src="https://ajax.googleapis.com/ajax/libs/jquery/3.5.1/jquery.min.js"></script>
+    <script>
+        $(document).ready(function () {
+            $('span').text("Loading...");
+            $.ajax({
+                type: "GET",
+                url: "/debug/api/sync_info",
+                success: function (data) {
+                    let sync_status = data.detailed_debug_status.sync_status;
+                    let network_info = data.detailed_debug_status.network_info;
+                    $('.js-sync-status').text(sync_status);
+                    $('.js-max-peers').text(network_info.peer_max_count);
+                    $('.js-num-peers').text(network_info.num_connected_peers);
+                    network_info.connected_peers.forEach(function(peer, index) {
+                        $('.js-tbody-peers').append($('<tr>')
+                            .append($('<td>').append(peer.addr))
+                            .append($('<td>').append(JSON.stringify(peer.account_id)))
+                            .append($('<td>').append(JSON.stringify(peer.height)))
+                            .append($('<td>').append(JSON.stringify(peer.tracked_shards)))
+                            .append($('<td>').append(JSON.stringify(peer.archival)))
+                        )
+                    });
+                    return;
+                },
+                dataType: "json",
+                error: function (errMsg, textStatus, errorThrown) {
+                    alert("Failed: " + textStatus + " :" + errorThrown);
+                },
+                contentType: "application/json; charset=utf-8",
+            })
+
+        });
+    </script>
+</head>
+<body>
+    <h1>
+        Welcome to the Sync Status page!
+    </h1>
+    <h2>
+        <p>
+            Current Sync Status:
+            <span class="js-sync-status"></span>
+        </p>
+        <p>
+            Number of peers: <span class="js-num-peers"></span>/<span class="js-max-peers"></span>
+        </p>
+    </h2>
+
+    <table>
+        <thead><tr>
+            <th>Address</th>
+            <th>Account ID</th>
+            <th>Height</th>
+            <th>Tracked Shards</th>
+            <th>Archival</th>
+        </tr></thead>
+        <tbody class="js-tbody-peers">
+        </tbody>
+    </table>
+</body>
+
+</html>
--- a/chain/jsonrpc/res/sync_info.html
+++ b/chain/jsonrpc/res/sync_info.html
@@ -51,7 +51,6 @@
                             .append($('<td>').append(JSON.stringify(peer.archival)))
                         )
                     });
-                    return;
                 },
                 dataType: "json",
                 error: function (errMsg, textStatus, errorThrown) {
--- a/chain/jsonrpc/src/lib.rs
+++ b/chain/jsonrpc/src/lib.rs
@@ -904,19 +904,6 @@ impl JsonRpcHandler {
         }
     }
 
-    pub async fn sync_info(
-        &self,
-    ) -> Result<
-        near_jsonrpc_primitives::types::status::RpcStatusResponse,
-        near_jsonrpc_primitives::types::status::RpcStatusError,
-    > {
-        Ok(self
-            .client_addr
-            .send(Status { is_health_check: false, detailed: true })
-            .await??
-            .into())
-    }
-
     /// Expose Genesis Config (with internal Runtime Config) without state records to keep the
     /// output at a reasonable size.
     ///
@@ -1397,7 +1384,7 @@ pub async fn prometheus_handler() -> Result<HttpResponse, HttpError> {
 }
 
 async fn sync_info_handler(handler: web::Data<JsonRpcHandler>) -> Result<HttpResponse, HttpError> {
-    match handler.sync_info().await {
+    match handler.debug().await {
         Ok(value) => Ok(HttpResponse::Ok().json(&value)),
         Err(_) => Ok(HttpResponse::ServiceUnavailable().finish()),
     }
--- a/chain/jsonrpc/src/lib.rs
+++ b/chain/jsonrpc/src/lib.rs
@@ -904,6 +904,19 @@ impl JsonRpcHandler {
         }
     }
 
+    pub async fn sync_info(
+        &self,
+    ) -> Result<
+        near_jsonrpc_primitives::types::status::RpcStatusResponse,
+        near_jsonrpc_primitives::types::status::RpcStatusError,
+    > {
+        Ok(self
+            .client_addr
+            .send(Status { is_health_check: false, detailed: true })
+            .await??
+            .into())
+    }
+
     /// Expose Genesis Config (with internal Runtime Config) without state records to keep the
     /// output at a reasonable size.
     ///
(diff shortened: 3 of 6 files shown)


chain/jsonrpc/src/lib.rs
```diff
@@ -1383,6 +1396,18 @@ pub async fn prometheus_handler() -> Result<HttpResponse, HttpError> {
     }
 }
 
+fn sync_info_handler(
```
mzhangmzz, line 1399: nit: a simpler way to write this is 
`async fn sync_info_handler(..) -> Result<HttpResponse, HttpError>`

chain/client/src/client_actor.rs
```diff
@@ -721,6 +722,30 @@ impl Handler<Status> for ClientActor {
         } else {
             None
         };
+        // Provide info about the current state of sync
+        let debug_sync_info = if msg.get_debug_sync_info {
+            Some(DebugSyncInfo {
+                network_info: NetworkInfoView {
+                    peer_max_count: self.network_info.peer_max_count,
```
mzhangmzz, line 729: Let's implement 
`impl From<PeerInfo> for PeerInfoView` and `impl From<NetworkInfo> for NetworkInfoView`, example: https://github.com/near/nearcore/blob/master/core/primitives/src/views.rs#L830-L861
Then you can simple call 
`network_info: self.network_info.into()`

mzhangmzz, line 729: Let's implement 
`impl From<PeerInfo> for PeerInfoView` and `impl From<NetworkInfo> for NetworkInfoView`, example: https://github.com/near/nearcore/blob/master/core/primitives/src/views.rs#L830-L861
Then you can simple call 
`network_info: self.network_info.into()`

core/primitives/src/views.rs
```diff
@@ -345,12 +346,39 @@ pub struct DebugBlockStatus {
...
+    pub peer_counter: usize,
+}
+
 #[cfg_attr(feature = "deepsize_feature", derive(deepsize::DeepSizeOf))]
 #[derive(Serialize, Deserialize, Debug)]
 pub struct DetailedDebugStatus {
     pub last_blocks: Vec<DebugBlockStatus>,
 }
 
+#[cfg_attr(feature = "deepsize_feature", derive(deepsize::DeepSizeOf))]
```
mzhangmzz, line 375: You don't need the DeepSizeOf decorator here

chain/client-primitives/src/types.rs
```diff
@@ -367,6 +367,7 @@ pub struct Status {
     pub is_health_check: bool,
     // If true - return more detailed information about the current status (recent blocks etc).
     pub detailed: bool,
+    pub get_debug_sync_info: bool,
```
mm-near, line 370: I'd re-use the "detailed" bool instead.

mm-near (changes requested):
I'd put this all under the "DetailedDebugStatus" (rather than creating another "high level" field).

yes, it might be a little bit less efficient - but will make the PR smaller and will simplify the flow and adding future fields/debug pages.

mm-near:
also - do you happen to have this code running already against mainnet or testnet ? Would be cool to see how the webpage looks like in "real life".
(go/own-mainnet)

chain/jsonrpc/src/lib.rs
jc-near:
```
error[E0308]: mismatched types
    --> chain/jsonrpc/src/lib.rs:1406:5
     |
1406 |     response.boxed()
     |     ^^^^^^^^^^^^^^^^ expected enum `Result`, found struct `Pin`
     |
     = note: expected enum `Result<actix_web::HttpResponse, actix_web::Error>`
              found struct `Pin<std::boxed::Box<dyn futures::Future<Output = Result<actix_web::HttpResponse, _>> + std::marker::Send>>`
```

jc-near: With the following signature:
`async fn sync_info_handler(handler: web::Data<JsonRpcHandler>) -> Result<HttpResponse, HttpError>`

Am I doing it right?

core/primitives/src/views.rs
jc-near: I realized that one of its parent struct (`DetailedDebugStatus`) needs this trait, so I am adding it back.

chain/jsonrpc/src/lib.rs
mzhangmzz: Yes, once you change the function signature, you don't need the "async" inside the function anymore. 
```
match handler.sync_info().await {
            Ok(value) => Ok(HttpResponse::Ok().json(&value)),
            Err(_) => Ok(HttpResponse::ServiceUnavailable().finish()),
        }
```
See examples from the above functions with async in their signature. For more reference, you can take a look here https://rust-lang.github.io/async-book/03_async_await/01_chapter.html

mzhangmzz:
> also - do you happen to have this code running already against mainnet or testnet ? Would be cool to see how the webpage looks like in "real life". (go/own-mainnet)

Could you start a node on mainnet that is falling behind so it's syncing, so we can see what the debug page will print when the node is syncing

chain/jsonrpc/src/lib.rs
jc-near: Great! Thanks for the explanation.

mzhangmzz:
And please don't squash your commits when it's not absolutely necessary, see https://docs.google.com/document/d/1trdvGsXyC8RPQRmCU1uATZrwg_caiSqpSK_dST0D2Eg/edit

chain/client/src/client_actor.rs
jc-near: Done!

jc-near:
> And please don't squash your commits when it's not absolutely necessary, see https://docs.google.com/document/d/1trdvGsXyC8RPQRmCU1uATZrwg_caiSqpSK_dST0D2Eg/edit

Thanks for the reminder!

core/primitives/src/views.rs
```diff
@@ -345,10 +345,33 @@ pub struct DebugBlockStatus {
...
+    pub archival: bool,
+}
+
+#[cfg_attr(feature = "deepsize_feature", derive(deepsize::DeepSizeOf))]
+#[derive(Serialize, Deserialize, Debug, PartialEq, Eq)]
+pub struct NetworkInfoView {
+    pub peer_max_count: u32,
+    pub num_connected_peers: usize,
+    pub connected_peers: Vec<PeerInfoView>,
+    pub sent_bytes_per_sec: u64,
```
mm-near, line 364: this is over last minute? or more?

core/primitives/src/views.rs
```diff
@@ -345,10 +345,33 @@ pub struct DebugBlockStatus {
...
+    pub received_bytes_per_sec: u64,
+    pub peer_counter: usize,
```
mm-near, line 366: what is 'peer_counter'?

chain/network/src/types.rs
```diff
@@ -327,6 +327,21 @@ pub struct FullPeerInfo {
     pub partial_edge_info: PartialEdgeInfo,
 }
 
+impl From<&FullPeerInfo> for PeerInfoView {
+    fn from(full_peer_info: &FullPeerInfo) -> Self {
+        PeerInfoView {
+            addr: match full_peer_info.peer_info.addr {
+                Some(socket_addr) => socket_addr.to_string(),
+                None => "N/A".to_string(),
```
mm-near, line 335: or maybe we can make 'addr' an Option field instead?

mm-near (approved):

matklad (changes requested):
I am confused here a bit. Right now, we have the following code:

```rust
    pub async fn debug(
        &self,
    ) -> Result<
        Option<near_jsonrpc_primitives::types::status::RpcStatusResponse>,
        near_jsonrpc_primitives::types::status::RpcStatusError,
    > {
        if self.enable_debug_rpc {
            Ok(Some(
                self.client_addr
                    .send(Status { is_health_check: false, detailed: true })
                    .await??
                    .into(),
            ))
        } else {
            return Ok(None);
        }
    }

    pub async fn sync_info(
        &self,
    ) -> Result<
        near_jsonrpc_primitives::types::status::RpcStatusResponse,
        near_jsonrpc_primitives::types::status::RpcStatusError,
    > {
        Ok(self
            .client_addr
            .send(Status { is_health_check: false, detailed: true })
            .await??
            .into())
    }
```

That is, these two method are identical, except that the new one doesn't check the `enable_debug_rpc` flag. 

This doesn't make sense to me, shouldn't we just use existing `debug` rather than adding an identical method?

jc-near:
> Member

Changed it according to your request.

matklad (approved):
Approving! It still feels odd to me that we have two distinct API endpoints, `/debug/api/sync_info` and `/debug/api/last_blocks` which return exactly the same data, but, as they are now both guarded behind a debug flag, it doesn't seem to be an immediate problem.

core/primitives/src/views.rs
mm-near: ack

mm-near (approved):

jc-near:
@mina86  @bowenwang1996 @nikurt  Can I get approvals from you three in order to proceed with a merge?

chain/network/src/types.rs
mina86: How it’s shown in the user interface is separate matter. You can have `foo.addr || 'N/A'`.

mina86 (approved):
Just minor comments. No strong feelings.

(Also, what am I an owner that mm-near isn’t? O_o)

core/primitives/src/views.rs
jc-near: I think rate-limiting is taken into account.

chain/network/src/types.rs
jc-near: The other thing is, this `addr` field needs to be deepsize-able. That is the other reason why a string is preferred here.

chain/jsonrpc/res/sync_info.html
```diff
@@ -0,0 +1,92 @@
...
+            background-color: lightgrey;
+        }
+
+        tr.active {
+            background-color: #eff8bf;
+        }
+    </style>
+    <script src="https://ajax.googleapis.com/ajax/libs/jquery/3.5.1/jquery.min.js"></script>
+    <script>
+        $(document).ready(function () {
```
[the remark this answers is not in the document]
jc-near: I think it is just a habit. Any style guide?

chain/jsonrpc/res/sync_info.html
```diff
@@ -0,0 +1,92 @@
+<html>
```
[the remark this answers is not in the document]
jc-near: Good to know!

core/primitives/src/views.rs
jc-near: See https://github.com/near/nearcore/blob/c43516499a972b066c0e6f649c88597192f2c7e8/chain/network/src/peer_manager/peer_manager_actor.rs#L188

chain/jsonrpc/res/sync_info.html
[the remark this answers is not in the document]
mina86: Arrow functions are considered to be less confusing since they don’t change what `this` is. (The syntax is also more concise).

chain/client/src/client_actor.rs
```diff
@@ -756,7 +756,11 @@ impl Handler<Status> for ClientActor {
                 last_block_height = block.header().height();
             }
 
-            Some(DetailedDebugStatus { last_blocks: blocks_debug })
+            Some(DetailedDebugStatus {
+                last_blocks: blocks_debug,
+                network_info: self.network_info.clone().into(),
+                sync_status: self.client.sync_status.as_variant_name().to_string(),
```
mzhangmzz, line 762: Sync status should display more information than just the enum itself. For example, in HeaderSync, it would be nice to know where the current header head is vs where it is trying to sync to. Please see `display_sync_status` for what it displayed there.

This can be done in a separate PR though

core/primitives/src/views.rs
```diff
@@ -347,10 +347,33 @@ pub struct DebugBlockStatus {
...
+#[derive(Serialize, Deserialize, Debug, PartialEq, Eq)]
+pub struct PeerInfoView {
+    pub addr: String,
+    pub account_id: Option<AccountId>,
+    pub height: BlockHeight,
+    pub tracked_shards: Vec<ShardId>,
+    pub archival: bool,
+}
+
+#[cfg_attr(feature = "deepsize_feature", derive(deepsize::DeepSizeOf))]
```
mzhangmzz, line 360: I'm not sure if the deepsize feature here is needed. I'm not too familiar with what deepsize, but I think it's used to measure struct sizes and I don't think we need to measure struct sizes here

core/primitives/src/views.rs
mzhangmzz: Frol has a point. Let's clean up the fields in this struct to remove the parts we are not displaying. We can always add more later if we find anything useful.

chain/network/src/types.rs
jc-near: I tried deleting the deepsize thing and it worked. So now I am able to change this one as well.

jc-near: See https://buildkite.com/nearprotocol/nearcore/builds/14403#be1e23a6-82fa-4c13-a569-48ff56616a36. Deepsize is required on `NetworkInfoView` and `PeerInfoView`. However, `SocketAddr` is not deep-sizeable, and must be replaced by an alternative type.

Since this struct is a view and is only responsible for storing the presentation of the data, rather than the data itself, a string should be okay.

mzhangmzz: You could get around this problem by implementing the deepsize trait manually. Example: https://github.com/near/nearcore/blob/master/chain/network-primitives/src/types.rs#L187-L191

mzhangmzz (approved):

PR #6411, opened 2022-03-10
endtoend: make more resilient to exceptions and other minor improvements

mina86:
Most notably, wrap body of the pinger loop in a try-except block which
logs exception information if any exception happens.  This way, even if
something goes wrong during one iteration, the test will continue trying.

And while editing the file, replace `for _ in range(len(_))` loops by
iterating over elements of lists.  This is more idiomatic and removes the
need to explicitly index the lists.

Lastly, initialise Prometheus client only after all arguments are validated.


nikurt (approved):
Thank you!

mm-near (approved):

PR #6414, opened 2022-03-11
feat: implement chunk nodes cache

Longarithm:
Introduce cache for storing items retrieved for a specific chunk to `TrieCachingStorage`.

Suggested model:
* By default, we are in the `CachingShard` state.
* On funcall execution, we will temporarily enable `CachingChunk` state.

Because all nodes touched in such state will be charged with 16 Ggas, the chunk cache size will be limited by 500 Tgas / 16 Ggas = 31250 elements. For each such node, we charge for touching it only once - on putting it into the cache.

The current implementation does not have effect on costs and cache chunk will remain empty. `CachingChunk` mode will be enabled in separate PR with protocol upgrade. With enabled costs, it will save ~33% of the cost for Aurora and ~87.5% of the touching trie node cost in particular.

The difference of receipt processing times is 2%, but this can be mitigated later, see https://near.zulipchat.com/#narrow/stream/295558-nearinc.2Fcore/topic/chunk.20nodes.20cache.20prototype for more details.

## Testing

* new trie cache tests in `trie_storage` 

## Reviewing
@EdvardD @matklad


matklad (commented):
Architecture-wise, this looks good to me now! Didn't do careful logic review yet.

matklad:
Took a closer look here, and it seems that we have way to many helper struct and types here. I *think* this should be equivalent to the following diff:

```diff
diff --git a/core/store/src/trie/trie_storage.rs b/core/store/src/trie/trie_storage.rs
index b337bf5b3..4075d0d72 100644
--- a/core/store/src/trie/trie_storage.rs
+++ b/core/store/src/trie/trie_storage.rs
@@ -165,32 +165,6 @@ const TRIE_MAX_SHARD_CACHE_SIZE: usize = 1;
 /// Note that Trie inner nodes are always smaller than this.
 const TRIE_LIMIT_CACHED_VALUE_SIZE: usize = 4000;
 
-/// Position of the value in cache.
-#[derive(Debug)]
-pub(crate) enum CachePosition {
-    /// Value is not present.
-    None,
-    /// Value is present **only** in the shard cache.
-    ShardCache(Arc<[u8]>),
-    /// Value is present in the chunk cache.
-    ChunkCache(Arc<[u8]>),
-}
-
-/// Cost of retrieving trie node from the storage. Used to compute gas cost for touching trie nodes.
-#[derive(Debug, Eq, PartialEq)]
-pub enum TrieNodeRetrievalCost {
-    Free,
-    Full,
-}
-
-#[derive(Debug, Eq, PartialEq)]
-pub(crate) struct RawBytesWithCost {
-    /// Bytes of the retrieved node. None if no value was found in cache.
-    pub(crate) value: Option<Arc<[u8]>>,
-    /// Cost of node retrieval.
-    pub(crate) cost: TrieNodeRetrievalCost,
-}
-
 pub struct TrieCachingStorage {
     pub(crate) store: Store,
     pub(crate) shard_uid: ShardUId,
@@ -246,52 +220,6 @@ impl TrieCachingStorage {
         self.counter.set(self.counter.get() + 1);
     }
 
-    /// Get position of the given key in the cache.
-    pub(crate) fn get_cache_position(&self, key: &CryptoHash) -> CachePosition {
-        match self.chunk_cache.borrow_mut().get(key) {
-            Some(value) => CachePosition::ChunkCache(value.clone()),
-            None => match self.shard_cache.get(key) {
-                Some(value) => CachePosition::ShardCache(value.clone()),
-                None => CachePosition::None,
-            },
-        }
-    }
-
-    /// If we are in the caching chunk state, put item to the chunk cache.
-    fn try_put_to_chunk_cache(&self, key: CryptoHash, value: Arc<[u8]>) {
-        if let TrieCacheState::CachingChunk = self.cache_state.borrow().get() {
-            self.chunk_cache.borrow_mut().insert(key, value);
-        };
-    }
-
-    /// Get value for the given key from cache.
-    /// Return value if it is present and cost of retrieving the trie node corresponding to this key.
-    pub(crate) fn get_from_cache(&self, key: &CryptoHash) -> RawBytesWithCost {
-        match self.get_cache_position(key) {
-            // If no value is present, return None. If node for this key will be retrieved from DB, we have to charge
-            // full cost for it.
-            CachePosition::None => {
-                RawBytesWithCost { value: None, cost: TrieNodeRetrievalCost::Full }
-            }
-            // If value is present only in shard cache, we can copy value to the chunk cache. The cost must be full
-            // anyway because we didn't access the node in the chunk yet.
-            CachePosition::ShardCache(value) => {
-                self.try_put_to_chunk_cache(key.clone(), value.clone());
-                RawBytesWithCost { value: Some(value), cost: TrieNodeRetrievalCost::Full }
-            }
-            // If value is present in chunk cache, the cost is free.
-            CachePosition::ChunkCache(value) => {
-                RawBytesWithCost { value: Some(value), cost: TrieNodeRetrievalCost::Free }
-            }
-        }
-    }
-
-    /// Put a key-value pair into cache.
-    pub(crate) fn put_to_cache(&self, key: CryptoHash, value: Arc<[u8]>) {
-        self.shard_cache.put(key.clone(), value.clone());
-        self.try_put_to_chunk_cache(key, value.clone());
-    }
-
     /// Set cache state.
     pub fn set_state(&self, state: TrieCacheState) {
         self.cache_state.set(state);
@@ -300,36 +228,36 @@ impl TrieCachingStorage {
 
 impl TrieStorage for TrieCachingStorage {
     fn retrieve_raw_bytes(&self, hash: &CryptoHash) -> Result<Arc<[u8]>, StorageError> {
-        // Try to get value from cache and determine node retrieval cost.
-        let RawBytesWithCost { value, cost } = self.get_from_cache(hash);
+        if let Some(val) = self.chunk_cache.borrow_mut().get(hash) {
+            return Ok(val.clone());
+        }
+        self.inc_counter();
 
-        // If full cost for retrieving the node must be charged, increment the counter.
-        if let TrieNodeRetrievalCost::Full = cost {
-            self.inc_counter();
+        if let Some(val) = self.shard_cache.get(hash) {
+            if let TrieCacheState::CachingChunk = self.cache_state.borrow().get() {
+                self.chunk_cache.borrow_mut().insert(*hash, val.clone());
+            };
+            return Ok(val);
         }
 
-        match value {
-            // If value is not present, get it from the store and put into the cache.
-            None => {
-                let key = Self::get_key_from_shard_uid_and_hash(self.shard_uid, hash);
-                let val = self
-                    .store
-                    .get(ColState, key.as_ref())
-                    .map_err(|_| StorageError::StorageInternalError)?;
-                if let Some(val) = val {
-                    let val: Arc<[u8]> = val.into();
-                    if val.len() < TRIE_LIMIT_CACHED_VALUE_SIZE {
-                        self.put_to_cache(*hash, val.clone());
-                    }
-                    Ok(val)
-                } else {
-                    // not StorageError::TrieNodeMissing because it's only for TrieMemoryPartialStorage
-                    Err(StorageError::StorageInconsistentState("Trie node missing".to_string()))
-                }
-            }
-            // If value is present, return it.
-            Some(val) => Ok(val),
+        let key = Self::get_key_from_shard_uid_and_hash(self.shard_uid, hash);
+        let val = self
+            .store
+            .get(ColState, key.as_ref())
+            .map_err(|_| StorageError::StorageInternalError)?
+            .ok_or_else(|| {
+                StorageError::StorageInconsistentState("Trie node missing".to_string())
+            })?;
+
+        let val: Arc<[u8]> = val.into();
+        if val.len() < TRIE_LIMIT_CACHED_VALUE_SIZE {
+            self.shard_cache.put(*hash, val.clone());
+            if let TrieCacheState::CachingChunk = self.cache_state.borrow().get() {
+                self.chunk_cache.borrow_mut().insert(*hash, val.clone());
+            };
         }
+
+        Ok(val)
     }
 
     fn as_caching_storage(&self) -> Option<&TrieCachingStorage> {
```

Note how in the above there's only one function, `retrieve_raw_bytes`, which contains all state mutation. Compare that with the current state of the PR where, eg, both `retrieve_raw_bytes` and `get_from_cache` mutate state.

core/store/src/trie/trie_storage.rs
```diff
@@ -184,30 +245,90 @@ impl TrieCachingStorage {
...
+                let key = Self::get_key_from_shard_uid_and_hash(self.shard_uid, hash);
+                let val = self
+                    .store
+                    .get(ColState, key.as_ref())
+                    .map_err(|_| StorageError::StorageInternalError)?;
+                if let Some(val) = val {
+                    let val: Arc<[u8]> = val.into();
+                    if val.len() < TRIE_LIMIT_CACHED_VALUE_SIZE {
+                        self.put_to_cache(*hash, val.clone());
+                    }
```
matklad, line 323: This condition feels wrong to me -- it's OK to skip `shard_cache` based on value size, but it's *not* ok to skip the `chunk_hash`. I think the property we want is "adjusting `TRIE_LIMIT_CACHED_VALUE_SIZE` is not  protocol change", but in the current implementation it **would** be one. WE also should have a test for this case.

matklad (changes requested):
https://github.com/near/nearcore/pull/6414/files#r826211333 feels important

core/store/src/trie/trie_storage.rs
```diff
@@ -16,7 +18,23 @@ pub struct TrieCache(Arc<Mutex<LruCache<CryptoHash, Arc<[u8]>>>>);
 
 impl TrieCache {
     pub fn new() -> Self {
-        Self(Arc::new(Mutex::new(LruCache::new(TRIE_MAX_CACHE_SIZE))))
+        Self::with_capacity(TRIE_MAX_SHARD_CACHE_SIZE)
```
mm-near, line 21: I'd mention above, that TrieCache is a wrapper over the LruCache, but doesn't hold too large elements

core/primitives/src/types.rs
```diff
@@ -977,3 +977,13 @@ pub trait EpochInfoProvider {
...
     fn minimum_stake(&self, prev_block_hash: &CryptoHash) -> Result<Balance, EpochError>;
 }
+
+/// State of the trie cache.
+#[derive(Debug, Copy, Clone)]
+pub enum TrieCacheState {
+    /// We put each visited node to LRU cache. It generally saves time, but existence of any node is not guaranteed;
+    CachingShard,
+    /// We put each visited node to the chunk cache which is a hash map. On the chunk processing, we guarantee that all
+    /// nodes visited on such state will remain in the chunk cache.
```
EdvardD, line 987: Probably the second sentence worth rephrasing. Like this:

This is needed to guarantee the all nodes are retrieved at most once during a single chunk processing. In other words we guarantee that any node visited during chunk processing lives in cache until the chunk processing is finished. The guarantee is needed so we can compute the gas in a way that user doesn't pay twice for a single node retrieval.

core/store/src/trie/trie_storage.rs
```diff
@@ -133,31 +151,74 @@ impl TrieStorage for TrieMemoryPartialStorage {
...
-const TRIE_MAX_CACHE_SIZE: usize = 1;
+const TRIE_MAX_SHARD_CACHE_SIZE: usize = 1;
 
 /// Values above this size (in bytes) are never cached.
 /// Note that Trie inner nodes are always smaller than this.
 const TRIE_LIMIT_CACHED_VALUE_SIZE: usize = 4000;
 
+/// Position of the value in cache.
+#[derive(Debug)]
+pub(crate) enum CachePosition {
```
EdvardD, line 170: IMO CachePosition should refer to the position in the cache but this looks more like CacheType

core/store/src/trie/trie_tests.rs
```diff
@@ -202,6 +202,137 @@ mod caching_storage_tests {
...
+            TrieCachingStorage::new(store, trie_cache, ShardUId::single_shard());
+        let value = &values[0];
+        let key = hash(value);
+
+        for _ in 0..2 {
+            let count_before = trie_caching_storage.get_touched_nodes_count();
+            let result = trie_caching_storage.retrieve_raw_bytes(&key);
+            let bytes: Arc<[u8]> = result.into();
+            let count_after = trie_caching_storage.get_touched_nodes_count();
+            assert_eq!(result, Some(bytes));
```
mm-near, line 253: but bytes are computed from 'result.into()'.

shouldn't you compare it with value instead?

mm-near (changes requested):

core/store/src/trie/trie_storage.rs
```diff
@@ -184,30 +245,90 @@ impl TrieCachingStorage {
...
+    }
+
+    /// Put a key-value pair into cache.
+    pub(crate) fn put_to_cache(&self, key: CryptoHash, value: Arc<[u8]>) {
+        self.shard_cache.put(key.clone(), value.clone());
+        self.try_put_to_chunk_cache(key, value.clone());
+    }
+
+    /// Set cache state.
+    pub fn set_state(&self, state: TrieCacheState) {
```
EdvardD, line 296: I would probably implement this in the following manner: instead of set_state have a method switch_to_chunk_caching which returns RAII object switching the state back. It seems important to explicitly (on the caller function side) make sure that the state will be switched back to shard caching.

EdvardD:
Overall looks good.
How did you test that this is no-op change (except for 2% degradation)?

EdvardD (approved):
Please consider using RAII object for state changes.

core/store/src/trie/trie_storage.rs
matklad: :thinking: yeah, it might be ok to make "values less than X bytes are cached" to be a property of the protocol. However, with the current implementation, we are not looking at the size of values, we are looking at the size of the whole encoding node. That is, we also leak our physical representation of the data on disk... 

That is, if, eg, we remove the `memory_usage` field from `RawTrieNodeWithSize`, that's also going to be a protocol change.

That is, the problem here is that we define the protocol not in terms of abstract shape of the trie (sizes of values or extension keys), but in terms of specific implementation.

matklad: :thinking: so, we actually **have** to make some such constant a part of the protocol. Originally I was thinking "let's just not cache leaf nodes (values)", but we of course can have an intermediate extension node with 4mb key, and we should have something in the protocol that prevents caching of such a gigantic node.

I feel pretty strongly that the condition for caching should be definable in terms of logical shape of the trie. That is, ideally we *first* decode the node, then look at the combined length of node's key and value, and make caching decision after that. This is in contrast to today's approach where we make caching decision prior to decoding.

However, taking a second look at the current implementation, it does seem that it can be reformulated in terms of logical shape of the trie. I think the rule is:

```
Trie node is cacheable if any of the following holds:

* it's an internal node
* it's a leaf node with key length <= (4000 - 4 - 4 - 4 - 32)
* it's an extension node with key length <= (4000 - 4 - 4 - 32)
* it's a value node with value length <= 4000
```

This seems workable, *but* we need tests for all those cases.

core/store/src/trie/trie_storage.rs
```diff
@@ -16,7 +18,23 @@ pub struct TrieCache(Arc<Mutex<LruCache<CryptoHash, Arc<[u8]>>>>);
...
+    pub fn with_capacity(cap: usize) -> Self {
+        Self(Arc::new(Mutex::new(LruCache::new(cap))))
+    }
+
+    pub fn get(&self, key: &CryptoHash) -> Option<Arc<[u8]>> {
+        self.0.lock().expect(POISONED_LOCK_ERR).get(key).cloned()
+    }
+
+    pub fn put(&self, key: CryptoHash, value: Arc<[u8]>) {
+        self.0.lock().expect(POISONED_LOCK_ERR).put(key, value);
```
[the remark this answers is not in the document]
Longarithm: I'm already doing it in `retrieve_raw_bytes`, as this condition was there before:
```
...
if val.len() < TRIE_LIMIT_CACHED_VALUE_SIZE {
    self.put_to_cache(*hash, val.clone());
}
```

We don't have problems with previously existing values, because key is always a hash of a value. I put a comment about it inside `TrieCachingStorage`.

core/primitives/src/types.rs
Longarithm: That's not exactly true. We don't (currently can't) cache all nodes visited in a chunk. What we want to achieve is the implication "CachingChunk state is enabled => Each node is retrieved from DB at most once".

But yeah, I use it to improve readability.

core/store/src/trie/trie_storage.rs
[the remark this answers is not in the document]
Longarithm: Added a comment with call into `execute_function_call`. I uncomment it in the following PR with the protocol upgrade.

The cache is dropped when `TrieCachingStorage` will be dropped. It is created in the beginning of the `apply_transactions_with_optional_storage_proof`, and dropped in the end.

core/store/src/trie/trie_storage.rs
```diff
@@ -133,31 +151,74 @@ impl TrieStorage for TrieMemoryPartialStorage {
...
 
+    pub(crate) cache_state: Cell<TrieCacheState>,
+    /// Caches ever requested items for the shard `shard_uid`. Used to speed up DB operations, presence of any item is
+    /// not guaranteed.
+    pub(crate) shard_cache: TrieCache,
+    /// Caches all items requested in the `TrieCacheState::CachingChunk` state. It must be empty when we start to apply
+    /// txs and receipts in the chunk. All items placed here must remain until applying txs/receipts ends.
+    /// Note that for both caches key is the hash of value, so for the fixed key the value is unique.
+    /// TODO (#5920): enable chunk nodes caching in Runtime::apply.
+    pub(crate) chunk_cache: RefCell<HashMap<CryptoHash, Arc<[u8]>>>,
```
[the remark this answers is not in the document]
Longarithm: Makes sense, however I think it is better to put `cache_state` after the caches.

core/store/src/trie/trie_storage.rs
Longarithm: Looks good. I just didn't understand how to make it work with the ownership structure we have (RuntimeExt -> TrieUpdate -> TrieStorage).

Will it be fine if I do it in a separate PR? I left a corresponding comment in `execute_function_call`.

EdvardD: sg

Longarithm:
@EdvardD 

How did you test that this is no-op change (except for 2% degradation)?

I run command like `neard view_state apply_range --start-index 55291778 --end-index 55291878 --shard-id 0 --sequential --verbose-output` for binaries compiled before and after the change. Block heights are taken from the end of my DB snapshot.

Then I parse resulting logs and check that `burnt_gas` and `touching_trie_node` cost didn't change. 
Same for shard 1.

[logs.zip](https://github.com/near/nearcore/files/8260965/logs.zip)

core/store/src/trie/trie_storage.rs
```diff
@@ -184,31 +220,53 @@ impl TrieCachingStorage {
     fn inc_counter(&self) {
         self.counter.set(self.counter.get() + 1);
     }
+
+    /// If we are in the caching chunk state, put item to the chunk cache.
+    fn try_put_to_chunk_cache(&self, key: CryptoHash, value: Arc<[u8]>) {
+        if let TrieCacheState::CachingChunk = self.cache_state.borrow().get() {
+            self.chunk_cache.borrow_mut().insert(key, value);
+        };
+    }
```
matklad, line 229:
```suggestion
    fn try_put_to_chunk_cache(&self, key: CryptoHash, value: &Arc<[u8]>) {
        if let TrieCacheState::CachingChunk = self.cache_state.borrow().get() {
            self.chunk_cache.borrow_mut().insert(key, Arc::clone(value));
        };
    }
```

Экономия на спичках (saving matches), but let's avoid needless clone here.

I am also not super happy about naming `try_` means "failable", not optional. We use `maybe_` elsewhere, so we might as well put it here

matklad: An alternative solution to the naming is to make sure there's only one call-site for the function, so that we can inline it:

```rust
    fn retrieve_raw_bytes(&self, hash: &CryptoHash) -> Result<Arc<[u8]>, StorageError> {
        // Try to get value from chunk cache containing free of charge nodes, otherwise increment the nodes counter.
        // let RawBytesWithCost { value, cost } = self.get_from_cache(hash);
        if let Some(val) = self.chunk_cache.borrow_mut().get(hash) {
            return Ok(val.clone());
        }

        // If full cost for retrieving the node must be charged, increment the counter.
        let val = match self.shard_cache.get(hash) {
            Some(val) => val,
            None => {
                // If value is not present in cache, get it from the storage.
                let key = Self::get_key_from_shard_uid_and_hash(self.shard_uid, hash);
                let val = self
                    .store
                    .get(ColState, key.as_ref())
                    .map_err(|_| StorageError::StorageInternalError)?
                    .ok_or_else(|| {
                        StorageError::StorageInconsistentState("Trie node missing".to_string())
                    })?;
                let val: Arc<[u8]> = val.into();

                // Insert value to shard and chunk caches.
                if val.len() < TRIE_LIMIT_CACHED_VALUE_SIZE {
                    self.shard_cache.put(*hash, val.clone());
                }
                val
            }
        };

        self.inc_counter();
        if let TrieCacheState::CachingChunk = this.cache_state.borrow().get() {
            self.chunk_cache.borrow_mut().insert(key, val);
        };

        Ok(val)
    }
```


One more thing I like about the above suggestion is that we move `self.inc_counter()` *after* fetcing from the DB. This is nice, because, if `.store.get` fails, we don't increment the counter (if we increment the counter but *dont* actually fetch the node, we break atomicity).

matklad (approved):
I think this is in a good shape now. I agree that, for chunk_cache, it's not unreasonable to remove the limit altogether, as the total size of the stuff we'll cache will indeed be limited by the costs.

Specifically, with `"storage_read_value_byte": 5611005,` (the lowest per-byte cost for storage ops), the total amount of IO in megabytes would be 

```
>>> (1000 * 10**12 / 5611005) / 2**20
170
```

That is, chunk cache would be at most 170 megs, which seems acceptable

core/store/src/trie/trie_storage.rs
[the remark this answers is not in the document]
mzhangmzz: why do we need to switch between the two mode/states? How will `TrieCachingStorage` be used outside of `apply_transactions_with_optional_storage_proof`?

core/store/src/trie/trie_storage.rs
```diff
@@ -184,31 +220,53 @@ impl TrieCachingStorage {
...
-                    guard.put(*hash, val.clone());
-                }
-                Ok(val)
-            } else {
-                // not StorageError::TrieNodeMissing because it's only for TrieMemoryPartialStorage
-                Err(StorageError::StorageInconsistentState("Trie node missing".to_string()))
-            }
+
+        // If full cost for retrieving the node must be charged, increment the counter.
+        if let Some(val) = self.shard_cache.get(hash) {
```
mzhangmzz, line 247: I think you should acquire lock in this function instead of inside the `get` and `put` functions because you want the actions in this function to be atomic. In fact, this struct is no longer thread safe with this modification because multiple threads can modify `self.chunk_cache` at the same time, and the `get` and `put` method should be guarded by the same lock, otherwise another thread could change the value between the `get` and `put` call.

It's possible that this function is still safe because how we share cache among different threads. I just feel that the story of lock here is very confusing and need to be flushed out more.

core/store/src/trie/trie_storage.rs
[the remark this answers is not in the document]
mzhangmzz: Instead of use `set_state`, how about we use `start_chunk_cache` and `finish_chunk_cache` to indicate the state(mode) transition. I think it's clearer that way for what the switch is for and where these methods should be called. And you also need to clear the chunk cache in `finish_chunk_cache`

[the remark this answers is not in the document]
Longarithm: > why do we need to switch between the two mode/states?

Ideally we would work only in `CachingChunk` mode. Unfortunately, there are many trie node touches for which we don't charge any costs (caused by `get_account`, `get_access_key`, etc.). So we enable this mode only during funcall processing, because all trie touches are charged there. It allows us to imply that cache chunk size is limited by ~31K.

> How will `TrieCachingStorage` be used outside of `apply_transactions_with_optional_storage_proof`?

Not sure I understood the question. `TrieCachingStorage` is created in the `get_trie_for_shard`, and is removed when the call ends. So its lifetime is similar to `apply_transactions_with_optional_storage_proof` call.

core/store/src/trie/trie_storage.rs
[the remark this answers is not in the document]
mm-near: sure you can have it there -- but I'd also add it here too..

(as you want to have 'clear' behaviors of your classes - and here one method allows large objects and the other one doesn't - which is confusing)

core/store/src/trie/trie_storage.rs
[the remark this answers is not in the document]
Longarithm: > Instead of use set_state, how about we use start_chunk_cache and finish_chunk_cache to indicate the state(mode) transition. I think it's clearer that way for what the switch is for and where these methods should be called. 

I agree, but with the current architecture we need to create "copies" of these methods in `RuntimeExt` and `TrieUpdate` to access the state from the outside. I think the current approach minimizes the introduced ugliness.

> And you also need to clear the chunk cache in finish_chunk_cache

I don't. This cache stores nodes for the whole chunk, so following receipts could spend less gas. It is done intentionally, and AFAIK it was discussed on some protocol meeting.

core/store/src/trie/trie_storage.rs
Longarithm: It is safe, because each key is a value hash, so there can be only one possible value for the given hash.

On the other hand, this change was not intentional and was caused by sequence of refactoring. I agree that it is confusing here, so I returned acquiring lock here again.

core/store/src/trie/trie_storage.rs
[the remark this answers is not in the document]
Longarithm: Put the explanation to `retrieve_raw_bytes`, because there is no need in `TrieCache::put` anymore.

core/store/Cargo.toml
```diff
@@ -23,6 +23,7 @@ tracing = "0.1.13"
 borsh = "0.9"
 thiserror = "1"
 lru = "0.7.2"
+assert_matches = "1.5.0"
```
matklad, line 26: This needs to be a dev-dependency?

core/store/src/trie/trie_storage.rs
```diff
@@ -133,31 +144,48 @@ impl TrieStorage for TrieMemoryPartialStorage {
...
     pub(crate) shard_uid: ShardUId,
 
+    /// Caches ever requested items for the shard `shard_uid`. Used to speed up DB operations, presence of any item is
+    /// not guaranteed.
+    pub(crate) shard_cache: TrieCache,
+    /// Caches all items requested in the mode `TrieCacheMode::CachingChunk`. It must be empty when we start to apply
+    /// txs and receipts in the chunk. All items placed here must remain until applying txs/receipts ends.
+    /// Note that for both caches key is the hash of value, so for the fixed key the value is unique.
+    /// TODO (#5920): enable chunk nodes caching in Runtime::apply.
+    pub(crate) chunk_cache: RefCell<HashMap<CryptoHash, Arc<[u8]>>>,
```
mzhangmzz, line 172: When are you going to clear the `chunk_cache`?

mzhangmzz (approved):

mm-near (approved):

Longarithm:
Nayduck run https://nayduck.near.org/#/run/2424

PolyProgrammist:
This code is a great contribution to the Near code, Near core and the Near company. I vote to double the salary for the author of this code.

PR #6415, opened 2022-03-11
added panic=abort to the whole workspace

pompon0:
Given that our storage-related code is crash-safe, we should be able to set panic=abort so that no panic goes unnoticed.
AFAICT the test targets are ignoring this setting:
https://github.com/alexcrichton/cargo/blob/master/src/cargo/core/profiles.rs#L114
therefore our "should_panic" tests still work (at least on my workstation).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -64,11 +64,16 @@ exclude = [ "neard" ]
 
 [patch.crates-io]
 
+# Note that "bench" profile inherits from "release" profile and
+# "test" profile inherits from "dev" profile.
+# https://doc.rust-lang.org/cargo/reference/profiles.html#test
+
 [profile.release]
 overflow-checks = true
+panic = 'abort'
 
-[profile.bench]
-overflow-checks = true
+[profile.dev]
+panic = 'abort'
 
 [profile.dev.package.hex]
 opt-level = 3 # BLS library is too slow to use in debug


Cargo.toml
```diff
@@ -61,11 +61,16 @@ exclude = [ "neard" ]
 
 [patch.crates-io]
 
+# Note that "bench" profile inherits from "release" profile and
+# "test" profile inherits from "dev" profile.
+# https://doc.rust-lang.org/cargo/reference/profiles.html#test
+
 [profile.release]
 overflow-checks = true
+panic = 'abort'
```
bowenwang1996, line 70: I don't think we should do it for release. Nearcore should tolerate crashes in some cases. For example if one jsonrpc thread panics or if one peer actor thread panics it should not cause neard to shut down completely.

pompon0: If panic occurs, then it means that there is a bug in the implementation and it is dangerous to run incorrect code. A failed jsonrpc should return an error, not panic. If an actor panics silently, you end up with a crippled process which is very hard to debug.

bowenwang1996:
> Implementation wise this LGTM from me as a `runtime/` stakeholder. Please have other stakeholders (especially people working on the storage stack, jsonrpc, etc.) look at this as well.
> 
> I generally agree that panicking should always remain a way to indicate failure to uphold critical invariants and any use of `panic!` that's otherwise considered recoverable is a bug. That said from my memory nearcore is pretty trigger happy with `unwrap`s and `panic!`s today so I'm not sure if e.g. the RPC code routinely observes recoverable panics.

Good point, but I think we should make sure that the panics we have in nearcore are intended before enabling this.

pompon0:
> Implementation wise this LGTM from me as a runtime/ stakeholder. Please have other stakeholders (especially people working on the storage stack, jsonrpc, etc.) look at this as well.

FYI I've postponed moving forward with this PR (but it is not abandoned) until I have a reasonable amount of confidence that the codebase doesn't rely on recoverable panics, i.e.
1) a panic won't corrupt the database
2) failed RPCs won't panic.

mm-near (approved):
LGTM - but let's announce it more widely, so that other folks are aware.

PR #6419, opened 2022-03-11
Metrics for execution time per Petagas

### LINKED ISSUE #6302: Wall clock time / gas histogram
As per https://near.zulipchat.com/#narrow/stream/297873-nearinc.2FnodeX/topic/chunk.20gas.20graphs/near/272146187

We should have histogram metrics that log wall clock time divided by gas with buckets centered around 1ms/1Tgas.

See also https://github.com/near/nearcore/pull/6268 for how to implement the metric itself.

nikurt:
As suggested by @Ekleog and @matklad , this PR adds metrics that correlate the amount of Tgas with the execution time.
Fix #6302 

The PR adds two metrics:
1) Execution time per 1000 Tgas.
2) Histogram of execution time for chunks with gas usage in buckets [0,100 Tgas), [100 Tgas, 200 Tgas), etc.

The metrics are useful because:
1) Gives us an overview of how well the execution time corresponds to the gas charged.
2) Provides insight into execution time of full chunks vs half-full chunks vs empty chunks, with the resolution of 100 Tgas per chunk.

Tested on betanet.

Dump of metrics from one node: https://gist.github.com/nikurt/062c54aecbf47d968a4bef7e7e9b15cb
Metric "execution seconds per 1000 tgas" for betanet: 
image


nearcore/src/metrics.rs
```diff
@@ -0,0 +1,22 @@
...
+        Some(prometheus::linear_buckets(0.0, 0.05, 50).unwrap()),
+    )
+    .unwrap()
+});
+
+pub static SECS_PER_THOUSAND_TGAS: Lazy<HistogramVec> = Lazy::new(|| {
+    try_create_histogram_vec(
+        "near_execution_secs_per_thousand_tgas",
+        "Execution time per teragas. Ignore label 'label'.",
+        &["label"],
```
nagisa, line 18: What is the purpose of this `label`? If it needs to be here, consider adding a comment?

nearcore/src/metrics.rs
```diff
@@ -0,0 +1,22 @@
...
+        Some(prometheus::linear_buckets(0.0, 0.05, 50).unwrap()),
```
nagisa, line 19: Is linear the best bucket distribution to use here? I suspect more bucket density around 1 second would help investigating problems around situations where the relationship doesn't hold just barely, and once we're looking at numbers like 2 seconds or more per Pgas, it is going to be pretty obvious things are bad no matter what, right?

nearcore/src/metrics.rs
```diff
@@ -0,0 +1,22 @@
...
+        "Time process a chunk. Tgas used by the chunk is a metric label.",
+        &["tgas_ceiling"],
+        Some(prometheus::linear_buckets(0.0, 0.05, 50).unwrap()),
+    )
+    .unwrap()
+});
+
+pub static SECS_PER_THOUSAND_TGAS: Lazy<HistogramVec> = Lazy::new(|| {
+    try_create_histogram_vec(
+        "near_execution_secs_per_thousand_tgas",
```
nagisa, line 16:
```suggestion
        "near_execution_seconds_per_petagas",
```

maybe?

---

[Metric&Label naming best practices](https://prometheus.io/docs/practices/naming/) also suggest that metrics should be using base units (which in this case would require splitting this histogram into two), but in our case computing the data in the runtime is the only way to actually get meaningful measurements out, so idk if its a good idea.

nearcore/src/runtime/mod.rs
```diff
@@ -562,9 +563,18 @@ impl NightshadeRuntime {
                 RuntimeError::ReceiptValidationError(e) => panic!("{}", e),
                 RuntimeError::ValidatorError(e) => e.into(),
             })?;
+        let elapsed = instant.elapsed();
 
         let total_gas_burnt =
             apply_result.outcomes.iter().map(|tx_result| tx_result.outcome.gas_burnt).sum();
+        metrics::APPLY_CHUNK_DELAY
+            .with_label_values(&[&format_total_gas_burnt(total_gas_burnt)])
```
nagisa, line 571: Won't this result in an explosion in cardinality, since the label value is going to be pretty much always different every time a chunk is processed? I believe it would effectively cause the prometheus instance to stop behaving quite quickly, and neard would also appear to leak memory since it'd need to store additional data in memory for each chunk processed.

nearcore/src/metrics.rs
nikurt: Done

nearcore/src/metrics.rs
nikurt: Done

nearcore/src/metrics.rs
nikurt: Added a comment

nearcore/src/metrics.rs
```diff
@@ -0,0 +1,28 @@
...
+        &["tgas_ceiling"],
+        Some(prometheus::linear_buckets(0.0, 0.05, 50).unwrap()),
+    )
+    .unwrap()
+});
+
+pub static SECS_PER_THOUSAND_TGAS: Lazy<HistogramVec> = Lazy::new(|| {
+    try_create_histogram_vec(
+        "near_execution_seconds_per_petagas_ratio",
+        "Execution time per teragas. Ignore label 'label'.",
```
nagisa, line 17: The description here also needs an update.

nearcore/src/metrics.rs
nagisa: Huh, that's a curious and needless limitation, which I didn't encounter before. Sounds like there's nothing stopping us from removing the label in the future, though.

nearcore/src/metrics.rs
```diff
@@ -0,0 +1,28 @@
+use near_metrics::{try_create_histogram_vec, HistogramVec};
+use once_cell::sync::Lazy;
+
+pub static APPLY_CHUNK_DELAY: Lazy<HistogramVec> = Lazy::new(|| {
+    try_create_histogram_vec(
+        "near_apply_chunk_delay",
```
nagisa, line 6: This metric name should contain a unit in it.

nearcore/src/runtime/mod.rs
nagisa: 13 labels is a very reasonable amount, I didn't notice the rounding, mostly because I missed the function being added.

nearcore/src/metrics.rs
nikurt: Actually, let me take that back. An empty list of labels works.

nearcore/src/metrics.rs
nikurt: Done

nearcore/src/metrics.rs
nikurt: Done

nagisa (approved):

nearcore/src/runtime/mod.rs
```diff
@@ -562,9 +563,18 @@ impl NightshadeRuntime {
...
+            .observe(elapsed.as_secs_f64());
+        if total_gas_burnt > 0 {
+            metrics::SECONDS_PER_PETAGAS
+                .with_label_values(&[])
```
matklad, line 575:
```suggestion
```

matklad (approved):

mm-near (approved):

nearcore/src/runtime/mod.rs
Ekleog: I'm late to the party, but this would probably be possible to resolve by replacing try_create_histogram_vec in the creation of this metric by try_create_histogram :)

PR #6420, opened 2022-03-12
pytest: add default timeout to utils.wait_for_blocks

mina86:
Calculate default timeout to the utils.wait_for_blocks function.  The
timeout is based on number of blocks the function waits for.  Having
the default allows us to simplify tests which no longer need to
specify explicit timeout.

### DIFF
--- a/pytest/tests/sanity/sync_chunks_from_archival.py
+++ b/pytest/tests/sanity/sync_chunks_from_archival.py
@@ -94,19 +94,19 @@ class Handler(ProxyHandler):
 # TODO(mina86): Make it a utility class
 class Timeout:
 
-    def __init__(self, timeout: float) -> None:
-        if timeout <= 0:
-            raise ValueError('timeout must be positive')
+    def __init__(self, sesconds: float) -> None:
+        if sesconds <= 0:
+            raise ValueError('sesconds must be positive')
         self.__start = time.monotonic()
-        self.__end = self.__start + timeout
+        self.__end = self.__start + sesconds
 
     def check(self) -> bool:
         return time.monotonic() < self.__end
 
-    def elapsed(self) -> bool:
+    def elapsed_seconds(self) -> float:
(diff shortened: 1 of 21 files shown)


pytest/tests/sanity/sync_chunks_from_archival.py
```diff
@@ -91,6 +91,25 @@ async def handle(self, msg, fr, to):
...
+    def __init__(self, timeout: float) -> None:
+        if timeout <= 0:
+            raise ValueError('timeout must be positive')
+        self.__start = time.monotonic()
+        self.__end = self.__start + timeout
+
+    def check(self) -> bool:
+        return time.monotonic() < self.__end
+
+    def elapsed(self) -> bool:
```
mm-near, line 106: change the return type

pytest/tests/sanity/sync_chunks_from_archival.py
```diff
@@ -91,6 +91,25 @@ async def handle(self, msg, fr, to):
...
+        return time.monotonic() - self.__start
+
+    def left(self) -> bool:
```
mm-near, line 109: same here

pytest/tests/sanity/sync_chunks_from_archival.py
mm-near, line 106: mention that this is in seconds

pytest/tests/sanity/sync_chunks_from_archival.py
```diff
@@ -91,6 +91,25 @@ async def handle(self, msg, fr, to):
         return True
 
 
+# TODO(mina86): Make it a utility class
+class Timeout:
+
+    def __init__(self, timeout: float) -> None:
```
mm-near, line 97: timeout is in seconds, right?

mm-near (approved):

PR #6427, opened 2022-03-14
fix(script): Fix params-estimator init in CI

jakmeier:
The old way of initializing a neard home does not install the test
contract on all accounts, which leads to estimations failing.
Giving an empty directory to the estimator as home will cause it to
initialize it properly.


matklad (approved):

matklad:
:heavy_check_mark: to fix CI, but I hope we can soon remove this completely in favor of continuous estimation. Would this be a good idea @jakmeier?

jakmeier:
Yes we can replace it soon. I currently think it will also live in buildkite in a similar fashion. But in any case, this script will be obsolete.

PR #6429, opened 2022-03-14
Estimator speed

jakmeier:
Some estimations took >1h to finish in qemu.
This change speeds them up by reducing unreasonable repetitions per
iteration.
(Total repetitions = iters_per_block * block_size * inner_repetitions)

- Deployment per byte: 100 -> 5 block size
- UTF8 / UTF16 decoding per byte: 10k -> 1k inner repetitions
    
The new time is around 5min per block for all 3 collectively.


runtime/runtime-params-estimator/src/lib.rs
```diff
@@ -569,7 +569,9 @@ fn deploy_contract_cost(
         let actions = vec![Action::DeployContract(DeployContractAction { code: code_factory() })];
         tb.transaction_from_actions(sender, receiver, actions)
     };
-    let total_cost = transaction_cost(testbed, &mut make_transaction);
+    // Use a small block size since deployments are gas heavy
```
matklad, line 572:
```suggestion
    // Use a small block size since deployments are gas heavy.
```

matklad (approved):
Awesome! This should make pretty much any other change to the estimator much easier!

PR #6431, opened 2022-03-14
chain: ShardManager: add metrics for partial encoded chunk response

mina86:
Add near_partial_encoded_chunk_request_processing_time metric which
returns how much time processing partial encoded chunk requests took.
The metric is split by the method used to create a response and also
whether in the end the response has been prepared or not.


chain/chunks/src/metrics.rs
```diff
@@ -0,0 +1,20 @@
+use once_cell::sync::Lazy;
+
+pub static PARTIAL_ENCODED_CHUNK_REQUEST_PROCESSING_TIME: Lazy<near_metrics::HistogramVec> =
+    Lazy::new(|| {
+        near_metrics::try_create_histogram_vec(
+            "near_partial_encoded_chunk_request_processing_time",
+            concat!(
+                "Time taken to prepare responses to partial encoded chuck ",
+                "requests.  The ‘method’ key describes how we tried to fulfil ",
```
mm-near, line 9: typo in fulfill

chain/chunks/src/lib.rs
```diff
@@ -999,32 +1000,62 @@ impl ShardsManager {
...
-        if let Some(response) = response {
-            self.peer_manager_adapter.do_send(PeerManagerMessageRequest::NetworkRequests(
-                NetworkRequests::PartialEncodedChunkResponse { route_back, response },
-            ))
+            let response =
+                self.prepare_partial_encoded_chunk_response_from_chunk(request, rs, chunk);
+            return (started, "chunk", response);
         }
+
+        (Instant::now(), "none", None)
```
mm-near, line 1058: this means that the entry is missing?

mm-near (approved):

chain/chunks/src/lib.rs
mina86: Yep.

PR #6433, opened 2022-03-15
refactor(params-estimator): Safe GasCost sub

jakmeier:
Replace GasCost - GasCost subtraction with a safe subtraction.
The safe subtraction does not panic if the result of components
is negative but instead marks the result as uncertain.
This is important for ICount based measurements, where IO bytes
are often close to 0 with some variance.


jakmeier:
I hope I caught all occurrences of unsafe subtraction... It did run through successfully once on both glcoud and my machine but it has always been failing inconsistently.

runtime/runtime-params-estimator/src/gas_cost.rs
```diff
@@ -104,6 +104,18 @@ impl GasCost {
...
     }
+
+    /// Subtracts two gas costs from each other without panicking on an arithmetic underflow.
+    /// If the given tolerance is breached, the result will be marked as uncertain.
+    pub(crate) fn safe_subtract(&self, rhs: &Self, tolerance: &NonNegativeTolerance) -> Self {
+        let mut pos = self.saturating_sub(rhs);
+        let neg = rhs.saturating_sub(self);
+        if !tolerance.tolerates(&pos, &neg) {
+            pos.set_uncertain(true);
+            pos.set_uncertain(true);
```
matklad, line 115:
```suggestion
            pos.set_uncertain(true);
```

runtime/runtime-params-estimator/src/gas_cost.rs
```diff
@@ -104,6 +104,18 @@ impl GasCost {
             }
         }
     }
+
+    /// Subtracts two gas costs from each other without panicking on an arithmetic underflow.
+    /// If the given tolerance is breached, the result will be marked as uncertain.
+    pub(crate) fn safe_subtract(&self, rhs: &Self, tolerance: &NonNegativeTolerance) -> Self {
```
matklad, line 110: Let's call this `saturating_sub` maybe?

matklad (commented):
Hm, I don't quite like that now you need to think whether you should use `-` or `safe_subtract`... This choice depends on non-local knowledge. 

What about the following alternative:

* we change the usual `-` to always saturate and mark the cost as uncertain
* we add a `reason: &'static str` to the uncertain costs, so that you can see that the cost is uncertain *because* it overflowed. Can even get fancy here and use `#[track_caller]` to show *where* this happened.

runtime/runtime-params-estimator/src/gas_cost.rs
jakmeier: Wouldn't that lead to confusion with the existing `impl num_traits::SaturatingSub for GasCost`?

matklad: As an aside, I'd just make that `saturating_sub` an inherent impl: using num_traits here feels like a case of [concrete abstraction](https://matklad.github.io/2020/08/15/concrete-abstraction.html)

jakmeier: Agreed on removing `num_traits` abstraction. And I think the calls to the old `saturating_sub` can be replaced with calls to the new function that merges `saturating_sub` and `safe_subtract`.

For the other change, I am not so sure. Personally, I don't like hiding important details that are not obvious in the function signature. The uncertain flag is a very important details, if it is set the entire estimation is essentially null and void.

Saturating subtraction or `ops::Sub` assuming a strict tolerance seems wrong, IO bytes as we measure it today will always have cases that underflow sometimes. Using 0.1% as default hard-coded tolerance would be ok but hard-coding such numbers smells to me.

I have two suggestions, either to swallow complexity or magic:

1. Complexity: Get rid of `ops::Sub` and embrace the fact that `GasCost` cannot be subtracted like a primitive. All subtractions should use `saturating_sub()` with a chosen tolerance.
2. Magic: `saturating_sub` has a tolerance argument and `ops::Sub` calls that with a default value of 0.1%. This way, choosing `-` has some magic hidden behind it but at least is never really wrong. Estimations functions MAY opt-in to use an explicit tolerance by calling `saturating_sub()` if their local knowledge requires it. (For example, when no tolerance is desired)

WDYT?

jakmeier:
> * we add a `reason: &'static str` to the uncertain costs, so that you can see that the cost is uncertain _because_ it overflowed. Can even get fancy here and use `#[track_caller]` to show _where_ this happened.

Independent of what we decide on the other discussion, big thumps up on this suggestion!

runtime/runtime-params-estimator/src/gas_cost.rs
matklad: >Saturating subtraction or ops::Sub assuming a strict tolerance seems wrong, I

Good point. I think there's also 3:

Make `ops::Sub` *strict* (always set uncertain on underflow), but also add `tolerance_sub` or something which allows to *not* set uncertain if it's close enough.

I don't have strong preference among 1, 2, 3 -- all seems workable and refactorabe. I suggest picking one and seeing how it goes

jakmeier: I decide for 3, I like strict `-` better than some hard-coded arbitrary number.

jakmeier:
The new uncertain flag shows up like this:

```
ActionSirReceiptCreation                                16_706_250 gas [        0.00i 4.95r 0.00w] UNCERTAIN (computed in 29.782716908s) SUBTRACTION-UNDERFLOW: runtime/runtime-params-estimator/src/lib.rs:286:36
```
It's getting a bit ugly IMO. But it's primarily debug output and the new information added is very useful at that.

Also, some scripts for continuous estimation use that output because, at least for now. Those search for `UNCERTAIN` and remain compatible with this change.

matklad:
>Also, some scripts for continuous estimation use that output, at least for now. Those search for UNCERTAIN and remain compatible with this change.

Ouch, that seems something we need to fix before it grows up. We should add json output in addition to debug output. That's why this into is printed to `stderr` -- so that we can have JSON output to stdout at the same time.

runtime/runtime-params-estimator/src/gas_cost.rs
```diff
@@ -32,8 +33,9 @@ pub(crate) struct GasCost {
     ///
     /// Each specific cost can use it's own criteria for uncertainty -- the end
     /// result here is just printing UNCERTAIN next to the corresponding cost in
-    /// the output.
-    uncertain: bool,
+    /// the output. `uncertain_message` can be called to display the reason and
+    /// code location of where the uncertainty has been set.
+    uncertain: Option<(&'static str, &'static Location<'static>)>,
```
matklad, line 38: It's probably worth it to add a named struct here, though no strong opinion.

runtime/runtime-params-estimator/src/gas_cost.rs
```diff
@@ -149,14 +191,16 @@ impl LeastSquaresTolerance {
...
-enum NonNegativeTolerance {
+pub(crate) enum NonNegativeTolerance {
     /// Allow no negative values
     Strict,
     /// Tolerate negative values if it changes the total gas by less than X times.
     RelativeTolerance(f64),
     /// Tolerate negative values if they are below X gas
     AbsoluteTolerance(Gas),
 }
+pub(crate) const PER_MILLE_TOLERANCE: NonNegativeTolerance =
```
matklad, line 202: might be an associated const of NonNegativeTolerance enum?

runtime/runtime-params-estimator/src/gas_cost.rs
matklad: No opinion -- renaming types is easy, so I tend to just rename things as I go, when I realize a better name is possible.

PR #6434, opened 2022-03-15
fuzz: disable runtime-tester for now as it's got lots of intermittent failures anyway

Ekleog:
This way we can scale up the fuzzing infrastructure for the other fuzz targets without getting overwhelmed by this fuzzer's intermittent failures that look mostly independent from its inputs.

### DIFF
--- a/nightly/fuzz.toml
+++ b/nightly/fuzz.toml
@@ -17,9 +17,11 @@ weight = 10
 flags = ["-len_control=0", "-prefer_small=0", "-max_len=4000000"]
 
 [[target]]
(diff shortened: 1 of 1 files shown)


nagisa:
Please file issues for the distinct failure modes before landing this.

matklad (approved):

Ekleog:
Aaah you reacted too fast before my push adding the reference to the github issue as a comment completed :sweat_smile: 

I'll just add a reminder to the issue to not forget reenabling the fuzzer then :)

PR #6437, opened 2022-03-15
Proper rocksdb shutdown if many run_actix threads are running

mm-near:


mm-near:
passing nayduck: https://nayduck.near.org/#/run/2422

nikurt (approved):

test-utils/actix-test-utils/src/lib.rs
```diff
@@ -64,8 +68,15 @@ pub fn spawn_interruptible<F: std::future::Future + 'static>(
     actix_rt::spawn(handle_interrupt!(f))
 }
 
+// Number of actix instances that are currently running.
+pub(crate) static ACTIX_INSTANCES_COUNTER: Lazy<Mutex<usize>> = Lazy::new(|| (Mutex::new(0)));
```
mina86, line 72: Just use `AtomicUsize`.

mm-near: I tried - but then this must be a 'mutable' static variable (as I need access to 'get_mut' - to make sure that I hold the lock while I'm cleaning up the rocksdb handles)

bowenwang1996: Atomic variables have interior mutability. You can use `load` to read and `store` to update it without requiring the variable to be mutable

PR #6439, opened 2022-03-15
chain: gc partial chunks on archival nodes

mina86:
Start garbage collecting ColPartialChunks and ColInvalidChunks on
archival nodes.  The former is quite sizeable column and its data can
be recovered from ColChunks.  The latter is only needed when operating
at head.

Note that this is likely insufficient for the garbage collection to
happen in reasonable time (since with current default options we’re
garbage collecting only two heights at a time).  It’s best to clean
out the two columns.


chain/chain/src/store.rs
```diff
@@ -1983,39 +1983,50 @@ impl<'a> ChainStoreUpdate<'a> {
...
-                let chunk = self.get_chunk(&chunk_hash)?.clone();
-                debug_assert_eq!(chunk.cloned_header().height_created(), height);
-                for transaction in chunk.transactions() {
-                    self.gc_col(ColTransactions, &transaction.get_hash().into());
-                }
-                for receipt in chunk.receipts() {
-                    self.gc_col(ColReceipts, &receipt.get_hash().into());
+                if !archival {
+                    let chunk = self.get_chunk(&chunk_hash)?.clone();
+                    debug_assert_eq!(chunk.cloned_header().height_created(), height);
```
matklad, line 1997: drive by, but

```suggestion
                    debug_assert_eq!(chunk.height_created(), height);
```

also, it feels like, rather than `debug_assert`, we should generally be using 

```
if !cond {
  if cfg(debug_assertions) { panic!() } else { tracing::error!() }
}
```

pattern, but this goes well beyond the scope of the PR

matklad (commented):
Absence of tests is concerning. I guess it's the usual "we don't have infra to test this conveniently", but the logic here looks quite finicky...

mina86:
> Absence of tests is concerning. I guess it's the usual "we don't have infra to test this conveniently", but the logic here looks quite finicky...

Yeah, I’ve tested this manually and pytest/tests/sanity/block_sync_archival.py checks that the node is able to fill in those requests but there is currently no test. I am contemplating a few approaches so I am planning to add something.

chain/chain/src/store.rs
```diff
@@ -1983,39 +1983,50 @@ impl<'a> ChainStoreUpdate<'a> {
...
 
                 // 2. Delete chunk_hash-indexed data
-                let chunk_header_hash = chunk_hash.clone().into();
-                self.gc_col(ColChunks, &chunk_header_hash);
+                let chunk_header_hash = chunk_hash.into();
+                if !archival {
+                    self.gc_col(ColChunks, &chunk_header_hash);
+                }
                 self.gc_col(ColPartialChunks, &chunk_header_hash);
                 self.gc_col(ColInvalidChunks, &chunk_header_hash);
```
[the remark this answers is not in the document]
mina86: > Seems that we are gc'ing ColInvalidChunks here too for archival node, is that on purpose?

Yes.  I figured we might just as well garbage collect that column as well. It’s contents isn’t needed in the archive as far as I can tell. It’s only used during syncing.

mina86:
@mzhangmzz, PTAL

bowenwang1996 (commented):
I am not very comfortable that we don't have a unit test for this. It feels to me that should be relatively easy to do with `TestEnv`

mina86:
> 

It’s kinda tested a bit in the Python test, at the moment so at least there’s that.

chain/chain/src/store.rs
```diff
@@ -2028,6 +2028,43 @@ impl<'a> ChainStoreUpdate<'a> {
...
+            for chunk_hash in chunk_hashes {
+                let chunk_header_hash = chunk_hash.into();
+                self.gc_col(ColPartialChunks, &chunk_header_hash);
+                // Data in ColInvalidChunks isn’t technically redundant (as it
+                // cannot be calculated from any other data) but it is data we
+                // don’t need for anything so it can be deleted as well.
+                self.gc_col(ColInvalidChunks, &chunk_header_hash);
+            }
+        }
+        self.update_chunk_tail(stop);
```
mm-near, line 2064: so here we actually start 'updating' chunk_tail for archival nodes, right?

are we 'abusing' it a little? shouldn't we use a separate 'pointer' for our garbage collection here? 

(as in theory we're removing only the partial chunks -- the real chunks are still there, so in theory the chunk_tail shouldn't have changed)

mm-near (dismissed):

chain/chain/src/store.rs
```diff
@@ -2028,6 +2028,43 @@ impl<'a> ChainStoreUpdate<'a> {
...
+    ///
+    /// `gt_height_limit` indicates limit of how many heights to process.  This
+    /// limit means that the method may stop garbage collection before reaching
+    /// `gc_stop_height`.
+    pub fn clear_redundant_chunk_data(
+        &mut self,
+        gc_stop_height: BlockHeight,
+        gc_height_limit: BlockHeightDelta,
+    ) -> Result<(), Error> {
+        let start = self.chunk_tail()?;
```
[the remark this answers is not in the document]
mina86: Nope. I don’t think it’s going to be a useful metric in the long run. There may be some fluctuation at the beginning when archival node starts garbage collecting partial chunks but then it’s going to be always five epochs behind head.

chain/chain/src/store.rs
mina86: I mean, I wouldn’t call it abuse, just repurposing. I’m not really sure if it’s worth adding a new pointer. I’ve talked with Bowen about turning of archival setting and we agreed that this is essentially unsupported operation so here we can just reuse the pointer. And in the future, e.g. with cold storage,n we’d need to make some custom migration step anyway at which point we’d know that data from genesis till chunk_tail is only partially garbage collected.

chain/chain/src/store.rs
```diff
@@ -2028,6 +2028,43 @@ impl<'a> ChainStoreUpdate<'a> {
...
+        let stop = gc_stop_height.min(start.saturating_add(gc_height_limit));
+        for height in start..stop {
```
mm-near, line 2053: in other GCs - we keep going until we gc exactly 'height_limit' blocks.
Here you  pick 'fixed' limits. (so if there were no blocks produced at this height - it means that we don't GC anything in this round).

Was this intended? if yes - maybe let's quickly mention it in a comment - as this is different behaviour than other GC.

chain/chain/src/store.rs
mm-near: > about turning off archival setting

this might be a nice starter project -- making sure that we 'fail' with a nice error if someone tries changing this config setting.

pytest/lib/cluster.py
```diff
@@ -466,9 +466,17 @@ def start(self, *, boot_node: BootNode = None, skip_starting_proxy=False):
...
-    def kill(self):
+    def kill(self, *, gentle=False):
+        """Kills the process.  If `gentle` sends SIGINT before killing."""
         if self._proxy_local_stopped is not None:
             self._proxy_local_stopped.value = 1
+        if self._process and gentle:
+            self._process.send_signal(signal.SIGINT)
+            try:
+                self._process.wait(5)
+                self._process = None
```
mm-near, line 477: but does it mean that we don't send a kill in such case?

pytest/tests/sanity/block_sync_archival.py
```diff
@@ -97,11 +94,63 @@ def __exit__(self, *_):
...
+        A `{key: count}` dictionary where key is in ‘method/success’ format.
+        The values correspond to the
+        near_partial_encoded_chunk_request_processing_time_count Prometheus
+        metric.
+    """
+    url = 'http://{}:{}/metrics'.format(*node.rpc_addr())
+    response = requests.get(url)
+    response.raise_for_status()
+
+    metric_name = 'near_partial_encoded_chunk_request_processing_time'
```
mm-near, line 117: I'm surprised that this is not an argument of the function (but I guess you'll change it once you refactor?)

I'd also generalize this method a lot more (to simply return a dict of all metrics)

pytest/tests/sanity/block_sync_archival.py
```diff
@@ -111,76 +160,51 @@ def get_all_blocks(node: cluster.BaseNode, *,
...
-            counts.setdefault(key, 0)
-        for key, count in counts.items():
-            if key in keys:
-                assert count, f'Expected {key} counter to be non-zero'
-            else:
-                assert not count, (
-                    f'Expected {key} counter to be zero but got {count}')
+def wait_for_node_to_sync(node: cluster.BaseNode,
+                          blocks: typing.Sequence[cluster.BlockId]) -> None:
+    """Waits for block to sync and asserts it sees the same blocks."""
```
mm-near, line 165: this function is empty? is it used anywhere?

mm-near (approved):

chain/chain/src/store.rs
mina86: https://github.com/near/nearcore/issues/6522

pytest/tests/sanity/block_sync_archival.py
mina86: Yes, that would be the idea. In this function it’s not worth generalising on the metric name since the function cannot even handle any other metric. Once this is refactored (as per #6458), we’d need a more general code which can handle any kind of metrics.

chain/chain/src/store.rs
mina86: I’ve changed it a bit so that we don’t count empty heights. This is more in line with how we treat blocks.

Issue #6432, opened 2022-03-15
Update to clap 3.1


mina86:
Simple search and replace in Cargo.toml is not enough to update clap. Newer version gives use ‘hide’ option which can be used for deprecated commands namely unsafe_reset_data, unsafe_reset_all and testnet.

Issue #4764, opened 2021-09-01
Migrate off clap2


matklad:
In nearcore, a bunch of crates uses `clap3`, and a bunch of crates uses `clap2`. As neard itself uses `clap3`, I think we should migrate everything else to it. The specific benefits are:

* no confusion about what to use for new tools
* better compile times (clap is pretty big, two claps are twice as big)
* easier to maintain code -- clap3 has derive API which makes adding or changing cli options much easier.

bings47:
Hi, would like to try my hand at this issue if it's still open for contribution.

bowenwang1996:
@bings47 sure. Assigned it to you

bings47:
should I also bump the v3 version out of the beta version and into the latest one? Seems to be [3.1.6](https://github.com/clap-rs/clap/tree/v3.1.6) now

matklad:
@bings47 probably makes sense to

* first upgrade all betas to proper v3
* then remove the remaining usages of v2

PR #6441, opened 2022-03-16
chore: migrate clap to v3

binggh:
Bump packages using clap v3 beta to use the latest version ‘3.1.6’ and
migrate packages still using v2 to ‘3.1.6’.

With that done, [`hide`](https://docs.rs/clap/latest/clap/struct.Arg.html#method.hide)
deprecated commands testnet, unsafe_reset_all and unsafe_reset_data as
discussed in #6432.

And while at it, for consistency rename ‘recompress-storage’ command
to ‘recompress_storage’.

Fixes: https://github.com/near/nearcore/issues/4764
Fixes: https://github.com/near/nearcore/issues/6432


integration-tests/Cargo.toml
```diff
@@ -17,7 +17,7 @@ actix-rt = "2"
 base64 = "0.11"
 borsh = "0.9"
 chrono = { version = "0.4.4", features = ["serde"] }
-clap = "=3.0.0-beta.2"
+clap ="3.1.6"
```
matklad, line 20:
```suggestion
clap = "3.1.6"
```

tools/indexer/example/Cargo.toml
```diff
@@ -9,7 +9,7 @@ edition = "2021"
 
 [dependencies]
 actix = "=0.11.0-beta.2"
-clap = "3.0.0-beta.1"
+clap ="3.1.6"
```
matklad, line 12:
```suggestion
clap = "3.1.6"
```

tools/restored-receipts-verifier/Cargo.toml
```diff
@@ -10,7 +10,7 @@ edition = "2021"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
-clap = "3.0.0-beta.1"
+clap ="3.1.6"
```
matklad, line 13:
```suggestion
clap = "3.1.6"
```

nagisa (approved):

Cargo.lock
```diff
@@ -3527,7 +3533,7 @@ dependencies = [
  "mime",
  "once_cell",
  "paperclip-macros",
- "parking_lot 0.10.2",
+ "parking_lot 0.11.2",
```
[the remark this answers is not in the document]
binggh: might've unintentionally bumped this dependency, is there a way to selectively roll this back?

binggh:
Added [hide](https://docs.rs/clap/latest/clap/struct.Arg.html#method.hide) to deprecated commands `testnet`, `unsafe_reset_all`, `unsafe_reset_data` in [14617bc](https://github.com/near/nearcore/pull/6441/commits/14617bc5630cafb5320b99e119789f756a7694c1) and [c7b298e](https://github.com/near/nearcore/pull/6441/commits/c7b298ef18177dd4b91489e25bc12f0be08c0bf0)

as discussed in #6432

integration-tests/Cargo.toml
binggh: Thanks @matklad , fixed formatting issues

neard/src/cli.rs
```diff
@@ -123,27 +123,28 @@ pub(super) enum NeardSubCommand {
...
+    #[clap(name = "unsafe_reset_data", hide(true))]
     UnsafeResetData,
     /// View DB state.
-    #[clap(name = "view_state")]
+    #[clap(subcommand, name = "view_state")]
     StateViewer(StateViewerSubCommand),
     /// Recompresses the entire storage.  This is a slow operation which reads
     /// all the data from the database and writes them down to a new copy of the
     /// database.
+    #[clap(name = "recompress_storage")]
```
mina86, line 143: This changes the name of the subcommand. I prefer dash to be honest but I guess this makes senses for consistency (alternative would be to rename view state to `view-state`) but this rename should be mentioned in commit message.

neard/src/cli.rs
```diff
@@ -123,27 +123,28 @@ pub(super) enum NeardSubCommand {
     /// DEPRECATED: this command has been renamed to 'localnet' and will be removed in a future
     /// release.
     // TODO(#4372): Deprecated since 1.24.  Delete it in a couple of releases in 2022.
-    #[clap(name = "testnet")]
+    #[clap(name = "testnet", hide(true))]
```
mina86, line 126: I was under the impression that `hide` is enough, i.e. that `(true)` is not necessary. Same with `forbid_empty_values`, `arg_required_else_help` etc.

neard/src/cli.rs
```diff
@@ -123,27 +123,28 @@ pub(super) enum NeardSubCommand {
...
+    #[clap(name = "unsafe_reset_all", hide(true))]
     UnsafeResetAll,
     /// (unsafe) Remove all the data, effectively resetting node to the genesis state (keeps genesis and
     /// config).
-    #[clap(name = "unsafe_reset_data")]
+    #[clap(name = "unsafe_reset_data", hide(true))]
     UnsafeResetData,
     /// View DB state.
-    #[clap(name = "view_state")]
+    #[clap(subcommand, name = "view_state")]
```
mina86, line 138: Why does this one get `subcommand` but others don’t?

mina86 (approved):

neard/src/cli.rs
mina86: OK, I took the liberty of editing the commit message a bit.

neard/src/cli.rs
binggh: `hide`, `forbid_empty_values`, `arg_required_else_help` are Command methods and their syntax has to be in either of the two formats:

```
method = arg
```
or

```
method(arg1, arg2)
```

https://github.com/clap-rs/clap/blob/v3.1.6/examples/derive_ref/README.md#terminology

mina86: kk, sg, though in that case for consistency with `name` I’d go for `hide = true` myself.

jakmeier (approved):

neard/src/cli.rs
binggh: correct me if i'm wrong but from my understand it seems like only state viewer is a _true_ subcommand, because from within state viewer you can call more commands from `StateViewerSubCommand` enum, versus for example `RecompressStorageSubCommand` where you only provide the  `--output-dir`. Perhaps it's more of a discussion of the naming/semantics(?)

neard/src/cli.rs
binggh: got it - will update changes to the `method = arg` format for consistency

khorolets (approved):

neard/src/cli.rs
mina86: > from my understand it seems like only state viewer is a true subcommand, because from within state viewer you can call more commands

Yes, that’s correct. If that’s the distinction for what is a ‘subcommand’ that that answers my question.

Cargo.lock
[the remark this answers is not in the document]
binggh: hmm, not sure how i bumped this accidentally, is there a way to revert this alone?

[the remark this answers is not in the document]
nagisa: `cargo update -p parking_lot --precise '=0.10.2'` might work?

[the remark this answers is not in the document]
nagisa: Got it, fine to leave as is in that case.

mm-near (approved):

nikurt (approved):

PR #6446, opened 2022-03-16
RocksDB stats re-exported as Prometheus metrics

nikurt:
RocksDB collects a wide range of statistics, currently 231, which provide a lot of insight into the rate of DB operations, cache hit ratio, sizes of DB operations and some timing information.
This PR parses and re-exports all available RocksDB statistics as Prometheus metrics.

Unfortunately, the stats level cannot be changed using the Rust RocksDB API.

Tested by running this change on a betanet node:
https://grafana.near.org/d/OphWx5Enk/rocksdb-db-timings?orgId=1&var-chain_id=betanet&var-node_id=.*
https://grafana.near.org/d/wRf6xcE7k/rocksdb-block-cache?orgId=1&var-chain_id=betanet&var-node_id=.*


core/store/src/db.rs
```diff
@@ -794,6 +804,12 @@ fn rocksdb_options() -> Options {
     opts.set_bytes_per_sync(bytesize::MIB);
     opts.set_write_buffer_size(256 * bytesize::MIB as usize);
     opts.set_max_bytes_for_level_base(256 * bytesize::MIB);
+    // Rust API doesn't permit choosing stats level. The default stats level is
+    // `kExceptDetailedTimers`, which is described as:
+    // "Collects all stats except time inside mutex lock AND time spent on compression.".
+    opts.enable_statistics();
+    opts.set_stats_persist_period_sec(0);
+    opts.set_stats_dump_period_sec(60);
```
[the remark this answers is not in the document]
nikurt: Probably not, removed.

chain/client/src/rocksdb_metrics.rs
```diff
@@ -0,0 +1,156 @@
...
+                }
+            }
+            result.push((stats_name, values));
+        }
+    }
+    Ok(result)
+}
+
+fn get_prometheus_metric_name(stats_name: &str) -> String {
+    format!("near_{}", stats_name.replace(".", "_"))
```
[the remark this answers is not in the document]
nikurt: That would be outside of scope of the PR. Would be better to update `core/metrics/src/lib.rs` to use a custom registry. We don't need two custom registries with the same prefix.

chain/client/src/info.rs
```diff
@@ -231,6 +237,19 @@ impl InfoHelper {
         };
         telemetry(&self.telemetry_actor, content);
     }
+
+    fn export_rocksdb_statistics(&mut self, statistics: Option<String>) {
+        if let Some(statistics) = statistics {
+            match parse_statistics(&statistics) {
```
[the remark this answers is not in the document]
nikurt: Let's address this in a future refactoring.

chain/client/src/rocksdb_metrics.rs
```diff
@@ -0,0 +1,156 @@
...
+                // - near_a_b_c
+                if let StatsValue::Count(value) = values[0] {
+                    let entry = self.int_gauges.entry(stats_name.to_string());
+                    entry
+                        .or_insert_with(|| {
+                            try_create_int_gauge(
+                                &get_prometheus_metric_name(stats_name),
+                                stats_name,
+                            )
+                            .unwrap()
```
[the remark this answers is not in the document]
nikurt: Updated.

chain/client/src/rocksdb_metrics.rs
```diff
@@ -0,0 +1,156 @@
...
+) -> Result<Vec<(&str, Vec<StatsValue>)>, anyhow::Error> {
+    let mut result = vec![];
+    // Statistics are given one per line.
+    for line in statistics.split('\n') {
+        let mut values = vec![];
+        let words: Vec<&str> = line.split(' ').collect();
+        // Each line follows one of two formats:
+        // 1) <stat_name> COUNT : <value>
+        // 2) <stat_name> P50 : <value> P90 : <value> COUNT : <value> SUM : <value>
+        // Each line gets split into words and we parse statistics according to this format.
```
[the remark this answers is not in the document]
nikurt: Thanks

chain/client/src/rocksdb_metrics.rs
```diff
@@ -0,0 +1,156 @@
...
+                        words[i][1..].parse::<u32>().map_err(|err| anyhow::anyhow!(err))?,
+                        words[i + 2].parse::<f64>().map_err(|err| anyhow::anyhow!(err))?,
+                    ));
+                } else {
+                    return Err(anyhow::anyhow!(
+                        "Unsupported stats value: {} in {}",
+                        words[i],
+                        line
+                    ));
+                }
```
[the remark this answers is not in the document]
nikurt: Thanks

chain/client/src/rocksdb_metrics.rs
```diff
@@ -0,0 +1,156 @@
...
+                    values.push(StatsValue::Count(
+                        words[i + 2].parse::<i64>().map_err(|err| anyhow::anyhow!(err))?,
+                    ));
+                } else if words[i] == "SUM" {
+                    values.push(StatsValue::Sum(
+                        words[i + 2].parse::<i64>().map_err(|err| anyhow::anyhow!(err))?,
+                    ));
+                } else if words[i].starts_with("P") {
+                    values.push(StatsValue::Percentile(
+                        words[i][1..].parse::<u32>().map_err(|err| anyhow::anyhow!(err))?,
```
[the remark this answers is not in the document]
nikurt: Done

bowenwang1996 (approved):
Are we confident that this change wouldn't visibly affect the performance of a node?

mm-near:
we could have it until a config flag - and then start by enabling only on couple of our nodes first.

core/store/src/db.rs
```diff
@@ -794,6 +804,13 @@ fn rocksdb_options() -> Options {
     opts.set_bytes_per_sync(bytesize::MIB);
     opts.set_write_buffer_size(256 * bytesize::MIB as usize);
     opts.set_max_bytes_for_level_base(256 * bytesize::MIB);
+    // Rust API doesn't permit choosing stats level. The default stats level is
+    // `kExceptDetailedTimers`, which is described as:
+    // "Collects all stats except time inside mutex lock AND time spent on compression.".
+    opts.enable_statistics();
+    // Disabling dumping stats to filse because the stats are exported to Prometheus.
```
mm-near, line 811: s/filse/files

core/store/src/lib.rs
```diff
@@ -151,6 +151,10 @@ impl Store {
     pub fn get_rocksdb(&self) -> Option<&RocksDB> {
         self.storage.as_rocksdb()
     }
+
+    pub fn get_rocksdb_statistics(&self) -> Option<String> {
```
mm-near, line 155: what is the format of the string?

chain/chain/src/store.rs
```diff
@@ -761,6 +761,10 @@ impl ChainStore {
             }
         })
     }
+
+    pub fn get_rocksdb_statistics(&self) -> Option<String> {
```
mm-near, line 765: do we need this method on this level?

if yes -- we're 'leaking' the implementation details ('rocksdb') - which will make it harder in the future to use different storage system.

core/store/src/lib.rs
mm-near: maybe rather than string, we should create out own struct? (even if it starts as a string wrapper at the beginning)

mm-near (changes requested):

mm-near:
IMHO - this PR exposes a little bit too much 'internals' of our classes - (for example it adds a rocksdb specific function to chain).  
Very soon we might have 2 rocksdb instances (and/or other storage systems) - so let's think how we could adjust the interfaces (maybe we should compute those stats at the bottom layer and just expose the map<string, string>? ) to be ready for this.

core/store/src/lib.rs
matklad: +1

we actually do create our own sturuct (see `StatsValue` and `parse_statistics`), it’s just that it happens couple of layers above. 

It’s better to parse strings in structured data as early as possible instead.

chain/chain/src/store.rs
matklad: +1, specifically, this leaks two things:

* name rocks (fixable by renaming to get_store_statisrics)
* the specific string format (fixable by parsing at the rocks layer and returning a `struct StoreStats(Map<String, String>)` )

chain/client/src/rocksdb_metrics.rs
```diff
@@ -0,0 +1,162 @@
...
+    gauges: HashMap<String, GaugeVec>,
+}
+
+impl RocksDBMetrics {
+    pub fn export_stats_as_metrics(
+        &mut self,
+        stats: &[(&str, Vec<StatsValue>)],
+    ) -> anyhow::Result<()> {
+        for (stat_name, values) in stats {
+            if values.len() == 1 {
```
matklad, line 20: Can fold the two conditions I think:

`if let [StatsValue::Count(value)] = values.as_slice()`

chain/client/src/rocksdb_metrics.rs
```diff
@@ -0,0 +1,162 @@
...
+                    }
+                }
+            }
+        }
+        Ok(())
+    }
+
+    fn set_int_value<F1, F2>(
+        &mut self,
+        key_fn: F1,
```
matklad, line 83: We always immediately call F1, we can just pass in a string instead.

chain/client/src/rocksdb_metrics.rs
```diff
@@ -0,0 +1,162 @@
...
+        metric_fn: F2,
```
matklad, line 84: Generic lambdas are a “death by a thousand cuts” way to increase compile times due to monomorphisation. In this case, we can use monomorphic fn pointer instead:

metric_fn: fn(&str) -> String

chain/client/src/rocksdb_metrics.rs
```diff
@@ -0,0 +1,162 @@
...
+    where
+        F1: Fn(&str) -> String,
+        F2: Fn(&str) -> String,
+    {
+        let key: &str = &key_fn(stat_name);
+        if !self.int_gauges.contains_key(key) {
+            self.int_gauges
+                .insert(key.to_string(), try_create_int_gauge(&metric_fn(stat_name), stat_name)?);
+        }
+        self.int_gauges.get(key).unwrap().set(value);
```
matklad, line 97: Here and elsewhere, please use the entry API:

https://doc.rust-lang.org/std/collections/hash_map/struct.HashMap.html#method.entry

it gets rid of unwraps and cuts down on the number of duplicate map lookiups.

chain/client/src/rocksdb_metrics.rs
```diff
@@ -0,0 +1,162 @@
...
+    Count(i64),
+    Sum(i64),
+    Percentile(u32, f64),
+}
+
+/// Parses a string containing RocksDB statistics.
+pub(crate) fn parse_statistics(statistics: &str) -> anyhow::Result<Vec<(&str, Vec<StatsValue>)>> {
+    let mut result = vec![];
+    // Statistics are given one per line.
+    for line in statistics.split('\n') {
```
matklad, line 113:
```suggestion
    for line in statistics.lines() {
```

chain/client/src/rocksdb_metrics.rs
```diff
@@ -0,0 +1,162 @@
...
+                match key {
+                    "COUNT" => values.push(StatsValue::Count(val.parse::<i64>()?)),
+                    "SUM" => values.push(StatsValue::Sum(val.parse::<i64>()?)),
+                    p if p.starts_with("P") => values.push(StatsValue::Percentile(
+                        key[1..].parse::<u32>()?,
+                        val.parse::<f64>()?,
+                    )),
+                    _ => {
+                        return Err(anyhow::anyhow!(
+                            "Unsupported stats value: {} in {}",
```
matklad, line 131:
```suggestion
                            "Unsupported stats value: {key} in {line}",
```

Or is it the case that anyhow doesn’t yet supports this syntax?

chain/client/src/rocksdb_metrics.rs
```diff
@@ -0,0 +1,162 @@
...
+use prometheus::{GaugeVec, IntGauge};
+use std::collections::HashMap;
+
+#[derive(Default)]
+/// Creates prometheus metrics on-demand for exporting RocksDB statistics.
+pub(crate) struct RocksDBMetrics {
+    // Contains counters and sums, which are integer statistics in RocksDB.
+    int_gauges: HashMap<String, IntGauge>,
+    // Contains floating point statistics, such as quantiles of timings.
+    gauges: HashMap<String, GaugeVec>,
```
matklad, line 11: I wonder if these hash maps should be static singletons? my understanding that it’s only possible to register one counter with a specific name per process (prometheus register thing), so the current code looks like a lie: it seems that one can Crete independent instances of this struct, but this isn’t the case.

chain/chain/src/store.rs
nikurt: Done

chain/client/Cargo.toml
```diff
@@ -11,6 +11,7 @@ edition = "2021"
 ansi_term = "0.12"
 actix = "=0.11.0-beta.2"
 actix-rt = "2"
+anyhow = "1.0.51"
```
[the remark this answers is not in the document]
nikurt: Done

chain/client/src/info.rs
```diff
@@ -161,6 +167,9 @@ impl InfoHelper {
             paint(ansi_term::Colour::Green, blocks_info_log),
             paint(ansi_term::Colour::Blue, machine_info_log),
         );
+        if let Err(err) = self.export_rocksdb_statistics(statistics) {
+            warn!("Failed to export RocksDB statistics: {:?}", err);
```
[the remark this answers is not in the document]
nikurt: Done

chain/client/src/info.rs
```diff
@@ -231,6 +240,19 @@ impl InfoHelper {
         };
         telemetry(&self.telemetry_actor, content);
     }
+
+    fn export_rocksdb_statistics(&mut self, statistics: Option<String>) -> anyhow::Result<()> {
+        if let Some(statistics) = statistics {
+            parse_statistics(&statistics).context("Failed to parse rocksdb statistics").and_then(
```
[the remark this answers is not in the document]
nikurt: Done

chain/client/src/rocksdb_metrics.rs
nikurt: Done

chain/client/src/rocksdb_metrics.rs
nikurt: Done.
Can't use `or_insert_with()`, because `try_create_int_gauge()` can fail.

chain/client/src/rocksdb_metrics.rs
nikurt: Done

chain/client/src/rocksdb_metrics.rs
nikurt: Done

nikurt:
@bowenwang1996 documentation mentions a non-negligible impact on performance.
@mm-near added a config parameter, but had to add it to every call of `create_store()`. Please let me know if you have a better interface suggestion.

chain/client/src/info.rs
```diff
@@ -100,6 +101,7 @@ impl InfoHelper {
         validator_epoch_stats: Vec<ValidatorProductionStats>,
         epoch_height: EpochHeight,
         protocol_upgrade_block_height: BlockHeight,
+        statistics: Option<StoreStatistics>,
```
matklad, line 104: Nice collateral benefit of strong types with `statistics: Option<String>` it wasn’t clear what the stats are, and now it is!

chain/client/src/rocksdb_metrics.rs
```diff
@@ -0,0 +1,151 @@
...
+    match ROCKSDB_METRICS.lock().unwrap().export_stats_as_metrics(stats) {
+        Ok(_) => {}
+        Err(err) => {
+            warn!(target:"stats", "Failed to export store statistics: {:?}",err);
+        }
+    }
+}
+
+/// Wrapper for re-exporting RocksDB stats into Prometheus metrics.
+pub(crate) static ROCKSDB_METRICS: Lazy<Mutex<RocksDBMetrics>> =
```
matklad, line 20: Probably can be private.

(I usually don’t care about `pub(crate)`, but for global *state* I do)

chain/client/src/rocksdb_metrics.rs
```diff
@@ -0,0 +1,151 @@
...
+                // A summary stats.
+                // A statistic 'a.b.c' creates the following prometheus metrics:
+                // - near_a_b_c_sum
+                // - near_a_b_c_count
+                // - near_a_b_c{quantile="0.95"}
+                for stats_value in values {
+                    match stats_value {
+                        StatsValue::Count(value) => {
+                            self.set_int_value(
+                                |stat_name| get_stats_summary_count_key(stat_name),
```
matklad, line 60:
```suggestion
                                get_stats_summary_count_key,
```


Just this might work here and elsewher

chain/client/src/rocksdb_metrics.rs
```diff
@@ -0,0 +1,151 @@
...
+                                |stat_name| get_metric_name_summary_sum_gauge(stat_name),
+                                &stat_name,
+                                value,
+                            )?;
+                        }
+                        StatsValue::Percentile(percentile, value) => {
+                            let key = &stat_name;
+
+                            let entry = self.gauges.entry(key.to_string());
+                            match entry {
```
matklad, line 78: It seems this can be compressed:

```
let gauge = match entry {
  Vacant(entry) => entry.insert(create()?),
  Occupied(entry) => entry.get_mut(),
};
gauge.with_label_values…
```

core/store/src/db.rs
```diff
@@ -547,9 +549,13 @@ impl RocksDBOptions {
     }
 
     /// Opens the database in read/write mode.
-    pub fn read_write<P: AsRef<std::path::Path>>(self, path: P) -> Result<RocksDB, DBError> {
+    pub fn read_write<P: AsRef<std::path::Path>>(
+        self,
+        path: P,
+        enable_statistics: bool,
```
matklad, line 555: `RocksDBOptions` is a builder object, it exists so that we don’t have to add dozens of raw Boolean arguments. let’s add an 

   pub fn enable_statistics(mut self, yea: bool) -> Self


Builder method instead

core/store/src/db.rs
```diff
@@ -882,8 +915,11 @@ impl RocksDB {
         RocksDBOptions::default().read_only(path)
     }
 
-    pub fn new<P: AsRef<std::path::Path>>(path: P) -> Result<Self, DBError> {
-        RocksDBOptions::default().read_write(path)
+    pub fn new<P: AsRef<std::path::Path>>(
+        path: P,
+        enable_statistics: bool,
```
matklad, line 920: I’d probably keep `new` as a method which creates db with default options, and let the callers use RocksDBOptions API explicitly if they want more control. 

It kinda doesn’t make sense to have both options builder and a bunch of parameters to new

core/store/src/lib.rs
```diff
@@ -295,8 +300,8 @@ pub fn read_with_cache<'a, T: BorshDeserialize + 'a>(
     Ok(None)
 }
 
-pub fn create_store(path: &Path) -> Store {
-    let db = Arc::new(RocksDB::new(path).expect("Failed to open the database"));
+pub fn create_store(path: &Path, enable_statistics: bool) -> Store {
```
matklad, line 303: Adding an `false` to the dozens of call sites doesn’t feel great. Perhaps we need api shaped like this: 

```
/// Creates store with default config.
/// Use this if you don’t know what you are doing.
pub fn create_store(path: &Path)

/// Opens store with extra control knobs
pub fn create_store_conf(path: &Path, confit: StoreConfig)

#[derive(Default)]
pub struct StoreConfig {
  pub read_only: bool,
  pub enable_stats: bool,
}

File: core/store/src/db.rs (hunk 25 above)
Comment (nikurt) [reply core/store/src/db.rs]: Good idea, I misunderstood the idea of `RocksDBOptions`.

File: core/store/src/db.rs (hunk 26 above)
Comment (nikurt) [reply core/store/src/db.rs]: Done

File: core/store/src/lib.rs (hunk 27 above)
Comment (nikurt) [reply core/store/src/lib.rs]: Done, this is indeed cleaner

**Review by jakmeier — APPROVED** [review]

**Review by matklad — APPROVED** [review]

File: core/store/src/lib.rs [hunk 28]
```diff
@@ -300,10 +305,22 @@ pub fn create_store(path: &Path) -> Store {
...
 }
 
-/// Creates a store which is unable to modify an existing RocksDB instance.
-/// Panics if a write operation is attempted.
-pub fn open_read_only_store(path: &Path) -> Store {
-    let db = Arc::new(RocksDB::new_read_only(path).expect("Failed to open the database"));
+#[derive(Default, Debug)]
+pub struct StoreConfig {
+    pub read_only: bool,
+    pub enable_statistics: bool,
```
mm-near, line 311: mention if this has any impact on performance

core/store/src/db.rs
```diff
@@ -1100,3 +1142,45 @@ mod tests {
...
+    Sum(i64),
+    Percentile(u32, f64),
+}
+
+pub struct StoreStatistics {
+    pub data: Vec<(String, Vec<StatsValue>)>,
+}
+
+/// Parses a string containing RocksDB statistics.
+fn parse_statistics(statistics: &str) -> Result<StoreStatistics, Box<dyn std::error::Error>> {
```
mm-near, line 1158: do we have a unittest for this?

chain/client/src/rocksdb_metrics.rs
```diff
@@ -0,0 +1,133 @@
...
+                                .set(value);
+                        }
+                    }
+                }
+            }
+        }
+        Ok(())
+    }
+
+    fn set_int_value(
```
mm-near, line 96: comment - how does key, metric and stat_name differ

mm-near (approved):

core/store/src/db.rs
nikurt: Added

core/store/src/lib.rs
nikurt: Done

Issue #6119, opened 2022-01-19
Mitigate archival node data size growth


bowenwang1996:
Context can be found <span class="error">[here]</span>([https://near.zulipchat.com/#narrow/stream/308695-nearinc.2Fprivate/topic/archival.20node.20data.20size](https://near.zulipchat.com/#narrow/stream/308695-nearinc.2Fprivate/topic/archival.20node.20data.20size)). TLDR: archival node data usage grows at a very fast pace and since we require node operators to use SSDs to operate an archival node, it creates a considerable amount of financial burden for them, which is highly undesirable and unsustainable. Some ideas to mitigate the growth include the following:

*   Move most of the archival node data usage to HDD and only use SSD for the head of the chain so that an archival node can keep up with the network. This could be implemented through garbage collection where instead of removing data, we write archival nodes' data to a separate storage.
*   Do not store data that can be reconstructed for archival nodes. For example, partial encoded chunks can be reconstructed from full chunks and we do not need to store them for archival nodes (behind what is needed for block processing). This alone can potentially save us 1/3 of the storage needed. When a request comes in for old partial encoded chunks, we can compute them on the fly.
*   (Longer-term) Create data snapshots periodically and do not store full history on archival nodes. The snapshot can be stored in a decentralized storage and its metadata can be posted on-chain. We can develop a separate service that serves historical data by routing its query to the relevant snapshot (effective shard historical data by epoch)

bowenwang1996:
cc @mina86 @posvyatokum

PR #6447, opened 2022-03-17
neard: add recompress-storage command

mina86:
Add recompress-storage command which reads all the data from storage
and saves it in a new location effectively recompressing all the SST
files.  This is meant as a temporary tool to speed up migration to
zstd compression.  Without it, especially on archival nodes,
recompressing data may take a long time as most SST files can stay
untouched for epochs.


nearcore/src/lib.rs
```diff
@@ -498,3 +499,88 @@ pub fn start_with_config_and_synchronization(
         arbiters: vec![client_arbiter_handle, arbiter.handle()],
     })
 }
+
+pub fn recompress_storage(home_dir: &Path, dst_dir: &Path) -> anyhow::Result<()> {
+    use strum::{EnumCount, IntoEnumIterator};
+
+    // We’re configuring each RocksDB to use 512 file descriptors.  Make sure we
+    // can open that many by ensuring nofile limit is large enough to give us
+    // some room to spare over 1024 file descriptors.
```
EdvardD, line 508: Probably it makes sense to add the command to the helper info of the command: `ulimit -n 1024`

nearcore/src/lib.rs
```diff
@@ -498,3 +499,88 @@ pub fn start_with_config_and_synchronization(
         arbiters: vec![client_arbiter_handle, arbiter.handle()],
     })
 }
+
+pub fn recompress_storage(home_dir: &Path, dst_dir: &Path) -> anyhow::Result<()> {
```
EdvardD, line 503: What about extracting this to a separate file so lib.rs doesn't grow huge?

EdvardD (approved):

nearcore/src/lib.rs
mina86: The file isn’t actually that large and my intention with this function is that we’ll remove it in a couple of releases once everyone who needed to use it used it. All new users won’t need to perform the recompression.

nearcore/src/lib.rs
```diff
@@ -498,3 +499,90 @@ pub fn start_with_config_and_synchronization(
...
+    // some room to spare over 1024 file descriptors.
+    let (soft, hard) = rlimit::Resource::NOFILE
+        .get()
+        .map_err(|err| anyhow::anyhow!("getrlimit: NOFILE: {}", err))?;
+    if soft < 3 * 512 {
+        rlimit::Resource::NOFILE
+            .set(3 * 512, hard)
+            .map_err(|err| anyhow::anyhow!("setrlimit: NOFILE: {}", err))?;
+    }
+
```
bowenwang1996, line 517: If 512 is a config value why do we hard code it here?

neard/src/cli.rs
```diff
@@ -132,6 +137,10 @@ pub(super) enum NeardSubCommand {
     /// View DB state.
     #[clap(name = "view_state")]
     StateViewer(StateViewerSubCommand),
+    /// Recompresses the entire storage.  This is a slow operation which reads
+    /// all the data from the database and writes them down to a new copy of the
+    /// database.
+    RecompressStorage(RecompressStorageSubCommand),
```
bowenwang1996, line 143: Why does it not have `#[clap(name = ...)]` :)

nearcore/src/lib.rs
mina86: It’s hard coded in `db.rs` in `rocksdb_options`:

```rust
/// DB level options
fn rocksdb_options() -> Options {
    let mut opts = Options::default();

    set_compression_options(&mut opts);
    opts.create_missing_column_families(true);
    opts.create_if_missing(true);
    opts.set_use_fsync(false);
    opts.set_max_open_files(512);  // <------ here
    opts.set_bytes_per_sync(bytesize::MIB);
    opts.set_write_buffer_size(256 * bytesize::MIB as usize);
    opts.set_max_bytes_for_level_base(256 * bytesize::MIB);
    // ...

    opts
}
```

It should be an option actually, but that’s a separate story.

neard/src/cli.rs
mina86: It’s not necessary. Clap is smart enough to convert it to `recompress-storage`.

nearcore/src/lib.rs
bowenwang1996: > As to why not make it a constant I honestly intend for the recompress storage command to be temporary (as in just couple of releases to let anyone who cares recompress) so admittedly I am cutting corners a little.

What is the reason for not keeping it permanent? Is it because recompression would happen eventually anyways?

mina86: The recompression would happen eventually but I don’t think that would happen in reasonable time.  I haven’t tested it and could be wrong but to recompress the whole database all of the SST files would need to be touched and that means a lot of compactions through all the levels.

It’s more that I hope that in six months there will be no use for this code. Anyone who starts a new archival node in six months will use the database with new compression (our snapshots will be recompressed and their node will use the new compression). And anyone who is running archival node will use this command to recompress their storage (or start anew from our snapshot depending on their trust requirements).

And even more than that, this code to me feels like a hack prepared just for the sake of the compression migration rather than a command that belongs to neard.

neard/src/cli.rs
```diff
@@ -132,6 +137,10 @@ pub(super) enum NeardSubCommand {
     /// View DB state.
     #[clap(name = "view_state")]
     StateViewer(StateViewerSubCommand),
+    /// Recompresses the entire storage.  This is a slow operation which reads
+    /// all the data from the database and writes them down to a new copy of the
+    /// database.
```
mm-near, line 142: also mention why we have it (because we changed compression) - and that we plan to remove it in Q4/2022

neard/src/cli.rs
```diff
@@ -454,6 +463,21 @@ fn init_logging(verbose: Option<&str>) {
         .init();
 }
 
+#[derive(Clap)]
+pub(super) struct RecompressStorageSubCommand {
```
mm-near, line 467: I'm wondering if we should it put it inside nearcore? or have a separate tool to do only this.

mm-near (approved):

neard/src/cli.rs
matklad: -1 on a separate tool, I feel it’s overall better to have a single Swiss Army knife binary:

* (much) faster linking times
* easier incident investigation, no need to download extra tools
* better project structure: it’s easier to notice a dead module in an actively worked upon crate than to note a dead crate.

PR #6448, opened 2022-03-17
feature(params-estimator): JSON output

jakmeier:
Optional JSON output of estimation results for scripts to consume

### DIFF
--- a/runtime/runtime-params-estimator/src/gas_cost.rs
+++ b/runtime/runtime-params-estimator/src/gas_cost.rs
@@ -166,7 +166,7 @@ impl GasCost {
                 "instructions": self.instructions.to_f64(),
                 "io_r_bytes": self.io_r_bytes.to_f64(),
                 "io_w_bytes": self.io_w_bytes.to_f64(),
-                "uncertain": self.uncertain.is_some(),
+                // `None` will be printed as `null`
                 "uncertain_reason": self.uncertain.map(|u| u.reason),
             }),
             GasMetric::Time => json!({
--- a/runtime/runtime-params-estimator/src/gas_cost.rs
+++ b/runtime/runtime-params-estimator/src/gas_cost.rs
@@ -6,6 +6,7 @@ use std::{fmt, ops};
 use near_primitives::types::Gas;
 use num_rational::Ratio;
 use num_traits::ToPrimitive;
+use serde_json::json;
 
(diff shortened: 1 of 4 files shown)


jakmeier:
Example output:
```json
LogBase                                                441_061_948 gas [                    441ns]           (computed in 826.929296ms) 
{"computed_in":{"nanos":826929296,"secs":0},"name":"LogBase","result":{"gas":441061948,"metric":"time","time_ns":441.061948,"uncertain":false,"uncertain_reason":null}}
LogByte                                                  2_743_748 gas [                      2ns]           (computed in 983.235753ms) 
{"computed_in":{"nanos":983235753,"secs":0},"name":"LogByte","result":{"gas":2743748,"metric":"time","time_ns":2.7437486640625,"uncertain":false,"uncertain_reason":null}}
Utf8DecodingBase                                       327_640_298 gas [                    327ns]           (computed in 829.946438ms) 
{"computed_in":{"nanos":829946438,"secs":0},"name":"Utf8DecodingBase","result":{"gas":327640298,"metric":"time","time_ns":327.640298,"uncertain":false,"uncertain_reason":null}}
```

runtime/runtime-params-estimator/src/gas_cost.rs
```diff
@@ -153,6 +154,30 @@ impl GasCost {
...
+    pub fn to_json(&self) -> serde_json::Value {
+        match self.metric {
+            GasMetric::ICount => json!({
+                "gas": self.to_gas(),
+                "metric": "icount",
+                "instructions": self.instructions.to_f64(),
+                "io_r_bytes": self.io_r_bytes.to_f64(),
+                "io_w_bytes": self.io_w_bytes.to_f64(),
+                "uncertain": self.uncertain.is_some(),
+                "uncertain_reason": self.uncertain.map(|u| u.reason),
```
matklad, line 170: feels like this can be just one field perhaps? No strong opinion though

matklad (approved):

runtime/runtime-params-estimator/src/gas_cost.rs
jakmeier: Yes that's actually nicer, it removes unnecessary redundancy from the output and streamlines how script should use it. I changed it so we only have `uncertain_reason` which is `null` if the result is not uncertain.

PR #6450, opened 2022-03-17
feat: activate chunk nodes cache on nightly

Longarithm:
Make nightly protocol feature with activated chunk nodes cache introduced in https://github.com/near/nearcore/pull/6414.

**Overview**

For each node, for which we charge touching trie node cost, we want to do it only once per chunk - on the first access. Motivation is that accessing nodes from memory is much faster than querying DB for this, and we see that it gives ~87% touching trie node cost reduction and ~30% overall cost reduction for Aurora.

This is achieved by inserting these nodes into `chunk_cache` inside `TrieCachingStorage`, which is automatically dropped during finishing chunk processing. We enable such caching by switching to `TrieCacheMode::CachingChunk` mode, and disable it by switching back to `TrieCacheMode::CachingShard` mode. This is currently done only during funcalls, because we don't actually charge for nodes accessed in other places.
  
Note that we don't have a size limit for chunk cache items. Overall size of chunk cache can't exceed 150 MB (see https://github.com/near/nearcore/pull/6450#issuecomment-1072375572).

**Test plan**
 
- [x] `process_blocks::chunk_nodes_cache_test` - send transaction modifying contract storage before / after protocol upgrade, check that number of touched nodes is lower.
- [x] `test_chunk_nodes_cache_across_receipts_runtime` - similarly, apply 3 receipts and check that cost of last 2 receipts is lower.
- [x] Run Aurora transactions and check costs
- [x] Same, but also check chunk cache size


matklad:
In #6414 I said

>That is, chunk cache would be at most 170 megs, which seems acceptable

There's an important correction: I *think* I've made a mistake in that calculation -- each key's byte adds **two** nodes, so we are looking at 340 megs here, which makes me mildly uncomfortable. (admitedly, both "acceptable" and "mildly uncomfortable" are gut feelings for me -- I don't have a great understanding of our memeory usage).

Could someone independently verify memory usage implications here.

matklad:
I remember that, when we discussed this, @EgorKulikov mentioned that we should have some kind of cost even for cached nodes. Did we explicitly decide one way or another?

Or, to put it in the more actionable form, what are the costs/performance for the contract which first writes, and than repeatedly reads a 4mb key?

cc @jakmeier you might potentially help here.

integration-tests/src/tests/standard_cases/mod.rs
```diff
@@ -1348,15 +1362,78 @@ pub fn test_contract_write_key_value_cost(node: impl Node) {
...
+            node_touches = receipt_hashes
+                .iter()
+                .map(|receipt_hash| {
+                    let result = node_user.get_transaction_result(receipt_hash);
+                    let touching_trie_node_cost = get_touching_trie_node_cost(&result.metadata);
+                    touching_trie_node_cost
+                        / runtime_config.wasm_config.ext_costs.touching_trie_node
+                })
+                .collect();
+        }
```
mm-near, line 1435: as we are 'disabling' the cache directly after each runner call:

 do we have a test that checks that the cost of accessing the nodes 'outside' of the runtime is still counted?

Longarithm: Hm, could you clarify what 'outside' means?
We don't charge these costs outside of `Runtime` or `NightshadeRuntime` (and even `action_function_call` scope)

integration-tests/src/tests/client/process_blocks.rs
```diff
@@ -4511,3 +4511,162 @@ mod contract_precompilation_tests {
         assert!(caches[1].get(&contract_key.0).unwrap().is_none());
     }
 }
+
```
EdvardD, line 4514: Can you please add some bugs (one by one in most interesting places) in the chunk_cache_nodes and make sure that those integration tests break?

EdvardD, line 4514: Also it would be interesting to take a look on chunk cache code coverage running those tests.

EdvardD: You can also try to fully disable chunk caching to see if those tests break.

Longarithm: This is interesting. Do you know a tool for that?

Longarithm:
> There's an important correction: I think I've made a mistake in that calculation -- each key's byte adds two nodes, so we are looking at 340 megs here, which makes me mildly uncomfortable

Good question @matklad, let me be more precise here.

We store two different types of items: `RawTrieNodeWithSize` and values, DB key for each of them is the value hash.
`RawTrieNodeWithSize` contains some data of limited size and sometimes a `Vec<u8>`.

At first, estimate limited size data. In the worst case, when node is a branch, we store `memory_usage`, 16 `Option<CryptoHash>` and 1 `Option<(u32, CryptoHash)>`. The total size is `8 + 16 * 40 + 4 + 8 + 36 = 696`. Because we can access only 31K nodes in a chunk, it gives us a limit of 31 KB * 700 ~= 22 MB.

Now estimate total size of `Vec`s in nodes. Each such vector is a part of storage key we are reading, multiplied by two because we store nibbles instead of bytes. Thus, it is limited by total size of all storage keys accessed in a chunk. `storage_read_key_byte` is ~30 MGas, so we have a limit 2 * 500 Tgas / 30 Mgas ~= 34 MB.

Finally, `storage_read_value_byte` helps us to limit the total size of all values by 500 Tgas / 5.5 Mgas ~= 91 MB.

In total, this gives a limit of ~150 MB per chunk.

I would be happy if someone could double-check this :) @mm-near

integration-tests/src/tests/client/process_blocks.rs
EdvardD: I guess this should help with that although I didn't try it out yet (particularly `A source-based code coverage`):
https://doc.rust-lang.org/nightly/rustc/instrument-coverage.html

(It's requires nightly rustc)

Longarithm:
> I remember that, when we discussed this, @EgorKulikov mentioned that we should have some kind of cost even for cached nodes. Did we explicitly decide one way or another?
>
> Or, to put it in the more actionable form, what are the costs/performance for the contract which first writes, and than repeatedly reads a 4mb key?

I changed params-estimator with manually added `TouchingTrieNodeFromChunkCache` cost: https://github.com/Longarithm/nearcore/compare/0185519...53798b7. Then I made two runs with/without enabled feature, took a delta of computed costs and divided them by number of nodes fell into cache. So I got **~1.5 Ggas saved per cached node**.

Comparing to `TouchingTrieNodeRead` cost from the same estimator run which is **~1.72 Ggas**, I conclude that the cost of reading node from cache is **0.22 Ggas**.

Given:
* "touching_trie_node_cached" cost is 72x times lower than current cost which is 16 Ggas;
* "touching_trie_node_cached" cost is dominated by `storage_read_base` cost which is 56 Ggas - Aurora measurements showed that trie depth of their keys is in `[7, 13]` range
* urgency of this feature

I would proceed without adding a separate cost for it. However, it should be reconsidered when we lower storage costs further. @bowenwang1996 

Example of command I run:

```
cargo run --release --package runtime-params-estimator --features nightly_protocol,protocol_feature_chunk_nodes_cache,required --bin runtime-params-estimator -- --accounts-num 20000 --additional-accounts-num 200000 --iters 1 --warmup-iters 1 --docker --metric icount --costs TouchingTrieNodeFromChunkCache
```

mzhangmzz (commented):
I don't see where the cache is cleared between blocks. Could you point out where that is?

mzhangmzz:
> I don't see where the cache is cleared. Could you point out where that is?

Oh I see, it's because the get_trie_for_shard function will create a new TrieCachingStorage, which creates a new ChunkCache. Could you add comments for this?

runtime/runtime/src/actions.rs
```diff
@@ -99,21 +99,30 @@ pub(crate) fn execute_function_call(
...
+            &config.wasm_config,
+            &config.transaction_costs,
+            promise_results,
+            apply_state.current_protocol_version,
+            apply_state.cache.as_deref(),
+        )
+    };
+
+    if checked_feature!("protocol_feature_chunk_nodes_cache", ChunkNodesCache, protocol_version) {
+        runtime_ext.set_trie_cache_mode(TrieCacheMode::CachingChunk);
```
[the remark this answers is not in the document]
mzhangmzz: Now I'm confused, why do you need to explicitly restrict the use of cache chunk to function calls?

mzhangmzz:
> > There's an important correction: I think I've made a mistake in that calculation -- each key's byte adds two nodes, so we are looking at 340 megs here, which makes me mildly uncomfortable
> 
> Good question @matklad, let me be more precise here.
> 
> We store two different types of items: `RawTrieNodeWithSize` and values, DB key for each of them is the value hash. `RawTrieNodeWithSize` contains some data of limited size and sometimes a `Vec<u8>`.
> 
> At first, estimate limited size data. In the worst case, when node is a branch, we store `memory_usage`, 16 `Option<CryptoHash>` and 1 `Option<(u32, CryptoHash)>`. The total size is `8 + 16 * 40 + 4 + 8 + 36 = 696`. Because we can access only 31K nodes in a chunk, it gives us a limit of 31 KB * 700 ~= 22 MB.
> 
> Now estimate total size of `Vec`s in nodes. Each such vector is a part of storage key we are reading, multiplied by two because we store nibbles instead of bytes. Thus, it is limited by total size of all storage keys accessed in a chunk. `storage_read_key_byte` is ~30 MGas, so we have a limit 2 * 500 Tgas / 30 Mgas ~= 34 MB.
> 
> Finally, `storage_read_value_byte` helps us to limit the total size of all values by 500 Tgas / 5.5 Mgas ~= 91 MB.
> 
> In total, this gives a limit of ~150 MB per chunk.
> 
> I would be happy if someone could double-check this :) @mm-near
> 
> Experiments on shards 0 and 1 shows that actual size is much lower. When I computed chunk cache size using deepsize feature, maximum was 392 KB with 86 KB on average for Aurora shard. So I may miss some simpler explanation here.
> 
> I run command like `./neard view_state apply_range --start-index 55290678 --end-index 55290878 --shard-id 1 --sequential --verbose-output`

This estimation looks reasonable to me

integration-tests/src/tests/client/process_blocks.rs
```diff
@@ -4511,3 +4511,162 @@ mod contract_precompilation_tests {
...
+                    }
+                };
+
+                touching_trie_node_cost / base_cost
+            })
+            .collect();
+
+        assert_eq!(tx_node_touches[1], 12);
+        #[cfg(feature = "protocol_feature_chunk_nodes_cache")]
+        {
```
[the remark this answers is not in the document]
Longarithm: The third assert does it. If that was the case, the number of touches there would be 4 instead of 8, because we would not count first 4 nodes **both times**.
I will clarify this in comments as well.

runtime/runtime/src/actions.rs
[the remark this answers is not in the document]
Longarithm: This is because we charge for touching trie nodes only during function calls.
I would use the cache for **all** trie accesses, but then I couldn't control chunk cache size, e.g. in case of multiple `get_account` calls.

integration-tests/src/tests/client/process_blocks.rs
```diff
@@ -4511,3 +4511,162 @@ mod contract_precompilation_tests {
...
+                    ExecutionMetadata::V2(profile_data) => {
+                        profile_data.get_ext_cost(ExtCosts::touching_trie_node)
+                    }
+                };
+
+                touching_trie_node_cost / base_cost
+            })
+            .collect();
+
+        assert_eq!(tx_node_touches[1], 12);
```
matklad, line 4660: should we assert `tx_node_touches[0]` ?

runtime/runtime/src/actions.rs
```diff
@@ -99,21 +99,30 @@ pub(crate) fn execute_function_call(
...
-        &config.transaction_costs,
-        promise_results,
-        apply_state.current_protocol_version,
-        apply_state.cache.as_deref(),
-    );
-    // runtime_ext.set_trie_cache_mode(TrieCacheMode::CachingShard);
-    result
+    // TODO (#5920): Consider using RAII for switching the state back
+    let protocol_version = runtime_ext.protocol_version();
+    let runner = |runtime_ext| {
```
matklad, line 104: I'd avoid adding a lambda here -- lambdas are hard to understand. I'd just

```
if checked_feature!("protocol_feature_chunk_nodes_cache", ChunkNodesCache, protocol_version) {
    runtime_ext.set_trie_cache_mode(TrieCacheMode::CachingChunk);
}
let result = near_vm_runner::run(...);
if checked_feature!("protocol_feature_chunk_nodes_cache", ChunkNodesCache, protocol_version) {
    runtime_ext.set_trie_cache_mode(TrieCacheMode::CachingBlock);
}

```

matklad (commented):
Do we have test which hit the case of "branch with value"? That is, when one key is a prefix of another.

jakmeier:
It's not obvious, but I am a bit concerned this might make writing values unreasonably cheap.

Updating the value for key "ABCD" causes the entire node chain ("A", "AB", "ABC", "ABCD") to be updated due to the hashes changing. I discussed this a bit with @Longarithm in PMs and our current consensus is this cost probably *should* be charged through `storage_write_key_byte` and not `touching_trie_node`.

However, `storage_write_key_byte` is only 0.07 Ggas, whereas `touching_trie_node` is 16.1 Ggas. And each key byte causes up to two extra nodes to be written to the storage! In other words, today's `storage_write_key_byte` does not cover the cost of inserting new nodes and would have to be increased massively, if it should do so.

So far, this was not really a problem.There was no way to write a node without touching all the ancestor nodes first. Hence, the cost of writing to "ABCD" was dominated by `touching_trie_node` costs, which implicitly covered the writing to DB cost. But now we only charge for touching nodes if the nodes are not already in the chunk cache. Therefore, writing 100 times to "ABCD" would only charge 4x `touching_trie_node` instead of previously 400x. (I am not 100% sure about the truth of the last sentence, please correct me if necessary.)

Of course, this would still be okay if we actually did less work. But we still do the same amount of writing to the DB and it does not seem to be covered anywhere.

---

### Potential idea for a solution
Use the trie-key instead of the crypto-hash as the key for the chunk cache. This would mean that multiple writes to the same trie-key are reduced to only one write to the DB. I *believe* we have no reason to persist trie nodes that are only used temporarily while processing a chunk, which justifies this optimization.

Longarithm:
@jakmeier Interesting points!

`storage_write_base` should cover the concern. I claim that each single key-value write inserts **no more than 4 nodes** into Trie.

If we process adding key-value items sequentially, the worst case would be:
* we descended into `Extension` node after which we need to add a new path below
* then, we replace `Extension` node with `Branch` node with two childs
* we create new `Extension` node for the old path and `Leaf` node for the new key-value item
* we add new "Value" node for the item.

 > Use the trie-key instead of the crypto-hash as the key for the chunk cache. This would mean that multiple writes to the same trie-key are reduced to only one write to the DB.

This is actually already done implicitly in the `TrieUpdate` - if we write a value, it will be stored there and we won't make a trie lookup again. Also, I'm a bit confused here because trie keys exists only for trie values (`Account`s, `ContractCode`s, etc.) and doesn't exist for the intermediate nodes - could you clarify what you mean?

Longarithm:
Another important point to mention: if we write new key-value pair, we also have to rewrite existing nodes. This is because when any child hash changes, we have to update the parent node, so we have to update all touched nodes up to the root.

But it is still covered by `touching_trie_node` cost, because we still charge it one time per each touched node.

jakmeier:
> This is actually already done implicitly in the `TrieUpdate`

Right, that changes everything. Thanks for pointing this out!

> Also, I'm a bit confused here because trie keys exists only for trie values (`Account`s, `ContractCode`s, etc.) and doesn't exist for the intermediate nodes - could you clarify what you mean?

Sorry if my wording is imprecise. What I meant by the "trie-key" is the position of a node in the trie. For example, take a trie with two keys "aaabb" and "aaacc". This should have 1 root node, 3 extension nodes, 1 branch node and 2 value nodes. I would say the branch node has the trie position *"aaa"*. I guess the extensions are a bit trickier but for example, the position could be described as *extension from "aaa" to "aaabb"*.

> `storage_write_base` should cover the concern. I claim that each single key-value write inserts **no more than 5 new nodes** into Trie.

That sounds right. If the number of trie nodes we write is limited by a constant, then it can be charged through `storage_write_base`. With 64 Ggas this is also in the right weight-class.

I think I have to dig even deeper through the code to fully understand this. But your arguments seem to be valid and should cover the case that I brought up.

What I am still confused is mostly this. Writing to a value calls `move_node_to_mutable()` on all nodes on the path to that value. This in turn calls `retrieve_raw_bytes()` which used to always charge `touching_trie_node`. But now `retrieve_raw_bytes()` will only charge it when it is not in the chunk cache. It seems like we vastly lower cost per mutated node while doing the same amount of work.

matklad:
> "touching_trie_node_cached" cost is dominated by storage_read_base cost which is 56 Ggas - Aurora measurements showed that trie depth of their keys is in [7, 13] range

Is this (7-13 range)a safe assertion to make? It seems that it’s possible to artificially create a key with trie depth=10_000.

Longarithm:
> What I am still confused is mostly this. Writing to a value calls move_node_to_mutable() on all nodes on the path to that value. This in turn calls retrieve_raw_bytes() which used to always charge touching_trie_node. But now retrieve_raw_bytes() will only charge it when it is not in the chunk cache. It seems like we vastly lower cost per mutated node while doing the same amount of work.

Oh, I see.
Here the point is that increments of counter caused by `move_node_to_mutable` and related function never was accounted in costs.

Trie is actually updated in `TrieUpdate::finalize` which happens after all txs/receipts are processed, and we use the counter for costs computation only during function calls. So for such kind of operations cost doesn't change.

jakmeier:
Ah that clears things up! Thanks for the explanation.

jakmeier (approved):

Longarithm:
> Is this (7-13 range)a safe assertion to make? It seems that it’s possible to artificially create a key with trie depth=10_000.

That's a good point.
I need to investigate if it can be abused, and I will post an update when I have more understanding.

Longarithm:
Returning to the PR with the updated solution for charging per trie node read from memory.

Because ~3 weeks passed, let me refresh latest discussion points:

## Testing

Tests were updated to reflect two different trie node access counts - from DB and memory.
In standard_cases, we test 2 trie shapes - when 2 nodes share a parent and when one is a prefix of another.
Also, test_chunk_nodes_cache_mode checks that chunk nodes cache is not enabled for all actions. There I execute deploy contract action for which the code should not be cached.
process_block test the correctness of costs on the protocol upgrade.

## Costs

Our last experiments show that upper bound cost of reading cached trie node is 760 Mgas. With the safety multiplier 3, we charge ~2.3 Ggas which is 7x lower than TTN cost. The estimation will be merged later, currently we prioritize releasing the feature itself.
This still helps us to save 27% of Aurora costs on average. Also, it allows users to read only 217K nodes from the cache.

The theoretical limitations on the chunk cache size remain the same - 150 MB. However, experiments show that actual size is bounded by 400 KB.

integration-tests/src/tests/standard_cases/mod.rs
```diff
@@ -1328,35 +1325,230 @@ pub fn test_smart_contract_free(node: impl Node) {
...
+                }
+                "READ_CACHED_TRIE_NODE" => {
+                    count.mem_reads +=
+                        cost.gas_used / runtime_config.wasm_config.ext_costs.read_cached_trie_node;
+                }
+                _ => {}
+            };
+            count
+        },
+    )
```
matklad, line 1349: Optional, but it doesn't seem like `fold` makes the code more readable here

```rust
let mut res = TrieNodesCount::default();
for cost in metadata.gas_profile.iter() {
  match cost { ... }
}
res
```

matklad (approved):

mm-near (approved):

jakmeier:
I will have a PR following this with the refined estimation code. (it requires this to be merged first) 

The bad news is, estimations increased again. The good news is, I think we should keep the current value of 2.3Ggas for the release. With the newest estimation, I have the following: In a fairly pessimistic setting, 2.3Ggas is in the 99th percentile of the estimated distribution.
Note that this has no "safety factor of 3" built in, but I think the percentile approach is better anyway. I will present the details in my follow-up PR.

jakmeier (approved):
(deleting outdated review comments that Github somehow remembered and submitted with my approval)

PR #6457, opened 2022-03-18
feat(rosetta): only return info about final blocks and transactions

bowenwang1996:
Rosetta only cares about final blocks.  Change the implementation to
return only final blocks and transactions.

Test plan
--------------------------------
* pytest/rosetta.py
* Deployed to a node on testnet


chain/rosetta-rpc/src/lib.rs
```diff
@@ -185,6 +185,80 @@ async fn network_options(
...
+        .send(near_client::GetBlock(near_primitives::types::BlockReference::Finality(
+            near_primitives::types::Finality::Final,
+        )))
+        .await?
+    {
+        Ok(block) => block,
+        Err(_) => {
+            return Err(errors::ErrorKind::InternalError("final block not found".to_string()).into())
+        }
+    };
```
mina86, line 227: We always make this call.  It might be more readable to restructure
this function to make this query first.  We can also put height check
here which gets rid of one fetch if querying non-final heights:

```
    let final_block = ...;
    let is_query_by_height = match block_id {
        ...::Finality(...) => return Some(final_block),
        ...::BlockId(...Height(height)) => {
            if height > final_block.header.height {
                return Ok(None);
            }
            true
        },
        _ => false
    };

    let block = ...;

    if is_query_be_height {
        return Ok(Some(block));
    }
    if block.header.height > final_block.header.height {
        return Ok(None);
    }

    ...
```

chain/rosetta-rpc/src/lib.rs
```diff
@@ -185,6 +185,80 @@ async fn network_options(
...
+    if block.header.height <= final_block.header.height {
```
mina86, line 228: I’d negate the condition and make an early return:

```
    if block.header.height > final_block.header.height {
        return Ok(None);
    }
```

Excessive indentation and control flow makes code harder to read.

chain/rosetta-rpc/src/lib.rs
```diff
@@ -185,6 +185,80 @@ async fn network_options(
...
+                ))
+                .await?
+            {
+                Ok(block) => block,
+                Err(_) => {
+                    return Err(errors::ErrorKind::InternalInvariantError(format!(
+                        "block {} not found",
+                        block.header.height
+                    ))
+                    .into())
```
mina86, line 246: Why is this an error and not `Ok(None)`?

mina86 (commented):
We also need to edit `fn network_status` I believe. It currently returns latest block but it should return latest final block.

frol:
I feel we should just change the status endpoint to return the height of the final block instead of the latest block, and leave these methods alone since Rosetta won't query the blocks beyond the height that is reported in the status response anyway.

mina86:
@bowenwang1996 I have some free cycles right now so I can take this PR over if you want.

bowenwang1996:
> @bowenwang1996 I have some free cycles right now so I can take this PR over if you want.

@firatNEAR Will work on it

chain/rosetta-rpc/src/lib.rs
firatNEAR: @bowenwang1996 I am also confused why this returns an error

chain/rosetta-rpc/src/lib.rs
```diff
@@ -185,6 +194,73 @@ async fn network_options(
...
+                    block.header.height
+                ))
+                .into())
+            }
+        };
+        if block.header.hash == block_on_canonical_chain.header.hash {
+            Ok(Some(block))
+        } else {
+            Ok(None)
+        }
```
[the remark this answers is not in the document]
firatNEAR: The answer I got from Coinbase: 
Ideally the node will ONLY return finalized blocks from all endpoints.
So I think we should keep it this way

chain/rosetta-rpc/src/lib.rs
```diff
@@ -185,6 +198,73 @@ async fn network_options(
...
+        Ok(block) => block,
+        Err(_) => return Ok(None),
+    };
+    // if block height is larger than the last final block height, then the block is not final
+    if block.header.height > final_block.header.height {
+        return Ok(None);
+    }
+    // check that this block is on the canonical chain
+    if is_query_by_height {
+        Ok(Some(block))
```
mina86, line 243: I’d convert this to a `return Ok(Some(block));` so that the else part can be dedented.

mina86 (approved):
LGTM just with minor comments.

chain/rosetta-rpc/src/lib.rs
```diff
@@ -185,6 +188,80 @@ async fn network_options(
...
+) -> Result<near_primitives::views::BlockView, models::Error> {
+    return match view_client_addr
+        .send(near_client::GetBlock(near_primitives::types::BlockReference::Finality(
+            near_primitives::types::Finality::Final,
+        )))
+        .await?
+    {
+        Ok(block) => Ok(block),
+        Err(_) => Err(errors::ErrorKind::InternalError("final block not found".to_string()).into()),
+    };
```
frol, line 262:
```suggestion
   match view_client_addr
        .send(near_client::GetBlock(near_primitives::types::BlockReference::Finality(
            near_primitives::types::Finality::Final,
        )))
        .await?
        .map_err(|_| errors::ErrorKind::InternalError("final block not found".to_string()))
```

chain/rosetta-rpc/src/lib.rs
```diff
@@ -185,6 +188,80 @@ async fn network_options(
...
+            if height == &final_block.header.height {
+                return Ok(Some(final_block));
+            }
+            true
+        }
+        _ => false,
+    };
+    let block = match view_client_addr.send(near_client::GetBlock(block_id.clone())).await? {
+        Ok(block) => block,
+        Err(_) => return Ok(None),
```
frol, line 218: We should ideally handle the specific error type instead of reporting any error as a missing block.

chain/rosetta-rpc/src/lib.rs
```diff
@@ -185,6 +188,80 @@ async fn network_options(
...
+        ) => {
+            if height > &final_block.header.height {
+                return Ok(None);
+            }
+            if height == &final_block.header.height {
+                return Ok(Some(final_block));
+            }
+            true
+        }
+        _ => false,
```
frol, line 214: Instead of a catch-all arm (`_`), prefer explicitly list all the enum variants, so in the future, when there are more variants added, Rust will force the contributor to review the unhandled conditions.

chain/rosetta-rpc/src/lib.rs
```diff
@@ -185,6 +188,80 @@ async fn network_options(
...
+            .await?
+        {
+            Ok(block) => block,
+            Err(_) => {
+                return Err(errors::ErrorKind::InternalInvariantError(format!(
+                    "block {} not found",
+                    block.header.height
+                ))
+                .into())
+            }
```
frol, line 241: Use `map_err()?` for better readability 🙏

frol (approved):
Besides the nits, this PR looks good to me

PR #6459, opened 2022-03-18
feat(state-viewer): add apply_tx and apply_receipt commands

marcelo-gonzalez:
This adds apply_tx and apply_receipt commands that search for the given tx or receipt, and run the equivalent of `view_state apply` or `view_state apply_chunk`, based on whether the tx or receipt has been applied in some block we've already saved in the chain.

This is a breaking change, since we're getting rid of the --txs and --receipts flags for the `apply_chunk` command, but it's not that likely that anybody depends on that new command at this point.


bowenwang1996 (commented):
How did you test this implementation?

marcelo-gonzalez:
> How did you test this implementation?

For the case where the tx/receipt is included in some block, I just ran the command with hashes of some txs/receipts that should exist. Also of course trying it with hashes that don't exist to see that it does what you'd want

then for the case where it doesn't exist in any block on disk, ran a localnet w/ this change:

```diff
diff --git a/runtime/runtime/src/lib.rs b/runtime/runtime/src/lib.rs
index 753cf36ad..b5f4d5c2b 100644
--- a/runtime/runtime/src/lib.rs
+++ b/runtime/runtime/src/lib.rs
@@ -226,6 +226,11 @@ impl Runtime {
             tracing::debug_span!(target: "runtime", "Runtime::process_transaction").entered();
         metrics::TRANSACTION_PROCESSED_TOTAL.inc();
 
+        if let Some(Action::Transfer(a)) = signed_transaction.transaction.actions.first() {
+            if a.deposit == 4 {
+                panic!("ai!");
+            }
+        }
         match verify_and_charge_transaction(
             &apply_state.config,
             state_update,

```
with this pytest code:

```python
#!/usr/bin/env python3

import sys, time, base58
import pathlib

sys.path.append(str(pathlib.Path(__file__).resolve().parents[2] / 'lib'))

import transaction
import utils

nodes = start_cluster(3, 0, 4,
                      None,
                      {},
                      {
                          0: { "tracked_shards": [0, 1, 2, 3] },
                          1: { "tracked_shards": [0, 1, 2, 3] },
                          2: { "tracked_shards": [0, 1, 2, 3] },
                      }
)

amount = 1
nonce = 1

for height, hash in utils.poll_blocks(nodes[0],
                                       timeout=100,
                                       poll_interval=0.1):
    print(height)
    if height >= 12:
        break
    if height > 1 and amount < 40:
        for node_idx in range(len(nodes)):
            tx = transaction.sign_payment_tx(
                nodes[0].signer_key, 'test%s' % node_idx, amount,
                nonce,
                base58.b58decode(hash.encode('utf8')))
            nodes[0].send_tx(tx)
            nonce += 1
        amount += 1
```

so that it should crash when trying to apply the block sending 4 tokens, but we do have chunks containing our txs/receipts

nikurt (approved):

bowenwang1996 (commented):
@marcelo-gonzalez could we add some integration tests that can be run in CI? This is to avoid accidentally breaking it in the future

bowenwang1996:
> @bowenwang1996 what did you have in mind as far as why the integration test infra would be needed instead of just unit tests? As far as I can tell there's no state viewer integration tests right now, @nikurt do you know if there are any?

Sorry I meant unit test in state viewer

marcelo-gonzalez:
> > @bowenwang1996 what did you have in mind as far as why the integration test infra would be needed instead of just unit tests? As far as I can tell there's no state viewer integration tests right now, @nikurt do you know if there are any?
> 
> Sorry I meant unit test in state viewer

@bowenwang1996 ok added test. btw this last commit basically contains https://github.com/near/nearcore/pull/6488, because it makes it a little bit easier to test (the test code doesn't have a `NearConfig` around), and feels more correct anyway

PR #6461, opened 2022-03-18
Fix memory leak in ChunksDelayTracker

nikurt:
The leak happens if an already processed block attempts to be processed again.

### DIFF
--- a/chain/client/src/client.rs
+++ b/chain/client/src/client.rs
@@ -701,7 +701,8 @@ impl Client {
         block: MaybeValidated<Block>,
         provenance: Provenance,
     ) -> (Vec<AcceptedBlock>, Result<Option<Tip>, near_chain::Error>) {
-        self.record_receive_block_timestamp(block.header().hash());
+        let processing_started = Clock::instant();
+        let block_hash = block.header().hash().clone();
         let is_requested = match provenance {
             Provenance::PRODUCED | Provenance::SYNC => true,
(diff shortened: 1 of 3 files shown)


chain/client/src/client.rs
```diff
@@ -746,6 +747,19 @@ impl Client {
...
+        let block_known_error = if let Err(ref e) = result {
+            match e.kind() {
+                ErrorKind::BlockKnown(_) => true,
+                _ => false,
+            }
+        } else {
+            false
+        };
+        if !block_known_error {
+            self.chunks_delay_tracker.received_block(&block_hash, processing_started);
```
mzhangmzz, line 760: Good catch! Although I think this fix doesn't fix the issue completely, there could be other errors such as `EpochOutOfBounds` or if the block is invalid that will cause the same issue. And this change changes the timestamp for receiving block to after the block is processed. I think a better way to do this is to still call `received_block` at the beginning of this function, but remove the block hash if result is not Ok, or Error::Orphan or Error::ChunksMissing.

mm-near (approved):

mzhangmzz (changes requested):

chain/client/src/client.rs
nikurt: You're right, thanks.

nikurt:
Obsolete because of https://github.com/near/nearcore/pull/6523

PR #6464, opened 2022-03-18
fix(params-estimator): TouchingTrieNodeWrite

jakmeier:
Write different values in each measurement to ensure DB gets used.


jakmeier:
The old implementation wrote the same value over and over again, which did not actually trigger writes all the way to the database. In the icount-based estimation, it clearly shows that we had no write IO before, which the new implementation has.

Before:
```
TouchingTrieNode                                     4_838_931_572 gas [    38711.45i 0.00r 0.00w]           (computed in 141.405717398s) 
TouchingTrieNodeRead                                 1_679_372_343 gas [    13423.83i 0.00r 0.24w]           (computed in 82.64µs) 
TouchingTrieNodeWrite                                4_838_931_572 gas [    38711.45i 0.00r 0.00w]           (computed in 59.475µs) 
Finished in 141.45s, output saved to:
```
After:
```
TouchingTrieNode                                     7_976_474_887 gas [  55972.79i 0.00r 166.79w]           (computed in 141.303352765s) 
TouchingTrieNodeRead                                 1_677_844_273 gas [    13411.61i 0.00r 0.24w]           (computed in 96.825µs) 
TouchingTrieNodeWrite                                7_976_474_887 gas [  55972.79i 0.00r 166.79w]           (computed in 64.347µs) 
```

Our parameter value is `16_101_955_926`. We are still below that, but the safety margin is now around `2x` instead of the `3x` we initially used to come up with the parameter.

Thanks @Longarithm to point this out!

matklad (approved):
Good catch!

Although this begs a question of whether we should also make sure that intermediate trie nodes are updated. At least my understanding is that after the PR, we update the leaf node, but the branch nodes stay the same.

jakmeier:
In a merkle-trie, updating a leaf implicitly changes all the parents anyway, since all children hashes are part of the intermediate node. At least that was my assumption here from the beginning.

matklad (approved):

PR #6465, opened 2022-03-19
db: use EnumCount instead of NUM_COLS

mina86:
Instead of manually setting value of NUM_COLS constant, use EnumCount
trait which can be automatically derived for DBCol.


bowenwang1996 (approved):
Nice!

PR #6466, opened 2022-03-21
Print more detailed information about upcoming blocks

mm-near:
Print more detailed information about the upcoming blocks:
```
stats: #   75005 Downloading blocks 72.40% (28589) 3 peers ⬇ 42 B/s ⬆ 34 B/s 6.01 bps 0 gas/s CPU: 0%, Mem: 346.7 MiB
Mar 25 18:21:29.137 DEBUG stats: EpochId(`7ToKmaSRpyTNXjkRqKAtEbw58nVDKQKPthAatA7NHzj2`) Blocks in progress: 13 Chunks in progress: 15 Orphans: 5
75006 BdRFKoRQTxwUUW5WBRmP2bruGHG5FZdpTh5HBMqh7Lsn in progress for: 1.023578296s orphan for 1.016246792s Chunks:(✔✔✔✔✔))
75007 8cnhCm7VW58NufmThz6XjbR7Cne4uvAwBNtMVyHYunXB in progress for: 1.030077032s orphan for 1.029875794s Chunks:(✔✔✔✔✔))
75008 6iz2sLmhEASJ7kB6esJ3vEdk6X1mJxYP6MCVykhgCkZp in progress for: 1.028685958s orphan for 1.028509392s Chunks:(✔✔✔✔✔))
75009 9YrxSt3YVqiFMXPBZ3vYY6ytVktThY5Kyv6Aqs8xjEwH in progress for: 1.027374049s orphan for 1.027200947s Chunks:(.....))
75010 HAMFWGyyLMTFeBWiJCgVPA4pLsREzJgQFMqzyHJ7sRga in progress for: 1.012854001s orphan for 1.01270946s Chunks:(.....))
```


bowenwang1996:
> 61871384 AjaQt764LW9GC17DN8hjhcXGgtq4RJ8GM75HCWLrZSnZ Some(8.960966961s) None Chunks:(⬆⬆⬆⬆))
> 61871385 FdgKjPHZhJGn9PQEurBRqd2Jy3fNPgtYTmgr5nnR2WFG Some(8.035524445s) Some(8.007929715s) Chunks:(⬆⬆⬆⬆))
> 61871386 12gz4U6uLAjh3MYvQQqjx3a9eWpVVYXYE8kAinGBLNA1 Some(5.877330635s) Some(5.847115045s) Chunks:(⬆⬆⬆⬆))
> 61871387 EXrycwW75BQrYS9dRd8VoB2x6JxMDEzA124HPDb26v5C Some(5.172500383s) Some(5.159181685s) Chunks:(....))
> 61871388 2iVirrtmKWGTtwSfJNnzwnE7kKUkv79yEdJBjtYi5TKf Some(4.22288073s) Some(4.191636519s) Chunks:(....))
> 61871389 73HUK9nYtdTyZhhPyg53SFwFkrYuozAUAQc94qMRoC9A Some(2.272475162s) Some(2.245923886s) Chunks:(....))
> 61871390 EU52fTHivzTN5Dtr1Kh7X86azvjRdFoDBfkXucFZXyvV Some(744.329542ms) Some(718.973276ms) Chunks:(....))

Feels like it is not immediately clear what the two pieces of time info here mean. Also do we always want to print this as part of info log?

bowenwang1996 (commented):
@mzhangmzz please review this PR

chain/client/src/client.rs
```diff
@@ -102,7 +104,35 @@ pub struct Client {
...
+    // Chunk statuses are below:
+    // We first sent the request to fetch the chunk
+    // Later we get the response from the peer and we try to reconstruct it.
+    // If reconstructions suceeds, the chunk will be marked as complete.
+    // If it fails (or fragments are missing) - we're going to re-request the chunk again.
+
+    // Chunks that we reqeusted (sent the request to peers).
+    pub chunks_requested: HashSet<ChunkHash>,
+    // Chunks for which we've received the response.
+    pub chunks_received: HashSet<ChunkHash>,
```
mzhangmzz, line 131: Not sure what's the difference between "chunk_received" and "chunk_not_completed" and "chunks_completed". It seems from the comments that "chunks_received" means some chunk parts are received, so it could be overlapping with "chunks_not_completed" and "chunks_completed". Please update comments to make it clearer

chain/client/src/client.rs
```diff
@@ -1798,11 +1832,134 @@ impl Client {
...
+                            if block_info.chunks_completed.len() == block_info.chunk_hashes.len() {
+                                paint(ansi_term::Colour::Green, Some(chunk_status.join("")))
+                            } else {
+                                paint(ansi_term::Colour::White, Some(chunk_status.join("")))
+                            };
+
+                        format!(
+                            "{} {:?} {:?} Chunks:({}))",
+                            entry.0,
+                            block_info.in_progress_for,
```
mzhangmzz, line 1944: Could you add "in progress for" and "in orphan for" before the timestamps? Otherwise it's hard to know what the time stands for.

chain/client/src/chunks_delay_tracker.rs
```diff
@@ -10,29 +13,48 @@ use crate::metrics;
...
-struct ChunkInProgress {
+pub struct ChunkInProgress {
     /// Timestamp of requesting a missing chunk.
-    chunk_requested: Instant,
+    pub chunk_requested: Instant,
     /// Timestamp of receiving a chunk.
     /// If a chunk is received multiple times, only the earliest timestamp is recorded.
-    chunk_received: Option<Instant>,
+    pub chunk_received: Option<Instant>,
+    pub requestor_block_hash: CryptoHash,
```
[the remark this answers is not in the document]
mm-near: done

chain/client/src/client.rs
mm-near: you're right - we can remove the 'not completed' HashSet.

chain/client/src/client.rs
mm-near: done

mm-near:
@bowenwang1996 - I've moved it do debug, and I've also changed the default RUST_LOG to include this. (this will allow us to disable it if it becomes too spammy/unncessary)

chain/client/src/client.rs
mzhangmzz: Actually I think it's good to add another state "chunks_missing" if there is no chunk included for this shard, so we don't have to wait for it

mzhangmzz (commented):
Please also update the PR description to reflect the most recent debug lines

chain/client/src/client.rs
mm-near: If chunk is not in the block - then it will be missing from chunk_hashes Vec - so it should be then skipped during print.

We could actually do something else: and in such case actually print a different symbol - would be useful if the RPC node doesn't track all the shards - but I'd do that in a separate PR.

mzhangmzz: Actually not in your current implementation. When chunks are missing, the block still contains chunk headers for all shards, for the shards that don't have new chunks, the chunk header will simply be the chunk header in the previous block. So to check when a chunk is missing or not, you need to check if "chunk.height_included() == block.height()"

mm-near: I see - in such case, the current code will mark such chunks as 'succeeded' (as it will see that they already exist in storage).

We might add a separate char in the future - if this one turns out to be too confusing.

chain/client/src/chunks_delay_tracker.rs
```diff
@@ -10,29 +13,49 @@ use crate::metrics;
...
 /// before it is received.
-struct ChunkInProgress {
+pub struct ChunkInProgress {
     /// Timestamp of requesting a missing chunk.
-    chunk_requested: Instant,
+    pub chunk_requested: Instant,
     /// Timestamp of receiving a chunk.
     /// If a chunk is received multiple times, only the earliest timestamp is recorded.
-    chunk_received: Option<Instant>,
+    pub chunk_received: Option<Instant>,
```
mzhangmzz, line 43: actually I think `chunk_received` here actually means the chunk is completed. It means all chunk parts are received, instead of some chunk parts are received. So the `chunks_received` and `chunk_completed` in `UpcomingBlockDebugStatus` are not different, except when some chunks are missing in a block.

mzhangmzz (approved):

chain/client/src/chunks_delay_tracker.rs
mm-near: well spotted. I'll fix in the next PR (probably by adding more state here)

PR #6467, opened 2022-03-21
Add metrics for delay between client actor and network

mzhangmzz:
This change adds two metrics:
- The delay between when ClientActor requests a partial chunk and when PeerManagerActor receives the request from ClientActor
- The delay between when PeerManagerActor sends a partial chunk response and when ClientActor receives the response


chain/client/src/metrics.rs
```diff
@@ -152,3 +152,10 @@ pub static CHUNK_SKIPPED_TOTAL: Lazy<IntCounterVec> = Lazy::new(|| {
     )
     .unwrap()
 });
+pub static PARTIAL_ENCODED_CHUNK_RESPONSE_DELAY: Lazy<Histogram> = Lazy::new(|| {
+    try_create_histogram(
+        "partial_encoded_chunk_response_delay",
+        "Delay between when a partial encoded chunk response is sent from PeerActor and when it is received by ClientActor",
+    )
```
bowenwang1996, line 159: This doesn't tell us whether the delay is between peer actor and peer manager actor or between peer manager actor and client actor

integration-tests/src/tests/client/chunks_management.rs
```diff
@@ -189,6 +189,7 @@ fn chunks_produced_and_distributed_common(
                     NetworkRequests::PartialEncodedChunkRequest {
                         target: AccountIdOrPeerTrackingShard { account_id: Some(to_whom), .. },
                         request: _,
+                        create_time: _,
```
bowenwang1996, line 192:
```suggestion
                        ..
```

chain/network/src/types.rs
```diff
@@ -385,7 +386,7 @@ pub enum NetworkClientMessages {
     /// Request chunk parts and/or receipts.
     PartialEncodedChunkRequest(PartialEncodedChunkRequestMsg, CryptoHash),
     /// Response to a request for  chunk parts and/or receipts.
-    PartialEncodedChunkResponse(PartialEncodedChunkResponseMsg),
+    PartialEncodedChunkResponse(PartialEncodedChunkResponseMsg, Instant),
```
bowenwang1996, line 389: Nit: named fields would be better

bowenwang1996 (approved):
Please fix CI

mm-near (approved):

chain/client/src/metrics.rs
mzhangmzz: The response is sent directly from PeerActor to ClientActor, there is no PeerManagerActor in between

PR #6468, opened 2022-03-21
refactor: Receipt handling to be done in VM logic

austinabell:
Refactors receipt handling from being done in all `External` implementations to being done in vm-logic. A few refactoring patterns were explored but this one solved the main issues of testability and deduplication the best I've found.

The tests are somewhat of a property-based test with all common cases included, but this can be expanded to be fuzzed as well. (might do this in separate PR though)


runtime/near-vm-logic/src/logic.rs
```diff
@@ -2670,6 +2702,19 @@ pub struct VMOutcome {
     pub logs: Vec<String>,
     /// Data collected from making a contract call
     pub profile: ProfileData,
+    action_receipts: ActionReceipts,
```
matklad, line 2705: Dual observation to that about `epoch_manager`: in this struct, all fields are `pub` -- it's a plain old data, value type with defined identity. Adding a private field here breaks valueness. 

I think it'll actually be ok to make `ActionReceipts` public, and introduce some kind of `VMLogicRecipt`. 
take_receipts
That is, rather than doing `VMOutcome::take_receipts`, the caller would themselves convert VMLogicRecipts to proper Receipts.

matklad (commented):
This looks beautiful!

I have only one overall comment -- let's just make receipts a public field of VMOutcome (burdening the caller with the task of converting them to actual recipts). Other than that, it looks basically good to me! Though, I didn't do a super thorough review, as this is still a draft.

runtime/near-vm-logic/src/logic.rs
austinabell: Sounds good! I was trying hard to make it so there wasn't breaking changes and exposing things that aren't needed by nearcore. Apparently too careful :)

I'll make this change now and open this up for review

matklad: >I was trying hard to make it so there wasn't breaking changes

We don't care about backwards compatability at this layer. On the contrary, we are willing to change all the call-sites to prevent tech debt from accumulating.

austinabell: The only question about this is should I update the `PartialEq` implementation for outcome to include the receipts? I avoided it for now to avoid breaking change, but don't have an opinion

runtime/near-vm-logic/src/receipt_manager.rs
```diff
@@ -0,0 +1,493 @@
...
+        receipt_indices: Vec<u64>,
+        receiver_id: AccountId,
+    ) -> ExtResult<u64> {
+        let mut input_data_ids = vec![];
+        for receipt_index in receipt_indices {
+            let data_id = ext.generate_data_id();
+            self.action_receipts
+                .0
+                .get_mut(receipt_index as usize)
+                .ok_or_else(|| HostError::InvalidReceiptIndex { receipt_index })?
```
matklad, line 130: :thinking: this is pre-existing, but it actually feels like it should be an assert. If/when we refactor ReceiptIndex to be a proper newtype index, rather than just a type alias, it'll be more clear.

matklad: No action required in this PR!

runtime/near-vm-logic/src/receipt_manager.rs
```diff
@@ -0,0 +1,493 @@
...
+    /// * `gas` - amount of unused gas to distribute
+    ///
+    /// # Returns
+    ///
+    /// Function returns a [GasDistribution] that indicates how the gas was distributed.
+    #[cfg(feature = "protocol_feature_function_call_weight")]
+    pub fn distribute_unused_gas(&mut self, gas: u64) -> GasDistribution {
+        let gas_weight_sum: u128 =
+            self.gas_weights.iter().map(|(_, GasWeight(weight))| *weight as u128).sum();
+        if gas_weight_sum != 0 {
```
matklad, line 450: I'd early return here, to save indentatin:

```
if gas_weight_sum == 0 {
    return ...
}
```

runtime/near-vm-logic/src/receipt_manager.rs
```diff
@@ -0,0 +1,493 @@
...
+            // Floor division that will ensure gas allocated is <= gas to distribute
+            let gas_per_weight = (gas as u128 / gas_weight_sum) as u64;
+
+            let mut distribute_gas = |metadata: &FunctionCallActionIndex, assigned_gas: u64| {
+                let FunctionCallActionIndex { receipt_index, action_index } = metadata;
+                if let Some(Action::FunctionCall(FunctionCallAction { ref mut gas, .. })) = self
+                    .action_receipts
+                    .0
+                    .get_mut(*receipt_index)
```
matklad, line 459: Let's extract this bit into `fn get_fuction_call_action_mut` function, it'll more clearl y explain the purpose of FunctionCallIndex construct.

runtime/near-vm-logic/src/receipt_manager.rs
```diff
@@ -0,0 +1,493 @@
...
+                    .and_then(|(_, receipt)| receipt.actions.get_mut(*action_index))
+                {
+                    *gas += assigned_gas;
```
matklad, line 462: let's add comment explaining that this can't ovreflow

runtime/near-vm-logic/src/receipt_manager.rs
```diff
@@ -0,0 +1,493 @@
...
+    ///
+    /// # Arguments
+    ///
+    /// * `gas` - amount of unused gas to distribute
+    ///
+    /// # Returns
+    ///
+    /// Function returns a [GasDistribution] that indicates how the gas was distributed.
+    #[cfg(feature = "protocol_feature_function_call_weight")]
+    pub fn distribute_unused_gas(&mut self, gas: u64) -> GasDistribution {
```
matklad, line 447: Let's name the `gas` here something else -- we have a local `gas` variable inside, and, in this case, this is confusing.

runtime/near-vm-logic/src/receipt_manager.rs
```diff
@@ -0,0 +1,493 @@
...
+
+                distribute_gas(action_index, assigned_gas);
+
+                distributed += assigned_gas
+            }
+
+            // Distribute remaining gas to final action.
+            if let Some((last_idx, _)) = self.gas_weights.last() {
+                distribute_gas(last_idx, gas - distributed);
+            }
```
matklad, line 486: :thinking: I am still not happy about this algorithm. A bad example here would be if we want to distribute `10` gas across `10` function calls evenly, but where each weight is `1000`. In that situation, `gas_per_weight` will be zero, and only the last call will get the gas. 

I suggest tackling this separately though, let's first land this much needed refactor.

runtime/near-vm-logic/src/tests/gas_counter.rs
```diff
@@ -103,54 +103,103 @@ fn test_hit_prepaid_gas_limit() {
...
-            static_gas,
-            GasWeight(gas_weight),
+            *static_gas,
+            GasWeight(*gas_weight),
         )
         .expect("batch action function call should succeed");
+        ratios.push((index, *gas_weight));
+    }
+
+    fn assert_with_gas(receipt: &ReceiptMetadata, cb: impl Fn(Gas) -> bool) {
```
matklad, line 132:
```suggestion
    #[track_caller]
    fn assert_with_gas(receipt: &ReceiptMetadata, cb: impl Fn(Gas) -> bool) {
```

runtime/near-vm-logic/src/receipt_manager.rs
austinabell: So with this, this is more of an edge case where the weight is greater or close to the amount of gas remaining.

The downside to this is just that it requires extra logic and computation to handle for this edge case. I'm also wondering in the case of splitting 12 gas among the 10 function calls, do the last two functions receive an extra 1 gas or how will this happen?

I don't have an opinion on which direction this goes, I was just defaulting to this for simplicity rather than creating a bunch of semantics that only does something subjectively meaningful with improper usage.

runtime/near-vm-logic/src/logic.rs
```diff
@@ -2670,6 +2701,7 @@ pub struct VMOutcome {
     pub logs: Vec<String>,
     /// Data collected from making a contract call
     pub profile: ProfileData,
+    pub action_receipts: ActionReceipts,
```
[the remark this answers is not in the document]
austinabell: Will do, the reason it was avoided before it to keep the API surface the same and not introduce extra public fields or types.

This also will make it easier to migrate from using primitives within vm-logic in the future

runtime/near-vm-logic/src/receipt_manager.rs
matklad: I don't think handlig this particular cases requires that much excar logic. I think what we need here is

```
let assigned_gas = (gas_to_distribute as u128 * weight as u128) / sum_weights;
```

That is, we just needlessly loosing precision here. 

Distributing the remainder evenly is a different, smaller problem, that I think would require more complex logic.

runtime/near-vm-logic/src/tests/gas_counter.rs
```diff
@@ -103,54 +103,103 @@ fn test_hit_prepaid_gas_limit() {
...
+
+    let (gas_per_weight, remainder_gas) = {
+        let gas_to_distribute = logic.gas_counter().remaining_gas();
+        let weight_sum: u64 = function_calls
+            .iter()
+            .map(|(_, weight)| weight)
+            // Saturating sum of weights
+            .fold(0, |sum, val| sum.checked_add(*val).unwrap_or(u64::MAX));
+        (gas_to_distribute / weight_sum, gas_to_distribute % weight_sum)
+    };
```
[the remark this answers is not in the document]
austinabell: Yeah, that's a good suggestion, I was avoiding doing specific values to avoid being brittle, but I will change this. (I might keep some of this for a prop based fuzz test though)

runtime/near-vm-logic/src/tests/gas_counter.rs
```diff
@@ -103,54 +103,103 @@ fn test_hit_prepaid_gas_limit() {
...
+        }
     }
+
+    let (gas_per_weight, remainder_gas) = {
+        let gas_to_distribute = logic.gas_counter().remaining_gas();
+        let weight_sum: u64 = function_calls
+            .iter()
+            .map(|(_, weight)| weight)
+            // Saturating sum of weights
+            .fold(0, |sum, val| sum.checked_add(*val).unwrap_or(u64::MAX));
```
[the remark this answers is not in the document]
austinabell: The implementation doesn't panic though, so this would cause inconsistency (one of the test cases has larger than u64::MAX ratio sum)

runtime/near-vm-logic/src/receipt_manager.rs
austinabell: The reason I didn't do this is because this closure is used within code blocks where `self` is borrowed immutably, so this wouldn't be able to be used. I'll extract it into a helper function but I'm unsure that's better

runtime/near-vm-logic/src/receipt_manager.rs
austinabell: This is what I'm referring to, the remainder distribution being more calculations.

My intuition is that it would divide the amount of remaining gas by the number of individual weights to get the distribution of gas per function call, then get the remainder of that division (`x`) to attach the remaining gas to final `x` number of function calls.

Seems like unnecessary logic and computation for what will be trivial amounts of gas unless specifically misused by specifying gas weights larger than the amount of gas, but I'm happy to make the change if there is consensus on this :)

matklad:
This looks good to me! As this is a larger refactor, I think it'll benefit from a second pair of eyes doing the review as well.

@Longarithm would you have time to take a look here?

Longarithm (approved):
LGTM, it's great that we are able to shorten the list of `RuntimeExt` fields.
However, I didn't look into all details, because the change touches many places in the code.

runtime/near-vm-logic/src/receipt_manager.rs
Longarithm: From what I see, here we choose between

```
1) let assigned_gas = ((gas_to_distribute as u128 * weight as u128) / sum_weights) as u64;
2) let assigned_gas = (gas_to_distribute / sum_weights) * weight;
```

I don't see a significant difference for regular usecases, so I would also choose (2) because it allows to make less operations.

matklad (approved):

Issue #6072, opened 2022-01-14
`tracing::{info, error, warn, trace, debug}` is slower by factor of 30 


pmnoxx:
We are currently using tracing for logging. This is done through `info, error, warn, trace, debug` macros.
By default, tracing logs data to `stderr`, or the data gets redirected through pipe to a file.

I did a few measurements, that shows that it's taking a considerable amount of time and that can be improved.

For example `tracing::info!` takes around 1600ns to execute, when it's output is logged to file, while `println!` takes "only" 800 ns.
```shell
// tracing::info!                   time:   [1.5989 us 1.6017 us 1.6065 us]
// log::info!                   time:   [1.8204 us 1.8230 us 1.8256 us]
// println!                time:   [809.49 ns 810.59 ns 811.76 ns]
```

However, when the data is redirected to `/dev/null`, then it takes less time to execute those commands.
```shell
// to /dev/null
// println!                time:   [265.99 ns 267.48 ns 269.30 ns]
// tracing::info!                   time:   [742.95 ns 743.34 ns 743.79 ns]
// log::info!                   time:   [992.79 ns 993.24 ns 993.76 ns]
```

This shows that, the `tracing:info!` is blocking, and does a `syscall` which blocks the thread from executing, and it practice thread can be frozen for a longer period of time while doing disk write.

On the other hand, doing only `format!` takes only around `40ns`. Due to, both the `tracing` being inefficient at both formatting the data and writing them to stdout, the time to write a log increases from `40ns` to `1600ns`
```
// only format!
// format!                 time:   [38.676 ns 38.705 ns 38.740 ns]
```

I did some experiments, when instead of using `tracing`, and instead of writing data to `stderr`, we write data to a concurrent, queue, which then gets consumed by another process. The result is that it takes only around `60ns` to write a log, instead of `1600ns`.

```
// use format! and send result to conqueue
// format+conqueue         time:   [61.043 ns 61.050 ns 61.059 ns]
```

My suggestion is to do the following.
1) Replace, `tracing::{info, error, warn, trace, debug}`, with our own code, which formats the data, and sends an event through MPSC `conqueue` to another thread
2) From another thread we have two choices
a) write data using `println!`, this is what `tracing` does internally. It will take around `700ns`, but at the very least we won't block the target thread
b) send data using UDP to `syslog` (https://en.wikipedia.org/wiki/Syslog)
A syslog service will be responsible for writing data to disk, and our process will not be blocked by disk writes anymore.

Drawbacks:
[...]

PR #6470, opened 2022-03-21
Use non-blocking log writer

nagisa:
This will utilize a separate thread to write out the spans and events
without while letting the main computation to proceed with its business.
Additionally, we are buffering the output by lines, thus reducing the
frequency of syscalls that can occur when the subscriber is writing out
parts of the message.

This should mitigate concerns of enabling debug logging as its impact on
performance should now be minimal (putting an event structure onto a
MPSC queue.) There are still costs associated with logging everything
however. Most notably formatting and construction of the
`tracing_core::ValueSet`s still occur on the caller side, so if
constructing those is expensive, the logging might remain expensive.
An example of code sketchy like that (although silly) could be:

```
debug!(message = { std::time::sleep(Duration::from_secs(1)); "hello" })
```

or for a less silly example:

```
debug!("{}", my_vector.iter().map(|...| {
  do_expensive_stuff()
}).collect::<String>())
```

These should be considered a bug (alas one that `tracing` does not have
any tooling to detect, sadly.)

I opted adding a new crate dedicated to observability utilities. From my
experience using things like prometheus will eventually result in a
variety of utilities being written, so this crate eventually would
likely expand in scope...

Fixes https://github.com/near/nearcore/issues/6072 (though I haven't made any actual measurements as to what the improvement really is)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2349,15 +2349,6 @@ dependencies = [
 
 [[package]]
 name = "matchers"
-version = "0.0.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f099785f7595cc4b4553a174ce30dd7589ef93391ff414dbb67f62392b9e0ce1"
-dependencies = [
- "regex-automata",
-]
-
-[[package]]
-name = "matchers"
 version = "0.1.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "8263075bb86c5a1b1427b5ae862e8889656f126e9f77c484496e8b47cf5c5558"
@@ -2943,7 +2934,7 @@ version = "0.1.0"
 dependencies = [
  "tracing",
  "tracing-appender",
- "tracing-subscriber 0.3.9",
+ "tracing-subscriber",
 ]
 
 [[package]]
@@ -5388,7 +5379,7 @@ checksum = "09d48f71a791638519505cefafe162606f706c25592e4bde4d97600c0195312e"
 dependencies = [
  "crossbeam-channel",
  "time 0.3.7",
- "tracing-subscriber 0.3.9",
+ "tracing-subscriber",
 ]
 
 [[package]]
@@ -5424,45 +5415,13 @@ dependencies = [
 ]
 
 [[package]]
-name = "tracing-serde"
-version = "0.1.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "fb65ea441fbb84f9f6748fd496cf7f63ec9af5bca94dd86456978d055e8eb28b"
-dependencies = [
- "serde",
- "tracing-core",
-]
-
-[[package]]
 name = "tracing-span-tree"
-version = "0.1.0"
+version = "0.1.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "877726b1570d7764022ef20cc61479b5bcfc1118b90521ce61f6cc9e4f5ffbd8"
+checksum = "c3809cda0328d505548af1ca24fac329f41c0c213d0a1e78908922d0b43e5c43"
 dependencies = [
  "tracing",
- "tracing-subscriber 0.2.25",
-]
-
-[[package]]
-name = "tracing-subscriber"
-version = "0.2.25"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "0e0d2eaa99c3c2e41547cfa109e910a68ea03823cccad4a0525dcbc9b01e8c71"
-dependencies = [
- "ansi_term",
- "chrono",
- "lazy_static",
- "matchers 0.0.1",
- "regex",
- "serde",
- "serde_json",
- "sharded-slab",
- "smallvec",
- "thread_local",
- "tracing",
- "tracing-core",
- "tracing-log",
- "tracing-serde",
+ "tracing-subscriber",
 ]
 
 [[package]]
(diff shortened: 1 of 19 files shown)


Cargo.lock
```diff
@@ -1977,11 +1977,12 @@ dependencies = [
  "anyhow",
  "clap 3.0.0-beta.2",
  "near-indexer",
+ "near-o11y",
  "openssl-probe",
  "serde_json",
  "tokio",
  "tracing",
- "tracing-subscriber",
+ "tracing-subscriber 0.2.25",
```
matklad, line 1985: Let’s try to dedup these

core/o11y/src/lib.rs
```diff
@@ -0,0 +1,31 @@
...
+     warn";
+
+/// Run the code with a default subscriber set to the option appropriate for the NEAR code.
+pub fn with_default_subscriber<R, F>(log_filter: EnvFilter, run: F) -> R
+where
+    F: FnOnce() -> R,
+{
+    // Do not lock the `stderr` here to allow for things like `dbg!()` work during development.
+    let stderr = std::io::stderr();
+    let lined_stderr = std::io::LineWriter::new(stderr);
```
matklad, line 20: Oh wow, til about LineWriter!

neard/src/cli.rs
```diff
@@ -27,54 +27,96 @@ pub(super) struct NeardCmd {
...
-            error!("THIS IS A NODE COMPILED WITH ADVERSARIAL BEHAVIORS. DO NOT USE IN PRODUCTION.");
-
-            if env::var("ADVERSARY_CONSENT").unwrap_or_default() != "1" {
-                error!("To run a node with adversarial behavior enabled give your consent by setting variable:");
-                error!("ADVERSARY_CONSENT=1");
-                std::process::exit(1);
+        let verbose = neard_cmd.opts.verbose.as_deref();
+
+        let rust_log = env::var("RUST_LOG");
+        let rust_log = rust_log.as_ref().map(String::as_str).unwrap_or(near_o11y::DEFAULT_RUST_LOG);
```
matklad, line 33:
```suggestion
        let rust_log = rust_log.as_deref().unwrap_or(near_o11y::DEFAULT_RUST_LOG);
```

matklad: It also feels like maybe we need a function in o11y crate to just return effective value of RusT_LOG?

matklad: Or maybe even default envfiler

neard/src/cli.rs
```diff
@@ -27,54 +27,96 @@ pub(super) struct NeardCmd {
...
+        let mut env_filter = EnvFilter::new(rust_log);
+        if let Some(module) = verbose {
+            env_filter = env_filter
+                .add_directive("cranelift_codegen=warn".parse().unwrap())
```
matklad, line 37: This one feels like dead code

tools/chainsync-loadtest/Cargo.toml
```diff
@@ -16,8 +16,7 @@ name = "chainsync-loadtest"
 clap = "=3.0.0-beta.2"
 actix = "=0.11.0-beta.2"
 parking_lot = "0.11.2"
-tracing = "0.1.13"
-tracing-subscriber = "0.2.4"
+tracing-subscriber = "0.3.9"
```
matklad, line 19: We might want to re-export tracing crates form o11y, to have fewer versions in Cargo.tomls

core/o11y/src/lib.rs
```diff
@@ -0,0 +1,31 @@
...
+     telemetry=info,\
+     delay_detector=info,\
+     near-performance-metrics=info,\
+     near-rust-allocator-proxy=info,\
+     warn";
+
+/// Run the code with a default subscriber set to the option appropriate for the NEAR code.
+pub fn with_default_subscriber<R, F>(log_filter: EnvFilter, run: F) -> R
+where
+    F: FnOnce() -> R,
```
matklad, line 16: I have a moderate preference towards guard-based API rather than closure-based API:

https://docs.rs/tracing/latest/tracing/subscriber/fn.set_default.html

As far as I understand, they are equivalent (closure seems to be implemented in terms of guard), but, syntactically, guards seems more light weight (smaller diff, no indents, etc)

matklad (approved):
<3

well overdue, thanks for just fixing this!

core/o11y/Cargo.toml
```diff
@@ -0,0 +1,11 @@
+[package]
+name = "near-o11y"
+description = "Observability helpers for the near codebase"
```
mm-near, line 3: please also add README.md file to this directory - and mention more details (of what we have here and what should belong here in the future)

And also maybe a code example on how to use o11y:

    let env_filter = EnvFilter::from_default_env().add_directive(LevelFilter::INFO.into());
    near_o11y::with_default_subscriber(env_filter, || {
    YOUR CODE
    });

mm-near (approved):

Cargo.lock
nagisa: tracing-span-tree would need an update... Not entirely sure what the motivation was behind maintaining our own.

neard/src/cli.rs
nagisa: The plan is to avoid making any globally functional changes in this PR. If this is dead code, it should be addressed as a followup.

Cargo.lock
matklad: Publishes the new version of tracing span tree!  It’s its own thing because I wasn’t able to find an existing subscriber which gives an output, useful for quick profiling.

now that we wave o11y crate in near, it might be a good idea to just copy the contents here: that’s a tiny amount of code, it isn’t worth it to keep a separate crate for that.

khorolets (approved):

Issue #6462, opened 2022-03-18
Disable state sync from validators


bowenwang1996:
State syncs are expensive and given that most nodes today do track all the shards, nodes don't need to do state sync from validators and it can reduce resource usage for validator nodes.

bowenwang1996:
@mzhangmzz this issue isn't addressed by #6471. A regular node can still state sync from a validator node and state sync requests can be expensive.

mzhangmzz:
@bowenwang1996 I guess I understood this wrong. Do you want regular nodes to send their state sync requests to non-validators? Are you worried that state sync requests sent to validators are slowing down validators?

bowenwang1996:
> Are you worried that state sync requests sent to validators are slowing down validators?

Yes

PR #6471, opened 2022-03-21
avoid catch up when no new shards are needed

mzhangmzz:
fixes https://github.com/near/nearcore/issues/6462

Test plan:
- Added a new test to test that state sync is not triggered when there are no new states to download
- Cases for when there are new states to download are covered by the existing sharding_udgrade tests


chain/chain/src/chain.rs
```diff
@@ -1331,17 +1283,12 @@ impl Chain {
...
                 chain_update.chain_store_update.save_block_height_processed(block_height);
                 chain_update.commit()?;
 
                 self.pending_states_to_patch = None;
 
-                if needs_to_start_fetching_state {
-                    debug!(target: "chain", "Downloading state for block {}", block.hash());
-                    self.start_downloading_state(me, &block)?;
-                }
-
```
bowenwang1996, line 1344: Where is `start_downloading_state` called if it is removed here?

chain/chain/src/chain.rs
```diff
@@ -4173,23 +4168,23 @@ impl<'a> ChainUpdate<'a> {
...
 
-                // For the first block of the epoch we never apply state for the next epoch, so it's
-                // always caught up.
-                (false, true)
-            } else {
-                (self.prev_block_is_caught_up(&prev_prev_hash, &prev_hash)?, false)
-            };
+            // For the first block of the epoch we check if we need to start download states for
+            // shards that we will care about in the next epoch, if so, we consider that we don't have
+            let has_state_to_download = self.start_downloading_state(me, block)?;
```
mzhangmzz, line 4181: @bowenwang1996  here

mzhangmzz, line 4181: @bowenwang1996  here

integration-tests/src/tests/client/process_blocks.rs
```diff
@@ -3429,6 +3429,29 @@ fn test_limit_contract_functions_number_upgrade() {
...
+        .build();
+    // run the chain to a few epochs and make sure no catch up is triggered and the chain still
+    // functions
+    for h in 1..20 {
+        let block = env.clients[0].produce_block(h).unwrap();
+        let (_, res) = env.clients[0].process_block(block.unwrap().into(), Provenance::PRODUCED);
+        res.unwrap();
+        assert_eq!(env.clients[0].chain.store().iterate_state_sync_infos(), vec![]);
+    }
+}
```
bowenwang1996, line 3453: Could we assert that blocks to catch up (`ColBlocksToCatchUp`) is also always empty?

bowenwang1996 (approved):

PR #6473, opened 2022-03-22
Make state-viewer optionally open the RocksDB in readwrite mode instead of readonly only

nikurt:


EdvardD (approved):

neard/src/cli.rs
```diff
@@ -78,6 +78,17 @@ impl NeardCmd {
     }
 }
 
+#[derive(Clap)]
+pub(super) struct StateViewerCommand {
+    /// By default state viewer opens rocks DB in the read only mode, which allows it to run
+    /// multiple instances in parallel and be sure that no unintended changes get written to the DB.
+    /// In case an operation needs to write to caches, a read-write mode may be needed.
```
Longarithm, line 85: To add more context here:

On apply_range, we may hit an event where Wasmer needs to write contract into `CompiledContractCache`. And if readonly mode is enabled, it will fail. cc @matklad 

Example: `neard view_state apply_range --start-index 55290678 --end-index 55290878 --shard-id 0 --sequential --verbose-output`

Could we say here that read-write mode is needed only on the first run? Because I believe this is the case (though I didn't test it) and it may be convenient for users to use readonly mode and know that key-values won't be overwritten since the second run.

matklad: It might be ok to just ignore “db not writable” errors in compiled contract cache, if we can clearly distinguish it.

Longarithm: Then it will cause contract to recompile on each funcall, and we will see unrealistic performance for such range. IMO it is less preferrable.

matklad: Good point. I *think* when opening a db there’s a way to specify which columns are read only. It should be safe to exclude the one which holds compiled contracts from that set.

nikurt: Is your suggestion to have 2 modes: read-only and only-compiled-contracts-are-read-write?
I think state viewer strongly needs a read-only mode to make sure we don't accidentally corrupt a DB.

nikurt: > Could we say here that read-write mode is needed only on the first run?

I don't think we should hardcode this assumption, because I like the tools to have a simple behavior.

mm-near (approved):

matklad (approved):

PR #6477, opened 2022-03-23
neard: recompress_storage: clean out unnecessary columns

mina86:
When recompressing database of an archival node, skip
ColPartialChunks, ColInvalidChunks and ColTrieChanges columns which
can be safely deleted.  Data in the first one can be reconstructed
from ColChunks, ColInvalidChunks is only needed at head and the last
is never read by archival nodes.

Mostly for testing, if someone wants to keep those columns,
offer --keep-partial-chunks, --keep-invalid-chunks
and --keep-trie-changes switches.  They are always on when dealing
with non-archival node.

Issue: https://github.com/near/nearcore/issues/6250


neard/src/cli.rs
```diff
@@ -493,12 +493,28 @@ pub(super) struct RecompressStorageSubCommand {
...
+
+    /// Delete data from ColPartialChunks column.  The option is ignored unless
+    /// running on archival node.
+    #[clap(long)]
+    clean_partial_chunks: bool,
+
+    /// Clean ColTrieChanges column.  The option is ignored unless running on
+    /// archival node.
+    #[clap(long)]
+    clean_trie_changes: bool,
```
bowenwang1996, line 505: I think it might be easier if they default to true. What do you think?

nearcore/src/lib.rs
```diff
@@ -500,9 +500,27 @@ pub fn start_with_config_and_synchronization(
...
+        .map_err(|err| anyhow::anyhow!("{}: {}", config_path.display(), err))?
+        .archive;
+    let mut skip_columns = Vec::new();
+    if archive && opts.clean_partial_chunks {
+        skip_columns.push(near_store::db::DBCol::ColPartialChunks);
+    }
+    if archive && opts.clean_trie_changes {
+        skip_columns.push(near_store::db::DBCol::ColTrieChanges);
+    }
+
```
bowenwang1996, line 523: Can we print a warning here that this may take a day to finish?

neard/src/cli.rs
```diff
@@ -493,12 +493,29 @@ pub(super) struct RecompressStorageSubCommand {
...
+    /// Keep data in ColTrieChanges column.  This is always true if node is
+    /// not an archival node.
+    #[clap(long)]
+    keep_trie_changes: bool,
 }
 
 impl RecompressStorageSubCommand {
     pub(super) fn run(self, home_dir: &Path) {
-        if let Err(err) = nearcore::recompress_storage(&home_dir, &self.output_dir) {
+        warn!(target: "neard", "Recompressing storage; note that this operation may take up to a day to finish.");
```
mm-near, line 510: and if it is disturbed, you have to start from scratch, right?

neard/src/cli.rs
```diff
@@ -493,12 +493,29 @@ pub(super) struct RecompressStorageSubCommand {
     /// Directory where to save new storage.
     #[clap(long)]
     output_dir: PathBuf,
+
+    /// Keep data in ColPartialChunks column.  This is always true if node is
+    /// not an archival node.
+    #[clap(long)]
+    keep_partial_chunks: bool,
```
mm-near, line 500: What is the default value here? false?

so if someone runs the recompress after we submit this PR (assuming they have archival node) - they will automatically 'lose' those columns, right?

Maybe we should wait with this PR until we finish the code for computing partial_chunks/trie changes on the fly?

neard/src/cli.rs
```diff
@@ -493,12 +493,29 @@ pub(super) struct RecompressStorageSubCommand {
...
+
+    /// Keep data in ColTrieChanges column.  This is always true if node is
+    /// not an archival node.
+    #[clap(long)]
+    keep_trie_changes: bool,
```
mm-near, line 505: maybe we should put 'archival' somewhere in the name?

mm-near (changes requested):

neard/src/cli.rs
mina86: Yep.

neard/src/cli.rs
mina86: > What is the default value here? false?

Yep.

> so if someone runs the recompress after we submit this PR (assuming they have archival node) - they will automatically 'lose' those columns, right?

Yep.

> Maybe we should wait with this PR until we finish the code for computing partial_chunks/trie changes on the fly?

The code to recompute partial chunks is already merged in #6377. For recomputing trie changes we decided we won’t bother at least for now. The column only matters when non-archival node is doing GC.

There is #6439 still pending but I think it doesn’t matter that much which one of those gets merged first.

neard/src/cli.rs
mina86: Do you have idea for the names? I think the current one is descriptive enough.

nearcore/src/lib.rs
```diff
@@ -589,9 +624,27 @@ pub fn recompress_storage(home_dir: &Path, dst_dir: &Path) -> anyhow::Result<()>
...
+        let tip: Option<near_primitives::block::Tip> =
+            src_store.get_ser(DBCol::ColBlockMisc, near_store::FINAL_HEAD_KEY)?;
+        if let Some(tip) = tip {
+            let chunk_tail: near_primitives::types::BlockHeight = tip.height;
+            info!("Setting chunk tail to {}", chunk_tail);
+            let mut store_update = dst_store.store_update();
+            store_update.set_ser(DBCol::ColBlockMisc, near_store::CHUNK_TAIL_KEY, &chunk_tail)?;
+            store_update.commit()?;
+        } else {
+            info!("Final head not set; leaving chunk tail be");
```
mm-near, line 641: what are the scenarios when we're missing the 'final head'?

neard/src/cli.rs
```diff
@@ -465,12 +465,37 @@ pub(super) struct RecompressStorageSubCommand {
...
 
 impl RecompressStorageSubCommand {
     pub(super) fn run(self, home_dir: &Path) {
-        if let Err(err) = nearcore::recompress_storage(&home_dir, &self.output_dir) {
+        warn!(target: "neard", "Recompressing storage; note that this operation may take up to a day to finish.");
+        let opts = nearcore::RecompressOpts {
+            dest_dir: self.output_dir,
+            keep_partial_chunks: self.keep_partial_chunks,
+            keep_invalid_chunks: self.keep_invalid_chunks,
+            keep_trie_changes: self.keep_trie_changes,
```
mm-near, line 494: for this column - let's figure out the cold storage migration first.

mm-near (dismissed):

nearcore/src/lib.rs
mina86: In any legitimate uses of recompress storage I can’t really think of any scenario. Do you think the code should just panic?

neard/src/cli.rs
mina86: After discussion we’ve agreed this column can go.

nearcore/src/lib.rs
```diff
@@ -589,9 +639,21 @@ pub fn recompress_storage(home_dir: &Path, dst_dir: &Path) -> anyhow::Result<()>
         store_update.commit()?;
     }
 
+    // If we’re not keeping ColPartialChunks, update chunk tail to point to
+    // current final block.  If we don’t do that, the gc will try to work its
+    // way from the genesis even though chunks at those heights have been
+    // deleted.
+    if skip_columns.contains(&DBCol::ColPartialChunks) {
+        let chunk_tail = final_head_height.unwrap();
+        info!("Setting chunk tail to {}", chunk_tail);
```
nagisa, line 648:
```suggestion
        info!(target: "recompress"; %chunk_tail, "Setting chunk tail");
```

nearcore/src/lib.rs
```diff
@@ -537,21 +561,47 @@ pub fn recompress_storage(home_dir: &Path, dst_dir: &Path) -> anyhow::Result<()>
...
 
     for (n, column) in DBCol::iter().enumerate() {
+        if skip_columns.contains(&column) {
+            info!(
+                "Clearing out col{} ‘{}’ ({:2} / {:2})",
+                column as usize,
+                column,
+                n + 1,
+                DBCol::COUNT
+            );
```
nagisa, line 601:
```suggestion
            info!(
                column_id = column as usize,
                %column,
                current_column = n + 1,
                column_count = DBCol::COUNT,
                "Clearing a column",                
            );
```

PR #6479, opened 2022-03-23
moved 'Peer bandwidth exceeded threshold' warning from WARN to DEBUG

pompon0:
Since logging bandwidth stats is just a temporary measure (the real one will be introduction of throttling + rate limiting) there is no point in investing time to implement conditional logging (depending on whether or not sync_status.is_syncing()). I'm just switching WARN to DEBUG to decrease the verbosity for now.


bowenwang1996 (approved):

chain/network/src/peer_manager/peer_manager_actor.rs
```diff
@@ -397,7 +397,7 @@ impl PeerManagerActor {
             if bandwidth_used > REPORT_BANDWIDTH_THRESHOLD_BYTES
                 || total_msg_received_count > REPORT_BANDWIDTH_THRESHOLD_COUNT
             {
-                warn!(
+                debug!(
```
mm-near, line 400: also add 'target: "bandwith" - this way we'll be able to enable this log selectively

mm-near (approved):

chain/network/src/peer_manager/peer_manager_actor.rs
pompon0: done

PR #6484, opened 2022-03-23
feat: estimator-warehouse

jakmeier:
@matklad  This is the first half of the proposed continuous estimation infrastructure. I feel like this is enclosed in itself enough to discuss as its own PR and merge it as is.

Included is:
*  A "data warehouse" where estimation results from many runs over a long period of time can be stored and easily accessed. 
* `run_estimation.sh` to glue the estimator and the warehouse together.
* PoC for sending messages to [pagoda/contract-runtime/ce](https://near.zulipchat.com/#narrow/stream/319057-pagoda.2Fcontract-runtime.2Fce)

Not included yet:
* Configuration of what notification we actually care about in Zulip and who to ping.
* Anything related to the website that displays the time series

I recommend looking at the 2 new READMEs first for an overview, before jumping into code.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -80,7 +80,7 @@ dependencies = [
  "h2",
  "http",
  "httparse",
- "itoa",
+ "itoa 0.4.8",
  "language-tags",
  "local-channel",
  "log",
@@ -213,7 +213,7 @@ dependencies = [
  "encoding_rs",
  "futures-core",
  "futures-util",
- "itoa",
+ "itoa 0.4.8",
  "language-tags",
  "log",
  "mime",
@@ -397,7 +397,7 @@ dependencies = [
  "cookie",
  "derive_more",
  "futures-core",
- "itoa",
+ "itoa 0.4.8",
  "log",
  "mime",
  "percent-encoding",
@@ -1183,7 +1183,7 @@ checksum = "22813a6dc45b335f9bade10bf7271dc477e81113e89eb251a0bc2a8a81c536e1"
 dependencies = [
  "bstr",
  "csv-core",
- "itoa",
+ "itoa 0.4.8",
  "ryu",
  "serde",
 ]
@@ -1899,7 +1899,7 @@ checksum = "1323096b05d41827dadeaee54c9981958c0f94e670bc94ed80037d1a7b8b186b"
 dependencies = [
  "bytes",
  "fnv",
- "itoa",
+ "itoa 0.4.8",
 ]
 
 [[package]]
@@ -1940,7 +1940,7 @@ dependencies = [
  "http-body",
  "httparse",
  "httpdate",
- "itoa",
+ "itoa 0.4.8",
  "pin-project-lite",
  "socket2",
  "tokio",
@@ -2148,6 +2148,12 @@ name = "itoa"
 version = "0.4.8"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "b71991ff56294aa922b450139ee08b3bfc70982c6b2c7562771375cf73542dd4"
+
+[[package]]
+name = "itoa"
+version = "1.0.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "1aab8fc367588b89dcee83ab0fd66b72b50b72fa1904d7095045ace2b0c81c35"
 
 [[package]]
 name = "jobserver"
@@ -4266,7 +4272,7 @@ dependencies = [
  "async-trait",
  "combine",
  "dtoa",
- "itoa",
+ "itoa 0.4.8",
  "percent-encoding",
  "sha1",
  "url",
@@ -4790,7 +4796,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "0f690853975602e1bfe1ccbf50504d67174e3bcf340f23b5ea9992e0587a52d8"
 dependencies = [
  "indexmap",
- "itoa",
+ "itoa 0.4.8",
  "ryu",
  "serde",
 ]
@@ -4802,7 +4808,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "edfa57a7f8d9c1d260a549e7224100f6c43d43f9103e06dd8b4095a9b2b43ce9"
 dependencies = [
  "form_urlencoded",
- "itoa",
+ "itoa 0.4.8",
  "ryu",
  "serde",
 ]


runtime/runtime-params-estimator/continuous-estimation/run_estimation.sh
```diff
@@ -0,0 +1,77 @@
...
+fi
+
+IMPORT_OPTIONS="--db ${SQLI_DB} import --commit-hash=${GIT_HASH} "
+
+"${ESTIMATOR_BINARY}" ${COMMON_ARGS} ${USE_LARGE_STATE} ${TIME_OPTIONS} --metric time \
+    | "${WAREHOUSE_BINARY}" ${IMPORT_OPTIONS};
+
+# Run icount measurements inside docker container.
+"${ESTIMATOR_BINARY}" ${COMMON_ARGS} ${ICOUNT_OPTIONS} --metric icount \
+    | "${WAREHOUSE_BINARY}" ${IMPORT_OPTIONS};
```
matklad, line 77: Can we pull this bash script into the wherehouse binary itself? Not a huge fan of mixing bash and Rust. You can use https://github.com/matklad/xshell to make the rust version of this relatively concise. 

Though, no strong opinion here -- if you prefer bash, then so be it!

runtime/runtime-params-estimator/estimator-warehouse/Cargo.toml
```diff
@@ -0,0 +1,17 @@
...
+rust-version = "1.56.0"
+
+[dependencies]
+anyhow = "1"
+chrono = "0.4"
+clap = {version = "3.1", features = ["derive"]}
+reqwest = {version = "0.11", features = ["blocking", "json"]}
+rusqlite = {version = "0.27", features = ["bundled", "chrono"]}
+serde = {version = "1", features = ["derive"]}
+serde_json = "1"
```
matklad, line 17: Minor, but, based on the following discussion:

https://users.rust-lang.org/t/psa-please-specify-precise-dependency-versions-in-cargo-toml/71277

I now prefer to always specify all three numbers in Cargo.toml

runtime/runtime-params-estimator/estimator-warehouse/src/check.rs
```diff
@@ -0,0 +1,135 @@
...
+    db::{EstimationRow, DB},
+    zulip::{ZulipEndpoint, ZulipReport},
+    Metric,
+};
+
+#[derive(Parser, Debug)]
+pub(crate) struct CheckConfig {
+    /// Send notifications from checks to specified server.
+    #[clap(long, default_value = "near.zulipchat.com")]
+    zulip_server: String,
```
matklad, line 15: Do we need this to be a CLI option?

runtime/runtime-params-estimator/estimator-warehouse/src/check.rs
```diff
@@ -0,0 +1,135 @@
...
+    } else {
+        let rows_a = EstimationRow::select_by_commit_and_metric(db, &commit_a, config.metric)?;
+        let rows_b = EstimationRow::select_by_commit_and_metric(db, &commit_b, config.metric)?;
+        let estimations_a = rows_a.into_iter().map(|row| row.name);
+        let estimations_b = rows_b.into_iter().map(|row| row.name);
+        let mut intersection = vec![];
+        let lut = estimations_a.collect::<BTreeSet<_>>();
+        for b in estimations_b {
+            if lut.contains(&b) {
+                intersection.push(b);
```
matklad, line 84: minor, but might use https://doc.rust-lang.org/stable/std/collections/struct.BTreeSet.html#method.intersection here

runtime/runtime-params-estimator/estimator-warehouse/src/check.rs
```diff
@@ -0,0 +1,135 @@
...
+    commit_a: &str,
+    commit_b: &str,
+    tolerance: f64,
+    metric: Metric,
+) -> anyhow::Result<Vec<Notice>> {
+    let mut warnings = Vec::new();
+    for name in estimation_names {
+        let a = &EstimationRow::get(db, name, commit_a, metric)?[0];
+        let b = &EstimationRow::get(db, name, commit_b, metric)?[0];
+        let rel_change = (a.gas - b.gas).abs() / a.gas;
```
matklad, line 124: Might want to protect against `/ 0` just in case?

runtime/runtime-params-estimator/estimator-warehouse/src/db.rs
```diff
@@ -0,0 +1,196 @@
...
+        Ok(Self { conn })
+    }
+    #[cfg(test)]
+    pub(crate) fn test() -> Self {
+        let conn = Connection::open_in_memory().unwrap();
+        let init_sql = include_str!("init.sql");
+        conn.execute(init_sql, []).unwrap();
+        Self { conn }
+    }
+    pub(crate) fn insert_estimation(&self, row: &EstimationRow) -> anyhow::Result<()> {
```
matklad, line 30: It's a bit inconsistent that other methods are on `EstimationRow`, but this one is on DB. I guess the difference is read-vs-update? Probably doesn't matter much either way.

runtime/runtime-params-estimator/estimator-warehouse/src/import.rs
```diff
@@ -0,0 +1,172 @@
...
+
+use crate::db::{EstimationRow, DB};
+
+/// Additional information required for import
+#[derive(Debug, Parser)]
+pub(crate) struct ImportConfig {
+    /// Required for importing estimation results, which source code commit it
+    /// should be associated with.
+    #[clap(long)]
+    pub commit_hash: Option<String>,
```
matklad, line 17: Feels like we can shell-out to git from this binary to get the hash?

matklad (approved):
LGTM!

I am not sure if this is the right arch, but we can always iterate here.

Though, tbh, it still feels that sqlite is probably an overkill here, and an .ndjson file would be enough. It seems that with serde and itertools, writing a bit of custom code to analyze stream of json events might actually be simpler than going through sqlite?

See also https://tailscale.com/blog/an-unlikely-database-migration/

But than, again, no strong opinion: we can try SQLite and if it just works, than it just works!

runtime/runtime-params-estimator/estimator-warehouse/src/zulip.rs
```diff
@@ -0,0 +1,122 @@
+use std::env;
+
+use anyhow::Context;
+use reqwest::blocking::Client;
+
+use crate::check::{Notice, RelativeChange, Status};
+
+pub(crate) struct ZulipEndpoint {
```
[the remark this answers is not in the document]
jakmeier: I was planning to run this on ALL estimations. Currently 78 estimations with two graphs each (for `icount` and `time`). The Zulip daily report will give an overview and links to relevant graphs.
Also, for the time being, I would expect myself to be the only person monitoring these reports actively.

runtime/runtime-params-estimator/continuous-estimation/run_estimation.sh
jakmeier: Xshell looks neat, I will give it a try and see how it goes. If we can remove bash from the equation, that seems overall positive.

Generally speaking though, for me personally bash and Rust are fundamentally different tools that often complement each other perfectly. (Rust) binaries have these clearly defined interfaces with CLI arguments and stdin/stdout. Bash is designed to use this interface effectively in clear and concise code. Rust's strong type system does not help much with that task and the re-compiling for every constant string value change is actively harming productivity.

runtime/runtime-params-estimator/estimator-warehouse/src/import.rs
jakmeier: For the daily job in buildkite, yes. But there are other cases to consider, where the binary can be from a different repository than the estimator. This has been the standard case while developing this (estimating master branch while working on my own branch).
Another case, at some point in the future we might figure out that feature X we just merged requires us to change some estimation functions. (E.g. new cache invalidates old estimation.) At that point, we want to compare "new estimation on new code" vs "new estimation on old code". It can always be done with some git history juggling but I would prefer my current solution.

runtime/runtime-params-estimator/estimator-warehouse/src/check.rs
jakmeier: Probably not. I will remove it. (I was thinking about development with a local Zulip instance. But then I discovered I can easily send to my private user instead of the stream.)

jakmeier:
> LGTM!
> 
> I am not sure if this is the right arch, but we can always iterate here.
> 
> Though, tbh, it still feels that sqlite is probably an overkill here, and an .ndjson file would be enough. It seems that with serde and itertools, writing a bit of custom code to analyze stream of json events might actually be simpler than going through sqlite?
> 
> See also https://tailscale.com/blog/an-unlikely-database-migration/
> 
> But than, again, no strong opinion: we can try SQLite and if it just works, than it just works!

Thanks for the feedback, I really appreciate to read your thoughts! SQLite was perfect for prototyping with bash + awk but maybe now it should be removed. I really don't want to over engineer this and pull in unnecessary dependencies/tools.

I completely agree that for what is included in this PR, SQLite is overkill by a landslide. But the exports from the warehouse that I will include in the next PR, will contain queries that are at the edge of what I find reasonable of implementing in Rust, when we could also use existing query tooling.

In the end, things should be as simple as possible but no simpler. (_paraphrased quote from unknown source, some say Einstein_)

I see two options to move forward.
1. Merge PR 1 with SQLite and remove later if possible
1. Wait until everything is ready

I thought merging in pieces will make reviewing easier. But maybe it just added confusion...

runtime/runtime-params-estimator/continuous-estimation/run_estimation.sh
matklad: > Bash is designed to use this interface effectively in clear and concise code. Rust's strong type system does not help much with that task and the re-compiling for every constant string value change is actively harming productivity.

Yeah, that's true. My argument here is that, *despite* bash being a better tool for this particular job, using Rust might be overall better, because we'll stick to a single language/ecosystem. Basically, this: https://twitter.com/id_aa_carmack/status/989951283900514304

No strong opinion though!

matklad:
> I see two options to move forward

Let's do 1 -- it seems that that's what you are leaning to, and I trust your judgement here. I think this has all the required reviews.

runtime/runtime-params-estimator/continuous-estimation/run_estimation.sh
jakmeier: Thanks to xshell, I was able to remove bash from the process without too much of a hassle. Although it was more work than the bash script and I am not sure which is more readable, I do really like the fact that now I can do `estimator-warehouse estimate` instead of calling a bash script. Overall it was worth it IMO.

Before merging this, can you please take a look at the new file `estimate.rs` that replaced `run_estimation.sh`? It feels like I probably missed better ways to do it in a few places, so if you see something obvious it would be appreciated.

jakmeier:
> Given
> 
> > First batch of continuous estimation ready to merge
> 
> I assume you forgot to add auto-merge label?

Well, I wanted to give you chance to look at `estimate.rs` and my first attempt of using xshell. But if it is merged now, then it is merged now. :)

matklad:
Ah, I missed that there was substantial update, thought that was just a rebase.

runtime/runtime-params-estimator/estimator-warehouse/src/estimate.rs
```diff
@@ -0,0 +1,101 @@
...
+            maybe_tmp.as_ref().unwrap().path().to_str().unwrap()
+        }
+    };
+
+    if let Some(external_repo) = &config.external_repo {
+        sh.change_dir(external_repo);
+    }
+    let output = cmd!(sh, "git rev-parse --show-toplevel").output()?;
+    let mut git_root = String::from_utf8_lossy(&output.stdout).to_string();
+    git_root.pop(); // \n
```
matklad, line 37:
```suggestion
    let output = cmd!(sh, "git rev-parse --show-toplevel").read()?;
```

It does `.pop` thing

rust-toolchain.toml
```diff
@@ -2,6 +2,6 @@
 # This specifies the version of Rust we use to build.
 # Individual crates in the workspace may support a lower version, as indicated by `rust-version` field in each crate's `Cargo.toml`.
 # The version specified below, should be at least as high as the maximum `rust-version` within the workspace.
-channel = "1.58.1"
```
matklad, line 5: Curious, was this intentional? :)

jakmeier: Oh yeah, that one.^^ I guess it was semi-intentional. Xshell wouldn't work with 1.58.1 so I had to lift the requirement. But I wanted to ask you explicitly about this since I didn't know what the best approach was. That, I completely forgot.

matklad: Heh, we both missed that. Yeah, it's better to land compiler upgrades in a separate PRs, as things might get wrong

PR #6486, opened 2022-03-23
feat: return error if building test contract fails

Longarithm:
I realized that if building test contract fails, we don't notice it, because it just results in non-zero status.

The PR should fix it.

### DIFF
--- a/runtime/runtime-params-estimator/src/main.rs
+++ b/runtime/runtime-params-estimator/src/main.rs
@@ -1,6 +1,6 @@
 #![doc = include_str!("../README.md")]
 
-use anyhow::Context;
+use anyhow::{Context, Error};
 use clap::Parser;
 use genesis_populate::GenesisBuilder;
 use near_chain_configs::GenesisValidationMode;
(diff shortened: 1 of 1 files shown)


runtime/runtime-params-estimator/src/main.rs
```diff
@@ -98,10 +99,19 @@ fn main() -> anyhow::Result<()> {
...
+        let result = std::process::Command::new(build_test_contract)
             .current_dir(estimator_dir)
             .output()
             .context("could not build test contract")?;
+        if !result.status.success() {
+            return Err(Error::from(std::io::Error::new(
+                std::io::ErrorKind::InvalidData,
+                format!(
+                    "Failed to build test contract: {}",
+                    str::from_utf8(&result.stderr).unwrap()
```
jakmeier, line 111:
```suggestion
                    str::from_utf8_lossy(&result.stderr)
```

jakmeier (approved):
LGTM, thanks for fixing this!

Longarithm:
I just remembered that `anyhow::bail` works here, which make the code even simpler.

PR #6487, opened 2022-03-24
Adding metrics for counting and timing client messages

mzhangmzz:
Metric here
https://grafana.near.org/d/zzGDUdcnk/block-processing-and-chunk-delay?orgId=1&refresh=1m&from=now-30m&to=now&var-chain_id=testnet&var-role=min-testing&var-node_id=All
Screen Shot 2022-03-23 at 10 00 45 PM


chain/client/src/metrics.rs
```diff
@@ -159,3 +159,20 @@ pub static PARTIAL_ENCODED_CHUNK_RESPONSE_DELAY: Lazy<Histogram> = Lazy::new(||
     )
         .unwrap()
 });
+pub static CLIENT_MESSAGES_COUNT: Lazy<IntCounterVec> = Lazy::new(|| {
+    try_create_int_counter_vec(
+        "nar_client_messages_count",
```
mm-near, line 164: typo? s/nar/near ?

mm-near (approved):

Issue #6436, opened 2022-03-15
Warn about unknown fields when deserialising JSON files


mina86:
Currently, if serde encounters an unknown field it silently ignores it. We might want to at least warn if this happens. I’ve just spent noticeable amount of time debugging an issue caused by a typo in `config.json` file.

kakoc:
If I understand correctly we may try to use that crate?:
https://github.com/dtolnay/serde-ignored

mina86:
Yes, looks exactly like what we want.

kakoc:
If so, am I right that we can do that in the following way:
We can create a de wrapper type and use it for debug deserializations?

```
impl<T> DeserializedWithDebug <T> {
    fn deserialize<'de, D>(deserializer: D) -> Result<T, D::Error>
    where
        T: Deserialize<'de>,
        D: Deserializer<'de>,
    {
        let mut unused = Set::new();
        let p: Package = serde_ignored::deserialize(jd, |path| {
            unused.insert(path.to_string());
        });

        debug!(unused);

        p
    }
}
```

mina86:
That looks reasonable. The exact formatting for the debug message would need some work of course and maybe even have the function return `(T, Vec<String>)` rather than printing the unrecognised keys, but in general seems to be what we want.

jancionear:
It looks like this has been implemented in https://github.com/near/nearcore/pull/6490, but I don't get any warnings when I add an invalid option to the config. Opened another issue about it: https://github.com/near/nearcore/issues/10309

PR #6490, opened 2022-03-25
feat(serde): show missed de fields in debug mode #6436

kakoc:
https://github.com/near/nearcore/issues/6436

@mina86

### DIFF
--- a/nearcore/src/config.rs
+++ b/nearcore/src/config.rs
@@ -15,7 +15,7 @@ use serde::{Deserialize, Serialize};
 #[cfg(test)]
 use tempfile::tempdir;
 use tokio::io::AsyncWriteExt;
-use tracing::{error, info};
+use tracing::{error, info, warn};
 
 use near_chain_configs::{
     get_initial_supply, ClientConfig, Genesis, GenesisConfig, GenesisValidationMode,
@@ -493,10 +493,19 @@ impl Default for Config {
 
 impl Config {
     pub fn from_file(path: &Path) -> anyhow::Result<Self> {
+        let mut unrecognised_fields = Vec::new();
         let s = std::fs::read_to_string(path)
             .with_context(|| format!("Failed to read config from {}", path.display()))?;
-        let config = serde_json::from_str(&s)
+        let config =
+            serde_ignored::deserialize(&mut serde_json::Deserializer::from_str(&s), |path| {
+                unrecognised_fields.push(path.to_string());
+            })
             .with_context(|| format!("Failed to deserialize config from {}", path.display()))?;
+
+        if !unrecognised_fields.is_empty() {
+            warn!("{}: encountered unrecognised fields: {:?}", path.display(), unrecognised_fields);
+        }
+
         Ok(config)
     }
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3306,6 +3306,7 @@ dependencies = [
  "rayon",
  "rlimit",
  "serde",
+ "serde_ignored",
  "serde_json",
  "smart-default",
  "strum",
@@ -4669,6 +4670,15 @@ dependencies = [
  "proc-macro2",
  "quote",
  "syn",
+]
+
+[[package]]
+name = "serde_ignored"
+version = "0.1.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "1c2c7d39d14f2f2ea82239de71594782f186fd03501ac81f0ce08e674819ff2f"
+dependencies = [
+ "serde",
 ]
 
 [[package]]
--- a/nearcore/Cargo.toml
+++ b/nearcore/Cargo.toml
@@ -26,6 +26,7 @@ rand = "0.7"
 rayon = "1.5"
 serde = { version = "1", features = ["derive"] }
 serde_json = "1"
+serde_ignored = "0.1"
 dirs = "3"
 borsh = "0.9"
 thiserror = "1.0"


core/primitives-core/src/serialize.rs
```diff
@@ -212,3 +216,27 @@ pub mod option_u128_dec_format {
...
+#[derive(Debug, Deserialize)]
+pub struct DeserializedWithDebug<T>(pub T);
+
+impl<T> DeserializedWithDebug<T> {
+    fn _deserialize<'de, D>(deserializer: D) -> Result<T, D::Error>
+    where
+        T: Deserialize<'de>,
+        D: Deserializer<'de>,
+    {
+        if cfg!(debug_assertions) {
```
mina86, line 229: We want the logging to be always present.

tools/rpctypegen/macro/src/lib.rs
```diff
@@ -38,8 +38,11 @@ impl Drop for Schema {
         let schema_json = serde_json::to_value(self).expect("Schema serialize failed");
         let new_schema_json = if let Ok(data) = std::fs::read(filename) {
             // merge to the existing file
-            let mut existing_schema = serde_json::from_slice::<Value>(&data)
-                .expect("cannot deserialize target/existing_schema.json");
+            let mut existing_schema = serde_json::from_slice::<
+                near_primitives_core::serialize::DeserializedWithDebug<Value>,
```
nagisa, line 42: `serde_json::Value` will never ignore any keys.

tools/rpctypegen/core/src/lib.rs
```diff
@@ -114,8 +116,9 @@ mod tests {
         })
         .unwrap();
         parse_error_type(&mut schema, &error_type);
-        let expected: BTreeMap<String, ErrorType> = serde_json::from_str(
-            r#"
+        let expected: BTreeMap<String, ErrorType> =
+            serde_json::from_str::<DeserializedWithDebug<_>>(
```
nagisa, line 120: Similarly, keys won't be ignored when deserializing to things like `BTreeMap`. I believe deserialization will just fail if the type is wrong or something.

runtime/near-vm-runner-standalone/src/script.rs
```diff
@@ -81,7 +82,7 @@ impl Script {
 
     pub(crate) fn vm_config_from_file(&mut self, path: &Path) {
         let data = fs::read(path).unwrap();
-        let vm_config = serde_json::from_slice(&data).unwrap();
+        let vm_config = serde_json::from_slice::<DeserializedWithDebug<_>>(&data).unwrap().0;
```
nagisa, line 85: Would it help to keep the code more straightforward if some sort of a helper like below was added? 

```
pub fn near_primitives::serialize::from_json_slice<T>(data: &[u8]) -> Result<T, ...> {
    serde_json::from_slice::<DeserializedWithDebug<T>>(data)
}
```

That way the straightforward thing to solve the problem would also by default do the right thing (introduce various log hooks) without anybody needing to go out of their way to opt-in.

core/primitives-core/src/serialize.rs
```diff
@@ -212,3 +216,27 @@ pub mod option_u128_dec_format {
...
+            let mut unused_fields = Vec::new();
+            let de_result: Result<T, D::Error> = serde_ignored::deserialize(deserializer, |path| {
+                unused_fields.push(path.to_string());
+            });
+
+            debug!("missed fields while deserializing: {:?}", unused_fields);
```
mina86, line 235: Probably `warn!`.

Also an issue here is that this does not include information about what we were deserialising. Ideally it would be: `warn!("{}: encountered unrecognised fields: {:?}", path.display(), unused_fields);`. Perhaps it would make sense to change `DeserializedWithDebug` to have two fields? E.g. `pub struct DeserializedWithDebug<T>(pub T, pub Vec<String>)` so that the caller can decide how to format the warning? Though then we would be repeating pretty much the same `warn!` code in multiple places. Now that I think about it perhaps what we actually want is `load_json_from_file` function which would read the file, deserialise, log the warning and return the object?

core/primitives-core/src/serialize.rs
nagisa, line 229: `debug_assertions` will largely be disabled 99/100 times, as running a non-optimized near code is not all that feasible, and rarely anybody will go out of their way to set `-Cdebug-assertions=yes`.

chain/jsonrpc/client/src/lib.rs
```diff
@@ -92,7 +93,11 @@ where
             response.body().map(|body| match body {
                 Ok(bytes) => String::from_utf8(bytes.to_vec())
                     .map_err(|err| format!("Error {:?} in {:?}", err, bytes))
-                    .and_then(|s| serde_json::from_str(&s).map_err(|err| err.to_string())),
+                    .and_then(|s| {
+                        serde_json::from_str::<DeserializedWithDebug<_>>(&s)
+                            .map(|v| v.0)
+                            .map_err(|err| err.to_string())
+                    }),
```
mina86, line 100: So this is data that comes from outside so we probably don’t actually care. We need the warnings when reading local files or data from the database.

core/crypto/src/signature.rs
```diff
@@ -912,22 +914,29 @@ mod tests {
         let pk = sk.public_key();
         let expected = "\"ed25519:DcA2MzgpJbrUATQLLceocVckhhAqrkingax4oJ9kZ847\"";
         assert_eq!(serde_json::to_string(&pk).unwrap(), expected);
-        assert_eq!(pk, serde_json::from_str(expected).unwrap());
+        assert_eq!(pk, serde_json::from_str::<DeserializedWithDebug<_>>(expected).unwrap().0);
```
mina86, line 917: I think I would leave tests as is to be honest.

core/primitives-core/src/serialize.rs
nagisa, line 235: This might want a `target` specific to this functionality and also perhaps a higher level (`info`? `warn`?). That way anybody who doesn't want to see these messages can disable them by setting a larger level filter for that target, while for people who are running `neard` or other utilities for the first time, the message would readily come up.

```suggestion
            info!(target: "near::serialization::missing_fields", message = "unknown fields while deserializing", ?unused_fields);
```

perhaps?

matklad:
I think the following point by mina86 is salient: https://github.com/near/nearcore/pull/6490#discussion_r835222395

Let's start with adding this to a specific place where we care about this most, and then scale it up to more places if it works great. I *think* the most impactful place to have this logic is

https://github.com/matklad/nearcore/blob/master/nearcore/src/config.rs#L496-L503

I'd also start with using the lightweight `serde_ignored::deserialize` API, without necessary generalizing it to a wrapper type.

chain/jsonrpc/client/src/lib.rs
matklad: @mina86 could you clarify a bit about

> data from the database.

I am confused here -- data in the db is borsh-encoded, it's not self-describing, so we can't apply this to db data?

mina86: Well, actually, VERSION_KEY is in JSON. ;) But yes, you’re right in general. I didn’t have anything specific in mind here. Just wanted to point out that we don’t want to warn on data controlled by third parties. Only on data that comes from us.

mina86:
> https://github.com/matklad/nearcore/blob/master/nearcore/src/config.rs#L496-L503

That’s 404.

matklad (approved):

nearcore/src/config.rs
```diff
@@ -487,10 +487,24 @@ impl Default for Config {
...
         let s = std::fs::read_to_string(path)
             .with_context(|| format!("Failed to read config from {}", path.display()))?;
-        let config = serde_json::from_str(&s)
+        let config =
+            serde_ignored::deserialize(&mut serde_json::Deserializer::from_str(&s), |path| {
+                unrecognised_fields.push(path.to_string());
+            })
             .with_context(|| format!("Failed to deserialize config from {}", path.display()))?;
+
+        if unrecognised_fields.len() > 0 {
```
mina86, line 499: `!unrecognised_fields.is_empty()`

nearcore/src/config.rs
```diff
@@ -487,10 +487,24 @@ impl Default for Config {
...
+            warn!(
+                "While trying to deserialize a config from {},
+		the following unrecognised fields were encountered: {:?}",
```
mina86, line 502: This should be single line. You can terminate line with ` \` if you want to wrap it. But then also the alignment is kinda weird.

mina86 (approved):

mm-near (approved):

PR #6491, opened 2022-03-25
feat(params-estimator): Block overhead consistency

jakmeier:
Most estimations execute transactions. The measurment includes
as many blocks as necessary to execute all receipts.

This change does two things.
1. Mark as uncertain if overhead of applying blocks is significant.
2. Force estimation function to define over how many blocks it executes.

### DIFF
--- a/runtime/runtime-params-estimator/src/estimator_context.rs
+++ b/runtime/runtime-params-estimator/src/estimator_context.rs
@@ -70,13 +70,13 @@ impl<'c> Testbed<'c> {
     /// might trigger multiple blocks in execution. The returned results are
     /// exactly one per input block, regardless of how many blocks needed to be
     /// executed. To avoid surprises in how many blocks are actually executed,
-    /// `blocks_per_block` must be specified and the function will panic if it
-    /// is wrong.
+    /// `block_latency` must be specified and the function will panic if it is
+    /// wrong. A latency of 0 means everything is done within a single block.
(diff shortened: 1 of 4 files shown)


runtime/runtime-params-estimator/src/estimator_context.rs
```diff
@@ -65,23 +65,34 @@ impl<'c> Testbed<'c> {
...
+    /// might trigger multiple blocks in execution. The returned results are
+    /// exactly one per input block, regardless of how many blocks needed to be
+    /// executed. To avoid surprises in how many blocks are actually executed,
+    /// `blocks_per_block` must be specified and the function will panic if it
+    /// is wrong.
+    #[track_caller]
     pub(crate) fn measure_blocks<'a>(
         &'a mut self,
         blocks: Vec<Vec<SignedTransaction>>,
+        blocks_per_block: usize,
```
matklad, line 79: Is there a better name here? Perhaps, `block_latency`?  We can than say that `0` latency means everything is in the same block.

matklad (approved):
Both are excellent ideas, thanks!

runtime/runtime-params-estimator/src/estimator_context.rs
jakmeier: Yeah, that's so much clearer. Thanks for the idea!

PR #6493, opened 2022-03-25
style: Add guidance on `tracing`

nagisa:


matklad:
Note that we have a similar guidence at https://github.com/near/nearcore/blob/master/docs/architecture.md#logging--observability

>Tracing supports format!-like arguments, but prefer key=value paris instead.

Now that we have style.md, let's maybe remove almost all tracing related contents from architecture.md, and leave just

```
# Observability

We use tracing for structured, hierarchical logging. Refer to the (style) document for usage guidance. 
```

docs/style.md
```diff
@@ -125,3 +126,89 @@ While convenient for editing, it may be poorly legible in unrendered form
...
+    }),
+    %hash,
+    "block.previous_hash" = %block.header().prev_hash(),
+    "block.height" = block.header().height(),
+    %peer_id,
+    was_requested
+);
+```
+
+Always specify the `target` explicitly.
```
nagisa, line 171: We should provide more guidance on how to pick a value for the `target`. Naming it after a crate in most cases makes sense, but sometimes it may make sense to diverge. We should also provide some example `target`s...

matklad: Yeah... The current rule is the minimal rule which keeps the code consistent with *pre-existing* practice.

docs/style.md
```diff
@@ -125,3 +126,89 @@ While convenient for editing, it may be poorly legible in unrendered form
...
+);
+```
+
+Most of the time an event will not need to format anything into the message text
+in order to pass more information:
+
+```rust
+debug!(
+    target: "client",
+    message = "Received block",
```
matklad, line 159:
```suggestion
    "Received block",
```

I think we currently consistently *don't* use `message =`?

docs/style.md
```diff
@@ -125,3 +126,89 @@ While convenient for editing, it may be poorly legible in unrendered form
...
+    target: "client",
+    "{:?} Received block {} <- {} at {} from {}, requested: {}",
+    self.client.validator_signer.as_ref().map(|vs| vs.validator_id()),
+    hash,
+    block.header().prev_hash(),
+    block.header().height(),
+    peer_id,
+    was_requested
+);
+```
```
matklad, line 151: Let's use the same style we use elsewhere in the document, with GOOD and BAD markers:

```rust
// GOOD

// BAD
```

Rationale: consistency. It also naively seems that *positive* guidelines, which explain what to do and show good example first, should work better.

docs/architecture.md
```diff
@@ -143,60 +143,10 @@ This file contains schema (DBCol) of our internal RocksDB storage - a good start
...
-```
-
-Always specify the `target` explicitly.
-
-The `INFO` level is enabled by default, use it for information useful for node
-operators. The `DEBUG` level is enabled on the canary nodes, use it for
-information useful in debugging testnet failures. The `TRACE` level is not
-generally enabled, use it for arbitrary debug output.
+The [tracing](https://tracing.rs) crate is used for structured, hierarchical event output
+and logging. We also integrate [Prometheus](https://prometheus.io) for light-weight metric output.
```
matklad, line 149: Let's link the style document here

matklad (approved):

PR #6494, opened 2022-03-25
network: introduce Blacklist type 

mina86:
Refactor code by moving blacklist creation and matching code into
a new Blacklist type.  While at it, print warning whenever blacklist
address is not parsed correctly (rather than silently ignoring it as
it was done previously).

### DIFF
--- a/chain/network-primitives/src/blacklist.rs
+++ b/chain/network-primitives/src/blacklist.rs
@@ -1,0 +1,170 @@
+/// A blacklist for socket addresses.  Supports adding individual IP:port tuples
+/// to the blacklist or entire IPs.
+#[derive(Debug, Clone)]
+pub struct Blacklist(std::collections::HashMap<std::net::IpAddr, PortsSet>);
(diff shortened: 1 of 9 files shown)


mm-near (approved):

chain/network-primitives/src/blacklist.rs
```diff
@@ -0,0 +1,169 @@
+/// A blacklist for socket addresses.  Supports adding individual IP:port tuples
+/// to the blacklist or entire IPs.
+#[derive(Debug, Clone)]
+pub struct Blacklist(std::collections::HashMap<std::net::IpAddr, PortsSet>);
+
+impl Default for Blacklist {
+    fn default() -> Self {
+        Self(Default::default())
+    }
+}
```
bowenwang1996, line 10: Isn't this derivable?

mina86: D’oh.

PR #6495, opened 2022-03-25
Chain debug page: head & header head info

jc-near:
### Description
This PR adds the skeleton of the chain info debug page , and fills in the heights & hashes of the current head & header head.

### JIRA Issue
[CP-18](https://nearinc.atlassian.net/browse/CP-18)

### Test plan

 - Build with make neard
 - Under `nearcore` directory, launch localnet using command `nearup stop && nearup run localnet --binary-path ./target/release`
 - Open your browser and go to http://[host_address]:3030/debug/
 - Expect to see the link "Chain Info"
 - Click the link "Chain Info"
 - Go back to http://[host_address]:3030/debug/
 - Make sure other links aren't broken, either

### Screenshot on localnet
Screen Shot 2022-03-25 at 1 19 29 PM

### First round reviewer
 - @mzhangmzz 
 - @mm-near

### DIFF
--- a/chain/jsonrpc/res/chain_info.html
+++ b/chain/jsonrpc/res/chain_info.html
@@ -45,7 +45,7 @@
                     $('.js-current-header-head-height').text(header_head.height);
                 },
                 dataType: "json",
-                error: function (errMsg, textStatus, errorThrown) {
+                error: (errMsg, textStatus, errorThrown) => {
                     alert("Failed: " + textStatus + " :" + errorThrown);
                 },
                 contentType: "application/json; charset=utf-8",
--- a/chain/jsonrpc/res/chain_info.html
+++ b/chain/jsonrpc/res/chain_info.html
@@ -1,0 +1,103 @@
+<html>
+<head>
+    <style>
+        table {
+            width: 100%;
+            border-collapse: collapse;
+        }
+
+        table,
+        th,
+        td {
+            border: 1px solid black;
+        }
+
+        td {
+            text-align: left;
+            vertical-align: top;
+            padding: 8px;
+        }
+
+        th {
+            text-align: center;
+            vertical-align: center;
+            padding: 8px;
+            background-color: lightgrey;
+        }
+
+        tr.active {
+            background-color: #eff8bf;
+        }
+    </style>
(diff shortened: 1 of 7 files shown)


chain/jsonrpc/src/lib.rs
```diff
@@ -1489,6 +1502,10 @@ pub fn start_http(
             .service(last_blocks_html)
             .service(web::resource("/debug/api/sync_info").route(web::get().to(sync_info_handler)))
             .service(sync_info_html)
+            .service(
+                web::resource("/debug/api/chain_info").route(web::get().to(chain_info_handler)),
```
mm-near, line 1506: I think we start to abuse the "separate APIs"s too much :-)

as we end up with 3 apis (debug/api/sync_info, last_blocks and chain_info) - that return exactly the same JSON.

One option it simply to rename it to "debug/api/status" and then have all the page use this.

The other option is for each of the APIs to return a different object - but that might be a lot more work for now.

jc-near: Cool, I think it is the time to do the refactor as well. Let me proceed with that refactor.

mm-near (approved):

matklad (approved):
LGTM!

One worry I have here is that we now have a bunch of `.html` files which look kinda similar and contain some logical duplication (eg, all call the `/api/status` endpoint). 

One alternative design would be to code all debug pages as a single .html document with tabbed interface. Probably doesn't make sense to do right now, but we should consider that if, at any point, we'll start thinking about html templates and static site generators :)

chain/jsonrpc/res/chain_info.html
```diff
@@ -0,0 +1,103 @@
...
+                success: data => {
+                    let head = data.detailed_debug_status.current_head_status;
+                    let header_head = data.detailed_debug_status.current_header_head_status;
+                    $('.js-current-head-hash').text(head.hash);
+                    $('.js-current-head-height').text(head.height);
+                    $('.js-current-header-head-hash').text(header_head.hash);
+                    $('.js-current-header-head-height').text(header_head.height);
+                },
+                dataType: "json",
+                error: function (errMsg, textStatus, errorThrown) {
```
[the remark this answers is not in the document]
jc-near: Let me update this one then.

PR #6507, opened 2022-03-29
Tool to grab snapshot of specific accounts for an epoch

EgorKulikov:
@mzhangmzz please take a look

### DIFF
--- a/nearcore/src/runtime/mod.rs
+++ b/nearcore/src/runtime/mod.rs
@@ -537,7 +537,7 @@ impl NightshadeRuntime {
                 );
             }
 
-            self.snapshot_grabber.grab_transactions(&epoch_id, transactions);
+            self.snapshot_grabber.grab_transactions(&epoch_id, shard_id, transactions);
         }
 
         let apply_state = ApplyState {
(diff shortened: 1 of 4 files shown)


nearcore/src/runtime/snapshot_grabber.rs
```diff
@@ -0,0 +1,92 @@
...
+use near_primitives::types::{AccountId, EpochId, ShardId, StateRoot};
+use near_store::TrieIterator;
+use std::cell::Cell;
+use std::collections::HashSet;
+use std::fs;
+use std::fs::File;
+use std::sync::Mutex;
+use tracing::debug;
+
+pub struct SnapshotGrabber {
```
mm-near, line 16: how do you use this tool?

I was expecting a stand-alone binary, which you can 'point' at the existing database, and pass some parameters (like which accounts to take etc).

nearcore/src/runtime/mod.rs
```diff
@@ -517,6 +525,21 @@ impl NightshadeRuntime {
                epoch_height, &epoch_id, current_protocol_version, is_first_block_of_version,
         );
 
+        #[cfg(feature = "snapshot_grabber")]
```
mm-near, line 528: I'm surprised that we have to modify runtime code for this tool - I'd rather have this tool 'separate'

mm-near (changes requested):

akhi3030:
@EgorKulikov: this PR has not seen any activity in a long time. I am assuming is no longer being actively developed. I will close it. Please reopen when you are ready to work on it again.

PR #6508, opened 2022-03-29
enable rpc endpoint for mock_network

mzhangmzz:
Add rpc endpoint for the binary start_mock_network to expose debug page and metrics when running the binary


integration-tests/src/mock_network/setup.rs
```diff
@@ -125,7 +125,7 @@ impl FromStr for MockNetworkMode {
 pub fn setup_mock_network(
     client_home_dir: &Path,
     network_home_dir: &Path,
-    config: &NearConfig,
+    config: NearConfig,
```
jc-near, line 128: Asking to learn: any specific reason why we remove the borrowing?

jc-near (commented):
Do you mind putting in some background in the PR description so that I can better understand what is happening?

integration-tests/src/mock_network/setup.rs
mzhangmzz: It's because the line I added below requires cloning the config if I use the reference. I removed the borrowing to save the clone.

jc-near (approved):
LGTMing to unblock. One more question for learning: based on my undersyanding mock network is something different from what we typically use, so we need this PR to turn on the RPC endpoint. Is that correct?

jc-near:
Also, don't forget to fix CI!

integration-tests/src/mock_network/setup.rs
```diff
@@ -313,6 +313,15 @@ pub fn setup_mock_network(
                 target_height,
             )
         });
+    #[cfg(not(feature = "test_features"))]
```
mm-near, line 316: I'm surprised that rpc server depends on test_features -- can you add a comment explaining why? (or consider simply enabling it always)

mm-near (approved):

mzhangmzz:
> LGTMing to unblock. One more question for learning: based on my undersyanding mock network is something different from what we typically use, so we need this PR to turn on the RPC endpoint. Is that correct?

yep, that is correct

integration-tests/src/mock_network/setup.rs
mzhangmzz: yeah it's because with test_features, rpc server requires PeerManagerActor, which is not available here, so I simply just require no test feature here.

PR #6514, opened 2022-03-31
create epoch info debug page to show validators

jc-near:
### Description
This PR adds the skeleton of the epoch info debug page , and fills in the information of account IDs of the validators, as the first step.

### JIRA Issue
[CP-20](https://nearinc.atlassian.net/browse/CP-20)

### Test plan

 - Build with make neard
 - Under `nearcore` directory, launch localnet using command `nearup stop && nearup run localnet --binary-path ./target/release`
 - Open your browser and go to http://[host_address]:3030/debug/
 - Expect to see the link "Epoch Info"
 - Click the link "Epoch Info"
 - Go back to http://[host_address]:3030/debug/
 - Make sure other links aren't broken, either

### Screenshot on localnet
Screen Shot 2022-04-07 at 2 22 19 PM

### First round reviewer
 - @mzhangmzz 
 - @mm-near

### DIFF
--- a/chain/jsonrpc/res/epoch_info.html
+++ b/chain/jsonrpc/res/epoch_info.html
@@ -38,6 +38,8 @@
                 url: "/debug/api/status",
                 success: data => {
                     let epoch_info = data.detailed_debug_status.epoch_info;
+                    $('.js-epoch-id').text(epoch_info.epoch_id);
+                    $('.js-epoch-start-time').text(epoch_info.start_time);
                     epoch_info.validators.forEach((validator, index) =>
                         $('.js-tbody-validators').append($('<tr>')
                             .append($('<td>').append(validator.account_id))
@@ -60,7 +62,15 @@
     </h1>
     <h2>
         <p>
-            Current Validators
+            Current Epoch ID:
+            <span class="js-epoch-id"></span>
+        </p>
+        <p>
+            First block started at:
+            <span class="js-epoch-start-time"></span>
+        </p>
+        <p>
+            Validators
         </p>
     </h2>
 
--- a/chain/jsonrpc/res/epoch_info.html
+++ b/chain/jsonrpc/res/epoch_info.html
@@ -1,0 +1,76 @@
+<html>
+<head>
+    <style>
+        table {
+            width: 100%;
+            border-collapse: collapse;
+        }
+
+        table,
+        th,
+        td {
+            border: 1px solid black;
+        }
+
+        td {
+            text-align: left;
+            vertical-align: top;
+            padding: 8px;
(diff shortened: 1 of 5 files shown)


core/primitives/src/views.rs
```diff
@@ -378,6 +378,14 @@ impl From<Tip> for BlockStatusView {
     }
 }
 
+#[cfg_attr(feature = "deepsize_feature", derive(deepsize::DeepSizeOf))]
+#[derive(Serialize, Deserialize, Debug)]
+pub struct EpochInfoView {
+    pub epoch_id: CryptoHash,
+    pub start_time: String,
```
mm-near, line 385: what is the timezone? UTC?

chain/jsonrpc/res/epoch_info.html
```diff
@@ -0,0 +1,86 @@
...
+        Welcome to the Epoch Status page!
+    </h1>
+    <h2>
+        <p>
+            Current Epoch ID:
+            <span class="js-epoch-id"></span>
+        </p>
+        <p>
+            First block started at:
+            <span class="js-epoch-start-time"></span>
```
mm-near, line 70: can we also print the first block height and block hash ?

mm-near (approved):

core/primitives/src/views.rs
mzhangmzz, line 385: Can we use the actual Instant here instead of String. You can get around the DeepSizeFeature requirement by using this https://github.com/near/nearcore/pull/6467/files#diff-c0cb71be9c57de08ac1ac5752b7af7a55f7ff6226435f786d8199baecc0bb4f4R58

chain/jsonrpc/res/epoch_info.html
mzhangmzz: And how many blocks have been produced already in this epoch, and the number of blocks left

core/primitives/src/views.rs
jc-near: I believe this one includes the timezone. Check out the string in the screenshot!

jc-near: I think of this struct as a ready-to-display payload for javascript. The purpose here is to provide a value that frontend can just print without having to do any conversion. In that regard, maybe a string is the easiest.

mzhangmzz: I think what the javascript is getting is a json object, which is serialized by this Serialize trait here, so you don't have to worry about converting the inner fields into string. Letting the serialization be handled automatically help with the consistency of your output.

mzhangmzz (approved):

PR #6515, opened 2022-03-31
fix(buildkite) Fix docker conflicts with previous runs

mhalambek:


mm-near (approved):

.buildkite/pipeline.yml
```diff
@@ -15,6 +15,7 @@ steps:
 
   - label: "cargo test not integration-tests*"
     command: |
+      docker rm -f $(docker ps --filter "publish=6379" -q) | true
```
mm-near, line 18: I'd also add a comment where does the 'magic' 6379 comes from

rtsai123 (approved):

Issue #6483, opened 2022-03-23
Implement StateChangesRequest::AllAccountsChanges and use it in Rosetta RPC


mina86:
Currently in convert_block_to_transactions we’re first sending near_client::GetStateChangesInBlock to get list of all changed accounts and then send near_client::GetStateChanges asking for changes for all those accounts. This can be optimised by introducing a StateChangesRequest::AllAccountsChanges which would send buck all changes which touched accounts.

PR #6517, opened 2022-04-02
feat(rpc): add AllAccountsChanges to StateChangesRequest #6483

kakoc:
#6483

### DIFF
--- a/chain/chain/src/store.rs
+++ b/chain/chain/src/store.rs
@@ -32,7 +32,7 @@ use near_primitives::types::{
     StateChangesRequest,
 };
 use near_primitives::utils::{get_block_shard_id, index_to_bytes, to_timestamp};
-use near_primitives::views::LightClientBlockView;
+use near_primitives::views::{LightClientBlockView, StateChangeKindView};
 use near_store::{
     read_with_cache, ColBlock, ColBlockExtra, ColBlockHeader, ColBlockHeight, ColBlockInfo,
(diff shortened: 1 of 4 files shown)


chain/chain/src/store.rs
```diff
@@ -706,6 +706,45 @@ impl ChainStore {
...
+
+                let all_accounts_changes = self
+                    .get_state_changes(
+                        &block_header.hash,
+                        &StateChangesRequest::AccountChanges { account_ids: touched_account_ids },
+                    )
+                    .expect("All accounts state changes received");
+
+                all_accounts_changes
+            }
```
mina86, line 747: Moving the code verbatim doesn’t really buy us much though. This needs to be reimplemented such that it does only one iteration over the trie. This can be done by adding `TrieKey::AllAccounts` variant which would correspond to prefix `block_hash + col::ACCOUNT`. (Currently we have `TrieKey::Account` which corresponds to `black_hash + col::ACCOUNT + account_id`). Then, `AllAccountsChanges` would be handled by something along the lines of:


```rust
let data_key = TrieKey::AllAccounts;
let storage_key = KeyForStateChanges::from_trie_key(block_hash, &data_key);
let changes_per_key_prefix = storage_key.find_iter(&self.store);
let changes = StateChanges::from_account_changes(changes_per_key_prefix)?
```

I think this should just work.

mina86 (changes requested):

akhi3030:
@kakoc: this PR has not seen any activity in a long time. I am assuming is no longer being actively developed. I will close it. Please reopen when you are ready to work on it again.

PR #6523, opened 2022-04-04
add metrics for how long a block stay in the orphan pool and the miss…

mzhangmzz:
…ing chunks pool

I also reorganized the original ChunksDelayTracker a little, but didn't change any of the essential logic.
- I renamed ChunksDelayTracker to BlocksDelayTracker, as now it includes information about various important timestamps in blocks lifetime, instead of just tracking chunks delays. 
- I renamed a few methods.
- I had to move it from client to chain because I need to access it in chain.


chain/chain/src/blocks_delay_tracker.rs
```diff
@@ -0,0 +1,204 @@
...
+        self.update_block_chunks_requested_metric(processed_block_hash, &chunks);
+        self.update_chunks_metric(&chunks);
+
+        self.blocks_in_progress.remove(&processed_block_hash);
+        for chunk_hash in chunks {
+            self.chunks_in_progress.remove(chunk_hash);
+        }
+    }
+
+    fn update_various_block_metrics(&mut self, block_hash: &CryptoHash) {
```
mina86, line 148: ‘various’ feels weird. ‘update_block_metrics’ should be fine.

chain/chain/src/blocks_delay_tracker.rs
```diff
@@ -0,0 +1,204 @@
...
+        &mut self,
+        block_hash: &CryptoHash,
+        chunks: &[ChunkHash],
+    ) {
+        if let Some(block_received) = self.blocks_in_progress.get(&block_hash) {
+            for (shard_id, chunk_hash) in chunks.iter().enumerate() {
+                if let Some(chunk_in_progress) = self.chunks_in_progress.get(chunk_hash) {
+                    let chunk_requested = chunk_in_progress.chunk_requested;
+                    metrics::BLOCK_CHUNKS_REQUESTED_DELAY
+                        .with_label_values(&[&format!("{}", shard_id)])
```
mina86, line 179: `shard_id.to_string()`?

chain/chain/src/blocks_delay_tracker.rs
```diff
@@ -0,0 +1,204 @@
...
+        chunks: &[ChunkHash],
+    ) {
+        self.update_various_block_metrics(processed_block_hash);
+        self.update_block_chunks_requested_metric(processed_block_hash, &chunks);
+        self.update_chunks_metric(&chunks);
+
+        self.blocks_in_progress.remove(&processed_block_hash);
+        for chunk_hash in chunks {
+            self.chunks_in_progress.remove(chunk_hash);
+        }
```
mina86, line 145: This does multiple unnecessary lookups in hash tables. E.g. the block hash is looked up in self.blocks_in_progress three times in this method. Dunno, but I’d do something like:

```rust
let processed_block = self.blocks_in_progress.remove(&block_hash);
self.update_block_metrics(&processed_block);
for chunk_hash in chunks {
    let chunk = self.chunks_in_progress.remove(chunk_hash);
    self.update_chunk_metrics(&processed_block, &chunk);
}
```

mina86 (commented):
LGTM

chain/chain/src/blocks_delay_tracker.rs
```diff
@@ -0,0 +1,204 @@
...
+use near_primitives::hash::CryptoHash;
+use near_primitives::sharding::ChunkHash;
+use near_primitives::types::BlockHeight;
+use std::collections::HashMap;
+use std::time::Instant;
+use tracing::warn;
+
+use crate::metrics;
+
+/// Provides monitoring information about the important timestamps in throughout the lifetime of
```
mm-near, line 12: nit:

- remove 'in' (timestamp thorough)
- next line: s/has/have
- 3 & 4th line: Blocks are added -> Block is added, Chunk is added

chain/chain/src/blocks_delay_tracker.rs
```diff
@@ -0,0 +1,204 @@
...
+/// blocks and chunks. It keeps information of all pending blocks and chunks that has not been fully processed yet.
+/// Blocks are added the first time it is received and removed when it finishes processing
+/// Chunks are added the first time it was requested and removed when the block it belongs finishes processing
+#[derive(Debug, Default)]
+pub struct BlocksDelayTracker {
+    /// Timestamps of the first attempt of block processing.
```
mm-near, line 18: update comment

mm-near (approved):

chain/chain/src/blocks_delay_tracker.rs
```diff
@@ -0,0 +1,204 @@
...
+                        .observe(end.saturating_duration_since(start).as_secs_f64());
+                }
+            } else {
+                metrics::BLOCK_ORPHANED_DELAY.observe(0.);
+            }
+            if let Some(start) = block.missing_chunks_timestamp {
+                if let Some(end) = block.removed_from_missing_chunks_timestamp {
+                    metrics::BLOCK_MISSING_CHUNKS_DELAY
+                        .observe(end.saturating_duration_since(start).as_secs_f64());
+                }
```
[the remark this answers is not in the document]
mzhangmzz: In this case, it won't be included in the chunk. But if the chain code is implemented correctly, this should never happen.

chain/chain/src/blocks_delay_tracker.rs
```diff
@@ -0,0 +1,184 @@
...
+    pub removed_from_missing_chunks_timestamp: Option<Instant>,
+    /// Height at which the block is going to be added.
+    pub height: BlockHeight,
+    /// Hashes of the chunks that belong to this block.
+    pub chunks: Vec<ChunkHash>,
+}
+
+#[derive(Debug)]
+/// Records timestamps of requesting and receiving a chunk. Assumes that each chunk is requested
+/// before it is received.
```
matklad, line 45: Unreleated to PR, but `#[derive]` goes after comment

chain/chain/src/blocks_delay_tracker.rs
```diff
@@ -0,0 +1,184 @@
...
+            chunks,
+        });
+    }
+
+    pub fn mark_block_orphaned(&mut self, block_hash: &CryptoHash, timestamp: Instant) {
+        if let Some(block_entry) = self.blocks_in_progress.get_mut(block_hash) {
+            block_entry.orphaned_timestamp = Some(timestamp);
+        } else {
+            debug_assert!(false);
+            warn!(target:"blocks_delay_tracker", "block {:?} was orphaned but was not marked received", block_hash);
```
matklad, line 78: This feels like it should be `error!`, as this signals bug in the code

matklad: Also, perhaps the four functions can be compressed into the following API:

```rust
pub fn mark_block_orphaned(&mut self, block_hash: &CryptoHash, timestamp: Instant) {
  self.update_block(block_hash, |block_progress| block_progress.orphaned_timestamp = Some(timestamp));
}
```

chain/chain/src/blocks_delay_tracker.rs
```diff
@@ -0,0 +1,184 @@
...
+        requestor_block_hash: &CryptoHash,
+    ) {
+        self.chunks_in_progress.entry(chunk_hash.clone()).or_insert_with(|| ChunkInProgress {
+            chunk_requested: timestamp,
+            chunk_received: None,
+            block_hash: requestor_block_hash.clone(),
+        });
+    }
+
+    pub fn finish_block_processing(&mut self, block_hash: &CryptoHash, chunks: &[ChunkHash]) {
```
matklad, line 132: (pretty uninformed comment, take with a grain of salt)

It's curious that we update metrics only when we finish block processing. That is, if we never finish a block, it bypasses our metrics. Not sure if that is being done elsewhere, but, eg, logging the len of two hash maps every time we process a block might be interesting

matklad:
(to clarify: non of my comments are blocking, you've already got all reviews)

chain/chain/src/blocks_delay_tracker.rs
mzhangmzz: Yeah actually I think most other metrics are updated instantly. This is one of the few cases where we gather data and print it at the end of processing blocks.

Issue #6520, opened 2022-04-04
Switch ext costs to strum


matklad:
We use the following open-codded pattern to "reflect" on the number of elements in an enum.

https://github.com/near/nearcore/blob/cf6c3177dd0b0bc883038d5c320d92197dd9f315/core/primitives-core/src/config.rs#L615-L617

We should switch that to the `strum` crate, as we use it elsewhere. 

I am not entirely sure that using `strum` for these is worth it (the usual argument -- a little boilerplate compiles much faster than a little of proc-macro), but we definitely should be consitent, and we use `strum` today.

AnhTuDo1998:
Hi, I am new to NEAR and would like to try out this problem. 

I do notice that the enum ActionCosts in the same file also has this open-codded pattern. Should we fix this also ?

matklad:
Yes!

PR #6526, opened 2022-04-05
Switch __count field in Enum to strum

AnhTuDo1998:
fix: Use strum crate for consistency

This PR is for problem #6520. It replaces ```__count``` field in ```ExtCosts``` & ```ActionCosts``` Enums with ```strum::EnumCount``` and change codes where count of fields in Enum is needed. As we used strum crates in a few place in our project, this change is meant for consistency.

Compilation and existing unit test or ```core/primitives-core``` passed


core/primitives-core/src/config.rs
```diff
@@ -643,7 +638,7 @@ impl fmt::Display for ActionCosts {
 
 impl ActionCosts {
     pub const fn count() -> usize {
-        ActionCosts::__count as usize
+        ActionCosts::COUNT as usize
     }
```
matklad, line 642: Let's remove this method altogether

core/primitives-core/src/config.rs
```diff
@@ -738,13 +733,11 @@ impl ExtCosts {
             alt_bn128_g1_sum_base => config.alt_bn128_g1_sum_base,
             #[cfg(feature = "protocol_feature_alt_bn128")]
             alt_bn128_g1_sum_byte => config.alt_bn128_g1_sum_byte,
-
-            __count => unreachable!(),
         }
     }
 
     pub const fn count() -> usize {
```
matklad, line 739: and this one

AnhTuDo1998:
To test the changes in using ```strum::Display```, I have hardcoded locally a few ```println!()``` in unit test of profile.rs that previously use ```name_of()```. Here is a snippet of ```cargo test -- --nocapture```, let me know if this is alright

image

matklad:
Yeah, this looks good!

matklad (approved):

core/primitives-core/src/config.rs
```diff
@@ -611,13 +611,10 @@ pub enum ExtCosts {
...
     #[cfg(feature = "protocol_feature_alt_bn128")]
     alt_bn128_g1_sum_byte,
-
-    // NOTE: this should be the last element of the enum.
-    __count,
 }
 
 // Type of an action, used in fees logic.
-#[derive(Copy, Clone, Hash, PartialEq, Eq, Debug, PartialOrd, Ord)]
+#[derive(Copy, Clone, Hash, PartialEq, Eq, Debug, PartialOrd, Ord, EnumCount, Display)]
```
nagisa, line 617: I'd argue that “Display” is way too generic of a name to use in `derive` without referring to it by a crate (i.e. `strum::Display`), but this is probably a discussion to be had outside of this PR.

matklad: Uhu, same thoughts here

matklad: I'd even say this is not specific to `Display`: `strum` is a minor crate, with a short name, so we should just `strum::EnumCount, strum::Display`, etc.

nagisa: Yeah, I kinda prefer taking that approach to the extreme of “all external derives must be prefixed, and not imported", but either way this seems like a discussion best had over a changeset to the style guide.

PR #6527, opened 2022-04-05
Update tail bug fix

mm-near:
Fixing the bug in update tail that was causing the GC to stop processing forks (if they happened too far away from the epoch boundary) - until the next epoch.

This caused GC running time to be unstable.

Also added additional tests to actually test GC behavior and started setting correct epochs for the blocks within tests.


chain/chain/src/store.rs
```diff
@@ -1979,7 +1979,7 @@ impl<'a> ChainStoreUpdate<'a> {
         self.tail = Some(height);
         let genesis_height = self.get_genesis_height();
         // When fork tail is behind tail, it doesn't hurt to set it to tail for consistency.
-        if self.fork_tail.unwrap_or(genesis_height) < height {
+        if self.fork_tail().unwrap_or(genesis_height) < height {
```
mm-near, line 1982: this is the culprit of the problem .. ;-)

All the other changes here are test improvements.

mina86 (approved):

chain/chain/src/store.rs
matklad: Hm, now that this is a Result rather than an option, I wonder if this should use `?` rather than `.unwrap_or_else`, to not mask db failures and such.

The `or_else` already happens in the implementation of the method:

```
    fn fork_tail(&self) -> Result<BlockHeight, Error> {
        self.store
            .get_ser(ColBlockMisc, FORK_TAIL_KEY)
            .map(|option| option.unwrap_or_else(|| self.genesis_height))
            .map_err(|e| e.into())
    }
```

So I think error here always means catastrophic erro

mzhangmzz (approved):

chain/chain/src/test_utils.rs
```diff
@@ -1034,12 +1034,33 @@ impl RuntimeAdapter for KeyValueRuntime {
...
+            // first block of the epoch.
+            // Unfortunately many tests are depending on this and not setting epochs when
+            // they produce blocks.
             let block_height = self
                 .get_block_header(block_hash)
                 .unwrap_or_default()
                 .map(|h| h.height())
                 .unwrap_or_default();
             block_height.saturating_sub(NUM_EPOCHS_TO_KEEP_STORE_DATA * self.epoch_length)
+        /*  // TODO: use this version of the code instead - after we fix the block creation
```
bowenwang1996, line 1047: @mm-near Is there an issue that tracks this TODO

mm-near: good catch. Created CP-35

PR #6529, opened 2022-04-05
feat: update cost model for alt_bn128

matklad:
Two changes here:

* change cost morel from per-byte to per-elemen (these host functions deal with array of things)
* adapt inputs to the new serialization format

Note that this doesn't actually change the values of the cost themselves. I suggest we mesure those separatelly. We might also fix the code for measuring nightly host functions -- it doesn't work out of the box today.


jakmeier (approved):
Makes sense to me, a necessary follow-up to #6505 I suppose.

matklad:
In some sense it's even the opposite this cost model is a part of motivation for #6505. Before, we tried to measure the size of inputs in bytes, but, as the input was borsh-serialized, there were extra constant 4 bytes of the input. 

With borsh removed, the byte-size of the input now actually precisely corresponds to the number of elements.

PR #6531, opened 2022-04-05
epoch_manager: refactor get_and_update_epoch_info_aggregator

mina86:


chain/epoch_manager/src/lib.rs
```diff
@@ -1084,19 +1085,22 @@ impl EpochManager {
                 )
             }
             ValidatorInfoIdentifier::BlockHash(ref h) => {
-                let aggregator = self.get_and_update_epoch_info_aggregator(&epoch_id, h, true)?;
+                // If we are here, `h` is hash of the latest block of the
+                // current epoch.
+                self.update_epoch_info_aggregator(&epoch_id, h, None)?;
```
mzhangmzz, line 1090: I think it's fine to use `self.epoch_info_aggregator` directly here, instead of calling `update_epoch_info_aggregator`, since we are always accessing the latest block, and after your change `epoch_info_aggregator` will always be updated to chain head. That way, we can also change the `ValidatorInfoIdentifier::BlockHash` option to `ValidatorInfoIdentifier::LatestBlock`.

chain/epoch_manager/src/lib.rs
```diff
@@ -329,24 +343,16 @@ impl EpochManager {
...
-            all_proposals,
-            version_tracker,
-            ..
-        } = self.get_and_update_epoch_info_aggregator(
-            last_block_info.epoch_id(),
-            last_block_hash,
-            false,
-        )?;
+
+        self.update_epoch_info_aggregator(last_block_info.epoch_id(), last_block_hash, None)?;
```
mzhangmzz, line 347: Similarly, here we can just use &self.epoch_info_aggregator directly. This is because this function is called https://github.com/near/nearcore/blob/master/chain/epoch_manager/src/lib.rs#L641, right after we just updated epoch_info_aggregator with the latest block

chain/epoch_manager/src/lib.rs
```diff
@@ -1464,73 +1470,105 @@ impl EpochManager {
...
-                .get_ser(ColEpochInfo, AGGREGATOR_KEY)
-                .map_err(EpochError::from)?
-                .unwrap_or_else(|| EpochInfoAggregator::new(epoch_id.clone(), *last_block_hash))
-        };
-        if &aggregator.epoch_id != epoch_id {
-            aggregator = EpochInfoAggregator::new(epoch_id.clone(), *last_block_hash);
-            epoch_change = true;
+        store_update: Option<&mut StoreUpdate>,
+    ) -> Result<(), EpochError> {
+        debug_assert_eq!(epoch_id, self.get_block_info(last_block_hash)?.epoch_id());
```
mzhangmzz, line 1484: why not just get rid of epoch_id from the argument?

chain/epoch_manager/src/lib.rs
```diff
@@ -1464,73 +1470,105 @@ impl EpochManager {
...
-            epoch_change = true;
-            self.store
-                .get_ser(ColEpochInfo, AGGREGATOR_KEY)
-                .map_err(EpochError::from)?
-                .unwrap_or_else(|| EpochInfoAggregator::new(epoch_id.clone(), *last_block_hash))
-        };
-        if &aggregator.epoch_id != epoch_id {
-            aggregator = EpochInfoAggregator::new(epoch_id.clone(), *last_block_hash);
-            epoch_change = true;
+        store_update: Option<&mut StoreUpdate>,
```
mzhangmzz, line 1482: You can get rid of the option here if we access self.epoch_info_aggregator directly in the two places I commented above.

chain/epoch_manager/src/lib.rs
```diff
@@ -1464,73 +1470,105 @@ impl EpochManager {
...
+                self.epoch_info_aggregator_needs_saving = true;
+            }
         }
+
+        Ok(())
+    }
+
+    /// Updates epoch info aggregator to block as of `last_block_hash`.  Assumes
+    /// that aggregator needs updating, i.e. that `last_block_hash` does not
+    /// equal `self.epoch_info_aggregator_needs_saving.last_block_hash`.
```
mzhangmzz, line 1506: s/epoch_info_aggregator_needs_saving/epoch_info_aggregator

chain/epoch_manager/src/lib.rs
```diff
@@ -1464,73 +1470,105 @@ impl EpochManager {
...
-        };
-        if &aggregator.epoch_id != epoch_id {
-            aggregator = EpochInfoAggregator::new(epoch_id.clone(), *last_block_hash);
-            epoch_change = true;
+        store_update: Option<&mut StoreUpdate>,
+    ) -> Result<(), EpochError> {
+        debug_assert_eq!(epoch_id, self.get_block_info(last_block_hash)?.epoch_id());
+
+        let needs_saving = self.update_epoch_info_aggregator_impl(epoch_id, last_block_hash)?
+            || self.epoch_info_aggregator_needs_saving;
```
mzhangmzz, line 1487: Why do you need this additional flag here?

chain/epoch_manager/src/lib.rs
```diff
@@ -1464,73 +1470,105 @@ impl EpochManager {
...
-        store_update: &mut StoreUpdate,
-        aggregator: EpochInfoAggregator,
-        write_to_storage: bool,
-    ) -> Result<(), EpochError> {
-        if write_to_storage {
-            store_update.set_ser(ColEpochInfo, AGGREGATOR_KEY, &aggregator)?;
+        if overwrite {
+            self.epoch_info_aggregator = aggregator;
+        } else {
+            self.epoch_info_aggregator.merge(aggregator);
```
mzhangmzz, line 1567: Nice! I like this way better too.

mzhangmzz (changes requested):

mzhangmzz:
Great work! I think it already looks much better than the current code. I added a few suggestions to further simplify it.

chain/epoch_manager/src/lib.rs
mina86: > That way, we can also change the ValidatorInfoIdentifier::BlockHash option to ValidatorInfoIdentifier::LatestBlock.

So this is something I'm not clear about.  Normally we call the update on final blocks (https://github.com/near/nearcore/blob/master/chain/epoch_manager/src/lib.rs#L602) which is a few blocks behind latest block.  When we’re called with ValidatorInfoIdentifier::BlockHash we’re called with a block which hasn’t been aggregated yet, no?

mzhangmzz: Ah great observation, I didn't notice that. I think it's fine to use the latest final block, since right now the only use cases for get_validator_info using block hashes are in log_summary and in GetValidatorInfo, either cases, I don't think the distinction matters that much and I'd choose to keep the code simpler and cleaner over getting these two cases really accurate. We could use `ValidatorInfoIndentifier::LastestFinalBlock` to signal the correct meaning.

mina86:
So the longer I think about it the more I’m convinced this is wrong. It seems to me that we need two functions:
1. `update_epoch_info_aggregator` which is called from `record_block_info` and is only ever called on final blocks. This would update the `self.epoch_info_aggregator` field.
2. `get_epoch_info_aggregator` which would be called from `collect_block_info` and `get_validator_info` and would be called on latest block.
This way, `self.epoch_info_aggregator` would always store state up to the final block so that `get_epoch_info_aggregator` would not need to iterate through many block until it reaches epoch start or state already cached.

mzhangmzz:
> So the longer I think about it the more I’m convinced this is wrong. It seems to me that we need two functions:
> 
> 1. `update_epoch_info_aggregator` which is called from `record_block_info` and is only ever called on final blocks. This would update the `self.epoch_info_aggregator` field.
> 2. `get_epoch_info_aggregator` which would be called from `collect_block_info` and `get_validator_info` and would be called on latest block.
>    This way, `self.epoch_info_aggregator` would always store state up to the final block so that `get_epoch_info_aggregator` would not need to iterate through many block until it reaches epoch start or state already cached.

Yes I think you are right, I overlooked that `collect_blocks_info` needs to be called on latest block, not latest final blocks

chain/epoch_manager/src/lib.rs
mina86: Let’s leave it as `ValidatorInfoIndentifier::BlockHash` for now. I’ll look into simplifying it in another PR.

chain/epoch_manager/src/lib.rs
```diff
@@ -1464,73 +1440,151 @@ impl EpochManager {
...
-        if write_to_storage {
-            store_update.set_ser(ColEpochInfo, AGGREGATOR_KEY, &aggregator)?;
-        }
-        self.epoch_info_aggregator = Some(aggregator);
-        Ok(())
+            let prev_info = self.get_block_info(&prev_hash)?;
+            let prev_height = *prev_info.height();
+            let prev_epoch = prev_info.epoch_id().clone();
+
+            let block_info = self.get_block_info(&cur_hash)?;
```
mm-near, line 1576: you already fetched it in 1558

chain/epoch_manager/src/lib.rs
```diff
@@ -1464,73 +1440,151 @@ impl EpochManager {
...
-        } else {
-            self.epoch_info_aggregator.take()
-        };
-        let mut epoch_change = false;
-        let mut aggregator = if let Some(aggregator) = epoch_info_aggregator_cache {
-            aggregator
+        if let Some((mut aggregator, replace)) = self.aggregate_epoch_info_upto(last_block_hash)? {
+            if !replace {
+                aggregator.merge_prefix(&self.epoch_info_aggregator);
+                aggregator.last_block_hash = *last_block_hash;
```
mm-near, line 1494: I'm surprised that you have to set it here. Is 'merge_prefix' overwriting it?

chain/epoch_manager/src/types.rs
```diff
@@ -104,46 +103,62 @@ impl EpochInfoAggregator {
         }
     }
 
-    pub fn merge(&mut self, new_aggregator: EpochInfoAggregator, overwrite: bool) {
-        if self.epoch_id != new_aggregator.epoch_id {
-            debug_assert!(false);
-            error!(target: "epoch_manager", "Trying to merge an aggregator with epoch id {:?}, but our epoch id is {:?}", new_aggregator.epoch_id, self.epoch_id);
-            return;
+    pub fn merge(&mut self, other: EpochInfoAggregator) {
```
mm-near, line 106: comments pls. (here and on the other methods too)

chain/epoch_manager/src/lib.rs
mina86: `get_block_info` returns a reference (thus `block_info` has a hold on
`self`) so this does not compile:

```rust
            let block_info = self.get_block_info(&cur_hash)?;
            let prev_hash = *block_info.prev_hash();
            let prev_info = self.get_block_info(&prev_hash)?;
            aggregator.update(block_info, &epoch_info, prev_height);
```

That’s why we have to either get `block_info` again or clone it the
first time we fetch it.  The object isn’t small so fetching it again
seems a better option.

chain/chain/src/types.rs
```diff
@@ -771,7 +771,11 @@ pub struct LatestKnown {
     pub seen: u64,
 }
 
-/// Either an epoch id or latest block hash
+/// Either an epoch id or latest block hash.  When `EpochId` variant is used it
+/// must be an identifier of a past epoch.  When `BlockHeight` is used it must
+/// be hash of the latest block in the current epoch.  Using current epoch id
+/// with `EpochId` or arbitrary block hash in past or present epochs will result
+/// in errors.
```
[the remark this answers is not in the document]
mina86: I want to keep it as is for now and look into changing it into another PR. This PR is already quite involved.

chain/epoch_manager/src/lib.rs
mina86: Good catch. I meant to remove this once I introduced `merge_prefix`.

chain/epoch_manager/src/lib.rs
```diff
@@ -1462,73 +1441,150 @@ impl EpochManager {
...
+                *block_info.height() % AGGREGATOR_SAVE_PERIOD == 0
+            };
+            if save {
+                store_update.set_ser(ColEpochInfo, AGGREGATOR_KEY, &self.epoch_info_aggregator)?;
+            }
+        }
+        Ok(())
+    }
+
+    /// Returns epoch info aggregate with state up to `last_block_hash`.
```
mzhangmzz, line 1480: Thanks for the comments!

chain/epoch_manager/src/lib.rs
```diff
@@ -1462,73 +1441,150 @@ impl EpochManager {
...
+        block_hash: &CryptoHash,
+    ) -> Result<Option<(EpochInfoAggregator, bool)>, EpochError> {
+        if block_hash == &self.epoch_info_aggregator.last_block_hash {
+            return Ok(None);
+        }
+
+        if cfg!(debug) {
+            let agg_hash = self.epoch_info_aggregator.last_block_hash.clone();
+            let agg_height = *self.get_block_info(&agg_hash)?.height();
+            let block_height = *self.get_block_info(block_hash)?.height();
```
mzhangmzz, line 1533: Should we log an error here for the release binary too?

mzhangmzz (commented):
Great work! The code is sooo much better, easier to read and reason through. And I can't express how much I appreciate your comments. 

My only comments is that we need to add a test that would catch the previous bug. I didn't check the existing tests, maybe they do after you added `epoch_info_aggregator_loop_counter`. And if there are no such tests yet, add some to test forks and epoch switch.

chain/epoch_manager/src/lib.rs
mina86: So I chose to only do that in debug because of the get_block_info calls which are somewhat expensive (I mean the are likely just hash lookup but still it felt like something that could be avoided on release).

mina86:
I have taking a closer look at tests on my TODO. I don’t think that the current tests cover the issue we’ve encountered.

chain/epoch_manager/src/lib.rs
```diff
@@ -1462,73 +1441,150 @@ impl EpochManager {
...
+            assert!(
+                agg_height < block_height,
+                "#{} {} >= #{} {}",
+                agg_hash,
+                agg_height,
+                block_hash,
+                block_height
+            );
```
matklad, line 1541:
```suggestion
            assert!(
                agg_height < block_height,
                "#{agg_hash} {agg_height} >= #{block_hash} {block_height}",
            );
```

doesn't really matter, but we have string interpolation now!

mina86: Oh, God, I hate that feature.

matklad: Curious, why? For me, it makes the code so much more readable. Though, the fact that only identifiers are allowed creates a significant discontinuity -- you can't *always* use interpolation. 

It's also implemented in a non-transparent for tooling way, which makes IDE life harder

chain/epoch_manager/src/types.rs
```diff
@@ -104,46 +103,104 @@ impl EpochInfoAggregator {
...
 
-    pub fn merge(&mut self, new_aggregator: EpochInfoAggregator, overwrite: bool) {
-        if self.epoch_id != new_aggregator.epoch_id {
-            debug_assert!(false);
-            error!(target: "epoch_manager", "Trying to merge an aggregator with epoch id {:?}, but our epoch id is {:?}", new_aggregator.epoch_id, self.epoch_id);
-            return;
+    /// Merges information from `other` aggregator into `self`.
+    ///
+    /// The `other` aggregator must hold statistics from blocks which
+    /// **proceed** the ones this aggregator has.  Both aggregators have to be
```
mm-near, line 109: nit: while 'proceed' is correct - I think i'd write 'follow'

mm-near (approved):

chain/epoch_manager/src/lib.rs
mina86: It magically pulls in variables from the local scope with no indication that it’s what it’s doing.  Why is `let bar = format!("{foo}");` interpolating variable `foo` but `let bar = "{foo}";` is just a string?  How do I know if `some_random_macro!("{foo}");` will interpolate the variable?

matklad: Heh yes, that's exactly what IDE struggle with -- it'd be much better if it were first-class syntax, like in Kotlin.

Though, I must say the same is true of the usual interpolation: `"{}"` vs `format!("{}")`.

Syntax highlighting works-around here a bunch (well, with ligatures me):

image

That should be available in Emacs as well: https://github.com/emacs-lsp/lsp-mode/pull/1721

But yeah, it's a shame that we need to expand macros to figure that out.

PR #6532, opened 2022-04-06
jsonrpc: drop flatten specification for `RpcGasPriceRequest`

miraclx:
`RpcGasPriceRequest` attempts to flatten an untagged enum into a struct. In turn, causing serialization issues.

Reproducible Playground Link: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=8de6a7b9e6ddd60c4e8511ebb321d18f


miraclx:
Weird, GitHub closed this. It's still relevant.

matklad:
So it literally fails to serialize, and we didn't notice this?

LGTM, but maybe let's add a smoke tests here which pins down the serialization format? Removing `flatten` is JSON-API breaking change, and it's weird that this isn't caught by the tests.

matklad:
I guess, as you are the primary consumer of that RPC, if you don't worry about potential breakage here, it's also fine to land as is!

matklad (approved):

miraclx:
Removing `flatten` doesn't change anything when deserializing. It's with serialization that we have this issue.

When `jsonrpc-primitives` types were built, serialization wasn't top priority. The checks were, instead, for deserialization.

It's the `jsonrpc-client`, that reuses those types for serialization, that would catch little issues like these.

PR #6533, opened 2022-04-06
refactor: remove needless closure

matklad:
We needed this for historical reasons, but now we can get without it just fine.


matklad:
cc @austinabell I assume there's no secret plan why we need this still?

austinabell (approved):
nope! Was just kept to be more flexible if other types of checks were wanted. Good catch, forgot to update

mina86 (approved):

PR #6536, opened 2022-04-06
fix!: adjust gas distribution formula for NEP264

austinabell:
Adjusting formula based on @matklad suggestion. It more closely represents distributing gas based on weights if the sum of weights is greater than the amount of gas remaining at the cost of performing slightly more arithmetic.

I think this makes more sense, but hard to think about the exact implications without some benchmark of the two.


matklad:
Let's merge this: I doubt an extra integral division would be visible among all the other stuff we do to create a promise.

*But* it's good to check our assumtion that we can actually re-use the same cost for this new host function, I filed #6537 for that!

matklad (approved):

PR #6538, opened 2022-04-06
fix a github CI nightly test fail on master

mzhangmzz:
The broken test was tests::simple_chain::build_chain


mina86 (approved):
Could you point out in commit message which test is being fixed?

mzhangmzz:
> Could you point out in commit message which test is being fixed?

I'll update the message. It's not a nayduck test but a github CI test, not sure why it was broken in the first place

PR #6541, opened 2022-04-06
pytest: better support ANSI codes in LogTracker

mina86:
Strip ANSI codes from the log files read via LogTracker so that they don’t
affect whether a string is found in the logs.

This fixes sanity/state_sync_routed.py test.

### DIFF
--- a/pytest/lib/utils.py
+++ b/pytest/lib/utils.py
@@ -5,6 +5,7 @@ import json
 import os
 import pathlib
 import random
+import re
 import shutil
 import subprocess
 import sys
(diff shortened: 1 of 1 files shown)


nagisa (approved):
LGTM.

That said, I wonder if it wouldn't be more straightforward if we added ability to make the log output [not use colors](https://docs.rs/tracing-subscriber/latest/tracing_subscriber/fmt/struct.SubscriberBuilder.html#method.with_ansi) in the first place, or perhaps even [output JSON](https://docs.rs/tracing-subscriber/latest/tracing_subscriber/fmt/struct.SubscriberBuilder.html#method.json) so that workarounds on the consumer side aren't necessary? For specifying colour use, I could imagine an [environment variable](https://no-color.org) too.

I recall hearing us wanting some more machine-parsable format for other reasons in the past, too.

mina86:
I’ve just filed https://github.com/near/nearcore/issues/6548 about that as a matter of fact.

Issue #6409, opened 2022-03-09
`near_block_head_height` is not updated if no blocks are received


mina86:
The `near_block_head_height` metric (a.k.a. `metrics::BLOCK_HEIGHT_HEAD`) is set only when head is updated (in `ChainUpdate::update_head`). This means that if a node is stuck after startup and does not make any progress syncing up the metric is never updated and never exported. This in turn means that even if the node is running it does not show up on some of our graphs. The metric should be initialised at some point at startup to the head value read from the storage so that it is exported. The same probably applies to `HEADR_HEAD_HEIGHT`.

PR #6543, opened 2022-04-06
Add gc tail metrics

mzhangmzz:
https://grafana.near.org/goto/s03bDZU7k
Screen Shot 2022-04-06 at 3 53 58 PM


chain/chain/src/chain.rs
```diff
@@ -810,6 +809,10 @@ impl Chain {
         let prev_epoch_id = self.get_block_header(&head.prev_block_hash)?.epoch_id();
         let epoch_change = prev_epoch_id != &head.epoch_id;
         let mut fork_tail = self.store.fork_tail()?;
+        metrics::TAIL_HEIGHT.set(tail as i64);
+        metrics::FORK_TAIL_HEIGHT.set(fork_tail as i64);
+        metrics::CHUNK_TAIL_HEIGHT.set(self.store.chunk_tail()? as i64);
+        metrics::GC_STOP_HEIGHT.set(gc_stop_height as i64);
```
mina86, line 815: Should this be called at the end of the method since the numbers will be updated?

chain/chain/src/chain.rs
```diff
@@ -810,6 +809,10 @@ impl Chain {
         let prev_epoch_id = self.get_block_header(&head.prev_block_hash)?.epoch_id();
         let epoch_change = prev_epoch_id != &head.epoch_id;
         let mut fork_tail = self.store.fork_tail()?;
+        metrics::TAIL_HEIGHT.set(tail as i64);
+        metrics::FORK_TAIL_HEIGHT.set(fork_tail as i64);
+        metrics::CHUNK_TAIL_HEIGHT.set(self.store.chunk_tail()? as i64);
```
mina86, line 814: chunk tail is also updated by `clear_archive_data` below.

chain/chain/src/metrics.rs
```diff
@@ -42,6 +42,14 @@ pub static NUM_ORPHANS: Lazy<IntGauge> =
...
     try_create_int_gauge("near_header_head_height", "Height of the header head").unwrap()
 });
+pub static TAIL_HEIGHT: Lazy<IntGauge> =
+    Lazy::new(|| try_create_int_gauge("near_tail_height", "Height of tail").unwrap());
+pub static CHUNK_TAIL_HEIGHT: Lazy<IntGauge> =
+    Lazy::new(|| try_create_int_gauge("near_chunk_tail_height", "Height of chunk tail").unwrap());
+pub static FORK_TAIL_HEIGHT: Lazy<IntGauge> =
+    Lazy::new(|| try_create_int_gauge("near_fork_tail_height", "Height of fork tail").unwrap());
+pub static GC_STOP_HEIGHT: Lazy<IntGauge> =
+    Lazy::new(|| try_create_int_gauge("near_gc_stop_height", "Target height of gc").unwrap());
```
mina86, line 52: Could you also add code which is executed when node starts and which updates those metrics. Without that, counters won’t be exported until gc happens. We already have that issue with HEADER_HEAD_HEIGHT, see https://github.com/near/nearcore/issues/6409.

chain/chain/src/chain.rs
mzhangmzz: I don't have a strong preference on this. I think either way is fine, since the metric is only pull once every 10 seconds anyways.

chain/chain/src/metrics.rs
mzhangmzz: I set TAIL_HEIGHT, CHUNK_TAIL_HEIGHT, FORK_TAIL_HEIGHT but not GC_STOP_HEIGHT when node starts, because I think the last one is related to the gc process and should only be set when gc is initiated.

chain/chain/src/chain.rs
```diff
@@ -908,6 +914,7 @@ impl Chain {
 
         let mut chain_store_update = self.store.store_update();
         chain_store_update.clear_redundant_chunk_data(gc_stop_height, gc_height_limit)?;
+        metrics::CHUNK_TAIL_HEIGHT.set(self.store.chunk_tail()? as i64);
```
mina86, line 917: Probably also set GC_STOP_HEIGHT since we’re using that here as well.

mina86 (approved):

PR #6544, opened 2022-04-07
bugfix: fix m1 sandbox

Ekleog:
We have not implemented build caching for wasmtime yet. As such, we also
need to not try using it, or the build will fail.

Partial fix for https://github.com/near/workspaces-rs/issues/110

@matklad I'm not sure how good I feel about exposing VMKind while it was previously marked as internal. It was already being used by at least near-vm-runner-standalone and runtime-param-estimator, but still. The issue is, if I'm not doing that it means that somewhere else down the call chain we would have to ignore a correctly-configured contract cache, which feels worse to me than exposing VMKind.

What do you think about that? Or is internal maybe for "anything in nearcore but semver doesn't apply"?

cc @ChaoticTempest


matklad:
I think it's better to ignore the cache inside wasmtime implementation -- the most convincing argument for that is that that'll allow us to keep all "hacks" inside near-vm-runner abstraction boundary.

matklad:
also, cc @nagisa as the architect of hiding VMKind :)

bowenwang1996:
@Ekleog what is the status of this PR? Should we close it?

Ekleog:
@bowenwang1996 No, I still need to come back to it to implement it the better way, I've just put it on the backburner for now to focus on one thing at a time

ChaoticTempest:
@Ekleog hey, any updates on this?

Ekleog:
@ChaoticTempest Sorry for the delay, here it comes! I tried the procedure to reproduce you listed in the linked issue and it seems to work for me, though I was stuck for a bit on the need to white down a full path (or it tries running with an amd64 sandbox instead of the aarch64 one).

@matklad when you get a tick to re-review, I'm not 100% sure of the semantics of Ok(CacheNotFound) but it seems to both typecheck and work so… :sweat_smile:

matklad:
Hm, I've just realized that it's suboptimal that this didn't get into the release. @ChaoticTempest how do sandbox release work? Are they synchronized with overall nearcore release, or is it separate infra? If this PR is important for sandbox users, we might want to do an extra sandbox release, but it's on the dev-platform team to see to it!

ChaoticTempest:
@matklad Currently sandbox releases just take roughly the top of master w/ the latest features. This is not a tagged version binary either since there were until now no tagged macos binaries that were being published. 

>  If this PR is important for sandbox users, we might want to do an extra sandbox release

Definitely should require a new version bump for this so M1 users can make use of it! But this also hinges on https://nearinc.atlassian.net/browse/DPI-148 to make M1 binaries available

PR #6547, opened 2022-04-07
doc: Some short descriptions

Buckram123:
Some descriptions asked by https://github.com/near/near-sdk-rs/issues/763

### DIFF
--- a/core/primitives-core/src/config.rs
+++ b/core/primitives-core/src/config.rs
@@ -5,7 +5,7 @@ use std::collections::hash_map::DefaultHasher;
 use std::hash::{Hash, Hasher};
 use strum::{Display, EnumCount};
 
-/// Describes сonfiguration of wasm operations.
+/// Describes cost of different wasm operations.
 #[derive(Clone, Debug, Hash, Serialize, Deserialize, PartialEq, Eq)]
 pub struct VMConfig {
     /// Costs for runtime externals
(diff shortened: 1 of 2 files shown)


core/primitives-core/src/config.rs
```diff
@@ -5,6 +5,7 @@ use std::collections::hash_map::DefaultHasher;
 use std::hash::{Hash, Hasher};
 use strum::{Display, EnumCount};
 
+/// Describes сonfiguration of wasm operations.
```
bowenwang1996, line 8: Configuration might be misleading. What is described here is cost of different operations

core/primitives-core/src/config.rs
```diff
@@ -5,6 +5,7 @@ use std::collections::hash_map::DefaultHasher;
 use std::hash::{Hash, Hasher};
 use strum::{Display, EnumCount};
 
+/// Describes cost of different wasm operations.
```
nagisa, line 8: I think we should make sure that we document not only the costs are a part of `VMConfig` but also limits. But also would be great if the description allowed for extending the structure without needing to change the description so that it doesn't go out of date.

So perhaps something like:

```suggestion
Configuration of the VM properties such as costs and limits.
```

nagisa (approved):
This overall LGTM I think it is better than status quo either way.

Thanks for the contribution and extremely sorry for taking so long to take a look at this!

akhi3030:
@Buckram123: sorry for dropping the ball on merging this PR in time after it was approved.  It seems like the changes to config.rs are no longer needed as there is some documentation that exists there now.  The changes to fees.rs are still relevant.  Do you mind updating your PR please?

akhi3030:
Looks like this is abandoned now.

PR #6549, opened 2022-04-07
perf: avoid locks in epoch manager

matklad:
This replaces `lru::LruCache` with `near_cache::SyncLruCache` in the implementation of the EpochManager, which in turn allows to make all the getters to use `&self`, which in turns allows the callers of EpochManager to actually take advantage of RwLock.

### DIFF
--- a/chain/chain/src/types.rs
+++ b/chain/chain/src/types.rs
@@ -1,5 +1,6 @@
 use std::cmp::Ordering;
 use std::collections::HashMap;
+use std::sync::Arc;
 
 use borsh::{BorshDeserialize, BorshSerialize};
 use chrono::DateTime;
@@ -545,7 +546,17 @@ pub trait RuntimeAdapter: Send + Sync {
         prev_epoch_last_block_hash: &CryptoHash,
         epoch_id: &EpochId,
         next_epoch_id: &EpochId,
-    ) -> Result<(BlockInfo, BlockInfo, BlockInfo, EpochInfo, EpochInfo, EpochInfo), Error>;
+    ) -> Result<
+        (
+            Arc<BlockInfo>,
+            Arc<BlockInfo>,
+            Arc<BlockInfo>,
+            Arc<EpochInfo>,
+            Arc<EpochInfo>,
+            Arc<EpochInfo>,
+        ),
+        Error,
+    >;
 
     // TODO #3488 this likely to be updated
     /// Hash that is necessary for prove Epochs in Epoch Sync.
--- a/chain/epoch_manager/src/lib.rs
+++ b/chain/epoch_manager/src/lib.rs
@@ -1580,28 +1580,25 @@ impl EpochManager {
 
 #[cfg(test)]
 mod tests2 {
-    use num_rational::Rational;
-
-    use near_primitives::challenge::SlashedValidator;
-    use near_primitives::hash::hash;
-    use near_primitives::types::ValidatorKickoutReason::NotEnoughBlocks;
-    use near_primitives::version::PROTOCOL_VERSION;
-    use near_store::test_utils::create_test_store;
-
+    use super::*;
+    use crate::reward_calculator::NUM_NS_IN_SECOND;
     use crate::test_utils::{
         block_info, change_stake, default_reward_calculator, epoch_config,
         epoch_info_with_num_seats, hash_range, record_block, record_block_with_final_block_hash,
         record_block_with_slashes, record_with_block_info, reward, setup_default_epoch_manager,
         setup_epoch_manager, stake, DEFAULT_TOTAL_SUPPLY,
     };
-
-    use super::*;
-    use crate::reward_calculator::NUM_NS_IN_SECOND;
+    use near_primitives::challenge::SlashedValidator;
     use near_primitives::epoch_manager::EpochConfig;
     use near_primitives::epoch_manager::ShardConfig;
+    use near_primitives::hash::hash;
     use near_primitives::shard_layout::ShardLayout;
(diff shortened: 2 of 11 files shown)


chain/epoch_manager/src/lib.rs
```diff
@@ -1496,22 +1491,20 @@ impl EpochManager {
...
                 // or we are at the start of an epoch. In this case, the new aggregator should
                 // overwrite the old one.
                 overwrite = true;
                 break;
             }
-            new_aggregator.update(block_info, &epoch_info, prev_height?);
-            cur_hash = *block_info.prev_hash();
+            let prev_height = *self.get_block_info(prev_hash)?.height();
+            new_aggregator.update(&cur_block_info, &epoch_info, prev_height);
+            cur_hash = *prev_hash;
```
mina86, line 1507: This loop can be further optimised so that this second get_block_info is not necessary. What’s previous block in one iteration becomes current block in the next iteration so really this loop should be something along the lines of:

```rust
let mut current_block = get_block_info(last_block_hash);
loop {
  ...
  let prev_block = get_block_info(current_block.prev_hash());
  ...
  current_block = prev_block;
}
```

It’s best if we wait for #6531 before going with this PR I think.  It
does a bit of refactoring around this method.

matklad: Yeah, I've noticed that, and decided to leave it for the follow up, for fear of actually messing up the logic.

tools/state-viewer/src/epoch_info.rs
```diff
@@ -87,13 +85,10 @@ fn get_block_height_range(
             cur_block_info,
             epoch_info.epoch_height()
         );
-        let prev_epoch_last_block_hash = epoch_manager
-            .get_block_info(cur_block_info.epoch_first_block())
-            .unwrap()
-            .prev_hash()
-            .clone();
+        let prev_epoch_last_block_hash =
+            *epoch_manager.get_block_info(cur_block_info.epoch_first_block()).unwrap().prev_hash();
```
mina86, line 89: We can avoid the CryptoHash cloning here:

```rust
let epoch_first_block = epoch_manager.get_block_info(cur_block_info.epoch_first_block()).unwrap();
let prev_epoch_last_block_info = epoch_manager.get_block_info(prev_epoch_first_block.prev_hash()).unwrap();
```

mina86 (approved):

chain/epoch_manager/src/lib.rs
```diff
@@ -497,13 +503,13 @@ impl EpochManager {
             Ok(next_next_epoch_info) => next_next_epoch_info,
             Err(EpochError::ThresholdError { stake_sum, num_seats }) => {
                 warn!(target: "epoch_manager", "Not enough stake for required number of seats (all validators tried to unstake?): amount = {} for {}", stake_sum, num_seats);
-                let mut epoch_info = next_epoch_info.clone();
+                let mut epoch_info = (*next_epoch_info).clone();
```
nagisa, line 506: `next_epoch_info` is already an `Arc` AFAICT, so `Arc::clone` here, and pull `Arc::new` into the relevant branches.

(And more generally, it is pretty hard to tell which `Clone` impl is being invoked here, and it sort of matters, so prudent to use `T::clone(...)` to invoke, over a method call)

chain/epoch_manager/src/lib.rs
```diff
@@ -1343,26 +1331,22 @@ impl EpochManager {
...
         self.store
-            .get_ser(ColEpochValidatorInfo, epoch_id.as_ref())
-            .map_err(|err| err.into())
-            .and_then(|value| value.ok_or_else(|| EpochError::EpochOutOfBounds(epoch_id.clone())))
+            .get_ser(ColEpochValidatorInfo, epoch_id.as_ref())?
+            .ok_or_else(|| EpochError::EpochOutOfBounds(epoch_id.clone()))
     }
 
+    // NOTE: beware, after calling `save_epoch_validator_info`,
+    // `get_epoch_validator_info` will return stale results.
```
mm-near, line 1349: nice catch -- we should add a cache to it to avoid this pitfall.

Can you create a github issue and mention it in this comment?

mm-near (approved):
great!

chain/chain/src/types.rs
```diff
@@ -545,7 +546,17 @@ pub trait RuntimeAdapter: Send + Sync {
...
         next_epoch_id: &EpochId,
-    ) -> Result<(BlockInfo, BlockInfo, BlockInfo, EpochInfo, EpochInfo, EpochInfo), Error>;
+    ) -> Result<
+        (
+            Arc<BlockInfo>,
+            Arc<BlockInfo>,
+            Arc<BlockInfo>,
+            Arc<EpochInfo>,
+            Arc<EpochInfo>,
+            Arc<EpochInfo>,
```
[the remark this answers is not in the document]
matklad: Yeah... I decided to not touch this for the time being -- I want to actually understand what the code is doing, maybe we can just refactor this away instead of just refactoring...

chain/epoch_manager/src/lib.rs
matklad: >AFAICT, so Arc::clone here, and pull Arc::new into the relevant branches.

Hm, I think this doesn't work (or I am misunderstanding the suggestion). Here, we want and independent copy to mutate it. Ie, even `Arc::make_mut` doesn't seem to be what we need here. 

> so prudent to use T::clone(...) to invoke, over a method call

Thanks, I'll add this to my idiom store!

PR #6552, opened 2022-04-07
perf: force processing of each task as a separate job.

matklad:
IMPORTANT: light local testing implies that this is not necessary --
tasks are always maximally split. However, I've heard from marcin that
we observe different effect in some real-world experiments, so this
commit is more of an experiment to check if this helps.

To re-iterate, I am 0.7 sure that we shouldn't actually merge this in
master.

### DIFF
--- a/chain/chain/src/chain.rs
+++ b/chain/chain/src/chain.rs
@@ -72,7 +72,7 @@ use delay_detector::DelayDetector;
 use near_primitives::shard_layout::{
     account_id_to_shard_id, account_id_to_shard_uid, ShardLayout, ShardUId,
 };
-use rayon::iter::{IntoParallelIterator, ParallelIterator};
+use rayon::iter::{IndexedParallelIterator, IntoParallelIterator, ParallelIterator};
 
 /// Maximum number of orphans chain can store.
 pub const MAX_ORPHAN_SIZE: usize = 1024;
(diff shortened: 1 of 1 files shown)


matklad:
Started discussion in rayon's chat to learn how this actually works: https://gitter.im/rayon-rs/Lobby?at=624f258d9a09ab24f3e05589

matklad:
For the refernce, the test i used to check rayon's scheduling behavior:

```rust
use std::time::{Duration, Instant};

use rayon::prelude::*;

fn main() {
    let t = Instant::now();
    let it = make_work().into_par_iter();
    it.for_each(|f| f());
    eprintln!("{:0.2?}", t.elapsed());
}

#[inline(never)]
fn make_work() -> Vec<fn()> {
    let slow = || {
        eprintln!(
            "std::thread::current().id() = {:?}",
            std::thread::current().id()
        );
        let start = std::time::Instant::now();
        let mut x = 92u64;
        while start.elapsed() < Duration::from_secs(5) {
            for _ in 0..100_000 {
                x = (x.wrapping_mul(124) ^ x) / 3;
            }
        }
    };
    let fast = || {
        eprintln!(
            "std::thread::current().id() = {:?}",
            std::thread::current().id()
        );
    };
    vec![fast, fast]
}
```

PR #6555, opened 2022-04-07
refeactor: split get_and_update_epoch_info_aggregator into get and up…

matklad:
…date.

Standard refactor -- split a function with a boolean param into two
differently named functions. This doesn't bring substantail benefit
right now, but in the future refactor, the `get` version would take
`&self`, while the `take` version would use `&mut self`.

Realized that this is needed for #6549


mina86:
This is obsolete by https://github.com/near/nearcore/pull/6531

matklad:
Indeed, that one is a much for thorough approach, :+1:

PR #6556, opened 2022-04-08
fix: Add v53 to config store

jakmeier:
When updating #6397 to protocol v53 right before merging,
I forgot to add the configuration JSON file to the config store.

Test plan
---------
Added a test to verify that cost indeed increases.

### DIFF
--- a/integration-tests/src/tests/client/process_blocks.rs
+++ b/integration-tests/src/tests/client/process_blocks.rs
@@ -3459,6 +3459,96 @@ fn test_catchup_no_sharding_change() {
     }
 }
 
+/// Tests if the cost of deployment is higher after the protocol update 53
+#[test]
+fn test_deploy_cost_increased() {
+    let old_protocol_version = 52;
+    let new_protocol_version = 53;
+
+    let contract_size = 1024 * 1024;
+    let test_contract = near_test_contracts::sized_contract(contract_size);
+
+    // Prepare TestEnv with a contract at the old protocol version.
+    let epoch_length = 5;
+    let mut env = {
+        let mut genesis = Genesis::test(vec!["test0".parse().unwrap()], 1);
+        genesis.config.epoch_length = epoch_length;
+        genesis.config.protocol_version = old_protocol_version;
+        let chain_genesis = ChainGenesis::from(&genesis);
+        let runtimes: Vec<Arc<dyn RuntimeAdapter>> =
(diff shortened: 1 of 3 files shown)


integration-tests/src/tests/client/process_blocks.rs
```diff
@@ -3459,6 +3459,96 @@ fn test_catchup_no_sharding_change() {
...
+    let old_outcome = {
+        let tip = env.clients[0].chain.head().unwrap();
+        let signed_transaction =
+            Transaction { nonce: 10, block_hash: tip.last_block_hash, ..tx.clone() }.sign(&signer);
+        let tx_hash = signed_transaction.get_hash();
+        env.clients[0].process_tx(signed_transaction, false, false);
+        for i in 0..epoch_length {
+            env.produce_block(0, tip.height + i + 1);
+        }
+        env.clients[0].chain.get_final_transaction_result(&tx_hash).unwrap()
```
mina86, line 3509: Perhaps wrap it in a closure so we can reuse it rather than duplicating the same code? This will make it obvious that we deploy the exact same contract.

mina86 (approved):

integration-tests/src/tests/client/process_blocks.rs
```diff
@@ -3459,6 +3459,96 @@ fn test_catchup_no_sharding_change() {
     }
 }
 
+/// Tests if the cost of deployment is higher after the protocol update 53
+#[test]
+fn test_deploy_cost_increased() {
+    let old_protocol_version = 52;
+    let new_protocol_version = 53;
```
Longarithm, line 3466: Alternatively I would introduce a protocol feature like `IncreaseDeploymentCost` and use its protocol version (- 1).

I think of a scenario when we can postpone this feature to the v54 due to emergency reasons. Then it will be natural to modify protocol version of feature in the `version.rs`, but it can be easy to forget to fix two numbers in some test.

Longarithm (changes requested):

integration-tests/src/tests/client/process_blocks.rs
jakmeier: Yeah I like that. Done.

integration-tests/src/tests/client/process_blocks.rs
jakmeier: You are right, I have now added this as an explicit protocol feature.

Longarithm (approved):

Longarithm:
Thank you!

PR #6557, opened 2022-04-08
neard: enable target:"recompress" info logs by default

mina86:
The target is used by recompress_storage.  Initially the command used
no target but then target was added which caused the log lines to be
ignored by default.


nagisa (approved):
I wonder if we should be prefixing our targets with `near::` so that the targets were more structured…

LGTM for now though.

PR #6558, opened 2022-04-08
chore: Update estimator docker image to Rust 1.60

jakmeier:


matklad:
@jakmeier approved, but maybe let's merge this after CE data collection is enabled (which it might already be!) just to make sure we observe regressions/wins

matklad (approved):

jakmeier:
Data collecting confirmed to work, so let's finally merge this

PR #6560, opened 2022-04-08
refactor: enforce the invariant that weights are non-zero at compile time

matklad:
(builds on top of #6559, ignore the first commit)

### DIFF
--- a/core/primitives-core/src/types.rs
+++ b/core/primitives-core/src/types.rs
@@ -1,4 +1,5 @@
 use crate::hash::CryptoHash;
+use std::num::NonZeroU64;
 
 /// Account identifier. Provides access to user's state.
 pub use crate::account::id::AccountId;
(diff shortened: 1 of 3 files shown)


matklad:
@austinabell do we actually want it? I am myself somewhat undecided -- I don't think it brings any super-noticeable win, but at the same time I think a little bit of pedantic typing won't harm in this relatively sensitive part of the code base.

core/primitives-core/src/types.rs
```diff
@@ -28,8 +29,19 @@ pub type Gas = u64;
...
+
+#[cfg(feature = "protocol_feature_function_call_weight")]
+impl GasWeight {
+    pub fn new(weight: u64) -> Option<GasWeight> {
+        NonZeroU64::new(weight).map(GasWeight)
+    }
+
+    pub fn as_u128(self) -> u128 {
+        self.0.get().into()
+    }
```
austinabell, line 43: If we do this, might be worth making these `const`

core/primitives-core/src/types.rs
```diff
@@ -28,8 +29,19 @@ pub type Gas = u64;
...
+#[derive(Clone, Copy, Debug, PartialEq)]
+pub struct GasWeight(NonZeroU64);
+
+#[cfg(feature = "protocol_feature_function_call_weight")]
+impl GasWeight {
+    pub fn new(weight: u64) -> Option<GasWeight> {
+        NonZeroU64::new(weight).map(GasWeight)
+    }
+
+    pub fn as_u128(self) -> u128 {
```
austinabell, line 41: prob better to name `to_u128`

austinabell (commented):
honestly don't have an opinion. No concerns here since we won't be using this type natively from the SDK anyway

matklad:
Let's do this: close this PR for now, to not interfere with stabilization efforts, and maybe revisit after that.

Issue #6443, opened 2022-03-16
Replace `debug_assert` with logging


matklad:
In #6439, I've noticed a couple of `debug_assert`, which I think were used in a defensive fashion, to not bring down the node in production. `debug_assert` is not the best tool here, as it is completely elliminated in release. What we want here is to panic in debug and to log in release:

<pre> if !cond {
 if cfg(debug_assertions) 

 { panic!() } 

 else 

 { tracing::log!() } 

}</pre>

In rust-analyzer, we use [https://docs.rs/always-assert/latest/always_assert/](https://docs.rs/always-assert/latest/always_assert/) for this purpose.

[https://www.sqlite.org/assert.html](https://www.sqlite.org/assert.html) is a relevant read.

mina86:
Question can also be asked whether we should just `assert`.  To me, `debug_assert` makes sense in hot paths or generally when checking the condition is too expensive for production code.

kakoc:
So, am I right that our intention is to basically replace all `debug_assert!` in the code with appropriate proposed variants with logging so that we will have as much debug info in logs(when invariants are broken) as we can even in production?

mina86:
I’m not sure it’s as simple as replacing all debug_asserts with
logging on non-debug builds.  For example:

```
chain/chain/src/chain.rs:300:                    debug_assert!(_visited.insert(*hash));
```

should stay as debug_assert I think because the cost of the check
involves creating a hash set.  Many places probably should just be
converted to plain asserts.  For example:

```
core/primitives/src/trie_key.rs:187:        debug_assert_eq!(res.len(), expected_len);
core/primitives/src/trie_key.rs:265:        debug_assert_eq!(ACCOUNT_DATA_SEPARATOR.len(), 1);
core/primitives/src/trie_key.rs:293:        debug_assert_eq!(col::ACCESS_KEY.len(), 1);
core/primitives/src/trie_key.rs:351:        debug_assert_eq!(ACCOUNT_DATA_SEPARATOR.len(), 1);
```

especially since here the last three will be compiled out (though
honestly in this case it’d be better to convert ACCOUNT_DATA_SEPARATOR
et al to be `u8` rather than `&[u8]`).  Or in:

```rust
    fn store_at(&mut self, handle: StorageHandle, node: TrieNodeWithSize) {
        debug_assert!(self.nodes.get(handle.0).expect(INVALID_STORAGE_HANDLE).is_none());
        self.nodes[handle.0] = Some(node);
    }
```

[...]

PR #6561, opened 2022-04-08
Add asserts for fn that should not be used with rc columns

mm-near:


EdvardD (approved):

matklad:
cc https://github.com/near/nearcore/issues/6443

I think debug asserts are suboptimal here -- if this happens in prod, would be great to see an error in the log.

matklad:
To be clear: this PR is definitely an improvement, we should land it, lest perfect be the enemy of the good. The above is FYI

matklad (approved):

core/store/src/lib.rs
```diff
@@ -184,10 +184,15 @@ impl StoreUpdate {
         self.transaction.update_refcount(column, key, value)
     }
 
+    // Saves a given value.
+    // Must not be used for RC columns - please use 'update refcount' instead.
     pub fn set(&mut self, column: DBCol, key: &[u8], value: &[u8]) {
+        debug_assert!(!column.is_rc());
```
mina86, line 190: So this would actually break what I know think might be correct usage in recompress_storage. It now reads data using iter_without_rc_logic and writes it to a new empty database using set. As I understand things now, it’s a bit hacky but correct way to do it.

mm-near: yeah, that's quite hacky.. 

how about we add a new iterator that returns (key, value, refcount) - and then we could use 'update_refcount' call rather than set?

matklad:
I think this was superceeded by https://github.com/near/nearcore/pull/6674 perhaps?

mm-near:
mostly yes -- but I see that we didn't cover the 'delete' case there. I'll update the PR.

mm-near:
@matklad @mina86 - updated the PR to assert on delete and set_ser

matklad (approved):
LGTM!

In the future, we *might* have a need for `delete_raw` (or making `set_raw_byte`'s value an `Option`), but afair we don't use `delete` like that today yet

mina86 (approved):

PR #6563, opened 2022-04-08
chain: allow chain_tail to be set w/o tail being set on archival nodes

mina86:
Since commit 6be2e0e8e: ‘gc partial chunks on archival nodes (#6439)’,
archival nodes set chunk_tail without setting tail.  However, store
validation expects both of those to be set or unset.  Change the code
to allow unset tail on archival nodes.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,9 +4,9 @@ version = 3
 
 [[package]]
 name = "actix"
-version = "0.11.0-beta.2"
+version = "0.12.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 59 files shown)


matklad:
reassigning to mm, as mzhangmzz is OOO, and this is P-high

chain/client/src/view_client.rs
```diff
@@ -87,6 +87,14 @@ pub struct AdversarialControls {
     pub adv_disable_header_sync: bool,
     pub adv_disable_doomslug: bool,
     pub adv_sync_height: Option<u64>,
+    pub is_archival: bool,
```
mm-near, line 90: are you using this anywhere? (I couldn't find it during the review)

mm-near (approved):

PR #6566, opened 2022-04-08
possible fix for the 'SyncRoutingTable' overload

mm-near:
Don't rebroadcast the AnnounceAccounts that you already know of.

Without this PR, if we had a 'batch' of AnnounceAccounts, we could have broadcasted them multiple times.


chain/network/src/peer_manager/peer_manager_actor.rs
```diff
@@ -352,6 +352,25 @@ impl PeerManagerActor {
...
+                    if announce_account.epoch_id == current_announce_account.epoch_id {
+                        None
+                    } else {
+                        Some(announce_account)
+                    }
+                } else {
+                    Some(announce_account)
+                }
+            })
+            .collect();
```
matklad, line 372: Better to use https://doc.rust-lang.org/stable/std/vec/struct.Vec.html#method.retain here. `into_iter().collect()` is a useless reallocation. Retain should also be somewhat shorter.

mm-near: I've tried - but as retain is 'mutable' - I'd need to get a mutable reference in the arguments - and then it is causing a lot larger change. (unless i'm missing something)

matklad: The following diff works for me:

```diff
diff --git a/chain/network/src/peer_manager/peer_manager_actor.rs b/chain/network/src/peer_manager/peer_manager_actor.rs
index 50764e72e..283b32c44 100644
--- a/chain/network/src/peer_manager/peer_manager_actor.rs
+++ b/chain/network/src/peer_manager/peer_manager_actor.rs
@@ -351,25 +351,15 @@ impl PeerManagerActor {
         }
     }
 
-    fn broadcast_accounts(&mut self, accounts: Vec<AnnounceAccount>) {
+    fn broadcast_accounts(&mut self, mut accounts: Vec<AnnounceAccount>) {
         // Filter the accounts again, so that we're sending only the ones that were not added.
         // without it - if we have multiple 'broadcast_accounts' calls queued up, we'll end up sending a lot of repeated messages.
-        let accounts: Vec<(AnnounceAccount)> = accounts
-            .into_iter()
-            .filter_map(|announce_account| {
-                if let Some(current_announce_account) =
-                    self.routing_table_view.get_announce(&announce_account.account_id)
-                {
-                    if announce_account.epoch_id == current_announce_account.epoch_id {
-                        None
-                    } else {
-                        Some(announce_account)
-                    }
-                } else {
-                    Some(announce_account)
-                }
-            })
-            .collect();
+        accounts.retain(|announce_account| {
+            match self.routing_table_view.get_announce(&announce_account.account_id) {
+                Some(current_announce_account) if announce_account.epoch_id == current_announce_account.epoch_id => false,
+                _ => true,
+            }
+        });
 
         if accounts.is_empty() {
             return;
```

matklad (approved):
:+1: as far as the Rut code goes, can't vouch for correctness though.

Although, I have a question -- can we test this behavior?

mm-near:
yes - added test in #6611 (that was a little bit more complex than I've expected)

nikurt (commented):
LGTM

PR #6567, opened 2022-04-09
add info print for GC and block production at epoch starts

mzhangmzz:


chain/client/src/client.rs
```diff
@@ -61,6 +61,9 @@ pub const EPOCH_SYNC_REQUEST_TIMEOUT: Duration = Duration::from_millis(1_000);
 // TODO #3488 set 60_000
 pub const EPOCH_SYNC_PEER_TIMEOUT: Duration = Duration::from_millis(10);
 
+/// number of blocks at the epoch start for which we will lock more detailed info
```
mm-near, line 64: s/lock/log

chain/chain/src/doomslug.rs
```diff
@@ -575,13 +577,24 @@ impl Doomslug {
                     DoomslugBlockProductionReadiness::NotReady => false,
                     DoomslugBlockProductionReadiness::ReadySince(when) => {
                         if has_enough_chunks {
+                            if log_block_production_info {
+                                info!("ready to produce block @ {}, have enough approvals for {:?}, has enough chunks", target_height, now.saturating_duration_since(when));
```
mm-near, line 581: s/have/has

mm-near (approved):

nikurt (commented):
Looks good

PR #6570, opened 2022-04-10
chain: remove failing debug asserts

mina86:
The asserts have been introduced in commit ed1a1b600bb4: ‘add metrics
for how long a block stay in the orphan pool and the missing chunks
pool’ and at least the one in mark_block_unorphaned have been failing
in some of the tests.  Remove the asserts to make the tests pass.


mina86:
Here’s an example of a failing test: https://nayduck.near.org/#/test/301361

mm-near (approved):

Issue #5962, opened 2021-12-23
protocol_feature_fix_contract_loading_cost tracking issue


matklad:
`protocol_feature_fix_contract_loading_cost` features changes the logic for charging for loading contract. Roughly, this is a per-byte cost which measures the work of "dynamic linker" (loading pre-compiled x86_64 machine code from the database into memory, doing neccesary adjustments and marking the relevant pages as executable). 

Specifically, the "fix" here covers: 

* charging the cost if contract execution was interrupted before the actual execution started (method not found and other such errors)
* charging the cost *before* fetching data from disk, or otherwise making sure that paying just a bit for a call to a big contract isn't exploitable.

Implementation history: 

* implementation https://github.com/near/nearcore/pull/6772
* first attempt at stabilization https://github.com/near/nearcore/pull/7177
* cost estimation https://github.com/near/nearcore/issues/7227
* stabilization reversal https://github.com/near/nearcore/pull/7241

Blockers: 

- [ ] https://github.com/near/nearcore/issues/7511

matklad:
related: #4826

matklad:
Note: we also should charge this cost if contract fails to compile (ie, it is invalid wasm)

jakmeier:
I started working on this. I wonder what the correct place to charge the new `contract_loading` is. I think ideally, it should be *before* we actually load the contract. Unfortunately we currently do not have the size of the contract in advance. So I see two practical options.

1. Simplest approach: Move it to the the start of each implementation of `runner::VM::run` where it currently is.
2. Lifting it out of the contract-runtime, charging it before even setting up `VMContext`.

jakmeier:
Having merged #6772 also, I consider this issue done.

jakmeier:
We also want to reshuffle costs at some point: #7741

Maybe we should release these things together. But we can continue to treat them separately, as the cost changes probably need some more optimizations before we can really move forward with them.

akashin:
Closing this as obsolete.

PR #6571, opened 2022-04-11
refactor: VM::run returns an enum instead of tuple

jakmeier:
This refactor primarily tries to make the `VM::run` signature clearer.
It is also a preparation step towards changing where contract loading
gas costs are charged.

### DIFF
--- a/runtime/near-vm-runner-standalone/src/main.rs
+++ b/runtime/near-vm-runner-standalone/src/main.rs
@@ -7,6 +7,7 @@ use clap::Parser;
 use near_vm_logic::ProtocolVersion;
 use near_vm_logic::VMOutcome;
 use near_vm_runner::internal::VMKind;
+use near_vm_runner::VMResult;
 use serde::{
     de::{MapAccess, Visitor},
     ser::SerializeMap,
@@ -105,7 +106,7 @@ struct CliArgs {
 #[allow(unused)]
 #[derive(Debug, Clone)]
 struct StandaloneOutput {
-    pub outcome: Option<VMOutcome>,
+    pub outcome: VMOutcome,
     pub err: Option<String>,
     pub state: State,
 }
@@ -163,28 +164,29 @@ fn main() {
     step.promise_results(promise_results);
 
     let mut results = script.run();
-    let (outcome, err) = results.outcomes.pop().unwrap();
+    let last_result = results.outcomes.pop().unwrap();
+    let maybe_error = match &last_result {
+        VMResult::NotRun(err) | VMResult::Aborted(_, err) => Some(err),
+        VMResult::Ok(_) => None,
+    };
 
-    println!(
-        "{:#?}",
-        StandaloneOutput {
-            outcome: outcome.clone(),
-            err: err.map(|it| it.to_string()),
-            state: State(results.state.fake_trie),
-        }
-    );
+    match &last_result {
+        VMResult::NotRun(_) => {}
+        VMResult::Aborted(outcome, _) | VMResult::Ok(outcome) => {
+            println!(
+                "{:#?}",
+                StandaloneOutput {
+                    outcome: outcome.clone(),
+                    err: maybe_error.map(|it| it.to_string()),
+                    state: State(results.state.fake_trie),
+                }
+            );
 
-    if let Some(outcome) = &outcome {
-        println!("\nLogs:");
-        for log in &outcome.logs {
-            println!("{}\n", log);
-        }
-    }
-
-    match &outcome {
-        Some(outcome) => {
+            println!("\nLogs:");
+            for log in &outcome.logs {
+                println!("{}\n", log);
+            }
             println!("{:#?}", outcome.profile);
         }
-        _ => {}
     }
 }
--- a/runtime/near-vm-runner-standalone/src/main.rs
+++ b/runtime/near-vm-runner-standalone/src/main.rs
@@ -165,10 +165,7 @@ fn main() {
 
     let mut results = script.run();
     let last_result = results.outcomes.pop().unwrap();
-    let maybe_error = match &last_result {
-        VMResult::NotRun(err) | VMResult::Aborted(_, err) => Some(err),
-        VMResult::Ok(_) => None,
-    };
+    let maybe_error = last_result.error();
 
     match &last_result {
         VMResult::NotRun(_) => {}
--- a/runtime/near-vm-runner/src/runner.rs
+++ b/runtime/near-vm-runner/src/runner.rs
@@ -125,8 +125,7 @@ impl VMResult {
     pub fn not_run(error: VMError) -> VMResult {
         VMResult::NotRun(error)
     }
-    #[cfg_attr(not(feature = "protocol_feature_function_call_weight"), allow(unused_mut))]
-    pub fn abort(mut logic: VMLogic, error: VMError) -> VMResult {
+    pub fn abort(logic: VMLogic, error: VMError) -> VMResult {
         let outcome = logic.compute_outcome_and_distribute_gas();
         VMResult::aborted(outcome, error)
     }
--- a/runtime/near-vm-runner/src/runner.rs
+++ b/runtime/near-vm-runner/src/runner.rs
@@ -122,19 +122,21 @@ pub enum VMResult {
 }
 
 impl VMResult {
-    pub fn not_run(error: VMError) -> VMResult {
-        VMResult::NotRun(error)
-    }
+    /// Consumes the `VMLogic` object and computes the final outcome with the
+    /// given error that stopped execution from finishing successfully.
     pub fn abort(logic: VMLogic, error: VMError) -> VMResult {
         let outcome = logic.compute_outcome_and_distribute_gas();
-        VMResult::aborted(outcome, error)
-    }
-    pub fn aborted(outcome: VMOutcome, error: VMError) -> VMResult {
         VMResult::Aborted(outcome, error)
     }
-    pub fn ok(outcome: VMOutcome) -> VMResult {
+
+    /// Consumes the `VMLogic` object and computes the final outcome for a
+    /// successful execution.
+    pub fn ok(logic: VMLogic) -> VMResult {
+        let outcome = logic.compute_outcome_and_distribute_gas();
         VMResult::Ok(outcome)
     }
+
+    /// Borrow the internal outcome, if there is one.
     pub fn outcome(&self) -> Option<&VMOutcome> {
         match self {
             VMResult::NotRun(_err) => None,
@@ -142,6 +144,8 @@ impl VMResult {
             VMResult::Ok(outcome) => Some(outcome),
         }
     }
+
+    /// Borrow the internal error, if there is one.
     pub fn error(&self) -> Option<&VMError> {
         match self {
             VMResult::NotRun(err) => Some(err),
@@ -149,6 +153,9 @@ impl VMResult {
             VMResult::Ok(_outcome) => None,
         }
     }
+
+    /// Unpack the internal outcome and error. This method mostly exists for
+    /// easy compatibility with code that was written before `VMResult` existed.
     pub fn outcome_error(self) -> (Option<VMOutcome>, Option<VMError>) {
         match self {
             VMResult::NotRun(err) => (None, Some(err)),
--- a/runtime/near-vm-runner/src/runner.rs
+++ b/runtime/near-vm-runner/src/runner.rs
@@ -114,6 +114,8 @@ impl VMKind {
 #[derive(Debug, PartialEq)]
 pub enum VMResult {
     /// Execution was not able to start because preconditions were not met.
+    /// TODO: Remove this with more refactoring or present a good reason why it
+    /// is needed.
     NotRun(VMError),
     /// Execution started but hit an error.
     Aborted(VMOutcome, VMError),
--- a/runtime/near-vm-runner/src/runner.rs
+++ b/runtime/near-vm-runner/src/runner.rs
@@ -6,7 +6,7 @@ use near_primitives::types::CompiledContractCache;
 use near_primitives::version::ProtocolVersion;
 use near_vm_errors::VMError;
 use near_vm_logic::types::PromiseResult;
-use near_vm_logic::{External, VMContext, VMOutcome};
+use near_vm_logic::{External, VMContext, VMLogic, VMOutcome};
 
 use crate::vm_kind::VMKind;
 
@@ -33,7 +33,7 @@ pub fn run(
     promise_results: &[PromiseResult],
     current_protocol_version: ProtocolVersion,
     cache: Option<&dyn CompiledContractCache>,
-) -> (Option<VMOutcome>, Option<VMError>) {
+) -> VMResult {
     let vm_kind = VMKind::for_protocol_version(current_protocol_version);
     if let Some(runtime) = vm_kind.runtime(wasm_config.clone()) {
         runtime.run(
@@ -72,7 +72,7 @@ pub trait VM {
         promise_results: &[PromiseResult],
         current_protocol_version: ProtocolVersion,
         cache: Option<&dyn CompiledContractCache>,
-    ) -> (Option<VMOutcome>, Option<VMError>);
+    ) -> VMResult;
 
     /// Precompile a WASM contract to a VM specific format and store the result into the `cache`.
     ///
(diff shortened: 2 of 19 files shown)


jakmeier:
Throughout last week, I spent some time wrapping my head around how and where to fix contract loading cost (see #5962). As discussed with @matklad , it seems that refactoring the output type of `VM::run()` might be necessary as a first step for that. Therefore, this PR is only a suggested refactor, without fixing or changing anything, yet.

After trying a few signatures, this is what I currently like best.

`(Option<VMOutcome>, Option<VMError>)` turns into
```
pub enum VMResult {
    NotRun(VMError),
    Aborted(VMOutcome, VMError),
    Ok(VMOutcome),
}
```

The mapping is as follows:
 - `(Some(_),None)` turns into `VMResult::Ok(_)`
 - `(Some(_),Some())` turns into `VMResult::Aborted(_,_)`
 - `(None,Some())` turns into `VMResult::NotRun(_)`
 - `(None,None)` does not exist

One alternative I considered was to just return `VMOutcome` and add `Option<VMError>` as a field. The main problem with that is how to represent `VMResult::NotRun(_)`. We have to manufacture an empty `VMOutcome`, which is possible but I didn't really like it. Also, this requires `Clone` for `VMError`, which in turn means we have to extend the `AnyError` trickery to also have `Clone`. (Today we only have `Eq`.) Or, we could try to get rid of the `AnyError`.

runtime/near-vm-runner/src/runner.rs
```diff
@@ -108,3 +108,53 @@ impl VMKind {
...
+    /// Execution started but hit an error.
+    Aborted(VMOutcome, VMError),
+    /// Execution finished without error.
+    Ok(VMOutcome),
+}
+
+impl VMResult {
+    pub fn not_run(error: VMError) -> VMResult {
+        VMResult::NotRun(error)
+    }
```
nagisa, line 127: Is this method necessary? I think it is better to have just one way to construct the variant, even if it isn't consistent with helpers for some of the other variants.

runtime/near-vm-runner/src/runner.rs
```diff
@@ -108,3 +108,53 @@ impl VMKind {
...
+        }
+    }
+    pub fn error(&self) -> Option<&VMError> {
+        match self {
+            VMResult::NotRun(err) => Some(err),
+            VMResult::Aborted(_outcome, err) => Some(err),
+            VMResult::Ok(_outcome) => None,
+        }
+    }
+    pub fn outcome_error(self) -> (Option<VMOutcome>, Option<VMError>) {
```
nagisa, line 153: Please add some space between- and doc-comments to the methods here.

runtime/near-vm-runner/src/wasmer2_runner.rs
```diff
@@ -570,19 +575,20 @@ impl crate::runner::VM for Wasmer2VM {
...
         }
         let import = imports::wasmer2::build(vmmemory, &mut logic, current_protocol_version);
         if let Err(e) = get_entrypoint_index(&*artifact.artifact, method_name) {
-            return (None, Some(e));
+            // TODO: This should return an outcome to account for loading cost
+            return VMResult::not_run(e);
+        }
+        match self.run_method(&artifact, import, method_name) {
+            Ok(()) => VMResult::ok(logic.compute_outcome_and_distribute_gas()),
+            Err(err) => VMResult::aborted(logic.compute_outcome_and_distribute_gas(), err),
```
nagisa, line 590: This could be calling `VMResult::abort(...)`.

runtime/near-vm-runner/src/wasmer_runner.rs
```diff
@@ -257,11 +256,13 @@ pub(crate) fn run_wasmer0_module<'a>(
...
     if let Err(e) = check_method(&module, method_name) {
-        return (None, Some(e));
+        return VMResult::not_run(e);
     }
 
-    let err = run_method(&module, &import_object, method_name).err();
-    (Some(logic.compute_outcome_and_distribute_gas()), err)
+    match run_method(&module, &import_object, method_name) {
+        Ok(()) => VMResult::ok(logic.compute_outcome_and_distribute_gas()),
+        Err(err) => VMResult::aborted(logic.compute_outcome_and_distribute_gas(), err),
```
nagisa, line 264: This could be calling `VMResult::abort`.

runtime/near-vm-runner/src/wasmer2_runner.rs
```diff
@@ -446,10 +447,14 @@ impl Wasmer2VM {
...
         let import = imports::wasmer2::build(vmmemory, &mut logic, current_protocol_version);
         if let Err(e) = get_entrypoint_index(&*artifact.artifact, method_name) {
-            return (None, Some(e));
+            return VMResult::not_run(e);
+        }
+        let status = self.run_method(artifact, import, method_name);
+        let outcome = logic.compute_outcome_and_distribute_gas();
+        match status {
+            Ok(()) => VMResult::ok(outcome),
+            Err(err) => VMResult::aborted(outcome, err),
```
nagisa, line 456: This could be calling `VMResult::abort`.

runtime/near-vm-runner/src/wasmer_runner.rs
```diff
@@ -345,23 +344,24 @@ impl crate::runner::VM for Wasmer0VM {
...
-            return (None, Some(e));
+            // TODO: This should return an outcome to account for loading cost
+            return VMResult::not_run(e);
         }
 
-        let err = run_method(&module, &import_object, method_name).err();
-        (Some(logic.compute_outcome_and_distribute_gas()), err)
+        match run_method(&module, &import_object, method_name) {
+            Ok(()) => VMResult::ok(logic.compute_outcome_and_distribute_gas()),
+            Err(err) => VMResult::aborted(logic.compute_outcome_and_distribute_gas(), err),
```
nagisa, line 363: `VMResult::abort`.

runtime/near-vm-runner/src/runner.rs
```diff
@@ -108,3 +108,53 @@ impl VMKind {
...
+    pub fn abort(mut logic: VMLogic, error: VMError) -> VMResult {
+        let outcome = logic.compute_outcome_and_distribute_gas();
+        VMResult::aborted(outcome, error)
+    }
+    pub fn aborted(outcome: VMOutcome, error: VMError) -> VMResult {
+        VMResult::Aborted(outcome, error)
+    }
+    pub fn ok(outcome: VMOutcome) -> VMResult {
+        VMResult::Ok(outcome)
+    }
```
nagisa, line 138: Similarly to above, I think this constructor-method provide a fair bit of value if it took `logic` and called the `compute_outcome_and_distribute_gas` for you. As-is it only serves to provide multiple ways to construct the same variant.

nagisa: I guess a part of an intent here is to make it possible to modify the ADT definition itself without changing the callers? I guess could see that being valuable if we anticipate further experimentation in this area.

runtime/near-vm-runner-standalone/src/main.rs
```diff
@@ -163,28 +164,29 @@ fn main() {
...
-        println!("\nLogs:");
-        for log in &outcome.logs {
-            println!("{}\n", log);
-        }
-    }
+    let last_result = results.outcomes.pop().unwrap();
+    let maybe_error = match &last_result {
+        VMResult::NotRun(err) | VMResult::Aborted(_, err) => Some(err),
+        VMResult::Ok(_) => None,
+    };
```
nagisa, line 171:
```suggestion
    let maybe_error = last_result.error();
```

nagisa (approved):
I really like this.

I guess one broader thought that I had while reviewing this is that `VMResult::NotRun` variant should not exist in the first place. Any time control flow reaches `near_vm_runner::run`, we are already executing code on the behalf of the user, and so any instance where we end up returning a `NotRun` means that we did some computing but failed to account for it.

I imagine this is largely addressed by the `FunctionCall` action having a flat processing fee, but it isn't immediately apparent to me if it ought to include the parts of pre-processing that gets executed when control flow reaches `near_vm_runner::run` or not. Either way it seems like it wouldn't hurt to always force a return of some cost counter just so that every time we want to exit from `run` early, we are forced to consider if we didn't miss charging any fees so far.

jakmeier:
I forgot to mention why this could help with planned changes: I think we might want to remove most, if not all cases of `NotRun` from the `VM::run` implementations, either by returning `Aborted` or by checking things outside (such as `MethodEmptyName`). But at the level of `execute_function_call`, I think we always need the `NotRun` option as well. However, moving things around means that the order of checks cannot always be preserved. So that would need to be a protocol change.

Having a clearer return type would make that change easier to implement and to discuss its implications. Although, I fully admit that the refactor is not necessary to do the actual changes.

runtime/near-vm-runner/src/runner.rs
jakmeier: Yes, this was the intent here. Or actually, I came from the other side, where the `enum` was a `struct` before.
That said, I think if we all agree on this being an enum, then it makes sense to just remove the constructors.

nagisa: An alternative approach would be to wrap the enum into a newtype so that the variants aren't directly exposed (thus forcing use of the methods to construct and inspect the state here), though that sounds like it might be too heavy handed.

jakmeier:
I cleaned up the constructor methods a bit. Now both `VMResult::ok()` and `VMResult::abort()` take a `VMLogic` and call `compute_outcome_and_distribute_gas()` internally.
But we could also remove them entirely, it only reduces code duplication slightly and I can see an argument for why constructing an enum directly would be considered less complex code. So no strong opinion from my side.

> I guess one broader thought that I had while reviewing this is that `VMResult::NotRun` variant should not exist in the first place. [...]

I would agree on that. I think we could get rid of it as a pure refactor but it would involve a couple of empty outcomes that do not quite make sense to me. I would rather do the protocol update that fixes contract load costs first, since that will also rearrange some things. If we are lucky, `VMResult::NotRun` pops out automatically as we do that change. :)

runtime/near-vm-runner/src/runner.rs
jakmeier: I considered this alternative after you mentioned it. But I agree, it's too clunky and not really necessary IMO.

runtime/near-vm-runner/src/runner.rs
jakmeier: Removed

runtime/near-vm-runner/src/runner.rs
jakmeier: Done

runtime/near-vm-runner-standalone/src/main.rs
```diff
@@ -163,28 +164,26 @@ fn main() {
...
-            println!("{}\n", log);
-        }
-    }
-
-    match &outcome {
-        Some(outcome) => {
+    let last_result = results.outcomes.pop().unwrap();
+    let maybe_error = last_result.error();
+
+    match &last_result {
```
nagisa, line 170: Minor minor nit, but since `last_result` is no longer used after this, you can match by value here and avoid the `clone` of the `VMOutcome` in the `Aborted | Ok` branch.

nagisa (approved):
LGTM

runtime/near-vm-runner-standalone/src/main.rs
jakmeier: Yeah that clone annoys me, too. The problem is that `output` is used again for printing afterwards. So it is not possible to move it when constructing `StandaloneOutput`, without either reordering the prints or changing the `StandaloneOutput::output` type to a borrow. Neither of which seems to be worth it just to avoid the clone here.

matklad:
>Having a clearer return type would make that change easier to implement and to discuss its implications

:+1: to a multi-stage refactor! (triangle inequality of software engineering: `a + b < c`). Let's add a comment to NoRun though, that that is intended to be a transient thing. 

>a couple of empty outcomes that do not quite make sense to me

Hm, empty outcomes is what I'd expect to see, to be honest. Neutral elements are quite a natural thing. That is, in the final state after a protocol change, I expect tho code to look like

```rust
let mut outcome = VMOutcome::default()
let error = MethodNotFound;
if protocol_feature_active {
    outcome.burnt_gas += loading_cost;
}
```

So, I'd expect this to evolve like this:

```rust
// Stage1
pub enum VMResult {
    NotRun(VMError),
    Aborted(VMOutcome, VMError),
    Ok(VMOutcome),
}

// Stage2
pub enum VMResult {
    Aborted(VMOutcome, VMError),
    Ok(VMOutcome),
}

// Stage3
struct VMOutcome {
  ...
  aborted: Option<VMError>
}

// Stage4
struct VMOutcome {
  ...
  aborted: Option<VMError>
}

type VMResult = Result<VMOutcome, DbIsHavingABadDate>
```

(the final stage splits of runtime errors like `CacheError` from `VmError` into an outer error layer).

runtime/near-vm-runner/src/runner.rs
```diff
@@ -108,3 +108,59 @@ impl VMKind {
...
+        let outcome = logic.compute_outcome_and_distribute_gas();
+        VMResult::Ok(outcome)
+    }
+
+    /// Borrow the internal outcome, if there is one.
+    pub fn outcome(&self) -> Option<&VMOutcome> {
+        match self {
+            VMResult::NotRun(_err) => None,
+            VMResult::Aborted(outcome, _err) => Some(outcome),
+            VMResult::Ok(outcome) => Some(outcome),
```
matklad, line 144:
```suggestion
            VMResult::Aborted(outcome, _err) | VMResult::Ok(outcome) => Some(outcome),
```

:+1: for `_err`, naming underscored things helps readability a bunch!

runtime/near-vm-runner/src/wasmer2_runner.rs
```diff
@@ -536,18 +540,18 @@ impl crate::runner::VM for Wasmer2VM {
...
+            return VMResult::NotRun(error);
         }
         let artifact =
             cache::wasmer2_cache::compile_module_cached_wasmer2(code, &self.config, cache);
         let artifact = match into_vm_result(artifact) {
             Ok(it) => it,
-            Err(err) => return (None, Some(err)),
+            Err(err) => {
+                return VMResult::NotRun(err);
+            }
```
matklad, line 554:
```suggestion
            Err(err) => return VMResult::NotRun(err),
```

runtime/runtime-params-estimator/src/gas_metering.rs
```diff
@@ -155,11 +155,11 @@ pub(crate) fn compute_gas_metering_cost(config: &Config, contract: &ContractCode
             PROTOCOL_VERSION,
             cache,
         );
-        if result.1.is_some() {
-            let err = result.1.as_ref().unwrap();
+        if result.error().is_some() {
+            let err = result.error().unwrap();
             eprintln!("error: {}", err);
         }
```
matklad, line 161: let's if-let it while we are here

matklad (approved):
<3

runtime/near-vm-runner/src/runner.rs
jakmeier: I actually tried that first but ran into this error:

```
   --> runtime/near-vm-runner/src/runner.rs:145:48
    |
145 |             VMResult::Aborted(outcome, _err) | VMResult::Ok(outcome)=> Some(outcome),
    |                                        ----    ^^^^^^^^^^^^^^^^^^^^^ pattern doesn't bind `_err`
    |                                        |
    |                                        variable not in all patterns
```

Do you know a way of having both, the named `_err` and the succinct syntax?

jakmeier:
Regarding the empty outcome, you are probably right.

It's a bit awkward since we currently go through the cycle of `VMContext` -> `VMLogic` -> `VMOutcome` to produce an outcome. Some errors happen before we even create the `VMContext` and it does not make sense to me to go through the cycle there. But this is probably solvable with some more refactoring. :)

runtime/runtime-params-estimator/src/gas_metering.rs
jakmeier: Done

runtime/near-vm-runner/src/wasmer2_runner.rs
jakmeier: Done

runtime/near-vm-runner/src/runner.rs
matklad: Ah, that indeed breaks the things, stupid Rust :)

matklad:
Just to clarify, I don't think we should go for empty outcome *in this PR*

Issue #5537, opened 2021-11-30
Update `actix` dependency to stable


lilses:
**Describe the bug**
This is causing extremely frustrating issues when running an indexer because we need to `Cargo.lock` from https://github.com/near/near-indexer-for-explorer to build successfully. Previously didn't need to do this prior to https://github.com/near/nearcore/releases/tag/1.23.0-rc.1

[`actix 11-beta`](https://github.com/near/nearcore/blob/master/chain/indexer/Cargo.toml#L9) is now stable https://github.com/actix/actix/releases/tag/v0.12.0

**To Reproduce**
Build indexer without `Cargo.toml` from https://github.com/near/near-indexer-for-explorer

**Expected behavior**
Deterministically build indexer without relying on anybody's`Cargo.lock`

**Why is this bad**
Getting creative and building any of such just become pointless when convention becomes a web of gotchas documented nowhere.

**Workaround**
This is not a workaround. Copying https://github.com/near/near-indexer-for-explorer is the only way but that also you are implicitly restricted to certain dependencies.

frol:
@miraclx May I ask you to take a look into this and potentially upgrade actix "one last time"? :pray: 

The extreme test are:
* get nearcore compilable after Cargo.lock is removed
* get near-indexer-for-explorer compilable when Cargo.lock is removed

PR #6573, opened 2022-04-11
chore: bump `actix` to `v0.12.0` stable

miraclx:
Closes https://github.com/near/nearcore/issues/5537

Updates `actix` to `v0.12.0`, resolving the compilation issue that required us to manually pin `actix_derive` in our dependency tree. Or downstream dependents of nearcore, having to copy the `Cargo.lock` into their projects.

Related: https://github.com/near/nearcore/pull/6246

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,9 +4,9 @@ version = 3
 
 [[package]]
 name = "actix"
-version = "0.11.0-beta.2"
+version = "0.12.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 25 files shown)


Cargo.lock
```diff
@@ -3639,7 +3637,7 @@ dependencies = [
  "paperclip-actix",
  "paperclip-core",
  "paperclip-macros",
- "parking_lot 0.11.2",
+ "parking_lot 0.10.2",
```
nagisa, line 3640: Really surprising to see the downgrade here.

nagisa (approved):
LGTM, thanks!

Cargo.lock
miraclx: reverted, nice catch.

PR #6580, opened 2022-04-11
refactor: move column definition to a separate file

matklad:
It takes a lot of code to define all our columns
and their properties. This code gets in a way when
reading the rest of the db code.

Moreover, *specific* columns are unrelated to an
*abstract* db we implement in db.rs.

So let's move them into a different file, cleaning
up the API a bit in the process.


matklad:
Note: I find myself utterly confused about semantich of two gc-related sets of columns

matklad:
Ok, I think I've unconfused myself, and, hopefully, future reader as well!

mm-near (approved):

PR #6582, opened 2022-04-11
feat: separate counter for chunk nodes cache

Longarithm:
Recent experiments show that cost for reading kv pair from the chunk cache should be somewhere between 37 and 200 Mgas. Given that we use a safety multiplier 3 and that we could build an example where ~1M nodes are read from the chunk cache, we have to add the new cost to the runtime config.

The first step is adding a separate counter for that, and updating trie storage tests appropriately. This PR shouldn't affect the protocol in any way.

Next steps will be:
* add new cost to runtime config with an estimation function
* return to https://github.com/near/nearcore/pull/6450 and charge this cost there

## Testing

* Updating trie storage tests.

### DIFF
--- a/core/primitives/src/types.rs
+++ b/core/primitives/src/types.rs
@@ -14,6 +14,7 @@ use crate::trie_key::TrieKey;
 use crate::receipt::Receipt;
 /// Reexport primitive types
 pub use near_primitives_core::types::*;
+use std::ops::Sub;
 
 /// Hash used by to store state root.
 pub type StateRoot = CryptoHash;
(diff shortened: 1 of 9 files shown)


core/primitives/src/types.rs
```diff
@@ -990,3 +991,23 @@ pub enum TrieCacheMode {
...
+/// Counts accessed trie nodes during tx/receipt execution for proper storage costs charging.
+#[derive(Debug)]
+pub struct TrieNodesCount {
+    /// Number of nodes read from storage or shard cache.
+    pub touches: u64,
+    /// Number of nodes read from the chunk cache.
+    pub chunk_cache_reads: u64,
+}
+
+impl Sub for TrieNodesCount {
```
matklad, line 1004: Micro nit, but, as a rule of thumb, I've found that `impl ops::Sub` usually works better (needs only one import if you impl, eg, both Sub and SubAssign)

core/primitives/src/types.rs
```diff
@@ -990,3 +991,23 @@ pub enum TrieCacheMode {
...
     CachingChunk,
 }
+
+/// Counts accessed trie nodes during tx/receipt execution for proper storage costs charging.
+#[derive(Debug)]
+pub struct TrieNodesCount {
+    /// Number of nodes read from storage or shard cache.
+    pub touches: u64,
+    /// Number of nodes read from the chunk cache.
+    pub chunk_cache_reads: u64,
```
matklad, line 1001: It seems that the names could be better:

* they seem to describe non-essential properties
* the two names should be symmetric, but they aren't

I'd maybe suggest something like

```
/// Cheap trie node reads which are guaranteed to be served from RAM.
pub mem_reads: u64,
/// Potentially expensive trie node reads which are served from disk in the worst case.
pub db_reads: u64,
```

matklad: Though, even if we do renaming along the lines, it *might* makes sense to do a separate PR for that and take a holistic look at the naming. It seems that the core problem is that we use chunk/block terminology inside the trie, but the trie exists on a level of abstraction *below* chunks and blocks.

core/primitives/src/types.rs
matklad, line 1004: Let's add some comment here explaining that this impl can potentially panic on underflow, and should be used carefully.

matklad (approved):

PR #6584, opened 2022-04-12
Parametrize max open files and col state cache size

EdvardD:
Create `store` section in the neard config so different parameters can be passed to the store;
Add max_open_files and ColState cache size parameters,
Some cleanup around read_only store and enabling statistics.


nearcore/src/config.rs
```diff
@@ -334,6 +334,33 @@ fn default_enable_rocksdb_statistics() -> bool {
...
+    fn default() -> Self {
+        StoreConfig {
+            max_open_files: default_max_open_files(),
+            col_state_cache_size: default_col_state_cache_size(),
+        }
+    }
+}
+
+fn default_max_open_files() -> i32 {
+    10 * 1000
```
mm-near, line 357: let's also mention how we got those numbers (that we run such and such experiment on such and such date).

Will be useful in the future.

nearcore/src/config.rs
```diff
@@ -334,6 +334,33 @@ fn default_enable_rocksdb_statistics() -> bool {
...
+}
+
+fn default_col_state_cache_size() -> usize {
+    512 * 1024 * 1024
```
mm-near, line 361: mention where did we get this number (I guess here, we can say, that it was set like this for a long time ;-) and we might consider tuning this number in the future.

nearcore/src/config.rs
```diff
@@ -334,6 +334,33 @@ fn default_enable_rocksdb_statistics() -> bool {
     false
 }
 
+#[derive(Serialize, Deserialize, Clone, Debug)]
+pub struct StoreConfig {
+    /// Maximum number of store files being opened simultaneously.
```
mm-near, line 339: let's also mention pros & cons of increasing/lowering.

(I guess for increasing we should also tell people to change the ulimit etc)

nearcore/src/config.rs
```diff
@@ -334,6 +334,33 @@ fn default_enable_rocksdb_statistics() -> bool {
...
+    #[serde(default = "default_max_open_files")]
+    pub max_open_files: i32,
+    /// Cache size for ColState column.
```
mm-near, line 342: same here - what are pros & cons of increasing/lowering this value.

core/store/src/db.rs
```diff
@@ -531,8 +535,9 @@ impl RocksDBOptions {
     /// Opens a read only database.
     pub fn read_only<P: AsRef<std::path::Path>>(self, path: P) -> Result<RocksDB, DBError> {
         use strum::IntoEnumIterator;
-        let options = self.rocksdb_options.unwrap_or_else(rocksdb_options);
-        let cf_with_opts = DBCol::iter().map(|col| (col_name(col), rocksdb_column_options(col)));
+        let options = self.rocksdb_options.unwrap_or_else(|| rocksdb_options(self.max_open_files));
```
mm-near, line 538: it is strange that we treat 'max_open_files' in a special way.

core/store/src/db.rs
```diff
@@ -590,6 +601,16 @@ impl RocksDBOptions {
         self.enable_statistics = true;
         self
     }
+
+    pub fn max_open_files(mut self, max_open_files: i32) -> Self {
+        self.max_open_files = max_open_files;
```
mm-near, line 606: maybe this should add/set the rocksdb options?

nearcore/src/lib.rs
```diff
@@ -570,7 +572,12 @@ pub fn recompress_storage(home_dir: &Path, opts: RecompressOpts) -> anyhow::Resu
     info!(target: "recompress", src = %src_dir.display(), dest = %opts.dest_dir.display(), "Recompressing database");
     let src_store = create_store_with_config(
         &src_dir,
-        StoreConfig { read_only: true, enable_statistics: false },
+        StoreConfig {
+            read_only: true,
```
mm-near, line 576: we should have a constructor instead (where you just pass config.store as parameter)

core/store/src/lib.rs
```diff
@@ -289,21 +289,38 @@ pub fn read_with_cache<'a, T: BorshDeserialize + 'a>(
...
+    /// Cache size for ColState column.
+    pub col_state_cache_size: usize,
+}
+
+impl Default for StoreConfig {
+    fn default() -> StoreConfig {
+        StoreConfig {
+            read_only: false,
+            enable_statistics: false,
+            max_open_files: 10 * 1000,
```
mm-near, line 310: avoid specifying this value here again -and use the one from config::default instead

core/store/src/db.rs
```diff
@@ -869,18 +890,18 @@ fn rocksdb_block_based_options(cache_size: usize) -> BlockBasedOptions {
 }
 
 // TODO(#5213) Use ByteSize package to represent sizes.
-fn choose_cache_size(col: DBCol) -> usize {
+fn choose_cache_size(col: DBCol, col_state_cache_size: usize) -> usize {
```
mm-near, line 893: consider either adding a config as a parameter, or moving this method into config

mm-near (changes requested):

EdvardD:
> The default on Debian and Ubuntu is 1024 RLIMIT_NOFILE so this will cause failures in production. If we’re increasing the count we should also set the limit at startup. I’m doing that in recompress_storage so we should use something similar at command startup.

I was thinking of only the running daemon but you're right the client can run it one-off. I returned 512 in this PR and will increase the number in the followup.

core/store/src/db.rs
```diff
@@ -486,6 +488,8 @@ impl Default for RocksDBOptions {
             free_space_threshold: bytesize::ByteSize::mb(16),
             warn_treshold: bytesize::ByteSize::mb(256),
             enable_statistics: false,
+            max_open_files: 10 * 1000,
```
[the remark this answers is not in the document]
EdvardD: PTAL how it looks now: I made neard to pick up defaults from core::store since the opposite is not possible

core/store/src/db.rs
EdvardD: Passing the whole store config now.

core/store/src/db.rs
```diff
@@ -934,12 +942,12 @@ impl RocksDB {
...
-    pub fn new_read_only<P: AsRef<std::path::Path>>(path: P) -> Result<Self, DBError> {
-        RocksDBOptions::default().read_only(path)
+    pub fn new_read_only<P: AsRef<std::path::Path>>(path: P, store_config: &StoreConfig) -> Result<Self, DBError> {
+        RocksDBOptions::default().read_only(path, &store_config)
     }
 
-    pub fn new<P: AsRef<std::path::Path>>(path: P) -> Result<Self, DBError> {
-        RocksDBOptions::default().read_write(path)
+    pub fn new<P: AsRef<std::path::Path>>(path: P, store_config: &StoreConfig) -> Result<Self, DBError> {
+        RocksDBOptions::default().read_write(path, &store_config)
```
mina86, line 950: This feels wrong. `StoreConfig` has a `read_only` flag so it feels those two methods should be merged into one which checks the config. `get_version` uses `new_read_only` so there may be reason to keep that one but it probably should be private.

nearcore/src/lib.rs
```diff
@@ -114,9 +120,19 @@ fn create_db_checkpoint(path: &Path, near_config: &NearConfig) -> Result<PathBuf
     Ok(checkpoint_path)
 }
 
+fn get_store_version_with_config(path: &Path, near_config: &NearConfig) -> u32 {
```
mina86, line 123: The function will open and then immediately close the database. I honestly don’t think we need to pass configuration from NearConfig here.

EdvardD: My thought was that in the future there could be other store parameters REQUIRED for db to correctly open even for just reading the version so it's better to open the store with the same parameters everywhere.

core/store/src/lib.rs
```diff
@@ -289,27 +289,49 @@ pub fn read_with_cache<'a, T: BorshDeserialize + 'a>(
...
+impl StoreConfig {
+    pub fn read_only() -> StoreConfig {
+        let mut config = StoreConfig::default();
+        config.read_only = true;
+        config
+    }
+}
+
+impl Default for StoreConfig {
+    fn default() -> StoreConfig {
```
mina86, line 314: Perhaps it would be better to remove `Default` implementation and rename this method `read_write`?

nearcore/src/config.rs
```diff
@@ -334,6 +334,43 @@ fn default_enable_rocksdb_statistics() -> bool {
...
+    /// Linux.
+    #[serde(default = "default_max_open_files")]
+    pub max_open_files: i32,
+    /// Cache size for ColState column.
+    /// Default value: 512MiB.
+    /// Increasing ColState cache size helps making storage more efficient. On the other hand we
+    /// don't want to increase hugely requirements for running a node so currently we use a small
+    /// default value for it.
+    #[serde(default = "default_col_state_cache_size")]
+    pub col_state_cache_size: usize,
```
mina86, line 354: By the way, since we’re making this configurable, should we make cache size for other columns configurable as well?

nearcore/src/config.rs
```diff
@@ -487,6 +525,7 @@ impl Default for Config {
             db_migration_snapshot_path: None,
             use_db_migration_snapshot: true,
             enable_rocksdb_statistics: false,
```
mina86, line 527: We should move `enable_rocksdb_statistics` to `store` as well IMO. This option is not present in 1.25.0 so we can still move it around.

nearcore/src/config.rs
```diff
@@ -334,6 +334,43 @@ fn default_enable_rocksdb_statistics() -> bool {
     false
 }
 
+#[derive(Serialize, Deserialize, Clone, Debug)]
+pub struct StoreConfig {
```
mina86, line 338: Alternatively make `near_store::StoreConfig` serialisable and refer to it directly. You can mark `read_only` field in the struct as `serde(skip)` and thus it won’t be present in the config.json. This reduces amount of code.

nearcore/src/config.rs
```diff
@@ -788,6 +827,17 @@ impl NearConfig {
...
+pub fn update_with(
+    store_config: near_store::StoreConfig,
+    config: &Config,
+) -> near_store::StoreConfig {
+    let mut store_config = store_config.clone();
+    store_config.enable_statistics = config.enable_rocksdb_statistics;
+    store_config.max_open_files = config.store.max_open_files;
+    store_config.col_state_cache_size = config.store.col_state_cache_size;
+    store_config
+}
```
mina86, line 839: This would work better as `Into<near_store::StoreConfig>` implementation IMO. Then add `pub fn with_read_only(self, read_only: bool) -> Self` method to `near_store::StoreConfig` which changes the value of the field and you can do `near_store::StoreConfig::from(near_config.store).read_only(true)`. This would also work in `get_version` where you’d just do `store_config.clone().read_only(true)`.

nearcore/src/lib.rs
mina86: I think we can worry about required options when we actually have any. For now I wouldn’t complicate it.

core/store/src/db.rs
```diff
@@ -923,8 +943,14 @@ impl RocksDB {
...
     /// Returns version of the database state on disk.
-    pub fn get_version<P: AsRef<std::path::Path>>(path: P) -> Result<DbVersion, DBError> {
-        let db = RocksDB::new_read_only(path)?;
+    pub fn get_version<P: AsRef<std::path::Path>>(
+        path: P,
+        store_config: &StoreConfig,
+    ) -> Result<DbVersion, DBError> {
+        if !store_config.read_only {
+            warn!("consider opening the database in read only mode to get the version");
+        }
```
mina86, line 952: I think it’s better to just make a copy of the config and update `read_only = true` here rather than making it caller’s burden.

core/store/src/db.rs
```diff
@@ -177,12 +177,27 @@ impl RocksDBOptions {
         self
     }
 
+    /// Opens the database either in read only or in read/write mode depending on the read_only
+    /// parameter specified in the store_config.
+    pub fn open(self, path: &Path, store_config: &StoreConfig) -> Result<RocksDB, DBError> {
```
mina86, line 182: I meant this can stay `path: impl AsRef<Path>` but then do `let path = path.as_ref();` and the other two methods can just accept `&Path`.

core/store/src/db.rs
```diff
@@ -177,12 +177,27 @@ impl RocksDBOptions {
...
         use strum::IntoEnumIterator;
-        let options = self.rocksdb_options.unwrap_or_else(rocksdb_options);
-        let cf_with_opts = DBCol::iter().map(|col| (col_name(col), rocksdb_column_options(col)));
-        let db = DB::open_cf_with_opts_for_read_only(&options, path, cf_with_opts, false)?;
+        let options = self.rocksdb_options.unwrap_or_else(|| rocksdb_options(store_config));
+        let cf_with_opts =
+            DBCol::iter().map(|col| (col_name(col), rocksdb_column_options(col, store_config)));
+        let db = DB::open_cf_with_opts_for_read_only(
+            &options,
+            AsRef::<Path>::as_ref(path),
```
mina86, line 197: Just `path`.

core/store/src/db.rs
```diff
@@ -199,20 +214,25 @@ impl RocksDBOptions {
...
+                .map(|col| {
+                    ColumnFamilyDescriptor::new(
+                        col_name(col),
+                        rocksdb_column_options(col, store_config),
+                    )
+                })
                 .collect()
         });
-        let db = DB::open_cf_descriptors(&options, path, cf_descriptors)?;
+        let db = DB::open_cf_descriptors(&options, AsRef::<Path>::as_ref(path), cf_descriptors)?;
```
mina86, line 235: Ditto, no need for `AsRef` shenanigans.

core/store/src/lib.rs
```diff
@@ -288,27 +289,84 @@ pub fn read_with_cache<'a, T: BorshDeserialize + 'a>(
...
-#[derive(Default, Debug)]
+#[derive(Serialize, Deserialize, Clone, Debug)]
 pub struct StoreConfig {
     /// Attempted writes to the DB will fail. Doesn't require a `LOCK` file.
+    #[serde(skip)]
     pub read_only: bool,
+
     /// Re-export storage layer statistics as prometheus metrics.
     /// Minor performance impact is expected.
+    #[serde(default = "default_enable_statistics")]
```
mina86, line 300: I think `#[serde(default)]` will just work since `bool` implements `Default` with `false` value.

core/store/src/lib.rs
```diff
@@ -288,27 +289,84 @@ pub fn read_with_cache<'a, T: BorshDeserialize + 'a>(
...
+    StoreConfig::COL_STATE_CACHE_SIZE
 }
 
-pub fn create_store_with_config(path: &Path, store_config: StoreConfig) -> Store {
-    let mut opts = RocksDBOptions::default();
-    if store_config.enable_statistics {
-        opts = opts.enable_statistics();
+impl StoreConfig {
+    // This is a value that we've used since 3 Dec 2019.
+    pub const MAX_OPEN_FILES: i32 = 512;
```
mina86, line 336: DEFAULT_MAX_OPEN_FILES

Also, why not turn the `//` comment into `///`.

core/store/src/lib.rs
```diff
@@ -288,27 +289,84 @@ pub fn read_with_cache<'a, T: BorshDeserialize + 'a>(
...
+
+    // We used to have the same cache size for all columns 32MB. When some RocksDB
+    // inefficiencies were found ColState cache size was increased up to 512MB.
+    // This was done Nov 13 2021 and we consider increasing the value.
+    pub const COL_STATE_CACHE_SIZE: usize = 512 * bytesize::MIB as usize;
```
mina86, line 341: DEFAULT_COL_STATE_CACHE_SIZE

core/store/src/lib.rs
```diff
@@ -288,27 +289,84 @@ pub fn read_with_cache<'a, T: BorshDeserialize + 'a>(
...
+    pub fn read_write() -> StoreConfig {
+        StoreConfig {
+            read_only: false,
+            enable_statistics: default_enable_statistics(),
+            max_open_files: default_max_open_files(),
+            col_state_cache_size: default_col_state_cache_size(),
+        }
+    }
+
+    pub fn with_read_only(&self, read_only: bool) -> Self {
```
mina86, line 356: I’d take it by value and avoid clone.

mina86 (approved):

nearcore/src/config.rs
```diff
@@ -456,8 +452,7 @@ pub struct Config {
     /// For example, setting "use_db_migration_snapshot" to "/tmp/" will create a directory "/tmp/db_migration_snapshot" and populate it with the database files.
     #[serde(skip_serializing_if = "Option::is_none")]
     pub db_migration_snapshot_path: Option<PathBuf>,
-    #[serde(default = "default_enable_rocksdb_statistics")]
-    pub enable_rocksdb_statistics: bool,
+    pub store: near_store::StoreConfig,
```
mm-near, line 455: it seems that we removed something (enable_rocksdb_statistics) -- did we set it in any config in any of our jobs?

mm-near: Also - add a comment about this field.

mm-near (approved):

PR #6585, opened 2022-04-12
Increase the secondary stack cleck limit

nagisa:
200*1024 is still just under 2MiB, which we can definitely afford, given a default stack size on glibc systems is 8MiB.


core/primitives/src/runtime/config_store.rs
```diff
@@ -22,6 +22,7 @@ static CONFIGS: &[(ProtocolVersion, &[u8])] = &[
     (50, include_config!("50.json")),
     // max_gas_burnt increased to 300 TGas
     (52, include_config!("52.json")),
+    // Increased deployment costs, increased wasmer2 stack_limit
```
matklad, line 25: :+1:

matklad (approved):

matklad:
Hm, CI's failing?

